PR #1184, opened 2020-01-28
Fix layout for structured buffers of matrices

tfoleyNV:
When using row-major layout (via command-line or API option), the following sort of declaration:

```hlsl
StructuredBuffer<float4x4> gBuffer;

... gBuffer[i] ...
```

Generates unexpected results when compiled to DXBC via fxc or DXIL via dxc, because the fxc/dxc compilers do not respect the matrix layout mode in this specific case (a structured buffer of matrices). Instead, they always use column-major layout, even if row-major was requested by the user.

A user can work around this behavior by wrapping the matrix in a `struct`:

```hlsl
struct Wrapper { float4x4 wrapped; }
SturcturedBuffer<Wrapper> gBuffer;

... gBuffer[i].wrapped ...
```

This change simply automates that workaround when compiling for an HLSL-based downstream compiler, so that we get the same behavior across all our backends.

The change adds a test case to confirm the behavior across multiple targets, but it turns out we also had a test checked in that confirmed the buggy (or at least surprising) fxc/dxc behavior, so that one had its baselines changed and can work as a regression test for this fix as well.


tfoleyNV:
Yes, the reliance on heuristics rather than hard data is worrying. I actually don't expect the heuristic for detecting the intrinsic functions to be much of a problem (famous last words...), but I'm more worried about the long-term risk of having more code that assumes type deduplication is working 100% of the time when we know it isn't.

Solving the deduplication issue in the IR is starting to float from the "technical debt" pile to a "maybe this is actually a serious lurking bug" pile in my brain.

PR #1186, opened 2020-01-28
Feature/test for double behavior

jsmall-nvidia:
* Added GPULikeBindRoot so can use BindLocation on GPU
* Added support for writing bindings using their underlying types
  * Use the -output-using-type option on your test
* Added the dxbc-double-problem test that allows easy demonstration of problem with dx12 dxbc double behavior being broken
  * Is disabled by default


tools/render-test/shader-input-layout.cpp
```diff
@@ -698,42 +698,182 @@ namespace renderer_test
...
 
-                const int size = int(entry.bufferData.getCount());
-                // Must be the same size or less than allocated buffer
-                SLANG_ASSERT(size * sizeof(unsigned int) <= buffer->m_sizeInBytes);
 
-                for (int i = 0; i < size; ++i)
+        switch (scalarType)
+        {
+            // TODO(JS):
+            // Bool is here, because it's not clear across APIs how bool is laid out in memory
```
tfoleyNV: One random thought is that we might want to output these values canonically as hex, and then include a "humane" version of the value as a comment, e.g.:

```
0x3f800000 // 1.0 (float32)
```

Ideally we would then spruce up the comparison logic so that for these tests we use a comparsion that ignores the `//` and the text that follows it, so that the canonical test output is just the hex values, to account for potential differences in printing across OSes (I guess that still doesn't deal with differences across platforms in the actual bits of the result...)

PR #1189, opened 2020-01-29
Check library modules for explicit entry point names as well

chuckgenome:
Also scan any entry points scraped from library modules when explicitly requesting


source/slang/slang-check-shader.cpp
```diff
@@ -913,19 +937,19 @@ namespace Slang
...
-            auto entryPoint = EntryPoint::createDummyForDeserialize(
-                linkage,
-                extraEntryPoint.name,
-                extraEntryPoint.profile,
-                extraEntryPoint.mangledName);
-            allComponentTypes.add(entryPoint);
+            // Also consider entry points that were introduced via adding
+            // a library reference...
+            //
+            for (auto extraEntryPoint : compileRequest->m_extraEntryPoints)
```
tfoleyNV: Moving this loop under the `if` changes the semantics in the case where *no* explicit entry points were added, because it makes it so that these "extra" entry points only get added if there were explicit entry point already (which should almost never be what you want).

Seeing this change here makes me wonder whether the whole change here isn't unnecessary.

If the original code that is invoking the Slang API here just didn't specify *any* explicit entry points, then it would auto-magically get these "extra" entry points from the linked libraries added in, and it could request kernel code from them just fine. This change seems to make it so that you can *only* use the "extra" entry points by specifying at least one explicit entry point, and that means you'd end up with duplication in `allComponentTypes` since the same entry point would be added both as an explicit entry point and as an "extra" one.

jsmall-nvidia:
I believe the behavior is that if there are no entry points specified, it will create entry point information in the front end for any located in the specified modules. 

Therefore in the use case where you want to use an entry point that's in a library, not specifying  any entry points, and then getting the entry point that will be auto added is the current mechanism.

It could be argued that mechanism is a bit surprising - which is touched on in the comments. That perhaps there should be a flag that adds library entry points (assuming no clashes with explicit entry points), so we don't rely on the implicit behavior and allows the mixing.

chuckgenome:
Incorrect approach. Solved this in host software

PR #1191, opened 2020-01-30
Support for 64 bit integer types

jsmall-nvidia:
Main fixes here was to make 64 bits work on GLSL, and to makes sure integer literals are output more appropriately. 

* For integer literals add postfix, and use unsigned/signed output appropriately
* Extend GLSL extension handling by type, and for adding 64 bit int extensions
* Added tests for int/uint64 types

This is preliminary support. More complete support would allow vector types to be used in slang source etc..

uint64_t, uint64_t2, uint64_t3, uint64_t4

https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12


source/slang/slang-emit-glsl-extension-tracker.cpp
```diff
@@ -26,18 +26,35 @@ void GLSLExtensionTracker::requireVersion(ProfileVersion version)
     }
 }
 
-void GLSLExtensionTracker::requireHalfExtension()
+void GLSLExtensionTracker::requireBaseTypeExtension(BaseType baseType)
 {
-    if (!m_hasHalfExtension)
+    uint32_t bit = 1 << int(baseType);
+    if (m_hasBaseTypeFlags & bit)
```
tfoleyNV: I suppose the micro-optimization here of not even trying to add the strings pays off for stuff like basic types, since we'd call into this logic a lot. Otherwise I'd point out that we already avoid re-adding an extension that was already added as part of `requireExtension`.

PR #1192, opened 2020-01-30
Fix for a macro expansion bug

tfoleyNV:
The work in #1177 added a notion of a macro being "busy" to prevent errors around recursively-defined macros:

    #define BAR(X) BAR(0)

    BAR(A) /* should yield BAR(0) and not infinite-loop */

The fix was to place an `isBusy` flag on each macro, setting it to `true` when an expansion of that macro gets pushed onto the stack of input streams, and back to `false` when the expansion gets popped from the stack of input streams.

That approach meant that we had to pre-expand macro arguments to avoid incorrectly treating a macro as busy for nested-but-not-recursive invocations:

    #define NEG(X) (-X)

    NEG(NEG(3)) // should expand to (-(-3)) and not (-NEG(3))

The subtle bug that arose with the `isBusy` flag is that the current preprocessor design doesn't always pop input streams when we might expect. For input like the following:

    BAR(A) BAR(B)

The preprocessor can be in a state where the stack of input streams look like:

    // top input stream:
    BAR(X) => BAR(0)
    ----------------^

    // bottom input stream
    BAR(A) BAR(B)
    ------^

That is, the first macro expansion is still "open," even though it is at the end of its input. When the preprocesor looks ahead and sees `BAR` as the next token and asks whether it should be expanded, the `isBusy` state on the BAR macro hasn't yet been unset, so expansion gets skipped.

Aside: it is completely reasonable to ask at this point whether we should just "fix" the behavior of not popping input streams that are at their end. We should consider doing that, but the current code goes to some lengths to preserve the current behavior and I do *not* recall why we had found it necessary. Any attempt to change that behavior should come along with lots of testing.

This change instead adds a different approach for tracking the busy-ness of macros that doesn't rely on mutable state in the macro, and thus works even without pre-expansion of macro arguments.

In the Slang preprocessor, macro names are always looked up in *environments*, where each environment maps names to macros, and has an option parent environment.
There is a global environment used for most cases, but when expanding a function-like macro we would create an "argument environment" for it that maps the parameter names to their argument tokens.

By expanding the environment type to include an (optional) field for a macro that is made "busy" by that environment, we can check if a macro is busy in a given environment by checking if the given macro has been associated with any of the environments on the chain of parent environments.

A function-like macro expansion could then attach the macro to the "argument environment" and automatically make itself busy for the purposes of expansion of its body.
To extend the same functionality to *all* macros, the "argument environment" was changes to an "expansion environment" that always gets used for a macro expansion and always has the identity of the macro attached.

Because the arguments to a function-like macro have their own environment for expansion that bypasses the argument/expansion environment of the function-like macro, the macro will show as busy in its own body, but not in the expansion of its arguments. Thus the pre-expansion of macro arguments is not needed with this change.

Aside: it might not be clear how this design avoids the problem of the unpopped input streams that aren't at their end. The answer is that the "current" environment for the preprocessor is already taken as the environment of the top-most input stream that is *not at its end*. This means that the new test for busy-ness benefits from the existing logic that was in place to deal with non popping input streams as soon as then end. (This is not to say that the current input-stream behavior isn't questionable)

This change includes a new test case for the behavior that was broken, and expands the test case from #1177 to also test object-like macros.


tfoleyNV:
This should handle the mutually-recursive macro case, so I think it is ready for review @jsmall-nvidia.

The code still doesn't correctly handle subtle cases that require macro arguments to be expanded before the macro body is expanded, such as:

```
#define A(x,y) x + y
#define B(z) A(z)
#define C 1,2

B(C) // should expand to `1 + 2`
```

That behavior has *never* been supported by our preprocessor, though, and would seemingly require expanding the macros for parameters (e.g., `z`) even in cases that currently suppress expansion (like when reading the arguments to a macro). I think we could make it work, but it is quite a corner case and I suspect variadic macros to come up before that.

PR #1195, opened 2020-01-31
Some Slang API additions

tfoleyNV:
These are additions to the public Slang API that came up while I was trying to write an example to demonstrate GPU printing. The additions aren't strictly necessary for the example, but I found these to be missing services when writing the code the way I wanted to.

The main public changes are:

* There is a new distinct `IEntryPoint` interface which inherit from `IComponentType` (much like `IModule` before). For now this doesn't expose any new functions on top of `IComponentType`, but I expect it to do so eventually.

* It is now possible to get the `IModule` for a specific translation unit in a compile request with `spCompileRequest_getModule`. Even for a compile request that had only one translation unit this is *not* the same object as gets returned by `spCompileRequest_getProgram`. The latter should probably be called `*_getLinkedProgram` because it returns a composite component type that links the module with everything it `import`s.

* An `IModule` can look up entry points declared in the module. Eventually a module should support looking up most of the declarations in the module (e.g., types) by name, but entry points are an obvious case.

* A new `link()` operation is added to `IComponentType`. It is possible to have component types that have unsatisfied dependencies, such that trying to generate kernel code from them will fail. The `link()` operation tries to produce a new composite component type that combines a component with its dependencies, to enable code generation. The implementation of end-to-end compilation was using a function like this internally, but it hadn't been exposed to the API.

Notes on the implementation:

* The list of entry points declared in a given translation unit has moved from `TranslationUnitRequest` to the `Module` inside of it.

* `EntryPoint` now has to do a song and dance much like `Module` to both inherit from `ComponentType` and support the `IEntryPoint` interface.

* The `Session* m_session` member in `Linkage` (in terms of public API, this is the `slang::ISession` holding a pointer to the `slang::IGlobalSession`) has been changed to a `RefPtr`. Without this change an application can't just hold onto a `ComPtr<slang::ISession>`; they also need to retain the `IGlobalSession` or things will crash. The new behavior seems more correct, but I worry that it might introduce a leak.

* The `asInternal` operation for `IComponentType` had to be updated to not just perform a cast. A type like `Module` has two `IComponentType` sub-objects, and only one of these is at the same address as the `ComponentType` base.

* Similarly, the `Module::getInterface` logic was changed to fall back to `Super::getInterface` for all the cases other than `IID_IModule`, so that it would be guaranteed to return the `IComponentType` at the same address as `ComponentType` in response to a `queryInterface`.


tfoleyNV:
Looks like I'm causing the rest running to crash on exit in debug builds, so I'd better resolve that before review.

tfoleyNV:
Yes, it appears that with my change to have `Linkage` store a strong pointer to `Session`, there is now some kind of leak going on in our test framework such that there is still an active reference to the session at the time we call `spDestroySession`.

The simplest fix would be to make `spDestroySession` not assert that the reference it is releasing is the last reference to the global session, but it would be good to ensure that the session is actually getting cleaned up sooner or later...

slang.h
```diff
@@ -3240,16 +3269,50 @@ namespace slang
     }
 }
 
-/**
+/** Get the (linked) program for a compile request.
+
+The linked program will include all of the global-scope modules for the
```
jsmall-nvidia: Would be nice to have SlangCompileRequest actually be an interface, that these functions could just be inline calls into that, if you wanted a 'c-like interface' still. Doing so would clean up some example code and probly other places cos can just use ComPtr.

source/slang/slang-compiler.cpp
```diff
@@ -196,6 +196,16 @@ namespace Slang
     // EntryPoint
     //
 
+    static const Guid IID_IEntryPoint = SLANG_UUID_IEntryPoint;
```
jsmall-nvidia: This might be a little rad - but if our API is a C++ interface (that may be C compatible in the future with some code gen/COM ABIness), perhaps we should remove the macro stuff around guids, and say have a static member function in the interface that can return the interface. Doing so would mean we can always do (say) IEntryPoint::getInterfaceGuid(); . Doing as a function means that all this still works as a header (the interface can just return a static local variable with a ref to that).

It also means we don't have to define static local variables in the impl files that use the interfaces. 

The style means template mechanisms can be empowered to do more useful work too.

source/slang/slang-compiler.h
```diff
@@ -937,6 +1018,26 @@ namespace Slang
 
         // List of filesystem paths this module depends on
         FilePathDependencyList m_filePathDependencyList;
+
+        // Entry points that were defined in thsi module
```
jsmall-nvidia: thsi

source/slang/slang.cpp
```diff
@@ -1725,14 +1727,51 @@ void Module::setModuleDecl(ModuleDecl* moduleDecl)
...
+    // Note: we use a `queryInterface` here instead of just a `static_cast`
+    // to ensure that the `IComponentType` we get is the preferred/canonical
+    // one, which shares its address with the `ComponentType`.
+    //
+    // TODO: An alternative choice here would be to have a "magic" IID that
+    // we pass into `queryInterface` that returns the `ComponentType` directly
+    // (without even `addRef`-ing it).
+    //
+    ComPtr<slang::IComponentType> componentType;
+    inComponentType->queryInterface(IID_IComponentType, (void**) componentType.writeRef());
```
jsmall-nvidia: Hmm. This assumes the inComponentType has an external ref count. It should probably check that. I mean it probably will... but there is potential for kaboom.

Better if we had the 'cast magic' where a UUID can return say RefObject* (and it does so without the refcount) you can do this without worrying about refcounts either. It would also allow the use of dynamic_cast (or some other mechanism) to sure the cast is to the impl.

source/slang/slang-compiler.cpp
tfoleyNV: Yeah, it could be good, so long as the functions are header-only (so they don't put C++ symbols into our DLL interface)

slang.h
tfoleyNV: Revamping the compile request part of the API is definitely a next step. The biggest sticking point (for me) is whether it should stick with the current stateful model of the `sp...` functions, or whether it should use some kind of `Desc` structure. I know the latter is what most clients of the API would prefer, but we have also found that we keep having to add more bells and whistles to compilation.

source/slang/slang.cpp
tfoleyNV: In this case it *must* have an external ref count because otherwise you are passing in an already-released pointer, no?

I agree that the magic cast would probably be better (hence the `TODO` comment). Would be a nice cleanup, but I'm ultimately just trying to get things working well enough to get back to my example program.

PR #1196, opened 2020-01-31
64 bit doc  and testing improvements

jsmall-nvidia:
* Improved and simplified 64 bit type documentation 
* Added tests to try and determine what intrinsics are really available for double for targets


docs/64bit-type-support.md
```diff
@@ -11,6 +11,42 @@ This also applies to vector and matrix versions of these types.
 
 Unfortunately if a specific target supports the type or the typical HLSL instrinsic functions (such as sin/cos/max/min etc) depends very much on the target. 
 
+Special attention has to be made with respect to literal 64 bit types. By default float and integer literals if they do not have an explicit postfix are assumed to be 32 bit. There is a variety of reasons for this design choice - the main one being around by default getting good performance. The postfixes required for 64 bit types are as follows
+
+```
+// double - 'l' or 'L'
+
+double a = 1.34e-200L;
+// WRONG!: This is the same as b = double(float(1.34e-200)) which will be 0
```
tfoleyNV: Minor nit, but it might be hard for a reader to know that this comment refers to the line that follows it rather than the one before. That said, I figured it out easily enough...

PR #1198, opened 2020-02-03
CUDA/C++ backend improvements 

jsmall-nvidia:
* Handling construction (via init) of a vector from a vector and scalars
* Improved 64bit documentation
* Added vector-float.slang test and made sure worked on all targets (with some caveats)
* Handle sincos for vector/matrix types 
* Allow comparisons in test to be tested in type sensitive way


source/slang/slang-emit-cpp.cpp
```diff
@@ -394,6 +394,23 @@ SlangResult CPPSourceEmitter::calcTypeName(IRType* type, CodeGenTarget target, S
 {
     switch (type->op)
     {
+        case kIROp_PtrType:
+        {
+            auto ptrType = static_cast<IRPtrType*>(type);
+            SLANG_RETURN_ON_FAIL(calcTypeName(ptrType->getValueType(), target, out));
+            // TODO(JS): It seems although it says it is a pointer, it can actually be output as a reference
+            // not clear where the ptr aspect is there, as in the definition it is just 'out', implying out
+            // is somewhere converted to a ptr?
```
tfoleyNV: At the IR level, there is no semantic difference between pointers and references, and indeed the `RefType` case and any other `IRPtrTypeBase` (there are some other cases that seemingly wouldn't be caught here) should probably all be emitted identically (and that should be as a pointer in the long run).

This is similar to how if you had to translate a function like this:

```c++
void doThings(int& ioVlaue);
```

over to plain C, you'd write it as:

```c
void doThings(int* ioValue);
```

The only reasons to emit things as C++ reference types in our output today are:

1. Because of the current over-simplifications in the emit logic that have been made for HLSL/GLSL which mean we don't properly handle pointer types in the middle of expressions (e.g., we emit `obj.field` instead of `obj->field` or `(*obj).field`).

2. Because we might want to emit C++ code (especially for headers) that follows a more conventional style in terms of making `out` and `inout` parameters use references. (This might matter if there are C++ ABIs where pointers and references aren't actually guaranteed to be binary-compatible in function signatures)

source/slang/slang-hlsl-intrinsic-set.h
```diff
@@ -146,26 +146,28 @@ struct HLSLIntrinsic
...
     bool operator!=(const ThisType& rhs) const { return !(*this == rhs); }
 
+    static bool isTypeScalar(IRType* type)
+    {
+        // Strip off ptr if it's an operand type
+        if (type->op == kIROp_PtrType)
+        {
+            type = as<IRType>(type->getOperand(0));
+        }
+        // If any are vec or matrix, then we
```
tfoleyNV: Comment trails off without reaching the...

PR #1202, opened 2020-02-05
Literal handling improvements

jsmall-nvidia:
* Display a warning if an integer literal loses bits because it cannot be represented 
  * Does allow uses such as -1u 
* Display a warning if a float literal is outside of range, turns into infinity
* Display if float literal is so small it becomes 0
* Enable source emitters to be able to handle infinity
* Fixed some diagnostic numbers so weren't all the same number
* Added diagnostic tests for float and int
* Added compute tests for float/int literals and for inf
* Made behavior of integer literals consistent across targets
* Fixed tests broken by changes

Note that the diagnostics do not include - for something like -10, and this is because Slang will lex as `-` and `10`. So when diagnostic is produced it doesn't have the -, which could be a little surprising. 

Work was done to make the the lexer lex including the -, but this broke in a test with `count -1`, which became unparsable, because -1 became a literal. So that was removed.


source/slang/slang-compiler.h
```diff
@@ -1922,6 +1922,34 @@ namespace Slang
...
+            };
+        };
+
+        SLANG_FORCE_INLINE static const BaseTypeInfo& getInfo(BaseType baseType) { return s_info[Index(baseType)]; }
+
+        static UnownedStringSlice asText(BaseType baseType);
+
+        uint8_t sizeInBytes;               ///< Size of type in bytes
+        Flags flags;
+        uint8_t baseType;
```
tfoleyNV: We could in theory extend this to store the number of bits for various parts of the value, e.g.:

```
int signBits; // zero or one for unsigned/signed
int exponentBits; // zero for integers
int valueBits; // for floating point these are the mantissa bits
```

It would be possible in principle to derive the rest of what we need (minimum/maximum values, etc.) from those. Of course, that would rely on us having an arbitrary-precision numeric type to store the literal values...

source/slang/slang-diagnostic-defs.h
```diff
@@ -317,44 +317,46 @@ DIAGNOSTIC(30600, Error, varWithoutTypeMustHaveInitializer, "a variable declarat
 // 307xx: parameters
 DIAGNOSTIC(30700, Error, outputParameterCannotHaveDefaultValue, "an 'out' or 'inout' parameter cannot have a default-value expression");
 
-DIAGNOSTIC(39999, Error, expectedIntegerConstantWrongType, "expected integer constant (found: '$0')")
-DIAGNOSTIC(39999, Error, expectedIntegerConstantNotConstant, "expression does not evaluate to a compile-time constant")
-DIAGNOSTIC(39999, Error, expectedIntegerConstantNotLiteral, "could not extract value from integer constant")
+DIAGNOSTIC(30701, Error, expectedIntegerConstantWrongType, "expected integer constant (found: '$0')")
+DIAGNOSTIC(30702, Error, expectedIntegerConstantNotConstant, "expression does not evaluate to a compile-time constant")
+DIAGNOSTIC(30703, Error, expectedIntegerConstantNotLiteral, "could not extract value from integer constant")
```
tfoleyNV: Hmm... The comment says `307xx` is for parameter-related errors, but your edits here have extended that range to include a *lot* of unrelated error messages (including error messages not related to your change).

I'd really prefer that we keep unrelated "cleanup" stuff out of PRs, partly to avoid merge conflicts (there is a non-zero chance that our two open PRs will end up with merge conflicts because of these edits), and partly to avoid having to spend time on design discussion that isn't really about the feature being worked on.

In this case, all the `39999` errors were indeed waiting for a good time to give them more permanent numbers. One problem is that our diagnostics need a lot of scrubbing (there are a bunch of diagnostics that are completely unused), and the numbering realistically needs a major rethink/reorg.

The intention with the header comments like `307xx : parameters` was to start reserving ranges of diagnostic numbers for related messages, so that we could keep the overall space somewhat sparse and not have to worry as much about fitting in a new message near the ones it relates to.

I would have greatly preferred to see this PR reserve `308xx` for literal-related diagnostics (or maybe even numeric literals), and put all its diagnostics in that range without rewriting any of the existing ones.

source/slang/slang-ir.h
```diff
@@ -638,6 +638,14 @@ typedef double IRFloatingPointValue;
 
 struct IRConstant : IRInst
 {
+    enum class FloatKind
+    {
+        Normal,
```
tfoleyNV: Maybe use `Finite` instead of `Normal`?

source/slang/slang-parser.cpp
```diff
@@ -3871,6 +3872,20 @@ namespace Slang
...
     }
 
+    static bool _isFinite(double value)
+    {
+        // Lets type pun double to uint64_t, so we can detect special double values
+        union
+        {
+            double d;
+            uint64_t i;
+        } u = { value };
```
tfoleyNV: Is using `{}` initialization for a `union` well-defined? I've definitely seen it work, but had always assumed it was some compiler-specific language extension stuff.

source/slang/slang-parser.cpp
```diff
@@ -4072,32 +4143,105 @@ namespace Slang
...
+                                fixedValue = float(INFINITY);
+                            }
+                            else if (fixedValue < -FLT_MAX)
+                            {
+                                fixedValue = -float(INFINITY);
+                            }
+                            else if (fixedValue && float(fixedValue) == 0.0f)
+                            {
+                                fixedValue = 0.0f;
+                            }
```
tfoleyNV: As somebody with almost no knowledge of the intricacies of floating-point numbers, I have to ask if this logic makes it impossible for a user to specify a "negative zero" literal. I guess since we only support positive literals right now and negate elsewhere it is unimportant (and negating a "positive zero" doesn't seem to give a negative zero...).

source/slang/slang-parser.cpp
```diff
@@ -4072,32 +4143,105 @@ namespace Slang
...
+                            parser->sink->diagnose(token, Diagnostics::floatLiteralUnrepresentable, BaseTypeInfo::asText(suffixBaseType), token.Content, fixedValue);
+                        }
+                    }
+                }
+
+                Session* session = parser->getSession();
+
+                Type* suffixType = (suffixBaseType == BaseType::Void) ? session->getErrorType() : session->getBuiltinType(suffixBaseType);
+
+                constExpr->value = fixedValue;
```
tfoleyNV: Is it strictly necessary to replace the value here, or should we be able to assume that back-ends will deal with that properly? That is, if we feed the HLSL/GLSL/C++ compiler a literal that is too big for a `float`, won't they produce infinity as well?

source/slang/slang.cpp
```diff
@@ -37,6 +37,61 @@
...
+        case BaseType::Int16:           return UnownedStringSlice::fromLiteral("int16_t");
+        case BaseType::Int:             return UnownedStringSlice::fromLiteral("int");
+        case BaseType::Int64:           return UnownedStringSlice::fromLiteral("int64_t");
+        case BaseType::UInt8:           return UnownedStringSlice::fromLiteral("uint8_t");
+        case BaseType::UInt16:          return UnownedStringSlice::fromLiteral("uint16_t");
+        case BaseType::UInt:            return UnownedStringSlice::fromLiteral("uint");
+        case BaseType::UInt64:          return UnownedStringSlice::fromLiteral("uint64_t");
+        case BaseType::Half:            return UnownedStringSlice::fromLiteral("half");
+        case BaseType::Float:           return UnownedStringSlice::fromLiteral("float");
+        case BaseType::Double:          return UnownedStringSlice::fromLiteral("double");
```
tfoleyNV: Unrelated to this change, but: we probably need to step back at some point and decide what the "official" naming for the various scalar types is inside the Slang implementation. We have a mix here of types that are explicit about their bit width (`uint16_t`) and types that leave it implicit (`half`, `int`).

I'd love to reach a point where the canonical name for `float` in the Slang stdlib is `Float32`, but I suppose that gets us into dangerous places when `float4` and `Float32` mean very different things by their numeric suffixes.

source/slang/slang-diagnostic-defs.h
jsmall-nvidia: Ok - will revert. Added a comment, so it doesn't just seem like an oversight.

source/slang/slang-parser.cpp
jsmall-nvidia: It's okay, because !!-0.0 is false

source/slang/slang-parser.cpp
jsmall-nvidia: If you have any constant folding then you are going to have problems.

source/slang/slang-compiler.h
jsmall-nvidia: Well... there is a similar type in the stdlib. That just having such information defined in one place, would be preferable, but wasn't clear if that was appropriate with stdlib use.

PR #1209, opened 2020-02-07
Code standard changes for Lexer

jsmall-nvidia:
* Module local functions/static functions _ prefix
* m_ prefix member variables for classes with ctor/ (significant) methods
* Method names upper camel -> lower camel

### DIFF
--- a/source/slang/slang-lexer.cpp
+++ b/source/slang/slang-lexer.cpp
@@ -31,44 +31,44 @@ namespace Slang
     }
 
     TokenSpan::TokenSpan()
-        : mBegin(NULL)
-        , mEnd  (NULL)
+        : m_begin(nullptr)
+        , m_end  (nullptr)
     {}
 
     TokenReader::TokenReader()
-        : mCursor(NULL)
-        , mEnd   (NULL)
+        : m_cursor(nullptr)
+        , m_end   (nullptr)
     {}
 
 
-    Token& TokenReader::PeekToken()
+    Token& TokenReader::peekToken()
     {
-        return nextToken;
+        return m_nextToken;
     }
 
-    TokenType TokenReader::PeekTokenType() const
+    TokenType TokenReader::peekTokenType() const
     {
-        return nextToken.type;
+        return m_nextToken.type;
     }
 
-    SourceLoc TokenReader::PeekLoc() const
+    SourceLoc TokenReader::peekLoc() const
     {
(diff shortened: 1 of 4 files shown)


source/slang/slang-lexer.cpp
```diff
@@ -989,182 +989,182 @@ namespace Slang
...
+            case '=': _advance(lexer); return TokenType::OpAddAssign;
             default:
                 return TokenType::OpAdd;
             }
 
         case '-':
-            advance(lexer);
-            switch(peek(lexer))
+            _advance(lexer);
+            switch(_peek(lexer))
```
tfoleyNV: I'm a bit uneasy about putting the `_` on functions like `peek()` and `advance()`, especially when I extrapolate from this to the equivalent refactor for the parser (which admittedly has a lot more legacy-related cruft that really merits refactoring).

The design pattern in both the lexer and parser is supposed to be that there is a set of "core" routines that deal with the tricky invariants and policies, while the majority of "user" routines are expressed entirely in terms of that core.

Concrete examples are how almost none of the lexer code needs to concern itself with source location tracking, backslash-escaped newlines, UTF-8 decoding (if we support it - I honestly can't recall and it makes no difference to the "user" code), and the tracking of which bytes make up the value of a token. For the parser, a big detail hidden by the "core" API is how error reporting and recovery occurs (you just call `ReadToken(TokenType::semicolon)` and it automatically handles error recovery).

As such, some of these functions that are getting an underscore are things that I think of as "public API" - it is just a public API between the "core" and "user" parts of a single system.

Anyway, I'm fine with this change as proposed, but it almost makes me inclined to factor the lexer into multiple files just to make the core/user distinction more clear and justify removing the `_`. :)

In my mind each of the lexer and parser is divided into a "core" implementation and a bunch of "user-space" code.

The "core" implementation manages the tricky state/invariants and hides details that the majority of code shouldn't have to think about. That includes stuff like tracking source location information and dealing with error recovery in the parser. The core implementation exposes a small set of key APIs that should be used for inspecting and consuming the input, and those APIs are carefully designed to hide many details. E.g., most lexer code doesn't even need to know/care if we support UTF-8 or just ASCII for the input stream.

The "user" code in the lexer/parser then comprises the majority of the logic that doesn't need to worry

jsmall-nvidia: Ok understood.

PR #1210, opened 2020-02-07
Add attributes to enable dual-source blending on Vulkan

tfoleyNV:
This change adds support for the `[[vk::location(...)]]` and `[[vk::index(...)]]` attributes, which can be used together to mark up shader outputs for dual-source blending on Vulkan. HLSL/Slang code like the following:

```hlsl
struct Output
{
    [[vk::location(0)]]
    float4 a : SV_Target0;

    [[vk::location(0), vk::index(1)]]
    float4 b : SV_Target1;
}

[shader("fragment")]
Output main(...) { ...}
```

can be used to set up dual-source blending on both D3D and Vulkan APIs. The output GLSL for the above will look something like:

```glsl
layout(location = 0)            out vec4 a;
layout(location = 0, index = 1) out vec4 b;

void main() { ... }
```

The more or less straightforward parts of this change were:

* Added new `attribute_syntax` declarations to the stdlib, for `[[vk::location(...)]]` and `[[vk::index(...)]]`

* Added new AST node types for the new attribute cases, sharing a base class so that argument checking can be shared

* Added checks for the arguments to the new attributes in `slang-check-modifier.cpp` (eventually this kind of logic shouldn't be needed for new attributes)

* Updated GLSL emit logic so that it treats the `index`/`space` parts of a variable layout as the `location`/`index` for varying parameters.

* Updated GLSL legalization so that when it translates entry-point parameters into globals (and scalarizes structures) it handles both a binding index and space for the parameters.

* Added a cross-compilation test case to verify that the basics of the feature work

The remaining work is all in `slang-parameter-binding.cpp`.

There is some work that isn't technically related to this change (and which could be reverted if it causes problems), around the detection and handling of fragment shader outputs with `SV_Target` semantics. The basic changes (which could be backed out and then merged separately) are:

* Made the special-case `SV_Target` logic only trigger for fragment shaders (that is the only place where `SV_Target` should appear, but we weren't guarding against it)

* Made the logic to reserve a `u<N>` register for `SV_Target<N>` only trigger for D3D Shader Model 5.0 and below (since it is not required for SM 5.1 and up). This could be a breaking change for some users, but that seems unlikely.

* Fixed one test case that relied on the behavior of reserving `u0` for `SV_Target0` even though it was a SM6.0 test.

* Also added more comments to the system-value handling logic.

The more interesting changes come up starting in `processEntryPointVaryingParameterDecl()`. The basic issue is that we have so far only supported implicit layout for varying parameters on GLSL/Vulkan, but the `[[vk::location(...)]]` attribute is a form of explicit layout annotation. Rather than try to kludge something that only works in narrow cases, I instead opted to try to fix things more generally.

In `processEntryPointVaryingParameterDecl()` we now check for the `location` and `index` attributes when we are on "Khronos" targets (Vulkan/OpenGL/GLSL) and immediately add them to the variable layout being constructed if they are found. There is nothing in this logic specific to fragment-shader outputs, so this feature now applies to any varying input/output on Khronos targets.

Allowing explicit layouts creates the potential for mixing implicit and explicit layout. For example, consider:

```hlsl
struct Output
{
    float4 color : COLOR;
    [[vk::location(0)]] float3 normal : NORMAL;
}
```

What `location` should `color` get? Should this code be an error? There are two cases where this conundrum can come up: when working with `struct` types used for varying parameters, and the entry-point parameter list itself.

For the varying `struct` case we currently make an expedient choice. We handle fields with both implicit or explicit layotu with appropriate logic, but logic that doesn't account for the case of mixing the two. Then at the end of layout for the `struct` we issue an error if there was a mix of implicit and explicit layout (such that our results aren't likely to be valid).

For the entry point varying parameter case, things were already using a `ScopeLayoutBuilder` type (that encapsulates some logic shared between entry-point and global parameters). The entry-point-specific bits were moved out into a `SimpleScopeLayoutBuilder` and it was updated so that rather than assuming all parameters use implicit layout it does a two-phase layout approach similar to what we use for the global scope:

* First all parameters are enumerated to collect explicit bindings and mark certain ranges as "used"

* Next the parameters are enumerated again and those without explicit bindings get allocated space using a "first fit" algorithm

In principle we could extend the two-phase approach to apply to `struct` types as well, but that would be best saved for a future refactoring of some of this parameter binding logic, since I would like to exploit more of the opportunities for sharing code across the uniform/varying and struct/entry-point/global cases.

By moving the point where entry point parameters get their offsets assigned, it was necessary to move around some of the logic that removes varying parameter usage (and other things that shouldn't "leak" out of an entry point) to a different point in the entry point layout process.

While adding these various pieces does not quite enable us to support explicit bindings on entry point parameters (e.g., putting `uniform Texture2D t : register(t0)` in an entry point parameter list) or in `struct` types (e.g., explicit `packoffset` annotations on fields), it starts to provide some of the infrastructure that we'd need in order to support those cases.


source/slang/slang-type-layout.cpp
```diff
@@ -3899,6 +3899,19 @@ void TypeLayout::removeResourceUsage(LayoutResourceKind kind)
...
 }
 
+void VarLayout::removeResourceUsage(LayoutResourceKind kind)
+{
+    Int infoCount = resourceInfos.getCount();
+    for( Int ii = 0; ii < infoCount; ++ii )
+    {
+        if( resourceInfos[ii].kind == kind )
+        {
+            resourceInfos.removeAt(ii);
```
jsmall-nvidia: Presumably you could use fastRemoveAt - doesn't seem to be an ordering of resourceInfos.

source/slang/slang-parameter-binding.cpp
```diff
@@ -1517,11 +1612,89 @@ static RefPtr<TypeLayout> processEntryPointVaryingParameterDecl(
...
+                // When using dual-source blending the GLSL/SPIR-V looks like:
+                //
+                //      layout(location = 0)            vec4 a;
+                //      layout(location = 0, index = 1) vec4 b;
+                //
+                // Thus for a parameter of kind `VaryingOutput` when targetting
+                // GLSL/SPIR-V, we need a way to encode the value that was pased
+                // for `index` on the secondary color output.
+                //
+                // We are already using the `index` field in the `VarLayout::ResourceInfo`
```
jsmall-nvidia: I didn't see this aspect of space usage change discussed on the space variable comment - which might be useful for maintainers.

source/slang/slang-type-layout.cpp
tfoleyNV: I don't think we have any code that relies on the ordering of the resourceInfos, but I've tried not to write code that would disturb their order once established. There are two different future directions where might want to guarantee an ordering of the resourceInfos, and I want to leave the door open for them:

1. We might decide that the first resourceInfo should always be the "primary" resource usage for a type. For example we might want a `ConstantBuffer<X>` to always report the constant buffer register/binding as the first resourceInfo (as a guarantee to the reflection API user), whether or not `X` contains textures or other resources.

2. Probably more interesting: if we guarantee that the resourceInfos are always sorted by `LayoutResourceKind` then we can look up a particular resource kind with a binary search instead of the linear search we do today.

PR #1211, opened 2020-02-07
Fixes to make all CPU compute shaders work on CUDA

jsmall-nvidia:
* Add support for SampleLevel
  * Fixed problem with Sample - used the wrong type
* Added $T 'special macro magic' option. 
* Fixed kernel launch for CUDA to take into account blocks
* Turned isCPUOnly hack into 'onlyCPULikeBinding'. Still horrible - but hey it now works on CUDA!

### DIFF
--- a/source/slang/slang-emit-c-like.cpp
+++ b/source/slang/slang-emit-c-like.cpp
@@ -1367,6 +1367,22 @@ void CLikeSourceEmitter::emitIntrinsicCallExprImpl(
                 }
                 break;
 
+            case 'T':
+                // Get the the 'element' type for the type of the param at the index
+                {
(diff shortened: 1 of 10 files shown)


source/slang/slang-emit-c-like.cpp
```diff
@@ -1367,6 +1367,22 @@ void CLikeSourceEmitter::emitIntrinsicCallExprImpl(
...
+
+                    IRType* type = args[argIndex].get()->getDataType();
+                    if (auto baseTextureType = as<IRTextureType>(type))
+                    {
+                        type = baseTextureType->getElementType();
+                    }
+                    emitType(type);
+                }
+                break;
+
```
tfoleyNV: I'd prefer not to burn `T` for this, since `$T0` is the most obvious syntax for referencing generic argument 0. Not a big deal though.

PR #1216, opened 2020-02-11
Small improvements around List

jsmall-nvidia:
* Fixed a typo in bounds checking, and improved bounds checks
* Converts tabs -> spaces
* Made List const correct
  * Fixed small amount of code which relied on non-const behavior
* Changed some (non-list) UpperCamel methods to lowerCamel
* Added hasContent() to FilteredMemberList, to simplify some code iterating over lists
   * Came up because change to list, made VS produce 'unreachable code' in some places


source/core/slang-list.h
```diff
@@ -43,126 +43,136 @@ namespace Slang
...
+        {
+            delete [] ptr;
+        }
+    };
+
+    // List is container of values of a type held consecutively in memory (much like std::vector)
+    // 
+    // Note that in this implementation, the underlying memory is backed via an allocation of T[capacity]
+    // This means that all values have to be in a valid state *even if they are not used* (ie indices >= m_count must be valid)
+    //
```
tfoleyNV: Should we try to fix this behavior? It seems to be obviously wrong as soon as `List<...>` gets used together with types that have destructors, which is something we do all the time right now. There is merit in a similar type that is designed to only work with POD types, but that is also one of those things that "modern C++" at least has an answer for.

source/slang/slang-check-decl.cpp
```diff
@@ -2779,11 +2779,7 @@ namespace Slang
         //      subscript(uint index) -> T { get; }
         //
 
-        bool anyAccessors = false;
-        for(auto accessorDecl : decl->getMembersOfType<AccessorDecl>())
-        {
-            anyAccessors = true;
-        }
+        bool anyAccessors = decl->getMembersOfType<AccessorDecl>().hasContent();
```
tfoleyNV: Not sure if it was added in this change, but `hasContent` is a slightly odd function name. I would have expected `!someList.isEmpty()` to be the more natural way to express that idea.

source/slang/slang-syntax.h
```diff
@@ -753,8 +753,8 @@ namespace Slang
 
         struct Iterator
         {
-            Element* m_cursor;
-            Element* m_end;
+            const Element* m_cursor;
+            const Element* m_end;
```
tfoleyNV: I don't think that `FilteredMemberList` had specifically implied it was a read-only iterator over the collection, but I would also concede/agree that there is no reasonable use case that would require it to provide mutable access.

source/slang/slang-syntax.h
```diff
@@ -819,8 +819,10 @@ namespace Slang
             return result;
         }
 
-        Element* m_begin;
-        Element* m_end;
+        bool hasContent() const { return m_end != m_begin; }
```
tfoleyNV: Okay, so `hasContent` *was* introduced here. I might argue for `isEmpty` as the more natural primitive, but I guess all the current call sites actually want the opposite. Would `isNonEmpty` be reasonable?

source/core/slang-list.h
jsmall-nvidia: Yes I think we should fix this aspect, and make it not backed via T[], as that just wastes memory and assumes everything is initialized, even when it doesn't need to be.

PR #1218, opened 2020-02-11
CUDA barrier/atomic support

jsmall-nvidia:
* First attempt at stdlib support for CUDA atomic/barriers
* Enabled barrier/atomic tests for CUDA

Not clear just yet if RWBuffer style atomic-buffer.slang will work for CUDA.


source/slang/hlsl.meta.slang
```diff
@@ -347,10 +347,13 @@ __generic<T : __BuiltinType, let N : int, let M : int> bool all(matrix<T,N,M> x)
 
 // Barrier for writes to all memory spaces (HLSL SM 5.0)
 __target_intrinsic(glsl, "memoryBarrier(), groupMemoryBarrier(), memoryBarrierImage(), memoryBarrierBuffer()")
+// TODO(JS): Doesn't seem to be weaker form of sync, so use this?
+__target_intrinsic(cuda, "__syncthreads()")
```
tfoleyNV: We should try to find out if we need to issue a memory barrier separate from the `__syncthreads()` because I think that function might only synchronize execution and shared memory access, but not anything like global memory access.

source/slang/hlsl.meta.slang
```diff
@@ -812,99 +816,132 @@ float2 GetRenderTargetSamplePosition(int Index);
 __target_intrinsic(glsl, "groupMemoryBarrier")
 void GroupMemoryBarrier();
 
+
 __target_intrinsic(glsl, "groupMemoryBarrier(), barrier()")
+// TODO(JS): I don't think __syncwarp can be used because a 'group' could be multiple warps
+__target_intrinsic(cuda, "__syncthreads()")
```
tfoleyNV: Correct. I don't think there is a way in the current HLSL programming model to do warp-level or sub-group synchronization.

source/slang/hlsl.meta.slang
```diff
@@ -812,99 +816,132 @@ float2 GetRenderTargetSamplePosition(int Index);
...
 void GroupMemoryBarrierWithGroupSync();
 
 // Atomics
 
 __target_intrinsic(glsl, "$atomicAdd($A, $1)")
+__target_intrinsic(cuda, "atomicAdd(&$0, $1)")
```
tfoleyNV: Having to put `&` in front of `$0` here is a byproduct of the known issue in our C++/CUDA codegen where we emit the `someVar` instead of `&someVar` and so forth for expressions of pointer type.

The fix here is the expedient one, but we will eventually need/want to revisit the original issue in the emit logic.

source/slang/hlsl.meta.slang
jsmall-nvidia: From the CUDA docs for __syncthreads()...

```waits until all threads in the thread block have reached this point and all global and shared memory accesses made by these threads prior to __syncthreads() are visible to all threads in the block.

__syncthreads() is used to coordinate communication between the threads of the same block. When some threads within a block access the same addresses in shared or global memory, there are potential read-after-write, write-after-read, or write-after-write hazards for some of these memory accesses. These data hazards can be avoided by synchronizing threads in-between these accesses.

__syncthreads() is allowed in conditional code but only if the conditional evaluates identically across the entire thread block, otherwise the code execution is likely to hang or produce unintended side effects. ```

PR #1223, opened 2020-02-13
Feature/cuda coverage

jsmall-nvidia:
For CUDA

* Added cube map support 
* Added fences
* Added support in harness for 1d textures
* Added gather texture functions
* Moved to driver API usage as much as possible

For CPU

* Added Texture1D support
* Made texture value alterable

Other

* Added simple compute texture test. 
* Made unit vector/matrices map to the element type on CPU/CUDA targets


prelude/slang-cpp-types.h
```diff
@@ -228,6 +228,23 @@ struct SamplerComparisonState
...
+    virtual void Sample(SamplerState samplerState, float loc, void* out) = 0;
+    virtual void SampleLevel(SamplerState samplerState, float loc, float level, void* out) = 0;
+};
+
+template <typename T>
+struct Texture1D
+{
+    T Load(const int2& v) const { T out; texture->Load(v, &out); return out; }
+    T Sample(SamplerState samplerState, float v) const { T out; texture->Sample(samplerState, v, &out); return out; }
+    T SampleLevel(SamplerState samplerState, float v, float level) { T out; texture->SampleLevel(samplerState, v, level, &out); return out; }
```
tfoleyNV: Should forwarding functions like this be marked with a `SLANG_PRELUDE_INLINE` or similar macro (that can be defined to be the appropriate force-inline flag for a given compiler)?

tfoleyNV: One issue here (that I don't think we need to solve in this PR) is that the virtual `ITexture2D` interface doesn't get any information about the type of the `out` parameter that is passed in.

There are three kinds of "mismatch" that can happen here:

1. Mismatch on the "kind" of the component type: e.g., trying to fetch a `uint` from a texture with `float` or `int` components.
2. Mismatch on the size of the component type: e.g., trying to fetch a `half` from a texture with `float` components
3. Mismatch on the number of components: e.g., trying to fetch a `float2` from a texture with `float4` texels

I'd need to double-check the Vulkan and D3D rules, but I suspect it would be okay to make (1) an "undefined behavior" or runtime error case, but (2) and (3) need to be handled correctly.

The easiest solution (it seems) would be to bin textures into three cases for "kinds" (issue 1), similar to how GLSL does: `float`, `uint`, and `int` components. Then the body of a forwarding function like `Texture1D<T>::Load` would need to declare and pass along an `out` variable that matches the kind, but has the default size and component count.

For example, the body of `Texture1D<half2>::Load` would need to declare a local `out` variable of type `float4`, pass that into the `ITexture2D::Sample` function, and then extract/convert the `out.xy` components.

Doing that inside the prelude code would involve a fair bit of template magic (all do-able, but ugly). It might be best to punt on this issue until we are at a point where we could consider doing this in an IR pass instead.

(It is worth noting that we already have logic for dealing with this in our GLSL emit, so anything we do for CPU would be at least partially redundant with that. It would be great to put something into the IR so it can be shared...)

source/slang/core.meta.slang
```diff
@@ -1028,7 +1029,9 @@ for (int tt = 0; tt < kBaseTextureTypeCount; ++tt)
                     sb << "constexpr int" << kBaseTextureTypes[tt].coordCount << " offset);\n";
                 }
 
-
+                // TODO(JS): Not clear how to map this to CUDA, because in HLSL, the gradient is a vector based on
+                // the dimension. On CUDA there is texNDGrad, but it always just takes ddx, ddy.
+                // I could just assume 0 for elements not supplied, and ignore z. For now will just leave
```
tfoleyNV: Last sentence seems to be incomplete here.

PR #1226, opened 2020-02-18
Added support for Targets to TypeTextUtil.

jsmall-nvidia:
* Moved getting targets by name, getting name from target into TypeTextUitl
* Expanded what names/extensions were supported.


source/core/slang-type-text-util.cpp
```diff
@@ -36,13 +37,43 @@ struct ScalarTypeInfo
...
+    { SLANG_TARGET_UNKNOWN, "",                                                 "unknown"},
+    { SLANG_TARGET_NONE,    "",                                                 "none"},
+    { SLANG_HLSL,           "hlsl,fx",                                          "hlsl"},
+    { SLANG_DXBC,           "dxbc",                                             "dxbc"},
+    { SLANG_DXBC_ASM,       "dxbc.asm",                                         "dxbc-asm,dxbc-assembly" },
+    { SLANG_DXIL,           "dxil",                                             "dxil" },
+    { SLANG_DXIL_ASM,       "dxil.asm",                                         "dxil-asm,dxil-assembly" },
+    { SLANG_GLSL,           "glsl,vert,frag,geom,tesc,tese,comp",               "glsl" },
+    { SLANG_GLSL_VULKAN,    "",                                                 "glsl-vulkan" },
+    { SLANG_GLSL_VULKAN_ONE_DESC, "",                                           "glsl-vulkan-one-desc" },
```
tfoleyNV: Unrelated, but these should probably be considered fully deprecated and ready for removal. No current client code is passing these in, and the `*_ONE_DESC` variant was specifically created for the Slang research effort.

If/when we start to distinguish Vulkan GLSL from OpenGL GLSL, we will need to invent a new way to do it.

source/core/slang-type-text-util.cpp
```diff
@@ -160,6 +191,54 @@ static const ScalarTypeInfo s_scalarTypeInfo[] =
     return UnownedStringSlice::fromLiteral("unknown");
 }
 
+/* static */SlangCompileTarget TypeTextUtil::asCompileTargetFromExtension(const UnownedStringSlice& slice)
```
tfoleyNV: Just my preference, but I'd prefer to call this something like `findCompileTargetForExtension` or `findCompileTargetFromExtension`, where "find" both implies that there is some searching going on, as well as that the search might not result in anything being found. Using `as` here implies a cast to me, which isn't quite what this is.

source/core/slang-type-text-util.cpp
```diff
@@ -160,6 +191,54 @@ static const ScalarTypeInfo s_scalarTypeInfo[] =
...
+    {
+        if (s_compileTargetInfos[i].target == target)
+        {
+            return i;
+        }
+    }
+    return -1;
+}
+
+UnownedStringSlice TypeTextUtil::asNameFromCodeGenTarget(SlangCompileTarget target)
```
tfoleyNV: Similarly, I'd call this just `getCompileTargetName` since this operation can't really fail (or at least it shouldn't - every target should have a canonical name that will be used in diagnostic messages, etc.).

tfoleyNV: Not really related to this change, but `SlangCompileTarget` is poorly named, so having more functions using "compile target" to refer to this type is a bit ugly.

Really the `SlangCompileTarget` enumeration covers a target *format* that the output is encoded in (e.g., DXBC vs. DXIL) which is only one aspect of what defines a compilation target in Slang today. The notion of a compilation target is much closer to the `TargetReq` type at this point, which uses `SlangCompileTarget` only to represent the output format.

source/core/slang-type-text-util.cpp
jsmall-nvidia: Fair point.

PR #1235, opened 2020-02-20
Initial support for explicit default initializers

tfoleyNV:
This change makes it so that for a suitable type `MyType`, a variable declaration like:

    MyType v;

is treated as if it were written:

    MyType v = MyType();

The definition of "suitable" here is that `MyType` needs to have an available `__init` declaration that can be invoked with zero arguments. I've added a test to confirm that the new behavior works in this specific case.

There are a bunch of caveats to the feature as it stands today:

* Just because `MyType` has a zero-parameter `__init`, that doesn't mean an array type like `MyType[10]` does, so arrays currently remain uninitialized by default. Fixing this gap requires careful consideration because some, but not all, array types should be default-initializable.

* The change here should mean that a `struct` type with a field like `MyType f;` should count as having a default initial-value expression for that field, but I haven't confirmed that.

* Even if a `struct` provides initial values for all its fields (e.g., `struct S { float f = 0; }`), that doesn't mean it has a default `__init` right now, so those `struct` types will still be left uninitialized by default. Converging all this behavior is still TBD.

Just to be clear: there is no provision or plan in Slang to support destructors, RAII, copy constructors, move constructors, overloaded assignment operations, or any other features that buy heavily into the C++ model of how construction and destruction of values gets done.

In fact, I'm not even 100% sure I like having this change in place at all, and I think we should reserve the right to revert it and say that only specific stdlib types get to opt in to default initialization along these lines.


tests/compute/default-initializer.slang
```diff
@@ -0,0 +1,35 @@
...
+
+// Confirm that a type with a default initializer gets
+// default-initialized where appropriate.
+
+//TEST(compute):COMPARE_COMPUTE:
+//TEST(compute):COMPARE_COMPUTE:-cpu
+
+struct Stuff
+{
+	int value;
```
jsmall-nvidia: The tabbing looks odd here.

tfoleyNV: I think the editor I'm using to write tests is inserting actual tab characters, and I sometimes end up with a mix of tabs and spaces as a result. At this point it might be best to have a standalone PR that just does a full tabs->spaces pass on all the test files.

PR #1243, opened 2020-02-25
Support for RWTexture types on CPU and CUDA

jsmall-nvidia:
* Moved location to be * type so only needed ITexture and IRWTexture interface
* Fix bug to pad out UniformState to match layout
* Add surf (RWTexture) types for CUDA to StdLib

The main remaining anomaly here is that writes to RWTexture1D on CUDA do not work as expected. The correct CUDA code appears to be emitted and is appropriately compiled by nvrtc, but when data is read back from a write it is 0. It works as expected for 2D, 3D. 

For now this means RWTexture1D read is disabled.


prelude/slang-cpp-types.h
```diff
@@ -228,136 +228,132 @@ struct SamplerComparisonState
...
-struct ITexture1D
+struct ITexture
 {
-    virtual void Load(const int2& v, void* out) = 0;
-    virtual void Sample(SamplerState samplerState, float loc, void* out) = 0;
-    virtual void SampleLevel(SamplerState samplerState, float loc, float level, void* out) = 0;
+    virtual void Load(const int* v, void* out) = 0;
+    virtual void Sample(SamplerState samplerState, const float* loc, void* out) = 0;
+    virtual void SampleLevel(SamplerState samplerState, const float* loc, float level, void* out) = 0;
 };
```
tfoleyNV: This seems like a really good simplification.

prelude/slang-cpp-types.h
```diff
@@ -228,136 +228,132 @@ struct SamplerComparisonState
...
+    T SampleLevel(SamplerState samplerState, const float4& loc, float level) { T out; texture->SampleLevel(samplerState, &loc.x, level, &out); return out; }
     
-    ITexture1DArray* texture;              
+    ITexture* texture;              
 };
 
-struct ITexture2DArray
+/* !!!!!!!!!!!!!!!!!!!!!!!!!!! RWTexture !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! */
+
+struct IRWTexture
```
tfoleyNV: It is correct/accurate that `RWTexture2D` would not have the `Sample` operations and similar stuff from `Texture2D`? Otherwise, I would kind of expect `IRWTexture` to inherit from `ITexture`

jsmall-nvidia: Yes the RWTexture seems like it should derive from 'read only' Texture, but it is sort of something different, and doesn't support sampleState. It's the same under CUDA with surfaces.

PR #1245, opened 2020-02-26
Start moving some built-in definitions to stdlib

tfoleyNV:
Some of the functions declared in the Slang standard library are built in on some targets (almost always the case for HLSL) but aren't available on other targets (often the case for GLSL, CUDA, and CPU). To date, the CUDA and CPU targets have worked around this issue by synthesizing definitions of the missing functions on the fly as part of output code generation, at the cost of some amount of code complexity in the emit pass.

This change tries to move things more in the direction that Slang intends to support longer-term, by having those functions be defined in the standard library itself (written in ordinary Slang code), so that they can be instantiated, specialized, and code-generated into the output just like any user-written function would be.

This change picks a few functions to port (largely just starting at the top of `hlsl.meta.slang` and looking for candidates) to make sure that the approach is workable.


tfoleyNV:
> For 2 it might be better to just have a decoration such that it is marked as allowed for those types, and the appropriate code generation is just done with some code in the slang codebase.

The trade-off there is that it would amount to having a new custom attribute to represent the idea of "this function should behave as if it has a body, and the body is just a component-wise (or row-wise) map over its arguments of a function with the same name," and then we'd need the downstream logic both to expand that into actual code (which wouldn't look that different from what we have now - at least not until `N` is known), which would also need a way to look up the "same" function one rank down (matrix->vector, vector->scalar). That would amount to a lot of ad hoc work.

What I might advocate for (and part of the reason why I used a macro so far) is having a construct like a `__vector_map` intstruction, where `__vector_map<T,N>(i, foo(i))` effectively acts as a constructor for a `vector<T,N>` where each element `i` is initialized with `foo(i)` (for some arbitrary expression in terms of `i`). We could then directly represent `__vector_map`s in the AST and IR, and then we could easily define an IR pass that expands them into straight-line code when `N` is constant, or into a loop when `N` isn't. The nice thing about such an approach is that it gets rid of the use of a local variable, looping, and mutation in these somewhat core subroutines.

Alternatively, as we'd discussed, we might just try to mark the loops introduced here with some internal-use-only `[__unroll_special]` attribute that turns on a special-purpose loop unroller that doesn't handle any difficult or corner cases.

I guess the point I'm making is that the more we can make these definitions just be ordinary code (that might make use of some internal-use-only expressions or attributes), the better I'll feel. The use of macros in the current code is definitely intended to give us the freedom to try out these alternatives down the line.

tfoleyNV:
As discussed elsewhere, I may not want to check this in until we can answer questions about the potential perf impact of having more code in the stdlib.

Given that I'm going to keep adding more of these declarations (now that the basic idea is approved), it might make sense for me to just close this PR and re-open it later with more code.

PR #1250, opened 2020-02-27
Constant time dynamic cast

jsmall-nvidia:
* Replaced getClass with getClassInfo on NodeBase
* Calculate a range for each ClassInfo that all sub classes must be in
* Implement dynamic cast/as inline
* Use range for dynamic casting

On Visual Studio nearly halves stdlib compile time on release.

### DIFF
--- a/source/slang/slang-syntax.cpp
+++ b/source/slang/slang-syntax.cpp
@@ -35,15 +35,14 @@ namespace Slang
     SyntaxClassBase::ClassInfo const SyntaxClassBase::Impl<NAME>::kClassInfo(#NAME, nullptr, &SyntaxClassBase::Impl<BASE>::kClassInfo);
 
 #define SYNTAX_CLASS(NAME, BASE)                                                \
-    void NAME::accept(NAME::Visitor* visitor, void* extra)                      \
-    { visitor->dispatch_##NAME(this, extra); }                                  \
     template<>                                                                  \
     void* SyntaxClassBase::Impl<NAME>::createFunc() { return new NAME(); }      \
-    SyntaxClass<NodeBase> NAME::getClass() { return Slang::getClass<NAME>(); }  \
-    const SyntaxClassBase::ClassInfo& NAME::getClassInfo() const { return SyntaxClassBase::Impl<NAME>::kClassInfo; } \
     template<>                                                                  \
-    SyntaxClassBase::ClassInfo const SyntaxClassBase::Impl<NAME>::kClassInfo( #NAME, &SyntaxClassBase::Impl<NAME>::createFunc, &SyntaxClassBase::Impl<BASE>::kClassInfo);
-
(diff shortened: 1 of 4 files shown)


source/slang/slang-syntax.cpp
```diff
@@ -71,26 +71,117 @@ ABSTRACT_SYNTAX_CLASS(GlobalGenericParamSubstitution, Substitutions);
...
-    return nullptr;
+    // Remove the warning about not referenced
+    SLANG_UNUSED(&_checkSubClassRange);
+
+    // TODO(JS):
+    // Note that the calculating of the ranges could be done more efficiently by adding to an array of struct { super, class }, sorting, by super classs
+    // and using a dictionary to map from class it's first in list of super class use. This works for now though.
+  
+    // We want to produce a map from a node that holds all of it's children
+    Dictionary<const ThisType*, List<const ThisType*> > childMap;
```
tfoleyNV: What seems like an even simpler alternative to me would be that instead of a single global linked list of all the `ClassInfo`s, you could have each `ClassInfo` hold a pointer for a singly-linked list of direct subclasses. Then the constructor for `ClassInfo` would add it to the list of the super class (if any), and you'd get a single tree structure that you could recursively walk in one go to assign the numeric ranges.

source/slang/slang-syntax.h
```diff
@@ -549,23 +572,18 @@ namespace Slang
         return SyntaxClass<T>::getClass();
     }
 
-    NodeBase* _dynamicCastImpl(NodeBase* node, SyntaxClassBase const& toClass);
-
```
tfoleyNV: Did you measure that moving the casting logic out of this subroutine and into the inline functions actually paid off? Usually I've been trying to put as *little* code into templated functions as possible, and always have them immediately delegate to a non-template function in order to add as little code size increase as possible.

jsmall-nvidia: The inlining like this originally showed a small benefit. With the range change - the difference seems less obvious. With it being inlined and the amount of code being pretty small I'd expect it to be ok - but in might produce enough more code to impact code size perhaps.

source/slang/slang-syntax.cpp
jsmall-nvidia: It's an interesting idea. One problem is construction order - well it's the order they are defined within TU, but here the order the variables are specified does not honor super/child relationship, so I'd be modifying an un-constructed variable. Might be ok - but probly a bit dodgy. But the observation made me make sure that s_first is constructed first. I'd note in practice that the that style works across TU because assigning 0 to a global becomes BSS, and so is 'executed' before anything else - but that specific behavior is not strictly part of C++ standard.

tfoleyNV: Construction order is a good point. Right now all these variables get defined in a single TU, and as far as I know the declaration order is consistent with the inheritance graph (for the simple reason that the same macro magic generated the `class` declarations too, and those need super-classes to come before subclasses also). It probably isn't worth refactoring this to take advantage of any such assumptions since the long-term goal would be to not require this logic in the first place.

PR #1252, opened 2020-02-28
Additional Wave Intrinsic Support

jsmall-nvidia:
* Additional support for wave intrinsics on CUDA
* Fix problem with Wave intrinsics not working because of __BuiltinIntegerType not being set
* Added wave.slang test


source/core/slang-nvrtc-compiler.cpp
```diff
@@ -307,6 +307,9 @@ SlangResult NVRTCDownstreamCompiler::compile(const CompileOptions& options, RefP
         // This is arguably too much - but nvrtc does not appear to have a mechanism to switch off individual warnings.
         // I tried the -Xcudafe mechanism but that does not appear to work for nvrtc
         cmdLine.addArg("-w");
+
+        //
+        cmdLine.addArg("-arch=compute_70");
```
tfoleyNV: We should eventually add a new "profile family" to capture the CUDA target capabilities that are based on architecture version. That would require adding logic to mark certain builtins as requiring a specific version/architecture, similar to how we already do things for GLSL.

Whether to do this soon, or wait for a more complete/holistic approach to handling profiles/capabilities is unclear.

source/slang/core.meta.slang
```diff
@@ -132,9 +132,12 @@ for (int tt = 0; tt < kBaseTypeCount; ++tt)
     case BaseType::Half:
     case BaseType::Float:
     case BaseType::Double:
-        sb << "\n    , __BuiltinFloatingPointType\n";
+        sb << "\n    ,  __BuiltinFloatingPointType\n";
         sb << "\n    ,  __BuiltinRealType\n";
-        ; // fall through to:
```
tfoleyNV: Good catch.

source/slang/hlsl.meta.slang
```diff
@@ -1395,35 +1395,51 @@ __generic<T : __BuiltinType> T QuadReadAcrossDiagonal(T localValue);
 __generic<T : __BuiltinType, let N : int> vector<T,N> QuadReadAcrossDiagonal(vector<T,N> localValue);
 __generic<T : __BuiltinType, let N : int, let M : int> matrix<T,N,M> QuadReadAcrossDiagonal(matrix<T,N,M> localValue);
 
-__generic<T : __BuiltinIntegerType> T WaveActiveBitAnd(T expr);
+__generic<T : __BuiltinIntegerType>
+__target_intrinsic(cuda, "_waveAnd(__activemask(), $0)")
```
tfoleyNV: Is there a specific reason to favor this lowering vs. just having the CUDA prelude define a `WaveActiveBitAnd` function that uses `__activemask()` in its definition? Is the `_waveAnd` function going to be used elsewhere in the prelude?

tests/hlsl-intrinsic/wave.slang
```diff
@@ -0,0 +1,36 @@
...
+    int sum = WaveActiveSum(idx);
+    value |= (sum << 4);
+    
+    // TODO(JS): 
+    // This result is unexpected. I expect 1 * 2 * 1 * 2 = 4. But we get 0 on CUDA and DX.
+    int product = WaveActiveProduct((idx & 1) + 1);
+    value |= (product << 8);    
+        
+    // TODO(JS): NOTE! This only works with uint, *NOT* int on HLSL/DXC. 
+    // We need to update the stdlib to reflect this.
```
tfoleyNV: That's surprising, but should be an easy fix (and the function would then be callable on `int` anyway, thanks to casting).

source/slang/hlsl.meta.slang
jsmall-nvidia: It was as I believed I could do the Prefix versions by anding with __lanemask_lt(). That's wrong in two ways though. 

1) __lanemask_lt() isn't an intrinsic it seems(!), even though we have this example https://devblogs.nvidia.com/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/
2) All _sync functions must be invoked by all threads in the mask, with the *same* mask otherwise undefined results. So that wouldn't work anyway.

 So may make sense to remove now.

PR #1254, opened 2020-03-02
Renamed UnownedStringSlice::size to getLength to make match String.

jsmall-nvidia:
Renamed 'size' on UnownedStringSlice to 'getLength()' to match String and elsewhere.

### DIFF
--- a/source/core/slang-gcc-compiler-util.cpp
+++ b/source/core/slang-gcc-compiler-util.cpp
@@ -20,7 +20,7 @@ namespace Slang
     {
         if (line.startsWith(prefix))
         {
-            const UnownedStringSlice remainingSlice = UnownedStringSlice(line.begin() + prefix.size(), line.end()).trim();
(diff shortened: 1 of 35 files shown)


tfoleyNV:
I can't remember where we landed on this, but I think we've talked about the naming here before.

The underlying operation here is a somewhat dangerous one because it measures the number of UTF-8 code units in the string, which is pretty much only useful if you want to allocate a buffer of matching size, *or* you want to iterate over the string as bytes (perhaps looking for specific ASCII code points). For both of those purposes I might claim that `getSize()` is the better name, and `String::getLength()` is poorly named.

Ideally we'd choose a name for this operation ("how many bytes does the string take up?") in a way that leaves the door open to having future APIs that are more Unicode-aware.

Maybe even `getSizeInBytes()` is more appropriate, although wordy.

For now I guess unifying around `getLength()` is appropriate since it is probably less churn than changing `String`.

PR #1257, opened 2020-03-05
Improve glslang binary compatibility

jsmall-nvidia:
Safer binary compatibility between 1.0 and 1.1 versions, without using struct embedding, which can produce different results because of structs aligment requirements.


jsmall-nvidia:
When I used a contained struct the idea was I would later remove it. But on thinking about this a little more - depending on target/alignment, doing such a transformation is unsafe, as the variables after the contained struct are now aligned to the largest contained alignment. 

It is possible that in practice that might have been okay. But it seemed like a rocky place to start trying to get binary compatibility right. So just took the pain.

PR #1263, opened 2020-03-06
Yet more definitions moved into the stdlib

tfoleyNV:
The only big catch that I ran into with this batch was that I found the `float.getPi()` function was being emitted to the output GLSL even when that function wasn't being used. This seems to have been a latent problem in the earlier PR, but was only surfaced in the tests once a Slang->GLSL test started using another intrinsic that led to the `float : __BuiltinFloatingPointType` witness table being live in the IR.

The fix for the gotcha here was to add a late IR pass that basically empties out all witness tables in the IR, so that functions that are only referenced by witness tables can then be removed as dead code. This pass is something we should *not* apply if/when we start supporting real dynamic dispatch through witness tables, but that is a problem to be solved on another day.

The remaining tricky pieces of this change were:

* Needed to remember to mark functions as target intrinsics on HLSL and/or GLSL as appropriate (hopefully I caught all the cases) so they don't get emitted as source there.

* The `msad4` function in HLSL is very poorly documented, so filling in its definition was tricky. I made my best effort based on how it is described on MSDN, but it is likely that if anybody wants to rely on this function they will need us to vet our results with some tests.


tfoleyNV:
It would have been clear to you when you reviewed, but for posterity I'll note that this PR only adds the stdlib declarations for these functions (which seem to be automatically getting picked up by the CPU and CUDA back-ends), but doesn't eliminate the support code for synthesizing the functions on those back-ends. That should be able to come as a pure cleanup PR to follow.

As discussed elsewhere, the next small-ish step would then be to eliminate the need to synthesize scalar versions of functions where we now auto-generate vector/matrix versions, and then the final push would be finding a solution for stdlib functions that currently map to single instructions (mostly these are operators today).

PR #1266, opened 2020-03-09
CUDA support for vector/matrix Wave intrinsics

jsmall-nvidia:
* Support vector and matrix types for Wave intrinsics
* Fix parsing issue for CUDA compiler output
* Better documentation of convergence issues
* Removed mask parameter from _wave functions in CUDA prelude

### DIFF
--- a/prelude/slang-cuda-prelude.h
+++ b/prelude/slang-cuda-prelude.h
@@ -441,6 +441,33 @@ __forceinline__ __device__ uint32_t _getLaneId()
 }
 #endif
 
+__forceinline__ __device__ int _getConvergedMask()
+{
+    // It appears that the active mask is not good enough because on later hardware threads can diverge.
+    // 
+    // For CUDA the article claims mask has to be used carefully
+    // https://devblogs.nvidia.com/using-cuda-warp-level-primitives/
+    // With the Warp intrinsics there is no mask, and it's just the 'active lanes'. 
+    // __activemask() though does not require there is convergence, so that doesn't work.
+    // 
+    // '__ballot_sync' produces a convergance. 
+    // 
+    // From the CUDA docs:
+    // For __all_sync, __any_sync, and __ballot_sync, a mask must be passed that specifies the threads 
+    // participating in the call. A bit, representing the thread's lane ID, must be set for each participating thread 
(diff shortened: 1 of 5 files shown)


prelude/slang-cuda-prelude.h
```diff
@@ -538,24 +564,63 @@ struct WaveOpMin
...
+// Scalar
+template <> struct ElementTypeTrait<int> { typedef int Type; };
+template <> struct ElementTypeTrait<uint> { typedef uint Type; };
+template <> struct ElementTypeTrait<float> { typedef float Type; };
+template <> struct ElementTypeTrait<double> { typedef double Type; };
+template <> struct ElementTypeTrait<uint64_t> { typedef uint64_t Type; };
+template <> struct ElementTypeTrait<int64_t> { typedef int64_t Type; };
+
+// Vector
+template <> struct ElementTypeTrait<int1> { typedef int Type; };
```
tfoleyNV: I was going to ask why the vector case can't be handled via a partial specialization, but then I remembered that the vector types are built-ins for CUDA, while we have to implement matrices in user space.

source/slang/hlsl.meta.slang
```diff
@@ -2228,123 +2228,134 @@ __generic<T : __BuiltinIntegerType>
 __glsl_extension(GL_KHR_shader_subgroup_arithmetic)
 __spirv_version(1.3)
 __target_intrinsic(glsl, "subgroupAnd($0)")
-__target_intrinsic(cuda, "_waveAnd(__activemask(), $0)")
+__target_intrinsic(cuda, "_waveAnd($0)")
```
tfoleyNV: Now that these functions are just 1-to-1 parallels to the HLSL builtin, we could simplify these `__target_intrinsic` lines in two ways:

1. Any case where you'd just write `__target_intrinsic(cuda, "someFunc($0)")` you can just write `__target_intrinsic(cuda, "someFunc")`

2. If we rename the `_waveWhatever` functions in the CUDA prelude to exactly match the HLSL names, then we can just write `__target_intrinsic(cuda)`. On a function named `WaveFoo` this should be equivalent to writing `__targert_intrinsic(cuda, "WaveFoo")` or `__target_intrinsic(cuda, "WaveFoo($0)")`.

PR #1270, opened 2020-03-10
Wave Prefix Product

jsmall-nvidia:
Most of this work is as it relates to the CUDA target. 

* Template mechanisms for doing prefixes
  * Needed to add doInverse to the op interface
* Implement Sum/Product/Xor etc using the mechanism
* Split out SM6.5 Wave intrinsics
* Added wave-prefix-product.slang test


prelude/slang-cuda-prelude.h
```diff
@@ -823,46 +828,121 @@ __inline__ __device__ T _waveReadLaneAtMultiple(T inVal, int lane)
     return outVal;
 }
 
-__device__ int _wavePrefixSum(int val)
+// Scalar 
+
+// Invertable means that when we get to the end of the reduce, we can remove val (to make exclusive), using 
+// the inverse of the op.
```
tfoleyNV: Inverting the operation to back out the value for the current lane seems really undesirable, and could be incorrect in many cases. What if you are doing a product and the value is zero? What if you are doing a sum of floating-point values and the value is not finite (or *is* finite but causes the final result to be non-finite).

It seems like we need an approach that works for any "exclusive" op, including ones where invertability isn't really a thing, and at that point we should probably use that approach for all the ops (even ones where intertability might be usable).

prelude/slang-cuda-prelude.h
```diff
@@ -823,46 +828,121 @@ __inline__ __device__ T _waveReadLaneAtMultiple(T inVal, int lane)
...
-        return result;
     }
+    return result;
 }
+    
+template <typename T>
+__inline__ __device__ T _wavePrefixProduct(T val) { return _wavePrefixScalar<WaveOpMul<T>, T>(val); }
+
+template <typename T>
+__inline__ __device__ T _wavePrefixSum(T val) { return _wavePrefixInvertableScalar<WaveOpAdd<T>, T>(val); }
```
tfoleyNV: Okay, it looks like you are careful not to rely on the "invertability" in most cases where it would be a problem, but the sum case is still technically an issue for floating-point inputs, both because of non-finite inputs and because of precision issues.

PR #1273, opened 2020-03-12
Define compound intrinsic ops in the standard library

tfoleyNV:
The current stdlib code has a notion of "compound" intrinsic ops, which use the `__intrinsic_op` modifier but don't actually map to a single IR instruction.
Instead, most* of these map to multiple IR instructions using hard-coded logic in `slang-ir-lower.cpp`.

(* One special case is `kCompoundOp_Pos` that is used for unary `operator+` and that maps to *zero* IR instructions)

All of the opcodes that used to use the `kCompoundOp_` enumeration values now have definitions directly in the stdlib and use the new `[__unsafeForceInlineEarly]` attribute to ensure that they get inlined into their use sites so that the output code is as close as possible to the original.

For the most part, generating the stdlib definitions for the compound ops is straightforward, but here's some notes:

* The unary `operator+` I just defined directly in Slang code, since it doesn't share much structure with other cases

* The unary increment/decrement ops are generated as a cross product of increment/decrement and prefix/postfix. The logic is a bit messy but given that we have scalar, vector, and matrix versions to deal with it still saves code overall

* Because all the compound/assignment cases got moved out, the existing code for generating unary/binary ops can be simplified a bit

* All the no-op bit-cast operations like `asfloat(float)` are now inline identity functions

* A few other small cleanups are made by not having to worry about the compound ops (which used to be called "pseudo ops") sometimes being encoded in to the same type of value as a real IR opcode.

The one big detail here is a fix for how IR lowering works for `let` declarations: they were previously being `materialize()`d which only guarantees that they've been placed in a contiguous and addressable location, but doesn't actually convert them to an r-value. As a result a `let` declaration could accidentally capture a mutable location by reference, which is definitely *not* what we wanted it to do. Fixing this was needed to make the new postfix `++` definition work (several existing tests end up covering this).

One important forward-looking note:

One subtle (but significant) choice in this change is that we actually reduce the number of declarations in the stdlib, because instead of having the compound operators include both per-type and generic overloads (just listing scalar cases for now):

    float operator+=(in out float left, float right) { ... }
    int operator+=(in out int left, int right) { ... }
    ...
    T operator+= <T:__BuiltinBlahBlah>(in out T left, T right) { ... }

We now have *only* the single generic version:

    T operator+= <T:__BuiltinBlahBlah>(in out T left, T right) { ... }

In running our current tests, this change didn't lead to any regressions (perhaps surprisingly).

Given that we were able to reduce the number of overloads for `operator+=` by a factor of N (where N is the number of built-in types), it seems worth considering whether we could also reduce the number of overloads of `operator+` by the same factor by only having generic rather than per-type versions.

One concern that this forward-looking question raises is whether the quality of diagnostic messages around bad calls to the operators might suffer when there are only generic overloads instead of per-type overloads. In order to feel out this problem I added a test case that includes some bad operator calls both to `+=` (which is now only generic with this change) and `+` (which still has per-type overloads). Overall, I found the quality of the error messages (in terms of the candidates that get listed) isn't perfect for either, but personally I prefer the output in the generic case.

As part of adding that test, I also added some fixups to how overload resolution messages get printed, to make sure the function name is printed in more cases, and also that the candidates print more consistently. These changes affected the expected output for one other diagnostic test.


tfoleyNV:
Ugh. The build failure on Travis here is because the baseline for the diagnostic tests is formatted like:

```
core.meta.slang(1740): note 39999: candidate: func +=<T, R:int, C:int>(matrix<T,R,C>, T) -> matrix<T,R,C>
```

while the output that gets produced in practice on Travis (possibly just the way our Linux build is configured) is:

```
../../../source/slang/core.meta.slang(1740): note 39999: candidate: func +=<T, R:int, C:int>(matrix<T,R,C>, T) -> matrix<T,R,C>
```

The most expedient "fix" right now would be to disable the test I just added (or at least disable it on Linux). A longer-term solution would be track down why the Linux build seems to be getting a different path to `core.meta.slang` in the generated stdlib files, and fix the build so that it prints the same kinds of diagnostics as Windows.

source/slang/core.meta.slang
```diff
@@ -1490,27 +1484,85 @@ for (auto op : unaryOps)
...
-        sb << "__intrinsic_op(" << int(op.opCode) << ") matrix<" << resultType << ",N,M> operator" << op.opName << "(" << qual << "matrix<T,N,M> value);\n";
+        sb << "__prefix __intrinsic_op(" << int(op.opCode) << ") matrix<" << resultType << ",N,M> operator" << op.opName << "(matrix<T,N,M> value);\n";
     }
 }
 
-for (auto op : binaryOps)
+}}}}
+
+__generic<T : __BuiltinArithmeticType>
+[__unsafeForceInlineEarly]
```
jsmall-nvidia: It's a small point but I'm now wondering if 'unsafe' seems to be gaining weight from it's meaning in rust, but that is not what is meant here.

tfoleyNV: Yes, we aren't using `unsafe` in a way that is consistent with the Haskell/Rust/Swift precedent, and we should maybe think about that.

For now I'm not too worried because this isn't a feature we ever plan to expose to users with its current name.

PR #1275, opened 2020-03-16
First pass support of WaveMultiPrefix SM6.5 intrinsics on CUDA

jsmall-nvidia:
* Made the prefix CUDA templates take a mask parameter, so can use with Multi style and just regular 'Prefix'
* Prefixed with _waveOp the array helpers for some of the multi functions
* Added cs_6_4/cs_6_5 to profiles
* Added WaveMatch intrinsics on CUDA
* Added wave-multi-prefix.slang test

Note that currently test is only performed on CUDA, because the version of dxc being used (probably the version installed in current version of win 10) does not support SM 6.5.


source/slang/slang-profile-defs.h
```diff
@@ -129,6 +129,8 @@ PROFILE(DX_Compute_6_0,				cs_6_0,				Compute,	DX_6_0)
 PROFILE(DX_Compute_6_1,				cs_6_1,				Compute,	DX_6_1)
 PROFILE(DX_Compute_6_2,				cs_6_2,				Compute,	DX_6_2)
 PROFILE(DX_Compute_6_3,				cs_6_3,				Compute,	DX_6_3)
+PROFILE(DX_Compute_6_4,				cs_6_4,				Compute,	DX_6_4)
+PROFILE(DX_Compute_6_5,				cs_6_5,				Compute,	DX_6_5)
```
tfoleyNV: We eventually need to make a decision about whether we continue to define stage-specific `-profile` names for each new shader model (creating an O(N*M) problem) or if we eventually keep things separate and only support using separate `-stage compute -profile sm_6_5`.

The only reason to support these as their own profiles is for compatibility with users who work with dxc, so I guess we are stuck following that precedent for now.

PR #1279, opened 2020-03-16
First pass at a language reference

tfoleyNV:
We already had the `language-guide.md` document under `docs/`, but this is an attempt to introduce a more full-featured reference to the Slang language and its features.

Right now it is mostly focused on the syntax and what the language allows to be declared, and it is a little light on semantic details throughout (mostly relying on familiarity with C to explain the things that are left unsaid). Even so, this hopefully provides a starting point to continue adding more detail.

### DIFF
--- a/docs/language-reference/02-lexcial-structure.md
+++ b/docs/language-reference/02-lexcial-structure.md
@@ -29,12 +29,17 @@ An _escaped line break_ comprises a backslack (`\`, U+005C) follow immediately b
 Comments
 --------
 
-A _comment_ is either a line comment or a block comment.
+A _comment_ is either a line comment or a block comment:
+
+```hlsl
+// a line comment
+/* a block comment */
+```
 
 A _line comment_ comprises two forward slashes (`/`, U+002F) followed by zero or more characters that do not contain a line break.
 A line comment extends up to, but does not include, a subsequent line break or the end of the source unit.
 
-A _block comment_ begins with a forward slahs (`/`, U+002F) followed by an asterisk (`*`, U+0052). 
+A _block comment_ begins with a forward slash (`/`, U+002F) followed by an asterisk (`*`, U+0052). 
 A block comment is terminated by the next instance of an asterisk followed by a forward slash (`*/`).
 A block comment contains all characters between where it begins and where it terminates, including any line breaks.
 Block comments do not nest.
@@ -60,7 +65,7 @@ Compilation of a source unit proceeds _as if_ the following steps are executed i
 Identifiers
 -----------
 
-An _identifier_ betgins with an upercase or lowercase ASCII leter (`A` through `Z`, `a` through `z`), or an underscore (`_`).
+An _identifier_ begins with an uppercase or lowercase ASCII letter (`A` through `Z`, `a` through `z`), or an underscore (`_`).
 After the first character, ASCII digits (`0` through `9`) may also be used in an identifier.
 
 The identifier consistent of a single underscore (`_`) is reserved by the language and must not be used by programs.
--- a/docs/language-reference/02-lexcial-structure.md
+++ b/docs/language-reference/02-lexcial-structure.md
@@ -1,0 +1,114 @@
+Lexical Structure
+=================
+
+Source Units
+------------
+
+A _source unit_ comprises a sequence of zero or more _characters_ which for purposes of this document are defined as Unicode scalars (code points).
+Implementations *may* accept source units stored as files on disk, buffers in memory, or any appropriate implementation-specified means.
+
+Encoding
+--------
+
+When encoding is required, source units *should* be encoded using UTF-8.
+Implementations *may* support additional implemented-specified encodings.
+
+Whitespace
+----------
+
+_Horizontal whitespace_ consists of space (U+0020) and horizontal tab (U+0009).
+
+A _line break_ consists of a line feed (U+000A), carriage return (U+000D) or a carriage return followed by a line feed (U+000D, U+000A).
+Line breaks are used as line separators rather than terminators; it is not necessary for a source unit to end with a line break.
+
+Escaped Line Breaks
+-------------------
+
+An _escaped line break_ comprises a backslack (`\`, U+005C) follow immediately by a line break.
+
+Comments
+--------
+
+A _comment_ is either a line comment or a block comment.
+
+A _line comment_ comprises two forward slashes (`/`, U+002F) followed by zero or more characters that do not contain a line break.
+A line comment extends up to, but does not include, a subsequent line break or the end of the source unit.
+
+A _block comment_ begins with a forward slahs (`/`, U+002F) followed by an asterisk (`*`, U+0052). 
+A block comment is terminated by the next instance of an asterisk followed by a forward slash (`*/`).
+A block comment contains all characters between where it begins and where it terminates, including any line breaks.
+Block comments do not nest.
+It is an error if a block comment that begins in a source unit is not terminated in that source unit.
+
+Phases
+------
+
+Compilation of a source unit proceeds _as if_ the following steps are executed in order:
+
+1. Line numbering (for subsequent diagnostic messages) is noted based on the locations of line breaks
+
+2. Escaped line breaks are eliminated. No new characters are inserted to replace them. Any new escaped line breaks introduced by this step are not eliminated.
+
+3. All comments are replaced with a single space (U+0020)
+
+4. The source unit is _lexed_ into a sequence of tokens according the lexical grammar in this chapter
+
+5. The lexed sequence of tokens is _preprocessed_ to produce a new sequence of tokens (Chapter 3)
+
+6. Subsequent processing is performed on the preprocessed sequence of tokens
+
+Identifiers
+-----------
+
+An _identifier_ betgins with an upercase or lowercase ASCII leter (`A` through `Z`, `a` through `z`), or an underscore (`_`).
+After the first character, ASCII digits (`0` through `9`) may also be used in an identifier.
+
+The identifier consistent of a single underscore (`_`) is reserved by the language and must not be used by programs.
+Otherwise, there are no fixed keywords or reserved words.
+Words that name a built-in language construct can also be used as user-defined identifiers and will shadow the built-in definitions in the scope of their definition.
+
+Literals
+--------
+
+### Integer Literals
+
+An _integer literal_ consists of an optional radix specifier followed by digits and an optional suffix.
+
+The _radix specifier_ may be:
+
+* `0x` or `0X` to specify a hexadecimal literal (radix 16)
+* `0b` or `0B` to specify a binary literal (radix 2)
+
+When no radix specifier is present a radix of 10 is used.
+
+Octal literals (radix 8) are not supported.
+A `0` prefix on an integer literal does *not* specify an octal literal as it does in C.
+Implementations *may* warn on integer literals with a `0` prefix in case users expect C behavior.
+
+The _digits_ of an integer literal may include ASCII `0` through `9`.
+In the case of a hexadecimal literal, digits may include the letters `A` through `F` (and `a` through `f`) which represent digit values of 10 through 15.
+It is an error for an integer literal to include a digit with a value greater than or equal to the radix.
+The digits of an integer literal may also include underscore (`_`) characters, which are ignored and have no semantic impact.
+
+The _suffix_ on an integer literal may be used to indicate the desired type of the literal:
+
+* A `u` suffix indicates the `uint` type
+* An `l` or `ll` suffix indicates the `int64_t` type
+* A `ul` or `ull` suffix indicates the `uint64_t` type
+
+### Floating-Point Literals
+
+> Note: This section is not yet complete.
+
+### String Literals
+
+> Note: This section is not yet complete.
+
+### Character Literals
+
+> Note: This section is not yet complete.
+
+Operators and Punctuation
+-------------------------
+
+> Note: This section is not yet complete.
--- a/docs/language-reference/04-types.md
+++ b/docs/language-reference/04-types.md
@@ -72,7 +72,18 @@ A vector type is written as `vector<T, N>` and represents an `N`-element vector 
 The _element type_ `T` must be one of the built-in scalar types, and the _element count_ `N` must be a specialization-time constant integer.
 The element count must be between 2 and 4, inclusive.
 
-A vector type allows subscripting of its elements like an array, but also supports component-wise arithmetic on its elements.
+A vector type allows subscripting of its elements like an array, but also supports element-wise arithmetic on its elements.
+_Element-wise arithmetic_ means mapping unary and binary operators over the elements of a vector to produce a vector of results:
+
+```hlsl
+vector<int,4> a = { 1, 2, 30, 40 };
+vector<int,4> b = { 10, 20, 3, 4 };
+
+-a; // yields { -1, -2, -30, -40 }
+a + b; // yields { 11, 22, 33, 44 }
+b / a; // yields { 10, 10, 0, 0 }
+a > b; // yields { false, false, true, true }
+```
 
 A vector type is laid out in memory as `N` contiguous values of type `T` with no padding.
 The alignment of a vector type may vary by target platforms.
@@ -111,7 +122,7 @@ The element type `T` must be one of the built-in scalar types.
 The _row count_ `R` and _column count_ `C` must be specialization-time constant integers.
 The row count and column count must each be between 2 and 4, respectively.
 
-A matrix type allwos subscripting of its rows, similar to an `R`-element array of `vector<T,C>` elements.
+A matrix type allows subscripting of its rows, similar to an `R`-element array of `vector<T,C>` elements.
 A matrix type also supports element-wise arithmetic.
 
 Matrix types support both _row-major_ and _column-major_ memory layout.
@@ -125,7 +136,7 @@ Under column-major layout, a matrix is laid out in memory equivalent to the row-
 This means it will be laid out equivalently to a `C`-element array of `vector<T,R>` elements.
 
 As a convenience, Slang defines built-in type aliases for matrices of the built-in scalar types.
-E.e.g, declarations equivalent to the following are provied by the Slang standard library:
+E.g., declarations equivalent to the following are provided by the Slang standard library:
 
 ```hlsl
 typealias float3x4 = matrix<float, 3, 4>;
@@ -211,6 +222,8 @@ This type represents an array of exactly `N` values of type `T`.
 
 An unknown-size array type is written `T[]` where `T` is a type.
 This type represents an array of some fixed, but statically unknown, size.
+
+> Note: Unlike in C and C++, arrays in Slang are always value types, meaning that assignment and parameter passing of arrays copies their elements.
 
 ### Declaration Syntax
 
@@ -314,10 +327,10 @@ Types with unknown size arise in a few ways:
 
 * A structure type has unknown size if any field type has unknown size
 
-The use of tyeps with unknown size is restricted as follows:
+The use of types with unknown size is restricted as follows:
 
 * A type with unknown size cannot be used as the element type of an array
 
 * A type with unknown size can only be used as the last field of a structure type
 
-* A type with unknown size cannot be used as a generic argument to specialize a type, function, etc.
+* A type with unknown size cannot be used as a generic argument to specialize a user-defined type, function, etc. Specific built-in generic types/functions may support unknown-size types, and this will be documented on the specific type/function.
--- a/docs/language-reference/04-types.md
+++ b/docs/language-reference/04-types.md
@@ -1,0 +1,323 @@
+Types
+=====
+
+This section defines the kinds of types supported by Slang.
+
+Types in Slang do not necessarily prescribe a single _layout_ in memory.
+The discussion of each type will specify any guarantees about layout it provides; any details of layout not specified here may depend on the target platform, compiler options, and context in which a type is used.
+
+Void Type
+---------
+
+The type `void` contains no data and has a single, unnamed, value.
+
+A `void` value takes up no space, and thus does not affect the layout of types.
+Formally, a `void` value behaves as if it has a size of zero bytes, and one-byte alignment.
+
+Scalar Types
+------------
+
+### Boolean Type
+
+The type `bool` is used to represent Boolean truth value: `true` and `false`.
+
+The size of a `bool` varies across target platforms; programs that need to ensure a matching in-memory layout between targets should not use `bool` for in-memory data structures.
+On all platforms, the `bool` type must be _naturally aligned_ (its alignment is its size).
+
+### Integer Types
+
+The following integer types are defines:
+
+| Name          | Description |
+|---------------|-------------|
+| `int8_t`      | 8-bit signed integer |
+| `int16_t`     | 16-bit signed integer |
+| `int`         | 32-bit signed integer |
+| `int64_t`     | 64-bit signed integer |
+| `uint8_t`     | 8-bit unsigned integer |
+| `uint16_t`    | 16-bit unsigned integer |
+| `uint`        | 32-bit unsigned integer |
+| `uint64_t`    | 64-bit unsigned integer |
+
+All signed integers used two's complement representation.
+All arithmetic operations on integers (both signed and unsigned) wrap on overflow/underflow.
+
+All target platforms must support the `int` and `uint` types.
+Specific target platforms may not support the other integer types.
+
+All integer types are stored in memory with their natural size and alignment on all targes that support them.
+
+### Floating-Point Types
+
+The following floating-point type are defined:
+
+| Name          | Description                   |
+|---------------|-------------------------------|
+| `half`        | 16-bit floating-point number (1 sign bit, 5 exponent bits, 10 fraction bits) |
+| `float`       | 32-bit floating-point number (1 sign bit, 8 exponent bits, 23 fraction bits) |
+| `double`      | 64-bit floating-point number (1 sign bit, 11 exponent bits, 52 fraction bits) |
+
+All floating-point types are laid out in memory using the matching IEEE 754 standard format (`binary16`, `binary32`, `binary64`).
+Target platforms may define their own rules for rounding, precision, denormals, infinities, and not-a-number values.
+
+All target platforms must support the `float` type.
+Specific targets may not support the other floating-point types.
+
+All floating-point types are stored in memory with their natural size and alignment on all targets that support them.
+
+Vector Types
+------------
+
+A vector type is written as `vector<T, N>` and represents an `N`-element vector with elements of type `T`.
+The _element type_ `T` must be one of the built-in scalar types, and the _element count_ `N` must be a specialization-time constant integer.
+The element count must be between 2 and 4, inclusive.
+
+A vector type allows subscripting of its elements like an array, but also supports component-wise arithmetic on its elements.
+
+A vector type is laid out in memory as `N` contiguous values of type `T` with no padding.
+The alignment of a vector type may vary by target platforms.
+The alignment of `vector<T,N>` will be at least the alignment of `T` and may be at most `N` times the alignment of `T`.
+
+As a convenience, Slang defines built-in type aliases for vectors of the built-in scalar types.
+E.g., declarations equivalent to the following are provided by the Slang standard library:
+
+```hlsl
+typealias float4 = vector<float, 4>;
+typealias int8_t3 = vector<int8_t, 3>;
+```
+
+### Legacy Syntax
+
+For compatibility with older codebases, the generic `vector` type includes default values for `T` and `N`, being declared as:
+
+```hlsl
+struct vector<T = float, let N : int = 4> { ... }
+```
+
+This means that the bare name `vector` may be used as a type equivalent to `float4`:
+
+```hlsl
+// All of these variables have the same type
+vector a;
+float4 b;
+vector<float, 4> c;
+```
+
+Matrix Types
+------------
+
+A matrix type is written as `matrix<T, R, C>` and represents a matrix of `R` rows and `C` columns, with elements of type `T`.
+The element type `T` must be one of the built-in scalar types.
+The _row count_ `R` and _column count_ `C` must be specialization-time constant integers.
+The row count and column count must each be between 2 and 4, respectively.
+
+A matrix type allwos subscripting of its rows, similar to an `R`-element array of `vector<T,C>` elements.
+A matrix type also supports element-wise arithmetic.
+
+Matrix types support both _row-major_ and _column-major_ memory layout.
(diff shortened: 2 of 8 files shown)


docs/language-reference/06-statements.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Break Statement
+
+A `break` statement transfers control to after the end of the closest lexically enclosing switch statement or loop statement:
+
+```hlsl
+break;
+```
+
+### Continute Statement
```
jsmall-nvidia: Continute

docs/language-reference/06-statements.md
```diff
@@ -0,0 +1,235 @@
...
+### Discard Statement
+
+A `discard` statement can only be used in the context of a fragment shader, in which case it causes the current invocation to terminate and the graphics system to discard the corresponding fragment so that it does not get combined with the framebuffer pixel at its coordintes.
+
+Operations with side effects that were executed by the invocation before a `discard` will still be performed and their results will become visible according to the rules of the platform.
+
+Compile-Time For Statement
+--------------------------
+
+A _compile-time for statement_ is used as an alternative to preprocessor techniques for loop unrolling.
```
jsmall-nvidia: Does this need to be marked as a 'may in the future be removed' feature?

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
+Declarations
+============
+
+Modules
+-------
+
+A module consists of one or more source units that are compiled together.
+The global declarations in those source units comprise the body of the module.
+
+In general, the order of declarations within a source unit does not matter; declarations can refer to other decalrations (of types, functions, variables, etc.) later in the same source unit.
```
jsmall-nvidia: Except for our current parser disambiguation via knowing a kind.

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+```hlsl
+let x = 7;
+var y = 9.0;
+```
+
+A `let` declaration introduces an immutable variable, which may not be assigned to or used as the argument for an `in out` or `out` parameter.
+A `var` declaration introduces a mutable variable.
+
+An explicit type may be given for a variable by placing it afte the variable name and a colon (`:`):
```
jsmall-nvidia: afte

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+A variable declared at function scoep and marked with `static` (but not `const`) is a _function-scope static variable_.
+Semantically, a function-scope static variable behaves like a global static variable except that its name is only visible in the local scope.
+
+The initial-value expression for a function-scope static variable may refer to non-static variables in the body of the function.
+In these cases initialization of the variable is guaranteed not to occur until at least the first time the function body is evaluated for a given invocation.
+
+#### Local Variables
+
+A variable declared at function scope and not marke with `static` (even if marked with `const`) is a _local variable_.
```
jsmall-nvidia: marke

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+```hlsl
+func add(x: in out int, y : float) { x += ... }
+```
+
+The available directions are:
+
+* `in` (the default) indicates typical pass-by-value (copy-in) semantics. The callee receives a *copy* of the argument passed by the caller.
+
+* `out` indicates copy-out semantics. The callee writes to the parameter and then a copy of that value is assigned to the argument of the caller after the call returns.
```
jsmall-nvidia: Might need some clarity as to how that differs from say C++.

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+```
+
+> Note: Slang does not require "forward declaration" of functions, although
+> forward declarations are supported as a compatibility feature.
+>
+> The only place where a function declaration without a definition should be
+> required is in the body of an `interface` declaration.
+
+
+The result type of a function mayb be specified after the parameter list using a _result type clause_ consisting of an arrow (`->`) followed by a type.
```
jsmall-nvidia: mayb

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+```hlsl
+// A terminating semicolon is allowed
+struct Stuff { ... };
+
+// The semicolon is not required
+struct Things { ... }
+```
+
+When a structure deeclarations ends without a semicolon, the closing curly brace (`}`) must be the last non-comment, non-whitespace token on its line.
```
jsmall-nvidia: deeclarations

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+    Red,
+    Green = 3,
+    Blue,
+}
+```
+
+### Cases
+
+The body of an enumeration type declaration consists of a comma-separated list of case declarations.
+An optional trailing comma may terminate the lis of cases.
```
jsmall-nvidia: lis

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+The body of a getter can access the parameters of the enclosing subscript, as a well as an implicit `this` parameter of the type that encloses the accessor.
+The `this` parameter of a getter is immutable; `[mutating]` getters are not currently supported.
+
+Extensions
+----------
+
+An extension declaration is introduced with the `extension` keyword:
+
+```hlsl
+extension MyVector
```
jsmall-nvidia: You might want to define MyVector in this example - to make clear what is going on here.

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+```
+
+An extension declaration adds behavior to an existing type.
+In the example above, the `MyVector` type is extended with an instance method `getLength()`, and a static method `getDimensionality()`.
+
+An extension declaration names the type being extended after the `extension` keyword.
+The body of an extension declaration may include type declarations, functions, initializers, and subscripts.
+
+> Note: The body of an extension may *not* include variable declarations.
+> An extension cannot introduce members that would change teh in-memory layout of the type being extended.
```
jsmall-nvidia: teh

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+The members of an extension are access through the type that is being extended.
```
jsmall-nvidia: access

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+    ...
+}
+```
+
+The inheritance clause of an extension declaration may only include interfaces.
+When an extension declaration lists an interface in its inheritance clause, it asserts that the extension introduces a new conformance, such that the type being extended now conforms to the given interface.
+The extension must ensure that the type being extended satisfies all the requirements of the interface.
+Interface requirements may be satisfied by the members of the extension, members of the original type, or members introduced through other extensions visible at the point where the conformance was declared.
+
+I is an error for overlapping conformances (that is, of the same type to the same interface) to be visible at the same point.
```
jsmall-nvidia: It?

docs/language-reference/07-declarations.md
```diff
@@ -0,0 +1,771 @@
...
+
+> Note: The syntax for generic value parameters is provisional and subject to possible change in the future.
+
+Generic parameters may declare a default value with `=`:
+
+```hlsl
+T anotherFunction<T = float, let N : int = 4>(vector<T,N> v);
+```
+
+For generic type parametters, the default value is a type to use if no argument is specified.
```
jsmall-nvidia: parametters

docs/language-reference/02-lexcial-structure.md
```diff
@@ -0,0 +1,114 @@
...
+Escaped Line Breaks
+-------------------
+
+An _escaped line break_ comprises a backslack (`\`, U+005C) follow immediately by a line break.
+
+Comments
+--------
+
+A _comment_ is either a line comment or a block comment.
+
```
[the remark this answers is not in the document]
tfoleyNV: Yes, more examples in these earlier sections would help, and in turn some of the later sections would benefit from some more specificity.

We should probably eventually have EBNF grammars for both the lexical grammar and syntax rules.

docs/language-reference/02-lexcial-structure.md
```diff
@@ -0,0 +1,114 @@
...
+6. Subsequent processing is performed on the preprocessed sequence of tokens
+
+Identifiers
+-----------
+
+An _identifier_ betgins with an upercase or lowercase ASCII leter (`A` through `Z`, `a` through `z`), or an underscore (`_`).
+After the first character, ASCII digits (`0` through `9`) may also be used in an identifier.
+
+The identifier consistent of a single underscore (`_`) is reserved by the language and must not be used by programs.
```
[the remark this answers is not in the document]
tfoleyNV: One note (that I think will come up again later, based on my quick scan of your comments) is that there is an important difference between what the Slang front-end accepts and whatever constraints the different downstream compilers place.

The actual comment about a single underscore being reserved is currently just a goal, but I'd like to have documentation warn people that we might snag that identifier later.

No matter what we allow in identifiers (we could even allow non-ASCII code points if we felt like it...) we should be able to "scrub" them to something that a downstream compiler can accept. We already do this in cases where a "name hint" might include a `.` (which we turn into a `_`), and for GLSL we already deal with cases of identifiers with multiple underscores in a row (so that `__` turns into `_U_`, and `___` into `_U1_` or something like that, IIRC).

I don't think a reference for the Slang *language* should include those kinds of details, although it would be good if we had a per-back-end document like what you've added for CPU and CUDA to describe those kinds of details.

docs/language-reference/02-lexcial-structure.md
```diff
@@ -0,0 +1,114 @@
...
+Otherwise, there are no fixed keywords or reserved words.
+Words that name a built-in language construct can also be used as user-defined identifiers and will shadow the built-in definitions in the scope of their definition.
```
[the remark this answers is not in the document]
tfoleyNV: Well, for right now this is a bit loose, but the short answer is that *yes* you can shadow a lot of the built-in language keywords.

From a mechanism standpoint: when the parser is trying to parse an expression, statement, or declaration and the lookahead token is an identifier, it attempts to look up that identifier in the current environment. If the lookup step finds a `SyntaxDecl` then it knows the identifier mapped to some kind of built-in syntax, and it will parse using the callback on that `SyntaxDecl`. If something in a more local scope shadows the same identifier, then lookup will find that declaration instead, and the parser won't parse using the `SyntaxDecl`.

There are a few pieces of rationale behind this design:

1. It is 100% desirable to have many keywords be usable as ordinary identifiers. This is both because HLSL has keywords with names like `triangle` and `sample` that are very likely to be things people want to use as local variables, *and* because Slang adds its own keywords (`extension`, `import`, etc.) that would make it impossible to be backwards-compatible with HLSL if users couldn't use those same identifiers to name functions/variables/etc.

2. It is super easy to work with and extend. Adding parsing logic a new language construct with a new keyword mostly amounts to adding a subroutine to do the parsing, plus registering its name/keyword at one place in the parser initialization. In comparison, approaches that recognize keywords during tokenization tend to require a lot more steps when adding a keyword.

3. Because syntax can be introduced as an ordinary kind of declaration, many of our keywords are just defined in the stdlib by naming the modifier keyword and the C++ AST node class that it should yield.

4. This flexibility can/will be required if we ever wanted to introduce a first-class hygienic macro system that allows user to introduce their own syntactic keywords.

I completely agree that this design choice could lead to confusion if people abuse it:

```hlsl
bool myFunc(bool true, bool false)
{
    // ... many lines later

    return true; // returns first parameter which might not be true
}

bool b = myFunc(false, true); // should return false
```

I can argue that people can always write confusing code when given flexible language rules (e.g., overloading has often been vilified for similar reasons), so this isn't *that* bad.

Still, the current realization of this feature has a lot of gotchas, like:

* The lookup rules that end up being used during parsing don't match those used for semantic checks. Perhaps most importantly, they don't work correctly for forward references (so if `true` is defined as a global variable later in your file, a reference to `true` earlier in the file won't necessarily see it).

* There are still some ad hoc rules in the parser that look for exact strings instead of always looking up `SyntaxDecl`s and these need to be cleaned up sooner or later (they are a maintenance headache).

I'd definitely be willing to consider changes that make some of our built-in syntax act more like keywords, perhaps by issuing an error on any attempt to use them as an ordinary identifier. We probably need to leave many of them as contextual keywords that can still be used to name functions/variables/etc.

I still maintain a dim hope of some day having a hygienic macro system in Slang, although I will freely admit that the longer we've been working on the language, the less demand I'm seeing for anything like that.

docs/language-reference/02-lexcial-structure.md
```diff
@@ -0,0 +1,114 @@
...
+
+When no radix specifier is present a radix of 10 is used.
+
+Octal literals (radix 8) are not supported.
+A `0` prefix on an integer literal does *not* specify an octal literal as it does in C.
+Implementations *may* warn on integer literals with a `0` prefix in case users expect C behavior.
+
+The _digits_ of an integer literal may include ASCII `0` through `9`.
+In the case of a hexadecimal literal, digits may include the letters `A` through `F` (and `a` through `f`) which represent digit values of 10 through 15.
+It is an error for an integer literal to include a digit with a value greater than or equal to the radix.
```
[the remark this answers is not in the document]
tfoleyNV: Yeah, in practice if we have `10a` as a radix-ten literal then we interpret `a` as a suffix on that literal (which in this case would be an unknown/unsupported suffix). The wording here could be clarified, I suppose.

docs/language-reference/02-lexcial-structure.md
```diff
@@ -0,0 +1,114 @@
...
+The digits of an integer literal may also include underscore (`_`) characters, which are ignored and have no semantic impact.
+
+The _suffix_ on an integer literal may be used to indicate the desired type of the literal:
+
+* A `u` suffix indicates the `uint` type
```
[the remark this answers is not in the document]
tfoleyNV: For me, the ideal answer is that we introduce a change so that numeric literals are untyped (or more accurately "multi-typed"/overloaded) by default and only resolve to a specific type based on context.

Under that model, anything that allows the desired type to be inferred from context would work:

```hlsl
half h = 1.23; // clearly a `half` literal is desired

float f = ...;
float g = f + 1.23; // clearly a `float` literal is desired
```

Basically, whenever a literal is being coerced/converted, we can decide that the type of the literal should be whatever type would minimize the conversion cost.

If we implement those rules then something that would just "fall out" of the implementation is that using a literal as the argument to a cast/constructor would work to ascribe a type to the literal:

```hlsl
let f = float(1.23); // `float` type for literal can be inferred from context

let h = (half) 1.23; // `half` type for literal can be inferred from context
```

For the integer case I think this already mirrors what people do in practice, with stuff like `uint(0)`.

We could certainly try to devise something that folds into the literal syntax, but given how easy it is to make cast/constructor syntax fill the same role, I'm inclined to not invent something new.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+The size of a `bool` varies across target platforms; programs that need to ensure a matching in-memory layout between targets should not use `bool` for in-memory data structures.
+On all platforms, the `bool` type must be _naturally aligned_ (its alignment is its size).
+
+### Integer Types
+
+The following integer types are defines:
+
+| Name          | Description |
+|---------------|-------------|
+| `int8_t`      | 8-bit signed integer |
```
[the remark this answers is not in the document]
tfoleyNV: Probably, and adding them as `typedef`s should be too hard.

One thing we need to decide is what the "real" names of all these types should be, and then what should be the type aliases.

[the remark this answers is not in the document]
tfoleyNV: One issue with adding `int32_t` as a `typedef` of `int` (or vice versa) is that it closes the door to us ever having `int` be a pointer/word-sized integer on each target platform. I suppose that was never going to fly given how ingrained (`int` == 32 bits) is for most of our target users.

Eventually we might want to have a distinct `intptr_t` type in the table here, sitting between `int32_t` and `int64_t`, so that conversions of 32- and 64-bit integers to/from pointer-sized integers are always explicit in the AST and initial IR (since our initial IR is supposed to be portable across targets).

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+Specific targets may not support the other floating-point types.
+
+All floating-point types are stored in memory with their natural size and alignment on all targets that support them.
+
+Vector Types
+------------
+
+A vector type is written as `vector<T, N>` and represents an `N`-element vector with elements of type `T`.
+The _element type_ `T` must be one of the built-in scalar types, and the _element count_ `N` must be a specialization-time constant integer.
+The element count must be between 2 and 4, inclusive.
```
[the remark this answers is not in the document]
tfoleyNV: Whether or not to allow 1-element vectors, or matrices with only 1 row or column is something we can fine tune over time.

There is definitely a practical issue that GLSL doesn't allow 1-element vectors or matrices with only 1 row/column. Thus if we allow these in the front-end we'd have to include a pass to lower single-element vectors into scalars (and have a similar treatment for matrices, although ideally we will desugar matrices into arrays and avoid the whole problem).

In terms of the Slang front-end, the immediate question is: where would one-element vectors ever come from?

We could try to simplify our handling of swizzles so that a swizzle with `M` components always yields a `vector<M,T>` instead of having special-case rules for scalars. That sounds like a nice generalization (and mirrors your point about generic code being easier to write if you support one-element vectors), but it has the problem that if somebody writes:

```hlsl
float f = float3(99).z;
```

we need to have an explanation for how that code type-checks, which means we need implicit conversions from the hypothetical `float1` type to/from `float`, and those conversions need to be extremely low cost to make sure that things like overload resolution treat `float1` as basically the same as `float`.

We can implement all of that extra stuff (and that is what Cg and the "real" HLSL language do), but it has only a marginal impact on users (sometimes they will see `float1` instead of `float` in overload resolution errors, and wonder where it came from).

You also end up needing careful rules so that if, say, a user adds a `float4` and a `float1` then we *don't* handle that as a case of mismatched dimensions (like we would for a `float4` plus a `float3`) and instead handle it by casting from `float1 -> float -> float4` which is a bit weird and needs to be special-cased in the guts of the compiler (since things like generic argument inference already need to understand the scalar->vector case).

In the long run, I think I agree with your sentiment, though. We should realistically support vectors with any (non-negative) number of elements, and then legalize these based on what a given target can support. The special case of one-element vectors is just interesting because we probably have to allow them to easily convert to/from scalars in all contexts, which makes one-element vectors a big special case.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+A vector type is laid out in memory as `N` contiguous values of type `T` with no padding.
+The alignment of a vector type may vary by target platforms.
+The alignment of `vector<T,N>` will be at least the alignment of `T` and may be at most `N` times the alignment of `T`.
+
+As a convenience, Slang defines built-in type aliases for vectors of the built-in scalar types.
+E.g., declarations equivalent to the following are provided by the Slang standard library:
+
+```hlsl
+typealias float4 = vector<float, 4>;
+typealias int8_t3 = vector<int8_t, 3>;
```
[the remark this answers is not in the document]
tfoleyNV: Don't complain to me; this is straight up what is in HLSL as implemented by dxc.

There is an important point here, though, in that there is a conflict between two different conventions for what suffix digits on a type name mean:

* The shading language convention is that the digits tell us the number of elements in a vector (or rows x columns in a matrix)

* The dominant convention for every other systems-y language on the planet is that the digits tells us the number of bits in a scalar type

The result of these dualing conventions is that you have things like an OpenCL `float32` being a 32-element vector of `float` while a Swift `Float32` is a 32-bit floating-point scalar.

I am not aware of any precedent for a language that has been able to marry the conventions successfully, and the solution of copy-pasting the `_t` suffix from C over to shading languages sure feels like a "clever" cheat rather than something pleasant.

Cg and CUDA seem to handle the problem by defining old-fashioned C-style names for all the types of different sizes (e.g, `char`, `short`, `int`, `long`) so that vector names don't feel clunky. That kind of works, but it doesn't seem to reflect the trajectory that general-purpose languages have taken of being more explicit about the size of fundamental types.

The GLSL convention actually has some promise here in that it actually *doesn't* just use a numeric suffix and instead uses the word `vec` or `mat` before the suffix, for all types. GLSL instead uses a per-type *prefix* before the `vec` or `mat`, with the shorter prefixes being reserved for more common types.

I could get behind a convention of defining stuff like `Int8Vec4`, `Float16Mat2x2`, etc. as a convention for these types. It is still a mouthful, but far better than `int8_t4` or `float16_t2x2`.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+D3D constant buffer layout is similar to standard layout with two differences:
+
+* The initial alignment is 16 instead of one
+
+* If a field would have _improper straddle_, where the interval `(fieldOffset, fieldOffset+fieldSize)` (eclusive on both sides) contains any multiple of 16, *and* the field offset is not already a multiple of 16, then the offset of the field is adjusted to the next multiple of 16
+
+Array Types
+-----------
+
+An _array type_ is either a statically-sized or dynamically-sized array type.
```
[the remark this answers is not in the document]
tfoleyNV: Good point about arrays being value types. There's a related point there that because arrays are value types there is no specific reason why they can't/shouldn't be used as function return types, but I don't recall if we actually added the IR pass required to make that Just Work.

As for passing an array of unknown size to a function, its the kind of thing that should work, but I don't think we'd implemented all the required IR infrastructure. Ideally a function should be able to define a parameter that is an array of unknown size, and then you can call that function with either known-size or unknown-size arrays.

At the IR level, the simplistic approach would be to treat each function parameter of unknown array size as introducing an implicit generic `int` parameter that gives the size of the array, and then specializing functions based on actual array sizes at call sites. That approach could of course lead to heavy code bloat.

A smarter approach is probably to assume that at the "mechanism" level unknown-size arrays get passed by `const` reference, but with a convention that the callee needs to make their own copy of the array if they want to do mutation. Actually implementing that copy becomes a problem for any target that doesn't have an `alloca()` equivalent, so it isn't really a full solution...

Maybe the simplest answer is to decide by fiat that function parameters that have an unknown-size array type are immutable (at which point they can be passed by-value or by-const-reference and there should be no difference).

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+
+> Note: when declaring arrays of arrays (often thought of as "multidimensional arrays") a programmer must be careful about the difference between the two declaration syntaxes.
+> The following two declarations are equivalent:
+>
+> ```hlsl
+> int[3][5] a;
+> int a[5][3];
+> ```
+>
+> In each case, `a` is a five-element array of three-element arrays of `int`s.
```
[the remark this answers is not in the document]
tfoleyNV: > Presumably the point here is overall there are the same amount of underlying T in both examples?

No, the point is that those two declarations result in variables with the *exact same type*, despite the order of the element counts in `[]` being different.

This is an obscure but known issue with C-style declaration syntax (one that GLSL ran into head-first when it finally enabled "arrays of arrays").

The C-style declaration rules informally require reading a declaration from the inside out, so we have:

```hlsl
// `a` is a five element array ...
    a[5]   ;

// ... of three-element arrays
    a[5][3];

// .. of ints
int a[5][3];
```

Reading things with types requires going in the other direction, building up the type piece by piece:

```hlsl
// ... int
int

// ... a three-element array of ints
int[3]

// ... a five-element array of three-element arrays of ints
int[3][5]

// `a` is a five-element array of three-element arrays of ints
int[3][5] a;
```

If we could arbitrarily change the syntax, then the Right Way to fix this problem is to make it so that array types are defined the other way around, with `[3]int` instead of `int[3]` for a three-element array of `int`s. With the alternative ordering, then we have:

```hlsl
[5][3]int a; // `a` is a five-element array of three-element arrays of int
```

which looks even better using a more modern `let` or `var` declaration:

```hlsl
var a : [5][3]int;
```

Long story short: the current text in the reference represents the reality of the language and possibly the best possible semantics. The reason I called things out in a note is because of just how confusing the "best possible" semantics can be.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+
+When a variable is declared with an unknown-size array type, and also includes an initial-value expression:
+
+```hlsl
+int a[] = { 0xA, 0XB, 0xC, 0xD };
+```
+
+The compiler will attempt to infer an element count based on the type and/or structure of the initial-value expression.
+In the above case, the compiler will infer an element count of 4 from the structure of the initializer-list expression.
+Thus the preceding declaration is equivalent to:
```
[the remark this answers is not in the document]
tfoleyNV: I assume your point here is that we should make clear that specifying a "wrong" element count would be an error?

Although in that case, I think we'd just have an initializer list that initializes `a[0]` to `1`, and then the remaining elements to zero.

A better error case would be:

```hlsl
int a[2] = { 1, 2, 3 };
```

The document already has a note about how the exact rules for initializer lists need to be spelled out (in the "expressions" section), so I'd like to leave the finer points to that.

There is a definite tension to the way things are broken apart in the document right now in that some topics cross-cut sections and seem to require saying the same information more than once.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+
+The _stride_ of a type is the smallest multiple of its alignment not less than its size.
+
+Using the standard layout for an array type `T[]` or `T[N]`:
+
+* The _element stride_ of the array type is the stride of its element type `T`
+* Element `i` of the array starts at an offset that is `i` times the element stride of the array
+* The alignment of the array type is the alignment of `T`
+* The size of an unknown-size array type is unknown
+* The size of a known-size array with zero elements is zero
```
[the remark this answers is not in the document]
tfoleyNV: Yeah, our reflection and layout stuff might kludge things and treat a zero-element array as being an unknown-size array. We should fix that at some point.

(And yes, the meta-level point is that at times this document veers into being aspirational rather than just documenting the status quo. We should probably do better about calling out things that ought to work but don't).

Zero-element arrays should be easy enough to support in principle. We just need an IR pass that removes all variables/fields/parameters with zero-element-array type and any attempts to index into a zero-element array with `undefined` values.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+Every type has either known or unknown size.
+Types with unknown size arise in a few ways:
+
+* An unknown-size array type has unknown size
+
+* A structure type has unknown size if any field type has unknown size
+
+The use of tyeps with unknown size is restricted as follows:
+
+* A type with unknown size cannot be used as the element type of an array
```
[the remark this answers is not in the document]
tfoleyNV: No problem parsing it, but it is semantically meaningless to our layout model (so it *should* be illegal).

From what I remember, in C/C++ if the element type of an array is itself an unknown-size array then the latter array "decays" into a pointer type, so that the result is equivalent to defining an array of pointers, which is of course legal, but isn't what a Slang array is meant to represent.

docs/language-reference/04-types.md
```diff
@@ -0,0 +1,323 @@
...
+
+* A type with unknown size can only be used as the last field of a structure type
```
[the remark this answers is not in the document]
tfoleyNV: Well, I think we want the trailing allocation style, and I would argue that we at least need to *support* it.

We have language mechanisms for introducing an indirection where people want one, so I'd argue against introducing indirections that the programmer didn't ask for.

docs/language-reference/05-expressions.md
```diff
@@ -0,0 +1,351 @@
...
+---------------
+
+A _this expression_ consists of the keyword `this` and refers to the implicit instance of the enclosing type that is being operated on in instance methods, subscripts, and initializers.
+
+The type of `this` is `This`.
+
+Parenthized Expression
+----------------------
+
+An expression wrapped in parentheses `()` is a _parenthized expression_ and evaluates to the same value as the wrapped expression.
```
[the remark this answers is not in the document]
tfoleyNV: I used "wrapped" because earlier in the same sentence I said that the expression is "wrapped in parentheses," so it seemed to be clear.

docs/language-reference/06-statements.md
tfoleyNV: Yes, it should probably be marked that way.

To be clear, the specific syntax of the compile-time for was based around the following ideas:

* We might want to support range based loops in the future, and I hoped that `for(i in Range(0,N))` would be something we'd be able to support in that case. The goal was that the compile-time loop would look a lot like the syntax for something we'd support at runtime as well.

* The `$` prefix here is meant to indicate a shift from run-time control flow to compile-time (and specifically metaprogramming/macro time). We could conceivably extend this support with `$if` and other stuff, and if/when we allow arbitrary user-defined metaprogramming the `$` sigil would be used for splicing/inteprolating.

So the long story short is that the syntax there is at least motivated by things we might do in the future, but I agree that it probably deserves a caveat about how everything there is provisional.

docs/language-reference/07-declarations.md
tfoleyNV: Yes, our current parser does some disambiguation based on being able to tell types and values apart.

The correct fix there is not to change the language reference (although a warning here would be nice), but rather to fix our parser so that the bodies of functions/structures/etc. don't get parsed at first (instead we just capture all the tokens between matching `{}` and save them for later). Then we can go back later to parse the bodies of things *after* we've built up the full global scope so that forward references can be resolved.

The trend in stuff like Rust and Swift is in the other direction of making sure that a context-free parse of the entire input is possible without any kind of type information, while still supporting out of order declarations. The argument is that allowing for a context-free parse makes it easier to write stand-alone tools for documentation generation, refactoring, highlighting, etc.

Those languages aren't wrong about the benefits, but we don't have a choice given that we've inherited C-like syntax with its ambiguities, and dxc has added some C++ ambiguities to their flavor of HLSL creating even more potential problems down the line.

One down-side to the "it must be possible to parse without knowing anything" rule is that it means that invocations of user-defined syntax macros need to be marked (e.g., `!someMacro` in Rust) *and* the bodies of macro invocations need to be handled as completely undifferentiated token streams by tools (significantly impairing their utility in macro-heavy code, which some Rust projects, including most that try to do anything with GPUs/graphics).

docs/language-reference/07-declarations.md
tfoleyNV: I think the stuff below helps to clarify that `out` and `in out` *may* be implemented using C++ references and their semantics, but that this is not a guarantee provided by the language.

docs/language-reference/07-declarations.md
tfoleyNV: Fair enough, I had hoped that the example that defined it was close enough that I could call back to it, but maybe it helps to have each section stand on its own.

PR #1284, opened 2020-03-20
Handling of switch with empty body

jsmall-nvidia:
* Fixed handling of empty switch (would lead to an access through a nullptr)
* Added test for empty switch.

The solution used here was to try and detect a switch with no cases/default in it's body. If that's the case then nothing inside of the switch body is executed, only the condition code. Instead of lowering to a switch in this case, the lowering just lowers to the condition code.


source/slang/slang-lower-to-ir.cpp
```diff
@@ -3592,6 +3592,41 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
...
+            stmt = blockStmt->body;
+            continue;
+        }
+
+        if (auto seqStmt = as<SeqStmt>(stmt))
+        {
+            // Walk through the children looking for cases
+            for (auto childStmt : seqStmt->stmts)
+            {
+                if (hasSwitchCases(childStmt))
```
tfoleyNV: Minor point, but we should technically *not* recurse through inner/child `switch` statements, since their `case` or `default` statements don't count for the outer/parent statement.

jsmall-nvidia: I'm just following the same logic as lowerSwitchCases, which seems to be finding cases that do belong to the switch?

PR #1288, opened 2020-03-24
Better diagnostics on failure on CUDA.

jsmall-nvidia:
* Made DISABLE_TEST (and DISABLE_DIAGNOSTIC_TEST) explicitly disable a test
* Added better reporting of errors from CUDAComputeUtil
* Defaulted to compute3_0 as arch type on nvrtc
* Disable CUDA SM7.0 tests - they fail on some CI targets with CUDA 

Previously a line with DISABLE_TEST did nothing. This changes with this PR, as the test and it's parameters are still parsed, but the test itself is marked as disabled. This is useful because it provides a mechanism to disable a test that might be synthesized otherwise and then fail.


tools/render-test/cuda/cuda-compute-util.cpp
```diff
@@ -18,13 +18,79 @@ using namespace Slang;
 SLANG_FORCE_INLINE static bool _isError(CUresult result) { return result != 0; }
 SLANG_FORCE_INLINE static bool _isError(cudaError_t result) { return result != 0; }
 
-#if 0
-#define SLANG_CUDA_RETURN_ON_FAIL(x) { auto _res = x; if (_isError(_res)) return SLANG_FAIL; }
-#else
+struct CUDAResult
```
tfoleyNV: Is this type necessary vs. just having two overloads of `_handleCUDAError`?

source/core/slang-nvrtc-compiler.cpp
```diff
@@ -309,7 +309,12 @@ SlangResult NVRTCDownstreamCompiler::compile(const CompileOptions& options, RefP
         cmdLine.addArg("-w");
 
         //
+#if 0
         cmdLine.addArg("-arch=compute_70");
+#else
+        // Needed for Warp intrinsics
+        cmdLine.addArg("-arch=compute_30");
+#endif
```
tfoleyNV: If having the `#ifdef`ed out code here intentional?

jsmall-nvidia: Just so I remember it's a place holder, and that I will have to pass something down to nvrtc depending on what arch is required, based on compiled code.

PR #1290, opened 2020-03-24
Parser changes to improve handling of static method calls

tfoleyNV:
Static Method Calls
-------------------

The main fix here is for parsing of calls to static methods. Given a type like:

    struct S
    {
        void doThing();
    }

the parser currently gets tripped up on a statement like:

    S::doThing();

The problem here is that the `Parser::ParseStatement` routine was using the first token of lookahead to decide what to do, and in the case where it saw a type name it assumed that must mean the statement would be a declaration.

It turns out that `Parser::ParseStatement` already had a more intelligent bit of disambiguation later on when handling the general case of an identifier (for which it couldn't determine the type-vs-value status at parse time), and simply commetning out the special-case handling of a type name and relying on the more general identifier case fixes the issue.

That catch-all case still has some issues of its own, and this change expands on the comments to make some of those issues clear so we can try to address them later.

Empty Declarators
-----------------

One reason why the static method call problem was hard to discover was that it was masked by the parser allowing for empty declarator. That is, given input like:

    S::doThing();

This can be parsed as a variable declaration with a parenthesized empty declarator `()`.

Practically, there is no reason to support empty declarators anwhere except on parameters, and allowing them in other contexts could make parser errors harder to understand.

This change makes the choice of whether or not empty declarators are allowed something that can be decided at each point where we parse a declarator, and makes it so that only parsing of parameters opts in to allowing them.

By disabling support for empty declarators in contexts where they don't make sense, we make code like the above a parse error when it appears at global scope, rather than a weird semantic error.

A more complete future version of this change might *also* make support for parenthesized declarators an optional feature, or remove that support entirely. Slang doesn't actually support pointers (yet) so there is no real reason to allow parenthesized declarators right now.

One note for future generations is that using an emptye declarator on a parameter of array type can actually create an ambiguity. If the user writes:

    void f(int[2][3]);

did they mean for it to be interpreted as:

    void f(int a[2][3]);

or as:

    void f(int[2][3] a);

or even as:

    void f(int[2] a[3]);

The first case there yields a different type for `a` than the other two, but is also what we pretty much *have* to support for backwards compatibility with HLSL. Requiring all function declarations to include parameter names would eliminate this potentially confusing case.

Layout Modifiers
----------------

One of the above two syntax changes led to a regression in the output for a diagnostic test for `layout` modifiers (which are a deprecated but still functional feature from back when `slangc` supported GLSL input).

The original output of the test case seemed odd, and when I looked at the parsing logic I saw that an early-exit error case was leading to spurious error messages because it failed to consume all the tokens inside the `layout(...)`. Fixing the logic to not use an early-exit (and instead rely on the built-in recovery behavior of `Parser`) produced more desirable diagnosic output.

I changed the input file to put the `binding` and `set` specifiers on differnet lines so that the error output could show that the compiler properly tags both of the syntax errors.


source/slang/slang-parser.cpp
```diff
@@ -3314,24 +3342,66 @@ namespace Slang
...
+
             // If the next token after we parsed a type looks like
             // we are going to declare a variable, then lets guess
             // that this is a declaration.
             //
             // TODO(tfoley): this wouldn't be robust for more
             // general kinds of declarators (notably pointer declarators),
             // so we'll need to be careful about this.
+            //
+            // If the lookahead token is `*`, then we have to decide
```
jsmall-nvidia: I guess you could operator overload * and do something that had side effects, but probably not a significant issue for now even if true.

tfoleyNV: Correct, an operator overload could violate the assumptions, but it would still be really odd to write `a * b;` as a statement, to the point that I don't think the language would be unreasonable in banning it.

It might still be surprising if you are refactoring/editing a shader and you end up with a stray subexpression on its own line and the compiler gives you an error like "hey, `a` is not a type!" and/or "hey, you already *have* a variable called `b` in this function!".

I suspect that the best long-term answer (if we continue to support C-style declaration syntax, which I don't think we have any choice about) will be to make our parsing of `{}`-enclosed bodies be context-sensitive, so that we intermix semantic checking and parsing just like a C/C++ compiler would. That would let us more easily parse tricky cases of the C/C++ syntax without error, and even in cases where we don't want to support some aspects of that syntax it would make our error messages more robust and tailored if we have the additional context.

The global scope would still be parsed in a context-free fashion to support out-of-order declarations, but most of the ambiguity is gone there because of the lack of support for statements/expressions. The one tricky bit that would come up at the global scope would be the "most vexing parse" sorts of situations which all use C++ variable declaration+construction syntax:

```c++
Foo bar(Baz); // a function that takes an (unnamed) parameter of type `Baz`
Foo bar(baz); // a variable that is being initialized/constructed with the argument value `baz`

// Ambiguous:
// * A function that takes an unnamed parameter of type `Baz`?
// * A variable that is initialized/constructed with a default-constructed `Baz` value?
// (the latter choice is arbitrarily selected)
//
Foo bar(Baz());
```

If we ban C++-style variable declarations with constructor arguments at global scope, this would be a non-issue, but unfortunately the DXR extensions technically introduced a few global types used for DXIL collection/linkage meta-data stuff that are designed to be used with constructor syntax.

We may need to define a plan for that stuff sooner or later, but the fact that C++ variable+constructor syntax is a complete mess makes me disinclined to spend too much effort on trying to support it.

PR #1291, opened 2020-03-25
Unroll target improvements

jsmall-nvidia:
* Add [unroll] support for CUDA
* Add a sort of placeholder support for [unroll] in C++
* Write up about unroll in target-compatibility.md
* Updated loop-unroll.slang so 
  * Doesn't assume all threads happen simultaneously (they don't currently on CPU)
  * Enable for other targets where it should work

That currently the DX12 DXC test is disabled, because it causes a crash. Not clear why yet (after all it works with DXBC).

### DIFF
--- a/source/slang/slang-emit-c-like.cpp
+++ b/source/slang/slang-emit-c-like.cpp
@@ -2592,37 +2592,8 @@ void CLikeSourceEmitter::emitRegion(Region* inRegion)
                 //
                 if (auto loopControlDecoration = loopInst->findDecoration<IRLoopControlDecoration>())
                 {
-                    const auto loopMode = loopControlDecoration->getMode();
-                    if (loopMode == kIRLoopControl_Unroll)
-                    {
-                        switch (getSourceStyle())
-                        {
-                            case SourceStyle::HLSL:
-                            {
-                                m_writer->emit("[unroll]\n");
-                                break;
-                            }
-                            case SourceStyle::GLSL:
-                            {
-                                // Note: loop unrolling control not available in GLSL
-                                break;
-                            }
-                            case SourceStyle::CUDA:
-                            {
-                                m_writer->emit("#pragma unroll\n");
-                                break;
-                            }
-                            case SourceStyle::C:
-                            case SourceStyle::CPP:
-                            {
-                                // This relies on a suitable definition in slang-cpp-prelude.h or defined in C++ compiler invocation.
-                                m_writer->emit("SLANG_UNROLL\n");
-                                break;
-                            }
-                            default: break;
-                        }
-                    }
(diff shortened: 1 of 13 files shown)


source/slang/slang-emit-c-like.cpp
```diff
@@ -2592,18 +2592,35 @@ void CLikeSourceEmitter::emitRegion(Region* inRegion)
...
-                            m_writer->emit("[unroll]\n");
+                            case SourceStyle::HLSL:
+                            {
+                                m_writer->emit("[unroll]\n");
+                                break;
+                            }
+                            case SourceStyle::GLSL:
+                            {
+                                // Note: loop unrolling control not available in GLSL
+                                break;
```
tfoleyNV: It looks like there is actually a [GLSL extension](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GL_EXT_control_flow_attributes.txt) defined for control-flow-related attributes so that GLSL can generate the SPIR-V stuff. We should consider mapping the Slang attributes over to those and adding the extension as a requirement when they are used (assuming glslang supports them).

The gotcha here is that just passing down an unrolling hint to SPIR-V won't necessarily solve the problem that certain loops needs to be unrolled *before* GLSL/SPIR-V gets generated, or else the code won't conform to the constraints of the language spec (e.g., that certain function call arguments need to be compile-time constants).

tests/compute/loop-unroll.slang
```diff
@@ -1,7 +1,15 @@
...
+//TEST(compute):COMPARE_COMPUTE:-cpu
+//TEST(compute):COMPARE_COMPUTE:-cuda
+// Note VK output is not loop unrolled
+//TEST(compute):COMPARE_COMPUTE:-vk
 
-//TEST_INPUT:ubuffer(data=[0 1 2 3], stride=4):out
-//TEST_INPUT:ubuffer(data=[1 2 3 0], stride=4):
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out, name buffers[0]
+//TEST_INPUT:ubuffer(data=[0 1 2 3], stride=4):name buffers[1]
+//TEST_INPUT:ubuffer(data=[1 2 3 0], stride=4):name buffers[2]
```
tfoleyNV: As discussed elsewhere, I suspect the root cause of the crash in the D3D12 + DXIL case is that these `//TEST_INPUT` lines are effectively declaring three different descriptor ranges (each of size one) while the generated DXIL is compiled to expect/require a single resource range of size three.

I'm not sure why that kind of issue would lead to a crash rather than some kind of validation error message, so maybe I'm mistaken and something more sinister is going on.

PR #1297, opened 2020-03-27
WaveReadLaneAt  with dynamic index for vk.

jsmall-nvidia:
* Support for WaveReadLaneAt with dynamic (but uniform across Wave) on Vk by enabling VK1.4.
* Fixed wave-lane-at.slang test to test with laneId that is uniform across the Wave.

NOTE! That as it stands the VK tests and the feature will not work without an updated (and patched) version of glslang, which is not yet available.


source/slang/hlsl.meta.slang
```diff
@@ -2722,14 +2722,17 @@ matrix<T,N,M> WaveReadLaneFirst(matrix<T,N,M> expr);
 
 // NOTE! On GLSL based targets the lane index *must* be a compile time expression!
 // See https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_shader_subgroup.txt
+// It is allowed to be 'dynamically uniform within the subgroup' if it's SPIR-V 1.4.
+// TODO(JS): For now we'll use 1.4, but aim for the future for the compiler to determine
+// if the line the is compile constant, and reduce requirement to 1.3
 __generic<T : __BuiltinType>
 __glsl_extension(GL_KHR_shader_subgroup_ballot)
-__spirv_version(1.3)
+__spirv_version(1.4)
```
tfoleyNV: When I look at the relevant [extension](https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_shader_subgroup.txt), I see (emphasis mine):

> ... must be an integral constant expression when targeting **SPIR-V 1.4 and below**, otherwise it must be dynamically uniform within the subgroup.

That seems to imply that we need SPIR-V 1.5 or higher to opt in to the more flexible behavior.

PR #1298, opened 2020-03-27
Adds WaveShuffle intrinsic

jsmall-nvidia:
On top of #1297

* Added WaveIntrinsic to stdlib 
   * Maps to WaveReadLaneAt on HLSL. 
* Added tests

NOTE! Using WaveReadLaneAt *will not work on some hardware*. That the HLSL spec requires only to work if the landId is 'dynamically uniform' across the Wave. WaveShuffle does not require this - and so if used on hardware that doesn't also allow for this behavior is undefined.


docs/target-compatibility.md
```diff
@@ -56,6 +57,14 @@ tex.GetDimensions is the GetDimensions method on 'texture' objects. This is not
 
 SM6.5 Wave Intrinsics are supported, but requires a downstream DXC compiler that supports SM6.5. As it stands the DXC shipping with windows does not. 
 
+## WaveShuffle
+
+WaveShuffle is an intrinsic added to the Slang stdlibrary to expose the glsl `subgroupShuffle` intrinsics and allow loosened requirements on laneId.
```
tfoleyNV: I'm not sure I like "stdlibrary" as a term. I might argue for consistently using either "standard library" or "stdlib," and it could probably be argued that the latter should be reserved for less formal documentation (like comments).

docs/target-compatibility.md
```diff
@@ -56,6 +57,14 @@ tex.GetDimensions is the GetDimensions method on 'texture' objects. This is not
 
 SM6.5 Wave Intrinsics are supported, but requires a downstream DXC compiler that supports SM6.5. As it stands the DXC shipping with windows does not. 
 
+## WaveShuffle
+
+WaveShuffle is an intrinsic added to the Slang stdlibrary to expose the glsl `subgroupShuffle` intrinsics and allow loosened requirements on laneId. 
+
+In `HLSL` requires the `laneId` must be 'dynamically uniform' across the wave. WaveShuffle has the same functionality but relaxes this restriction.
```
tfoleyNV: Unrelated to this change, but we'll probably eventually need to normalize on one dialect of English when it comes to single-vs-double quotation marks, etc. The British conventions are certainly more logical, and I can't really complain when you are writing so much of our documentation, so I'm not inclined to try to impose anything too soon.

If/when we have resources to spare it would be good to assign somebody to scrub all our documents and get them into an explicitly chosen style.

docs/target-compatibility.md
```diff
@@ -56,6 +57,14 @@ tex.GetDimensions is the GetDimensions method on 'texture' objects. This is not
 
 SM6.5 Wave Intrinsics are supported, but requires a downstream DXC compiler that supports SM6.5. As it stands the DXC shipping with windows does not. 
 
+## WaveShuffle
+
+WaveShuffle is an intrinsic added to the Slang stdlibrary to expose the glsl `subgroupShuffle` intrinsics and allow loosened requirements on laneId. 
+
+In `HLSL` requires the `laneId` must be 'dynamically uniform' across the wave. WaveShuffle has the same functionality but relaxes this restriction. 
+
+`WaveReadLaneAt` most obviously maps to `subgroupBroadcast` in GLSL. This has the extra restriction the index must be compile time consts. With SPIR-V 1.4 it is allowed to be 'dynamically uniform', but doesn't work on current glslang.
```
tfoleyNV: The comment about SPIR-V 1.4 here seems inaccurate w.r.t. the extension spec, as covered in the review of the previous PR. We should double check the facts here before checking this in.

source/slang/hlsl.meta.slang
```diff
@@ -2722,21 +2722,47 @@ matrix<T,N,M> WaveReadLaneFirst(matrix<T,N,M> expr);
 
 // NOTE! On GLSL based targets the lane index *must* be a compile time expression!
 // See https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_shader_subgroup.txt
+// It is allowed to be 'dynamically uniform within the subgroup' if it's SPIR-V 1.4.
```
tfoleyNV: More 1.4 references here.

PR #1301, opened 2020-03-30
CUDA version handling

jsmall-nvidia:
* Mark up intrinsics with __cuda_sm_version
* Ability to pass a version requirement to the Downstream compiler
* Added CUDAExtensionTracker - to track highest SM version needed
* Fix handling of CUdevice being mixed with device index
* Enabled CUDA tests which would fail because not high enough SM (they will be ignored on CI now)


source/core/slang-downstream-compiler.h
```diff
@@ -207,6 +208,16 @@ class DownstreamCompiler: public RefObject
...
     };
 
+    struct Version
+    {
+        enum class Kind
+        {
+            CUDASM,                     ///< What the version is for
+        };
+        Kind kind;
+        SemanticVersion version;
```
tfoleyNV: Hmm... So `SemanticVersion` here is meant to reference Semantic Versioning (SemVer) I assume? That might be a bit of a risky/misnomer since not all of the projects we'd be applying these versions to claim to follow SemVer.

There's also the gotcha that this codebase already has to deal with the legacy HLSL/Cg terminology of what a "semantic" is, along with the fact that a compiler already has stuff like semantic analysis... I worry that "semantic" is too overloaded here.

tfoleyNV: I might also argue for `Version` here to be something like a `FeatureVersion` or `CapabilityVersion` to capture that it is marking the version of a particular feature or capability supported by a target.

That would turn `requiredVersions` below into `requiredCapabilities`, which I think is more accurate to what we might want to specify.

Something like a GLSL extension could then be marked as a capability where version is immaterial.

jsmall-nvidia: I changed to CapabilityVersion, and requiredCapabilityVersion. 

On SemanticVersion - I think it's ok. It doesn't enforce any of the more subjective aspects. It's just a version that can consist of up to 3 parts. 'Version' would be too vague.

PR #1307, opened 2020-04-02
Initial work to support OptiX output for ray tracing shaders

tfoleyNV:
This change represents in-progress work toward allowing Slang/HLSL ray-tracing shaders to be cross-compiled for execution on top of OptiX. The work as it exists here is incomplete, but the changes are incremental and should not disturb existing supported use cases.

One major unresolved issue in this work is that the OptiX SDK does not appear to set an environment variable

Changes include:

* Modified the premake script to support new options for adding OptiX to the build. Right now the default path to the OptiX SDK is hard-coded because the installer doesn't seem to set an environment variable. We will want to update that to have a reasonable default path for both Windows and Unix-y platforms in a later chance.

* I ran the premake generator on the project since I added new options, which resulted in a bunch of diffs to the Visual Studio project files that are unrelated to this change. Many of the diffs come from previous edits that added files using only the Visual Studio IDE rather than by re-running premake, so it is arguably better to have the checked-in project files more accurately reflect the generated files used for CI builds.

* The "downstream compiler" abstraction was extended to have an explicit notion of the kind of pipeline that shaders are being compiled for (e.g., compute vs. rasterization vs. ray tracing). This option is used to tell the NVRTC case when it needs to include the OptiX SDK headers in the search path for shader compilation (and also when it should add a `#define` to make the prelude pull in OptiX). This code again uses a hard-coded default path for the OptiX SDK; we will need to modify that to have a better discovery approach and also to support an API or command-line override.

  * One note for the future is that instead of passing down a "pipeline type" we could instead pass down the list/set of stages for the kernels being compiled, and the OptiX support could be enabled whenever there is *any* ray tracing entry point present in a module. That approach would allow mixing RT and compute kernels during downstream compilation. We will need to revisit these choices when we start supporting code generation for multiple entry points at a time.

* The CUDA emit logic is currently mostly unchanged. The biggest difference is that when emitting a ray-tracing entry point we prefix the name of the generated `__global__` function with a marker for its stage type, as required by the OptiX runtime (e.g., a `__raygen__` prefix is required on all ray-generation entry points).

* The `Renderer` abstraction had a bare minimum of changes made to be able to understand that ray-tracing pipelines exist, and also that some APIs will require the name of each entry point along with its binary data in order to create a program.

* The `ShaderCompileRequest` type was updated so that only a single "source" is supported (rather than distinct source for each entry point), and also the entry points have been turned into a single list where each entry identifies its stage instead of a fixed list of fields for the supported entry-point types.

* The CUDA compute path had a lot of code added to support execution for the new ray-tracing pipeline type. The logic is mostly derived from the `optixHello` example in the OptiX SDK, and at present only supports running a single ray-generation shader with no parameters. The code here is not intended to be ready for use, but represents a signficiant amount of learning-by-doing.

* The `slang-support.cpp` file in `render-test` was updated so that instead of having separate compilation logic for compute vs. rasterization shaders (which would mean adding a third path for ray tracing), there is now a single flow to the code that works for all pipeline types and any kind of entry points.

  * Implicit in the new code is dropping support for the way GLSL was being compiled for pass-through render tests, which means pass-through GLSL render tests will no longer work. It seems like we didn't have any of those to begin with, though, so it is no great loss.

  * Also implicit are some new invariants about how shaders without known/default entry points need to be handled. For example, the ray tracing case intentionally does not fill in entry points on the `ShaderCompileRequest` and instead fully relies on the Slang compiler's support for discovering and enumerating entry points via reflection. As a consequence of those edits the `-no-default-entry-point` flag on `render-test` is probably not working, but it seems like we don't have any test cases that use that flag anyway.

Given the seemingly breaking changes in those last two bullets, I was surprised to find that all our current tests seem to pass with this change. If there are things that I'm missing, I hope they will come up in review.


source/slang/slang-emit-cuda.cpp
```diff
@@ -606,29 +606,65 @@ void CUDASourceEmitter::emitModuleImpl(IRModule* module)
...
+                //
+                auto stage = entryPointDecor->getProfile().GetStage();
+                switch( stage )
                 {
-                    if (ii != 0) m_writer->emit(", ");
-                    m_writer->emit(sizeAlongAxis[ii]);
+                default:
+                    break;
+
+            #define CASE(STAGE, PREFIX) \
```
jsmall-nvidia, line 633: Doesn't this need a break;?

source/slang/slang-emit-cuda.cpp
```diff
@@ -606,29 +606,65 @@ void CUDASourceEmitter::emitModuleImpl(IRModule* module)
...
+                // OptiX, and in that case the name of the global symbol
+                // must be prefixed to indicate to the OptiX runtime what
+                // stage it is to be compiled for.
+                //
+                auto stage = entryPointDecor->getProfile().GetStage();
+                switch( stage )
                 {
-                    if (ii != 0) m_writer->emit(", ");
-                    m_writer->emit(sizeAlongAxis[ii]);
+                default:
```
jsmall-nvidia, line 630: I guess for this as it stands having the switch (or if) without the #define CASE might be preferable.

jsmall-nvidia:
It appears that 

serialized-module-entry-point-test
library-entry-point-test.slang 

uses -no-default-entry-point.

jsmall-nvidia:
On the point about GLSL and pass through, it could be argued that it is an issue that we don't have a GLSL passthrough test, to make sure that works, even though there doesn't appear to be one now. 

Well it looks like we have one... glslang-error.slang

tfoleyNV:
> It appears that
> 
> serialized-module-entry-point-test
> library-entry-point-test.slang
> 
> uses -no-default-entry-point.

I wonder how I missed that when I searched for that text, and also how the tests appeared to work on my machine...

tfoleyNV:
> On the point about GLSL and pass through, it could be argued that it is an issue that we don't have a GLSL passthrough test, to make sure that works, even though there doesn't appear to be one now.
> 
> Well it looks like we have one... glslang-error.slang

To be clear, this change does *not* break our ability to test GLSL pass-through. It just breaks our ability to use pass-through GLSL as part of a render-test. Pass-through GLSL is used for many of our cross-compilation tests already.

source/slang/slang-emit-cuda.cpp
tfoleyNV: Oops. Yeah that needs a fix. I did some cleanup edits before pushing the PR and must have missed that.

source/slang/slang-emit-cuda.cpp
tfoleyNV: Well, the `CASE` macro will get a lot more uses as I fill this out, since every since RT stage needs its own prefix and there are 6 of those in DXR.

PR #1310, opened 2020-04-07
Remove static struct members from layout and reflection

jsmall-nvidia:
* Added MemberFilterStyle - controls action of FilteredMemberList and FilteredMemberRefList
* Splt out template implementations
* Use more standard method names dofr FilteredMemberRefList
* Added reflect-static.slang test
* Added isNotEmpty/isEmpty to filtered lists
* Added ability to index into filtered list (so not require building of array)


jsmall-nvidia:
Ok - I defaulted to MemberFilterStyle::All. It means slightly less verbose, but perhaps not quite as clear on call site the specifics. I didn't default when doing the change because I wanted to inspect all the call sites.

Having a predicate version would be easy to do. If we end up with much more complexity (say by expanding the 'filter styles') that may make sense and certainly is more flexible, although it is likely to be slower.

On third point perhaps a different naming would make the intention more clear?

tfoleyNV:
> I didn't default when doing the change because I wanted to inspect all the call sites.

Yes, I suspected that was the reason it was written that way, and it is good to have confidence that we surveyed all the call sites. With the default it is now possible for us to make a mistake at future call sites, but I think it is reasonable that `getMembersOfType` by default returns *all* members of the given type (both static and instance), since we aren't specifying any other kind of assumption at the call site.

> On third point perhaps a different naming would make the intention more clear?

It would be more explicit to have `::WithStaticModifier` and `::WithoutStaticModifier` as the cases rather than `::Static` and `::Instance`, but those are admittedly somewhat clunky.

source/slang/slang-check-constraint.cpp
```diff
@@ -284,7 +284,7 @@ namespace Slang
         // These seem more reasonable to have influence constraint solving, since it could
         // conceivably let us specialize a `X<T> : IContainer` to `X<Int>` if we find
         // that `X<T>.IndexType == T`.
-        for( auto constraintDeclRef : getMembersOfType<GenericTypeConstraintDecl>(genericDeclRef) )
+        for( auto constraintDeclRef : getMembersOfType<GenericTypeConstraintDecl>(genericDeclRef, MemberFilterStyle::All) )
```
tfoleyNV, line 287: One case of using `MemberFilterStyle::All` slipped through the cleanup change. Probably not worth editing to fix unless you really want to.

tfoleyNV:
One option is to stick with the current `::Instance` and `::Static` names, but to then have a stated goal of making the "is this a static member?" test be refined in a future version so that it is accurate for all kinds of members...

We probably need to clean up that whole "is this static" stuff in the codebase at some point, because it seems like different code paths currently use different queries with slightly different meanings. I looked at the `isEffectivelyStatic` function recently and was surprised to see that (IIRC) it did *not* consider things like global variables to be "effectively static" which is really weird.

PR #1311, opened 2020-04-07
Fixes for IR generics

tfoleyNV:
There are a few different fixes going on here (and a single test that covers all of them).

1. Fix optionality of trailing semicolon for `struct`s
======================================================

We have logic in the parser that tries to make a trailing `;` on a `struct` declaration optional. That logic is a bit subtle and couild potentially break non-idiomatic HLSL input, so we try to only trigger it for files written in Slang (and not HLSL). For command-line `slangc` this is based on the file extension (`.slang` vs. `.hlsl`), and for the API it is based on the user-specified language.

The missing piece here was that the path for handling `import`ed code was *not* setting the source language of imported files at all, and so those files were not getting opted into the Slang-specific behavior. As a result, `import`ed code couldn't leave off the semicolon.

2. Fix generic code involving empty `interface`s
================================================

We have logic that tries to only specialize "definitions," but the definition-vs-declaration distinction at the IR level has historically been slippery. One corner case was that a witness table for an interface with no methods would always be considered a declaration, because it was empty. The notion of what is/isn't a definition has been made more nuanced so that it amounts to two main points:

* If something is decorated as `[import(...)]`, it is not a definition

* If something is a generic/func (a declaration that should have a body), and it has no body, it is a declaration

Otherwise we consider anything a definition, which means that non-`[import(...)]` witness tables are now definitions whether or not they have anything in them.

3. Fix IR lowering for members of generic types
===============================================

The IR lowering logic was trying to be a little careful in how it recurisvely emitted "all" `Decl`s to IR  code. In particular, we don't want to recurse into things like function parameters, local variables, etc. since those can never be directly referenced by external code (they don't have linkage).

The existing logic was basically emitting everything at global scope, and then only recursing into (non-generic) type declarations. This created a problem where a method declared inside a generic `struct` would not be emitted to the IR for its own module at all *unless* it happened to be called by other code in the same module.

The fix here was to also recurse into the inner declaration of `GenericDecl`s. I also made the code recurse into any `AggTypeDeclBase` instead of just `AggTypeDecl`s, which means that members in `extension` declarations should not properly be emitted to the IR.

Conclusion
==========

These fixes should clear up some (but not all) cases where we might emit an `/* unhandled */` into output HLSL/GLSL. A future change will need to make that path a hard error and then clean up the remaining cases.


tests/language-feature/generics/struct-generic-value-param.slang
```diff
@@ -0,0 +1,63 @@
...
+	return data.state + val*16;
+}
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+[numthreads(4, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+    uint tid = dispatchThreadID.x;
```
jsmall-nvidia, line 59: Tabbing seems to be going between 2 and 4 spaces?

tests/language-feature/generics/struct-generic-value-param-import.slang
```diff
@@ -0,0 +1,19 @@
...
+// and also incidentally tests that a trailing `;` is optional
+// for `struct` decalrations in Slang files, including
+// any `import`ed code.
+
+interface IData {}
+
+struct Data<let kCount : int> : IData
+{
+    int state;
+
```
jsmall-nvidia, line 14: Here as well the tabs seem to change.

PR #1314, opened 2020-04-09
Literal folding on other operators

jsmall-nvidia:
* Previously on did literal folding around negation. Expand the ops to cover prefix +, --, ++, ~, !.
* Added a test for the vector case, that has lots of unary ops


source/slang/slang-parser.cpp
```diff
@@ -4706,9 +4706,45 @@ namespace Slang
         }
     }
 
+    static IRIntegerValue _foldIntegerPrefixOp(TokenType tokenType, IRIntegerValue value)
+    {
+        switch (tokenType)
+        {
+            case TokenType::OpInc:      return ++value;
+            case TokenType::OpDec:      return --value;
```
tfoleyNV, line 4714: I may be missing something, but wouldn't the `++` and `--` operators always be invalid on something that is a literal constant (since it can't also be an l-value), and if it *were* valid wouldn't trying to "fold" them like this actually drop the side effect?

jsmall-nvidia: Actually yeah cos they'd have to be writable.... will remove.

tfoleyNV:
Looks like `tests/compute/static-const-array.slang` is enabled for CPU here, but doesn't have a `name` specified for its output.

PR #1320, opened 2020-04-14
CUDA global scope initialization of arrays without function calls.

jsmall-nvidia:
* Changed CUDA makeArray to try to generate code that has no function calls
* Fixed bug in 'convert' generation
* Added matrix static const array test
* Fixed unbounded-array-of-array-syntax.slang  to work on CUDA

The underlying problem that this PR aims to help resolve is around CUDA not being able to call functions to initialize global `static` variables. The compiler complains that it is not able to call __device__ code dynamically. This contrasts to C++ (which the CUDA emitter is based on), it can call functions to set up global variables. 

One of the reasons this is a problem is that, what seem like very simple constructions do by default generate function calls. Take 

```
matrix<float, 2, 2> mat = { float2(1 2), float2(3, 4) };
```

Will generate calls to make_ functions to create the vectors on CUDA, and that is not permitted. What is permitted would be the struct initialization that produces the same result, which is somewhat convoluted, but in essence means using struct/array initialization and not function calls. 

The solution used here, is not complete in covering all the cases. For example having

```
static const matrix<float, 4, 4> kIdentity = { 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 };
static const matrix<float 2, 2> kArray[1] = kIdentity; 
```

Does not work because conversion currently produces a function call. 

```
static const matrix<float 2, 2> kArray[1] = { 1, 2 }; 
```
Where 1, 2 become vectors 1, 1 and 2, 2 does not work either because they become function calls. 

Such a transformation could be made, but doing so would require more code, and it was felt that clean up in terms of how this is represented in the IR at the point of emit, might be more appropriate, or at least discussed.

A final point, is that it might seem that doing something like 

```
static const matrix<float, 2, 2> kMatrix = { 1, 0, 0, 1 }; 
```

Would also be a problem - in that it would create a function call at global scope. In practice this doesn't happen, as kMatrix does not exist as a variable, but is inlined where used, and so the function call happens in program flow. 

Whether its safe/reasonable to rely on such behavior is perhaps arguable. A test was modified to check that this does work though.


source/slang/slang-emit-cuda.cpp
```diff
@@ -369,6 +513,23 @@ bool CUDASourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+            IRType* dataType = inst->getDataType();
+            IRArrayType* arrayType = as<IRArrayType>(dataType);
+
+            IRType* elementType = arrayType->getElementType();
+
+            // Emit braces for the FixedArray struct. 
+            m_writer->emit("{\n");
+            m_writer->indent();
+
+            _emitInitArray(elementType, inst->getOperands(), Index(inst->getOperandCount()));
```
tfoleyNV, line 527: I might argue for this being called `_emitInitializerList` to indicate the part of the C grammar that it is emitting.

source/slang/slang-emit-cuda.h
```diff
@@ -76,6 +76,9 @@ class CUDASourceEmitter : public CPPSourceEmitter
     
     SlangResult _calcCUDATextureTypeName(IRTextureTypeBase* texType, StringBuilder& outName);
 
+    void _emitInitArray(IRType* elementType, IRUse* operands, Index operandCount);
+    void _emitInitArrayValue(IRType* elementType, IRInst* value);
```
tfoleyNV, line 80: These could arguably be hoisted up to `CLikeSourceEmitter` since initializer lists are a common feature across many of these languages.

source/slang/slang-emit-cuda.cpp
```diff
@@ -349,6 +349,150 @@ void CUDASourceEmitter::emitLoopControlDecorationImpl(IRLoopControlDecoration* d
...
+                        return;
+                    }
+                }
+            }
+                      
+            break;
+        }
+    }
+
+    // All other cases we just use the default emitting - might now work on arrays defined in global scope on CUDA though
```
tfoleyNV, line 477: "might now" -> "might not"

source/slang/slang-emit-cuda.cpp
```diff
@@ -349,6 +349,150 @@ void CUDASourceEmitter::emitLoopControlDecorationImpl(IRLoopControlDecoration* d
...
+    switch (value->op)
+    {
+        case kIROp_Construct:
+        case kIROp_MakeMatrix:
+        case kIROp_makeVector:
+        {
+            IRType* type = value->getDataType();
+
+            // If the types are the same, we can can just break down and use
+            if (_areEquivalent(dstType, type))
```
tfoleyNV, line 412: Are there cases where this check doesn't pass (that is, where we are initializing something from an expression of the wrong type)? If so, that would seem to be a problem with the IR code we are generating, since it should always be fully "type correct" at the IR level.

source/slang/slang-emit-cuda.cpp
```diff
@@ -349,6 +349,150 @@ void CUDASourceEmitter::emitLoopControlDecorationImpl(IRLoopControlDecoration* d
...
+                if (auto vecType = as<IRVectorType>(type))
+                {
+                    if (UInt(GetIntVal(vecType->getElementCount())) == value->getOperandCount())
+                    {
+                        _emitInitArray(vecType->getElementType(), value->getOperands(), value->getOperandCount());
+                        return;
+                    }
+                }
+                else if (auto matType = as<IRMatrixType>(type))
+                {
```
tfoleyNV, line 423: The logic in the matrix case is kind of complicated. Ideally we should be going back and vetting the IR to make sure things emit in a more predictable way:

* `makeVector` should always apply to `N` scalars and produce an `N`-component vector

* `makeMatrix` should always apply to `R` vectors (each with `C` components) and produce an `R` by `N` matrix

* `Construct` should be eliminated completely, and get replaced with more specific opcodes and/or explicit code in the stdlib.

(None of the above is a request for changes in this PR)

source/slang/slang-emit-cuda.cpp
```diff
@@ -349,6 +349,150 @@ void CUDASourceEmitter::emitLoopControlDecorationImpl(IRLoopControlDecoration* d
...
+
+void CUDASourceEmitter::_emitInitArrayValue(IRType* dstType, IRInst* value)
+{
+    // When constructing a matrix or vector from a single value this is already handled by the default path
+
+    switch (value->op)
+    {
+        case kIROp_Construct:
+        case kIROp_MakeMatrix:
+        case kIROp_makeVector:
```
tfoleyNV, line 407: Should `kIROp_makeStruct` be in here as well? It seems like initializer lists should also be used for `struct` types.

jsmall-nvidia: Can look into this and moving to CLikeSourceEmitter perhaps as another PR.

source/slang/slang-emit-cuda.cpp
jsmall-nvidia: Yes this would simplify things.

tfoleyNV:
> Such a transformation could be made, but doing so would require more code, and it was felt that clean up in terms of how this is represented in the IR at the point of emit, might be more appropriate, or at least discussed.

To add a note on this point, I think the Right Way to handle this in the long run is:

* We should have a more codified notion in the IR of instructions that represent "values." All literals are values, and `makeStruct|Array|Matrix|Vector` where the operands are values should also count as values.

* At the AST level we should have a notion of `constexpr` functions (or something akin to it), which are guaranteed to be evaluable at the IR level to produce values. We should define a base set of `constexpr` functions in the stdlib (the ones that our constant-folding can handle), and eventually allow users to define their own `constexpr` functions so long as they obey the right restrictions (e.g., only allowed to call other `constexpr` functions, can't read/write globals, etc.)

* The definition of "compile-time constant" expressions in the front-end should expand to include any expression where all the calls/operators resolve to calling `constexpr` functions, such that the whole thing can be evaluated to a value in the IR (though perhaps not earlier).

* Finally, the initial-value expression for any `static const` expression should be limited to a "compile-time constant" expression.

If we do a reasonable vertical slice of the above, then we should be able to guarantee that by the time we need to emit source code from the IR, we have fully evaluated any calls that might appear as part of a `static const` initial-value expression, so that the result is an IR value, and the `_emitInitializerList` logic should be able to handle any IR value.

PR #1323, opened 2020-04-16
Add support for global shader parameters to OptiX path

tfoleyNV:
There are two main pieces here.

First, we specialize the code generaiton for CUDA kernels to account for the way that shader parameters are passed differently for ordinary compute kernels vs. ray-tracing kernels. Both global and entry-point shader parameters in Slang are translated to kernel function parameters for CUDA compute kernels, while for OptiX ray tracing kernels we need to use a global `__constant__` variable for the global parameters, and the SBT data (accessed via an OptiX API function) for entry-point shader parameters.

This choice bakes in a few pieces of policy when it comes to how Slang ray-tracing shaders translate to OptiX:

* It fixes the name used for the global `__constant__` variable for global shader parameters to be `SLANG_globalParams`. Since that name has to be specified when creating a pipeline with the OptiX API, the choice of name effectively becomes an ABI contract for Slang's code generation.

* It fixes the choice that global parameters in Slang map to per-launch parameters in OptiX, and entry-point parameters in Slang map to SBT-backed parameters in OptiX. This is a reasonable policy, and it is also one that we are likely to need to codify for Vulkan as well, but it is always a bit unfortunate to bake policy choices like this into the compiler (especially when shaders compiled for D3D can often decouple the form of their HLSL/Slang code from how things are bound in the API).

The second piece is a lot of refactoring of the logic in `render-test/cuda/cuda-compute-util.cpp`, so that the logic for setting up (and reading back) the buffers of parameter data can be shared between the compute and ray-tracing paths. The result may not be a true global optimum for how the code is organized, but it at least serves the goal of not duplicating the parameter-binding logic between compute and ray-tracing.


tools/render-test/cuda/cuda-compute-util.cpp
```diff
@@ -964,514 +988,611 @@ static SlangResult _compute(CUcontext context, CUmodule module, const ShaderComp
...
-                                // On CPU we just use the memory in the BindSet buffer, so don't need to create anything
-                                RefPtr<MemoryCUDAResource> resource = new MemoryCUDAResource;
-                                SLANG_CUDA_RETURN_ON_FAIL(cuMemAlloc(&resource->m_cudaMemory, value->m_sizeInBytes));
-                                value->m_target = resource;
-                                break;
-                            }
+                        case SLANG_BYTE_ADDRESS_BUFFER:
+                        case SLANG_STRUCTURED_BUFFER:
+                        {
+                            // On CPU we just use the memory in the BindSet buffer, so don't need to create anything
```
jsmall-nvidia, line 1379: Looks like comment is out of date.

tfoleyNV: Looks like it, yeah. That is code I copy-pasted during the refactoring, so I didn't even look at the logic inside the function.

I'm inclined to leave the stale comment in this PR so the diffs can in theory be more pure (not that git/GitHub diff is able to figure out the edits).

PR #1325, opened 2020-04-17
Feature/wave mask review

jsmall-nvidia:
* Improve target-compatibility.md doc around Wave intrinsic
* Started wave-intrinsics document 
* WaveActiveMask -> WaveConvergedMask
* WaveMaskSync -> AllMemoryBarrierWithWaveMaskSync
* WaveMaskSharedSync -> GroupMemoryBarrierWithWaveMaskSync
* Add more WaveMask intrinsics
* Improve wave.slang test

This PR doesn't cover all aspects of the review, other/s to follow.


docs/wave-intrinsics.md
```diff
@@ -0,0 +1,74 @@
+Wave Intrinsics
+===============
+
+Slang has support for Wave intrinsics introduced to HLSL in SM6.0 and SM6.5. All intrinsics are available on D3D12, and a subset on Vulkan. On CUDA 'WaveMask' intrinsics are introduced which map more directly to the CUDA model of requiring a `mask` of participating lanes. On D3D12 and Vulkan the WaveMask instrinsics can be used, but the mask is effectively ignored. For this to work across targets including CUDA, the mask must be calculated such that it exactly matches that of HLSL defined 'active' lanes, else the behavior is undefined. 
+
+Another wrinkle in compatibility is that on GLSL targets such as Vulkan, the is not built in language support for Matrix versions of Wave intrinsics. Currently this means that Matrix is not a supported type for Wave intrinsics on Vulkan, but may be in the future.
```
tfoleyNV, line 6: Naively it seems like it would be semantically correct to just generate the matrix versions of `Wave*()` operations by iterating row-wise as we already do for many emulated matrix ops. Is there a particular concern there, other than the hypothetical cost of doing more warp-level syncs than strictly necessary?

docs/wave-intrinsics.md
```diff
@@ -0,0 +1,74 @@
...
+
+Most of the `WaveMask` functions are identical to the regular Wave intrinsics, but they take a WaveMask as the first parameter, and the intrinsic name starts with `WaveMask`. 
+
+```
+WaveMask GetConvergedMask();
+```
+
+Gets the mask of lanes which are converged within the Wave. Note that this is *not* the same as Active threads, and may be some subset of that. It is equivalent to the `__activemask()` in CUDA.
+
+On non CUDA targets the the function will return all lanes as active - even though this is not the case. This is 'ok' in so far as on non CUDA targets the mask is ignored. It is *not* okay if the code uses the value other than as a superset of the 'really converged' lanes. For example testing the bit's and changing behavior would likely not work correctly on non CUDA targets.
```
tfoleyNV, line 59: I don't think this is necessary, because it can return `WaveActiveBallot(true)` which should be exactly the HLSL/GLSL "active" mask.

docs/wave-intrinsics.md
```diff
@@ -0,0 +1,74 @@
...
+
+In the future Slang may have the capability to work out the masks required such that the regular HLSL Wave intrinsics work. As it stands there does not appear to be any way to implement the regular Wave intrinsics directly. To work around this problem we introduce 'WaveMask' intrinsics, which are essentially the same as the regular HLSL Wave intrinsics with the first parameter as the WaveMask which identifies the participating lanes. 
+
+The WaveMask intrinsics will work across targets, but *only* if on CUDA targets the mask captures exactly the same lanes as the 'Active' lanes concept in HLSL. If the masks deviate then the behavior is undefined. On non CUDA based targets currently the mask is ignored. This behavior may change on GLSL which has an extension to support a more CUDA like behavior.  
+
+Most of the `WaveMask` functions are identical to the regular Wave intrinsics, but they take a WaveMask as the first parameter, and the intrinsic name starts with `WaveMask`.
```
tfoleyNV, line 51: It would be nice to have a before/after example of some of the paired `WaveActive*()` and `WaveMask*()` functions, so that users can get a more visual example of what this means.

source/slang/hlsl.meta.slang
```diff
@@ -2484,7 +2484,7 @@ matrix<T, N, M> trunc(matrix<T, N, M> x)
 typedef uint WaveMask;
 
 __target_intrinsic(cuda, "__activemask()")
-WaveMask WaveGetActiveMask() { return 0xffffffff; }
+WaveMask WaveGetConvergedMask() { return 0xffffffff; }
```
tfoleyNV, line 2487: I'm happy with the rename here. My two questions:

* Do we want to *also* have a `WaveGetActiveMask()` that returns `WaveActiveBallot(true)`? It seems like that would be useful as a parallel accessor, even if it isn't strictly necessary (since HLSL chose to leave out a getter for the active mask, despite it being computable)

* Do we want to make `WaveGetConvergedMask()` return something more "accurate" on targets that don't support masks? I suspect that having it return the same value as `WaveGetActiveMask()` would on targets without more flexible scheduling would be the safest option. In particular it seems like it would be nice to have an invariant that the "converged" mask seen by a thread is always a subset of the "active" mask.

tests/hlsl-intrinsic/wave-mask/wave.slang
```diff
@@ -37,25 +37,34 @@ int exclusivePrefixSum(WaveMask mask, int index, int waveLaneId, int originalVal
...
     
     return 0;    
 }
 
+// It matters how kernels with WaveMask intrinsics are launched(!). 
+// TODO(JS):
+// If I launch with an numthreads amount that is not the size of the Wave on the device, then some 
+// lanes will not be executing at startup, and the kernel will have to know that is the case. 
+// This works currently though because the mask is only used
+// on CUDA, and it's Wave size is 32.
```
tfoleyNV, line 51: We should probably figure out what the best practices would be for developers to compute the "initial" active mask at the start of their program. My hope would be that your `~WaveMask(0)` idiom, or something like it would be something we would *guarantee* will always work.

PR #1327, opened 2020-04-20
WaveMask remaining intrinsics and tests

jsmall-nvidia:
On top of #1325 

* Added missing WaveMask intrinsics (mainly the analogous to SM6.5 additions)
* Added WaveMask tests that test the same surface broadly as wave- tests in hlsl-intrinsics directory


docs/wave-intrinsics.md
```diff
@@ -5,6 +5,110 @@ Slang has support for Wave intrinsics introduced to HLSL in SM6.0 and SM6.5. All
...
+
+// Barriers 
+
+void AllMemoryBarrierWithWaveSync();
+
+void GroupMemoryBarrierWithWaveSync();
+
+// Across Lanes
+
+__generic<T : __BuiltinIntegerType>
```
tfoleyNV, line 38: We probably ought to scrub these declarations so that they don't use `__generic` in our docs:

```
T WaveActiveBitAnd<T>(T expr);
```

I'm on the fence about whether we should just write `<T>` (which is a little bit misleading) or `<T : __BuiltinIntegerType>`, which is accurate but exposes the `__BuiltinIntegerType` interface which is currently not really intended for the eyes of users. It seems like it might be best to go with `<T : __BuiltinIntegerType>` but to let this light a fire under us to rename those built-in `interface`s to something we are ready to have users see and think about.

docs/wave-intrinsics.md
```diff
@@ -70,5 +177,93 @@ void GroupMemoryBarrierWithWaveMaskSync(WaveMask mask);
...
+// Barriers
+
+void AllMemoryBarrierWithWaveMaskSync(WaveMask mask);
+
+void GroupMemoryBarrierWithWaveMaskSync(WaveMask mask);
+
+// Across lane ops
+
+__generic<T : __BuiltinIntegerType>
+T WaveMaskBitAnd(WaveMask mask, T expr);
```
tfoleyNV, line 211: Same basic point in this file.

PR #1328, opened 2020-04-20
Small Improvements around Wave Intrinsics

jsmall-nvidia:
On top of #1327

* Improved documentation
* Added WaveGetCoverged/Active Mask/Multi
* Added tests


docs/wave-intrinsics.md
```diff
@@ -1,14 +1,231 @@
...
-T can be scalar, vector or matrix, except on Vulkan which doesn't support Matrix.
+There are three styles of wave intrinsics...
+
+## WaveActive
+
+The majority of 'regular' HLSL Wave intrinsics which operate on implicit 'active' lanes. 
+
+In the [DXC Wiki](https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics) active lanes are described as
+
+"These intrinsics are dependent on active lanes and therefore flow control. In the model of this document, implementations
```
tfoleyNV, line 18: Note on Markdown syntax. You can write something with a leading `>`:

```
> Like this.
```

and it will render as an HTML block quote:

> Like this.

tools/render-test/cuda/cuda-compute-util.cpp
```diff
@@ -1550,15 +1550,19 @@ SlangResult _loadAndInvokeKernel(
...
-    default:
-        return SLANG_FAIL;
-
     case PipelineType::Compute:
         return _loadAndInvokeComputeProgram(cudaContext, cudaStream, outputAndLayout, dispatchSize, outContext);
 
     case PipelineType::RayTracing:
+#ifdef RENDER_TEST_OPTIX
         return _loadAndInvokeRayTracingProgram(cudaContext, cudaStream, outputAndLayout, dispatchSize, outContext);
+#endif
```
tfoleyNV, line 1559: Good catch, thanks.

PR #1329, opened 2020-04-21
Fix for a generic definition, followed by a declaration with target intrisic causing a crash

jsmall-nvidia:
* Make a 'definition' if a function has a body or a target intrinsic defined
* Added test for this situation.


source/slang/slang-check-decl.cpp
```diff
@@ -2555,7 +2572,7 @@ namespace Slang
         // If both of the declarations have a body, then there
         // is trouble, because we wouldn't know which one to
         // use during code generation.
-        if (newDecl->Body && oldDecl->Body)
+        if (_isDefinition(newDecl) && _isDefinition(oldDecl))
```
tfoleyNV, line 2575: I was going to comment that this change to the rules should create problems in principle because we actually *should* allow multiple declarations with `__target_intrinsic` so long as they are for distinct targets, but then I noticed a comment shortly before this code:

```
        // TODO: if the two declarations are specialized for
        // different targets, then skip the body checks below.
        //
        // ???: Why isn't this problem showing up in practice?
```

It seems like we are already missing the logic that I'm saying it needed, and if/when we add it it seems like you are correct that the "is this a definition" logic should probably consider `__target_intrinsic` like a function body.

The one big gotcha here (and it has nothing to do with this PR) is that we still have an annoying implicit rule that a function declaration with no body gets treated more or less like a `__target_intrinsic` without us having to opt in. That rule is pretty much broken, and we are maybe at the point where we should remove it and deal with the fallout (which is mostly just adding `__target_intrinsic(hlsl)` to a bunch of stdlib declarations).

PR #1333, opened 2020-04-23
Small improvements around atomics

jsmall-nvidia:
* Improve target-compatibility.md around atomics
* Enable atomics test for targets where appropriate
* Added Buffer and RWBuffer to C++ prelude
* Use originalValue in test coverage.


docs/target-compatibility.md
```diff
@@ -161,3 +162,10 @@ On GLSL and VK targets loop unrolling uses the [GL_EXT_control_flow_attributes](
 
 Slang does have a cross target mechanism to [unroll loops](language-reference/06-statements.md), in the section `Compile-Time For Statement`.
 
+## Atomics on RWBuffer
```
tfoleyNV, line 165: Are we sure that atomics on `RWBuffer` (as opposed to `RWStructuredBuffer` or `RWByteAddressBuffer`) are legal at all?

The `RWBuffer` type is a "formatted" resource, like a texture, and as such I wouldn't expect it to support atomic operations even when it is mutable (because the atomics would potentially have to perform format conversion).

tests/compute/atomics-buffer.slang
```diff
@@ -2,6 +2,13 @@
 
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx12
```
tfoleyNV, line 4: I guess this test proves me wrong about `RWBuffer` and atomics, but I'm honestly very confused by that. I wonder what the restrictions are in D3D to make that workable.

PR #1334, opened 2020-04-24
Add support for generic load/store on byte-addressed buffers

tfoleyNV:
Introduction
============

The HLSL `*ByteAddressBuffer` types originaly only supported loading/storing `uint` values or vectors of the same, using `Load`/`Load2`/`Load3`/`Load4` or `Store`/`Store2`/`Store3`/`Store4`. More recent versions of dxc have added support for generic `Load<T>` and `Store<T>`, which adds a two main pieces of functionality for users.

The first and more fundamental feature is that `T` can be a type that isn't 32 bits in size (or a vector with elements of such a type), thus exposing a capability that is difficult or impossible to emulate on top of 32-bit load/store (depending on what guarantees `*StructuredBuffer` makes about the atomicity of loads/stores).

The secondary benefit of having a generic `Load<T>` and `Store<T>` is that it becomes possible to load/store types like `float` without manual bit-casting, and also becomes possible to load/store `struct` types so long as all the fields are loadable/storable.

This change adds generic `Load<T>` and `Store<T>` to the Slang standard library definition of byte-address buffers, and tries to bring those same benefits to as many targets as possible. In particular, the secondary benefits become available on all targets, including DXBC: byte-address buffers can be used to directly load/store types other than `uint`, including user-defined `struct` types, so long as all of the fields of those types can be loaded/stored.

The ability to load/store non-32-bit types depends on target capabilities, and so is only available where direct support for those types is available. For 16-bit types like `half` this includes both Vulkan and D3D12 DXIL with appropriate extensions or shader models.

The implementation is somewhat involved, so I will try to explain the pieces here.

Standard Library
================

The changes to the Slang standard library in `hlsl.meta.slang` are pretty simple. We add new `Load<T>` and `Store<T>` generic methods to `*ByteAddressBuffer`, and route them through to a new IR opcode.

Right now the generic `Load<T>` and `Store<T>` do *not* place any constraints on the type `T`, although in practice they should only work when `T` is a fixed-size type that only contains "first class"
uniform/ordinary data (so no resources, unless the target makes resource types first class). Our front-end checking cannot currently represent first-class-ness and validate it (nor can it represent fixed-size-ness), so these gaps will have to do for now.

Rather than directly translate `Load<T>` or `Store<T>` calls into a single instruction, we instead bottleneck them through internal-use-only subroutines. The design choice here is intended to ensure that for some large user-defined type like `MassiveMaterialStruct` we only emit code for loading all of its fields *once* in the output HLSL/GLSL rather than once per load site. While downstream compilers are likely to inline all of this logic anyway, we are doing what we can to avoid generating bloated code.

Emit and C++/CUDA
=================

Over in `slang-emit-c-like.cpp` we translate the new ops into output code in a straightforward way. A call like `obj.Load<Foo>(offset)` will eventually output as a call like `obj.Load<Foo>(offset)` in the generated code, by default.

For the CPU C++ and CUDA C++ codegen paths, this is enough to make a workable implementation, and we add suitable templated `Load<T>` and `Store<T>` declarations to the prelude for those targets.

Legalization
============

For targets like DXBC and GLSL there is no way to emit a load operation for an aggregate type like a `struct`, so we introduce a legalization pass on the IR that will translate our byte-address-buffer load/store ops into multiple ops that are legal for the target.

Scalarization
-------------

The big picture here is easy enough to understand: when we see a load of a `struct` type from a byte-address buffer, we translate that into loads for each of the fields, and then assemble a new `struct` value from the results. We do similar things for arrays, matrices, and optionally for vectors (depending on the target).

Bit Casting
-----------

After scalarization alone, we might have a load of a `float` or a `float3` that isn't legal for D3D11/DXBC, but that *would* be legal if we just loaded a `uint` or `uint3` and then bit-casted it. The legalization pass thus includes an option to allow for loads/stores to be translated to operate on a same-size unsigned integer type and then to bit-cast.

To make this work actually usable, I had to add some more details to the implementation of the bit-cast op during HLSL emit and, more importantly, I had to customize the way that the byte-address buffer load/store ops get emitted to HLSL so that it prefers to use the existing operations like `Load`/`Load2`/`Load3`/`Load4` instead of the generic one, whenever operating on `uint`s or vectors of `uint`.

Translation to Structured Buffers
---------------------------------

Even after scalarizing all byte-address-buffer loads/stores, we still have a problem for GLSL targets, because a single global `buffer` declaration used to back a byte-address buffer can only have a single element type (currently always `uint`), so the granularity of loads/stores it can express is fixed at declaration time. If we want to load a `half` from a byte-address buffer, we need a dedicated `buffer` declaration in the output GLSL with an element type of `half`.

The solution we employ here is to translate all byte-address buffer loads into "equivalent" structured-buffer ops when targetting GLSL. We add logic to find the underlying global shader parameter that was used for a load/store and introduce a new structured-buffer parameter with the desired element type (e.g., `half`) and then rewrite the load/store op to use that buffer instead. We copy layout information from the original buffer to the new one, so that in the output GLSL all the various `buffer`s will use a single `binding` and thus alias "for free."

We don't want to create a new global buffer for every load/store, so we try to cache these "equivalent" structured buffers as best as we can. For the caching I ended up needing a pair to use as a key, so I tweaked the `KeyValuePair<K,V>` type in `core` so that it could actually work for that purpose.

Because we are working at the level of IR instructions instead of stdlib functions at this work I had to add new IR opcodes to represent structured-buffer load/store that only (currently) apply to GLSL.

Layout
======

In order to translate a load/store of a `struct` type into per-field load/store we need a way to access layout information for the types of the fields. Previously layout information has been an AST-level concern that then gets passed down to the IR only when needed and only on global parameters, so layout information isn't always available in cases like this, at the actual load/store point.

As an expedient move for now I've introduced a dedicated module that does IR-level layout and caches its results on the IR types themselves. This approach *only* supports the "natural" layout of a type, and thus is usable for structured buffers and byte-address buffers (or general pointer load/store on targets that support it), but which is *not* usable for things like constant buffer layout.

We've known for a while that the Right Way to do layout going forward is to have an IR-based layout system, and this could either be seen as a first step toward it, or else as a gross short-term hack. YMMV.

Details
=======

The GLSL "extension tracker" stuff around type support needed to be tweaked to recognize that types like `int16_t` aren't actually available by default. I switched it from using a "black list" of unavailable types at initialization time over to using a "white list" of types that are known to always be available without any extensions.

Tests
=====

There are two tests checked in here: one for the basic case of a `struct` type that has fields that should all be natively loadable, and one that stresses 16-bit types. Each test uses both load and store operations.

Future Directions
=================

Right now we translate vector load/store to GLSL as load/store of individual scalars, which means the assumed alignment is just that of the scalars (consistent with HLSL byte-address buffer rules). We could conceivably introduce some controls to allow outputting the vector load/store ops more directly to GLSL (e.g., declaring a `buffer` of `float4`s), which might enable more efficient load/store based on the alignment rules for `buffer`s.

The IR layout work has a number of rough edges, but the most worrying is probably the assumption that all matrices are laid out in row-major order. Slang really needs an overhaul of its handling of matrices and matrix layout, so I don't know if we can do much better in the near term.

At some point the IR-based layout system needs to be reconciled with our current AST-base layout, and we need to figure out how "natural" layout and the currently computed layouts co-exist (in particular, we need to make sure that the IR-based layout and the existing layout logic for structured buffers will agree). This probably needs to come along once we have moved the core layout logic to operate on IR types instead of AST types (a change we keep talking about).

As part of this work I had to touch the implementation of bit-casting for HLSL, and it seems like that logic has some serious gaps. We really ought to consider a separate legalization pass that can turn IR bitcast instructions into the separate ops that a target actually supports so that we can implement `uint64_t`<->`double` and other conersions that are technically achievable, but which are hard to express in HLSL today.


source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -0,0 +1,924 @@
...
+    {
+        m_sharedBuilder.session = m_session;
+        m_sharedBuilder.module = module;
+
+        m_builder.sharedBuilder = &m_sharedBuilder;
+
+        processInstRec(module->getModuleInst());
+    }
+
+    // We recursively walk the entire IR structure (except
```
jsmall-nvidia, line 49: This isn't something for this PR, but based on something we discussed earlier - the appeal of traversing IR with recursive traversals of the entire module tree. It's very appealing because it's so simple.

That an advantage to handling off traversal to some kind of 'traversal manager' even with the homogeneity that IR gives us, is that it can achieve two things (depending on the type of transformation).

1) Could allow a single traversal to do multiple (typically simple) transformations
2) Can use knowledge about the structure to speed traversal - if we are looking for a node type that we know can't exist other node types, we can just skip that
3) Could hold a cache that could speed other kinds of searches/transitions

How much this helps in practice is arguable, and you could imagine you might want more than one style of traversal manager, and the baked in knowledge could be quite esoteric.

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -0,0 +1,924 @@
...
+            // it would complicate the generated code, and also
+            // make the rules about when a type is supported
+            // for byte-address load/store much more complicated.
+
+            // We collect the loaded per-field values into an
+            // array, which we will then use to construct the
+            // full value of the `struct` type.
+            //
+            List<IRInst*> fieldVals;
+            for( auto field : structType->getFields() )
```
jsmall-nvidia, line 222: That I'm assuming we don't need to 'flatten' the structure, because it has already been flattened at this point?

jsmall-nvidia:
Something I wasn't totally clear on was on glsl structured buffer/buffer comment 

"We copy layout information from the original buffer to the new one, so that in the output GLSL all the various buffers will use a single binding and thus alias "for free.""

I can bind to different types of buffer to the same binding?

tests/hlsl-intrinsic/byte-address-buffer/byte-address-16bit.slang
```diff
@@ -0,0 +1,51 @@
...
+struct Data
+{
+	int16_t	a;
+	int16_t b;
+}
+
+int test(int val)
+{
+	int offsetX = val*4;
+	int offsetY = offsetX & 0xFF;
```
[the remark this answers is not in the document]
tfoleyNV: I probably should have added a comment about this.

The idea is that to a human it is clear that `offsetX` and `offsetY` will always be equal, so that the loads and stores that follow all touch the same memory, but to a downstream compiler that *won't* be clear, so it should be impossible for a compiler to optimize away a sequence like:

```
buffer[offsetX] = valueA;
valueB = buffer[offsetY];
```

and eliminate the load so you just have:

```
buffer[offsetX] = valueA;
valueB = valueA;
```

The point here is to ensure that all the operations we wrote in the function actually get executed.

In some similar codebases I've worked with in the past, tests would often have a constant buffer of inputs with names like `zero`, `one`, `two`, etc. that are always bound to the integer values the names imply. That means that you could write a test that uses `someVal + zero` and be sure the compiler couldn't optimize based on knowing that is equal to `someVal`.

tests/hlsl-intrinsic/byte-address-buffer/byte-address-16bit.slang
```diff
@@ -0,0 +1,51 @@
...
+//TEST(compute):COMPARE_COMPUTE_EX:-slang -cuda -compute
+
+// Unavailable on D3D and fxc-based targets, because fxc doesn't
+// support loading anything besides `uint` from a byte-address
+// buffer.
+//
+//TEST_DISABLED(compute):COMPARE_COMPUTE_EX:-slang -compute
+
+
+//TEST_INPUT:ubuffer(data=[0x00010002 0x00030004 0x00050006 0x00070008]):name=inputBuffer
```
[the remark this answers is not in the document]
tfoleyNV: That's a fair point. Name-based binding of test inputs should help avoid the possible issues there.

I'm not sure if we have any big-endian platforms that are likely to come up as a request any time soon.

source/slang/slang-ir-byte-address-legalize.cpp
tfoleyNV: I agree that it makes some sense to offload the main recursive walk to subroutine, whether or not we decide to make the kinds of improvements you are talking about. Repeating that code over and over is a waste of time and energy, really.

For my part I'd be interested to see such a change comes a part of codifying the notion of a "pass" in the IR, so that things like optimization first construct an explicit list of `Pass` objects to apply, and then a central execution engine goes through and applies those passes.

The execution engine could then own things like:

* Running validation before/after passes
* Dumping of IR before/after passes (including controls for picking where in the process to dump)
* Ganging together passes that can share the recursive walk over the tree (as you describe)
* Parallelizing passes over functions, where that is possible
* Computing and then re-using side-band structures like the dominator tree across passes when intermediate passes do not invalidate it

All of that is squarely in the "would be nice" pile rather than the "must have" pile right now.

tfoleyNV:
> I can bind to different types of buffer to the same binding?

Sorta kinda. There's a bunch of distinct buffer types in HLSL that all map to the same kind of `buffer` declaration in the output GLSL, including all of the `*StructuredBuffer` and the `*ByteAddressBuffer` types. From a GLSL language and Vulkan API standpoint those are all "shader storage buffers."

The Vulkan SPIR-V rules allow for multiple global shader parameters that have the same `binding`, and it is okay to access multiple parameters that overlap like that so long as the type of the parameter in GLSL/SPIR-V is consistent with the descriptor that actually got bound via the API.

What we are doing in the Slang IR here is a little slippery in terms of layout, in that the layout information for a structured buffer might have extra data about the layout of the elements, which a layout for byte-address buffer would lack, but this pass occurs late enough in the compiler that no code would look for or use that additional information (and I'm not even sure we represent that information in the IR layout representation today).

PR #1335, opened 2020-04-30
Improve GLSL coverage of boolean binary ops

tfoleyNV:
This change ensures that the `&&`, `||`, `&`, `|`, and `^` apply correctly to vectors of `bool` values when targetting GLSL.

Most of the changes are in the GLSL emit path, where the IR instructions for these operators are bottlenecked through a small set of helper routines to cover the different cases. In general:

* The vector variants of the operations are implemented by casting to `uint` vectors, performing bitwise ops, then casting back
* The scalar variants are handled by conveting the bitwise operations to their equivalent logical operator (the one interesting case there is bitwise `^` where the equivalent logical operation on `bool` is `!=`)

This change makes it clear that our IR really shouldn't have distinct opcodes for logical vs. bitwise and/or/xor, and instead should just have a single family of operations where the behavior differs based on the type of the operand. That is already *de facto* the way things work (a user can always write `&`, `|` and `^` and expect them to work on `bool` and vectors of `bool`), so that the GLSL output path has to deal with the overlap. Having two sets of IR ops here actually makes for more code instead of less.


source/slang/slang-emit-glsl.cpp
```diff
@@ -1171,32 +1261,11 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
-                m_writer->emit(prec.op);
-                emitOperand(inst->getOperand(1), rightSide(outerPrec, prec));
-
-                maybeCloseParens(needClose);
-                return true;
-            }
-            break;
-        }
+            return _tryEmitBitBinOp(inst, getInfo(EmitOp::BitOr), getInfo(EmitOp::Or), inOuterPrec);
+        case kIROp_BitXor:
```
jsmall-nvidia, line 1267: I know this is covered in you write up on the PR - but could do with a comment, as it is surprising.

tfoleyNV: Good point. I will go back and comment this case, as well as document the declarations of the helper functions.

tfoleyNV:
> Does this mean ~ and ! are ok for vectors?

I don't know for sure. It looks like we correctly translate a logical not on a boolean vector into a call to `not(...)`, but I don't know about the `~` case.

I can try to tweak the test case in this PR to cover negation and then see what the results are. I might as well fix those ones (if they are broken) at the same time.

tfoleyNV:
Follow-up: The `!` operator is already working on vectors of `bool` in GLSL output, and I modified the test to cover that case.

The `~` operator on vectors of `bool` (and on scalar `bool` as well) has surprising behavior on the fxc and dxc compilers, where it promotes the operand to `int` and does the bitwise negation there. As a result `(bool)(~someBool)` is *always* true on those compilers.

(I noew suspect that the reality is that operators like `&` and `|` only work on `bool` vectors in existing HLSL code because of implicit promotion to `int` followed by implicit demotion back to `bool`)

For now, Slang doesn't support `~` on a `bool` (or vector of `bool`) operand at all, and that might be the best option for now. In the long run I'd prefer it if the bitwise operations on `bool` operated on it as if it were a one-bit unsigned integer type, without intermediate promotion. Our current supported operations are consistent with that.

PR #1336, opened 2020-05-01
Make stdlib WaveActive* call WaveMask*

tfoleyNV:
This change makes the various `WaveActive*()` functions have default implementations that call `WaveMask*()` passing `WaveActiveMask()`. The new definitions will be used during CUDA code generation, which simplifies some of the duplication that was occuring in the `__target_intrinsic` modifiers.

This change does *not* add logic to make computation of `WaveGetActiveMask()` corect on CUDA, so these functions will still fail to provide the behavior that users need/expect. A future change will need to add logic to synthesize the value of `WaveGetActiveMask()` automatically.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2531,6 +2531,7 @@ __glsl_extension(GL_KHR_shader_subgroup_ballot)
 __spirv_version(1.3)
 __target_intrinsic(glsl, "subgroupBallot(true).x")
 __target_intrinsic(hlsl, "WaveActiveBallot(true).x")
+__target_intrinsic(cuda, "__activemask()") // Note: semantically incorrect, but best we can do for now.
 WaveMask WaveGetActiveMask();
 
 __glsl_extension(GL_KHR_shader_subgroup_basic)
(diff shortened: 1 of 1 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -2920,7 +2921,7 @@ matrix<T,N,M> WaveMaskReadLaneFirst(WaveMask mask, matrix<T,N,M> expr);
 __generic<T : __BuiltinType>
 __target_intrinsic(hlsl, "WaveMatch($1).x")
 __cuda_sm_version(7.0)
-__target_intrinsic(cuda, "_waveMatchScalar($0, $1)")
+__target_intrinsic(cuda, "_waveMatchScalar($0, $1).x")
```
jsmall-nvidia, line 2924: What's with adding the .x?

tfoleyNV: We maybe need to talk about this more and potentially have a follow-on change for some of the functions right below this in the file, but:

The stdlib `WaveMaskMatch` function was defined to return a `WaveMask`, but the CUDA prelude `_waveMatchScalar` function that was being used to implement it returned a `uint4`, so I got a downstream compilation error when I introduced a call to `WaveMaskMatch` inside the `WaveMatch` function.

It seems like some of these functions in the stdlib wouldn't actually work on CUDA if/when we call them, and we should survey the mask-returning functions to double-check this. I'd like to imagine that we could fold a fix like that into a change where we make `WaveMask` be a target-dependent type, such that all of these operations will take/return a `WaveMask` and the CUDA prelude never has to think about anything besides 32-bit masks.

PR #1337, opened 2020-05-01
C++ Extractor

jsmall-nvidia:
* C++ extractor  
* Improved Path functions 
* Ability to check if TokenReader is at cursor position
* Moved some diagnostic functions so as to separate Lexer from rest of Slang code

The majority of the work here is around the C++ extractor. The main motivation for this work is to change how AST nodes are defined and in particular to make them C++ classes, rather than a 'macro file' which is included multiple times, to generate different things. 

A reasonable chunk of the work was around converting the AST types such that they are just C++ classes - those changes are not part of this PR and will be part of a subsequent PR once it is in a place to replace the ASTs.

When AST nodes are defined as C++ classes we also need to be able to 

* Support reflection - including Super class, name and a create function
* Adding of visitor functions to classes, optionally
* Generate visitor types, for all of the derived types of a type

These were all generated by changing macros and multiply including files in clever ways previously.

A final goal here was to try and make the amount of 'markup' needed for a class minimal. In practice this typically means adding a 'SLANG_CLASS(name)' or 'SLANG_ABSTRACT_CLASS(name)' at the start of the classes definition.

So for example previously we might have
```
    // Resources that contain "elements" that can be fetched
    ABSTRACT_SYNTAX_CLASS(ResourceType, BuiltinType)
    // The type that results from fetching an element from this resource
    SYNTAX_FIELD(RefPtr<Type>, elementType)

    // Shape and access level information for this resource type
    FIELD(TextureFlavor, flavor)

    RAW(
        TextureFlavor::Shape GetBaseShape()
        {
            return flavor.GetBaseShape();
        }
        bool isMultisample() { return flavor.isMultisample(); }
        bool isArray() { return flavor.isArray(); }
        SlangResourceShape getShape() const { return flavor.getShape(); }
        SlangResourceAccess getAccess() { return flavor.getAccess(); }

    )
END_SYNTAX_CLASS()
```

Is now written as

```C++
// Resources that contain "elements" that can be fetched
class ResourceType : public BuiltinType 
{
    SLANG_ABSTRACT_CLASS(ResourceType)

    // The type that results from fetching an element from this resource
    RefPtr<Type> elementType;

    // Shape and access level information for this resource type
    TextureFlavor flavor;

    TextureFlavor::Shape GetBaseShape()
    {
        return flavor.GetBaseShape();
    }
    bool isMultisample() { return flavor.isMultisample(); }
    bool isArray() { return flavor.isArray(); }
    SlangResourceShape getShape() const { return flavor.getShape(); }
    SlangResourceAccess getAccess() { return flavor.getAccess(); }
};
```

That it is not necessary to mark up fields that the extractor, using some fairly simple heuristics can determine most types, including templated types and so determine what are fields. The underlying here was to not have to mark up all fields, but put markers such that all fields in-between will be marked so.  The field marking is not implemented here, because fields are not currently needed for the extractor.

The mechanism for marking a class for the extractor is by specifying the prefix and postfix text of an identifier. The default used here is SLANG_ and _CLASS. The value inbetween is recorded and reflected on the output and so can be used as appropriate.

The extractor currently generates two files - a header file and a 'macro' file. The header file is a standard C++ file that defines an enum that determines a types identity and such that testing if a type is derived is as simple as testing a range. For an 'inner' type, we need to know the end value of the range, and this is output in a second enum class. Finally we want to be able to determine the Super class automatically. To do this a struct is created with the super class to class x is a typedef of x_Super.

Some examples

```C++
enum class ASTNodeType
{
    NodeBase = 0,
        SyntaxNodeBase = 1,
            SyntaxNode = 2,
                Modifier = 3,
                    InModifier = 4,
                    OutModifier = 5,
                        InOutModifier = 6,
                    ConstModifier = 7,
                    InstanceModifier = 8,
```

Last values

```C++
enum class ASTNodeLast
{
    NodeBase = int(ASTNodeType::TaggedUnionSubtypeWitness),
        SyntaxNodeBase = int(ASTNodeType::ThisTypeExpr),
            SyntaxNode = int(ASTNodeType::ThisTypeExpr),
                Modifier = int(ASTNodeType::ImplicitConversionModifier),
                    InModifier = int(ASTNodeType::InModifier),
                    OutModifier = int(ASTNodeType::InOutModifier),
                        InOutModifier = int(ASTNodeType::InOutModifier),
                    ConstModifier = int(ASTNodeType::ConstModifier),
                    InstanceModifier = int(ASTNodeType::InstanceModifier),
                    BuiltinModifier = int(ASTNodeType::BuiltinModifier),
                    InlineModifier = int(ASTNodeType::InlineModifier),
```

Super types

```C++
struct ASTNodeSuper
{
    typedef RefObject NodeBase_Super;
    typedef NodeBase SyntaxNodeBase_Super;
    typedef SyntaxNodeBase SyntaxNode_Super;
    typedef SyntaxNode Modifier_Super;
    typedef Modifier InModifier_Super;
    typedef Modifier OutModifier_Super;
    typedef OutModifier InOutModifier_Super;
    typedef Modifier ConstModifier_Super;
    typedef Modifier InstanceModifier_Super;
    typedef Modifier BuiltinModifier_Super;
    typedef Modifier InlineModifier_Super;
    typedef Modifier PublicModifier_Super;
    typedef Modifier RequireModifier_Super;
    typedef Modifier ParamModifier_Super;
...
```

The macro file looks like

```
#define SLANG_DERIVED_ASTNode_NodeBase(x, param) \
    SLANG_ASTNode_SyntaxNodeBase(x, param) \
    SLANG_ASTNode_Val(x, param) \
/* */
#define SLANG_ASTNode_NodeBase(x, param) \
    x(NodeBase, ABSTRACT, BASE, param) \
    SLANG_DERIVED_ASTNode_NodeBase(x, param) 

#define SLANG_DERIVED_ASTNode_SyntaxNodeBase(x, param) \
    SLANG_ASTNode_SyntaxNode(x, param) \
/* */
#define SLANG_ASTNode_SyntaxNodeBase(x, param) \
    x(SyntaxNodeBase, ABSTRACT, INNER, param) \
    SLANG_DERIVED_ASTNode_SyntaxNodeBase(x, param) 

#define SLANG_DERIVED_ASTNode_SyntaxNode(x, param) \
    SLANG_ASTNode_Modifier(x, param) \
    SLANG_ASTNode_ModifiableSyntaxNode(x, param) \
    SLANG_ASTNode_Expr(x, param) \
/* */
#define SLANG_ASTNode_SyntaxNode(x, param) \
    x(SyntaxNode, ABSTRACT, INNER, param) \
    SLANG_DERIVED_ASTNode_SyntaxNode(x, param) 

```

Note the split of SLANG_ASTNode, with SLANG_DERIVED_ASTNode_. This allows injecting code for just the derived types easily. 

Thus the classes can be generated via 

```
#define SLANG_CLASS_REFLECT(x) \
    public:     \
    static const ASTNodeType kType = ASTNodeType::x; \
    typedef ASTNodeSuper::x##_Super Super; \
    SLANG_FORCE_INLINE static bool isDerivedFrom(ASTNodeType type) { return int(type) >= int(kType) && int(ASTNodeLast::x) <= int(type); } \
    static const ReflectClassInfo kReflectClassInfo;

#define SLANG_ABSTRACT_CLASS(x) SLANG_CLASS_REFLECT(x)

#define SLANG_CLASS(x) SLANG_CLASS_REFLECT(x) \
     virtual void accept(x::Visitor* visitor, void* extra) override;  
```

To complete the types first include the generated header file, then define the macros, and finally include the class headers.

To generate all the implementations can be achieved through the macros file

```
template <typename T>
struct CreateImpl
{
    static void* create() { return new T; }
};

// For special handling of BASE types, INNER (ie not leaf) and LEAF types
// For base we could make point to a 'RefObject' type info if we wanted, here I just make nullptr
#define SLANG_GET_SUPER_BASE(x) nullptr
#define SLANG_GET_SUPER_INNER(x) &ASTNodeSuper::x##_Super::kReflectClassInfo
#define SLANG_GET_SUPER_LEAF(x) &ASTNodeSuper::x##_Super::kReflectClassInfo

// This is using some paste hackery. That something the abstract style (probly needs a better name) is the text between the marker prefix and postfx if nothing, it's just 'NORMAL' so SLANG_ABSTRACT_CLASS -> ABSTRACT and SLANG_CLASS -> NORMAL.

#define SLANG_GET_CREATE_FUNC_ABSTRACT(x) nullptr
#define SLANG_GET_CREATE_FUNC_NORMAL(x) &CreateImpl<x>::create

#define SLANG_REFLECT_CLASS_INFO(name, abstractStyle, nodeStyle, param) \
    /* static */const ReflectClassInfo name::kReflectClassInfo = { uint32_t(ASTNodeType::name), uint32_t(ASTNodeLast::name), SLANG_GET_SUPER_##nodeStyle(name), #name, SLANG_GET_CREATE_FUNC_##abstractStyle(name)  };

// Generates implementations for all the nodes. Has two invocations, because their common base class is 'RefObject' which we don't have injection on.
SLANG_ASTNode_NodeBase(SLANG_REFLECT_CLASS_INFO, _)
SLANG_ASTNode_Substitutions(SLANG_REFLECT_CLASS_INFO, _)
```

The x macro magic is 'nice' in so far as you can just produce code directly without multiple includes. If something goes wrong it might be a little more obtuse as to where than with the #include mechanism. The extractor could of course generate '-defs.h' files if desired, and could use the previous mechanism, without having to deal with those files by hand.

When this work was originally started, a lexer was written for the purpose. It only understood enough to extract the class information - not fields. When the lexer was improved to handle strings, this was cribbed from the Slang lexer. When field extraction was tried, the amount of implementation overlap became sort of ridiculous. So an attempt was then made to try and use the Slang lexer and see how separable everything was. As it turns out with only minor changes were required to make the extractor able to use just a few source files from the Slang compiler. 

Moreover using the slang lexer brings the DiagnosticSink mechanism, as well as the SourceManager/SourceLoc. Both of which make error reporting much simpler and more precise. It should also be easy to run the C preprocessor using slangs implementation, although that isn't tried here yet. Doing so would mean #if 0 and the like would work, as well as other macro tricks (although that's perhaps less desirable - as part of the point was to make the IDE understand these types).

Subsequent PRs will need to integrate this into build, and have AST nodes defined in Slang this way. Both should be relatively easy.


source/slang/slang-diagnostics.h
```diff
@@ -99,8 +99,6 @@ namespace Slang
     void printDiagnosticArg(StringBuilder& sb, TokenType tokenType);
     void printDiagnosticArg(StringBuilder& sb, Token const& token);
     void printDiagnosticArg(StringBuilder& sb, CodeGenTarget val);
-    void printDiagnosticArg(StringBuilder& sb, Stage val);
-    void printDiagnosticArg(StringBuilder& sb, ProfileVersion val);
```
tfoleyNV, line 103: Is there a reason for moving this logic, which is all about printing of information in diagnostic messages, out of the diagnostics system and over to the profile stuff?

jsmall-nvidia: Yes - it removed the slang-profile.cpp dependency from DiagnosticSink. 

Which admittedly was only desirable so as to make the amount of dependencies to the internals of Slang source minimal.

tools/slang-cpp-extractor/slang-cpp-extractor-diagnostic-defs.h
```diff
@@ -0,0 +1,45 @@
+//
+
+// The file is meant to be included multiple times, to produce different
```
tfoleyNV, line 3: Okay, I think I get it. You are re-using the existing diagnostics infrastructure for the extractor, which means that it is helpful to have the diagnostics system not have too many dependencies on the rest of the compiler (which relies on the output of the extractor to build).

It seems like we need to figure out a holistic policy for this kind of stuff where the Slang compiler proper needs a service that some of our utility/testing/build code may also need, but the version for the Slang compiler may benefit from a degree of customization for and integration with the specifics of the compiler. Right now there are three main strategies we've got going on:

1. Have two copies of some service (e.g., the way that there is a tokenizer/lexer in `core`, and then the real one in `slang`)
2. Move the Slang compiler functionality out to `core`, even in cases where it isn't a very "universal" piece of code (e.g., the downstream compiler abstraction), and bring along any pieces of Slang-specific infrastructure/types that it needs
3. This case, where a piece of the Slang compiler proper is set up to be usable without the rest of the compiler in place, so that another tool/system can use it.

I worry that none of those solutions are really ideal.

tools/slang-cpp-extractor/slang-cpp-extractor-main.cpp
```diff
@@ -0,0 +1,1944 @@
...
+    }
+    return false;
+}
+
+bool CPPExtractor::advanceIfStyle(IdentifierStyle style, Token* outToken)
+{
+    if (m_reader.peekTokenType() == TokenType::Identifier)
+    {
+        IdentifierStyle readStyle = m_identifierLookup->get(m_reader.peekToken().Content);
+        if (readStyle == style)
```
tfoleyNV, line 758: One note is that we *might* be able to avoid having a context-sensitive parse in the C++ extractor by having it rely on a combination of heuristic rules and convention to disambiguate the interesting cases. The worry with using a "proper" context-sensitive parse is that you can never really have all of the context without processing `#include` directives to discover other user-defined names.

tools/slang-cpp-extractor/slang-cpp-extractor-main.cpp
```diff
@@ -0,0 +1,1944 @@
...
+
+            out << node->m_name.Content << " = " << typeIndex << ",\n";
+
+            typeIndex++;
+        }
+
+        out << "};\n\n";
+
+        out << "\n";
+        out << "enum class " << reflectTypeName << "Last\n";
```
tfoleyNV, line 1723: This kind of stuff is where the "template engine" style of `slang-generate` could actually make for cleaner code, but as discussed already, entangling this new tool with the existing one would probably complicate things more than necessary for right now.

tfoleyNV:
This is looking great so far. Let me know when you want me to review it for real, but I think this kind of tooling should be a real boon for a lot of parts of our code.

One open question in my mind is whether we can/should fold other cases where we currently rely on `-defs.h` files into this system, even when they aren't directly related to `class`/`struct` declarations. E.g., there is a certain amount of "include the file multiple times" magic going on for things like profiles, base types, etc. that could maybe be cleaned up by generating that code rather than hand-writing it, but in that case there is no actual class hierarchy to use as the source of truth.

On the issue of X-macros vs `-defs` style files, I think I mostly prefer the latter because it is easier to maintain manually (which would be moot here), but also because they tend to make error messages a tiny bit easier to track down to the right line (as you note).

While I don't think it needs to be in the first version of things, I also thing that it would be nice if the extractor/generator tool had a provision for associating classes in a hierarchy (or their IDs) with additional type-level metadata (e.g., stuff like "is this IR opcode guaranteed to have no side effects?").

tfoleyNV:
One additional detailed thought that probably is too much "magic" for kind of system: rather than have the `SLANG_CLASS(Foo)` macro always expand to the same thing for every class, you can in theory have a system where:

* Every file `foo-bar.h` to be processed always includes `foo-bar.generated.h` which has all the generated macros

* Every class `Foo` in `foo-bar.h` always includes a macro invocation `SLANG_GENERATED_CLASS_BODY()` (or similar) in its body

* The `SLANG_GENERATED_CLASS_BODY()` macro expands to something like `SLANG_GENERATED_CLASS_BODY_##__LINE__`, so each invocation expands to a distinct macro invocation

* The `foo-bar.generated.h` file defines each `SLANG_GENERATED_CLASS_BODY_<NNN>` to be the desired declarations to introduce for the given class, which can differ based on the class and need not have any circumlocutions involved in identifying the super class, dealing with abstract vs. concrete classes, etc.

The main gotchas with that approach are that you need to be confident that all compilers will agree on the `__LINE__` value within the macro expansions, and then you also need to make sure that the include order set up for the `.generated.h` files isn't one that risks collisions.

It can be nice working with that style, though, because you can in principle synthesize things like accessor functions, constructors, etc. on a per-class basis.

PR #1340, opened 2020-05-05
Enhanced C++ extractor

jsmall-nvidia:
In the process of trying to replace -defs.h for defining AST nodes a few problems came up. With the defs mechanism types in different files could be dealt with in different ways - as macros could be changed and then only some files included. The extractor had something to deal with this with the 'marker', but using would require all of the types in a file have different marks set on them.

The second problem was that only limited information about the type was available at the class definition, which limited what kind of injection could be performed at that point.

The observation here is there is a need for a mechanism to allow arbitrary injection, based on extracted information at the class declaration. The trick used here is that the the class name as defined in the SLANG_CLASS line is a unique identifier. So the extractor generates a x-macro for each class name that holds the information. By adding to that macro the super class, and the last type, we no longer need the super typedef, or the Last enum.

The -macro file is changed to hold 'CHILDREN' and 'ALL' macros defined to just use the types definition macros. 

Some other small changes:

* Support for indexOf on UnownedStringSlice
* More work to split dependencies from DiagnosticSink/Lexer
* Each source file now creates a SourceOrigin, type that tracks nodes as they are defined in the file
* There is now an option to generate -defs.h files


tools/slang-cpp-extractor/slang-cpp-extractor-main.cpp
```diff
@@ -193,6 +195,10 @@ class Node : public RefObject
 
         /// True if has a derived type that is reflected
     bool hasReflectedDerivedType() const;
+        /// Stores in out any reflected derived types
+    void getReflectedDerivedTypes(List<Node*>& out) const;
+
+    static void filterReflectedClassLike(List<Node*>& io);
```
tfoleyNV, line 201: Nits:

Why not `outDerivedTypes` instead of just `out` for the `getReflectedDerivedTypes` parameter?

Similarly, could `io` be changed to something more descriptive? The function name/signature isn't very clear here (probably needs a comment). It took me a while to come to the conclusion that this operation probably takes as input the list of nodes `io`, and filters it down to just those nodes that represent "class-like" nodes, returning the resulting filtered list back in `io`.

PR #1342, opened 2020-05-11
Add GLSL translation for HLSL fmod()

tfoleyNV:
The existing code was assuming `fmod()` was available as a builtin in GLSL, which isn't true. It also isn't possible to translate the HLSL `fmod()` to the GLSL `mod()` since the two have slightly different semantics.

This change introduces a definition for `fmod(x,y)` that amounts to `x - y*trunc(x/y)` which should agree with the HLSL version except in corner cases (e.g., there are some cases where the HLSL version returns `-0` and this one will return `0`).


tfoleyNV:
> It's nice you can just define a default impl in std library like this now.

Yes, it feels a bit more like a "standard library" now and less like a meta-file that just tells the compiler what the intrinsics are.

One thing we need to decide for operations like `fmod` here is whether the vector/matrix versions should be defined in terms of the same formula as the scalar one, or if they should tend to be defined in terms of iteration over elements/rows.

PR #1344, opened 2020-05-11
WIP: Have ASTNodeType as set field in all AST types.

jsmall-nvidia:
Have ASTNodeType set on all AST nodes, such that casting can be performed without calling virtual function getClassInfo().

The mechanism used here is to set a default Ctor to set the field. If a class implements a non default constructor - they must also set the m_astType = kType in their constructor.

This mechanism is not ideal in that the m_astType might be set multiple times during construction. Really what we want is to set the type once, but there doesn't seem like a straight forward way to do that. Doing it through threading the type through constructors also seems fragile.

If we had say a 'create' function on the base class we could default construct and set the parameter. With perhaps variadic template, or a range of parameter sizes.

### DIFF
--- a/source/slang/slang-ast-base.h
+++ b/source/slang/slang-ast-base.h
@@ -26,8 +26,9 @@ struct ReflectClassInfo;
 class NodeBase : public RefObject
 {
     SLANG_ABSTRACT_CLASS(NodeBase)
-   
+
     SyntaxClass<NodeBase> getClass() { return SyntaxClass<NodeBase>(&getClassInfo()); }
+    ASTNodeType m_astType;
 };
 
 // Casting of NodeBase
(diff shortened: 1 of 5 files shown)


tfoleyNV:
> Really what we want is to set the type once, but there doesn't seem like a straight forward way to do that. Doing it through threading the type through constructors also seems fragile.

If I were writing this logic by hand I would do it by "threading the type through constructors," so I'm wondering what about that is fragile in our case.

> If we had say a 'create' function on the base class we could default construct and set the parameter. With perhaps variadic template, or a range of parameter sizes.

This definitely seems workable. It would mean auditing all the sites where we `new` an AST node and changing then to something like `astBuilder->create<SomeNodeType>(...)`, which could then easily set the appropriate tag based on `SomeNodeType::kTag`, etc.

(If we are leaning on code generation more heavily this could of course be `createSomeNodeType` instead of `create<SomeNodeType>`...)

I'm assuming that if we go down this road we'd want to allocate everything under some umbrella object like an "AST builder" or an "AST context" so that we can more easily track memory usage for now. In the longer run, the builder/context would own the pool allocator(s) being used, and could also be responsible for deduplication for stuff like `Type`s.

PR #1348, opened 2020-05-15
AST dumping via C++ Extractor reflection

jsmall-nvidia:
* Added fields to what's output from C++ extractor
* Added -dump-ast option to slang command line
* Fixed bug in StringSlicePool - needs to make sure the slice is held in the arena if it's in the map
* Added -output-fields option to C++ extractor
* Added 'SLANG_UNREFLECTED' and 'SLANG_REFLECTED' can be used to mark fields (and types) that aren't to be reflector or reflected. Work similarly to public:  etc applying to subsequent fields/types

When dump-ast is enabled will dump the ModuleDecl for each parsed file. Writes the file in current directory with the same name, but with slang-ast extension.

AST dumping supports a 'hierarchical' and 'flat' modes. Hierarchical dumps in a depth first manner. Flat dumps each object independently and then uses references to specify the relationship.

The output can be very large - as by default it's seems to end up dumping all of the standard library, so will need to look at more suitable restricting the output.


source/slang/slang-ast-dump.cpp
```diff
@@ -0,0 +1,615 @@
...
+
+#include "slang-compiler.h"
+
+#include "../core/slang-string.h"
+
+#include "slang-ast-generated-macro.h"
+
+// There could be an argument for using the SourceWriter, but whilst that has some useful stuff
+// like the automatic indentation, it is a little clumsy in some other ways, like how to build up
+// a result via << operators is not available.
```
tfoleyNV, line 13: Interesting point. It might be good to create a unified abstraction for outputting formatted text (especially around things like indention handling), and then put all the necessary convenience features into it.

source/slang/slang-ast-dump.cpp
```diff
@@ -0,0 +1,615 @@
...
+    void dump(const UnownedStringSlice& slice)
+    {
+        m_writer->emitChar('\"');
+
+        {
+            ScopeWrite scope(this);
+            auto& buf = scope.getBuf();
+            for (const char c : slice)
+            {
+                if (c < 0x20 || c >= 0x80)
```
tfoleyNV, line 158: Probably worth factoring into a helper subroutine just so this can be written as `if(!_isPrintable(c))` or similar.

PR #1349, opened 2020-05-19
Reduce the size of Token

jsmall-nvidia:
The 64 bits Token type was taking up 40 bytes per token, the changes here bring this down to 24. This was due to 3 reasons

* The token type/flags were each 32 bits by default
* Identifers can hold a pointer to a Name
* The lexeme is held as a UnownedStringSlice

To save space therefore we

* Use 8 bits for type/flags
* Have a union that either holds a Name* or chars* which is identified by a flag (rather than TokenType)
* We don't need to use a pointer sized value for the length of the lexeme, 32 bit is enough

The  current layout is still somewhat wasteful, with 6 bytes not being used on 64 bits.

```
struct Token
{
    TokenType   type = TokenType::Unknown;
    TokenFlags  flags = 0;
   SourceLoc   loc;
    uint32_t charsCount = 0;              ///< Amount of characters. Is set if name or not.
    void* charsOrName;
};
```

Unfortunately there doesn't seem like a straightforward way to improve on this for now, if we want a single structure.

### DIFF
--- a/source/slang/slang-check-shader.cpp
+++ b/source/slang/slang-check-shader.cpp
@@ -396,7 +396,7 @@ namespace Slang
                 {
                     const auto& semanticToken = semantic->name;
 
-                    String lowerName = String(semanticToken.Content).toLower();
+                    String lowerName = String(semanticToken.getContent()).toLower();
 
                     if(lowerName == "sv_dispatchthreadid")
                     {
--- a/source/slang/slang-diagnostics.cpp
+++ b/source/slang/slang-diagnostics.cpp
@@ -70,7 +70,7 @@ void printDiagnosticArg(StringBuilder& sb, TokenType tokenType)
 
 void printDiagnosticArg(StringBuilder& sb, Token const& token)
 {
-    sb << token.Content;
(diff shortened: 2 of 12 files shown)


tfoleyNV:
> Unfortunately there doesn't seem like a straightforward way to improve on this for now, if we want a single structure.

The only ideas I can think of would be more drastic than what you are proposing here:

* We could eliminate the `charsOrName` field entirely, and require looking up the text of a token to go through the `SourceLoc` to find the text in the source buffer. This would eliminate the potential speedup of directly referencing a `Name*`, and would require a workaround for any tokens that needed to be "scrubbed" because of backslash-escaped newlines. Both of those could in theory be dealt with by having a separate sparse `SourceLoc`->value mapping just for the tokens that need it.

* We could keep a more bloated `Token` type but then have the `TokenStream` type store the data in more of an SoA rather than AoS fashion, so that bulk token storage can be more efficient. That would eliminate the 6 bytes per `Token` of waste in your improved layout, and it could also be improved on by making the storage for optional values that not all tokens need be sparse.

PR #1350, opened 2020-05-19
AST dump improvements

jsmall-nvidia:
This is on top of #1348

* Made some AST fields unreflected (for example transitory fields and m_session back pointer)
* Improved how pointers are dumped
* Special case around ModuleDecl, to only output as references except for starting module


source/slang/slang-ast-base.h
```diff
@@ -144,7 +144,8 @@ class Type: public Val
 
     virtual RefPtr<Type> CreateCanonicalType() = 0;
     Type* canonicalType = nullptr;
-    
+
+    SLANG_UNREFLECTED
```
tfoleyNV, line 148: Unrelated to this change (it was actually part of a previous PR):

I wonder if we should bikeshed the name of this macro since it might be surprising that it works the way it does. Something like `SLANG_BEGIN_UNREFLECTED` might better convey the meaning of changing the default, but that also invites the question of where `SLANG_END_UNREFLECTED` is, or make it confusing that the behavior ends at the end of the enclosing type.

I might argue that it is better in terms of having self-documenting code to just have `SLANG_UNREFLECTED` apply to only the immediately following field/member and not everything that follows, even if it means making the header files more verbose.

PR #1354, opened 2020-05-22
Working matrix swizzle

Checkmate50:
Supports multiple elements and zero and one indexing of the forms:

```
float4x4 myMatrix = ...;
myMatrix._11_13;   // vec2(myMatrix[0][0], myMatrix[0][2])
myMatrix._m11_m13; // vec2(myMatrix[1][1], myMatrix[1][3])
```

Performs semantic checking of the swizzle.  Matrix swizzles are transformed into a vector of indexing operations during lowering to the IR.

This change does not handle matrix swizzle as lvalues.


source/slang/slang-check-expr.cpp
```diff
@@ -1286,39 +1286,39 @@ namespace Slang
                             // that takes a single `int` parmaeter and means to call that instead.
                             //
                             // For now that should be a non-issue, and in a pinch such a user
-                            // could use `T(0)` instead of `(T) 0` to get around this special
```
jsmall-nvidia, line 1289: The tabbing here seems to have changed - or been removed? We use 4 spaces for for each indent (we don't use tabs).

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+        swizExpr->base = memberRefExpr->baseExpression;
+
+        int elementIndices[4][2];
+        int elementCount = 0;
+
+        bool anyDuplicates = false;
+        int oneIndexed = -1;
+
+        String swizzleText = getText(memberRefExpr->name);
+        auto coursor = swizzleText.begin();
```
jsmall-nvidia, line 1375: Looks like I typo coursor -> cursor?

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+
+        while (coursor != swizzleText.end())
+        {
+            // Throw out swizzling with more than 4 output elements
+            if (elementCount >= 4) {
```
jsmall-nvidia, line 1380: Coding conventions can be somewhat arbitrary - and the { on the same line has some good properties, like making better use of vertical space. On the Slang project we've gone with the { } on the same indentation.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+                getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
+                return CreateErrorExpr(memberRefExpr);
+            }
+            int elementIndex[2];
```
jsmall-nvidia, line 1384: Nit pick - could perhaps used = { 0, 0 } to make shorter/cleaner.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+    RefPtr<Expr> SemanticsVisitor::CheckMatrixSwizzleExpr(
+        MemberExpr* memberRefExpr,
+        RefPtr<Type>      baseElementType,
+        IntegerLiteralValue baseElementRowCount,
+        IntegerLiteralValue baseElementColCount)
+    {
+        RefPtr<MatrixSwizzleExpr> swizExpr = new MatrixSwizzleExpr();
+        swizExpr->loc = memberRefExpr->loc;
+        swizExpr->base = memberRefExpr->baseExpression;
+
```
jsmall-nvidia, line 1367: This could do with a comment to say what the makeup for the array means.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+        int elementIndices[4][2];
+        int elementCount = 0;
+
+        bool anyDuplicates = false;
+        int oneIndexed = -1;
```
jsmall-nvidia, line 1372: Nit pick: This variable name reads like a boolean, and it sort of looks like it works that way until later it's used to subtract to make 1 indexed value zero indexed. I might be better to be named as zeroIndexOffset, zeroIndexFix or something like that.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+                oneIndexed = 1;
+            }
+
+            // Check for the ij components
+            for (Index j = 0; j < 2; j++) {
+                auto ch = *coursor++;
+                int subIndex;
+
+                // Manually enumerating cases to throw out 5+ and letters
+                switch (ch)
```
jsmall-nvidia, line 1419: Nit pick: Could perhaps be more concise with 
```
if (!(ch >= '0' && ch <= '5'))
{ 
    ... diagnose ... 
}
subIndex = ch - '0';

File: source/slang/slang-check-expr.cpp [hunk 8]
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+        int elementIndices[4][2];
+        int elementCount = 0;
+
+        bool anyDuplicates = false;
+        int oneIndexed = -1;
+
+        String swizzleText = getText(memberRefExpr->name);
+        auto coursor = swizzleText.begin();
+
+        while (coursor != swizzleText.end())
```
jsmall-nvidia, line 1377: As it turns out here things work out ok - but it probably worth pointing out that this surrounding loop check is looking for end, but that checking is not honored by the subsequent code. Which can step 1 or more characters and therefore could in principal step past end. If you know the cursor is a character pointer you might be better off with < (although that does not fix the problem). This works out because as it turns out the buffer held by String is 0 terminated, and so the subsequent code will hit that, and report as an 'invalid character'. 
That's probably not quite the error you want. 
Note that UnownedStringBuffer is not zero terminated, and so bad things would happen. 

So how might you improve this? Doing this kind of char checking whilst constantly checking for end can be a little tedious. Here you know it's zero terminated, that being the case you probably don't want to mix the mechanisms. 

More concretely something like

```
// the contents of String are zero terminated
const char* cursor = swizzleText.getBuffer(); 
while (*cursor) 
{ 
    ... 

```

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+                }
+                else {
+                    elementLimit = baseElementColCount;
+                    elementIndex[1] = subIndex;
+                }
+                // Make sure the index is in range for the source type
+                // Account for off-by-one and reject 0 if oneIndexed
+                if (subIndex >= elementLimit + oneIndexed || subIndex < oneIndexed)
+                {
+                    getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
```
jsmall-nvidia, line 1446: This might be improved with an error reporting the value is out of range and giving the value, and the range.

source/slang/slang-check-expr.cpp
tfoleyNV: FWIW, I specifically advocated to @Checkmate50 to use this style and rely on the fact that the buffer for a `String` is null-terminated. I believe the error messages won't specifically call out a single invalid character, but will instead reference the entire swizzle string as being invalid. I might be wrong though, and you may have looked at the code more closely than I have.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
         }
     }
 
+    RefPtr<Expr> SemanticsVisitor::CheckMatrixSwizzleExpr(
+        MemberExpr* memberRefExpr,
+        RefPtr<Type>      baseElementType,
+        IntegerLiteralValue baseElementRowCount,
+        IntegerLiteralValue baseElementColCount)
+    {
```
jsmall-nvidia, line 1363: On reading through this code it seems like this code could be improved from a readability point of view by using a struct to represent the coord. Doing so would make clearer the intent with 4 coordinates in an array. So something like

```
struct MatrixCoord
{
   bool operator==(...);
   bool operator!=(...) const { return !(*this == rhs); }
   // Rows and columns are zero indexed 
   int row;   
   int col;
};
```
You use the ability to access the row or col via an index. You could do with an accessor that returns a reference on the struct.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+            if (*coursor++ != '_') {
+                getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
+                return CreateErrorExpr(memberRefExpr);
+            }
+
+            // Check for one or zero indexing            
+            if (*coursor == 'm') {
+                // Can't mix one and zero indexing
+                if (oneIndexed == 1) {
+                    getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
```
jsmall-nvidia, line 1397: Nit pick: A message describing the problem here a little more precisely might be nice.

tests/diagnostics/matrix-swizzle.slang
```diff
@@ -0,0 +1,24 @@
+//DIAGNOSTIC_TEST:SIMPLE:
+
+int doSomething(int a)
+{
```
jsmall-nvidia, line 4: Like the test covering lots of situations.

tests/language-feature/swizzles/matrix-swizzles.slang
```diff
@@ -0,0 +1,34 @@
+// matrix-swizzle.slang
+
+//TEST(compute):COMPARE_COMPUTE:
+
```
jsmall-nvidia, line 4: Great to have a compare compute test to check it works on the different targets and on downstream compilers and APIs.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -2960,6 +2967,53 @@ struct LValueExprLoweringVisitor : ExprLoweringVisitorBase<LValueExprLoweringVis
...
+                subscript1,
+                base,
+                index1);
+            // Second index expression
+            irExtracts[ii] = getSimpleVal(context, subscriptValue(
+                subscript2,
+                irExtract1,
+                index2));
+        }
+        auto irArray = builder->emitMakeVector(
```
jsmall-nvidia, line 3008: Why irArray? Perhaps irVector?

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+                // Make sure the index is in range for the source type
+                // Account for off-by-one and reject 0 if oneIndexed
+                if (subIndex >= elementLimit + oneIndexed || subIndex < oneIndexed)
+                {
+                    getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
+                    return CreateErrorExpr(memberRefExpr);
+                }
+            }
+            // Check if we've seen this index before
+            for (int ee = 0; ee < elementCount; ee++)
```
jsmall-nvidia, line 1451: Nit pick - it seems like this check should be outside this loop.

source/slang/slang-check-expr.cpp
```diff
@@ -1355,6 +1355,162 @@ namespace Slang
...
+        return swizExpr;
+    }
+
+    RefPtr<Expr> SemanticsVisitor::CheckMatrixSwizzleExpr(
+        MemberExpr* memberRefExpr,
+        RefPtr<Type>		baseElementType,
+        RefPtr<IntVal>				baseRowCount,
+        RefPtr<IntVal>				baseColCount)
+    {
+        if (auto constantRowCount = as<ConstantIntVal>(baseRowCount))
```
jsmall-nvidia, line 1503: Nice, that it checks the size is reified.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -2960,6 +2967,53 @@ struct LValueExprLoweringVisitor : ExprLoweringVisitorBase<LValueExprLoweringVis
 
 struct RValueExprLoweringVisitor : ExprLoweringVisitorBase<RValueExprLoweringVisitor>
 {
+    // A matrix swizzle in an r-value context can save time by just
```
jsmall-nvidia, line 2970: Looks good.

source/slang/slang-check-expr.cpp
Checkmate50: I decided to skip on this requested change to match the HLSL style of just stating that the swizzle is invalid (and avoiding complicating the code with several error message types).  I can add a couple of error message types if you still disagree and think it's worth the overhead.

source/slang/slang-check-expr.cpp
jsmall-nvidia: The issue wasn't so much about relying on zero termination - that makes a lot of sense for this style of lexing. I realize my comment might waffle around this - as I'm trying to figure out what's going on. The problem is the *mix* where the outside loop is written to use the start/end style, which implies that the remaining code is follows the same mechanism which it doesn't, so that raises alarms. So just making the loop rely an the the same zero termination style would make it ok.

source/slang/slang-ast-expr.h
```diff
@@ -176,6 +176,27 @@ class StaticMemberExpr: public DeclRefExpr
     RefPtr<Expr> baseExpression;
 };
 
+struct MatrixCoord
+{
+    bool operator==(const MatrixCoord rhs) const {
```
jsmall-nvidia, line 181: For a short single line expression like this it's ok to just do as 
```
bool operator==(const MatrixCoord& rhs) const { return row == rhs.row && col == rhs.col; };
```

You might want to make the rhs a reference.

source/slang/slang-check-expr.cpp
jsmall-nvidia: Within our coding standards the opening brace indentation should line up with the closing brace, so we need to have that fixed.

tfoleyNV: In case the phrasing isn't clear: the Slang codebase prefers the "Allman" convention:

```c++
if(condition)
{
    stuff();
}
```

and does not use the "K&R" convention:

```c++
if(condition) {
    stuff();
}
```

The main exception is when you can reduce a block to a single line:

```c++
if(condition) { stuff(); }
```

source/slang/slang-check-expr.cpp
```diff
@@ -1365,107 +1365,97 @@ return coerce(typeExp.type, initListExpr);
...
         {
             // Throw out swizzling with more than 4 output elements
             if (elementCount >= 4) {
                 getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
                 return CreateErrorExpr(memberRefExpr);
             }
-            int elementIndex[2];
-            elementIndex[0] = 0;
-            elementIndex[1] = 0;
+            MatrixCoord elementIndex = { 0, 0 };
```
jsmall-nvidia, line 1387: Nit pick: I guess it could be argued that the MatrixCoord is an 'index', but when I read index it means an integer based value. That matrixCoord, or just coord avoids this.

PR #1355, opened 2020-05-23
Improvements around hashing

jsmall-nvidia:
* Made hash types explicit
* Added 'Stable' hash for hash functions that need to be the same across targets (and not often changed as exposed via api/serialization)
* Use lowerCamel for function/method names
* Some small standards improvements around AST nodes
* Some fixes to make compilation work on latest GCC (implicit operator= is now depreciated)
* Fixed issue with GCC and exception catching - need to be references else slicing

Strictly speaking to get this to compile on latest GCC I also need to add "-Wno-class-memaccess" to the premake5. I can test and make a PR incorporating those fixes later.

### DIFF
--- a/examples/model-viewer/main.cpp
+++ b/examples/model-viewer/main.cpp
@@ -1018,13 +1018,13 @@ struct ShaderCache : RefObject
             return true;
         }
 
-        Slang::HashCode GetHashCode() const
+        Slang::HashCode getHashCode() const
         {
-            auto hash = ::GetHashCode(effect);
-            hash = combineHash(hash, ::GetHashCode(parameterBlockCount));
+            auto hash = ::getHashCode(effect);
+            hash = combineHash(hash, ::getHashCode(parameterBlockCount));
             for( UInt ii = 0; ii < parameterBlockCount; ++ii )
             {
-                hash = combineHash(hash, ::GetHashCode(parameterBlockLayouts[ii]));
+                hash = combineHash(hash, ::getHashCode(parameterBlockLayouts[ii]));
             }
             return hash;
         }
--- a/examples/model-viewer/main.cpp
+++ b/examples/model-viewer/main.cpp
@@ -1018,7 +1018,7 @@ struct ShaderCache : RefObject
             return true;
         }
 
-        UInt GetHashCode() const
+        Slang::HashCode GetHashCode() const
         {
             auto hash = ::GetHashCode(effect);
             hash = combineHash(hash, ::GetHashCode(parameterBlockCount));
--- a/source/core/slang-hash.h
+++ b/source/core/slang-hash.h
@@ -18,7 +18,9 @@ namespace Slang
     // A fixed 32bit wide hash on all targets.
     typedef uint32_t HashCode32;
 
-    // TODO(JS): We might want to implement HashCode as just an alias to HashCode32 or HashCode64 based on target.
+    SLANG_FORCE_INLINE HashCode32 toHash32(HashCode value) { return (sizeof(HashCode) == sizeof(int64_t)) ? (HashCode32(uint64_t(value) >> 32) ^ HashCode(value)) : HashCode32(value); }
+    SLANG_FORCE_INLINE HashCode64 toHash64(HashCode value) { return (sizeof(HashCode) == sizeof(int64_t)) ? HashCode(value) : ((HashCode64(value) << 32) | value); }
+
     SLANG_FORCE_INLINE HashCode GetHashCode(int64_t value)
     {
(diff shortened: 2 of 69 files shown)


examples/model-viewer/main.cpp
```diff
@@ -1018,13 +1018,13 @@ struct ShaderCache : RefObject
             return true;
         }
 
-        UInt GetHashCode() const
+        Slang::HashCode getHashCode() const
```
tfoleyNV, line 1021: It is probably overkill to try to add to this PR, but I'd like to see us eventually move to a model where hashable types implement something like:

```c++
void hashInto(Hasher& hasher);
```

And then there can be a single generic `getHashCode()` operation that is:

```c++
template<typename T>
HashCode getHashCode(T const& value)
{
    Hasher hasher;
    value.hashInto(hasher);
    return hasher.getResult();
}
```

The reason for the alternative design is that we currently have a lot of cases where an aggregate type computes its hash by calling `GetHashCode` recursively on its fields/contents and then using `combineHash` to reduce things down. This approach has two big drawbacks:

1. There's an annoying amount of boilerplate code because you need both the `GetHashCode()` and `combineHash()` calls for each field, and you often need to thread through a local variable like `hash` below, which gets uttered twice for each field.

2.  The quality of the hash is compromised in that when computing the hash of a type with many bits of state, you reduce each of its fields down to the number of bits in `HashCode` and *then* combine then, instead of combining them into an intermediate hash with a larger number of bits, and then reducing it down at the end.

Having a distinct `Hasher` type solves both of those problems:

1. Calling `hasher.hash(something)` can both handle the recursive hashing of `something` into the hasher, and deal with combining it with the hash state that was already in place.

2. Because the `Hasher` type is opaque and only used during the computation of hash codes, it can use more bits for its internal state than `HashCode` does.

I'm not arguing for this change to get added to this PR, but it might be worth considering as a follow-up if we are changing all these places in the code anyway.

source/core/slang-hash.h
```diff
@@ -7,56 +7,102 @@
 
 namespace Slang
 {
-    typedef int HashCode;
+    // Ideally Hash codes should be unsigned types - makes accumulation simpler (as overflow/underflow behavior are defined)
+    // Only downside is around multiply, where unsigned multiply can be slightly slower on some targets.
 
-	inline int GetHashCode(double key)
+    // HashCode - size may vary by platform. Typically has 'best' combination of bits/performance. Should not be exposed externally as value from same input may change depending on compilation platform.
+    typedef unsigned int HashCode;
```
tfoleyNV, line 14: I strongly agree with the move to make `HashCode` be unsigned.

Is there a strong argument for moving to a `HashCode` that is always 64 bits, or always pointer-sized, or do we think that sticking with the "natural" `int` size is always going to be the right choice?

jsmall-nvidia: When performance is key and you support 32 and 64 word targets, having the 'normal' hash type be pointer sized makes a lot of sense. 

It could be now argued that the only targets we really care about are 64 bit. That making HashCode always 64 bits might diminish perf on 32 bit, and perhaps significantly (as in a few percent of overall execution), but if 32 bit targets are not a priority, then perhaps thats ok in the context of simplicity.

examples/model-viewer/main.cpp
jsmall-nvidia: I mean following this logic, it would argue for having the virtual methods be not getHashCode but intoHasher also. Then the final, getting of the HashCode is only something that is done at the very end. 

A nice advantage of such a design is that it allows for more bits for the hash state, and that the final hash mixing can be a one off performed when the hash code is produced at the end.

I also agree the combineHash mechanism adds a lot of cruft.

You could also use operator<< overload on the hasher to allow doing 

```
void Thing::intoHasher(Hasher& hasher)
{
    hasher << m_fieldA << m_anotherField << m_andAnother;
}
```
```

tfoleyNV: I think we are in violent agreement here.

source/core/slang-hash.h
tfoleyNV: I guess the central question for me is whether sticking with an `int`-sized hash value is a kind of "worst of both worlds" thing, in that there are 64-bit platforms where it will end up being 32 bits and others where it will be 64 bits, and that results in three different cases of hashing behavior that we need to think about for stuff like `Dictionary` and `HashSet` (ILP32, LP64, and ILP64).

jsmall-nvidia: Yes - unsigned int isn't arguably the best choice. That this PR was in part about making it so you could easily have the option of changing the size. 

If you want a combination of perf/most bits across targets (that include 32/64 bits) it should probly be ptr sized. If you want some simplicity at the cost of perf on 32 bit, an argument can be made for it being 64 bit. 

To see what the impact should now be pretty easy to check.

PR #1358, opened 2020-05-27
WIP: ASTBuilder

jsmall-nvidia:
This is work in progress for the following goals for AST

* Remove reference counting
* Use a memory arena for allocation 
* Use ASTBuilder for node lifetime management/construction
* Use the node class id for rtti (rather than slower getClassType virutal call)
* Types/structures that are shared are moved into SharedASTBuilder (some coming from Session)
* Make the AST serializable

Most of the foundation for the above goals is here, but the final steps to complete have yet to be taken.

Whilst it's marked as WIP, it would be good to have reviewed and merged because the change hit's a lot of source.

Note that currently the ASTBuilder is not owned by the Module/ModuleDecl which means everything is owned by Linkage and Session, which isn't right.


source/slang/slang-ast-builder.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    if (0) {}
+#define CASE(NAME) \
+        else if(name == #NAME) return create<NAME>()
+
+    CASE(GLSLBufferModifier);
+    CASE(GLSLWriteOnlyModifier);
+    CASE(GLSLReadOnlyModifier);
+    CASE(GLSLPatchModifier);
+    CASE(SimpleModifier);
```
tfoleyNV, line 183: Are these really the only cases that need create-from-name functionality?

jsmall-nvidia: I guess that's kind of silly - as we have the reflection mechansism that creates any of them. I just copied it over...

jsmall-nvidia: Ok removed that.

PR #1363, opened 2020-05-29
NodeBase types constructed with astNodeType member set

jsmall-nvidia:
* Made Substitutions derive from NodeBase
* Added astNodeType member to NodeBase
* Made getClassInfo non virtual (can do so because of astNodeType member)
* Set the astNodeType on creation via ASTBuilder - via init function (that can also set ASTBuilder - replacing setASTBuilder)
* Make all NodeBase construction through ASTBuilder create functions
* Made the ReflectClassInfo create function type take an ASTBuilder parameter, such that it is constructed through the ASTBuilder

### DIFF
--- a/source/slang/slang-ast-base.h
+++ b/source/slang/slang-ast-base.h
@@ -23,10 +23,19 @@ class NodeBase : public RefObject
 {
     SLANG_ABSTRACT_CLASS(NodeBase)
 
-        // By default AST types do *not* store the builder. This is called when constructed tho.
-    SLANG_FORCE_INLINE void setASTBuilder(ASTBuilder* astBuilder) { SLANG_UNUSED(astBuilder); }
+        // MUST be called before used. Called automatically via the ASTBuilder.
+        // Note that the astBuilder is not stored in the NodeBase derived types by default.
+    SLANG_FORCE_INLINE void init(ASTNodeType inAstNodeType, ASTBuilder* /* astBuilder*/ ) { astNodeType = inAstNodeType; }
+
+        /// Get the class info 
+    SLANG_FORCE_INLINE const ReflectClassInfo& getClassInfo() const { return *ReflectClassInfo::getInfo(astNodeType); }
 
     SyntaxClass<NodeBase> getClass() { return SyntaxClass<NodeBase>(&getClassInfo()); }
+
+        /// The type of the node. ASTNodeType(-1) is an invalid node type, and shouldn't appear on any
+        /// correctly constructed (through ASTBuilder) NodeBase derived class. 
+        /// The actual type is set when constructed on the ASTBuilder. 
+    ASTNodeType astNodeType = ASTNodeType(-1);
 };
 
 // Casting of NodeBase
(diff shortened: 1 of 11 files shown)


source/slang/slang-ast-base.h
```diff
@@ -23,39 +23,49 @@ class NodeBase : public RefObject
 {
     SLANG_ABSTRACT_CLASS(NodeBase)
 
-        // By default AST types do *not* store the builder. This is called when constructed tho.
-    SLANG_FORCE_INLINE void setASTBuilder(ASTBuilder* astBuilder) { SLANG_UNUSED(astBuilder); }
+        // MUST be called before used. Called automatically via the ASTBuilder.
+        // Note that the astBuilder is not stored in the NodeBase derived types by default.
+    SLANG_FORCE_INLINE void init(ASTNodeType inAstNodeType, ASTBuilder* /* astBuilder*/ ) { astNodeType = inAstNodeType; }
```
tfoleyNV, line 28: Unrelated to this change, but: I'd really like us to decide on and document a policy for how to name "tag" types like `ASTNodeType`.

I'm not a fan of using "type" or "kind" for these, and have usually used "flavor" for the ones I introduce (because the terms "type" and "kind" have more specialized meaning in a compiler or type system). That said, I also know that using "type" for this kind of field is something like the 99% case for codebases in the wild.

Using `astNodeType` instead of just `type` is a nice step because it makes it clear that it represents the type of the AST node object and not something like the type of an expression (which is represented as an AST node...).

There's nothing to change in this PR, so this is just my thoughts meandering...

source/slang/slang-ast-base.h
```diff
@@ -23,39 +23,49 @@ class NodeBase : public RefObject
...
+    ASTNodeType astNodeType = ASTNodeType(-1);
 };
 
 // Casting of NodeBase
 
 template<typename T>
 SLANG_FORCE_INLINE T* dynamicCast(NodeBase* node)
 {
-    return (node && node->getClassInfo().isSubClassOf(T::kReflectClassInfo)) ? static_cast<T*>(node) : nullptr;
+    return (node && ReflectClassInfo::isSubClassOf(node->astNodeType, T::kReflectClassInfo)) ? static_cast<T*>(node) : nullptr;
```
tfoleyNV, line 46: Would it make sense to have `astNodeType` be a wrapper type around the `enum` so that it could provide member functions like `isSubClassOf` rather than requiring separate free/static functions?

source/slang/slang-ast-dump.cpp
```diff
@@ -679,33 +682,11 @@ void Context::dumpObjectFull(NodeBase* node)
     }
 }
 
-void Context::dumpObjectFull(Substitutions* subs)
```
tfoleyNV, line 682: I like that just putting `Substitutions` into the `NodeBase` hierarchy allows us to remove this code.

source/slang/slang-ast-base.h
jsmall-nvidia: That perhaps using the ReflectClassInfo, in a better way is preferable as it's a class already, and can have suitable methods, allowing the node type to just be an enum.

PR #1368, opened 2020-06-03
Devirtualize AST types

jsmall-nvidia:
* Made all virtual methods (other than dtor) on AST nodes non virtual
* Moved AST node function definitions to associated .cpp files

The mechanism used for this was to replace the virtual function in the base class with a non virtual one, and make it's contents something like

```
RefPtr<Val> Val::substituteImpl(ASTBuilder* astBuilder, SubstitutionSet subst, int* ioDiff)
{
    SLANG_AST_NODE_VIRTUAL_CALL(Val, substituteImpl, (astBuilder, subst, ioDiff))
}
```

This code generates a switch that will call (in this case) _substituteImplOverride for all types derived from Val. If `substituteImpl` in the original code was = 0, we cannot directly simulate this, but we can make it assert or fail along a non implemented path - for example using SLANG_UNEXPECTED. The transformation doesn't have compiler support to check the overriding is correct, or not allowing abstract construction. Here if an 'abstract' type is constructed, it will fail when the devirtualized method is called at runtime.

This was achieved by making the C++ extractor macros generate a switch to do the virtual dispatch. Unfortunately this also introduced a problem, as the Override functions require to be accessible to a base classes implementation. The work around here was to just make Override functions public. Work was performed to generate the friend definitions as part of SLANG_CLASS and SLANG_ABSTRACT_CLASS and this nearly works via using the ORIGIN mechanism. This doesn't work for getSup though because it is not defined in the base classes. We could of course make work by requiring if a new 'virtual' method is added to a class, it's friendness is added to the derived classes generated code. This was easy for the base classes via origin - but seemed to fiddly for other places so was removed here.

This PR does not remove all of the virtual functions from AST nodes - the dtor is still virtual. Making destruction non virtual is possible, because we can expand the ReflectClassInfo have a pointer to the inplace destructor. This is not enough though - because whilst it seems like we could do something like

```
class NodeBase
{
     public:
     ~NodeBase()
     {
          ReflectTypeInfo::getInfo(astTypeNode).dtor(this);
     }
};
```

It unfortunately leads the ~NodeBase being invoked recursively, because ~NodeBase will be invoked by dtor().

If the lifetimes of AST nodes is say managed by the ASTBuilder, then it can correctly invoke the dtor without using delete or calling the destructor.

To make the AST nodes non virtual, also requires them not deriving from RefObject, and implies they should not be refcounted if managed by ASTBuilder.


source/slang/slang-ast-reflect.h
```diff
@@ -28,4 +28,26 @@
...
+
+#define SLANG_AST_NODE_VIRTUAL_CALL(base, methodName, methodParams) \
+    switch (astNodeType) \
+    { \
+        SLANG_ALL_ASTNode_##base(SLANG_AST_NODE_CASE, (methodName, methodParams)) \
+        default: return SLANG_AST_NODE_INVOKE (methodName, methodParams); \
+    }
+
+// Same but for a method that's const
+#define SLANG_AST_NODE_CONST_CASE(NAME, SUPER, ORIGIN, LAST, MARKER, TYPE, param)      case ASTNodeType::NAME: return static_cast<const NAME*>(this)-> SLANG_AST_NODE_INVOKE param;
```
tfoleyNV, line 45: Minor design nit: it would eliminate some duplication in this code to not bother having a separate `const` and non-`const` cases. My general philosophy is that `const`-ness is mostly only of interest when it pertains to "value" types, and of almost no benefit when it comes to "reference" / "object" types (which is what the AST nodes are).

I know there are differences of opinion on this stuff, so I'm fine with future proofing things for `const` use cases, but I'm also no excited about a future where those distinctions start to matter for stuff like the AST.

PR #1371, opened 2020-06-04
Emit [loop] attribute to output HLSL.

csyonghe:
This PR adds HLSL emit for the `[loop]` attribute.

This is needed because `fxc` will emit an error if it cannot unroll a large loop that has not been marked with `[loop]`.


csyonghe:
While working on the test case, I found another issue. If we have two loops like this:
```
for (int i = 0;  i < 5; i++) doSomething(i);
for (int i = 0; i < 5; i++) doSomethingElse(i);
```

The generated HLSL will cause fxc to issue an error about i being redefined.

tfoleyNV:
I'm surprised that our output HLSL in that case causes an error, because I'd expect the `i` variables to have unique names in the generated code (e.g., `i_0` and `i_1`... and probably more than that thanks to SSA).

Is it possible that the error message is coming from fxc during a "pass-through" compilation, rather than a true Slang compilation?

source/slang/slang-emit-hlsl.cpp
```diff
@@ -673,9 +673,16 @@ void HLSLSourceEmitter::emitVectorTypeNameImpl(IRType* elementType, IRIntegerVal
...
+    switch (decl->getMode())
     {
+    case kIRLoopControl_Unroll:
         m_writer->emit("[unroll]\n");
+        break;
+    case kIRLoopControl_Loop:
+        m_writer->emit("[loop]\n");
+        break;
+    default:
+        SLANG_UNREACHABLE("emit loop control decoration");
```
tfoleyNV, line 685: I might leave out the `default` here in case we ever add more loop control modes, where only a subset of them are implemented on each target.

tests/cross-compile/loop-attribs.slang
```diff
@@ -0,0 +1,19 @@
+// loop-attribs.slang
+// Test that loop attributes are correctly emitted to the resulting HLSL.
+
+//TEST:SIMPLE:-target hlsl  -entry main -stage fragment -profile sm_6_0
```
tfoleyNV, line 4: It is usually best to avoid having tests where the expected output is the generated HLSL or GLSL text from the compiler, because that can end up changing in a bunch of ways that shouldn't affect the validity of a given test.

My preference for stuff like this is to define is as a `//TEST:CROSS_COMPILE:` test and then have a `*.slang.hlsl` file that contains the equivalent raw HLSL (you could use your current `.slang.expected` to set that up), and then the test will check against the expected output DXBC and/or DXIL.

Under that model, we don't care if the HLSL produced by Slang is exactly the same; we just care if the output HLSL compiles to the same DXBC/DXIL.

(Note that I'm also *not* advocating for establishing a raw DXBC/DXIL baseline, since that is also a source of problems since fxc/dxc version changes can alter the exact DXBC/DXIL produced. It might help to hash out the details here in a separate conversation thread...)

csyonghe:
> I'm surprised that our output HLSL in that case causes an error, because I'd expect the `i` variables to have unique names in the generated code (e.g., `i_0` and `i_1`... and probably more than that thanks to SSA).
> 
> Is it possible that the error message is coming from fxc during a "pass-through" compilation, rather than a true Slang compilation?

Seems like it might have been a case of "pass-through" compilation. Now that I moved the test to cross-compile I can no longer reproduce the issue.

PR #1373, opened 2020-06-04
Checkin .clang-format and an example file for discussion

csyonghe:
Let's have a discussion on what changes needs to be made and I will remove the example cpp file before we checkin.


examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+                              TestClass* param4,
+                              float* param5,
+                              std::pair<int, float> param6)
+        {
+            if (param1 == 0)
+                longParameterList(param1,
+                                  param1 + param2 + param3,
+                                  param2 - param1 + (1 << 23) / sizeof(TestClass),
+                                  param4 + (param1) * sizeof(unsigned),
+                                  param6);
```
tfoleyNV, line 126: These bits are the main place where I'd object to the indentation choices in this example. I *think* I want `AlignAfterOpenBracket: AlwaysBreak`). Aligning all the parameters to the opening `(` has the problem that a find-replace of a function name results in changing the indent of all the parameters at all call sites.

In terms of my personal style, I'd actually like to have some of the other `Align*` settings turned on (e.g., `AlignConsecutiveAssignments`) because I think they lead to much more readable code, but I'm nervous about enabling this globally for similar reasons. I *think* the consequences of changing a variable name (and thus re-indenting its peers) are more local in practice than the consequences of changing a function name, but I'm not sure how that will work in practice.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+                {
+                    return normalFunction(0, nullptr);
+                }
+                break;
+                case 1:
+                {
+                    float b = *param2;
+                    i--;
+                }
+                break;
```
tfoleyNV, line 60: I'm not a fan of this style which leads to nested `{}` blocks without indenting, but I will concede that there are few good options possible for `switch` statements. I'd favor something where we go ahead and indent all code beneath each `case`, and if that leads to a double indent if the code under the `case` is a `{...}` block, so be it.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+                default:
+                    break;
+                }
+            }
+            if (param1)
+                return *param2;
+            else
+                return -1;
```
tfoleyNV, line 68: Do we need some specific rules about whether or not to allow unbraced statements in control-flow constructs like this? I end up using these sorts of things when they are one-liners:

```
if(skipThisOne) continue;
```

or *almost* one-liners like:

```
if(someComplexConditionThatTakesUpALotOfHorizontalSpace)
    return;
```

However, I'm aware that many people would rather avoid any possibility for hidden errors and insist that the sub-statements must always use `{}`.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+        class LoweredValInfo
+        {
+        };
```
tfoleyNV, line 72: I'd *strongly* prefer to be able to have more compact styles when we have empty `{}` blocks:

```c++
struct MyDerived : MyBase {}; // no need for more lines, please

virtual void handleSomeCase(...)
{} // nothing to do by default
```

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
+
+#include "internalGroup1.h"
+#include "slang.h"
+#include <algorithm>
```
tfoleyNV, line 4: I'd advocate for `SortIncludes: true`, since we mostly use that style today. I think keeping `IncludeBlocks: Preserve` gives us the flexibility to override the ordering in cases where we care.

As long as we are at it, I'd also advocate for `SortUsingDeclarations: true`.

.clang-format
```diff
@@ -0,0 +1,21 @@
+---
+BasedOnStyle: LLVM
+IndentWidth: 4
```
tfoleyNV, line 3: I assume we want `UseTab: Never`, or is that implied by the base `LLVM` style?

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        int shortFunction() { return 0; }
+
+        /// <summary>
+        /// A normal function.
+        /// </summary>
+        /// <param name="param1">The first parameter.</param>
+        /// <param name="param2">The second parameter</param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
csyonghe: No they are not, they happen to be something that VS automatically generates. I am not advocating for an XML comment.

.clang-format
csyonghe: I think so, but it doesn't hurt to add it.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,130 @@
...
+namespace slang
+{
+    /// <summary>
+    /// Summary of TestClass
+    /// </summary>
+    class TestClass
+    {
+    private:
+        int examplePrivateMember; // Trailing comment1
+        int* examplePointerMember = nullptr; // trailing coment2
```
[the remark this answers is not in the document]
csyonghe: I am not a fan of AlignTrailingComments for the same reason to keep changes local. When you change the variable name of one line, you don't want to cause nearby lines to change because of that.

examples/hello-world/format-example.cpp
csyonghe: Do you prefer such rules? clang-format doesn't enforce any of them, and I don't have a strong opinion on this.

examples/hello-world/format-example.cpp
csyonghe: Added `IndentCaseBlocks:true`

examples/hello-world/format-example.cpp
csyonghe: Added `AlignAfterOpenBracket: AlwaysBreak`

csyonghe:
I updated the format settings. However `IndentCaseBlocks` is very new clang feature that doesn't exist in the clangformat.exe that comes with VS19 so I commented it out. If we feel strongly in need of it, we can turn it on but that requires everyone to download and use the latest clangformat.exe.

examples/hello-world/format-example.cpp
tfoleyNV: I don't think we need to enforce anything on that kind of point. As much as possible I want to trust everybody to make responsible decisions about how they structure their code, and only use automated tools to enforce consistency on stuff that isn't interesting (whitespace, indentation, newlines, etc.).

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+#endif
+
+namespace slang
+{
+/// <summary>
+/// Summary of TestClass
+/// </summary>
+class TestClass
+{
+private:
```
jsmall-nvidia, line 24: The style of having private members at the start of a class is the opposite of what is desired in general. Typically the member variables are an 'implementation detail'.They should typically be at the end of a class definition. 

That the first/main thing you want to see in a class definition is the interface. Under this logic unless the methods implementation is small, it should be placed later, either in a .cpp file or in the header, but afterwards/inlined.

If there are public member variables, they should be before protected:/private:

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+        , examplePointerMember(nullptr)
+        , member4(1.0f)
+        , member5(200.0f)
+    {}
+
+    virtual void emptyFunction() {}
+
+    int shortFunction() { return 0; }
+
+    /// <summary>
```
jsmall-nvidia, line 42: Where we do have parameters explicitly defined it's less verbose to use @param, @return etc.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+        // The `materialize` operation should ensure that we only have to
+        // deal with the small number of base cases for lowered value
+        // representations.
+        //
+        baseVal = materialize(context, baseVal);
+
+        switch (baseVal.flavor)
+        {
+        case LoweredValInfo::Flavor::Simple:
```
jsmall-nvidia, line 112: Personally find switches more readable/consistent if they are indented (like they are in most circumstances - with the exception namespaces). 
I also generally prefer having braces surrounding case contents as a generality. It doesn't work always - if you just have returns, then they are placed directly after the : for example.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+    int longParameterList(
+        int param1,
+        int param2,
+        int param3,
+        TestClass* param4,
+        float* param5,
+        std::pair<int, float> param6)
+    {
+        if (param1 == 0)
+            longParameterList(
```
jsmall-nvidia, line 135: I know Tim likes lists of parameters in this style. If you have a really long list of parameters, it makes sense - but I would also claim that having a really long list of parameters has a code smell. There is a balance between code density and readability - typically, but by no means always, density is the more important.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+namespace slang
+{
+/// <summary>
+/// Summary of TestClass
+/// </summary>
+class TestClass
+{
+private:
+    int examplePrivateMember; // Trailing comment1
+    int* examplePointerMember = nullptr; // trailing coment2
```
jsmall-nvidia, line 26: Is there a parameter to keep these comments lined up?

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "internalGroup1.h"
+#include "slang.h"
+#include <algorithm>
+
+#include "includeGroup2.h"
+#include <vector>
+
+#include <vulkan.h>
+
+#ifdef __linux__
```
jsmall-nvidia, line 11: I do think it's a benefit having indentation on pre processor usage.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+    TestClass()
+        : examplePrivateMember(0)
+        , examplePointerMember(nullptr)
+        , member4(1.0f)
+        , member5(200.0f)
+    {}
+
+    virtual void emptyFunction() {}
+
+    int shortFunction() { return 0; }
```
jsmall-nvidia, line 40: Having short methods on a single line is handy. It also typical helps in code density, and overall readabily.

examples/hello-world/format-example.cpp
```diff
@@ -0,0 +1,144 @@
...
+        {
+            int col = 6;
+            col--;
+        }
+        if (param1)
+            return *param2;
+        else
+            return -1;
+    }
+    class LoweredValInfo
```
jsmall-nvidia, line 85: Not sure if this is a style issue with empty class definitions always taking more than one line? Having it take two lines doesn't seem to be a benefit.

tfoleyNV:
To make sure I understand, is the concern that automated formatting raises the barrier to entry for new contributors, since they need to have additional tools and processes before they can create a PR? Or is it just the amount of friction that it creates for contributors even after they have the tools set up?

I'm certainly not keen on having more tools involved in my process, which is why having something that can ideal work in the context of VS2019 would be appealing (assuming all the current and near-term contributors can use that IDE).

I agree on the point of trying to accept contributions from people who might have different styles of code they produce, and I think we want to err on the side of being flexible (if somebody comes along with a great PR but we don't agree with their naming convention choices, we can/should accept it and then consider changing the names down the line).

Some minimal automated enforcement of style around whitespace--indentation, alignment, newlines, etc.--seems like the best way to dip our toes in the water of automated formatting without getting prescriptive. If we can't find a way to make automated formatting work for us on trivial matters, I definitely wouldn't want to expand it to anything more serious.

csyonghe:
The thing is if we don't have this .clang-format file, Visual Studio will still apply its default formatting for every line of code you type. So my thinking is why not having some formatting tool that can do more for you. I am probably wrong about going this way but so far I haven't noticed any friction caused by having a .clang-format file in the repo. Visual Studio will discover and use it to format your code automatically. If anyone dislikes auto formatting they can turn the setting off. But for the ones who uses this feature, it provides great convenience since you will not need to worry about where to break the lines etc. any more, so I see that as a productivity gain instead of a measure to enforce any kind of formatting rules.

examples/hello-world/format-example.cpp
csyonghe: This is not a style that clang-format enforces. It is just that my sample code happen to be this way. Clang-format will not change the order things are declared in classes.

examples/hello-world/format-example.cpp
csyonghe: There is. But I preferred not to use them because they cause more than necessary changes and lead to overhead in code reviews. If you are strongly against this style we can change.

tfoleyNV: My two thoughts on this:

* I would object strongly if the automated formatting *stops* me from aligning comments/names/etc. across multiple declarations. If I can manually align things and have the automatic formatting not mess that up, I'm okay with leaving its default behavior as not enforcing alignment.

* The agree that having alignment create extra diffs is a bit worrying. For code reviews you can use options to ignore whitespace (GitHub's diff viewer includes that as an option), but for things like merges the result could be more painful.

One potential saving grace for this kind of automated alignment is that it should only lead to whitespace changes when you modify at least one declaration in the block, which hopefully wouldn't collide with another change unless it *also* modified a declaration in the same block. The main case where that could lead to merge conflicts is any kind of bulk renaming of types, which is the kind of thing that can easily create conflicts anyway.

jsmall-nvidia:
Well maybe I misunderstood the PRs purpose. If the idea is to have a clang format file that largely fits what we think is an appropriate style, and developers can optionally use that, then fair enough. 

If it's purpose is to define what 'correct formatting' is - and then that is enforced either at check-in time or a requirement of developers, or regular syncs, then I think the friction will probably outweigh the benefit. 

That more philosophically there is the problem of nuance. On one level we are communicating with a machine, but more importantly we are communicating with our later selves, other team members, and future contributors. If enforced formatting adds friction, or lessens clarity because of it's rigidity it may not be worth it. That on a recent project enforced clang format did lead to imho a drop in code readability and significant friction. When I had it's automatic formatting enabled - perhaps with an earlier version than is in 2019 - it was infuriating.

I turn off most of Visual Studios formatting because the friction is too high. For me the odd benefit where it does something that helps me is far outweighed for all the times it gets it wrong and I have to fight it.

tfoleyNV:
>Well maybe I misunderstood the PRs purpose. If the idea is to have a clang format file that largely fits what we think is an appropriate style, and developers can optionally use that, then fair enough.

I *think* that is what is being proposed right now, with the caveat that depending on what the default out-of-the-box experience is with VS2019 we may find that it is a de facto more aggressive change.

> If it's purpose is to define what 'correct formatting' is - and then that is enforced either at check-in time or a requirement of developers, or regular syncs, then I think the friction will probably outweigh the benefit.

I don't think that strict enforcement is being proposed. I would not expect that running the clang-format script would be a hard requirement on developers, and I would not want to have any automated tooling (e.g. in CI) that tries to push formatting changes back on the repository.

Aside: I do think that any effort like this will de facto define what "correct" formatting is for the codebase, at least by default. Whether or not we expect all developers to conform to that "correct" formatting is a separate issue, but a machine-readable file like this will always be a stake in the ground about what is right/wrong (even if we don't apply it).

> That on a recent project enforced clang format did lead to imho a drop in code readability and significant friction. When I had it's automatic formatting enabled - perhaps with an earlier version than is in 2019 - it was infuriating.

The issue of what VS2019 does or does not do (especially by default) is the main deal-breaker question for me right now.

LIke you, I've had an experience where a codebase using clang-format support in VS2019 led to an infuriating editing experience, where the automatic formatting would reflow blocks as I was writing them, and often re-position my cursor to a different location (on a different line!) while I was just typing ordinary code (or going through the keyboard motions I naturally do for setting up, e.g., `{}`-enclosed blocks). Automatic formatting made it impossible for me to code productively, so I had to turn it off and only apply it selectively.

The other potential deal-breaker issue for me is what to do about the larger fraction of code already in the repository that doesn't follow the rules that clang-format will try to enforce. I see two main alternatives:

1. Do a stop-the-world pass to reformat all the existing files and establish a baseline that is at least close to matching the expected conventions.

2. Don't try to reformat existing code, and rely on developers to take care not to introduce spurious formatting changes when doing other work.

I worry about whether (1) will lead to a ton of churn and pain, but I also worry that (2) is going to cause some real pain of its own, both for developers who enable automatic formatting (which might then reformat some files/functions/whatever without them realizing it, and lead to massive diffs), and those who don't (who find their code routinely getting clobbered in merge-conflict-y ways by developers in that first group).

csyonghe:
If most of us aren't willing to see such a file checked-in and aren't having a pleasant experience dealing with clang-format, then I don't see a reason to use it. I don't really care about this issue that much, just thought it may make life easier because I personally really don't like formatting the code myself and I would accept anything that a machine does. But if this is not the case for the majority of us, then let's pull back on this.

tfoleyNV:
The ground rules I'd like to make sure are enshrined in the codebase and our policies are:

* Respect the ability of individual contributors to write good code, and don't put them in a situation where they feel like they spend a lot of their time/effort on appeasing a pedantic set of coding conventions rather than on solving real problems. This may include letting them check in "bad code" sometimes, because there are other priorities in play (so long as the code reviewer(s) are convinced, that should be enough).

* As much as possible, developers should refrain from changing code that is not the focus of their work "just because." When undertaken carefully, large-scale refactoring can make the codebase more consistent, and benefits the team over time, but there is also a risk that well-meaning "tidying" of the code can create merge conflicts, or that two or more programmers with different ideas of what good code looks like end up in an edit war.

The second rule is kind of an extension of the first (whatever code you are looking at was writing by another contributor, so default to assuming they wrote things that way on purpose).

I'm fine with checking in some kind of automated formatting so long as it doesn't inadvertantly lead to us breaking either of those two rules, which basically means:

* The automated formatting needs to be an optional thing that programmers *can* use, and not something that becomes obtrusive to their workflow (e.g., commits getting rejected over formatting issues).

* The automated formatting needs to be localized so that people who turn it on don't accidentally create large diffs that "tidy" up code unrelated to their changes. Auto-formatting new code is fine, and so is *intentionally* auto-formatting old code in cases where we mutually agree the cleanup is warranted. If regular PRs end up filled with unrelated auto-formatting diffs then I'll be worried.

I honestly don't know how the clang-format support in VS2019 works against these goals. I can try out the file from this PR on my own and see what I think...

csyonghe:
Yeah I can imagine issues may arise when someone cut a large page of code out and then paste them back again, that will trigger the format-on-paste the feature leading to large blocks of changes. Also format on save can be another feature that will cause problems.

jsmall-nvidia:
Can we put this file in say a extras folder or some other suitable name, as a thing people can use if they want to? It seems useful to have and something we want in repro and up to date if people use it. 

It could stay where it is too potentially - I guess my worry is that it is automatically picked up by Visual Studio (which depending on version might do some odd stuff), and that could be irritating confusing to new users if they don't know what's going on or how to disable easily.

tfoleyNV:
So where did we all land on this?

* Can/should we check in the file to a location that VS2019 won't pick up by default so that we can use it for manual formatting while also revamping the configurations?

* Can/should we check the file in at the top level, but make sure that everybody using VS2019 knows that they *must* disable its automatic whole-file formatting and at worst should leave on as-you-type formatting (if they find it works for them)?

I'm fine with either resolution since I don't use VS2019 right now, but I will plan to very vigorously point out and request resolution of PRs that include formatting changes on code that isn't related to the work being done. I would expect our existing golden rule of "leave well enough alone" will continue to trump the benefits of automatic formatting for a while longer.

That said, I'd like to get this checked in if only so that I can play with it and see if I can get the automated formatting to be something to my liking and then make my own PR to commit those tweaks. :)

csyonghe:
Checked in in `extras` sub dir.

tfoleyNV:
Okay, let's plan to merge this at some point when the dust settles on other PRs.

PR #1375, opened 2020-06-05
Fix FindTypeByName reflection API not finding stdlib types.

### LINKED ISSUE #1357: Namespace support breaks Reflection::FindTypeByName
`FindTypeByName("ParameterBlock<MyStruct>")` returns `ErrorType` since #1304 

It seems that lookup for `ParameterBlock` failed when type-checking the parsed expression tree.

csyonghe:
Fixes #1357

### DIFF
--- a/source/slang/slang-check-shader.cpp
+++ b/source/slang/slang-check-shader.cpp
@@ -1290,7 +1290,7 @@ namespace Slang
         // it defines keywords like `true` and `false`).
         //
         RefPtr<Scope> scope = new Scope();
-        scope->parent = getLinkage()->getSessionImpl()->baseLanguageScope;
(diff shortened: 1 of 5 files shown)


source/slang/slang-check-shader.cpp
```diff
@@ -1290,7 +1290,7 @@ namespace Slang
         // it defines keywords like `true` and `false`).
         //
         RefPtr<Scope> scope = new Scope();
-        scope->parent = getLinkage()->getSessionImpl()->baseLanguageScope;
+        scope->parent = getLinkage()->getSessionImpl()->slangLanguageScope;
```
tfoleyNV, line 1293: Great catch here.

PR #1376, opened 2020-06-05
ASTNodes use MemoryArena

jsmall-nvidia:
* Ast nodes no longer derive from RefObject
* There is no reference counting on AST Nodes
* The TypeCheckingCache is moved from Session to the Linkage 
  * Otherwise there is a scoping problem with ASTBuilder
* Some other small fixes and improvements
* Associate a name with ASTBuilder to better track scope + id (for debugging)
* Memory for backing AST nodes is from MemoryArena on ASTBuilder
* Only track (and dtor) AST types that have a non trivial dtor

On testing on the profile timing goes from 12.5 seconds to 8.3 seconds. Dll size goes from 2.49Mb to 2.17Mb.

### DIFF
--- a/source/slang/slang-ast-support-types.h
+++ b/source/slang/slang-ast-support-types.h
@@ -53,7 +53,7 @@ namespace Slang
     SourceLoc const& getDiagnosticPos(SyntaxNode const* syntax);
     SourceLoc const& getDiagnosticPos(TypeExp const& typeExp);
 
-    typedef RefPtr<RefObject> (*SyntaxParseCallback)(Parser* parser, void* userData);
+    typedef NodeBase* (*SyntaxParseCallback)(Parser* parser, void* userData);
 
     typedef unsigned int ConversionCost;
     enum : ConversionCost
--- a/source/slang/slang-ast-support-types.h
+++ b/source/slang/slang-ast-support-types.h
@@ -43,6 +43,8 @@ namespace Slang
     struct TypeExp;
     class Val;
 
+    class NodeBase;
+
     void printDiagnosticArg(StringBuilder& sb, Decl* decl);
     void printDiagnosticArg(StringBuilder& sb, Type* type);
     void printDiagnosticArg(StringBuilder& sb, TypeExp const& type);
@@ -454,6 +456,7 @@ namespace Slang
         typedef ReflectClassInfo ThisType;
 
         typedef void* (*CreateFunc)(ASTBuilder* astBuilder);
+        typedef void (*DestroyFunc)(void* ptr);
 
         /// A constant time implementation of isSubClassOf
         SLANG_FORCE_INLINE bool isSubClassOf(const ThisType& super) const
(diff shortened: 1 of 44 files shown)


source/slang/slang-ast-support-types.h
```diff
@@ -485,6 +494,7 @@ namespace Slang
         const ReflectClassInfo* m_superClass;       ///< The super class of this class, or nullptr if has no super class. 
         const char* m_name;                         ///< Textual class name, for debugging 
         CreateFunc m_createFunc;                    ///< Callback to use when creating instances
+        DestroyFunc m_destroyFunc;                  ///< Called to inplace destroy (ie not backing memory) of this type
```
tfoleyNV, line 497: The lack of symmetry between the `Create` and `Destroy` cases here is a bit subtle and unfortunate, but probably not worth addressing at this point.

source/slang/slang-check-expr.cpp
```diff
@@ -1324,9 +1324,11 @@ namespace Slang
                             // of explicit default initializers for `struct` fields to
                             // make this a major concern (since they aren't supported in HLSL).
                             //
-                            RefPtr<InitializerListExpr> initListExpr = m_astBuilder->create<InitializerListExpr>();
+                            InitializerListExpr* initListExpr = m_astBuilder->create<InitializerListExpr>();
                             auto checkedInitListExpr = visitInitializerListExpr(initListExpr);
-                            return coerce(typeExp.type, initListExpr);
+
+                            // TODO(JS): I changed this to checkInitListExpr form initListExpr
```
tfoleyNV, line 1330: Your change here is more correct than the original code, which probably worked by accident, since `visitInitializerListExpr` would always return the same AST node that was passed in.

source/slang/slang-check-impl.h
```diff
@@ -899,42 +899,43 @@ namespace Slang
...
-        RefPtr<Val> createTypeWitness(
-            RefPtr<Type>            subType,
-            DeclRef<AggTypeDecl>    superTypeDeclRef,
-            TypeWitnessBreadcrumb*  inBreadcrumbs);
+        /// Create a withness that `subType` is a sub-type of `superTypeDeclRef`.
+        ///
+        /// The `inBreadcrumbs` parameter represents a linked list of steps
+        /// in the process that validated the sub-type relationship, which
+        /// will be used to inform the construction of the witness.
+        ///
```
tfoleyNV, line 933: Indenting and spacing changed here from the original.

(Also I now see a really glaring typo in my comment: `withness`).

source/slang/slang-check-shader.cpp
```diff
@@ -1455,12 +1455,14 @@ namespace Slang
         EndToEndCompileRequest::EntryPointInfo const&   entryPointInfo)
     {
         auto sink = endToEndReq->getSink();
-        auto entryPointFuncDecl = unspecializedEntryPoint->getFuncDecl();
+
+        // TODO(JS): Not used
+        //auto entryPointFuncDecl = unspecializedEntryPoint->getFuncDecl();
```
tfoleyNV, line 1460: This is fine to delete, then.

PR #1378, opened 2020-06-06
Filter lookup results from interfaces in `visitMemberExpr`.

### LINKED ISSUE #1377: Regression in multi-level associated type lookup
```
// assoctype-nested.slang

// Confirm that an associated type can be correctly looked up.

//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -cpu
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute

interface IBoneWeightSet
{
    associatedtype PackedType;
};

struct StandardBoneWeightSet : IBoneWeightSet
{
    struct PackedType
    {
        uint boneIds : BONEIDS;
        uint boneWeights : BONEWEIGHTS;
    };
    PackedType field;
};

interface IVertexFormat
{
    associatedtype BoneWeightSet : IBoneWeightSet;
};

struct VertexFormat<TBoneWeightSet : IBoneWeightSet> : IVertexFormat
{
    typedef TBoneWeightSet BoneWeightSet;
    TBoneWeightSet.PackedType weightSet;
};

struct OutputType
{
    VertexFormat<StandardBoneWeightSet>.BoneWeightSet.PackedType field;
};

int test(int val)
{
    OutputType rs;
    rs.field.boneIds = 256;
    return rs.field.boneIds;
}

[numthreads(4, 1, 1)]
{
    int inputVal = tid;
    int outputVal = test(inputVal);
}
```

Expected result: outputs 256.
Actual result: compilation failed, error mesesage: "tests/compute/assoctype-lookup.slang(38): error 30027: 'PackedType' is not a member of 'overload group'."

csyonghe:
Fixes #1377

This change is intended to fix only #1377. It might not be general enough. For example, I used `declRef.parent.as<InterfaceDecl>()` to test if a lookup result is an abstract requirement. There might already be some helper functions out there for such tests. There might also be other cases that need filtering.


tfoleyNV:
Our approach to resolving which declaration is being referred to in cases of overloading is starting to have a lot of ad hoc tweaks to deal with the combination of inherited and "direct" declarations in the candidate set. This kind of issue would only get worse if we start to introduce more general forms of inheritance and/or the ability to override inherited declarations.

At the same time, we also have some related issues when a method on a type wants to reference a global function with the same name as a member function of its type (such that overload resolution should be able to disambiguate things).

It seems that we are close to the point of needing a general-purpose system for representing the "priority" of a particular lookup result relative to the others, so that we can easily collapse an overloaded result by dismissing everything that doesn't have the "best" priority (to be used for things like direct member reference as is happening here). That priority could be attached to overload candidates at call sites and be used as a final disambiguation metric (after all the type-based resolution has occured).

The biggest missing thing here is a really concrete plan for what the priority should look like.

PR #1382, opened 2020-06-10
Fix problem with C++ extractor ernoneous concating of type tokens 

jsmall-nvidia:
When producing text from a stream of token care must be taken, as some concats can change meaning.

For example previously

```
'Thing' 'const' '*' -> Thingconst*
```
The change checks consecutive token types and adds a space if needed.


jsmall-nvidia:
> I assume this comes up because the representation of types in the extractor is currently just an unstructured sequence of tokens rather than anything like C++ declarator syntax?

Yes. This is a fairly ad hoc solution. If the complexity gets much higher, just outputting spaces might be ticket.

PR #1383, opened 2020-06-10
Add compiler flag to disable specialization pass.

csyonghe:


source/slang/slang-options.cpp
```diff
@@ -557,6 +557,10 @@ struct OptionsParser
                 {
                     requestImpl->getFrontEndReq()->useSerialIRBottleneck = true;
                 }
+                else if (argStr == "-allow-dynamic-code")
+                {
+                    requestImpl->getBackEndReq()->allowDynamicCode = true;
+                }
```
tfoleyNV, line 563: We should probably "bikeshed" the name for this option.

The current behavior of this flag is more like `-disable-specialization`, rather than `-allow-dynamic-code`, but I guess you might be picking the option name figuring that it will eventually enable a more nuanced mode in the compiler that might specialize some things but not others, rather than this current all-or-nothing behavior.

Issue #1374, opened 2020-06-05
Variable referencing itself will cause stack overflow


jsmall-nvidia:
If we have a static variable in a struct that references itself, Slang will stack overflow

```
struct Thing
{
    static const int kValue = kValue;
};
```

With a local variable that references itself (such as z = z;) we do get an appropriate error

```
tests/bugs/variable-self-reference.slang(21): fatal error 39999: local variable 'z' is being used before its declaration.
```

Trying something a little more out there 

```
struct Thing
{
    static const int kOther = kValue;
    static const int kValue = kOther;
};
```

Also leads to a stack overflow.

csyonghe:
This is fixed.

PR #1384, opened 2020-06-11
Diagnose circularly-defined constants

tfoleyNV:
Work on #1374

This change diagnoses cases like the following:

```hlsl
static const int kCircular = kCircular;

static const int kInfinite = kInfinite + 1;

static const int kHere = kThere;
static const int kThere = kHere;
```

By diagnosing these as errors in the front-end we protect against infinite recursion leading to stack overflow crashes.

The basic approach is to have front-end constant folding track variables that are in use when folding a sub-expression, and then diagnosing an error if the same variable is encountered again while it is in use. In order to make sure the error occurs whether or not the constant is referenced, we invoke constant folding on all `static const` integer variables.

Limitations:

* This only works for integers, since that is all front-end constant folding applies to. A future change can/should catch circularity in constants at the IR level (and handle more types).

* This only works for constants. Circular references in the definition of a global variable are harder to diagnose, but at least shouldn't result in compiler crashes.

* This doesn't work across modules, or through generic specialization: anything that requires global knowledge won't be checked


source/slang/slang-check-expr.cpp
```diff
@@ -795,8 +796,56 @@ namespace Slang
         return result;
     }
 
-    IntVal* SemanticsVisitor::TryConstantFoldExpr(
-        Expr* expr)
+    bool SemanticsVisitor::_checkForCircularityInConstantFolding(
+        Decl*                           decl,
+        ConstantFoldingCircularityInfo* circularityInfo)
```
jsmall-nvidia, line 801: As an alternative it could use a bit in DeclCheckStateExt, or have some variable (as lots of unused bytes) in the decl that could be used to check without having to track a stack held linked list.

source/slang/slang-check-expr.cpp
```diff
@@ -795,8 +796,56 @@ namespace Slang
...
+        if(!decl->hasModifier<HLSLStaticModifier>())
+            return nullptr;
+        if(!decl->hasModifier<ConstModifier>())
+            return nullptr;
+
+        auto initExpr = getInitExpr(m_astBuilder, declRef);
+        if(!initExpr)
+            return nullptr;
+
+        ConstantFoldingCircularityInfo newCircularityInfo;
```
jsmall-nvidia, line 839: Using a ctor for ConstantFoldingCircularityInfo would be more compact in usage and less fragile.

source/slang/slang-check-decl.cpp
```diff
@@ -775,6 +775,24 @@ namespace Slang
         return true;
     }
 
+    void SemanticsVisitor::_validateCircularVarDefinition(VarDeclBase* varDecl)
+    {
+        // The easiest way to test if the declaration is circular is to
+        // validate it as a constant.
+        //
+        // TODO: The logic here will only apply for `static const` declarations
```
jsmall-nvidia, line 783: Whilst this does root out the more subtle scenarios - this problem was originally identified with the most straightforward 

```
static const SomeType kThing = kThing;
```

Can't we also have an additional simple test for this case and it doesn't have the requirement of only being constant foldable (and therefore currently int)?

source/slang/slang-check-expr.cpp
tfoleyNV: That's true, although I'd like us to be slowly moving away from storing semantic-checking state on the AST nodes at all. Eventually the `DeclCheckState` for declarations should be stored in a `Dictionary<Decl*, DeclCheckStateExt>` owned by semantic checking, since it doesn't need to outlive the semantic checking step.

The state in this case is sort of local to the checking of a single declaration, so if we stored it globally (either as a global dictionary on the semantic checking context, or on the AST nodes themselves) we'd have to clear it out when we are done (in an RAII-like way that mirrors what we are constructing on the stack here anyway).

source/slang/slang-check-expr.cpp
tfoleyNV: Yes, it is probably worth doing that as a convenience/cleanup.

source/slang/slang-check-decl.cpp
tfoleyNV: To be clear, there are two main places in the compiler where this kind of circularity can cause a crash:

1. In the front-end, when folding `int` constants

2. In the AST->IR lowering step, when dealing with `static const`s of all types

This PR is intentionally just addressing (1) by adding protection to the front-end constant-folding path (and then hijacking that logic to also provide a guaranteed error message even if a constant is never referenced).

I'd like to attack (2) with a second PR, and that PR would catch the case you show as well as others.

The current PR isn't meant to dismiss the importance of non-`int` types; it is only as narrowly scoped as it is because the constant folding logic that we are trying to protect is only applicable to integers.

PR #1387, opened 2020-06-12
Generate IRType for interfaces, and reference them as `operand[0]` in IRWitnessTable values

csyonghe:
This results the following IR for the included test case:
```
[export("_S3tu010IInterface7Computep1pii")]
let  %1 : _     = key
[export("_ST3tu010IInterface")]
[nameHint("IInterface")]
let  %IInterface        : _     = interface(%1)
witness_table %2        : _(%IInterface);
```


source/slang/slang-ast-support-types.h
```diff
@@ -1285,6 +1285,7 @@ namespace Slang
     struct WitnessTable : RefObject
     {
         RequirementDictionary requirementDictionary;
+        Type* baseType;
```
tfoleyNV, line 1288: This probably merits a comment. I assume that `baseType` represents the type that the witness table witnesses conformance *to* (currently an interface), and not the type it witnesses conformance *of*.

source/slang/slang-ir.cpp
```diff
@@ -4382,6 +4391,7 @@ namespace Slang
 
         case kIROp_WitnessTable:
         case kIROp_StructType:
+        case kIROp_InterfaceType:
```
tfoleyNV, line 4394: I'm not 100% convinced that the way we are dumping IR interfaces here is the ideal, since they are currently *not* parent instructions (they have zero children, always), and they operate more like ordinary instructions at this point.

Of course, the dumping logic is not critical, so I'm fine with changing this up later.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -4489,6 +4489,10 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         Dictionary<WitnessTable*, IRWitnessTable*>  mapASTToIRWitnessTable)
     {
         auto subBuilder = subContext->irBuilder;
+        SLANG_ASSERT(as<DeclRefType>(astWitnessTable->baseType));
+        auto baseTypeLowerInfo = ensureDecl(subContext, as<DeclRefType>(astWitnessTable->baseType)->declRef.getDecl());
+        SLANG_ASSERT(baseTypeLowerInfo.val);
+        irWitnessTable->setFullType(as<IRType>(baseTypeLowerInfo.val));
```
tfoleyNV, line 4495: We should be able to streamline this logic and avoid the assertions:

```
IRType* irBaseType = lowerType(subContext, astWitnessTable->baseType);
```

tfoleyNV: Also, the `irWitnessTable->setFullType()` here worries me.

Given a witness table `w` that witnesses the conformance of concrete type `Foo` to interface `IBar`, this code appears to be saying that the type of `w` is `IBar`, but that doesn't seem accurate. You couldn't pass `w` to a function that expects an `IBar` parameter.

The current IR logic instead says that an `IRWitnessTable` always has the singular "witness table type" (all witness tables have the same type), and I think that is well motivated: when an `IRGeneric` has a parameter for a witness table, it is expecting you to pass a value that has the matching type.

I believe what is desired here is to have the `IRWitnessTable` instruction have an *operand* that represents the interface being conformed to, something like:

```
witness_table %w : WitnessTable ( %IBar )
{
    ...
}
```

This seems more accurate: `w` is a witness table (so its type is `WitnessTable`), and we can find the interface that it is representing a conformance to as its first (only) operand, while the actual requirement witnesses are referenced as children of the witness table (for now).

I can imagine that we might later update witness table to also reference the type for which they represent a conformance, so that a witness table for `Foo : IBar` would reference both `IBar` and `Foo` as operands. (There may be arguments against taking that step...).

source/slang/slang-ast-support-types.h
```diff
@@ -1285,6 +1285,9 @@ namespace Slang
     struct WitnessTable : RefObject
     {
         RequirementDictionary requirementDictionary;
+
+        // The type that the witness table witnesses conformant to (e.g. an Interface)
```
tfoleyNV, line 1289: Typo: conformant->comformance.

PR #1389, opened 2020-06-13
Specialize function calls involving array arguments.

### LINKED ISSUE #890: Poor performance when reading float4x4 uniform with dynamic array index
This may or may not be an issue of Slang, but I have observed a strange performance issue in my shader that is probably worth noting here.

On Vulkan, I defined a `uniform float4x4` array in my shader code (in this case,  the `lightMatrix` for cascaded shadow maps), and as long as I am accessing the array in a loop, the performance drops dramatically (from ~200fps to 20fps), even after I changed loop to do nothing else but accessing the data.

The performance issue is solved by manually unrolling the loop, and make sure the shader is only using constants as indices into the array. I also tried to just define `float4` arrays, and construct a `float4x4` by manually loading 4 `float4`s, but that doesn't seem to work either. However, the same issue does not exist for `float4x3` arrays. 

Note that it does not matter whether the code is actually executed or not, the existence of such code causes the slow down of entire pixel shader execution, even if the access is in a branch that is never executed.

I feel that this may have something to do with `row_major`, and if that is the case, then we definitely should generate code to handle all input fetching ourselves (as we discussed last time).

csyonghe:
Fixes #890.

On Vulkan, the Slang generated code performs poorly if a `ParameterBlock` contains a large array. The driver compiler seemed unable to optimize out the intermediate copies if the array is accessed from a loop. Per discussion in #890, we need to specialize functions to call sides if the function takes in a struct type with a large array as argument. This PR does that.

This fix brings a 2x speedup to SpireEngine (https://github.com/spire-engine/spire-engine) when using Vulkan.


source/slang/slang-ir-specialize-resources.cpp
```diff
@@ -171,6 +171,27 @@ struct ResourceParameterSpecializationContext
         return anySpecializableParam;
     }
 
+    // Returns true if `type` is an `IRStructType` with array-typed fields.
+    bool isStructTypeWithArray(IRType* type)
```
tfoleyNV, line 175: Hypothetically, we could cache the results of this check on the `IRType` by introducing a decoration like `[containsArrays(true)]` or `[containsArrays(false)]`. That would avoid the need to re-run this recursive analysis on the same type over and over again.

csyonghe:
On testing, it will be nice if we have something like llvm FileCheck and run pattern matching on the resulting HLSL/GLSL. For now it seems difficult to include such a test.

tfoleyNV:
Yes, switching our test system away from requiring an exact/complete match over to something closer to FileCheck is on the to-do list. As you know, though, that starts to invite questions of why we aren't just depending on LLVM and using FileCheck as-is.

PR #1391, opened 2020-06-16
Generate dynamic C++ code for the minimal test case.

### LINKED ISSUE #1386: Implement minimal support for dynamic dispatch code gen
Here's a minimal test case for dynamic dispatch code gen:

```
//TEST(compute):COMPARE_COMPUTE:-cpu -xslang -allow-dynamic-code

// Test basic dynamic dispatch code gen

interface IInterface
{
	static int Compute(int inVal);
};

int GenericCompute<T:IInterface>(int inVal)
{
	return T.Compute(inVal);
}

struct Impl : IInterface
{
	static int Compute(int inVal) { return inVal * inVal; }
};

int test(int inVal)
{
	return GenericCompute<Impl>(inVal);
}

//TEST_INPUT:ubuffer(data=[0 1 2 3], stride=4):out,name=outputBuffer
RWStructuredBuffer<int> outputBuffer : register(u0);

[numthreads(4, 1, 1)]
{
	int inVal = outputBuffer[tid];
	int outVal = test(inVal);
	outputBuffer[tid] = outVal;
}
```

csyonghe:
This PR includes two major changes to enable C++ code generation for the minimal dynamic-dispatch test case.

The first part is to add an IR pass to lower generics into ordinary functions. To generate dynamic dispatch code for generic functions, we need to turn generic functions into ordinary functions. This include two steps:

1. Identify `call(specialize(genericValue, TArgs), args)` instruction sequence and turn it into an ordinary function `call(ordinaryFunc, args+TArgs)`.
2. Turn all generic functions into ordinary functions by appending the generic parameters at the end of the inner functions's parameter list.

Intrinsic generic functions should not be transformed in this pass.

The second part is to implement the code emitting logic in `CPPSourceEmitter`. Each `interface` type translates into a `struct` of function pointers for each requirement in the interface. A witness table translates to a global variable of this `struct` type. Because all functions are emitted as a member function of the `Context` class, the actual definition of the witness table must come after the `Context` struct has been emitted.

Fixes #1386.


source/slang/slang-ir-inst-defs.h
```diff
@@ -166,7 +166,14 @@ INST(Nop, nop, 0, 0)
 INST(StructType, struct, 0, PARENT)
 INST(InterfaceType, interface, 0, PARENT)
 
-INST_RANGE(Type, VoidType, InterfaceType)
+// A TypeType-typed IRValue represents a IRType.
+// It is used to represent a type parameter/argument in a generics.
+INST(TypeType, type_t, 0, 0)
```
tfoleyNV, line 171: There is already a `TypeKind` instruction, so that having both `TypeType` and `TypeKind` could potentially be confusing. There are really two main options here, and we need to think carefully about which seems correct:

1. Only keep `TypeKind`, and have it so that after your transformation we may have IR functions with parameters classified by `TypeKind`. This means that the distinction between the type and value levels gets a bit muddy, since we might have value-level operations that are manipulation type-level entities.

2. Have both `TypeKind` as the classifier for generic type parameters, and `TypeType` as the classifier for value-level type parameters (with the understanding that is is really the type of RTTI/reflection information). In this case your pass would translate a generic parameter classified by `TypeKind` over to a value parameter classified by `TypeType`. In this case we might need an explicit operation to go from a type-level instruction classified by `TypeKind` to a value-level instruction classified by `TypeType`, which represents the operation of looking up RTTI/reflection information for a given type.

My gut instinct is to go for option (1) there, and only consider the more complicated option if we run into serious challenges. That would argue for *not* introducing `TypeType` and instead using the existing `TypeKind`.

source/slang/slang-ir-inst-defs.h
```diff
@@ -166,7 +166,14 @@ INST(Nop, nop, 0, 0)
...
+
+// An `IRWitnessTable` has type `WitnessTableType`.
+INST(WitnessTableType, witness_table_t, 0, 0)
+
+INST_RANGE(Type, VoidType, WitnessTableType)
```
tfoleyNV, line 176: Similarly, but less important: we should consider whether witness tables are classified by a `WitnessTableType` or by a `WitnessTableKind`. For the most part they are type-level entities in practice, so I suppose that `WitnessTableKind` might be more accurate.

The same basic choice space arises as for `TypeType` and `TypeKind`, where we need to decide whether to use a single classifier for all cases (both generic and value-level parameters), or to distinguish the use cases as different classifiers.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -0,0 +1,159 @@
...
+        IRModule* module;
+
+        Dictionary<IRInst*, IRInst*> loweredGenericFunctions;
+
+        SharedIRBuilder sharedBuilderStorage;
+
+        // We will use a single work list of instructions that need
+        // to be considered for specialization or simplification,
+        // whether generic, existential, etc.
+        //
```
tfoleyNV, line 25: Some of these comments seem to be copy-pasted from the specialization pass, and we might want to tailor them a bit more to what is going on.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -0,0 +1,159 @@
...
+                return result;
+            auto genericParent = as<IRGeneric>(genericValue);
+            SLANG_ASSERT(genericParent);
+            auto func = as<IRFunc>(findGenericReturnVal(genericParent));
+            SLANG_ASSERT(func);
+            if (!func->isDefinition())
+            {
+                loweredGenericFunctions[genericValue] = genericValue;
+                return genericValue;
+            }
```
tfoleyNV, line 62: It seems like we would eventually want to support the case where the generic function is an imported declaration rather than a definition, so that we can lower its signature as expected and then rewrite call sites to invoke it as expected.

I guess that won't work in the short term where we have a *lot* of built-in functions that we really need to invoke as generics with implicit full specialization using the templates in the "prelude" file. Trying to invoke those things as dynamic-dispatch generics today would require massive changes to the prelude file and completely tank the performance of our CPU code.

source/slang/slang.vcxproj
```diff
@@ -1,415 +1,417 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
tfoleyNV, line 1: We should double-check why this diff is so massive.

In general, the Right Way to generate the project files is to invoke the copy of `premake5.exe` that is checked into the `slang-binaries` submodule. By doing this, we can guarantee that all the developers/contributors are on a consistent version of the tool, which we hope will generate project files with minimal diffs.

csyonghe:
For reference, here's the generated C++ code for the test case:
```
#include "D:/git_repo/slang/prelude/slang-cpp-prelude.h"


struct IInterface_0;
struct IInterface_0
{
    int32_t (Context::*_S3tu010IInterface7Computep1pii)(int32_t inVal_0);

};
extern IInterface_0 _SW3tu04Impl3tu010IInterface;
struct Impl_0
{
};

struct UniformState
{
    RWStructuredBuffer<int32_t> outputBuffer_0;



};

struct Context
{
    UniformState* uniformState;
    uint3 dispatchThreadID;
    uint3 groupID;
    uint3 groupDispatchThreadID;
    uint3 calcGroupThreadID() const 
    {
        uint3 v = { dispatchThreadID.x - groupDispatchThreadID.x, dispatchThreadID.y - groupDispatchThreadID.y, dispatchThreadID.z - groupDispatchThreadID.z };
        return v;
    }
    int32_t Impl_Compute_0(int32_t inVal_0)
    {
        return inVal_0 * inVal_0;
    }

    int32_t GenericCompute_0(int32_t inVal_1, IInterface_0* _S1)
    {
        int32_t _S2 = (this->*(_S1->_S3tu010IInterface7Computep1pii))(inVal_1);
        return _S2;
    }

    int32_t test_0(int32_t inVal_2)
    {
        int32_t _S3 = GenericCompute_0(inVal_2, (&_SW3tu04Impl3tu010IInterface));
        return _S3;
    }

    void _computeMain()

    {
        uint32_t tid_0 = dispatchThreadID.x;
        int32_t outVal_0 = test_0((uniformState->outputBuffer_0)[tid_0]);
        (uniformState->outputBuffer_0)[tid_0] = outVal_0;
        return;
    }

};

IInterface_0 _SW3tu04Impl3tu010IInterface = {
    &Context::Impl_Compute_0
};
// [numthreads(4, 1, 1)]
SLANG_PRELUDE_EXPORT
void computeMain_Thread(ComputeThreadVaryingInput* varyingInput, UniformEntryPointParams* params, UniformState* uniformState)
{
    Context context = {};
    context.uniformState = uniformState;
    context.dispatchThreadID = {
        varyingInput->groupID.x * 4 + varyingInput->groupThreadID.x,
        varyingInput->groupID.y * 1 + varyingInput->groupThreadID.y,
        varyingInput->groupID.z * 1 + varyingInput->groupThreadID.z
    };
    context._computeMain();
}
// [numthreads(4, 1, 1)]
SLANG_PRELUDE_EXPORT
void computeMain_Group(ComputeVaryingInput* varyingInput, UniformEntryPointParams* params, UniformState* uniformState)
{
    Context context = {};
    context.uniformState = uniformState;
    const uint3 start = {
        varyingInput->startGroupID.x * 4,
        varyingInput->startGroupID.y * 1,
        varyingInput->startGroupID.z * 1
    };
    context.dispatchThreadID = start;
    for (uint32_t x = start.x; x < start.x + 4; ++x)
    {
        context.dispatchThreadID.x = x;
        context._computeMain();
    }
}
// [numthreads(4, 1, 1)]
SLANG_PRELUDE_EXPORT
void computeMain(ComputeVaryingInput* varyingInput, UniformEntryPointParams* params, UniformState* uniformState)
{
    Context context = {};
    context.uniformState = uniformState;
    const uint3 start = {
        varyingInput->startGroupID.x * 4,
        varyingInput->startGroupID.y * 1,
        varyingInput->startGroupID.z * 1
    };
    const uint3 end = {
        varyingInput->endGroupID.x * 4,
        varyingInput->endGroupID.y * 1,
        varyingInput->endGroupID.z * 1
    };
    for (uint32_t z = start.z; z < end.z; ++z)
    {
        context.dispatchThreadID.z = z;
        for (uint32_t y = start.y; y < end.y; ++y)
        {
            context.dispatchThreadID.y = y;
            for (uint32_t x = start.x; x < end.x; ++x)
            {
                context.dispatchThreadID.x = x;
                context._computeMain();
            }
        }
    }
}

```

source/slang/slang-emit-c-like.cpp
```diff
@@ -217,8 +217,34 @@ void CLikeSourceEmitter::emitSimpleType(IRType* type)
...
-}
-
+}
+
+List<IRWitnessTableEntry*> CLikeSourceEmitter::getSortedWitnessTableEntries(IRWitnessTable* witnessTable)
+{
+    List<IRWitnessTableEntry*> sortedWitnessTableEntries;
+    auto interfaceType = cast<IRInterfaceType>(witnessTable->getOperand(0));
+    auto witnessTableItems = witnessTable->getChildren();
+    for (UInt i = 0; i < interfaceType->getOperandCount(); i++)
```
tfoleyNV, line 227: For now I think the simple code is warranted, but this logic is yet another case of the Slang compiler being "accidentally quadratic" in ways that could bite us later. In this case, the performance of `getSortedWitnessTableEntries` is going to be quadratic in the number of requirements/entries.

We can obviously fix this down the line if the performance becomes an issue, but it might be worth making a note of the lurking performance problem and a  basic idea of how to resolve it at the cost of more complicated code.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1849,6 +1882,27 @@ void CLikeSourceEmitter::emitIntrinsicCallExprImpl(
...
 
+void CLikeSourceEmitter::_emitCallArgList(IRCall* inst)
+{
+    m_writer->emit("(");
+    UInt argCount = inst->getOperandCount();
+    for (UInt aa = 1; aa < argCount; ++aa)
+    {
+        auto operand = inst->getOperand(aa);
+        if (as<IRVoidType>(operand->getDataType()))
+            continue;
```
tfoleyNV, line 1893: Ideally we should have an earlier IR pass that is responsible for eliminating all `void`-type parameters, local variables, fields, etc. rather than the current logic we have in the emit passes to tiptoe around `void`.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1849,6 +1882,27 @@ void CLikeSourceEmitter::emitIntrinsicCallExprImpl(
...
+
+        // TODO: [generate dynamic dispatch code for generics]
+        // Pass RTTI object here. Ignore type argument for now.
+        if (as<IRType>(operand))
+            continue;
+
+        if (aa != 1) m_writer->emit(", ");
```
tfoleyNV, line 1900: How do these `continue` cases work together with the `a != 1` logic for emitting `,`? It seems like if the first argument is `void` or a type then we won't emit anything, and then the next argument (which is the first *actual* argument) will start out by emitting a `,`.

source/slang/slang-emit-cpp.cpp
```diff
@@ -1561,7 +1571,117 @@ void CPPSourceEmitter::emitParamTypeImpl(IRType* type, String const& name)
...
+    emitSimpleType(interfaceType);
+    m_writer->emit(" ");
+    m_writer->emit(getName(witnessTable));
+    m_writer->emit(";\n");
+
+    // The actual definition of this witness table global variable
+    // is deferred until the entire `Context` class is emitted, so
+    // that the member functions are available for reference.
+    // The witness table definition emission logic is defined in the
+    // `_emitWitnessTableDefinitions` function.
```
tfoleyNV, line 1590: I was initially going to say that needing to work around the `Context` type here was unfortunate because it stops the witness tables from being `static const` data, but then I realized that in the pursuit of separate compilation we would eventually need the witness tables to have `extern` linkage so that another compiled module can link against them by their mangled name.

source/slang/slang-emit-cpp.cpp
```diff
@@ -1561,7 +1571,117 @@ void CPPSourceEmitter::emitParamTypeImpl(IRType* type, String const& name)
...
+    }
+}
+
+void CPPSourceEmitter::emitInterface(IRInterfaceType* interfaceType)
+{
+    // The current IRInterfaceType defintion does not contain
+    // sufficient info for emitting a witness table struct by itself
+    // Instead, it defines the order of entries in a witness table.
+    // Therefore, we emit a forward declaration here, and actual definition
+    // for the witness table type during emitWitnessTable.
```
tfoleyNV, line 1634: This seems a bit worrying in the long run, since it amounts to saying that the first witness table for an interface gets to dictate the signature of the interface methods for all other implementations of that interface.

It seems like we will need to work towards a situation where the interface type records not only the keys for its requirements, but also the expected type of the value that satisfies the requirement. The big challenge there is what type should be used to stand in for the `This` type, associated types, etc.

This is obviously something we should defer solving until it becomes critical.

source/slang/slang-emit.cpp
```diff
@@ -273,6 +274,18 @@ Result linkAndOptimizeIR(
     if (!compileRequest->allowDynamicCode)
         specializeModule(irModule);
 
+    switch (target)
+    {
+    case CodeGenTarget::CPPSource:
+        // For targets that supports dynamic dispatch, we need to lower the
+        // generics / interface types to ordinary functions and types using
+        // function pointers.
+        lowerGenerics(irModule);
```
tfoleyNV, line 283: Is `lowerGenerics` a sufficiently descriptive name? I can't think of something strictly better.

Something like `lowerGenericFunctionsToOrdinaryFunctions` is more explicit about what happens right now, but ignores the fact that the pass is likely to also handle lowering of generic types, witness tables, etc. in the long run.

PR #1395, opened 2020-06-17
Bug fix: Include prelude when outputting source

jsmall-nvidia:
When appending a prelude, during emit, the downstream compiler is looked up, and then the prelude associated with that compiler is used.

That when the output is source though no downstream compiler is *required* and so no prelude was being inserted. The change here tries to determine what downstream compiler will be used, but does not require it (requiring also implies loading it and other work).

This is probably fine for now, but probably means that associating a prelude with the downstream compiler was probably not the right answer. The idea was you could vary the prelude on compiler and take that into account. This added flexibility doesn't really add much so in the end is probably worse than just associating a prelude with an output language. 

Will do separate PR for associating with a language at future point.


source/slang/slang-compiler.cpp
```diff
@@ -418,18 +418,26 @@ namespace Slang
...
+                // Use the default compiler for the source language if set, 
+                DownstreamCompiler* downstreamCompiler = session->getDefaultDownstreamCompiler(SourceLanguage::HLSL);
+                if (downstreamCompiler)
+                {
+                    return PassThroughMode(downstreamCompiler->getDesc().type);
+                }
+                else
+                {
+                    // This is ambiguous, because we could use dxc or fxc. For now we'll go with Dxc. 
+                    return PassThroughMode::Dxc;
```
tfoleyNV, line 435: Do we eventually need to add `PassThroughMode::GenericHLSL` and then have it select between fxc/dxc based on the shader model?

source/slang/slang-emit.cpp
```diff
@@ -714,17 +714,35 @@ SlangResult emitEntryPointSourceFromIR(
...
-            // Get the compiler used for the language
-            DownstreamCompiler* compiler = session->getDefaultDownstreamCompiler(sourceLanguage);
-            if (compiler)
+            // TODO(JS):
+            // NOTE! This makes the *assumption* that if we are outputting source, we don't want the
+            // prelude for a specific compiler, we are happy to just use the GenericCCpp
+            // If we didn't do this, we would have to do the work out what's available etc.
+            //
+            // This is all a bit unfortunate. The decision to associate preludes on compilers seemed like a good idea, but was perhaps a mistake.
+            // That it would be simpler if prelude was based on output source type.
```
tfoleyNV, line 729: Yes, I think a follow-up change to make the prelude simply be determined by the `CodeGenTarget` could be nice to have.

jsmall-nvidia: You would prefer CodeGenTarget over SlangSourceLanguage?

tfoleyNV: Of the top of my head, `CodeGenTarget` seems more strictly correct, in that the prelude is always something we combine with generated (target) code.

I suppose one concern you might be thinking about is when the `CodeGenTarget` is the final destination target (e.g., `HostCallable`) while the prelude should be associated with the target we selected as a stepping stone to get there (e.g., `Cpp`). No matter what it seems like that scenario requires us to compute what the stepping-stone language/target should be, and the choice of whether that is a `CodeGenTarget` or a `SourceLanguage` isn't the most important.

(In practice, every textual language we support as a target is in *both* `CodeGenTarget` and `SourceLanguage`, so this is more of a philosophical point than anything else; either shoudl work)

Realistically, if we could revisit the fundamental design choices, it might be nice to have both `CodeGenTarget` and `SlangSourceLanguage` collapsed into a single `enum` that represents formats in which code can be expressed. Some day we might like Slang to be able to ingest something like a SPIR-V module as an input, at which point we'd either have to represent inputs using `CodeGenTarget` or extend `SlangSourceLanguage` to include things that aren't really "source" languages...

PR #1398, opened 2020-06-17
Prelude is associated with SourceLanguage

jsmall-nvidia:
This is on top of  #1395

This changes the prelude association to be with SlangSourceLanguage. 

The SourceStyle enum in the CLikeSourceEmitter was removed and SourceLanguage is now used instead.

The previous set/getDownstreamCompilerPrelude methods are depreciated, but still work, because it associates a default source language with each compiler.


slang.h
```diff
@@ -2909,6 +2913,25 @@ namespace slang
             @return The downstream compiler for that source language */
         virtual SlangPassThrough SLANG_MCALL getDefaultDownstreamCompiler(
             SlangSourceLanguage sourceLanguage) = 0;
+
+            /* Set the 'prelude' placed before generated code for a specific language type.
```
tfoleyNV, line 2917: I hadn't realized when we discussed `CodeGenTarget` vs `SlangSourceLanguage` that we were talking about public API.

I definitely think it is a bit odd to have this key on a "source language" but then the comment talks about "generated code."

I'm fine with this insofar as it works, but I definitely feel like this API would be less confusing if we dropped the word `Source` from `SlangSourceLanguage` (and/or merged `SlangSourceLanguage` with `CodeGenTarget`).

source/core/slang-test-tool-util.cpp
```diff
@@ -59,36 +59,19 @@ static SlangResult _addCPPPrelude(const String& parentPath, slang::IGlobalSessio
...
-        SLANG_PASS_THROUGH_CLANG,                   ///< Clang C/C++ compiler 
-        SLANG_PASS_THROUGH_VISUAL_STUDIO,           ///< Visual studio C/C++ compiler
-        SLANG_PASS_THROUGH_GCC,                     ///< GCC C/C++ compiler
-        SLANG_PASS_THROUGH_GENERIC_C_CPP,
-    };
-    for (auto downstreamCompiler : downstreamCompilers)
-    {
-        session->setDownstreamCompilerPrelude(downstreamCompiler, prelude.getBuffer());
-    }
+    session->setLanguagePrelude(SLANG_SOURCE_LANGUAGE_CPP, prelude.getBuffer());
```
tfoleyNV, line 64: This seems like a nice indication that the model of associating the prelude with a language instead of a compiler is the Right Choice.

PR #1399, opened 2020-06-17
Heterogeneous example

Checkmate50:
Introduced heterogeneous example.  Example includes C++ source and header files, and does not currently make use of the associated slang file when building.  The intent of this commit is to introduce the example as a baseline for later updates as the heterogeneous model is expanded.


examples/heterogeneous-hello-world/main.cpp
```diff
@@ -0,0 +1,108 @@
+// main.cpp
+
+#include <stdio.h>
+
+// This file implements an extremely simple example of loading and
```
jsmall-nvidia, line 5: This comment should probably be updated to explain the point of the example.

examples/cpu-hello-world/main.cpp
```diff
@@ -40,6 +40,8 @@ using namespace Slang;
 #define SLANG_PRELUDE_NAMESPACE CPPPrelude
 #include "../../prelude/slang-cpp-types.h"
 
+struct UniformState;
```
jsmall-nvidia, line 43: It doesn't look like this is needed?

PR #1403, opened 2020-06-18
Work on struct inheritance and interfaces

tfoleyNV:
The main new feature that works here is that a derived `struct` type can satisfy one or more interface requirements using methods it inherited from a base `struct` type:

```hlsl
interface ICounter { [mutating] void increment(); }

struct CounterBase { int val; [mutating] void increment() { val++; } }

struct ResetableCounter : CounterBase, ICounter
{
    [mutating] void reset() { val = 0; }
}
```

Here the derived `ResetableCounter` type is satisfying the `increment()` requirement from `ICounter` using the inherited `CounterBase` method instead of one defined on `ResetableCounter`.

The crux of the problem here was that after lowering to HLSL/GLSL, the above code looks something like:

```hlsl
struct CounterBase { int val; };

void CounterBase_increment(in out CounterBase this) { this.val++; }

struct ResetableCounter { CounterBase base; }

void ResetableCounter_reset(in out ResetableCounter this) { this.base.val = 0; }
```

The central problem is that `CounterBase_increment` here is not type-compatible what we expect to find in the witness table for `ResetableCounter : ICounter`: the `this` parameter has the wrong type!

The basic solution strategy here is to intercept the search for a witness to sastify an interface requirement in `findWitnessForInterfaceRequirement` (those witnesses get collected into a witness table). The revised logic first looks for an exact match, which will only consider members introduced for the type itself, and not those introduced by base types.

If an exact match for a method requirement is not found, the semantic checker then tries to *synthesize* a witness for the requirement, which more or less amounts to generating a function like:

```hlsl
[mutating] void ResetableCounter::synthesized_increment()
{
    this.increment();
}
```

The body of that synthesized method will type-check just fine in this case (because it desugars into `this.base.increment()`, more or less), and thus the synthesized method declaration can be used as the actual witness that drives downstream code generation.

Details:

* I added some options to lookup to allow us to explicitly skip member lookup through base interfaces; this should make sure that we don't accidentally satisfy an interface requirement using a member of the same or another interface (since such members are conceptually `abstract`).

* As it originally stood, the semantic checker was allowing `CounterBase.increment()` to satisfy the `increment()` requirement of `ResetableCounter` directly, with the result that we got invalid HLSL/GLSL code as output. In order to avoid this and other bad cases, I made sure that the "exact match" case of requirement satisfaction ignores members that included any "breadcrumbs" in the lookup result item (since the breadcrumbs would all indicate transformations that needed to be applied to `this` to find the right member).

* If we eventually have targets where `this` is passed by pointer/reference in all cases, then all of this work is not needed for the common case of single inheritance, and the base-type method should be usable as a witness directly. I don't see any easy way to handle that special case without producing target-dependent code in the front-end. It might be that we need an IR pass that can detect functions that are trivial "forwarding" functions and replace them with the function they forward to.

* This change includes a test case that should have come along with the original PR that started adding struct inheritance

Caveats:

* The comments in this change talk about things like allowing a method with a default parameter to satisfy a requirement without that parameter. That scenario won't actually work at present because we still have an enormous hack in our logic for checking methods against requirements: we don't actually consider their signatures! I couldn't fold a fix for that issue into this change because there are subtle corner cases around associated types that we need to handle correctly (which were part of the reason why the checking is as hacked as it is)

* This change does *not* try to test or address the case where we want to have a `Derived` type conform to `ISomething` because it inherits from `Base` and `Base : ISomething`. That case has its own details that need to be worked out, but ideally can follow a similar implementation strategy when it comes to re-using methods from `Base` to satisfy requirement on `Derived`.


source/slang/slang-check-decl.cpp
```diff
@@ -1443,6 +1443,325 @@ namespace Slang
...
+        RefPtr<WitnessTable>        witnessTable)
+    {
+        // The situation here is that the context of an inheritance
+        // declaration didn't provide an exact match for a required
+        // method. E.g.:
+        //
+        //      interface ICounter { [mutating] int increment(); }
+        //      struct MyCounter : ICounter
+        //      {
+        //          int increment(int val = 1) { ... }
```
csyonghe, line 1459: Is `[mutation]` needed for implementations?

source/slang/slang-check-decl.cpp
```diff
@@ -1443,6 +1443,325 @@ namespace Slang
...
+        // The body of our synthesized method is going to try to
+        // make a call using the name of the method requirement (e.g.,
+        // the name `increment` in our example at the top of this function).
+        //
+        // The caller already passed in a `LookupResult` that represents
+        // an attempt to look up the given name in the type of `this`,
+        // and we really just need to wrap that result up as an overloaded
+        // expression.
+        //
+        auto synBase = m_astBuilder->create<OverloadedExpr>();
```
csyonghe, line 1597: Having to emit checked AST directly is something that makes complex translation logic difficult to implement. I wonder if we can build some infra to make it easy, e.g.

`newAst = compileAST("return %1(%2);", funcAst, argsAst);`

Given our existing architecture, this kind of flow may not be possible...

source/slang/slang-check-decl.cpp
```diff
@@ -1525,29 +1844,72 @@ namespace Slang
...
+            //
+            getSink()->diagnose(inheritanceDecl, Diagnostics::typeDoesntImplementInterfaceRequirement, type, requiredMemberDeclRef);
+            return false;
+        }
 
         // Iterate over the members and look for one that matches
         // the expected signature for the requirement.
         for (auto member : lookupResult)
         {
+            // To a first approximation, any lookup result that required a "breadcrumb"
```
csyonghe, line 1878: I guess this is another case that `trySynthesizeMethodRequirementWitness` will handle.

source/slang/slang-check-decl.cpp
tfoleyNV: Oh yeah, it should be required there.

One note is that my current test case doesn't cover `static` or `[mutating]` methods, so I can't pretend to have full confidence that they will work (I'd like to expand out the test coverage bit by bit so that I can keep the PRs/fixes small).

source/slang/slang-check-decl.cpp
tfoleyNV: I agree 100% that manually building checked AST is a mess. Your approach of hijacking the parser to get it done could work, but probably requires a lot of infrastructure we don't have today.

One thing that I think can/does help is to frame our semantic checking in terms of "abstract interpretation," with a split between:

* Abstract interpretation functions that take checked syntax as input and produce checked syntax as output, potentially emitting diagnostics along the way (and/or producing an `ErrorExpr` or similar)

* AST visitor functions which orchestrate the post-order traversal of first checking sub-terms and then invoking an appropriate abstract interpretation function to do the main work and produce output syntax.

In the current semantic checking you can see some rudiments of this approach with stuff like `coerce()`, which handles implicit type conversions on already-checked syntax, or `ResolveInvoke()` which handles overload resolution for a call where all the sub-expressions are already checked.

If we wanted to go fully down that kind of road we might find that we need/want a completely separate AST hierarchy for checked vs. unchecked syntax, so that semantic checking produces a new AST rather than mutating an existing one in-place.

Certainly this one use case isn't quite enough to motivate any massive architecture changes just yet.

source/slang/slang-check-decl.cpp
tfoleyNV: The hope is that `trySynthesize` can (eventually) handle the arbitrary general case: if your type supports `this.whateverName(whateverArgs...)` syntactically and produces a value of the right type, it will be able to satisfy the `whateverName` requirement.

The loop there is then only responsible for finding "exact match" cases, where the type already has a declaration that is guaranteed to match the exact signature (down to ABI, I guess) of the operation in the interface.

We could in theory eliminate the "exact match" logic if `trySynthesize...` was generalized to handle all kinds of witnesses, at the cost of potentially introducing a bunch of unnecessary synthesized wrapper declarations.

PR #1404, opened 2020-06-18
Dynamic dispatch for static member functions of associatedtypes.

csyonghe:


source/slang/slang-emit-c-like.cpp
```diff
@@ -237,18 +237,15 @@ List<IRWitnessTableEntry*> CLikeSourceEmitter::getSortedWitnessTableEntries(IRWi
...
         auto reqKey = cast<IRStructKey>(interfaceType->getOperand(i));
-        bool matchingEntryFound = false;
         IRWitnessTableEntry* entry = nullptr;
         if (witnessTableEntryDictionary.TryGetValue(reqKey, entry))
         {
-            if (entry->requirementKey.get() == reqKey)
-            {
-                matchingEntryFound = true;
-                sortedWitnessTableEntries.add(entry);
-                break;
```
csyonghe, line 248: This code is wrong. There is no need for the comparison given the use of dictionary, and should not `break`.

tfoleyNV: I had wondered in the earlier change, but assumed that the logic was being left in for some defensive-programming reason. It is certainly easier to not have it.

PR #1408, opened 2020-06-24
Dynamic dispatch for generic interface requirements and `associatedtype`

csyonghe:
This PR includes several changes to enable dynamic dispatching a generic interface method and calling a method via associated types.

The major changes are:
- Include requirement type info in `IRInterfaceType` insts. Currently `IRInterfaceType` is just a set of `IRStructKey`s that is used to define the order of witnesstable entries. This is insufficient when a requirement entry is a generic function, where additional type info from the requirement itself is required for code generation. Including type info in `IRInterfaceType` also allows us to emit the witness table type without requiring a witness table value.
- Lower `DeclRef<AssocTypeDecl>` into `IRAssociatedType`. This allows an associatedtype that is used as parameter type or return type to lower as `void*`.
- Generate proper IRType for generic functions. Since `IRInterfaceType` now stores a type for each requirement entry, We need a proper type for generic functions that are lowered as `IRGeneric`. Currently there is no type associated with `IRGeneric` insts. This changes makes it so that generic functions will have an `IRGeneric{IRParam genericParam; IRFuncType ft(...); return ft;}` as the type of the resulting `IRGeneric`. However, deduplicating a generic type is not implemented in this PR.
- Fix `finishOuterGenerics` incorrectly returning the inner value if the inner value has already been hoisted out of the generic. `finishOuterGenerics` now takes an explicit `outerGeneric` parameter instead of it trying to find it through the parent chain.
- Add `IRRawPointerType` that represents a typeless pointer. `IRRawPointerType` will lower to `void*` in CPP and CUDA backends.
- Add `IRThisType` and lower all `This` in interface definitions to `IRThisType`. `IRThisType` will be translated into `IRRawPointer` type in `lowerGenerics` pass.
- Add `IRAssociatedType`, all references to an `associatedtype` in a interface definition will lower to a global `IRAssociatedType` object, and will be translated to `IRRawPointer` in `lowerGenerics` pass.
- Add a test case exercising dynamic dispatching a generic static function through an associated type.
- Bug fixes for the test case.


source/slang/slang-emit-c-like.cpp
```diff
@@ -1962,6 +1962,10 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
         are hashed with 'getStringHash' */
         break;
 
+    case kIROp_undefined:
```
csyonghe, line 1965: Uninitialized local variables can get optimized out into an undefined value. We probably need to rethink whether it is a good idea to include such a hack here. 

Specifically, this is for the following case:
```
int GenericCompute<T:IInterface>(T obj, int inVal)
{
    GetterImpl getter;
	return obj.Compute(inVal) + T.Assoc.getBase(getter);
}
```
In this code, `getter` could be come `undefined`.

source/slang/slang-ir-inst-defs.h
```diff
@@ -507,14 +509,12 @@ INST(HighLevelDeclDecoration,               highLevelDecl,          1, 0)
...
-        /// is marked `[polymorphic]`, so that the code gen logic can emit it as a `void*` parameter,
-        /// allowing the function to be used at sites that are agnostic of the actual object type.
-    INST(PolymorphicDecoration, polymorphic, 0, 0)
 
         /// A `[this_ptr]` decoration marks a function parameter that serves as `this` pointer.
-    INST(ThisPointerDecoration, this_ptr, 0, 0)
+        /// `[this_ptr]` decoration is also used to mark an `IRFunc` as a non-static function.
+        /// The argument is an integer value that represents the index of the `this` parameter,
+        /// which is always 0.
+    INST(ThisPointerDecoration, this_ptr, 1, 0)
```
csyonghe, line 517: We could have all `this` typed parameters to have `IRThisType` instead of using this decoration. But I would like to do that in another PR instead.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6470,7 +6547,53 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
         irFunc->moveToEnd();
-        return LoweredValInfo::simple(finishOuterGenerics(subBuilder, irFunc));
+
+        // If this function is defined inside an interface, add a reference to the IRFunc from
+        // the interface's type definition.
+        auto finalVal = finishOuterGenerics(subBuilder, irFunc);
+
+        if (auto genericVal = as<IRGeneric>(finalVal))
+        {
+            auto funcType = lowerFuncType(decl);
```
csyonghe, line 6557: For generic functions I'll have to run the `lowerFuncType` logic all over again. This is unfortunate but I can't come up with a better way.

source/slang/slang-ir-inst-defs.h
tfoleyNV: There is a design question here of how `IRThisType` would be used in the encoding. I can see two big options:

1. There is a single global `IRThisType` instance, and all interfaces just use it as a placeholder in their requirement type signatures for wherever the `this` paramter, or another parameter of type `This` goes.

2. Each `IRInterfaceType` `I` has an associated `ThisType<I>` that is specific to it. This allows us to know the interface that each `This`-typed parameter was associated with, in case that matters.

Option (2) potentially creates a circular-reference situation (the interface includes operands that reference the `ThisType` while the type references the interface). That circularity would be easier to navigate in the case where we change the encoding of interfaces so that they are a parent instruction and the entries are child instructions (similar to how witness tables are encoded at the moment).

source/slang/slang-ir-insts.h
```diff
@@ -410,9 +408,13 @@ struct IRLookupWitnessMethod : IRInst
 {
     IRUse witnessTable;
     IRUse requirementKey;
+    IRUse interfaceType;
 
     IRInst* getWitnessTable() { return witnessTable.get(); }
     IRInst* getRequirementKey() { return requirementKey.get(); }
+    IRInst* getInterfaceType() { return interfaceType.get(); }
+
+    IR_LEAF_ISA(lookup_interface_method)
```
tfoleyNV, line 417: It really shouldn't be required for a `lookup_interface_method` to have the interface type as an explicit operand.

A better solution in the long run would be to have an `IRWitnessTableType` that takes the interface being conformed to as an operand (e.g., `WitnessTable<IFoo>`). In that case, the `lookup_interface_method` instruction could have an invariant that the `witnessTable` operand needs to have an `IRWitnessTableType`, and in that case it could produce the interface type by looking at the operand of the witness table's type.

source/slang/slang-ir-link.cpp
```diff
@@ -89,6 +89,25 @@ struct IRSpecContextBase
...
+{
+    switch (value->op)
+    {
+    case kIROp_StructKey:
+        return GlobalValueClass::StructKey;
+    default:
+        return GlobalValueClass::Other;
+    }
+}
+
```
tfoleyNV, line 110: This seems like a reasonable workaround for now, but the question I continue to have is: why does anything other than the key for a requirement need a mangled name and/or linkage?

It seems like the fundamental problem here stems from trying to emit extra "stuff" from an interface requirement, and then giving that "stuff" a mangled name (that happens to conflict with the key, which was the thing that was original there and using the mangled name).

I don't see any evidence that the `Other` case here is being used for things that ever need linkage.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -45,6 +46,20 @@ namespace Slang
             workListSet.Add(inst);
         }
 
+        bool isPolymorphicType(IRInst* typeInst)
+        {
+            if (as<IRParam>(typeInst) && as<IRTypeType>(typeInst->getFullType()))
+                return true;
+            switch (typeInst->op)
+            {
+            case kIROp_AssociatedType:
```
tfoleyNV, line 55: It seems like `IRAssociatedType` is the case causing problems for linking, but it doesn't seem like any code outside the module that defined an interface needs to refer to this representation of an associated type; it exists only to allow us to put signatures on interface requirements.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -64,6 +79,7 @@ namespace Slang
             builder.sharedBuilder = &sharedBuilderStorage;
             builder.setInsertBefore(genericParent);
             auto loweredFunc = cloneInstAndOperands(&cloneEnv, &builder, func);
+            loweredFunc->setFullType(lowerGenericFuncType(&builder, cast<IRGeneric>(genericParent->typeUse.get())));
```
tfoleyNV, line 82: `->typeUse.get()` should be the same as `->getFullType()`, right?

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -91,32 +107,160 @@ namespace Slang
...
+                if (isPolymorphicType(paramType))
+                {
+                    newOperands.add(builder->getPtrType(builder->getVoidType()));
+                    translated = true;
+                }
+                else if (paramType->op == kIROp_Specialize)
+                {
+                    // TODO: handle static specialized type here.
+                    // For now treat all specialized types as dynamic.
+                    // In the future, we need to turn things like Array<IDynamic> into Array<void*>.
```
tfoleyNV, line 156: As discussed briefly on another channel, there is a question of whether a parameter of type `Array<T>` should translate to an `Array<void*>` or to a single `void*`. There are arguments for and against each option.

source/slang/slang-ir.cpp
```diff
@@ -2821,6 +2837,16 @@ namespace Slang
         return structType;
     }
 
+    IRAssociatedType* IRBuilder::createAssociatedType()
```
tfoleyNV, line 2840: Given that `IRAssociatedType` has no operands, it seems like we face the same decision point as with `IRThisType`:

1. We could have a single global instance of `IRAssociatedType`, and all references to associated types in the signature of an interface requirement will refer to that one instance.

2. We could have a distinct instance of `IRAssociatedType` for each interface and each associated type requirement on that interface. In that case, we could imagine `IRAssociatedType` having operands that name the interface and the requirement key.

For option (2), it might actually work really well if an interface type was a parent isntruction, and then an `IRAssociatedType` was a child of that interface, so that we could be careful that references to the associated type in this fashion are properly restricted to only occur in the scope of the interface's requirement signatures.

Going even deeper, it is possible that an `IRAssociatedType` could actually just *be* the `IRInterfaceRequirementEntry` that represents the associated type!

source/slang/slang-ir.h
```diff
@@ -458,7 +458,7 @@ struct IRInst
 
     void setOperand(UInt index, IRInst* value)
     {
-        getOperands()[index].set(value);
+        getOperands()[index].init(this, value);
```
tfoleyNV, line 461: This change seems like something we need to be very careful about.

The current `setOperand()` operates under the assumption that all operands of the instruction have already been initialized (even if it was to a `null` operand value), which means that calling `init(this, ...)` here is working around some piece of code that wasn't following the rules.

I'd rather ass an assert in `set` that the user for the `IRUse` has already been set, and then catch whatever code was violating the rules.

source/slang/slang-ir.h
```diff
@@ -1190,6 +1190,25 @@ struct IRStructType : IRType
...
+    IR_LEAF_ISA(AssociatedType)
+};
+
+struct IRInterfaceRequirementEntry : IRInst
+{
+    // The AST-level requirement
+    IRUse requirementKey;
+
+    // The IR-level value that represents the declaration of the requirement
+    IRUse requirementVal;
```
tfoleyNV, line 1204: It isn't any kind of documented policy, but I've largely moved away from bothering to declare the `IRUse` members in derived `IRInst` types (even if that makes their memory layout unclear to somebody reading the code).

We probably need to establish an official policy on this and then make all the `IRInst`-derived types follow that policy.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -986,6 +998,8 @@ IRStructKey* getInterfaceRequirementKey(
     IRGenContext*   context,
     Decl*           requirementDecl)
 {
+    if (auto genericDecl = as<GenericDecl>(requirementDecl))
+        return getInterfaceRequirementKey(context, genericDecl->inner);
```
tfoleyNV, line 1002: This doesn't seem right. Why did you need to make this edit?

For a generic interface requirement, the `GenericDecl` itself is the requirement, not whatever its inner declaration is, right?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5275,34 +5291,37 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+        auto subBuilder = nestedContext.getBuilder();
+        auto subContext = nestedContext.getContext();
+        List<IRInterfaceRequirementEntry*> requirementEntries;
+
         for (auto requirementDecl : decl->members)
         {
-            requirementKeys.add(getInterfaceRequirementKey(requirementDecl));
-
+            auto key = getInterfaceRequirementKey(requirementDecl);
+            auto entry = subBuilder->createInterfaceRequirementEntry(key, nullptr);
```
tfoleyNV, line 5302: This seems to be part of the problem, insofar as we are not creating the lowered type of the requirement at this point in the code, despite this being an obvious/appropriate time to do so. I'm confused about when this stuff gets wired up if it doesn't happen here.

It seems like all this code needs/wants to do is to lower the type of `requirementDecl` in an appropriately-construct `IRGenContext` - one that will map the AST `ThisType` -> `IRThisType`, and references to any associated types to `IRAssociatedType`. It should be possible to construct an appropriate `IRGenEnv`/`IRGenContext` before entering this loop.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5387,65 +5406,78 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+        auto outerGeneric = emitOuterGenerics(subContext, decl, decl);
 
-        IRStructType* irStruct = subBuilder->createStructType();
-        addNameHint(context, irStruct, decl);
-        addLinkageDecoration(context, irStruct, decl);
+        IRInst* resultType = nullptr;
+        if (as<AssocTypeDecl>(decl))
+        {
+            resultType = subBuilder->createAssociatedType();
+        }
```
tfoleyNV, line 5415: Yeah, this bit seems broken to me. We shouldn't be emitting a global `IRInst` to represent each associated type and giving it a mangled name (and thus linkage), when the only context that needs to refer to it will be the `IRInterfaceType` that declares it. These `IRAssociatedType` insts really seem like they should be scoped as narrowly as possible, and it seems like the lowering for the interface type could easily just create all of them as a pre-process before it dives into emitting the requirement entries.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5387,65 +5406,78 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
-        subBuilder->setInsertInto(irStruct);
+        addNameHint(context, resultType, decl);
+        addLinkageDecoration(context, resultType, decl);
 
-        // A `struct` that inherits from another `struct` must start
-        // with a member for the direct base type.
-        //
-        for( auto inheritanceDecl : decl->getMembersOfType<InheritanceDecl>() )
+        if (resultType->op == kIROp_StructType)
         {
```
tfoleyNV, line 5425: The sheer amount of code that is specific to the non-`AssociatedType` case here argues that even *if* we keep going down this road of having an associated type emit as a top-level entity with linkage, we probably want to break it out as a separate `visitAssociatedTypeDecl` method instead of trying to have `visitAggTypeDecl` handle two very different cases.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5995,29 +6028,17 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
     }
 
-    LoweredValInfo lowerFuncDecl(FunctionDeclBase* decl)
+    void _lowerFuncResultAndParameterTypes(
+        ParameterLists& parameterLists,
+        List<IRType*>& paramTypes,
+        IRType*& irResultType,
+        IRBuilder* subBuilder,
+        IRGenContext* subContext,
+        FunctionDeclBase* decl)
```
tfoleyNV, line 6037: Interesting. My work on `property` declaration also ended up needing to refactor this operation to make some of the pieces be more reusable.

I have a feeling there is a painful merge ahead for me, but it seems like there is a good sign that the refactoring is something the codebase needed (since both of us seem to have made it).

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6107,11 +6120,83 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
 
-        auto irFuncType = subBuilder->getFuncType(
+        auto irFuncType = builder->getFuncType(
             paramTypes.getCount(),
             paramTypes.getBuffer(),
             irResultType);
+
+        if (parameterLists.params.getCount() && parameterLists.params[0].isThisParam)
+            builder->addThisPointerDecoration(irFuncType, 0);
+        return irFuncType;
```
tfoleyNV, line 6148: There is a potential problem with this logic. What if I have:

```
struct S {
    int doA(float x) { ... }
}
int doB(S s, float y) { ... }
```

At the IR level, both `doA` and `doB` would currently lower to use the exact same `IRFuncType`: `(S, y) -> int`.

So what happens if you decorate that single `IRFuncType` with your `IRThisPointerDecoration`? It seems like it will incorrectly flag the `s` parameter of `doB` as a `this` pointer, which isn't correct?

Maybe that isn't actually a problem (I don't fully understand where/how the decoration impacts codegen), but it doesn't seem semantically correct.

tfoleyNV: To follow up on this a bit, we *do* have the starting point of a solution for cases where you want to attach "extra" data to an instruction (like you would do with decorations), but the instruction in question is one that can get hoisted/deduplicated and you don't want the versions with and without your decoration to get mixed up.

There is an `IRAttr` type for "attributes," which are similar to `IRDeocration`s in that they store extra/optional data, but are very different than decorations in that decorations are *children* of the thing they decoration, while attributes are *operands* of the thing they decorate.

You can think of `IRAttr`s as extra operands that can follow the "ordinary" operands of certain instructions (not all instructions are designed to work with attributes today). It is possible to query for the attributes on an instruction, and to search for attributes of a certain type (like you can with deocrations). When creating an instruction of type kind that gets deduplicated, you need to put together both the ordinary operands and all attributes before you construct the instance (you can't go back and add attribute slater), so that the deduplication logic can take the attributes into account (two instructions with the same "ordinary" operands, but different attributes are distinct from a deduplication standpoint).

I don't actually think attributes are a perfect fit for what you are doing in this case (I'm not sure what an ideal design would be) but I wanted to note that they are there in case you ever need to apply them to some challenge.

(E.g., if/when we add a notion of different calling conventions to Slang, we'd have to make a choice of whether to make calling convention a required property of an `IRFuncType` (so that it is always one of the operands), or to make an `IRAttr` that can associate an `IRFuncType` with a non-default calling convention. No matter what, we'd need to ensure that function types with different calling conventions never get confused for one another and/or deduplicated into a single type.)

source/slang/slang-lower-to-ir.cpp
tfoleyNV: It seems like a general-purpose fix might be to make `finishOuterGenerics` handle the problem of giving a type to each outer `IRGeneric` using the following approach:

* Take the type of the inner instruction (the thing you are going to `return` from the generic
* Construct a new `IRGeneric` that will represent the type of this generic
* Clone all the parameters of this `IRGeneric` into the new one, and build up an `IRCloneEnv` that maps the original parameters over to the new one
* Now clone the type of the inner instruction in that newly-construct `IRCloneEnv` and make it the return value of our new `IRGeneric`
* Now we can make the type of the original `IRGeneric` be the new `IRGeneric`
* TODO: figure out what type to give to the new `IRGeneric` so that this doesn't result in infinite regress.

tfoleyNV: Oh! One thing I just realized is an issue with using an `IRGeneric` to represent a `forall` type to be used as the type of another generic is that we can't deduplicate such types using their operands, so we'd have no way to easily and robustly answer questions like "do these two generic functions have the same (`forall`-based) type?"

That issue might be reason enough to start thinking about how we can represent `forall` types more directly. In the worst case I believe we can do something using [De Bruijin indices](https://en.wikipedia.org/wiki/De_Bruijn_index).

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6470,7 +6548,52 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+        auto irInterfaceType = cast<IRInterfaceType>(loweredVal->val);
+        auto key = getInterfaceRequirementKey(decl);
+        for (UInt i = 0; i < irInterfaceType->getOperandCount(); i++)
+        {
+            auto operand = cast<IRInterfaceRequirementEntry>(irInterfaceType->getOperand(i));
+            if (operand->getOperand(0) == key)
+            {
+                operand->setOperand(1, irFuncType);
+                return;
+            }
```
tfoleyNV, line 6593: This answers my earlier question about how things are getting "wired up."

The first note is that this seems like it can/should use `operand->getOperands()[1]->init(operand, irFuncType)` instead of going through `setOperand` and then having to tweak that to remove preconditions.

The second note is that it seems like this should be written in a "gather" rather than a "scatter" fashion - the code that builds up the requirement entries should be able to pull these types off of the requirements themselves rather than have the lowering of the requirements push things back into the interface.

source/slang/slang-emit-cpp.cpp
```diff
@@ -390,12 +390,27 @@ static UnownedStringSlice _getResourceTypePrefix(IROp op)
     }
 }
 
+static bool isVoidPtrType(IRType* type)
+{
+    auto ptrType = as<IRPtrType>(type);
+    if (!ptrType) return false;
+    return ptrType->getValueType()->op == kIROp_VoidType;
+}
```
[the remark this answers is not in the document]
csyonghe: Done.

source/slang/slang-ir-link.cpp
csyonghe: Done. Reverted.

source/slang/slang-ir.cpp
csyonghe: Done.

source/slang/slang-lower-to-ir.cpp
csyonghe: Fixed.

source/slang/slang-ir.h
csyonghe: Removed.

source/slang/slang-lower-to-ir.cpp
csyonghe: Cleaned up.

source/slang/slang-ir-insts.h
csyonghe: Removed, I just need to handle the `TransitiveSubtypeWitness` lowering case and ensure a correct `IRWitnessTable` type is emitted there so I don't get nullptr for the type of the resulting `lookup_interface_method` inst.

source/slang/slang-emit-c-like.cpp
```diff
@@ -236,9 +236,9 @@ List<IRWitnessTableEntry*> CLikeSourceEmitter::getSortedWitnessTableEntries(IRWi
     // Get a sorted list of entries using RequirementKeys defined in `interfaceType`.
     for (UInt i = 0; i < interfaceType->getOperandCount(); i++)
     {
-        auto reqKey = cast<IRStructKey>(interfaceType->getOperand(i));
+        auto reqEntry = cast<IRInterfaceRequirementEntry>(interfaceType->getOperand(i));
```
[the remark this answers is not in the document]
csyonghe: Done.

source/slang/slang-ir.h
csyonghe: Done.

csyonghe: I added forced initialization of `IRUse`s in `createInst`, no matter whether the passed in `type` or `args` is null or not.

source/slang/slang-lower-to-ir.cpp
csyonghe: It seems that `getInterfaceRequirementKey` is called on the generic decl when using an interface type (e.g. when generating the `lookup_interface_method` inst, and called on the inner function decl when lowering the interface type, thus two duplicate keys with the same name will be created and the lookup will fail.

tfoleyNV: That's a bit unfortunate, and it would be nice to figure out if we can make the cases more consistent without a workaround like this, but if the workaround prevents a bug for now then I'm fine with it. Maybe it is worth adding a comment to this conditional to be clear about why you had to add it?

source/slang/slang-lower-to-ir.cpp
csyonghe: Interface lowering is now all done in `visitInterfaceDecl` without this injecting logic.

source/slang/slang-ir-inst-defs.h
csyonghe: Done. Option 1 is implemented.

source/slang/slang-lower-to-ir.cpp
csyonghe: This stems from that `WitnessTable` are constructed from the inner decls in `slang-check-decl.cpp`. Making it consistent would require changing `doesGenericSignatureMatchRequirement`, currently `doesGenericSignatureMatchRequirement` just calls `doesMemberSignatureMatchRequirement(genDecl->inner)` so the AST witness table will always contain the inner decl as the requirement key. Not sure what would be the cleanest way to fix this.

csyonghe: I added a comment for this issue.

source/slang/slang-lower-to-ir.cpp
csyonghe: Will address in follow-up PRs.

csyonghe:
@tfoleyNV This PR is ready for review again. I have created two issues for follow-up work.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5275,34 +5300,95 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+                    {
+                    case kIROp_Func:
+                    case kIROp_Generic:
+                        requirementVal->removeAndDeallocate();
+                        break;
+                    default:
+                        break;
+                    }
+                }   
+            }
```
tfoleyNV, line 5369: This seems fine for now, but it would be cleaner to not even emit the `requirementDecl` in the first place, and instead just emit the *type* of `requirementDecl`.

The biggest problem in doing that right now is that the front-end doesn't store `Type` information on every `Decl`, and instead it has this complicated logic in `getTypeForDeclRef` that is used to compute the type based on the class of declaration.

We could either introduce similar logic in a subroutine here (something like `lowerDeclType()`) that looks for the different cases of declaration (or uses a visitor) and emits the appropriate type, *or* we could try to be better about having all front-end `Decl`s have a type directly stored on them, if it would make things easier downstream.

csyonghe: I thought about adding `lowerDeclType` but decided that it is adding too much complexity. It's likely that may turn out to be the better choice down the road, but for now I think we can get by with this less efficient hack.

PR #1409, opened 2020-06-24
AST Serialize Reading 

jsmall-nvidia:
* Handle alignment issues during writing
* Ability to write serialized data to a Stream
* Fast pass implementation of ASTSerialReader
  * Load directly from in memory serial data


jsmall-nvidia:
Yes needs more documentation/description for sure will add.

PR #1410, opened 2020-06-24
Initial work on property declarations

tfoleyNV:
Introduction
============

The main feature added here is support for `property` declarations, which provide a nicer experience for working with getter/setter pairs.

If existing code had something like this:

```hlsl
struct Sphere
{
	float4 centerAndRadius; // xyz: center, w: radius

	float3 getCenter() { return centerAndRadius.xyz; }
	void setCenter(float3 newValue) { centerAndRadius.xyz = newValue; }

	// similarly for radius...
}
void someFunc(in out Sphere s)
{
	float3 c = s.getCenter();
	s.setCenter(c + offset);
}
```

It can be expressed instead using a `property` declaration for `center`:

```hlsl
struct Sphere
{
	float4 centerAndRadius; // xyz: center, w: radius

	property center : float3
	{
		get { return centerAndRadius.xyz; }
		set(newValue) { centerAndRadius.xyz = newValue; }
	}

	// similarly for radius...
}
void someFunc(in out Sphere s)
{
	float3 c = s.center;
	s.center = c + offset;
}
```

The benefits at the declaration site aren't that signficiant (e.g., in the example above we actually have slightly more lines of code), but the improvement in code clarity for users is significant.

Having `property` declarations should also make it easier to migrate from a simple field to a property with more complex logic without having to first abstract the use-site code using a getter and setter.

An important future benefit of `property` syntax will be if we allow `interface`s to include `property` requirements, and then also allow those requirements to be satisfied by ordinary fields in concrete types.

Subscripts
----------

The Slang compiler already has limited (stdlib-use-only) support for `__subscript` declarations, which are conceptually similar to `operator[]` from the C++ world, but are expressed in a way that is more in line with `subscript` declarations in Swift. A `SubscriptDecl` in the AST contains zero or more `AccessorDecl`s, which correspond to the `get` and `set` clauses inside the original declaration (there is also a case for a `__ref` accessor, to handle the case where access needs to return a single address/reference that can be atomically mutated).

A major goal of the implementation here is to re-use as much of the infrastructure as possible for `__subscript` declarations when implementing `property` declarations.

Nonmutating Setters
-------------------

One additional thing added in this change is the ability to mark a `set` accessor on either a subscript or a property as `[nonmutating]`, and indeed all of the existing `set` accessors declared in the stdlib have been marked this way.

The need for this modifier is a bit subtle. If we think about a typical subscript or property:

```hlsl
struct MyThing
{
    int f;
    property p : int
    {
        get { return f; }
        set(newValue) { f = newValue; }
    }
}
```

it is clear we want the `set` accessor to translate to output HLSL as something like:

```
void MyThing_p_set(inout MyThing this, int newValue)
{
    this.f = newValue;
}
```

Note how the implicit `this` parameter is `inout` even though we didn't mark anything as `[mutating]`. This is the obvious thing a user would expect us to generate given a property declaration.

Now consider a case like the following:

```hlsl
struct MyThing
{
    RWStructuredBuffer<int> storage;

    property p : int
    {
        get { return storage[0]; }
        set(newValue) { storage[0] = newValue; }
    }
}
```

This new declaration doesn't require (or want) an `inout` `this` parameter at all:

```
void MyThing_p_set(MyThing this, int newValue)
{
    this.storage[0] = newValue;
}
```

In fact, given the limitations in the current Slang compiler around functions that return resource types (or use them for `inout` parameters), we can only support a `set` operation like this if we can ensure that the `this` parameter is considered to be `in` instead of `inout`. This is exactly the behavior we allow users to opt into with a `[nonmutating] set` declaration.

All of the subscript operations in the stdlib today have `set` accessors that don't actually change the value of `this` that they act on (e.g., storing into a `RWStructuredBuffer` using its `operator[]` doesn't change the value of the `RWStructuredBuffer` variable -- just its contents).

We'd gotten away without this detail so far just because `set` accessors were only being declared in the stdlib and they were all implicitly `[nonmutating]` anyway, so it never surfaced as an issue that the code we generated assumed a setter wouldn't change `this`.

Implementation
==============

Parser and AST
--------------

Adding a new AST node for `PropertyDecl` and the relevant parsing logic was mostly straightforward. The biggest change was allowing a `set` declaration to introduce an explicit name for the parameter that represents the new value to be set.

This change also adds a `[nonmutating]` attribute as a dual to `[mutating]`, for reasons I will get to later.

Semantic Checking
-----------------

The `getTypeForDeclRef` logic was updated to allow references to `property` declarations.

Some of the semantic checking work for subscripts was pulled out into re-usable subroutines to allow it to be shared by `__subscript` and `property` declarations.

The checking of accessor declarations, which sets their result type based on the type of the outer `__subscript` was changed to also handle an outer `property`.

Some special-case logic was added for checking of `set` declarations to make sure that their parameter is given the expected type.

Some logic around deciding whether or not `this` is mutable had to be updated to correctly note that `this` should be mutable by default in a `set` accessor, with an explicit `[nonmutating]` modifier required to opt out of this default. (This is the inverse of how a typical method or `get` accessor works).

IR Lowering
-----------

The good news is that after IR lowering, access to properties turns into ordinary function calls (equivalent to what hand-written getters and setters would produce), so that subsequent compiler steps (including all the target-specific emit logic) doesn't have to care about the new feature.

The bad news is that adding `property` declarations has revealed a few holes in how IR lowering was handling `__subscript` declarations and their accessors, so that it didn't trivially work for the new case as-is.

The IR lowering pass already has the `LoweredValInfo` type that abstractly represents a value that resulted from lowering some AST code to the IR. One of the cases of `LoweredValInfo` was `BoundSubscript` that represented an expression of the form `baseVal[someIndex]` where the AST-level expression referenced a `__subscript` declaration. The key feature of `BoundSubscript` is that it avoided deciding whether to invoke the getter, the setter, or both "too early" and instead tried to only invoke the expected/required operations on-demand.

This change generalizes `BoundSubscript` to handle `property` references as well, so it changes to `BoundStorage`. Making the type handle user-defined property declarations required fixing a bunch of issues:

* When building up argument lists in the IR, we need to know whether an argument corresponds to an `in` or an `out`/`inout` parameter, to decide whether to pass the value directly or a pointer to the value. Some of the logic in the lowering pass had been playing fast and loose with this, so this change tries to make sure that whenever we care computing a list of `IRInst*` that represent the arguments to a call we have the information about the corresponding parameter.

* Similarly, when emitting a call to an accessor in the IR, the information about the expected type of the callee was missing/unavailable, and the code was incorrectly building up the expected type of the callee based on the types of the arguments at the call site. The logic has been changed so that we can extract the expected signature of an accessor (how it will be translated to the IR) using the same logic that is used to produce the actual `IRFunc` for the accessor (so hopefully both will always agree).

* Dealing with `in` vs. `inout` differences around parameters means also dealing with the "fixup" code that is used to assign from the temporary used to pass an `inout` argument back into the actual l-value expression that was used. That logic has all been hoisted out of the expression visitor(s) and into the global scope.

Future Work
===========

The entire approach to handling l-values in the IR lowering pass is broken, and it is in need a of a complete rewrite based on new first-principles design goals. While something like `LoweredValInfo` is decent for abstracting over the easy cases of r-values, addresses, and a few complicated l-value cases like swizzling, it just doesn't scale to highly abstract l-values like we get from `__subcript` and `property` declarations, nor other corner cases of l-values that we need to handle (e.g., passing an `int` to an `inout float` parameter is allowed in HLSL, and performs conversions in both directions!).

It Should be Easy (TM) to extend the logic that tries to synthesize an interface conformance witness method when there isn't an exact match to also support synthesizing a property declaration (plus its accessors) to witness a required property when the type has a field of the same name/type.


source/slang/slang-check-decl.cpp
```diff
@@ -3652,19 +3682,19 @@ namespace Slang
...
+    void SemanticsDeclHeaderVisitor::visitAbstractStorageDeclCommon(ContainerDecl* decl)
     {
-        decl->returnType = CheckUsableType(decl->returnType);
-
-        // If we have a subscript declaration with no accessor declarations,
+        // If we have a subscript or property declaration with no accessor declarations,
         // then we should create a single `GetterDecl` to represent
         // the implicit meaning of their declaration, so:
         //
         //      subscript(uint index) -> T;
```
jsmall-nvidia, line 3691: A niggle: The use of the colon is presumably to try and match a 'modern var' variable : type? It seems a little inconsistent with the use of -> on subscript, in light of the getter/setting being 'functions'. That some languages use : instead of -> also adds to the worry.

source/slang/slang-check-decl.cpp
```diff
@@ -3702,6 +3749,42 @@ namespace Slang
...
+        visitAccessorDecl(decl);
+
+        auto params = decl->getParameters();
+        ParamDecl* param = nullptr;
+        if( params.getCount() >= 1 )
+        {
+            param = params.getFirst();
+            if( params.getCount() > 1 )
+            {
+                // TODO: diagnose it!
```
jsmall-nvidia, line 3764: Is it okay to just continue here?

source/slang/slang-parser.cpp
```diff
@@ -2665,6 +2669,20 @@ namespace Slang
...
+        parser->PushScope(decl);
+
+        // A `set` declaration should support declaring an explicit
+        // name for the parameter representing the new value.
+        //
+        // We handle this by supporting an arbitrary parameter list
+        // on any accessor, and then assume that semantic checking
+        // will diagnose any cases that aren't allowed.
+        //
+        if(parser->tokenReader.peekTokenType() == TokenType::LParent)
```
jsmall-nvidia, line 2681: Should it only support the modern param list style?

source/slang/slang-check-decl.cpp
csyonghe: I think we should avoid mixing different syntax styles. I agree that we should also support the same C# property syntax `property Type Name`. When you are actively developing using C++/C#, switching to another syntax is a mental overhead and reduces producitivity.

tfoleyNV: There's a few distinct points being discussed here, so I want to tackle each separately:

### On using `:` vs `->`

In the world of "modern" languages,  the form `x : Y` is usually used to indicate that the type of `x` is `Y`. That is consistent with how it is being used here. `property x : Y` is saying that the type of property `x` is `Y`.

The `->` operator is used to construct function types, so that `Y -> Z` is the type of a function from `Y` to `Z`. If we had a variable of function type, we might say `f : Y -> Z`, and that all Just Works.

Many languages (including Swift, where we get much of our inspiration for new syntax) use `->` as the way to introduce the return type of a function:  e.g.,`func f(x: Y) -> Z`. In this case the use of `->` is to reinforce that the type of a function declared this way matches the type of a function-typed variable. E.g.:

```
// These two declarations both have function types, and the types those functions return is `Z`.
func f(x : Y) -> Z { ... }
let g : Y -> Z = ...
```

Using a colon would then be misleading:

```
func f(x : Y) : Z { ... }
let g : Y -> Z = ...
```

This second option starts to look like we are saying `f` has type `Z`, when that isn't the case. You could argue that it should be read as saying that `f(y)` has type `Z` (that is, once we apply the function the type is `Z`), but at the very least we should acknowledge that each choice has arguments in its favor.

### On using `name : Type` vs. `Type name`

The mental overhead to switch styles is a valid point. We obviously need to support the legacy C-like syntax in Slang for HLSL compatibility, so if we were only going to support one syntax that would have to be the one. There is an argument to be made that having two different declaration syntaxes co-exist in Slang is a Bad Thing.

All that said, C/C++-style declaration syntax is a train wreck, and there are things you can't do within its constraints. Even C++ ran into this problem and had to introduce a trailing return-type syntax as a band-aid.

My opinion has been that Slang should provide a slow and gentle ramp from current HLSL syntax/style toward something cleaner and more modern. Whenever a 100% new piece of syntax is being added and exposed to users, I've been trying to expose it in the way that we'd like it to look and not in the way that most closely mimics the legacy of HLSL.

It seems like we should probably have a discussion about the goals here, and arrive at a consensus about the design principles for how to evolve Slang forward. So far I've added things like `let`, `var`, and `func` declarations that all use this "modern" declaration style. If we want to change things now we will have to revisit those designs and either remove them or find a way to align them with whatever design we pick.

source/slang/slang-check-decl.cpp
tfoleyNV: I probably need to actually implement that `TODO` comment and diagnose the problem.

source/slang/slang-parser.cpp
tfoleyNV: This seems strongly related to the other comment, insofar as we need to make some decisions about Slang syntax going forward.

In general, for a given syntactic form we need to know up front whether it will use "modern" or "legacy" style. Right now the rules are that if you start a declaration with a keyword that indicates "modern" form, then you are locked into modern form for the rest of that declaration (this applies to `let`, `var`, and `func` today, and would apply to `property` and `set` with these changes)*.

For the particular case this is enabling, the detail is almost moot. The only thing we want is for somebody to be able to write:

```
set(newValue) { _someBackingField = newValue; }
```

We only really want them to specify a name for the parameter, not a type. It turns out that if you use a bare identifier in a "legacy" parameter list, then we have to interpret that as an unnamed parameter, like:

```
void foo(int); // this is a function with a single unnamed parameter of type `int`
```

whereas in a "modern" parameter list, a bare identifier is a parameter name that you forgot to give a type to:

```
func bar(x); // this is a function with a single named parameter `x`, which doesn't have a type (and thus will be an error)
```

For this specific case, interpreting things as the "modern" case gives us the right behavior for a single identifier, while still also letting the parser consume extra parameters and/or an explicit type in case we ever want to expand the scope of what is allowed.

If we did decide to use "legacy" parameter list syntax here then it would be required for the user to give the "new value" parameter an explicit type, and then we would definitely need to check that the type they gave it matches the type of the outer property.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6087,6 +6408,7 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         return as<IRStringLit>(builder->getStringValue(stringLitExpr->value.getUnownedSlice()));
     }
 
+#if 0
```
csyonghe, line 6411: Should we just remove them?

tfoleyNV: Yes. I failed to clean these up after the merge. I will plan to do a bit more cleanup before calling this done.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -1827,28 +1882,186 @@ LoweredValInfo createVar(
...
+            // Now we can pass the address of the temporary variable
+            // to the callee as the actual argument for the `in out`
+            SLANG_ASSERT(tempVar.flavor == LoweredValInfo::Flavor::Ptr);
+            IRInst* tempPtr = getAddress(context, tempVar, loc);
+            addInArg(context, ioArgs, LoweredValInfo::simple(tempPtr));
+
+            // Finally, after the call we will need
+            // to copy in the other direction: from our
+            // temp back to the original l-value.
+            OutArgumentFixup fixup;
```
csyonghe, line 2010: For properties this can lead to potential performance issues if calling a function with an `inout` parameter that does not actually writes to it. Given the current lowering logic, it is difficult to optimize out the call to `set` in future passes even if the function is inlined, because at that point it is too late to know that the `set` call is related with the `out` parameter.

tfoleyNV: I meant to follow up on this. The code here is old code that just got moved around in this change, so it wasn't in scope to change the details as part of this PR.

That said, I'm not sure I follow the details of your comments, so I don't get which option you are arguing for or against here. Right now at call sites to things with `inout` parameters we take input like:

```
foo(someVal.someField);
```

and translate it to something like (in C++ pseduo-code):

```
auto tmp = someVal.someField; copy from storage to temporary
foo(&tmp); // pass temporary by reference
someVal.someField = tmp; // copy from temporary back to storage
```

The comments associated with the code mention that *if* we can declare by fiat that the language doesn't support a function mutating a storage location that aliases with an `inout` argument (which I think we can), then we can translate this to the following:

```
foo(&someVal.someField); // pass storage directly by reference
```

This latter translation has the benefit of eliminating some copies, but it potentially inhibits breaking up lifetimes in our SSA pass because our analysis isn't smart enough to know that `foo()` won't (and cannot) allow the address that was passed into "escape." Keeping the temporary around except when true by-reference passing is required helps us make sure SSA can do its job, but leaves us with extra temporaries.

Ideally we could eventually have a virtual "register allocation" pass in the IR that collapses independent local variables together when their lifetimes don't overlap (if doing so would eliminate moves between them), and similarly decides to allocate SSA temporaries into those same locations for the purposes of code emit when doing so would eliminate copies.

Anyway, that is all further off...

In the specific case of `property` declarations, the same idiom currently translates to something even more silly:

```
auto tmpProp = someProperty_get(someVal);
foo(&tmp);
auto tmpVal = someVal;
someProperty_set(&tmpVal, tmpProp);
someVal = tmpVal;
```

The reason for the extra verbiage is that the setter for `someProperty` is also going to have an `inout` parameter (for `this`) and so the logic for introducing these copies will apply to that too.

That is clearly a bit of a mess, and we will be reliant on downstream optimization passes (or ones we introduce to Slang) to clean up the resulting logic into something more idiomatic.

Even if we eliminate the extra copies due to our handling of `inout`, I think your point was that we still end up with:

```
auto tmpProp = someProperty_get(someVal);
foo(&tmpProp);
somePropety_set(&someVal, tmpProp);
```

That is, we cannot eliminate the temporary variable for `someProp` because the property itself might not have a fixed address that can be passed along to `foo`.

Your point seems to be that we cannot eliminate the call to `someProperty_set` even if we discover that `foo()` never writes to that `inout` parameter, because the compiler wouldn't have a way to know that assigning back `tmpProp` would do nothing (and indeed there is no reason for the compiler to require/assume that a setter can't have observable side effects).

AFAICT there are two basic ways to deal with that specific concern:

1. We can implement true call-by-reference and have by-reference accessors. It turns out we actually have this in Slang, although it is only supported for Slang stdlib code. Operations can have `ref` parameters instead of just `inout`, such that by-reference parameter passing is guaranteed/required, and subscripts/properties can have `ref` accessors that simply return a reference to storage instead of needing paired setters/getters. Exposing these as first-class constructs for users would require some non-trivial work, but could probably be done.

2. We could implement "call-by-name" parameter passing, so that instead of receiving a pointer (as in the current Slang IR) the IR version of a function with an `inout` parameter receives a kind of object; something like an `IReadWriteCell<T>` that it can use to read and write the argument, invoking the getter/setter logic on each such access (so every line of code in `foo()` that writes to its `inout` parameter would invoke the full setter logic).

Implementing call-by-name would be a bit of a mess in the implementation, and would seemingly move our design further away from HLSL (which is all based on copy-in/copy-out). I honestly don't think of call-by-name as a realistic option.

There is a remaining option in this space that addresses some of the overhead of getter/setter pairs, but that we aren't ready to try and tackle. As usual, this is an idea coming from the Swift world, where they have `modify` accessors a not-yet-final feature to go along with `get` and `set`. The idea of a `modify` accessors is that it is a coroutine, and the value it `yield`s is a reference to storage.

A `modify` accessor is thus a lot like the current `ref` accessor feature in Slang (they both expose a reference to storage to the caller), and both would allow the code generation at a call site to directly use the storage location that was returned/yielded to pass to a function with an `inout` parameter. The key difference is that because it is a coroutine, a `modify` accessor can have code that logically runs *after* the operation that needed access to the storage, which can handle cleanup, writing back to another location, validation, etc.

I like to think that `property` declarations as we are building them right now (with only `get` and `set`) are still a clear syntactic improvement over traditional getter/setter function pairs, even if we don't address some of these thornier challenges. We can hopefully make them perform better over time, but it is reasonable to say that in code that cares a lot about fine-grained performance details for `inout` parameter passing, programmers should favor ordinary fields over `property` declarations for now.

source/slang/slang-check-decl.cpp
```diff
@@ -3724,64 +3732,155 @@ namespace Slang
...
         {
-            param = m_astBuilder->create<ParamDecl>();
-            param->nameAndLoc.name = getName("newValue");
-            param->nameAndLoc.loc = decl->loc;
+            // If the user didn't declare any explicit parameters,
+            // then we create an implicit one and add it into
+            // the AST.
+            //
+            newValueParam = m_astBuilder->create<ParamDecl>();
+            newValueParam->nameAndLoc.name = getName("newValue");
```
csyonghe, line 3846: Should we stay consistent with C# and just use `value` instead?

source/slang/slang-check-decl.cpp
```diff
@@ -3724,64 +3732,155 @@ namespace Slang
...
         else
         {
             getSink()->diagnose(decl, Diagnostics::accessorMustBeInsideSubscriptOrProperty);
         }
+    }
 
-        checkCallableDeclCommon(decl);
+    void SemanticsDeclHeaderVisitor::visitAccessorDecl(AccessorDecl* decl)
+    {
+        _visitAccessorDeclCommon(decl);
```
csyonghe, line 3771: A property cannot have `void` type, right?

source/slang/slang-check-decl.cpp
tfoleyNV: I'd be fine with changing this default down the line, but I won't do it in this PR. We can revisit the default name when we add support for parsing the type in the traditional order you wanted.

source/slang/slang-check-decl.cpp
tfoleyNV: I'd like to start removing as many assumptions as possible about `void` being special in Slang, so that we can start treating it as an ordinary (but empty and zero-byte) type.

Is there a particular reason that concern came up here?

source/slang/slang-lower-to-ir.cpp
csyonghe: I did not mean that we should address the issue now. The comment is only there to remind us that this could be a place for potential performance improvement. A possible alternative to `ref` is we can have functions to take an `Assignable<T>` instead of `inout<T>`, where `Assinable<T>` provides `get` and `set` methods, and the compiler can synthesize such an `Assignable` object for properties or ordinary fields.

source/slang/slang-check-decl.cpp
csyonghe: That sounds reasonable to me.

source/slang/slang-check-decl.cpp
csyonghe: Sounds good.

PR #1411, opened 2020-06-24
Backend for Multiple Entry Points

Checkmate50:
Introduces the basic backend on the compiler for zero or more entry
points.  Entry points have been extended to lists for several functions,
with loopFunctions have been extended to take in entry points and
indices as appropriate, to allow for multiple entry points once the
frontend is expanded.  Several functions are currently being assumed to
have a single entry point for simplicity and provide a work in progress
commit.


source/slang/slang-dxc-support.cpp
```diff
@@ -82,7 +82,8 @@ namespace Slang
         // Now let's go ahead and generate HLSL for the entry
         // point, since we'll need that to feed into dxc.
         SourceResult source;
-        SLANG_RETURN_ON_FAIL(emitEntryPointSource(compileRequest, entryPointIndex, targetReq, CodeGenTarget::HLSL, endToEndReq, source));
+        //SLANG_RETURN_ON_FAIL(emitEntryPointsSource(compileRequest, List<Int>(entryPointIndex),
+        //    targetReq, CodeGenTarget::HLSL, endToEndReq, source));
```
tfoleyNV, line 86: Seems like this needs a fix.

source/slang/slang-options.cpp
```diff
@@ -1457,6 +1460,10 @@ struct OptionsParser
             // point was handled above, and the user is expected to
             // follow the ordering rules when using multiple entry points.
             //
+            // TODO(DG): check if output format is left unassociated
+            // If (known format for that output && that format supports whole program) {
+            //    set isWholeProgram bit on the TargetRequest }
+            // else { diagnose what happened }
```
tfoleyNV, line 1466: Just to note: I think it will be okay to check this PR in with these `TODO` comments in place, if we can reach a good checkpoint. I would just like to make sure that we address all the open `TODO`s related to this feature in follow-on changes.

source/slang/slang-emit.cpp
```diff
@@ -604,19 +608,23 @@ Result linkAndOptimizeIR(
...
+    // Temporary assertion for checkpoint
+    SLANG_ASSERT(entryPointIndices.getCount() == 1);
     outSource.reset();
 
     auto sink = compileRequest->getSink();
     auto program = compileRequest->getProgram();
 
-    auto entryPoint = program->getEntryPoint(entryPointIndex);
+    // TODO(DG): Update from assertion
+    auto entryPoint = program->getEntryPoint(entryPointIndices[0]);
```
tfoleyNV, line 627: Style-wise, I think it works a lot better to do:

```
SLANG_ASSERT(entryPointIndices.getCount() == 1);
auto entryPointIndex = entryPointIndices[0];
```

at the top of functions like this, and thereby avoid having to change any of the existing references to `entryPointIndex` downstream.

source/slang/slang-dxc-support.cpp
```diff
@@ -82,7 +82,10 @@ namespace Slang
         // Now let's go ahead and generate HLSL for the entry
         // point, since we'll need that to feed into dxc.
         SourceResult source;
-        SLANG_RETURN_ON_FAIL(emitEntryPointSource(compileRequest, entryPointIndex, targetReq, CodeGenTarget::HLSL, endToEndReq, source));
+        List<Int> entryPointIndices;
+        entryPointIndices.add(entryPointIndex);
+        SLANG_RETURN_ON_FAIL(emitEntryPointsSource(compileRequest, entryPointIndices,
+            targetReq, CodeGenTarget::HLSL, endToEndReq, source));
```
tfoleyNV, line 88: I would definitely advocate for introducing `emitEntryPointSource` (singular) back into the codebase so that call sites like this can be left unchanged, rather than having to build the singleton `List<Int>` in each case.

source/slang/slang-dxc-support.cpp
```diff
@@ -190,7 +193,9 @@ namespace Slang
             args[argCount++] = L"-enable-16bit-types";
         }
 
-        const String sourcePath = calcSourcePathForEntryPoint(endToEndReq, entryPointIndex);
+        List<Int> entryPointIndices2;
+        entryPointIndices.add(entryPointIndex);
+        const String sourcePath = calcSourcePathForEntryPoints(endToEndReq, entryPointIndices2);
```
tfoleyNV, line 198: It would be a non-issue if we introduce the singular `calcSourcePathForEntryPoint` back as a wrapper, but I will note that the need to declare `entryPointIndices2` here is redundant. You could just use the `entryPointIndices` you already constructed earlier in the function.

Alternatively, you could wrap up these first call to `emitEntryPointsSource` in its own scope with `{ ... }` in order to narrow the lifetime of the first `entryPointIndices` variable. Either way, needing to use a variable name like `entryPointIndices2` doesn't feel clean.

source/slang/slang-compiler.h
```diff
@@ -1944,7 +1956,10 @@ namespace Slang
     @param endToEndReq The end-to-end compile request which might be using pass-through compilation
     @param entryPointIndex The index of the entry point to compute a filename for.
     @return the appropriate source filename */
-    String calcSourcePathForEntryPoint(EndToEndCompileRequest* endToEndReq, UInt entryPointIndex);
+    // TODO(DG): Note to reviewer; this was changed from UInt to List<Int> -- let me know if that's a problem
+    //   and I can work out the appropriate casts
```
tfoleyNV, line 1960: The new versions look good at this point, so I'd remove this `TODO` comment.

source/slang/slang-compiler.cpp
```diff
@@ -1686,159 +1761,160 @@ SlangResult dissassembleDXILUsingDXC(
...
-                {
-                    maybeDumpIntermediate(compileRequest, downstreamResult, target);
+            if (SLANG_SUCCEEDED(emitWithDownstreamForEntryPoints(
+                compileRequest,
+                getEntryPointIndices(entryPoints),
+                targetReq,
+                endToEndReq,
+                downstreamResult)))
+            {
+                maybeDumpIntermediate(compileRequest, downstreamResult, target);
```
tfoleyNV, line 1774: More formatting changes on `switch` statements here that are making it harder to evaluate the actual diffs.

PR #1412, opened 2020-06-25
AST serialize improvements

jsmall-nvidia:
* Initial documentation on AST serialization mechanism
* Fix some bugs and special cases 
* Added controls for ASTDumpUtil to control whats dumped
* Gave AST Context type a unique name (as opposed to anonymous Context)

With these changes the AST serialization can now save and reconstruct ASTs as generated by typical Slang compiler usage. This was tested via using the ASTDumpUtil to dump an AST tree to compare the original AST tree to the deserialized tree. 

The main thing missing in that comparison is SourceLoc information.


source/slang/slang-ast-serialize.h
```diff
@@ -7,13 +7,188 @@
...
+
+It is useful to have a structure that holds the type information, so it can be stored. That is achieved with
+
+```
+template <typename T>
+struct ASTSerialGetType;
+```
+
+This template can be specialized for a specific native types - but all it holds is just a function getType, which returns a ASTSerialType*,
+which just holds the information held in the ASTSerialTypeInfo template, but additionally including the size of the SerialType.
```
tfoleyNV, line 66: This is a difficult sentence to work through before finishing my morning coffee. I *think* what you are getting at is that you have a custom RTTI system for serialization, so that every type to be serialized is associated with some `ASTSerialType*`, and this template is the thing that handles fetching the RTTI for a given type `T`, while the `ASTSerialTypeInfo<T>` template is what is used to *define* the policy stuff or a given type.

source/slang/slang-ast-serialize.h
```diff
@@ -7,13 +7,188 @@
...
+
+So we need to define a specialized ASTSerialTypeInfo for each type that can be a field in a NodeBase derived type. We don't need to define
+anything explicitly for the NodeBase derived types, as we will just generate the layout from the fields. How do we know the fields? We just
+used the macros generated from the C++ extactor.
+
+So first a few things to observe...
+
+1) Some types don't need any conversion to be serializable - int8_t, or float the bits can just be written out and read in (1)
```
tfoleyNV, line 74: I assume that we are implicitly following the policy of the rest of the codebase to juse assume little endian has taken over enough of the world that worrying about endian conversion doesn't pay off.

source/slang/slang-ast-serialize.h
```diff
@@ -7,13 +7,188 @@
...
+
+For (3) there are a few scenarios. For any field in a serial type we must store in the serialized type such that the representation
+will work across all processors/compilers. So one problematic type is `bool`. It's not specified how it's laid out in memory - and
+some compiles have stored it as a word. Most recently it's been stored as a byte. To make sure bool is ok for serialization therefore
+we store as a uint8_t.
+
+Another example would be double. It's 64 bits, but on some arches/compilers it's SLANG_ALIGN_OF is 4 and on others it's 8. On some
+arches a non aligned read will lead to a fault. To work around this problem therefore we have to ensure double has the alignment that
+will work across all targets - and that alignment is 8. In that specific case that issue is handled via ASTSerialBasicTypeInfo, which
+makes the SerialAlignment the sizeof the type.
```
tfoleyNV, line 105: An alternative approach would seemingly be to have the "backing storage" for the serialized form not provide any alignment guarantees, and then to use `memcpy` whenever read/writing so that we don't have to spend any time/effort on alignment.

I'm not advocating for this option right now, but just noting that it exists.

source/slang/slang-ast-serialize.h
```diff
@@ -7,13 +7,188 @@
...
+
+Lazy reconstruction may be useful in the future to partially reconstruct a sub part of the serialized data. In the current implementation, lazy
+evaluation is used on Strings. The m_objects array holds all of the recreated native 'objects'. Since the objects can be derived from different
+base classes the associated Entry will describe what it really is.
+
+For the String type, we initially store the object pointer as null. If a string is requested from that index, we see if the object pointer is null,
+if it is we have to construct the StringRepresentation that will be used.
+
+An extra wrinkle is that we allow accessing of a serialized String as a Name or a string or a UnownedSubString. Fortunately a Name just holds a string,
+and a Name remains in scope as long as it's NamePool does which is passed in.
```
tfoleyNV, line 188: I'm not sure I fully follow this bit. Are you saying we serialize the `NamePool` as part of the AST, or that we will have a `NamePool` available when deserializing so that we can use it to construct `Name`s from strings as needed?

tfoleyNV:
Now that I follow the logic of the format a bit better, I want to make one note that we might care about for the future:

Because we serialize most nodes as just a list of their field values, it could be important in the future that we have a way to identify the fields of a type in a binary-stable way across compiler versions. That is, we might want the C++ extractor to work with meta-data macros that assigns unique IDs to the (serializable) fields of a type, and then when we serialize an AST part of the serialized data would be information about the IDs of the fields (and their order) that is being written for each AST node type.

That logic would complicate deserialization of AST nodes because it would need to indirect through that serialized type information to know what fields are present in the serialized data and which aren't, but it would seem to be required if we ever want the serialized ASTs to be portable across compiler versions (which isn't needed right away, but might be important some day).

source/slang/slang-ast-serialize.h
jsmall-nvidia: For ir serialization - I did nod to endianess in that with that serialized format it is relatively easy to handle endianness. Here it could be supported but would probably want some other template stuff to define a per serialized type function that could handle endianess. That it only needs to work at the field type level (and only with serialized types) - as NodeBase types can just use that field knowledge to flip those types.

source/slang/slang-ast-serialize.h
jsmall-nvidia: Indeed. Whilst this approach takes up slightly more space, and requires extra care, just being able to access directly in memory makes this approach fast, more easily debuggable and in many ways simpler.

source/slang/slang-ast-serialize.h
jsmall-nvidia: The latter we have a NamePool to deserialize Names into but additionally (and perhaps less obviously) it is using the NamePool to scope UnownedStringSlices (for example as held in Token type in AST). That when a 'string' is accessed it can be accessed in 3 ways - a String, a Name and an UnownedStringSlice. So how it's stored may change depending on how it is accessed.

tfoleyNV: > we have a NamePool to deserialize Names into but additionally (and perhaps less obviously) it is using the NamePool to scope UnownedStringSlices (for example as held in Token type in AST)

There might be an argument there to do some collapsing/unification between the `NamePool` and the string pool type(s) that you'd introduced as part of IR serialization. We could have a base layer that just handles deduplication and scoping the lifetime of strings, and a higher-level layer that simply wraps those deduplicated strings up as `Name`s (which mostly exist just so that AST-level code doesn't get confused and try to use plain strings as names, thus breaking our deduplication assumptions)

PR #1416, opened 2020-06-26
Clean up unused code for IR object ownership

tfoleyNV:
There was a small but non-trivial amount of code across `IRModule`, the `ObjectScopeManager`, and `StringRepresentationCache` that had to do with managing the lifetimes of `RefObject`s that might be referenced by IR instructions (and thus need to be kept alive for the lifetime of the IR module).

We have long since migrated to a model where IR instruction do not include owned references to `RefObject`s, so these facilities weren't actually needed. This streamlines `IRModule`'s declaration, and trims code that we aren't actually using.

One note for the future is that the `StringRepresentationCache` no longer does what its name implies (it is not a cache of `StringRepresentation`s), so we should consider giving it a more narrowly scoped name. I didn't include that in this change because I wanted to keep the diffs narrow and easy to review.

A follow-on renaming change should be trivial if/when we can agree on what the type should be called at this point. Alternatively, we could simply bake the functionality of `StringRepresentationCache` into he IR deserialiation logic itself, since that is the only code using it.


source/slang/slang-ir-serialize-types.h
```diff
@@ -22,36 +21,29 @@ enum class IRSerialCompressionType : uint8_t
 
 class StringRepresentationCache
 {
+    // TODO: The name of this type is no longer accurate to its function.
```
csyonghe, line 24: Do we plan to rename it in a future PR?

tfoleyNV: That would be my goal, yeah. I like to avoid putting bulk renaming/refactoring changes in the same PR as functionality changes, since it makes it harder to review.

A small PR that requires line-by-line review is usually easy.

A large PR that only has renaming changes is also easy to review.

The combination of the two is usually not easy to review.

PR #1418, opened 2020-07-01
Emit pointers for CPU target

csyonghe:
With this change, `IRPtrType` now translates to actual pointers in C++ backend.

All `var` definitions will still emit as `Type var`, but all references to `var` insts will now emit as `&var_name`, since they have pointer type in the IR.

A `*` operator will be added before the following operands:
- The pointer operand of `load` inst
- The pointer operand of 'store' and `swizzledStore` inst

Meanwhile, the C++ prelude has changed so that all reference-typed function parameters are now pointer-typed, since references are represented as poiners in the IR.

This handles most cases. One thorny case is the handling of `__subscript` for HLSL builtin types such as `StructuredBuffer` or `RWTexture2D` etc. Before this change, everything just works with C++ references. With explicit pointers, now it matters to distinguish between the `getter`, `setter` and `ref` accessor of the subscript call, since they will have different argument return types. For this reason, additional code has been added to `emitIntrinsicCallExprImpl` to determine which accessor is being used, and generate different code accordingly.


prelude/slang-cpp-types.h
```diff
@@ -20,7 +20,8 @@ struct FixedArray
 {
     const T& operator[](size_t index) const { SLANG_PRELUDE_ASSERT(index < SIZE); return m_data[index]; }
     T& operator[](size_t index) { SLANG_PRELUDE_ASSERT(index < SIZE); return m_data[index]; }
-    
+    T* elementPtr(size_t index) { SLANG_PRELUDE_ASSERT(index < SIZE); return m_data + index; }
+
```
tfoleyNV, line 24: Should we lower things to use `&array[index]` instead of `array.elementPtr(index)`? Avoiding an explicit `elementPtr` would mean that it is easier to swap out `FixedArray<T,N>` for an actual `T[N]` if we want to (or if we eventually want to target plain C).

source/slang/slang-emit-c-like.cpp
```diff
@@ -969,10 +969,11 @@ bool CLikeSourceEmitter::shouldFoldInstIntoUseSites(IRInst* inst)
         type = arrayType->getElementType();
     }
 
-    // Don't allow temporaries of pointer types to be created.
+    // Don't allow temporaries of pointer types to be created,
+    // if target langauge doesn't support pointers.
     if(as<IRPtrTypeBase>(type))
     {
-        return true;
+        return !supportPtrTypes();
```
tfoleyNV, line 976: `supportPtrTypes()` is maybe a bit ambiguous. I might argue for either `doesTargetSupportPtrTypes()` or `arePtrTypesSupported()` since both of those make it clear just from the name that they return a `bool` based on whether or not pointer types are supported.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1161,6 +1162,40 @@ bool CLikeSourceEmitter::shouldFoldInstIntoUseSites(IRInst* inst)
     return true;
 }
 
+void CLikeSourceEmitter::emitDereferenceOperand(IRInst* inst, EmitOpInfo const& outerPrec)
+{
```
tfoleyNV, line 1166: The obvious alternative here would be to have `emitDereferenceOperand` be a customization point on `CLikeSourceEmitter` where subtypes override these operations. I think you are probably making the right choice here since there are currently only two options (actually emit the op, or ignore it).

source/slang/slang-emit-c-like.cpp
```diff
@@ -2019,18 +2057,35 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
...
+
+            auto base = ii->getBase();
+            emitOperand(base, leftSide(outerPrec, prec));
+            m_writer->emit(".");
+            if(getSourceLanguage() == SourceLanguage::GLSL
+                && as<IRUniformParameterGroupType>(base->getDataType()))
+            {
+                m_writer->emit("_data.");
+            }
+            m_writer->emit(getName(ii->getField()));
```
tfoleyNV, line 2087: Some of this could conceivably be simplified by treating this as emitting `&((*base).field)` and then using subroutines like `emitDereferenceOperand` to emit both the `*` and `&`. For a pointer-based target both of those would emit, and for a non-pointer target neither would.

Of course that approach wouldn't get you the nice `->` instead of `(*base).field`, but we need to strike a balance here between the clarity of the generated code and the clarity of the emit logic.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1161,6 +1162,40 @@ bool CLikeSourceEmitter::shouldFoldInstIntoUseSites(IRInst* inst)
...
+{
+    if (supportPtrTypes())
+    {
+        auto dereferencePrec = EmitOpInfo::get(EmitOp::Prefix);
+        EmitOpInfo newOuterPrec = outerPrec;
+        bool needClose = maybeEmitParens(newOuterPrec, dereferencePrec);
+        m_writer->emit("*");
+        emitOperand(inst, rightSide(dereferencePrec, newOuterPrec));
+        maybeCloseParens(needClose);
+    }
```
tfoleyNV, line 1175: You could add a special case here for when `inst->op == kIROp_Var`, and emit `varName` instead of `*&varName` like this would do by default. That would be the kind of "peephole" optimization I'm talking about.

A more complicated solution (with a lot of chances for errors) would be to pass down a "dereference level" alongisde `outerPrec` (maybe even fold that *into* `EmitOpInfo`). That would be an integer that tracks the pending number of dereference (positive) or address-of (negative) operations that are needed in front of an operand.

Most emit cases would immediately flush all of those by emitting prefix `*` or `&` operations head of the op they normally emit. An operation like `load` would just emit their operand with a `+ 1` to the dereference level, and when emitting a reference to a `var` we'd apply a `-1` before emitting the prefix operators to indicate that it is name embeds an implicit dereference.

I'm not advocating for anything that complicated right now, but we might want to think about it if we decide that producing idiomatic-looking output is important.

source/slang/slang-emit-c-like.cpp
```diff
@@ -2169,13 +2224,41 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
...
+                {
+                    const auto info = getInfo(EmitOp::Add);
+
+                    needClose = maybeEmitParens(outerPrec, info);
+                    emitOperand(inst->getOperand(0), leftSide(outerPrec, info));
+                    m_writer->emit(" ");
+                    m_writer->emit(info.op);
+                    m_writer->emit(" ");
+                    emitOperand(inst->getOperand(1), rightSide(outerPrec, info));
+                }
```
tfoleyNV, line 2249: This seems like it can just emit as `&operand0[operand1]` for both cases, rather than needing to special-case either side.

source/slang/slang-emit-c-like.cpp
```diff
@@ -2169,13 +2224,41 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
...
 
-            emitOperand( inst->getOperand(0), leftSide(outerPrec, prec));
-            m_writer->emit("[");
-            emitOperand(inst->getOperand(1), getInfo(EmitOp::General));
-            m_writer->emit("]");
+                emitOperand(inst->getOperand(0), leftSide(outerPrec, prec));
+                m_writer->emit("[");
+                emitOperand(inst->getOperand(1), getInfo(EmitOp::General));
+                m_writer->emit("]");
+            }
```
tfoleyNV, line 2261: If we have an operation like `emitDereferenceOperand` that handles emitting a `&` if needed, and skipping it if not, we could collapse all these cases down to one.

The main difficulty with an operation that emits prefix `&` or `*` as needed is that it might lead to parentheses being needed. It seems like overall the `needClose` setup in this code should change from `bool needClose` to `int closeCount` which tracks the number of closing `)` needed. Then we can have steps that do `closeCount += maybeEmitPrefixAddrOf(...)`.

In the far-flung future, a major alternative here would be to have the emit logic build up a "rope" structure in a bottom-up fashion, rather than the current approach that emits text directly in a top-down fashion.

source/slang/slang-emit-c-like.cpp
csyonghe: Done.

source/slang/slang-emit-c-like.cpp
csyonghe: Done.

prelude/slang-cpp-types.h
csyonghe: Done.

source/slang/slang-emit-c-like.cpp
csyonghe: This was originally a virtual function that can be customized like you said. But then I find out it really doesn't give you much and turn it back into just an ordinary function.

source/slang/slang-emit-c-like.cpp
csyonghe: Change of `needClose` is a big change on its own, so I'd rather leave that as a future improvement if we determine it is necessary.

source/slang/slang-emit-c-like.cpp
csyonghe: Done.

source/slang/slang-emit-c-like.cpp
csyonghe: Done.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2095,21 +2093,39 @@ void CPPSourceEmitter::emitIntrinsicCallExprImpl(
...
             // The user is invoking a built-in subscript operator
-            auto prec = getInfo(EmitOp::Postfix);
-            needClose = maybeEmitParens(outerPrec, prec);
 
-            emitOperand(args[0].get(), leftSide(outerPrec, prec));
-            m_writer->emit("[");
-            emitOperand(args[1].get(), getInfo(EmitOp::General));
-            m_writer->emit("]");
+            // Determine if we are calling the `ref` accessor:
+            // `ref` accessor returns a pointer of element type.
```
tfoleyNV, line 2098: This isn't a problem with this PR, but the steps required to special-case subscript calls here are out of control, and we really need to go back and triage the Slang stdlib so that all the built-in subscripts and their accessors have their emit logic defined in the stdlib instead of here.

PR #1427, opened 2020-07-02
Multiple Entry Point Cleanup

Checkmate50:
This commit provides some in-code cleanup of the previous multiple entry point PR (#1411).  Specifically, this change provides refactoring of multiple entry point functions into helper functions, the removal of the EntryPointAndIndex struct, and various stylistic improvements.


source/slang/slang-compiler.cpp
```diff
@@ -1645,8 +1619,9 @@ SlangResult dissassembleDXILUsingDXC(
         List<uint8_t>&          spirvOut);
 
     SlangResult emitSPIRVForEntryPointsViaGLSL(
+        ComponentType*                  program,
         BackEndCompileRequest*          slangRequest,
-        List<EntryPointAndIndex> const& entryPoints,
+        List<Int>                       entryPointIndices,
```
tfoleyNV, line 1624: Reminder: please prefer `List<...> const&` over just passing a `List<...>`.

source/slang/slang-compiler.cpp
```diff
@@ -1699,8 +1673,9 @@ SlangResult dissassembleDXILUsingDXC(
     }
 
     SlangResult emitSPIRVForEntryPoints(
+        ComponentType*                  program,
         BackEndCompileRequest*          slangRequest,
-        List<EntryPointAndIndex> const& entryPoints,
+        List<Int>                       entryPointIndices,
```
tfoleyNV, line 1678: `List<Int> const&`

source/slang/slang-compiler.cpp
```diff
@@ -1709,32 +1684,35 @@ SlangResult dissassembleDXILUsingDXC(
...
                 endToEndReq,
                 spirvOut);
         }
     }
 
     SlangResult emitSPIRVAssemblyForEntryPoints(
+        ComponentType*                  program,
         BackEndCompileRequest*          slangRequest,
-        List<EntryPointAndIndex> const& entryPoints,
+        List<Int>                       entryPointIndices,
```
tfoleyNV, line 1706: `List<Int> const&`

source/slang/slang-compiler.cpp
```diff
@@ -1748,8 +1726,9 @@ SlangResult dissassembleDXILUsingDXC(
 
     // Do emit logic for a zero or more entry points
     CompileResult emitEntryPoints(
+        ComponentType*                  program,
         BackEndCompileRequest*          compileRequest,
-        List<EntryPointAndIndex> const& entryPoints,
+        List<Int>                       entryPointIndices,
```
tfoleyNV, line 1731: `List<Int> const&`

source/slang/slang-compiler.h
```diff
@@ -1984,6 +1982,14 @@ namespace Slang
         EndToEndCompileRequest* endToEndReq,
         SourceResult&           outSource);
 
+    SlangResult emitEntryPointSource(
+        BackEndCompileRequest*  compileRequest,
+        Int                     entryPointIndices,
```
tfoleyNV, line 1987: This should probably be `entryPointIndex` instead of `...Indices`.

source/slang/slang-dxc-support.cpp
```diff
@@ -46,8 +46,8 @@ namespace Slang
     }
 
     SlangResult emitDXILForEntryPointUsingDXC(
+        ComponentType*          program,
         BackEndCompileRequest*  compileRequest,
```
tfoleyNV, line 50: I believe that we can technically extract the `ComponentType` that represents the program from the `BackEndCompileRequest` rather than have to pass it down, but this is fine in the name of being more explicit.

PR #1430, opened 2020-07-02
ShortList<T> and core.natvis improvements.

csyonghe:
This change adds `ShortList` to `core`, and improves debugging experience of `Dictionary` types. Empty items in a `Dictionary` will no longer be displayed.


source/core/slang-short-list.h
```diff
@@ -0,0 +1,487 @@
...
+
+        template <typename IterateFunc>
+        void forEach(IterateFunc f) const
+        {
+            for (Index i = 0; i < m_count; i++)
+                f(m_buffer[i]);
+        }
+
+    private:
+        T* m_buffer = nullptr;           ///< A new T[N] allocated buffer. NOTE! All elements up to capacity are in some valid form for T.
```
tfoleyNV, line 454: Eventually it seems like we will want to change the invariants on `List<T>` and `ShortList<T>` so that they do *not* initialize elements up to the capacity, and instead behave more like standard containers where only the elements used are initialized.

source/core/slang-short-list.h
```diff
@@ -0,0 +1,487 @@
...
+            {
+                result.ownsStorage = true;
+                result.arrayView.m_count = count;
+                result.arrayView.m_buffer = ThisType::_allocate(count);
+                for (Index i = start; i < shortListSize; i++)
+                    result.arrayView.m_buffer[i - start] = m_shortBuffer[i];
+                for (Index i = shortListSize; i < start + count; i++)
+                    result.arrayView.m_buffer[i - start] = m_buffer[i - shortListSize];
+                return result;
+            }
```
tfoleyNV, line 218: Couldn't we simplify things by always moving all elements over to the allocated `m_buffer` as soon as you need more than `shortListSize` elements? It seems like we could streamline a lot of logic by having an invariant that things are always contiguous (e.g., it would also make the iterator a bit simpler - just a pointer).

csyonghe: I thought about that and struggled between the two implementations. Maybe I made the wrong decision.

tfoleyNV: I don't think we have to change it right away, but I do think that contiguous storage is an important property that we should only give up when there isn't a reasonable alternative.

One unrelated note is that sooner or later we will probably need/want to revisit the naming of the `List` type (which would affect `ShortList` too). Naming these as `Array<T>` and `ShortArray<T,N>` is probably the front-runner, but it would be good to reach consensus. The `List` name ends up confusing many people because most examples of "list" data structures are node-based rather than contiguous. Even something like Java `ArrayList` would be more clear about the underlying implementation.

PR #1434, opened 2020-07-08
Dynamic code gen for generic local variables.

csyonghe:
This PR adds basic support of generic local variables when generating dynamic dispatch code.
For example,
```
interface IInterface
{
    [mutating]
    int Compute(int inVal);
};

int GenericCompute<T:IInterface>(int inVal)
{
    T obj;
    obj.Compute(3);
    return obj.Compute(inVal); // 3 + inVal
}
```

When compiling a generic function, the generic type parameter `T` is lowered into an ordinary `RTTI` parameter, i.e. `__rtti_t rttiObj_T`, which is used to store type information of `T`. In this PR, the only info stored in a RTTI object is the type size. A RTTI object is represented as a set of key-value pairs (`IRRRTIEntry`) as operands in an `IRRTTIObject` inst. 

A local variable declaration `T obj` is then translated into a stack allocation:
```
obj = alloca(rtti_extract_size(rttiObj_T))
```
When calling a generic function with a concrete type `S`, the compiler construct an  RTTI object for `S` at global scope, and pass a pointer to the RTTI object as argument to the lowered generic type parameter. An `alloca` instruction returns `RTTIPointer(ptr_to_rtti_object)` type. `Load`s of pointers of `RTTIPointer` type can only occur as the `value` argument in a `Store` inst (for assignments) or when calling another generic/interface function. In both cases, we don't really need the load and can just use the pointer instead. `lowerGenerics` pass is changed to include the logic to eliminate such `Load`s.

This PR implements RTTI object as a new type of entity in the IR: `IRRTTIObject`, which has type `IRRTTIType`. Related instructions are added to construct and extract info from a RTTI object: `IRMakeRTTIObject`, `IRRTTIExtractSize`.

`Store`s of generic variables are translated into `Copy`s in the IR, and `Copy` instructions will be emitted as `memcpy` calls in `CPPEmitter`.


csyonghe:
There are still some fundamental issues, this PR is not ready for review yet.

csyonghe:
@tfoleyNV This PR is ready for review.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -208,6 +237,133 @@ namespace Slang
...
+            // 2. call(f, load(genVar)) when calling a generic function or a member function
+            //    via an interface witness lookup. In this case, we need to replace with
+            //    just `genVar`, since that function has already been lowered to take
+            //    raw pointers.
+            // Here we replace all uses of load to just the pointer itself.
+            // After this, all arguments in `call`s will be in its correct form.
+            // All `store`s will become `store(x, genVar)`, and still need
+            // to be translated into a `copy`, we leave that step when we get to
+            // process the `store` inst.
+            loadInst->replaceUsesWith(loadInst->getOperand(0));
```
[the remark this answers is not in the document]
tfoleyNV: A final note here is that at some level the ideal situation would be to completely eliminate all `var`s of generic/assocaited type, and somehow have code that only uses SSA temporaries of those types. Doing so doesn't eliminate the need for `alloca()` or register allocation strategies, but it does eliminate the issue of mutation.

Ideally the one exception to this rule would be the `var`s we currently allocate to manage the storage for `inout` and `out` arguments. We could conceivably mark those variables with a decoration (or even us a special opcode/type/something) to make their role clear, and then have special-case strategies to handle those `var`s (e.g., the `var` for an `inout` argument will never be mutated after its use as an argument, so it is safe to use its pointer as a replacement for downstream `load`s.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -208,6 +237,133 @@ namespace Slang
...
+            auto typeSize = builder->getIntValue(builder->getIntType(), sizeAndAlignment.size);
+            auto sizeEntry = builder->emitRTTIEntry(kRTTISize, typeSize);
+            result = builder->emitMakeRTTIObject(makeArrayView(sizeEntry));
+
+            // Give a name to the rtti object.
+            if (auto exportDecoration = typeInst->findDecoration<IRExportDecoration>())
+            {
+                String rttiObjName = "__rtti_" + String(exportDecoration->getMangledName());
+                builder->addExportDecoration(result, rttiObjName.getUnownedSlice());
+            }
```
tfoleyNV, line 362: Can the RTTI object just use the mangled name of the type itself for export? At this point we are late enough in the compiler that we are concerned with the linker of the target machine and not the Slang linker, so there is no need to worry about clashes with the mangled name of the type itself. Prefixing the name is fine for now, but seems like overkill.

(Also: one minor note is that some platforms/toolchains apparently support compressing string tables for binaries based on the idea that many symbol names will have long prefixes in common. As a result, we should tend to pick mangling strategies that make related symbols have the same prefix, which means adding the text that differentiates symbols to the end of the name instead of the beginning.)

source/slang/slang-ir.cpp
```diff
@@ -634,6 +634,13 @@ namespace Slang
         return entryBlock->getParams();
     }
 
+    IRInst* IRGlobalValueWithParams::getFirstNonParamInst()
+    {
+        auto lastParam = getLastParam();
+        if (lastParam == nullptr)
+            return getFirstBlock()->getFirstInst();
+        return lastParam->next;
+    }
```
tfoleyNV, line 643: As discussed elsewhere, I *think* this function does what `getFirstOrdinaryInst` already does.

source/slang/slang-ir.cpp
```diff
@@ -5043,6 +5132,8 @@ namespace Slang
         case kIROp_Nop:
         case kIROp_Specialize:
         case kIROp_lookup_interface_method:
+        case kIROp_RTTIExtractSize:
+        case kIROp_getAddr:
```
tfoleyNV, line 5136: Is `getAddr` still needed? I thought that with the explicit pointer-ification of things we could remove that because it was a bit of a hack...

source/slang/slang-ir.h
```diff
@@ -729,6 +729,16 @@ struct IRPtrLit : IRConstant
...
 };
 
+// Range of keys of an RTTI entry.
+// For an `RTTIEntry(op0, op1)`, `op0` is an `IRIntLit` whose value
+// must be one of the enums defined in `RTTIEntryKeys`.
+enum RTTIEntryKeys : int32_t
+{
+    kRTTISize, // Size in bytes of the type represented by the RTTI object.
+    kRTTIKeys_End, // Marks the count of possible key values.
+                   // Not a valid value to appear in the IR.
```
tfoleyNV, line 739: This is a rathole of naming convention stuff, but:

* My strong prefenece is that `enum` type names should be singular: e.g., `RTTIEntryKey` because when used as a type they represent a single value (this makes function signatures read better).

* My personally preference is that non-`enum class` enumerants should be of the form `k<TypeName>_<CaseName>`. So `kRTTIEntryKey_Size` would be my preference.

* There are dueling conventions in the codebase today about how to name the sentinel value past the end. I tend to use `k<TypeName>Count` while Jonathan tends to use `k<TypeName>_CountOf` (which doesn't parse grammatically for me...). This change would add a third option of `k<TypeName>_End` assuming you meant to write the full type name instead of just `RTTIKeys`...

source/slang/slang-ir-insts.h
```diff
@@ -1712,6 +1764,11 @@ struct IRBuilder
         UInt            argCount,
         IRInst* const* args);
 
+    IRInst* emitRTTIEntry(RTTIEntryKeys key, IRInst* value);
```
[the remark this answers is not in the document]
csyonghe: I removed `RTTIEntry` and use decorations instead. Now `RTTIObject` becomes an inst that takes one operand -- for the type this RTTI is about, and all type info comes in the form of decorations. For now there is only one `RTTITypeSizeDecoration`.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -15,6 +16,9 @@ namespace Slang
         // we are processing.
         IRModule* module;
 
+        // RTTI objects for each type used to call a generic function.
+        Dictionary<IRInst*, IRInst*> rttiObjects;
```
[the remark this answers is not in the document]
csyonghe: Done.

source/slang/slang-ir.h
csyonghe: This is removed. We are using decorations instead of these entries so they are no longer needed.

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -95,79 +130,73 @@ namespace Slang
...
-            // Turn generic parameters into void pointers.
-            for (auto param : cast<IRFunc>(loweredFunc)->getParams())
+            // Lower generic typed parameters into RTTIPointers.
+            auto firstInst = loweredFunc->getFirstNonParamInst();
+            builder.setInsertBefore(firstInst);
+
+            for (IRInst* param = loweredFunc->getFirstParam();
+                param && param->op == kIROp_Param;
+                param = param->getNextInst())
             {
```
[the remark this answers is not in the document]
csyonghe: I can change that, but I will need an explicit `if (param->op != kIROp_Param) break;` statement in the for loop, so I am not sure if that is better.

source/slang/slang-emit-cpp.cpp
```diff
@@ -1695,11 +1703,22 @@ void CPPSourceEmitter::_emitWitnessTableWrappers()
                             if (reqParamType->op == kIROp_RawPointerType &&
                                 param->getFullType()->op != kIROp_RawPointerType)
                             {
-                                m_writer->emit("*static_cast<");
-                                emitType(param->getFullType());
-                                m_writer->emit("*>(");
-                                m_writer->emit(getName(param));
-                                m_writer->emit(")");
+                                if (as<IRPtrTypeBase>(param->getFullType()))
+                                {
```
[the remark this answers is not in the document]
tfoleyNV: I don't think we need to do anything right now, but the thing I'd argued for is having a named predicate that tests for whatever property of a type we care about here:

```
bool doesTypeNeedAddedIndirectionWhenPassedGenerically(IRType* type);
```

Part of my point is that I don't actually know exactly what property of the types involved is being tested for here, and that means the code isn't really self-documenting right now.

The eventual goal should be to have a well-defined vocabulary for talking about types, and predicates that can test for whether types do or do not have the terms in that vocabulary.

Some examples of properties we might eventually need/want:

* A type can either have a fixed/known size, or not. Non-concrete generic, associated, and `This` types are all not fixed-size.

* A type can be "memory-only" meaning that in final generated code it cannot be stored used as the type of an SSA temporary. Types that are not fixed-size must be memory-only. Our compiler needs to ensure that we eliminate any SSA temporaries of memory-only types before emitting code, either by specializing them out or by applying you `alloca()`-based legalization.

* Two types can be layout-compatible if there is a bidirectional value-preserving conversion between the two that consists of just reinterpreting bits (e.g., a `T*` and a `RawPtr` are always layout-compatible).

I'm not saying those are the properties we nee because, again, I'm not 100% clear on the predicate we are testing for here. The fact that the predicate needs to agree between callers and callees means that having a name for the properties and subroutines to check them seems important.

source/slang/slang-ir-lower-generics.cpp
tfoleyNV: Right now we don't use the mangled names for symbols *at all* when emitting code, so the mangled names don't really affect anything downstream from the IR linker.

Where this would start to be an issue is if we wanted to compile `a.slang` -> `a.cpp` and then `b.slang` -> `b.cpp` and then compile and link both `a.cpp` and `b.cpp` together (resolving cross-references between them), which seems to be exactly where all your current work is headed (in terms of capabilities).

For `a.cpp` to refer to a function that will be deifned in `b.cpp` they will need a way to agree on the name that will be used for the platform linker, and the current strategy of generating unique names by taking the "name hint" decoration and adding a numeric suffix obviously won't work (one module could name something `foo_1` and the other could name the same thing `foo_3` because there happened to be other things named `foo`).

The right fix would be that when we are in "separate compilation mode" (or however we want to enable this stuff), we emit all the symbols in our generated C/C++ as `extern "C"` symbols using the Slang-generated mangled name:

```c++
// a.cpp
extern "C" int _S1b3foo...(float x);

// b.cpp
extern "C" int _S1b3foo...(float x)
{
    ...
}
```

All that matters for making the platform linker do the right thing is that modules agree on the mangled name for symbols that have linkage. Because we are using `extern "C"` the name is all that matters, and the names of the types of parameters/results don't matter.

The point I was getting at is then that *types* do not have linkage in terms of traditional platform linkers; there is no such thing as an `extern "C" struct` in C/C++. As such, we would never need to emit a `struct Bar` using a mangled name in generated C/C++ code, and there would be no problem with using `Bar_0` in one generated file and `Bar_1` in another (again, because we use `extern "C"` linkage for all functions, so that the names of parameter types don't affect the name of the function's symbol).

Because the type itself doesn't need to be emitted using a mangled name, the corresponding mangled name is available and free to use for any `extern "C"` RTTI object declarations/definitions we want to emit.

PR #1437, opened 2020-07-10
Multiple Entry Point Backend

Checkmate50:
This PR introduces changes to the IR linking, emitting, and options for
multiple entry points.  Specifically, this PR updates several locations
to support a (potentially empty) list of entry points, adding list 
infrastructure and looping over entry points as appropriate.


examples/heterogeneous-hello-world/shader.slang
```diff
@@ -3,9 +3,8 @@
 //TEST_INPUT:ubuffer(random(float, 4096, -1.0, 1.0), stride=4):name=ioBuffer
 RWStructuredBuffer<float> ioBuffer;
 
-[shader("compute")]
 [numthreads(4, 1, 1)]
-void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+public void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
```
tfoleyNV, line 7: I'm not sure that changing this signature in the way that was done here is exactly what we want, but I suppose it doesn't really matter to the implementation of the feature.

source/slang/slang-emit-spirv.cpp
```diff
@@ -1115,19 +1115,21 @@ struct SPIRVEmitContext
 SlangResult emitSPIRVFromIR(
     BackEndCompileRequest*  compileRequest,
     IRModule*               irModule,
-    IRFunc*                 irEntryPoint,
+    List<IRFunc*>           irEntryPoints,
```
tfoleyNV, line 1118: `List<IRFunc*> const&`

source/slang/slang-emit.cpp
```diff
@@ -805,7 +802,7 @@ SlangResult emitEntryPointSourceFromIR(
 SlangResult emitSPIRVFromIR(
     BackEndCompileRequest*  compileRequest,
     IRModule*               irModule,
-    IRFunc*                 irEntryPoint,
+    List<IRFunc*>           irEntryPoints,
```
tfoleyNV, line 805: `List<IRFunc*> const&`

source/slang/slang-ir-link.cpp
```diff
@@ -1448,12 +1447,12 @@ LinkedIR linkIR(
...
     //
-    // TODO(DG): spot to generalize to multiple entry points
-    // Note that only stuff referenced by an entry point gets linked here
-    // Temporary assertion for checkpoint
-    SLANG_ASSERT(entryPointIndices.getCount() == 1);
-    auto entryPointMangledName = program->getEntryPointMangledName(entryPointIndices[0]);
-    auto irEntryPoint = specializeIRForEntryPoint(context, entryPointMangledName);
+
+    List<IRFunc*> irEntryPoints;
+    for (auto entryPointIndex : entryPointIndices) {
```
tfoleyNV, line 1452: Formatting: opening and closing `{}` should be in the same column.

source/slang/slang-options.cpp
```diff
@@ -143,12 +143,14 @@ struct OptionsParser
 
     SlangCompileFlags flags = 0;
 
+    //TODO(DG): Update this type?
```
tfoleyNV, line 146: Probably safe to remove this `TODO` comment now.

source/slang/slang-options.cpp
```diff
@@ -1421,9 +1424,6 @@ struct OptionsParser
         //
         for(auto& rawOutput : rawOutputs)
         {
-            // TODO(DG): As noted below, change support for output formats
-            // with multiple entry points
-
             // For now, all output formats need to be tightly bound to
```
tfoleyNV, line 1427: There's a comment here about how "down the road" we will need to allow output formats to support multiple entry points. That comment should probably be updated to reflect the new reality.

source/slang/slang-options.cpp
```diff
@@ -1459,14 +1459,21 @@ struct OptionsParser
...
+                    switch (outputFormat)
+                    {
+                    case CodeGenTarget::CPPSource:
+                        rawOutput.isWholeProgram = true;
+                        break;
+                    default:
+                        sink->diagnose(SourceLoc(), Diagnostics::cannotMatchOutputFileToEntryPoint, rawOutput.path);
+                        break;
+                    }
+                }
```
tfoleyNV, line 1476: This new logic probably deserves some commenting. Most importantly, we should probably be noting that the `switch` is here to check whether the given target format supports multiple entry points in the same output file, and we should expand the list of supported cases here as/when additional targets add such support.

PR #1438, opened 2020-07-10
CUDA/CPU varying compute inputs as IR pass

tfoleyNV:
The main change here is that the CPU and CUDA C++ emit paths now rely on an earlier IR pass to legalize the varying parameter list of a kernel and translate references to varying parameters with semantics like `SV_DispatchThreadID`. Doing so removes a lot of special-case logic from the emit passes.

This work moves us even closer to being able to eliminate `KernelContext` from the CPU/CUDA emit logic, because it removes the issue of state related to varying inputs being stored in `KernelContext`.

The new pass that handles the legalization is in `slang-ir-legalize-varying-params.cpp`, and it borrows heavily from the existing `slang-ir-glsl-legalize.cpp` pass. The new pass factors out the target-independent and target-dependent logic, so that both CPU and CUDA can share much of the same code despite having very different rules for how the system-value parameters are being provided.

An eventual goal is to have the new pass also handle the GLSL case, but doing so requires copying even more logic out of the GLSL-specific pass, and doing so seemed like a step to far for what was meant to be a stepping-stone change as part of other work. As a result of the incomplete nature of the pass, certain cases don't work for compute shader inputs for CPU/CUDA (e.g., wrapping your varying inputs in a `struct` type parameter), but those were cases that also didn't work in the existing `emit`-based logic.

One major consequence of this change is that the logic for emitting the various different functions that represent an entry point for our CPU back-end has been streamlined and simplified. The original logic had a fair bit of cleverness built in to try and avoid unnecessary math ops when computing the various IDs/indices, while the new logic is much more simplistic (the main dispatch function loops over threadgroups with a triply-nested `for` and then delegates to the  group-level function loops over threads with its own nested `for`s).

Longer term, it will be important to simplify the CPU functions we emit further, by eliminating things like the `_Thread` function that should never really be exposed to users (the minimum granularity of invoking a CPU compute kernel should be a single threadgroup). We may eventually decide to synthesize all of the extra code that is being generated in the `emit` pass as IR instead.


source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -0,0 +1,1321 @@
...
+                entryPointLayout->getResultLayout(),
+                LayoutResourceKind::VaryingOutput);
+
+            // Now that we have a representation of the value(s) that will
+            // be used to hold the entry-point result we need to transform
+            // any `returnVal(r)` instructions in the function body to
+            // instead assign `r` to `legalResult` and then `returnVoid`.
+            //
+            IRBuilder builder(m_sharedBuilder);
+            for( auto block : entryPointFunc->getBlocks() )
```
csyonghe, line 329: Does `getBlocks` function allows you to recursively traverse all blocks in a function? After reading the implementation of `IRInstList` I still don't see why that is the case.

PR #1439, opened 2020-07-11
Dynamic code gen for functions returning generic types.

csyonghe:
This change allows generating dynamic dispatch code for generic functions with generic return types.

The basic idea is that for every generic function that returns a generic value, e.g.
```
T SomeGenericFunc<T:IInterface>(int x);
```
we lower it into a normal function with additional `out T` parameter for the return value:
```
void SomeGenericFunc<T:IInterface>(out T retVal, int x);
```

This requires us to:
1. Translate the function type to reflect the correct signature. This affects `lowerFuncType`.
2. Transform the function body. This includes adding a new `RawPointer` typed parameter to the begining of the parameter list. `IRBlock` is extended with a `insertParamAtHead` to allow inserting new `IRParam`s to the begining of the list. The implementation just inserts an un-lowered `out T` type to the parameter list, and reuse existing parameter lowering logic to futhre lower that to a `RawPointer` typed parameter with type casts in the function body to convert it to `RTTIPointer`.
3. Translate the call sites. This is implemented in `lowerCall`. We need to identify all `call` insts that calls into affected function. This can be determined by checking the return type of the callee and the type of the `call` inst, if the return type of callee is `void` but the `call` inst's type is not `void`, it means the callee has been translated to expect the caller to allocate space for the return value. In this case, we allocate a temporary variable (again just emit a `IRVar` and reuse exiting logic to translate that `IRVar` into a `alloca` if necessary), and insert the pointer of the temporary variable to the begining of the `call` inst's argument list. Then we need to replace all uses of the `call` with a `load` of the temporary variable. The newly created `load` inst then needs to be inserted to the work list so it can processed further if the temporary variable is generic typed itself.


source/slang/slang-ir-lower-generics.cpp
```diff
@@ -124,6 +133,41 @@ namespace Slang
                 }
             }
             cloneInstDecorationsAndChildren(&cloneEnv, &sharedBuilderStorage, func, loweredFunc);
+            
+            // If the function returns a generic typed value, we need to turn it
+            // into an `out` parameter, since only the caller can allocate space
+            // for it.
+            auto oldFuncType = cast<IRFuncType>(func->getDataType());
+            if (isPolymorphicType(oldFuncType->getResultType()))
```
tfoleyNV, line 141: One question I have as this pass starts to grow is: should some of these steps be handled as their own distinct IR passes? E.g., should we have a stand-alone IR pass does the transformation for functions with generic return types, and thus allow the generics lowering pass to ignore this detail (because it would already have been handled upstream)?

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -382,110 +437,146 @@ namespace Slang
...
+            builder->setInsertBefore(callInst);
+
+            List<IRInst*> args;
+
+            // Indicates whether the caller should allocate space for return value.
+            // If the lowered callee returns void and this call inst has a type that is not void,
+            // then we are calling a transformed function that expects caller allocated return value
+            // as the first argument.
+            bool shouldCallerAllocateReturnValue = (funcType->getResultType()->op == kIROp_VoidType &&
+                callInst->getDataType() != funcType->getResultType());
```
tfoleyNV, line 498: This is more of the logic that could be pulled out into its own pass instead, right?

source/slang/slang-ir-insts.h
```diff
@@ -1956,6 +1956,8 @@ struct IRBuilder
         IRType* type);
     IRParam* emitParam(
         IRType* type);
+    IRParam* emitParamAtHead(
+        IRType* type);
```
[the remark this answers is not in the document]
csyonghe: OK.

source/slang/slang-ir-lower-generics.cpp
csyonghe: You are right. I may do it together with my associated types change.

PR #1440, opened 2020-07-14
Remove KernelContext wrapper from CPU/CUDA emit

tfoleyNV:
Currently, the CPU and CUDA C++ targets rely on a `KernelContext` type that is generated during emit, as a way to provide implicit access to things that were global in the input Slang code, but that can't actually be emitted as globals in the target language (because the semantics of global declarations differ).

For example, input like:

```hlsl
ConstantBuffer<Stuff> gStuff; // shader parameter
groupshared int gData[1024];  // thread-group shared variable
static int gCounter = 0;      // "thread-local" global-scope variable

void subroutine() { ... }

[shader("compute")] void computeMain() { ... }
```

would translate to output C++ for CPU a bit like:

```c++
struct KernelContext
{
    ConstantBuffer<Stuff> gStuff;
    int gData[1024];
    int gCounter = 0;

    void subroutine() { ... }

    void computeMain() { ... }
};
```

Note that both `computeMain()` and `subroutine()` are non-`static` members functions on `KernelContext`, so they have an implicit `this` parameter of type `KernelContext`, which allows the bodies of those functions to implicitly reference `gStuff`, etc. by name in their bodies.

Because `KernelContext::computeMain()` is a member function, we end up emitting an additional global-scope function to expose the entry point to the outside world, and that function is responsible for declaring a local `KernelContext` and invoking the generated entry point on it.

This approach has several important drawbacks:

* It complicates the emit logic for CPU and CUDA, with many special cases around when/how things get emitted

* It complicates the implementation of dynamic dispatch, because what seems like a function pointer in Slang IR needs to be a pointer-to-member-function in C++.

* It makes it difficult to have a non-kernel-oriented mode of compilation for CPU where a Slang function with a given signature gets output as a C++ CPU function with the "same" signature (not wrapped up as a member function of `KernelContext`.

This change makes a step toward addressing these issues by making the introducing of the `KernelContext` type be something that is done in an explicit IR pass instead of being handled as part of the last-mile emit logic.

The most important change is the removal of code related to `KernelContext` from the `slang-emit-{cpp,cuda}.{h,cpp}` files, with the equivalent logic instead being handled in a new pass in `slang-ir-explicit-global-context.{h,cpp}`. It should be noted that further cleanups to the emit logic should now be possible; in particular, both the CPU and CUDA emit paths are manually sequencing the `EmitAction`s instead of relying on the default logic, but at this point they should be able to just use the default. The additional cleanups are left for future work.

The explicit IR pass does more or less what one would expect: it identifies global-scope entities (global variables and parameters) that need to be wrapped and turns them into fields of a `KernelContext` type. It then modifies all entry points to initialize a `KernelContext` as part of their startup. Finally, any code that used to refer to the global entities is changed to refer to a field of the context, with the context passed via new function parameters (the new parameter is only added to functions that need it for now).

Transforming global variables into fields of a `KernelContext` type in the IR pass ends up dropping their initial-value expressions (since those were attached as basic blocks on the `IRGlobalVar`). To avoid breaking code that relies on global-scope (but thread-local) variables, this change also adds an explicit pass that takes the initialization logic on all global variables and moves it to explicit logic that runs at the start of every entry point in a linked module (`slang-ir-explicit-global-init.{h,cpp}`). This pass would also be useful when we get back to direct SPIR-V emit, since SPIR-V also requires initialization logic for globals to be emitted into entry points.

One complication that arises when the IR is introducing the types for entry-point parameters, global-scope parameters, and the `KernelContext` type is that it becomes harder for the emit logic to utter the names of those types (they might not even have names, since `IRNameHint`s might get stripped). This created a problem since the wrapper operations that were being generated for CPU were taking `void*` parameters and casting them to the appropriate type. To work around this issue, we have added an explicit IR pass (`slang-ir-entry-point-raw-ptr-params.{h,cpp}`) that transforms the signature of entry points so that any pointer parameters instead become raw pointer (`void*`) parameters, with the casting being handled inside the entry point itself.

One consequence of all the above changes is that for the CUDA target we no longer need a wrapper function to invoke the generated entry point any more, because the IR function for the entry point ends up having the correct/expected signature already. This is also the case for CPU when it comes to the `*_Thread` wrapper function, but this change doesn't try to eliminate the wrapper because of a belief that the `*_Thread`-level interface is going away anyway.

Because the IR is now responsible for ensuring the signature of the IR entry point for CUDA and CPU is what is expected, I needed to modify the `slang-ir-entry-point-uniforms` pass to always create an explicit parameter for the entry point uniforms when compiling for CUDA/CPU, even if there were no `uniform` parameters on the entry point as written. This also ended up requiring some tweaks to the parameter layout logic to ensure that CPU/CUDA targets always treat `ConstantBuffer<T>` as a `T*` even in the case where `T` is an empty `struct` type (which happens when we construct a `struct` type to represent the uniform parameters of an entry point with no uniform parameters...).

There are several future changes that can/should build on this work:

* We should change the generated signatures for CUDA kernels, so that they don't rely on `KernelContext` for global-scope parameters. At that point we can avoid generating a `KernelContext` at all for CUDA, except when a program uses global-scope thread-local variables.

* We should figure out how to make the "ABI" for dynamic-dispatch calls ensure that the kernel context is either always passed, or always *not* passed. Making a hard-and-fast rule as part of the calling convention for dynamic calls would ensure that they access through the context continues to work with dynamic calls (this change might break it in some cases).

* We should figure out how to handle the layout for the `KernelContext` in cases where a program is composed of multiple separately-compiled modules. Right now the layout of the `KernelContext` requires global knowledge (as does the pass that introduces explicit initialization for global-scope thread-locals).

* We should try to further clean up the CPU/CUDA C++ emit logic to fall back on the default emit behavior more, now that the various special-case approaches that were taken are no longer needed

### DIFF
--- a/source/slang/slang-emit-c-like.cpp
+++ b/source/slang/slang-emit-c-like.cpp
@@ -2341,16 +2341,20 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
 
     case kIROp_BitCast:
         {
-            // TODO: we can simplify the logic for arbitrary bitcasts
-            // by always bitcasting the source to a `uint*` type (if it
-            // isn't already) and then bitcasting that to the destination
-            // type (if it isn't already `uint*`.
+            // Note: we are currently emitting casts as plain old
+            // C-style casts, which may not always perform a bitcast.
             //
-            // For now we are assuming the source type is *already*
-            // a `uint*` type of the appropriate size.
-            //
-            //  auto fromType = extractBaseType(inst->getOperand(0)->getDataType());
+            // TODO: This operation should map to an intrinsic to be
+            // provided in a prelude for C/C++, so that the target
+            // can easily emit code for whatever the best possible
+            // bitcast is on the platform.
          
+            auto prec = getInfo(EmitOp::Prefix);
+            needClose = maybeEmitParens(outerPrec, prec);
+
+            m_writer->emit("(");
+            emitType(inst->getDataType());
+            m_writer->emit(")");
             m_writer->emit("(");
             emitOperand(inst->getOperand(0), getInfo(EmitOp::General));
             m_writer->emit(")");
--- a/source/slang/slang-ir-entry-point-raw-ptr-params.cpp
+++ b/source/slang/slang-ir-entry-point-raw-ptr-params.cpp
@@ -1,0 +1,121 @@
+// slang-ir-entry-point-raw-ptr-params.cpp
+#include "slang-ir-entry-point-raw-ptr-params.h"
+
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+
+// This pass transforms the entry points in a module
+// so that any entry-point parameters of pointer
+// type (or a pointer-like type like `ConstantBuffer<T>`)
+// are replaced with parameters of raw pointer (`void*`)
+// type, with a cast in teh function body used to
+// produce a value of the expected type.
+
+struct ConvertEntryPointPtrParamsToRawPtrsPass
+{
+    IRModule* m_module;
+
+    void processModule()
+    {
+        SharedIRBuilder sharedBuilder(m_module);
+        IRBuilder builder(&sharedBuilder);
+
+        // We start by getting and caching the raw pointer type.
+        //
+        auto rawPtrType = builder.getRawPointerType();
+
+        // Now we loop over global-scope instructions searching
+        // for any entry points.
+        //
+        for( auto inst : m_module->getGlobalInsts() )
+        {
+            auto func = as<IRFunc>(inst);
+            if(!func)
+                continue;
+
+            if( !func->findDecoration<IREntryPointDecoration>() )
+                continue;
+
+            // We can only modify entry points with definitions here.
+            //
+            auto firstBlock = func->getFirstBlock();
+            if(!firstBlock)
+                continue;
+
+            // Any code we introduce for casts will need to be inserted
+            // before the first ordinary instruction in the first block
+            // of the function (right after the parameters).
+            //
+            builder.setInsertBefore(firstBlock->getFirstOrdinaryInst());
+
+            // Note: because we are inserting code right after the parameters
+            // it doesn't work here to use `firstBlock->getParams()`, because
+            // that captures a begin/end range where the "end" is the 
+            // first ordinary instruction at the time of the call, which will
+            // chane when we insert code.
+            //
+            // TODO: We chould probably change the represnetation of ranges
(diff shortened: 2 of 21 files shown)


source/slang/slang-emit-c-like.cpp
```diff
@@ -2341,16 +2341,20 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
...
-            // For now we are assuming the source type is *already*
-            // a `uint*` type of the appropriate size.
-            //
-            //  auto fromType = extractBaseType(inst->getOperand(0)->getDataType());
+            // TODO: This operation should map to an intrinsic to be
+            // provided in a prelude for C/C++, so that the target
+            // can easily emit code for whatever the best possible
+            // bitcast is on the platform.
          
+            auto prec = getInfo(EmitOp::Prefix);
```
csyonghe, line 2352: I was also needing this for #1444. I added them to CPPSourceEmitter but it seems doing it here is better.

source/slang/slang-ir-entry-point-raw-ptr-params.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+
+// This pass transforms the entry points in a module
+// so that any entry-point parameters of pointer
+// type (or a pointer-like type like `ConstantBuffer<T>`)
+// are replaced with parameters of raw pointer (`void*`)
+// type, with a cast in teh function body used to
```
csyonghe, line 13: typo "teh"->"the"

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -0,0 +1,523 @@
...
+                        continue;
+
+                    m_entryPoints.add(func);
+                }
+                break;
+            }
+        }
+
+        // Now that we've capture all the relevant global entities from the IR,
+        // we can being to transform them in an appropriate order.
```
csyonghe, line 136: "being"->"begin"

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -0,0 +1,523 @@
...
+            //
+            auto call = as<IRCall>(use->getUser());
+            if(!call)
+                continue;
+
+            // We are going to construct a new call to `func`
+            // that has all of the arguments of the original call...
+            //
+            UInt originalArgCount = call->getArgCount();
+            List<IRInst*> args;
```
csyonghe, line 474: At some point we should add an overload of `emitCallInst` so we can start to use `ShortList`.

PR #1443, opened 2020-07-15
 IR pass to generate witness table wrappers.

csyonghe:
This change adds an IR pass to generate witness table wrappers, and remove the existing wrapp emitting logic from `CPPSourceEmitter`. This change is based on #1442.

Functions that are used to satisfy interface requirement have concrete type signatures for `this` and `associatedtype` parameters/return values. However, when they are called from a witness table, the callee only have a raw pointer for this arguments, since the conrete type is not known to the callee. Therefore, we need to generate wrappers for each member function
callable through a witness table, so that the wrapper functions take general `void*` for arguments whose type is unknown at call sites, and convert them to concrete types and calls the actual implementation.

Before this change, these wrapper functions are not explicitly created at IR level, instead `CPPSourceEmitter` includes some adhoc logic to emit these wrappers are part of witness table emission. With this change, all these wrapper related hackery code in `CPPSourceEmitter` has been cleaned up, as wrapper functions would just emit the same way as any other normal functions.


source/slang/slang-ir-witness-table-wrapper.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+            IRBuilder builderStorage;
+            auto builder = &builderStorage;
+            builder->sharedBuilder = &sharedContext->sharedBuilderStorage;
+            builder->setInsertBefore(func);
+
+            auto wrapperFunc = builder->createFunc();
+            wrapperFunc->setFullType((IRType*)interfaceRequirementVal);
+            if (auto name = _getWitnessTableWrapperFuncName(func))
+                builder->addNameHintDecoration(wrapperFunc, name);
```
tfoleyNV, line 46: Is it important for these wrappers to have distinguished names, or can we just leave them having the same name as the original, and then allow the emit pass to give them a unique suffix?

csyonghe: Not that important, this is just to make it easier to identify a wrapper function.

PR #1446, opened 2020-07-15
Remove debug build and conduct only quick tests on appveyor

csyonghe:
Since we are mostly relying on TeamCity for tests, we can remove debug build and full test runs on appveyor to speed things up.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,6 @@ platform:
 
 configuration:
   - Release
-  - Debug
 
 # In the interests of time, go ahead and immediately fail a build
 # if any job fails, rather than keep on building to discover the
@@ -55,18 +54,7 @@ build:
 
 test_script:
   - ps: |
-      if ($env:CONFIGURATION -eq "Debug")
-      {
-        $testCategory = "smoke"
-      }
-      elseif($env:PLATFORM -eq "x64")
-      {
-        $testCategory = "full"
-      }
-      else
-      {
-        $testCategory = "quick"
-      }
+      $testCategory = "quick"
       .\test.bat -platform %PLATFORM% -configuration %CONFIGURATION% -appveyor -category $testCategory
 
 # Package up the stuff we want to deploy into a single .zip file


jsmall-nvidia:
On the upside this makes what is one of the slowest parts of CI testing significantly faster. 12mins instead of 38 mins. Although this isn't perhaps a fair direct comparison, because it can take more the 30 mins to just start the tests.

As it stands on other CI, we use premake to make a project and then build it. We don't do this on appveyor - so appveyor provides a test for our VS project, which actually catches a fair few problems, most often of just forgetting to update the vs project. 

It could be argued that having appveyor being the *only* path that tests this is probably not a good idea, especially if it's slow. 

If we had a TC build/s that tested the visual studio projects, then appveyor would perhaps be less important. 

The one other thing the appveyor build has going for it, is it allows external parties to see breaking changes through CI. 

So perhaps instead of disabling theses tests on appveyor, we should leave it as it is. Add TC tests to test the VS projects, and then make merging not depend on appveyor tests passing. Then perhaps (assuming the TC VS builds are much faster than appveyor) we have close to the best of both worlds)?

csyonghe:
Do the problems you mentioned affects only builds? I cannot imagine the scenarios where you missed some files in vs projects and it still allows the solution to build but leads to incorrect binary.

jsmall-nvidia:
I mean yeah - if the project is broken somehow it's most likely going to not build any binaries and not run any tests. So an argument could be made that just running the minimal tests is enough. That in that scenario you would still want to build debug binaries too. 

The problem of external parties not being able to see failed tests and the reason for failed tests on visual studio though would remain. How much of a problem that is is arguable, but seems desirable at least in principal.

Thinking about it I suppose another tack to speed things up a little would be to drop 32bit x86 target.

csyonghe:
For your point on visual studio project build, we can have a test that runs premake and see if the visual studio files are changed. If any of them is changed, then that test fails, this should allow us to guarantee the projects are latest.

jsmall-nvidia:
Doing a comparison between VS projects and projects is a good idea. That unfortunately we can't even rely on that right now - because premake doesn't handle our natvis files correctly, and so the premake generated file doesn't match what's checked into the project. This would ideally be fixed though. 

For external devs, we'd still probably need to run appveyor tests across the whole suite. 

If we disabled requiring appveyor checks passing to allow merging, then the speed of appveyor slowness is not so much an issue.

csyonghe:
Closing this for now, since our TeamCity builds often have a longer wait time than Appveyor.

PR #1447, opened 2020-07-16
Run SSA pass to clean up temporary variables during generics lowering.

csyonghe:


tests/vkray/raygen.slang.glsl
```diff
@@ -1,58 +1,30 @@
...
-#define tmp_dot             _S23
-#define tmp_sat             _S24
-#define tmp_trace2_A        _S25
-#define tmp_trace2_B        _S26
-#define tmp_trace2_C        _S27
-#define tmp_trace2_D        _S28
-#define tmp_trace2_E        _S29
-#define tmp_trace2_ray      _S30
-#define tmp_trace2_payload  _S31
-#define tmp_storeIdx        _S32
```
tfoleyNV, line 41: I'm not happy to see all of this work go out the window. I'd set up this `.slang.glsl` file with the hope that it would be easy enough to change the Slang-generated temporary names that match these `#define`s in order to get the code back to working after a change to the GLSL emit logic.

For a larger test, it is always a bit worrying to just check in a new GLSL baseline to use for comparison, because we might not always take the time to carefully validate that GLSL against what it should be. Keeping the GLSL more or less the same, and only allowing changes to temporary variable names helps ensure that we don't introduce larger changes without realizing it.

If actually getting this back into the format it had before is out of the question, then I can live with checking in this new baseline, so long as it has been carefully compared against what we expect.

csyonghe: Fixed.

PR #1449, opened 2020-07-17
Run array specialization in a sperate pass.

csyonghe:
GLSL targets have a performance problem when calling a function with a parameter of large arrays.
For example
```
struct SkinningParams
{
    float4 quaternions[256];
};

ParameterBlock<SkinningParams> params;

float4 compute(SkinningParams p)
{
   use(p.quaternions);
}

float4 main()
{
    compute(params);
}
```

The downstream compiler doesn't seem to be able to optimize out the intermediates related to `params`, and leads to code that allocates buffers in thread-local memory. The solution is to specialize such calls to fetch the array directly from global variables, just like how we did for resource typed parameters in `specializeResourceParameters`:
```
struct SkinningParams
{
    float4 quaternions[256];
};

ParameterBlock<SkinningParams> params;

float4 compute()
{
   use(params.quaternions);
}

float4 main()
{
    compute_specialized(params);
}
```

Because the logic of lowering array parameters and resource parameters are the same, the previous implementation just directly modifies `specializeResourceParameters` to also include arrays. This has led to an unexpected bug when the array parameter cannot be specialized, as the pass will bail out and does not specialize the rest of the resource parameters that are necessary for the target.

This change address this issue by conducting array specialization, which is an optimization rather than a requirement to produce correct code, into a seprate pass. To be able to reuse all the existing `specializeResourceParameters` logic, this PR refactors this function into `specailizeFunctionCall`, with an additional callback to specify which types of parameters should be specialized. Then both `specializeResourceParameters` and `specializeArrayParameters` become a call to `specializeFunctionCall` with different callbacks.


source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -0,0 +1,830 @@
...
+    // Of course, now we need to back-fill the predicates that
+    // the above function used to evaluate prameters and arguments.
+
+    bool doesParamNeedSpecialization(IRParam* param)
+    {
+        return condition->doesParamNeedSpecialization(param);
+    }
+
+    bool isArgSuitableForSpecialization(IRInst* inArg)
+    {
```
tfoleyNV, line 187: I suppose it might eventually make sense to make the conditions under which an argument is "suitable for specialization" be something a derived pass can specialize.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -0,0 +1,830 @@
...
+        //
+        return anySpecializableParam;
+    }
+
+    // Of course, now we need to back-fill the predicates that
+    // the above function used to evaluate prameters and arguments.
+
+    bool doesParamNeedSpecialization(IRParam* param)
+    {
+        return condition->doesParamNeedSpecialization(param);
```
tfoleyNV, line 183: Was there a specific reason chosen for making the condition be a separate object rather than have this pass be an abstract base class that other passes can inherit from and then override this as a `virtual` function?

source/slang/slang-ir-specialize-resources.cpp
```diff
@@ -263,647 +70,16 @@ struct ResourceParameterSpecializationContext
...
 {
-    ResourceParameterSpecializationContext context;
-    context.compileRequest = compileRequest;
-    context.targetRequest = targetRequest;
-    context.module = module;
-
-    context.processModule();
+    ResourceParameterSpecializationCondition condition;
+    condition.targetRequest = targetRequest;
+    specializeFunctionCall(compileRequest, targetRequest, module, &condition);
```
tfoleyNV, line 82: I just noticed that we should probably name the function `specializeFunctionCalls` since it can (and does) specialize multiple function calls in one go.

source/slang/slang-ir-specialize-function-call.cpp
csyonghe: Yeah that is more extensible and we probably need to change to that form if needed in the future.

PR #1461, opened 2020-07-24
Fix bugs related to mutating implementations of interface methods

tfoleyNV:
There are two main bug fixes here:

* We were failing to diagnose when code calls a `[mutating]` method on a value that doesn't support mutation (that is an r-value instead of an l-value).

* We had a bug in the synthesis logic for interface requirements where we used the *result* type of the requirement in place of each of the *parameter* types.

The second bug made synthesis often produce incorrect signatures with `void` parameters.

The first bug meant that even though a `[mutating]` method should not be able to satisfy a non-`[mutating]` method (and we had code to enforce this for the "exact match" case), when we go on to try and synthesize a non-`[mutating]` method that satisfies the requirement by delegating to the user-written one, it would end up succeeding, because nothing was stopping a non-`[mutating]` method from calling a `[mutating]` one.

In each case this code adds a fix and a test case to confirm it.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -358,11 +358,54 @@ namespace Slang
         return true;
     }
 
-    bool SemanticsVisitor::TryCheckOverloadCandidateDirections(
-        OverloadResolveContext&		/*context*/,
-        OverloadCandidate const&	/*candidate*/)
+    bool isEffectivelyMutating(CallableDecl* decl)
     {
-        // TODO(tfoley): check `in` and `out` markers, as needed.
+        if(decl->hasModifier<MutatingAttribute>())
+            return true;
+
+        if(decl->hasModifier<NonmutatingAttribute>())
+            return false;
+
+        if(as<SetterDecl>(decl))
+            return true;
+
+        return false;
+    }
+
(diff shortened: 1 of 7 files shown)


source/slang/slang-check-overload.cpp
```diff
@@ -358,11 +358,54 @@ namespace Slang
...
+        // For now we will only use this step to check the
+        // mutability of the `this` parameter where necessary.
+        //
+        if(!isEffectivelyStatic(funcDeclRef.getDecl()))
+        {
+            if(isEffectivelyMutating(funcDeclRef.getDecl()))
+            {
+                if(context.baseExpr && !context.baseExpr->type.isLeftValue)
+                {
+                    if(context.mode == OverloadResolveContext::Mode::ForReal)
```
csyonghe, line 399: I wonder if it is possible to also provide additional diagnostic for unsatisfied interface requirement to point out specifically the reason, rather than just return `false` here and print out a general message.

tfoleyNV: Yes, we could use an approach like what we do for overload resolution where we keep track of one or more candidates that are “best” but not necessarily usable. When we fail to find a usable candidate we can take whatever the best candidates are and explain why they weren’t applicable.

PR #1464, opened 2020-07-28
Generalize lowerSimpleIntrinsicType to include generic arguments

csyonghe:
This change generalizes `lowerSimpleIntrinsicType` to lower all generic arguments of a `DeclRefType` into operands of the resulting IR type. By doing so, we are able to remove the special handling logic in `ValLoweringVisitor::visitPtrType`. In the future we can further cleanup the code handling `elementType`s of resource types.

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -1484,6 +1484,19 @@ struct ValLoweringVisitor : ValVisitor<ValLoweringVisitor, LoweredValInfo, Lower
         }
     }
 
+    // Lower substitution args and collect them into a list of IR operands.
+    void _collectSubstitutionArgs(List<IRInst*>& operands, Substitutions* subst)
+    {
+        if (!subst) return;
+        _collectSubstitutionArgs(operands, subst->outer);
+        if (auto genSubst = as<GenericSubstitution>(subst))
+        {
+            for (auto arg : genSubst->args)
+            {
+                operands.add(lowerVal(context, arg).val);
+            }
+        }
+    }
     // Lower a type where the type declaration being referenced is assumed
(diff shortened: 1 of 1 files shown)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -1498,7 +1492,24 @@ struct ValLoweringVisitor : ValVisitor<ValLoweringVisitor, LoweredValInfo, Lower
         auto intrinsicTypeModifier = type->declRef.getDecl()->findModifier<IntrinsicTypeModifier>();
         SLANG_ASSERT(intrinsicTypeModifier);
         IROp op = IROp(intrinsicTypeModifier->irOp);
-        return getBuilder()->getType(op);
+        List<IRInst*> operands;
+        // If there are any substitutions attached to the declRef,
+        // add them as operands of the IR type.
+        for (auto subst = type->declRef.substitutions.substitutions;
+            subst; subst = subst->outer)
```
tfoleyNV, line 1499: It shouldn't matter for any type we have today, but I think this would end up with a backwards result in the case of nested generics. If we have:

```hlsl
struct Outer<T>
{
    struct Inner<U>
    {
    }
}
```

then the iteration here would produce an IR instruction for `Outer<X>.Inner<Y>` that is of the form `InnerOp(Y, X)`. That is because the `substitutions` linked list for a `DeclRef` always starts with the inner-most substitutions and proceeds outward (so that it can share the outer substitutions with the parent declaration).

I suspect that the official Right Way to handle this would be to have a recursive function on the substitutions, that adds arguments for the outer substitution before adding arguments for the current one.

For now this really doesn't matter, since we have no cases of nested generic types that need to map to intrinsic operations in this fashion.

csyonghe: Done.

PR #1467, opened 2020-07-29
Binary for Heterogeneous Example

Checkmate50:
This PR introduces the ability to insert the binary of a non-CPU target
by using the -heterogeneous flag.  Specifically, this PR updates the
emitting logic to produce a variable of name `__[name_of_entryPoint]`
when the heterogeneous flag is present.


examples/heterogeneous-hello-world/main.cpp
```diff
@@ -63,103 +63,15 @@ struct gfx_DescriptorSet_0;
 struct gfx_PipelineState_0;
 
 bool executeComputation_0();
+extern unsigned char __computeMain[];
+extern size_t __computeMainSize;
```
tfoleyNV, line 67: These names don't include the format of the generated code (e.g., `dxbc`), so there would be a conflict if the user specified two or more GPU target formats at once (`-target dxbc -target spirv`). It would be nice to clean that up in a follow-on.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -1416,6 +1416,12 @@ struct ValLoweringVisitor : ValVisitor<ValLoweringVisitor, LoweredValInfo, Lower
             resultType);
     }
 
+    IRType* visitPtrType(PtrType* type)
+    {
+        IRType* valueType = lowerType(context, type->getValueType());
+        return getBuilder()->getPtrType(valueType);
+    }
+
```
tfoleyNV, line 1424: I *think* Yong added catch-all logic such that this diff is no longer needed.

PR #1468, opened 2020-07-30
Fix issues arising around DXR 1.1 RayQuery usage

tfoleyNV:
This change includes a few different fixes for issues that arose in a user shader that made use of DXR 1.1.

The existing solution we had for handling the DXR 1.1 `RayQuery` type relied on the fact that a declaration like:

```hlsl
RayQuery<0> myRayQuery;
```

Looks like an undefined variable to existing Slang, while to dxc it is a variable declaration that runs an implicit default constructor (sneaking a bit of C++ into HLSL, but only in a way the standard library can use).

Slang was getting away with the fact that this maps to an undefined variable because it turns out that our emit logic would output the exact same declaration for an undefined value (since declaring a variable without initializing it is the simplest way to get an undefined value of a given type in a C-like language).

The main bug that arose here was that if the `RayQuery<...>`-typed variable was declared under control flow, then the `undefined` instructions introduced by our SSA pass would actually get inserted into the wrong block. Basically, when a block was trying to read a variable, and there was no preceding `store` to that variable in the block, we'd start looking for incoming values from its predecessor block(s). In the case where the variable *never* gets stored to, this search would eventually reach the first block of the function, where we'd realize the value must be `undefined`. The result was that we might insert an `undefined` instruction of some `T` into the first block of a function, but the type `T` might be the result of a lookup operation performed later in that function. This ends up creating a use of `T` that isn't dominated by the definition, which violates the SSA property. This violation of the SSA properties lead to us generating incorrect code in a later pass that deals with scoping differences between SSA form and our structured output statements; that code would end up creating a local variable to hold a *type* instead of a value.

The main fix is in `slang-ir-ssa.cpp`, where we catch the case of trying to read a variable in the block that declares it, if there we no preceding `store`s. We simply insert an `undefined` instruction before the first such read and write that out as the value of the variable to be used for subsequent instructions (up to the next `store`). This fixes the SSA dominance property for the `undefined` values that get introduced and thus technically fixes the output code for the user shader.

A secondary issue is that it is kind of gross to be relying on the behavior of `undefined` instructions in the IR for the semantics of an important standard library type like `RayQuery<...>`. A preceding change already added basic support for Slang to run default initializers (declared as `__init()`) on variables that are declared without an initial-value expression. This change adds such a default initializer to `RayQuery<...>` and maps it to a dedicated IR instruction that is intended to represent the idea of running a C++-style default constructor to produce a value. It turns out that the code we need to emit in that cse is identical to what we currently emit for `undefined` instructions, so that is helpful.

A tertiary issue is that when trying to run the user shader in debug mode, I ran into an assertion because our type layout logic for reflection had never dealt with the issue of user-defined `enum` types being used in constant buffers or other memory that needs layout. I added a quick fix that lays out any `enum` types as their "tag type" (which defaults to `int`).

Unfortunately, there is no easy way to check in a regression test for the user issue, because official `dxcompiler` versions with support for DXR 1.1 are not yet released (at least as of last time I checked).


source/slang/slang-ir-ssa.cpp
```diff
@@ -776,6 +776,19 @@ IRInst* readVar(
...
 
+    if( blockInfo->block == var->parent )
+    {
+        // If this is the block that actually *introduces*
+        // the variable, then there is no reason to keep
+        // searching, because its value cannot have been
+        // established in a predecessor block.
+        //
+        auto type = var->getDataType()->getValueType();
+        val = blockInfo->builder.emitUndefined(type);
```
csyonghe, line 787: How does this ensure the `undefined` is emitted after the type inst, if that type inst is defined in this same block?

tfoleyNV: For this bug, it doesn’t matter if it is actually emitted in the right order, so long as the *block* if is emitted into is valid to use.

I *think* the logic here in the SSA pass sets the insertion point of the builder to be right before the instruction that reads the value, so the ordering would be fine anyway. If that isn’t the case now, we can fix it to be the case if/when it becomes relevant.

PR #1470, opened 2020-07-30
First pass support for Sampler Feedback

jsmall-nvidia:
This PR adds preliminary support for [Sampler Feedback](https://microsoft.github.io/DirectX-Specs/d3d/SamplerFeedback.html) 

* Added SLANG_TEXTURE_FEEDBACK_FLAG to identify flags that are texture feedback for reflection
* Added SLANG_RESOURCE_ACCESS_WRITE as the texture feedback resource is 'write only'
* Added the SAMPLER_FEEDBACK_MIP_REGION_USED and SAMPLER_FEEDBACK_MIN_MIP HLSL types
* Added sampler-feedback-basic.slang test
* Support around dumping/reflection/reflection testing.

Note that the intrinsics are described in hlsl explicitly as in 

```
__target_intrinsic(hlsl, "($0).WriteSamplerFeedbackBias($1, $2, $3, $4)")
```

This is because the methods on feedback textures, ended up being emitted as a function call taking two generic parameters and not a method call. An improved version of this would fix this issue, and not require this work around - and could just use

```
__target_intrinsic(hlsl)
```


slang.h
```diff
@@ -1841,7 +1842,11 @@ extern "C"
         SLANG_RESOURCE_UNKNOWN              = 0x08,
         SLANG_ACCELERATION_STRUCTURE        = 0x09,
 
-        SLANG_RESOURCE_EXT_SHAPE_MASK       = 0xF0,
+        //SLANG_RESOURCE_EXT_SHAPE_MASK       = 0xF0,
+
+        SLANG_TEXTURE_FLAG_MASK             = 0xF0,
+
+        SLANG_TEXTURE_FEEDBACK_FLAG         = 0x10,
```
tfoleyNV, line 1849: Hmm... at the very least it seems like we need to keep the old `SLANG_RESOURCE_EXT_SHAPE_MASK` enumeration so that existing source can continue working.

I'm a bit confused about the rename, though. Whether or not a texture is a feedback texture isn't really part of its "shape," but once we've committed to storing the relevant bit inside `SlangResourceShape` we are kind of stuck with pretending like it is, right?

The idea of "shape" here was that you had the "base shape" (2D, 3D, etc.) and then the "extended shape" included any modifiers that might lead to it requiring new texture coordinate values on lookups (which both array-ness and multisample-ness do).

Does feedback-ness of a texture need to be exposed through reflection? If it does, does it need to be exposed as part of the "shape" of a texture?

source/slang/slang-ast-type.h
```diff
@@ -107,6 +107,19 @@ class BuiltinType : public DeclRefType
     SLANG_ABSTRACT_CLASS(BuiltinType)
 };
 
+class FeedbackType : public BuiltinType
+{
+    SLANG_CLASS(FeedbackType)
+
+    enum class Kind : uint8_t
```
tfoleyNV, line 114: I'm always a bit uneasy using "kind" in a way that doesn't really refer to the "type of a type."

AFAICT, in this case the `Kind` enumeration is used to tell which of two concrete types we have, such that the information in `kind` is redundant with the `DeclRef` that the type refers to. Is it possible to just remove this field entirely?

source/slang/slang-emit-cpp.cpp
```diff
@@ -340,21 +340,31 @@ SlangResult CPPSourceEmitter::_calcCPPTextureTypeName(IRTextureTypeBase* texType
...
-        case TextureFlavor::Shape::Shape2D:		outName << "Texture2D";		break;
-        case TextureFlavor::Shape::Shape3D:		outName << "Texture3D";		break;
-        case TextureFlavor::Shape::ShapeCube:	outName << "TextureCube";	break;
-        case TextureFlavor::Shape::ShapeBuffer: outName << "Buffer";         break;
-        default:
-            SLANG_DIAGNOSE_UNEXPECTED(getSink(), SourceLoc(), "unhandled resource shape");
-            return SLANG_FAIL;
+        SLANG_ASSERT(texType->GetBaseShape() == TextureFlavor::Shape::Shape2D);
+        outName << "FeedbackTexture2D";
+    }
```
tfoleyNV, line 354: Why not just emit `Feedback` as a prefix based on the flag in the shape? We already use a suffix to handle `Array` and `MS` from the shape.

source/slang/slang-ast-type.h
tfoleyNV: Having looked through the rest of the PR, I'm not seeing this field being used, so I'd be inclined to remove it for now. I know that we might want to expose the difference here via the reflection API, but I'd rather not do so in a way that relies on storing extra information in the AST nodes that reference these types.

(The way that texture/resource types use the special extra "payload" data field for magic types is a gross hack that I am not happy with at all. I'd like to see us trend in the direction of using it less rather than using it for more types.)

slang.h
tfoleyNV: I think I'm convinced that this solution is the best one available right now, but I'd encourage *not* renaming the existing mask enumeration (both returning it to the original name *and* removing the new name).

We probably need something cleaner for exposing aspects of resource types that aren't conceptually part of the "shape," but that seems best to do later as part of an overhaul of this logic (e.g., we are probably close to the point where squeezing all this into one byte turns out to have been a short-sighted optimization on my part).

source/slang/slang-emit-cpp.cpp
jsmall-nvidia: Ok changed that.

slang.h
jsmall-nvidia: I reinstated SLANG_RESOURCE_EXT_SHAPE_MASK, and see there is argument for it remaining if just for compatibility.  I changed it because it's name doesn't seem to match it's use.

PR #1477, opened 2020-08-04
`AnyValue` based dynamic dispatch code gen

csyonghe:
This is the second part of the change to use fixed sized `AnyValue` types to replace raw pointers in dynamic functions following #1469 

The detailed design reasoning is documented in #1469, here are the details related with this PR:
1. Added `IRPackAnyValue` and `IRUnpackAnyValue` instructions.
2. Implemented the pack and unpack instructions in C++ prelude with `memcpy`s. This is temporary and will be replaced by calls to actual synthesized per-type packing and unpacking functions in a future change.
3. Inject `pack` and `unpack` at necessary places around function calls and witness table wrapper functions.
4. Remove `lowerGenericVar` pass, since it can now be replaced by simply replacing all polymorphic types with `AnyValue` types.
5. Change `IRInterfaceType` lowering logic to create a new interface type instead of doing in-place modifications.
6. Add several diagnostic message for invalid uses around anyValueSize attribute.
7. Updated test cases to use the `[anyValueSize]` attribute.


prelude/slang-cpp-types.h
```diff
@@ -41,6 +41,26 @@ struct Array
     size_t count;
 };
 
+template<size_t N>
+struct AnyValue
+{
+    uint8_t data[N];
```
tfoleyNV, line 47: I was going to ask about alignment constraints (what if the user wants to store a `double` in here?) but I then realized that we don't care about alignment for these buffers at all: they are only accessed by copying in/out, rather than by casting the memory to the desired type.

Some of this policy stuff really needs to be handled by adding comments to the declaration of this type and the pack/unpack operations.

source/slang/slang-diagnostic-defs.h
```diff
@@ -482,6 +483,12 @@ DIAGNOSTIC(40007, Internal, irValidationFailed, "IR validation failed: $0")
 
 DIAGNOSTIC(40008, Error, invalidLValueForRefParameter, "the form of this l-value argument is not valid for a `ref` parameter")
 
+DIAGNOSTIC(40009, Error, dynamicInterfaceLacksAnyValueSizeAttribute, "interface '$0' is being used in dynamic dispatch code but has no [anyValueSize] attribute defined.")
+DIAGNOSTIC(40010, Note, seeInterfaceUsage, "see usage of interface '$0'.")
+
+DIAGNOSTIC(40011, Error, unconstrainedGenericParameterNotAllowedInDynamicFunction, "unconstrained generic paramter '$0' is not allowed in a dynamic function.")
```
tfoleyNV, line 489: I assume this error could be removed if/when we allow `[anyValueSize(...)]` to be applied to a generic type parameter.

prelude/slang-cpp-types.h
csyonghe: This is supposed to be temporary code. All these `AnyValue` related stuff in prelude will go away when I have the packing pass in place.

PR #1478, opened 2020-08-05
Refactor enumerating directory contents

jsmall-nvidia:
Motivation for this change was the valgrind found resource leakages. 

The os.cpp/.h contain a platform specific implementation of iterating over the contents of a directory. The implementation created a type that could be iterated over. The original refactor emulated it's behavior but fixed up the leaks, as well as handling copies. Whilst this ended up simpler than the original, the code was still rather complicated. 

The solution presented here does not provide a mechanism for outside code to iterate over contents whilst platform specific state is held in scope. Instead we have a windows/linux function that just calls a visitor for each matching file/directory. This code was moved to Path, which already had platform specific code.  The os.cpp file then became directory-util, which just uses a simple implementation of the visitor to get the results that is used by the test framework. Which just outputs a list of String paths. This is slightly more clumsy at usage point, but is significantly simpler in implementation.

The other fix here was to make the linux implementation handle pattern matching. 

A small improvement is the Result mechanism has a simply way of indicating that the directory wasn't found.


source/core/slang-io.h
```diff
@@ -25,8 +25,39 @@ namespace Slang
...
+                File = TypeFlags(1) << int(Type::File),
+                Directory = TypeFlags(1) << int(Type::Directory),
+            };
+        };
+
+        class Visitor
+        {
+        public:
+            virtual void accept(Type type, const UnownedStringSlice& filename) = 0;
+        };
```
tfoleyNV, line 51: A simple callback function that takes a "user data" pointer would seem to be simpler here, and can easily be wrapped with a template variant that adapts a lambda, allowing the use site to be very similar to a loop.

PR #1479, opened 2020-08-05
Emit spir-v using MemoryArena to stop memory leak

jsmall-nvidia:
The motivation for this fix was to remove memory leaks found via valgrind.

This also includes a fix for uninitialized variable `alwaysCreateCollectedParam` that valgrind reports.

* Use MemoryArena to store SPIR-V instructions and operands
* Small improvements around writing out of UnownedStringSlice

Allocating the SpvInst on the MemoryArena is trivial, the problem is operands, and that an instruction might need to be constructed whilst creating the current instruction. This is trivial with a List<SpvWord>, because the result can just be added at the end. This isn't directly possible with the arena, because it is necessary to know how much space is needed. The other problem is that an operand addition may not in general take a single operand word - with a string, it can consume an arbitrary amount of words. So we can't just have something line 

```
SpvWord operands[]  = { getOperand(a), getOperand(b), getOperand(c) }; 
setOperands(dst, operands, SLANG_COUNT_OF(operands));
```

Another way of approaching this would be to make the operands allocation a little like a list with a capacity, and then allocate new space as needed. That is not what is done here, because unlike List/malloc, with the MemoryArena when we reallocate, we just consume more memory - the previous memory remains allocated.

Here we use a stack on SpvWord that holds the operands list as they are begin constructed. Only one can be constructed at at time, so if we are part way constructing instruction A, that references B and B needs operands, they can just be placed on top of the stack and released, before returning to As construction.

This mechanism for stack operands is handled here with a RAII type InstConstructScope, which manages the stack and on destruction moves the tail of operands from the m_operandStack into the instruction itself. 

That using this mechanism the construction of instructions is nested, and each nested insts construction typically must be completed (ie it's operands determined) before the parents can complete. This may be a construction limitation that is somewhat limiting - but if some construction cannot be made that way, then this is still possible. For example the instruction can be constructed without operands, and the operands added/allocated on the arena at some later point.


source/slang/slang-emit-spirv.cpp
```diff
@@ -579,14 +636,14 @@ struct SPIRVEmitContext
         IRInst* irInst = nullptr;
     };
 
-        /// Emit operand words for all the operands of a given IR instruction
-    void emitOperand(SpvInst* dst, OperandsOf const& other)
+        /// Emit operand words for all the operands of the current IR instruction
```
tfoleyNV, line 639: The change to the comment here is incorrect.

The code change means that the operand words are emitted *to* the implicit destination instruction instead of the one given explicitly.

But the operands that get emitted are still those *of* the IR instruction that is given as `other`, and that is what the comment is referring to.

source/slang/slang-ir-entry-point-uniforms.h
```diff
@@ -9,7 +9,9 @@ struct IRModule;
 
 struct CollectEntryPointUniformParamsOptions
 {
-    bool alwaysCreateCollectedParam;
+    // TODO(JS): Not sure if it makes sense to initialize to true or false. Go with false as
+    // seems to fit usage.
+    bool alwaysCreateCollectedParam = false;
```
tfoleyNV, line 14: Yes, false is a reasonable default here.

PR #1482, opened 2020-08-07
GPU Foreach Parsing and Checking Retry

Checkmate50:
This PR introduces parsing and semantic checking for a GPU foreach loop
for heterogeneouis programming. A GPU foreach loop takes the form:
```
__GPU_FOREACH(renderer, gridDims, LAMBDA(uint3 dispatchThreadID) {
        kernelCall(args, ...); });
```

And will allow the host code to call into a kernel with the correct
renderer and grid dimensions. This commit also introduces a hack to
unify types in the heterogeneous hello world file, which will hopefully
be amended in the future.


source/slang/slang-ast-stmt.h
```diff
@@ -117,6 +117,17 @@ class DefaultStmt : public CaseStmtBase
     SLANG_CLASS(DefaultStmt)
 };
 
+// a `default` statement inside a `switch`
```
tfoleyNV, line 120: The comment here is inaccurate. This AST node is unique enough that it probably deserves a decent comment.

PR #1483, opened 2020-08-07
Improvements to Casting

jsmall-nvidia:
The motivation for this change was that 

```
struct Outer
{
    enum Channel
    {
        Red,
        Green,
        Blue,
        Alpha,
    }
}

void something()
{
    Outer::Channel a = (Outer::Channel)1;
}
```

Would not compile. The problem is that the cast is determined previously if it contains a single identifier and if that identifier is a type. The cast we have here is not a single identifier.

There are several methods used to disambiguate

* A more sophisticated test of if the expression in brackets specifies a type
* A heuristic that looks at tokens after the ) to determine if its a cast
* If the prior heuristic doesn't work - a whitespace heuristic. 

The last heuristic is needed because the looking at tokens after ) can't disambiguate all situations for example for + - (or & and * if we allow pointers). 

The whitespace heuristic means that `(..)+a` or `(..) + a` are determined to say + is a binary op, and if it's a binary op, then it cannot be a cast.

 A type lookup disambiguate clearly if a type is found, but not if it's not, because it could be an out-of-order declaration that happens later in the file. 

Note there may be an inconsistency around how the heuristic test is applied. Currently it is only on a 'complicated' (ie non single id), that is followed by + or - that is a StaticMemberExpr uses the heuristic. That we may want to apply it in the simple case too. We also may want to expand when it's applied outside StaticMemberExpr. That is left for a future PR.


source/slang/slang-parser.cpp
```diff
@@ -4310,6 +4361,130 @@ namespace Slang
...
+                return true;
+            }
+            case TokenType::LParent:
+            {
+                // If we are followed by ( it might not be a cast - it could be a method invocation.
+                //
+                // It's hard to know what a heuristic would look like here to try and disambiguate (as we do for + and -)
+                // so this *only* works if the type is defined before the cast.
+                
+                return (_tryResolveTypeDecl(parser, expr) != nullptr);
```
tfoleyNV, line 4424: Note that in the case where there is ambiguity between a cast and a function/method call, it is always safe to default to the call, because "calling" a type is already supported (the semantic checker can already handle `int(7)` so there is no reason it can't handle `(int)(7)` as the same thing.

source/slang/slang-parser.cpp
```diff
@@ -4310,6 +4361,130 @@ namespace Slang
...
+                    // Skip the + or -
+                    advanceToken(parser);
+                    // Peek the next token to see if it was preceded by white space
+                    const Token nextToken = peekToken(parser);
+
+                    // Rewind
+                    parser->tokenReader.setCursor(cursor);
+
+                    // If there isn't any whitespace prior, we assume unary, and that means it must be a cast
+                    return ((nextToken.flags & TokenFlag::AfterWhitespace) == 0);
```
tfoleyNV, line 4477: I believe that the correct heuristic for identifying whether an operator is binary or unary should be something like:

* Whitespace before, whitespace after -> binary
* No whitespace before, no whitespace after -> binary
* Otherwise -> unary

This basically breaks things down as:

```hlsl
a + b // assume binary
a+b // assume binary
a+ b // assume unary (looks like a postfix op)
a +b // assume unary (looks like a prefix op)
```

jsmall-nvidia: Changed to this style.

source/slang/slang-parser.cpp
jsmall-nvidia: Ok - that makes that a little simpler. Change in PR.

PR #1488, opened 2020-08-11
Bugfix: WaveActiveCountBits on glsl output.

jsmall-nvidia:
The problem here is that `bitCount(subgroupBallot($0)` isnt quite right for the intrinsic for GLSL. That subgroupBallot returns uint4. The fix is a little convoluted, because I don't want to implement something super complicated in an 'intrinsic' definition - I'd prefer to use slang. The fix here was to add the function _WaveActiveBallot which exposes the return type as uint4, and then have a special BitCounts function `_WaveCountBit` that attempts to limit the amount of work based on warp/lane width. 

This is perhaps not an ideal solution. That a more elegant answer would just operate on WaveMask, and have a bit count that operated on that. Then the default impl could be more general, where here it's quite specific to this scenario. 

More broadly this appears to show issues around how to write target specific code within the standard library, that it would be nice if there could be (say) more than one implementation, so we can have a target specific implementation say.

This PR also contains a fix for the issue of glslang outputting an unable to inline warning. The problem was in the default optimization, the MergeReturnPass needed to be executed before the inline pass.


source/slang/hlsl.meta.slang
```diff
@@ -2568,12 +2568,24 @@ __target_intrinsic(cuda, "__ballot_sync($0, $1)")
 __target_intrinsic(hlsl, "WaveActiveBallot($1)")
 WaveMask WaveMaskBallot(WaveMask mask, bool condition);
 
+// Predeclare this function
+uint _WaveCountBits(uint4 value);
```
tfoleyNV, line 2572: Forward declarations should not be required in Slang.

source/slang/hlsl.meta.slang
```diff
@@ -2568,12 +2568,24 @@ __target_intrinsic(cuda, "__ballot_sync($0, $1)")
...
+
+// Ideally we'll do this with WaveMask backing type changing in the future,
+// but for the moment it's stuck on uint. So this is a work around, such that we can make
+// the code Slang code.
 __glsl_extension(GL_KHR_shader_subgroup_ballot)
 __spirv_version(1.3)
-__target_intrinsic(glsl, "bitCount(subgroupBallot($1))")
+__target_intrinsic(glsl, "subgroupBallot($0)")
+uint4 _WaveActiveBallot(bool condition);
```
tfoleyNV, line 2580: Why can't this just be/use the existing `WaveActiveBallot`? Is there a reason why an underscore-prefixed version is needed?

source/slang/hlsl.meta.slang
```diff
@@ -3292,6 +3301,25 @@ bool WaveIsFirstLane()
...
 
+// It's useful to have a wave uint4 version of countbits, because some wave functions return uint4.
+// This implementation tries to limit the amount of work required by the actual lane count.
+uint _WaveCountBits(uint4 value)
+{
+    // Assume since WaveGetLaneCount should be known at compile time, the branches will boil
+    // away
+    uint waveLaneCount = WaveGetLaneCount();
+    
+    switch (waveLaneCount / 32)
```
tfoleyNV, line 3312: I almost asked if this should be `((waveLaneCount + 31) / 32)`, but then I saw that your `case`s are set up to match this style.

Given this approach, though, it seems like you *do* want to change to `((waveLaneCount - 1) / 32)`, because otherwise the case where you have a 32-lane wave would map to `case 1` and add up the bits from both `.x` and `.y`.

There is also the argument to be made that if the downstream compiler can optimize for `WaveGetLaneCount()` returning a constant, then they can probably also optimize for `WaveActiveBallot()` and similar operations always returning zero in the `.y`, `.z`, and `.w` components...

tfoleyNV:
> More broadly this appears to show issues around how to write target specific code within the standard library, that it would be nice if there could be (say) more than one implementation, so we can have a target specific implementation say.

I'm not 100% sure what the limitation is that drives this comment. We can already express target-specific function implementations in the stdlib using `__specialized_for_target`:

```hlsl
__target_intrinsic(cpu, "1")
int magicFunction()
{
    return 2;
}

__specialized_for_target(glsl)
int magicFunction()
{
    int i = 0;
    i++; i++; i++;
    return i;
}
```

The above code defines a stdlib function called `magicFunction()`. In the default case, `magicFunction()` lowers to an ordinary IR function that returns 2. On CPU targets, it lowers as an *intrinsic* function, that happens to map to the string `"1"` in output. For GLSL targets, it lowers as an ordinary function, but uses a distinct definition that is customized to based on what GLSL (seemingly) needs, and it effectively computes `3` (although it does so using ordinary/arbitrary Slang code).

The catch that seems to keep coming up is that the rules for how stdlib functions need to be declared are not clear to everybody. So far I haven't really run into too many cases where they flat out can't express what people want.

source/slang/hlsl.meta.slang
jsmall-nvidia: Good catch, fixed in PR.

jsmall-nvidia:
Ok didn't realize '__specialized_for_target' was a thing, that'll help in the future.

PR #1492, opened 2020-08-12
IR support for Tuple types.

csyonghe:
This change adds IR support for tuples. Tuples is a useful tool for implementing dynamic dispatch code for existential types, and can also be used to simplify our existing type legalization pass.

Specifically, this PR adds a `IRTupleType` and two related instructions: `IRMakeTuple` and `IRGetTupleElement`.

The change also includes an IR pass that lowers `IRTupleType`s into ordinary `struct` types.

For example,`IRTupleType(T0, T1, T2)` lowers into
```
struct Tuple
{
    T0 value0;
    T1 value1;
    T2 value2
};
```
`IRMakeTuple(v0, v1, v2)` lowers into:
```
var t: ptr(Tuple);
addr0 = getFieldAddr(t, StructKey("value0"));
store(addr0, v0);
addr1 = getFieldAddr(t, StructKey("value1"));
store(addr1, v01);
addr2 = getFieldAddr(t, StructKey("value2"));
store(addr2, v2);
return load(t);
```

`IRGetTupleElement(t, 0)` lowers into:
```
extractField(t, StructKey("value0")
```

To ensure correctness, the `lower-tuple-types` pass explictly deduplicates types, so two equavalent `IRTupleType`s will lower into the same `struct` type.

Deduplication is implemented in `SharedIRBuilder::deduplicateAndRebuildGlobalValueMapping()`.


source/slang/slang-ir-deduplicate.cpp
```diff
@@ -0,0 +1,80 @@
...
+        {
+            IRConstantKey key = { value };
+            if (auto newValue = builder->constantMap.TryGetValue(key))
+                return *newValue;
+            builder->constantMap[key] = value;
+            return value;
+        }
+        IRInst* addTypeValue(IRInst* value)
+        {
+            // Do not deduplicate struct types.
```
tfoleyNV, line 27: This same exception should apply to interface types, right?

I feel like we've made these same distinctions in other passes, and we really need a way to identify structural-vs-nominal types that is shared across parts of the IR.

source/slang/slang-ir-inst-defs.h
```diff
@@ -235,6 +236,8 @@ INST(makeVector, makeVector, 0, 0)
 INST(MakeMatrix, makeMatrix, 0, 0)
 INST(makeArray, makeArray, 0, 0)
 INST(makeStruct, makeStruct, 0, 0)
+INST(MakeTuple, makeTuple, 0, 0)
+INST(GetTupleElement, getTupleElement, 2, 0)
```
tfoleyNV, line 240: Can `getTupleElement` just use the existing `getElement` (or is it `extractElement`?) instruction? I guess the reason to have a distinct case is that this operation is index-based (like array `getElement` and unlike struct ` getField`), but the type of the result depends on the index/key (like struct `getField` but unlike array `getElement`).

source/slang/slang-ir-lower-tuple-types.cpp
```diff
@@ -0,0 +1,197 @@
...
+                nameSb << "_";
+                appendTypeName(nameSb, elementType);
+                auto key = builder->createStructKey();
+                fieldNameSb.Clear();
+                fieldNameSb << "value" << i;
+                builder->addNameHintDecoration(key, fieldNameSb.getUnownedSlice());
+                auto field = builder->createStructField(structType, key, (IRType*)elementType);
+                info->fields.add(field);
+            }
+            builder->addNameHintDecoration(structType, nameSb.getUnownedSlice());
```
tfoleyNV, line 74: While I appreciate the work that has been put into giving the generated tuple types decent names, it might be easier and more consistent with the direction of the IR to give them all the same name hint of `Tuple` and then let the emit logic turn those into `Tuple_0`, `Tuple_1`, etc.

source/slang/slang-ir-lower-tuple-types.cpp
```diff
@@ -0,0 +1,197 @@
...
+            {
+                SLANG_ASSERT(i < (Index)inst->getOperandCount());
+                auto ptrType = builder->getPtrType(info->fields[i]->getFieldType());
+                auto addr = builder->emitFieldAddress(ptrType, var, info->fields[i]->getKey());
+                builder->emitStore(addr, inst->getOperand((UInt)i));
+            }
+            auto load = builder->emitLoad(var);
+            inst->replaceUsesWith(load);
+            inst->removeAndDeallocate();
+        }
```
tfoleyNV, line 115: I'm a bit confused about why the `makeTuple` operation doesn't translate into a `makeStruct` instruction?

Is the answer that we don't have `makeStruct` (possible), or that there would be a concern over the order of operands (since if we *do* have a `makeStruct` I suspect it doesn't include the field keys as operands)?

Generating a new read/write temporary instead of an SSA value puts a burden on subsequent passes to undo the memory load/store operations before further optimizations become possible.

csyonghe: Yeah we do have a `makeStruct`, I should probably use it.

source/slang/slang-ir-inst-defs.h
csyonghe: This makes it easier for the lowering pass to identify the instruction. Otherwise it needs to also examine the type of the first operand of the `getElement` inst to tell if it is dealing with a tuple or not.

source/slang/slang-ir-deduplicate.cpp
csyonghe: I was not quite sure on interfaces, as it seems fine to deduplicate equivalent interface types?

csyonghe: We don't deduplicate `struct`s because its fields are defined as children, not as operands. But `interface` types are using operands so they can deduplicate correctly.

source/slang/slang-ir-inst-defs.h
tfoleyNV: Sounds good.

source/slang/slang-ir-deduplicate.cpp
tfoleyNV: Well, we could switch the handling of `struct` types around to use operands now, but we'd still want to avoid applying deduplication to them.

I guess nothing would go drastically wrong right now if we deduplicate `interface` types, but I think it is wrong on principle, and we'd probably want to avoid potential problems down the line. E.g., if we start emitting RTTI into output binaries, then we'd want distinct interfaces to get distinct RTTI objects, even if they are structurally identical. This would be especially important if we starting allowing RTTI queries to check for things like the types that implement an interface, or the `extension`s that apply to one.

source/slang/slang-ir-lower-tuple-types.cpp
tfoleyNV: The code you have right now works, so I wouldn't push hard to change it unless we run into issues.

source/slang/slang-ir-lower-tuple-types.cpp
csyonghe: Done.

PR #1494, opened 2020-08-12
Support property declarations in interfaces

tfoleyNV:
There are two main features in this change. First, we allow for `interface`s to declare `property` requirements, which can be satisfied by matching `property` declarations in a type that conforms to the interface:

    interface IRectangle
    {
        property float width { get; }
        property float height { get; }
    }

    struct Square : IRectangle
    {
        float size;
        property float width { get { return size; } }
        property float height { get { return size; } }
    }

Second, we allow a type to satisfy a `property` requirement with an ordinary field of the same name:

    struct Rectangle : IRectangle
    {
        float width;
        float height;
        // no explicit `property` declarations needed
    }

The implementation of these features is mostly in `slang-check-decl.cpp` in the logic for checking conformance of a type to an interface.

The first feature simply requires adding logic to checking whether a candidate satisfying `property` declaration matches a required `property` declaration. To do so, it must have the same type, and an accessor to satisfy each of the required accessors.

The second feature requires adding logic to synthesize an AST `property` declaration for a type, based on a required `property` declaration and its accessors. This means that, more or less, any type where `this.name` yields a storage location that does what is needed can satisfy a property requirement (there is no specific rule that says the storage needs to be a field, although that is the most likely case).

The way that witnesses are stored for property declarations probably merits some description. During IR lowering, an abstract storage declaration like a subscript or `property` more or less desugars away, so that the actual interface requirements correspond to the accessors within it (the `get`, `set`, etc.). This means that a witness table should have entries/keys corresponding to the accessors and not the property itself. The process of finding/recording witnesses for `property` requirements thus installs entries for the individual accessors (with care taken to only install accessor witnesses once we are sure we have witnesses for all the requirements). Currently, the code also installs an entry for the property itself, although that is not strictly required, and might not be something we continue to do long-term.

(Aside: it was somewhat surprising that an end-to-end test of `property` declarations in `interface`s Just Worked without any changes to IR lowering.)

As we continue to write more code that synthesizes and checks AST expressions/statements, it becomes necessary to refactor the semantic checking logic so that it splits the recursive part (e.g., checking the operands of an assignment) from the validation part (e.g., checking that the assignment itself is valid). It is probably too big of a change to justify at this point, but it might be valuable in the future to have distinct hierarchies that represent unchecked and checked ASTs, with semantic checking mostly being a transformation from one to the other. The benefit of such a change is we could factor out a distinct "builder" API for constructing validated/checked AST nodes, with both semantic checking and AST synthesis being clients of that API.


source/slang/slang-check-decl.cpp
```diff
@@ -1758,6 +1874,348 @@ namespace Slang
...
+        // Leaving out the name here doesn't help matters, becaues then
+        // *all* synthesized members on a given type would share the same
+        // mangled name.
+        //
+        synPropertyDecl->nameAndLoc = requiredMemberDeclRef.getDecl()->nameAndLoc;
+
+        // The type of our synthesized property will be the expected type
+        // of the interface requirement.
+        //
+        // TODO: This logic can/will run into problems if the type is,
```
csyonghe, line 1947: Is there an assumption that synthesized code will not get type checked again? This seems to be one place that you'll want to check synthesized code for simplicity.

source/slang/slang-check-decl.cpp
```diff
@@ -1758,6 +1874,348 @@ namespace Slang
...
+            //
+            // Note: in practice we expect that only `set` accessors will have any parameters,
+            // and they will only have a single parameter.
+            //
+            List<Expr*> synArgs;
+            for( auto requiredParamDeclRef : getParameters(requiredAccessorDeclRef) )
+            {
+                auto paramType = getType(m_astBuilder, requiredParamDeclRef);
+
+                // The synthesized parameter will ahve the same name and
```
csyonghe, line 1988: "have"

source/slang/slang-check-decl.cpp
tfoleyNV: Yes, this logic runs late enough that we shouldn’t have problems with things getting re-checked.

PR #1496, opened 2020-08-13
GPU Foreach Loop

Checkmate50:
This PR introduces the completed GPU foreach loop and updates the
heterogeneous-hello-world example to use it.  This PR builds on the
previous introduction of the GPU Foreach loop parsing and semantic
checking PR (#1482) by introducing IR lowering and emitting.  The new
feature can be used by having a GPU_Foreach loop interacting with a
named non-CPP entry point, and using the -heterogeneous flag.


Checkmate50:
Somehow the submodules got messed up again, so trying this from the top

PR #1500, opened 2020-08-15
Support initializing an existential value with a generic value in dynamic code gen.

csyonghe:
This change implements support for initializing an existential value (function argument or local variable) with a generic value in dynamic code gen.

For example,
```
[anyValueSize(16)]
interface IInterface
{
	int Compute(int inVal);
};

int GenericCompute0(IInterface obj, int inVal)
{
    return obj.Compute(inVal);
}

int GenericCompute1<T:IInterface>(T obj, int inVal)
{
    IInterface iobj = obj;
	return iobj.Compute(inVal) + 
        GenericCompute0(obj, inVal) -
        GenericCompute0(iobj, inVal);
}
```
In `GenericCompute1` function, a generic value `obj` is being assigned to an existential value `iobj`, and is being used to invoke the `GenericCompute0` function that accepts an existential typed parameter.

Conceptually, the current existential type support should just work. However, there is a major unimplemented piece: when initializing an existential value, we need to create a tuple that includes a packed `AnyValue`, a witness table, and an RTTI object. While the former two is trivially available at the site of an `IRMakeExistential`, the RTTI object is not immediately available since the concrete type of the generic value is passed in as a different argument. In the existing implementation, a generic function parameter `T obj` is lowered into `AnyValue<N> obj`, and therefore the link to the RTTI parameter (which comes as a lowered parameter of `T` itself) is lost during the translation. To address this issue, this change introduces an `IRAnyValueTypeWithRTTI` IR type that behaves the same as `IRAnyValue`, but with an additional operand that refers to the associated RTTI object. With this change, `obj` in function `GenericCompute1` will have type `IRAnyValueTypeWithRTTI(16, rtti_T)`, where `rtti_T` represents the type info parameter for `T` in the lowered function.

The next issue is that `IRAnyValueTypeWithRTTI` cannot appear in a function parameter list due to a limitation in our IR that prevents parameter lists to have internal depedencies (i.e. a parameter has a type that is dependent on another parameter in the same parameter list). The solution is to still use `IRAnyValueType` in the parameter list, but insert a `BitCast` that converts the parameter into `IRAnyValueTypeWithRTTI` and have all the uses refer to the bitCast instead. 

Note that the bitcast instruction will then have type `Ptr(IRAnyValueTypeWithRTTI(16, rtti_T))`. Since this type depends on `rtti_T`, which is another parameter of the function, the type must be defined in the local function scope instead of the global scope. However, after translating all `IRAnyValueType` and `IRAnyValueTypeWithRTTI` into a specialized `AnyValue` struct type in `anyvalue-marshalling` pass, we will leave behind a type definition like `Ptr(AnyValueStruct)` in local scope. Since our emitting logic will have problems dealing with local type definitions, this change also introduces an IR pass `hoistLocalTypes` to hoist these types back to global scope to avoid problems.


source/slang/slang-ir.h
```diff
@@ -1265,15 +1265,30 @@ struct IRRTTIType : IRType
...
+};
+
+struct IRAnyValueTypeWithRTTI : IRAnyValueTypeBase
+{
+    IR_LEAF_ISA(AnyValueTypeWithRTTI);
+    IRInst* getRTTI()
+    {
+        return getOperand(1);
+    }
+};
```
tfoleyNV, line 1290: The main question I have on this PR is: why have a pass that transforms `AnyValueType` to `AnyValueTypeWithRTTI` instead of just transforming the `makeExistential` operations themselves?

The current encoding of `makeExistential(v, w)` relies on the fact that the type that needs to go into the existential can be read off of the `v` operand, and thus we don't bother making it an explicit operand.

An alternative encoding would be to have `makeExistential(v, T, w)` take the type as an explicit operand, and then a pass could replace the type operand `T` with an RTTI operand instead.

Alternatively, we could also have `makeExistentialWithRTTI(v, RTTI, w)` and have the pass transform `makeExistential` to `makeExistentialWithRTTI`.

I'm just a bit confused about why `AnyValue` needs to sacrifice simplicity and clarity for this feature, when it seems like the design issue might just be on `makeExistential`.

csyonghe: That's true. Let me redo this PR. Thanks for pointing it out!

PR #1508, opened 2020-08-19
Allow calling a generic function with an existential value (dynamic dispatch)

csyonghe:
This change allows calling a generic function with an existential value when generating dynamic dispatch code.
It enables the following use case:
```
[anyValueSize(16)]
interface IInterface
{
	int Compute(int inVal);
};

int GenericCompute0(IInterface obj, int inVal)
{
    return GenericCompute1(obj, obj, inVal);
}

int GenericCompute1<T:IInterface>(T obj, IInterface obj2, int inVal)
{
    return obj.Compute(inVal) + obj2.Compute(inVal);
}

```
Upon seeing the call to `GenericCompute1` from `GenericCompute0`, our current implementation would attempt to specialize `GenericCompute1` with `T=IInterface`, which is not correct. In this case, the right behavior is to "open" the existential value `obj` first, and use the `ExtractExistentialType` obtained from opening the existential value to specialized the generic function.

This change implements the right behavior by adding a call to `maybeOpenExistential` on every argument at the beginning of `SemanticsVisitor::ResolveInvoke`.  The result expr type from `openExistential` is a `DeclRefType` to the interface decl, with a `ThisTypeSubstitution` referring to an `ExtractExistentialSubtypeWitness` and results in the following generic argument inference to fail. This PR changes `openExistential` so that it returns an expression with `ExtractExistentialType`, and the `ThisTypeSubstitution` is available from the new `interfaceDeclRef` member of the `ExtractExistentialType`. This allows the rest of resolving logic to ignore `ThisTypeSubstitutions`. A new case is added to `lookupMemberInSuperType` to handle looking up through `ExtractExistentialType`s. In addition, a case for `ExtractExistentialType` is added to `_isDeclaredSubtype` and `TryJoinTypeWithInterface` to get our type checking logic recognize the new use case. 

This change can introduce unnecessary `openExistential` operations. For example, the `obj` argument in `GenericCompute0` can be used directly to fullfil the `obj2` parameter of `GenericCompute1`. However, with this change, since all arguments are opened conditionlessly, it needs to be packed into another existential value before serving to the `obj2` parameter. This is redundent code that can be eliminated in an optimization pass (not included in this PR).

This change doesn't enable the compiler to generate statically specialized code for the introduced test case. I haven't investigated why it doesn't work yet.


source/slang/slang-check-constraint.cpp
```diff
@@ -158,6 +158,11 @@ namespace Slang
                 return bestType;
         }
 
+        if (auto extractExistentialType = as<ExtractExistentialType>(type))
+        {
+            if (extractExistentialType->interfaceDeclRef.equals(interfaceDeclRef))
+                return extractExistentialType;
+        }
```
tfoleyNV, line 165: Hmm... I would have expected this to be a check that `type` is a `DeclRefType` referring to an `InterfaceDecl` and with a `ThisTypeSubstitution`, and then it would `return` the underlying type from the `ThisTypeSubstitution`.

source/slang/slang-check-overload.cpp
```diff
@@ -1517,6 +1517,25 @@ namespace Slang
...
+            {
+                arg->type->substitute(getASTBuilder(), declRefType->declRef.substitutions);
+                if (auto interfaceRef = declRefType->declRef.as<InterfaceDecl>())
+                {
+                    if (auto thisTypeSubst = findThisTypeSubstitution(
+                            declRefType->declRef.substitutions, interfaceRef.getDecl()))
+                    {
+                        arg->type = thisTypeSubst->witness->sub;
+                    }
+                }
```
tfoleyNV, line 1535: Basically, this is the logic I expected to see in the constraint inference part of things.

Here in the argument-related stuff I'm not 100% sure it is valid to make this change, because we are overwriting the type of an `Expr` to something other than its original type. Maybe it happens to work out all right in practice, but this makes me nervous, because a type of `IFoo` with a `ThisType` of `Bar` is distinct from the type `Bar` in our system today (although maybe that is actually a broken concept in our representation), and here you are replacing the first type with the latter.

source/slang/slang-check-conformance.cpp
```diff
@@ -268,6 +268,24 @@ namespace Slang
...
+                    auto witness = getASTBuilder()->create<ExtractExistentialSubtypeWitness>();
+                    *outWitness = witness;
+                    witness->declRef = extractExistentialType->declRef;
+                    witness->sub = extractExistentialType;
+                    witness->sup = DeclRefType::create(getASTBuilder(), superTypeDeclRef);
+                }
+                return true;
+            }
+            return false;
+        }
```
tfoleyNV, line 288: I like this bit, since it is a clear typing rule that we should support.

tfoleyNV (dismissed):

PR #1512, opened 2020-08-21
NVAPI improvements

jsmall-nvidia:
* Detect NVAPI ops for features on D3D11/12
* Made Renderer::Desc contain the 'nvapiExtnSlot'. If -1 then nvapi is not enabled.
* Fixed bug around the setting of the NVAPI slot used.


tools/gfx/open-gl/render-gl.cpp
```diff
@@ -717,9 +717,12 @@ SlangResult GLRenderer::initialize(const Desc& desc, void* inWindowHandle)
...
     }
 
-    if (m_desc.requiredFeatures.indexOf("nvapi") >= 0 && SLANG_SUCCEEDED(NVAPIUtil::initialize()))
+    if (m_desc.nvapiExtnSlot >= 0)
     {
-        m_features.add("nvapi");
+        if (SLANG_FAILED(NVAPIUtil::initialize()))
+        {
+            return SLANG_E_NOT_AVAILABLE;
+        }
```
tfoleyNV, line 725: I may have missed this before, but why are we initializing NVAPI in our OpenGL back-end? I assume there are no NVAPI features for GL, because those would all be GL extensions instead...

PR #1514, opened 2020-08-24
Export witness table and RTTI objects in compiled libraries.

csyonghe:
This PR includes a series of changes to expose witness tables and RTTI objects in the compiled DLL, and provide a reflection function to get the mangled name of a type witness. These changes allows the host to fill in an existential value.

Detailed changes:

- Ensure that witness tables and RTTI objects are defined with the `extern "C"` modifier in the generated C++ code.
- RTTI objects use the mangled name of the type directly, so that can be queried using the type's mangled name directly from the resulting DLL.
- Expose `Linkage::getTypeConformanceWitnessMangledName` to return the mangled name of witness tables to the host.
- Ensure that all witness tables (including those for associated types) have proper mangled name.

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -5170,7 +5170,11 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
                         // Need to construct a sub-witness-table
                         auto irWitnessTableBaseType = lowerType(subContext, astReqWitnessTable->baseType);
                         irSatisfyingWitnessTable = subBuilder->createWitnessTable(irWitnessTableBaseType);
-
+                        auto mangledName = getMangledNameForConformanceWitness(
+                            subContext->astBuilder,
+                            astReqWitnessTable->witnessedType,
+                            astReqWitnessTable->baseType);
+                        subBuilder->addExportDecoration(irSatisfyingWitnessTable, mangledName.getUnownedSlice());
                         // Recursively lower the sub-table.
                         lowerWitnessTable(
                             subContext,
--- a/source/slang/slang.cpp
+++ b/source/slang/slang.cpp
@@ -623,6 +623,17 @@ SLANG_NO_THROW slang::TypeLayoutReflection* SLANG_MCALL Linkage::getTypeLayout(
     return asExternal(typeLayout);
 }
 
+SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::getTypeConformanceWitnessMangledName(
+    slang::TypeReflection* type, slang::TypeReflection* interfaceType, ISlangBlob** outNameBlob)
+{
+    auto subType = asInternal(type);
+    auto supType = asInternal(interfaceType);
+    auto name = getMangledNameForConformanceWitness(subType->getASTBuilder(), subType, supType);
+    Slang::ComPtr<ISlangBlob> blob = Slang::StringUtil::createStringBlob(name);
+    *outNameBlob = blob.detach();
+    return SLANG_OK;
+}
+
 SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::createCompileRequest(
     SlangCompileRequest**   outCompileRequest)
 {
--- a/slang.h
+++ b/slang.h
@@ -3102,6 +3102,13 @@ namespace slang
             LayoutRules     rules = LayoutRules::Default,
             ISlangBlob**    outDiagnostics = nullptr) = 0;
 
+            /** Get the mangled name for a type witness.
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getTypeConformanceWitnessMangledName(
+            TypeReflection* type,
+            TypeReflection* interfaceType,
+            ISlangBlob** outNameBlob) = 0;
+
             /** Create a request to load/compile front-end code.
             */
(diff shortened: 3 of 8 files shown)


source/slang/slang.cpp
```diff
@@ -623,6 +623,17 @@ SLANG_NO_THROW slang::TypeLayoutReflection* SLANG_MCALL Linkage::getTypeLayout(
...
 
+SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::getTypeConformanceWitnessMangledName(
+    slang::TypeReflection* type, slang::TypeReflection* interfaceType, ISlangBlob** outNameBlob)
+{
+    auto subType = asInternal(type);
+    auto supType = asInternal(interfaceType);
+    auto name = getMangledNameForConformanceWitness(subType->getASTBuilder(), subType, supType);
+    Slang::ComPtr<ISlangBlob> blob = Slang::StringUtil::createStringBlob(name);
+    *outNameBlob = blob.detach();
+    return SLANG_OK;
```
tfoleyNV, line 634: Wait a second... This logic isn't actually validating that the given type actually does conform to the interface. It is just trying to return what the mangled name for the conformance *would be* if it exists, and leaves it as a linker error if/when the conformance doesn't actually exist.

I'm worried that this isn't the right way to go about things.

Conceptually, the Right Way to handle this would be to call into the AST/semantics level with a `doesTypeConformToInterface` or `getSubtypeWitness` or whatever the operation is called, to either produce a `SubtypeWitness*` value, or `nullptr`. If a subtype witness can be found, then we can get its mangled name directly and we are happy/done (although at that point we can also just return the `SubtypeWitness*` as a reified object that represents the conformance.

Such a lookup needs to be done with specific modules (and their `ExtensionDecl`s) in scope, so it would need to be an operation on the `ISession` (`Linkage`).

If we want to have an API that eventually interacts with linking, then the user needs a way to say "my compiled program needs to include the code for the conformance of `MyType` to `IFoo`." It may be that the conformance is *not* in the module that defines `MyType`, nor the module that defines `IFoo`, but is instead in some third module `M`.

Just because we can form the mangled name that represents a `MyType : IFoo` conformance, that does not do anything to help us identify the module `M`, and when it comes time to link together the code for our program, nothing about `MyType`, `IFoo`, or that mangled name will help ensure we link in the code from `M`.

What the user needs is a way to get a handle to the `MyType : IFoo` conformance (which is a `SubtypeWitness`) and to specify that the conformance represented by that witness had better be linked in. Inside the compiler, that witness eventually refers to an `InheritanceDecl`, and we simply need to ensure that the given `InheritanceDecl` (and the module it comes from) show up in the linked code.

I'm coming away convinced that this API *needs* to be written in terms of an interaction with the compiler front-end to do proper semantic checking of the conformance (which will yield a `SubtypeWitness`).

I'm fine with this as a short-term hack, but I don't think it should be the basis for our longer-term API for this feature.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5170,7 +5170,11 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
                         // Need to construct a sub-witness-table
                         auto irWitnessTableBaseType = lowerType(subContext, astReqWitnessTable->baseType);
                         irSatisfyingWitnessTable = subBuilder->createWitnessTable(irWitnessTableBaseType);
-
+                        auto mangledName = getMangledNameForConformanceWitness(
+                            subContext->astBuilder,
+                            astReqWitnessTable->witnessedType,
+                            astReqWitnessTable->baseType);
```
tfoleyNV, line 5176: This isn't relevant to this PR, but we actually have a problem in Slang today in terms of how we represent the mangled names of conformance witnesses. Right now we generate the mangled name for a witness in a way that doesn't care where it came from; all that matters is the base type and the derived type.

That's actually technically incorrect, and can cause problems if two modules introduce the same conformance via `extension`s.

If the module that declares the conformance is the same as the one that declared the concrete type, or the interface, the current naming is reasonable.

If the module that declares the conformance is distinct from the ones that declare the concrete type and the interface, we really need the mangled name to include the module name that introduced the conformance. We could always apply this rule if we want to keep things simple.

The problem with including the module name, of course, is that it breaks the logic added to the public API here, where we assume that the mangled name for a conformance only depends on the types involved.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5170,7 +5170,11 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+                        subBuilder->addExportDecoration(irSatisfyingWitnessTable, mangledName.getUnownedSlice());
```
tfoleyNV, line 5177: The second issue I'm pointing out here is more closely related to a statement in the PR description: it is mentioned that this change makes it so that even the witness tables for associated types get mangled names (and this logic seems to be the part that is doing that...).

I don't think that is correct/necessary.

If the only way a witness table is referenced is as an entry in another witness table, then there is no need for the child witness table to have an `[export(...)]` decoration and linkage.

I suspect that the logic here in IR lowering should be backed out, and we should retain the current rules for when declarations get an `[export(...)]` decoration as-is.

Instead, the emit logic should pay attention to whether a global-scope entity (like a witness table) has linkage, and use that to decide between emitting it as `extern "C"` with the mangled name, or as a `static` with a synthesized name).

I don't like the idea of forcing things to have linkage that don't actually need it, just to simplify our code generation.

PR #1518, opened 2020-08-26
Enable lower-generics pass universally.

csyonghe:
This change renames the `-allow-dynamic-code` compiler flag to `-disable-specialization`, and make `lower-generics` pass always-run.

In theory, the IR specialization pass that runs before `lower-generics` will replace all existential values and generic values with specialized value if there is enough info at compile time for the specialization. For code that is fully specialization, there will be no more generic or existential values after specialization, so the subsequent lower-generics pass will be a no-opt. In the case of existential values whose type is unknown, or specializing a generic function with dynamic type info, the specialization will leave the existing generics code unmodified, and therefore they will be picked up by lower-generics pass and lowered into ordinary functions that takes `IRAnyValue`, `IRRTTIObject` and `IRWitnessTable` parameters.

The main piece of work is to correctly identify and exclude builtin interfaces and intrinsic generic functions from the `lower-generics` pass. This is done by 1) ignoring calls to functions with an `IRTargetIntrinsicsDecoration`; 2) Mark all builtin interfaces (e.g. `__BuiltinArithmeticType`) with an `[builtin]` attribute that translates to an `IRBuiltinDecoration`, and do not lower the builtin interfaces types or emit witness table wrapper functions for them.

One thing that needs future attention is how to do with the `stripWitnessTable` pass. This change make it so that it never runs for CPU/CUDA targets. We may need a more systematic approach to run it when we can determine there is no dynamic dispatch code in the IR.


tangent-vector:
IIRC, the main reason we have the `stripWitnessTables` pass is because of messes related to the stdlib around:

* There are stdlib target-intrinsics functions that are generic over `<T : IBuiltinWhatever>` and calls to those functions can appear in generated IR, which means they reference the witness table for `T : IBuiltinWhatever`

* The `T : IBuiltinWhatever` witness table will mean that certain declarations are kep alive in the IR, even when they aren't needed/wanted, and in some cases those declarations were causing problems when emitted to HLSL/GLSL.

The easiest solution in a case like that was to completely eliminate all witness tables, since at the time they weren't needed in final codegen. Once the `T : IBuiltinWhatever` witness table was removed, any operations it referenced what weren't otherwise used could be DCE'd.

A simple short-term fix would be to make it so that we don't have any stdlib builtin operations that are generic based on an interface constraint (by having a non-intrinsic wrapper function for `<T : IBuiltinWhatever>` that calls an "unsafe" intrinsic function that is generic on `<T>` without constraints).

The longer-term fix involves solving the root cause, and one catch there is that I fear we may need to introduce the notion of some interface requirements (or conformances) being target-specific, so that we don't emit them on targets where they can't/shouldn't be used.

tfoleyNV:
Is this PR ready to review? It seems to still be running into failures on the tests.

csyonghe:
I am still working on it, seems like there are issues with the SamplerFeedback interface and related intrinsic functions since those are the only cases where they are multi-level generics and the target_intrinsic decoration is somehow missing.

tangent-vector:
There's some really gross stuff we do to try and find the `__target_intrinsic` decoration on things, since I think it ends up being placed on the `IRFunc` at the deepest level, so there are places in the code where if we have an `IRGeneric` we attempt to find the thing that the generic would return. I have fixed *some* places in the code to deal with there being a possibility of multiple levels of nested generics, but I would not be surprised at all if there are still some that haven't been fixed.

PR #1524, opened 2020-08-28
AST Serialization in Modules

jsmall-nvidia:
* Added ability to filter what ASTSerialization serializes out 
  + In particular replaces Decls outside of this module with ImportExternalDecl
  + Doesn't serialize Stmt
* Added ImportExternalDecl, which contains mangled name of something referenced out 
* Fixed some problems/improved AST dump
* Added ability to serialize AST data into RIFF container
* Made serialization read/write ir with associated AST data

Note:

*  That no attempt is made to replace ImportExternalDecl with the actual decl currently.
* That the association between IRModule/ASTModule is implicit in ordering in the RiffList. Perhaps needs module name or other information to be stored.
* Currently nothing is done with the deserialized AST data.


source/slang/slang-ast-serialize.cpp
```diff
@@ -46,6 +49,60 @@ static void _toNativeValue(ASTSerialReader* reader, const SERIAL_TYPE& src, NATI
...
+            return writer->writePointer(ptr);
+        }
+    }
+
+    // TODO(JS): What we really want to do here is to ignore bodies functions.
+    // It's not 100% clear if this is even right though - for example does type inference
+    // imply the body is needed to say infer a return type?
+    // Also not clear if statements in other scenarios (if there are others) might need to be kept.
+    //
+    // For now we just ignore all stmts
```
tfoleyNV, line 93: Given our current setup, I believe it is valid to skip all `Stmt`s when serializing, because they only affect the "body" of various declarations.

Right now we don't infer things like the return type of a function from its body, but even if we did I would expect AST serialization to be done *after* the inference has run and the result type has been recorded on the function declaration itself.

The main sticking point I anticipate for serialization right now comes from `Expr`s instead of `Stmt`s.. An expression can end up in the AST in a context that isn't just a function "body" in two ways:

1. A type can in principle include arbitrary expression as a generic argument (`vector<float, 1 + foo(N)>`). Right now those are basked down to `Val`s, and we don't support `Val`s that are full expressions, so this is a non-issue today just because of other limitations in the compiler.

2. A function can include a default-value expression for one or more of its parameters, and these are represented as a bare `Expr` today.

Item (2) is the main reason that I expect we need to have serialization of `Expr`s in the near term, because lowering a *call* to a function in another module currently requires being able to emit the value of its default arguments as needed. Something more clever is possible, but we'd have to change how we do code generation for default-value expressions.

tfoleyNV:
It seems like the next important step will be to wire up those `ImportExternalDecl`s to the actual `Decl`s coming from other modules during the load process.

PR #1525, opened 2020-08-31
Support dynamic existential shader parameters in render-test

csyonghe:
This change adds support for specifying dynamic existential shader parameter values in render-test.

Specifically:
1. allow two new types of element in a `data=[]` block: `rtti(typename`) and `witness(c, I)`, so this is now valid:
    ```
    //TEST_INPUT:cbuffer(data=[rtti(MyImpl) witness(MyImpl, IInterface) 1 0], stride=4):
    ```
2. Add a new `spCompileRequest_getSession` to get a `ISession` handle from the legacy end-to-end compile request object. This allows the test infra to call `Linkage::getTypeConformanceMangledName`.
3. Add `CPUComputeUtil::populateRTTIEntries` that is run as a setup step, which fills in actual rtti pointer values in the data buffer of the provided `shaderInputLayout`.

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -891,6 +891,7 @@ extern "C"
             @return The function pointer related to the name or nullptr if not found 
             */
         virtual SLANG_NO_THROW SlangFuncPtr SLANG_MCALL findFuncByName(char const* name) = 0;
+        virtual SLANG_NO_THROW void* SLANG_MCALL findObjectByName(char const* name) = 0;
     };
     #define SLANG_UUID_ISlangSharedLibrary { 0x9c9d5bc5, 0xeb61, 0x496f,{ 0x80, 0xd7, 0xd1, 0x47, 0xc4, 0xa2, 0x37, 0x30 } };
 
--- a/slang.h
+++ b/slang.h
@@ -3374,6 +3374,11 @@ SLANG_API SlangResult spCompileRequest_getModule(
     SlangInt                translationUnitIndex,
     slang::IModule**        outModule);
 
+/** Get the `ISession` handle behind the `SlangCompileRequest`.
+*/
+SLANG_API SlangResult spCompileRequest_getSession(
+    SlangCompileRequest* request,
+    slang::ISession** outSession);
 #endif
 
 /* DEPRECATED DEFINITIONS
--- a/slang.h
+++ b/slang.h
@@ -892,7 +892,7 @@ extern "C"
             */
         inline SlangFuncPtr SLANG_MCALL findFuncByName(char const* name)
         {
-            return static_cast<SlangFuncPtr>(findSymbolAddressByName(name));
+            return reinterpret_cast<SlangFuncPtr>(findSymbolAddressByName(name));
         }
             /** Get a symbol by name. If the library is unloaded will only return nullptr. 
             @param name The name of the symbol 
--- a/slang.h
+++ b/slang.h
@@ -890,8 +890,15 @@ extern "C"
             @param name The name of the function 
             @return The function pointer related to the name or nullptr if not found 
             */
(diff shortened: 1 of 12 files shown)


slang.h
```diff
@@ -3374,6 +3375,11 @@ SLANG_API SlangResult spCompileRequest_getModule(
     SlangInt                translationUnitIndex,
     slang::IModule**        outModule);
 
+/** Get the `ISession` handle behind the `SlangCompileRequest`.
+*/
+SLANG_API SlangResult spCompileRequest_getSession(
+    SlangCompileRequest* request,
+    slang::ISession** outSession);
```
tfoleyNV, line 3382: I don't mind having this, but it should also be unnecessary because a client that is using the newer COM-like API would be calling `ISession::createCompileRequest`, so that there would be no question about which session a compile request belongs to.

slang.h
```diff
@@ -891,6 +891,7 @@ extern "C"
             @return The function pointer related to the name or nullptr if not found 
             */
         virtual SLANG_NO_THROW SlangFuncPtr SLANG_MCALL findFuncByName(char const* name) = 0;
+        virtual SLANG_NO_THROW void* SLANG_MCALL findObjectByName(char const* name) = 0;
```
tfoleyNV, line 894: There's a couple issues I want to note here.

The first issue is that `ISlangSharedLibrary` is not just an interface that Slang vends out to applications, but it is also an interface that application code can use when passing things into Slang. I believe we have APIs that let the user say "here is the implementation of dxcompiler I want you to use" and the user passes in an `ISlangSharedLibrary`. Simply adding new requirements to interfaces that get passed in the app->Slang direction is a binary-incompatible change, so we need to be careful about it.

The second issue is that the naming for `findObjectByName` makes it a little tricky for the user to figure out what it does. Is it related to "objects" in the C++ sense? Object files? The lack of documentation here doesn't help, but I assume the intention was to piggyback on the documentation for `findFuncByName`, since they really represent the same operation.

That last point brings me to what I *think* is a reasonable solution:

We can actually just replace the `findFuncByName` operation entirely, and instead have a single `findSymbolByName` (or maybe `findSymbolAddressByName`) operation, which is intended to work for both functions and data.

Because the new operation would be replacing the old one, the change will be binary-compatible on any platform where function pointers and data pointers are the same size (which is a constraint I'm willing to live with since all platforms we are likely to care about have that guarantee).

We can then provide `findFuncByName` as an inline wrapper function so that existing client code continues to work.

Th change would be binary-compatible but source-breaking for any client code that defined their own `ISlangSharedLibrary`, but I can live with that.

A more extreme alternative is to leave `findFuncByName` as-is, and simply introduce a `findSymbolByName` that wraps it and returns a `void*` for the cases where somebody wants to look up a non-function symbol.

source/slang/slang.cpp
```diff
@@ -3652,6 +3652,14 @@ SLANG_API SlangResult spCompileRequest_getModule(
     return SLANG_OK;
 }
 
+SLANG_API SlangResult spCompileRequest_getSession(
+    SlangCompileRequest* request,
+    slang::ISession** outSession)
+{
+    *outSession = Slang::asInternal(request)->getLinkage();
```
tfoleyNV, line 3659: This code doesn't seem to be correctly increasing the reference count of the session/linkage.

csyonghe: Fixed.

slang.h
csyonghe: Renamed `findFuncByName` to `findSymbolAddressByName` and provided an inline wrapper for `findFuncByName`.

slang.h
csyonghe: Kept this because our test infra uses the old API and I don't want to change that in this PR.

PR #1529, opened 2020-09-01
Allow unspecialized existential shader parameters (dynamic dispatch).

csyonghe:
This change enables generating dynamic dispatch code when the user is not specifying the concrete type of an existential shader parameter.

It enables this test case:
```
[anyValueSize(8)]
interface IInterface
{
    int run(int input);
}

//TEST_INPUT:cbuffer(data=[rtti(MyImpl) witness(MyImpl, IInterface) 1 0], stride=4):name=gCb

[numthreads(4, 1, 1)]
void computeMain(uint3       dispatchThreadID : SV_DispatchThreadID)
{
    let tid = dispatchThreadID.x;

    let inputVal : int = tid;
    let outputVal = gCb.run(inputVal);

}

// No type input for dynamic dispatch //TEST_INPUT: globalExistentialType MyImpl
// Type must be marked `public` to ensure it is visible in the generated DLL.
public struct MyImpl : IInterface
{
    int val;
    int run(int input)
    {
        return input + val;
    }
};
```

When this test case is run, no `globalExistentialType` input is passed to `compileRequest`, therefore `bindExistential` pass will not rewrite the `gCb` global parameter. The global parameter will then be left untouched by the subsequenced `specializeIR` pass, and get captured and lowered by the `lower-generics` pass.

This PR implements the following changes:
1. Type layout for an existential typed struct field now occupies a fixed size uniform space in addition to `ExistentialTypeParam`  and `ExistentialObjectParam` slots when targeting CPU/CUDA.
2. Expose `ISession::getTypeRTTIMangledName` API to query the mangled name of a type RTTI object.
3. Add a `-disable-dynamic-dispatch` compiler flag and only report diagnostic errors for unspecified existential type parameters when the flag is enabled, otherwise the compiler will generate dynamic dispatch code around the parameter.
4. Respect `public` modifier on `struct`s, and make sure RTTI objects and witness tables for `public` `struct` types are always included in the generated DLL.
5. Bug fixes for testing framework and cpp-emit logic to ensure type RTTI objects and witness table objects can be discovered.


source/slang/slang-compiler.cpp
```diff
@@ -2483,13 +2483,14 @@ SlangResult dissassembleDXILUsingDXC(
...
+        // When dynamic dispatch is disabled, the program must
+        // be fully specialized by now. So we check if we still
         // have unspecialized generic/existential parameters,
-        // then there is a problem.
+        // and report them as an error.
         //
         auto program = compileRequest->getProgram();
         auto specializationParamCount = program->getSpecializationParamCount();
-        if( specializationParamCount != 0 )
+        if (compileRequest->disableDynamicDispatch && specializationParamCount != 0)
```
tfoleyNV, line 2493: This works for the two extreme cases:

* All type parameters have been specialized
* *No* type parameters have been specialized

The missing piece at that point is some kind of partial specialization. I don't think that the code at this site needs to change at all to support partial specialization, but we do need to figure out what it means for the public API that supports specialization.

My take is that akin to how we can specialize a parameter to a single type right now, we eventually need a way to "specialize" a parameter to a value representing:

* "Do not specialize this parameter; use dynamic dispatch instead"
* "Treat this parameter as if it could be one of these N types, but nothing else"

PR #1531, opened 2020-09-02
Rework type layout for ExistentialSpecializedType

csyonghe:
As a part of dynamic dispatch support, this change reworks the type layout rules for existential typed shader parameters.
Currently, a type layout contains a `base` part and a `pending` part. An incomplete type (that contains existential fields) will have all the ordinary parts laid out first in the `base` part, and all existential fields placed in the `pending` part. When the type is specialized (and becomes a `ExistentialSpecailizedType`, the concrete type used to specialize the existential gets filled in in the `pending` part of the layout and comes after all the ordinary fields.

The original `base` and `pending` layout mechanism is designed to allow variable sized fields on targets that does not support pointers, since if we had pointers, any values that has unknown size at layout calculation time can be accessed via a pointer and therefore become a known sized value. The dynamic dispatch support requires existential values be stored in a fixed sized box on GPU for performance considerations. Therefore, any user code that is written to allow dynamic dispatch will guarantee that an existential value of any allowed type can fit within a block of a predefined `anyValueSize`. This means that the layout rule can be simplified to just use the fixed size in the place of existential values and we no longer need the `base` and `pending` mechanism.

This change does not remove the `base-pending` mechanism, but instead extends the layout logic so that:
1) If generating code for CPU and CUDA, always reserve a fixed size uniform slot for existential values. The size is derived from the `anyValueSize` attribute on the interface type, plus 16 bytes for RTTI and witness table pointers. For GLSL and HLSL targerts, do not reserve the slot.
2) If the reserved slot is smaller than the actually laid out specialized type, fall back to the `base` and `pending` mechanism and access the specialized type from the `pending` region. If the actually laid out type fits in the reserved slot, just access it from the reserved slot.

A detail is that values stored in the fixed sized `anyValue` slot should be laid out according to `anyValueRules` regardless of the enclosing environment's layout rule. Currently we don't enforce this for GLSL/HLSL platforms since doing so requires us implementing the load/store logic following this layout rule on these targets first.

### DIFF
--- a/source/slang/slang-type-layout.cpp
+++ b/source/slang/slang-type-layout.cpp
@@ -687,6 +687,7 @@ struct HLSLRayTracingLayoutRulesImpl : DefaultVaryingLayoutRulesImpl
     {}
 };
 
+DefaultLayoutRulesImpl kDefaultLayoutRulesImpl;
 Std140LayoutRulesImpl kStd140LayoutRulesImpl;
 Std430LayoutRulesImpl kStd430LayoutRulesImpl;
 HLSLConstantBufferLayoutRulesImpl kHLSLConstantBufferLayoutRulesImpl;
@@ -710,6 +711,7 @@ HLSLRayTracingLayoutRulesImpl kHLSLHitAttributesParameterLayoutRulesImpl(LayoutR
 
 struct GLSLLayoutRulesFamilyImpl : LayoutRulesFamilyImpl
 {
+    virtual LayoutRulesImpl* getAnyValueRules() override;
     virtual LayoutRulesImpl* getConstantBufferRules() override;
     virtual LayoutRulesImpl* getPushConstantBufferRules() override;
     virtual LayoutRulesImpl* getTextureBufferRules() override;
@@ -730,6 +732,7 @@ struct GLSLLayoutRulesFamilyImpl : LayoutRulesFamilyImpl
 
 struct HLSLLayoutRulesFamilyImpl : LayoutRulesFamilyImpl
 {
+    virtual LayoutRulesImpl* getAnyValueRules() override;
     virtual LayoutRulesImpl* getConstantBufferRules() override;
     virtual LayoutRulesImpl* getPushConstantBufferRules() override;
     virtual LayoutRulesImpl* getTextureBufferRules() override;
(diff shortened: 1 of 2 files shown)


source/slang/slang-type-layout.cpp
```diff
@@ -3613,43 +3661,65 @@ static TypeLayoutResult _createTypeLayout(
...
+                    fixedExistentialValueSize = anyValueAttr->size;
                 }
-                else
-                {
-                    // The interface type does not have an `[anyValueSize]` attribute,
-                    // assume a default of 8 bytes.
-                    fixedSize += 8;
-                }
-                typeLayout->addResourceUsage(LayoutResourceKind::Uniform, fixedSize);
+                typeLayout->addResourceUsage(LayoutResourceKind::Uniform, fixedExistentialValueSize + 16);
```
tfoleyNV, line 3675: The `+ 16` here could probably do with a comment. That is for the RTTI and witness table fields, but it might not be clear to the reader of this code.

source/slang/slang-type-layout.cpp
```diff
@@ -3613,43 +3661,65 @@ static TypeLayoutResult _createTypeLayout(
...
+                auto uniformLayoutInfo =
+                    concreteTypeLayout->FindResourceInfo(LayoutResourceKind::Uniform);
+                if (!useFixedSizedExistentialValue ||
+                    (uniformLayoutInfo && uniformLayoutInfo->count > fixedExistentialValueSize))
+                {
+                    // TODO: for targets that supports pointers, oversized existential values
+                    // should be placed in an overflow region and only a pointer is needed in
+                    // the place of the fixed sized uniform slot.
+                    // We only need the "pending layout" mechanism for targets that does not
+                    // support pointers.
```
tfoleyNV, line 3712: I suspect that we don't want to use `pendingDataTypeLayout` *at all* for CPU/CUDA.

In the case where things don't fit, we'd have to put a pointer to the actual data into the fixed-size region, and that seems like the only important layout constraint. The code we generate should not care about where the pointed-to bytes reside, so trying to encode that in the layout information seems unnecessary.

I'd like to think of the `pendingDataTypeLayout` thing as *only* being a fix for legacy targets that don't support pointers and have all types be POD-able.

tfoleyNV (approved):

PR #1533, opened 2020-09-04
Allow mixing unspecialized and specialized existential parameters.

csyonghe:
This change makes it possible to mix unspecialized and specialized existential shader parameter when emitting CPU code. It enables the following simple test case:
```
// Test using interface typed shader parameters with dynamic dispatch.

//TEST(compute):COMPARE_COMPUTE:-cpu
//DISABLE_TEST(compute):COMPARE_COMPUTE:-cuda -xslang -disable-specialization

[anyValueSize(8)]
interface IInterface
{
    int run(int input);
}

//TEST_INPUT:cbuffer(data=[rtti(MyImpl) witness(MyImpl, IInterface) 1 0], stride=4):name=gCb

//TEST_INPUT:cbuffer(data=[rtti(MyImpl) witness(MyImpl, IInterface) 1 0], stride=4):name=gCb1
ConstantBuffer<IInterface> gCb1;

[numthreads(4, 1, 1)]
void computeMain(uint3       dispatchThreadID : SV_DispatchThreadID)
{
    let tid = dispatchThreadID.x;

    let inputVal : int = tid;
    let outputVal = gCb.run(inputVal) + gCb1.run(inputVal);

}

// Specialize gCb1, but not gCb2
//TEST_INPUT: globalExistentialType MyImpl
//TEST_INPUT: globalExistentialType __Dynamic
// Type must be marked `public` to ensure it is visible in the generated DLL.
public struct MyImpl : IInterface
{
    int val;
    int run(int input)
    {
        return input + val;
    }
};
```

### Part 1
The first part of this change is to allow the user to indicate not to specialize an existential parameter. 

First, to allow expressing the intention to not specailize an existential type parameter, we introduced a new builtin type called `__Dynamic`. This type can only be used as type specialization argument provided through the Slang API. This `__Dynamic` type informs the compiler to not generate specailized code for where the argument is used.

The `__Dynamic` type is defined in `core.meta.slang` and parsed into a new builtin front-end type `DynamicType`, and lowers into `IRDynamicType`. A `DynamicSubtypeWitness` is synthesized by the front-end to mark the fact that `__Dynamic` conforms to the associated interface type parameter.

Next, we check to see if all type arguments provided to a `BindExistential` operation is a dynamic type, if so, we do not insert a `BindExistentialDecoration` during IR lowering. Without a `BindExistentialDecoration`, the `bind-existential` pass will simply ignore the global parameter, leaving it to be picked up and handled by `lower-generics` and will just work. Note that this change does not implement the more general case of partial specialization, where only some arguments in a `BindExistential` operation is `__Dynamic`. We will need to investigate `bind-existential` pass systematically for that to work.

### Part 2
The second part of this change is to make sure specialized existential parameters continue to work given the new layout rule introduced in #1531. After running `specialize` pass, all specialized existential values are referenced through a `ExistentialBox` type, which works as a pointer type in the IR. This change first add an explicit reference to the corresponding interface type in a `ExistentialBox` type, and then introduce an explicit `GetValueFromExistentialValue` instruction to retrieve the true pointer to the value from the `ExistentialBox`. After `specialize` pass is run, there will be `ExistentialBox` typed parameters left in the places for the specialized parameter values. These types are translated the same way as interface types by the `lower-generics` pass, into a `tuple(RTTI, WitnessTable, AnyValue)` object. The `GetValueFromExistentialValue` then translates into operations to load and bit-cast the pointer to the `AnyValue` field. The current implementation `GetValueFromExistentialValue` is unsound in that it attempts to retrieve the address of a tuple element, which is in theory a temporary reigster value. We need a more systematic approach to deal with the problem of wrapping temporary values in pointers.

### Next Steps
1. Allow partial specailization of a `BindExistential` typed parameter.
2. When the size of the concrete specialization type does not fit in the fixed size slot, it is supposed to be loaded from a value stored at the fixed uniform slot. The logic to fetch the uniform slot as a pointer and load from it is not yet implemented. This can come as a follow-up PR.


source/slang/slang-ir.cpp
```diff
@@ -2524,6 +2532,14 @@ namespace Slang
         }
     }
 
+    IRInst* IRBuilder::emitGetValueFromExistentialBox(IRType* type, IRInst* existentialBox)
+    {
+        auto inst =
+            createInst<IRInst>(this, kIROp_GetValueFromExistentialBox, type, 1, &existentialBox);
```
tfoleyNV, line 2538: I guess technically the `getValueFromExistentialBox` instruction is effectively just a `load` because the `ExistentialBox<T>` type is a kind of pointer type.

I suspect that having a distinct op here makes sense, but I wonder if maybe I need to remove the pointer-ness from the box type eventually.

I think I made `ExistentialBox<T>` act like a `Ptr<T>` because at the time I thought that a value of existential type could just be a single pointer, and that would mean that a `Ptr<Derived>` would have the same layout as an `IBase`.

That is very much *not* the case today, both because an `IBase` is now a tuple, and also because the bits of the `Derived` value might not be stored indirectly.

As such, the pointer-like-ness of `ExistentialBox` is debatable at best. The main argument in favor of it is just for the "legacy" layout model where we might need to shuffle around data to make up for the lack of pointers.

I expect that in a future evolution of this logic, we might need to make a clean separation between `ExistentialBox<Bar, IFoo>`, which is simply indicating a type that is layout-compatible with `IFoo`, but that is known to contain a `Bar` in its any-value part, and a separate type named something like `PseudoPtr<Bar>` which represents the idea of a *zero-size* field that logically acts as a pointer to a value of type `Bar`.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -7859,21 +7863,30 @@ struct SpecializedComponentTypeIRGenContext : ComponentTypeVisitor
...
-                    irSlotArgs.getBuffer());
+                // Only insert a `BindExistentialSlots` decoration when there are at least
+                // one non-dynamic type argument.
+                if (hasConcreteTypeArg)
+                {
+                    builder->addBindExistentialSlotsDecoration(
+                        irParam,
+                        irSlotArgs.getCount(),
+                        irSlotArgs.getBuffer());
+                }
```
tfoleyNV, line 7889: The existential specialization pass should incrementally make the granularity of `BindExistentials<...>` smaller and smaller, until eventually everything is an `ExistentialBox<...>`.

It should be possible to add logic to the existential specialization pass that notices when we are about to introduce an `ExistentialBox<__Dynamic, IFoo>` and decides *not* to introduce that box and directly use `IFoo` instead. That would seemingly help in the case where a top-level type is being specialized for some parameters and not others.

tfoleyNV (approved):

source/slang/slang-lower-to-ir.cpp
csyonghe: Agreed, I was trying to avoid touching too many fronts in this PR. This logic here doesn't hurt even when we make the bind-existential pass capable of ha doing partial specialization.

source/slang/slang-ir-specialize.cpp
```diff
@@ -678,6 +705,10 @@ struct SpecializationContext
             workListSet.Remove(inst);
             cleanInsts.Add(inst);
 
+            // If inst represents a value of ExistentialBox type, all its uses
+            // must be through a `GetValueFromExistentialBox` inst.
+            maybeInsertGetExistentialValue(inst);
+
```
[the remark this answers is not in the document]
csyonghe: Right. There isn't code that deals with a value of ExistentialBox type, instead it is just transforming the ExistentialBox type insta directly without looking at where it is being used. So it is a bigger change if want to implement in this way.

PR #1535, opened 2020-09-08
Improvements around CompileResult::getBlob 

jsmall-nvidia:
* If a blob is not returned from CompileResult return a failure
* Test for failures when getBlob is used
* Make spGetEntryPointCode handle lack of blob gracefully


tfoleyNV (approved):
Looks like a good fix.

PR #1536, opened 2020-09-10
Allow existential types in `StructuredBuffer` element type.

csyonghe:
This change enables proper specialization of `StructuredBuffer` (and all other StructuredBuffer variants) shader parameters whose element type is an existential type / contains existential fields.

With this change, a `StructuredBuffer<IFoo>` parameter is handled the same way as a `ConstantBuffer<IFoo>` in that:
1. `BindExistential(StructuredBuffer<SomeExistentialType>, t, w)`  is transformed into `StructuredBuffer<BindExistential<SomeExistentialType, t, w>>`, similar to the `ConstantBuffer<SomeExistentialType>` case.
2. `StructuredBufferLoadMethod(wrapExistential(existentialStructuredBufferValue, ...), index)` is transformed into `wrapExistential(StructuredBufferLoadMethod(existentialStructuredBufferValue, index))`, similar to the handling of `load(wrapExistential(...))` case, where `StructuredBufferLoadMethod` can be `operator[]`, `Load` or `Consume`.

This change also contains a clean-up that removes the direct interface-type reference from an `IRWitnessTable` value, instead, we always retrieve the underlying interface type from a witness table value via its type, i.e. `witnessTable->getConformanceType() ` is a helper function of `cast<IRWitnessTableType>(witnessTable->getDataType())->getConformanceType()`.


source/slang/slang-ir-specialize.cpp
```diff
@@ -770,17 +770,91 @@ struct SpecializationContext
         }
     }
 
+    // Finds any `IRTargetDecoration` from `inst`. Recursively chasing `specialize` chains.
+    IRTargetIntrinsicDecoration* findTargetIntrinsicDecorationRec(IRInst* inst)
```
csyonghe, line 774: This is duplicated with `CLikeSourceEmitter::findTargetIntrinsicDecoration`, also there is a global scope `findTargetInstrinsicDecoration(IRInst* inst, String targetName)` method but it does not look through `specialize` and `generic` insts. Let me know if you would like to clean this up to have a single `findTargetIntrinsicDecoration` method.

tfoleyNV: I think we should aspire to do  the cleanup, but it should probably be its own PR just to make sure that we can test/validate just the refactoring in isolation.

source/slang/slang-ir-specialize.cpp
```diff
@@ -770,13 +770,99 @@ struct SpecializationContext
...
+        }
+        if (auto decor = inst->findDecoration<IRTargetIntrinsicDecoration>())
+            return decor;
+        return nullptr;
+    }
+
+    // Returns true if the call inst represents a call to
+    // StructuredBuffer::operator[]/Load/Consume methods.
+    bool isBufferLoadCall(IRCall* inst)
+    {
```
tfoleyNV, line 792: We have a growing problem in the compiler that we either need to represent an operation as its own `IROp` value, in which case we can easily detect it and special-case it through optimizations/transformations, *or* we need to represent it as a built-in function that can use the existing `call` opcode and not need to be special-cased in emit logic for each target.

It seems like we need a better long-term solution here than looking for things like target-intrinsic decorations. After all, shouldn't the target-intrinsic decorations be target-specific? How can we be sure we will find the target-intrinsic decoration that has `.Load` as its value (the HLSL one), and not one for another target?

I don't think this PR needs to introduce the end-all-be-all solution to this problem, but it might be good to take a stand on the issue.

Maybe rather than use the target-intrinsic decoration, we should introduce a new decoration (with supporting syntax) for operations that have "known semantics" (something like `[knownSemantics(...)]` and the operand of the `knownSemantic` attribute can be an integer value that names a case from an `enum KnownSemantics { ... }`. That would let us identify when a function call is invoking an operation with some known semantics to the compiler, and allow passes to take those semantics into account.

tfoleyNV (approved):
I'm nervous about the way that the buffer load operations are being detected; it doesn't seem like it would work across targets.

If this doesn't break anything I'm fine with it landing as-is, but we should start thinking about a better way to identify calls to functions with compiler-known semantics such that they can be optimized/transformed.

PR #1541, opened 2020-09-11
Dynamic dispatch bug fixes.

csyonghe:
Bug fixes for dynamic dispatch compiler passes.

- Prevent redundant witness table types being emitted.
- Prevent lower-generic passes that does in-place instruction modification from corrupting `IRSharedBuilder::globalValueNumberingMap` and `IRSharedBuilder::constantMap`s.

### DIFF
--- a/source/slang/slang-emit-cpp.cpp
+++ b/source/slang/slang-emit-cpp.cpp
@@ -1724,6 +1724,10 @@ void CPPSourceEmitter::emitRTTIObject(IRRTTIObject* rttiObject)
 void CPPSourceEmitter::_maybeEmitWitnessTableTypeDefinition(
     IRInterfaceType* interfaceType)
 {
+    if (m_interfaceTypesEmitted.Contains(interfaceType))
+        return;
+    m_interfaceTypesEmitted.Add(interfaceType);
+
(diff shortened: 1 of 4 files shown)


source/slang/slang-emit-cpp.cpp
```diff
@@ -1724,6 +1724,10 @@ void CPPSourceEmitter::emitRTTIObject(IRRTTIObject* rttiObject)
 void CPPSourceEmitter::_maybeEmitWitnessTableTypeDefinition(
     IRInterfaceType* interfaceType)
 {
+    if (m_interfaceTypesEmitted.Contains(interfaceType))
+        return;
+    m_interfaceTypesEmitted.Add(interfaceType);
+
```
tfoleyNV, line 1730: It's a bit unfortunate to have to add this logic. Why can't we just emit the witness table type when we encounter the actual instruction that represents the type?

tfoleyNV (approved):

source/slang/slang-emit-cpp.cpp
csyonghe: Yeah, I think it is just me being lazy to refactor the code that way.

PR #1542, opened 2020-09-11
Support shader parameters that are an array of existential type.

csyonghe:
Similar to #1536, this change adds the missing pieces in supporting using existential types in arrays, i.e. enabling this:
```
struct Params
{
    IInterface values[2];
};
ConstantBuffer<Params> gCb;
``` 

This PR implements the following missing pieces:
1. Hoisting `getElement[Ptr](wrapExistential(array, wrapArgs), index)` into `wrapExistential(getElement[Ptr](array, index), wrapAargs)`.
2. `BindExistentialType(ArrayType(ElementType), bindArgs)` ==> `ArrayType(BindExistentialType(ElementType, bindArgs))`
3. Make sure the type layout for an interface type correctly report the uniform slot in its `TypeLayoutInfo`.
4. Update the test infra to work with reflection info for existential values.


tools/render-test/bind-location.cpp
```diff
@@ -328,6 +328,23 @@ void BindSet::calcValueLocations(const BindLocation& location, Slang::List<BindL
     }
 }
 
+// Finds the first category from layout reflection that represents an actual value
+// i.e. that is not ExistentialType or ExistentialObject.
+template<typename LayoutReflectionType>
+slang::ParameterCategory findOrdinaryValueCategory(LayoutReflectionType* layout)
```
tfoleyNV, line 334: Right now, the layout/binding logic tries to use "ordinary" as a synonym for `LayoutResourceKind::Uniform` (just because "uniform" can imply other things, and we need a term to refer to "just bytes"). This function is using "ordinary" to mean something different, and maybe we need a better term.

How about `firstNonExistentialValueCategory` since that would describe exactly what is going on?

tfoleyNV (dismissed):

tfoleyNV (approved):

PR #1546, opened 2020-09-16
Embed default prelude for CUDA

tfoleyNV:
Slang supports the notion of a "prelude" that gets prepended to the source code we generate in language. For some targets, a prelude is not necessary (e.g., we compile to HLSL/GLSL and then on to DXBC/DXIL/SPIR-V just fine without a prelude), but some targets have been implemented in a way that makes a prelude necessary (notably CPU and CUDA). For the targets that require a prelude, the Slang codebase includes usable preludes under the `prelude/` directory.

Prior to this change, if a user was compiling for such a target (whether via command-line or API), there had to take responsibility for specifying the prelude to use (usually by passing in the contents of the prelude file(s) already included in the Slang distribution).

It is reasonable for a user to expect an out-of-the-box experience where compilation to CUDA PTX or native CPU code should Just Work, similarly to how compilation to SPIR-V Just Works. This change is a step in the direction of providing a user experiene that Just Works for common cases.

The main addition here is a tool called `slang-embed` that we run during our build to turn the `prelude/*.h` files into `prelude/*.h.cpp` files that embed the contents of the original `.h` file as a `const` variable.

By compiling and linking in the generated `.h.cpp` file for the CUDA prelude, we are then able to set the default prelude to use for CUDA at the time a session/linkage is created. That default prelude will be used unless the user manually specifies their own prelude (which current users of the CUDA back-end must be doing).

This change only sets up a default prelude for CUDA because of the way that the CPU prelude is split across multiple files. A strategy that provides a good default prelude for CPU may take more work, but that work might also be unnecessary if we switch to a strategy of using LLVM to generate native code.

The implementation of the `slang-embed` tool is intentionally simple, and it will likely run into issues if/when we need to embed binary files or larger text files. The assumption being made here is that we can address those issues when they arise, and there is no reason to over-engineer the tool right now.

The way that `slang-embed` is integrated into our build process is likely to require some iteration to make sure that it works across all platforms. I expect that this change will have multiple follow-up fixes related to trying to get the build to work as expected across all targets on CI.


tools/slang-embed/slang-embed.cpp
```diff
@@ -0,0 +1,197 @@
...
+            fprintf(stderr, "%s: error: failed to open '%s' for reading\n", appName, inputPath);
+            exit(1);
+        }
+
+        // We derive an output path simply by appending `.cpp` to the input path.
+        //
+        // TODO: If we start adding more complicated options, a `-o` option
+        // to specify a desired output path would be an obvious choice.
+        //
+        char* outputPath = (char*) malloc(strlen(inputPath) + strlen(".cpp") + 1);
```
csyonghe, line 77: May I request that we don't leak even in these kind of run-once tools. The reason is that if we enable any memory testing tools, they could complain these leaks during the build process and cause the build to fail.

csyonghe (approved):

tools/slang-embed/slang-embed.cpp
tfoleyNV: Good point. I'll fix this up to properly clean up the scratch memory.

csyonghe (approved):

PR #1547, opened 2020-09-16
Share debug information between AST and IR

jsmall-nvidia:
* Split out debug information such that it can be shared across IR and/or AST
* Small improvements around RiffContainer, to make a little more convenient to use
* Made 'Container' a distinct type - in SerialContainerData
* Use a string pool for serializing entry points, to make each entry point fixed size
* Made serialize files conform to slang-serialize-xxx to group together
* Made payload in RiffContainer alignment 8 bytes 
* Removed SerialStringTable, just use StringSlicePool instead

The main effort here was to put the main missing feature in AST serialization in place - that it can also serialize out debug information. This is done here by allowing the IR and AST serialization to use DebugSerialReader/Writer optionally and save off debug info into it's own riff section.


source/core/slang-riff.h
```diff
@@ -177,6 +177,14 @@ need to be recalculated, before serialization.
 class RiffContainer
 {
 public:
+    enum 
+    {
+        // This alignment is only made for arena based allocations.
+        // For external blocks it's client code to have appropriate alignment.
+        // This is needed because when reading a RiffContainer, all allocation is arena based, and
+        // if the payload contains 8 byte aligned data, the overall payload needs to be 8 byte aligned.
+        PAYLOAD_MIN_ALIGNMENT = 8,
```
tfoleyNV, line 186: This isn't really a style we've used for constants so far.

source/slang/slang-serialize-container.h
```diff
@@ -0,0 +1,107 @@
...
+/* Struct that holds all the data that can be held in a 'container' */
+struct SerialContainerData
+{
+    struct Target
+    {
+        CodeGenTarget       codeGenTarget = CodeGenTarget::Unknown;
+        SlangTargetFlags    flags = 0;
+        Profile             profile;
+        FloatingPointMode   floatingPointMode = FloatingPointMode::Default;
+    };
```
tfoleyNV, line 42: It feels like having this type is redundant. We both have the `TargetReq` type that holds the same basic data (and is what we'd likely use this to create) and also the public API's `slang::TargetDesc` type that is nearly identical to this struct except that the fields are in a different order, and there is an extra `optimizationLevel` field not present here.

Is it possible for the (de)serialization logic to directly work with a `slang::TargetDesc` as supported by the public API?

source/slang/slang-serialize-container.h
```diff
@@ -0,0 +1,107 @@
...
+    
+    struct TargetModule
+    {
```
tfoleyNV, line 45: The equivalent type in `slang-compiler.h` is currently called `TargetProgram`, and in reality it acts more like a `TargetComponentType` or `TargetSpecializedComponentType`.

The name `TargetModule` is a bit misleading here, because while this thing does hold an `IRModule`, it doesn't relate to the notion of a "module" as understood by the non-IR parts of the compiler.

It is probably not necessary to come up with a different name right now, but we will eventually need a consistency pass where we try to get our terminology the same across as many parts of the compiler as possible.

source/slang/slang-serialize-container.h
```diff
@@ -0,0 +1,107 @@
...
+        Target target;
+        RefPtr<IRModule> irModule;
+    };
+
+    struct TranslationUnit
+    {
+        RefPtr<IRModule> irModule;
+        RefPtr<ASTBuilder> astBuilder;
+        NodeBase* astRootNode = nullptr;
+    };
```
tfoleyNV, line 56: Similarly, this `TranslationUnit` type is precisely what the front-end logic in `slang-compiler.h` calls a `Module`, and it is *not* the same as a `TranslationUnitRequest`.

tfoleyNV (approved):

source/slang/slang-serialize-container.h
jsmall-nvidia: I'm happy to rename - as with some of these other types, it wasn't really clear what was appropriate.

PR #1550, opened 2020-09-17
Control container serialization with SerialOptionFlags

jsmall-nvidia:
On top of #1547

* Add flags to control what is output in SerialContainer.
* Turn off AST output for obfuscated code.
* Lazily create astClasses when doing write container serialization.


source/slang/slang-serialize-types.h
```diff
@@ -0,0 +1,232 @@
...
+namespace Slang {
+
+// Options for IR/AST/Debug serialization
+
+struct SerialOptionFlag
+{
+    typedef uint32_t Type;
+    enum Enum : Type
+    {
+        RawSourceLocation   = 0x01,     ///< If set will store directly SourceLoc - only useful if current source locs will be identical when read in (typically this is *NOT* the case)
```
tfoleyNV, line 21: It seems like we are basically at the point where we need/want to drop `RawSourceLocation` support, since our long-term goals for serialization are almost all incompatible with it.

source/slang/slang-serialize-types.h
```diff
@@ -0,0 +1,232 @@
...
+        DebugInfo           = 0x02,     ///< If set will output debug information, that can be reconstructed when read after being stored.
```
tfoleyNV, line 22: I wonder if `DebugInfo` should more specifically be `SourceLocationInfo`, because that is really the main debugging-related information we emit today.

I guess the question is: if we added support for other debugging-related information, would we want it to all be controlled by the same flag, or would we want different flags for different kinds of debug information?

tfoleyNV (approved):
It is a bit hard to review this when the preceding PR that it builds on hasn't landed, but everything I can find looks okay.

PR #1555, opened 2020-09-21
Allow #include of absolute paths

jsmall-nvidia:


source/slang/slang-include-system.cpp
```diff
@@ -9,15 +9,26 @@ namespace Slang
...
 SlangResult IncludeSystem::findFile(SlangPathType fromPathType, const String& fromPath, const String& path, PathInfo& outPathInfo)
 {
-    // Get relative path
-    ComPtr<ISlangBlob> combinedPathBlob;
-    SLANG_RETURN_ON_FAIL(m_fileSystemExt->calcCombinedPath(fromPathType, fromPath.begin(), path.begin(), combinedPathBlob.writeRef()));
-    String combinedPath(StringUtil::getString(combinedPathBlob));
-    if (combinedPath.getLength() <= 0)
+    String combinedPath;
+
+    if (fromPath.getLength() == 0 || Path::isAbsolute(path))
```
tfoleyNV, line 14: Unrelated to the specifics of this PR, but I wonder if we should make `Path` be a type that wraps a `String` and provides these operations in a more fluent-feeling way.

tfoleyNV (approved):

PR #1560, opened 2020-09-23
Enable default cpp prelude.

csyonghe:
This PR changes slang-embed to allow it to handle `#include` directives, and process C++ preludes the same way as cuda and hlsl preludes.

This change also removes the inclusion of `slang.h` from C++ preludes to reduce its size.


tools/slang-embed/slang-embed.cpp
```diff
@@ -92,13 +94,99 @@ struct App
         // could lead to a difference in the embedded bytes based on
         // the line ending convention of the host platform)
         //
-        FILE* inputFile = fopen(inputPath, "r");
+        Slang::StreamReader streamReader(inputPath);
+        FILE* inputFile = fopen(inputPath.getBuffer(), "r");
```
tfoleyNV, line 98: Is there a reason the `inputFile` variable is still being used?

tools/slang-embed/slang-embed.cpp
```diff
@@ -160,7 +248,11 @@ struct App
         // task of outputting the generated source file is simple.
         //
         fprintf(outputFile, "// generated code; do not edit\n");
-        fprintf(outputFile, "const char* %s =\n", variableName);
+        fprintf(outputFile, "#include \"../source/core/slang-basic.h\"\n");
+
+        fprintf(outputFile, "Slang::String get_%s()\n", variableName);
```
tfoleyNV, line 253: I'm not 100% happy with the approach that computes the file contents into a heap-allocated `String` on the fly. I would much rather that we stick with a `static const` array and just move it to be a `static const uint8_t kWhatever[] = { ... }` with explicit array initializer. My (limited) investigations have shown that array initializers don't have the same size limitations as string literals in VS, so it ought to avoid the size limitation just the same.

The eventual goal would be for `slang-embed` to be usable for binary payloads as well (not just text files), so finding a solution that lets us embed binary data of any size without heap allocation would be nice.

tfoleyNV (approved):

tools/slang-embed/slang-embed.cpp
csyonghe: Good catch.

tools/slang-embed/slang-embed.cpp
csyonghe: I didn't know that you are not subject to the same limitation on normal arrays. I am going to leave it as is now, since in this case we are not using more heap allocations since the result will get assigned to a `String` anyways.

PR #1562, opened 2020-09-24
Add API for whole program compilation.

csyonghe:
This change exposes a new target flag: `SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM` that can be set on a target with `spSetTargetFlags`. When this flag is set, `spCompile` function generates target code for the entire input module instead of just the specified entrypoints. The resulting code will include all the entrypoints defined in the input source.

The resulting whole program code can be retrieved with two new functions: `spGetTargetCodeBlob` and `spGetTargetHostCallable`.

This change also cleans up the unnecessary `entryPointIndices` parameter of `TargetProgram::getOrCreateWholeProgramResult`, and modifies the `cpu-hello-world` example to make use of the new whole-program compilation API to simplify its logic.

### DIFF
--- a/examples/cpu-hello-world/main.cpp
+++ b/examples/cpu-hello-world/main.cpp
@@ -85,6 +85,10 @@ static SlangResult _innerMain(int argc, char** argv)
     // If we wanted a just a shared library/dll, we could have used SLANG_SHARED_LIBRARY.
     int targetIndex = spAddCodeGenTarget(slangRequest, SLANG_HOST_CALLABLE);
 
+    // Set the target flag to indicate that we want to compile all the entrypoints in the
+    // slang shader file into a library.
+    spSetTargetFlags(slangRequest, targetIndex, SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM);
+
     // A compile request can include one or more "translation units," which more or
     // less amount to individual source files (think `.c` files, not the `.h` files they
     // might include).
@@ -99,14 +103,6 @@ static SlangResult _innerMain(int argc, char** argv)
     // There are also variations of this API for adding source code from application-provided buffers.
     //
     spAddTranslationUnitSourceFile(slangRequest, translationUnitIndex, "shader.slang");
-    
-    // Next we will specify the entry points we'd like to compile.
-    // It is often convenient to put more than one entry point in the same file,
-    // and the Slang API makes it convenient to use a single run of the compiler
-    // to compile all entry points. 
-    // 
-    const char entryPointName[] = "computeMain";
-    int computeIndex = spAddEntryPoint(slangRequest, translationUnitIndex, entryPointName,  SLANG_STAGE_COMPUTE);
 
     // Once all of the input options for the compiler have been specified,
     // we can invoke `spCompile` to run the compiler and see if any errors
@@ -133,7 +129,7 @@ static SlangResult _innerMain(int argc, char** argv)
     // Get the 'shared library' (note that this doesn't necessarily have to be implemented as a shared library
     // it's just an interface to executable code).
     ComPtr<ISlangSharedLibrary> sharedLibrary;
-    SLANG_RETURN_ON_FAIL(spGetEntryPointHostCallable(slangRequest, 0, 0, sharedLibrary.writeRef()));
+    SLANG_RETURN_ON_FAIL(spGetTargetHostCallable(slangRequest, 0, sharedLibrary.writeRef()));
 
     // Once we have the sharedLibrary, we no longer need the request
     // unless we want to use reflection, to for example workout how 'UniformState' and 'UniformEntryPointParams' are laid out
@@ -141,10 +137,10 @@ static SlangResult _innerMain(int argc, char** argv)
     spDestroyCompileRequest(slangRequest);
 
     // Get the function we are going to execute 
+    const char entryPointName[] = "computeMain";
     CPPPrelude::ComputeFunc func = (CPPPrelude::ComputeFunc)sharedLibrary->findFuncByName(entryPointName);
     if (!func)
     {
-        spDestroyCompileRequest(slangRequest);
         return SLANG_FAIL;
     }
 
--- a/slang.h
+++ b/slang.h
@@ -584,6 +584,7 @@ extern "C"
            @deprecated This behavior is now enabled unconditionally.
         */
         SLANG_TARGET_FLAG_PARAMETER_BLOCKS_USE_REGISTER_SPACES = 1 << 4,
+        SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM = 1 << 8
     };
 
     /*!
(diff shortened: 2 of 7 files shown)


examples/cpu-hello-world/main.cpp
```diff
@@ -133,18 +129,18 @@ static SlangResult _innerMain(int argc, char** argv)
...
     // unless we want to use reflection, to for example workout how 'UniformState' and 'UniformEntryPointParams' are laid out
     // at runtime. We don't do that here - as we hard code the structures. 
     spDestroyCompileRequest(slangRequest);
 
     // Get the function we are going to execute 
+    const char entryPointName[] = "computeMain";
     CPPPrelude::ComputeFunc func = (CPPPrelude::ComputeFunc)sharedLibrary->findFuncByName(entryPointName);
     if (!func)
     {
-        spDestroyCompileRequest(slangRequest);
```
tfoleyNV, line 147: Why did you remove the cleanup code on the failure path?

slang.h
```diff
@@ -584,6 +584,7 @@ extern "C"
            @deprecated This behavior is now enabled unconditionally.
         */
         SLANG_TARGET_FLAG_PARAMETER_BLOCKS_USE_REGISTER_SPACES = 1 << 4,
+        SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM = 1 << 8
```
tfoleyNV, line 587: Probably deserves a comment?

slang.h
```diff
@@ -1659,6 +1660,36 @@ extern "C"
...
         ISlangSharedLibrary**   outSharedLibrary);
 
+    /** Get the output code associated with a specific target.
+
+    @param request   The request 
+    @param targetIndex The index of the target to get code for (default: zero).
+    @param outBlob A pointer that will receive the blob of code
+    @returns A `SlangResult` to indicate success or failure.
+
+    The lifetime of the output pointer is the same as `request`.
```
tfoleyNV, line 1670: This comment seems inaccurate; I assume the output blob is properly reference-counted and will be retained as long as the caller retains it.

source/slang/slang-compiler.cpp
```diff
@@ -2197,24 +2197,23 @@ SlangResult dissassembleDXILUsingDXC(
...
-            if (*entryPointIndex >= m_entryPointResults.getCount())
-                m_entryPointResults.setCount(*entryPointIndex + 1);
+        // It is possible that entry points goot added to the `Program`
+        // *after* we created this `TargetProgram`, so there might be
+        // a request for an entry point that we didn't allocate space for.
+        //
+        // TODO: Change the construction logic so that a `Program` is
+        // constructed all at once rather than incrementally, to avoid
+        // this problem.
+        //
```
tfoleyNV, line 2210: Note: this comment seems to have been copy-pasted over from `_createEntryPointResult`, and doesn't seem accurate to anything that is going on here. It might be good to change it to a more accurate comment about how the current codegen logic requires a list of entry point indices to generate, so that a "whole-program" result requires us to manually specify all entry points for inclusion.

tfoleyNV: In the context of talking about specialization and linking, it is worth noting that the list of entry points here is *not* exactly the same as the list of entry points declared via `[shader(...)]` attributes in the input.

What is being collected here is a list of all entry points that were linked into `m_program`, which could include explicitly-specialized entry points if this `TargetProgram` was created via the `IComponentType` part of the API.

For the older C-style API, the `m_program` will always consist of all referenced modules, plus all the entry points that were declared/found, each specialized to any specialization arguments that were specified via API/command-line.

source/slang/slang-compiler.h
```diff
@@ -1132,7 +1132,10 @@ namespace Slang
         SlangTargetFlags    targetFlags = 0;
         Slang::Profile      targetProfile = Slang::Profile();
         FloatingPointMode   floatingPointMode = FloatingPointMode::Default;
-        bool                isWholeProgramRequest = false;
+        bool isWholeProgramRequest()
+        {
+            return (targetFlags & SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM) != 0;
+        }
```
tfoleyNV, line 1138: We might eventually need/want to have certain targets that *always* generate a whole-program output. CPU and CUDA targets would seem like obvious candidates for this, but SPIR-V and DXIL might also qualify if we can work through any complication.

For formats and targets that support multiple entry points, we already have the rule that the client of Slang needs to look up entry points by name in the output blob we produce (at least, I think that is the case...), so we could always return a multiple-entry-point blob in response to a request for a single entry point.

I guess that defaulting to such behavior might confuse users, but I also don't like having to make users explicitly opt into what is obviously the right behavior for targets like CPU/CUDA...

tfoleyNV: Actually, I think what we *might* want to do here is to have the public API operations for getting a whole-program result take an (optional) output parameter for a blob of diagnostics, and then they can get `getOrCreateWholeProgramResult` instead of `getExistingWholeProgramResult`, which would mean that they could be used to query the whole-program result *and* generate it on demand if it wasn't already created. That approach would mean that even if an application leaves out the `SLANG_TARGET_FLAG_GENERATE_WHOLE_PROGRAM` flag, they can still query the whole-program result.

source/slang/slang.cpp
```diff
@@ -3613,6 +3613,33 @@ static SlangResult _getEntryPointResult(
...
+    if ((targetIndex < 0) || (targetIndex >= targetCount))
+    {
+        return SLANG_ERROR_INVALID_PARAMETER;
+    }
+    auto targetReq = linkage->targets[targetIndex];
+
+    auto targetProgram = program->getTargetProgram(targetReq);
+    if (!targetProgram)
+        return SLANG_FAIL;
+    *outCompileResult = &targetProgram->getExistingWholeProgramResult();
```
tfoleyNV, line 3639: Should this step yield an error if the result doesn't have any data in it?

source/slang/slang.cpp
```diff
@@ -3648,6 +3675,43 @@ SLANG_API SlangResult spGetEntryPointHostCallable(
...
+{
+    using namespace Slang;
+    if (!outBlob)
+        return SLANG_ERROR_INVALID_PARAMETER;
+    Slang::CompileResult* compileResult = nullptr;
+    SLANG_RETURN_ON_FAIL(
+        _getWholeProgramResult(request, targetIndex, &compileResult));
+
+    ComPtr<ISlangBlob> blob;
+    SLANG_RETURN_ON_FAIL(compileResult->getBlob(blob));
```
tfoleyNV, line 3691: Never mind, I guess that the failure on a missing/invalid result will come here.

tfoleyNV (dismissed):

examples/cpu-hello-world/main.cpp
csyonghe: It is already destroyed in the line above regardless of failure here.

tfoleyNV (approved):

PR #1563, opened 2020-09-29
Generalizing Serialization

jsmall-nvidia:
This is the first and main part of a refactor of the AST serialization such that it can be used generally. The changes

* Split out the conversion between native/serial types (misc and more general)
* Created SerialReader/SerialWriter that can serialize types other than AST types
  + Through using SerialExtraObjects - to attach extra types needed to serialize
  + Use of a SerialObjectFactory to construct types
* Make types identified by their TypeKind/SubType combination
* Debug info -> SourceLoc types (SerialSourceLocData/Reader/Writer)

In order to work around type information for RefObject derived types a hack is used here. That the type is determined by lots of casts(!), and construction just uses a switch. This is okay for now because there is only one 'regular' RefObject derived type that needs this functionality (other types such as Name/String are specially handled).

The next stage is to make it so that RefObject derived classes can provide reflection information. The most straight forward way to do this would be to add a SerialRefObject base class that has a method that can return a ReflectClassInfo. The next problem is the serialization identifies types via an integral value. There is already code that given a list of ReflectClassInfos will use the super class information to correctly work out casting appropriate ids. For now the list of all the types will need to be determined together such that this can be produced. 

This can be supported via some SLANG_CLASS style macros, and the C++ extractor to extract fields.


tfoleyNV (approved):
I can't pretend to fully follow the serialization system and its design, but the refactoring stuff here seems fine to me.

PR #1566, opened 2020-09-30
Use new vulkan debug layer.

csyonghe:
`"VK_LAYER_LUNARG_standard_validation"` is deprecated and no longer included in latest NVIDIA drivers. Changing it to `"VK_LAYER_KHRONOS_validation"`


tangent-vector (approved):
I'm glad that this code checks for both the old and new extension names; I was concerned that we'd make our code incompatible with older drivers.

PR #1570, opened 2020-10-02
Fix C++ emit for `bit_cast` inst.

csyonghe:
We currently emit a C style type cast for the `bit_cast` inst, this is incorrect.

This change defines a `slang_bit_cast` function in the C++ prelude and have `bit_cast` inst emit as a call to it.


tfoleyNV (approved):
Looks fine. We probably need to all reach some consensus on what approach we use when we want "type punning." I know we have some logic doing pointer casts like this, and others using `union`s, and both of those options are officially undefined in C/C++. Having a single rule for how we do this stuff should at least make it easier to audit things if/when we ever need to change the behavior for certain compilers.

(As far as I understand, the only portable way to do type-punning is to call `memcpy` or do the moral equivalent of it)

csyonghe:
I am not convinced that the compiler can always optimized that memcpy out, especially in CUDA. For now there isn't a real standard compliant way to do type punning in C++. What we need is std::bit_cast, unfortunately we can't assume C++2x in our generated code yet.

tfoleyNV:
To be clear, I agree 100% that we should not default to using `memcpy` in things like our generated code (I *do* believe that we probably ought to default to `memcpy` in our host/compiler code, just to play it safe).

The point I was making is more that we want to bottleneck as many things as possible through a small number of routines that handle type punning. Having everything in generated C/C++ code go through a `slang_bit_cast` is actually a great step. The next step would be to make everything in our host code do something similar.

The biggest problem I can see with this PR is the fact that we have two definitions of `slang_bit_cast` (one for CUDA and one for host CPU), when one would probably suffice. It would be great to eventually merge the shared parts of the CPU/CUDA prelude, *especially* if we decide that we are going to go ahead and keep relying on C++ support.

PR #1574, opened 2020-10-08
NVAPI support doc

jsmall-nvidia:
* Split out from target-compatibility.md
* Attempt to describe update NVAPI arrangement, doc was out of date.


docs/nvapi-support.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+In order for the include to work, it is necessary for the include path to include the folder that contains the nvHLSLExtns.h and associated headers.
+
+Implicit usage of NVAPI
+=======================
+
+It is convenient and powerful to be able to directly use NVAPI calls, but will only work on such targets that support the mechansism, even if there is a way to support the functionality some other way.
+
+Slang provides some cross platform features on HLSL based targets that are implemented via NVAPI. For example RWByteAddressBuffer atomics are supported on Vulkan, DX12 and CUDA. On DX12 they are made available via NVAPI, whilst CUDA and Vulkan have direct support. When compiling Slang code that uses RWByteAddressBuffer atomice Slang will emit HLSL code that use NVAPI. In order for the downstream compiler to be able to compile this HLSL it must be able to include the NVAPI header `nvHLSLExtns.h`.
```
csyonghe, line 27: atomice->atomics

csyonghe (dismissed):

PR #1575, opened 2020-10-09
Support CUDA bindless texture in dynamic dispatch code.

csyonghe:
This change adds the `anyValue` packing and unpacking for resource types (64bit `int`s).

To run a test case, the existing testing framework is extended to create bindless resource objects and fill their handles into buffers, just like what is done for RTTI object pointers.

The following resource specifier is now supported in test cases:
```
//TEST_INPUT: Texture2D(size=8, content = one):name t2D,bindless
//TEST_INPUT:ubuffer(data=[rtti(MyImpl) witness(MyImpl, IInterface) handle(t2D) 0 0], stride=4):name=gCb
```


tangent-vector (approved):
This looks good. I'm a bit concerned that the "bindless" terminology might be confusing, but it seems like a reasonable way to fit this behavior in with the existing support.

Eventually we really need the `TEST_INPUT` system to be split to allow two different constructs:

1. The existing *assignment* behavior would take a left-hand-side that references a shader parameter or a sub-piece of one, and a right-hand side that describes a value to assign there.

2. A new *creation* behavior would use a left-hand side that defines a named temporary, and a right hand-side that describes a value to put into that temporary. The name of that temporary could then be used in other assignment and/or creation steps.

The above pretty much describes how this change works at a functional level, but I think we could aspire to make things more clear. I recommend that we not try to dig into that kind of cleanup until after we've managed to unify the shader parameter binding logic across CPU/CUDA and graphics targets.

PR #1583, opened 2020-10-20
Single pass C++ extraction

jsmall-nvidia:
The main objective here was to make the C++ extractor able to parse all input header files in one go, and output multiple output files. Each node the C++ extractor parses can belong to one type set - which one determined by the 'magic macro' name. More concretely

```
class Thing
{
    SLANG_ABSTRACT_AST_CLASS(Thing)
};
```

(assuming a prefix of SLANG_ and a suffix of _CLASS), produces a TypeSet name of 'AST'. That we output generated files for all the AST types in file with 'ast' as part of the name. 

To make this work the command line options either have to work for every input/output across all files, or we need a mechanism to specify output information. The only real problem here was with the 'Type' put on the command line. Here this is specified on a TypeSet basis, in the source via the SLANG_TYPE_SET(set, typename)

* Changed premake5.lua to use the new mechanism - and only have to run C++ extractor once.
* Added CharUtil
* Added a bunch of functions to StringUtil to convert string naming conventions (may remove to own util)
* Allow multiple base types per type set


source/core/slang-char-util.h
```diff
@@ -0,0 +1,51 @@
+#ifndef SLANG_CORE_CHAR_UTIL_H
+#define SLANG_CORE_CHAR_UTIL_H
+
+#include "slang-string.h"
+
+namespace Slang {
+
+struct CharUtil
+{
```
tfoleyNV, line 9: We've talked about it many many times, but I will continue to raise my objection that I think the stuff here would be easier to discover and use if it was written as global functions rather than namespaced into a type.

source/core/slang-char-util.h
```diff
@@ -0,0 +1,51 @@
...
+        };
+    };
+
+    SLANG_FORCE_INLINE static bool isDigit(char c) { return c >= '0' && c <= '9'; }
+    SLANG_FORCE_INLINE static bool isLower(char c) { return c >= 'a' && c <= 'z'; }
+    SLANG_FORCE_INLINE static bool isUpper(char c) { return c >= 'A' && c <= 'Z'; }
+    SLANG_FORCE_INLINE static bool isLineWhitespace(char c) { return c == ' ' || c == '\t'; }
+
+        /// True if it's alpha
+    SLANG_FORCE_INLINE static bool isAlpha(char c) { return (getFlags(c) & (Flag::Upper | Flag::Lower)) != 0; }
```
tfoleyNV, line 29: Have you benchmarked that using the flags this way is faster than just doing `isLower(c) | isUpper(c)`? My intuition would be that having to do the load to index into `g_charFlagMap` would cost more than doing both comparisons and the OR.

source/core/slang-char-util.h
```diff
@@ -0,0 +1,51 @@
...
+            Digit           = 0x04,         
+            LineWhitespace  = 0x08,         ///< Whitespace that can appear in a line (ie excluding CR/LF)
+            HexDigit        = 0x10,         ///< 0-9, a-f, A-F
+        };
+    };
+
+    SLANG_FORCE_INLINE static bool isDigit(char c) { return c >= '0' && c <= '9'; }
+    SLANG_FORCE_INLINE static bool isLower(char c) { return c >= 'a' && c <= 'z'; }
+    SLANG_FORCE_INLINE static bool isUpper(char c) { return c >= 'A' && c <= 'Z'; }
+    SLANG_FORCE_INLINE static bool isLineWhitespace(char c) { return c == ' ' || c == '\t'; }
```
tfoleyNV, line 26: Should this be `HorizontalWhitespace`? That's what I assume "line whitespace" means here, and from my googling I don't see the term "line whitespace" used much other than as part of "end-of-line whitespace."

source/core/slang-char-util.h
```diff
@@ -0,0 +1,51 @@
...
+    SLANG_FORCE_INLINE static char toLower(char c) { return (c >= 'A' && c <= 'Z') ? (c -'A' + 'a') : c; }
+        /// Given a character return the upper case equivalent
+    SLANG_FORCE_INLINE static char toUpper(char c) { return (c >= 'a' && c <= 'z') ? (c -'a' + 'A') : c; }
+
+    struct CharFlagMap
+    {
+        Flags flags[0x100];
+    };
+
+    static const CharFlagMap g_charFlagMap;
```
tfoleyNV, line 46: Given that all of this stuff is very ASCII-specific, should we make this whole thing be `ASCIIUtil` instead of `CharUtil`? At some point I'd like us to have at least basic support for Unicode text in parts of Slang, but at the same time I expect that ASCII-specific utilities like this will be worth keeping around even in that hypothetical future (because things like the C++ extractor can and should always focus on ASCII only).

source/core/slang-name-convention-util.cpp
```diff
@@ -0,0 +1,193 @@
...
+                        // TODO(JS): This doesn't catch situations where there is a single letter like
+                        // ABox, because of this rule will be split to AB, ox
+                        // If we have acronyms not being capitalized this would not be an issue
+
+                        // Could be an acronym, if the next character is lower
+                        if (CharUtil::isLower(cur[1]))
+                        {
+                            out.add(UnownedStringSlice(start, cur));
+                            start = cur;
+                        }
```
tfoleyNV, line 57: I'm not sure I follow this comment, because the obvious splitting is `A`, `Box`.

The informal rules seem to be:

* If the previous char was lower-case, and the next is upper-case, split. (you definitely have that rule)

* If the previous char char was upper-case, and the next is upper-case, and the one after *that* is lowercase, split. (That seems to be what you implemented here)

In the case of `ABox`, we'd have the point where the `A` has been read, so the cursor is placed like `A|Box`. In that case, `flags & CharFlag::Upper` is true, `prevFlags & CharFlag::Upper` is true, `cur + 1 < end` is true, and `CharUtil::isLower(cur[1])` should be true as well. It seems like that would split at the `|` (`cur`) and lead to us getting `A`, `Box` as the output, which is what we'd want.

Am I misunderstanding the code, or the comment?

source/core/slang-name-convention-util.h
```diff
@@ -0,0 +1,49 @@
...
+
+Split is used to split up a name into it's constituent 'words' based on a convention.
+Join is used to combine words based on a convention/character case
+
+Convert uses split and join to allow easy conversion between conventions. 
+*/
+struct NameConventionUtil
+{
+        /// Given a slice and a naming convention, split into it's constituent parts. 
+    static void split(NameConvention convention, const UnownedStringSlice& slice, List<UnownedStringSlice>& out);
```
tfoleyNV, line 35: It's not really important to the current PR, but it seems like it wouldn't be too big of a stretch to have a version of `NameConventionUtil::split` that infers the convention from a name (or a separate function that infers convention).

source/slang/slang-ast-base.h
```diff
@@ -4,7 +4,7 @@
 
 #include "slang-ast-support-types.h"
 
-#include "slang-ast-generated.h"
+#include "slang-generated-ast.h"
```
tfoleyNV, line 7: Not very important, but if we ever change these names again, I might argue for a blanket policy of having names like `slang-ast.generated.h` so that files of the form `*.generated.*` always represent generated code and then we can set up our `.gitignore` and things like that to always ignore them.

(In fact, I suspect we *do* already have `.gitignore` lines that ignore `*-generated.h`, so we might want to make sure that this PR updates those)

tfoleyNV (approved):

source/core/slang-char-util.h
jsmall-nvidia: I have benchmarked on x64, if single range test it's slightly faster, with two ranges it is typically (although perhaps not always) faster.

source/core/slang-name-convention-util.cpp
jsmall-nvidia: The comments not right - the limitation around upper acronyms is two acronyms next to one another. Like say IAABBSystem.

source/slang/slang-ast-base.h
jsmall-nvidia: Ok fixed .gitignore - which was actually already out of date.

tfoleyNV (dismissed):

PR #1585, opened 2020-10-21
Generate `if` based dispatch logic on GPU targets.

csyonghe:
Our previous implementation of dynamic dispatch makes use of function pointers on both CPU and GPU. Upon initial performance benchmarks, we observed very slow performance of function-pointer calls on CUDA. Combined with CUDA's restriction that all functions callable from a function pointer must be defined in the same CUDA module, the function-pointer based implementation is not beneficial in anyways in practice.

This change adds a new subpass `specializeDispatchFunctions` to the `lower-generics` pass that modifies the dispatch functions introduced in #1584 with `if` statement based dispatch logic. This subpass is enabled on GPU targets only.

Currently the dispatch function produced by this pass will perform cascaded `if` comparisons on the witness table argument to all witness tables of the same interface type defined in the current module, and calls the corresponding witness table entry when the comparison result is true. The pass can be further improved to use a `switch` statement to match type ids instead of witness table pointers, but that requires coordination with the user-facing compiler API and front-end logic to determine type ids and pass them through to the lowering pass.

The `lower-generic-call` subpass will identify all interface method dispatch calls and generate intial dispatch function implementations that are of the form 
```
func dispatch_func(IRWitnessTableType(IRInterfaceType) witnessTable, OrignalFuncParams args)
{
    return call(lookup_interface_method(witnessTable, memberFuncKey), args);
}
```

These functions will be hold in `mapInterfaceRequirementKeyToDispatchFunction` dictionary in `SharedGenericsLoweringContext`, and be processed in the subsequent newly added `specializeDispatchFunctions` pass, which translates them into the form of
```
func dispatch_func(IRWitnessTableType(IRInterfaceType) witnessTable, OrignalFuncParams args)
{
    if (bit_cast<uint64_t>(witnessTable) == bit_cast<uint64_t>(type1_witnessTable))
          return call(type1_memberFunc, args);
    if (bit_cast<uint64_t>(witnessTable) == bit_cast<uint64_t>(type2_witnessTable))
          return call(type2_memberFunc, args);
    ...
    return call(typeN_memberFunc, args);
}
```

A minor detail is that since we did not have a comparison intrinsic function for `IRWitnessTable` type, the witness table argument is first `bit_cast`ed into a `uint64_t`, and then do a integer comparison.

### DIFF
--- a/source/slang/slang-ir-generics-lowering-context.h
+++ b/source/slang/slang-ir-generics-lowering-context.h
@@ -31,6 +31,9 @@ namespace Slang
         // Dictionaries for interface type requirement key-value lookups.
         // Used by `findInterfaceRequirementVal`.
         Dictionary<IRInterfaceType*, Dictionary<IRInst*, IRInst*>> mapInterfaceRequirementKeyValue;
+        
+        // Map from interface requirement keys to its corresponding dispatch method.
+        OrderedDictionary<IRInst*, IRFunc*> mapInterfaceRequirementKeyToDispatchMethods;
 
         SharedIRBuilder sharedBuilderStorage;
 
--- a/source/slang/slang-ir-lower-generic-call.cpp
+++ b/source/slang/slang-ir-lower-generic-call.cpp
@@ -8,9 +8,6 @@ namespace Slang
     {
         SharedGenericsLoweringContext* sharedContext;
 
-        // Map from interface requirement keys to its corresponding dispatch method.
-        OrderedDictionary<IRInst*, IRFunc*> mapInterfaceRequirementKeyToDispatchMethods;
-
         // Represents a work item for unpacking `inout` or `out` arguments after a generic call.
         struct ArgumentUnpackWorkItem
         {
@@ -91,8 +88,8 @@ namespace Slang
         // Create a dispatch function for a interface method.
         // On CPU, the dispatch function is implemented as a witness table lookup followed by
         // a function-pointer call.
-        // TODO: On GPU targets, we should implement the dispatch function with a `switch` statement
-        // based on the type ID.
+        // On GPU targets, we can modify the body of the dispatch function in a follow-up
+        // pass to implement it with a `switch` statement based on the type ID.
         IRFunc* _createInterfaceDispatchMethod(
             IRBuilder* builder,
             IRInterfaceType* interfaceType,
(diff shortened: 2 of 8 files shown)


source/slang/slang-ir-generics-lowering-context.h
```diff
@@ -31,6 +31,9 @@ namespace Slang
         // Dictionaries for interface type requirement key-value lookups.
         // Used by `findInterfaceRequirementVal`.
         Dictionary<IRInterfaceType*, Dictionary<IRInst*, IRInst*>> mapInterfaceRequirementKeyValue;
+        
+        // Map from interface requirement keys to its corresponding dispatch method.
+        OrderedDictionary<IRInst*, IRFunc*> mapInterfaceRequirementKeyToDispatchMethods;
```
tfoleyNV, line 36: Eventually we need to figure out how this will work in the presence of generic interfaces. If you have:

```
interface IHolder<T>
{
    T getValue();
}
```

Then there will be a single key for `IHolder<...>::getValue`, and even if we do explicit specialization, that key won't change. That means that witness tables for `IHolder<int>` and `IHolder<double>` would use the same key.

As a result, it seems like we have two choices here:

1. Ensure that the dispatch method we generate is usable for any instantiation of a generic interface. That would seem to require that the dispatch method would itself be generic (e.g., `IHolder_getValue<T>(...)`), or be equivalent to the dynamic-dispatch version of such a generic.

2. Generate distinct dispatch methods for each instantiation of `IHolder<...>`. In that case, it seems like the `mapInterfaceRequirementKeyToDispatchMethod` approach wouldn't be enough, since it would need to map the pair of `(specializedInterface, requirementKey)` to the dispatch method.

Naively, it seems to me like option (1) is closest to what we'd want long-term, but you'd probably have a better intuition given that you've implemented most of the logic involved. Thoughts?

tfoleyNV (approved):

source/slang/slang-ir-generics-lowering-context.h
csyonghe: None of the other parts of `lower-generics` pass considers/handles the possibility of generic interfaces. I think one possible answer is to have a pre-pass that turns all generic interfaces into interfaces with generic methods, and then lower those generic methods as usual.

PR #1589, opened 2020-10-27
Handling imported/exporting symbols from serialized modules

jsmall-nvidia:
* Removed ImportExternalDecl as added an import mechanism to serialization mechanism
* Added mechanism (and cache) for looking up Decls based on mangled names
* Added (perhaps incomplete) mechanism to extract module name from MangledName
* Fixed bugs around deserializing AST module correctly/SourceLocs
* Added the stdlib modules to mapNameToLoadedModules for a new Module, so they can be looked up

That with this we should be able to do a couple of things

1) Deserialize the stdlib modules and that work (!) maybe ;)
2) When linkage demands module, we can make it look for a serailized binary module and load that (as well as look for source)


source/slang/slang.cpp
```diff
@@ -468,6 +467,15 @@ Linkage::Linkage(Session* session, ASTBuilder* astBuilder)
...
     setFileSystem(nullptr);
+
+    // Copy of the built in linkages modules
+    if (builtinLinkage)
+    {
+        for (const auto& pair : builtinLinkage->mapNameToLoadedModules)
+        {
+            mapNameToLoadedModules.Add(pair.Key, pair.Value);
+        }
+    }
```
tfoleyNV, line 478: Hmm... I'm not sure I like the idea of the module names from the builtin linkage being automatically pulled into any child linkages, but I also can't pinpoint a case where that would cause problems.

I guess one note here is that this change would make it possible for a user file to write `import hlsl;` and have that be interpreted as an import of the module from `hlsl.meta.slang` that is part of the stdlib. At the same time, we don't actually treat the visibility of symbols from `hlsl` or `core` in user code as them having being imported, so doing an import like that would likely create an error for the user. That's probably fine, so long as our stdlib module names are unlikely to conflict with user-defined modules.

I'd have a slight preference here for having any code that wants to search for a module by name first look in the current linkage and then in whatever builtin linkage that refers to, but I'm also fine with this solution as a kind of quick fix.

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
     m_entryPoints.add(entryPoint);
 }
 
+static bool _canExportDeclSymbol(ASTNodeType type)
```
tfoleyNV, line 1977: The new `static` functions here probably deserve documentation.

tfoleyNV: I might argue for an even more straightforward-but-wordy name for this function. Something like `_canDeclsOfTypeBeExported` to match the semantics of "this function returns `true` if declarations of the given `type` can be exported from a module"

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
...
+{
+    switch (type)
+    {
+        case ASTNodeType::ModuleDecl:
+        case ASTNodeType::EmptyDecl:
+        case ASTNodeType::NamespaceDecl:
```
tfoleyNV, line 1983: I'm kind of surprised at why `NamespaceDecl` would be on the excluded list here, since a `namespace` declaration isn't very different from a `struct` declaration that just contains `static` members.

Is there a specific reason it had to be left out?

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
...
+        {
+            return false;
+        }
+        default: break;
+    }
+
+    return true;
+}
+
+static bool _canRecurseExportSymbol(Decl* decl)
```
tfoleyNV, line 1993: I'm not 100% sure this subroutine is needed (since it only gets called in one place), but if we do keep it I'd argue for it having a worry-but-precise name like `_canContainExportedDecls`.

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
...
+}
+
+void Module::_processFindDeclsExportSymbolsRec(Decl* decl)
+{
+    if (_canExportDeclSymbol(decl->astNodeType))
+    {
+        // It's a reference to a declaration in another module, so first get the symbol name. 
+        String mangledName = getMangledName(getASTBuilder(), decl);
+
+        Index index = Index(m_mangledExportPool.add(mangledName));
```
tfoleyNV, line 2010: It's not relevant right now, but I want to keep bringing up that eventually we will probably want to store any large tables of mangled names in a compressed for that collapses out common prefixes.

The reason to collapse out common prefixes is that all the child declarations of some parent type declaration `SomeModule.SomeType` are going to have names of the form `SomeModule.SomeType.*`, so that storing the substring `SomeModule.SomeType` (or rather, the mangling of it) redundantly will get costly.

There are things we can do in the mangling scheme itself to make things more compact, but those are in some ways orthogonal to redundancy across names. The main thing we will want from our mangling scheme when we get aroudn to doing it right is a guarantee that the similarity or "closeness" of two entities is reflected in their mangled names sharing a common prefix. (As a concrete example, we probably *don't* want all mangled function names to start with `F` and all mangled type names to start with `T`, because that gets rid of potential shared prefixes).

I'll need to look up the references on this stuff, but once you have a list of names that you'd like to build into a string pool/table, you can then sort them, and for each entry in the sorted list compute the number N of prefix bytes it shares with its immediate predecessor in the list, and the suffix S that is needs to append to those bytes. You can also compute the index of the earliest entry in the list with which it shares the given prefix. You can then transform each string into a form like

```
struct Entry
{
    size_t prefixBytesSharedWithPrevious;
    size_t indexOfEarliestPredecessorWithMatchingPrefix;
    size_t suffixBytesToAdd;
    char const* suffixBytes;
};
```

Expanding out entries like that into full strings is pretty easy in a linear scan, but more importantly it is also possible to compare a string against an entry in the table without ever fully expanding all the strings. Given an index in the table and a string to test, you can check whether it ends with the given `suffixBytes` for the entry, and then jump to the `indexOfEarliestPredecessorWithMatchingPrefix` and continue matching (with a little bit of careful indexing to make sure you only compare against the bytes that make up the prefix for the chosen entry). IIRC the check should be linear in the length of the string.

A compressed table like that could be used as a way to confirm hit-vs-miss for lookups into a serialized hash table of mangled named in a module.

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
...
+
+        // TODO(JS): It appears that more than one entity might have the same mangled name.
+        // So for now we ignore and just take the first one.
```
tfoleyNV, line 2013: This seems worrying, so we should probably set aside some time for one of us to debug why we are ending up with symbols that have the exact same mangled name.

source/slang/slang.cpp
```diff
@@ -1965,6 +1974,87 @@ void Module::_addEntryPoint(EntryPoint* entryPoint)
...
+    {
+        for (auto child : containerDecl->members)
+        {
+            _processFindDeclsExportSymbolsRec(child);
+        }
+    }
+    else if (auto genericDecl = as<GenericDecl>(decl))
+    {
+        _processFindDeclsExportSymbolsRec(genericDecl->inner);
+    }
```
tfoleyNV, line 2037: I think the order of the `if`s here may be wrong, since a `GenericDecl` is also a `ContainerDecl`.

tfoleyNV (approved):

PR #1595, opened 2020-11-05
Include hierarchy output

jsmall-nvidia:
* Added a new command line option -output-includes (perhaps needs a better name)
* Improved output when a token paste is used
* Added GetOrAddValue to Dictionary - simplifies the construction of the heirarchy structure

The core idea here was to make a SourceView to optionally record a SourceLoc that defined where it was initiated from in other source. This 'initiation' could be from a #include or from token pasting. 

With this extra information it is possible to display #includes in the order they are taken. The initial version used the token stream to define the order.. but of course that doesn't work if there aren't any tokens. The solution used here is to just just build up the hierarchy defined by initiating parent/child relationships and then order the children by source locs. This of course assumes the SourceLocs increase in source order (which they do right now). 

With the hierarchy, it is fairly straightforward to walk the tree. 

The token paste was actually the initiating idea to the PR - it seemed something could be knocked together such that it would be faster to write it than write an issue. That intuition was false though - outputting where a paste initiated (in effect which macro), is perhaps somewhat useful, but that's not really enough - at a minimum knowing where the original macro application occurred would be very useful. More useful but perhaps yet more complex would be to know the history of how the token was produced. Will add an issue - as this PR doesn't really fix the the original motivating problem well enough.


jsmall-nvidia:
I'm going to refactor this a bit - I don't like my hierarchy type it's overkill.

docs/command-line-slangc.md
```diff
@@ -147,6 +147,8 @@ For completeness, here are the options that `slangc` currently accepts:
 
 * `--`: Stop parsing options, and treat the rest of the command line as input paths
 
+* `-output-includes`: After pre-processing has been performed will output to via the diagnostics the hierarchy of paths to source files reached
```
tfoleyNV, line 150: I don't think we need to do anything right now, but it seems like we might eventually want to think about a common prefix to add to command-line options that are mostly there for debugging purposes.

source/slang/slang-diagnostics.cpp
```diff
@@ -177,6 +177,42 @@ static void formatDiagnostic(
             humaneLoc = sourceView->getHumaneLoc(sourceLoc);
         }
         formatDiagnostic(humaneLoc, diagnostic, sb);
+
+        {
+            SourceView* currentView = sourceView;
+
+            while (currentView && currentView->getInitiatingSourceLoc().isValid() && currentView->getSourceFile()->getPathInfo().type == PathInfo::Type::TokenPaste)
```
tfoleyNV, line 184: To make sure I understand; this is specifically only working up the stack of source views in the case where the error location is inside the results of a token paste?

If that is correct, then this seems like a good way to introduce this functionality without making large-scale changes to our diagnostic output in all cases.

Eventually we will likely want to turn on the printing of the full "backtrace" on more error messages, but it seems like there are open questions about what the policy for that should be (every message? every non-note message?)

source/slang/slang-options.cpp
```diff
@@ -972,9 +976,10 @@ struct OptionsParser
                             String name;
                             SLANG_RETURN_ON_FAIL(tryReadCommandLineArgument(sink, arg, &argCursor, argEnd, name));
 
-                            String slice = argStr.subString(1, index - 1);
+                            // Skip the initial -, up to the last -
+                            UnownedStringSlice passThruSlice(argStr.begin() + 1, argStr.begin() + index);
```
tfoleyNV, line 980: It might be nice to add a `.subString()` method to `UnownedStringSlice`  so that the line could just change to:

```
auto slice = argStr.subString(1, index - 1);
```

and thus be independent of whether `argStr` is a `String` or `UnownedStringSlice` or whatever else we might define down the line.

source/slang/slang-options.cpp
```diff
@@ -972,9 +976,10 @@ struct OptionsParser
...
                             SlangPassThrough passThrough = SLANG_PASS_THROUGH_NONE;
-                            if (SLANG_SUCCEEDED(TypeTextUtil::findPassThrough(slice.getUnownedSlice(), passThrough)))
+                            if (SLANG_SUCCEEDED(TypeTextUtil::findPassThrough(passThruSlice, passThrough)))
```
tfoleyNV, line 982: Unrelated to this change (since this change eliminates the `.getUnownedSlice()` call, rather than adding it): it is worth considering whether conversion from `String` to `UnownedStringSlice` should be implicit.

tests/diagnostics/token-paste-location.slang.expected
```diff
@@ -0,0 +1,7 @@
+result code = -1
+standard error = {
+token paste(1): error 20001: unexpected '%', expected identifier
+tests/diagnostics/token-paste-location.slang(10): note: see token pasted location
```
tfoleyNV, line 4: Pasting `%` and `%` is perhaps a bit confusing. The traditional C preprocessor rules are that the result of a token paste must be a single valid token, but in this case we are pasting `%` and `%` to get the text `%%` and then tokenizing them back into `%` and `%` (hence why the error says `unexpected '%'` instead of `unexpected '%%'`).

This behavior in Slang (where we re-tokenize after a paste, and (seemingly) don't give an error when more than one token is produced) could be considered a bug, and users might want us to fix it down the line (even if I'm inclined to keep it since it is internally consistent).

If we want this test case to not care about this semantic wrinkle, it might be best to use another character than `%`; using `+` would seemingly produce similar results without the semantic question.

tfoleyNV (approved):

source/slang/slang-diagnostics.cpp
jsmall-nvidia: Yes - it's only working up 'stack' from pastes. That I suspect from there we want to do the backtrace of macro applications, as is touched on in issue #1600. You might want also to back trace includes - and I would agree there are open issues on how this should be done, where, and how to control it.

PR #1598, opened 2020-11-06
Use integer RTTI/witness handles in existential tuples.

csyonghe:
This change makes existential parameters lower into a `tuple(rtti_handle, witness_id, anyvalue)` instead of `tuple(rtti, witness_table, anyvalue)`. `rtti_handle` and `witness_id` types will be lowered into integers by an optional specialization pass that specializes witness tables and RTTI objects into `switch` based dispatch functions and removes the concepts of interfaces, witness tables and RTTI objects from the resulting code, making it suitable for GPU execution.

With this change, the resulting dynamic dispatch code no longer contains things that can't map to HLSL, e.g. pointers and function pointers.

The change organizes the witness-table specialization pass into an optional pass, without which the emit logic will pick up the interfaces, RTTI objects and witness tables that are left from the `lower-generics` pass and emit pointer-based code. The `strip-witness-table` pass is changed to always-execute, since none of the interface or witness table definitions are used after the witness-table specialization pass.


source/slang/slang-emit-cpp.cpp
```diff
@@ -1652,8 +1658,10 @@ void CPPSourceEmitter::_emitWitnessTableDefinitions()
         m_writer->emit(" = {\n");
         m_writer->indent();
         auto seqIdDecoration = witnessTable->findDecoration<IRSequentialIDDecoration>();
-        SLANG_ASSERT(seqIdDecoration);
-        m_writer->emit((UInt)seqIdDecoration->getSequentialID());
+        if (seqIdDecoration)
+            m_writer->emit((UInt)seqIdDecoration->getSequentialID());
+        else
+            m_writer->emit("0");
```
tfoleyNV, line 1664: Does this `0` case arise in practice? Should it be something where we give an error message?

source/slang/slang-emit-cpp.cpp
```diff
@@ -1669,7 +1677,7 @@ void CPPSourceEmitter::_emitWitnessTableDefinitions()
                 m_writer->emit(getName(witnessTableVal));
             }
             else if (entry->getSatisfyingVal() &&
-                     isPointerOfType(entry->getSatisfyingVal()->getDataType(), kIROp_RTTIType))
+                     entry->getSatisfyingVal()->getDataType()->op, kIROp_RTTIHandleType)
```
tfoleyNV, line 1680: It sure looks like there is a typo bug here, since the new code seems to be using "operator comma" in a way that would make this condition be true in a lot of cases where it shouldn't be.

Do you mean to be comparing `...->op == kIROp_RTTIHandleType` like you do in the code below?

source/slang/slang-emit.cpp
```diff
@@ -664,18 +664,7 @@ Result linkAndOptimizeIR(
...
-    case CodeGenTarget::CUDASource:
-        break;
-    default:
-        // For all targets that don't support true dynamic dispatch through
-        // witness tables, we need to eliminate witness tables from the IR so
-        // that they don't keep symbols live that we don't actually need.
-        stripWitnessTables(irModule);
-        break;
-    }
+    stripWitnessTables(irModule);
```
tfoleyNV, line 667: Maybe a `TODO` or `Note` comment is warranted, to point out that we currently always strip witness tables because our dynamic dispatch no longer needs them, but if/when we have targets that support dynamic dispatch with separately-compiled functions then we might need to leave them in place?

source/slang/slang-ir-lower-generics.cpp
```diff
@@ -16,6 +16,94 @@
 
 namespace Slang
 {
+    void specializeRTTIObjectReferences(SharedGenericsLoweringContext* sharedContext)
```
tfoleyNV, line 19: More comments would be more better. :)

I've been allowing a lot of this generics lowering and dynamic dispatch work to go in without extensive comments because it is important that we get it working, but I think it would be healthy for us to go back at some point and make sure it gets the kind of extensive comments that will make it easier (or even possible) to bring up new maintainers on this code.

I realize that we have a *lot* of code in the project that is impenetrable to anybody but the person who wrote it (e.g., a lot of the current type legalization stuff), but that is something I'd like us to push back on whenever we write new code.

tfoleyNV (approved):

tfoleyNV (changes requested):
This should be okay to go once the place that looks like a typo bug is addressed.

source/slang/slang-emit-cpp.cpp
csyonghe: Good catch here. Thanks!

source/slang/slang-emit-cpp.cpp
csyonghe: It arises when I am using the code to generate function-pointer based dispatch, where this field is not used and sequential IDs are not assigned.

tfoleyNV (approved):

PR #1599, opened 2020-11-06
Fix IR serialization to use variable length encoding for opcode.

csyonghe:
The number of IR instructions has expanded beyond 256, so 8 bits is no longer sufficient to represent all possible IR opcodes.

The `kIROpMeta_OpMask` value has been changed to 0x3FF and number of opcode bits has been changed to 10 previously.

This change modifies the IR encoding/decoding logic and use variable length integer encoding for opcodes to fix serialization errors.

### DIFF
--- a/source/slang/slang-serialize-ir-types.h
+++ b/source/slang/slang-serialize-ir-types.h
@@ -129,7 +129,7 @@ struct IRSerialData
         
         SLANG_FORCE_INLINE bool operator!=(const ThisType& rhs) const { return !(*this == rhs); }
 
-        uint8_t m_op;                       ///< For now one of IROp 
+        uint16_t m_op;                      ///< For now one of IROp 
         PayloadType m_payloadType;	 		///< The type of payload 
         uint16_t m_pad0;                    ///< Not currently used             
 
--- a/source/slang/slang-serialize-ir-types.h
+++ b/source/slang/slang-serialize-ir-types.h
@@ -131,7 +131,7 @@ struct IRSerialData
 
         uint16_t m_op;                      ///< For now one of IROp 
         PayloadType m_payloadType;	 		///< The type of payload 
-        uint16_t m_pad0;                    ///< Not currently used             
+        uint8_t m_pad0;                    ///< Not currently used             
 
         InstIndex m_resultTypeIndex;	    //< 0 if has no type. The result type of this instruction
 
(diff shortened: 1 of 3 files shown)


source/slang/slang-serialize-ir-types.h
```diff
@@ -129,7 +129,7 @@ struct IRSerialData
         
         SLANG_FORCE_INLINE bool operator!=(const ThisType& rhs) const { return !(*this == rhs); }
 
-        uint8_t m_op;                       ///< For now one of IROp 
+        uint16_t m_op;                      ///< For now one of IROp
```
jsmall-nvidia, line 132: In order to keep the current packing, it would be better to make m_pad0 into a uint8_t.

source/slang/slang-serialize-ir.cpp
```diff
@@ -190,7 +190,7 @@ Result IRSerialWriter::write(IRModule* module, SerialSourceLocWriter* sourceLocW
             IRInst* srcInst = m_insts[i];
             Ser::Inst& dstInst = m_serialData->m_insts[i];
 
-            dstInst.m_op = uint8_t(srcInst->op & kIROpMeta_OpMask);
```
jsmall-nvidia, line 193: Although it wasn't in the original source it would be an improvement to have a typedef for the the op integral type. So not a problem with this PR, just an observation.

source/slang/slang-serialize-ir.cpp
```diff
@@ -357,8 +357,8 @@ Result _encodeInsts(SerialCompressionType compressionType, const List<IRSerialDa
             encodeOut = encodeArrayOut.begin() + offset;
             encodeEnd = encodeArrayOut.end();
         }
+        encodeOut += ByteEncodeUtil::encodeLiteUInt32(inst.m_op, encodeOut);
```
jsmall-nvidia, line 360: Using the encoder is probably a reasonable way to go here. For the future it would be better to make the IR encoding perhaps more like AST format, where it is more stream like and has some natural alignment (like 32 bits). Not an issue with this PR though.

jsmall-nvidia: The line maxInstSize will need to change - because now it can take up an extra 2 bytes. That we could add 

ByteEncodeUtil::kMaxLiteEncodeUInt16 to be 3. If so that line would be 

```
const size_t maxInstSize = 1 + ByteEncodeUtil::kMaxLiteEncodeUInt16 + Math::Max(sizeof(insts->m_payload.m_float64), size_t(2 * ByteEncodeUtil::kMaxLiteEncodeUInt32));
```

Where the 1 is for the payload type

jsmall-nvidia (changes requested):
Making _pad0 be a uint8_t and fixing maxInstSize will be good to go.

jsmall-nvidia (approved):

PR #1601, opened 2020-11-10
Integrate github action for linux build+test.

csyonghe:


csyonghe:
If no one objects, I'd like to get this checked in first just to see how github actions work in real. I am still learning how to use it to publish releases, but that can come as a separate PR.

tfoleyNV (approved):

PR #1605, opened 2020-11-13
Test for serializing out and reading back Stdlib

jsmall-nvidia:
On top of #1603

* Exposed by adding some 'experimental' API functions to Slang API 
* Tests via compute rendering test using stdlib
* Added -load/save/compile-stdlib as options to slang
* Added TestToolUtil to test for load/compile options


slang.h
```diff
@@ -2988,6 +2988,22 @@ namespace slang
         virtual SLANG_NO_THROW void SLANG_MCALL getLanguagePrelude(
             SlangSourceLanguage sourceLanguage,
             ISlangBlob** outPrelude) = 0;
+
+        
+            /** Compile from (embedded source) the StdLib on the session.
+            Will return a failure if there is already a StdLib available
+            NOTE! API is experimental and not ready for production code 
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL compileStdLib() = 0;
```
tfoleyNV, line 2997: As already predicted, I'm not a huge fan of having these in the public API, and I'm even less of a fan of having them on a core interface like this (so that we are potentially stuck having them in the binary interface "forever").

Realistically, I don't expect these to create a binary-compatibility issue, and we can probably pull them in a later version/release without breaking anything.

slang.h
```diff
@@ -3378,12 +3394,31 @@ namespace slang
...
+
+@param apiVersion Pass in SLANG_API_VERSION
+@param outGlobalSession (out)The created global session that doesn't have a StdLib setup.
+
+NOTE! API is experimental and not ready for production code 
+*/
+SLANG_API SlangResult slang_createGlobalSessionWithoutStdLib(
+    SlangInt                apiVersion,
+    slang::IGlobalSession** outGlobalSession);
+
```
tfoleyNV, line 3421: It is probably fine to have this operation for now, but this points to the fact that we *probably* need to start migrating to a model where `slang_createGlobalSession` or whatever we use takes a flexible/extensible "descriptor" of some kind (whether it is an array of key-value pairs, or just a `struct` that embeds a size or version field).

source/core/slang-test-tool-util.cpp
```diff
@@ -37,6 +37,19 @@ namespace Slang
     }
 }
 
+/* static */bool TestToolUtil::hasDeferredStdLib(Index argc, const char*const* argv)
+{
```
tfoleyNV, line 41: I will continue to register my complaint that I don't like the test-runner looking at the command line of tests to figure out things about them. There's not a great alternative right now, but if/when we get around to a grand refactor of `slang-test`, I think that one of the big goals will be making it not have to know/understand the policy of each and every test type it supports.

tfoleyNV (approved):

PR #1610, opened 2020-11-17
Fix constant folding in attributes

csyonghe:
This PR fixes a crash during constant folding triggerred from an attribute value.
For example,
```
const int A = 2;
const int B = 3;
const int C = A+B;
[numthreads(C, 1, 1)]
float maint() {}
```
This is because when the compiler tries to determine the value of `C` while checking the `numthreads` attribute, the init expr of the `VarDecl` referenced by `C` has not been checked yet, and attempt to access the invoke expr's `decl` or modifiers results in a null pointer dereference.

The fix is to make sure a `VarDeclBase` has been checked before attempting to extract its value. Change is implemented in `SemanticsVisitor::tryConstantFoldDeclRef`.


tests/compute/groupshared.slang
```diff
@@ -1,39 +1,32 @@
...
 //TEST(compute):COMPARE_COMPUTE_EX:-cuda -compute
-
 //TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out, name=gBuffer
 RWStructuredBuffer<int> gBuffer;
-
-#define THREAD_COUNT 4
-
+static const int A = 4;
+static const int B = 1;
+static const int THREAD_COUNT = A + B;
```
tfoleyNV, line 10: Shouldn't this be `A * B` so that the thread count is the same as in the original?

tests/compute/groupshared.slang
```diff
@@ -1,39 +1,32 @@
 // groupshared.slang
-
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx12
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute
 //TEST(compute):COMPARE_COMPUTE_EX:-cuda -compute
-
```
tfoleyNV, line 7: There are some spurious whitespace diffs in this change (the removed lines here, and then indentation changes later on), possibly as a result of how the modified test file got shared. It might be nice if the only diffs in the test where the ones to set up `THREAD_COUNT` as a `static const` variable instead of a `#define`.

tfoleyNV (approved):
Would nice to see the smaller issues cleaned up, but the actual bug fix is good.

PR #1613, opened 2020-11-19
Make witness and RTTI handles lower to `uint2`.

csyonghe:
In previous PRs, we have already implemented a `switch` based dynamic dispatch without relying on pointers. This should allow all dynamic dispatch code to be supported for D3D/VK targets.

With PR #1612, the only remaining block to support `switch` based dynamic dispatch on D3D/VK is that the witness table and RTTI handles lower to `uint64_t`, which isn't supported by all platforms.

This PR makes all such handles lower to `uint2` instead, and adds the vector packing logic in places that are necessary.

One small issue with fxc is that if a `switch` statement contains only one `default` case, it sometimes complains that not all control paths return a value. To fix this, we omit the `switch` statement if there is only 1 case.

This PR also enables some dynamic dispatch tests on D3D/VK. Will enable all dynamic dispatch tests on D3D as a follow up PR.

### DIFF
--- a/source/slang/slang-ir-link.cpp
+++ b/source/slang/slang-ir-link.cpp
@@ -1481,21 +1481,19 @@ LinkedIR linkIR(
             cloneValue(context, bindInst);
         }
     }
-    if (target == CodeGenTarget::CPPSource || target == CodeGenTarget::CUDASource)
-    {
-        for (IRModule* irModule : irModules)
-        {
-            for (auto inst : irModule->getGlobalInsts())
-            {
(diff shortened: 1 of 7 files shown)


tangent-vector:
Is the plan to re-enable the static specialization tests that I had to disable because of the `uint64_t` usage as part of this PR, or would that come as a follow-on?

csyonghe:
Let's do that as a follow up.
There are still major issues with this PR.

tangent-vector (approved):

PR #1615, opened 2020-11-20
Enable all dynamic-dispatch tests on D3D/VK.

csyonghe:


tests/compute/dynamic-dispatch-10.slang
```diff
@@ -1,3 +1,5 @@
+//TEST(compute):COMPARE_COMPUTE:-dx11
+//TEST(compute):COMPARE_COMPUTE:-vk
 //TEST(compute):COMPARE_COMPUTE:-cpu -xslang -disable-specialization
```
jsmall-nvidia, line 3: I *think* if we have a cpu definition that will synthesize the cuda test.

jsmall-nvidia: That it would be good if we had a better way of specifying tests - such that we can just list what it requires, or where it can't work and everything else is synthesized. That's not anything at issue in this PR tho.

jsmall-nvidia (approved):

PR #1617, opened 2020-11-20
Zip FileSystem support

jsmall-nvidia:
On top of #1616

* Added miniz external dependency 
* Added a ISlangMutableFileSystem implementation that is backed by an in memory zip archive
* Added unit test for impl

The main difficultly here is around the mutable part. That miniz doesn't directly support this feature. You can have an archive that is a reader or a writer. When writing it acts as a 'stream' - you can write the same file multiple times for example(!). If you want to remove a file you would need to copy everything into a new archive without the file/directory. To work around this, this implementation has

* Lazy remove 
* It's own dictionary to lookup by name 
** This helps perf (writers do linear lookup) as well as holding the 'latest' entry
* Handling of implicit directories (a directory implied by a path to an object, but not explicitly created)

The dictionary always points to the 'latest' version of an entity(file/directory). That a UIntSet is used to indicate if an entity has been removed. This can be done on a reader archive. 

The upside of the approach is that if items are removed or added, we don't need to constantly recreate the archive. The downside is that the old version remain in the archive. When the archive is requested (via getArchive) all of the laziness is removed and a single reader archive is produced, that can just be written to disk. 

A fairly simple addition might want to see how much redundant memory is consumed, and then 'flatten' the archive out again. It would only be necessary to do this on write operations.

It's also worth noting to make this lazy evaluation work it was necessary to initialize the writer such that it had a compatible 'read' function. As the writer just has a flat block of memory for the archive, this is easy to implement - but not so compatible with the rest on miniz. The solution used here is pretty ugly - that we grab the reader function by constructing a reader from a writer.


source/core/slang-zip-file-system.h
```diff
@@ -0,0 +1,35 @@
...
+public:
+
+    enum class CompressionType
+    {
+        BestSpeed,
+        BestCompression,
+    };
+
+        /// Get as an archive (that can be saved to disk)
+    virtual ArrayView<uint8_t> getArchive() = 0;
```
tfoleyNV, line 22: Should this be defined to return a "blob" that snapshots the current contents of the archive?

I was about to say that such a thing would be easy with our `Array<T>` type, but then I realized that I only changed our `String` to be copy-on-write, and not `Array<T>`.

With `String` it should be relatively easy at this point to change the heap-allocated `StringRepresentation` to support the `IBlob` interface, which would mean that creating a blob that snapshots the current state of a `String` would be a simple constant-time operation (just retain a reference to the `StringRepresnetation` and cast it to the blob interface). There'd be no extra heap allocation or copies, but then if somebody subsequently tries to mutate the original string our copy-on-write logic would automatically make a copy at that point.

If we changed `Array<T>` to use a similar implementation, where the actual `Array<T>` is a single pointer in size, and points to a heap-allocated `ArrayRepresentation` object, then we could make the operation of creating a blob from an array just as simple, and it would have the same property of being able to snapshot an immutable blob from a mutable array.

None of this is very relevant to the current PR, but it does seem a bit unsafe to be directly returning a view into the bits that make up the archive, when those bits might change with subsequent filesystem calls (or even the address of the allocation might change, right?).

source/core/slang-zip-file-system.h
```diff
@@ -0,0 +1,35 @@
...
+        /// Set the compression - used for any subsequent items added
+    virtual void setCompressionType(CompressionType type) = 0;
+
+        /// Create a zip with the contents of data/size (the contents of a zip file)
+    static SlangResult createZip(const void* data, size_t size, RefPtr<CompressedFileSystem>& out);
```
tfoleyNV, line 27: The naming here might confuse people. The `createZip` call isn't creating a zip archive or anything like that, but rather is creating a zip *file system* from the data in an archive. I guess at a call site it would be more clear, written `CompressedFileSystem::createZip`, but it would obviously be even more clear if it was written `ZipFileSystem::create()`. I assume we can't use the latter because `ZipFileSystem` is declared in a `.cpp` file to hide its implementation details.

(C++ is a bad language for the way that its `.h`/`.cpp` separation actually forces you to compromise on API design if you care about information hiding)

tfoleyNV (approved):

PR #1620, opened 2020-12-01
Make SlangCompileRequest COM type

jsmall-nvidia:
* Use slang::ICompileRequest COM interface to implement SlangCompileRequest
* Add more 'session' functionality to the ISlangGlobalSession interface
* Make where appropriate C exported functions, just call the interface methods
* m_ prefixed member variables in EndToEndCompileRequest (which implements slang::ICompileRequest)
* Made the slangc-tool use the COM based interface
* Moved comments from slang c interface to the COM method (with @see pointing to the method, so comment not written twice)

There are two things driving this change 
1) That having the request as a COM type allows for simpler invocation/IDE usage is nicer/consistency with other types
2) That the C-like export interface introduces issues with exports/static libraries

On the second point. There is a desire to be able to build a Slang static library, such that this can be used to create multiple dlls - in particular one with serialized stdlib embedded in it, and perhaps a 'bootstrap' version without it. Note that this static library does not probably want to contain the sp C like functions, because they would be embedded in the static lib *without* exporting decoration. It is not clear if there is a mechanism where just using the signature (say) with the decoration when combined with the static lib will export - or even if it does work on one compiler does it work across compilers. NOTE: There is a windows specific way around this by defining the exports in a separate file - but this is much more inconvenient than using SLANG_API magic.

One way to avoid this issue is the the Slang static lib to not contain any API c export functions. Then when we build the dll to have that as the source. In this scenario it is desirable for the export functions to really be shims - as we would like a build of Slang static lib to be the implementation of Slang as much as possible. Or put another way there is a desire to limit the surface of the C API and how it interacts with Slang internals - and the COM interface achieves this, by the COM interface not having the export issue, and additionally by allowing all the C style functionality, by just having it trivially defer to the interface.

Note that additionally we might want to have a Slang *API* static lib. This can just be the Slang static lib (+ serialized stdlib say) and the shims without dll export etc.

Some issues:

* Perhaps some methods shouldn't be part of the regular COM interface (say like the setCommandLineMode) because they aren't strictly 'public API' material - perhaps that should be accessed some other way (for example with a special separate interface)
* The C function like interface allows for a more flexible kind of binary compatibility - as functions can be added at will. COM interfaces to keep binary compatibility require significant work/cruft.


slang.h
```diff
@@ -3051,6 +2837,541 @@ namespace slang
...
+            TODO(JS): Arguably this should just return the session pointer.
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getSession(
+            slang::ISession** outSession) = 0;
+
+            /** get reflection data from a compilation request */
+        virtual SLANG_NO_THROW SlangReflection* SLANG_MCALL getReflection() = 0;
+
+            /** Make output specially handled for command line output */
+        virtual SLANG_NO_THROW void SLANG_MCALL setCommandLineCompilerMode() = 0;
```
tfoleyNV, line 3370: As discussed elsewhere, this is probably something we'd want to put into a hidden interface that is in a non-public header, so that `slangc` can perform a `queryInterface` and then toggle a given compile request into command-line mode without having users be able to do that and shoot themselves in the foot.

This still raises the issue that command-line compilation is fundamentally linked to what we call "end-to-end" compilation in the current API, and there are certain features that are only allowed in "end-to-end mode."

source/slang/slang-check-shader.cpp
```diff
@@ -1513,7 +1513,7 @@ namespace Slang
         List<Expr*> globalSpecializationArgs;
         parseSpecializationArgStrings(
             endToEndReq,
-            endToEndReq->globalSpecializationArgStrings,
+            endToEndReq->m_globalSpecializationArgStrings,
```
tfoleyNV, line 1516: Grumble: I'm not a fan of introducing a "cleanup" like this into a change that isn't really related to coding convention or cleanup work. I'm still not 100% bought in on the idea that the `m_` prefix should be something that our convention *requires* as opposed to something it merely *allows*.

tfoleyNV (approved):
As discussed separately, most of my review feedback here is based on the (mis)understanding that the `ICompileRequest` being introduced here is intended to be part of our public API in the long run.

As I now understand it, the primary goal is to minimize the number of C functions that are necessary parts of the DLL interface of Slang, and to make many of the existing C functions layer on top of the COM-lite API.

I would like it is the newly introduced interface had a name that reflects its temporary/provisional nature, but that isn't strictly needed so long as we don't end up with users talking to it directly and creating source/binary compatibility issues.

Along those lines, it would be great if the new interface went in a header that is something like `slang-internal.h` representing internal bits of our "public" API that might be used as part of the binary interface of tools within the project, but which clients should not try to depend on.

The biggest open question I have from looking at this code is how exactly we are going to break away from having all the `sp*` functions be part of the binary interface of `slang.dll`, given that existing client code fundamentally depends on those functions being present.

source/slang/slang-check-shader.cpp
jsmall-nvidia: It may be worth saying that I did not add the m_ without cause - perhaps I should have made more explicit what the cause was. With the sp functions we have asInternal produce variable called typically req, and then we access things via req->. With the move to the functionality to methods on EndToEndRequest in several methods the parameter name was the same as the member variable name - and confusion bugs ensued. 

That did not require *all* of the variables to be m_ prefixed it's true, but it seemed entirely arbitrary to have some have it and others not. There was also a variable with the other style.

slang.h
```diff
@@ -1169,143 +1168,121 @@ extern "C"
...
+    SLANG_API const char* spGetBuildTagString();
+
+    /* @see slang::IGlobalSession::createCompileRequest
     */
     SLANG_API SlangCompileRequest* spCreateCompileRequest(
         SlangSession* session);
 
     /*!
     @brief Destroy a compile request.
+    Note a request is a COM object and can be destroyed via 'Release'.
```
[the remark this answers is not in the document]
jsmall-nvidia: Um - yes it's equivalent to release(). That on spSessionDestroy we actually check (on debug) on destroy that the ref count is 1, such that calling it does always 'destroy' it. We could do that here. Or not. I think either answer is fine - but perhaps not having both behaviors...

source/slang/slang-check-shader.cpp
tfoleyNV: That makes more sense given the context. Thanks.

Honestly, if we could just agree on a `_` prefix for both internal functions and internal data, I'd be 100% on board with a sweeping cleanup of the code to change this stuff. It's that little `m` that ties me up in knots. :)

PR #1622, opened 2020-12-02
Add shader object parameter binding to renderer_test.

csyonghe:
This PR pulls in the on-going work from `tfoleyNV/render-test-name-based-binding`, and makes the change co-exists with the current (legacy) `DescriptorSet` based binding logic. To use shader object binding in `renderer_test`, add `-shaderobj` argument to 
`//TEST` line of the test, otherwise legacy binding will be used.

Reverted back to legacy binding for the following tests since the shader object binding logic is not yet fully implemented for global type parameters and interface parameters:
```
tests/compute/global-type-param2.slang
tests/compute/interface-shader-param.slang 
```

This pulls in the reflection API change for shader object binding. In the next set of changes I will move `ShaderObject` to be part of the `Renderer` API instead of building it on top of existing `DescriptorSet` abstraction, since it is not possible to implement a CUDA backend of the `Renderer` interface (which assumes bindless resources) without `ShaderObject`.


multiple-definitions.hlsl
```diff
@@ -0,0 +1,11 @@
...
+
+#line 21 "tests/bugs/multiple-definitions.slang"
+[numthreads(1, 1, 1)]
+void main()
+{
+
+#line 21
+    return;
+}
+
```
tfoleyNV, line 11: Is there a reason for this file to be checked in? It seems like maybe it is being checked in on accident.

If this is really needed by the test case, then I have to wonder why this file isn't under `test/bugs/*` and also why it needs to be checked in as plain HLSL.

tfoleyNV (approved):
I'm excited to see that you found a way to land these changes. Hopefully it will be a lot faster to iterate from here and get more of the test cases working with the new approach.

multiple-definitions.hlsl
csyonghe: Checked in by accident. removed.

tfoleyNV (dismissed):

tangent-vector (approved):

PR #1624, opened 2020-12-03
Projects in 'build' and Slang API separation

jsmall-nvidia:
* Files that contain API functions are now suffixed by '-api'
* Made premake generate project files in 'build' directory
* Explicit specifying of source directory for 'standardProject'
  * Can pass nil and let project configure how source is added

The slang-reflection.h file was removed, it had signatures in it, but none were actually implemented. The slang-reflection-api.cpp only exported functions are that of the slang relection external API.


tangent-vector (approved):
This seems reasonable.

One major issue that I'm just now remembering though is that changing the locations of our `.vcxproj` files will change the default behavior when running projects through the VS debugger. The default working directory is (IIRC) the directory with the project file, and for our examples this ends up being important for them to run correctly (because they look up files by the relative path).

I'm not clear on whether that is enough reason to change the plan here, but I am a bit worried.

tfoleyNV:
We are landing this as-is and will work through later what to do for example projects so that people can hit F5 from the Visual Studio IDE.

PR #1625, opened 2020-12-03
Add github action to verify vs project file consistency.

csyonghe:


premake5.lua
```diff
@@ -302,6 +302,16 @@ function addSourceDir(path)
         path .. "/*.hpp",       -- C++ style headers (for glslang)
         path .. "/*.natvis",    -- Visual Studio debugger visualization files
     }
+    removefiles
```
jsmall-nvidia, line 305: Can't we just use "**/*generated.h",
        "**/*generated-macro.h"
To cover ast and ref-object scenarios?

premake5.lua
```diff
@@ -302,6 +302,16 @@ function addSourceDir(path)
...
+    {
```
jsmall-nvidia, line 306: As discussed elsewhere I find having these in the project useful. Perhaps they are only removed with an option? So can run premake on local machine.

Alternatively there is logic in premake that knows what all the extractor files are, and so we could add so they are always there, even if the file is not present. 

Anyway I guess none of this should stop this PR.

jsmall-nvidia (approved):

PR #1626, opened 2020-12-03
Source level support for stdlib embedding

jsmall-nvidia:
On top of #1624

* Use the COM interface to access API in api-less-slang
* Use debugdir to make examples run without having to set the working directory
* Make slang-test run without setting working directory
* Added ability to save and load std lib
* Added ability to write out serialized stdlib as source
* Added StaticBlob type - and use for returning serialized stdlib 

This provides all the source level functionality for stdlib embedding. The remaining issues are around the embedding process via premake. An attempt was made to make that work - which did on windows, but for now that is removed because of linking issues on non windows targets.

Originally this PR had..

* Split out 'api-less-slang' project. This is all of Slang *except* client facing API
* Build slang dll using api-less-slang
* Static library version of Slang (again built using api-less-slang)

These aspects have been removed - because when building with gcc/slang linkage did not work correctly in multiple ways. With there being some kind of clash around DSO 'hidden' symbols, and not being able to link with core symbols correctly.


jsmall-nvidia:
Subsumed into #1634

PR #1633, opened 2020-12-08
Move ShaderObject to be under renderer interface.

csyonghe:
This change moves the `ShaderObject` implementation, currently sits in `render-test-main.cpp` to be part of the `Renderer` interface implementation.

The change makes existing renderers (renderer-gl, renderer-vk, renderer-d3d11, renderer-d3d12) all inherit from a shared parent class: `GraphicsAPIRenderer`, which provides the `ShaderObject` interface implementation that is layered on top of the `DescriptorSet` abstraction.

A future step is to remove the `DescriptorSet` abstraction and have all these renderers implement the `ShaderObject` API directly.


tools/gfx/render.h
```diff
@@ -938,23 +1013,23 @@ class Renderer: public Slang::RefObject
     }
 
     virtual Result createGraphicsPipelineState(
-        const GraphicsPipelineStateDesc&    desc,
+        GraphicsPipelineStateDesc&    desc,
```
tfoleyNV, line 1016: Why are the `const`s getting removed here? That seems incorrect. Creating an object shouldn't modify the `desc` that was used to create it, right?

tools/gfx/shader-cursor.h
```diff
@@ -0,0 +1,359 @@
...
+/// Given a cursor pointing to a value of some "primitive" type, we can set or get the value
+/// using operations like `setResource`, `getResource`, etc.
+///
+/// Because type information for shader parameters is being reflected dynamically, all type
+/// checking for shader cursors occurs at runtime, and errors may occur when attempting to
+/// set a parameter using a value of an inappropriate type. As much as possible, `ShaderCursor`
+/// attempts to protect against these cases and return an error `Result` or an invalid
+/// cursor, rather than allowing operations to proceed with incorrect types.
+///
+struct ShaderCursor
```
tfoleyNV, line 26: Many of the methods here probably ought to be moved into a `shader-cursor.cpp` implementation file. There is no strong reason for them to be inline.

(I realize that the current setup is a byproduct of copy-pasting them from the older location, but it is worth bringing this up because that move changed it from the class being in a `.cpp` file to it being in a `.h` file)

tfoleyNV (changes requested):
I can mostly approve this, but I don't understand why `const` got removed in several places.

If there is a good reason for that change I would be fine with this landing as-is so long as we have a plan for how to get the `const&` back in a future change.

tfoleyNV (approved):

PR #1634, opened 2020-12-08
Building with embedded stdlib

jsmall-nvidia:
On top of #1624

* Use the COM interface to access API in non slang 'API' source files
* Use debugdir to make examples run without having to set the working directory
* Make slang-test run without setting working directory
* Added ability to save and load std lib
* Added ability to write out serialized stdlib as source
* Added StaticBlob type - and use for returning serialized stdlib

The remaining effort is around actual embedding.

Embedding is currently disabled by default. To enable use 

```
premake --enable-embed-stdlib=true
```

The way the embedding occurs is by building `slangc-bootstrap` which compiles the internal Slang source along with slangc. This is then executed in a later step into slang-stdlib-generated.h which is the zip contents of the stdlib.

That client can use a pre compiled stdlib, without using embedding by using to save off the stdlib in a zip

```
-save-stdlib stdlib.zip
```

And then on command lines load the zip with

```
-load-stdlib stdlib.zip
```

Without the above options if there is an embedded stdlib, it will be used over compiling the stdlib at startup.

That disabling embedding by default is probably reasonable - because if it is enabled, any change to the Slang source will trigger a stdlib embedding, and that can be very slow. With the way this currently works - Slang source is compiled twice making it even slower. We do need to have a CI test out this path though. 

A previous effort tried to stop the compiling twice problem by using  a 'api-less-slang', but that caused many as yet unexplained linking issues.


tfoleyNV (approved):
I won't lie: I'm a little scared of what is going to happen with this change.

I'm also excited: the actual performance improvement should be good and this is the culmination of a ton of work in many parts of the compiler. We should be able to leverage the infrastructure that has been developed to make this change possible in a bunch of other places.

PR #1637, opened 2020-12-11
Implements CUDA renderer in gfx.

csyonghe:
This change adds a CUDA implementation of the `gfx::Renderer` interface, and routes CUDA tests with `-shaderobj` enabled through the standard gfx framework instead of through the `cuda-compute-util` functions.

The CUDA renderer provides only compute functionality, it does not implement any graphics pipeline features. It also only supports `ShaderObject` binding instead of the legacy `DescriptorSet` API. Samplers are not supported, but Texture and CombinedTextureSampler are supported.

This change enables two tests that is currently using this path: `array-param` which only uses global parameters, and `entry-point-uniform-params`, which uses global parameters and entry point parameters, and exercises sub-object binding via a `ConstantBuffer` parameter.

### DIFF
--- a/tools/gfx/cuda/render-cuda.cpp
+++ b/tools/gfx/cuda/render-cuda.cpp
@@ -1,0 +1,1506 @@
+#include "render-cuda.h"
+
+#ifdef GFX_ENABLE_CUDA
+#include "../render.h"
+#include <cuda.h>
+#include <cuda_runtime_api.h>
+#include "../../source/core/slang-std-writers.h"
+#include "slang.h"
+#endif
+
+namespace gfx
+{
+#ifdef GFX_ENABLE_CUDA
+using namespace Slang;
+
+SLANG_FORCE_INLINE static bool _isError(CUresult result) { return result != 0; }
+SLANG_FORCE_INLINE static bool _isError(cudaError_t result) { return result != 0; }
+
+// A enum used to control if errors are reported on failure of CUDA call.
+enum class CUDAReportStyle
+{
+    Normal,
+    Silent,
+};
+
+struct CUDAErrorInfo
+{
+    CUDAErrorInfo(
+        const char* filePath,
+        int lineNo,
+        const char* errorName = nullptr,
+        const char* errorString = nullptr)
+        : m_filePath(filePath)
+        , m_lineNo(lineNo)
+        , m_errorName(errorName)
+        , m_errorString(errorString)
+    {}
+    SlangResult handle() const
+    {
+        StringBuilder builder;
+        builder << "Error: " << m_filePath << " (" << m_lineNo << ") :";
+
+        if (m_errorName)
+        {
+            builder << m_errorName << " : ";
+        }
+        if (m_errorString)
+        {
+            builder << m_errorString;
+        }
+
+        StdWriters::getError().put(builder.getUnownedSlice());
+
+        // Slang::signalUnexpectedError(builder.getBuffer());
+        return SLANG_FAIL;
+    }
+
+    const char* m_filePath;
+    int m_lineNo;
+    const char* m_errorName;
+    const char* m_errorString;
+};
+
+#    if 1
+// If this code path is enabled, CUDA errors will be reported directly to StdWriter::out stream.
+
+static SlangResult _handleCUDAError(CUresult cuResult, const char* file, int line)
+{
+    CUDAErrorInfo info(file, line);
+    cuGetErrorString(cuResult, &info.m_errorString);
+    cuGetErrorName(cuResult, &info.m_errorName);
+    return info.handle();
+}
+
+static SlangResult _handleCUDAError(cudaError_t error, const char* file, int line)
+{
+    return CUDAErrorInfo(file, line, cudaGetErrorName(error), cudaGetErrorString(error)).handle();
+}
+
+#        define SLANG_CUDA_HANDLE_ERROR(x) _handleCUDAError(_res, __FILE__, __LINE__)
+
+#    else
+// If this code path is enabled, errors are not reported, but can have an assert enabled
+
+static SlangResult _handleCUDAError(CUresult cuResult)
+{
+    SLANG_UNUSED(cuResult);
+    // SLANG_ASSERT(!"Failed CUDA call");
+    return SLANG_FAIL;
+}
+
+static SlangResult _handleCUDAError(cudaError_t error)
+{
+    SLANG_UNUSED(error);
+    // SLANG_ASSERT(!"Failed CUDA call");
+    return SLANG_FAIL;
+}
+
+#        define SLANG_CUDA_HANDLE_ERROR(x) _handleCUDAError(_res)
+#    endif
+
+#    define SLANG_CUDA_RETURN_ON_FAIL(x)              \
+        {                                             \
+            auto _res = x;                            \
+            if (_isError(_res))                       \
+                return SLANG_CUDA_HANDLE_ERROR(_res); \
+        }
+#    define SLANG_CUDA_RETURN_WITH_REPORT_ON_FAIL(x, r)                               \
+        {                                                                             \
+            auto _res = x;                                                            \
+            if (_isError(_res))                                                       \
+            {                                                                         \
+                return (r == CUDAReportStyle::Normal) ? SLANG_CUDA_HANDLE_ERROR(_res) \
+                                                      : SLANG_FAIL;                   \
+            }                                                                         \
+        }
+
+#    define SLANG_CUDA_ASSERT_ON_FAIL(x)           \
+        {                                          \
+            auto _res = x;                         \
+            if (_isError(_res))                    \
+            {                                      \
+                SLANG_ASSERT(!"Failed CUDA call"); \
+            };                                     \
+        }
+
+#    ifdef RENDER_TEST_OPTIX
+
+static bool _isError(OptixResult result) { return result != OPTIX_SUCCESS; }
+
+#        if 1
+static SlangResult _handleOptixError(OptixResult result, char const* file, int line)
+{
+    fprintf(
+        stderr,
+        "%s(%d): optix: %s (%s)\n",
+        file,
+        line,
+        optixGetErrorString(result),
+        optixGetErrorName(result));
+    return SLANG_FAIL;
+}
+#            define SLANG_OPTIX_HANDLE_ERROR(RESULT) _handleOptixError(RESULT, __FILE__, __LINE__)
+#        else
+#            define SLANG_OPTIX_HANDLE_ERROR(RESULT) SLANG_FAIL
+#        endif
+
+#        define SLANG_OPTIX_RETURN_ON_FAIL(EXPR)           \
+            do                                             \
+            {                                              \
+                auto _res = EXPR;                          \
+                if (_isError(_res))                        \
+                    return SLANG_OPTIX_HANDLE_ERROR(_res); \
+            } while (0)
+
+void _optixLogCallback(unsigned int level, const char* tag, const char* message, void* userData)
+{
+    fprintf(stderr, "optix: %s (%s)\n", message, tag);
+}
+
+#    endif
+
+class MemoryCUDAResource : public BufferResource
+{
+public:
+    MemoryCUDAResource(const Desc& _desc)
+        : BufferResource(_desc)
+    {}
+
+    ~MemoryCUDAResource()
+    {
+        if (m_cudaMemory)
+        {
+            SLANG_CUDA_ASSERT_ON_FAIL(cudaFree(m_cudaMemory));
+        }
+    }
+
+    uint64_t getBindlessHandle() { return (uint64_t)m_cudaMemory; }
+
+    void* m_cudaMemory = nullptr;
+};
+
+class TextureCUDAResource : public TextureResource
+{
+public:
+    TextureCUDAResource(const TextureResource::Desc& desc)
+        : TextureResource(desc)
+    {}
+    ~TextureCUDAResource()
+    {
+        if (m_cudaSurfObj)
+        {
+            SLANG_CUDA_ASSERT_ON_FAIL(cuSurfObjectDestroy(m_cudaSurfObj));
+        }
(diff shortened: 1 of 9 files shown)


tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+    struct BindingRangeInfo
+    {
+        slang::BindingType bindingType;
+        Index count;
+        Index baseIndex;
+        Index descriptorSetIndex;
+        Index rangeIndexInDescriptorSet;
+        Index uniformOffset;
+        //        Index   subObjectRangeIndex = -1;
+    };
```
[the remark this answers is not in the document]
tfoleyNV: Scanning ahead, I see that `baseIndex` is that I referred to as `flatIndex` in the previous comment, so that field is the other one that seemingly should stay.

It then seems like `descriptorSetIndex`, `rangeIndexInDescriptorSet`, and `uniformOffset` are *probably* redundant? I guess maybe `uniformOffset` is maybe useful.

(One thing to note is that there is no strong reason to store information in `m_bindingRanges` here that is redundant with what can be queried on `typeLayout`.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+
+            case slang::BindingType::VaryingOutput:
+                baseIndex = m_varyingOutputCount;
+                m_varyingOutputCount += count;
+                break;
+
+            default:
+                baseIndex = m_resourceViewCount;
+                m_resourceViewCount += count;
+                break;
```
tfoleyNV, line 361: The rationale for this kind of `switch` statement in a `ShaderObjectLayout` implementation is that it handles the N-to-M mapping from the N `slang::BindingType` cases that binding ranges use in Slang reflection data over to the M distinguished cases that a particular API cares about.

E.g., for a pure D3D11 implementation, the main cases would be CBV, SRV, UAV, and Sampler. Vulkan would have a much higher number of cases, since it recognizes a greater number of cases for descriptor types.

In the CUDA case, a lot of these binding types are either expected to never occur (e.g., `VaryingInput`) or don't really matter when they do (`Sampler`).

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+            subObjectRanges.add(subObjectRange);
+        }
+    }
+};
+
+class CUDAProgramLayout : public CUDAShaderObjectLayout
+{
+public:
+    slang::ProgramLayout* programLayout = nullptr;
+    List<RefPtr<CUDAShaderObjectLayout>> entryPointLayouts;
```
tfoleyNV, line 411: It might be worth keeping this as a `List` for generality (especially when we add OptiX support into this path), but for the pure CUDA compute case it seems redundant to have multiple entry points here.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+            layout->m_bindingRanges[offset.bindingRangeIndex].baseIndex + offset.bindingArrayIndex;
+        SLANG_ASSERT(
+            offset.uniformOffset ==
+            layout->m_bindingRanges[offset.bindingRangeIndex].uniformOffset +
+                offset.bindingArrayIndex * sizeof(void*));
+        auto cudaObject = dynamic_cast<CUDAShaderObject*>(object);
+        if (subObjectIndex >= objects.getCount())
+            objects.setCount(subObjectIndex + 1);
+        objects[subObjectIndex] = cudaObject;
+        return setData(offset, &cudaObject->bufferResource->m_cudaMemory, sizeof(void*));
```
tfoleyNV, line 494: I honestly don't know the answer to this: are we guaranteed that the pointer size always matches between a CUDA device and the host program? Are there cases where we could have a 32-bit app and 64-bit device pointers (or vice versa)?

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+    {
+        auto subObjectIndex =
+            layout->m_bindingRanges[offset.bindingRangeIndex].baseIndex + offset.bindingArrayIndex;
+        SLANG_ASSERT(
+            offset.uniformOffset ==
+            layout->m_bindingRanges[offset.bindingRangeIndex].uniformOffset +
+                offset.bindingArrayIndex * sizeof(void*));
+        auto cudaObject = dynamic_cast<CUDAShaderObject*>(object);
+        if (subObjectIndex >= objects.getCount())
+            objects.setCount(subObjectIndex + 1);
```
tfoleyNV, line 492: This step shouldn't be needed, right? We should be able to set the size of the `resources` array to `layout->m_resourceViewCount` at the time we create a `ShaderObject`, rather than try to dynamically resize stuff.

There's a certain amount of defensive code here trying to handle the case of the arrays being the wrong size, but it seems better to try and allocate the arrays at the exact *right* size at creation time.

(If we ever want to drill deep into performance stuff, we could switch to a model where the arrays of bindings on a `ShaderObject` are tail-allocated as part of the same memory as the object itself, so that we can use a single heap allocation for any object. In that case, the `ShaderObjectLayout` could pre-compute the byte size required for objects with that layout (including tail allocation), as well as byte offsets for each resource range.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+        SLANG_UNUSED(offset);
+        return SLANG_OK;
+    }
+    virtual SlangResult setCombinedTextureSampler(
+        ShaderOffset const& offset, ResourceView* textureView, SamplerState* sampler)
+    {
+        SLANG_UNUSED(sampler);
+        setResource(offset, textureView);
+        return SLANG_OK;
+    }
```
tfoleyNV, line 542: This case is the tricky one. If we want CUDA to seamlessly support combined texture/sampler views in this way we need to account for the combined texture/sampler pairs when computing the `ShaderObjectLayout`.

I guess in this case it shouldn't be too bad, but we'd need to make sure that the combined case also increments `m_resourceViewCount` and uses it in computing the `baseIndex`. Right now I think the logic in `ShaderObjectLayout` is acting like the combined case is using completely separate arrays.

(One longer-term question here is whether the combined case should somehow be creating and/or caching a CUDA surface object that combines the texture and sampling state provided here. It seems incorrect to just ignore the `sampler` in this case.)

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+
+        // Copy global parameter data to the `SLANG_globalParams` symbol.
+        {
+            CUdeviceptr globalParamsSymbol = 0;
+            size_t globalParamsSymbolSize = 0;
+            cuModuleGetGlobal(
+                &globalParamsSymbol,
+                &globalParamsSymbolSize,
+                currentPipeline->shaderProgram->cudaModule,
+                "SLANG_globalParams");
```
tfoleyNV, line 1211: Note that a lot of the data here is stuff that should eventually be cached on the `shaderProgram` and/or `programLayout`.

E.g., there is no reason to call `cuModuleGetGlobal` on every launch, since the address of that symbol should not change for the lifetime of `shaderProgram`.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+                (void*)globalParamsCUDAData,
+                globalParamsSymbolSize,
+                cudaMemcpyDeviceToDevice,
+                0);
+        }
+        //
+        // In the case of the entry-point parameters, we have to deal with
+        // two different wrinkles.
+        //
+        // First, the `bindRoot` will have the entry-point argument data
```
tfoleyNV, line 1228: Note that some of the comments are now stale, with references to stuff like `bindRoot` from the Old Way of doing things.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+        void* entryPointHostData = nullptr;
+        if (entryPointDataSize)
+        {
+            entryPointHostData = alloca(entryPointDataSize);
+            cudaMemcpy(
+                entryPointHostData,
+                (void*)entryPointBuffer->getBindlessHandle(),
+                entryPointDataSize,
+                cudaMemcpyDeviceToHost);
+        }
```
tfoleyNV, line 1245: Note that one option we could consider now that `ShaderObject` is under the API boundary is to have two cases of CUDA `ShaderObject`s: an ordinary `CUDAShaderObject` for the 99% case and a `CUDAEntryPointShaderObject` to handle the entry-point case. The latter would differ only in that it stores its data in host memory instead of device memory.

By having a distinct type for CUDA entry point shader objects, we could avoid the allocation and device->host copy here.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+    virtual void presentFrame() override {}
+    virtual TextureResource::Desc getSwapChainTextureDesc() override
+    {
+        return TextureResource::Desc();
+    }
+    
+    virtual Result createSamplerState(SamplerState::Desc const& desc, SamplerState** outSampler) override
+    {
+        SLANG_UNUSED(desc);
+        *outSampler = nullptr;
```
tfoleyNV, line 1319: Should the CUDA implementation allocate and return a kind of placeholder sampler object, with no actual functionality?

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+        // cannot pre-allocate the object(s) to go into that
+        // range, since we can't possibly know what to allocate
+        // at this point.
+        //
+        if (!subObjectLayout)
+            continue;
+        //
+        // Otherwise, we will allocate a sub-object to fill
+        // in each entry in this range, based on the layout
+        // information we already have.
```
tfoleyNV, line 1471: I'm not 100% certain, but it feels like we'd eventually want to move the automatic allocation of sub-objects into a higher layer that works across all the `ShaderObject` implementations, since it is mostly a utility for applications that don't want to have to think about explicitly managing this stuff.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -0,0 +1,1506 @@
...
+        return SLANG_OK;
+    }
+
+    virtual Result createRootShaderObject(ShaderObjectLayout* layout, ShaderObject** outObject) override
+    {
+        RefPtr<CUDARootShaderObject> result = new CUDARootShaderObject();
+        SLANG_RETURN_ON_FAIL(result->init(this, dynamic_cast<CUDAShaderObjectLayout*>(layout)));
+        *outObject = result.detach();
+        return SLANG_OK;
+    }
```
tfoleyNV, line 1140: I may have missed it in the previous change that refactored the `ShaderObject` API to be part of `Renderer`, but it seems like we may want to think about tweaking the "root shader object" idea in a few ways:

* It seems like a `Program` that is being loaded/created from Slang could eventually store the `slang::ProgramLayout` directly, at which point any API that currently takes a `slang::ProgramLayout` could just take a `gfx::Program`.

* It seems like it is redundant to have a `createRootShaderObjectLayout` step, because if a `gfx::Program` stores the `slang::ProgramLayout`, then it should also be able to create and store the `gfx`-level "root layout."

* At that point `createRootShaderObject` can just take the `gfx::Program` instead of a `ShaderObjectLayout`.

* At that point, it makes sense that `createRootShaderObject` returns a distinguished sub-type of `ShaderObject` to represent the fact that it is representing a `ProgramObject` or `RootShaderObject` or whatever we want to call it.

* If a `ComputePipelineState` only stores a `Program` (since the layout is now determined by the program), there is no reason to have `ComputePipelineState`.

* If a `RootShaderObject`/`ProgramObject` is always allocated from a `Program` then, binding a root shader object already determined the program to use, so that there need not be distinct calls to set the program and the root shader object.

* If a `RootShaderObject`/`ProgramObject` fully determines the program and the argument data to use for  dispatch, then it seems reasonable to have an overload of `dispatchCompute` that directly takes the root shader object to use for the dispatch, and is thus completely stateless.

One additional longer-term note is that somebody might look at the current `gfx` abstraction and argue that it is inefficient because you end up having to bind all the "root" parameters into an explicit root `ShaderObject`, and *then* bind it with `bindRootShaderObject`, which has the effect of copying state from the `ShaderObject` over to the "real" pipeline/context state (e.g., the copying of data from a `ShaderObject` to the `SLANG_globalParams`, which is equivalent to copying bindings into root parameters).

What I would advocate as a potential long-term solution there is to have an interface where intead of binding a specific root `ShaderObject` (that would have already been filled in), the application binds a specific `Program` and the operation returns a "transient root shader object":

```
RootShaderObject* bindTransientRootShaderObject(Program* program);
```

The user-facing constraint for a "transient root shader object" would be that as soon as another root program/object gets bound, the transient object is no longer valid to use, and is subject to cleanup on the GPU timeline.

Because of that constraint, the implementation of a transient root shader object can know/assume that it is currently bound to a particular context/command-buffer, and that there is no need to persist its state after it get unbound. A dedicated root shader object implementation could thus map operations like `setResource` directly to low-level graphics API calls that bind things to root parameters.

Implementing a dedicated "transient root shader object" abstraction would allow this API layer to potentially get some of the benefits of the API approach advocated for in the "shader components" paper.

tfoleyNV (approved):
Looks pretty good. I'm interested to see what we can iterate toward with this foundation in place.

It seems like the next major priority on this push will be:

* Add a CPU path
* Add minimal ray tracing to this stuff (initially optix-only)
* Slowly port all our existing test cases to use the new mode
* Remove the old paths from the `Renderer` abstraction once all the graphics-API tests are ported
* Remove the old CUDA/CPU paths once all their tests are ported

That's a lot of work, but I'm happy to see us digging into it.

Once everything is using a single consistent name-based binding approach, it should be a lot easier to consider an overhaul of the `TEST_INPUT` system for specifying the values of all these parameters.

Issue #1638, opened 2020-12-11
Heterogeneous flag hiding errors


Checkmate50:
Currently, the `-heterogeneous` flag attempts to write the bytecode blob of a program to to C++ (if possible).  However, if compilation fails, the error produced is the internal error:

`(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unexpected: No blob to emit`

This error message should probably be made more descriptive, and at least give an indication of what the compilation error is.

tangent-vector:
I suspect the big fix would be to not diagnose an internal compiler error in that case. Diagnosing an internal error triggers logic in the compiler that tries to exit as quickly as possible (kind of like a kernel panic), while an "ordinary" error would just append to the list of diagnostics and keep going. In the non-internal error case, all the preceding error messages would still show up in the output.

I suspect there are two main options here:

1. We could try to ensure that all the kernels that would be embedded in the output C++ code are generated *before* we go to generate C++ output, and then have logic to bail out of compilation if any of them produced an error. This option might be tricky given that we don't currently control the relative order of the targets, and thus can't easily guarantee that DXBC output gets written before CPU C++.

2. We could simply write defensive code in the C++ output case, byt replacing the internal error with an ordinary error message that says more or less the same thing ("No generated code was found for kernel X in format Y; this may be because of another compilation error") and then make sure that we conditionalize the logic that required the blob so that it produces some kind of valid workaround code when the blob isn't present.

Checkmate50:
Sounds good; I think option (2) makes sense to me, I'll start working on implementing that.  It should be straightforward to just output the same C++ code without the blob (which obviously won't compile), but an error at the end of the output should make that clear.

PR #1642, opened 2020-12-15
Heterogeneous Flag Error Visibility

Checkmate50:
PR to fix issue #1638.  This change introduces a diagnostic sink to the
emitModule function, and updates all associated calls to that function.
Additionally, this commit updates the heterogeneous hello world example
to not need the entry and stage flags for simplicity.

### DIFF
--- a/source/slang/slang-emit-cpp.cpp
+++ b/source/slang/slang-emit-cpp.cpp
@@ -2589,25 +2589,24 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module, DiagnosticSink* sink)
                             default:
 
                                 auto targetProgram = program->getTargetProgram(targetRequest);
-                                //DiagnosticSink sink(linkage->getSourceManager());
                                 CompileResult result =
                                     targetProgram->getOrCreateEntryPointResult(index, sink);
 
                                 Slang::ComPtr<ISlangBlob> blob;
                                 if (SLANG_FAILED(result.getBlob(blob)))
                                 {
-                                    StringBuilder builder;
-                                    sink->diagnoseRaw(Severity::Error, UnownedStringSlice
-                                        ("Slang heterogeneous error: No blob to emit\n"));
+                                    sink->diagnoseRaw(Severity::Error,
+                                        "Slang heterogeneous error: No blob to emit\n");
                                     m_writer->emit("size_t __");
                                     m_writer->emit(entryPointName);
                                     m_writer->emit("Size = 0;\n");
                                     m_writer->emit("unsigned char __");
                                     m_writer->emit(entryPointName);
-                                    m_writer->emit("[];\n");
+                                    m_writer->emit("[1];\n");
                                 }
 
-                                else {
+                                else
+                                {
                                     auto ptr = (const unsigned char*)blob->getBufferPointer();
 
                                     m_writer->emit("size_t __");
@@ -2619,11 +2618,18 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module, DiagnosticSink* sink)
                                     m_writer->emit("unsigned char __");
                                     m_writer->emit(entryPointName);
                                     m_writer->emit("[] = {");
-                                    for (unsigned int i = 0; i < blob->getBufferSize() - 1; i++) {
+                                    // every 20 bytes, emit a newline
+                                    size_t j = 0;
+                                    for (size_t i = 0; i < blob->getBufferSize(); i++) {
                                         m_writer->emitUInt64(ptr[i]);
                                         m_writer->emit(", ");
+                                        if (j == 20)
+                                        {
+                                            m_writer->emit("\n");
+                                            j = 0;
+                                        }
+                                        j++;
                                     }
-                                    m_writer->emitUInt64(ptr[blob->getBufferSize() - 1]);
                                     m_writer->emit("};\n");
                                 }
                             }
--- a/source/slang/slang-emit-cpp.cpp
+++ b/source/slang/slang-emit-cpp.cpp
@@ -2553,7 +2553,7 @@ void CPPSourceEmitter::_emitForwardDeclarations(const List<EmitAction>& actions)
     }
 }
 
-void CPPSourceEmitter::emitModuleImpl(IRModule* module)
+void CPPSourceEmitter::emitModuleImpl(IRModule* module, DiagnosticSink* sink)
 {
     // If we are emitting a heterogeneous program
     // Emit the binary blob of each non-CPP target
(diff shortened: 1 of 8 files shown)


source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
                             default:
 
                                 auto targetProgram = program->getTargetProgram(targetRequest);
-                                DiagnosticSink sink(linkage->getSourceManager());
+                                //DiagnosticSink sink(linkage->getSourceManager());
```
jsmall-nvidia, line 2592: We probably just want to remove this line, as it doesn't seem to add any future value.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
                                 CompileResult result =
-                                    targetProgram->getOrCreateEntryPointResult(index, &sink);
+                                    targetProgram->getOrCreateEntryPointResult(index, sink);
 
                                 Slang::ComPtr<ISlangBlob> blob;
                                 if (SLANG_FAILED(result.getBlob(blob)))
                                 {
-                                    SLANG_UNEXPECTED("No blob to emit");
-                                    return;
+                                    StringBuilder builder;
```
jsmall-nvidia, line 2599: The builder here doesn't look like it's used. It *might* be more convenient to build up your result using 

```
builder << "size_t __" << entryPointName << "Size = 0;\n"; 
```

etc, and just write out to writer at the end.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
+                                    sink->diagnoseRaw(Severity::Error, UnownedStringSlice
```
jsmall-nvidia, line 2600: The use of UnownedStringSlice ctor works, but internally does a strlen. If you wanted to avoid that, you can do 

```
UnownedStringSlice::fromLiteral("Some literal");
```

And it doesn't need to do the strlen - it can compile time determine length. There's not much in this I'll grant you - but it seems that there is already a diagnoseRaw overload that just takes a const char*, which is slightly less unsightly to invoke.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
-                                m_writer->emitInt64(blob->getBufferSize());
-                                m_writer->emit(";\n");
-
-                                m_writer->emit("unsigned char __");
-                                m_writer->emit(entryPointName );
-                                m_writer->emit("[] = {");
-                                for (unsigned int i = 0; i < blob->getBufferSize() - 1; i++) {
-                                    m_writer->emitUInt64(ptr[i]);
-                                    m_writer->emit(", ");
+                                else {
```
jsmall-nvidia, line 2610: We have opening { on next line.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
-                                    return;
+                                    StringBuilder builder;
+                                    sink->diagnoseRaw(Severity::Error, UnownedStringSlice
+                                        ("Slang heterogeneous error: No blob to emit\n"));
+                                    m_writer->emit("size_t __");
+                                    m_writer->emit(entryPointName);
+                                    m_writer->emit("Size = 0;\n");
+                                    m_writer->emit("unsigned char __");
+                                    m_writer->emit(entryPointName);
+                                    m_writer->emit("[];\n");
```
jsmall-nvidia, line 2607: Does the line with just [] - actually work? It seems like it looks like a forward declaration or a variable, but not the actual declaration that will need to include the size. I think you want [1] there.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
+                                    m_writer->emit("size_t __");
+                                    m_writer->emit(entryPointName);
+                                    m_writer->emit("Size = ");
+                                    m_writer->emitInt64(blob->getBufferSize());
+                                    m_writer->emit(";\n");
+
+                                    m_writer->emit("unsigned char __");
+                                    m_writer->emit(entryPointName);
+                                    m_writer->emit("[] = {");
+                                    for (unsigned int i = 0; i < blob->getBufferSize() - 1; i++) {
```
jsmall-nvidia, line 2622: unsigned int should presumably be size_t?

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
+                                        m_writer->emitUInt64(ptr[i]);
```
jsmall-nvidia, line 2623: That here I don't think we want the -1? Oh okay maybe you have that so there isn't a hanging , at the end. You don't need to worry about this though as C/C++ allow this extra comma (just for these kinds of situations). 
Also note special handling means that if getBufferSize() == 0, would be a bug/crash. That the blob being 0 sized arguably doesn't make much sense here in practice but, the simpler loop with a , at the end would not have this issue.

```
// To avoid repeatedly calling getBufferSize in loop (the C++ compiler can't generally see into blob implementation, that the value will not change).
const size_t bufferSize = blob->getBufferSize();
for (size_t i = 0; i < bufferSize; ++i)
{
   // If I don't cast then it won't find the overload unambiguously (arguably we should have some explicit sized methods)
  // But I don't want to use emitUInt64 because that's likely more unnecessary work 
   m_writer->emit(int(ptr[i]));
}
```

source/slang/slang-emit-cpp.cpp
```diff
@@ -2589,34 +2589,43 @@ void CPPSourceEmitter::emitModuleImpl(IRModule* module)
...
+                                        m_writer->emit(", ");
+                                    }
+                                    m_writer->emitUInt64(ptr[blob->getBufferSize() - 1]);
+                                    m_writer->emit("};\n");
```
jsmall-nvidia, line 2627: As a suggestion it might be simpler to have a function that handled this writing out that could be used for either case. It would need a bit of special case code around the 0 size (for failed getting of a blob say), but overall should cut down the code. Say something like...

```
void writeBlob(const void* data, size_t dataCount, const char* entryPointName, SourceWriter* writer)
```

That the function would have to test if the size is 0, and still output [1] for the data.

You might also want to output a cr every n bytes - as seen in HexDumpUtil::dumpSourceBytes.

jsmall-nvidia (changes requested):
That much of this is suggested improvements, not requirements. I think that outputting [1] for the empty case is required.

Checkmate50:
Thanks for the suggestions! I think you're right on all points, I'll make the above changes.

tfoleyNV (approved):

PR #1644, opened 2020-12-15
Cleanup CUDA renderer.

csyonghe:
- Remove unnecesary fields kept in `BindingRangeInfo`.
- `CUDARootShaderObject` stores uniform buffer in host memory so we no longer need the device->host copy before launching a kernel.


tfoleyNV (approved):
These are nice improvements to the clarity of the code.

It is really nice to see that the new `CUDAEntryPointObject` eliminates a GPU->CPU copy without really adding any significant complexity.

PR #1649, opened 2021-01-05
Refactor GUI/Window utils out of gfx library

csyonghe:
This change moves the GUI and windowing logic out of the `gfx` library, so that all remains in `gfx` is for abstracting over GPU platforms only. This is a necessary step to make `gfx` an abstract layer that can be used outside the Slang testing framework.


tfoleyNV (approved):
This is fine, insofar as it is mainly a refactoring change.

I might argue for something more short-and-sweet like `platform` as the name for the library since it's main job is to abstract over platform-specific facilities. Eventually it might make sense for certain things like file I/O to move out of `core` and into `platform` so that we have a bit more of a clean distinction there (or alternatively just lump *everything* into `core` and have done).

PR #1656, opened 2021-01-13
COM-ify all slang-gfx interfaces.

csyonghe:
This change follows #1653, and makes all interfaces provided by `slang-gfx`  comply to COM standard. This is a required step to compile gfx library into a DLL.

This change doesn't actually change the compiler settings to compile gfx into a DLL. The remaining work is to fix the API around `Surface` so `captureScreenSurface` doesn't lead to allocation across DLL boundary.


multiple-definitions.hlsl
```diff
@@ -0,0 +1,11 @@
+#pragma pack_matrix(column_major)
```
tfoleyNV: This seems like an accidental file addition.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -222,16 +225,32 @@ class TextureCUDAResource : public TextureResource
     CUmipmappedArray m_cudaMipMappedArray = CUmipmappedArray();
 };
 
-class CUDAResourceView : public ResourceView
+class CUDAResourceView : public IResourceView, public RefObject
```
tfoleyNV, line 228: Is the inheritance from `RefObject` on this stuff just for the convenience of not having to hand-write the reference-counting?

It seems like we eventually want to have some kind of `ComponentImpl<IWhatever>` base type that can be inherited from that handles all the boring bits of COM if possible. One of the key benefits of using that idiom would be that you don't end up with multiple inheritance and thus with every object containing multiple virtual function table pointers.

(The base type could even be `ComponentImpl<DerivedType, IWhatever>` if that gets rid of more implementation cruft)

tools/gfx/render.h
```diff
@@ -1,28 +1,36 @@
...
 namespace gfx {
 
-using Slang::RefObject;
-using Slang::RefPtr;
-using Slang::Dictionary;
-using Slang::List;
-
-using Slang::getHashCode;
-using Slang::combineHash;
+using Slang::ComPtr;
```
tfoleyNV, line 33: We may eventually want to make the dependency on `Slang::ComPtr` optional, but we can handle that down the line.

tools/gfx/render.h
```diff
@@ -277,36 +261,48 @@ class Resource: public Slang::RefObject
...
-        /// For a usage gives the required binding flags
-    static const BindFlag::Enum s_requiredBinding[];    /// Maps Usage to bind flags required
-
-    protected:
-    Resource(Type type):
-        m_type(type)
-    {}
-
-    static void compileTimeAsserts();
+    inline static BindFlag::Enum getDefaultBindFlagsFromUsage(IResource::Usage usage)
```
tfoleyNV, line 264: It seems like `BindFlag` and `IResource::Usage` are overlapping implementations of the same basic idea. I suspect they were added by two different people who had differing ideas about how all this stuff should work.

Once we get to refactoring, I would like it if we could drop down to just having one or the other.

tools/gfx/render.h
```diff
@@ -315,33 +311,53 @@ class BufferResource: public Resource
...
+    {                                                                                  \
+        0x1b274efe, 0x5e37, 0x492b, { 0x82, 0x6e, 0x7e, 0xe7, 0xe8, 0xf5, 0xa4, 0x9b } \
     }
 
-    protected:
-    Desc m_desc;
-};
+template <typename T> T _slang_gfx_max(T v0, T v1) { return v0 > v1 ? v0 : v1; }
+
+static inline unsigned int _slang_gfx_ones32(unsigned int x)
```
tfoleyNV, line 334: Having this stuff in the header sure comes across like a code smell. It is a sign that we have too much logic in the `Desc` types.

Hypothetically, we could eventually split things so that there is `gfx.h` which is the minimal header that just defines the COM interface-y stuf, and then `gfxu.h/.cpp` which is "user"/utility code that is meant to be directly added to your project (similar to D3DX, but distributed as source).

tools/gfx/render.h
```diff
@@ -366,9 +382,38 @@ class TextureResource: public Resource
...
             /// Given a size, calculates the size at a mip level
-        Size calcMipSize(int mipLevel) const;
+        Size calcMipSize(int mipLevel) const
+        {
+            Size size;
+            size.width = calcMipSize(this->width, mipLevel);
+            size.height = calcMipSize(this->height, mipLevel);
+            size.depth = calcMipSize(this->depth, mipLevel);
+            return size;
+        }
```
tfoleyNV, line 416: This stuff definitely doesn't seem like public API.

tools/gfx/render.h
```diff
@@ -860,223 +1083,229 @@ class IRenderer: public ISlangUnknown
...
         // Will return with SLANG_E_NOT_AVAILABLE if NVAPI can't be initialized and nvapiExtnSlot >= 0
-    virtual SLANG_NO_THROW SlangResult SLANG_MCALL initialize(const Desc& desc, void* inWindowHandle) = 0;
+    virtual SLANG_NO_THROW Result SLANG_MCALL initialize(const Desc& desc, void* inWindowHandle) = 0;
+
+    virtual SLANG_NO_THROW bool SLANG_MCALL hasFeature(const char* feature) = 0;
 
-    bool hasFeature(const Slang::UnownedStringSlice& feature) { return getFeatures().indexOf(Slang::String(feature)) != Slang::Index(-1); }
-    virtual SLANG_NO_THROW const Slang::List<Slang::String>& SLANG_MCALL getFeatures() = 0;
+        /// Returns a list of features supported by the renderer.
+    virtual SLANG_NO_THROW Result SLANG_MCALL getFeatures(const char** outFeatures, UInt bufferSize, UInt* outFeatureCount) = 0;
```
tfoleyNV, line 1098: Couldn't this be defined so that it doesn't need the `bufferSize`? That is, we could conceivably return a pointer to storage allocated when the `Renderer` was created that will have the same lifetime as the renderer, since the list of supported features is not going to change after initialization.

tools/gfx/render.h
```diff
@@ -1113,37 +1342,36 @@ class IRenderer: public ISlangUnknown
...
 {
-    typedef SlangResult (*CreateFunc)(IRenderer** outRenderer);
+    typedef SlangResult(SLANG_MCALL * SGRendererCreateFunc)(IRenderer** outRenderer);
 
-        /// Gets the size in bytes of a Format type. Returns 0 if a size is not defined/invalid
-    SLANG_FORCE_INLINE static size_t getFormatSize(Format format) { return s_formatSize[int(format)]; }
-        /// Given a renderer type, gets a projection style
-    static ProjectionStyle getProjectionStyle(RendererType type);
+    /// Gets the size in bytes of a Format type. Returns 0 if a size is not defined/invalid
+    SLANG_GFX_API size_t SLANG_MCALL gfxGetFormatSize(Format format);
```
tfoleyNV, line 1357: This is a good candidate for a function that should be in a utility library rather than part of the DLL interface.

tools/gfx/render.h
```diff
@@ -1113,37 +1342,36 @@ class IRenderer: public ISlangUnknown
...
 
-        /// Given the projection style returns an 'identity' matrix, which ensures x,y mapping to pixels is the same on all targets
-    static void getIdentityProjection(ProjectionStyle style, float projMatrix[16]);
+    /// Gets the binding style from the type
+    SLANG_GFX_API BindingStyle SLANG_MCALL gfxGetBindingStyle(RendererType type);
 
-        /// Get the binding style from the type
-    static BindingStyle getBindingStyle(RendererType type) { return s_rendererTypeToBindingStyle[int(type)]; }
+    /// Given a renderer type, gets a projection style
+    SLANG_GFX_API ProjectionStyle SLANG_MCALL gfxGetProjectionStyle(RendererType type);
```
tfoleyNV, line 1363: I am not a fan of these functions, and I'm not convinced we ever should have added them. If we can find a way to remove them as part of the future refactoring work, that would be ideal.

tools/gfx/render.h
```diff
@@ -1113,37 +1342,36 @@ class IRenderer: public ISlangUnknown
...
 
-        /// Get as text
-    static Slang::UnownedStringSlice toText(RendererType type);
+    /// Given the projection style returns an 'identity' matrix, which ensures x,y mapping to pixels
+    /// is the same on all targets
+    SLANG_GFX_API void SLANG_MCALL
+        gfxGetIdentityProjection(ProjectionStyle style, float projMatrix[16]);
```
tfoleyNV, line 1368: If we end up needing to keep this one, it should be moved out to utility code (not part of DLL interface)

tools/gfx/render.h
```diff
@@ -1113,37 +1342,36 @@ class IRenderer: public ISlangUnknown
...
 
-        /// Given a type returns a function that can construct it, or nullptr if there isn't one
-    static CreateFunc getCreateFunc(RendererType type);
+    /// Get the name of the renderer
+    SLANG_GFX_API const char* SLANG_MCALL gfxGetRendererName(RendererType type);
```
tfoleyNV, line 1371: This seems like it should obviously be part of an eventual interface that is used to represent different implementations o the graphics layer (the next level above "adapter", but below what VK calls a "loader").

tools/gfx/render.h
```diff
@@ -1113,37 +1342,36 @@ class IRenderer: public ISlangUnknown
...
 
-    private:
-    static void compileTimeAsserts();
-    static const uint8_t s_formatSize[]; // Maps Format::XXX to a size in bytes;
-    static const BindingStyle s_rendererTypeToBindingStyle[];           ///< Maps a RendererType to a BindingStyle
-};
+    /// Given a type returns a function that can construct it, or nullptr if there isn't one
+    SLANG_GFX_API SGRendererCreateFunc SLANG_MCALL gfxGetCreateFunc(RendererType type);
+}
```
tfoleyNV, line 1375: It seems like we eventually want a more nuanced way to enumerate and find implementations of the graphics layer, but this works for now.

One thing I kind of have to wonder is: why have a function that returns a function pointer, that you then have to call? Why not just have a function that creates/returns a *thing* that you can work with?

tfoleyNV (changes requested):
Most of this is fine by me, but I notice that the stdlib `.zip` file seems to be added as part of the PR, and that shouldn't be allowed into the repository!

tools/gfx/render.h
csyonghe: This is just what it is now, and this PR doesn't try to change anything meaningfully.

tfoleyNV: I agree with the goals, and none of this nit-picks need to be addressed in this PR (the added `.zip` file does, but I see you've fixed that as I'm typing).

tfoleyNV (approved):

PR #1661, opened 2021-01-19
Make  `ShaderCursor` no longer depend on core.

csyonghe:
This change removes the `ShaderCursor`'s dependency on `core`, so that it may be used by users who do not wish to use`core`.

### DIFF
--- a/tools/render-test/render-test-main.cpp
+++ b/tools/render-test/render-test-main.cpp
@@ -190,13 +190,13 @@ SlangResult _assignVarsFromLayout(
             return SLANG_E_INVALID_ARG;
         }
 
-        auto entryCursor = rootCursor.getPath(entry.name);
+        auto entryCursor = rootCursor.getPath(entry.name.getBuffer());
 
         if(!entryCursor.isValid())
         {
             for(gfx::UInt i = 0; i < shaderObject->getEntryPointCount(); i++)
             {
-                entryCursor = ShaderCursor(shaderObject->getEntryPoint(i)).getPath(entry.name);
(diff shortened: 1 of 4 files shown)


tools/render-test/render-test-main.cpp
```diff
@@ -190,13 +190,13 @@ SlangResult _assignVarsFromLayout(
             return SLANG_E_INVALID_ARG;
         }
 
-        auto entryCursor = rootCursor.getPath(entry.name);
+        auto entryCursor = rootCursor.getPath(entry.name.getBuffer());
```
tfoleyNV, line 193: I have a change similar to this one in my local branch, so the one thing that jumps out to me is that this version of things ends up assuming null-termination of the string for the `getPath/followPath` case, with no option for explicit termination.

The logic only works in this case because (1) our `String`s are null-terminated and (2) it isn't necessary for `render-test` to support names with embedded null bytes in them. I somehow `entry.name` got changed to something like a `UnownedStringSlice` then this code would risk breaking.

Ideally we can/should get back to a point where `ShaderCursor` can work with explicitly-terminated string slices in a convenient fashion soon (e.g., by moving the relevant bits of the `ShaderCursor` API to be inline functions in the public header).

tfoleyNV (approved):
This seems good enough that I am okay with throwing away my local changes to these files to merge what is in this PR.

I am a bit concerned about whether the `gfx-util` library is going to end up needing to rely on `core` in the long run, such that we should be figuring out the details of how that would work.

(I guess the main problem there is that having `gfx-util` end up with yet another dependency makes it that much harder to adopt, so maybe a bit of slightly-ugly code here is worth it)

PR #1667, opened 2021-01-21
Fix reflection to correctly report descriptor ranges of `StructureBuffer`s of existential types.

csyonghe:
If a shader parameter is a `StructuredBuffer<IExistential>`, the `resourceInfos` associated with it will contain 3 entries: one for the shader resource descriptor slot, one for `ExistentialTypeParam` and one for `ExistentialObjectParam`. The two existential entries doesn't affect descriptor set related reflection, so we update the reflection logic to be capable of handling parameters with more than one `resourceInfo`s.

### DIFF
--- a/source/slang/slang-reflection-api.cpp
+++ b/source/slang/slang-reflection-api.cpp
@@ -1587,8 +1587,13 @@ namespace Slang
                     //
                     return;
                 }
-                else if(resourceKindCount == 1)
+                else
                 {
+                    // `resourceKindCount` should be 1 in most cases.
+                    // However if this field is a buffer of existential type,
+                    // the resourceInfo array will contain additional ExistentialTypeParam
+                    // and ExistentialObjectParam entries. These entries doesn't affect the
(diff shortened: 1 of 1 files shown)


source/slang/slang-reflection-api.cpp
```diff
@@ -1587,8 +1587,13 @@ namespace Slang
...
                     return;
                 }
-                else if(resourceKindCount == 1)
+                else
                 {
+                    // `resourceKindCount` should be 1 in most cases.
+                    // However if this field is a buffer of existential type,
+                    // the resourceInfo array will contain additional ExistentialTypeParam
+                    // and ExistentialObjectParam entries. These entries doesn't affect the
+                    // logic here, so we only need to care about the first entry.
```
tfoleyNV, line 1596: Should we do something like scanning through the list of resource kinds until we find one that is *not* the existential object/type case?

I worry that a change somewhere else that results in the resource kinds being listed in a different order could break this logic.

tfoleyNV (approved):

PR #1676, opened 2021-01-26
Obfuscation naming issue fix

jsmall-nvidia:
With the obfuscation feature enabled, it is possible for Slang to output erroneous GLSL code if a function is duplicated because it accesses different resources. In the normal code path (without obfuscation) the name is made unique for every version of a function.

The problem here is that when a function is specialized (say because of resource access requires it), the linkage decoration is copied. This is then used for the function name on emit - and so leads to name clashes. 

The fix here is to remove any linkage decorations after function specialization. 
This could perhaps be performed more cleanly by not cloning all decorations during cloning, most obviously by having some kind of filter. On trying this path originally it was significantly more complicated. 

Other changes:

* Improve code path for appendChar on String/Builder
* Removed the extraneous string construction by scrubName -> appendScrubbedName
* Remove one of two lookups on unique name count by using GetOrAddValue
* A test to check the change works

There is also an argument for replacing m_uniqueNameCounters Dictionary type with StringSliceIndexMap. Doing so would mean the names are not held as individual allocations, and lookup can be performed without needing a String.


source/slang/slang-emit-c-like.cpp
```diff
@@ -667,63 +711,24 @@ String CLikeSourceEmitter::generateName(IRInst* inst)
...
     {
-        return linkageDecoration->getMangledName();
+        if (inst->findDecoration<IRUseLinkageNameDecoration>())
+        {
+            // Just use the linkage name directly
+            return linkageDecoration->getMangledName();
+        }
+        else
+        {
+            return _generateUniqueName(linkageDecoration->getMangledName());
```
tfoleyNV, line 730: I don't think it is a workable solution to generate a unique name based on the mangled name; it just obscures the underlying bug that is going on.

On platforms that can/do support true separate compilation, the mangled name should be the actual name that a Slang symbol (function, variable, whatever) gets emitted as, so that when separately-compiled modules get linked by the system linker (or dynamic library loader), they get matched up by that mangled name. Being able to exploit system linkers in this way requires that we always emit a symbol with linkage using *exactly* its mangled name.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -691,6 +691,9 @@ ScalarizedVal createSimpleGLSLGlobalVarying(
     {
         builder->addImportDecoration(globalParam, UnownedTerminatedStringSlice(systemValueInfo->name));
 
+        // Marks the instruction as using the import decorations name as is.
+        builder->addDecoration(globalParam, kIROp_UseLinkageNameDecoration);
```
tfoleyNV, line 695: I don't think this new decoration should be required. *Every* instruction with linkage should use its mangled name as-is, and we should fix the underlying problem of why those aren't unique.

tfoleyNV (changes requested):
I think we need to try a different approach to this problem, by finding the places where we are generating speicalized copies of functions (e.g., for resource-based specialization) and ensuring that the copies don't get linkage decorations.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -762,6 +762,44 @@ struct FunctionParameterSpecializationContext
...
+                if (as<IRLinkageDecoration>(decoration))
+                {   
+                    decoration->removeFromParent();
+                    decoration->removeArguments();
+
+                    // We DON'T deallocate the children/decorations, as they will be used on the original
+                    // This is okay as it stands, because linkage decorations just contain a string literal  
+
+                    // Run destructor to be sure...
+                    decoration->~IRDecoration();
```
tfoleyNV, line 797: I *think* it would be safe to just use `decoration->removeAndDeallocate()` here.

The comment here seems to be confused about the distinction between operands and children/decorations:

* An instruction does not own its operands, because they could be shared with other instructions (the use/def relationship forms a directed graph - hopefully a DAG). The `removeAndDeallocate()` operation unlinks an insruction as a user of its operands, but does not otherwise try to delete/remove the instructions that are used as oprands.

* An instruction full owns its children (and decorations and parameters are just special cases of children), and they cannot be shared with other instructions (the parent/child relationship is always a tree). While the children of a parent instruction P might be used as operands of other instructions, those other instructions must also be descendents of P given our SSA formulaton.

Instructions should pretty much always be removed using `removeInstAndDeallocate()`.

tfoleyNV (dismissed):
Looks a lot nicer now.

I'd like to see the removal logic cleaned up to be just `removeInstAndDeallocate()`, but that could be a follow-on if it is really needed.

source/slang/slang-ir-specialize-function-call.cpp
jsmall-nvidia: Okay fixed this, thanks for the clarification.

tfoleyNV (approved):

PR #1678, opened 2021-01-27
Make `gfx::IRenderer` own a slang session.

csyonghe:
This change makes `gfx::IRenderer` responsible for maintaining a slang `ISession` for shader reflection and compilation. This is a step necessary to allow `gfx` layer take control of shader compilation to properly support shader-object based specialization.

The `hello-world` and `shader-toy` examples have been modified to use the `ISession` from `IRenderer` instead of creating its own slang session.

### DIFF
--- a/tools/gfx/render.h
+++ b/tools/gfx/render.h
@@ -1085,15 +1085,32 @@ struct Viewport
 class IRenderer: public ISlangUnknown
 {
 public:
+    struct SlangDesc
+    {
+        slang::IGlobalSession* slangGlobalSession = nullptr; // (optional) A slang global session object. If null will create automatically.
+
+        SlangMatrixLayoutMode defaultMatrixLayoutMode = SLANG_MATRIX_LAYOUT_ROW_MAJOR;
+
+        char const* const* searchPaths = nullptr;
+        SlangInt            searchPathCount = 0;
+
+        slang::PreprocessorMacroDesc const* preprocessorMacros = nullptr;
+        SlangInt                        preprocessorMacroCount = 0;
+
+        const char* targetProfile = nullptr; // (optional) Target shader profile. If null this will be set to platform dependent default.
+        SlangFloatingPointMode floatingPointMode = SLANG_FLOATING_POINT_MODE_DEFAULT;
+        SlangOptimizationLevel optimizationLevel = SLANG_OPTIMIZATION_LEVEL_DEFAULT;
+    };
 
     struct Desc
     {
-        int width = 0;                                  ///< Width in pixels
-        int height = 0;                                 ///< height in pixels
-        const char* adapter = nullptr;                  ///< Name to identify the adapter to use
-        int requiredFeatureCount = 0;                   ///< Number of required features.
-        const char** requiredFeatures = nullptr;        ///< Array of required feature names, whose size is `requiredFeatureCount`.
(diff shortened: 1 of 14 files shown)


tools/gfx/render.h
```diff
@@ -1085,15 +1085,32 @@ struct Viewport
...
+        char const* const* searchPaths = nullptr;
+        SlangInt            searchPathCount = 0;
+
+        slang::PreprocessorMacroDesc const* preprocessorMacros = nullptr;
+        SlangInt                        preprocessorMacroCount = 0;
+
+        const char* targetProfile = nullptr; // (optional) Target shader profile. If null this will be set to platform dependent default.
+        SlangFloatingPointMode floatingPointMode = SLANG_FLOATING_POINT_MODE_DEFAULT;
+        SlangOptimizationLevel optimizationLevel = SLANG_OPTIMIZATION_LEVEL_DEFAULT;
+    };
```
tfoleyNV, line 1103: Could this just use `slang::SessionDesc` instead? Some of the fields there might need to be ignored, but it seems like a safer way to make the `IRenderer` API independent of the details of the Slang API.

This would be especially important if/when we start to make APIs like `creaseSession` "extensible" in a binary-compatible fashion; we'd want to ensure that the `gfx` API doesn't need any changes to expose new facilities that are part of Slang session creation.

tools/gfx/slang-context.h
```diff
@@ -0,0 +1,43 @@
...
+            slangSessionDesc.searchPathCount = desc.searchPathCount;
+            slangSessionDesc.searchPaths = desc.searchPaths;
+            slangSessionDesc.preprocessorMacroCount = desc.preprocessorMacroCount;
+            slangSessionDesc.preprocessorMacros = desc.preprocessorMacros;
+            slang::TargetDesc targetDesc = {};
+            targetDesc.format = compileTarget;
+            auto targetProfile = desc.targetProfile;
+            if (targetProfile == nullptr)
+                targetProfile = defaultProfileName;
+            targetDesc.profile = globalSession->findProfile(targetProfile);
```
tfoleyNV, line 34: I guess this is one detail that makes it tricky to just use `SessionDesc`: that type requires the integer version of the profile, while we want users of this API to be able to use the string.

That gets into a messier issue which is that we probably need to make one of two choices in the Slang API:

* We could support/encourage using strings directly to represent profiles, rather than having to go through `spFindProfile()` or `IGlobalSession->findProfile()`. The cost of the string lookups is realistically neglibible at this point compared to other work being done inside the Slang library.

* We could decide to give most (or all) profiles fixed integer ID values so that application code is not required to look them up by string name at runtime. Instead, they could hard-code in a value using an `enum` defined in `slang.h`.

I'm not sure which of these we should prefer, but it seems like either one would be part of the way to get around mess like this.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -1033,6 +1033,8 @@ VkPipelineShaderStageCreateInfo VKRenderer::compileEntryPoint(
 
 SlangResult VKRenderer::initialize(const Desc& desc, void* inWindowHandle)
 {
+    SLANG_RETURN_ON_FAIL(slangContext.initialize(desc.slang, SLANG_SPIRV, "sm_5_1"));
```
tfoleyNV, line 1036: It seems a bit odd to use `sm_5_1` as the profile here. Surely it should be `glsl_440` or whatever is the appropriate GLSL version for what we need/want to support?

tfoleyNV (approved):

PR #1679, opened 2021-01-27
README.md update

jsmall-nvidia:
To include section on 'trying out' Slang - that points to shader compiler explorer website.


tfoleyNV (approved):
My biggest concern with including a link to Shader Playgrounds here is that, IIRC, Shader Playground does not include DXBC, DXIL, or SPIR-V as output formats for Slang, and instead expects the user to build a compilation pipeline that includes them. That usage model is *not* one we want to encourage, but Shader Playground makes it look like it is the intended way to use Slang.

Obviously that concern has nothing to do with this PR, and the right way to resolve it is to file issues against Shader Playground requesting that they more accurately show the output formats that Slang is designed to support.

jsmall-nvidia:
I think we can quite easily add support for other output targets on shader explorer. It does assume that the appropriate dlls are available, but they went out of their way to make NVRTC available so that PTX worked. So seems very doable.

tfoleyNV:
I opened an issue on Shader Playground about adding more direct support for our DXBC/DXIL/SPIR-V targets. Obviously it might be simpler to get movement on that issue if we submit a PR for it. :)

PR #1683, opened 2021-02-02
Diagnostic comparison using parsing

jsmall-nvidia:
This is on top of #1682

* Add parsing for Slang diagnostic output
* Added compiler detection 
* Parsing of Slang test output (std/error result)
* Removed ad hoc special case handling for diagnostic comparison 
* Use parsing result as basis for diagnostic comparison 
* Added ability to parse diagnostics with and without line prefixes identifying the compiler


source/core/slang-downstream-compiler.h
```diff
@@ -18,6 +18,8 @@ namespace Slang
 
 struct DownstreamDiagnostic
 {
+    typedef DownstreamDiagnostic ThisType;
+
     enum class Type
```
tfoleyNV, line 23: Unrelated to this PR, but we should probably name this "severity" eventually, to match what we use for our own diagnostics. It might also be nice to share some subset of those fundamental types between these cases, but I guess the issue here is that we are in `core` here rather than `slang`.

tfoleyNV (approved):
This looks good to me, modulo the fact that I can't pretend to know the details of the output diagnostic formats for the different compilers being handled here.

source/core/slang-downstream-compiler.h
tfoleyNV: (I'm referring to the `enum class Type` here, and not the `ThisType` declaration that the PR added, just to be clear.)

(And part of the context here is my ongoing effort to avoid using the term "type" to refer to anything that is not actually or arguably a type in the PL sense of the term)

PR #1688, opened 2021-02-04
[gfx] Shader-object driven shader compilation.

csyonghe:
Adds the following responsibilities to `gfx` layer:
- Shader/pipeline specialization based on the type of bound shader-objects
- Caching of specialized shader kernels
Note: none of these new `gfx` capabilities are used and tested by any
part of the code base, and the implementation and interface involving
these new capabilities are likely to be changed in future commits.
The implementation only handles the simplest case of shader specialization:
via a sub-object defined by an interface-typed field.
A `ParameterBlock<IInterface>` or `StructuredBuffer<IInterface>` typed
field is not implemented yet. As a result, one of the dynamic-dispatch
tests has to be disabled for now since it uses `ConstantBuffer<IInterface>`.

Refactors to the `gfx` layer:
- Remove `gfxGetCreateFunc` and just expose `gfxCreateRenderer`.
- Remove `IRenderer::initialization` method and performs initialization
  directly at renderer creation time.
- Move many logic in `renderer-graphics-common.h/cpp` to `renderer-shared.h/cpp`,
  where things in `renderer-graphics-common` are shared by all graphics
  API implementations, and things in `renderer-shared` are shared by all
  renderer implementations, including CUDA and CPU. This increases code
  reuse and allow most shader compilation and specialization logic to be
  shared among all renderers.
- `IRenderer::setPipelineState` no longer takes a `PipelineType` parameter.

General changes:
- Change from C style type cast to `static_cast<RefObject*>` in
  `RefPtr<T>::releaseReference` to capture all `RefPtr` mis-uses.

### DIFF
--- a/source/core/slang-dictionary.h
+++ b/source/core/slang-dictionary.h
@@ -135,15 +135,17 @@ namespace Slang
 			}
 
 		};
-		inline int GetHashPos(TKey& key) const
+        template<typename KeyType>
+		inline int GetHashPos(KeyType& key) const
         {
             SLANG_ASSERT(bucketSizeMinusOne > 0);
             const unsigned int hash = (unsigned int)getHashCode(key);
             return (hash * 2654435761u) % (unsigned int)(bucketSizeMinusOne);
 		}
-		FindPositionResult FindPosition(const TKey& key) const
+        template<typename KeyType>
+		FindPositionResult FindPosition(const KeyType& key) const
 		{
-			int hashPos = GetHashPos(const_cast<TKey&>(key));
+			int hashPos = GetHashPos(const_cast<KeyType&>(key));
 			int insertPos = -1;
 			int numProbes = 0;
 			while (numProbes <= bucketSizeMinusOne)
@@ -380,14 +382,16 @@ namespace Slang
                 throw InvalidOperationException("Inconsistent find result returned. This is a bug in Dictionary implementation.");
         }
 
-		bool ContainsKey(const TKey& key) const
+        template<typename KeyType>
+		bool ContainsKey(const KeyType& key) const
 		{
 			if (bucketSizeMinusOne == -1)
 				return false;
 			auto pos = FindPosition(key);
 			return pos.ObjectPosition != -1;
 		}
-		bool TryGetValue(const TKey& key, TValue& value) const
+        template<typename KeyType>
+		bool TryGetValue(const KeyType& key, TValue& value) const
 		{
 			if (bucketSizeMinusOne == -1)
 				return false;
@@ -399,7 +403,8 @@ namespace Slang
 			}
 			return false;
 		}
-		TValue* TryGetValue(const TKey& key) const
+        template<typename KeyType>
+		TValue* TryGetValue(const KeyType& key) const
 		{
 			if (bucketSizeMinusOne == -1)
 				return nullptr;
--- a/tests/compute/dynamic-dispatch-11.slang
+++ b/tests/compute/dynamic-dispatch-11.slang
@@ -1,9 +1,9 @@
 // Test using interface typed shader parameters with dynamic dispatch.
 
-//TEST(compute):COMPARE_COMPUTE:-dx11 -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-dx11 -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
 //TEST(compute):COMPARE_COMPUTE:-cpu -xslang -disable-specialization -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-cuda -xslang -disable-specialization -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-cuda -xslang -disable-specialization -shaderobj
 
 [anyValueSize(8)]
 interface IInterface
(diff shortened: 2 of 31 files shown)


source/core/slang-dictionary.h
```diff
@@ -135,15 +135,17 @@ namespace Slang
 			}
 
 		};
-		inline int GetHashPos(TKey& key) const
+        template<typename KeyType>
+		inline int GetHashPos(KeyType& key) const
```
tfoleyNV, line 139: We have a few operations on `Dictionary` that allow these kinds of operations, where the user can pass something that isn't actually of the correct key type. This choice seems to be designed around things like smart pointers where the value you use as the key might not be exactly what the dictionary stores.

The concern I have is that this approach seems to allow a value to be used for the key that is different from the actual key type, but is convertible to it, and the implicit constraint is that the hash of the value before/after conversion should not be changed.

What would happen if I had a `Dictinary<int, int>` and then I tried to use a `float` as a key? Hashing my key as a `float` instead of an `int` would seemingly lead to different values. How does the code here protected against that kind of stuff?

tests/compute/dynamic-dispatch-11.slang
```diff
@@ -1,9 +1,9 @@
 // Test using interface typed shader parameters with dynamic dispatch.
 
-//TEST(compute):COMPARE_COMPUTE:-dx11 -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-dx11 -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
 //TEST(compute):COMPARE_COMPUTE:-cpu -xslang -disable-specialization -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-cuda -xslang -disable-specialization -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-cuda -xslang -disable-specialization -shaderobj
```
tfoleyNV, line 6: Are we okay with these tests being disabled for now?

tools/gfx/render-graphics-common.cpp
```diff
@@ -859,16 +800,66 @@ class GraphicsCommonShaderObject : public IShaderObject, public RefObject
...
+                // A binding type of `ExistentialValue` means the sub-object represents a interface-typed field.
+                // In this case the specialization argument for this field is the actual specialized type of the bound
+                // shader object. If the shader object's type is an ordinary type without existential fields, then the
+                // type argument will simply be the ordinary type. But if the sub object's type is itself a specialized
+                // type, we need to make sure to use that type as the specialization argument.
+
+                // TODO: need to implement the case where the field is an array of existential values.
+                SLANG_ASSERT(bindingRange.count == 1);
+                ExtendedShaderObjectType specializedSubObjType;
+                SLANG_RETURN_ON_FAIL(m_objects[subObjIndex]->getSpecializedShaderObjectType(&specializedSubObjType));
```
tfoleyNV, line 841: Same issue as for the case in the CUDA back-end. I'd expect to see two levels of nesting here: one over the sub-object ranges and one over the sub-objects in each range. The actual `subObjIndex` should be something computed from the `bindingRange` for the sub-object range plus the index of the particular sub-object in the inner loop.

This is probably fine to leave as-is for now, but it definitely isn't 100% correct.

tools/gfx/render.h
```diff
@@ -1104,18 +1105,17 @@ class IRenderer: public ISlangUnknown
...
     struct Desc
     {
+        RendererType rendererType; // The underlying API/Platform of the renderer.
         int width = 0;                                  // Width in pixels
         int height = 0;                                 // height in pixels
         const char* adapter = nullptr;                  // Name to identify the adapter to use
         int requiredFeatureCount = 0;                   // Number of required features.
         const char** requiredFeatures = nullptr;        // Array of required feature names, whose size is `requiredFeatureCount`.
         int nvapiExtnSlot = -1;                         // The slot (typically UAV) used to identify NVAPI intrinsics. If >=0 NVAPI is required.
+        ISlangFileSystem* shaderCacheFileSystem = nullptr; // The file system for loading cached shader kernels.
```
tfoleyNV, line 1115: As discussed elsewhere, I like this idea, but it also has a lot of implications for how we do `.zip`-based shader caching if we want to support that. This is probably good to have as a stake in the ground, but shader caching needs a lot of top-down design work.

tools/gfx/renderer-shared.cpp
```diff
@@ -96,4 +145,396 @@ Result createProgramFromSlang(IRenderer* renderer, IShaderProgram::Desc const& o
...
+}
+
+Slang::RefPtr<ShaderBinary> ShaderCache::tryLoadShaderBinary(ShaderComponentID componentId)
+{
+    Slang::ComPtr<ISlangBlob> entryBlob;
+    Slang::RefPtr<ShaderBinary> binary;
+    if (shaderBinaries.TryGetValue(componentId, binary))
+        return binary;
+
+    if (fileSystem && fileSystem->loadFile(String(componentId).getBuffer(), entryBlob.writeRef()) == SLANG_OK)
```
tfoleyNV, line 308: Of course I *say* that the caching needs a lot of thought, and then you already have a working version of it implemented. :)

The main detail here is that the caching can only work because you've designed a keyspace that is entirely string-based. Maybe that is enough, but it is possible that sooner or later we will need a manifest/index format independent of the actual cached kernels.

tfoleyNV (approved):
High-order bit: it is great to see so much progress on this stuff!

csyonghe:
I agree that there are a lot of design questions around the shader cache. For now I am just doing a rough and fast first version to get at least basic examples working. Once all the priorities are sorted out we could evolve this over time.

PR #1690, opened 2021-02-04
Fix line offset problem

jsmall-nvidia:
On top of #1687 

* Small improvements to GLSL output
* Split out the intrinsic expansion code into separate class/files
* Output intrinsic expansion using 'emit' so as line number handling/indenting is handled correctly
* Added some more line number tests
* Made __SyntaxError `return` an int so can be used within an expression

These changes mean that line numbers are consistent across backends and match the __SyntaxError location correctly in original source.


tfoleyNV (approved):
Unfortunately, the move of code into a new file makes it a bit hard to understand what exactly about the emission logic for intrinsics changed. This seems fine overall, but I could maybe use a pointer to the relevant lines in `slang-intrinsic-expand.cpp`.

PR #1691, opened 2021-02-04
Initial implementation of interface conjunctions

tfoleyNV:
The basic feature here is the ability to use the `&` operator to produce the conjunction/intersection of two interfaces. That is, you can have interfaces:

    interface IFirst { int getFirst(); }
    interface ISecond { int getSecoond(); }

and if you need a generic function where the type parameter `T` must conform to *both* of these interfaces, you express that by constraining the parameter to the intersection of the interfaces:

    void someFunction<T : IFirst & ISecond>(T value) { ... }

Without this feature, the main alternative an application would have is to define an intermediate interface, like:

    interface IBoth : IFirst, ISecond {}

Forcing users to deal with an intermediate interface creates more work for type authors (they need to remember to inherit from the right combined interface(s)), or for `extension` authors (when you add `ISecond` to a type that used to just support `IFirst`, you had better also add `IBoth`). In the worst case, a family of N related "leaf" interfaces would give rise to an exponential number of intermediate interfaces to represnt the possible combinations.

A conjunction like `IFirst & ISecond` is officially its own type, and can be used to declare a type alias:

    typealias IBoth = IFirst & ISecond;

This change only includes the first pass of work on this feature, so there are several caveats to be aware of:

* Using a conjunction as part of an inheritance clause is not yet supported (e.g., `struct X : IFirst & ISecond`). This is true even if the conjunction was introduced by an intermediate `typealias`

* The `&` syntax introduced here is only parsed in places where only a type (not an expression) is possible. This means you cannot do things like cast to a conjunction with `(IFirst & ISecond)(someValue)`.

* This work *should* apply to conjunctions of more than two interfaces (like `IA & IB & IC`) but that has not yet been tested

* In the long run it may be sensible to allow conjunctions that use concrete types, but we really ought to have the semantic checking logic rule that out for now.

* During testing, I encountered compiler crashes when trying to use this feature together with `property` declarations. Further investigation and debugging is called for.

* The handling of conjunction types is currently incomplete, in that there are many equivalences the compiler does not yet understand. For example, it is clear that `IA & IB` is equivalent to `IB & IA`, but the compiler currently does not understand this and will treat them as different types. A deeper implementation approach is called for.

* Conjunctions are currently only supported for generic type parameter constraints, when performing full specialization. Use of conjunctions for existential-type value parameters or with dynamic dispatch is not yet supported.


source/slang/slang-lower-to-ir.cpp
```diff
@@ -6228,6 +6310,79 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+            //
+            auto param = subBuilder->emitParam(subBuilder->getWitnessTableType(supType));
+            addNameHint(context, param, constraintDecl);
+
+            // In order to support some of the "any-value" work in dynamic dispatch
+            // we have to attach the interface that was used as a constraint onto the
+            // type that is being constrained (which we expect to be a generic type
+            // parameter).
+            //
+            // TODO: It feels a bit gross to be doing this here; perhaps the front-end
```
csyonghe, line 6362: This definitely seems something better done in the front-end. And we probably don't need the general `TypeConstraintDecoration`, and just use a simple `ValueSizeDecoration`.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6228,6 +6310,79 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+            // the `IRConunctionType` can support more than just binary
+            // conjunctions) and recursively add constraints for each.
+            //
+            List<IRInst*> caseVals;
+            auto caseCount = andType->getCaseCount();
+            for(Int i = 0; i < caseCount; ++i)
+            {
+                auto caseType = andType->getCaseType(i);
+                auto caseVal = emitGenericConstraintValue(subContext, constraintDecl, caseType);
+                caseVals.add(caseVal);
```
csyonghe, line 6342: So instead of generating a flattened list of constraint values, we are still grouping them into hierarchies of tuples here? Will they get flattened later?

csyonghe (approved):

source/slang/slang-check-constraint.cpp
```diff
@@ -659,6 +659,21 @@ namespace Slang
         return false;
     }
 
+    bool SemanticsVisitor::TryUnifyConjunctionType(
+        ConstraintSystem&   constraints,
+        AndType*            fst,
```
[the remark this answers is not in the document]
tfoleyNV: I tend to agree, but in this case I was copy-pasting logic from the function below that already established the variable names, so I figured I'd keep them.

The original rationale was that the code was working with two types, and wanted to treat them symmetrically. It was easier to visually scan to see that they were being treated the same when the identifier names for the two types were the same length.

It is probably better to just name them something like `type0` and `type1` instead.

source/slang/slang-lower-to-ir.cpp
tfoleyNV: It's a little thorny, yeah.

When we are generating code for a `GenericDecl` and it has a `GenericConstraintDecl` that represents something like `T : L & R`, then this code will recursively emit distinct IR parameters for witness tables representing `T : L` and `T : R`. That works reasonably well and ensures that our `IRGeneric`s just have ordinary witness-table parameters and code that works with them doesn't need to know about the tuples.

The problem is that the AST-level code inside of that `GenericDecl` will include references directly to the original `GenericConstraintDecl` (the one for `T : L & R`), and we somehow need to be able to resolve what those references should actually refer to. The decision I made is that given the way a constraint parameter for `T : L & R` is being lowered into the IR as two witness-table parameters for `T : L` and `T : R`, then it made sense for the IR-level value of the original constraint parameter to be represented by the IR tuple `( T:L, T:R )`.

Any piece of AST code that references that `GenericConstraintDecl` is likely to be referencing it via an `ExtractFromConjunctionSubtypeWitness`. For example, if we have a value `value : T` and `T : IFirst & ISecond` and somebody writes `value.getFirst()`, then the front-end will know that it needs to cast `value` to the type `IFirst` to access the `getFirst()` method. In order to do that cast we need a witness that `T : IFirst`, which the AST forms (during lookup) by extracting the left half of the `T : IFirst & ISecond` witness.

Given the encoding of a `GenericConstraintDecl` for `T : L & R` as the tuple `( T:L, T:R )`, it becomes easy to see that an `ExtractFromConjunctionSubtypeWitness` can easily extract the `T:L` part of that constraint by grabbing the first element of the tuple.

Thus, while the lowering of the `GenericConstraintDecl` introduces a few IR tuples, use sites almost always start by extracting one or the other element from it. This PR also includes a change to the `IRBuilder::emitGetTupleElement` logic so that it detects an attempt to emit a `getTupleElement()` on a `makeTuple()` instruction and immediately simplifies it to a direct reference to the appropriate element. As a result, most of the existing IR machinery never has to contend with the possibility of a tuple.

(I originally thought that I might be able to consistently "flatten" these conjunctions everywhere in the IR, but there is a pretty deep assumption in the IR lowering that most AST-level declarations/types/expressions can always lower to a single IR instruction. Yes, there is the `LoweredValInfo` approach, but that really only exists to support l-value vs r-value distinctions during expression lowering (and even *that* might be wearing out its welcome as an approach). In practice, for something like `typealias IBoth = IFirst & ISecond` we fundamentally needed to be able to construct a single `IRInst*` that represents the type being aliased there.)

PR #1692, opened 2021-02-05
Copy SourceLoc when inlining

jsmall-nvidia:
This PR attempts to improve sourceLoc information around inlining. Previously no source loc information was copied over when inlining. The change copies over sourceLoc information from the call location. It only does this because it assumes all [__unsafeForceInlineEarly] are not in user code, but in stdlib. That being the case reporting line numbers from stdlib is likely confusing.

The example uses [__unsafeForceInlineEarly], in 'user' code and so less helpfully identifies the problem to be on the caller line.

A better fix would alter the sourceLoc depending if the thing being inlined is part of stdlib or not.


source/slang/slang-ir-inline.cpp
```diff
@@ -355,6 +355,52 @@ struct InliningPassBase
...
+        // to be a decoration for this.
+        // We could find out by looking at the source loc and checking if it's in the range of stdlib - this would actually be
+        // a fast and easy but to do properly this way you'd want a way to mark that source range that would also work across
+        // serialization.
+        // 
+        // For now this punts on this, and just assumes [__unsafeForceInlineEarly] is not in user code.
+    static IRInst* _cloneInstWithSourceLoc(CallSiteInfo const& callSite,
+        IRCloneEnv*     env,
+        IRBuilder*      builder,
+        IRInst*         inst)
```
tfoleyNV, line 383: This function seems like a good solution to the problem, and my only question is whether we can make it something more integrated/systematic.

One obvious tweak would be to make the `cloneInst` operation default to copying the `sourceLoc` of the instruction being cloned (which makes sense for a "clone" operation anyway). That would mean that the `else if` bit below is not needed.

Another potentially interesting bit would be to have the `IRBuilder` type include a field called something like `sourceLocToUseForOverride` that can optionally be set and, when it is set, becomes the source location of all instructions emitted for that `IRBuilder`. That would mean that the `inlineTrivialFuncBody()` below could just set up the "override location" based on the `callSite` and allow other machinery to do the rest of the work.

tfoleyNV (approved):
This is a good fix for one of our classes of `#line` issues. My only feedback is around maybe migrating parts of the solution deeper into the infrastructure so that more cases in the future can Just Work without us doing quite as much manual tweaking.

tfoleyNV:
One really ambitious/scary thing for us to think about is whether we should just make it so that *every* operation that creates an `IRInst` receives some kind of explicit source location, or explicit information on how to *derive* a source location. By doing so we could try to ensure that at every point in the code where we are generating IR instructions we are forced to think about how the locations of those instructions should be derived.

PR #1697, opened 2021-02-09
Add getting started documentation

csyonghe:


docs/user-guide/01-get-started.md
```diff
@@ -0,0 +1,93 @@
...
+
+The easiest way to start using Slang is to download a [binary release](https://github.com/shader-slang/slang/releases/) from the github repository. Once you have downloaded and extracted the files from a release package, you can find the `slangc.exe` executable under `/bin/windows-x64/release/`. In this tutorial we will use the `slangc` standalone Slang compiler included in a release package. Note that `slang.dll` and `slang-glslang.dll` must be placed in the same directory as `slangc.exe` as they are required by the standalone executable.
+
+If you are interested in building from source, please refer to the [documentation on building Slang](../building.md). 
+
+## Your first Slang shader
+
+In this section we demonstrate how to write a simple compute shader in Slang that adds numbers from two buffers and writes the results into a third buffer. To start, create a text file named `hello-world.slang` in any directory, and paste the following content in the newly created file:
+
+```
```
tfoleyNV, line 15: Consider changing to ````` ```hlsl ````` to indicate the source language that should be used/assumed for syntax highlighting (and similarly for the GLSL block below).

docs/user-guide/01-get-started.md
```diff
@@ -0,0 +1,93 @@
...
+
+## The full example
+
+The full Vulkan example that sets up and runs the `hello-world.slang` shader in located in the `/examples/hello-world` directory of the Slang repository. The example code initializes a Vulkan context and runs the compiled SPIRV code.
```
tfoleyNV, line 93: It is a little thorny to check in a statement like this that is false at the time it is written.

We probably need some kind of convention in our markdown files for annotating things as callouts/warnings. A fairly simple policy would be to use block quotes where the first line is a header with giving the kind of callout. E.g., the following Markdown:

```
> #### Note ####
> The current `hello-world` implementation does not use raw Vulkan API calls;
> this will be addressed in a future revision so that it matches the documentation.
```

renders as:

> #### Note ####
> The current `hello-world` implementation does not use raw Vulkan API calls;
> this will be addressed in a future revision so that it matches the documentation.

which seems like a reasonable callout for now. If/when we get more serious about how we process our `.md` files into documentation this is the kind of thing that we could detect and customize the formatting for when translating Markdown to HTML.

tfoleyNV (approved):

PR #1699, opened 2021-02-09
Add convenience features chapter in user-guide doc

csyonghe:


tfoleyNV (approved):
We probably need to go through and renumber the chapters after merging this.

In the outline form I'd been thinking of Chapter 2 as an outline for the rest of the document, with forward links to other chapters, and then Chapter 3 is the part about "conventional" language features.

I think what you have here is good, and it makes sense to lump together some of the more straightforward convenience features (the ones without larger implications for the API, runtime, etc.). The only problem is that it doesn't make sense to describe this convenience features *before* describing the features they build on.

Probably what we need to do after this lands is:

* Add an "Overview of the rest of this guide" section to the end of `00-introduction` that gives a summary of the other chapters so that people who want to can skip ahead.
* Renumber the Conventional Language Features chapter from 3 to 2
* Renumber the Convenience Features chapter from 2 to 3
* Carefully consider whether "Conventional Features" and "Convenience Features" are too similar of names. Figure out which we should change, and how.

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
+# Basic Convenience Features
```
tfoleyNV, line 1: Minor nit: stylistically, I find that the "ATX-style" headers are nicer to look at when you have the cases they cover (1st- and 2nd-level headers):

```
First Level
===========

Second Level
------------
```

They will render the same in the output HTML, but when reading or editing the plain text of the Markdown file, they tend to stand out more and make it easier to visually parse the document structure.

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
...
+> ```
+> property uint highBits {get;set;} // Not supported yet.
+> ```
+
+## Initializers
+> #### Note ####
+> The syntax for defining initializers are subject to future change.
+
+
+Slang supports defining initializers in `struct` types. You can write:
```
tfoleyNV, line 125: We probably need to convene the dev team and decide on an official name for this feature. The main contenders I know of are "initializers" and "constructors," but both of these have their own unwanted baggage.

The term "constructor" is inextricably linked to C++ and its particular semantics of constructions/destruction. If we have constructors in Slang people will expect them to behave exactly like constructors in C++, filing bugs when they don't. It also seems inevitable that this road will lead to people wanting destructors, an overridable assignment operator, etc.

The term "initializer" is loaded because languages like C have used it as a synonym for an initial-value expression on an variable declaration, and there is also the specific "initializer list" form that is only allowed in that context.

I think that the better option is to use "initializer" like you have here, but then to zealously avoid using the term "initializer" to refer to anything like an initial-value expression for a variable (and then to come up with an alternative name for C-style "initializer lists").

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
...
+{
+    int myVal;
+    __init(int a, int b)
+    {
+        myVal = a + b;
+    }
+}
+```
+
+You can use an initializer to construct a new instance by using the type name in a function call expression:
```
tfoleyNV, line 137: We should probably refer to this form as just a *call expression* and not a *function call expression*, because of the fact that clearly one can call things that are not functions.

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
...
+MyType instance = {1, 2};
+```
+
+If an initializer does not define any parameters, it will be recognized as *default* initializer that will be automatically called at the definition of a variable:
+
+```C#
+struct MyType
+{
+    int myVal;
+    __init(i)
```
tfoleyNV, line 153: First note: the `i` here seems to be an accident.

tfoleyNV: Second note: I want us to be *very* careful about whether to advertise this feature. Right now, I think of our support for default initializers as a bit of hack.

Almost every single non-C++ language (including all older versions of HLSL) has made a consistent choice between one of two options:

1. A variable without an initial-value expression is uninitialized, *always*. Access to a potentially-uninitialized variable may be a static error. or not. Access to an uninitialized variable at runtime may be a dynamic error, or it might yield an undefined value, or it might result in the dreaded Undefined Behavior.
2. A variable without an initial-value expression is default-initialized, *always*. Numeric values are initialized to zero, and references are initialized to null.

To my knowledge, C++ is the only mainstream language that doesn't pick one of these two internally-consistent options, and I'm reluctant to drag Slang down that road just because dxc has used some C++ features in the way it exposes new pipeline features.

Current HLSL does not support constructors/initializers in user code, and there is (to my knowledge) exactly *one* type in the current HLSL standard library that makes use of default construction: `RayQuery<M>`. (There is also the `HitGroup` type that uses C++ variable construction syntax as part of its idiom, but it doesn't rely on *default* construction)

I am very hesitant to introduce this feature as something our users are expected to know about. I would rather flag that `RayQuery<M>` initializer as an extreme special case, and perhaps even warn users when they declare a `RayQuery<M>` without initializing it (and advise them to do `var query = RayQuery<M>();` or similar.

Let's plan to talk about this and come to a conclusion.

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
...
+}
+
+int test()
+{
+    MyType rs = MyType(1) + MyType(2);
+    return rs.val; // returns 3.
+}
+```
+
+Slang currently supports overloading the following operators: `+`, `-`, `*`, `/`, `%`, `&`, `&&`, `|`, `||`, `<`, `>`, `<=`, `>=`, `==`, `!=`, unary `-`, `~` and `!`.
```
tfoleyNV, line 188: This list is missing the compound assignment operators like `+=`.

Realistically, we probably want to have a POR of defining some built-in `interface`s for the typical interpretation of the binary/unary operators, and then allow types to opt in to operator support by conforming to those interfaces (that is how Rust works). This approach minimizes the total number of operator overloads that the semantic checker needs to deal with, and also means that opting into the `IAddable` interface (or whatever it is called) automatically confers both support for `+` and `+=`. A similar statement could be made for an `IPartiallyOrdered` and `ITotallyOrdered` protocol pair, which could be used to implement all the comparison ops (giving many of the same benefits that recent C++ adds with the "spaceship operator").

docs/user-guide/02-convenience-features.md
```diff
@@ -0,0 +1,227 @@
+# Basic Convenience Features
+
+This topic covers a series of nice-to-have language features in Slang. These features are not supported by HLSL but are introduced to Slang to simplify code development. Many of these features are added to Slang per request of our users.
```
tfoleyNV, line 3: One thing that is missing from this section that might be nice is the `let` and `var` syntax for declaring variables. These cover 99% of the cases where people tend to reach for `auto`, so they seem nice to discuss.

(Also, using `let` automatically makes variables immutable by default, which is nice for those users who currently add `const` to every single variable they declare)

PR #1700, opened 2021-02-10
Diagnostic location highlighting

jsmall-nvidia:
Slang diagnostics display the location of a SourceLoc as a filename and a line number. This PR adds to that capability by displaying the the line from the source (if available), and highlighting the location of the error on that line.

Displaying a caret where a SourceLoc is, is relatively straightforward, and just requires finding the location in the SourceFile, and outputting something suitable. This PR also adds the ability to mark the caret and the surrounding characters of the token being pointed at. This is achieved by giving the DiagnosticSink a SourceLocationLexer callback, that given a Slice, will return the token located there. This feature is optional. The functionality is achieved by callback, so that any suitable implementation can be used, and so the DiagnosticSink is not tied to a specific implementation.

To take an example from tests

```
int foo()
{
	return 1;

	// The following statements can never be reached.
	int x = 0;
	return x;
}
```

Previously would output...

```
tests/diagnostics/unreachable-code.slang(10): warning 41000: unreachable code detected
```

Now outputs 

```
tests/diagnostics/unreachable-code.slang(10): warning 41000: unreachable code detected
    int x = 0;
    ^~~
```

Two additional tests were added to check behavior was reasonable with line continuations. The expected test output has been updated to have the additional information.

Note that by displaying the information this way, it can be seen that some of the tests location information is less than ideal. Examples of this are

* Being on the wrong line (x-macro-line-continuation.slang)
* The location is 'slightly off' (constexpr-error.slang)

Some other surrounding issues

* If there is a line continuation, the whole token is not typically shown (doesn't seem unreasonable tho)
* No command line control of the feature currently
* Assumes a tab size of 4 - we may want that controllable from command line
* May want a mechanism to specify a terminal width, so for lines which are too long we handle suitably

Lines are only output like this for non-note/info lines. This is because it generally appeared to output too much extra stuff. Then when a diagnostic refers back to a previous definition though - it may be useful to display the line. Some other heuristic could perhaps capture this scenario, or have a Severity that captures this, or only have SourceLoc on note when it is reasonable to output.

Diagnostic location that could be improved

local-used-before-declared.slang
matrix-swizzle.slang - location is on the ., seems more appropriate on the _m
mutating-method-on-rvalue.slang - seems more appropriate location would be on method name, not `(`
static-ref-to-nonstatic-member.slang - should be on name not .
vk-bindings.slang - not terrible but a bit surprising
x-macro-line-continuation.slang - wrong line

define-function-like.slang - the location for this error is arguably confusing


tfoleyNV:
This is a cool feature to have, and I'll look forward to reviewing this.

One note is that once we have this kind of feature we *probably* want our AST nodes to start tracking a `SourceRange` instead of just a `SourceLoc`, so that we can rely on the parsing logic that constructs AST nodes to fill in source ranges in an appropriate fashion, rather than trying to reverse-engineer a reasonable source range from a single location.

We can in theory even go so far as to have certain AST nodes store additional source locations/ranges beyond their overall range. For example, a binary operator expression like `+` could store both the entire source range of the expression (from the start of the left operand to the end of the right one), and also the exact location of the `+` token. In a case like that, we'd really want to mark the entire range with `~` but put the `^` caret on the `+` token.

This whole feature might push us toward a model where emitting a diagnostic follows more of a "builder" approach, and a diagnostic comprises:

* Required: the diagnostic ID
* Optional: zero or more `SourceLoc`s where `^`s will be placed
* Optional: zero or more `SourceRange`s which will be covered by `~` if not already using `^`
* Optional: zero or more additional arguments for printing as part of the formatted diagnostic
* ... (eventually stuff like "fixit" hints)

Under that model, each AST node might actually have both a `SourceRange` *and* a `SourceLoc` by default, where the range ideally covers every byte of input that the AST node covers, while the location is the ideal single location to use.

source/slang/slang-diagnostics.cpp
```diff
@@ -160,6 +161,160 @@ static void formatDiagnostic(const HumaneSourceLoc& humaneLoc, Diagnostic const&
...
+
+    if (start < end)
+    {
+        out.append(start, end);
+    }
+}
+
+// Given multi-line text, and a position within the text (as a pointer into the memory of text)
+// extract the line that contains pos
+static UnownedStringSlice _extractLineContainingPosition(const UnownedStringSlice& text, const char* pos)
```
tfoleyNV, line 217: We already have logic very similar to this in the `SourceManager` (or something similar... I forget the details), and it would be great to be able to use that logic instead of a linear search.

At least as of whenever I last touched the relevant code, each source file/unit would have a lazily-built lookup to able that could be used to map a source location (basically just an offset into the file) to the appropriate line number. That same structure could seemingly be used to extract the text of a line as well. The main benefit there is that the linear search over the text of the file can then be done once and amortized, while the actual lookup only uses a binary search.

source/slang/slang-lexer.h
```diff
@@ -106,6 +106,16 @@ namespace Slang
...
 
+            /// Runs the lexer to try and extract a single token, which is returned.
+            /// This can be used by the DiagnosticSink to be able to display more appropriate
+            /// information when displaying a source location - such as underscoring the
+            /// token at that location.
+            ///
+            /// NOTE! This function is relatively slow, and is designed for use around this specific
+            /// purpose. It does not return a token or a token type, because that information is
+            /// not needed by the DiagnosticSink.
+        static UnownedStringSlice sourceLocationLexer(const UnownedStringSlice& in);
```
tfoleyNV, line 117: Ideally we will not need this in the long run if we make diagnostics use `SourceRange`s for the range highlighting, rather than trying to auto-infer them from the lexer.

tests/bugs/generic-type-arg-overloaded.slang.expected
```diff
@@ -1,11 +1,17 @@
 result code = -1
 standard error = {
 tests/bugs/generic-type-arg-overloaded.slang(14): error 30200: declaration of 'Stuff' conflicts with existing declaration
+struct Stuff {}
+^~~~~~
```
tfoleyNV, line 5: This is a case where it would be great to have the diagnostic indicate the `Stuff` identifier. This comes down to our decision right now to have the location of a declaration usually point at the declaration keyword (`struct`, `var`, etc.) instead of the name (it stores a separate `NameLoc` that can identify the location of the name token).

tfoleyNV (approved):
This seems like a nice boost to the quality of our diagnostics.

It would be great if we could consolidate the logic for mapping a location/offset back to a line (including its text) into a single place, rather than having to reproduce that logic in a few different places. That could be addressed in later changes, though.

PR #1702, opened 2021-02-11
Support `bit_cast` between complex types.

csyonghe:
This change adds a `bit_cast` intrinsic operation to the language that supports reinterpreting the bits between two arbitrary types.

For example,
```
struct A{int a, b;}
struct B{uint2 val;}
A a;
B b = bit_cast<B,A>(a); // OK
float2 v = bit_cast<float2, A>(a); // OK
```

Currently the user need to provide both the from-type and to-type generic arguments because the front-end cannot automatically deduce partially specified type list. Bit_cast for resource types are not supported, even on targets where resource types that maps to actual data fields.

`bit_cast`s between types with different sizes are allowed. For example:
```
bit_cast<float2, float>(1.0) == float2(1.0, 0.0)
bit_cast<float, float2>(float2(1.0, 2.0)) == float(1.0)
```

This feature is implemented by introducing a general IR operation `extractValueAtOffset(obj, offset, size)` that reads `size` bytes into an `uint` value from `offset` in `obj`. A bit cast that returns a complex type is translated to constructing the complex type from leaf values that are read from the source object using `extractValueAtOffset` operation. This same operation can also be used to implement `AnyValue` `pack` and `unpack` operations, but we leave that in a future change.


source/slang/slang-ir-extract-value-from-type.cpp
```diff
@@ -0,0 +1,278 @@
...
+        {
+            resultValue =
+                builder.emitBitAnd(uintType, resultValue, builder.getIntValue(uintType, bitMask));
+        }
+        return resultValue;
+    }
+    else
+    {
+        // The requested value crosses the boundaries of different fields.
+        // We need to extract first and second half separately, and combine them together.
```
tfoleyNV, line 247: In principle, we might have to extract more than twice, right?

It seems like the logic here eventually needs to be something like:

```
remainingBits = totalSizeInBits;
foundBits = 0;
offset = ...;
accumulator = nullptr;
while(remainingBits) {
    leafValue = findleafValue(offset);
    leafBits = leafValue.bits;
    if(leafBits > remainingBits) {
        leafBits = remainingBits;
        leafValue = mask(leafValue, leafBits);
    }
    if(foundBits > 0) {
        leafValue = shiftLeft(leafValue, foundBits);
    }
    accumulator = accumulator == null ? leafValue : or(leafValue, accumulator);
    foundBits += leafBits;
    remainingBits -= leafBits;
}
return accumulator
```

tfoleyNV (approved):

source/slang/slang-ir-extract-value-from-type.cpp
csyonghe: The call itself can recurve more than once to take care of all the cases.

tfoleyNV: Ah, I hadn't thought about that. So long as all the leaf values are power-of-two sizes, then that approach would seemingly ensure a minimal number of reads. Nice.

PR #1704, opened 2021-02-12
First part of interfaces and generics doc.

csyonghe:


tfoleyNV (approved):
This looks like a good start.

One thing we may want to think about in this chapter is having a section that specifically addresses C++ templates, both to explain how generics differ, but also to show how to accomplish certain template-y idioms in a generics-based language.

PR #1713, opened 2021-02-16
More #line improvements

jsmall-nvidia:
The following PR is a collection of ad-hoc #line output improvements. 

* Track the block closing location, so source loc can be used - used on adding return void to set loc appropriately
* Added advanceToSourceLocationIfValid, which only advances if the source location is valid
* Made sourceLoc the only way to advance
* Do a location update before a type is output (otherwise you get `#line` inbetween line type and variable in declaration)
* If we have a param, only advance if it has a defined source loc 
* Use previously set sloc if advanced with an invalid loc
* Added local-line.slang test that was used to test out the changes and that we get reasonable results
* Added source loc output to IR dumping
* Added copying sloc when creating ssa terminators 

The names of some of these functions could perhaps be improved. That in order to output more and more suitable locs, we want to output a source location on every suitable instruction. So we may want to 'advance' a source location is invalid, as doing so doesn't change the location, but it will trigger the output of the #line on output.

Another open question could be around the closingSourceLoc in BlockStmt used to fix the return void issue. It may be better placed higher in the derivation tree or elsewhere. Perhaps (as touched on in another PR) there should be a source loc range on AST nodes. That isn't touched on here, and the simplest most direct location was used.


source/slang/slang-parser.cpp
```diff
@@ -557,6 +557,16 @@ namespace Slang
...
+    bool AdvanceIf(Parser* parser, TokenType tokenType, Token* outToken)
+    {
+        if (parser->LookAheadToken(tokenType))
+        {
+            *outToken = parser->ReadToken();
+            return true;
+        }
+        return false;
+    }
+
```
tfoleyNV, line 569: Not really related to this change, but: it would be nice in theory if `AdvanceIf` *always* returned a `Token`, and then we could have a custom `bool` conversion where an invalid token evaluates as `false` and any valid token evaluates as `true`.

The same could be done for the functions like `ReadToken()`  so that when the expected token isn't there, it can return a "falsey" invalid token, allowing parser code to detect and handle certain cases of errorneous input more succintly.

Making `Token` convertible to `bool` is probably not a step to take lightly, so I wouldn't propose adding it to another change like this. (Realistically, we'd probably want a custom wrapper *subtype* of `Token` that has the conversion, so that we don't accidentally make all `Token`s convertible to `bool` by accident).

source/slang/slang-parser.cpp
```diff
@@ -3925,7 +3953,9 @@ namespace Slang
         {
             FillPosition(blockStatement);
         }
-        while (!AdvanceIfMatch(this, TokenType::RBrace))
+
+        Token closingBraceToken;
+        while (!AdvanceIfMatch(this, TokenType::RBrace, &closingBraceToken))
```
tfoleyNV, line 3958: If we do a follow-up change where we add a `SourceRange` to each and every AST node, we might want to consider making some of the tracking of source ranges in the parser more systematic/automatic. Having to capture a token from a subroutine here so that we can poke it into the `blockStatement` feels like the kind of thing we want to make more automatic if we are going to apply it broadly throughout the compiler.

(No concrete request here; just making notes for when we move on to future stuff we've talked about like documentation comments)

tfoleyNV (approved):

PR #1719, opened 2021-02-18
Make gfx library visible to external user.

csyonghe:
This change surfaces the gfx library header `render.h` to root directory and renamed to `slang-gfx.h`, and includes `slang-gfx.h` and the gfx binary in releases.


.github/workflows/release-windows.yml
```diff
@@ -50,10 +50,12 @@ jobs:
           7z a "$binArchive" slang-com-helper.h
           7z a "$binArchive" slang-com-ptr.h
           7z a "$binArchive" slang-tag-version.h
+          7z a "$srcArchive" slang-gfx.h
```
tfoleyNV, line 53: I don't think we need to derail this PR over it, but it is getting to be about time for us to decide as a team what we want our policy to be for headers. Do we want to continue dumping everything into the top level of the project?

Also, should this be `gfx.h` or `slang-gfx.h`? I suppose it depends on whether `gfx` is conceptually a stand-alone thing that just happens to depend on Slang (and just happens to be getting packaged/released with Slang for now), or if it is conceptually a sub-project of Slang.

tfoleyNV (approved):
Big +1 to not having `render.h` as the main header any more.

We should probably land this as-is and then have a separate conversation across the dev team about the location/name for this header.

PR #1724, opened 2021-02-22
Documentation markup extraction

jsmall-nvidia:
This PR provides the basis for tooling for documentation extraction. Documentation is associated with AST nodes. The documentation styles supported for the moment follow doxygens, allowing markup to be associated with declarations either before or after the declaration.

The underlying mechanism for all this to work is a small change that allows the Lexer to output comment tokens. The parser processes the token stream as before without comments. If we want to extract documentation (currently indicated on command line with -doc), after the AST has been parsed, we find all of the Decls in a module. We now want to find all the markup, and we can do this by for each source file included, re tokenizing with comment output enabled. We can then lookup the the decls location in that stream, and with some heuristics search for the the markup associated with each declaration. 

This is currently all held in ModuleMarkup. This could be used as a basis for outputting documentation, using the contents of the markup (and some suitable standard around that) and the declarations. 

That piece of work has been started in DocumenationUtil - although it doesn't do anything too interesting just yet. 

Also included is a doc.slang test, which is currently disabled, as suitable output isn't available just yet.


source/slang/slang-doc.cpp
```diff
@@ -0,0 +1,972 @@
...
+
+* If Decls hand SourceRange, then we could use the range to simplify getting the Post markup, as will be trivial to get to the 'end'
+* Need to handle preceeding * in some markup styles
+* If we want to be able to disable markup we need a mechanism to do this. Probably define source ranges.
+
+* Need a way to take the extracted markup and produce suitable markdown
+** This will need to display the decoration appropriately
+*/
+
+class MarkupExtractor
```
tfoleyNV, line 18: For terminology, should this be a `DocumentationCommentExtractor` (or `DocCommentExtractor` if we need to abbreviate)? It seems like there is potential for confusion (like I immediately had when reading the type name) between "markup" as in "we have marked up these declarations with documentation comments" vs. "markup" as in "the markup language we are using for documentation comments is markdown."

source/slang/slang-lexer.h
```diff
@@ -134,6 +144,7 @@ namespace Slang
 
         TokenFlags      m_tokenFlags;
         LexerFlags      m_lexerFlags;
+        OptionFlags     m_optionFlags;
```
tfoleyNV, line 147: I don't think that we need/want to have two different sets of flags that control the lexer's behavior. The existing `LexerFlags` are already being used by code that sits above the lexer to impose higher-level policy on it, and it seems like we can just fold something like a `kLexerFlag_PreserveComments` into the mix without trouble. Or am I missing something that makes the two kinds of flags here different?

source/slang/slang-options.cpp
```diff
@@ -531,6 +531,10 @@ struct OptionsParser
                 {
                     requestImpl->getFrontEndReq()->shouldDumpAST = true;
                 }
+                else if (argStr == "-doc")
+                {
+                    requestImpl->getFrontEndReq()->shouldDocument = true;
+                }
```
tfoleyNV, line 537: This is probably fine/reasonable for now, but it would be really nice in the long run if `slang-doc` where a different tool than `slangc`. I guess it creates the problem that they would end up needing/wanting a lot of the same options, though, so yeah, it seems best to not tackle that just yet.

tfoleyNV (dismissed):
This seems like a good start at extracting all the information.

source/slang/slang-lexer.h
jsmall-nvidia: I actually did have it like this originally, but when I made the change, but found things broke because other code assumed it could just overwrite the contents. So I split out such that 'immutable' options are separate.

PR #1725, opened 2021-02-23
Add support for GetAttributeAtVertex for D3D

tfoleyNV:
This operation was added along with the `SV_Barycentrics` system-value input, and allows for a `nointerpolation` varying input to a fragment shader to be fetched at a specific vertex index within the primitive that is causing the fragment shader to be invoked.

This change adds support for the new operations in the standard library, and also includes a test case to make sure that we emit it correctly when producing HLSL/DXIL.

This change also includes a small bug fix to our emission logic for function parameters so that we properly emit layout-related attributes for varying parameters declared directly on an entry point.
(Note that most attribute end up being declared in `struct` types in existing HLSL shaders, and our IR passes produce only global variables for attributes on GLSL; the only case this affects is inidividual scalar/vector attributes declared declared as entry-point parameters, when outputting HLSL)

Note that this change only adds support for the new function on the HLSL/DXIL path, and doesn't yet add any cross-compilation support for GLSL/SPIR-V. The reason for this is that the equivalent GLSL feature(s) appear to use a different model to the HLSL version, and we need to invent a suitable approach to align them to make portable code possible.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -1805,6 +1805,61 @@ matrix<T, N, M> fwidth(matrix<T, N, M> x)
     MATRIX_MAP_UNARY(T, N, M, fwidth, x);
 }
 
+/// Get the value of a vertex attribute at a specific vertex.
+///
+/// The `GetAttributeAtVertex()` function can be used in a fragment shader
+/// to get the value of the given `attribute` at the vertex of the primitive
+/// that corresponds to the given `vertexIndex`.
+///
+/// Note that the `attribute` must have been a declared varying input to
+/// the fragment shader with the `nointerpolation` modifier.
+///
+/// This function can be applied to scalars, vectors, and matrices of
+/// built-in scalar types.
+///
+/// Note: these functions are not curently implemented for Vulkan/SPIR-V output.
+///
+__generic<T : __BuiltinType>
+[__readNone]
+T GetAttributeAtVertex(T attribute, uint vertexIndex);
+
+/// Get the value of a vertex attribute at a specific vertex.
+///
+/// The `GetAttributeAtVertex()` function can be used in a fragment shader
+/// to get the value of the given `attribute` at the vertex of the primitive
+/// that corresponds to the given `vertexIndex`.
+///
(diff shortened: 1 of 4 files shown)


csyonghe (approved):
I wonder what will happen if `GetAttributeAtVertex` is called on a local variable. We probably need a way to distinct a direct reference to an attribute from ordinary values in the type system.

tfoleyNV:
@csyonghe Yes, we eventually need some kind of semantic checking for the validity of calls to `GetAttributeAtVertex()`, `EvaluateAttrributeAtCentroid()`, etc.

The simplest validation rule would be entirely syntactical: a valid argument must be a "attribute expression" where:

* Varying inputs to a fragment shader entry point are attribute expressions
* Member access `base.field` on an attribute expression yields an attribute expression'
* Array indexing on an attribute expression yields an attribute expression

That syntactic formulation might be a bit limiting since it doesn't allow for these operations anywhere except the top level of a fragment shader entry point, but they'd be easy-ish to enforce (we could add an attribute on the appropriate function parameter and then perform these checks around the same time we do the checks for `out` and `inout` arguments).

(The big thing that the above discussion leaves out is that some of these evaluation operations may place restrictions on what interpolation qulaifier(s) are allowed on the attribute, implying that our type system should also be propagating and tracking those qualifiers - but *only* for accesses that start with a fragment-shader varying input)

A more complex but complete solution would introduce one or more type qualifiers that indicate attribute-ness of the appropriate kind (e.g., flat interpolation vs. perspective-corrected linear interpolation, etc.) and then have type-system features for propagating those qualifiers through the relevant operations. Basically an `attribute T` should be able to be coerced to a plain `T`, but not vice versa, and then specific ops (like field access and array indexing) would know to propagate such attributes from the base type through to the result. We'd probably need to make overload resolution aware of such conversions, so that they can prefer an exact qualifier match over a match on type but not qualifier.

The thing that keeps getting at me is that such attributes are basically just rate qualifiers, so if we are going to introduce a notion of a `QualifiedType` we might as well call it something like a `RateQualifiedType` and have the rates themselves come rom syntactic declarations that can have their own associated rules...

PR #1729, opened 2021-02-26
Doc improvements

jsmall-nvidia:
* Change toString -> toText and had it append to a StringBuilder
* Added ASTPrinter, so that printing of AST types is available outside of SemanticVisitor
* Added toSlice as simpler way to get slices from string literals
* Made DeclRef/Val* have operator<<
* Improved markdown output

The ASTPrinter has a mechanism to mark where in the built up string different 'parts' are located. This can then be used to extract parts (and link etc) from a complex node, such that the details don't need to be known by the DocumentUtil. 

The change of toString, to toText is for performance/simplicity reasons. It could be argued the name 'toText' should be more like 'appendText', but toText is short and sweet, it's not going to be confused with returning a string, and the signature makes it pretty clear how it works. 

The test file (tests/doc/doc.slang) now produces:

# struct SomeStruct

## Fields

* _aField_ A field 
* _anotherField_ Multi-line  is a thing 
* _yetAnother_ A field with stuff  

## Description

A struct with some fields

# enum AnEnum

## Values 

* _Value_ An enum 
* _AnotherValue_ Another value With a multi-line comment 

## Description

An enum

# doThing

## Signature 
```
int doThing(
    int                  a,
    int                  b);
```

## Parameters

* _a_ a parameter 
* _b_ b parameter 

## Description

doThing!


tfoleyNV (approved):
This is great stuff! Thanks for doing the refactoring work included in this.

PR #1731, opened 2021-03-02
Fix issue with long identifier names in GLSL output

jsmall-nvidia:
* Added reduceLength function to String
* Added code to CLikeSourceEmitter, when scrubbing names on GLSL, will if the name is too long, reduce it's length and concatenate a hash
* Added test to test identifier length


tfoleyNV (approved):
Nice fix.

PR #1735, opened 2021-03-04
Refactor `gfx` to surface `CommandBuffer` interface.

csyonghe:
This change refactors `gfx` to expose `ICommandQueue` and `ICommandBuffer` interfaces.
The user can now explicitly control command buffer recording and submission.

This change adopts Metal's model for command buffer recording: the user must first request a `CommandEncoder` object from a command buffer, and then record concrete commands with the `CommandEncoder`. Currently we implemented three types of `CommandEncoder`s: `RenderCommandEncoder` for recording all graphics pipeline related commands, `ComputeCommandEncoder` for compute dispatch commands, and `ResourceCommandEncoder` for resource transfers.

The D3D12 and Vulkan implementation maps this interface to actual API command buffers. D3D11, OpenGL and CUDA implementations records commands into a host-side buffer and replays the commands in the immediate execution context upon command buffer submission. The immediate mode targets share the same command recording and replay infrastructure, implemented in `command-writer.h` and `immediate-renderer-base.h/cpp`.

Note that the `tests/bugs/rwstructuredbuffer-existential-in-struct.slang` test no longer passes with this change. There is still an issue in the shader-object implementation that needs to be fixed. This change temporarily disables this test to get this refactor merged in first. We need to fix shader-object issues in a future change.

With this change, `setClearColor` and `clearFrame` commands are removed from public API. The user need to specify clear color during resource creation and indicate a clear operation is need in a separate `RenderPassLayout` object. The clear operation is executed at the beginning of a render pass. The render pass model also makes it possible for the user to specify initial and final resource states.


examples/heterogeneous-hello-world/main.cpp
```diff
@@ -1,3 +1,4 @@
+#if 0
```
tfoleyNV, line 1: Maybe we should include a comment here about why the example is being disabled?

slang-gfx.h
```diff
@@ -1183,6 +1178,266 @@ struct WindowHandle
...
+        AttachmentAccessDesc* renderTargetAccess;
+        AttachmentAccessDesc* depthStencilAccess;
+    };
+};
+#define SLANG_UUID_IRenderPassLayout                                                   \
+    {                                                                                  \
+        0xdaab0b1a, 0xf45d, 0x4ae9, { 0xbf, 0x2c, 0xe0, 0xbb, 0x76, 0x7d, 0xfa, 0xd1 } \
+    }
+
+class ICommandEncoder : public ISlangUnknown
```
tfoleyNV, line 1237: Because the validity of encoders isn't really coupled to reference counts, should we make them not use `ISlangUnknown` and require `ComPtr` or similar? We could conceivably make them be non-reference-counted objects that just "magically" go away on `endEncoding`.

I don't know whether that would actually be wise, or if keep the reference-counting around is a net win.

slang-gfx.h
```diff
@@ -1183,6 +1178,266 @@ struct WindowHandle
...
+    // object by calling one of the `encode*Commands` methods again.
+    virtual SLANG_NO_THROW void SLANG_MCALL encodeRenderCommands(
+        IRenderPassLayout* renderPass,
+        IFramebuffer* framebuffer,
+        IRenderCommandEncoder** outEncoder) = 0;
+    ComPtr<IRenderCommandEncoder>
+        encodeRenderCommands(IRenderPassLayout* renderPass, IFramebuffer* framebuffer)
+    {
+        ComPtr<IRenderCommandEncoder> result;
+        encodeRenderCommands(renderPass, framebuffer, result.writeRef());
```
tfoleyNV, line 1352: This is in no way a criticism of this PR, but we should eventually decide whether the interfaces in `gfx` should contain these convenience operations or not, and if we *do* keep them, then we probably need to change them so that they do something with the `Result`s of the operations they call (something like wrap the call in `GFX_CHECK_RESULT` or similar, so that the user can define that macro to be something that `throw`s or `assert`s).

I notice in this case the operations return `void`, and I don't know if that is intentional, because it seems like a command buffer created for a compute-only renderer/device/queue should be able to fail on `encodeRenderCommands()`.

slang-gfx.h
```diff
@@ -1183,6 +1178,266 @@ struct WindowHandle
...
+    {                                                                                  \
+        0x5d56063f, 0x91d4, 0x4723, { 0xa7, 0xa7, 0x7a, 0x15, 0xaf, 0x93, 0xeb, 0x48 } \
+    }
+
+class ICommandQueue : public ISlangUnknown
+{
+public:
+    enum class QueueType
+    {
+        Graphics
```
tfoleyNV, line 1386: Any reason not to include the other cases here? I guess it is more implementation work to support all the case and our apps only ever want graphics for now.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -954,13 +956,367 @@ class CUDARenderer : public RendererBase
...
+                cudaMemcpyDefault);
+        }
+
+        void uploadBufferData(IBufferResource* dst, size_t offset, size_t size, void* data)
+        {
+            auto dstImpl = static_cast<MemoryCUDAResource*>(dst);
+            cudaMemcpy((uint8_t*)dstImpl->m_cudaMemory + offset, data, size, cudaMemcpyDefault);
+        }
+
+        void execute(CommandBufferImpl* commandBuffer)
```
tfoleyNV, line 1279: It seems like the implementations that want to use user-space command buffers could all implement a common interface like `IImmediateModeCommands` and then we could use a single `execute()` implementation that can work with any of them rather than need to implement the walking of the command list multiple times.

Maybe that wouldn't be worth the cost, because then the CUDA renderer would need stub functions for all the commands that could conceivably be recorded but that it doesn't support.

tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -51,7 +50,7 @@ using namespace Slang;
 
 namespace gfx {
 
-class D3D11Renderer : public GraphicsAPIRenderer
+class D3D11Renderer : public ImmediateRendererBase
```
tfoleyNV, line 53: Oh, never mind, you did introduce a base class, and it is just that CUDA was different enough to not warrant using it. Good.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -841,93 +638,1214 @@ class D3D12Renderer : public GraphicsAPIRenderer
...
+                    D3D12_INDEX_BUFFER_VIEW indexBufferView;
+                    indexBufferView.BufferLocation =
+                        m_boundIndexBuffer->m_resource.getResource()->GetGPUVirtualAddress() +
+                        m_boundIndexOffset;
+                    indexBufferView.SizeInBytes =
+                        UINT(m_boundIndexBuffer->getDesc()->sizeInBytes - m_boundIndexOffset);
+                    indexBufferView.Format = m_boundIndexFormat;
+
+                    m_d3dCmdList->IASetIndexBuffer(&indexBufferView);
+                }
```
tfoleyNV, line 1189: Not relevant to this PR, but we may be close to a point where we can move some of the logic that is in `prepareDraw()` to just be in the relevant state-setting function. E.g., there seems to be no reason why we can't call `IASetIndexBuffer` as part of the `setIndexBuffer` implementation at this point.

The main sticking point that remains is dealing with the way that the root shader object plus the pipeline state combine to produce the final state, but that seems almost inevitable to have to deal with.

tools/gfx/dummy-render-pass-layout.h
```diff
@@ -0,0 +1,30 @@
...
+// concept.
+
+#include "slang-gfx.h"
+#include "slang-com-helper.h"
+#include "core/slang-basic.h"
+
+namespace gfx
+{
+
+class DummyRenderPassLayout
```
tfoleyNV, line 15: Naming nit: the use of "dummy" as a name in code is something people could find offensive even if we don't intend it. It might be better to have this be called a "placeholder," "proxy," or "simple" render pass layout.

tfoleyNV (approved):
This is a huge change so I can't claim to have read every line, but the overall design is good and the implementations all make sense in the big picture.

examples/gpu-printing/main.cpp
```diff
@@ -175,19 +175,26 @@ Result execute()
...
+    auto commandBuffer = queue->createCommandBuffer();
+    auto encoder = commandBuffer->encodeComputeCommands();
     // TODO: need to copy a zero into the start of the print buffer!
 
     gDescriptorSet->setResource(0, 0, printBufferView);
-    gRenderer->setDescriptorSet(PipelineType::Compute, gPipelineLayout, 0, gDescriptorSet);
-
-    gRenderer->setPipelineState(gPipelineState);
-    gRenderer->dispatchCompute(1, 1, 1);
+    encoder->setDescriptorSet(PipelineType::Compute, gPipelineLayout, 0, gDescriptorSet);
```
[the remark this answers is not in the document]
csyonghe: This is not needed. Removed.

tools/gfx/dummy-render-pass-layout.h
csyonghe: Renamed to "Simple".

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: That's right, should clean up in a future pass.

examples/heterogeneous-hello-world/main.cpp
csyonghe: Commented and moved to `examples/experimental` subdirectory along with `model-viewer`.

slang-gfx.h
csyonghe: Yeah, I simply don't have time to implement all the other cases yet.

PR #1736, opened 2021-03-05
Cache stdlib when creating global session.

### LINKED ISSUE #1655: Enable stdlib embedding for development cycles
We already have a build process that generates compiled binary for stdlib and embeds it in slang.dll. This build process is used on CI servers for testing/deployment, but not enabled by default in our project files for daily development. It would be nice to have daily Slang development benefit from the stdlib serialization work by reducing the frequency of a stdlib compile.

There are some difficulties enabling stdlib embedding for daily development builds. On one hand we don't want to run slangc to build stdlib on every build (as it does on CI servers) since that would greatly lengthen our build times, on the other hand we don't want to always reuse an existing stdlib binary since that can become out-of-date due to any Slang source changes.

One strategy that offers most of the benefits is to enable a development build process where `slang.dll` attempts to find a pre-build stdlib binary file from a certain file system location (e.g. the same path as the slang.dll module, or the current path or some pre-defined system temp path), and verifies if the binary is generated by the same version of the `slang.dll`. If the pre-existing stdlib binary passes verifcation then just load the binary instead of generating one. This process is strictly a win compared to our current development process since 1) does not involve generating stdlib binary during a Slang build so it doesn't lengthen build time; 2) saves us the time to repetitively compile stdlib if the same slang.dll is being used in consecutive debugging launches.

csyonghe:
Makes slang cache the compiled stdlib during global session creation.
The cache file contains a timestamp of the `slang.dll/so` that creates the stdlib along with a serialization of the stdlib. The cache file is stored along side the slang shared library.
Upon global session creation, slang will look for the cache file and attempt to use it. The timestamp stored in the cache file will be compared against the timestamp of the current running `slang.dll/so` before actually using it.

Fixes #1655.

### DIFF
--- a/source/core/slang-shared-library.cpp
+++ b/source/core/slang-shared-library.cpp
@@ -119,12 +119,12 @@ String SharedLibraryUtils::getSharedLibraryFileName(void* symbolInLib)
     return String::fromWString(filenameBuffer);
 
 #elif defined(__linux__)
-    Dl_info DlInfo;
-    if (!dladdr((void*)func, &DlInfo))
+    Dl_info dllInfo;
+    if (!dladdr(symbolInLib, &dllInfo))
     {
         return String();
     }
-    return DlInfo.dli_fname;
+    return dllInfo.dli_fname;
 
 #else
     return String();
--- a/source/core/slang-shared-library.cpp
+++ b/source/core/slang-shared-library.cpp
@@ -5,6 +5,18 @@
 #include "slang-io.h"
 #include "slang-string-util.h"
 
+#if defined(_WIN32)
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#elif defined(__linux__)
+#include <dlfcn.h>
+#endif
+#include <sys/stat.h>
+#include <sys/types.h>
+#ifndef WIN32
+#    include <unistd.h>
+#endif
+
 namespace Slang
 {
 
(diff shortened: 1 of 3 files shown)


source/core/slang-shared-library.h
```diff
@@ -94,6 +94,13 @@ class TemporarySharedLibrary : public DefaultSharedLibrary
     String m_path;
 };
 
+class SharedLibraryUtils
```
jsmall-nvidia, line 97: Perhaps these functions should be placed in 'SharedLibrary' in slang-platform.h/.cpp? It could be argued that SharedLibrary class name is not great in those files. 

It could be argued that slang-shared-library should contain all this stuff (not just the 'object' version of the interface.

source/core/slang-shared-library.h
```diff
@@ -94,6 +94,13 @@ class TemporarySharedLibrary : public DefaultSharedLibrary
...
+{
+public:
+    static String getSharedLibraryFileName(void* symbolInLib);
+    static uint64_t getSharedLibraryTimestamp(void* symbolInLib);
```
jsmall-nvidia, line 101: Putting the getting at time stamp from a file into the slang-io.h/cpp (say into Path). And having typedef def the timstamp type might be nice.

source/core/slang-shared-library.cpp
```diff
@@ -87,4 +99,50 @@ void* DefaultSharedLibrary::findSymbolAddressByName(char const* name)
     return SharedLibrary::findSymbolAddressByName(m_sharedLibraryHandle, name);
 }
 
+
+String SharedLibraryUtils::getSharedLibraryFileName(void* symbolInLib)
```
jsmall-nvidia, line 103: Having this return a SlangResult and take a StringBuffer for output, does have the benefit of the function being to communicate more about it's failure. Admittedly it might be a bit clumsier to use.

source/slang/slang-api.cpp
```diff
@@ -17,6 +19,65 @@ SLANG_API SlangSession* spCreateSession(const char*)
     return globalSession.detach();
 }
 
+// Attempt to load a previously compiled stdlib from the same file system location as the slang dll.
+// Returns SLANG_OK when the cache is sucessfully loaded.
+// Also returns the filename to the stdlib cache and the timestamp of current slang dll.
+SlangResult tryLoadStdLibFromCache(
```
jsmall-nvidia, line 25: May want to make static and perhaps _ prefix.

jsmall-nvidia (approved):
This is a nice improvement to get benefits of serialized stdlib during development.

PR #1739, opened 2021-03-08
Refactor window library.

csyonghe:
Renames `graphics-app-framework` to `platform` and redesigned the window library interface to get ready for adding a linux implementation.


examples/gpu-printing/main.cpp
```diff
@@ -201,21 +198,12 @@ Result execute()
 
 };
 
-// This "inner" main function is used by the platform abstraction
-// layer to deal with differences in how an entry point needs
-// to be defined for different platforms.
-//
-void innerMain(ApplicationContext* context)
+int main()
```
tfoleyNV, line 201: Design nitpick: the choice to wrap the `main()` function even for console applications was a conscious choice.

Most importantly, on Windows we are likely to need/want a `wmain()` entry point for Unicode support, and that entry point will need to convert all the command-line arguments from their wide character encoding to UTF-8. The inner main should not have to think about those details, and we want our command-line apps to be Unicode-compliant by default.

There's also some details here on many Linux platforms where we likely want to set the C locale to a region-independent but Unicode-compatible locale like `"C.UTF08"`.

Having a distinct "context" type here that gets passed down into the code also avoids the issue of having an "ambient authority" - it makes it so that application code (including things like plugins) need to have the relevant platform-specific objects passed in directly in order to interact with them.

Eventually we may need/want things like the application context to be a COM interface (like what we do for the compiler and `gfx` APIs), so that we can easily expose varying per-target facilities (including ones that cannot just be detected at compile time).

examples/hello-world/main.cpp
```diff
@@ -230,12 +229,12 @@ Slang::Result initialize()
...
     //
-    WindowDesc windowDesc;
+    platform::WindowDesc windowDesc;
     windowDesc.title = "Hello, World!";
     windowDesc.width = gWindowWidth;
     windowDesc.height = gWindowHeight;
-    gWindow = createWindow(windowDesc);
-
+    gWindow = platform::Application::createWindow(windowDesc);
+    gWindow->events.mainLoop = [this]() { renderFrame(); };
```
tfoleyNV, line 237: I'd kind of rather see the event-handling stuff handled by having a "delegate" sort of COM interface that handles events for the application/window/whatever. Using fine-grained event-handler callbacks like this seems like overkill.

(I guess it also weird me out to be attaching the handlers after creation, rather than as part of creation)

examples/hello-world/main.cpp
```diff
@@ -323,8 +322,9 @@ Slang::Result initialize()
     swapchainDesc.height = gWindowHeight;
     swapchainDesc.imageCount = kSwapchainImageCount;
     swapchainDesc.queue = gQueue;
-    gSwapchain = gRenderer->createSwapchain(
-        swapchainDesc, gfx::WindowHandle::FromHwnd(getPlatformWindowHandle(gWindow)));
+    gfx::WindowHandle windowHandle;
+    memcpy(&windowHandle, &gWindow->getNativeHandle(), sizeof(windowHandle));
```
tfoleyNV, line 326: Should `getNativeHandle()` be something like:

```
Result getNativeHandle(size_t dataSize, void* outData);
```
?

It seems like we don't have a lot of protection against copying out the native handle expecting one kind and then getting another, but maybe that is unavoidable. I guess in theory we could have `gfx::WindowHandle` include a tag to identify the flavor of the handle it holds, and then the `platform` layer could specifically return a `gfx::WindowHandle` instead of something opaque.

examples/hello-world/main.cpp
```diff
@@ -545,18 +539,17 @@ void innerMain(ApplicationContext* context)
...
     app.finalize();
+
+    return 0;
 }
 
 // This macro instantiates an appropriate main function to
 // invoke the `innerMain` above.
 //
-GFX_UI_MAIN(innerMain)
+PLATFORM_UI_MAIN(innerMain)
```
tfoleyNV, line 555: Should this be `SLANG_PLATFORM_UI_MAIN` so that we always have the `SLANG_` prefix on our code, other than `gfx`?

examples/shader-toy/main.cpp
```diff
@@ -82,7 +85,7 @@ void diagnoseIfNeeded(slang::IBlob* diagnosticsBlob)
 {
     if( diagnosticsBlob != nullptr )
     {
-        reportError("%s", (const char*) diagnosticsBlob->getBufferPointer());
+        printf("%s", (const char*) diagnosticsBlob->getBufferPointer());
     }
```
tfoleyNV, line 89: What is the rationale for removing this as a bit of utility code that applications can use? I realize that `reportError` isn't a full-fledged logging system, but I would kind of expect us to move in that direction (making our logging more centralized and feature-full) rather than in the opposite direction.

I feel like I'm missing part of the rationale for this PR.

tools/platform/performance-counter.h
```diff
@@ -0,0 +1,30 @@
+#ifndef PLATFORM_PERFORMANCE_COUNTER_H
+#define PLATFORM_PERFORMANCE_COUNTER_H
+
+#include <chrono>
+
+namespace platform
+{
+typedef std::chrono::high_resolution_clock::time_point TimePoint;
+typedef std::chrono::high_resolution_clock::duration Duration;
```
tfoleyNV, line 9: It seems a little weird that there is code for a `std::function` replacement earlier in this PR, but then here we are just relying on `std::chrono` directly.

I have a general preference for *not* relying on services from `std` if we can get them from native platform APIs instead without too much trouble. If this is really the best way to get a high-performance timer that we can use everywhere, then so be it.

tfoleyNV (approved):
I'm totally happy with the renaming of this code, but I'm a bit confused about some of the refactoring just because I don't see why it was an important change/improvement over what was already there.

PR #1741, opened 2021-03-09
Swapchain resize and rename to `IDevice`

csyonghe:
This PR is a combination of these changes:
1. Add resize support to `ISwapchain`.
2. Refactor the implementations of Swapchain so that D3D11 and D3D12 share the same implementation, and remove `VulkanSwapchain` class from the Vulkan swapchain implementation (unnecessary indirection).
3. Rename `IRenderer` to `IDevice`.
4. Add `IDevice::getDeviceInfo()` to replace the exising C `gfxGetXXX` functions.
5. Rename `RendererType` enum to `DeviceType`.
6. Add `DeviceType::Default` enum that allows gfx to automatically select the underlying graphics API on each platform.
7. Allow user to create resizeable or fixed size `platform::Window`.
8. Modified `ShaderToy` example to demonstrate resizing swapchain.


slang-gfx.h
```diff
@@ -1427,16 +1428,47 @@ class ISwapchain : public ISlangUnknown
...
+    /// Returns the back buffer image at `index`.
+    virtual SLANG_NO_THROW Result SLANG_MCALL
+        getImage(uint32_t index, ITextureResource** outResource) = 0;
+
+    /// Present the next image in the swapchain.
+    virtual SLANG_NO_THROW Result SLANG_MCALL present() = 0;
+
+    /// Returns the index of next back buffer image that will be presented in the next
+    /// `present` call. If the swapchain is invalid/out-of-date, this method returns -1.
+    virtual SLANG_NO_THROW int SLANG_MCALL acquireNextImage() = 0;
```
tfoleyNV, line 1441: Should this use a `Result` to indicate the invalidity cases, and then return the image index via an output parameter?

slang-gfx.h
```diff
@@ -1427,16 +1428,47 @@ class ISwapchain : public ISlangUnknown
...
+{
+    DeviceType deviceType;
+
+    BindingStyle bindingStyle;
+
+    ProjectionStyle projectionStyle;
+
+    /// An projection matrix that ensures x, y mapping to pixels
+    /// is the same on all targets
+    float identityProjectionMatrix[16];
```
tfoleyNV, line 1462: It is a bit weird to call this an "identity" matrix. I have usually referred to it as a "fixup" matrix, but I'm not even sure that is a great name. Basically we are mapping from the clip space of D3D11/12 to the proper clip space for the chosen target, right?

slang-gfx.h
```diff
@@ -1427,16 +1428,47 @@ class ISwapchain : public ISlangUnknown
...
     {                                                                                \
         0xbe91ba6c, 0x784, 0x4308, { 0xa1, 0x0, 0x19, 0xc3, 0x66, 0x83, 0x44, 0xb2 } \
     }
 
-class IRenderer: public ISlangUnknown
+struct DeviceInfo
+{
+    DeviceType deviceType;
+
+    BindingStyle bindingStyle;
```
tfoleyNV, line 1456: I'd like to believe that `BindingStyle` is going to be completely superfluous once shader objects are under the hood of the API, right?

I guess the hope is that the same is true of `ProjectionStyle`.

slang-gfx.h
tfoleyNV: I keep wondering if we can/should just do the fixup math as part of how Slang compiles vertex shaders, because pushing this back on the application is so gross.

tfoleyNV (approved):

PR #1744, opened 2021-03-10
Add Linux support to `platform` and `gfx`

csyonghe:
`platform` now supports creating windows on Linux via `Xlib`.
`gfx` now supports creating Vulkan devices on Linux.

Latest version of glslang uses `pthread`, therefore we need to ensure pthread is loaded before loading `libslang-glslang.so`.

### DIFF
--- a/build/visual-studio/platform/platform.vcxproj
+++ b/build/visual-studio/platform/platform.vcxproj
@@ -27,25 +27,25 @@
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v140</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v140</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v140</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v140</PlatformToolset>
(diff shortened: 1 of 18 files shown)


build/visual-studio/platform/platform.vcxproj
```diff
@@ -27,25 +27,25 @@
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
```
tfoleyNV, line 30: I'm a bit nervous about introducing Yet Another DLL, but if this is only impactful to our examples and tests, then this is fine.

In the long term we really need to figure out how we want to factor our code into dynamic libraries for long-term use. My tendency is to assume we want a kind of Slang "core" DLL that includes what amounts to the language "runtime" plus all these platform-abstraction services like window creation (basically the core/stdlib stuff), and then have a separate slang compiler DLL that provides the current public compilation API. It seems like this `platform` code is stuff that basically belongs in the core/stdlib DLL in the long run.

source/core/slang-platform.cpp
```diff
@@ -171,7 +171,7 @@ SLANG_COMPILE_TIME_ASSERT(E_OUTOFMEMORY == SLANG_E_OUT_OF_MEMORY);
 {
     handleOut = nullptr;
 
-	void* h = dlopen(platformFileName, RTLD_NOW | RTLD_LOCAL);
+	void* h = dlopen(platformFileName, RTLD_NOW | RTLD_GLOBAL);
```
tfoleyNV, line 174: This is a subtle change to be making, and we should be careful about it.

The use of `RTLD_LOCAL` was intentional here because it means that the compiler can load something like `dxcompiler.so` (for a Unix-y build of dxc) from the specified path and *not* have to worry about other code in the same process that might also want to load dxc (but which might want a different version).

I guess the tricky bit is that for some of the stuff like `pthread` we really do want to have a global load, since we are actually loading it in order to influence what subsequently gets loaded.

I think we can let this change in for right now, but it seems like we actually need/want to update the shared library loader interface to support some minimal flags (or a `loadGlobal` sort of entry point), so that we can specify which of the two modes we are looking for.

tools/gfx/immediate-renderer-base.cpp
```diff
@@ -57,8 +57,8 @@ class CommandBufferImpl : public ICommandBuffer, public RefObject
             *outObject = nullptr;
             return SLANG_E_NO_INTERFACE;
         }
-        virtual SLANG_NO_THROW uint32_t SLANG_MCALL addRef() { return 1; }
-        virtual SLANG_NO_THROW uint32_t SLANG_MCALL release() { return 1; }
+        virtual SLANG_NO_THROW uint32_t SLANG_MCALL addRef() override { return 1; }
+        virtual SLANG_NO_THROW uint32_t SLANG_MCALL release() override { return 1; }
```
tfoleyNV, line 61: Is there a reason for using `override` over `SLANG_OVERRIDE`. I guess we might be at the point where all the compilers we can work with support `override` so that there is no point in trying to be defensive against older compilers...

tools/gfx/render.cpp
```diff
@@ -90,6 +90,12 @@ extern "C"
                 return SLANG_FAIL;
             }
             break;
+#elif SLANG_LINUX_FAMILY
+            case DeviceType::Default:
+            case DeviceType::Vulkan:
+            {
+                return createVKDevice(desc, outDevice);
+            }
```
tfoleyNV, line 98: Shouldn't we also support creating a CUDA device on Linux? That option seems easy enough to support, right?

tfoleyNV (approved):

tools/gfx/render.cpp
csyonghe: You are right.

tools/gfx/immediate-renderer-base.cpp
csyonghe: Yeah, I doubt if there are any compilers that doesn't support override.

source/core/slang-platform.cpp
csyonghe: Agreed, we should change the interface of dll loader to support the flag.

build/visual-studio/platform/platform.vcxproj
csyonghe: The current way means that Slang itself doesn't need to depend on things like XLib, which is nice.

PR #1745, opened 2021-03-10
stdlib documentation

jsmall-nvidia:
* Lots of fixes and improvements to doc output
* Added requirements to list underlying requirements for a function or method
* Added target listing the source targets that will output for 
* Control line visibility
* Used visibility controls on stdlib to remove some of the internal stuff
* Combine both stdlib files into single output
* First pass of stdlib output

Issues:

* Types are not in general output correctly (just uses the toText of Type derived types)
* No links/navigation
* The 'targets' can only determine whats available based on markup - for C++ there's none. For HLSL very little is marked up and so code assumes it's always available, which is occasionally wrong.

An example is below

# `Texture1DMS<T>.Load`

## Signature 

```
/// See Target Availability 1
T Texture1DMS<T>.Load(
    vector<int,1>        location,
    int                  sampleIndex);
T Texture1DMS<T>.Load(
    vector<int,1>        location,
    int                  sampleIndex,
    vector<int,1>        offset);
/// See Target Availability 2
T Texture1DMS<T>.Load(
    vector<int,1>        location,
    int                  sampleIndex,
    vector<int,1>        offset,
    out uint             status);
```

## Requirements

`GLSL GL_EXT_samplerless_texture_functions`

## Target Availability

1. `GLSL`, `HLSL`
2. `HLSL`

## Parameters

* `location`
* `sampleIndex`
* `offset`
* `status`


tfoleyNV:
I think the target availability stuff will get better over time as we make the capability system more codified and can thus use its information very directly to drive all of this.

What I'd like to see in terms of the availability information is something like a list of targets+capabilities for which the function/type/whatever is available, e.g.:

* `hlsl & sm_6_1`
* `glsl_440 & GL_EXT_whatever & spirv_1_4`
* ...

That kind of listing would seemingly allow us to both list where the feature is available and also what the conditions are on that availability.

source/slang/slang-ast-print.cpp
```diff
@@ -75,7 +80,27 @@ void ASTPrinter::_addDeclPathRec(const DeclRef<Decl>& declRef)
...
+    else if (auto namespaceDeclRef = parentDeclRef.as<NamespaceDecl>())
+    {
+        _addDeclPathRec(namespaceDeclRef);
+        // Hmm, it could be argued that we follow the . as seen in AggType as is followed in some other languages
+        // like Java.
+        // That it is useful to have a distinction between something that is a member/method and something that is
+        // in a scope (such as a namespace), and is something that has returned to later languages probably for that
+        // reason (Slang accepts . or ::). So for now this is follows the :: convention.
+        //
+        // It could be argued them that the previous '.' use should vary depending on that distinction.
```
tfoleyNV, line 94: FWIW I personally think of `.` as the correct operator in Slang, and `::` is just there for compatibility. Sometimes I will use `::` when discussing language features to draw the distinction between `someObj.someInstanceMethod` and `SomeType::someInstanceMethod` where the latter is a static reference to a per-instance method, but I'm not aware of any cases where a `::` operator is strictly *needed* in the language.

The main beneficial thing `::` provides in C++ is an ability to explicitly name the global scope (Rust ran into the problem at some point of having no way to unambiguously name a top-level module - not sure how they fixed it). The big problem is that using `::` as an infix operator at times and a prefix operator-like thing at other times is itself a source of ambiguity.

If possible I'd like to avoid sending our users down the C++-minded `::` path any more than they are already inclined to go that way. However, if we discover a strong reason why having `::` as a distinct scoping operator is important then I will gladly retract all of this.

tfoleyNV (dismissed):

PR #1747, opened 2021-03-11
Change representation of initial data for textures

tfoleyNV:
Before this change, initial data for a texture has been provided with the `ITextureResource::Data` type, where a call to `IDevice::createTexture()` would take zero or one `Data` and, if present, use it to initialize all the subresources of a texture.

The organization of `Data` was not actually quite how its own documentation comment described it (the implementations didn't agree with the comment), and while it aggressively factored out redundancies (e.g., only storing the stride for each mip level once, instead of once per subresource for large arrays), the result was that setting up a `Data` correcty was a bit confusing.

This change makes the initial data for a texture using a `SubresourceData` type that is almost identical to what D3D11 uses, so that developers are more likely to be comfortable filling it in. All of the existing implementations were easily adapted to use the new type, so it seems like a net win.

Note: Both Vulkan and D3D11 do away with the idea of initializing a texture with data as part of allocating it, and we might eventually want to do the same given the complexity that this system entails. The main reason to preserve this detail is for better compatibility with D3D11, where immutable textures/buffers need to have their data specified at creation time. It seems good to preserve the ability to have immutable resources on target APIs where this distinction could affect performance (e.g., immutable resources do not need state/transition tracking on APIs like D3D11).

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -519,14 +519,9 @@ public:
             {
             case IResource::Type::Texture1D:
             case IResource::Type::Texture2D:
-                {
-                    return numMipMaps * arrSize;
-                }
             case IResource::Type::Texture3D:
                 {
-                    // can't have arrays of 3d textures
-                    assert(this->arraySize <= 1);
-                    return numMipMaps * this->size.depth;
+                    return numMipMaps * arrSize;
                 }
             case IResource::Type::TextureCube:
                 {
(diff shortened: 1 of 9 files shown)


csyonghe (dismissed):

tfoleyNV:
I neglected to fix CUDA here, because it doesn't build by default. I'll push a fix shortly.

tfoleyNV:
One thing I found when pushing through these changes is that several of our targets don't properly handle texture input data that isn't already tightly packed (they assume the row/layer stride are the same as the row/layer size). We should probably fix that sooner or later, but it is out of scope for these changes.

csyonghe (approved):

PR #1750, opened 2021-03-11
Add a CPU renderer implementation

tfoleyNV:
This change adds a CPU back-end to `gfx` and ensures that most of our existing CPU tests pass when using it.

Detailed notes:

* Most of the CPU renderer implementation is copy-pasted from the CUDA case, so they share a lot of similar logic

* The main addition to the CPU renderer is a semi-complete implementation of host-memory textures. The logic here handles all the main shapes (Buffer, 1D, 2D, 3D, Cube) and all the currently-supported `Format`s that are sample-able as-is (no D24S8). The implementation is not intended to be fast, and it currently only does nearest-neighbor sampling, but otherwise it tries to avoid cutting too many corners and should be ar reasonable starting point for a more complete (but not performance-oriented) implementation.

* Refactored the CPU prelude `IRWTexture` interface to inherit from `ITexture`, since in most cases a single type will end up implementing both. It might be worth it to collapse it all down to a single interface later.

* Changed the CPU prelude `ITexture`/`IRWTexture` interface so that it takes both a pointer *and* a size for output arguments. This change seems necessary to allow a shader variable declared as a `Texture2D<float>` to fetch a single `float` when the underlying texture might be using RGBA32F.

* Added to the `IComponentType` public API so that we can query a "host callable" for an entry point and not just a binary.

* Turned off the `-shaderobj` flag on two tests that weren't yet compatible with shader objects but still had the flag left in on the path (since previously the CPU path always used the non-`gfx` non-shader-object logic anyway)

* Disabled one test (`dynamic-dispatch-11`) that relied on the `ConstantBuffer<IInterface>` idiom that we know we are planning to chagne soon anyway.

* Made a few changes to the CUDA path to bring it into line with what I added for the CPU path. These were mostly bug fixes around indexing logic for sub-objects and resources.

### DIFF
--- a/tools/gfx/cpu/render-cpu.cpp
+++ b/tools/gfx/cpu/render-cpu.cpp
@@ -1,0 +1,1796 @@
+// render-cpu.cpp
+#include "render-cpu.h"
+
+#include "slang.h"
+#include "slang-com-ptr.h"
+#include "slang-com-helper.h"
+#include "core/slang-basic.h"
+#include "core/slang-blob.h"
+
+#include "../command-writer.h"
+#include "../renderer-shared.h"
+#include "../slang-context.h"
+
+#define SLANG_PRELUDE_NAMESPACE slang_prelude
+#include "prelude/slang-cpp-types.h"
+
+namespace gfx
+{
+using namespace Slang;
+
+class CPUBufferResource : public BufferResource
+{
+public:
+    CPUBufferResource(const Desc& _desc)
+        : BufferResource(_desc)
+    {}
+
+    ~CPUBufferResource()
+    {
+        if (m_data)
+        {
+            free(m_data);
+        }
+    }
+
+    SlangResult init()
+    {
+        m_data = malloc(m_desc.sizeInBytes);
+        if(!m_data) return SLANG_E_OUT_OF_MEMORY;
+        return SLANG_OK;
+    }
+
+    SlangResult setData(size_t offset, size_t size, void const* data)
+    {
+        memcpy((char*)m_data + offset, data, size);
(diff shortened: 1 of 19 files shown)


tools/gfx/cpu/render-cpu.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    virtual SLANG_NO_THROW Result SLANG_MCALL
+        init(IDevice* device, CPUShaderObjectLayout* typeLayout);
+
+    CPUShaderObjectLayout* getLayout()
+    {
+        return static_cast<CPUShaderObjectLayout*>(m_layout.Ptr());
+    }
+
+#if 0
+    virtual SLANG_NO_THROW Result SLANG_MCALL initBuffer(IDevice* device, size_t bufferSize)
```
csyonghe, line 773: Remove?

tools/gfx/cpu/render-cpu.cpp
```diff
@@ -0,0 +1,1796 @@
...
+                if (uuid == GfxGUID::IID_ISlangUnknown ||
+                    uuid == GfxGUID::IID_IResourceCommandEncoder)
+                {
+                    *outObject = static_cast<IResourceCommandEncoder*>(this);
+                    return SLANG_OK;
+                }
+                *outObject = nullptr;
+                return SLANG_E_NO_INTERFACE;
+            }
+            virtual SLANG_NO_THROW uint32_t SLANG_MCALL addRef() { return 1; }
```
csyonghe, line 1234: Need override here to silence warnings on Linux

csyonghe (approved):

PR #1751, opened 2021-03-12
Improvements in Docs requirements/availability

jsmall-nvidia:
* Have single 'Availability' concept in doc - that lists availability of a callable, with it's requirements on that target
* Incorporate the capability system
* Simpler mechanism of finding requirements

It might make more sense to use a CapabilityAtom instead of CodeGenTarget.
There may be a better way to convey this information - as 'compatibility' is now associated with an output (source) target. Which doesn't make much sense if say outputting directly without source. So perhaps associating with the binary format - DXIL, SPIR-V, DXBC, PTX, (CPU?). We'd want the nuance of what is possible on DXIL and DXBC represented too in that scenario.

**NEW OUTPUT**

# `RWByteAddressBuffer.InterlockedAddF32`

## Signature 

```
void RWByteAddressBuffer.InterlockedAddF32(
    uint                 byteAddress,
    float                valueToAdd,
    out float            originalValue);
void RWByteAddressBuffer.InterlockedAddF32(
    uint                 byteAddress,
    float                valueToAdd);
```

## Availability

**HLSL** `NVAPI` **CUDA** `SM 2.0` 

## Parameters

* `byteAddress`
* `valueToAdd`
* `originalValue`

**PREVIOUS OUTPUT**

# `RWByteAddressBuffer.InterlockedAddF32`

## Signature 

```
void RWByteAddressBuffer.InterlockedAddF32(
    uint                 byteAddress,
    float                valueToAdd,
    out float            originalValue);
void RWByteAddressBuffer.InterlockedAddF32(
    uint                 byteAddress,
    float                valueToAdd);
```

## Requirements

`CUDA SM 2.0`, `NVAPI`

## Target Availability

`CUDA`, `HLSL`

## Parameters

* `byteAddress`
* `valueToAdd`
* `originalValue`


source/slang/slang-capability.cpp
```diff
@@ -127,6 +127,31 @@ CapabilityAtom findCapabilityAtom(UnownedStringSlice const& name)
     return CapabilityAtom::Invalid;
 }
 
+bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base)
```
tfoleyNV, line 130: This is good to have, but I believe the capability-set stuff also has support for querying a compacted/filtered version of a capability set. That set should only include atoms that aren't already implied by other atoms in the set (e.g., it should not include both `spirv_1_3` and `spirv_1_4` because the latter implies the former). It might be enough to use that filtering logic and then directly enumerate the result knowing it won't have any atoms that should be skipped.

(I'll have to look at the actual logic later to see, and it is possible that I'm imagining that I actually implemented stuff in `CapabilitySet`)

tfoleyNV (approved):
Looks good for now. I think a lot of this can be refined once we replace all the distinct mechanisms with just capabilities.

PR #1755, opened 2021-03-12
Preliminary docs on 'Doc System'.

jsmall-nvidia:


docs/doc-system.md
```diff
@@ -0,0 +1,114 @@
...
+/**
+ ... text ... (JavaDoc style)
+ */
+void someFunctionA() {} 
+ 
+/*!
+ .. text .. (QT style)
+ another line
+ */ 
+void someFunctionB() {}
```
tfoleyNV, line 16: My two cents: I think we should make the documentation of the doc system primarily show/encourage a consistent style for how doc comments are written (e.g., only using `/** ... */` and `/// ...`), and then we can document the other available conventions as a later section on something like "syntax compatibility" or something like that.

While the `slang-doc` mode/tool is currently only something we are using internally, the goal would be for it to eventually be something Slang users are encouraged to make use of, and in that case we are responsible or showing them what we think "good" Slang codebases will look like.

docs/doc-system.md
```diff
@@ -0,0 +1,114 @@
...
+/// another line
+void someFunctionC() {}
+ 
+//!... text ...  (QT Multi line)
+//! another line
+void someFunctionD() {}
+ 
+```
+ 
+All of the above examples will add the documentation for the declaration that appears after them. Also note that this slightly diverges from doxygen in that an empty line before and after in a multi line comment is *not* required.
```
tfoleyNV, line 28: Can we have a "comparison with doxygen" section and move all the detailed differences there? The way this is phrased here kind of ends up assuming that the reader is already familiar with doxygen, and anticipates that they will already be thinking of detailed questions like that.

A reader who doesn't already know doxygen's conventions might just be confused by this bit, since all of the above being equivalent seems like the obvious behavior anyway.

docs/doc-system.md
```diff
@@ -0,0 +1,114 @@
...
+    int b);
+```
+
+If you just need a single line comment to describe something, you can place the documentation after the parameter as in
+
+```
+
+/// My function
+void myFunction(    int a,      //< The A parameter     
+                    int b)      //< The B parameter
```
tfoleyNV, line 47: It isn't related to the documentation, but I'd like for us to eventually allow this kind of syntax with ordinary `///` comments and not needing `//<`. It is "obvious" to a reader of the code that a trailing comment like this pertains to the preceding declaration and not the next one.

If we just using the leading/trailing trivia definition, then we can say that a comment like these is trailing trivia on the declaration (or the `,` / `)` after it), and thus belongs to the declaration.

(Yes, I have an agenda here to get rid of all the `<`, `!`, etc. nonsense that doxygen pushes onto people. I want our solution to first and foremost be something we can describe as "humane")

docs/doc-system.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+
+To actually get Slang to output documentation you can use the `-doc` option from the `slangc` command line, or pass it in as parameter to `spProcessCommandLineArguments` or `processCommandLineArguments`. The documentation is currently output by default to the same `ISlangWriter` stream as diagnostics. So for `slangc` this will generally mean the terminal/stderr. 
+
+Currently the Slang doc system does not support any of the 'advanced' doxygen documentation features. If you add documentation to a declaration it is expected to be in [markdown](https://guides.github.com/features/mastering-markdown/). 
+
+Currently the only documentation style supported is a single file 'markdown' output. Future versions will support splitting into multiple files and linking between them. Also future versions may also support other documentation formats/standards.
+
+It is possible to generate documentation for Slangs internal `stdlib`. This can be achieved with `slangc` via
```
tfoleyNV, line 88: Do we need to document this in the public docs? Ideally only devs on the project would care about how to document the stdlib.

docs/doc-system.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+slangc -doc -compile-stdlib
+```
+
+The documentation will be written to a file `stdlib-doc.md`.
+
+It should be noted that it is not necessary to add markup to a declaration for the documentation system to output documentation for it. Without the markup the documentation is going to be very limited, in essence saying the declaration exists and other aspects that are available from the source. This may not be very helpful. For this reason and other reasons there is a mechanism to control the visibility of items in your source. 
+
+There are 3 visibility levels 'public', 'internal' and 'hidden'/'private'. There is a special comment that controls visibility for subsequent lines. The special comment starts with `//@` as shown below.
```
tfoleyNV, line 98: It seems like we are at a point where we should go ahead and add `public` and `internal` (at least) to the language itself and use those instead of comment-based conventions.

Also, we probably need to discuss what `hidden` would mean in terms of actual access control. I still strongly prefer the option where the `__` prefix is how we do "hidden"-ness, since that makes it clear that it is purely a convention thing (no actual compiler enforcement), and avoids needing to have these `@`-commands.

tfoleyNV (approved):

docs/doc-system.md
jsmall-nvidia: That without some extra heuristics/limitations this introduces an ambiguity.  Take

```
enum Blah
{
   A, /// Some documentation
   B, 
};
```
Is it B or As? You could say ah - because it's after A *on the same line* it belongs to A. Ok now

```
enum Blah
{ 
   A,  /// A 
        /// multi-line comment 
   
   B,
};
```

That //< allows (fairly simply) multi line is the point here. 
I don't bring this up to make out this is some insurmountable challenge, just that given the features of the doxygen style/s it isn't without reasonable reason that they have the //< and associated styles. 

I admit I perhaps also do find the //< style particularly objectionable having used it in many code bases for many years.

docs/doc-system.md
jsmall-nvidia: I kind of see this as documentation for ourselves too, and doesn't seem problematic that users at large can know this also. That said it probably isn't very important to users, so perhaps should be relegated to a more lowly location within the docs. Or perhaps it needs to go in another folder/and or file that isn't so user facing.

docs/doc-system.md
jsmall-nvidia: Being able to have a language features around accessibility seems like a good thing. I would perhaps argue there is a distinction between a language level access/usability and documentation visibility. I wouldn't fight that tooth and nail - as it could be argued there is perhaps largely suitable overlap.

When you have spoken about this previously to my recollection you have advocated for such keywords to be used on each declaration. I think it is useful to at least have the choice 

```
private int something(); 
public:
int aPublicThing(); 
int andAnother(); 
``` 

That having to write public/private (or whatever) on every single declaration smells like unwarranted drudgery to me. I also appreciate that having both opens questions on the interplay - does the non colon style *just* effect the current decl, or is it really the same as : style and so a slight short hand? I think probably the right answer is that the : style sets the default, and the non : style just sets for that decl, and the 'default' for that scope remains.

When this mechanism was in stdlib used it was useful to just wipe out whole sections of the output with just a couple of lines. 

That said using `public:` style outside of say a class where it's scope of effect is limited doesn't seem quite right. And seems really something more doc related. Which again makes me wonder if combining the uses into a single feature is really that desirable.

tfoleyNV: On the first issue (whether documentation public/internal and symbol public/internal should be coupled or not): it seems like the somewhat obvious default that:

* Anything that is not visible from outside a module probably shouldn't be included in the public documentation for that module. Hence anything non-`public` probably shouldn't be in the public documentation, right?
* Anything that is publicly visible outside the module should probably be documented *unless* it is specifically annotated as not being suitable for documentation. The `__` prefix convention seems like a great way to mark this since it is already used to indicate "you *can* use this, but you really *shouldn't*."

(There's a weird third case of symbols that aren't part of the public API, but *are* part of the public ABI, because they are allowed to be used by public `inline` functions. I don't think those need to affect documentation, but they need to be part of a robust visibility system)

tfoleyNV: On the second issue, I understand the concern over having to write `public` over and over again, but my take is that this is one of those subtle places where the C++ style is privileging the experience of the person who has to *write* the code over all the future people who have to *read*/*maintain* the code.

Making the meaning of a declaration (even including meta-level meaning like how it gets documented) depend on where it appears in the sequence of a file is dangerous because simple cut-paste editing operations could greatly change the meaning (or meta-meaning).

I don't know the full rationale for the Java switch to `public` on each declaration, but in the case of C# there's actually a lot of discussion of rationale for some of their syntax choices and there was a big focus (IIRC) on the kinds of syntax choices that make copy-paste programming more likely to introduce silent semantic changes. Given that we can't currently do our own research into the habits of programmers and the classes of bugs they introduce, I'm inclined to try to leverage the research done by other language designers whenever possible. :)

tfoleyNV: IIRC the D language tries to split the difference by both supporting per-declaration modifiers and then also allowing a kind of scoped modifier construct like:

```
public
{
    /* declarations go here */
}
```

The use of an explicit nesting/grouping syntax makes that option at least a bit less prone to copy-paste errors, although it is the kind of solution that strikes me as just not being willing to commit to one option.

(All of this said, I fully expect our users to ask for the C++ style of visibility markers until we relent and provide them)

jsmall-nvidia: I didn't touch on the `__` aspect. In the context of documenting our stdlib - that seems fine. 

As a mechanism for users to control documentation in their source code, it seems ... not fine.

PR #1761, opened 2021-03-18
Remove `DescriptorSet` from D3D11 and GL devices.

csyonghe:
Refactored the shader-object implementation so that d3d11 and gl no longer uses `GraphicsCommonShaderObject` classes to implement shader-object API, so d3d11 and gl implementations are now free of descriptor-set related infrastructure.

Note that currently the implementation still requires `RootShaderObject` to be persistent and will crash if it is used as a transient object that gets released immediately after command buffer recording. Properly supporting transient root objects require systematic fix to the public interface and the implementations that are beyond the scope of this PR.

### DIFF
--- a/tools/gfx/d3d11/render-d3d11.cpp
+++ b/tools/gfx/d3d11/render-d3d11.cpp
@@ -99,13 +99,48 @@ public:
         IInputLayout** outLayout) override;
 
     virtual SLANG_NO_THROW Result SLANG_MCALL createDescriptorSetLayout(
-        const IDescriptorSetLayout::Desc& desc, IDescriptorSetLayout** outLayout) override;
+        const IDescriptorSetLayout::Desc& desc, IDescriptorSetLayout** outLayout) override
+    {
+        SLANG_UNUSED(desc);
+        SLANG_UNUSED(outLayout);
+        return SLANG_FAIL;
+    }
     virtual SLANG_NO_THROW Result SLANG_MCALL
-        createPipelineLayout(const IPipelineLayout::Desc& desc, IPipelineLayout** outLayout) override;
+        createPipelineLayout(const IPipelineLayout::Desc& desc, IPipelineLayout** outLayout) override
+    {
+        SLANG_UNUSED(desc);
+        SLANG_UNUSED(outLayout);
+        return SLANG_FAIL;
+    }
     virtual SLANG_NO_THROW Result SLANG_MCALL createDescriptorSet(
         IDescriptorSetLayout* layout,
         IDescriptorSet::Flag::Enum flag,
-        IDescriptorSet** outDescriptorSet) override;
+        IDescriptorSet** outDescriptorSet) override
+    {
+        SLANG_UNUSED(layout);
+        SLANG_UNUSED(flag);
+        SLANG_UNUSED(outDescriptorSet);
+        return SLANG_FAIL;
+    }
+    virtual SLANG_NO_THROW void SLANG_MCALL setDescriptorSet(
+        PipelineType pipelineType,
+        IPipelineLayout* layout,
+        UInt index,
+        IDescriptorSet* descriptorSet) override
+    {
+        SLANG_UNUSED(pipelineType);
+        SLANG_UNUSED(layout);
+        SLANG_UNUSED(index);
+        SLANG_UNUSED(descriptorSet);
+    }
+
+    virtual Result createShaderObjectLayout(
+        slang::TypeLayoutReflection* typeLayout,
+        ShaderObjectLayoutBase** outLayout) override;
+    virtual Result createShaderObject(ShaderObjectLayoutBase* layout, IShaderObject** outObject) override;
+    virtual SLANG_NO_THROW Result SLANG_MCALL
+        createRootShaderObject(IShaderProgram* program, IShaderObject** outObject) override;
+    virtual void bindRootShaderObject(PipelineType pipelineType, IShaderObject* shaderObject) override;
 
     virtual SLANG_NO_THROW Result SLANG_MCALL
         createProgram(const IShaderProgram::Desc& desc, IShaderProgram** outProgram) override;
@@ -127,12 +162,6 @@ public:
 
     virtual SLANG_NO_THROW void SLANG_MCALL
         setPrimitiveTopology(PrimitiveTopology topology) override;
-
-    virtual SLANG_NO_THROW void SLANG_MCALL setDescriptorSet(
-        PipelineType pipelineType,
-        IPipelineLayout* layout,
-        UInt index,
-        IDescriptorSet* descriptorSet) override;
 
     virtual SLANG_NO_THROW void SLANG_MCALL setVertexBuffers(
         UInt startSlot,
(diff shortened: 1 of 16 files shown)


tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -542,22 +423,1101 @@ class D3D11Device : public ImmediateRendererBase
...
+            //
+            SLANG_RETURN_ON_FAIL(_writeOrdinaryData(device, m_ordinaryDataBuffer, 0, specializedOrdinaryDataSize, specializedLayout));
+
+            return SLANG_OK;
+        }
+
+        /// Bind the buffer for ordinary/uniform data, if needed
+        Result _bindOrdinaryDataBufferIfNeeded(
+            D3D11Device* device,
+            RootBindingState* bindingState)
```
tfoleyNV, line 1282: I'm a little worried/suspcicious when I see `RootBindingState` here, because it seems to imply that we are copying/flattening the state from the shader objects into a secondary data structure that then goes on to be written to the context state. I would have expected the `bind*()` operations here to take a `D3D11DeviceContext` and directly write state into the context, but maybe there's details I'm missing right now.

tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -542,22 +423,1101 @@ class D3D11Device : public ImmediateRendererBase
...
+
+        List<RefPtr<UnorderedAccessViewImpl>> m_uavs;
+
+        List<RefPtr<ShaderObjectImpl>> m_objects;
+
+        /// A constant buffer used to stored ordinary data for this object
+        /// and existential-type sub-objects.
+        ///
+        /// Created on demand with `_createOrdinaryDataBufferIfNeeded()`
+        RefPtr<BufferResourceImpl> m_ordinaryDataBuffer;
```
tfoleyNV, line 1336: Again, this is awesome to see and exactly what we'd want/expect in this case.

Maybe the only detailed difference we could look for is to have a flat array next to `m_srvs` that holds just a bunch of `ID3D11ShaderResourceView*` so that it can be passed directly to API calls with no extra translation.

tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -1994,10 +2969,74 @@ D3D11_COLOR_WRITE_ENABLE translateRenderTargetWriteMask(RenderTargetWriteMaskT m
...
+        m_immediateContext->CSSetConstantBuffers(0, (UINT)m_rootBindingState.constantBuffers.getCount(), m_rootBindingState.constantBuffers.getBuffer());
+        break;
+    default:
+        m_immediateContext->VSSetShaderResources(0, (UINT)m_rootBindingState.srvBindings.getCount(), m_rootBindingState.srvBindings.getBuffer());
+        m_immediateContext->PSSetShaderResources(0, (UINT)m_rootBindingState.srvBindings.getCount(), m_rootBindingState.srvBindings.getBuffer());
+        m_immediateContext->VSSetSamplers(0, (UINT)m_rootBindingState.samplerBindings.getCount(), m_rootBindingState.samplerBindings.getBuffer());
+        m_immediateContext->PSSetSamplers(0, (UINT)m_rootBindingState.samplerBindings.getCount(), m_rootBindingState.samplerBindings.getBuffer());
+        m_immediateContext->VSSetConstantBuffers(0, (UINT)m_rootBindingState.constantBuffers.getCount(), m_rootBindingState.constantBuffers.getBuffer());
+        m_immediateContext->PSSetConstantBuffers(0, (UINT)m_rootBindingState.constantBuffers.getCount(), m_rootBindingState.constantBuffers.getBuffer());
```
tfoleyNV, line 3031: This is an obvious statement, but we'd eventually need to make this calls for whichever pipeline stages the selected program uses.

tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -1994,10 +2969,74 @@ D3D11_COLOR_WRITE_ENABLE translateRenderTargetWriteMask(RenderTargetWriteMaskT m
...
+    m_rootBindingState.constantBuffers.clear();
+    static_cast<ShaderObjectImpl*>(shaderObject)->bindObject(this, &m_rootBindingState);
+    switch (pipelineType)
+    {
+    case PipelineType::Compute:
+        m_immediateContext->CSSetShaderResources(0, (UINT)m_rootBindingState.srvBindings.getCount(), m_rootBindingState.srvBindings.getBuffer());
+        m_immediateContext->CSSetUnorderedAccessViews(0, (UINT)m_rootBindingState.uavBindings.getCount(), m_rootBindingState.uavBindings.getBuffer(), nullptr);
+        m_immediateContext->CSSetSamplers(0, (UINT)m_rootBindingState.samplerBindings.getCount(), m_rootBindingState.samplerBindings.getBuffer());
+        m_immediateContext->CSSetConstantBuffers(0, (UINT)m_rootBindingState.constantBuffers.getCount(), m_rootBindingState.constantBuffers.getBuffer());
+        break;
```
tfoleyNV, line 3024: This big thing I'm missing here (and it applies to the GL case as well) is why this operation needs to be broken into two steps like this.

With this logic, we recursively walk the shader-object hierarchy in the "right" order and do a sequence of `.add()` operations to the relevant arrays, which eventually pipe into a single monolithic `CSSet*` operation per category of binding.

It seems equivalent to walk the same hierarchy and instead of calling `.add()` at each step, directly call `CSSet*()` as appropriate. This would eliminate creating the copy entirely.

It is possible that the choice here is an intentional one, in that it minimizes the total number of D3D11 API calls we need to make by setting everything en masse. If that's what this is about then the solution being used seems reasonable.

One down-side to this approach is that it doesn't seem to translate as cleanly to having a transient root shader object or anything, but I guess that doesn't apply for D3D11 anyway.

(Aside: should we consider having some kind of special support in `gfx` for single-threaded applications that just want an "immediate context" since that could be implemented natively on D3D11/GL/CPU/CUDA and emulated for VK/D3D12? A lot of applications only do single-threaded rendering a lot of the time.)

tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -542,22 +423,1101 @@ class D3D11Device : public ImmediateRendererBase
...
+            SLANG_RETURN_ON_FAIL(_bindOrdinaryDataBufferIfNeeded(device, bindingState));
+
+            for (auto sampler : m_samplers)
+                bindingState->samplerBindings.add(sampler ? sampler->m_sampler.get() : nullptr);
+            
+            for (auto srv : m_srvs)
+                bindingState->srvBindings.add(srv ? srv->m_srv : nullptr);
+
+            for (auto uav : m_uavs)
+                bindingState->uavBindings.add(uav ? uav->m_uav : nullptr);
```
tfoleyNV, line 1313: In the hypothetical model where this operation directly talks to the `ID3D11DeviceContext`, each of these loops could be a single `CSSet*()` call in the compute case. We'd just need to store parallel arrays that are already in D3D11-API-friendly format, list a `List<ID3D11ShaderResourceView*>`.

I think having a really clear operation on a shader object that shows how it "splats" its state onto the context in as few operations as possible will go a long way to showing that they are actually a highly efficient way to drive the API.

tfoleyNV (approved):
Looks great. I think we can iterate further from here, but this is exactly the kind of thing I was hoping we'd see for D3D11 and GL.

tools/gfx/d3d11/render-d3d11.cpp
csyonghe: You are right, I didn't realize you can call *SSetXXX multiple times, since all UAVs needs to be set at a single time, but that is just an anomaly.

tfoleyNV: Oh yeah, I forgot that UAVs were a special case in D3D11 (and an even more special case when it comes to RTVS + UAVs for PS).

One thing that I guess I forgot to note in the original review is that it is kind of subtle but this logic in the D3D11 handling of shader objects seems to be assuming that applications only use the completely default layout rules (no explicit `register` stuff). We probaby need/want to address that sooner or later if we want to have tests that also make use of explicit bindings (even if they are basically always a Bad Idea).

PR #1763, opened 2021-03-22
`gfx` D3D12 shader objects rewrite.

csyonghe:
This change reimplements D3D12 shader objects in the `gfx` layer,
so that it is no longer layered over the `DescriptorSet`
abstraction. The new implementation also takes advantage of
D3D12's flexibility to map a `space` and `register` binding to an
arbitrary descriptor address to facilitate a more efficient
binding algorithm.

In this implementation, each shader object
(representing an entry-point, a constant buffer or a parameter
block) maintains one descriptor table set (one table for each
descriptor heap type) that is used to hold all bindings for the
object, excluding bindings of sub-objects. For example, if the
global scope defines the following parameters:
```
struct CB
{
    Texture2D texture;
    SamplerState s;
};
ConstantBuffer<CB> cb;
RWStructuredBuffer<int> bufferOut;
```
Then we will use two shader objects to hold all parameters:
```
ShaderObject_0 (root):
    ResourceDescriptorTable(slotCount = 1)
        DescriptorRangeType: UAV
        DescriptorCount: 1
ShaderObject_1 (cb):
    ResourceDescriptorTable(slotCount = 1)
        DescriptorRangeType: SRV
        DescriptorCount: 1
    SamplerDescriptorTable(slotCount = 1)
        DescriptorRangeType: Sampler
        DescriptorCount: 1
```
We will create the following root signature:
```
RootParam[0]:
    DescriptorTable
        Range[0]: UAV, Count = 1
        Range[1]: CBV, Count = 1
RootParam[1]:
    DescriptorTable
        Range[0]: Sampler, Count = 1
```

Note that in this root signature, the descriptor for the
constant buffer `cb` appears *after* the UAV for `bufferOut`,
despite that `cb` is defined before `bufferOut` in the shader
code. This is intentional: by laying out the descriptors in
this order, we can simply copy the descriptor tables from all
shader objects to form the final GPU descriptor table.
In this example, when we are generating the GPU descriptor
from the hierarchy of shader objects, we simply append the
descriptor tables for the global scope and for `cb`.

By doing so, the user's `space` and `register` bindings will
only has an effect on the root signature's descriptor range
descriptions, and will not affect the descriptor table layout.
This allows the `gfx` layer to always operate the in most
efficient way.

### DIFF
--- a/tools/gfx/d3d12/render-d3d12.cpp
+++ b/tools/gfx/d3d12/render-d3d12.cpp
@@ -5,7 +5,6 @@
 
 //WORKING:#include "options.h"
 #include "../renderer-shared.h"
-#include "../render-graphics-common.h"
 #include "../simple-render-pass-layout.h"
 #include "../d3d/d3d-swapchain.h"
 #include "core/slang-blob.h"
@@ -66,7 +65,7 @@ using namespace Slang;
 
 static D3D12_RESOURCE_STATES _calcResourceState(IResource::Usage usage);
 
-class D3D12Device : public GraphicsAPIRenderer
+class D3D12Device : public RendererBase
 {
 public:
     // Renderer    implementation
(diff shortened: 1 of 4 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -752,3283 +580,4160 @@ class D3D12Device : public GraphicsAPIRenderer
...
+                    cpuHeap.getCpuHandle(cpuDescriptorTable),
                     D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER);
-
-                submitter->setRootDescriptorTable(
-                    int(rootParameterIndex++), gpuHeap.getGpuHandle(gpuDescriptorTable));
+                bindingState->offset.sampler += descSet.m_samplerCount;
+                descTableIndex++;
+            }
+            bindingState->futureRootParamOffset =
+                Math::Max(descTableIndex, bindingState->futureRootParamOffset);
```
tfoleyNV, line 2085: I think I see what is going on here, and in my own versions of this code (that is, what I've written for Falcor in the past) I think I handled the same issue by just having two different recursive walks. Well, three actually:

* Case (1) is the recursive walk is to write the uniform/ordinary data of a shader object. This one copies the directly-contained data and then recurses on any interface-type sub-objects.

* Case (2) is the recursive walk is to write descriptors *into* descriptor a descriptor set (or in the case of D3D12, a pair of descriptor sets). Case (2a) writes the default CBV binding (if needed), filling it in using (1), and then delegates to case (2b) which writes all the directly-contained SRV/UAV/sampler descriptors. This path recurses on interface-type sub-objects via (2b) and on `ConstantBuffer<X>`-type sub-objects via (2a).

* Case (3) is a recursive walk to write whole descriptor sets. It creates the descriptor set(s) for the current object and then relies on (2a) to fill it in. Then it recurses on any `ParameterBlock<X>`-type sub-objects.

With that decomposition, I think case (3) can just keep a running counter of how many root parameters have been filled in so far and it will do just fine.

tfoleyNV (approved):
Unfortunately this diff is pretty hard to read so I can't be sure I saw everything relevant, but I'm going to approve this with the assumption we'll be continuing to iterate on the design no matter what, so the details maybe don't matter as much.

PR #1764, opened 2021-03-24
Reimplement Vulkan shader objects.

csyonghe:
This change reimplements Vulkan shader objects in the `gfx` layer so that it is no longer layered on top of the `DescriptorSet` abstraction. Since this is the last implementation that uses `DescriptorSet`, the change also removes all `DescriptorSet` related API from public `gfx` interface.

The Vulkan implementation now passes all test cases, but it still have two issues:
1. The PushConstant setting is not correct, this is because we don't seem to be able to get correct reflection data about the size of push constants for an entry-point.
2. The `shader-toy` example can't run on Vulkan, because it currently sets nullptr to `Texture` bindings, and this change doesn't properly handle setting resource to null in `ShaderObject`s yet. If we can use the `nullDescriptor` feature on vulkan, this implementation will be simple. However we still want to decide whether we want to use a Vulkan 1.2 feature for this.


examples/hello-world/main.cpp
```diff
@@ -244,6 +244,7 @@ Slang::Result initialize()
     // platforms/APIs.
     //
     IDevice::Desc deviceDesc = {};
+    deviceDesc.deviceType = DeviceType::Vulkan;
```
tfoleyNV, line 247: We should probably remove these lines for check-in or shortly after.

slang-gfx.h
```diff
@@ -113,35 +113,11 @@ enum class BindingStyle
...
-    };
-
     struct Desc
     {
         PipelineType        pipelineType;
-
-        KernelDesc const*   kernels;
-        Int                 kernelCount;
-
             /// Use instead of `kernels`/`kernelCount` if loading a Slang program.
```
tfoleyNV, line 119: Comment will probably need updating.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -38,14 +38,14 @@
...
+class VKDevice : public RendererBase
 {
 public:
     enum
     {
         kMaxRenderTargets = 8,
         kMaxAttachments = kMaxRenderTargets + 1,
-
+        kMaxPushConstantSize = 256,
         kMaxDescriptorSets = 8,
```
tfoleyNV, line 49: We eventually need to query these via VK API calls and be ready for them to be different across implementations.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -879,322 +676,2084 @@ class VKDevice : public GraphicsAPIRenderer
...
+            write.pImageInfo = (VkDescriptorImageInfo*)infos;
+            for (Index i = 0; i < samplers.getCount(); i++)
+            {
+                auto& imageInfo = ((VkDescriptorImageInfo*)infos)[i];
+                imageInfo.imageView = 0;
+                imageInfo.imageLayout = VK_IMAGE_LAYOUT_GENERAL;
+                imageInfo.sampler = samplers[i]->m_sampler;
+            }
+            bindingState->descriptorSetWrites.add(write);
+        }
```
tfoleyNV, line 2181: Hypothetically, the shader object *layout* could pre-compute all these `VkWriteDescriptorSet` and related objects into a single compact buffer, and then a particular object would just need to use a scratch-allocated copy that it then patches a few pointers into. (One challenge here is that the VK API loves its little structure types, but then misses a lot of the potential benefit of re-using those structures once written by mixing up pointer and non-pointer data all over the place)

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -879,322 +676,2084 @@ class VKDevice : public GraphicsAPIRenderer
...
+            // Copy entry point uniform data into global push constant buffer.
+            if (m_ordinaryData.getCount())
+            {
+                Index dstOffset = bindingState->pushConstantBuffer.getCount();
+                ::memcpy(
+                    bindingState->pushConstantBuffer.getBuffer() + dstOffset,
+                    m_ordinaryData.getBuffer(),
+                    m_ordinaryData.getCount());
+                bindingState->pushConstantBuffer.setCount(dstOffset + m_ordinaryData.getCount());
+            }
```
tfoleyNV, line 2496: We've discussed it in some of these other PRs, so I won't belabor it here, but it seems weird to me to copy all the uniform/ordinary data from all the entry points into a distinct buffer so that we can make a single VK call to write the push constants instead of just making the calls right here. We have the data, and we know the offset where we want to write it, so why not just make the right VK API call here?

tfoleyNV (approved):

PR #1765, opened 2021-03-24
Improve Vulkan shader-objects implementation.

csyonghe:
1. Null bindings no longer crashes.
2. No longer copies push constants to staging CPU buffer before setting it into command buffer. The entry-point shader object now directly sets it into command buffer upon `bindObject` call.
3. Improved vulkan implementation so that each shader object is responsible for creating descriptor sets on-demand.
4. Fixed slang reflection to correctly report `ParameterBlock` binding.
5. Re-enabled 3 tests that were previously disabled.

### DIFF
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -700,10 +700,11 @@ public:
             Index bindingRangeIndex;
         };
 
-        struct DescriptorSetInfo : public RefObject
+        struct DescriptorSetInfo
         {
             List<VkDescriptorSetLayoutBinding> vkBindings;
             Slang::Int space = -1;
+            VkDescriptorSetLayout descriptorSetLayout = VK_NULL_HANDLE;
         };
 
         struct Builder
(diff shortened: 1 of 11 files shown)


tools/gfx/vulkan/render-vk.cpp
```diff
@@ -826,7 +829,16 @@ class VKDevice : public RendererBase
                         }
                         descriptorSetInfo.vkBindings.add(vkBindingRangeDesc);
                     }
+                    VkDescriptorSetLayoutCreateInfo createInfo = {};
+                    createInfo.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
+                    createInfo.pBindings = descriptorSetInfo.vkBindings.getBuffer();
+                    createInfo.bindingCount = (uint32_t)descriptorSetInfo.vkBindings.getCount();
+                    VkDescriptorSetLayout vkDescSetLayout;
+                    SLANG_RETURN_ON_FAIL(m_renderer->m_api.vkCreateDescriptorSetLayout(
```
tfoleyNV, line 837: I had to go and convince myself that applying `SLANG_RETURN_ON_FAIL` to a Vulkan API call actually works, but the documentation for `VkResult` is consistent with `HRESULT` and our Slang/`gfx` `Result` in that non-negative values represent success. That's kind of convenient/cool.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -3788,17 +3800,6 @@ Result VKDevice::PipelineCommandEncoder::bindRootShaderObjectImpl(
...
-    for (auto pushConstantRange : specializedLayout->m_pushConstantRanges)
-    {
-        m_device->m_api.vkCmdPushConstants(
-            m_commandBuffer->m_commandBuffer,
-            specializedLayout->m_pipelineLayout,
-            pushConstantRange.stageFlags,
-            pushConstantRange.offset,
-            pushConstantRange.size,
-            bindState.pushConstantBuffer.getBuffer() + pushConstantRange.offset);
-    }
```
tfoleyNV, line 3801: It is good to have this move out of the `bindState`, and the next step would seemingly (eventually) be to move the `vkUpdateDescriptorSet` and `vkCmdBindDescriptorSets` calls out as well.

Realistically, further cleanup can/should probably wait until we start to implement the more advanced design for root shader objects where the application cannot arbitrarily persist them, and then behave closer to the command encoders in practice (created for short-term use and then dropped).

tfoleyNV (approved):

Issue #1762, opened 2021-03-21
Add support for __LINE__ and __FILE__


savage309:
Using __LINE__ give the following error:

`error 30015: undefined identifier '__LINE__'`

Expected behavior is like in C/C++ compilers

PR #1772, opened 2021-03-30
Support for __LINE__ and __FILE__ in preprocessor

jsmall-nvidia:
* Add support for `__LINE__` and `__FILE__` in preprocessor
* Added -E option to output preprocessor stream
* Made writers available to FrontEndRequest so it can output to stdout/err etc
* Added tests for variety of of situations 
* Fixed issues around empty stderr for diagnostics tests

Note that in doing this testing a problem appeared with pasting, so added a test that demonstrates the incorrect pasting behavior.

The behavior here is most akin to gcc/clang

* `__LINE__` and `__FILE__` can be `#define`d (produces a warning) 
* undefining after defining does *not* revert to previous behavior 
 * Ie doing `#define __LINE__ #undef __LINE__ __LINE__` produces `__LINE__` in output

This addresses issue  #1762


source/slang/slang-compiler.h
```diff
@@ -1564,8 +1568,11 @@ namespace Slang
     class FrontEndCompileRequest : public CompileRequestBase
     {
     public:
+            /// Note that writers can be parsed as nullptr to disable output,
```
tfoleyNV, line 1571: typo: Parsed -> passed

source/slang/slang-compiler.h
```diff
@@ -1564,8 +1568,11 @@ namespace Slang
...
+            /// and individual channels set to null to disable them
         FrontEndCompileRequest(
             Linkage*        linkage,
+            StdWriters*     writers,
```
tfoleyNV, line 1575: I'm not a fan of having `StdWriters` reach its way deeper into the Slang library, because it creates a potential for confusion about "where should I send this output?".

Ideally we should only have two cases of output:

1. Diagnostic output. This is always written to a diagnostic sink, which emits to `stderr` for `slangc` and can be captured into buffers for API-based use.
2. File output ("compile results"). These are always captures into buffers for API-based used, and for `slangc` are always either written to filesystem paths or `stdout`.

The redirection feature really only exists to support the test-runner using in-process execution while still running Slang in "`slangc` mode." I feel like we need to push back and start solving more of this at the test-runner level when we next get time to work on the test system.

source/slang/slang-preprocessor.cpp
```diff
@@ -2539,6 +2638,25 @@ TokenList preprocessSource(
...
+
+        for (Index i = 0; i < SLANG_COUNT_OF(builtinNames); i++)
+        {
+            auto name = namePool->getName(builtinNames[i]);
+
+            PreprocessorMacro* macro = CreateMacro(&preprocessor);
+            macro->flavor = builtinFlavors[i];
+            macro->nameAndLoc = NameLoc(name);
+
+            preprocessor.globalEnv.macros[name] = macro;
```
tfoleyNV, line 2656: Note: We could in theory have a "super-global" macro environment that sits above all user-defined macros. That might allow us to avoid having `#undef` remove these built-in macros.

source/slang/slang-compiler.h
tfoleyNV: I don't think it is worth trying to change anything for this PR, since none of this is really related to the `__LINE__`/`__FILE__` feature.

tests/preprocessor/special-macro-simple.slang.expected
```diff
@@ -0,0 +1,6 @@
+result code = 0
+standard error = {
+}
+standard output = {
+5 "tests/preprocessor/special-macro-simple.slang"
```
tfoleyNV, line 5: It is a minor point (and not something we need to rush to "fix" any time soon), but the `-E` mode for most C/C++ compilers produces preprocessed output with `#line` directives to ensure that file/line information is correct when the output is passed into a compiler again.

(I know it is weird the running a C preprocessor on a file results in a file that still has preprocessor directives, but `#line` is just special that way.)

tfoleyNV (approved):

PR #1774, opened 2021-03-30
`gfx` explicit transient resource management.

csyonghe:
This change adds an `ITransientResourceHeap` interface to the `gfx` layer to expose transient resource mangement directly to the user.

The user is reponsible for creating transient resource heaps for command buffer, shader-object constant buffer and descriptor table allocation. The user must also periodically call `ITransientResourceHeap::synchronizeAndReset()` method to reset the heap and reuse the space for new allocations. A typical use of `ITransientResourceHeap` is to create one heap per swapchain image, and reset the corresponding heap at each frame, then allocate command buffers from the heap for the draw commands of that frame.

This change also refactors the examples so that they share common code via the `example-base` project.

Note that the D3D11 and GL implementation currently does not reuse constant buffers for different transient heaps, but they should be doing so since these APIs already provide versioning support for resources.


examples/gpu-printing/main.cpp
```diff
@@ -113,9 +113,13 @@ Result execute()
...
     auto printBufferView = gDevice->createBufferView(printBuffer, printBufferViewDesc);
 
+    ITransientResourceHeap::Desc transientResourceHeapDesc = {};
+    transientResourceHeapDesc.constantBufferSize = 256;
+    auto transientHeap = gDevice->createTransientResourceHeap(transientResourceHeapDesc);
+
     ICommandQueue::Desc queueDesc = {ICommandQueue::QueueType::Graphics};
     auto queue = gDevice->createCommandQueue(queueDesc);
-    auto commandBuffer = queue->createCommandBuffer();
+    auto commandBuffer = transientHeap->createCommandBuffer();
```
tfoleyNV, line 122: Minor: when we get around to supporting Metal, we will need to deal with command buffers being allocated from a specific queue, so this either needs to be `queue->createCommandBuffer(heap)` or `heap->createCommandBuffer(queue)` to make sure that both objects are available.

I lean toward `queue->createCommandBufer(heap)` because in many cases we can/should support allocations with a null heap, which would be used to indicate that the user want a persistent allocation.

examples/gpu-printing/main.cpp
```diff
@@ -113,9 +113,13 @@ Result execute()
     printBufferViewDesc.type = IResourceView::Type::UnorderedAccess;
     auto printBufferView = gDevice->createBufferView(printBuffer, printBufferViewDesc);
 
+    ITransientResourceHeap::Desc transientResourceHeapDesc = {};
+    transientResourceHeapDesc.constantBufferSize = 256;
+    auto transientHeap = gDevice->createTransientResourceHeap(transientResourceHeapDesc);
```
tfoleyNV, line 118: I'm a bit surprised that there is nothing here to indicate the "depth" of the queue of scopes/frames inside the heap.

examples/shader-toy/main.cpp
```diff
@@ -407,13 +340,9 @@ float clickMouseY = 0.0f;
...
 
-void renderFrame()
+virtual void renderFrame(int frameIndex) override
 {
-    auto frameIndex = gSwapchain->acquireNextImage();
-    if (frameIndex == -1)
-        return;
-
-    auto commandBuffer = gQueue->createCommandBuffer();
+    auto commandBuffer = gTransientHeaps[frameIndex]->createCommandBuffer();
```
tfoleyNV, line 345: Okay, now I'm realizing that maybe I'm not understanding things. This line implies that we need N transient heaps to match an N-deep swap chain, and that the application will manually ring-buffer through the transient heaps?

My assumption had been that we'd want to have a *single* object that could be created up front with a conceptual "queue depth" and then the application could poke it to swap which if its internal frames/scopes it is on. Maybe the difference is minor, but it would be great if we can free applications from needing to always allocate these N copies of things for an N-deep swap chain.

slang-gfx.h
```diff
@@ -1268,6 +1255,34 @@ class ICommandQueue : public ISlangUnknown
         0x14e2bed0, 0xad0, 0x4dc8, { 0xb3, 0x41, 0x6, 0x3f, 0xe7, 0x2d, 0xbf, 0xe } \
     }
 
+class ITransientResourceHeap : public ISlangUnknown
+{
+public:
+    struct Desc
+    {
+        size_t constantBufferSize;
```
tfoleyNV, line 1263: Comment? It isn't clear what precisely this represents. I'd have to guess it is a hint on the amount of constant buffer memory to pre-allocate for use on ring-buffered CB allocations.

Ideally I'd like it if all the values in the `Desc` here were hints and not requirements, so that performance-conscious apps could set a value they know will be sufficient to avoid runtime allocations, while simple applications can just leave things zeroed at the cost of suffering some on-the-fly allocations during the first few frames that use a given heap.

tfoleyNV (approved):

PR #1775, opened 2021-03-31
Added compiler-core project

jsmall-nvidia:
* Moved Lexer/Diagnostics/SourceLocation into compiler-core
* Moved DownstreamCompiler support into compiler-core
* Moved slang-file-system to core

Contains a hack around CharUtil to make linking work on linux.


source/compiler-core/slang-token.cpp
```diff
@@ -1,13 +1,10 @@
 // slang-token.cpp
 #include "slang-token.h"
 
-#include <assert.h>
+//#include <assert.h>
```
tfoleyNV, line 4: Delete if not needed?

tfoleyNV (approved):

Issue #1645, opened 2020-12-15
Slang API ABI Ideas


jsmall-nvidia:
This mainly discusses the issues around value types and ABI compatibility. ABI compatibility could potentially be backward and/or forward compatible. 

Backward compatible: System accepts input for an earlier version of itself.
Forward compatible: System accepts input for a later version of itself.

Both would be ideal. Of the two backward compatibility is probably the more important and easier to achieve. 

The main discussion here is around the use of value types - typically structs within interfaces. There are many mechanisms for ABI compatibility around the interfaces themselves - such as COM, and others. 

We would desire a system that:

* Had type safety
* We need to be able to extend types by adding fields
* We need to be able to change types by removing fields or other tranformations
* Convenient to use
* Hard to use incorrectly
* Prefer to keep a single parameter is possible
* Minimize the amount of 'cruft' around usage
* Ideally has a mechanism for both forward and backward compatibility, but at a minimum have backward compatibility
* Code internal to the API generally does *not* have to traffic in different versions. 
** ie that the API call implemention is the point of that trafficing takes place, internal the 'latest version' is used
* It is desirable to have the possibility of using static const structs

Other APIs Struct Handling
==========================

# DirectX sizeof

In DirectX and other windows based API, structs are passed with the first member being the size of the struct. There is type safety in so far as it assumes subsequent value types are extensions of the previous type. Fields can be added - but not removed. If a struct breaks this it will require a binary compatibility breaking change. 

Knowing the struct size, does make the code on entry to the API call somewhat simple. If the struct isn't the latest size, all that is required is to allocate and initialize the current larger version and copy the fields from the value passed in. 

It's not failsafe in use - in that the wrong size can be set, or the setting of the size forgotten.

This mechanism might also allow for an array of value structs - as long as they have all the same type/size.

The DirectX mechanism doesn't allow for a binary incompatible value type to be used - it requires a COM style interface change. 

For subsequent value types, an increasing integer value is suffixed to the type.

This mechanism allows backward and forward compatibility if it's a 'minor' change (only adds fields) and we allow zero initialization.

There is no mechanism to add extra chunks of struct such as Vulkan has. 

There is type safety for the main type.

# Vulkan

In Vulkan structs are passed with a first member of VkStructureType that identifies them. That to allow multiple structures to be passed the structure contains a `pNext` member. The `pNext` member is a `void*` and so no type safety around these types. 

[...]

PR #1776, opened 2021-04-01
Associating GUID (or UUID) with types

jsmall-nvidia:
* Added macros SLANG_COM_INTERFACE and SLANG_CLASS_GUID that can be used to setup a guid in interface/class
* Made the #define SLANG_UUID_ just use the other macros
* Removed the places where guids where defined in C++ files as no longer required

This wasn't a new idea - it's covered in #1645

The main observation is that whilst it's not possible to define a Guid directly within an interface, without it being implemented in a C++ file, it *is* possible to do so within a function. If we inline the function we get much the same effect - although there are no assurances around what the pointers will be across files. 

I tested on compiler explorer, and all major compiler do correctly inline and behave as if it is a global in release.

I did not replace in gfx until deemed appropriate and since lots of changes there.


tfoleyNV (approved):
This looks pretty good, although I suppose there is a chance that a client would rather embed the GUIDs/UUIDs (I'm never totally clear on the distinction, but I know that UUID tends to be used outside of the Microsoft ecosystem) into their application via their own approach rather than rely on the accessor functions generated here. We can probably cross that bridge when we come to it, since the convenience here is high.

One other thing is that we could consider having another case of `as<T>` for `ISlangUnknown`-derived types that calls `queryInterface` with `T::getTypeGuid()` to streamline some of the casting sites. Not a real priority any time soon, though.

PR #1777, opened 2021-04-01
cygwin - disable VK/CUDA

jsmall-nvidia:
Doing so makes most of Slang build on cygwin, but there remains an issue around `relocation truncated to fit: R_X86_64_PC32 against undefined symbol`


tfoleyNV (approved):
I suppose following our current model we should have a `SLANG_CYGWIN` macro instead of directly using `defined(__CYGWIN__)` but I'm not concerned with the details for now.

PR #1782, opened 2021-04-03
Transient root shader object.

csyonghe:
This change refactors `gfx` layer's public interface around root shader objects so that they are transient objects returned by `CommandEncoder::bindPipeline` method. With this change the user is no longer responsible for creating root shader objects themselves.

This biggest change is that D3D12 and VK shader object implementations now sub-allocate constant buffer ranges from `TransientResourceHeap` instead of creating individual API buffer resources. In the D3D12 implementation where each shader object maintains a CPU descriptor table, the descriptor table of a root shader object is now allocated from a per-command-buffer linear, auto-resizable descriptor heap and reset after each use, so the more frequent root shader object reset is now more efficient (since it does not go through the general free-list descriptor heap). Note that even when root shader objects are transient, we still need to stage the root level descriptors in a CPU heap because the full size of the final descriptor table is not known until the root object can be specialized, and therefore we can't preallocate the GPU descriptor table prior to parameter binding.

This PR also fixed a previously un-discovered bug regarding the use of `ID3D12Resource::Map` function. When passing in a range to the `Map` function, the pointer returned is still from the begining of the resource and not the beginning of the input range. So we still need to add the offset to the mapped pointer if it is not zero.

D3D11, GL, CPU and  CUDA implementations still follows the old "creating-root-object" model, since the root object needs to be stored in the faked command buffers anyways to support deferred execution. To allow more reuse of the shared logic for adapting inmmediate targets to the `gfx` interface, `CPUDevice` is now changed to base on `ImmediateRendererBase` instead of rolling its own implementation directly on top of `RendererBase`. `CUDADevice` is left unchanged since it has a non-trivial command queue implementation.


tools/platform/gui.cpp
```diff
@@ -273,7 +273,7 @@ void GUI::endFrame(ITransientResourceHeap* transientHeap, IFramebuffer* framebuf
     auto renderEncoder = cmdBuf->encodeRenderCommands(renderPass, framebuffer);
     renderEncoder->setViewportAndScissor(viewport);
 
-    renderEncoder->setPipelineState(pipelineState);
+    renderEncoder->bindPipeline(pipelineState);
```
tfoleyNV, line 276: Out of curiosity, why the naming change here (from `set` to `bind` and from `PipelineState` to `Pipeline`)? We probably need to figure out a policy for what our terminology is going to be in `gfx` and try to be consistent about it. Does `bind` mean something different from `set`? If so, what?

tools/render-test/render-test-main.cpp
```diff
@@ -400,29 +399,38 @@ SlangResult _assignVarsFromLayout(
     return context.assign(rootCursor, layout.rootVal);
 }
 
-void RenderTestApp::applyBinding(PipelineType pipelineType, ICommandEncoder* encoder)
+Result RenderTestApp::applyBinding(PipelineType pipelineType, ICommandEncoder* encoder)
 {
+    auto slangReflection = (slang::ProgramLayout*)spGetReflection(
+        m_compilationOutput.output.getRequestForReflection());
```
tfoleyNV, line 405: Hmm... I don't really like us having to add this bit. What do we need the layout for here, and is it possible that we can just query the layout from the root shader object instead?

tfoleyNV: I looked at the relevant code to see why this comes up, and it vaguely makes sense because the `//TEST_INPUT` lines may allocate shader objects and thus need to be able to look up types by name. This code here doesn't feel like the appropriate place to be querying the program layout, but I guess its is reasonable for now.

tfoleyNV (approved):

tools/platform/gui.cpp
csyonghe: Originally I was thinking about a name like `beginPipelineBinding`, since the method is really a start of parameter binding. `setPipelineState` sounds more like an one shot, all on its own operation and doesn't feel like what is really going on semantically. I end up with `bindPipeline` which means a little more than just set pipeline state. But can decide to have a more accurate name here.

tools/render-test/render-test-main.cpp
csyonghe: This is what is already there and not newly added in this PR. It is definitely good to figure out how to simplify them.

PR #1788, opened 2021-04-07
Improve robustness of gfx lifetime management.

csyonghe:
This change revists the liftetime management of API objects in `gfx` to make it robust against arbitrary order of `release` calls.
For the most part, the change is about making sure we use `RefPtr<XXXDevice>` to hold a strong reference to the device that creates an object, to make sure that any objects will be freed before `XXXDevice`. This is particularly important for Vulkan, which requires that all objects created from a `VkDevice` must be freed before destroying the `VkDevice` itself.

One major challenge that arise when adding these kind of strong references to `Device` is that for some type of objects, `Device` is also storing a strong reference to them, resulting a cyclic referencing relation that breaks the reference counting GC mechanism. An example is that a `Device` contains a cache for all created / specialized `PipelineState` objects. With `PipelineState` objects also holding strong references to `Device`, they are effectively referencing each other. Note we cannot simply change one of the reference to weak reference in this case: we can't use weak reference to `PipelineState`s in `Device`'s pipeline cache because if the user releases a `PipelineState`, the entry in the cache will become invalid; we can't use weak reference to `Device` in a `PipelineState` either, because this will lead to crash if the user releases a `PipelineState` object after a `Device` object.

The solution to this problem is to use a separate reference counter for the `ISlangUnknown` reference counting. This reference counter represents the number of references from the user to a `gfx` object (i.e. references via `ComPtr`s). When the public reference counter goes from 0 to 1, we increase the private `RefObject` reference counter by 1. When the public reference counter goes from 1 to 0, we decrease the private reference counter by 1. In the specific case where a `PipelineState` object references `Device` and `Device` object references `PipelineState`s, we can simply break the `PipelineState`'s reference to `Device` when the `PipelineState` object no longer has any public references. As long as we know that `Device` holds references to `PipelineState` objects, this will always work regardless whether the user release `Device` first or `PipelineState` first.

With this implementation, we must be careful to increase the public reference counter when returning to the user an interface pointer instead of the private `RefObject` reference counter. To make sure we handle things correctly, this change introduces two sets of helper functions `returnComPtr`, `returnRefPtr` for returning a ComPtr and a RefPtr via return arguments. These helps ensure the correct reference counter is being changed.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -258,13 +258,11 @@ workspace "slang"
     -- optimization off, make sure debug symbols are output,
     -- and add the same preprocessor definition that VS
     -- would add by default.
-    -- We also use dynamic CRT in debug configuration to help
-    -- detect any memory leaks across DLL boundaries.
     filter { "configurations:debug" }
         optimize "Off"
         symbols "On"
         defines { "_DEBUG" }
-        staticruntime "Off"
+        -- staticruntime "Off"
     -- For the release configuration we will turn optimizations on
     -- (we do not yet micro-manage the optimization settings)
     -- and set the preprocessor definition that VS would add by default.
--- a/premake5.lua
+++ b/premake5.lua
@@ -258,11 +258,13 @@ workspace "slang"
     -- optimization off, make sure debug symbols are output,
     -- and add the same preprocessor definition that VS
     -- would add by default.
+    -- We also use dynamic CRT in debug configuration to help
+    -- detect any memory leaks across DLL boundaries.
     filter { "configurations:debug" }
         optimize "Off"
         symbols "On"
         defines { "_DEBUG" }
-
+        staticruntime "Off"
     -- For the release configuration we will turn optimizations on
     -- (we do not yet micro-manage the optimization settings)
     -- and set the preprocessor definition that VS would add by default.
--- a/tools/render-test/render-test-main.cpp
+++ b/tools/render-test/render-test-main.cpp
@@ -117,7 +117,7 @@ protected:
     // variables for state to be used for rendering...
     uintptr_t m_constantBufferSize;
 
-    ComPtr<IDevice> m_device;
+    IDevice* m_device;
     ComPtr<ICommandQueue> m_queue;
     ComPtr<ITransientResourceHeap> m_transientHeap;
     ComPtr<IRenderPassLayout> m_renderPass;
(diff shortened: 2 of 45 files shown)


tools/render-test/render-test-main.cpp
```diff
@@ -840,7 +830,11 @@ static SlangResult _setSessionPrelude(const Options& options, const char* exePat
         SLANG_RETURN_ON_FAIL(TestToolUtil::getRootPath(exePath, rootPath));
 
         String includePath;
-        SLANG_RETURN_ON_FAIL(TestToolUtil::getIncludePath(rootPath, "external/nvapi/nvHLSLExtns.h", includePath));
```
jsmall-nvidia, line 843: What's this change about? Strictly speaking to test for success you need SLANG_SUCCEEDED(). That testing for equality of SLANG_OK, doesn't work on non zero success codes. Admittedly that's rarely used. 

Also this change seems to make it more verbose? Was it part of debugging?

premake5.lua
```diff
@@ -262,7 +262,7 @@ workspace "slang"
         optimize "Off"
         symbols "On"
         defines { "_DEBUG" }
-
+        -- staticruntime "Off"
```
jsmall-nvidia, line 265: I'm assuming this is so that we use dll CRT when debug so we get the superior leak tracking.

jsmall-nvidia (approved):
I guess my worry here is that this solution seems rather subtle and far reaching. 

That for these kinds of ownership scenarios, it's not uncommon to make the 'owner' have the potential to keep track of owned objects. When the owner (which isn't ref'd by the ownees) goes out of scope it invalidates the ownees and releases them. If there refs to them, they stay in scope, but will assert if used. The invalidation requires a mechanism for the owner to put in this state. The underlying idea here is that when the owners ref hits zero, then all things created on it become invalidated. 

That on a release build you could not do any of this tracking. On a debug build, it's not necessary to 'invalidate' even - it can assert all of the active objects, to say they should be destroyed when the device is destroyed. This solution is somewhat more limited than what you have here in that things created on the device can keep the device in scope. But it's not clear to me that the extra flexibility is worth the complexity.

All that being said I don't want to slow down your efforts, over these observations, so will approve.

csyonghe:
This flexibility is needed to match the same semantics of the d3d API, which allows arbitrary release ordering on objects. This kind of flexibility frees the user from writing their own destructors and can just rely on ComPtrs to do the clean up. There is also the use scenario that a user may call getDesc on an object, which can return the user back the device/pool object that creates the resource even when the user has already released the device. To support these scenarios we have to implement this double referencing counting strategy. The change has been tested against memory leaks and shouldn't incur any significant performance overhead.

tools/render-test/render-test-main.cpp
csyonghe: In debug builds, SLANG_RETURN_ON_FAIL triggers an assert that prevents further execution, which is not the desired behavior here.

PR #1794, opened 2021-04-14
Fix for vector initialization combinations

jsmall-nvidia:
The following code wouldn't compile. 

```
float3 x = float3(3, float2(1, 2));
```

The issue being stdlib didn't add all possible combinations of __init. At first a new version of code generating the vector init combinations was written. On looking at the size of the output compared to the code to generate it, it seemed simpler to just explicitly have the combinations which is what ends up here.


source/slang/core.meta.slang
```diff
@@ -427,60 +427,34 @@ struct ParameterBlock {}
...
+__generic<T> __extension vector<T, 4>
+{
+    __init(T x, T y, T z, T w);
+    __init(vector<T,2> xy, T z, T w);
+    __init(T x, vector<T,2> yz, T w);
+    __init(T x, T y, vector<T,2> zw);
+    __init(vector<T,2> xy, vector<T,2> zw);
+    __init(vector<T,3> xyz, T w);
+    __init(T x, vector<T,3> yzw);
+}
```
tfoleyNV, line 451: I am almost *shocked* that this approach works. It is great to see the features of our language actually being usable like this.

tfoleyNV (approved):

PR #1795, opened 2021-04-14
Update `model-viewer` example and fixing compiler bugs.

csyonghe:
This change brings `model-viewer` example up-to-date, so that it uses the `shader-object` system implemented in `gfx` instead of rolling its own solution on the same matter.

Several compiler fixes are made in this change towards making this example to work, but there are more issues to fix before this example can be runnable (it currently compiles but crashes when compiling shader code). The remaining issue is that `PseudoPtr`s slip through code gen and appearing in final HLSL output. We need a closer look at our "pending data layout" implementation and make sure things still work if we specialize a existential parameter with a type that does not fit in the `anyValueSize` of the corresponding interface.

The fixes included in this change are:
1. Make `cast<IRType>` work for `IRSpecializeInst`.
2. Allow `dumpIR` to be specified from the COM interface and from `gfx`.
3. Make `IRBoundInterfaceType` not holding the witness table operand alive.
4. Ensure nested `IRGeneric` values have proper type at all levels.
5. Do not lower a generic intrinsic function, if the `IRTargetIntrinsicDecoration` is only defined on the inner func value but not on the outter generic value.
6. Implement a break-alloc debug mechanism on `IRInst`s, enabled by the `SLANG_ENABLE_IR_BREAK_ALLOC` control macro that is automatically enabled for debug builds (set in `premake5.lua`). When enables, all `IRInst`s will have a `_debugUID` field that contains a unique allocation ID during this run. Natvis and ir-dump will include these IDs. This makes it easy for us to break when a specific inst is being created/processed. The ID counter is properly set so that user code always starts from 0 and stdlib code starts from 0x80000000 so that the IDs for user code is stable regardless of stdlib is being loaded or being compiled during current run.
7. Make natvis visualize `OrderedHashSet` type.


build/visual-studio/slang/slang.vcxproj
```diff
@@ -98,7 +98,7 @@
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
-      <PreprocessorDefinitions>_DEBUG;SLANG_DYNAMIC_EXPORT;SLANG_WITHOUT_EMBEDDED_STD_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;SLANG_DYNAMIC_EXPORT;SLANG_WITHOUT_EMBEDDED_STD_LIB;SLANG_ENABLE_IR_BREAK_ALLOC=1;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
tfoleyNV, line 101: Minor naming question: is it better to have `SLANG_IR_ENABLE_*` or `SLANG_ENABLE_IR_*`? The latter option (which you currently have) reads more naturally, but means that any flags related to a sub-module of the compiler are not naturally grouped by a prefix. The former option means that all config options/defines related to the IR would start with `SLANG_IR_*` which could be nice.

I have no strong opinion on this right now given that we only have the one option being introduced here.

examples/model-viewer/main.cpp
```diff
@@ -0,0 +1,922 @@
...
+
+    // Create a shader object that contains the type info and parameter values
+    // that represent an instance of `SimpleMaterial`.
+    IShaderObject* createShaderObject(RendererContext* context) override
+    {
+        auto program = context->slangReflection;
+        auto shaderType = program->findTypeByName("SimpleMaterial");
+        shaderObject = context->device->createShaderObject(shaderType);
+        gfx::ShaderCursor cursor(shaderObject);
+        cursor["diffuseColor"].setData(&diffuseColor, sizeof(diffuseColor));
```
tfoleyNV, line 164: Not needed for this PR at all, but it would be great if we can get the syntax `cursor["name"] = value;` to work more broadly. Ideally that operation could be a templated `operator=` in `ShaderCursor` that routes through some user-overridable policy function that the user can then specialize for their own types. This would allow something like `model-viewer` to make use of an adapter that allows setting `glm` matrix/vector types directly into shader cursors.

examples/model-viewer/main.cpp
```diff
@@ -0,0 +1,922 @@
...
+{
+    RefPtr<Material>    material;
+    int                 firstIndex;
+    int                 indexCount;
+};
+struct Model : RefObject
+{
+    typedef platform::ModelLoader::Vertex Vertex;
+
+    ComPtr<IBufferResource>     vertexBuffer;
```
tfoleyNV, line 189: Unrelated to this change, but we eventually need to sit down and work out what the ideal system is for making rasterization shaders in Slang be adaptable to work with many vertex layouts (and combinations of vertex attributes). We have solved this with metaprogramming sorts of approaches before, but it seems like we eventually need to demonstrate a more complete solution to the problem.

examples/model-viewer/main.cpp
```diff
@@ -0,0 +1,922 @@
...
+        return shaderObject.get();
+    }
+
+    virtual slang::TypeReflection* getShaderType(RendererContext* context) override
+    {
+        return ::getShaderType<DirectionalLight>(context);
+    }
+};
+
+struct PointLight : Light
```
tfoleyNV, line 326: It's kind of an obvious statement, but if/when we start to support more complete heterogeneous programs, we will need to decide on the best way to support types like this where there is a host side `X` and a GPU-side `X` that have basically the same members/data. It seems clear that we don't want to have to duplicate the declarations here and deal with manually copying from CPU to GPU, but there are also many reasons why it can be useful to have distinct host and GPU-memory allocations for a shader object, and to make creating one from the other explicit. This needs careful thought as we continue to iterate on these examples.

examples/model-viewer/main.cpp
```diff
@@ -0,0 +1,922 @@
...
+// so that we can scope its allocations for easy cleanup, rather than
+// use global variables.
+//
+struct ModelViewer : WindowedAppBase
+{
+
+RendererContext context;
+
+// Most of the application state is stored in the list of loaded models,
+// as well as the active light source (a single light for now).
```
tfoleyNV, line 633: This comment is incorrect when it refers to a "single light". The comments in general will need a pass before we treat this example as being enabled/published again.

examples/model-viewer/shaders.slang
```diff
@@ -458,14 +447,14 @@ float4 fragmentMain(
...
-    // Note that the return type here is `TMaterial.BRDF`,
+    // Note that the return type here is `gMaterial.BRDF`,
     // which is the `BRDF` type *associated* with the (unknown)
     // `TMaterial` type. When `TMaterial` gets substituted for
     // a concrete type later (e.g., `SimpleMaterial`) this
     // will resolve to a concrete type too (e.g., `SimpleMaterial.BRDF`
     // which is an alias for `BlinnPhong`).
     //
-    TMaterial.BRDF brdf = gMaterial.prepare(g);
+    let brdf = gMaterial.prepare(g);
```
tfoleyNV, line 457: Out of curiosity, does it work to write `gMaterial.BRDF` here? I would hope that it Just Works in Slang to use types that are members of values, but I haven't checked how well the compiler handles it.

source/slang/slang-ir-dce.cpp
```diff
@@ -120,6 +120,15 @@ struct DeadCodeEliminationContext
...
             {
+                switch (inst->getOp())
+                {
+                case kIROp_BoundInterfaceType:
+                    if (inst->getOperand(ii)->getOp() == kIROp_WitnessTable)
+                        continue;
+                    break;
+                default:
+                    break;
+                }
```
tfoleyNV, line 131: It would be good to eliminate special cases like this from the DCE pass, by making sure the source instructions actually represent the dependencies of the code. Allowing things to be removed as dead code even when they are referenced like this seems like it could break some assumptions/invariants in the DCE pass.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6598,6 +6608,34 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         return nullptr;
     }
 
+    static bool isChildOf(IRInst* child, IRInst* parent)
```
tfoleyNV, line 6611: Shouldn't this technically be `isDescendentOf`?

tfoleyNV, line 6611: Shouldn't this technically be `isDescendentOf`?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6612,9 +6650,94 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+
         while (parentGeneric)
         {
+            // Create a universal type in `outterBlock` that will be used
+            // as the type of this generic inst. The return value of the
+            // generic inst will have a specialized type.
+            // For example, if we have a generic function
+            // g0 = generic<T> { return f: T->int }
+            // The type for `g0` should be:
+            // g0Type = generic<T1> { return IRFuncType{T1->int} }
```
tfoleyNV, line 6664: Doesn't this lead to infinite regress, since now `g0Type` needs a type? I guess at that point we have:
```
g0TypeType = generic<T2> { return IRType; }
```
And then `g0TypeType` can be used as the "type" of all generics with a single type parameter that return a type (that is, for the kind `Type -> Type`). We'd have to make sure to actually re-use those types.

That's definitely the bit where I'm still concerned that having "the type of a generic is a generic" isn't 100% right. We really need something that can represent `forall T . T->int` as a *type* and not a specific generic *value*.

This is getting to be a worrying hole in our IR as we start to care more and more about getting the types of generics "right."

tfoleyNV: Note: The challenge for making `forall` work as a type in our IR is basically that we've wanted types (other than nominal types) to be defined 100% by their operands (no need to inspect decorations/children). To support a type like `forall T . T->int` that means that `T->int` needs to somehow be an *operand* of the `forall` instruction, but that means that the declaration of `T` cannot somehow be a nested `IRParam` scoped under the `forall`.

I *think* the solution in this case is an encoding with something a bit like De Brujin indices:

* We have an IR instruction like `placeholderParam(index) : Type` which is uniquely defined by the `index` and `Type`.
* A `forall T0, T1, ... . whatever` instruction is encoded as `univeralQuantify(whatever, T0, T1, ...)` where each of the `Ti` is a `placeholderParam` that is suitable to the kind of the generic parameter and its index `i`. E.g., the `T0` might be `placeholderParam(0) : Type` for the generic in this example.
* The `whatever` result of a `universalQuantify` should then be an ordinary IR instruction (usually a type), but one that may make use of the `placeholderParam`s from whatever quantifies it. In the example above, that would be `IRFuncType(placeHolderParam(0)->int)`

Thus for the example, we might have:

```
let T1 = placeholderParam(0) : Type;
let F = IRFuncType(T1 -> int);
let g0Type = universalQuantify(T1 -> F);
let g0 : g0Type = generic<...> { ... } // unchanged from current encoding
```

This approach avoids using IR generics to encode the types of things, and seems like it is logically more consistent.

tfoleyNV (approved):
There's definitely more work to be done, but this looks like a good start.

PR #1805, opened 2021-04-20
Add `ISession::getParameterBlockLayout()`

csyonghe:
This allows a user to obtain a `TypeReflectionLayout` for `ParameterBlock<X>` from `X`. This API is added while I am fixing vulkan shader object issues, but it was not used in the end. However this addition may be useful in the future, so it doesn't hurt to add it anyways.


slang.h
```diff
@@ -3933,6 +3933,12 @@ namespace slang
             LayoutRules     rules = LayoutRules::Default,
             ISlangBlob**    outDiagnostics = nullptr) = 0;
 
+        virtual SLANG_NO_THROW TypeLayoutReflection* SLANG_MCALL getParameterBlockLayout(
+            TypeReflection* elementType,
+            SlangInt        targetIndex = 0,
+            LayoutRules     rules = LayoutRules::Default,
+            ISlangBlob**    outDiagnostics = nullptr) = 0;
+
```
tfoleyNV, line 3941: It seems like nine times out of ten what the user actually wants is a layout for `ConstantBuffer<X>` instead of `ParameterBlock<X>`, because the parameter block layout is really no different. I guess we can/should have both eventually if we keep going down this road.

Obviously the more powerful thing is to have an API way to form `X<Y>` for arbitrary generic type name `X` but, as discussed elsewhere, that API can't be supported easily today because `X` in that case is not a proper type (and isn't even a `Type` internall) and thus can't fit through the bottleneck of our current `Type` reflection system.

tfoleyNV (approved):

PR #1807, opened 2021-04-22
Add doc on matrix layout

csyonghe:


docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -0,0 +1,56 @@
...
+The differences on default matrix layout or storage conventions between GLSL (OpenGL/Vulkan) and HLSL has been an issue that frequently causes confusion among developers. When writing applications that work on different targets, one important goal that developers frequently seek is to make it possible to pass the same matrix generated by host code to the same shader code, regardless of what graphics API is being used (e.g. Vulkan, OpenGL or Direct3D). As a solution to shader cross-compilation, Slang provides necessary tools for developers navigate around the differences between GLSL and HLSL targets.
+
+Row-major vs Column-major storage layouts
+-------------------------
+In this article, the term matrix layouts or storage layouts refers to the order of matrix elements in memory. Specifically, we use the following umambiguous technique to determine whether a matrix is in `row-major` and `column-major` layout: if we have a 4x4 matrix defined as `float m[16]`, and the translation terms are stored at `m[12], m[13], m[14]`, then we say this matrix is in `column-major` layout. If the translation terms are stored at `m[3], m[7], m[11]`, then we say this matrix is in `row-major` layout.
```
jsmall-nvidia, line 12: This definition of row and column major seems to be the opposite of what I'd expect? 

https://en.wikipedia.org/wiki/Row-_and_column-major_order

It is consistent with the later code in this doc. 

Though if I was using 'row major' on HLSL, I'd expect to be doing 

```
float4 result = mul(vertexPosition, modelViewProjectionMatrix);
```

Perhaps all the flipping comes out in the wash, in that the code above in Slang, will be flipped in the HLSL output (to give what you have below), and then the flipping of the definition of row major makes that all alright...

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -0,0 +1,56 @@
...
+```
+gl_Position = vertexPosition * modelViewProjectionMatrix;
+```
+The correct HLSL code should be:
+```
+float4 result = mul(modelViewProjectionMatrix, vertexPosition);
+```
+
+As you can see, the ordering of matrix multiplication is always reversed between GLSL and HLSL.
+Slang will always translate `mul(a, b)` to `b * a` when generating GLSL code. By default, Slang assumes all matrices to be in `row-major` layout, so a shader code written in Slang should be written as `mul(modelViewProjectionMatrix, vertexPosition)`, and the host application should always make sure to pass matrices in row-major layout to shaders. However, the user can always override this behavior with a compiler flag to make Slang compiler assume `column-major` matrix layout by default. This compiler flag can be specified during the creation of a `Session`:
```
jsmall-nvidia, line 35: This again seems to be the opposite of 'normal' HLSL behaviour - if it's row major I expect to write `mul(vertexPosition, modelViewProjectionMatrix)`.

Perhaps I'm thinking fxc and in the dxc they flipped it?

jsmall-nvidia (approved):
Clarified in discussion elsewhere.

Issue #1796, opened 2021-04-15
float16/half support for CUDA


jsmall-nvidia:
CUDA has support for the half/float16 type, but currently the Slang language doesn't support it. At the moment half actually becomes float in CUDA output. This might seem a little odd, but HLSL has a precedent of this behavior, with sm6_2 bringing actual support. 

In order to get half support in CUDA is typically achieved by  `#include "cuda_fp16.h"` from the CUDA SDK. This introduces a variety of intrinsics as well as the `__half` and `__half2` types. It may be possible/desirable to not include cuda_fp16.h and just copy the appropriate definitions into the `cuda-prelude.h`. If they are defined in prelude - this would preclude them being included with "cuda_fp16.h" unless this is somehow guarded. 

The introduced types are essentially opaque - it is not possible to directly operate on them, there are intrinsics to do that. Moreover there are yet other intrinsics that perform operations on __half2 types.

An initial implementation might not want to deal with supporting half maths, but being able to convert between half types and other types that arithmetic can be performed on. In such an implementation it would be useful if it supported the HLSL vector type for half (if not any operations on it). It does open a question on how `vector<half,3>` would be implemented... 

```
struct half3_1 { __half2 xy; __half2 z_; }
struct half3_2{ __half2 xy; __half z; }
struct half3_3{ __half x, y, z; }
```
half3_3 has the potential for tightest packing, but loses the performance advantages of being able to do half2 ops in parallel. If implemented halfX with __half2 type it would require different behavior, as it would require different intrinsics, based on X. 

Once this decision is made it is necessary to have support for doing the conversions. The HLSL intrinsics are

```
uint16_t asuint16(float16_t value);
vector<uint16_t,N> asuint16<let N : int>(vector<float16_t,N> value);
matrix<uint16_t,R,C> asuint16<let R : int, let C : int>(matrix<float16_t,R,C> value);

float16_t asfloat16(uint16_t value);
vector<float16_t,N> asfloat16<let N : int>(vector<uint16_t,N> value);
matrix<float16_t,R,C> asfloat16<let R : int, let C : int>(matrix<uint16_t,R,C> value);

// signed

int16_t asint16(float16_t value);
vector<int16_t,N> asint16<let N : int>(vector<float16_t,N> value);
matrix<int16_t,R,C> asint16<let R : int, let C : int>(matrix<float16_t,R,C> value);

float16_t asfloat16(int16_t value);
[...]

PR #1808, opened 2021-04-22
Preliminary CUDA Half support

jsmall-nvidia:
* Added Float16/Half enable option to DownstreamCompiler::CompileOptions
* Used CUDAExtensionTracker to determine if half is used
* Automatic mechanism for looking up cuda_f16.h 
** Can be overridden by specifying the path to the include with the standard Slang include mechanism
* Made DefaultSharedLibrary queryable (not actually needed in the end)
* Added preliminary support for half and vector types to stdlib and cuda-prelude.h

There remains a somewhat odd problem around how HLSL deals with half. 

```
uint f32tof16(float value);
```

Returns uint not half. That's a problem because on CUDA half is a distinct (it's 16 bits) and real type. Similarly f32tof16 takes a uint. For the moment I have overridden this with taking half. 

Overriding won't work for f32tof16 - so have a special version that does return half (currently just use _ suffix - that will need to change).

Note also that initially in  #1796 it seemed plausible to not include cuda_f16.h, but it seems that the 'intrinsics' are not actually intrinsics, they are defined as inline asm in cuda_f16.hpp(!).


tests/compute/half-opaque-convert.slang
```diff
@@ -0,0 +1,28 @@
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-dx12 -compute -use-dxil -profile cs_6_2 -render-features half -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-vk -compute -profile cs_6_2 -render-features half -shaderobj
+//TEST(compute):COMPARE_COMPUTE:-cuda -compute -render-features half -shaderobj
+
+//DISABLE_TEST(compute):SIMPLE:-target ptx -stage compute -entry computeMain
+
+// The following test shows if half can be processed as an opaque type. This means
+// it can be copied, and converted to and from float types, but nothing else. 
+// No maths, no swizzling.
```
csyonghe, line 9: It should be allowed to load a `half` from a `StructuredBuffer<half>`, right?

csyonghe (approved):

tests/compute/half-opaque-convert.slang
jsmall-nvidia: In principal yes - although I need to check for sure as DX (and to a lesser degree GLSL) are somewhat vague around the half/float16 type. On CUDA half should be fine - but half3 is an issue in that the definition in prelude is 8 bytes, but perhaps it should be 6 but be slower.

PR #1809, opened 2021-04-22
Fix errors in matrix-layout documentation

csyonghe:


docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -7,34 +7,22 @@ Handling Matrix Layout Differences on Different Platforms
...
 
-If the host code generates a `column-major` matrix, then a vertex shader written in GLSL should be like this:
-```
-gl_Position = modelViewProjectionMatrix * vertexPosition;
-```
-The same vertex shader written in HLSL should be:
-```
-float4 result = mul(vertexPosition, modelViewProjectionMatrix);
-```
+By default, Slang assumes all matrices to be in **row-major** layout, since this is the most nature layout to work with in CPU code: each row of the matrix occupies contiguous space in memory. A user should stick to one of the above practices to get correct result. Note that this is different from `fxc` which assumes `column_major` layout by default. As an example, if the host code uses `glm` library to generate transform matrices, the translation terms will be stored in `[12], [13], [14]` locations in memory. Therefore, the shader code should stick to the `mul(v,m)` convention to ensure correctness.
```
jsmall-nvidia, line 18: nature -> natural

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -7,34 +7,22 @@ Handling Matrix Layout Differences on Different Platforms
...
 
-If the host code generates a `row-major` matrix, the correct GLSL code should be:
-```
-gl_Position = vertexPosition * modelViewProjectionMatrix;
-```
-The correct HLSL code should be:
-```
-float4 result = mul(modelViewProjectionMatrix, vertexPosition);
-```
+Slang automatically handles the convention differences when cross-compiling code to GLSL. For example, a `float3x4` matrix will be translated to `mat4x3` in the resulting GLSL. Correspondingly, `mul(v, m)` will be translated to `m*v` in GLSL. Therefore, as long as the user is sticking to the above practices consistently, they will get correct result with the same matrix value in memory regardless of what graphics API they are actually using.
```
jsmall-nvidia, line 20: sticking -> sticks

PR #1815, opened 2021-04-23
Matrix docs update

jsmall-nvidia:


docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -5,24 +5,48 @@ layout: user-guide
...
+The differences between default matrix layout or storage conventions between GLSL (OpenGL/Vulkan) and HLSL has been an issue that frequently causes confusion among developers. When writing applications that work on different targets, one important goal that developers frequently seek is to make it possible to pass the same matrix generated by host code to the same shader code, regardless of what graphics API is being used (e.g. Vulkan, OpenGL or Direct3D). As a solution to shader cross-compilation, Slang provides necessary tools for developers navigate around the differences between GLSL and HLSL targets.
+
+A high level summary:
+
+* Default matrix **layout** in memory for Slang is `column-major`.
```
csyonghe, line 12: I believe the default in Slang is `row-major`.

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -5,24 +5,48 @@ layout: user-guide
...
+This behavior is *independent* of how a matrix layout in memory. Host code needs to be aware of how a shader code will interpret a matrix stored in memory, it's layout, as well as the vector interpretation convention used in shader code (ie `mul(v,m)` or `mul(m, v)`).
+
+[Matrix layout](https://en.wikipedia.org/wiki/Row-_and_column-major_order) can be either `row-major` or `column-major`. The difference just determines which elements are contiguous in memory. `Row-major` means the rows elements are contiguous. `Column-major` means the column elements are contiguous.
+
+Another way to think about this difference is in terms of where translation terms should be placed in memory when filling a typical 4x4 transform matrix. For `row-major` matrix layout, translation will be at `m + 12, 13, 14`. For a `column-major` matrix layout, translation will be at `m + 3, 7, 11`.
```
csyonghe, line 41: This paragraph is not accurate. `row-major` or `column-major` by itself does not dictate where translation terms are stored, it must be combined with whether you are sticking to row-vector or column-vector convention. When using row-vector convention (writing `v*m` in shader code) and using `row-major` matrix layout, then translation terms are stored in 12,13,14. Fliping either one of the math convention or layout majorness flips the matrix storage.

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -33,37 +57,40 @@ slang::ISession* session;
...
--------------
+Examples
+--------
 
-In summary, we put together all options you have to ensure correct result:
+There are 4 variables:
 
-**Option 1: using row-major matrix layout, and `mul(m, v)` math convention**
+* Host vector interpretation (row or column) - and therefore effective tranform order (column) `m * v` or (row) `v * m`
+* Host matrix memory layout
```
csyonghe, line 74: I am not sure if it makes sense to separate matrix memory layout to host and shader, because it only makes logical sense if host and shader assumes the same layout, no?

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -33,37 +57,40 @@ slang::ISession* session;
...
+* Shader vector interpretation (as determined via `mul(v, m)` or `mul(m, v)`
+* Shader matrix memory layout 
 
-- Make sure the host code fills in matrices in the odering so that translation terms are specified in `m[3], m[7], m[11]` elements.
-- Leave `defaultMatrixLayoutMode` as default value when creating a Slang session, or specify `SLANG_MATRIX_LAYOUT_ROW_MAJOR`.
-- Write `mul(Matrix, Vector)` in shader code to transform `Vector` by `Matrix`.
+The simplest scenario is that they are all `row-major`/`row vector`. In that case code will be most easily portable. NOTE! Currently this requires setting Slangs default matrix layout mode to `row-major`, as it's default is `column-major`.
```
csyonghe, line 78: I believe slang's default has already changed to `row-major`.

docs/user-guide/a1-01-matrix-layout.md
```diff
@@ -33,37 +57,40 @@ slang::ISession* session;
...
-**Option 4: using column-major matrix layout, and `mul(v, m)` math convention**
+* row vector host interpretation
+* row-major matrix host layout
+* column vector shader interpretation
+* column-major matrix shader layout
 
-- Make sure the host code fills in matrices so that translations terms are specified in `m[3], m[7], m[11]` elements.
-- Set `defaultMatrixLayoutMode` to `SLANG_MATRIX_LAYOUT_COLUMN_MAJOR` when creating a Slang session.
-- Write `mul(Vector, Matrix)` in shader code.
+Will be ok as are 2 (ie *even*) 'row' (one vector interpretation and the other matrix layout).
```
csyonghe, line 93: I would still prefer we use the original wording that just lists out all viable options a developer can have, so readers can directly refer to them and use it as a checklist for the right set-up.

csyonghe (changes requested):
The callout of `row-vector` and `column-vector` is a good one, but I think Slang has already changed its default to `row-major`. I still think that the original enumeration of all viable set of set-up is a reference for developers so I'd prefer we keep it there.

docs/user-guide/a1-01-matrix-layout.md
jsmall-nvidia: On doing tests in #1814 it appears to be column-major? I can see an argument for making the default row-major.

docs/user-guide/a1-01-matrix-layout.md
jsmall-nvidia: True. That the original idea loses some of it's appeal when having to make the distinction. Added row vector.

docs/user-guide/a1-01-matrix-layout.md
jsmall-nvidia: I think it's a useful distinction - especially if we don't have the ability to just set the convention, as is the case for C++/CUDA. That as is discussed now host matrix layout is not arbitrary if performance is a concern.

docs/user-guide/a1-01-matrix-layout.md
csyonghe: If that is the case, it is likely render-test is explicitly specifying column-major. We should double check if that is the case, since it contradicts with my experience when doing the examples.

csyonghe (approved):

PR #1822, opened 2021-04-27
`gfx` DebugCallback and debug layer.

csyonghe:
This change adds a `IDebugCallback` interface to `gfx`. The user can supply a debug callback with a call to `gfxSetDebugCallback()`, the layer will use the callback to report error messages (including errors generated in underlying API levels).

The change also adds a debug layer to `gfx`, that can be enabled by calling `gfxEnableDebugLayer()` before calling `gfxCreateDevice()`. The debug layer validates calls and report errors using the debug callback provided by the user.

Currently the debug layer contains only a few checks. More checks will be added in the future.


examples/example-base/example-base.cpp
```diff
@@ -25,6 +25,10 @@ Slang::Result WindowedAppBase::initializeBase(const char* title, int width, int
     gWindow->events.sizeChanged = Slang::Action<>(this, &WindowedAppBase::windowSizeChanged);
 
     // Initialize the rendering layer.
+#ifdef _DEBUG
+    // Enable debug layer in debug config.
+    gfxEnableDebugLayer();
+#endif
```
tfoleyNV, line 31: Whey can't the debug layer stuff be something passed in via `IDevice::Desc`? I'm confused why it needs to be global state.

examples/example-base/example-base.cpp
```diff
@@ -170,6 +174,53 @@ int64_t getCurrentTime() { return std::chrono::high_resolution_clock::now().time
...
+        OutputDebugStringW(L"\n");
+#endif
+    }
+};
+
+void initDebugCallback()
+{
+    static DebugCallback callback = {};
+    gfxSetDebugCallback(&callback);
+}
```
tfoleyNV, line 222: Yeah, it seems like the `IDevice::Desc` type could just have an `IDebugCallback* debugCallback = nullptr;` field, and then seting that field implicitly enables the debug layer, right? Optionally we could have a distinct `enum` field for setting the "level" of debugging that the user wants.

tools/gfx/debug-layer.cpp
```diff
@@ -0,0 +1,948 @@
...
+#endif
+
+thread_local const char* _currentFunctionName = nullptr;
+struct SetCurrentFuncRAII
+{
+    SetCurrentFuncRAII(const char* funcName) { _currentFunctionName = funcName; }
+    ~SetCurrentFuncRAII() { _currentFunctionName = nullptr; }
+};
+#define SLANG_GFX_API_FUNC SetCurrentFuncRAII setFuncNameRAII(SLANG_FUNC_SIG)
+#define SLANG_GFX_API_FUNC_NAME(x) SetCurrentFuncRAII setFuncNameRAII(x)
```
tfoleyNV, line 26: Setting a thread-local variable on entry to every `gfx` function seems like it might introduce a lot of overhead. Have we confirmed that the the performance impact isn't actually a concern?

tools/gfx/debug-layer.cpp
```diff
@@ -0,0 +1,948 @@
...
+    UInt offset)
+{
+    SLANG_GFX_API_FUNC;
+    auto innerBuffer = static_cast<DebugBufferResource*>(buffer)->baseObject.get();
+    baseObject->setIndexBuffer(innerBuffer, indexFormat, offset);
+}
+
+void DebugRenderCommandEncoder::draw(UInt vertexCount, UInt startVertex)
+{
+    SLANG_GFX_API_FUNC;
```
tfoleyNV, line 685: Obvious opportunity: we could validate here that `draw()` operaitons are only being encoded when a valid graphics pipeline state is in place.

tools/gfx/debug-layer.cpp
```diff
@@ -0,0 +1,948 @@
...
+        return SLANG_FAIL;
+    }
+    returnComPtr(entryPoint, m_entryPoints[index]);
+    return SLANG_OK;
+}
+
+Result DebugShaderObject::setData(ShaderOffset const& offset, void const* data, size_t size)
+{
+    SLANG_GFX_API_FUNC;
+    return baseObject->setData(offset, data, size);
```
tfoleyNV, line 878: Obvious validation opportunity here is to calculate the ordinary data size for an object at creation time and validate that the range being set here is inside the object.

tools/gfx/debug-layer.cpp
```diff
@@ -0,0 +1,948 @@
...
+    SLANG_GFX_API_FUNC;
+    auto objectImpl = static_cast<DebugShaderObject*>(object);
+    m_objects[ShaderOffsetKey{offset}] = objectImpl;
+    return baseObject->setObject(offset, objectImpl->baseObject.get());
+}
+
+Result DebugShaderObject::setResource(ShaderOffset const& offset, IResourceView* resourceView)
+{
+    SLANG_GFX_API_FUNC;
+    auto viewImpl = static_cast<DebugResourceView*>(resourceView);
```
tfoleyNV, line 915: Yeah, we really want operations like this to be able to do validation on the `offset` being passed in and make sure that it properly identifies a binding range of an appropriate type for the review being passed in. We'd want to validate that the resource was allocated to support whatever state/usage the range requires.

It would also seem important to validate at draw/dispatch time that the objects bound to the pipeline via shader objects are in the required state for the operations being done.

tfoleyNV (approved):
This seems like a great start, and it would be good to figure out just how much validation we can do in this layer.

PR #1824, opened 2021-04-28
Add gfx user's guide.

csyonghe:
Setup new documentation folder for `gfx`.

The existing `user-guide` layout is updated to allow sharing between different user-manual style documentations that has a TOC.

The new documentation will be accessible from https://shader-slang.github.io/slang/gfx-user-guide/ once this PR is in.

Note: the current setup requires manually running `build_toc.ps1` to update table of contents before checking in. We need to make sure to run this script whenever we update the title/subsection structure of any document. We may want to figure out a way to automatically run this script before check in, or have the current `verify-vsproject` CI check to also check consistency of table of contents.


docs/gfx-user-guide/01-getting-started.md
```diff
@@ -0,0 +1,342 @@
...
+{
+    gfxEnableDebugLayer();
+    gfxSetDebugCallback(&gCallback);
+
+    IDevice::Desc deviceDesc = {};
+    gfxCreateDevice(&deviceDesc, &gDevice);
+}
+```
+
+Creating a Pipeline State
```
tfoleyNV, line 82: We probably need/want to iterate on the ordering a bit here. This flow jumps right into creating a pipeline state before setting up queues or any of that stuff. I would have expected the flow to be more organized around the order in which an application needs to initialize/create the objects at runtime (you probably create your queue(s) shortly after creating your device, etc.).

The code blocks here also seem to go a bit long, so we will eventually want to edit things to have more prose in between to explain what is going on.

In the specific case of the pipeline state creation here, I might argue that we want to leave out the issue of how the user creates a `slang::IComponentType` to represent their program, at least at first. If we assume that the user already has the component type, then it becomes much easier to show how to create a pipeline state, since it is almost trivial for the compute case.

tfoleyNV (approved):

PR #1825, opened 2021-04-29
Simplify CommandLine by removing Escaping

jsmall-nvidia:
#1823 provided mechanisms to traffic between escaped and unescaped strings. In particular it provides a platform specific interface to traffic. Having this mechanism allows removing the need to know if a string is escaped or not in the CommandLine type - we can require it is always unescaped. If we have a source that is escaped, we can use the the appropriate logic to unescape it.

Note that in practice this doesn't devolve into just using ProcessUtil::getEscapeHandler() and StringEscapeUtil everywhere, because for many sources (such as the test files), the escaping used has to work across targets, so a cross platform handler has to be used.


tools/slang-test/slang-test-main.cpp
```diff
@@ -1438,12 +1412,16 @@ TestResult runCompile(TestContext* context, TestInput& input)
     CommandLine cmdLine;
     _initSlangCompiler(context, cmdLine);
 
+    StringEscapeHandler* escapeHandler = StringEscapeUtil::getHandler(StringEscapeUtil::Style::Space);
+
     for (auto arg : input.testOptions->args)
     {
         // If there is a quote in the string, assume it is 'escaped'. 
-        if (arg.indexOf('"') >= 0)
+        if (arg.indexOf(escapeHandler->getQuoteChar()) >= 0)
```
csyonghe, line 1420: Why is this logic not part of StringEscapeUtil? I feel like there are a lot of design complexity involved around EscapeHandler, but most use sides really just want to deal with a simple set of helper functions.

csyonghe (approved):

PR #1826, opened 2021-04-29
Update gfx back-ends to handle static specialization

tfoleyNV:
The main goal here is to make the D3D11, D3D12 and Vulkan back-ends support static specialization of interface types in the case where the data for the type won't "fit" in the pre-allocated space for existential values. This includes all cases where the concrete type being specialized to has resources/samplers/etc., as well as any cases where its ordinary/uniform data exceeds the space available.

(Note that the CPU and CUDA targets don't need this work since they can (in theory) support arbitrary-size data in the fixed-size existential payload by using pointer indirection. Actually supporting indirection in those cases should be a distinct change)

The Slang compiler already performs layout for programs that have this kind of data that doesn't "fit," and it lays them out using an idea of "pending" type layouts. Basically, a type that contains some amount of specialized interface-type fields will produce both a "primary" type layout that just covers the data for the unspecialized case, as well as "pending" type layout that describes the layout for all the extra data needed by specialization.
When laying out a `ConstantBuffer<X>` or `ParameterBlocK<X>` ("CB" or "PB"), the front-end will try to place as much of that "pending" data into the layout of the buffer/block itself as is possible. That means that both CBs and PBs will be able to allocate trailing bytes for any ordinary data in the "pending" layout. PBs will be able to allocate any trailing resources/samplers into their layout, but for CBs they will spill out to be part of the pending layout for the buffer itself.

In order for the back-ends to properly handle pending data, they need to *either* assume the exact layout rules used by the front-end and try to reproduce them (e.g., by iterating over binding ranges and sub-objects in the exact same order that front-end layout would enumerate them), *or* they need to respect the reflection information produced by the front-end. This change takes the latter approach, trying to make only minimal assumptions about the layout rules being used. This choice is motivated by wanting to decouple the `gfx` implementation from the compiler front-end, especially insofar as this work has made me question whether the current layout rules are the best ones possible.

A common theme across all the implementations is to have a fixed-size type that can represent "binding offsets" for the chosen back-end. The offset type has fields that depend on the API-specific way bindings are indexed; e.g., for D3D11 it has offsets for CBV, SRV, UAV, and sampler bindings. This fixed-size offset type can be filled in based on Slang reflecton information, and then used to compute derived offsets with just a few add operations.

The simple offset type for each API is then extended to produce an offset type that includes both the offsets for "primary" data and also the offsets for "pending" data. Most logic that traffics in offsets doesn't have to know about this more complicated representation.

Making consistent use of these offsets required that I pretty much rewrite the logic that actually applies shader objects to the API state. Doing so might be lowering the efficiency of the system in the near term, but the increase in clarity was important for getting the work done, and it seems like it will also be important if/when we start trying to perform special-case optimizations around root and entry-point parameter setting.

While there are many API-specific differences, we can identify a repeated pattern where many steps, whether applying parameters to the pipeline stage or constructing signatures / layouts, can be broken down into three main operations on `ShaderObject`s or their layouts:

* `*AsValue()` is the core operation, and is the one used for the `ExistentialValue` case most of the time. It ignores the ordinary data in the object, and instead processes all nested binding ranges (for resources/smaplers) and sub-objects.

* `*AsConstantBuffer()` handles the `ConstntBuffer<X>` case, by dealing with the implicit buffer for ordinary data (if it is needed) and then delegates to the `*AsValue()` case.

* `*AsParameterBlock()` handles the `ParameterBlock<X>` case, by allocating/preparing/etc. any descriptor tables/sets that would be required for the current object/layout and then delegating to `*AsConstantBuffer()` to do the rest

The idea is that by having the parameter block case delegate to the constant buffer case, which delegates to the value/existential case, we can streamline a lot of the logic so that it doesn't seem quite as full of special cases.

Note: When preparing this pull request I spent a reasonable amount of time trying to clean up the D3D11 and Vulkan implementations, so they are probably the easiest to read and understand when it comes to the new code. Doing the cleanup work also helped to work out some weird corner case bugs/issues. In contrast, the D3D12 path hasn't had as much attention given to cleanliness and comments, so it really needs some attention down the line to get things into a state that is easier to understand.


examples/example-base/example-base.cpp
```diff
@@ -26,6 +26,7 @@ Slang::Result WindowedAppBase::initializeBase(const char* title, int width, int
 
     // Initialize the rendering layer.
     IDevice::Desc deviceDesc = {};
+    deviceDesc.deviceType = gfx::DeviceType::DirectX12;
```
csyonghe, line 29: We can remove this line since dx12 is already the default. We need to make our examples able to take command line arguments to determine what API to use for our debugging purposes.

examples/model-viewer/shaders.slang
```diff
@@ -85,7 +85,13 @@ struct BlinnPhong : IBRDF
         float3 H = normalize(L + V);
         float nDotH = saturate(dot(N, H));
 
-        return kd*nDotL + ks*pow(nDotH, specularity);
+        // TODO: The current model loading has a bug that is leading
```
csyonghe, line 88: Is this still true?

slang.h
```diff
@@ -1991,6 +1991,8 @@ extern "C"
     SLANG_API SlangReflectionTypeLayout* spReflectionTypeLayout_getPendingDataTypeLayout(SlangReflectionTypeLayout* type);
 
     SLANG_API SlangReflectionVariableLayout* spReflectionTypeLayout_getSpecializedTypePendingDataVarLayout(SlangReflectionTypeLayout* type);
+    SLANG_API SlangInt spReflectionType_getSpecializedTypeArgCount(SlangReflectionType* type);
```
csyonghe, line 1994: Should these be moved to be around other SlangTypeReflection_* functions instead of mixing in the SlangReflectionTypeLayout functions?

source/slang/slang-reflection-api.cpp
```diff
@@ -2033,6 +2140,25 @@ SLANG_API SlangInt spReflectionTypeLayout_getSubObjectRangeSpaceOffset(
     return extTypeLayout->m_subObjectRanges[subObjectRangeIndex].spaceOffset;
 }
 
+SLANG_API SlangReflectionVariableLayout* spReflectionTypeLayout_getSubObjectRangeOffset(
```
csyonghe, line 2143: Should it be named getSubObjectRangeVarLayout?

csyonghe (dismissed):
I have not have a close look into vk and d3d12. The d3d11 implementation looks very clean. Is OpenGL not implemented, or is it that simple to get OpenGL support pendinglayout?

examples/example-base/example-base.cpp
tfoleyNV: Sorry, that is debugging code that slipped in.

examples/model-viewer/shaders.slang
tfoleyNV: Yes. I didn't try to fix the model loading bug as part of this change. If you set a breakpoint at the places where `SimpleMaterial` instances are being created in C++ you will find that one of them is being filled in with garbage data.

slang.h
tfoleyNV: Probably, but the C reflection API is such a messy implementation detail at this point that I wasn't inclined to bother.

source/slang/slang-reflection-api.cpp
tfoleyNV: I suppose it could. It's a little tricky because it doesn't really represent the entire range so much as the "leaf" variable where the range starts.

I guess there's actually one high-order-bit design thing here that could impact the entire "flat" reflection API: it is possible that we should just have every binding range support a `getBindingRangeLeafVarLayout()` to match the `getBindingRangeLeafTypeLayout()` and at that point we realistically don't need the have a separate reflection setup for "descriptor ranges" at all.

examples/model-viewer/main.cpp
```diff
@@ -282,16 +282,17 @@ RefPtr<Model> loadModel(
...
+    //
+    virtual void writeTo(ShaderCursor const& cursor) = 0;
 
     // Retrieves the shader type for this light object.
     virtual slang::TypeReflection* getShaderType(RendererContext* context) = 0;
 
     // The shader object for a light will be stashed here
     // after it is created.
-    ComPtr<IShaderObject> shaderObject;
+//    ComPtr<IShaderObject> shaderObject;
```
csyonghe, line 295: I guess we can just remove this?

csyonghe (approved):

examples/model-viewer/main.cpp
tfoleyNV: Yes. I missed that one during cleanup.

One note here is that we could in theory have an application-side base class for all things that are "shader-object-able," and share that across light/material/etc. In that case the ability to create/cache a shader object could be a built-in feature of that base class (no `virtual` functions needed). E.g.:

```
// needs a real name
class ShaderObjectifiable : RefObject
{
    virtual slang::TypeReflection* getShaderType(RenderContext* context) = 0;
    virtual void writeInto(gfx::ShaderCursor const& cursor) = 0;

    IShaderObject* getShaderObject(gfx::Device* device)
    {
        if(shaderObject) return shaderObject;

        shaderObject = device->createShaderObject(getShaderType());
        writeInto(shaderObject);
        return shaderObject;
    }

protected:
    ComPtr<IShaderObject> shaderObject;
}
```

PR #1829, opened 2021-04-29
Clean up the way we stride over sub-object ranges for D3D11 and Vulkan

tfoleyNV:
This change applies to the case where we have a sub-object range with more than one object in it (so arrays of constant buffers, parameter blocks, or existentials). It is worth noting that we don't really seem to have any tests covering this case right now, so it is entirely possible that things are busted already, and/or that this change doesn't work the way I think it does.

When we go to actually bind the state from a sub-object range into the pipeline, we care about:

* The offset to where the first object should be written
* The "stride" between consecutive objects

We were already capturing the offset information from Slang reflection data, and the same was true for the part of the offset that pertains to "pending" ordinary data. For other cases, though, we were manually incrementing the offset by values computed manually in `gfx` for Vulkan, and we were just skipping the offsetting step entirely for D3D11.

With this change we extract more complete stride information from the Slang reflection data and use that instead of ad hoc computations. Hopefully this data is correct, and if it isn't we can consider whether it needs fixing at the Slang reflection level rather than in `gfx`.

This change doesn't apply to the OpenGL back-end (which hasn't been updated to match the new approach to static specialization at all) or to the D3D12 back end (which has been updated a bit, but probably needs other cleanup work to be done first to bring it in line).

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -1968,6 +1968,7 @@ extern "C"
     SLANG_API SlangReflectionType* spReflectionTypeLayout_GetType(SlangReflectionTypeLayout* type);
     SLANG_API SlangTypeKind spReflectionTypeLayout_getKind(SlangReflectionTypeLayout* type);
     SLANG_API size_t spReflectionTypeLayout_GetSize(SlangReflectionTypeLayout* type, SlangParameterCategory category);
+    SLANG_API size_t spReflectionTypeLayout_GetStride(SlangReflectionTypeLayout* type, SlangParameterCategory category);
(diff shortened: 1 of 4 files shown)


tfoleyNV:
Note: There's more cleanup that can be done on top of this, since a lot of the functions that were used for the old offset logic are now no longer needed, and that means we might be able to skip computing the values that drive them during creation of shader object layouts.

csyonghe (approved):
I assume the next step is to make sure this actually works by adding some tests?

PR #1834, opened 2021-05-04
CUDA half comparison support

jsmall-nvidia:
This is on top of #1833

CUDA does have support for comparison of half and half2 types in cuda_fp16.hpp file. Unfortunately the problem is that these functions return true if *both* elements of the half2 compare succeeds. We need to have the comparison return something like a vector<bool> such that the combination can be tested with `any`, `all`.

This means we can't use any of the half2 comparison operators. We can replace the half2 operators by defining `__CUDA_NO_HALF2_OPERATORS__`. This means we have to move the implementations that were in the cuda_fp16.hpp into the prelude. 

Whilst it might not be the most efficient, we can use the C++ code generation to create the appropriate comparison functions, as we can compare `__half`. The problem now is that the names of the elements changed depending on if it's a half2,3,4. To work around this we add a method that returns the element names for a vector, and then vary this depending on half types.


prelude/slang-cuda-prelude.h
```diff
@@ -5,6 +5,9 @@
 // are passed down. 
 
 #ifdef SLANG_CUDA_ENABLE_HALF
+// We don't want half2 operators, because it will implement comparison operators that return a bool(!). We want to generate
+// those functions. Doing so means that we will have to define all the other half2 operators.
+#   define __CUDA_NO_HALF2_OPERATORS__
```
tfoleyNV, line 10: It is possible that we can/should emit all the vector/matrix comparison operators for C++ and CUDA as function calls rather than directly as operators.

tfoleyNV (approved):

prelude/slang-cuda-prelude.h
jsmall-nvidia: We could but would end up being more complicated - as we'd need to make the 'generators' able to know about CUDA specifics. 

What would be *nice* would be more flexible/powerful ways to build types out of 'builtin' types targets have. Ideally I could implement half and half2 and then everything else could be inferred (with some other info perhaps) from that.

PR #1835, opened 2021-05-04
Add support for returning structures that contain opaque types

tfoleyNV:
Introduction
============

Several of our target platforms share a concept of "opaque" types, including resources (`Texture2D`) and samplers (`SamplerState`), which are restricted in how they can be used. GLSL and SPIR-V place very severe restrictions, in that opaque types cannot be used for the type of:

* (mutable) local variables
* (mutable) global variables
* structure fields
* Function result/return
* `out` or `inout` parameters

The HLSL language allows all of these cases, but with the practical caveat that the compiler front-end must be able to statically analyze how opaque types have been used and "optimize away" all of the above cases. For example, it is legal to have a local variable of an opaque type, but at any point where the variable gets used it must be statically known which top-level shader parameter the variable refers to.

Existing Work
=============

In the Slang compiler we need to implement our own passes to detect these "illegal" uses of opaque types and legalize them. The work is basically broken into two distinct steps:

* The existing `legalizeResourceTypes()` pass detects illegal types (e.g., a `struct` that has a field of type `Texture2D`) and replaces them with legal types, sometimes by splitting apart declarations (e.g., a parameter using such a `struct` type gets split into multiple parameters). At a high level, we can think of this as "exposing" opaque types so that they are not hidden inside of nested structures.

* Next, the `specializeResourceOutputs()` pass detects calls to functions that output opaque types (whether by the function return value of `out` / `inout` parameters). The pass analyzes the body of such functions, and tries to isolate the logic that determines their resource-type outputs and hoise that logic into call sites (so that the opaque-type outputs can then be eliminated).

This Change
===========

One important missing case was that the type legalization step was incapable of legalizing types that appear in the result/return type of functions. The existing logic would simply diagnose an internal/unimplemented error if it ecountered a non-simple type in the return position.

At a high-level, supporting this case seems simple enough. Given a function signature like:

```
struct Things { int a; Texture2D b; }

Things myFunc(int x) { ... }
```

we want to split the result type into an "ordinary" result type and then `out` parameters for any opaque-type fields:

```
struct Things_Legal { int a; }

Things_Legal myFunc(int x, out Texture2D result_b) { ... };
```

Similarly, at a call site to a function like this:

```
Things t = myFunc(99);
```

we split the function result into ordinary and opaque-type parts, and pass the latter as `out` parameters:

```
Texture2D t_b;
Things_Legal t = myFunc(99, /*out*/ t_b);
```

The main place where things get tricky is when dealing with `return` sites within the body of a function that needs legalization:

```
Things myFunc(int x) {
    ...
    Things things = ...;
    ...
    return things;
}
```

In theory the answer is simple: a `return` translates into writes to the `out` parameters for any opaque-type data, followed by a return of the ordinary-type part:

```
Things_Legal myFunc(int x, out Texture2D result_b) {
    ...
    Things_Legal things = ...;
    Texture2D things_b = ...;
    ...
    result_b = things_b;
    return things;
}
```

The sticking point here is that this step requires tracking data between the legalization of the parameter list for `myFunc` and legalization of the `return`s in its body, so that we can identify the `result_b` parameter to be able to write to it. The existing type legalization pass was not built with the idea that such communication is commonly needed; it assumes that each instruction can be legalized in isolation, so long as dependencies are respected.

This change adds logic such that the `legalizeFunc()` step sets up a data structure that it used to represent information about how a function (and its parameter list) got legalized, so that the logic for a `return` can make use of that legalized information. Right now the information we track consists of just the list of parameters that were introduced to represent a return/result type.

Testing
=======

In order to confirm what features do/don't work, I added a set of tests that cover a cross-product of opaque type use cases:

* The opaque type can be used in the function result type, an `out` parameter, or an `inout` parameter
* The opaque type can be used "directly" or nested inside a `struct`.

These tests are helpful to make sure we handle the most important cases, but it is worth noting that the coverage is still lacking in that we do not sufficiently test all the options for what the function body might do. An opaque-type function result could be derived from many different sources:

* It could be a global shader parameter
* It could be an `in` or `inout` parameter of the function itself
* It could be wrapped up in one or more structure types
* It could be wrapped up in one or more array types (such that the output of specialization needs to pass around array indices)
* It could involve use of the type as a local variable (including passing it into other functions with result/`out`/`inout` outputs of opaque types)

This change makes it so that we can handle the simplest cases involving result/return types with a wrapper `struct`, and adds test cases that confirm we handle several other cases for `out` and `inout` parameters. Gaining confidence that we cover all the cases that arise in practical shaders will require more work over following changes.


source/slang/slang-ir-legalize-types.cpp
```diff
@@ -194,90 +194,515 @@ static LegalVal legalizeOperand(
...
+        /// The context for legalization
+    IRTypeLegalizationContext* m_context = nullptr;
+
+        /// The `call` instruction we are legalizing
+    IRCall* m_call = nullptr;
+
+        /// The legalized arguments for the call
+    List<IRInst*> m_args;
+
+        /// Add a logical argument to the call (which may map to zero or mmore actual arguments)
```
csyonghe, line 216: mmore

csyonghe (approved):
Looks good to me. I don't see why it wouldn't work, but will be good if we cover generic functions in the tests.

PR #1836, opened 2021-05-04
Half texture support

jsmall-nvidia:
* Added ability to get info about formats from gfx
* Add float16 format types
* Added float16 support to CPU texture type
* Added float16 support to the test infrastructure TextureData
* Added support for different amount of channels


slang-gfx.h
```diff
@@ -146,6 +172,12 @@ enum class Format
     CountOf,
 };
 
+struct FormatInfo
+{
+    uint8_t channelCount;       ///< The amount of channels in the format. Only set if the channelType is set 
+    uint8_t channelType;        ///< One of SlangScalarType None if type isn't made up of elements of type.
```
tfoleyNV, line 178: Why isn't this declared as a `SlangScalarType` if that's what it represents? It seems like packing this all into 16 bits might be over-optimizing?

slang-gfx.h
```diff
@@ -1342,6 +1374,9 @@ extern "C"
     /// Gets the size in bytes of a Format type. Returns 0 if a size is not defined/invalid
     SLANG_GFX_API size_t SLANG_MCALL gfxGetFormatSize(Format format);
 
+    /// Gets information about the format 
+    SLANG_GFX_API FormatInfo gfxGetFormatInfo(Format format);
+
```
tfoleyNV, line 1379: This change isn't introducing the issue, but I am worried that we are adding more and more global functions to the API surface area. It seems like we really need an umbrella COM interface to represent an "instance" of the `gfx` library or whatever, so that we can get back to the point where the user only needs to load a single `extern "C"` function from the shared library to get started.

tfoleyNV (approved):

slang-gfx.h
jsmall-nvidia: I agree there is a strong argument to having a COM interface that provides the global facilities. It does raise a few questions - should there global function just return the interface (ie the user doesn't have to worry about life times), or return a reference counted COM interface. 
For convenience I think you probably want to just return the interface, but that does mean you lose something in that instance being able to hold global state, and be able to control the lifetime of that state outside of shared libraries. 
A helper library outside of the COM interface, that's liked with your binary could handle this lifetime, such that you get the best of both worlds.

slang-gfx.h
jsmall-nvidia: If we had a C++ sized type then this would be a no brainer - it could be a byte and type safe. That it is just a typedef to an int without any type safety makes the above choice less problematic. 
That said there is some small readability/lookup benefit perhaps in having it's type be SlangScalarType as opposed to reading in the comment.

tfoleyNV: Well, even if we used explicitly-sized `enum` types in the `slang.h` API, I suspect they'd all be 32 bits in size. In fact, many of the C `enum` declarations use a seaprate `typedef` already to make sure they are explicitly-sized (and all those use 32 bits IIRC).

I think I'm just far more concerned with the stability of binary interfaces than with making things really tightly bit-packed. I'm already nervous about stuff like our texture "shape" and "access" bitfield where we might be in danger of running out of bits. I think my mind is in "better safe than sorry" mode.

slang-gfx.h
tfoleyNV: I personally like the idea of that the global C function vends a proxy global/singleton object that never gets released, but that clients are still expected to reference-count it correctly (since that lets us switch up the implementation later if we like).

The one annoying design problem is that once we start wanting to get into the API-specific ways that device/adapter enumeration work, etc. we are going to need a multi-tier creation step:

1. The application gets a global "instance" of the `gfx` library (basically just a COM object representing the `gfx` implementation in a particular DLL)
2. The application uses the instance to create an API-specific "factory" or "session" object, which represents a connection to a particular one of the back-end implementations (VK, D3D12, etc.)
3. The application uses that factory to create their swap chain(s), device(s), etc.

It might seem like a lot of steps to users, and they might prefer that we just let them skip over as many of these steps as possible for common cases.

Issue #1847, opened 2021-05-17
Downstream Option Mechanism


jsmall-nvidia:
Downstream Option Mechanism
====================

During a Slang compilation work may be performed by multiple other stages including downstream compilers and linkers. It isn't possible in general or perhaps even desirable to provide Slang equivalents of every option available at every stage of compilation. It is useful to be able to set options specific to a particular complation stage - to alter code generation, linkage and other options. 

## Prior art

The [gcc toolchain](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html) has a somewhat analagous issue around passing options to the downstream linker. The solution used there is 

```
-Xlinker option
```

The mechanism requires -Xlinker to be used in front of every option. If an option and value are separate it requires two -Xlinker options. 

```
-Xlinker someOption -Xlinker someOptionsValue
```

It suggests using the `=` style setting which makes the setting a parameter somewhat less painful.

```
-Xlinker -someOption=someOptionValue
```

This is an improvement. Unfortunately Slang command line processing does not currently support = style.

Gcc also has the `-Wl` option which allows a comma-separated list of tokens to be passed to the linked.  

> Pass option as an option to the linker. If option contains commas, it is split into multiple options at the commas. You can use this syntax to pass an argument to the option. For example, -Wl,-Map,output.map passes -Map output.map to the linker. When using the GNU linker, you can also get the same effect with -Wl,-Map=output.map. 

The option presumably exists to work around the verbosity of having to write every option with the preceeding `-Xlinker`. A nice property of the `-Xlinker` style is that it's pretty clear what's going on. It also allows for easy extension around other kinds of downstream tools by having different names other than `linker`.  

(Clang)[https://clang.llvm.org/docs/ClangCommandLineReference.html#linker-flags] is similar but we also have `--for-linker`.

```
-X
-Xlinker <arg>, --for-linker <arg>, --for-linker=<arg>
```

If there are multiple options using this mechanism could get somewhat monotonous. You could imagine an extension such that all of the parameters up to the next '-X' are taken to go to the target so the above option would become

```
-Xlinker -someOption someOptionValue
```

That an empty `-X` could be used to indicate return to normal command line processing. This is somewhat appealing except the behavior deviates from the -X in gcc/clang, so the additional behavior needs to be a slightly different mechanism. 

Discussion
==========

In the Slang scenario it could be argued that there were more 'stages' than just linking.

* Source generation 
* Source compilation (typically but not necessarily) via a downstream compiler
[...]

PR #1848, opened 2021-05-18
SourceLoc use in command line processing

jsmall-nvidia:
* Added CommandLineArgs/Reader to add SourceLoc tracking and simplify parsing
* Parse slang command line options use CommandLineArgs/Reader
* Added ability to limit diagnostic source line length
* Made DiagnosticSink able to have a 'ParentSink' were diagnostics can be routed to 
* Moved DiagnosticSink methods in slang.cpp into slang-diagnostic-sink.cpp. To do so required moving the internal error diagnostic into MiscDiagnostics, and have a slightly more vague error. 

This is the start of some more broad work to control where command line args are channeled (discussed in  #1847). Whilst not necessary strictly speaking, an idea for handling the processing of -X options is to have a utility that will strip them out and standardizing 'CommandLineArgs' simplifies that. 

It might seem CommandLine and CommandLineArgs are very similar - which they are in some respects, but CommandLine is really for launching via command line through OS shell (and through ProcessUtil), CommandLineArgs/Reader is about parsing and tracking.

These idea for these changes seemed like the logical extension of the use of DiagnosticSink in command line parsing. The implementation required some, somewhat odd additions, and DiagnosticSink doesn't seem like quite the obvious fit that it did when this was started. That said this still seems like an improvement.


source/compiler-core/slang-command-line-args.h
```diff
@@ -0,0 +1,95 @@
...
+
+    List<Arg> m_args;                       ///< The args
+    SourceManager* m_sourceManager;         ///< The source manager and associated diagnostics sink
+    SourceView* m_sourceView;        ///< contains the command line as source
+};
+
+struct CommandLineReader
+{
+        /// Peek the current location
+    SourceLoc peekLoc() const { return m_index < m_args->getArgCount() ? (*m_args)[m_index].loc : SourceLoc(); }
```
tfoleyNV, line 50: Ideally we'd want to be able to indicate a location "one past the end" similar to how the EOF token lets us have error messages that point to the end of the argument list.

Assuming that the synthesized command-line string is registered as a source file, we should be able to produce a location that represents its last possible location.

source/compiler-core/slang-diagnostic-sink.h
```diff
@@ -138,8 +138,9 @@ class DiagnosticSink
     {
         enum Enum: Flags
         {
-            VerbosePath = 0x1,                  ///< Will display a more verbose path (if available) - such as a canonical or absolute path
-            SourceLocationLine = 0x2,           ///< If set will display the location line if source is available
+            VerbosePath         = 0x1,           ///< Will display a more verbose path (if available) - such as a canonical or absolute path
+            SourceLocationLine  = 0x2,           ///< If set will display the location line if source is available
+            HumaneLoc           = 0x4,           ///< If set will display humane locs (filename/line number) information
```
tfoleyNV, line 143: It seems like the `HumaneLoc` part doesn't specifically deal with humane locations, but rather with whether source locations are output *at all*.

(We probably need to go through and clarify all our terminology around source locations so that the compiler is always consistent about what things like "humane" locations even mean, but that's clearly out of scope here)

source/compiler-core/slang-diagnostic-sink.h
```diff
@@ -238,10 +249,17 @@ class DiagnosticSink
 
 protected:
     void diagnoseImpl(SourceLoc const& pos, DiagnosticInfo const& info, int argCount, DiagnosticArg const* const* args);
+    void diagnoseImpl(DiagnosticInfo const& info, const UnownedStringSlice& formattedMessage);
+
+        /// If set all diagnostics (as formatted by *this* sink, will be routed to the parent).
```
tfoleyNV, line 254: Seems like the parenthetical here should end earlier:
```
/// If set all diagnostics (as formatted by *this* sink), will be routed to the parent.
```

tfoleyNV: It's not something we should do write away, but this almost suggests that we'd eventually want some distinction between:

* The interface that clients talk to when they want to emit diagnostics, which handles formatting policy
* The interface that handles actually writing diagnostics (already formatted) to a buffer/console/whatever

source/slang/slang-options.cpp
```diff
@@ -425,11 +403,58 @@ struct OptionsParser
...
+        // source lines.
+        //
+        // The *problem* is that we still need to communicate to the requestSink in some suitable way.
+        //
+        // 1) We could have some kind of scoping mechanism (and only one sink)
+        // 2) We could have a 'parent' diagnostic sink, that if we set we route output too
+        // 3) We use something like the ISlangWriter to always be the thing output too (this has problems because
+        // some code assumes the diagnostics are accessible as a string)
+        //
+        // The solution used here is to have DiagnosticsSink have a 'parent' that also gets diagnostics reported to.
```
tfoleyNV, line 433: This comment seems to reinforce the idea that we need/want to tease apart some of the myriad things `DiagnosticSink` does into distinct types so that we can more easily swap in custom "adapters" for stuff like this.

source/slang/slang-options.cpp
```diff
@@ -425,11 +403,58 @@ struct OptionsParser
...
+        {
+            parseSink.setFlags(requestSink->getFlags());
+            // Allow HumaneLoc - it won't display much for command line parsing - just (1):
+            // Leaving allows for diagnostics to be compatible with other Slang diagnostic parsing.
+            //parseSink.resetFlag(DiagnosticSink::Flag::HumaneLoc);
+            parseSink.setFlag(DiagnosticSink::Flag::SourceLocationLine);
+        }
+
+        // We don't know how big the terminal is.. let's guess 120 for now 
+        parseSink.setSourceLineMaxLength(120);
```
tfoleyNV, line 446: Should the sinks we use for ordinary errors on code also use this setting?

source/slang/slang-options.cpp
```diff
@@ -440,59 +465,57 @@ struct OptionsParser
...
                 {
                     flags |= SLANG_COMPILE_FLAG_NO_MANGLING;
                 }
-                else if (argStr == "-load-stdlib")
+                else if (argValue == "-load-stdlib")
                 {
-                    String fileName;
-                    SLANG_RETURN_ON_FAIL(tryReadCommandLineArgument(sink, arg, &argCursor, argEnd, fileName));
+                    CommandLineArg fileName;
+                    SLANG_RETURN_ON_FAIL(reader.expectArg(fileName));
```
tfoleyNV, line 482: It's not a new change here, but we might want to consider what the error-reporting policy for the command line should be. If we liken this to the actual Slang parser (and semantic checking, etc.), there's two principles:

1. Report as many errors as possible in a single pass
2. Do not report "cascading" errors that are actually the byproduct of earlier errors

There's obviously a tension between those two, but right now the use of `SLANG_RETURN_ON_FAIL` during parsing of command-line arguments prioritizes (2) over (1). It almost seems like these cases should `continue` the loop that is processing top-level arguments.

tfoleyNV (approved):

PR #1850, opened 2021-05-19
Downstream option handling

jsmall-nvidia:
On top of #1848 

* Added DownstreamArgs to hold args for downstream targets
* Strip downstream args in slang-option parsing
* Added compilerSpecificArgs to DownstreamCompiler::CompileOptions
* Pass extracted args to downstream compiler

Tested on dxc - although this PR doesn't have a test included. 

This adds to the -Xname... -X. idea talked about in the original issue, and allows scoping of options. Allowing scoping provides an easy mechanism to pass down other -X args (for example to gcc), without any special 'escaping'.


source/compiler-core/slang-misc-diagnostic-defs.h
```diff
@@ -24,7 +24,11 @@
 
 DIAGNOSTIC(-1, Note, seeTokenPasteLocation, "see token pasted location")
 
-DIAGNOSTIC(21, Error, expectedArgumentForOption, "expected an argument for command-line option '$0'")
+DIAGNOSTIC(100000, Error, downstreamNameNotKnown, "downstream tool name not known")
```
tfoleyNV, line 27: Could this be something like `unknown downstream tool '$0'` so that it includes which tool is unknown?

source/compiler-core/slang-misc-diagnostic-defs.h
```diff
@@ -24,7 +24,11 @@
...
+DIAGNOSTIC(100001, Error, expectedArgumentForOption, "expected an argument for command-line option '$0'")
+DIAGNOSTIC(100002, Error, unbalancedDownstreamArguments, "unbalanced downstream arguments")
+DIAGNOSTIC(100003, Error, closeOfUnopenDownstreamArgs, "close of an unopen downstream argument scope")
```
tfoleyNV, line 30: The `close of an unopen` case where is also a little confusing. I assume this is something like `attempting to close a scope for downstream-tool arguments when no matching scope is open`.

source/slang/slang-compiler.cpp
```diff
@@ -964,6 +964,23 @@ namespace Slang
         typedef DownstreamCompiler::CompileOptions CompileOptions;
         CompileOptions options;
 
+        /* Let's set the compiler specific options 
+
+          We can only do this if the endToEndReq is set. */
```
tfoleyNV, line 969: This might be worth fixing in the long run, since downstream compiler arguments seem like the kind of thing that could arguably belong on a `TargetReq`, and thus could be used by any compile including API-based ones.

tfoleyNV (approved):

PR #1851, opened 2021-05-20
[gfx] Support StructuredBuffer<IInterface>.

csyonghe:
This change adds support for`StructuredBuffer<IInterface` values in gfx layer's shader-object implementation. This allows us to reenable the test cases that were previously disabled due to lack of this feature.

The main idea of this change is to allow user to create a shader object from a `StructuredBuffer<IInterface>` type, and then fill in the elements into the buffer with the shader object's `setObject` call.

The main change to Slang is to make sure `StructuredBuffer` fields are also reported as sub-objects in the sub-object range list.

To allow the logic of `setObject` and `getObject` and `collectSpecializationArgs` to be shared across different targets, the shader object implementation is refactored in this change so that these function are moved to `renderer-shared.h` under a `ShaderObjectImplBase<TShaderObject, TShaderObjectLayout, TShaderObjectData>` class, where `TShaderObject` is the derived shader object type, `TShaderObjectLayout` is the derived shader object layout type, and `TShaderObjectData` is a class that implements the storage of the shader object's buffer data.

In addition, this change fixes a bunch of bugs a long the way:
1. Fixed GLSL output that sometimes `#line xxx` directive is appearing before `#version`, causing glslang to fail.
2. Fixed render-test that incorrectly sets a SRV resource view into a shader object for a field that is `RWTexture` type.
3. Fixed render-test to pass consistent default matrix layout settings to `gfx`'s Slang context to make sure the reflection info agrees with the actual code being generated.
4. Fixed a constant buffer can sometimes be missing from binding range reflection for CUDA.


slang-gfx.h
```diff
@@ -505,6 +505,11 @@ struct ShaderOffset
     SlangInt bindingArrayIndex = 0;
 };
 
+enum class ShaderObjectContainerType
+{
+    None, Array, StructuredBuffer
+};
+
```
tfoleyNV, line 512: This has already been discussed separately, but I want to start by noting that a lot of the feedback here doesn't need to be addressed immediately or in the near term. Mostly I'm thinking about where we want the design to land in the long run.

My ideal would be that we don't need this distinction to be part of the API for creating shader objects, for two reasons:

* The user should (eventually) be able to take a reflection type `T` and form the type `T[]` or `StructuredBuffer<T>` using the reflection API, at which point they can directly construct the type they want to form an object for

* Ideally, the user should only need the `Array` case here, and the runtime would allow an object allocated based on an array type to be bound as a structured buffer

slang-gfx.h
```diff
@@ -516,6 +521,7 @@ class IShaderObject : public ISlangUnknown
     }
 
     virtual SLANG_NO_THROW slang::TypeLayoutReflection* SLANG_MCALL getElementTypeLayout() = 0;
+    virtual SLANG_NO_THROW ShaderObjectContainerType SLANG_MCALL getContainerType() = 0;
```
tfoleyNV, line 524: This *definitely* feels like something that shouldn't be public API.

slang.h
```diff
@@ -3950,11 +3955,23 @@ namespace slang
...
+            ISlangBlob** outDiagnostics = nullptr) = 0;
+
+            /** Return a `TypeReflection` that represents the `__Dynamic` type.
+                This type can be used as a specialization argument to indicate using
+                dynamic dispatch.
+            */
+        virtual SLANG_NO_THROW TypeReflection* SLANG_MCALL getDynamicType() = 0;
 
             /** Get the mangled name for a type RTTI object.
             */
```
tfoleyNV, line 3977: A minor note here is that when possible we should add any new `virtual` methods to the *end* of an interface declaration, so that it preserves binary compatibility with client code compiled against the old declaration.

slang.h
```diff
@@ -3950,11 +3955,23 @@ namespace slang
...
 
-        virtual SLANG_NO_THROW TypeLayoutReflection* SLANG_MCALL getParameterBlockLayout(
+            /** Get a container type from `elementType`. For example, given type `T`, returns
+                a type that represents `StructuredBuffer<T>`.
+
+                @param `elementType`: the element type to wrap around.
+                @param `containerType`: the type of the container to wrap `elementType` in.
+                @param `outDiagnostics`: a blob to receive diagnostic messages.
+            */
+        virtual SLANG_NO_THROW TypeReflection* SLANG_MCALL getContainerType(
```
tfoleyNV, line 3965: I think of this operation as a short-term necessary evil. Ideally the user should be able to do something like:

```
auto structuredBufferGeneric = session->getStandardLibraryDecl("StructuredBuffer");
auto structuredBufferOfMyType = structuredBufferGeneric->specialize({myType});
```

I get why that's not easy to support right now.

One smaller question would be whether `getContainerType` could use a string name for the container type instead of an enum:

```
auto structuredBufferOfMyType = ...->getContainerType("StructuredBuffer", myType);
```

Technically all of the types in that `enum` are named types in the Slang standard library...

source/slang/slang-reflection-api.cpp
```diff
@@ -1485,7 +1489,7 @@ namespace Slang
                 Index bindingRangeIndex = m_extendedInfo->m_bindingRanges.getCount();
                 SlangBindingType bindingType = SLANG_BINDING_TYPE_CONSTANT_BUFFER;
                 Index spaceOffset = -1;
-                bool usesIndirectAllocation = false;
+                bool shouldAllocDescriptorSet = true;
```
tfoleyNV, line 1492: The terminology here is a bit confusing (both in the original code and the new).

We probably need to be super clear about what is going on here, because there's really three distinct cases we are trying to juggle here:

1. The constant-buffer-like case. That includes `ConstantBuffer<X>` for D3D12/VK, but also includes `ParameterBlock<X>` for D3D11/GL. Any opaque types used by `X` "leak" into the resource usage of the container, but any uniform/ordinary data does not.

2. The descriptor-set-like case. This is obviously for `PB<X>` on D3D12/VK, and pretty much only those.

3. The CUDA/CPU "it's just memory" case. This covers both `PB<X>` and `CB<X>` for both CUDA and CPU, where they are basically just `X*`.

The old flag named `usesIndirectAllocation` was meant to detect/capture cases (2) (via `RegisterSpace`) and (3) (via `Uniform`).

The new flag named `shouldAllocDescriptorSet` seems to only capture case (2). AFAICT this seems to greatly change how binding ranges for CBs/PBs are reflected for CPU/CUDA, and I'm having trouble seeing how it could possibly work.

source/slang/slang-reflection-api.cpp
```diff
@@ -1888,7 +1894,19 @@ namespace Slang
...
+
                 m_extendedInfo->m_bindingRanges.add(bindingRange);
+
+                // For `StructuredBuffer` fields, we also make sure to report it as a sub-object range.
+                if (auto structuredBufferTypeLayout = as<StructuredBufferTypeLayout>(typeLayout))
+                {
+                    TypeLayout::ExtendedInfo::SubObjectRangeInfo subObjectRange;
+                    subObjectRange.bindingRangeIndex = bindingRangeIndex;
+                    subObjectRange.offsetVarLayout = createOffsetVarLayout(typeLayout, path);
+                    subObjectRange.spaceOffset = 0;
```
tfoleyNV, line 1907: Is a `spaceOffset` of zero necessarily correct here?

source/slang/slang.cpp
```diff
@@ -895,35 +895,65 @@ SLANG_NO_THROW slang::TypeLayoutReflection* SLANG_MCALL Linkage::getTypeLayout(
...
+            }
+            break;
+        case slang::ContainerType::StructuredBuffer:
+            {
+                HLSLStructuredBufferType* sbType =
+                    getASTBuilder()->create<HLSLStructuredBufferType>();
+                sbType->elementType = type;
+                containerTypeReflection = sbType;
+            }
+            break;
```
tfoleyNV, line 932: I'm concerned about how this kind of thing works in practice. All the types here are derived from `DeclRefType`, but nothing here is setting up a valid `DeclRef`. That means that the types that come out of this function will be missing key pieces (including generic specialization arguments) that would be present on equivalent types coming from the parsing and semantic checking logic of the full compiler.

If you look at a similar operation like `ASTBuilder::getVectorType`, you'll find that it goes to some lengths to ensure that when it constructs a type for `vector<T,N>` it does so by constructing a full decl-ref to the declaration of `Vector` in the Slang stdlib, with proper generic arguments.

Obviously if this approach works for now, we can run with it.

tfoleyNV (approved):

source/slang/slang-reflection-api.cpp
csyonghe: I think I intend to run the `if (shouldAllocDescriptorSet)` case for case 3 here, because we need the step to reflect the CB<X> itself as a binding range. Inside the `if` block, we already have further sub `if` statements to not do anything for case 3, so this will not change the current behavior except that `CB<X>` will appear as a binding range in the its containing descriptor set.

source/slang/slang-reflection-api.cpp
csyonghe: I think this is just a copy paste from existing code, so I am not sure if it is needed or correct...

source/slang/slang.cpp
csyonghe: Right, currently no where in the compiler is using that decl ref so it runs correctly. If we run into any issues we should make sure to fix this.

PR #1852, opened 2021-05-20
Improvements in -X support

jsmall-nvidia:
On top of #1850 

This work was in order to try and have an 'end to end' dxc arg test. 

* Put the TokenReader/TokenType etc in Slang::Misc namespace to avoid clash with Slang::Token/TokenType
* Use CommandLineReader to parse options for render test
* Use -X extraction in renderer test (supports backwards compatible -xslang and -compile-arg options)
* Moved DownstreamArgs to Linkage so they can be seen when not using EndToEndRequest
* Added an 'end to end' dxc test that passes a DXC specific arg through render-test to slang compilation to DXC


docs/command-line-slangc.md
```diff
@@ -149,6 +149,68 @@ For completeness, here are the options that `slangc` currently accepts:
...
+```
+-Xlinker option
+```
+
+When used, option is not interpreted by GCC, but is passed to the linker once compilation is complete. Slang extends this idea in several ways. First there are many more 'downstream' stages available to Slang. These different stages are known as `SlangPassThrough` types in the API and have the following names
+
+* `fxc` - FXC HLSL compiler
+* `dxc` - DXC HLSL compiler
+* `glslang` - GLSLANG GLSL compiler
+* `visualstudio` - Visual Studio C/C++ compiler
```
tfoleyNV, line 169: I think some people would want to nit-pick the naming here, since Visual Studio is the IDE while the `cl.exe` compiler (which is developed by a distinct team) is often just "Visual C/C++." I'm not sure what is the ideal here, and we can probably keep `visualstudio` as an alias even if we add something that would be more "correct" according to the team that developers the compiler in question.

(This is one of those things like how there is no "DirectX 11" only "Direct3D 11" where I waste way too many brain cycles on trying to be accurate to the brand names of all these APIs and tools)

docs/command-line-slangc.md
```diff
@@ -149,6 +149,68 @@ For completeness, here are the options that `slangc` currently accepts:
...
+* `clang` - Clang C/C++ compiler
+* `gcc` - GCC C/C++ compiler
+* `genericcpp` - A generic C++ compiler (can be any one of visual studio, clang or gcc depending on system and availability)
```
tfoleyNV, line 172: Can we allow `cpp` or `cxx` here as a shorthand?

That said, it would be tricky for somebody to actually form options that any of those compilers would accept.

Also: can the `genericcpp` args go to NVRTC as well? It seems like it accepts many of the same things, right?

tfoleyNV (approved):

PR #1857, opened 2021-05-25
Rework shader object specialization control interface.

csyonghe:
This PR changes the shader object's `setSpecializationArgs` implementation to operate on a sub-object range instead of on the parent shader object itself. This design preserves better modularity.

The `render_test` comment syntax is also updated so you user can write:
```
//TEST_INPUT: set gCb1 = dynamic new StructuredBuffer<IInterface>{new MyImpl{1}}
RWStructuredBuffer<IInterface> gCb1;
```
or
```
//TEST_INPUT: set gCb1 = specialize(__Dynamic) new StructuredBuffer<IInterface>{new MyImpl{1}}
RWStructuredBuffer<IInterface> gCb1;
```
This code sets the specialization arg on the `gCb1` object instead of on the `StructuredBuffer` objects.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -545,12 +545,13 @@ public:
     virtual SLANG_NO_THROW Result SLANG_MCALL setCombinedTextureSampler(
         ShaderOffset const& offset, IResourceView* textureView, ISamplerState* sampler) = 0;
 
-        /// Manually setting the specialization arguments for the shader object, overriding
-        /// the default arguments computed from the sub-objects.
+        /// Manually overrides the specialization argument for the sub-object binding at `offset`.
         /// Specialization arguments are passed to the shader compiler to specialize the type
         /// of interface-typed shader parameters.
-    virtual SLANG_NO_THROW Result SLANG_MCALL
-        setSpecializationArgs(const slang::SpecializationArg* args, uint32_t count) = 0;
+    virtual SLANG_NO_THROW Result SLANG_MCALL setSpecializationArgs(
+        ShaderOffset const& offset,
+        const slang::SpecializationArg* args,
+        uint32_t count) = 0;
 };
 #define SLANG_UUID_IShaderObject                                                       \
     {                                                                                 \
--- a/tools/gfx/renderer-shared.h
+++ b/tools/gfx/renderer-shared.h
@@ -473,10 +473,10 @@ class ShaderObjectBaseImpl : public ShaderObjectBase
 protected:
     TShaderObjectData m_data;
     Slang::List<Slang::RefPtr<TShaderObjectImpl>> m_objects;
+    Slang::List<Slang::RefPtr<ExtendedShaderObjectTypeListObject>> m_userProvidedSpecializationArgs;
 
     // Specialization args for a StructuredBuffer object.
     ExtendedShaderObjectTypeList m_structuredBufferSpecializationArgs;
-    Slang::RefPtr<ExtendedShaderObjectTypeListObject> m_userProvidedSpecializationArgs;
 
 public:
     TShaderObjectLayoutImpl* getLayout()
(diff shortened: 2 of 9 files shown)


slang-gfx.h
```diff
@@ -545,12 +545,13 @@ class IShaderObject : public ISlangUnknown
...
-        /// the default arguments computed from the sub-objects.
+        /// Manually overrides the specialization argument for the sub-object binding at `offset`.
         /// Specialization arguments are passed to the shader compiler to specialize the type
         /// of interface-typed shader parameters.
-    virtual SLANG_NO_THROW Result SLANG_MCALL
-        setSpecializationArgs(const slang::SpecializationArg* args, uint32_t count) = 0;
+    virtual SLANG_NO_THROW Result SLANG_MCALL setSpecializationArgs(
+        ShaderOffset const& offset,
+        const slang::SpecializationArg* args,
+        uint32_t count) = 0;
```
tfoleyNV, line 554: I assume the `count` is required because of the structured-buffer case, where the structure type could have multiple interface-type fields?

Having to handle that case here is a little unfortunate when it won't come up in the 99% case, but probably not the end of the world.

tools/gfx/renderer-shared.h
```diff
@@ -718,20 +716,57 @@ class ShaderObjectBaseImpl : public ShaderObjectBase
...
+        ShaderOffset const& offset,
+        const slang::SpecializationArg* args,
+        uint32_t count) override
     {
-        if (m_userProvidedSpecializationArgs)
+        auto layout = getLayout();
+
+        // If the shader object is a container, delegate the processing to
+        // `setSpecializationArgsForContainerElements`.
+        if (layout->getContainerType() != ShaderObjectContainerType::None)
```
tfoleyNV, line 733: I'm still a little confused about how the property of being or not being a container works, so I'm mostly reading this as "what to do in the structured-buffer case."

tools/gfx/renderer-shared.h
```diff
@@ -718,20 +716,57 @@ class ShaderObjectBaseImpl : public ShaderObjectBase
...
+            return SLANG_E_INVALID_ARG;
+
+        auto bindingRangeIndex = offset.bindingRangeIndex;
+        auto bindingRange = layout->getBindingRange(bindingRangeIndex);
+        auto objectIndex = bindingRange.subObjectIndex + offset.bindingArrayIndex;
+        if (objectIndex >= m_userProvidedSpecializationArgs.getCount())
+            m_userProvidedSpecializationArgs.setCount(objectIndex + 1);
+        if (!m_userProvidedSpecializationArgs[objectIndex])
+        {
+            m_userProvidedSpecializationArgs[objectIndex] =
```
tfoleyNV, line 754: I'm not clear here why `m_userProvidedSpecializationArgs` is indexed by the `objectIndex`. It seems that for a sub-object range that uses an array:

```
struct MyParams
{
    ...
    IThing things[10];
}
```

there would only be a single specialization argument needed/wanted for the `things` array.

It almost seems like this is a lurking bug in other parts of the specialization logic: we may be acting as if each *object* in a sub-object range yields a distinct specialization parameter, while the reality is that the entire *range* will have a single specialization parameter (and thus a single argument as well).

It seems that in the long run it will be simplest if each sub-object range pre-computes a "flat" index into a single array of specialization arguments related to that range, much like each binding range stores a flat index like `bindingRange.subObjectIndex` here.

tfoleyNV (approved):

tools/gfx/renderer-shared.h
csyonghe: You are right, this is likely wrong for the array case. However I am not getting into array support yet, so I guess I can let it slip a little bit right now until I need to make it work for arrays.

tools/gfx/renderer-shared.h
csyonghe: Yeah, this is just for the structured-buffer case for now.

slang-gfx.h
csyonghe: I was thinking about the case where you set an object to a `ParameterBlock<X>` binding range, where X is a struct that contains more than one interface typed fields, then there could be more than one arguments. You can argue that the user should call `setSpecializationArgs` on the parameter block object's interface-typed binding range instead, but this interface is not doing harm to allow the user to set things directly at higher level.

tools/gfx/renderer-shared.h
csyonghe: I should also note that in the case of an array, this code branch will not be executed, and the code where your previous comment is will be used instead, which should be correct.

PR #1859, opened 2021-05-26
JSON Parser and Writer

jsmall-nvidia:
* Uses JSONLexer to parse into a JSONListener
* JSONWriter is an implementation of JSONListener which outputs as formatted JSON text
* Unit test checks parsing/writing
* Supports an 'Allman' and 'K&R' (like) styles in terms of formatting

The JSONListener will be used for transversing a (to be added) JSON DOM, so the DOM can use the JSONWriter implementation. This will require more methods to be added to the listener, as current interface is only desived for 'Lexeme' style - which assumes already quoted/in text format.

The DOM will be able to values as references to the source code, such an end to end conversion will not lose any precision.  

Allman-like 

```
{
    "Hello" : "Json", 
    "!" : 10, 
    "array" : 
    [
        1, 
        2, 
        3.0
    ]
}
```

K&R like

```
{
    "Hello" : "Json", 
    "!" : 10, 
    "array" : [
        1, 2, 3.0
    ]
}
```


source/compiler-core/slang-json-parser.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+/*
+https://www.json.org/json-en.html
+*/
+
+namespace Slang {
+
+SlangResult JSONParser::_parseObject()
+{
+    SLANG_RETURN_ON_FAIL(m_lexer->expect(JSONTokenType::LBrace));
```
tfoleyNV, line 14: Eventually we may want this code to support a "recovery" system closer to what the actual Slang parser does, as a QOI feature.

source/compiler-core/slang-json-parser.cpp
```diff
@@ -0,0 +1,384 @@
...
+        SLANG_RETURN_ON_FAIL(m_lexer->expect(JSONTokenType::StringLiteral, keyToken));
+        m_listener->addLexemeKey(m_lexer->getLexeme(keyToken));
+
+        SLANG_RETURN_ON_FAIL(m_lexer->expect(JSONTokenType::Colon));
+
+        SLANG_RETURN_ON_FAIL(_parseValue());
+        if (m_lexer->advanceIf(JSONTokenType::Comma))
+        {
+            continue;
+        }
```
tfoleyNV, line 36: This gets into the same slippery slope as adding comment support: I'd personally like it if our JSON parser could work robustly when objects or arrays include a trailing comma. We can make the JSON lexer/parser have some flag(s) to mark extensions as disabled and go into a pedantic mode where it diagnoses them as errors. For ordinary development, I imagine that trailing commas will be the most valuable feature after comments.

source/compiler-core/slang-json-parser.cpp
```diff
@@ -0,0 +1,384 @@
...
+        case JSONTokenType::Invalid:
+        {
+            // It's a lex error, so just fail
+            return SLANG_FAIL;
+        }
+    }
+}
+
+SlangResult JSONParser::parse(JSONLexer* lexer, SourceView* sourceView, JSONListener* listener, DiagnosticSink* sink)
```
tfoleyNV, line 109: A small design question is whether the decomposition into `JSONLexer` and `JSONParser` pays off, or if we could write a single combined lexer/parser (a so-called "scannerless" parser) that handles both tasks. This is akin to how many Lisp implementations minimize code size/complexity by combining lexing and parsing.

source/compiler-core/slang-json-parser.h
```diff
@@ -0,0 +1,188 @@
+// slang-json-parser.h
+#ifndef SLANG_JSON_PARSER_H
+#define SLANG_JSON_PARSER_H
+
+#include "slang-json-lexer.h"
+
+
+namespace Slang {
+
+class JSONListener
```
tfoleyNV, line 10: I'm on the fence about the "listener" sort of approach as opposed to just committing to always building some kind of AST.

This design is clearly a big win for building up the infrastructure in pieces, since having the `JSONWriter` provides a way to test that the parser is working even before implementing the AST.

source/compiler-core/slang-json-parser.h
```diff
@@ -0,0 +1,188 @@
...
+{
+public:
+        /// Start an object
+    virtual void startObject() = 0;
```
tfoleyNV, line 14: An interesting design question here is how to pass source location information through to the listener, so that we can build a JSON AST with location information and properly diagnose errors when mapping from JSON to/from user-defined types.

The simplicity of the current operations is nice, so it might be best if source-location information (and comments, if we want to go that route) could be passed through a side-band interface.

tfoleyNV (approved):
Looking good. This is very clean and self-explanatory stuff, and the design should be usable in a lot of contexts.

PR #1865, opened 2021-06-01
JSONBuilder

jsmall-nvidia:
On top of #1864 

* JSONBuilder implements JSONListener so as to produce a JSONValue (belonging to a JSONContainer)
* Widened JSONListener interface to be able to take values
* Added SourceLoc to JSONListener interface
* Expanded the unit test coverage


source/compiler-core/slang-json-value.cpp
```diff
@@ -808,4 +808,318 @@ bool JSONContainer::areEqual(const JSONValue& a, const JSONValue& b)
...
+        case Type::False:           return listener->addBoolValue(false, value.loc);
+        case Type::Null:            return listener->addNullValue(value.loc);
+
+        case Type::StringLexeme:    return listener->addLexemeValue(JSONTokenType::StringLiteral, getLexeme(value), value.loc);
+        case Type::IntegerLexeme:   return listener->addLexemeValue(JSONTokenType::IntegerLiteral, getLexeme(value), value.loc);
+        case Type::FloatLexeme:     return listener->addLexemeValue(JSONTokenType::FloatLiteral, getLexeme(value), value.loc);
+
+        case Type::IntegerValue:    return listener->addIntegerValue(value.intValue, value.loc); 
+        case Type::FloatValue:      return listener->addFloatValue(value.floatValue, value.loc);
+        case Type::StringValue:
```
tfoleyNV, line 827: It's a little unfortunate to need these distinct cases for what amounts to "processed" and "unprocessed" values of the various leaf types. What would be the down-side to having the `JSONValue` type only support the `IntegerValue`, etc. cases and not the `IntegerLexeme` etc. cases?

source/compiler-core/slang-json-parser.cpp
```diff
@@ -375,10 +414,45 @@ void JSONWriter::addLexemeValue(JSONTokenType type, const UnownedStringSlice& va
...
+    m_builder << value;
+    _postValue();
+}
+
+void JSONWriter::addFloatValue(double value, SourceLoc loc)
+{
+    _preValue(loc);
+    m_builder << value;
+    _postValue();
+}
```
tfoleyNV, line 433: Probably not important right away, but eventually this logic would need to decide how to handle non-finite values, since the JSON spec doesn't specify how they should be encoded. A typical approach is to emit strings like `"inf"`, `"-inf"`, and `"nan"` instead of numeric values.

This goes down an entire rabbit-hole if you need to produce/consume JSON that is compatible with other systems that might have their own conventions (is it `"nan"` or `"NaN"`?), and thus potentially becomes a configuration option on the `JSONWriter` ("here are the literal string values to use when representing inf, -inf, and NaNs").

(There's a whole separate rabbit hole for representing timestamps because there are actual specs/RFCs for that case, IIRC. Hopefully that's not something we need right away.)

tfoleyNV (approved):
Looks good. I might feel more comfortable if the in-memory data structure didn't have both the value and "lexeme" cases (or if it was able to hide those as an implementation detail), but its not a huge concern.

PR #1867, opened 2021-06-01
Various Fixes to gfx, reflection and emit.

csyonghe:
- Fix GLSL emit to properly output `*bitsTo*` functions for `IRBitCast` insts.

- Add line directive mode setting in `TargetDesc`, moves line directive mode from `ICompileRequest` to `TargetRequest`, add `spSetTargetLineDirectiveMode` and deprecates `spSetLineDirectiveMode`.

- Add `structureSize` field to `TargetDesc` and `SessionDesc` for future binary compatibility.

- Extend `TypeLayout::getElementStride` to handle `VectorType` case.

- Fix `IDevice::readBufferResource` 's D3D12  implementation to copy only the requested bytes out.

- Fix `render-test` to use the `ISession` from `gfx` instead of creating its own `ISession` to make sure `gfx` and `render-test` agree on WitnessTable and RTTI IDs.

- Extend `render-test` to support filling vector and matrix values in the new `set x = ...` TEST_INPUT syntax.

- Add a `dynamic-dispatch-15` test case to make sure packing / unpacking works correctly across all targets, and to make sure render-test's RTTI/WitnessTable ID filling logic is correct for non-trivial cases.

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -3857,6 +3857,9 @@ namespace slang
         PreprocessorMacroDesc const*    preprocessorMacros = nullptr;
         SlangInt                        preprocessorMacroCount = 0;
 
+            /** The line directive mode for output source code.
+            */
+        SlangLineDirectiveMode lineDirectiveMode = SLANG_LINE_DIRECTIVE_MODE_DEFAULT;
     };
 
     enum class ContainerType
--- a/slang.h
+++ b/slang.h
@@ -1317,9 +1317,16 @@ extern "C"
         SlangCompileRequest*    request,
         const char* prefix);
 
-    /*! @see slang::ICompileRequest::setLineDirectiveMode */
+    /*! DEPRECATED: use `spSetTargetLineDirectiveMode` instead.
+        @see slang::ICompileRequest::setLineDirectiveMode */
     SLANG_API void spSetLineDirectiveMode(
         SlangCompileRequest*    request,
+        SlangLineDirectiveMode  mode);
+        
+    /*! @see slang::ICompileRequest::setTargetLineDirectiveMode */
+    SLANG_API void spSetTargetLineDirectiveMode(
+        SlangCompileRequest*    request,
+        int targetIndex,
         SlangLineDirectiveMode  mode);
 
     /*! @see slang::ICompileRequest::setCodeGenTarget */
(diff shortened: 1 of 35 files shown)


slang.h
```diff
@@ -3857,6 +3857,9 @@ namespace slang
         PreprocessorMacroDesc const*    preprocessorMacros = nullptr;
         SlangInt                        preprocessorMacroCount = 0;
 
+            /** The line directive mode for output source code.
+            */
+        SlangLineDirectiveMode lineDirectiveMode = SLANG_LINE_DIRECTIVE_MODE_DEFAULT;
```
tfoleyNV, line 3862: This realistically belongs in `TargetDesc`.

Also note that adding fields to these `Desc` types creates a binary backwards-compatibility issue, and it isn't actually something we've done before. In principle, this change would mean we can't drop a new `slang.dll` into an application compiled against an older version of `slang.h`.

I'm tentatively okay with letting this slip through for now just to see what happens, but we really need to put a true binary-compatibility plan into place for this stuff.

source/slang/slang-compiler.h
```diff
@@ -1408,6 +1410,8 @@ namespace Slang
         /// Set if fileSystemExt is a cache file system
         RefPtr<CacheFileSystem> m_cacheFileSystem;
 
+        LineDirectiveMode m_lineDirectiveMode = LineDirectiveMode::Default;
```
tfoleyNV, line 1413: Okay, I'm clearly missing something here. Why does the linkage/session need a line-directive mode stored on it? It really belongs on the targets, AFAICT. Or does the existing logic not correctly associate line-directive modes with specific targets?

source/slang/slang-emit-glsl.cpp
```diff
@@ -1404,21 +1404,48 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
         case kIROp_BitCast:
         {
             auto toType = extractBaseType(inst->getDataType());
+            auto fromType = extractBaseType(inst->getOperand(0)->getDataType());
```
tfoleyNV, line 1407: Maybe the original form of the logic got lost, but at some point I had intended for this conversion operations to implement the conversion in two steps to keep things simple:

1. Convert `fromType` to the same-size `uint` type, which we can consider the "intermediate" type.
2. Convert the intermediate type to `toType`, with the knowledge that the intermediate is always a `uint` type of matching size

The alternative would seem to be having nested `switch` statements in the long run, which is obviously not ideal.

source/slang/slang.cpp
```diff
@@ -3842,6 +3854,7 @@ void EndToEndCompileRequest::setLineDirectiveMode(SlangLineDirectiveMode mode)
 {
     // TODO: validation
     getBackEndReq()->lineDirectiveMode = LineDirectiveMode(mode);
+    m_linkage->setLineDirectiveMode(LineDirectiveMode(mode));
```
tfoleyNV, line 3857: Okay, I see. It seems like the existing line-directive mode is coming through from the end-to-end (command line) request, and thus maps most directly to a linkage-wide setting.

It would probably be good to have a later change make this a per-target-req setting instead, because that would seem to be the appropriate granularity for setting it, *especially* since the line directive mode is really part of the implementation of "debug information" for a particular target.

tfoleyNV (approved):

tfoleyNV (approved):

PR #1868, opened 2021-06-02
Fix cyclic reference in `ExtendedTypeLayout`.

csyonghe:
This change fixes a cyclic reference in `ExtendedTypeLayout`.
The problematic reference is in `TypeLayout::m_extendedLayout::subObjectRange::varOffsetLayout::typeLayout`, which references back to the initial `TypeLayout` object in a `RefPtr`. The solution is to introduce a `TransformablePtr` that can be demoted into a weak pointer, and demote the `varOffsetLayout::typeLayout` into a weak pointer when creating the `varOffsetLayout`.


tfoleyNV (approved):
The `TransformablePtr` is not the most beautiful thing ever, but it seems to get the job done.

PR #1869, opened 2021-06-03
Fix D3D11 `uploadBufferResource`.

csyonghe:
Also: `[gfx]` nows sets define per-platform preprocessor macro when compiling shaders, this is needed to write platform-dependent shader code.

### DIFF
--- a/tools/gfx/cuda/render-cuda.cpp
+++ b/tools/gfx/cuda/render-cuda.cpp
@@ -1176,7 +1176,11 @@ public:
 public:
     virtual SLANG_NO_THROW SlangResult SLANG_MCALL initialize(const Desc& desc) override
     {
-        SLANG_RETURN_ON_FAIL(slangContext.initialize(desc.slang, SLANG_PTX, "sm_5_1"));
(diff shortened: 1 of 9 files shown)


tools/gfx/cuda/render-cuda.cpp
```diff
@@ -1176,7 +1176,11 @@ class CUDADevice : public RendererBase
 public:
     virtual SLANG_NO_THROW SlangResult SLANG_MCALL initialize(const Desc& desc) override
     {
-        SLANG_RETURN_ON_FAIL(slangContext.initialize(desc.slang, SLANG_PTX, "sm_5_1"));
+        SLANG_RETURN_ON_FAIL(slangContext.initialize(
+            desc.slang,
+            SLANG_PTX,
+            "sm_5_1",
```
jsmall-nvidia, line 1182: Is sm_5_1 an suitable profile for CUDA? If it doesn't need a profile can I just pass nothing?

jsmall-nvidia (dismissed):

PR #1870, opened 2021-06-03
Escape `\` in slang-embed.

csyonghe:


jsmall-nvidia (approved):
Whilst not necessary, we could use the StringEscapeUtil to do this escaping. It would be DRY and we could be more confident our implementation covers all the cases.

PR #1871, opened 2021-06-04
Enable tracing rays with OptiX backend

n8vm:
This pull request introduces the necessary functionality to allow users to call `TraceRay` for PTX output. 

Additionally, hit attributes composed of vectors, matrices, arrays, and structs are now supported. This enables users to query the barycentrics of a triangle in a closest hit program, but down the line should also support reading from custom user geometry hit attributes. 

Note that custom intersection programs still require some work.


prelude/slang-cuda-prelude.h
```diff
@@ -1615,26 +1615,63 @@ struct UniformState;
...
 static __forceinline__ __device__
 void *unpackOptiXRayPayloadPointer(uint32_t i0, uint32_t i1)
 {
-	const uint64_t uptr = static_cast<uint64_t>(i0) << 32 | i1;
-	void*           ptr = reinterpret_cast<void*>(uptr);
-	return ptr;
+    const uint64_t uptr = static_cast<uint64_t>(i0) << 32 | i1;
+    void*           ptr = reinterpret_cast<void*>(uptr);
+    return ptr;
 }
```
n8vm, line 1632: Note, I converted these tabs to spaces to stay consistent with the rest of the prelude file

prelude/slang-cuda-prelude.h
```diff
@@ -1615,26 +1615,63 @@ struct UniformState;
...
 
 static __forceinline__ __device__ void *getOptiXRayPayloadPtr()
 {
-	const uint32_t u0 = optixGetPayload_0();
-	const uint32_t u1 = optixGetPayload_1();
-	return unpackOptiXRayPayloadPointer(u0, u1);
+    const uint32_t u0 = optixGetPayload_0();
+    const uint32_t u1 = optixGetPayload_1();
+    return unpackOptiXRayPayloadPointer(u0, u1);
+}
```
n8vm, line 1647: Same correction of tabs to spaces here

prelude/slang-cuda-prelude.h
```diff
@@ -1615,26 +1615,63 @@ struct UniformState;
...
+    uint32_t RayFlags,
+    uint32_t InstanceInclusionMask,
+    uint32_t RayContributionToHitGroupIndex,
+    uint32_t MultiplierForGeometryContributionToHitGroupIndex,
+    uint32_t MissShaderIndex,
+    RayDesc Ray,
+    T *Payload
+) {
+    uint32_t r0, r1;
+    packOptiXRayPayloadPointer((void*)Payload, r0, r1);
```
n8vm, line 1661: In the future, I intend to optimize the ray payload storage here to use as many registers as possible before then switching to this pointer strategy. As a result, a lot of this prelude code here will likely move to slang-emit-cuda.cpp in a subsequent revision.

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -1214,6 +1291,17 @@ struct CUDAEntryPointVaryingParamLegalizeContext : EntryPointVaryingParamLegaliz
                 return diagnoseUnsupportedUserVal(info);
             }*/ 
         }
+        case LayoutResourceKind::HitAttributes: {
+            IRBuilder builder(m_sharedBuilder);
+            builder.setInsertBefore(m_firstOrdinaryInst);
+            int ioBaseAttributeIndex = 0;
+            IRInst* getHitAttributes = emitOptiXAttributeFetch(/*ioBaseAttributeIndex*/ ioBaseAttributeIndex, /* type to fetch */info.type, /*the builder in use*/ &builder);
+            if (ioBaseAttributeIndex > 8) {
+                m_sink->diagnose(m_param, Diagnostics::unexpected, "the supplied hit attribute exceeds the maximum hit attribute structure size (32 bytes)");
```
n8vm, line 1300: Is there a better way to report errors like these to the user?

source/slang/slang-emit-cuda.cpp
```diff
@@ -788,6 +788,29 @@ bool CUDASourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
             m_writer->emit(")getOptiXRayPayloadPtr()");
             return true;
         }
+        case kIROp_GetOptiXHitAttribute:
+        {
+            auto typeToFetch = inst->getOperand(0);
+            auto idxInst = as<IRIntLit>(inst->getOperand(1));
+            IRIntegerValue idx = idxInst->getValue();
+            if (typeToFetch->getOp() == kIROp_FloatType) {
```
jsmall-nvidia, line 796: This is a pedantic point but our coding standard has allman rather than K&R indentation style. So { should be on next line and tabbed to the if. https://github.com/shader-slang/slang/blob/master/docs/design/coding-conventions.md

source/slang/slang-emit-cuda.cpp
```diff
@@ -788,6 +788,29 @@ bool CUDASourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+                m_writer->emit("__int_as_float(optixGetAttribute_");
+            }
+            else
+            {
+                m_writer->emit("optixGetAttribute_");
+            }
+            m_writer->emit(idx);
```
jsmall-nvidia, line 803: Nitpic - it might be slightly neater to have the special float logic all together rather (including the emit idx) instead of checking again for special casing the close.

jsmall-nvidia (approved):

source/slang/slang-emit-cuda.cpp
tfoleyNV: My guess is that eventually we'd want `GetOptiXHitAttribute` to always return a `uint`, and then have the code that inserts the new IR instructions also insert a bit-cast when it is needed. That would avoid mixing up the logic for fetching attributes and the logic for converting bits between types.

PR #1874, opened 2021-06-04
Fixed issue around 4xFloat16 texture on CUDA

jsmall-nvidia:
Bug in CUDA impl, where the element size is incorrect if there is more than one channel.

Small surrounding fixes. Re-enabled one of rw-texture tests for CUDA.


csyonghe (approved):
Great to see this bug figured out!

PR #1876, opened 2021-06-07
Fix RWTexture issues on CUDA

jsmall-nvidia:
* Enable RWTexture tests on CUDA
* Fixed bugs around byte addressing for surface access on CUDA


tests/compute/rw-texture-simple.slang
```diff
@@ -6,14 +6,14 @@
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx12 -profile cs_6_0 -use-dxil -output-using-type -shaderobj
 // TODO(JS): Doesn't work on vk currently, because createTextureView not implemented on vk renderer
 //DISABLE_TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -output-using-type -shaderobj
```
csyonghe, line 8: Can we try to enable the test on VK today? I think our gfx layer has already implemented texture views.

csyonghe (approved):

PR #1877, opened 2021-06-07
Various fixes to CUDA backend.

csyonghe:
- Fix emitting `StructuredBuffer<ISomething>::Load`, which triggers emitting for `IROp_WrapExistential` that is previously unhandled.

- Fix cuda layout around vectors, they should be aligned to 1,2,4,8,16 bytes instead of just using element type's alignment. That means `float4` has alignment of 16 instead of 4.

- Fix `SLANG_CUDA_HANDLE_ERROR` macro definition.

- Fix navis sometimes fail to find `Slang::kIROp_*` enum values when debugging external projects.


source/slang/slang-type-layout.cpp
```diff
@@ -479,8 +491,8 @@ struct CUDALayoutRulesImpl : DefaultLayoutRulesImpl
         SimpleLayoutInfo vectorInfo;
         vectorInfo.kind = elementInfo.kind;
         vectorInfo.size = elementInfo.size * elementCount;
-        vectorInfo.alignment = elementInfo.alignment;
-    
+        vectorInfo.alignment = getVectorAlignment(
```
jsmall-nvidia, line 494: It looks like size/alignment on CUDA is unfortunately more subtle than this. Can be improved on in subsequent PR.

source/slang/slang-ir.cpp
```diff
@@ -6019,3 +6019,15 @@ namespace Slang
     }
 } // namespace Slang
 
+#if SLANG_VC
+#ifdef _DEBUG
+// Natvis sometimes cannot find enum values.
+// Export symbols for them to make sure natvis works correctly when debugging external projects.
```
jsmall-nvidia, line 6025: Why not put these symbols into a namespace (like say SlangDebug )? Then they could be (say) `SlangDebug::kIROpNameHint`. Or if you don't want to use a namespace as you are worried about implication of name export/mangling how about `Slang_kIROpNameHint.

That the __ usage is worrying as

https://en.cppreference.com/w/cpp/language/identifiers


```
     the identifiers with a double underscore anywhere are reserved;
    the identifiers that begin with an underscore followed by an uppercase letter are reserved;
    the identifiers that begin with an underscore are reserved in the global namespace. 
```

jsmall-nvidia (approved):

PR #1879, opened 2021-06-08
Fix CUDA vector layout logic.

csyonghe:
And rename debug symbols for navis.


jsmall-nvidia (approved):
Some small edge cases remain with CUDA - will add additional PR covering those cases.

PR #1880, opened 2021-06-09
Support timestamp queries in `gfx`.

csyonghe:
This changes add an `IQueryPool` interface that is similar to Vulkan's `VkQueryPool` to allow querying GPU timestamps.
This feature can be used to obtain performance data.

### DIFF
--- a/tools/gfx/debug-layer.cpp
+++ b/tools/gfx/debug-layer.cpp
@@ -140,6 +140,8 @@ SLANG_GFX_DEBUG_GET_INTERFACE_IMPL(ShaderObject)
 SLANG_GFX_DEBUG_GET_INTERFACE_IMPL(ShaderProgram)
 SLANG_GFX_DEBUG_GET_INTERFACE_IMPL(Swapchain)
 SLANG_GFX_DEBUG_GET_INTERFACE_IMPL(TransientResourceHeap)
+SLANG_GFX_DEBUG_GET_INTERFACE_IMPL(QueryPool)
+
 
 #undef SLANG_GFX_DEBUG_GET_INTERFACE_IMPL
 #undef SLANG_GFX_DEBUG_GET_INTERFACE_IMPL_PARENT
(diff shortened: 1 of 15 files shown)


tools/gfx/debug-layer.cpp
```diff
@@ -477,6 +479,16 @@ const DeviceInfo& DebugDevice::getDeviceInfo() const
     return baseObject->getDeviceInfo();
 }
 
+Result DebugDevice::createQueryPool(const IQueryPool::Desc& desc, IQueryPool** outPool)
+{
+    SLANG_GFX_API_FUNC;
+    RefPtr<DebugQueryPool> result = new DebugQueryPool();
```
jsmall-nvidia, line 485: Super nitpic point - why not make the desc a param to a constructor.

tools/gfx/debug-layer.cpp
```diff
@@ -706,12 +724,24 @@ void DebugRenderCommandEncoder::setStencilReference(uint32_t referenceValue)
     return baseObject->setStencilReference(referenceValue);
 }
 
+void DebugRenderCommandEncoder::writeTimestamp(IQueryPool* pool, SlangInt index)
+{
+    SLANG_GFX_API_FUNC;
+    baseObject->writeTimestamp(static_cast<DebugQueryPool*>(pool)->baseObject, index);
```
jsmall-nvidia, line 730: It's possible to do something like a dynamic cast for incoming com interfaces by giving the impls their own guid, and querying for that, This can be seen implemented in CacheFileSystem::queryInterface(SlangUUID const& uuid, void** outObject). By adding `SLANG_CLASS_GUID(0x2f4d1d03, 0xa0d1, 0x434b, { 0x87, 0x7a, 0x65, 0x5, 0xa4, 0xa0, 0x9a, 0x3b })` and then doing 

```
 if (uuid == CacheFileSystem::getTypeGuid())
    {
        *outObject = this;
        return SLANG_OK;
    }
```

Note does not increment reference count. 

This some what outside of 'COM' but is 'safe' under the assumption that Guids don't clash (they are unique!), and that the GUID you are using is private to your impl. 

This is not *necessary* but thought I'd throw it out there.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -3448,7 +3451,7 @@ class VKDevice : public RendererBase
         // a `CommandBuffer` created from the heap. We need to break the cycle when
         // the public reference count of a command buffer drops to 0.
         SLANG_COM_OBJECT_IUNKNOWN_ALL
-        ICommandBuffer* getInterface(const Guid& guid)
+            ICommandBuffer* getInterface(const Guid& guid)
```
jsmall-nvidia, line 3454: Tab?

jsmall-nvidia (dismissed):
Nice functionality to have.

PR #1882, opened 2021-06-10
Properly fill `declref` in `Linkage::getContainerType`.

csyonghe:
`Linkage::getContainerType` is introduced in #1851 to allow users to conveniently obtain a specialized `StructuredBuffer<T>` or `ParameterBlock<T>` types with a given element type.

The implementation was not complete in that it does not fill in the `declRef` field of the resulting `Type`. This can lead to crashes when the `declRef` is used, such as in `Type::getName()`. This change fixes this issue by properly looking up and fill in the declref.

Also includes a small fix to the cpu timestamp query implementation.

### DIFF
--- a/source/slang/slang-ast-builder.cpp
+++ b/source/slang/slang-ast-builder.cpp
@@ -257,6 +257,29 @@ VectorExpressionType* ASTBuilder::getVectorType(
     return as<VectorExpressionType>(DeclRefType::create(this, declRef));
 }
 
+DeclRef<Decl> ASTBuilder::getBuiltinDeclRef(const char* bubiltinMagicTypeName, ConstArrayView<Val*> genericArgs)
+{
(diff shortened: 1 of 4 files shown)


source/slang/slang-ast-builder.cpp
```diff
@@ -257,6 +257,29 @@ VectorExpressionType* ASTBuilder::getVectorType(
     return as<VectorExpressionType>(DeclRefType::create(this, declRef));
 }
 
+DeclRef<Decl> ASTBuilder::getBuiltinDeclRef(const char* bubiltinMagicTypeName, ConstArrayView<Val*> genericArgs)
```
jsmall-nvidia, line 260: bubiltinMagicTypeName typo?

jsmall-nvidia (approved):

PR #1888, opened 2021-06-17
StructTag versioning

jsmall-nvidia:
This PR takes ideas from #1645

The main idea is to identify struct types with an 'id' that includes a type identity and a semantic version. 

Due to wanting to support arrays of structs and be able to handle structs from 'the future' the mechanism here *also* has the structure size. Having the the struct size allows an array of structs 'from the future' to be resized down amongst other features and checks.

What the system can do 

* Versioning backwards and forwards in minor semantic versions
* Zero extending if needed (fairly simply could have a 'default fields' feature)
* Handles instances, arrays, and pointer arrays
* Supports 'extension' and 'primary' struct types
* Allows structs to contain pointers/counts to other versioned structures
* Provides two conversion mechanisms - Lazy and Copy

The Lazy mechanism tries to do the minimum amount of work to provide a struct version hierarchy that is compatible with the current environment. In the limit it can do this by not copying or converting anything, but just returning the pointer to the thing passed in. 

The Copy mechanism always copies/converts everything. This provides an easy way to keep around a struct hierarchy without having to copy and memory track.

What it doesn't have

* Is a mechanism to convert from current (presumably internal) types to external types (this is covered in later PRs)
* To do this we need a mechanism of knowing what types the outside *wants* - and that probably entails them passing that information in (because of extensions and pointers/arrays) (discussed in later PRs)
* Doesn't have a delta mechanism to convert between major types

Converting to external types is quite straight forward if we know their tags and sizes. Doing so just requires making another system, which holds those versions (and removes fields if not in range). Then we just convert from the internal to external. This implies that we would want a method that creates some opaque type that contains the system, that can be passed to methods that output tagged structs.


slang.h
```diff
@@ -450,6 +450,178 @@ convention for interface methods.
...
+     |Primary | Category     | Type      | Major version | Minor Version
+-----|--------|--------------|-----------|---------------|----------------
+Bits | 31     | 30-24        | 23-16     |     15-8      |     0-7
+Type |        | AbiCategory  | Cat Spec  |               |
+
+Type will be specified as something specific for the category. A typical implementation will have an enum that lists
+types for that category. 
+
+TODO(JS): This layout may need to be altered - too may bits are perhaps used by major/minor version for example. This should
+be fine for the immediate future though.
```
tfoleyNV, line 501: One note is that the `Primary`-ness of a structure may not need to be encoded as its own bit, since it would be implied by the `Type`. The bit is thus redundant for any recognized `Category`+`Type` pair, and for any case where we don't recognize the category and/or type, just knowing whether something is primary or not is only of limited value (we could specialize a diagnostic message based on that bit, but I'm not clear it is worth distinguishing from other "I don't recognize this type" cases).

tfoleyNV: (I was about to put forward a bunch more ideas on how we could allocate the bits for all of this, but a PR review is probably not the best place to bikeshed all the design stuff)

slang.h
```diff
@@ -450,6 +450,178 @@ convention for interface methods.
...
+    slang::StructSize structSize = slang::StructSize(sizeof(ThisType)); \
+    const slang::StructTag** exts = nullptr; \
+    int32_t extsCount = 0; 
+
+#define SLANG_API_TYPE_ENUM(x) x,
+
+struct TaggedStructBase
+{
+    StructTag structTag;            ///< Identity for the type
+    StructSize structSize;             ///< Size of the type in bytes
```
tfoleyNV, line 584: Why are we *also* including `structSize` if we already have `structTag`? I very much thought of this as an either/or.

My thought had been that the validation logic traffics in two main cases:

1. The `structTag` is one that is allowed + recognized (and thus have a declaration for) and we can freely use that declaration directly to copy things: `ConcreteType& concrete = *(ConcreteType*)structPtr;`

2. The `structTag` is one that is not allowed (whether recognized or not), so we simply diagnose an error about the mismatch, which relies on name/version/whatever information from the tag (and doesn't care about the actual data in the struct)

I can't see how either of those cases is improved by having a `structSize` field, and having both seems wasteful to me. I'm happy to be convinced otherwise, of course.

slang.h
```diff
@@ -450,6 +450,178 @@ convention for interface methods.
...
+// Enumerate all of the types that want ABI handling as part of Slang API.
+//
+// NOTE! That care is needed using the enumeration, because we want IDs to remain stable.
+// New types should be added to the end. Removed types should become Depreciated_Name (say).
+//
+// NOTE! The names must match the type name for an active type exactly for the SLANG_PRIMARY_TAGGED_STRUCT and SLANG_EXTENSION_TAGGED_STRUCT
+// macros to work.
+#define SLANG_TAGGED_STRUCTS(x) \
+    x(TargetDesc) \
+    x(SessionDesc)
```
tfoleyNV, line 612: Style note: I always like to have multiline macro definitions like this terminate each content line with `\` and then have a stand-along line to mark the actual end of the macro, e.g.:

```
#define SLANG_TAGGED_STRUCTS(x) \
    x(TargetDesc) \
    x(SessionDesc) \
    /* end */
```

Formatting things that avoids spurious diffs to the existing lines when adding new cases, and helps avoid the kind of editing error where somebody adds a new case *with* a trailing `\` and accidentally makes the macro definition include a subsequent code line by accident.

source/compiler-core/slang-struct-tag-converter.h
```diff
@@ -0,0 +1,153 @@
+#ifndef SLANG_COMPILER_CORE_STRUCT_TAG_CONVERTER_H
+#define SLANG_COMPILER_CORE_STRUCT_TAG_CONVERTER_H
+
+#include "slang-struct-tag-system.h"
+#include "slang-diagnostic-sink.h"
+
+namespace Slang {
+
+class StructTagConverterBase
```
tfoleyNV, line 9: Maybe I'm missing something but this seems like a lot of infrastructure compared to the simpler thing of just having each relevant function do `switch(tag) { ... }` and either handle each allowed case directly (there's only one right now!) or convert the data for each case into a common representation.

For handling extensions, it seems like we'd often want code along the lines of:

```
// one local per recognized extension and version thereof
ExtA* a = nullptr;
ExtB* b = nullptr;
for(auto extPtr : primary->exts) {
    switch(extPtr->tag) {
        case A: a = (A*) extPtr; break;
        ...
    }
}
// later code can easily do:
if(a) { /* a-specific stuff */ }
```

I'm beginning to see that we maybe had very different ideas of how much of an underlying *system* needed to exist behind this. In my mind the tags are just a glorified `enum` to indicate which of a list of `struct` types was passed in, and the logic for actually handling those cases would be simplest and fastest if it is ad hoc...

tfoleyNV (approved):

slang.h
jsmall-nvidia: Having the structSize was added so I could support arrays of structs as opposed to arrays of pointers to structs. Having it allows for struct from the 'future' or the 'past' relative to current.

source/compiler-core/slang-struct-tag-converter.h
jsmall-nvidia: This is nice in so far as allows for programmatic transformations between versions. The PR solution provides mechanical transformations around semantic versioning (for now at least).

On the other hand the mechanism above implies that all the versions are available, and that any handling is ad-hoc. 

Whilst the mechanical semantic versioning is more limited, once it exists - it does not require keeping previous versions around, or ad-hoc processing or error handling. You do it once and you are done.

Issue #1891, opened 2021-06-22
Implementation of SamplerTexture2D is incomplete


GlaireDaggers:
My engine relies on combined texture samplers in Vulkan to do texturing, but the implementation in Slang appears to be missing some functionality - mainly, attempting to _sample_ from a SamplerTexture2D using the `texture` function gives the following error:

```
core.meta.slang(689): internal error 99999: unimplemented feature in Slang compiler: unexpected IR opcode during code emit
```

tfoleyNV:
Thanks for posting this as its own issue. I can probably try to tackle this, but I'm wondering what kind of timeline you have on when you'd need the implementation for this to be useful.

GlaireDaggers:
I'm not in any particular _hurry_ for what it's worth, since right now this is just for my own projects, but it _is_ a blocker for remaining work on the rendering pipeline in the engine.

tfoleyNV:
I'm starting to work on this and have made some positive steps toward getting things fixed in my branch. When this stuff lands the names and signatures are likely to be different than what you are working with right now, so I will try to document stuff in this issue when that time comes.

tfoleyNV:
PR #1894 should fix up a lot of the compiler's behavior around these combined types, and those fixes should be available in the `v0.19.0` and later releases. However, the change completely overhauls how combined texture/sampler type are written in input code.

Now, given an existing HLSL texture type like `Texture2D` the equivalent combined texture/sampler type has a name with `Sampler` in place of `Texture`. So the equivalent of `Texture2D` is the combined texture/sampler type `Sampler2D`.

Similarly, for any operations on a given `Texture*` type, the equivalent `Sampler*` type should have the same operations, just with any `SamplerState` parameter removed.

So if you would write the following with separate textures and samplers:

```hlsl
Texture2D t;
SamplerState s;

float4 test(float2 uv) { return t.Sample(s, uv); }
```

you would instead write the following to use the new combined types:

```hlsl
Sampler2D s;

float4 test(float2 uv) { return s.Sample(uv); }
```

Hopefully this large change doesn't cause too much disruption for your codebase @GlaireDaggers. Please feel free to let me know of any issues you run into with the newly re-enabled functionality. It is entirely possible that I've missed some important cases.

GlaireDaggers:
Ah, sorry I've been silent for a bit. Been on vacation, just got back last night. Will take a look at this hopefully later tonight or tomorrow! Thanks for looking into it :)

PR #1894, opened 2021-06-23
Fixes related to combined texture/sampler types

tfoleyNV:
Work on #1891

Our intention has always been to support combined texture/sampler types in Slang, both for targets like OpenGL where that is the only option available and for targets like Vulkan where it can be beneficial to performance. Because Slang's current users mostly focus on D3D12+Vulkan codebases, they strongly prefer separate textures and samplers, and the relevant support code in Slang has "bit-rotted" over many releases until the functionality that was there isn't useful any more.

This change significantly overhauls the implementation of combined texture/sampler types and adds a test that uses them in the hopes of avoiding future regressions.

The new combined texture/sampler types use the prefix `Sampler`, so where there is an existing standalone `Texture2D` type the equivalent combined texture/sampler type will be `Sampler2D`. The intention is that this naming mirrors the GLSL conventions (where the type is `sampler2d`) while following HLSL naming precedent (to the extent it exists).

The operations available on a `Sampler2D` are intended to be those that are available on a `Texture2D`, and it is just that in cases where the `Texture2D` operation would take a separate sampler argument:

    Texture2D t = ...;
    SamplerState s = ...;
    float4 result = t.Sample(s, uv);

the equivalent `Sampler2D` operation just elides that argument:

    Sampler2D s = ...;
    float4 result = s.Sample(uv);

In terms of implementation, there are a lot of subtle details here:

* I've tried to use the same metaprogramming logic that generates all the stdlib declarations for `Texture*` to also generate `Sampler*` in the hopes that this helps keep them in sync.

* The big catch to the above is that it means that for certain operations the indidces of parameters depend on whether or not an explicit sampler parameter is used. Rather than try to tweak the indices in the stdlib generation logic (which is already complicated) I went and added Yet Another Hack to the logic that handles intrinsic definition strings. Basically, the special-case handling of `$p` has been modified so that it *also* applies a negative offset to future parameter references in the same intrinsic string.

* Trying to actually bring this up in our test framework revealed that the "flat" reflection API was seemingly not reflecting combined texture/sampler types correctly at all (it was reflecting them as just plain textures). Other than that issue, the Vulkan path seems to work fine with combined texture/samplers.

* I also had to add logic to the `TEST_INPUT` parsing to re-introduce handling of the combined types (that was something I consciously left out to reduce the amount of code in the earlier refactor there). Luckily, the architecture is such that a combined texture/sampler can leverage most of the existing logic for the separate cases.


source/slang/core.meta.slang
```diff
@@ -596,6 +596,7 @@ static const struct {
 };
 static const int kBaseTextureAccessLevelCount = sizeof(kBaseTextureAccessLevels) / sizeof(kBaseTextureAccessLevels[0]);
 
+#if 0
```
jsmall-nvidia, line 599: Did you want to leave this in?

source/slang/slang-intrinsic-expand.h
```diff
@@ -29,6 +29,26 @@ struct IntrinsicExpandContext
     Int m_argCount = 0;
     Index m_openParenCount = 0;
     CLikeSourceEmitter* m_emitter;
+
```
jsmall-nvidia, line 32: This would be nice, though it assumes the name are available in IR.

tools/render-test/shader-input-layout.cpp
```diff
@@ -673,10 +683,43 @@ namespace renderer_test
                 maybeParseOptions(parser, val.Ptr());
                 return val;
             }
-            // TODO: We can include combined texture/sampler cases
-            // here, but we don't really have tests for them and
-            // it might be better to save it for a point where
-            // we support hierarchical binding values more directly.
+            else if (parser.AdvanceIf("TextureSamplerSampler1D"))
```
jsmall-nvidia, line 686: SamplerSampler?

source/slang/core.meta.slang
```diff
@@ -701,22 +715,34 @@ for (int tt = 0; tt < kBaseTextureTypeCount; ++tt)
...
-                sb << name;
+                sb << baseName;
+                sb << baseShapeName;
                 if (isMultisample) sb << "MS";
                 if (isArray) sb << "Array";
     //                        if (isShadow) sb << "Shadow";
                 sb << "\n{";
 
+                char const* samplerStateParam = prefixInfo.combined ? "" : "SamplerState s, ";
+//                int samplerStateParamCount =  prefixInfo.combined ? 0 : 1;
```
jsmall-nvidia, line 738: Remove?

source/slang/core.meta.slang
```diff
@@ -652,23 +653,36 @@ for (int tt = 0; tt < kBaseTextureTypeCount; ++tt)
...
+{
+    char const* name;
+    bool        combined;
+} kTexturePrefixes[] =
+{
+    { "Texture", false },
+    { "Sampler", true },
+};
 
+for(auto& prefixInfo : kTexturePrefixes)
```
jsmall-nvidia, line 681: It seems perhaps a little irksome it's not indented - but in the context of the complexity/interation over variety of cases I guess it's fair enough. It's like you want 'iterate over the combinations of some things'.

source/slang/slang-intrinsic-expand.cpp
```diff
@@ -462,33 +484,41 @@ const char* IntrinsicExpandContext::_emitSpecial(const char* cursor)
...
+                if (elementCount < 4)
                 {
-                    // What other cases are possible?
+                    char const* swiz[] = { "", ".x", ".xy", ".xyz", "" };
+                    m_writer->emit(swiz[elementCount]);
                 }
             }
             else
             {
-                SLANG_UNEXPECTED("bad format in intrinsic definition");
```
jsmall-nvidia, line 491: Having the 'assert' doesn't seem unreasonable.

jsmall-nvidia (changes requested):
Think there is a typo in 'TextureSamplerSampler1D'. Other issues are mostly musing - although it's not clear why the big #if 0 section is left.

source/slang/core.meta.slang
tfoleyNV: Oops. No. Will delete.

source/slang/slang-intrinsic-expand.h
tfoleyNV: I was thinking that *if* we did this we would hypothetically move the parsing+validation step to the front-end (or maybe IR lowering?) where it could work on the AST.

Basically, we could parse the definition string using the existing logic, and translate it into a flat list of "ops" (akin to what we use for macros now):

* Raw string ops would reference a range of bytes from the original definition string
* Param reference ops would reference a parameter of the (IR-level) function by index
* Various other opcodes would be defined for the numerous special cases

During IR lowering the sequence of ops would be translated to produce zero or more operands to the `IRTargetIntrinsicDecoration`, and the original string that the front-end used would no longer be needed. The translation could just use the AST to look up parameters by name, but then write out the index of the parameter to the IR-level data.

It would be a lot of code just to remove a bit of hackery, and maybe there is a better top-down design for how to handle all of this (e.g., something like "inline assembly" and a `__switch_target` construct so that all this stuff lives inside the function body instead?).

tools/render-test/shader-input-layout.cpp
tfoleyNV: Oops. Find-and-replace bit me there. Good catch.

source/slang/core.meta.slang
tfoleyNV: Yep...

source/slang/core.meta.slang
tfoleyNV: I'm not *proud* of doing that, but I also wanted to avoid the massive diff due to re-indenting some of our most complicated stdlib generation logic.

It actually seems like we might benefit from factoring the logic that generates the texture type definitions into one or more subroutines, or a "context" type with overloadable policy methods. There's just *so much* detail there and it is basically impossible to get anything right on the first try.

source/slang/slang-intrinsic-expand.cpp
tfoleyNV: Yeah, I think this was an example of me trying to do defensive coding. There is in principle the possibility somebody does:

```
struct MyStuff { float4 f; }
Texture2D<MyStuff> gMyStuffTex;
```

But realistically I don't expect the rest of the Slang compiler to handle that robustly anyway. It is something that I think historically fxc allowed, but that was partly due to the "anything goes" rules fxc seemed to employ.

It is probably healthier for us to say that texel types are always scalar basic types or vectors (which is what that assert enforces, in a roundabout way), and then if we want to relax that later actually put in the necessary design effort.

tfoleyNV:
Thanks for the feedback. I think this shows that I was maybe a bit hasty to get this PR up by the end of my work day and didn't put in the due diligence. I will fix the issues and update soon (hopefully later today, but possibly Friday).

PR #1898, opened 2021-06-29
Fix uninitialized access identified by Valgrind

jsmall-nvidia:
On looking at the possible flow for this the uninitialized usage happens in 

```
case Region::Flavor::Switch: 
```

When accessing `defaultCase`. The Ctor for SwitchRegion does not set it - the change here just sets to nullptr (which the comments says is ok), but it's not clear if it is correct for the flow that leads to the access.

It looks like it is left unintialized when 

```
                    // If the `default` case just branches to
                    // the join point, then we don't need to
                    // do anything with it.
                    //
                    if(defaultLabel == breakLabel)
                        defaultLabelHandled = true;
```

Means it is not set.

With this change valgrind no longer issues a warning.


tfoleyNV (approved):
Nice catch.

PR #1899, opened 2021-06-29
[gfx] Add inline ray tracing support.

csyonghe:
This PR adds inline ray tracing to `gfx` by adding a new `IAccelerationStructure` interface and its associated commands
via `IRayTracingCommandEncoder`.

This PR provides a Vulkan implementation and introduces a new `ray-tracing` example.


slang-gfx.h
```diff
@@ -871,16 +1028,12 @@ class IQueryPool : public ISlangUnknown
     { 0xc2cc3784, 0x12da, 0x480a, { 0xa8, 0x74, 0x8b, 0x31, 0x96, 0x1c, 0xa4, 0x36 } }
 
 
-class ICommandEncoder : public ISlangUnknown
+class ICommandEncoder
```
jsmall-nvidia, line 1031: By removing ISlangUnknown reference counting is removed (which could be seen as a benefit) - but you also lose being able to queryInterface, which breaks the main mechanisms for future non breaking binary compatibility. 

Moreover in this specific scenario since it has interfaces deriving from it, it means there is no mechanism to upcast.

slang-gfx.h
```diff
@@ -973,9 +1118,50 @@ class IResourceCommandEncoder : public ICommandEncoder
...
+};
+
+class IRayTracingCommandEncoder : public ICommandEncoder
+{
+public:
+    virtual SLANG_NO_THROW void SLANG_MCALL buildAccelerationStructure(
+        const IAccelerationStructure::BuildDesc& desc,
+        int propertyQueryCount,
+        AccelerationStructureQueryDesc* queryDescs) = 0;
+    virtual SLANG_NO_THROW void SLANG_MCALL copyAccelerationStructure(
```
jsmall-nvidia, line 1143: Pedantically in COM interfaces the 'output' is typically the last parameter. It could be argued that this 'out' is really a modify so that's fine. I actually think the reverse order is better 

```
copyAccelerationStructure( mode, src, dst)
```

None of this is super important I'll grant.

jsmall-nvidia (approved):
Looks good.

slang-gfx.h
```diff
@@ -495,11 +498,163 @@ class IResourceView : public ISlangUnknown
...
+        {
+            None = 0,
+            TriangleFacingCullDisable = 0x00000001,
+            TriangleFrontCounterClockwise = 0x00000002,
+            ForceOpaque = 0x00000004,
+            NoOpaque = 0x00000008
+        };
+    };
+
+    struct InstanceDesc
```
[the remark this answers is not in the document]
csyonghe: This struct is basically copied entirely from Vulkan header/ dx spec and is supposed to be filled in and copied to a gpu buffer. I guess we don't need to worry too much about compatibility for the platforms we cared about.

slang-gfx.h
csyonghe: Having this interface being a ISlangUnknown encourages the use of ComPtr on the user side, which can lead to wrong code since the object is transient and could be deleted when the user calls release via ComPtr. If we ever need queryInterface we should introduce a ITemporary or something like that in the future.

slang-gfx.h
csyonghe: This is to stay consistent with the d3d api. The call maps 1:1 to d3d's BuildAccelerationStructure function so I'd just keep the interface the same.

PR #1906, opened 2021-07-09
Enable testing with Swiftshader.

csyonghe:
This change enables gfx Vulkan implementation to find and load swiftshader if hardware support is absent.

Also makes github CI to download swiftshader binaries before running the tests, so that we can have VK coverage on GitHub.

Add `render-feature` requirements in several test cases to prevent running them when using swiftshader since many features such as `double`, `int64` are not supported by swiftshader.


jsmall-nvidia (approved):
Nice

PR #1909, opened 2021-07-14
Enable swiftshader in linux CI builds

csyonghe:


jsmall-nvidia (approved):
Nice to have swiftshader on linux. Comments are just some observations.

PR #1913, opened 2021-07-19
Minor refactor to gfx D3D12 implementation.

csyonghe:
- Allow more flexible collection of shader stages in a shader program.
- Add `createRayTracingPipelineState` public interface. (no implementation).
- Fix a bug in vulkan initialization that causes a crash when vulkan is not available.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -909,6 +909,25 @@ struct ComputePipelineStateDesc
     IShaderProgram*  program;
 };
 
+struct RayTracingPipelineFlags
+{
+    enum Enum : uint32_t
+    {
+        None = 0,
+        SkipTriangles = 1,
+        SkipProcedurals = 2,
+    };
+};
+
+struct RayTracingPipelineStateDesc
+{
+    IShaderProgram* program = nullptr;
+
+    int maxRecursion;
+    int maxRayPayloadSize;
+    RayTracingPipelineFlags::Enum flags;
+};
+
 class IPipelineState : public ISlangUnknown
 {
 };
@@ -1596,6 +1615,9 @@ public:
         SLANG_RETURN_NULL_ON_FAIL(createComputePipelineState(desc, state.writeRef()));
         return state;
     }
+
+    virtual SLANG_NO_THROW Result SLANG_MCALL createRayTracingPipelineState(
+        const RayTracingPipelineStateDesc& desc, IPipelineState** outState) = 0;
 
         /// Read back texture resource and stores the result in `outBlob`.
     virtual SLANG_NO_THROW SlangResult SLANG_MCALL readTextureResource(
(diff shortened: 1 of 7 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5407,8 +5422,8 @@ Result D3D12Device::createComputePipelineState(const ComputePipelineStateDesc& i
         D3D12_COMPUTE_PIPELINE_STATE_DESC computeDesc = {};
         computeDesc.pRootSignature = programImpl->m_rootObjectLayout->m_rootSignature;
         computeDesc.CS = {
-            programImpl->m_computeShader.getBuffer(),
```
jsmall-nvidia, line 5410: Should there be a check that there is a m_shaders.getCount() == 1 and it's the right type?

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5294,9 +5287,31 @@ Result D3D12Device::createGraphicsPipelineState(const GraphicsPipelineStateDesc&
     D3D12_GRAPHICS_PIPELINE_STATE_DESC psoDesc = {};
 
     psoDesc.pRootSignature = programImpl->m_rootObjectLayout->m_rootSignature;
-
-    psoDesc.VS = { programImpl->m_vertexShader.getBuffer(), SIZE_T(programImpl->m_vertexShader.getCount()) };
-    psoDesc.PS = { programImpl->m_pixelShader .getBuffer(), SIZE_T(programImpl->m_pixelShader .getCount()) };
+    for (auto& shaderBin : programImpl->m_shaders)
```
jsmall-nvidia, line 5290: We could validate here - for multiple definitions (say), or for if the stage is appropriate for the pipeline types.

slang-gfx.h
```diff
@@ -909,6 +909,25 @@ struct ComputePipelineStateDesc
     IShaderProgram*  program;
 };
 
+struct RayTracingPipelineFlags
```
jsmall-nvidia, line 912: Elsewhere we've done this as 

```
typedef uint32_t RayTracingPipelineFlags;
struct RayTracingPipelineFlag
{
    enum Enum : RayTracingPipelineFlags
    {
        None = 0,
        SkipTriangles = 1,
        SkipProcedurals = 2,
    };
};
```
Such that flags can be easily combined, or a single flag specified.
An issue is that 'Flags' then doesn't have type safety.

jsmall-nvidia (approved):

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: All these checks are supposed to be done in debug layer, and are omitted here to avoid overhead.

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: Yeah, that would be another validation to check in the debug layer.

PR #1915, opened 2021-07-23
Experimental DXR1.0 support in gfx.

csyonghe:
- Add `dispatchRays` command.
- Add `createRayTracingPipelineState` method to construct a D3D ray tracing state object from a linked slang program and user specified shader table.

Limitations/simplifications: no local root signature support, shader table entries contains only shader identifiers and is specified at pipeline creation time, owned by the pipeline state object.


slang-gfx.h
```diff
@@ -1191,6 +1201,17 @@ class IRayTracingCommandEncoder : public ICommandEncoder
         IAccelerationStructure* const* structures,
         AccessFlag::Enum sourceAccess,
         AccessFlag::Enum destAccess) = 0;
+
+    virtual SLANG_NO_THROW void SLANG_MCALL
+        bindPipeline(IPipelineState* state, IShaderObject** outRootObject) = 0;
+    /// Issues a dispatch command to start ray tracing worload with a ray tracing pipeline.
```
jsmall-nvidia, line 1207: worload

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5801,7 +5847,290 @@ void D3D12Device::CommandBufferImpl::RayTracingCommandEncoderImpl::memoryBarrier
...
+            {
+            }
+            virtual void setPipelineState(PipelineStateBase* pipeline) override
+            {
+                auto pipelineImpl = static_cast<RayTracingPipelineStateImpl*>(pipeline);
+                m_cmdList4->SetPipelineState1(pipelineImpl->m_stateObject.get());
+            }
+        };
+        RayTracingSubmitter submitter(m_commandBuffer->m_cmdList4);
+        if (SLANG_FAILED(_bindRenderState(&submitter, newPipeline)))
```
jsmall-nvidia, line 5879: This assumes that newPipeline is held by a ref elsewhere (else it will go out of scope when used in later cast). I'm assuming this is the case. 
That said a comment to the effect, or just making sure it is kept in scope all whilst used might make more clear the intent.

jsmall-nvidia (approved):

PR #1920, opened 2021-08-11
Further implementation of SPIRV direct emit.

csyonghe:
This change implements:
- Struct, Vector, Matrix and Unsized Array types.
- Basic arithmetic opcodes, vector construct, swizzle etc.
- getElementPtr, getElement, fieldAddress, extractField.
- SPIRV target intrinsics with SPIRV asm code in stdlib.
- RWStructuredBuffer and StructuredBuffer.
- Pointer storage class propagation.
- Control flow.


source/slang/hlsl.meta.slang
```diff
@@ -199,13 +199,15 @@ struct StructuredBuffer
         out uint numStructs,
         out uint stride);
 
-    __target_intrinsic(glsl, "$0._data[$1]") 
+    __target_intrinsic(glsl, "$0._data[$1]")
+    __target_intrinsic(spirv_direct, "%addr = 65 resultType*StorageBuffer resultId _0 _1; 61 resultType resultId %addr;")
```
tfoleyNV, line 203: I like the human-readable syntax here. I haven't looked at the implementation yet, but I can already guess from what is written here how most of it works.

My initial notes, upon looking at this:

* It might be sensible to have something like `_r` instead of `resultId`, to communicate that it is similar to `_0`, `_1`, etc.

* It might be sensible to have a uniform way to refer to the type of an operand or result, like `_t0` for the type of operand 0, and `_tr` for the type of the result.

* It might eventually make sense to allow simple identifiers for the opcode names instead of magic numbers, even if doing so requires us to have a string->id mapping table somewhere. This could always be added later, though, since we'd want to support hard-coded numbers as well.

* The `resultType*StorageBuffer` bit is probably the messiest thing there, obviously. It might be nice to have the SPIR-V "assembly" format being used here support a simple but uniform form of Lisp-y nested expressions, so that `(Ptr resultType StorageBuffer)` would be valid. Allowing nested expressions would seemingly allow this example to be written as:

```
61 resultType resultId (65 (Ptr resultType StorageBuffer) resultId _0 _1)
```

The implied rules here would be that a `(...)`-enclosed range represents another SPIR-V instruction that will be emitted into the flow. The logical evaluation order would be left-to-right depth-first (no ambiguity). For any nested instruction with an opcode that represents a type, etc. it would go through the existing deduplication we'd have to do for SPIR-V types anyway, and would be emitted into the relevant section (instead of directly into a function body).

Again, none of these "need" to happen, as there are no actual problems with what's here now.

source/slang/slang-capability-defs.h
```diff
@@ -55,6 +55,7 @@ SLANG_CAPABILITY_ATOM0(GLSL,     glsl,      Concrete,TargetFormat,0)
 SLANG_CAPABILITY_ATOM0(C,        c,         Concrete,TargetFormat,0)
 SLANG_CAPABILITY_ATOM0(CPP,      cpp,       Concrete,TargetFormat,0)
 SLANG_CAPABILITY_ATOM0(CUDA,     cuda,      Concrete,TargetFormat,0)
+SLANG_CAPABILITY_ATOM0(SPIRV_DIRECT, spirv_direct, Concrete, TargetFormat, 0)
```
tfoleyNV, line 58: There's some interesting questions about how capabilities like `vulkan`, `glsl`, `spirv`, and `spirv_direct` should relate. I don't think we need to resolve them right now, but it is probably necessary to resolve them in the long run in case we need overload resolution to pick between GLSL->SPIRV and direct-to-SPIRV code, while also still having some functions that are logically shared between the two cases.

source/slang/slang-emit-base.h
```diff
@@ -0,0 +1,29 @@
...
+#include "../core/slang-basic.h"
+
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+#include "slang-ir-restructure.h"
+
+namespace Slang
+{
+
+class SourceEmitterBase : public RefObject
```
tfoleyNV, line 14: Should this be `EmitterBase` since for SPIR-V we aren't emitting source? I guess I'm jumping ahead, but I'm just *assuming* that's how it will be used...

source/slang/slang-ir-insts.h
```diff
@@ -2734,6 +2735,14 @@ IRTargetSpecificDecoration* findBestTargetDecoration(
         IRInst*         val,
         CapabilityAtom  targetCapabilityAtom);
 
+inline IRTargetIntrinsicDecoration* findBestTargetIntrinsicDecoration(
+    IRInst* inInst,
+    CapabilitySet const& targetCaps)
+{
+    return as<IRTargetIntrinsicDecoration>(findBestTargetDecoration(inInst, targetCaps));
+}
```
tfoleyNV, line 2743: I wonder if the function name here is a little inaccurate to what the function does. Probably merits a comment describing what it means.

To me, this seems to mean "find the best target decoration for the given instruction on a target with the given capabilities, and *if* that is a target-intrinsic decoration, return it."

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+void legalizeIRForSPIRV(
+    SPIRVEmitSharedContext* context,
+    IRModule* module,
+    DiagnosticSink* sink)
+{
+    SLANG_UNUSED(sink);
+    propagateStorageClass(context, module);
+}
```
tfoleyNV, line 256: Would it be possible to make this pass not be explicitly SPIR-V related? Could we define a Slang-side `enum` for storage classes and emit a propagation pass like this for them without any reference to SPIR-V or its rules, and then once that it done just translate the Slang storage classes to the SPIR-V ones as part of emission?

I guess the question I'm asking here is whether there are multiple targets that might need/want storage class propagation and, if so, whether they can all share a common propagation pass and `enum` for the storage classes?

Realistically, it is probably best to leave this logic exactly as it is here until we actually have a second use case and worry about the details then.

source/slang/slang-ir-spirv-snippet.h
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    struct ASMOperand
+    {
+        ASMOperandType type;
+
+        // The value of the spv word when type is `SpvWord`, or
+        // the reference name when type is `ObjectReference`
+        // (e.g. an argument reference (_1) has `content` == 1).
+        int content;
```
tfoleyNV, line 43: Should this be a `SpvWord`?

tfoleyNV (approved):
This looks like good progress toward our long-term goals for SPIR-V emission.

PR #1921, opened 2021-08-12
Add GLSL450 intrinsics to SPIRV direct emit.

csyonghe:
This change adds SPIRV-direct-emit intrinsics for extended instructions defined in `GLSL.core.450`.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -200,14 +200,14 @@ struct StructuredBuffer
         out uint stride);
 
     __target_intrinsic(glsl, "$0._data[$1]")
-    __target_intrinsic(spirv_direct, "%addr = 65 resultType*StorageBuffer resultId _0 _1; 61 resultType resultId %addr;")
+    __target_intrinsic(spirv_direct, "%addr = 65 resultType*StorageBuffer resultId _0 const(int, 0) _1; 61 resultType resultId %addr;")
     T Load(int location);
     T Load(int location, out uint status);
 
     __subscript(uint index) -> T
     {
         __target_intrinsic(glsl, "$0._data[$1]")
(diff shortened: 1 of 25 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -720,6 +721,7 @@ T abs(T x);
 __generic<T : __BuiltinSignedArithmeticType, let N : int>
 __target_intrinsic(hlsl)
 __target_intrinsic(glsl)
+__target_intrinsic(spirv_direct, "12 resultType resultId glsl450 fi(4,5) _0")
```
tfoleyNV, line 724: To state the semi-obvious, we are reaching the point with this pseudo-assembler syntax where:

1. We need to start documenting the syntax and all these options in a top-down way

2. We need to make sure we are happy with its design choices because we are starting to write more and more code in this dialect and it will be hard to change later

tfoleyNV: While it might not be exactly what we want in all ways, we could certainly consider the SPIRV-Tools assembler syntax as a decent starting point: https://github.com/KhronosGroup/SPIRV-Tools/blob/master/docs/syntax.md

In particular, I wonder if we need to follow their model where "raw" words injected into the stream should use a `!` prefix so that plain constants can use a simpler syntax.

I'm also keen to work out our rules for what general nesting in this syntax should look like, and I really do think that something Lisp-y may be the best option to keep things simple to parse while supporting flexible nesting.

A hugely disruptive alternative (which I don't think we want to pursue right now) would be to shift our focus from these `__target_intrinsic` modifiers and instead try to support some kind of inline `asm` blocks inside function definitions. We could then use some kind of target-dependent `if` or `switch` to pick the `asm` block to execute for each target (or to include an ordinary definition as pure Slang code as a fallback).

tfoleyNV (approved):

PR #1929, opened 2021-09-03
Fix crash: dynamic dispatch of generic interface method.

csyonghe:
When generating dynamic dispatch code for an generic interface method whose function signature doesn't reference any generic parameters, the logic for computing the `IRType` of the generic method incorrectly omits the generic parameters, causing `lower-generics` to produce invalid code.

This change fixes this issue and includes a test case to cover the scenario.


csyonghe:
Merging now for Falcor.

PR #1931, opened 2021-09-07
Fix `spSetDumpIntermediates`.

csyonghe:
This change fixes a bug that `spSetDumpIntermediates` is not effective when using the new `IComponentType` based compilation API. The new logic ensures the setting is conserved through `TargetRequest` from the `Linkage` backing the legacy `SlangCompileRequest` object.


tfoleyNV (approved):
This is fine as a stopgap, but eventually we need to start putting together a more clear model for where flags like this should live in the software architecture.

PR #1933, opened 2021-09-09
`reinterpret` and 16-bit value packing.

csyonghe:
This changes adds support for 16-bit values in `packAnyValue`/`unpackAnyValue` operations.

Also adds a `reinterpret<T,U>` built-in operation that translates into a `packAnyValue` followed by a `unpackAnyValue` operation.

The packing rules for `packAnyValue` is currently implements as:
1. 32-bit and 64bit types are aligned to 4 byte boundaries.
2. 16-bit values are aligned to 2-byte boundaries.
3. Arrays/structs/composite types are laid out the same way as if individual elements are directly included in the outter type, i.e. no special alignments for arrays and structs.


source/slang/slang-ir-any-value-marshalling.cpp
```diff
@@ -541,4 +651,97 @@ namespace Slang
...
+        case kIROp_StructType:
+        {
+            auto structType = cast<IRStructType>(type);
+            for (auto field : structType->getFields())
+            {
+                offset = _getAnyValueSizeRaw(field->getFieldType(), offset);
+                if (offset < 0) return offset;
+            }
+            return offset;
+        }
```
tfoleyNV, line 716: Hmm... I'm a bit concerned if we end up having so many distinct pieces of layout logic that are *nearly* the same:

1. The AST-level layout rules for structured buffers
2. The IR-level type layout logic (implemented in `ir-type-layout`)
3. This logic, which defines a de facto distinct set of layout rules for "any-value" packing

The intention is that (2) is exactly compatible with (1) on all targets, but we don't currently test/enforce that (so it could be a source of bugs).

Looking at the code here, I'm worried because the struct case here isn't actually compatible with (2) or (1). It is going to be mostly the same, but then fail to match in a few cases, and I could see those creating issues for users.

This is especially notable because we don't have AST-level layout logic for this new format under (3), and that means that there is no way for a client of the reflection API (like `gfx`) to fill in a buffer using that layout.

I think this is a fine solution for now, but in the long run we really need the packing of data in an any-value to exactly match the layout for (1) and (2) above, so that we have a consistent notion of what "natural" layout is, and use that in as many cases as possible.

tfoleyNV: Note that a very real problem with this kind of approach to structure layout is that by not computing an alignment for the whole struct and applying that, you end up in a situation where different fields using the same struct type may lay out differently because they started at differently aligned offsets.

This logic having to loop through all fields or all array elements also bloats the time complexity of this code. If it was about to rely on the existing IR type layout it could take advantage of the  way those memo-cache the layout information on the IR nodes to speed up subsequent queries.

source/slang/slang-ir-inst-defs.h
```diff
@@ -548,105 +548,105 @@ INST(HighLevelDeclDecoration,               highLevelDecl,          1, 0)
     INST(OutputTopologyDecoration,          outputTopology,         1, 0)
     INST(PartitioningDecoration,            partioning,             1, 0)
     INST(DomainDecoration,                  domain,                 1, 0)
-    INST(MaxVertexCountDecoration,          maxVertexCount,         1, 0)
-    INST(InstanceDecoration,                instance,               1, 0)
-    INST(NumThreadsDecoration,              numThreads,             3, 0)
+INST(MaxVertexCountDecoration, maxVertexCount, 1, 0)
+INST(InstanceDecoration, instance, 1, 0)
+INST(NumThreadsDecoration, numThreads, 3, 0)
```
tfoleyNV: Seems like an over-zealous automatic formatting pass/tool grabbed this file.

tfoleyNV (approved):
There's some formatting stuff that should probably be fixed before commit.

I think we can check this in with its current layout rules for now, but I strongly believe we should unify the layout rules for structured buffers, `union`s, `reinterpret`, and any-value packing to all use the "natural" layout rules that the current `ir-type-layout` file tries to implement.

If we really need packing and reinterpreting to use different logic, then we should talk about that.

source/slang/slang-ir-any-value-marshalling.cpp
csyonghe: Right, the compiler behavior is overly complex due to this specific kind of layout. We should change them altogether to use (2) in a separate PR.

PR #1935, opened 2021-09-11
Bring heterogeneous-hello-world back up to date.

dsiher:
* Reintroduced heterogeneous-hello-world into the premake

* No longer uses compiled bytecode for entry point, instead a loadModule
call is hardocoded with the slang file name.

* Entry point is, similarly, hardcoded for now.

* Added a bypass to slang-legalize-types for an unneeded GPUForeach check


csyonghe:
You need to run `.\premake.bat vs2019` to update the solution/project files.
Please also change the absolute include path to avoid build failures on CI servers.

csyonghe:
Can you remove the old `experimental/heterogeneous-hello-world` or modify `premake5.lua` to point to `experimental/heterogeneous-hello-world` instead of `example/heterogeneous-hello-world` ?

csyonghe (approved):

PR #1936, opened 2021-09-13
Bug fix in 16bit type emit, vk validation error fix.

csyonghe:
In PR #1933 I made a mistake by changing our 16-bit value HLSL emit logic to use `min16float` instead of just `half` while not aware of `float16_t` and real `half` types have already been introduced to dxc for `sm_6_2` as well as in `glslang`. That change also breaks existing `bit-cast-16bit` test.

This PR fixes the issue by reverting that change. Also fixes the gfx/vulkan for a validation layer error.
Also implemented `bit_cast` between 16bit integer types and `float16` in GLSL emit, which enabled the `pack-any-value-16bit` test to run on Vulkan.


csyonghe:
Merging without review to get out a release ASAP.

PR #1939, opened 2021-09-15
Fix incorrect WIN32 macros and missing Windows.h inclusion

pablode:


csyonghe:
Thanks for contributing the fix. It is probably better to use `_MSC_VER` together with `_InterlockedIncrement` intrinsic here instead of the `windows.h` header.

csyonghe (approved):

PR #1940, opened 2021-09-15
Allow D3D12Device to use an existing device handle

lucy96chen:
Changes:
- Added a new field `existingDeviceHandle` to `IDevice::Desc`, which is 0 by default to indicate that a device does not yet exist and will need to be created
- Modified `D3D12Device::initialize` to check if `desc.existingDeviceHandle` is 0. If not 0, then it stores the value in `m_device` instead of creating a new device through `_createDevice`

@csyonghe Here are the changes I made based on the conversation and notes from yesterday.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -1424,8 +1424,13 @@ public:
         DeviceType deviceType = DeviceType::Default;
         // Name to identify the adapter to use
         const char* adapter = nullptr;
-        // The device handle. If 0, a device will need to be created.
-        uint64_t existingDeviceHandle = 0;
+        // Device handles (if they already exist)
+        struct existingDeviceHandles
+        {
+            // For D3D12, this only contains a single value for the ID3D12Device.
+            // For Vulkan, the first value is the VkInstance and the second is the VkDevice.
+            uint64_t values[2] = { 0 };
+        } existingDeviceHandles;
         // Number of required features.
         int requiredFeatureCount = 0;
         // Array of required feature names, whose size is `requiredFeatureCount`.
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -1424,6 +1424,8 @@ public:
         DeviceType deviceType = DeviceType::Default;
         // Name to identify the adapter to use
         const char* adapter = nullptr;
+        // The device handle. If 0, a device will need to be created.
+        uint64_t existingDeviceHandle = 0;
         // Number of required features.
         int requiredFeatureCount = 0;
         // Array of required feature names, whose size is `requiredFeatureCount`.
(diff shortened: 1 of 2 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4287,37 +4287,45 @@ Result D3D12Device::initialize(const Desc& desc)
...
-    if (!m_deviceInfo.m_adapter)
+        // Set the device
+        m_device = m_deviceInfo.m_device;
+    }
+    else
     {
-        // Couldn't find an adapter
-        return SLANG_FAIL;
+        // Set the existing device stored in desc
+        m_device = (ID3D12Device*)desc.existingDeviceHandle;
```
csyonghe, line 4326: I think you'll also need to assign to pointer to `m_deviceInfo.m_device`.

slang-gfx.h
```diff
@@ -1424,6 +1424,8 @@ class IDevice: public ISlangUnknown
         DeviceType deviceType = DeviceType::Default;
         // Name to identify the adapter to use
         const char* adapter = nullptr;
+        // The device handle. If 0, a device will need to be created.
```
csyonghe, line 1427: This is OK for now, but for vulkan we may need to pass in both a `VkInstance` and `VkDevice`, so perhaps we can use something like:
```
uint64_t existingDeviceHandles[2];
```
Or even better we can define a `struct` type:
```
struct ExistingDeviceHandle
{
      uint64_t values[2];
};
```

csyonghe (approved):
Looks good to me. We can either implement the suggested changes in this PR or in your next change.

tools/gfx/d3d12/render-d3d12.cpp
lucy96chen: Yeah, I missed that. Done.

slang-gfx.h
lucy96chen: Done. I'll leave the helpers you mentioned for later.

slang-gfx.h
```diff
@@ -1424,6 +1424,13 @@ class IDevice: public ISlangUnknown
         DeviceType deviceType = DeviceType::Default;
         // Name to identify the adapter to use
         const char* adapter = nullptr;
+        // Device handles (if they already exist)
+        struct existingDeviceHandles
```
csyonghe, line 1428: We do have a convention that type names are BigCamel. So here should be ExistingDeviceHandle.

csyonghe (approved):
Would be good to follow the BigCamel naming convention for type names.

slang-gfx.h
lucy96chen: Derp, fixed.

csyonghe (approved):

PR #1942, opened 2021-09-21
Support for existing device/instance handles in Vulkan

lucy96chen:
Changes:
- Added static helpers for constructing `ExistingDeviceHandles`
- Added getter functions for retrieving device handles from `ExistingDeviceHandles` when they exist
- Changed `D3D12Device::initialize()` to utilize new getter functions instead of directly indexing into `existingDeviceHandles.values`
- Added support for existing device handles to `VkDevice::initialize()` and changed the dtor to no longer destroy the device and instance if they were passed in

@csyonghe Changes to the Vulkan end based on what I remember from our discussion, let me know if I missed something big.

### DIFF
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -53,7 +53,7 @@ public:
         kMaxDescriptorSets = 8,
     };
     // Renderer    implementation
-    Result initVulkanInstanceAndDevice(bool useValidationLayer);
+    Result initVulkanInstanceAndDevice(Desc::ExistingDeviceHandles handles, bool useValidationLayer);
     virtual SLANG_NO_THROW Result SLANG_MCALL initialize(const Desc& desc) override;
     virtual SLANG_NO_THROW Result SLANG_MCALL createTransientResourceHeap(
         const ITransientResourceHeap::Desc& desc,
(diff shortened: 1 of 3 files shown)


tools/gfx/vulkan/render-vk.cpp
```diff
@@ -5221,101 +5222,108 @@ VkPipelineShaderStageCreateInfo VKDevice::compileEntryPoint(
...
-    uint32_t apiVersionsToTry[] = {VK_API_VERSION_1_2, VK_API_VERSION_1_1, VK_API_VERSION_1_0};
-    for (auto apiVersion : apiVersionsToTry)
+    else
     {
-        applicationInfo.apiVersion = apiVersion;
-        if (m_api.vkCreateInstance(&instanceCreateInfo, nullptr, &instance) == VK_SUCCESS)
-        {
-            break;
-        }
+        instance = *(VkInstance*)handles.getVkInstance();
```
csyonghe, line 5326: I don't think we need the indirection here. The user is supposed to give us an VkInstance represented in a uint64. So this just need to be (VkInstance)handles.getVkInstance().

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -5332,49 +5340,59 @@ Result VKDevice::initVulkanInstanceAndDevice(bool useValidationLayer)
...
-        {
-            // Device not found
-            return SLANG_FAIL;
-        }
+
+        physicalDevice = physicalDevices[selectedDeviceIndex];
+    }
+    else
+    {
+        physicalDevice = *(VkPhysicalDevice*)handles.getVkPhysicalDevice();
```
csyonghe, line 5387: Same here, we do not need the indirection here.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -5601,19 +5619,27 @@ Result VKDevice::initVulkanInstanceAndDevice(bool useValidationLayer)
...
-    deviceCreateInfo.ppEnabledExtensionNames = deviceExtensions.getBuffer();
+        deviceCreateInfo.enabledExtensionCount = uint32_t(deviceExtensions.getCount());
+        deviceCreateInfo.ppEnabledExtensionNames = deviceExtensions.getBuffer();
+
+        if (m_api.vkCreateDevice(m_api.m_physicalDevice, &deviceCreateInfo, nullptr, &m_device) != VK_SUCCESS)
+            return SLANG_FAIL;
+    }
+    else
+    {
+        m_device = *(VkDevice*)handles.getVkDevice();
```
csyonghe, line 5640: No indirection needed here.

csyonghe (commented):
Looks good to me overall, the only comment is to remove the handle indirection.

tools/gfx/vulkan/render-vk.cpp
lucy96chen: Oh, right. Fixed.

csyonghe (approved):

PR #1943, opened 2021-09-21
Add gfx unit testing framework.

csyonghe:
This  changes adds a unit-testing framework for gfx tests.

The basic idea is to define a series of unit test functions in a dll, then `slang-test` can have the ability to find such dlls, query the test functions and run the tests one by one.

Each unit test dll can have their own dependencies.

Next step is to migrate existing unit-tests to a separate dll within this new framework.

Further improvements can be done in the future, such as:
1. Allow `slang-test` to automatically discover unit test dlls and run them.
2. Forward more complete testing options from `slang-test` to each unit test function.
3. Allow individual unit test functions to declare its test category to allow it to be filtered by `slang-test`.


jsmall-nvidia (approved):
Looks good.

PR #1945, opened 2021-09-23
Move existing unit tests to a standalone dll.

csyonghe:
This change migrates existing unit tests into a standalone `slang-unit-test-tool.dll`.

Renames `gfx-test-tool` to `gfx-unit-test-tool`.

Forwards the `TestReporter` object to unit test dlls so they can make use of `SLANG_CHECK` utility macros.


tools/unit-test/slang-unit-test.h
```diff
@@ -3,36 +3,73 @@
...
+};
+
+enum class TestMessageType
+{
+    Info,                   ///< General info (may not be shown depending on verbosity setting)
+    TestFailure,           ///< Describes how a test failure took place
+    RunError,              ///< Describes an error that caused a test not to actually correctly run
+};
+
+class ITestReporter
```
jsmall-nvidia, line 22: I guess this interface isn't strictly 'COM', but to avoid ABI issues specifying SLANG_NO_THROW and SLANG_MCALL would be a good idea on the methods.

tools/unit-test/slang-unit-test.h
```diff
@@ -3,36 +3,73 @@
...
-    SlangResult name(slang::UnitTestContext* context); \
-    slang::UnitTestRegisterHelper _##name##RegisterHelper(#name, name); \
-    SlangResult name(slang::UnitTestContext* context)
-}
+void name(UnitTestContext* unitTestContext); \
+UnitTestRegisterHelper _##name##RegisterHelper(#name, name); \
+void name(UnitTestContext* unitTestContext)
+
+#define SLANG_CHECK(x) getTestReporter()->addResultWithLocation((x), #x, __FILE__, __LINE__);
+#define SLANG_CHECK_ABORT(x)                                                                     \
```
jsmall-nvidia, line 69: A nitpik - but macros that return without that being part of the name are perhaps a little opaque. When I read 'abort' my brain interprets as terminate.

jsmall-nvidia (approved):
Looks good.

PR #1946, opened 2021-09-24
Add existing device handle support unit test

lucy96chen:
Changes:
- As title

@csyonghe Not quite ready yet as the test is failing on the `loadShaderProgram` call on line 22. Could you take a look?


tools/gfx-unit-test/existing-device-handle-test.cpp
```diff
@@ -0,0 +1,169 @@
...
+        if (SLANG_FAILED(createDeviceResult))
+        {
+            return SLANG_IGNORE_TEST;
+        }
+
+        IDevice::NativeHandle handle = {};
+        GFX_CHECK_CALL_ABORT(device->getNativeHandle(&handle));
+        Slang::ComPtr<IDevice> testDevice;
+        IDevice::Desc testDeviceDesc = {};
+        switch (api)
```
csyonghe, line 131: Can just be `IDevice::Desc testDeviceDesc = deviceDesc;` and then modify `testDeviceDesc.existingDeviceHandle`, so that you don't need to duplicate line 131-149 again.

csyonghe (approved):
Looks good. Only a minor suggestion to reduce code duplication.

csyonghe (approved):

PR #1951, opened 2021-09-27
Fix ARM64 detection for MSVC

englercj:
Small one to fix ARM64 not being detected correctly for MSVC building for windows ARM64.

Would be awesome to see windows arm64 binaries attached to releases as well, but I wasn't sure what needed to change to make that happen.


csyonghe (approved):

csyonghe:
Thanks for your fix, @englercj ! Our releases are built by GitHub actions, you can take a look at `.github/workflows/release-windows.yml` to see you can make it build win/arm64 binaries. However I am not sure if GitHub has server nodes that runs ARM Windows.

englercj:
> However I am not sure if GitHub has server nodes that runs ARM Windows.

I don't think they do, but generally you build ARM64 binaries on x86 systems (cross-compile) anyway so it should be fine. I'll take a peek and see if I can add it and verify this change.

csyonghe:
That makes sense. Thank you for taking a look into this!

englercj:
I made some local modifications to add `aarch64` for windows and for the most part it compiles, the catch is that it tries to run some tools as part of the build (`slang-generate.exe`,  `slang-embed.exe`, `slang-cpp-extractor.exe`) but since these are built for ARM64 they can't be run on the x86_64 machine.

Some options:

1. Build for x64, then use those tools to build the ARM64 version.
2. Build for ARM64 natively on ARM64 machines (GitHub does not provide these AFAIK).
3. Commit the generated headers only regenerate them for Win32/x64 builds.

englercj:
I added my premake changes so you can try it out if you want.

csyonghe (dismissed):
Looks good, we can merge this in if it passes all tests.

csyonghe:
Yeah, I think we could probably modify the script we use to build for arm64 by building x64 first, run and get the generated files, then run arm64 build next. We've been trying to avoid committing generated files since it creates consistency problems and bloats the size of changes.

englercj:
OK I made the changes necessary to get ARM64 builds working using x64 tools. The github action workflow builds the x64 tools first then runs aarch64. Locally however, you just have to build x64 first yourself for aarch64 to pass.

There was a bunch of trailing whitespace in the premake5 file so that got trimmed causing a lot of whitespace changes. Sorry about that, probably best to view with whitespace ignored in the diff.

The last issue is that `slang-binaries` doesn't include a windows aarch64 version of the necessary binaries. Once it does, this may be ready.

csyonghe:
Thanks for pulling through this! We really appreciate your contribution. Would you update slang-binaries as part of this change, or do you want to land what we have here first, assuming we temporarily disable the aarch64 build until slang-binaries are ready? Note that we don't necessarily need everything in slang-binaries for a release. We don't need premake.exe if we don't care about a native build. Mostly we just need slang-glslang.dll but that is just a cache to speed up our builds, and we can build it along with slang by setting a config in premake.lua. That we I think we can just set the Slang repo to build slang-glslang every time when we build for aarch64 and get away with not having a prebuilt slang-glslang.dll for a while if that helps landing this change sooner.

csyonghe:
@englercj See premake5.lua line 78 to enable slang-glslang.dll as part of build.

englercj:
> Would you update slang-binaries as part of this change, or do you want to land what we have here first, assuming we temporarily disable the aarch64 build until slang-binaries are ready?

I'd rather do it before merging this so we can make sure all the things come together correctly.

> we can build it along with slang by setting a config in premake.lua

I'll check this out to make sure it compiles correctly. Look out for an aarch64 PR for slang-binaries though.

csyonghe:
Sounds good! I'll make sure to merge your changes there.

csyonghe (approved):

englercj:
I can't actually view the failures on Team City, is there something here I broke?

csyonghe:
Looks like it is more of a server config issue than an actual problem with this PR. We will investigate tomorrow. If everything works, will merge.

PR #1953, opened 2021-09-29
Removing exceptions from core/compiler-core

jsmall-nvidia:
The purpose of this PR is to be able to build core and compiler-core without exception support. This is desirable for a variety of reasons - but the motivation here is so that core/compiler-core can be linked against libraries without exceptions. LLVM on linux being the main issue.

The PR leaves a single file that has exceptions - slang-token-reader.cpp/.h. This could be made to not use exceptions, but for the moment it doesn't get in the way of LLVM integration, so is just excluded from core build when building LLVM.

In core something like exception support is needed, because it contains DiagnosticSink which with throw when a fatal error is hit. To work around this the PR introduces slang-signal.cpp/.h which replaces the signal mechanism that was in slang-string. It uses an enum to identify the 'signal'. This file can be compiled with exceptions disabled - in that case it won't throw, but will print the message and terminate. 

There were exceptions used elsewhere, where something more 'assert-like' would perhaps be more suitable. This is now handled via SLANG_ASSERT_FAILURE (or others defined in slang-signal.h), which can be passed a message. It remains in release, and will throw when exceptions are available, or terminate if they are not.

That the StreamWriter/Reader and unicode handling was somewhat refactored around this change. Some issues around character encoding, which aren't addressed in this PR are discussed in #1954.


source/core/slang-char-encode.h
```diff
@@ -0,0 +1,200 @@
...
+
+        /// Get the encoding type
+    CharEncodeType getEncodingType() const { return m_encodingType; }
+
+        /// Given some bytes determines a character encoding type, based on the initial bytes.
+        /// If can't be determined will assume UTF8.
+        /// Outputs the offset to the first non mark in outOffset
+    static CharEncodeType determineEncoding(const Byte* bytes, size_t bytesCount, size_t& outOffset);
+
+        /// Get the
```
csyonghe, line 183: incomplete comment

csyonghe (approved):

PR #1955, opened 2021-09-30
VK Resource Barrier

kyaoNV:
- Added resource barrier API to `IResourceCommandEncoder`
- Implemented in VK, stubbed in other implementations

No integration in this PR. Test coming in another PR, or can be added here if requested.

### DIFF
--- a/tools/gfx/debug-layer.h
+++ b/tools/gfx/debug-layer.h
@@ -317,7 +317,8 @@ public:
     virtual SLANG_NO_THROW void SLANG_MCALL
         uploadBufferData(IBufferResource* dst, size_t offset, size_t size, void* data) override;
     virtual SLANG_NO_THROW void SLANG_MCALL writeTimestamp(IQueryPool* pool, SlangInt index) override;
-
+    virtual SLANG_NO_THROW void SLANG_MCALL textureBarrier(ITextureResource* texture, ResourceState src, ResourceState dst) override;
+    virtual SLANG_NO_THROW void SLANG_MCALL bufferBarrier(IBufferResource* buffer, ResourceState src, ResourceState dst) override;
 public:
     DebugCommandBuffer* commandBuffer;
     bool isOpen = false;
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -3976,7 +3976,7 @@ public:
             }
         }
 
-        static VkAccessFlagBits calcAccessFlagBits(ResourceState state)
+        static VkAccessFlagBits calcAccessFlags(ResourceState state)
         {
             switch (state)
             {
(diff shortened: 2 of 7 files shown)


tools/gfx/debug-layer.h
```diff
@@ -317,7 +317,8 @@ class DebugResourceCommandEncoder : public UnownedDebugObject<IResourceCommandEn
     virtual SLANG_NO_THROW void SLANG_MCALL
         uploadBufferData(IBufferResource* dst, size_t offset, size_t size, void* data) override;
     virtual SLANG_NO_THROW void SLANG_MCALL writeTimestamp(IQueryPool* pool, SlangInt index) override;
-
+    virtual SLANG_NO_THROW void SLANG_MCALL textureBarrier(ITextureResource* texture, ResourceState src, ResourceState dst) override;
```
csyonghe, line 320: I think we should consider making the interface similar to D3D12, where we have a `struct ResourceBarrier`, and a `resourceBarrier(uint32_t count, ResourceBarrier* barriers)` function to reduce the amount of function calls needed for a large number of barriers.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -3943,6 +3943,118 @@ class VKDevice : public RendererBase
...
+                return VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL;
+            case ResourceState::Present:
+                return VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
+            default:
+                assert(!"Unsupported");
+                return VK_IMAGE_LAYOUT_UNDEFINED;
+            }
+        }
+
+        static VkAccessFlagBits calcAccessFlags(ResourceState state)
```
csyonghe, line 3979: This looks good, but we probably need to verify against the spec that we are producing compatible access flags and stage flags.

tools/gfx/debug-layer.h
csyonghe: Another concern is whether or not we want a resource to be in a state that is a combination of `ResourceState` values, and that will certainly complicate the translation logic.

csyonghe (approved):

csyonghe:
You need to provide a placeholder implementation for CUDA as well. By default the solution files are set up to not include CUDA platform. To enable it, you need to run `.\premake.bat vs2019 --enable-cuda=true` locally.

csyonghe (approved):

tools/gfx/debug-layer.h
kyaoNV: Yes, I think that interface would be better in this case because it's a closer mapping to VK/DX. I can make that change.

csyonghe: Let's change the interface then. We can do it in the next PR.

PR #1960, opened 2021-10-02
Get native handles for TextureResource and BufferResource

lucy96chen:
Changes:
- Added `getNativeHandle()` to `ITextureResource` and `IBufferResource`
- Implemented `getNativeHandle()` for Vulkan and D3D12 as well as in `renderer_shared.cpp`
- Added two unit tests testing the above functionality
- Made small modifications to previously added unit tests

@csyonghe


tools/gfx-unit-test/existing-device-handle-test.cpp
```diff
@@ -100,7 +100,7 @@ namespace gfx_test
     {
         if ((api & context->enabledApis) == 0)
```
csyonghe, line 101: You can use the new runTestImpl function to eliminate all the device creation logic here. See the current compute-smoke test for an example.

csyonghe (approved):
Looks good, you may consider simplify the code by making use of the latest runTestImpl function.

tools/gfx-unit-test/existing-device-handle-test.cpp
lucy96chen: I'm not seeing that function anywhere. Where is it supposed to be?

csyonghe: Ah, never mind, I just realized that it is still in my local branch and I haven't committed it to master yet.

lucy96chen: Ah, okay. I'll make this change in a later PR once it is then.

PR #1968, opened 2021-10-08
Generalize heterogenous code emit

dsiher:
* Pass in entryPoint and Slang file name to loadShaderProgram via compiler

* Add Slang file name to entryPoint decoration

* Modify premake to automatically compile shader.slang -> shader.cpp for hetereogeneous-hello-world


dsiher:
I think the only test that fails is due to adding the file's directory to the entry point decoration, the comparison doesn't expect that extra string in the decoration. Unless something else is failing in the TeamCity builds, that is the only problem I found in the reports I could see, and it is expected.

PR #1982, opened 2021-10-16
Expanded gfx::Format to include additional formats

arquelion:
Changes:
- `gfx::Format` now includes every format shared between D3D and Vulkan, including BC and SRGB formats. All currently unsupported formats are listed in `unsupported-formats.md` under `docs/gfx-user-guide`.
- `D3DUtil::getMapFormat()`, `VulkanUtil::getVkFormat()`, and `FormatInfoMap::FormatInfoMap()` have been updated to include their corresponding mappings.
- `GFX_FORMAT()` has been modified to track the number of bytes per block and the number of pixels per block (instead of only tracking bytes per pixel) in order to accommodate the BC formats and now includes size information for all newly added formats. To that end, all other functions which call `GFX_FORMAT()` have also been modified.
- The process for creating a texture in D3D and Vulkan have been slightly updated to account for 4x4 blocks when the texture is in any BC format, including adding a new global helper function `gfxIsCompressedFormat()`.
- `createTextureView()` now correctly respects the format in the view's Desc struct if the texture's format is typeless.
- Multiple unit tests have been added covering all added formats.

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -40,9 +40,9 @@ jobs:
           export SLANG_BINARY_ARCHIVE=slang-${SLANG_TAG}-${SLANG_OS_NAME}-${SLANG_ARCH_NAME}.zip
           export SLANG_BINARY_ARCHIVE_TAR=slang-${SLANG_TAG}-${SLANG_OS_NAME}-${SLANG_ARCH_NAME}.tar.gz
           echo "creating zip"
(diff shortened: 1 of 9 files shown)


lucy96chen:
> Also, please include a documentation (can be just in the description of this PR) on all the remaining unsupported format, and the reason they are not yet supported for future reference.

There's a comment at the very top of the Format enum list with all the currently omitted formats. I used the numerical values listed in dxgiformat.h though, so I can change those to the actual formats if you'd prefer.

PR #1983, opened 2021-10-18
Attempt to get CI consistently passing

jsmall-zzz:
This is on top of previously merged PR that disabled swiftshader.

It's not clear, but it seems like something related to swiftshader and/or gfx unit tests may be causing issues on CI testing.


csyonghe:
We may not need this temporary fix if #1984 fixes the issues. I suggest we merge in #1984 first and if we are still seeing problems.

PR #2003, opened 2021-11-03
Add interface for new gfx features.

csyonghe:
This change adds serveral new features to gfx header.

Implementations are left empty for now.


slang-gfx.h
```diff
@@ -1314,6 +1399,18 @@ class IRenderCommandEncoder : public ICommandEncoder
     virtual SLANG_NO_THROW void SLANG_MCALL draw(UInt vertexCount, UInt startVertex = 0) = 0;
     virtual SLANG_NO_THROW void SLANG_MCALL
         drawIndexed(UInt indexCount, UInt startIndex = 0, UInt baseVertex = 0) = 0;
+    virtual SLANG_NO_THROW void SLANG_MCALL drawIndirect(
+        uint32_t maxDrawCount,
+        IBufferResource* argBuffer,
+        uint64_t argOffset,
```
tfoleyNV, line 1405: Minor question: should  there be a utility `BufferAndOffset` type to capture this idiom since it appears a few times?

slang-gfx.h
```diff
@@ -1345,6 +1449,21 @@ class IResourceCommandEncoder : public ICommandEncoder
...
         size_t srcOffset,
         size_t size) = 0;
+    virtual SLANG_NO_THROW void SLANG_MCALL copyTexture(
+        ITextureResource* dst,
+        ITextureResource::SubresourceRange dstSubresource,
+        ITextureResource::Offset3D dstOffset,
+        ITextureResource* src,
+        ITextureResource::SubresourceRange srcSubresource,
+        ITextureResource::Offset3D srcOffset,
+        ITextureResource::Size extend) = 0;
```
tfoleyNV, line 1459: I think this wants to be `extent` and not `extend`.

slang-gfx.h
```diff
@@ -1282,6 +1362,11 @@ class IRenderCommandEncoder : public ICommandEncoder
         return rootObject;
     }
 
+    // Sets the current pipeline state and bind an existing root shader object.
+    virtual SLANG_NO_THROW Result SLANG_MCALL bindPipelineAndRootObject(
+        IPipelineState* state,
+        IShaderObject* rootObject) = 0;
```
tfoleyNV, line 1368: The biggest question(s) I have are all around this operation and what its semantics are.

I assume that `rootObject` needs to have been allocated using `createMutableRootShaderObject`, and this operation should be an error if it wasn't.

Am I correct to assume that while a pre-existing root shader object like this is bound any changes made to the object will automatically be mirrored to the active pipeline state (used by the next draw/dispatch/whatever)? That would make sense given how the existing root shader object case works, but I'd want to confirm.

slang-gfx.h
```diff
@@ -1333,7 +1430,14 @@ class IComputeCommandEncoder : public ICommandEncoder
         SLANG_RETURN_NULL_ON_FAIL(bindPipeline(state, &rootObject));
         return rootObject;
     }
+
+    // Sets the current pipeline state and bind an existing root shader object.
+    virtual SLANG_NO_THROW Result SLANG_MCALL bindPipelineAndRootObject(
+        IPipelineState* state,
+        IShaderObject* rootObject) = 0;
```
tfoleyNV, line 1437: For the compute case specifically, it seems redundant to have to specify both the pipeline state *and* a root shader object. The root shader object would have been created from a specific `IShaderProgram`, and pretty much the only things the `IPipelineState` stores for the compute case is the `IShaderProgram`.

I guess this is really just me getting to the point that maybe we don't need `IPipelineState`s for compute or, if we do, there could be some streamlined/simplified options where people can dispatch compute just using the `IShaderProgram` (or a pre-existing root object) more directly.

tfoleyNV (dismissed):

slang-gfx.h
csyonghe: Yes, that is the intention here.

slang-gfx.h
csyonghe: This will need to be done in a separate pass to get rid of all compute pipeline states.

slang-gfx.h
csyonghe: We can add it, but using it across the entire header would be a very pervasive change so I'd rather not do it now.

slang-gfx.h
csyonghe: Fixed.

tfoleyNV (approved):

PR #2005, opened 2021-11-04
Allow buffers to be shared between D3D12 and CUDA

lucy96chen:
Changes:
- Introduced a new `InteropHandle` struct and accompanying `InteropHandleAPI` enum containing a handle and the API the handle originated from, which is intended to be used by all methods that interact with either native or shared handles
- Added the `getSharedHandle()` method to `IResource`, which returns a Windows system handle for the resource that can then be shared between multiple APIs, and added destructors to D3D12's `BufferResourceImpl` and `TextureResourceImpl` which are responsible for closing the shared handle if it exists - This PR contains a tested implementation for D3D12 ONLY.
- Added `createTextureFromNativeHandle()` and `createBufferFromNativeHandle()` to `IDevice`, which creates a buffer or texture resource using the provided handle - This PR contains a tested implementation of the buffer version of these methods for D3D12 ONLY. The texture version has an implementation in D3D12 as well but has not yet been tested for correctness.
- Added `createBufferFromSharedHandle()` to `IDevice` and implemented it for `CUDADevice`, which creates a buffer resource using the provided system handle - This PR contains a tested implementation that shares a buffer from D3D12 to CUDA ONLY.
- Provided a proper implementation for `CUDADevice::getNativeHandle()`
- Added several new tests testing the aforementioned implementations
- Moved `getNativeHandle()` for `IBufferResource` and `ITextureResource` up a level into `IResource` and renamed it to `getNativeResourceHandle()`
- Renamed `IDevice::getNativeHandle()` to `IDevice::getNativeDevicesHandles()` and propagated this change where necessary

**IMPORTANT: `existingDeviceHandleCUDA` and `sharedHandleD3D12ToCUDA` are temporarily disabled in this PR. TeamCity reports failures from both tests (failure to create a transient resource heap for one, incorrect buffer values for the other) despite both tests passing locally.**

This is the first PR addressing interops between D3D12, Vulkan, and CUDA in gfx. As such, it is incomplete, largely only containing implementations for D3D12, and the in the case of shared resources, only containing a full implementation for sharing a D3D12 buffer with CUDA. There are still MANY implementations missing (e.g. sharing in the other direction from CUDA), the next of which to be addressed will be sharing a (buffer) resource from Vulkan to CUDA.


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -243,12 +267,24 @@ class D3D12Device : public RendererBase
...
+            return SLANG_OK;
+        }
+
+        virtual SLANG_NO_THROW Result SLANG_MCALL getSharedHandle(SharedHandle* outHandle) override
         {
-            *outHandle = (uint64_t)m_resource.getResource();
+            ComPtr<ID3D12Device> pDevice;
+            auto pResource = m_resource.getResource();
+            pResource->GetDevice(IID_PPV_ARGS(pDevice.writeRef()));
+            SLANG_RETURN_ON_FAIL(pDevice->CreateSharedHandle(pResource, NULL, GENERIC_ALL, (L"Texture" + std::to_wstring(id)).c_str(), outHandle));
```
csyonghe, line 286: We need to have a policy on who is responsible for freeing the shared handle. My suggestion is that the Buffer/Texture resource itself is responsible for that.
To do that the BufferResource/TextureResource implementation need to have a member that stores the shared handle, and we call `CreateSharedHandle` only when that handle hasn't been created yet. If it has we just return the previously created handle without calling CreateSharedHandle again. We also need to make sure to call `CloseHandle` on the shared handle when the resource is destroyed.

tools/gfx/debug-layer.cpp
```diff
@@ -712,9 +756,14 @@ DeviceAddress DebugBufferResource::getDeviceAddress()
     return baseObject->getDeviceAddress();
 }
 
-Result DebugBufferResource::getNativeHandle(NativeHandle* outHandle)
+Result DebugBufferResource::getNativeResourceHandle(NativeResourceHandle* outHandle)
 {
-    return baseObject->getNativeHandle(outHandle);
+    return baseObject->getNativeResourceHandle(outHandle);
```
csyonghe, line 761: Add `SLANG_GFX_API_FUNC;`

tools/gfx/debug-layer.cpp
```diff
@@ -729,9 +778,14 @@ ITextureResource::Desc* DebugTextureResource::getDesc()
     return baseObject->getDesc();
 }
 
-Result DebugTextureResource::getNativeHandle(NativeHandle* outHandle)
+Result DebugTextureResource::getNativeResourceHandle(NativeResourceHandle* outHandle)
 {
-    return baseObject->getNativeHandle(outHandle);
+    return baseObject->getNativeResourceHandle(outHandle);
```
csyonghe, line 783: Add `SLANG_GFX_API_FUNC;`

tools/gfx/debug-layer.cpp
```diff
@@ -729,9 +778,14 @@ ITextureResource::Desc* DebugTextureResource::getDesc()
...
+}
+
+Result DebugTextureResource::getSharedHandle(SharedHandle* outHandle)
+{
```
csyonghe, line 787: Add `SLANG_GFX_API_FUNC;`

tools/gfx/renderer-shared.cpp
```diff
@@ -75,7 +75,14 @@ IResource* BufferResource::getInterface(const Slang::Guid& guid)
...
+Result BufferResource::getNativeResourceHandle(NativeResourceHandle* outHandle)
+{
+    outHandle->handle = NULL;
+    outHandle->type = NativeHandleType::Unknown;
+    return SLANG_OK;
+}
+
+Result BufferResource::getSharedHandle(SharedHandle* outHandle)
 {
     *outHandle = NULL;
```
csyonghe, line 87: We probably want to return SLANG_FAIL here.

csyonghe, line 87: Use `nullptr` instead of `NULL`.

tools/gfx/renderer-shared.cpp
```diff
@@ -92,7 +99,14 @@ IResource* TextureResource::getInterface(const Slang::Guid& guid)
 SLANG_NO_THROW IResource::Type SLANG_MCALL TextureResource::getType() { return m_type; }
 SLANG_NO_THROW ITextureResource::Desc* SLANG_MCALL TextureResource::getDesc() { return &m_desc; }
 
-Result TextureResource::getNativeHandle(NativeHandle* outHandle)
+Result TextureResource::getNativeResourceHandle(NativeResourceHandle* outHandle)
+{
+    outHandle->handle = NULL;
+    outHandle->type = NativeHandleType::Unknown;
+    return SLANG_OK;
```
csyonghe, line 106: return SLANG_FAIL;

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4871,7 +4930,24 @@ Result D3D12Device::createBufferResource(const IBufferResource::Desc& descIn, co
...
 
     const D3D12_RESOURCE_STATES initialState = buffer->m_defaultState;
-    SLANG_RETURN_ON_FAIL(createBuffer(bufferDesc, initData, srcDesc.sizeInBytes, buffer->m_uploadResource, initialState, buffer->m_resource));
+    SLANG_RETURN_ON_FAIL(createBuffer(bufferDesc, initData, srcDesc.sizeInBytes, buffer->m_uploadResource, initialState, buffer->m_resource, descIn.isShared));
+
+    returnComPtr(outResource, buffer);
+    return SLANG_OK;
+}
+
+Result D3D12Device::createBufferFromNativeHandle(IResource::NativeResourceHandle handle, const IBufferResource::Desc& srcDesc, IBufferResource** outResource)
```
csyonghe, line 4939: No need to do this now, but we should plan to implement `createBufferFromSharedHandle` and `createTextureFromSharedHandle` by using `ID3D12Device::ID3D12Device::OpenSharedHandle`.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -225,9 +238,20 @@ class D3D12Device : public RendererBase
...
+        }
+
+        virtual SLANG_NO_THROW Result SLANG_MCALL getSharedHandle(SharedHandle* outHandle) override
         {
-            *outHandle = (uint64_t)m_resource.getResource();
+            ComPtr<ID3D12Device> pDevice;
+            auto pResource = m_resource.getResource();
+            pResource->GetDevice(IID_PPV_ARGS(pDevice.writeRef()));
+            SLANG_RETURN_ON_FAIL(pDevice->CreateSharedHandle(pResource, NULL, GENERIC_ALL, (L"Buffer" + std::to_wstring(id)).c_str(), outHandle));
+            id++;
```
csyonghe, line 254: I think we can just make `id` a static local variable in this function instead of a static member. Just declare `static uint32_t id = 0;` in the function.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3792,6 +3829,9 @@ class D3D12Device : public RendererBase
     bool m_nvapi = false;
 };
 
+uint32_t D3D12Device::BufferResourceImpl::id = 0;
```
csyonghe, line 3832: As mentioned, you can delete this and declare it as a local static variable in `getSharedHandle`.

tools/gfx/cuda/render-cuda.cpp
```diff
@@ -892,6 +895,12 @@ class CUDADevice : public RendererBase
     String m_adapterName;
 
 public:
+    Result getNativeHandle(NativeHandle* outHandle)
```
csyonghe, line 898: virtual SLANG_NO_THROW Result SLANG_MCALL getNativeHandle(...)

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4003,9 +4043,11 @@ Result D3D12Device::createBuffer(const D3D12_RESOURCE_DESC& resourceDesc, const
         heapProps.CreationNodeMask = 1;
         heapProps.VisibleNodeMask = 1;
 
+        D3D12_HEAP_FLAGS flags = isShared ? D3D12_HEAP_FLAG_SHARED : D3D12_HEAP_FLAG_NONE;
```
tfoleyNV, line 4046: Coding style thing (this is *not* a "coding convention" rule or anything, just a thing I've learned over years): very often when you have flags like this, the reality is that you end up wanting to add/include different flags under different conditions and trying to write it all as a big conditional with `?:` actually harms readability.

My strong preference when I end up having to write something like this is to break it down into multiple lines such that it can be a lot more future-proof:

```
D3D12_HEAP_FLAGS flags = D3D12_HEAP_FLAG_NONE;
if(isShared) flags |= D3D12_HEAP_FLAG_SHARED;
```

For now the effect is the exact same, but the difference is that the alternative way of writing it can easily handle a future where we find out that sometimes we need to allocate heaps with the new "nifty" flag:

```
if(isNifty) flags |= D3D12_HEAP_FLAG_NIFTY;
```

Adding support for a new flag becomes a one-line change that doesn't disrupt the existing logic, instead of requiring the next programmer who touches this to sweat about breaking apart a `?:` thing and wonder what to replace it with. I always like writing code that will minimize the diffs I have to write in the future.

tools/gfx/d3d12/render-d3d12.cpp
lucy96chen: That's the plan after we get buffer sharing from Vulkan to CUDA working. `createTextureFromSharedHandle` also needs to be added for CUDA, but the process seems to be very similar to sharing a buffer (with an extra step added to handle mipmaps) and seemed a little lower priority at the moment.

tools/gfx/d3d12/render-d3d12.cpp
lucy96chen: I'm pretty sure I did this the first time I added this flag because that's how it was done everywhere else I looked and then forgot on the redo. Makes sense though.

slang-gfx.h
```diff
@@ -436,6 +436,14 @@ class IResource: public ISlangUnknown
         CountOf,
     };
 
+    enum NativeHandleType
```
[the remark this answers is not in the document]
lucy96chen: It did occur to me yesterday that this might be the same thing as `DeviceType` since it stands to reason that the source API for a native handle ought to be the same as the API of the source device that's providing it (and `NativeHandleType` as is looks suspiciously like a subset of `DeviceType`). In hindsight, this enum is poorly named and should probably be something like `HandleSourceAPI`.

I do wonder if maybe it'd be better to at least merge native and shared handles into a single struct though. Perhaps like
```
struct ResourceHandles
{
    uint64_t nativeHandle = 0;
    void* sharedHandle = nullptr;
    DeviceType srcApi = DeviceType::Unknown;
    Type type = Type::Unknown; // This is the enum class under IResource
```
This would let us track all the handles for a single resource in one place, and the getters would just fill in the appropriate fields. Though having written this, I think I should probably give `SharedHandle` the same treatment as `NativeResourceHandle`...

tools/gfx/d3d12/render-d3d12.cpp
lucy96chen: > we can just pass in nullptr for the name argument and it will create an unnamed shared handle, with no chance of collision with other handles for other allocations

I think I added it initially to ensure unique names that would also indicate if a resource was a buffer or texture. But if that's the case and I also make the changes I mention above to `SharedHandle`, then yeah this won't be needed at all.

slang-gfx.h
```diff
@@ -445,9 +453,19 @@ class IResource: public ISlangUnknown
...
         int cpuAccessFlags = 0;     ///< Combination of Resource::AccessFlag
+        bool isShared = false;
+    };
+
+    struct NativeResourceHandle
+    {
+        uint64_t handle; // A pointer to the resource.
+        NativeHandleType type;
     };
+    typedef void* SharedHandle;     // System handle for the underlying resource that can be shared between APIs.
```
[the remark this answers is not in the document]
lucy96chen: Oh, this is basically what I wrote in one of my responses above. Ha. I think at minimum I'll give `SharedHandle` the same treatment as `NativeResourceHandle`.

slang-gfx.h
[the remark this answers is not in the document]
csyonghe: I am thinking of something like
```
enum class InteropHandleType
{
     Unknown, D3D12, Vulkan, CUDA, Win32, FileDescriptor
};

struct InteropHandle
{
     InteropHandleType type;
     uint64_t handleValue = 0;
};
```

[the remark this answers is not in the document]
lucy96chen: So everything would basically store two instances of this struct, one for the native handle and one for the shared handle? I still think that it should also track whether the handle is for a buffer/texture/etc though, especially if we also take Tess's suggestion to have a single unified struct for handles instead of having resource handles use a separate struct from device handles.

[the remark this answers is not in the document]
csyonghe: I didn't quite get on why we need to track if a handle is for buffer or texture? I'd assume it is a user's error error if they attempt to use the shared handles with mismatching types.

tools/gfx/d3d12/render-d3d12.cpp
lucy96chen: Added for D3D12 BufferResource, will gradually add more as the rest of the shared handle implementations come in

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: Let's get rid of the string name generation logic here.

csyonghe (dismissed):
The changes looks pretty clean. Thanks for addressing all the comments!

slang-gfx.h
```diff
@@ -445,9 +453,19 @@ class IResource: public ISlangUnknown
...
 
     virtual SLANG_NO_THROW Type SLANG_MCALL getType() = 0;
+    virtual SLANG_NO_THROW Result SLANG_MCALL getNativeResourceHandle(NativeResourceHandle* outHandle) = 0;
```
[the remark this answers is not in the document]
lucy96chen: I'll make a note of this somewhere and we can revisit later

slang-gfx.h
```diff
@@ -1648,24 +1658,34 @@ class IDevice: public ISlangUnknown
...
 
+        static NativeHandle fromCUDAHandle(int64_t device)
+        {
+            NativeHandle handles = {};
+            handles.values[0] = device;
+            return handles;
+        }
+
         // The following functions provide a way of getting handles from values.
         uint64_t getD3D12Device() const { return values[0]; }
```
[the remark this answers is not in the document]
lucy96chen: Same as above, made a note and revisit later

tfoleyNV (approved):

PR #2008, opened 2021-11-10
Add support for buffer sharing from Vulkan to CUDA

lucy96chen:
Changes:
- Implemented `BufferResource::getSharedHandle()` and `BufferResource::createBufferFromNativeHandle` in render-vk.cpp
- Modified `Buffer::init()` and `initVulkanInstanceAndDevice()` to include additional code paths and add required Vulkan extensions, respectively, if the buffer resource being created is shareable
- Modified `createBufferFromSharedHandle()` in CUDA to account for the handle's source API when filling in `externalMemoryHandleDesc.type`
- Added two additional tests that fully test creating a shareable buffer in Vulkan, getting the shared handle, and creating a buffer using that handle in CUDA as well as creating a buffer using a native Vulkan handle

**IMPORTANT: This PR includes a couple hacky solutions to deal with CUDA tests which fail during CI that will need to be removed later once better solutions (likely fences/sempahores and Linux support) are implemented. Namely, an additional call to `compareComputeResult` needs to be made for both buffer sharing tests to ensure synchronization between devices, and several tests are currently wrapped with either `SLANG_WIN64` OR `SLANG_WINDOWS_FAMILY` macros due to containing only 64-bit and Windows-specific code segments.**

This PR addresses buffer sharing from Vulkan to CUDA ONLY. It does not currently go back the other direction nor does it allow for texture sharing.


csyonghe (dismissed):
Looks good to me.

tfoleyNV (approved):

csyonghe (approved):

PR #2009, opened 2021-11-10
Interprocess communication via pipes

jsmall-nvidia:
The main intent here was to allow communication to a subprocess via pipes. Functionality was already available in ProcessUtil to execute an executable and capture the stdout/stderr. The additional functionality desired here is to be able to treat an executable as a 'server' and communicate with via stdin/stdout/stderr. To do so we make a Process object type that holds streams to communicate and the current state of the underlying process.

The initial problem here was that by default pipes act in a blocking manner. This lead to the 'reflect' test on test-proxy locking up when both the client and server block.

To work around this problem the read streams non blocking - via using mechanisms to check the underlying streams state before reading. Doing so allows the client or server to incrementally read information. Writes can still block, so care is required.

For the initial use scenario this should be enough as the client and server can be seen as working in lockstep. Client writes (which might block), server reads until it gets the 'command'. Performs the command it then writes the result (which could block), but the client is reading. In essence client server is either reading or writing.

For a more complex scenario with (say) JSON RPC where initiating and getting a result can be out of step (via sequence numbers), this scenario isn't necessarily the case. If reads and writes are 'complete', if we start reading, we continue to read until we have the complete 'packet'. If we do this, we can operate without worrying about blocking leading to deadlock.

There is a performance aspect to the read not blocking - in that a simple implementation can just poll in a tight loop, consuming cycles. This wasn't a problem with blocking, as if there wasn't any data the thread would just block. To allow yielding if data is not received Process::sleepCurrentThread is added. This is not as efficient as being able to (say) block with a timeout.    

Note also that the previous implementation had a timeout, that could be used to handle an infinite loop. This is not implemented here yet, but could/can be added pretty simply using a timeout measured via a system clock.

* Split ProcessUtil into Process/ProcessUtil. Process is the platform specific functionality and interface/ProcessUtil holds platform independent functionatlity.
* Added StreamUtil 
* Split CommandLine into own source file
* Simplified use of CommandLine around directory/name style (where extension is sometimes needed)
* Added BufferedReadStream - for simplifying incremental use of a Stream of data
* Added 'flush' to stream interface
* Added exeDirectory to UnitTestContext
* Added some test scenarios to test-proxy for checking process/crashing behavior
* Removed using threads on windows stream handling


source/core/slang-process.h
```diff
@@ -0,0 +1,73 @@
...
+        };
+    };
+
+        /// Get the stream for the type
+    Stream* getStream(StreamType type) const { return m_streams[Index(type)]; }
+    int32_t getReturnValue() const { return m_returnValue;  }
+
+        /// True if the process has terminated
+    virtual bool isTerminated() = 0;
+        /// Blocks until the process has completed
```
tfoleyNV, line 42: Probably not the most important, but we almost want to clarify for stuff like this that it blocks the calling thread (but not the entire process) until the subprocess terminates.

source/core/slang-process.h
```diff
@@ -0,0 +1,73 @@
...
+    virtual void waitForTermination() = 0;
+
+
+        /// The quoting style used for the command line on this target. Currently just uses Space,
+        /// but in future may take into account platform sec
```
tfoleyNV, line 47: I'm confused what "platform sec" is here.

source/core/slang-process.h
```diff
@@ -0,0 +1,73 @@
...
+    static StringEscapeHandler* getEscapeHandler();
+
+        /// Get the suffix used on this platform
+    static UnownedStringSlice getExecutableSuffix();
+
+        /// Create a process using the executable/args defined from the commandLine
+    static SlangResult create(const CommandLine& commandLine, Process::Flags flags, RefPtr<Process>& outProcess);
+
+        /// Sleep the current thread for time specified in milliseconds. 0 indicates to OS ok to yield this thread.
+    static void sleepCurrentThread(Index timeInMs);
```
tfoleyNV, line 57: I suppose we'd eventually want this in `Thread` or `ThreadUtils` rather than `Process`.

tfoleyNV (approved):

source/core/slang-process.h
jsmall-nvidia: I guess that was a typo from previous impl. Probably "platform specific" quoting.

source/core/slang-process.h
jsmall-nvidia: It's a little awkward as it is. Didn't want to implement 'Thread' interface so just threw it in 'Process' for now.

PR #2019, opened 2021-11-17
gfx ShaderObject interface update, getTextureAllocationInfo()

csyonghe:
Add `IShaderObject::getSize`, `getRawData()` and `setConstantBufferOverride()` methods.

Add `IDevice::getTextureAllocationInfo()` method and its d3d12 and vulkan implementations.


lucy96chen (commented):
Seems good to me, should work fine since I just need the size of the texture.

PR #2021, opened 2021-11-18
RTTI/JSON

jsmall-nvidia:
The motivation for this change was to simplify the use of JSON-RPC. Writing functions that convert native types and JSON is tedious, error prone, and typically has poor error checking diagnostics. This PR uses RTTI to represent types that are converted to and from JSON. It adds functionality to perform check conversions and write diagnostics. 

* Added Rtti support 
* Added functionality to convert native types to and from JSON types via Rtti reflection
* Work in progress 'test-server' designed to communicate over JSON-RPC

In order to make this work needed to make List backing memory allocation use malloc/free, rather than new T[]/delete[].
It *could* be made to work with new T[] etc ,but doing so would require all types to implement functions for such an array allocator. I did attempt to get this to work with T[] style assuming that only the size of T was important (once we dtor/ctor independently). On Visual studio at least this is not the case. 

Note:

* Hand written JSON-RPC functions are currently handwritten, but will be replaced using reflection
* The client part of JSON-RPC communication is not implemented
* Not all types are yet handled - enums, 8/16 u/int, Dictionaries, can be handled subsequently


tools/slang-unit-test/unit-test-rtti.cpp
```diff
@@ -0,0 +1,74 @@
...
+    float b = 2.0f;
+    String s;
+    List<String> list;
+
+    static const StructRttiInfo g_rttiInfo;
+};
+
+} // anonymous
+
+static const StructRttiInfo _makeSomeStructRtti()
```
csyonghe, line 23: I guess it is intended to have some build tool that generates these functions in the future?

csyonghe (approved):

tools/slang-unit-test/unit-test-rtti.cpp
jsmall-nvidia: Yes it would be better if we could generate these automatically, as they are rather tedious to write. The C++ extractor could do this, but would need some way to convey some of the options appropriately.

PR #2038, opened 2021-12-02
D3D12 and Vulkan to CUDA Texture Sharing

lucy96chen:
Changes:
- Added `getSharedHandle()` to Vulkan's `TextureResourceImpl` and made a couple small changes to the `getSharedHandle()` implementation for D3D12
- Added `createTextureFromSharedHandle()` to render-cuda, which handles CUDA Surface creation from a shared handle
- Added a basic test to test texture sharing functionality from D3D12 and Vulkan to CUDA. See note below regarding testing coverage.
- Added a templated `setData()` function to `ShaderCursor` to facilitate setting simple uniform values (e.g. width, height)
- Added `compareComputeResult()` for textures to gfx-test-util. The templated version may need some tweaking in a future PR.
- Factored out the section in `CUDADevice::createTextureResource()` that determines the equivalent CUDA format of a texture into a separate `getCUDAFormat()` function. This covers more formats than it did initially but DOES NOT CURRENTLY CONTAIN ALL SUPPORTED FORMATS.
- Implemented `CUDADevice::readTextureResource()` for debugging purposes
- Renamed `get-shared-handle` to `shared-buffers-tests` to better reflect the test's contents and to distinguish it from tests for shared textures

**Note: Many supported formats are still missing on the CUDA end as are many new `CUarray_format` enums added since CUDA 10.5 (most notably, UNORM, SNORM, and BC). Changes to incorporate these as well a full suite of tests to test texture sharing with all supported formats will be coming in a later PR.**


csyonghe:
The failure on TeamCity is due to the attempt to run the new test in x86 build. We need to wrap the new unit test with `#if SLANG_PROCESSOR_X86_64` to make it run only on x64.

tools/gfx-unit-test/shared-textures-tests.cpp
```diff
@@ -0,0 +1,223 @@
...
+        sharedTextureTestImpl(srcDevice, dstDevice, context);
+    }
+#if SLANG_WIN64
+    SLANG_UNIT_TEST(sharedTextureD3D12ToCUDA)
+    {
+        sharedTextureTestAPI(unitTestContext, Slang::RenderApiFlag::D3D12, Slang::RenderApiFlag::CUDA);
+    }
+#endif
+
+#if SLANG_WINDOWS_FAMILY // TODO: Remove when Linux support is added
```
csyonghe, line 217: This needs to be SLANG_WIN64.

lucy96chen: Woops, accidentally flipped the WINDOWS_FAMILY and WIN64 checks.

csyonghe:
The two new CUDA tests are still being run on x86 build. Could you double check it is only included for x64? You can verify this locally by switching to Win32 platform.

tools/gfx-unit-test/shared-buffers-tests.cpp
```diff
@@ -107,19 +107,19 @@ namespace gfx_test
             SLANG_IGNORE_TEST;
         }
 
-        sharedHandleTestImpl(srcDevice, dstDevice, context);
+        sharedBufferTestImpl(srcDevice, dstDevice, context);
     }
-#if SLANG_WIN64
-    SLANG_UNIT_TEST(sharedHandleD3D12ToCUDA)
+#if SLANG_WINDOWS_FAMILY
```
csyonghe, line 112: This needs to be `SLANG_WIN64` since we can't run any cuda related stuff on x86.

lucy96chen: Fixed again

csyonghe:
Can you merge with top of tree?

tfoleyNV:
@csyonghe Have all your concerns from review been resolved? It would be good to land this.

csyonghe (approved):

PR #2042, opened 2021-12-02
gfx Mutable Root shader object implementation.

csyonghe:
This change implements a mutable root shader interface in gfx that allows the user to create a shader object that represents the parameters for a shader program, set and mutate it, and then when it comes to draw calls and dispatches, the user can copy contents from the mutable object into the transient root object returned by `IResourceCommandEncoder::bindPipeline()`.


tools/gfx-unit-test/root-mutable-shader-object.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "tools/gfx-util/shader-cursor.h"
+#include "source/core/slang-basic.h"
+
+using namespace gfx;
+
+namespace gfx_test
+{
+    void mutableRootShaderObjectTestImpl(IDevice* device, UnitTestContext* context)
+    {
+        Slang::ComPtr<ITransientResourceHeap> transientHeap;
```
jsmall-nvidia, line 14: Pedantic point - but Slang:: seems unrequired here as used later without.

jsmall-nvidia (approved):

PR #2048, opened 2021-12-07
gfx: D3D12 and VK Fence implementation.

csyonghe:
This change implements the `gfx::IFence` interface for d3d12 and vulkan.

On Vulkan it uses the timeline semaphore feature to simplify implementation.


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -6205,6 +6285,30 @@ Result D3D12Device::createQueryPool(const IQueryPool::Desc& desc, IQueryPool** o
...
+    RefPtr<FenceImpl> fence = new FenceImpl();
+    SLANG_RETURN_ON_FAIL(fence->init(this, desc));
+    returnComPtr(outFence, fence);
+    return SLANG_OK;
+}
+
+Result D3D12Device::waitForFences(
+    uint32_t fenceCount, IFence** fences, uint64_t* fenceValues, bool waitForAll, uint64_t timeout)
+{
+    List<HANDLE> waitHandles;
```
jsmall-nvidia, line 6299: Perhaps use a ShortList?

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -6205,6 +6285,30 @@ Result D3D12Device::createQueryPool(const IQueryPool::Desc& desc, IQueryPool** o
...
 
+Result D3D12Device::createFence(const IFence::Desc& desc, IFence** outFence)
+{
+    RefPtr<FenceImpl> fence = new FenceImpl();
+    SLANG_RETURN_ON_FAIL(fence->init(this, desc));
+    returnComPtr(outFence, fence);
+    return SLANG_OK;
+}
+
+Result D3D12Device::waitForFences(
```
jsmall-nvidia, line 6296: I'm assuming the timeout is in milliseconds - because thats what WaitForMultipleObjects takes, but it doesn't seem to be documented as a comment in the interface, or as part of the parameter name.

jsmall-nvidia (approved):
Looks good.

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: Yeah, I need to make sure the timeOut paameter means the same thing across d3d and vulkan here.

PR #2049, opened 2021-12-07
gfx Fence implementation improvements.

csyonghe:


jsmall-nvidia (approved):

jsmall-nvidia (changes requested):
Err it just occurred to me I missed something...

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -6296,14 +6298,14 @@ Result D3D12Device::createFence(const IFence::Desc& desc, IFence** outFence)
 Result D3D12Device::waitForFences(
     uint32_t fenceCount, IFence** fences, uint64_t* fenceValues, bool waitForAll, uint64_t timeout)
```
jsmall-nvidia, line 6299: Isn't the timeout here in nanoseconds (from the updated comment)? 

WaitForMultipleObjects is in milliseconds

https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3945,7 +3945,9 @@ class D3D12Device : public RendererBase
             for (uint32_t i = 0; i < fenceCount; ++i)
             {
                 auto fenceImpl = static_cast<FenceImpl*>(fences[i]);
-                m_d3dQueue->Wait(fenceImpl->m_fence.get(), waitValues[i]);
```
jsmall-nvidia, line 3948: I think? what you had here before was correct. 

That you want to move the 
```
waitValues[i] == kTimeoutInfinite ? INFINITE : waitValues[i] / 1000000);
```

Logic down to D3D12Device::waitForFences.

I suppose there is some confusion because they both have the same method name. Perhaps you want to rename this function to waitForFenceValues?

csyonghe: Good catch, this is the wrong place.

jsmall-nvidia (approved):
Looks good.

PR #2051, opened 2021-12-08
Remove `PipelineType` from gfx header.

csyonghe:
`PipelineType` is used to create a `IShaderProgram`, but this is redundant since the slang `IComponentType` already contains info about what type of shader it is.

### DIFF
--- a/examples/gpu-printing/main.cpp
+++ b/examples/gpu-printing/main.cpp
@@ -73,7 +73,6 @@ ComPtr<gfx::IShaderProgram> loadComputeProgram(slang::IModule* slangModule, char
     gGPUPrinting.loadStrings(linkedProgram->getLayout());
 
     gfx::IShaderProgram::Desc programDesc = {};
-    programDesc.pipelineType = gfx::PipelineType::Compute;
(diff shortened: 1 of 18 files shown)


jsmall-nvidia (commented):
I find it more readable at the call sites to have the/a enum instead of a bool. That the enum still exists. It could be argued that by using an enum that can hold more states than are possible, it requires checking if the enum is valid at the function. I think on balance though that is preferable.

Not a big deal - so can approve, but it's niggles me enough I don't want to without at least pointing this out.

csyonghe:
Note that this use is in an example, which is considered user code. PipelineType is not used anywhere else in the gfx interface so it doesn't make sense to define it in the gfx header. We could still define it in our example code if that is preferred.

jsmall-nvidia:
"Note that this use is in an example, which is considered user code. PipelineType is not used anywhere else in the gfx interface so it doesn't make sense to define it in the gfx header. We could still define it in our example code if that is preferred."

> We could still define it in our example code if that is preferred.

I personally think that would be preferable - will approve, and you can decide.

jsmall-nvidia (approved):

PR #2054, opened 2021-12-09
gfx: Implement remaining resource commands on D3D12.

csyonghe:
Includes: `textureBarrier`, `copyTexture`, `uploadTextureData`, `copyTextureToBuffer`, and `textureSubresourceBarrier`.

### DIFF
--- a/tools/gfx/d3d12/render-d3d12.cpp
+++ b/tools/gfx/d3d12/render-d3d12.cpp
@@ -3632,7 +3632,49 @@ public:
                 ResourceState src,
                 ResourceState dst) override
             {
-                assert(!"Unimplemented");
+                ShortList<D3D12_RESOURCE_BARRIER> barriers;
+
+                for (size_t i = 0; i < count; i++)
+                {
+                    auto textureImpl = static_cast<TextureResourceImpl*>(textures[i]);
+                    auto d3dFormat = D3DUtil::getMapFormat(textureImpl->getDesc()->format);
+                    auto textureDesc = textureImpl->getDesc();
+                    D3D12_RESOURCE_BARRIER barrier;
+                    barrier.Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
+                    if (src == dst && src == ResourceState::UnorderedAccess)
+                    {
+                        barrier.Type = D3D12_RESOURCE_BARRIER_TYPE_UAV;
+                        barrier.UAV.pResource = textureImpl->m_resource.getResource();
+                    }
+                    else
+                    {
+                        barrier.Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+                        barrier.Transition.StateBefore = D3DUtil::getResourceState(src);
+                        barrier.Transition.StateAfter = D3DUtil::getResourceState(dst);
+                        barrier.Transition.pResource = textureImpl->m_resource.getResource();
+                        auto planeCount = D3DUtil::getPlaneSliceCount(
+                            D3DUtil::getMapFormat(textureImpl->getDesc()->format));
+                        for (uint32_t planeIndex = 0; planeIndex < planeCount; planeIndex++)
+                        {
+                            for (int layer = 0; layer < textureDesc->arraySize; layer++)
+                            {
+                                for (int mip = 0; mip < textureDesc->numMipLevels; mip++)
+                                {
+                                    barrier.Transition.Subresource = D3DUtil::getSubresourceIndex(
+                                        mip,
+                                        layer,
+                                        planeIndex,
+                                        textureImpl->getDesc()->numMipLevels,
+                                        textureImpl->getDesc()->arraySize);
+                                    barriers.add(barrier);
+                                }
+                            }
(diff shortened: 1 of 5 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3673,13 +3771,100 @@ class D3D12Device : public RendererBase
...
+                    bufferImpl->m_resource.getResource()->Map(0, &mapRange, (void**)&bufferData);
+                    for (uint32_t z = 0; z < footprint.Footprint.Depth; z++)
+                    {
+                        auto imageStart = bufferData + footprint.Footprint.RowPitch *
+                                                           footprint.Footprint.Height * (size_t)z;
+                        auto srcData =
+                            (uint8_t*)subResourceData->data + subResourceData->strideZ * z;
+                        for (uint32_t row = 0; row < footprint.Footprint.Height; row++)
+                        {
+                            memcpy(
```
jsmall-nvidia, line 3853: Stride is generally used to mean the amount of typically bytes between data, but is *not* necessarily the size of the data.
The memcpy here makes the assumption the stride here does make that assumption. If I read the comment around the member it also states

> In particular, they may only support strictly positive strides.

Also supporting the idea the stride is different from the size of the data in row.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3673,13 +3771,100 @@ class D3D12Device : public RendererBase
...
+
+                    D3D12_TEXTURE_COPY_LOCATION srcRegion = {};
+                    srcRegion.Type = D3D12_TEXTURE_COPY_TYPE_PLACED_FOOTPRINT;
+                    D3D12_PLACED_SUBRESOURCE_FOOTPRINT& footprint = srcRegion.PlacedFootprint;
+                    
+                    footprint.Offset = 0;
+                    footprint.Footprint.Format = texDesc.Format;
+                    uint32_t mipLevel = D3DUtil::getSubresourceMipLevel(
+                        subresourceIndex, dstTexture->getDesc()->numMipLevels);
+                    if (extent.x != 0xFFFFFFFF)
```
jsmall-nvidia, line 3804: ITextureResource::Size doesn't seem to say anything about 0xffffffff special value? Is it something that is set at the API level?

It would benefit from a constant describing what it means (even if not part of public API).

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3809,11 +4073,55 @@ class D3D12Device : public RendererBase
...
-                SLANG_UNUSED(dst);
-                SLANG_UNIMPLEMENTED_X("textureSubresourceBarrier");
+                auto textureImpl = static_cast<TextureResourceImpl*>(texture);
+
+                if (subresourceRange.mipLevelCount == 0)
+                    subresourceRange.mipLevelCount = textureImpl->getDesc()->numMipLevels;
+                if (subresourceRange.layerCount == 0)
+                    subresourceRange.layerCount = textureImpl->getDesc()->arraySize;
+
+                auto d3dFormat = D3DUtil::getMapFormat(textureImpl->getDesc()->format);
```
jsmall-nvidia, line 4083: Theres no checking for if d3dFormat is valid. Perhaps that's okay because it will fail when calling into D3D12. Perhaps there is an argument to check prior to that point, with an assert at a minimum.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -3809,11 +4073,55 @@ class D3D12Device : public RendererBase
...
+                    barrier.Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+                    barrier.Transition.StateBefore = D3DUtil::getResourceState(src);
+                    barrier.Transition.StateAfter = D3DUtil::getResourceState(dst);
+                    barrier.Transition.pResource = textureImpl->m_resource.getResource();
+                    auto aspectMask = (int32_t)subresourceRange.aspectMask;
+                    if (subresourceRange.aspectMask == TextureAspect::Default)
+                        aspectMask = (int32_t)TextureAspect::Color;
+                    while (aspectMask)
+                    {
+                        auto aspect = (int32_t)aspectMask &
```
jsmall-nvidia, line 4104: This might benefit by being in a function, with a suitable name, such that doesn't rely on a comment to explain what it's doing.

tools/gfx/transient-resource-heap-base.h
```diff
@@ -45,6 +47,26 @@ class TransientResourceHeapBaseImpl : public TransientResourceHeapBase
         return (value + alignment - 1) / alignment * alignment;
     }
 
+    Result allocateStagingBuffer(size_t size, IBufferResource*& outBufferWeakPtr, ResourceState state)
```
jsmall-nvidia, line 50: Not in this PR, but alignUp might be better off being `(value + alignment - 1) & ~(alignment - 1)`. Assuming alignment is power of 2.

jsmall-nvidia (approved):

PR #2057, opened 2021-12-11
Fully implemented test for drawInstanced()

lucy96chen:
Changes:
- Fully implemented the test for `drawInstanced()` with instancing - This test was incomplete when the PR with draw calls was opened due to issues getting the basic pipeline working. Instancing has since been added and this test now runs as advertised.

Remaining draw tests and the implementation for Vulkan's `readTextureResource` to allow these tests to run on Vulkan are coming later.

### DIFF
--- a/tools/gfx-unit-test/draw-instanced-test.cpp
+++ b/tools/gfx-unit-test/draw-instanced-test.cpp
@@ -5,22 +5,58 @@
 #include "tools/gfx-util/shader-cursor.h"
 #include "source/core/slang-basic.h"
 
+#include <stdlib.h>
+#include <stdio.h>
+
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include "external/stb/stb_image_write.h"
+
 using namespace gfx;
 
 namespace gfx_test
 {
+    /* static */ Slang::Result writeImage(
+        const char* filename,
+        ISlangBlob* pixels,
+        uint32_t width,
+        uint32_t height)
+    {
+        int stbResult =
+            stbi_write_hdr(filename, width, height, 4, (float*)pixels->getBufferPointer());
+
+        return stbResult ? SLANG_OK : SLANG_FAIL;
+    }
+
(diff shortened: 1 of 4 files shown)


tools/gfx-unit-test/draw-instanced-test.cpp
```diff
@@ -86,8 +116,8 @@ namespace gfx_test
         renderPassDesc.renderTargetAccess = &renderTargetAccess;
         ComPtr<IRenderPassLayout> renderPass = device->createRenderPassLayout(renderPassDesc);
 
-        const int width = 2;
-        const int height = 2;
+        const int width = 256;
```
csyonghe, line 119: Any particular reason to use 256 by 256? We didn't check the value of those pixels.

tools/gfx-unit-test/graphics-smoke.slang
```diff
@@ -3,7 +3,8 @@
 // Per-vertex attributes to be assembled from bound vertex buffers.
 struct AssembledVertex
 {
-    float3	position : POSITION;
+    float3	positionA : POSITIONA;
```
csyonghe, line 6: Instead of changing this shader, which is also used by another test, can you create a new shader file to keep things clean?

lucy96chen: This shader is only being used by this test as far as I'm aware. This test was still missing actual instancing when I opened the PR for the draw calls.

tools/gfx-unit-test/draw-instanced-test.cpp
lucy96chen: To have a large enough texture for adding in the instancing bits that were missing before, mostly. The test pixel locations are taken with one pixel from each instanced triangle.

lucy96chen:
Closing this PR and opening a new one as I've finished the test for `drawIndexedInstanced`, which involves a major refactoring of the code in this file.

PR #2060, opened 2021-12-15
gfx: Add tests for instanced, indexed instanced, and indirect draw calls

lucy96chen:
Changes:
- Renamed draw-instanced-test.cpp to instanced-draw-tests.cpp and factored out all shared boilerplate code into a base test class
- Added tests for `drawInstanced()`, `drawIndexedInstanced()`, `drawIndirect()`, and `drawIndexedIndirect()`
- Moved the required command signatures for indirect draw calls into `D3D12Device` and moved initialization of these signatures into `D3D12Device::initialize()`

This PR is currently still missing the implementation of `readTextureResource()` for Vulkan, which is required for checking test results. All D3D12 tests should work correctly.

### DIFF
--- a/tools/gfx-unit-test/instanced-draw-tests.cpp
+++ b/tools/gfx-unit-test/instanced-draw-tests.cpp
@@ -62,8 +62,8 @@ namespace gfx_test
     static const int kIndexCount = 6;
     static const uint32_t kIndexData[kIndexCount] =
     {
+        0, 2, 5,
         0, 1, 2,
-        3, 4, 5,
     };
 
     const int kWidth = 256;
@@ -99,8 +99,8 @@ namespace gfx_test
         IBufferResource::Desc indexBufferDesc;
         indexBufferDesc.type = IResource::Type::Buffer;
         indexBufferDesc.sizeInBytes = kIndexCount * sizeof(uint32_t);
-        indexBufferDesc.defaultState = ResourceState::VertexBuffer;
-        indexBufferDesc.allowedStates = ResourceState::VertexBuffer;
+        indexBufferDesc.defaultState = ResourceState::IndexBuffer;
+        indexBufferDesc.allowedStates = ResourceState::IndexBuffer;
         ComPtr<IBufferResource> indexBuffer = device->createBufferResource(indexBufferDesc, &kIndexData[0]);
         SLANG_CHECK_ABORT(indexBuffer != nullptr);
         return indexBuffer;
@@ -143,8 +143,11 @@ namespace gfx_test
             this->context = context;
         }
 
-        void createRequiredDrawCallResources()
+        void createRequiredResources()
         {
+            vertexBuffer = createVertexBuffer(device);
+            colorBuffer = createColorBuffer(device);
+
             ITransientResourceHeap::Desc transientHeapDesc = {};
             transientHeapDesc.constantBufferSize = 4096;
             GFX_CHECK_CALL_ABORT(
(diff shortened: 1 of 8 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4549,6 +4527,9 @@ class D3D12Device : public RendererBase
     PFN_D3D12_SERIALIZE_ROOT_SIGNATURE m_D3D12SerializeRootSignature = nullptr;
 
     bool m_nvapi = false;
+
+    ComPtr<ID3D12CommandSignature> drawIndirectCmdSignature;
+    ComPtr<ID3D12CommandSignature> drawIndexedIndirectCmdSignature;
```
tfoleyNV, line 4532: Comment these, maybe?

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5328,6 +5309,32 @@ Result D3D12Device::initialize(const Desc& desc)
...
+        D3D12_INDIRECT_ARGUMENT_DESC args[1];
+        args[0].Type = D3D12_INDIRECT_ARGUMENT_TYPE_DRAW;
+
+        D3D12_COMMAND_SIGNATURE_DESC desc;
+        desc.ByteStride = 36;
+        desc.NumArgumentDescs = 1;
+        desc.pArgumentDescs = args;
+        desc.NodeMask = 0;
+
+        SLANG_RETURN_ON_FAIL(m_device->CreateCommandSignature(&desc, nullptr, IID_PPV_ARGS(drawIndirectCmdSignature.writeRef())));
```
tfoleyNV, line 5322: Comment this block, maybe, to describe what it is doing? I think the short version is something like:

```
// Allocate a D3D12 "command signature" object that matches the behavior
// of a D3D11-style `DrawInstancedIndirect` operation.
```

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5328,6 +5309,32 @@ Result D3D12Device::initialize(const Desc& desc)
...
+        D3D12_INDIRECT_ARGUMENT_DESC args[1];
+        args[0].Type = D3D12_INDIRECT_ARGUMENT_TYPE_DRAW_INDEXED;
+
+        D3D12_COMMAND_SIGNATURE_DESC desc;
+        desc.ByteStride = 36;
+        desc.NumArgumentDescs = 1;
+        desc.pArgumentDescs = args;
+        desc.NodeMask = 0;
+
+        SLANG_RETURN_ON_FAIL(m_device->CreateCommandSignature(&desc, nullptr, IID_PPV_ARGS(drawIndexedIndirectCmdSignature.writeRef())));
```
tfoleyNV, line 5335: This block could also use a comment, obviously.

tfoleyNV (approved):
This all looks fine to me. If we get the test failures resolved then we should be able to check this in without further review.

I don't think any of my comments here need to be addressed before this lands, but I *do* want to continue to emphasize the importance of reasonable comments in explaining what exactly we are trying to accomplish in the code.

tools/gfx-unit-test/instanced-draw-tests.cpp
```diff
@@ -0,0 +1,581 @@
...
+    void drawIndexedIndirectTestImpl(IDevice* device, UnitTestContext* context)
+    {
+        DrawIndexedIndirectTest test;
+        test.init(device, context);
+        test.run();
+    }
+
+    SLANG_UNIT_TEST(drawInstancedD3D12)
+    {
+        runTestImpl(drawInstancedTestImpl, unitTestContext, Slang::RenderApiFlag::D3D12);
```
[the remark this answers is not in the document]
lucy96chen: Along those lines, I think I'll at least add a templated `drawTestImpl()` for `drawInstancedTestImpl`, `drawIndexedInstancedTestImpl`, etc. since all of them just create the appropriate test struct, call `init()`, then call `run()`.

PR #2073, opened 2022-01-05
Draw call tests for Vulkan

lucy96chen:
Changes:
- Fully implemented vertex buffer binding and input layouts for Vulkan - This comprises the bulk of the changes in this PR, which boil down to providing all vertex stride info when the `InputLayout` is created instead of being passed to `setVertexBuffers()` at call time because Vulkan requires that strides are known by the time pipeline states are created. To handle this, some elements of the current `InputElementDesc` were split off into a new `VertexStreamDesc` struct, and a new `Desc` struct was added under `IInputLayout` which contains both of these descs as well a count of both input elements and vertex streams. All other backends have had their `createInputLayout` and `setVertexBuffers` functions updated to account for these changes.
- Implemented `readTextureResource()` for Vulkan, making it possible to check the results of the draw call tests
- Moved Vulkan index and vertex buffer binding into `setIndexBuffer` and `setVertexBuffers`, respectively, instead of doing it inside each draw call
- Fixed an issue causing the Vulkan draw tests to render upside-down due to viewport orientation
- Added support for instanced and indexed instanced draws to D3D11 as well as tests for both calls

All four draw tests pass on both D3D12 and Vulkan now, and instanced and indexed instanced draws also pass on D3D11.

### DIFF
--- a/examples/model-viewer/main.cpp
+++ b/examples/model-viewer/main.cpp
@@ -748,6 +748,7 @@ Result initialize()
         {"UV",       0, Format::R32G32_FLOAT,  offsetof(Model::Vertex, uv) },
     };
     auto inputLayout = gDevice->createInputLayout(
+        sizeof(Model::Vertex),
         &inputElements[0],
         3);
     if(!inputLayout) return SLANG_FAIL;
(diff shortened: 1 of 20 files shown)


lucy96chen:
There is still some cleanup (on the D3D11 end, in particular) before all tests will pass, but the code itself should otherwise be ready for review.

csyonghe (approved):

PR #2078, opened 2022-01-09
Enable running tests in parallel.

csyonghe:
Allows slang-test to start multiple instances of TestServer and dispatch tests to them in parallel.
Modifies `TestContext` to make most its methods thread-safe.

This significantly reduces the amount of testing time. Using 4 threads, the test time on team city reduced from 13 minutes to 5 minutes.

This change only affects TestServer spawn mode, others test spawn modes are unaffected and still runs in a single thread.

The caveat is that because tests are running in more separate processes, there is a chance that TeamCity would mistake the test report from one build configuration as from another build configuration that is concurrently running on the same physical machine. For example, if a machine is running tests for both x64 debug and x64 release, it may some time report 1-2 less tests in x64debug and show them up in the test list of x64release.

### DIFF
--- a/source/core/slang-writer.cpp
+++ b/source/core/slang-writer.cpp
@@ -71,6 +71,7 @@ ISlangUnknown* BaseWriter::getInterface(const Guid& guid)
 
 SLANG_NO_THROW char* SLANG_MCALL AppendBufferWriter::beginAppendBuffer(size_t maxNumChars)
 {
+    mutex.lock();
     m_appendBuffer.setCount(maxNumChars);
     return m_appendBuffer.getBuffer();
 }
@@ -82,6 +83,7 @@ SLANG_NO_THROW SlangResult SLANG_MCALL AppendBufferWriter::endAppendBuffer(char*
     SlangResult res = write(buffer, numChars);
     // Clear so that buffer can't be written from again without assert
     m_appendBuffer.clear();
+    mutex.unlock();
     return res;
 }
 
--- a/tools/slang-test/slang-test-main.cpp
+++ b/tools/slang-test/slang-test-main.cpp
@@ -1759,6 +1759,7 @@ static TestResult runCPPCompilerSharedLibrary(TestContext* context, TestInput& i
     DownstreamCompiler* compiler = context->getDefaultCompiler(SLANG_SOURCE_LANGUAGE_CPP);
     if (!compiler)
     {
+        std::lock_guard<std::mutex> lock(context->mutex);
         return TestResult::Ignored;
     }
 
@@ -1878,6 +1879,7 @@ static TestResult runCPPCompilerExecute(TestContext* context, TestInput& input)
     // If we are just collecting requirements, say it passed
     if (context->isCollectingRequirements())
     {
+        std::lock_guard<std::mutex> lock(context->mutex);
         context->getTestRequirements()->addUsedBackEnd(SLANG_PASS_THROUGH_GENERIC_C_CPP);
         return TestResult::Pass;
     }
--- a/tools/slang-test/slang-test-main.cpp
+++ b/tools/slang-test/slang-test-main.cpp
@@ -3461,7 +3461,9 @@ void runTestsInDirectory(
         break;
     }
     if (context->options.serverCount == 1)
+    {
         useMultiThread = false;
+    }
     if (!useMultiThread)
     {
         for (auto file : files)
@@ -3471,7 +3473,8 @@ void runTestsInDirectory(
     }
     else
     {
-        std::atomic<int> consumePtr = 0;
+        std::atomic<int> consumePtr;
+        consumePtr = 0;
         auto threadFunc = [&](int threadId)
         {
             TestReporter reporter;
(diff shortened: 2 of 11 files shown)


source/core/slang-writer.cpp
```diff
@@ -71,6 +71,7 @@ ISlangUnknown* BaseWriter::getInterface(const Guid& guid)
 
 SLANG_NO_THROW char* SLANG_MCALL AppendBufferWriter::beginAppendBuffer(size_t maxNumChars)
 {
+    mutex.lock();
```
tfoleyNV, line 74: Can we agree that the locking/unlocking stuff here maybe requires a comment or two given that:

* The `lock()` and `unlock()` calls aren't scoped or even in the same method, which makes sense given the invariants of the writer approach, but is maybe still surprising. (One might argue that the `beginAppend`/`endAppend` thing merits having some dedicated RAII type that captures the scoping and allows for an exception to flow out, perhaps aborting an append operation that was started but not completed...)
* There isn't a whole lot of thread-safety happening in the Slang codebase right now (for better or worse), and so it is probably worth making clear our intentions here: are we saying that all these basic utilities should start to have some minimal threading guarantees (and the question then becomes: what should those guarantees be?), or are we just doing it on an as-needed basis depending on which parts of our code need threading support?
* There's a final detail here that (I guess because the threading policy stuff here isn't stated), I'm left wondering if there are other operations here that should be mutex'd.

tfoleyNV: Note that an alternative viewpoint here would be that the code that is creating an `AppendBufferWriter` and then handing it out to multiple threads should probably be the one that owns the mutex, and the threaded code itself should have a scoped lock on the mutex around where it does the append operations.

I'm guessing that isn't possible right now because the actual appends are deep in existing code that wasn't intended for threading.

tfoleyNV: All of this rambling has me realizing that maybe the cleanest solution here from a design perspective is to have a wrapper `MutexControlledWriter` or similar, that wraps another arbitrary writer with this `lock()`/`unlock()` stuff.

tools/slang-test/slang-test-main.cpp
```diff
@@ -3400,43 +3421,101 @@ static bool shouldRunTest(
...
+    if (context->options.serverCount == 1)
+    {
+        useMultiThread = false;
+    }
+    if (!useMultiThread)
+    {
+        for (auto file : files)
+        {
+            processFile(file);
+        }
```
tfoleyNV, line 3484: I suspect there is a way we could architect this to avoid some of the tangled complexity that is showing up here.

Ideally the logic in the host/parent process can just be that it walks the directory hierarchy and hands of a sequence of file paths to another component that it *thinks of* as the actual test runner (which it can invoke asynchronously, knowing that status will be reported back eventually). Then we can have a variety of implementations of that test-runner component abstraction:

* A simple test runner that just takes the file paths as they come in and directly tries to execute them as a test (what `processFile` seems to do here?)

* A JSONRPC test runner that maintains a connection with a (single!) sub-process for running tests and sends JSONRPC messages to that process to get it to run the test and report back status.

* A mutli-threaded runner that adds incoming path names to a work queue so that they can be consumed by its pool of worker threads, where each worker thread owns a sub-test-runner that it will pass the path names along to as it grabs them.

I really think we are reaching a point where our test tool has had a *lot* of new functionality bolted onto it, and we haven't always stepped back to make sure that we are keeping the overall architecture of the thing in a reasonable shape.

tools/slang-test/test-context.cpp
```diff
@@ -138,24 +174,24 @@ SlangResult TestContext::_createJSONRPCConnection(RefPtr<JSONRPCConnection>& out
 
 void TestContext::destroyRPCConnection()
 {
-    if (m_jsonRpcConnection)
+    if (m_jsonRpcConnections[slangTestThreadIndex])
     {
-        m_jsonRpcConnection->disconnect();
-        m_jsonRpcConnection.setNull();
+        m_jsonRpcConnections[slangTestThreadIndex]->disconnect();
+        m_jsonRpcConnections[slangTestThreadIndex].setNull();
```
tfoleyNV, line 180: To be clear, this code here is a sign to me that we are letting these different features get tangled together in ways that are unhealthy.

tools/slang-test/test-reporter.cpp
```diff
@@ -82,9 +84,10 @@ TestReporter::TestReporter() :
     m_isVerbose = false;
 }
 
-Result TestReporter::init(TestOutputMode outputMode)
+Result TestReporter::init(TestOutputMode outputMode, bool isSubReporter)
 {
     m_outputMode = outputMode;
+    m_isSubReporter = isSubReporter;
```
tfoleyNV, line 90: This `m_isSubReporter` flag seems to be a sign that we need to step back and work on the architecture a bit. Can we separate out the functionality of the `TestReporter` into two different objects that support a 1-to-N relationship? Then the non-threaded cases is just the simple case when N is 1.

tools/slang-test/test-reporter.cpp
```diff
@@ -493,6 +505,8 @@ void TestReporter::addTest(const String& testName, TestResult testResult)
 
 void TestReporter::message(TestMessageType type, const String& message)
 {
+    static std::mutex mutex;
+    std::lock_guard<std::mutex> lock(mutex);
```
tfoleyNV, line 509: Does this mutex definitely need to be process-wide, or was it just the simplest way to inject the needed synchronization?

tfoleyNV (approved):
I've fine with this landing, but I want us to all recognize that we are accumulating technical debt in the test runner and we probably need to take as much time for cleanup work as we do for adding features.

csyonghe:
Agree that slang-test needs an in-depth redesign to stay clean and correct, and to scale further. I am still going to land this since it significantly speeds up our test time.

PR #2083, opened 2022-01-18
Fixes to GFX.

csyonghe:
* Fix binding null resource views in d3d12.
* Fix array render target view creation.
* Add support for more primitive topologies.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -371,7 +371,7 @@ enum class PrimitiveType
 
 enum class PrimitiveTopology
 {
-    TriangleList,
+    TriangleList, TriangleStrip, PointList, LineList, LineStrip
 };
 
 enum class ResourceState
@@ -668,14 +668,14 @@ public:
 
 enum class ComparisonFunc : uint8_t
 {
-    Never           = 0,
-    Less            = 0x01,
-    Equal           = 0x02,
-    LessEqual       = 0x03,
-    Greater         = 0x04,
-    NotEqual        = 0x05,
-    GreaterEqual    = 0x06,
-    Always          = 0x07,
+    Never           = 0x0,
+    Less            = 0x1,
+    Equal           = 0x2,
+    LessEqual       = 0x3,
+    Greater         = 0x4,
+    NotEqual        = 0x5,
+    GreaterEqual    = 0x6,
+    Always          = 0x7,
 };
 
 enum class TextureFilteringMode
--- a/tools/gfx/d3d12/render-d3d12.cpp
+++ b/tools/gfx/d3d12/render-d3d12.cpp
@@ -981,6 +981,9 @@ public:
                 /// The type of binding in this range.
             slang::BindingType bindingType;
 
+                /// The shape of the resource
+            SlangResourceShape resourceShape;
+
                 /// The number of distinct bindings in this range.
             uint32_t count;
 
@@ -1118,6 +1121,7 @@ public:
                         typeLayout->getBindingRangeLeafTypeLayout(r);
                     BindingRangeInfo bindingRangeInfo = {};
                     bindingRangeInfo.bindingType = slangBindingType;
+                    bindingRangeInfo.resourceShape = slangLeafTypeLayout->getResourceShape();
                     bindingRangeInfo.count = count;
 
                     switch (slangBindingType)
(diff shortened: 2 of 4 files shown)


slang-gfx.h
```diff
@@ -371,7 +371,7 @@ enum class PrimitiveType
 
 enum class PrimitiveTopology
 {
-    TriangleList,
+    TriangleList, TriangleStrip, PointList, LineList, LineStrip
```
tfoleyNV, line 374: I assume we are consciously ignoring all uncommon GS- and tessellation-specific topologies for now?

jsmall-nvidia (approved):

slang-gfx.h
```diff
@@ -668,14 +668,14 @@ class ITextureResource: public IResource
...
-    GreaterEqual    = 0x06,
-    Always          = 0x07,
+    Never           = 0x0,
+    Less            = 0x1,
+    Equal           = 0x2,
+    LessEqual       = 0x3,
+    Greater         = 0x4,
+    NotEqual        = 0x5,
+    GreaterEqual    = 0x6,
+    Always          = 0x7,
```
tfoleyNV, line 678: Unrelated to this PR: at some point this warrants a block comment about how the cases here and their numeric values are actually cleverly arranged, using a scheme that dates at least back to the origins of OpenGL (and thus probably earlier).

The cases for comparisons in OpenGL have values such that `LessEqual = Less | Equal`, etc. Bitwise operations on the comparison functions correspond to logical operations on the conditions. One can take an unknown condition `c` and form the condition that matches in the same cases except *not* on equality with `c & ~Equal` or `c & NotEqual`.

This same style of enumeration can be extended to include an additional bit for `Unordered` to handle floating-point types, which don't have a total order. I don't think any graphics APIs currently support unordered comparisons in fixed-function units, but it is a relevant thing when setting up things like comparison ops in a compiler IR.

Documenting this kind of thing isn't the most important thing in the world, but it can help pass on knowledge that might otherwise get lost in successive generations of systems. For example, the D3D11/12 equivalents of this enumeration have the same cases, in the same order, but do *not* have the same properties described above because the `D3D11_NEVER`/`D3D12_NEVER` cases start at a value of `1` instead of `0` (presumably to leave `0` as an out-of-band value).

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -981,6 +981,9 @@ class D3D12Device : public RendererBase
                 /// The type of binding in this range.
             slang::BindingType bindingType;
 
+                /// The shape of the resource
+            SlangResourceShape resourceShape;
```
tfoleyNV, line 985: I'll be interested to see how this field is used/needed. I've gone to some trouble to try to keep the type related to ranges as small/lean as I can get away with, in order to avoid accidental bloat accumulating over time. Every single binding range will pay for this new field, even for non-resource ranges.

At the same time, this is probably a premature optimization concern on my part.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -7517,6 +7524,103 @@ Result D3D12Device::RayTracingPipelineStateImpl::createShaderTables(
...
+            srvDesc.Shader4ComponentMapping = D3D12_DEFAULT_SHADER_4_COMPONENT_MAPPING;
+            d3dDevice->CreateShaderResourceView(nullptr, &srvDesc, destDescriptor);
+        }
+        break;
+    case slang::BindingType::Texture:
+        {
+            D3D12_SHADER_RESOURCE_VIEW_DESC srvDesc = {};
+            srvDesc.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
+            srvDesc.Shader4ComponentMapping = D3D12_DEFAULT_SHADER_4_COMPONENT_MAPPING;
+            switch (bindingRange.resourceShape)
```
tfoleyNV, line 7583: I see. This is interesting, because in most cases D3D doesn't care about the difference between various SRV bindings, but obviously this matters for the case of a null binding.

I wonder if we can still avoid storing the `resourceShape` on the binding range...

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -7525,8 +7629,18 @@ Result D3D12Device::ShaderObjectImpl::setResource(ShaderOffset const& offset, IR
...
+
     auto& bindingRange = layout->getBindingRange(offset.bindingRangeIndex);
 
+    if (resourceView == nullptr)
+    {
+        // Create null descriptor for the binding.
+        auto destDescriptor = m_descriptorSet.resourceTable.getCpuHandle(
+            bindingRange.baseIndex + (int32_t)offset.bindingArrayIndex);
+        return createNullDescriptor(d3dDevice, destDescriptor, bindingRange);
+    }
```
tfoleyNV, line 7642: One interesting thing here is that if the user was using `ShaderCursor` to compute the offset they are using, then the `ShaderCursor` would know the Slang type of the parameter it is trying to set, and that same type information would be *exactly* the type information we need in order to create the null descriptor here.

The challenge there is that we probably don't want to require that users always bind via `ShaderCursor`.

I suppose the Right Way to deal with the hypothetical bloat down the road is to ensure that the `BindingType` and `resourceShape` are both stored compactly, rather than burning 32 bits each (or more with padding), but that can way until later (much later).

tfoleyNV (approved):

PR #2084, opened 2022-01-19
More fixes to GFX d3d12.

csyonghe:
* D3D12 feature report on conservative raster, programmable sample position, barycentrics and ROV.
* Add QueryPool::reset.
* Fix resource setDebugName.
* Dynamically expanding GPU descriptor heap.
* Render passes without render targets (null frame buffer).

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -1393,6 +1393,7 @@ public:
     };
 public:
     virtual SLANG_NO_THROW Result SLANG_MCALL getResult(SlangInt queryIndex, SlangInt count, uint64_t* data) = 0;
+    virtual SLANG_NO_THROW Result SLANG_MCALL reset() = 0;
 };
 #define SLANG_UUID_IQueryPool                                                         \
     { 0xc2cc3784, 0x12da, 0x480a, { 0xa8, 0x74, 0x8b, 0x31, 0x96, 0x1c, 0xa4, 0x36 } }
@@ -1769,9 +1770,23 @@ public:
 class ITransientResourceHeap : public ISlangUnknown
 {
 public:
+    struct Flags
+    {
+        enum Enum
+        {
+            None = 0,
+            AllowResizing = 0x1,
+        };
+    };
     struct Desc
     {
+        Flags::Enum flags;
         size_t constantBufferSize;
+        uint32_t samplerDescriptorCount;
+        uint32_t uavDescriptorCount;
+        uint32_t srvDescriptorCount;
+        uint32_t constantBufferDescriptorCount;
+        uint32_t accelerationStructureDescriptorCount;
     };
     virtual SLANG_NO_THROW Result SLANG_MCALL synchronizeAndReset() = 0;
 
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -371,7 +371,7 @@ enum class PrimitiveType
 
 enum class PrimitiveTopology
 {
-    TriangleList,
+    TriangleList, TriangleStrip, PointList, LineList, LineStrip
 };
 
 enum class ResourceState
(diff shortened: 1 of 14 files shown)


slang-gfx.h
```diff
@@ -1769,9 +1770,23 @@ class ICommandQueue : public ISlangUnknown
...
+    };
     struct Desc
     {
+        Flags::Enum flags;
         size_t constantBufferSize;
+        uint32_t samplerDescriptorCount;
+        uint32_t uavDescriptorCount;
+        uint32_t srvDescriptorCount;
+        uint32_t constantBufferDescriptorCount;
+        uint32_t accelerationStructureDescriptorCount;
```
tfoleyNV, line 1789: I'd like us to think about how to make these size hints be relatively API-independent *and* extensible to support new descriptor types as they come along.

Nothing that needs to happen today, of course.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -2590,39 +2719,46 @@ class D3D12Device : public RendererBase
...
-                auto& table = descriptorSet.resourceTable;
+                auto descriptorHeap = &transientHeap->getCurrentViewHeap();
+                auto& table = outDescriptorSet.resourceTable;
 
                 // Allocate the table.
                 //
-                table.allocate(descriptorHeap, descriptorCount);
+                if (!table.allocate(descriptorHeap, descriptorCount))
+                {
+                    context->outOfMemoryHeap = D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV;
```
tfoleyNV, line 2736: Should handling the out-of-memory case be callback-oriented in the future? I'm not clear if it is better to let the application intercept this and try to allow the app to solider on, or whether it is always easier to have the operation fail and then just retry at the application level.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4736,25 +4916,60 @@ Result D3D12Device::PipelineCommandEncoder::_bindRenderState(Submitter* submitte
...
+            d3dheap = m_transientHeap->getCurrentSamplerHeap().getHeap();
+            m_commandBuffer->bindDescriptorHeaps();
+            break;
+        default:
+            assert(!"shouldn't be here");
+            return SLANG_FAIL;
+        }
+
+        // Try again.
+        SLANG_RETURN_ON_FAIL(rootObjectImpl->bindAsRoot(&context, rootLayoutImpl));
```
tfoleyNV, line 4951: Never mind, I see the retry logic has been implemented here. That seems nice.

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5346,13 +5561,69 @@ Result D3D12Device::initialize(const Desc& desc)
...
             {
+                // Check double precision support
                 if (options.DoublePrecisionFloatShaderOps)
                     m_features.add("double");
+
+                // Check conservative-rasterization support
+                auto conservativeRasterTier = options.ConservativeRasterizationTier;
+                if (conservativeRasterTier == D3D12_CONSERVATIVE_RASTERIZATION_TIER_3)
+                {
+                    m_features.add("conservative-rasterization-3");
```
tfoleyNV, line 5576: Unrelated to this PR: we probably need a systematic way to derive the names for features, and probably even want to have a `gfx`-internal representation of features that is richer than just strings.

Ideally we would have something like:

```
// in gfx.h

// A single enumeration that lists all the optional GPU API
// features that gfx understands, including both multi-API
// and API-specific features.
//
enum class FeatureID : int32_t { ... }

// We don't have this today, but probably need it at some
// point:
//
// A connection to the gfx library/implementation (gfx.dll)
struct IInstance : ISlangUnknown
{
    const char* getFeatureName(FeatureID id) = 0;
    FeatureID findFeatureByName(const char* name) = 0;
    ...
};
```

```
// Inside the implementation:

struct FeatureInfo
{
    // the id number and name for the feature
    FeatureID id;
    const char* name;

    // Optional: the other features that this feature depends on
    // or implies the existence of. Allows recursively walking all
    // the features transitively implied by a starting-point feature,
    // without having to encode that information as code.
    int requireCount;
    FeatureID const* requires;
};
static const FeatureInfo kFeatures[] = { ... };
```

For API-specific features, we'd want to have ones named very closely after the API-specific ones:

```
FeatureID::D3D12_CONSERVATIVE_RASTERIZATION_TIER_2
FeatureID::VK_some_vulkan_thing
```

For the API-specific ones, I'd argue that the string names should probably match the enumerant exactly, *especially* for Vulkan where string names for extensions are already needed.

Then we'd also want to have ones based on our own naming scheme that unifies across APIs:

```
FeatureID::CrossPlatformFeature
```

For this case it is harder to decide if we should have the name `cross-platform-feature` or have the external name also be `CrossPlatformFeature` for consistency... No string opinion right now.

Anyway, none of this really has anything to do with this PR.

tfoleyNV (approved):

PR #2085, opened 2022-01-20
Update heterogeneous-hello-world to use gfx changes + delete shader.cpp

dsiher:


dsiher:
Removing shader.cpp from the repo now works for Windows, but I can't figure out what is causing an issue on the Linux build. 

I think the build stops due to this error:
`make[1]: *** No rule to make target '../../../examples/heterogeneous-hello-world/shader.cpp', needed by '../../../intermediate/linux-x64/debug/heterogeneous-hello-world/shader.o'.  Stop.`
which (searching tells me) is usually due to the .cpp file not existing, but shader.cpp should be generated by the prebuild step I added (and works in Windows).

premake5.lua
```diff
@@ -151,14 +151,6 @@ newoption {
      default     = "true",
      allowed     = { { "true", "True"}, { "false", "False" } }
   }
-
```
csyonghe, line 154: Since this is still a very experimental featur, can we keep this flag and have the heterogeneous-hello-world example behind this flag?

dsiher: Sounds good, I can add it back in the next commit. Also have to clean the file structure and generate the new solution.

jsmall-nvidia (approved):

PR #2090, opened 2022-01-21
GFX: separated ShaderTable.

csyonghe:
This change adds a new `IShaderTable` object to gfx, which allows the user to specify shader entry points in a shader binding table used in a DXR1.0 ray tracing pipeline. D3D12 implementation is included in this change, but Vulkan implemenetation is left empty.

This separates out the existing shader binding table logic in RayTracingPipelineState.

### DIFF
--- a/tools/gfx/renderer-shared.h
+++ b/tools/gfx/renderer-shared.h
@@ -39,6 +39,7 @@ struct GfxGUID
     static const Slang::Guid IID_IQueryPool;
     static const Slang::Guid IID_IAccelerationStructure;
     static const Slang::Guid IID_IFence;
+    static const Slang::Guid IID_IShaderTable;
 };
 
 // We use a `BreakableReference` to avoid the cyclic reference situation in gfx implementation.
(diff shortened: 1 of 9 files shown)


tools/gfx/renderer-shared.h
```diff
@@ -1186,14 +1187,56 @@ class TransientResourceHeapBase : public ITransientResourceHeap, public Slang::C
...
         return nullptr;
     }
 };
 
+class ShaderTableBase
+    : public IShaderTable
+    , public Slang::ComObject
+{
+public:
+    Slang::List<Slang::String> m_entryPointNames;
```
jsmall-nvidia, line 1203: Perhaps benefit for comment to say the names are in the ordered contiguously over rayGenShaderCount, missShader etc.

tools/gfx/renderer-shared.h
```diff
@@ -1186,14 +1187,56 @@ class TransientResourceHeapBase : public ITransientResourceHeap, public Slang::C
...
             return static_cast<ITransientResourceHeap*>(this);
         return nullptr;
     }
 };
 
+class ShaderTableBase
+    : public IShaderTable
+    , public Slang::ComObject
+{
+public:
```
jsmall-nvidia, line 1202: Perhaps these should be stored as wstr.

PR #2094, opened 2022-01-24
Add implementations for resolveResource() to D3D12 and Vulkan backends

lucy96chen:
Changes:
- Added implementations for `resolveResource()` in both D3D12 and Vulkan backends
- Added a simple test that ensures basic functionality
- Fixed several functions in the Vulkan end that were hard setting sample count to 1, preventing multisampling from working entirely

@csyonghe

### DIFF
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -4780,15 +4780,43 @@ public:
 
             virtual SLANG_NO_THROW void SLANG_MCALL resolveResource(
                 ITextureResource* source,
+                ResourceState sourceState,
                 SubresourceRange sourceRange,
                 ITextureResource* dest,
+                ResourceState destState,
                 SubresourceRange destRange) override
             {
-                SLANG_UNUSED(source);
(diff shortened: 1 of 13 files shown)


lucy96chen:
Unsure why the Vulkan version of the `resolveResource` test fails here as it passes locally in both debug and release.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -561,6 +561,7 @@ class VKDevice : public RendererBase
         VkAttachmentReference m_depthReference;
         bool m_hasDepthStencilAttachment;
         uint32_t m_renderTargetCount;
+        VkSampleCountFlagBits sampleCount = VK_SAMPLE_COUNT_1_BIT;
```
csyonghe, line 564: m_sampleCount

csyonghe (approved):
Looks good. Minor comment on member name.

tools/gfx/vulkan/render-vk.cpp
lucy96chen: I've made this change locally, will be in the next PR I open.

PR #2097, opened 2022-01-26
Turn off edit and continue to prevent warnings.

csyonghe:


jsmall-nvidia (approved):
It's not clear from the description what these errors (and from what VS versions etc) are and so it's hard to say if this issue can be mitigated some other way. Ideally we'd be able to compile with edit&continue even if that's not the default. 

That said it doesn't seem unreasonable that edit&continue isn't the default.

tfoleyNV (approved):

PR #2098, opened 2022-01-26
Set numerous fields in VKDevice::createGraphicsPipelineState using the information in GraphicsPipelineStateDesc

lucy96chen:
Changes:
- Numerous fields in the pipeline descs filled in inside `VKDevice::createGraphicsPipelineState()` now take their values where appropriate from the corresponding fields in the `GraphicsPipelineStateDesc` provided in the arguments instead of being hard-coded
- Small naming convention fix to `sampleCount` added alongside `resolveResource`

Note: There are a number of TODOs for some of the fields as either gfx does not currently support them or Vulkan and D3D12 treat them differently. These will be addressed at a later date.

@csyonghe

### DIFF
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -119,6 +119,7 @@ public:
     virtual SLANG_NO_THROW Result SLANG_MCALL createComputePipelineState(
         const ComputePipelineStateDesc& desc,
         IPipelineState** outState) override;
+    // TODO: Add implementation for createRayTracingPipelineState() - calls VkCreateRayTracingPipelinesKHR
     virtual SLANG_NO_THROW Result SLANG_MCALL createQueryPool(
         const IQueryPool::Desc& desc,
         IQueryPool** outPool) override;
@@ -5142,6 +5143,7 @@ public:
                 _memoryBarrier(count, structures, srcAccess, destAccess);
             }
 
+            // TODO: Bind ray tracing pipeline state
             virtual SLANG_NO_THROW void SLANG_MCALL
                 bindPipeline(IPipelineState* pipeline, IShaderObject** outRootObject) override
             {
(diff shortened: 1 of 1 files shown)


tools/gfx/vulkan/render-vk.cpp
```diff
@@ -8220,6 +8223,118 @@ Result VKDevice::createMutableRootShaderObject(
...
+    case CullMode::Back:        return VK_CULL_MODE_BACK_BIT;
+    default:
+        assert(!"Unsupported cull mode");
+        return VK_CULL_MODE_NONE;
+    }
+}
+
+VkFrontFace translateFrontFaceMode(FrontFaceMode frontFaceMode)
+{
+    // TODO: May need to be reversed due to the viewport flip
```
csyonghe, line 8258: That's right, this needs to be reversed.

lucy96chen: Fixed locally, but doesn't seem to be the cause of nothing being drawn in some of those tests.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -8305,43 +8399,87 @@ Result VKDevice::createGraphicsPipelineState(const GraphicsPipelineStateDesc& in
...
+    }
+    
     VkPipelineColorBlendStateCreateInfo colorBlending = {};
     colorBlending.sType = VK_STRUCTURE_TYPE_PIPELINE_COLOR_BLEND_STATE_CREATE_INFO;
-    colorBlending.logicOpEnable = VK_FALSE;
+    colorBlending.logicOpEnable = VK_FALSE; // TODO: D3D12 has per attachment logic op (and both have way more than one op)
     colorBlending.logicOp = VK_LOGIC_OP_COPY;
-    colorBlending.attachmentCount = 1;
-    colorBlending.pAttachments = &colorBlendAttachment;
+    colorBlending.attachmentCount = (targetCount == 0) ? 1 : targetCount;
```
csyonghe, line 8466: This can just be (uint32_t)colorBlendAttachments.getCount()

csyonghe (approved):

tools/gfx/vulkan/render-vk.cpp
lucy96chen: Fixed locally, will be in next PR

PR #2108, opened 2022-01-29
Added Vulkan implementation of getFormatSupportedResourceStates()

lucy96chen:
Changes:
- Added a Vulkan-specific implementation of `getFormatSupportedResourceStates()`

@csyonghe There's no test for this yet, any ideas? Tess suggested maybe querying each format then creating a texture and a buffer using the returned resource states and then transition them between each state (and maybe doing something between transitions to ensure everything works as expected).

### DIFF
--- a/tools/gfx/vulkan/render-vk.cpp
+++ b/tools/gfx/vulkan/render-vk.cpp
@@ -7958,27 +7958,73 @@ Result VKDevice::createTextureView(ITextureResource* texture, IResourceView::Des
     return SLANG_OK;
 }
 
+// TODO: Buffer vs texture specific?
 Result VKDevice::getFormatSupportedResourceStates(Format format, ResourceStateSet* outStates)
 {
     VkFormatProperties supportedProperties;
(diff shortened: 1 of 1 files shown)


csyonghe:
Yes, Tess's suggestion is a good idea. Let's add a test that does that and use it to test the d3d12 implementation as well.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -7956,6 +7958,78 @@ Result VKDevice::createTextureView(ITextureResource* texture, IResourceView::Des
     return SLANG_OK;
 }
 
+// TODO: Buffer vs texture specific?
```
csyonghe, line 7961: We only care about images at this point.
We need to make sure we return correct bits for ShaderResource, RenderTarget, UnorderedAccess, DepthWrite, Present and TransferSrc/TransferDest states.

csyonghe (commented):
Looks good but let's make sure we return correct bits for Present state.

tfoleyNV (approved):

PR #2114, opened 2022-01-31
Fixed naming conflicts in heterogeneous-hello-world

dsiher:
Added 3 new modifiers (`__unmangled`, `__exportDirectly`, `__externLib`)

`__unmangled`  causes mangleName() to return the normal name of the decl.

`__exportDirectly` changes parent decl name concatenation behavior to use
"::" instead of "." (for Name Hint) and emits the name hint when it exists,
otherwise it emits the mangled name.

`__externLib` stops Slang from emitting the corresponding struct.

Also made necessary changes to heterogeneous-hello-world so that this new
functionality is shown off.


source/slang/slang-lower-to-ir.cpp
```diff
@@ -1986,7 +1994,12 @@ static String getNameForNameHint(
 
     StringBuilder sb;
     sb.append(parentName);
-    sb.append(".");
+    if (decl->hasModifier<__exportDirectly>()) {
+        sb.append("::");
+    }
+    else {
+        sb.append(".");
+    }
```
tfoleyNV, line 2002: I do not believe this code should be here.

As discussed earlier, this logic relates to the (ab)use of the `IRNameHintDecoration` as a way of establishing linkage between C++ and Slang, and once we stop using name hints that way, there is no need for the name hint of a symbol to change in this way.

Secondarily, even if we were to consider adding logic like this for *mangled* names, we'd run into the problem that the use of `::` as a static-lookup-in-scope operation is language-specific (basically just C++ and languages that strive to match C++). Having a bare `::` in the middle of a symbol name only helps when emitting C++ (and even then perhaps only on narrow cases), and for many other potentially-interesting output languages the emit logic would need to parse the symbol name looking for `::`s and replace them with something suitable anyway.

The use of `.` in Slang is intentional as it provides us with a single separator character for all scoped names that is supported and encouraged by the Slang language itself. The `.` also has the advantage of being a single ASCII character, so that it can quickly be scanned for in a symbol name. If the emit logic for a target *did* want to do something like map `.`s to `::`s, they could do it. In fact, when generating HLSL/GLSL-compatible symbol names from a name hint, we already replace `.` with `_`.

tfoleyNV:
Unrelated to this PR: we have a big-picture challenge that we'll need to work through where there are really two big cases that we need our "no-mangle" logic to handle, which are most (only?) visible when working with C++:

1. In one case we want a Slang declaration of `void foo(...)` to emit in C/C++ as `extern "C" void foo(...)`, and thus we want to have linkage according to the platform C ABI and its rules (which typically means `foo` is not mangled, even by the platform linker).

2. In the other case, we want a Slang declaration of `void foo(int a)` to emit in C++ as `extern void foo(int a)`, which means that it will get linkage according to the platform/compiler C++ ABI and its rules. This will mean `foo(int)` gets mangled, and that we can export both `foo(int)` and `foo(float)` without their symbol names clashing.

The current implementation and example application are using (2), but seemingly could get away with (1) as well. I suspect Slang needs to support both. It seems like (1) is the more fundamental building-block piece that we'd want to have securely in place *first*, while (2) opens up questions around stuff like `namespace`s, as arose in this PR.

I'm fine with this PR being more about (2) for now, but we clearly still have some design iteration to do.

PR #2115, opened 2022-02-01
GFX: Add API interception mechanism for pipeline creation.

csyonghe:
This allows the user application to intercept API calls to create pipeline states. This feature can be used to integrate NVAPI in the user application.

### DIFF
--- a/tools/gfx/d3d12/render-d3d12.cpp
+++ b/tools/gfx/d3d12/render-d3d12.cpp
@@ -7222,7 +7222,15 @@ Result D3D12Device::createGraphicsPipelineState(const GraphicsPipelineStateDesc&
     psoDesc.PrimitiveTopologyType = D3DUtil::getPrimitiveType(desc.primitiveType);
 
     ComPtr<ID3D12PipelineState> pipelineState;
-    SLANG_RETURN_ON_FAIL(m_device->CreateGraphicsPipelineState(&psoDesc, IID_PPV_ARGS(pipelineState.writeRef())));
+    if (m_pipelineCreationAPIDispatcher)
+    {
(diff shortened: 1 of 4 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -7222,7 +7222,15 @@ Result D3D12Device::createGraphicsPipelineState(const GraphicsPipelineStateDesc&
...
-    SLANG_RETURN_ON_FAIL(m_device->CreateGraphicsPipelineState(&psoDesc, IID_PPV_ARGS(pipelineState.writeRef())));
+    if (m_pipelineCreationAPIDispatcher)
+    {
+        SLANG_RETURN_ON_FAIL(m_pipelineCreationAPIDispatcher->createGraphicsPipelineState(
+            this, programImpl->slangProgram.get(), &psoDesc, (void**)pipelineState.writeRef()));
+    }
+    else
+    {
+        SLANG_RETURN_ON_FAIL(m_device->CreateGraphicsPipelineState(&psoDesc, IID_PPV_ARGS(pipelineState.writeRef())));
+    }
```
tfoleyNV, line 7233: It would not be a very significant cleanup, but we could in principle make the `D3D12Device` class implement the `IPipelineCreationAPIDispatcher` interface (by directly doing the `m_device->Create*` cases), and set the `m_pipelineCreationAPIDispatcher` member to `this` as a default.

tfoleyNV (approved):

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: Yeah that is a good cleanup to have.

PR #2117, opened 2022-02-03
Add gfx interop to allow more direct D3D12 usage scenarios.

csyonghe:
Add several `getNativeHandle` methods to allow using D3D12 API directly.

Add root signature override when creating a compute pipeline state.


jsmall-nvidia (dismissed):

lucy96chen (commented):
Looks good to me.

slang-gfx.h
```diff
@@ -435,11 +435,13 @@ enum class MemoryType
...
-    CUDA,
-    Win32,
-    FileDescriptor,
+    D3D12, // A D3D12 object pointer.
+    Vulkan, // A general Vulkan object handle.
+    CUDA, // A general CUDA object handle.
+    Win32, // A general Win32 HANDLE.
+    FileDescriptor, // A file descriptor.
+    DeviceAddress, // A device address.
+    D3D12CpuDescriptorHandle, // A D3D12_CPU_DESCRIPTOR_HANDLE value.
```
tfoleyNV, line 444: Unrelated to the meat of the PR:

Okay.... I've said it before and I'm going to say it again: the `InteropHandleAPI` should never have been named/described as an enumeration of different APIs. It is an enumeration of different *types* or *formats* for handles.

The cases in `InteropHandleType` (or `InteropHandleFormat`) could, if desired, be stored in a bit representation where the actual API is the high-order bits and then the low-order bits hold the API-specific type/format:

```
enum class InteropPlatform
{
    D3D12,
    Vulkan,
    Win32,
    POSIX,
    ...
};
enum class InteropHandleType
{
    D3D12_IResource = (uint(InteropPlatform::D3D12) << 16),
    D3D12_IShaderResourceView,
    D3D12_ISwapChain,
    D3D12_CpuDescriptorHandle,
    // ...

    Win32_HANDLE = (uint(InteropPlatform::Windows) << 16),
};
```

Obviously we could decide not to bother with that kind of scoping, and just have a flat enum that lists all the `<API>_<HandleType>` pairs that matter in arbitrary order. The key point is that the entries here should really all be in the form `<API>_<HandleType>`.

slang-gfx.h
```diff
@@ -720,6 +722,11 @@ class ISamplerState : public ISlangUnknown
         float                   minLOD          = -FLT_MAX;
         float                   maxLOD          = FLT_MAX;
     };
+
+    /// Returns a native API handle representing this sampler state object.
+    /// When using D3D12, this will be a D3D12_CPU_DESCRIPTOR_HANDLE.
+    /// When using Vulkan, this will be a VkSampler.
+    virtual SLANG_NO_THROW Result SLANG_MCALL getNativeHandle(InteropHandle* outNativeHandle) = 0;
```
tfoleyNV, line 729: As a small TODO kind of thing: we probably want our `getNativeHandle` operations to support a second parameter like `const InteropHandleType* requestedType = nullptr`, where the semantics are:

* If `requestedHandleType` is non-null, the back-end will determine if it *can* return a handle of the requested type and do so if it can. It will return an error otherwise.

* If `requestedType` is `nullptr`, the back-end is allowed to pick whatever handle type it prefers to return (of the ones it supports)

This would both let an application proactive ask for just the handle type it expects/needs (rather than needing to validate the type of the returned handle separate from the `getNativeHandle` call itself), and also make sure we are future-proof for cases where more than one handle type may be possible for a single object (e.g., the way our Vulkan device ends up wrapping multiple Vulkan API objects).

slang-gfx.h
```diff
@@ -1232,7 +1246,8 @@ struct GraphicsPipelineStateDesc
 
 struct ComputePipelineStateDesc
 {
-    IShaderProgram*  program;
+    IShaderProgram*  program = nullptr;
+    void* d3d12RootSignatureOverride = nullptr;
```
tfoleyNV, line 1250: It would obviously be great if we eventually start working on the "extension mechanism" that will allow `gfx` to support stuff like this without it polluting each and every `Desc` type.

(Honestly, stuff like this is why I actually don't like the `*Desc` API approach, and think that there are some things OpenGL actually got right... Oh well, I am okay with being in the minority on this line of thinking...)

tfoleyNV (approved):

slang-gfx.h
csyonghe: Yes, we should be implementing this refactor sooner than later.

slang-gfx.h
csyonghe: Agreed.

PR #2120, opened 2022-02-09
Various fixes to gfx.

csyonghe:
* Add interface for creating counter UAVs.
* Allow passing entry point components separately to create a shader program. This enables the user to specialize each entry point differently (e.g. with different type conformances).
* Make reflection data aware of entrypoint name override.
* Allow the CPU descriptor heap in d3d12device to expand when it runs out of space.
* Fixing ChunkedList implementation
* Other fixes around invalid uses of D3D12.

Improvements needed in the future:
* The CPU descriptor table allocation in D3D12 needs to be optimized. Currently if the heap runs out of space, we will allocate a new heap and always allocate new tables from the latest heap, even after the descriptor tables allocated from previous heaps are freed. This can create implicit leaking.
* Need to implement counter UAVs for Vulkan. This requires a policy being implemented in Slang on how to generate a separate shader parameter for the counter buffer if `IncrementCounter` or `DecrementCounter` methods are being called on a RWStructuredBuffer resource.


jsmall-nvidia (approved):
Looks good.

PR #2122, opened 2022-02-09
gfx: support d3d12 root parameters

csyonghe:
This change adds the ability to pass shader parameters via root descriptors.

The user has the option to specify the name of the attribute that is used to mark a shader parameter as root parameter.
GFX will treat a parameter decorated with the specified attribute as root parameter and passing it via root descriptor instead of in a descriptor table.
Root descriptors is the only way to pass in buffer views that extend beyond 4GB.
Creating an SRV/UAV whose size is 4GB will cause a D3D12 device removal. This change also include an work around that detects this case and not to actually create an SRV/UAV descriptor for large buffer views. The resulting buffer view can no longer be bound to an ordinary shader parameter, but may still be used to bind into a root shader parameter.


jsmall-nvidia (approved):
Can't say I understand the nuance around this. 
Leaving that aside seems good.

PR #2123, opened 2022-02-09
gfx: support shader record overwrite and fix QueryPool.

csyonghe:
This changes adds the ability to specify shader record overwrites that modifies a shader record in a shader table via gfx.


slang-gfx.h
```diff
@@ -1313,16 +1318,27 @@ struct RayTracingPipelineStateDesc
 class IShaderTable : public ISlangUnknown
 {
 public:
+    // Specifies the bytes to overwrite into a record in the shader table.
+    struct ShaderRecordOverwrite
+    {
+        uint32_t offset; // Offset within the shader record.
+        uint32_t size; // Number of bytes to overwrite.
+        uint8_t data[8]; // Content to overwrite.
+    };
```
tangent-vector, line 1327: I was going to ask if this should be "override" instead of "overwrite," but as I look at the details it sure seems like it really is about overwriting specific bytes.

I can imagine future extensions to this might need to:

*  make `data` a pointer rather than a fixed-size array
* add a parallel `mask` array/pointer to allow the overwrite to touch only specific bits in the shader record

I think it would be premature to add those generalizations right now, though.

slang-gfx.h
```diff
@@ -1313,16 +1318,27 @@ struct RayTracingPipelineStateDesc
...
+
     struct Desc
     {
         uint32_t rayGenShaderCount;
         const char** rayGenShaderEntryPointNames;
+        const ShaderRecordOverwrite* rayGenShaderRecordOverwrites;
```
tangent-vector, line 1333: Why wouldn't these use the same `extendedDesc` mechanism used below? I suppose it doesn't matter for now.

When we get around to doing cleanup and renaming passes, it is going to be really important to think about how we want to handle this stuff. We have the luxury of seeing how the extensibility mechanisms in VK have fared in the wild, so we can avoid some of the same mistakes (e.g., see how almost all the bit-mask stuff in VK is at or dangerously close to the maximum number of bits even *after* they expanded most of the masks to 64 bits!).

slang.h
```diff
@@ -2652,6 +2653,12 @@ namespace slang
                 index);
         }
 
+        VariableReflection* getBindingRangeVariable(SlangInt index)
```
tangent-vector, line 2656: Should this be `getBindingRangeLeafVariable`, to match `getBindingRangeLeafTypeLayout`?

tools/gfx-unit-test/gfx-test-util.cpp
```diff
@@ -202,6 +202,14 @@ namespace gfx_test
         const char* searchPaths[] = { "", "../../tools/gfx-unit-test", "tools/gfx-unit-test" };
         deviceDesc.slang.searchPathCount = (SlangInt)SLANG_COUNT_OF(searchPaths);
         deviceDesc.slang.searchPaths = searchPaths;
+
+        gfx::D3D12DeviceExtendedDesc extDesc = {};
+        extDesc.rootParameterShaderAttributeName = "root";
```
tangent-vector, line 207: Should we go ahead and codify the `[root]` attribute and turn it into a feature directly supported by Slang? It sure seems that at the point where we need dedicated support for it in `gfx`, we might as well just consider it a Slang/`gfx` feature.

tools/gfx-unit-test/root-shader-parameter.cpp
```diff
@@ -0,0 +1,135 @@
...
+        device->createMutableShaderObject(
+            slangReflection->findTypeByName("S1"), ShaderObjectContainerType::None, s1.writeRef());
+        device->createMutableShaderObject(
+            slangReflection->findTypeByName("S1"), ShaderObjectContainerType::None, s2.writeRef());
+
+        {
+            auto cursor = ShaderCursor(s1);
+            cursor["c0"].setResource(srvs[2]);
+            cursor["c1"].setResource(uavs[3]);
+            cursor["c2"].setResource(srvs[4]);
```
tangent-vector, line 88: Unrelated/random: when it comes time to add convenience features, it seems important that the `IShaderObject` interface should support convenience methods that work with `ShaderCursor`*.

In general, it would be good if a call like `someObject->setResource(<whatever>, srv)` is equivalent to:

```
auto cursor = ShaderCursor(someObject);
cursor[<whatever>].setResource(srv);
```

At the very least, that should support simple strings and indices (it is debatable whether it should support more general "path" expressions like `"a.b.c[3]"`, but it probably makes sense to allow those for convenience).

With such a convenience feature, this code could just become:

```
s1->setResource(0, srvs[2]);
s1->setResource(1, uavs[3]);
s1->setResource(2, srvs[4]);
```

The use of field indices in place of string names is something we should support and possibly even encourage for performance-minded programmers using `gfx`, since it avoids any string-based lookup while still being 100% portable across API back-ends.

tools/gfx-unit-test/root-shader-parameter.cpp
```diff
@@ -0,0 +1,135 @@
...
+            {
+                auto encoder = commandBuffer->encodeComputeCommands();
+                auto root = encoder->bindPipeline(pipelineState);
+                root->copyFrom(rootObject, transientHeap);
+                encoder->dispatchCompute(1, 1, 1);
+                encoder->endEncoding();
+            }
+
+            commandBuffer->close();
+            queue->executeCommandBuffer(commandBuffer);
```
tangent-vector, line 123: In terms of further convenience stuff, it seems like we will want some RAII-style types to deal with command buffers and encoders.

At the very least, it would be good for a user to be able to do something like (ignoring naming):
```
{
    CommandBufferRAII commandBuffer(queue, transientHeap);
    ComputeEncoderRAII encoder(commandBuffer);
    ...
    encoder.dispatchCompute(1, 1, 1);
    // end
}
```
The destructor for the RAII encoder would then automatically close it, and the destructor for the RAII command buffer would automatically close *and* execute it on the provided queue.

For common cases used by simple apps, the RAII encoder types should then also support a convenience constructor that works directly with a queue:

```
{
    ComputeEncoderRAII encoder(queue, transientHeap);
    ...
    encoder.dispatchCompute(1, 1, 1);
}
```

That variant should have the encoder automatically manage its own constant buffer and close/execute the buffer when the RAII encoder is destructed.
// end

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -1797,8 +1886,34 @@ class D3D12Device : public RendererBase
...
+                    case D3D12_DESCRIPTOR_RANGE_TYPE_UAV:
+                        rootParam.ParameterType = D3D12_ROOT_PARAMETER_TYPE_UAV;
+                        break;
+                    default:
+                        getDebugCallback()->handleMessage(
+                            DebugMessageType::Error,
+                            DebugMessageSource::Layer,
+                            "A shader parameter marked as root parameter is neither SRV nor UAV.");
+                        return SLANG_FAIL;
+                    }
```
tangent-vector, line 1909: Doesn't D3D12 support binding constant buffers as root parameters as well?

tangent-vector (dismissed):

tools/gfx/d3d12/render-d3d12.cpp
csyonghe: They do, but supporting it in gfx requires some more careful handling because they are ShaderObjects instead of a plain resource, so I leave it unimplemented for now.

slang.h
csyonghe: Done.

tangent-vector (approved):

PR #2127, opened 2022-02-10
Add interface for querying downstream compiler time

csyonghe:


jsmall-nvidia (approved):
I guess I worry a little about the interface/mechanism, but it gets the job done for now.

PR #2136, opened 2022-02-17
Fully implement the ray tracing pipeline for Vulkan

lucy96chen:
Changes:
- Added implementation for `dispatchRays()`, `bindPipeline()`, and `endEncoding()` and added the `ShaderTableImpl` class, which implements the inherited function `createDeviceBuffer()` - Combined with the implementation of `createRayTracingPipelineState()`, these allow the `ray-tracing-pipeline` example to successfully run.
- Changed `ShaderTableBase::m_entryPointNames` to `m_shaderGroupNames` to better reflect its contents
- Added an `extendedFeatures` check that adds Vulkan's ray tracing pipeline extension if supported and added code to get the ray tracing pipeline properties, which is saved to the new member variable `m_rtProperties` in `VulkanApi`

@csyonghe

### DIFF
--- a/examples/ray-tracing-pipeline/main.cpp
+++ b/examples/ray-tracing-pipeline/main.cpp
@@ -293,7 +293,7 @@ void onMouseUp(platform::MouseEventArgs args) { isMouseDown = false; }
 
 Slang::Result initialize()
 {
-    initializeBase("Ray Tracing Pipeline", 1024, 768);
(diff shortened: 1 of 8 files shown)


csyonghe (approved):
Looks good, but please change the call to `getName` to `getNameOverride` in order to get the correct behavior if the user shader program is using the entry-point renaming feature.

csyonghe (approved):

PR #2139, opened 2022-02-20
gfx: defer downstream shader compilation until draw/dispatch.

csyonghe:
This change defers the shader compilation and actual pipeline state creation until it is needed by a draw call.

The reason for this is that creating a `MutableRootShaderObject` requires a `IShaderProgram`. If the user is creating a mutable root shader object, they will first need to create a `IShaderProgram`. Before this change, creating a `IShaderProgram` with a slang program that has 0 specialization arguments will immediately trigger downstream shader compilation. However there are cases where the user is creating MutableRootShaderObjects but never actually use them. Actively compiling the shaders is causing a slow down in user application's start up time. To address this issue, we always compile shaders and create pipeline state objects lazily until it is definitely needed.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -110,6 +110,8 @@ enum class AccessFlag
     Write,
 };
 
+const uint32_t kMaxRenderTargetCount = 8;
+
 class ITransientResourceHeap;
 
 class IShaderProgram: public ISlangUnknown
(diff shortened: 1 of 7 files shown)


csyonghe:
I am merging this in without review to get a release for a Falcor MR.

PR #2140, opened 2022-02-21
gfx: d3d12 performance optimizations.

csyonghe:
* Remove the logic for waitable objects in d3d12 swapchain implementation.
* Fixed a frame fence syncing bug (signaling the wrong fence after present)
* Reset the command allocator for QueryHeap to prevent leaking
* Refactored `TransientResourceHeap` implementation to reuse the same logic for transient constant buffer and staging buffer allocation.
* Make D3D12 `TransientResourceHeap` implement `ID3D12TransientResourceHeap` interface that provides method for allocating descriptor tables on the currently bound GPU heap.
* Improve `textureBarrier` implementation to issue a single barrier when transitioning the entire resource instead of one barrier for each subresource.

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -1821,8 +1821,6 @@ public:
     virtual SLANG_NO_THROW void SLANG_MCALL close() = 0;
 
     virtual SLANG_NO_THROW Result SLANG_MCALL getNativeHandle(InteropHandle* outHandle) = 0;
-
-    virtual SLANG_NO_THROW Result SLANG_MCALL resetDescriptorHeaps() = 0;
 };
 #define SLANG_UUID_ICommandBuffer                                                      \
     {                                                                                  \
@@ -1912,6 +1910,23 @@ public:
         0xcd48bd29, 0xee72, 0x41b8, { 0xbc, 0xff, 0xa, 0x2b, 0x3a, 0xaa, 0x6d, 0xeb } \
     }
 
+class ID3D12TransientResourceHeap : public ISlangUnknown
+{
+public:
+    enum class DescriptorType
+    {
+        ResourceView, Sampler
+    };
+    virtual SLANG_NO_THROW Result SLANG_MCALL allocateTransientDescriptorTable(
+        DescriptorType type,
+        uint32_t count,
+        uint64_t& outDescriptorOffset,
+        void** outD3DDescriptorHeapHandle) = 0;
+};
+#define SLANG_UUID_ID3D12TransientResourceHeap                                             \
+    {                                                                                  \
+        0x9bc6a8bc, 0x5f7a, 0x454a, { 0x93, 0xef, 0x3b, 0x10, 0x5b, 0xb7, 0x63, 0x7e } \
+    }
 
 class ISwapchain : public ISlangUnknown
 {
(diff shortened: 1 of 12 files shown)


slang-gfx.h
```diff
@@ -1912,6 +1910,23 @@ class ITransientResourceHeap : public ISlangUnknown
...
+public:
+    enum class DescriptorType
+    {
+        ResourceView, Sampler
+    };
+    virtual SLANG_NO_THROW Result SLANG_MCALL allocateTransientDescriptorTable(
+        DescriptorType type,
+        uint32_t count,
+        uint64_t& outDescriptorOffset,
+        void** outD3DDescriptorHeapHandle) = 0;
```
tangent-vector, line 1924: I think this is completely reasonable to have as a low-level interface, but I wonder if this is a case where we ultimately need to have a "mid-level" `gfx` interface that represents a D3D12/Vulkan descriptor table/set, and allow allocation of those from a transient heap in an API-independent manner (ignoring the way that can't be ported to the other APIs easily).

*If* we went down the road of such a mid-level API it would not make sense to have this `DescriptorType` enumeration, which requires every table to either be a resource-view table or a sampler table, given that Vulkan supports sets that mix these cases. In the portable mid-level version, a D3D12 "descriptor set" would consist of zero or one resource-view table and zero or one sampler table.

I guess the problem is that going down that road quickly requires us to have both an abstraction for descriptor sets/tables *and* one for the layouts for those sets/tables (where VK would require the layout, but D3D largely wouldn't care). A user who knows they are on D3D12 could then end up needing/wanting this lower-level API anyway... sigh.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -2482,22 +2482,29 @@ class VKDevice : public RendererBase
...
+                &mappedData));
+            memcpy((char*)mappedData + stagingBufferOffset, data, size);
             api.vkUnmapMemory(api.m_device, stagingBufferImpl->m_buffer.m_memory);
 
             // Copy from staging buffer to real buffer
             VkBufferCopy copyInfo = {};
             copyInfo.size = size;
             copyInfo.dstOffset = offset;
-            copyInfo.srcOffset = 0;
+            copyInfo.srcOffset = stagingBufferOffset;
```
tangent-vector, line 2507: It's a minor thing, but looking at all these changes, it seems clear that we really need a type (probably even in the public header for `gfx`) to represent a pair of a buffer and offset:

```
// Needs thought on naming
struct BufferAndOffset
{
    IBufferResource* buffer;
    Offset offset;
};

// Probably needed until we can join the resource and view hierarchies
struct BufferViewAndOffset
{
    IBufferResourceView* view;
    Offset offset;
};

// Needed?
struct BufferSpan
{
    IBufferResource* buffer;
    Offset begin;
    Offset end; // or offset+size?
};
```

A lot of our public API functions that currently take a buffer+offset as distinct parameters should probably take one of those types, and then have convenience overloads that take just a buffer, or a buffer and offset pair. For C++ we can obviously make the `BufferAndOffset` type include an implicit-conversion constructor that takes a buffer pointer, which could eliminate the need for some of the wrappers.

PR #2143, opened 2022-02-23
Better SCCP and inlining to work with more resource type usage for GLSL

csyonghe:
The changes adds more constant folding capability to SCCP pass, improved the inlining pass so it can handle functions with control flows as long as there is only one return statement.

Also added a `simplify-cfg` pass to merge trivial basic blocks linked by `unconditionalBranch`, this will allow more code to get SSA'ed and SCCP'ed. Not done in this PR, but it seems that SCCP, SSA, DeadCodeElimination and SimplifyCFG should be combined into a single pass that loops over these passes until no more change can be done.

These two combined allow us to compile more code to GLSL with more complicated use of resource type variables and return values. We will now inline all functions that has resource return types when emitting GLSL if `specializeResourceOutput` failed to  remove them.


source/slang/slang-ir-inline.cpp
```diff
@@ -374,81 +381,138 @@ struct InliningPassBase
...
+                case kIROp_ReturnVal:
+                    // A return with a value is similar to `returnVoid` except
+                    // that we need to note the (clone of the) value being
+                    // returned, so that we can use it to replace the value
+                    // of the original call.
+                    //
+                    {
+                        auto returnBranch = builder->emitBranch(afterBlock);
+                        _setSourceLoc(returnBranch, inst, callSite);
+                        returnedValue = findCloneForOperand(env, inst->getOperand(0));
```
tangent-vector, line 510: I'm trying to think this through, and I'm not yet able to convince myself we can handle the more general class of functions being handled here.

Consider this case:

```
int f( int x )
{
    while( true )
    {
        while(true)
        {
            if( B() ) return C(); // single return site
            D();
        }
    }
}

int g(int y)
{
    while( E() )
    {
        F();
        y = f( y );
        if( G(y) ) break;
    }
    return y;
}
```

From what I can tell, this inlining pass is going to effectively replace the `return C();` statement with a branch to the code after the `y = f( y )` call site. This will *seem* to work if the code that is logically "after" the call site is restricted enough, but in a case like the above, we seem to end up with something like:

```
int g_optimized(int y)
{
    g_loop: while( E() )
    {
        F();
// begin inlined body of `f`
        outer_f_loop: while( true )
        {
            while(true)
            {
                if( B() ) {
                    y = C(); // assign directly to ressult... this part is easy...

                    // What do do here?!?!?!?!
                    //
                    // If we include the code after the call site directly,
                    // we need a multi-level `break` to implement the `break`
                    // from after the call site...
                    if( G(y) ) break g_loop;
                    //
                    // If instead we wanted to include the `if( G() )` code
                    // in its original location, we need to be able to do
                    break f_outer_loop;
                    //
                    // Either way we need the semantics of multi-level `break`!!! 
                }
                D();
            }
        }
// end inlined body of `f`
    }
    return y;
}
```

Am I missing something here?

source/slang/slang-ir-inline.cpp
```diff
@@ -492,4 +556,44 @@ void performMandatoryEarlyInlining(IRModule* module)
...
+    GLSLResourceReturnFunctionInliningPass(IRModule* module)
+        : Super(module)
+    {}
+
+    bool shouldInline(CallSiteInfo const& info)
+    {
+        if (isResourceType(info.callee->getResultType()))
+        {
+            return true;
+        }
```
tangent-vector, line 578: This probably makes it okay: if we are only inlining functions that we *had* to inline anyway, then it is better to inline and risk producing a function we can't emit to GLSL rather than not inline and have an invalid function with a resource `return`.

tangent-vector: I'm still surprised by this situation, though, because we have an existing pass that is supposed to be optimizing these cases. I am left wondering why the existing pass that was supposed to specialize resources in function outputs isn't applying for these ones.

source/slang/slang-ir-sccp.cpp
```diff
@@ -228,6 +273,460 @@ struct SCCPContext
...
+        case kIROp_Int8Type:
+        case kIROp_Int16Type:
+        case kIROp_IntType:
+        case kIROp_Int64Type:
+        case kIROp_UInt8Type:
+        case kIROp_UInt16Type:
+        case kIROp_UIntType:
+        case kIROp_UInt64Type:
+        case kIROp_BoolType:
+            resultVal = getBuilder()->getIntValue(type, intFunc(c0->value.intVal, c1->value.intVal));
```
tangent-vector, line 401: One issue here (not worth addressing right now) is that we don't properly limit the range constant value we store based on the type. We should probably have a function that takes an `IRIntegerLiteralVal` and an `IROp` for an integer type, and limits the value to one representable in that type (and the same for floating-point).

The logic for handling `kIROp_Construct` above already handles most of that, so it could in theory just be pulled out to its own function.

source/slang/slang-ir-sccp.cpp
```diff
@@ -288,10 +793,116 @@ struct SCCPContext
...
+        if (!as<IRBasicType>(inst->getDataType()))
+            return LatticeVal::getAny();
+
+        switch (inst->getOp())
+        {
+        case kIROp_Construct:
+            return evalConstruct(inst->getFullType(), getLatticeVal(inst->getOperand(0)));
+        case kIROp_Add:
+            return evalAdd(
+                inst->getFullType(),
```
tangent-vector, line 807: Note: using `getFullType` here might lead to issues, since the distinction between `getType()` and `getFullType()` is that `getFullType()` can return a rate-qualified type. We use rate-qualified types *very* rarely in the IR right now - basically just as our way to encode `groupshared` variables (`@GroupShared`) and `constexpr` (`@ConstExpr`).

I suspect that in practice we actually just want `getType()` for all of these, but I can't be entirely certain; if we ended up using rates for more stuff down the line we might decide that the folding logic here needs to take them into account.

source/slang/slang-ir-sccp.cpp
```diff
@@ -288,10 +793,116 @@ struct SCCPContext
...
+                getLatticeVal(inst->getOperand(0)),
+                getLatticeVal(inst->getOperand(1)));
+        case kIROp_Select:
+            return evalSelect(
+                getLatticeVal(inst->getOperand(0)),
+                getLatticeVal(inst->getOperand(1)),
+                getLatticeVal(inst->getOperand(2)));
+        default:
+            break;
```
tangent-vector, line 904: This all seems good, and we can keep adding to it. I think there are opportunities to factor the folding and simplification logic to be more streamlined, and usable from outside the SCCP pass if we ever want that. E.g., it would be really nice if we could have access to these kind of simplifications as part of the `IRBuilder`, since it could mean that in truly simple cases we simply don't even emit any IR for these operations.

source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -0,0 +1,74 @@
...
+        while (block)
+        {
+            // If `block` does not end with an unconditional branch, bail.
+            if (block->getTerminator()->getOp() != kIROp_unconditionalBranch)
+                break;
+            auto branch = as<IRUnconditionalBranch>(block->getTerminator());
+            auto successor = branch->getTargetBlock();
+            // Only perform the merge if `block` is the only predecessor of `successor`.
+            if (successor->hasMoreThanOneUse())
+                break;
```
tangent-vector, line 31: I don't know if we *need* to note this in the comment, but it is important that the main case where `successor->hasMoreThanOneUse()` could be true even when `successor` might have only one predecessor is if `successor` is the "merge" point for some control-flow construct. E.g., if you have:

```
while( true )
{
    if(A())
    {
        B();
        break;
    }
    C();
}
D();
```

then the block with `D()` is the only successor of the block with `B()`, but we wouldn't be able to merge the two as long as the `while` loop remains in the IR, because the instruction for that loop references the `D()` block as its merge point. It is *required* that we don't perform a merge in a case like the above, because as written `B()` and `D()` will typically execute with different "active sets" of threads, and thus with different wave masks. Any merge of those two blocks could affect the value returned by wave-mask-dependent operations under `B()` and/or `D()`.

tangent-vector (approved):
This is great to have. There's room to keep building on what is in this PR, but this is already a significant step up from what we had before.

source/slang/slang-ir-inline.cpp
csyonghe: I need to go double check our code, but could it be that the `specializeResourceOutputs` pass only look at "trivial" functions? If I have a function that returns a resouce type, and that there is an unrelated `if` statement in the function, will that pass still work?

source/slang/slang-ir-simplify-cfg.cpp
csyonghe: Yeah, and this is exactly one of the bugs I have fixed when implementing this.

source/slang/slang-ir-inline.cpp
csyonghe: Right, we don't have a way to support multi-level break now.

csyonghe:
Sorry I messed up my local branch and now it is very hard for me the separate my changes with different PRs. This PR is now succeeded by #2146

PR #2145, opened 2022-02-24
Add isOccluded() and setFullScreenMode() to ISwapchain

lucy96chen:
Changes:
- Added `isOccluded()` and `setFullScreenMode()` to `ISwapchain`
- Added implementations for D3D12 (all others currently return `false`/`SLANG_FAIL`)

@csyonghe

### DIFF
--- a/tools/gfx/d3d/d3d-swapchain.h
+++ b/tools/gfx/d3d/d3d-swapchain.h
@@ -143,6 +143,7 @@ public:
     ISwapchain::Desc m_desc;
     ComPtr<IDXGISwapChain2> m_swapChain;
     Slang::ShortList<Slang::RefPtr<TextureResource>> m_images;
+    bool isWindowOccluded = false;
 };
 
 }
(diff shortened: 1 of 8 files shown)


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -5436,6 +5436,18 @@ class D3D12Device : public RendererBase
 
             return SLANG_OK;
         }
+        virtual SLANG_NO_THROW bool SLANG_MCALL isOccluded() override
+        {
+            if (isWindowOccluded)
```
csyonghe, line 5441: Let's not cache this. We should always call `m_swapChain3->Present` directly.

tools/gfx/d3d/d3d-swapchain.h
```diff
@@ -143,6 +143,7 @@ class D3DSwapchainBase
     ISwapchain::Desc m_desc;
     ComPtr<IDXGISwapChain2> m_swapChain;
     Slang::ShortList<Slang::RefPtr<TextureResource>> m_images;
+    bool isWindowOccluded = false;
```
csyonghe, line 146: Shouldn't need this.

csyonghe (approved):

csyonghe (approved):

PR #2146, opened 2022-02-25
Improved SCCP, inlining and resource specialization passes, legalize `ImageSubscript` for GLSL

csyonghe:
The changes adds more constant folding capability to SCCP pass, improved the inlining pass so it can handle functions with control flows as long as there is only one return statement.

Also added a simplify-cfg pass to merge trivial basic blocks linked by unconditionalBranch, this will allow more code to get SSA'ed and SCCP'ed. 

This PR also combines SCCP, SSA, DeadCodeElimination and SimplifyCFG into a single `simplifyIR` pass that loops over these passes until no more change can be done.

The interfaces for several existing passes, including dead code elimination, ssa, specializeFunction, inline, etc. have been changed to report whether any changes has been applied, so they can now be part of a iterative cycle.

Resource specialization is now done in a `specializeResourceUsage` pass that invokes the current `specializeResourceOutput` and `specializeResourceParameter` together with `simplifyIR` iteratively. It then calls `inlineCall` to inline any resource return functions that cannot be specialized, and then rerun the entire cycle to do parameter specialization for remaining functions. This will minimize the chance that we have to fall back to inlining.

GLSL does not support the `[]` syntax on `image` types, and this PR adds a legalization pass to convert all `ImageSubscript` insts into newly defined `ImageLoad` and `ImageStore`.

We should probably also change the stdlib definition to translate `Texture::operator[]` into the `ImageLoad` and `ImageStore` instructions instead of relying on the current `__target_intrinsic` hack for glsl generation.

The current implementation still generates an extra `imageLoad` in the added test case, and there is not easy way to workaround that. Still this is better than generating invalid glsl.


source/slang/slang-ir-inline.cpp
```diff
@@ -33,17 +35,18 @@ struct InliningPassBase
     }
 
         /// Consider all the call sites in the module for inliing
-    void considerAllCallSites()
+    bool considerAllCallSites()
```
tangent-vector, line 38: Definitely not needed right now, but it might be better if the change-tracking going on here were more explicit and passed down. We could have a type `ChangeTracker` or similar that can be passed by pointer/reference down the chain, and then it has a `.setChanged()`, `.hasChanged()`, and `.reset()` members.

Certainly I'd argue that a bare `bool` is not the return type we want in the long run. :)

tangent-vector (approved):

PR #2147, opened 2022-02-28
Use GLSL scalar layout for constant buffers.

csyonghe:
This change makes Slang to use `scalar` constant buffer layout when `forceGLSLScalarBufferLayout` target flag is set.
Also updates the type reflection logic to respect this target flag.

Adds `createShaderObjectFromTypeLayout` method in gfx to make gfx shader object use the consistent type layout.

### DIFF
--- a/source/slang/slang-parameter-binding.cpp
+++ b/source/slang/slang-parameter-binding.cpp
@@ -678,7 +678,7 @@ RefPtr<TypeLayout> getTypeLayoutForGlobalShaderParameter(
     // An "ordinary" global variable is implicitly a uniform
     // shader parameter.
     return createTypeLayout(
-        layoutContext.with(rules->getConstantBufferRules()),
+        layoutContext.with(rules->getConstantBufferRules(context->getTargetRequest())),
         type);
 }
 
@@ -2080,7 +2080,8 @@ static RefPtr<TypeLayout> computeEntryPointParameterTypeLayout(
         // constant buffer (e.g., the `$Params` constant buffer seen in fxc/dxc output).
         //
         return createTypeLayout(
-            context->layoutContext.with(context->getRulesFamily()->getConstantBufferRules()),
+            context->layoutContext.with(
+                context->getRulesFamily()->getConstantBufferRules(context->getTargetRequest())),
             paramType);
     }
     else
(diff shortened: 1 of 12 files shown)


source/slang/slang-parameter-binding.cpp
```diff
@@ -678,7 +678,7 @@ RefPtr<TypeLayout> getTypeLayoutForGlobalShaderParameter(
     // An "ordinary" global variable is implicitly a uniform
     // shader parameter.
     return createTypeLayout(
-        layoutContext.with(rules->getConstantBufferRules()),
+        layoutContext.with(rules->getConstantBufferRules(context->getTargetRequest())),
```
tangent-vector, line 681: I think that the original plan had been that the "rules" objects would not be parameterized on anything (including the target), and that instead it would be up to the target to construct/return a suitable set of rules objects to describe itself.

As it stands that would get really messy because most of the rules objects are allocated as global-scope values, so I agree with the approach taken here in the short term.

In the long run we should probably have an explicit operation where we take a `TargetRequest` and return a `LayoutRuleFamily*` that implements all its rules, and where the resulting rules objects never need to refer back to the target.

tools/gfx/debug-layer.cpp
```diff
@@ -669,6 +669,42 @@ Result DebugDevice::createMutableRootShaderObject(
...
+    outObject->m_slangType = type;
+    if (SLANG_FAILED(result))
+        return result;
+    returnComPtr(outShaderObject, outObject);
+    return result;
+}
+
+Result DebugDevice::createMutableShaderObjectFromTypeLayout(
+    slang::TypeLayoutReflection* typeLayout, IShaderObject** outShaderObject)
+{
```
tangent-vector, line 692: It seems like we will eventually want to have most of the exposed `create*ShaderObject*()` operations bottleneck through a single bottleneck routine to make things like the `DebugDevice` less onerous to implement.

I suppose that could either be done by having some of those operations just be helper/wrapper operations that sit above the binary interface of the COM objects, or by using even more subclassing in the implementations such that even `DebugDevice` can derive some of these implementations such that they bottleneck into one another.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -6393,7 +6398,7 @@ class VKDevice : public RendererBase
                 m_desc.format = Format::B8G8R8A8_UNORM;
             }
 
-            SLANG_RETURN_ON_FAIL(createSwapchainAndImages());
+            createSwapchainAndImages();
```
tangent-vector, line 6401: It seems surprising to omit the error check on this.

tools/gfx/vulkan/render-vk.cpp
```diff
@@ -8740,7 +8748,10 @@ Result VKDevice::getFormatSupportedResourceStates(Format format, ResourceStateSe
     }
     // AccelerationStructure
     if (bufferFeatures & VK_FORMAT_FEATURE_ACCELERATION_STRUCTURE_VERTEX_BUFFER_BIT_KHR)
+    {
         allowedStates.add(ResourceState::AccelerationStructure);
```
tangent-vector, line 8752: It seems like the `AccelerationStructure` state (the one that was in this code originally) is actually the one that *doesn't* belong here. Might be good to confirm.

tangent-vector (approved):

PR #2151, opened 2022-03-02
Expose API-specific row alignment values

lucy96chen:
Changes:
- Added `getTextureRowAlignment()` to `IDevice` - This function returns the row alignment value as defined by the device's specific API.
- Changed `copyTextureToBuffer()` to also take a `dstRowStride` value, which indicates the user's desired row stride to be used for the copy, and assign it to `Footprint.RowPitch` instead of calculating it inside the function - Failing to pass in a properly aligned value results in an assertion failure.

@csyonghe This was Tess's idea while we were looking at some problems I was having trying to refactor the texture copying test code to accommodate testing all formats and texture types


tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -4462,9 +4464,9 @@ class D3D12Device : public RendererBase
                         footprint.Footprint.Depth =
                             Math::Max(1, (textureSize.depth >> mipLevel)) - srcOffset.z;
                     }
-                    footprint.Footprint.RowPitch = (UINT)D3DUtil::calcAligned(
-                        footprint.Footprint.Width * (UInt)formatInfo.blockSizeInBytes,
-                        (uint32_t)D3D12_TEXTURE_DATA_PITCH_ALIGNMENT);
+
+                    assert(dstRowStride % D3D12_TEXTURE_DATA_PITCH_ALIGNMENT == 0);
```
csyonghe, line 4468: This assert is fine to keep here, but we should add this validation in our debug layer.

csyonghe, line 4468: This assert is fine to keep here, but we should add this validation in our debug layer.

csyonghe (approved):

PR #2153, opened 2022-03-07
GFX Vulkan: deferred shader compilation and pipeline creation.

csyonghe:
This change defers pipeline creation and shader compilation until they are actually needed by a draw call in gfx vulkan implementation.

Before this change, the implementation will compile the shader immediately at the creation of a `IShaderProgram` if the user program is already fully specialized.

This is a quick solution to avoid unnecessary shader compiles when the user creates a `IShaderProgram`, use it to create a mutable root shader objects but then never use them.


source/core/slang-dictionary.h
```diff
@@ -167,7 +167,7 @@ namespace Slang
 					return FindPositionResult(hashPos, -1);
 				}
 				numProbes++;
-				hashPos = (hashPos + GetProbeOffset(numProbes)) & bucketSizeMinusOne;
```
jsmall-nvidia, line 170: Shouldn't that be `bucketSize` or `(bucketSizeMinusOne + 1)` ? That anding if bucketSize is a power of 2, gives values between 0 and bucketSize - 1. Mod will give values (assuming bucketSize >=2) between 0 and  bucketSize - 1 - 1

tools/gfx/d3d12/render-d3d12.cpp
```diff
@@ -2478,68 +2478,26 @@ class D3D12Device : public RendererBase
...
-            {
-                // If the user specifies entry point components via the separated entry point array,
-                // compile code from there.
-                for (auto& entryPoint : linkedEntryPoints)
-                {
-                    SLANG_RETURN_ON_FAIL(compileShader(
-                        entryPoint->getLayout()->getEntryPointByIndex(0), entryPoint, 0));
-                }
-            }
+        virtual Result createShaderModule(
```
jsmall-nvidia, line 2491: This seems more like `addShaderModule`. 
Passing in kernelCode as just ISlangBlob seems reasonable.

tools/gfx/renderer-shared.cpp
```diff
@@ -821,6 +821,62 @@ void ShaderProgramBase::init(const IShaderProgram::Desc& inDesc)
...
+            entryPointIndex, 0, kernelCode.writeRef(), diagnostics.writeRef());
+        if (diagnostics)
+        {
+            getDebugCallback()->handleMessage(
+                compileResult == SLANG_OK ? DebugMessageType::Warning : DebugMessageType::Error,
+                DebugMessageSource::Slang,
+                (char*)diagnostics->getBufferPointer());
+        }
+        SLANG_RETURN_ON_FAIL(compileResult);
+        SLANG_RETURN_ON_FAIL(createShaderModule(entryPointInfo, kernelCode));
```
jsmall-nvidia, line 844: When I saw `create` it made it seem as if kernelCode was the output which made this suspicious.  Same point as earlier - perhaps better as `addShaderModule`.

source/core/slang-dictionary.h
jsmall-nvidia: It seems that the bucketSize is a power of 2 still? So why not stick with & - it would be faster. Not using a power2 might provide better hash map storage distribution.

jsmall-nvidia (approved):
As long as wrap around on probing on Dictionary is consistent with the hashPos being off by one (and min table size is >= 2, it seems to be 16) this probably doesn't introduce a bug. It would seem if the bucket has to be power of 2, then it would be best to use & - it would be faster and access all entries.

PR #2154, opened 2022-03-08
gfx: restructure render-d3d12.cpp

csyonghe:
The existing gfx implementations follow a structural pattern that defines a `class *Device`, all implementation classes for other interfaces are nested inside the `*Device* class. The entire implementation for a target API is defined as gigantic class sitting in a single cpp file.

While defining everything inside the scope of a single parent class saves us from the trouble of separating out declaration and definition, it creates the problem that the implementation cannot be separated apart and must be ordered in a specific way so that some types are defined first to allow an implementation of another type to compile. It is also impossible to sparate some pieces of the implementation (e.g. shader-object related stuff) into a different file.

The complicated nesting structure also makes the code difficult to navigate and it can take a long time to find where a member is defined when intellisense is not working.

This changes restructures the d3d12 implementation to make it follow a more conventional C++ code structure: all classes are declared in a header file, then all implementations goes to a cpp file. No complicated type nesting. All d3d12 implementation code are placed under a `gfx::d3d12` namespace to avoid naming conflict with other targets.

Since this change touches every line of code in d3d12-render.cpp, I also ran clang-format on the entire file to make its formatting consistent.


tools/gfx/d3d11/render-d3d11.cpp
```diff
@@ -2071,7 +2071,7 @@ SlangResult SLANG_MCALL createD3D11Device(const IDevice::Desc* desc, IDevice** o
     return SLANG_OK;
 }
 
-static void _initSrvDesc(IResource::Type resourceType, const ITextureResource::Desc& textureDesc, DXGI_FORMAT pixelFormat, D3D11_SHADER_RESOURCE_VIEW_DESC& descOut)
```
jsmall-nvidia, line 2074: Nit pick: Is there a reason to what to make impls that are private to a cpp file, not use `_`? The `_` prefixing is of course not necessary, and is more typically used for non public methods, but here it seems appropriate.

source/core/slang-dictionary.h
```diff
@@ -167,7 +167,7 @@ namespace Slang
 					return FindPositionResult(hashPos, -1);
 				}
 				numProbes++;
-				hashPos = (hashPos + GetProbeOffset(numProbes)) & bucketSizeMinusOne;
+				hashPos = (hashPos + GetProbeOffset(numProbes)) % bucketSizeMinusOne;
```
jsmall-nvidia, line 170: As discussed in other PR it seems this should either be `& bucketSizeMinusOne` (if bucketSize is pow 2) or `% (bucketSizeMinusOne + 1)`, if it's not.

jsmall-nvidia (commented):
I'll wait til #2153 is merged so I can more easily see what the diffs are.

tools/gfx/d3d11/render-d3d11.cpp
csyonghe: I intended to change the name in render-d3d12.cpp but somehow the refactoring feature mistakenly changes the name here. All these functions will eventually be moved out to separate files so perhaps it is fine to keep it as is.

tools/gfx/d3d12/render-d3d12.h
```diff
@@ -1,11 +1,1799 @@
...
+
+// Define function pointer types for PIX library.
+typedef HRESULT(WINAPI* PFN_BeginEventOnCommandList)(
+    ID3D12GraphicsCommandList* commandList, UINT64 color, _In_ PCSTR formatString);
+typedef HRESULT(WINAPI* PFN_EndEventOnCommandList)(ID3D12GraphicsCommandList* commandList);
+
+static const Int kMaxNumRenderFrames = 4;
+static const Int kMaxRTVCount = 8;
+
+struct D3D12DeviceInfo
```
jsmall-nvidia, line 47: Does it need D3D12 prefix when in the d3d12 namespace?

jsmall-nvidia (approved):
Why does the implementation of the methods move to the header? To be more like a regular C++ project - large method impls would typically remain in cpp file unless there is an important reason not to.

tools/gfx/d3d12/render-d3d12.h
```diff
@@ -1,11 +1,1799 @@
...
+
+    virtual SLANG_NO_THROW Result SLANG_MCALL unmap(MemoryRange* writtenRange) override;
+
+    virtual SLANG_NO_THROW Result SLANG_MCALL setDebugName(const char* name) override;
+};
+
+class TextureResourceImpl : public TextureResource
+{
+public:
+    typedef TextureResource Parent;
```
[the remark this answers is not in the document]
csyonghe: Yes, I noticed this as well. This is just original code and I will change it later.

tools/gfx/d3d12/render-d3d12.h
csyonghe: There is already a DeviceInfo type in `gfx` namespace, perhaps I should just rename it to something else.

PR #2156, opened 2022-03-08
Initial support for documentation extraction in C++

jsmall-nvidia:
* Split the markdown extractor, so main functionality is available to C++ extractor
* Fixed some bugs/issues around extraction
* Added support for documentation extraction to the C++ extractor
* Ran on stdlib to get changes.


tangent-vector (approved):
I'm interested to see where this goes (I'm assuming this is for the benefit of auto-generating our C/C++ API docs).

PR #2157, opened 2022-03-09
gfx: restructure render-vk.cpp

csyonghe:
Based on #2154, this change restructures render-vk.cpp to get rid of the nested class definition structure and move all class declarations to `render-vk.h` and all implementations to `render-vk.cpp`.

Reorders the definitions so that the functions for each type are all placed contiguously.

Renames `VKDevice` -> `vk::DeviceImpl`.


tangent-vector (approved):
These changes are good and necessary at this point. It will be healthy to be able to decompose the back-end implementations across more files now, in order to streamline some of these very heavy individual files.

PR #2158, opened 2022-03-11
gfx: Add `ITransientResourceHeap::finish()`

csyonghe:
This method allows `gfx` to know the user has finished using the resources in the transient heap for the current frame, so it can insert a fence signal operation.

Without this method, the implementation has to proactively signal a fence after every command buffer submit, which slightly reduces performance.

### DIFF
--- a/tools/gfx/d3d12/render-d3d12.h
+++ b/tools/gfx/d3d12/render-d3d12.h
@@ -490,9 +490,11 @@ public:
     {
         uint64_t waitValue;
         HANDLE fenceEvent;
+        ComPtr<ID3D12CommandQueue> queue;
         ComPtr<ID3D12Fence> fence = nullptr;
     };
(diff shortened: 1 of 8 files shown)


tools/gfx/d3d12/render-d3d12.h
```diff
@@ -490,9 +490,11 @@ class TransientResourceHeapImpl
     {
         uint64_t waitValue;
         HANDLE fenceEvent;
+        ComPtr<ID3D12CommandQueue> queue;
         ComPtr<ID3D12Fence> fence = nullptr;
     };
     ShortList<QueueWaitInfo, 4> m_waitInfos;
+    Array<HANDLE, 16> m_waitHandles;
```
tangent-vector, line 497: Should this be `ShortList`, or is an `Array` okay here?

tangent-vector (approved):

PR #2159, opened 2022-03-14
C++ extractor callable support

jsmall-nvidia:


tools/slang-cpp-extractor/parser.cpp
```diff
@@ -427,6 +419,97 @@ SlangResult Parser::_parseEnum()
...
+{
+    // Skip the current 'template' token.
+    m_reader.advanceToken();
+
+    // Consume everything in <>
+    SLANG_RETURN_ON_FAIL(expect(TokenType::OpLess));
+
+    {
+        Index arrowCount = 1;
+        while (true)
```
csyonghe, line 432: We need to be careful with these `while(true)` loops and make sure we exit even when there is an error, e.g. when reaching the end of file.

tools/slang-cpp-extractor/parser.cpp
```diff
@@ -215,7 +218,7 @@ SlangResult Parser::_maybeConsumeScope()
         }
         else if (type == TokenType::LBrace)
         {
-            m_reader.advanceToken();
+            //m_reader.advanceToken();
```
csyonghe, line 221: Just remove?

csyonghe (approved):

PR #2161, opened 2022-03-15
Add -depfile option to save dependency info

apanteleev:
This change adds a `-depfile` command line option to `slangc` to save the dependency information, i.e. which source files does every output of the current compile request depend on. The deps are saved in a regular Makefile-type format, similar to e.g. `glslangValidator`.

### DIFF
--- a/source/slang/slang-compiler.cpp
+++ b/source/slang/slang-compiler.cpp
@@ -2177,12 +2177,12 @@ namespace Slang
         return SLANG_OK;
     }
 
-    static void writeString(Stream& stream, const char* string)
+    static void _writeString(Stream& stream, const char* string)
     {
         stream.write(string, strlen(string));
     }
 
-    static void escapeDependencyString(const char* string, StringBuilder& builder)
+    static void _escapeDependencyString(const char* string, StringBuilder& outBuilder)
     {
         // make has unusual escaping rules, but we only care about characters that are acceptable in a path
         for (const char* p = string; *p; ++p)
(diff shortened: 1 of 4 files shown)


source/slang/slang-compiler.cpp
```diff
@@ -2177,6 +2177,103 @@ namespace Slang
...
+            case '[':
+            case ']':
+            case '\\':
+                builder.appendChar('\\');
+                break;
+
+            case '$':
+                builder.appendChar('$');
+                break;
+            }
```
jsmall-nvidia, line 2205: Is the '$' case necessary?

source/slang/slang-compiler.cpp
```diff
@@ -2177,6 +2177,103 @@ namespace Slang
         return SLANG_OK;
     }
 
+    static void writeString(Stream& stream, const char* string)
+    {
+        stream.write(string, strlen(string));
+    }
+
+    static void escapeDependencyString(const char* string, StringBuilder& builder)
```
jsmall-nvidia, line 2185: Nitpik: For variables that are 'output' we have a convention of prefixing with `out` so `outBuilder` here. If it's in/out we prefix with `io`.

source/slang/slang-compiler.cpp
```diff
@@ -2177,6 +2177,103 @@ namespace Slang
...
+                break;
+            }
+
+            builder.appendChar(c);
+        }
+    }
+
+    // Writes a line to the file stream, formatted like this:
+    //   <output-file>: <dependency-file> <dependency-file...>
+    static void writeDependencyStatement(Stream& stream, EndToEndCompileRequest* compileRequest, const String& outputPath)
```
jsmall-nvidia, line 2213: Nitpik: By convention (although certainly not always used by any means) for functions that are local to a .cpp file (or are not part of the public interface of a class) we prefix with _. So `_writeDependencyStatement` here.

source/slang/slang-compiler.cpp
apanteleev: Yes: per Makefile conventions, `$` is escaped as `$$`, unlike other characters.

source/slang/slang-compiler.cpp
apanteleev: Done.

source/slang/slang-compiler.cpp
apanteleev: Done.

source/slang/slang-compiler.cpp
jsmall-nvidia: Ok makes sense.

jsmall-nvidia (approved):

jsmall-nvidia (approved):

PR #2163, opened 2022-03-17
Fix type truncation during SCCP.

csyonghe:
This fixes the type truncations errors in SCCP.


source/slang/slang-emit-c-like.cpp
```diff
@@ -884,14 +883,32 @@ void CLikeSourceEmitter::emitSimpleValueImpl(IRInst* inst)
                     // This little hack is needed for gcc that if we have the expression
                     // int(-0x80000000) we get the warning: warning :  integer overflow in expression [-Woverflow]
                     // 0x80000000 and -0x80000000 mean the same thing when casted to 32 bit int, so we just flip the value here.
-                    IRIntegerValue value = litInst->value.intVal;
-                    value = (value == -0x80000000ll) ? -value : value;
```
jsmall-nvidia, line 888: Why have the mask here, when the cast later will mean it has no effect?

source/slang/slang-emit-c-like.cpp
```diff
@@ -884,14 +883,32 @@ void CLikeSourceEmitter::emitSimpleValueImpl(IRInst* inst)
...
                     m_writer->emit("int(");
-                    m_writer->emit(value);
+                    m_writer->emit(int32_t(value));
                     m_writer->emit(")");
                     return;
                 }
+                case BaseType::Int8:
+                    {
+                        IRIntegerValue value = (litInst->value.intVal & 0xFF);
+                        value = (value == -0x80ll) ? -value : value;
```
jsmall-nvidia, line 895: Why does Int8/Int16 have the ```value = (value == -0x80ll) ? -value : value;``` and the Int32 version removes it.

From the comment above it seems it's purpose is to avoid issues in gcc around int(-0x80000000). 

It's if the sequence that is desired (for say int8_t) is

```
// Does mask/sign extension
IRIntegerValue value = (int8_t)litInst->value.intVal;
// Makes sure MIN_INT8 is turned into unsigned form (to remove warning from gcc)
value = (value == -0x80ll) ? -value : value;
m_writer->emit("int8_t(");
m_writer->emit(value);
m_writer->emit(")");
```

source/slang/slang-ir.cpp
```diff
@@ -2081,7 +2081,32 @@ namespace Slang
...
+        case kIROp_IntType:
+            keyInst.value.intVal = static_cast<int32_t>(inValue);
+            break;
+        case kIROp_UInt8Type:
+            keyInst.value.intVal = static_cast<uint8_t>(inValue);
+            break;
+        case kIROp_UInt16Type:
+            keyInst.value.intVal = static_cast<uint16_t>(inValue);
+            break;
+        case kIROp_BoolType:
```
jsmall-nvidia, line 2102: On bool shouldn't it do something like IRIntegerValue(inValue != 0) to ensure it's only 1 or 0?

source/slang/slang-ir-sccp.cpp
```diff
@@ -309,24 +309,6 @@ struct SCCPContext
             case kIROp_BoolLit:
                 {
                     IRIntegerValue intVal = irConstant->value.intVal;
-                    switch (type->getOp())
```
jsmall-nvidia, line 312: If irConstants type is same as type, can just return directly?

source/slang/slang-emit-c-like.cpp
```diff
@@ -884,14 +883,32 @@ void CLikeSourceEmitter::emitSimpleValueImpl(IRInst* inst)
...
-                    value = (value == -0x80000000ll) ? -value : value;
-
+                    IRIntegerValue value = (litInst->value.intVal & 0xFFFFFFFF);
                     m_writer->emit("int(");
-                    m_writer->emit(value);
+                    m_writer->emit(int32_t(value));
                     m_writer->emit(")");
                     return;
                 }
+                case BaseType::Int8:
```
jsmall-nvidia, line 892: It seems from below that the construction of the integer literal handles masking/sign extension as part of construction. If that's the case doing the sign extension/masking here isn't necessary?

csyonghe: Currently I am not 100% confident that the masking we done at construction covers 100% sources of constant insts, so adding the logic here is just to be safe.

source/slang/slang-ir.cpp
csyonghe: Bool on many platforms can have arbitrary values so the compiler probably shouldn't alter the value here.

source/slang/slang-emit-c-like.cpp
csyonghe: Yeah, I think that using the mask alone is sufficient to cover this case, so we should just remove all this negating logic and just use masking.

source/slang/slang-ir.cpp
jsmall-nvidia: Hmm, not sure about this. This is the Slang IR - if code is going to process Slang IR it's typically not going to be able able to handle or it be desirable for *arbitrary* values for bools. It seems that part of the point of this bit of code is to convert values into valid values for the type.

Subsequent stages of IR for handling for targets such as emitting, converting the type or outputting to a target encoding seem more appropriate.

source/slang/slang-emit-c-like.cpp
jsmall-nvidia: The hack was there because there was a gcc version on a system we tested that failed to compile int(-0x80000000) in the way discussed. So that is a thing. That said I did try this out on a bunch of compilers/options on godbolt, and didn't repeat it.

So I guess it's ok, but it's possible it will arise again as an issue. On balance I'd leave it, but not  a show stopper.

jsmall-nvidia (approved):
I put my concerns in comments, but they aren't show stoppers in terms of fixing the bug.

csyonghe:
We should have a test case to make sure we have the correct behavior around int(-0x80000000). This PR should work just fine for this specific case, but will need a test case to confirm.

PR #2166, opened 2022-03-21
Fix most (if not all) issues that came up in previous heterogeneity PR

dsiher:
* All(?) unintentional style changes reverted.
* Decoration naming changed to fit Slang naming scheme
* Removed __externLib decoration, and with it the hacky NameHint based
  solution to emitting (now using __target_intrinsic instead).

For some reason the switch to using __target_intrinsic causes this exception to show up, and was hoping someone might have a better idea of what's going on:
image
I decided to PR despite this because the code that is being generated from the .slang file is identical to the code that was being generated before switching to __target_intrinsic, so I think the issue doesn't have to do with heterogeneity(?)


source/slang/core.meta.slang
```diff
@@ -63,25 +63,24 @@ syntax unorm : UNormModifier;
...
+/// TODO: This currently contains a bad hack for passing through
+///       namespaces that relies on name hints. Potentially can be
+///       relegated to another existing decoration (intrinsic)
+/// TODO: The behavior that this modifier implies should technically
+///       be the default, whereas the current compiler behavior of
+///       generating unique names for most things should be
+///       considered a bit of a bug.
 ///
-syntax __externLib : __externLib;
+syntax __exportDirectly : ExportDirectlyModifier;
```
csyonghe, line 83: Why do we still need this? It seems that after switching to __target_intrinsic this modifier is no longer used. We should consider remove it if that is the case.

build/visual-studio/heterogeneous-hello-world/obj/x64/Debug/heteroge.3E8C6F8B.tlog/heterogeneous-first-gen.lastbuildstate
```diff
@@ -0,0 +1,2 @@
+PlatformToolSet=v142:VCToolArchitecture=Native32Bit:VCToolsVersion=14.29.30037:VCServicingVersionATL=14.29.30038:VCServicingVersionCrtHeaders=14.29.30038:VCServicingVersionCompilers=14.29.30038:TargetPlatformVersion=10.0.19041.0:
```
csyonghe, line 1: Remove these files.

source/slang/core.meta.slang
dsiher: __exportDirectly is used in combination with __unmangled on the rest of the functions defined in shader.slang so that they are emitted unmangled. (Unless __target_intrinsic can also solve that problem?)

dsiher: I can swap to __target_intrinsic and remove the unneeded modifiers later today then (also remove the files you mentioned.)

Any idea what's causing the exception?

csyonghe: We have made significant changes to the gfx library, so it is likely that this example is using gfx in an unsupported way that triggers D3D error.

dsiher: I can look into that as well then.

Is there some interaction between gfx and __target_intrinsic though? Because the example was running fine prior to switching from the __externLib modifier (I had fixed some gfx issues before this which I assume were from those gfx modification, but I very well could have missed a more recent change.)

csyonghe: There shouldn't be any. I think it is just a gfx issue. If the crash is seen after you rebased / merged with master, you are also using a very different gfx library.

dsiher: Swapping __exportDirectly with __unmangled to __target_intrinsic is not viable because we want the function declaration to be emitted in the resulting shader.cpp (whereas we did not want the structs to be emitted.) Doing the swap led to identifier not found errors for the functions used in executeComputation (and when __target_intrinsic was applied to executeComputation then it was not emitted altogether.)

I can also confirm that my example works when I remove __target_intrinsic but keep the same previously generated shader.cpp, so the issue definitely is definitely caused (somehow) by the existence of __target_intrinsic. My gfx library usage matches all the other examples that were updated when the library changed, I updated to the new syntax before trying to swap to __target_intrinsic.

dsiher:
I'm not sure if this is relevant, but I'm also getting errors when trying to run the gpu-printing example. Are the changes to gfx recent enough that other examples are not corrected to use them correctly? If so that could be the same case for the heterogeneous example.

PR #2170, opened 2022-03-24
Allow slangc to generate exe from .slang file.

csyonghe:
This change allows `slangc` to produce standalone exe directly from a *.slang file.
Updated heterogeneous-hello-world example to make use of this feature.

With this change, you can write a slang file to represent a CPU program:
```
// main.slang

__target_intrinsic(cpp, "printf(\"%s\", ($0).getBuffer())")
public void writeln(String text);

{
    writeln("hello world");
    return 0;
}
```

And then compile it with `slangc`:

```
slangc main.slang -o hello.exe
```

New `HostCppSource` and `HostExecutable` code gen targets are added to represent a Slang source intended as a normal CPU program instead of a shader kernel. When the user requests `slangc` to produce a `*.exe` file, `slangc` will automatically use `HostCppSource` target to emit C++ source for host.

This change also removes `__externLib`, `__unmangled` syntax, and added `__extern_cpp` to make a symbol use unmangled name in output source.


slang.h
```diff
@@ -551,13 +551,14 @@ extern "C"
         SLANG_DXIL,
         SLANG_DXIL_ASM,
```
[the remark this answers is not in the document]
csyonghe: `SLANG_SHADER_SHARED_LIBRARY` are meant to represent the scenario where the source slang code is treated as kernels and the compiler generates a dll that contains entry-points to the kernel.

`SLANG_HOST_SHARED_LIBRARY` should be used if the whole thing is a general CPU library. So if you are compiling general C++ this is the target that should be used. As for linking slang-rt, currently `SLANG_HOST_SHARED_LIBRARY` implies such linking, but it does make sense to make it a separate option or more intelligent. We can have linking requirement decorations in stdlib and track them in a way similar to glsl extensions so we only link when the user code is using slang-rt features.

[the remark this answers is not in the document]
csyonghe: When compiling slang source, `SLANG_HOST_SHARED_LIBRARY` means a very different set of passes and code emit logic so it isn't just there to tell downstream compiler to link `slang-rt`.

source/core/slang-token-reader.cpp
```diff
@@ -681,9 +681,6 @@ namespace Misc {
         {
             switch (data[i])
             {
-            case ' ':
```
jsmall-nvidia, line 684: There are more complete versions of this functionality in StringEscapeUtil. Not directly an issue for PR, but would be nice to have have DRY.

source/slang/slang-compiler.cpp
```diff
@@ -55,7 +55,19 @@ namespace Slang
 
     // !!!!!!!!!!!!!!!!!!!!!! free functions for DiagnosicSink !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-    void printDiagnosticArg(StringBuilder& sb, CodeGenTarget val)
+bool isHeterogeneousTarget(CodeGenTarget target)
```
jsmall-nvidia, line 58: Along with the larger around targets it not clear to me how these types indicate heterogeneity?

source/slang/slang-emit-cpp.cpp
```diff
@@ -2326,6 +2356,14 @@ bool CPPSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOut
             m_writer->emit("))");
             return true;
         }
+        case kIROp_StringLit:
+        {
+            m_writer->emit("String(");
+            m_writer->emit(Slang::Misc::EscapeStringLiteral(as<IRStringLit>(inst)->getStringSlice()));
```
jsmall-nvidia, line 2362: As above would probly be better to use SlangEscapeUtil.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2629,130 +2666,7 @@ void CPPSourceEmitter::_emitForwardDeclarations(const List<EmitAction>& actions)
 
 void CPPSourceEmitter::emitModuleImpl(IRModule* module, DiagnosticSink* sink)
 {
-    // If we are emitting a heterogeneous program
```
jsmall-nvidia, line 2632: Have we lost some functionality - in terms of emitting GPU kernels as blobs? 

I understand the desire to want to remove the gfx specifics from an emit.

Also do we really have a 'heterogenous' example anymore?

jsmall-nvidia (commented):
I do really like how clean and simple the 'hello world' CPU sample became.

Two things worry me here. 

1) The additional concepts added to `target`
2) We don't appear to have a heterogeneous example anymore

The first point is perhaps not a show stopper, and is just a stage to having a more nuanced way to describe a target. 

The second point I think might need some discussion.

slang.h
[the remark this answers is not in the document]
jsmall-nvidia: Perhaps I'm missing something, or the git diff is, but I didn't see `SLANG_HOST_SHARED_LIBRARY` in the list?

That having perhaps helps, and maybe is ok to merge as such. 

I think having a more clear separation of concepts would be better in general for the future.

source/slang/slang-emit-cpp.cpp
csyonghe: Yeah, I did notice this class but I am not sure how to use it.

slang.h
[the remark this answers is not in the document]
csyonghe: It is currently missing since it is not directly needed by this change. Should add it in later.

jsmall-nvidia (approved):

PR #2177, opened 2022-04-06
Fixed the mapping of the ray tracing instance functions to GLSL

apanteleev:
It is confusing, but HLSL (and by extension Slang) and GLSL define these functions in opposite ways.

In HLSL, `InstanceID` is the user-provided index coming from the TLAS instance descriptor, and `InstanceIndex` is the autogenerated index. 

[DXR spec for ray queries](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#rayquery-committedinstanceindex)
[DXR spec for ray pipelines](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#instanceindex)

In GLSL, `InstanceId` is the autogenerated index, and `InstanceCustomIndex` is the user-provided index. 

[GLSL spec for ray queries](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_ray_query.txt)
[Vulkan spec for ray pipelines](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap15.html#VUID-InstanceCustomIndexKHR-InstanceCustomIndexKHR)

This applies both to ray query functions (candidate and committed) and to ray pipeline functions.

The OptiX mappings are correct.
[OptiX spec for ray pipelines](https://raytracing-docs.nvidia.com/optix7/api/group__optix__device__api.html#ga4d70272186506b1a5cb4813f1047021f)

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -4736,12 +4736,12 @@ uint RayFlags();
 
 // 10.4.3 - Primitive/Object Space System Values
 
-__target_intrinsic(GL_NV_ray_tracing, "(gl_InstanceCustomIndexNV)")
-__target_intrinsic(GL_EXT_ray_tracing, "(gl_InstanceCustomIndexEXT)")
+__target_intrinsic(__glslRayTracing, "(gl_InstanceID)")
 __target_intrinsic(cuda, "optixGetInstanceIndex")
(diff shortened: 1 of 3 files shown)


csyonghe (approved):
Looks good to me. Thanks for the fixes!

PR #2178, opened 2022-04-06
Refactor: eliminate BackEndCompileRequest

tangent-vector:
An earlier refactoring pass over the compiler codebase split the
type that had been called `CompileRequest` into three distinct
pieces:

* `FrontEndCompileRequest` which was supposed to own state and
  options related to running the compiler front end and producing
  IR + reflection (e.g., what translation units and source
  files/strings are included).

* `BackEndCompileRequest` which was supposed to own state and options
  related to running the compiler back end to translate the IR
  for a `ComponentType` (program) into output code. (Note that the
  `BackEndCompileRequest` was conceived of as orthogonal to the
  `TargetRequest`s, which store per-target and target-specific
  options.)

* `EndToEndCompileRequest` which was an umbrella object that owns
  separate front-end and back-end requests, plus any state that is
  only relevant when doing a true end-to-end compile (such as the
  kinds of compiles initiated with `slangc`). As originally conceived,
  the only state that this type was supposed to own was stuff related
  to "pass-through" compilation, as well as state related to writing
  of generated code to output files.

That refactoring work was very useful at the time, because it allowed
us to "scrub" the back end compilation steps to remove all
dependencies on front-end and AST state (this was important for our
goals of enabling linking and codegen from serialized Slang IR).

At this point, however, it is clear that the hierarchy that was built
up serves very little purpose:

* The `BackEndCompileRequest` type is only used in two places:

    * As part of an `EndToEndCompileRequest`, where the settings on
      the `BackEndCompileRequest` can be configured, but only through
      the `EndToEndCompileRequest`

    * As part of on-demand code generation through the `IComponentType`
      APIs. In this case, the settings stored on the
      `BackEndCompileRequest` are not accessible to the application
      at all, and will always use their default values, so that
      instantiating a "request" object doesn't really make any sense.

* The `FrontEndCompileRequest` type has a similar situation:

    * Front-end compilation as part of an `EndToEndCompileRequest`
      supports user configuration of `FrontEndCompileRequest` settings,
      but only through the `EndToEndCompileRequest`

    * Front-end compilation triggered by an `import` or a `loadModule()`
      call does not support user configuration of settings at all. It
      will always derive all relevant settings from thsoe on the
      session ("linkage").

In addition, subsequent changes have been made to the compiler that
show a bit of a "code smell" and/or forward-looking worries for this
decomposition:

* In some cases we've had to add the same setting to multiple types
  in the breakdown (front-end, back-end, end-to-end, linkage, target,
  etc.) which makes it harder for us to validate that all the possible
  mixtures of state work correctly.

* Related to the above, in some cases we have manual logic that copies
  state from one of the objects in the breakdown to another, in order
  to ensure that the user's intention is actually followed.

* As a forward-looking concern, it seems that developers have sometimes
  added new configuration options and state to places that don't really
  make sense according to the rationale of the original decomposition
  (e.g., we probably don't want to have a lot of state that is only
  available via end-to-end requests, given that the API structure is
  meant to push users *away* from end-to-end compiles).

As a result of all of the above, I've been planning a large refactor
with the following big-picture goals:

* Eliminate `BackEndCompileRequest`

    * Move all relevant state/options from the back-end request to
      the end-to-end request, since that is the only place they could
      be set anyway.

    * Introduce a transient "context" type to be used for the duration
      of code generation that serves the main functions that back-end
      requests really served in the codebase

* Make `EndToEndCompileRequest` be a subclass of
  `FrontEndCompileRequest`

    * Consider addding a transient "context" type for front-end
      compiles that can be used in `import`-like cases rather than
      needing a full front-end request object. If this works, then
      eliminate `FrontEndCompileRequest` and be back to world with
      just a single `CompileRequest` type

* Move *all* compiler configuration options to a distinct type (named
  something like `CompilerConfig` or `CompilerOptions` or whatever)
  which stores setting as key-value pairs, and has a notion of
  "inheritance" such that one configuration can extend or build on top
  of another. Make all the relevant types use this catch-all structure
  instead of redundantly storing flags in many places.

This change deals with the first of those bullets: removeal of
`BackEndCompileRequest`. The addition of the `CodeGenContext` type is
perhaps an unncessary additional step, but making that change helps
clean up a bunch of the code related to per-target code generation,
so I think it is the right choice.


source/slang/slang-compiler.h
```diff
@@ -2185,40 +2186,257 @@ namespace Slang
...
+            {}
+
+//            Shared(
+//                TargetProgram*              targetProgram,
+//                EndToEndCompileRequest*     endToEndReq);
+
+            TargetProgram*          targetProgram = nullptr;
+            EntryPointIndices       entryPointIndices;
+            DiagnosticSink*         sink = nullptr;
+            EndToEndCompileRequest* endToEndReq = nullptr;
```
csyonghe, line 2246: What are the situations where we need the back reference to `EndToEndCompileRequest` here? Is it only for pass through? Is it possible to get rid of this to make things cleaner?

csyonghe: A search for `isEndToEndCompile` in this PR reveals that it is really only needed to get dump ir options. Seems like this can just be cleaned up with the `IConfig` refactor we talked about earlier.

csyonghe (approved):
Thanks for putting together this PR! Things looks much cleaner and we can definitely further clean up the back reference to EndToEndCompileRequest in CodeGenContext once we have the `IConfig` refactor.

source/slang/slang-compiler.h
tangent-vector: The `endToEndReq` is really used in two main cases:

1. Pass-through compilation, in which case it is the only piece that has all the relevant state (since otherwise we cannot access the front-end configuration from within codegen).
2. Writing of output files for command-line compilation, in which case the end-to-end request is the only object that knows about the output paths.

It is likely we can split out the logic for (2) so that most of code generation doesn't need to care.

Pass-through (1) is a much more complicated case, though, and it will probably take a lot of thought to design how it ought to work more properly. I am worried that we are getting to a point where we may need to treat pass-through compilation as an entirely separate mode, rather than trying to integrate it into the ordinary functioning of the compiler as we have so far done.

PR #2181, opened 2022-04-11
Support `[DllImport]`

csyonghe:
This add a new attribute `[DllImport]` to heterogeneous c++ target.

This allows the user to call dll functions with the following syntax:
```
[DllImport("User32")]
int MessageBoxA(Ptr<void> hwnd, String text, String caption, uint flags);
```

The above declaration will turn into a function that loads the dll and finds the function pointer, then invokes it:
```
Slang_FuncType<int32_t, void *, char *, char *, uint32_t>  _S1_init()
{
    return nullptr;
}
Slang_FuncType<int32_t, void *, char *, char *, uint32_t>  _S1 = _S1_init();
int32_t MessageBoxA_0(void * _S2, String _S3, String _S4, uint32_t _S5)
{
    char * _S6 = (const_cast<char*>((_S3).getBuffer()));
    char * _S7 = (const_cast<char*>((_S4).getBuffer()));
    if(*(&_S1) == nullptr)
    {
        void * _S9 = (_slang_rt_load_dll((String("User32"))));
        void * _S10 = (_slang_rt_load_dll_func((_S9), (String("MessageBoxA"))));
        *(&_S1) = (slang_bit_cast<Slang_FuncType<int32_t, void *, char *, char *, uint32_t> >(_S10));
    }
    else
    {
    }
    int32_t _S11 = (*(&_S1))(_S2, _S6, _S7, _S5);
    return _S11;
}
```

To make this work, this PR needs to add `nullptr` to the front-end, supporting emitting function pointer types, and clean up the type emit logic in cpp-emit to reuse the delcarator mechanism in c-like-emit.


source/slang-rt/slang-rt.h
```diff
@@ -0,0 +1,22 @@
...
+#    define SLANG_RT_API
+#endif
+
+extern "C"
+{
+    SLANG_RT_API void SLANG_MCALL _slang_rt_abort(Slang::String errorMessage);
+    SLANG_RT_API void* SLANG_MCALL _slang_rt_load_dll(Slang::String modulePath);
+    SLANG_RT_API void* SLANG_MCALL
+        _slang_rt_load_dll_func(void* moduleHandle, Slang::String modulePath);
+}
```
tangent-vector, line 20: This file probably merits a comment of some kind to clarify what exactly `slang-rt` is meant to encapsulate in scope.

Also: looking at these operations, I assume we probably want the `String`s to be passed by-reference (or maybe even by-pointer, to be even more explicit about ABI considerations).

I expect that in the long run we will need to think carefully about how we carve up functionality that generated programs need across:

* The "standard library" code, which is currently written in Slang, compiled to our IR, and subsequently gets emitted as part of the generated code of programs using those routines (effectively getting statically linked into output files via duplication of the code).

* The "prelude" code, which is written in the downstream non-Slang language, and gets prepended to the generated code of programs (such that it is *also* statically linked by duplication)

* The "runtime" code, which differs in practice by not being included directly into the code we generate, and is instead linked to that code further downstream (in this case by the runtime library loader of the host platform).

We should probably be looking to unify the handling of these different cases down the line, so that we have a clear mental model like:

* The standard library declares all the operations that are visible to client code (including runtime operations that we might not always want them to access directly), as well as any lower-level operations that are used to implement the standard library itself. Some operations in the stdlib will be declared but not defined, and those operations are likely to have their definitions provided by the prelude and/or runtime for a target. While the stdlib currently gets compiled into generated output code from the Slang compiler, it should eventually be possible for parts of the stdlib to be precompiled into binaries for appropriate targets (binaries such as the runtime library).

* The prelude is simply used as a utility to generated code; it provides declarations and (inline) definitions for those operations that are needed by generated code, and which are best implemented by inlining into the output for a given target. In some cases the prelude may simply declare operations that are actually provided by the runtime, or may provide inline implementations that forward to the runtime.

* The runtime library provides the target-specific (and platform-specific) implementations of those services that are needed by generated code and that can be best provided by a pre-built binary that is deployed with the compiler.

I fully understand that for now we are unlikely to have any "runtime library" for most of our GPU targets, but we should start thinking about how to architect things for a world where these different pieces are designed to clearly work together (rather than the current situation where they sometimes come across as competing solutions to the same basic problems).

source/slang/core.meta.slang
```diff
@@ -342,11 +342,18 @@ ${{{{
 // Declare built-in pointer type
 // (eventually we can have the traditional syntax sugar for this)
 }}}}
+
+__magic_type(NullPtrType)
+struct NullPtr
+{
+};
+
```
tangent-vector, line 350: This is fine for now, but we should have a longer-term design discussion around how null should ideally be handled in Slang.

source/slang/slang-check-conversion.cpp
```diff
@@ -753,6 +753,18 @@ namespace Slang
             return true;
         }
 
+        // nullptr_t can be cast into any pointer type.
+        if (as<NullPtrType>(fromType) && as<PtrType>(toType))
+        {
+            if (outCost)
+            {
+                *outCost = kConversionCost_None;
```
tangent-vector, line 761: Not very important, but: making the conversion cost `None` here means that we can't detect an overload where the parameter has type `NullPtr` as better than an overload with `Ptr<T>`, even though we should seemingly clearly prefer the former (like how `nullptr_t` works with overloading in C++).

This isn't very important because I'd like to change how we handle `null` in the long run so that we don't have this weird typing behavior.

source/slang/slang-emit-c-like.cpp
```diff
@@ -159,6 +159,14 @@ void CLikeSourceEmitter::emitDeclarator(DeclaratorInfo* declarator)
...
         break;
 
+    case DeclaratorInfo::Flavor::Ref:
+        {
+            auto refDeclarator = (RefDeclaratorInfo*)declarator;
+            m_writer->emit("&");
+            emitDeclarator(refDeclarator->next);
+        }
+        break;
+
```
tangent-vector, line 169: I would have assumed we'd need additional cases here for pointer and function types...

source/slang/slang-emit-c-like.cpp
```diff
@@ -214,6 +220,8 @@ void CLikeSourceEmitter::emitSimpleType(IRType* type)
 
         case kIROp_FloatType:   return UnownedStringSlice("float");    
         case kIROp_DoubleType:  return UnownedStringSlice("double");
+
+        case kIROp_CharType:    return UnownedStringSlice("uint8_t");
```
tangent-vector, line 224: We probably also need to have a design discussion and some written rules about what we want for the whole menagerie of 8-bit integer types. This is especially true in a world where interop with C/C++ probably requires us to be able to interact with functions that use any of `char`, `unsigned char`, and `signed char`, knowing that there are no explicit guarantees about the signed-ness of `char`. Of course that leaves aside the way that most of the time `char` is being used to represent a *byte* rather than a character, and `uint8_t`/`int8_t` don't do that significantly better.

We need to pick some terminology here and then zealously enforce it throughout the compiler stack or we are going to end up with a mess of well-intentioned but broken types.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2059,15 +1999,69 @@ void CPPSourceEmitter::emitSimpleTypeImpl(IRType* inType)
     m_writer->emit(slice);
 }
 
-void CPPSourceEmitter::emitTypeImpl(IRType* type, const StringSliceLoc* nameLoc)
+void CPPSourceEmitter::_emitType(IRType* type, DeclaratorInfo* declarator)
```
tangent-vector, line 2002: I'm confused why this can't just share its logic with the `_emitType` in the base `CLikeSourceEmitter`?

source/slang/slang-emit-cpp.cpp
```diff
@@ -2363,7 +2346,23 @@ bool CPPSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOut
             m_writer->emit(")");
             return true;
         }
-
+        case kIROp_PtrLit:
+        {
+            auto ptrVal = as<IRPtrLit>(inst)->value.ptrVal;
```
tangent-vector, line 2351: We should probably have a distinct IR inst to represent a null pointer, and not just use `IRPtrLit`. In practice we don't expect to ever see a non-null `IRPtrLit` duirng emit anyway (at least I hope not).

source/slang/slang-ir-dll-import.cpp
```diff
@@ -0,0 +1,249 @@
+// slang-ir-dll-import.cpp
+#include "slang-ir-dll-import.h"
```
tangent-vector, line 2: Minor point, but we probably need to think about naming for this attribute and associated functionality over time. `DllImport` is very Windows-specific terminology (even in the C# feature works on other platforms).

source/slang/slang-ir-dll-import.cpp
```diff
@@ -0,0 +1,249 @@
...
+            IRBuilder builder(sharedBuilder);
+            builder.setInsertInto(module->getModuleInst());
+
+            IRType* stringType = builder.getStringType();
+            IRType* paramTypes[] = {builder.getPtrType(builder.getVoidType()), stringType};
+            loadFuncPtrFunc = createBuiltinIntrinsicFunc(
+                2,
+                paramTypes,
+                builder.getPtrType(builder.getVoidType()),
+                UnownedStringSlice("_slang_rt_load_dll_func($0, $1)"));
```
tangent-vector, line 63: This seems a bit messy, but I agree that the alternative isn't ideal either.

The "correct" way to handle this would, I believe, be to have a specific IR instruction opcode for the `_slang_rt_load_dll_func` operation, and to simply emit an instruction with that opcode directly instead of emitting a call to a generated builtin function. That approach avoids a lot of the boilerplate here, at the cost of needing to have emit logic for uses of that new instruction (which ideally isn't too bad because it can be put in `CLikeSourceEmitter` and shared).

I like the instruction option better than any alternative I can think of to try to generalize the pattern going on here...

source/slang/slang-parser.cpp
```diff
@@ -6296,6 +6302,7 @@ namespace Slang
         _makeParseExpr("This",  parseThisTypeExpr),
         _makeParseExpr("true",  parseTrueExpr),
         _makeParseExpr("false", parseFalseExpr),
+        _makeParseExpr("nullptr", parseNullPtrExpr),
```
tangent-vector, line 6305: I understand why you use `nullptr` here, but I would personally greatly prefer `null`. I suspect that we will end up needing `nullptr` as well, even if we add `null`, but just copy-pasting the C++ name for things like this always worries me.

source/slang/slang-type-system-shared.h
```diff
@@ -19,6 +19,7 @@ namespace Slang
     X(Half)                     \
     X(Float)                    \
     X(Double)                   \
+    X(Char)                     \
```
tangent-vector, line 22: As mentioned earlier, I think we really need to have a discussion of what we need/want from a type like this, and clearly document it.

tangent-vector (approved):

source/slang/slang-emit-c-like.cpp
csyonghe: These are placed under the CPPEmit override. We can't get rid of that for now because of the existing logic of emitting `FixedArray<T>` and `Array<T>` in C++, and this is different from the default implementation.

Once we clean this up we should be able to just put all these logic in CLikeEmit. But before then I would just implement it in CppEmit to reduce the chance of running into issues.

source/slang/slang-parser.cpp
csyonghe: I reserved `null` for `ref` classes, and `nullptr` is just for pointers. This is likely unnecessary, but we can change it if we want.

PR #2183, opened 2022-04-11
GFX renaming work part 1

lucy96chen:
Changes:
- Added new typdefs `Size`, `Offset`, `GfxCount`, and `GfxIndex` to `slang-gfx.h` (distinctions for gfx made to avoid conflict with otherwise similarly named typedefs in Slang)
- Replaced numerous instances of `size_t` in `slang-gfx.h`, `render-d3d12.h/cpp`, `render-vk.h/cpp` and `renderer-shared.h/cpp` with `Size`/`Offset` where appropriate
- Fixed numerical conversion warnings in `texture-types-tests.cpp` and `copy-texture-tests.cpp`

@csyonghe @tangent-vector Checkpoint PR, mostly. Lots of renames to go through so the more eyes checking I haven't missed any size/offset `size_t`s, the better

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -31,6 +31,11 @@
 #undef Always
 #undef None
 
+// GLOBAL TODO: doc comments
+// GLOBAL TODO: Rationalize integer types (not a smush of uint/int/Uint/Int/etc)
+//    - need typedefs in gfx namespace for Count, Index, Size, Offset (ex. DeviceAddress)
+//    - Index and Count are for arrays, and indexing into array - like things(XY coordinates of pixels, etc.)
+//    - Offset and Size are almost always for bytes and things measured in bytes.
 namespace gfx {
 
 using Slang::ComPtr;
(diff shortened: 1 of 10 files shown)


slang-gfx.h
```diff
@@ -41,6 +46,10 @@ typedef SlangResult Result;
 typedef SlangInt Int;
```
csyonghe, line 46: Let's define this to be `int32_t` and `uint32_t` for now.

lucy96chen: I tried fiddling around with this change and it causes some major issues because there's also a `typedef SlangInt Int;`, which results in a large number of ambiguous symbol errors in some of my test files. I think this would be better left as a separate PR, especially since many of these could potentially be replaced with `Index` or `Count`.

csyonghe: I see. Can we at least make Index and Count 32bit then? I know Tess likes 64bit ints, but I am worried that using 64bit across the board could make things slower due to their larger memory footprint.

lucy96chen: Also causes similar issues at the moment. There's a typedef in `slang-common.h` for both `Index` and `Count`. I think it'd still be better to leave this for a separate PR and properly hash out how to handle all these integer types and what to typedef them as first.

lucy96chen: Ahh never mind, I'll just make the change here since nothing will actually be affected by this typedef rename. I'll leave the actual renaming changes throughout the rest of the header for later.

csyonghe (approved):

PR #2184, opened 2022-04-12
Make translation units in the same CompileReq visible to `import`.

csyonghe:
This changes makes translation units processed in a front end compile request visible to `import` decls in the later translation units in the same compile request.

This allows the user to mix generated sources and file sources in a same compile request, without having to output the generated source to file first.

The main body of this change is to propagate a list of additional modules to use as loaded modules from the `FrontEndCompileRequest` all the way to `Linkage::findOrImportModule()`.

### DIFF
--- a/source/slang/slang-check.cpp
+++ b/source/slang/slang-check.cpp
@@ -164,10 +164,11 @@ namespace Slang
         TranslationUnitRequest* translationUnit,
         LoadedModuleDictionary& loadedModules)
     {
-        SharedSemanticsContext sharedSemanticsContext(translationUnit->compileRequest->getLinkage(),
-                                                      translationUnit->getModule(),
-                                                      translationUnit->compileRequest->getSink(),
-                                                      &loadedModules);
+        SharedSemanticsContext sharedSemanticsContext(
+            translationUnit->compileRequest->getLinkage(),
+            translationUnit->getModule(),
+            translationUnit->compileRequest->getSink(),
+            &loadedModules);
 
         SemanticsDeclVisitorBase visitor(&sharedSemanticsContext);
 
--- a/source/slang/slang-check.cpp
+++ b/source/slang/slang-check.cpp
@@ -161,12 +161,13 @@ namespace Slang
     }
 
     void checkTranslationUnit(
-        TranslationUnitRequest* translationUnit)
+        TranslationUnitRequest* translationUnit,
+        LoadedModuleDictionary& loadedModules)
     {
-        SharedSemanticsContext sharedSemanticsContext(
-            translationUnit->compileRequest->getLinkage(),
-            translationUnit->getModule(),
-            translationUnit->compileRequest->getSink());
+        SharedSemanticsContext sharedSemanticsContext(translationUnit->compileRequest->getLinkage(),
+                                                      translationUnit->getModule(),
+                                                      translationUnit->compileRequest->getSink(),
+                                                      &loadedModules);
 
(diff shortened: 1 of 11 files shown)


source/slang/slang-check.cpp
```diff
@@ -161,12 +161,13 @@ namespace Slang
...
+        LoadedModuleDictionary& loadedModules)
     {
-        SharedSemanticsContext sharedSemanticsContext(
-            translationUnit->compileRequest->getLinkage(),
-            translationUnit->getModule(),
-            translationUnit->compileRequest->getSink());
+        SharedSemanticsContext sharedSemanticsContext(translationUnit->compileRequest->getLinkage(),
+                                                      translationUnit->getModule(),
+                                                      translationUnit->compileRequest->getSink(),
+                                                      &loadedModules);
```
tangent-vector, line 170: We don't have a lot of enforced coding-convention stuff, but this is one of the few cases where I'm inclined to get militant. The "indent all subsequent arguments to match the indent of the first" style is not robust against find-replace changes of type/function names, while the "indent all arguments by one level" style is.

I assume this was introduced by your editor, and that we should put more settings into the `.editorconfig` or whatever files we can to get better matching. I think the latest Visual Studio supports putting more of the code formatting settings in `.editorconfig` so we should try to leverage that.

source/slang/slang-compiler.h
```diff
@@ -1568,7 +1568,9 @@ namespace Slang
                 Slang::getHashCode(elementType), Slang::getHashCode(containerType));
         }
     };
-    
+
+    typedef Dictionary<Name*, Module*> LoadedModuleDictionary;
```
tangent-vector, line 1572: Comment?

source/slang/slang.cpp
```diff
@@ -2689,6 +2693,16 @@ RefPtr<Module> Linkage::findOrImportModule(
...
 
+    // If the user is providing an additional list of loaded modules, we find
+    // if the module being imported is in that list. This allows a translation
+    // unit to use previously checked translation units in the same
+    // FrontEndCompileRequest.
+    Module* previouslyLoadedModule = nullptr;
+    if (loadedModules && loadedModules->TryGetValue(name, previouslyLoadedModule))
+    {
+        return previouslyLoadedModule;
+    }
```
tangent-vector, line 2704: The one thing this doesn't handle is the case where the user specifies something like a single compile with:

* Translation unit 0 is for a module named `A` and includes the code `import B`
* Translation unit 1 is for a module named `B`

Because we process the translation units in the order they are specified, the `import B` in translation unit 0 will not be able to see module `B` because it hasn't been compiled/loaded yet. The user might be a bit surprised by that fact.

This implementation is fine for unblocking the user who requested this feature, but we should probably put some kind of `TODO` comment in here noting that a more complete long-term solution should detect cases where an earlier module in a compile is trying to import a later one from the same compile and either:

* Issue an error, indicating that the requested module was found, but isn't loaded yet because the compiler respects the user-specified order

* Allow such cases, by demand-loading the later translation unit and doing a recursive compile (just like we do for the more ordinary `import` case), and then skipping that translation unit when the compiler gets to it in the ordinary loop

The second option is implementable in principle, so it is *probably* the one we'd want to do in the long run, but it sounds kind of annoying.

The first option could actually be added to this solution pretty easily by installing a `nullptr` in `loadedModules` for all the module names specified in a compilation, and then replacing those entries as they get loaded. The logic here could then detect the null and issue an error.

tangent-vector (approved):

tangent-vector (approved):

PR #2186, opened 2022-04-14
DXIL library support and Artifact type

jsmall-nvidia:
This PR adds support for compiling to a DXIL library.

I addition it adds the `Artifact` type. From the comments:

The Artifact type is a type designed to represent some Artifact of compilation. It could be input to or output from a compilation.

An abstraction is desirable here, because depending on the compiler the artifact/s could be

* A file on the file system
* A blob
* Multiple files
* Some other (perhaps multiple) in memory representations 

The artifact uses the Blob as the standard representation of in memory data. 

Some downstream compilers require the artifact to be available as a file system file, or to produce artifacts that are files. The Artifact type allows to abstract away this difference, including the ability to turn an in memory representation into a temporary file on the system file. 

The mechanism also allows for 'Containers' which allow for Artifacts to contain other Artifacts (amongst other things). Those artifacts may be other files. For example a downstream compilation that produces results as well as temporary files could be a Container containing artifacts for

* Diagnostics
* Temporary files (of known and unknown types)
* Files that contain known types
* Callable interface (an ISlangSharedLibrary)

A more long term goal would be to

* Make Artifact an interface (such that it can work long term over binary boundaries)
* Make Diagnostics into an interface (such it can be added to a Artifact result)
* Use Artifact and related types for downstream compiler 
* Ability to convert between representations in an ad hoc manner


source/slang/slang-artifact.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+#if __CYGWIN__
+    SLANG_KIND_EXTENSION(SharedLibrary, "dll")
+#elif SLANG_APPLE_FAMILY
+    SLANG_KIND_EXTENSION(SharedLibrary, "dylib")
+#else
+    SLANG_KIND_EXTENSION(SharedLibrary, "so")
+#endif
+
+#endif
```
tangent-vector, line 47: Should we eventually try to detect all of these extensions in all compile configurations, but then somehow associate them with the appropriate target platform(s) (or in some cases the toolchain(s)) that use them?

source/slang/slang-artifact.cpp
```diff
@@ -0,0 +1,535 @@
...
+    {
+        case CodeGenTarget::Unknown:                return make(Kind::Unknown, Payload::None, Style::Unknown, 0);
+        case CodeGenTarget::None:                   return make(Kind::None, Payload::None, Style::Unknown, 0);
+        case CodeGenTarget::GLSL_Vulkan:
+        case CodeGenTarget::GLSL_Vulkan_OneDesc:
+        case CodeGenTarget::GLSL:
+        {
+            // For the moment we make all just map to GLSL, but we could use flags
+            // or some other mechanism to distinguish the types
+            return make(Kind::Text, Payload::GLSL, Style::Kernel, 0);
```
tangent-vector, line 90: It is almost certainly not necessary to differentiate these. The `*_Vulkan*` variants are basically just there as a legacy thing (in fact ,the `OneDesc` variant was only created as part of research work as part of the original Slang paper). We should really remove those from `CodeGenTarget` (by marking them as `DEPRECATED_*` or something like that).

source/slang/slang-compiler.cpp
```diff
@@ -1193,24 +1193,46 @@ void printDiagnosticArg(StringBuilder& sb, CodeGenTarget val)
...
-            auto profile = getEffectiveProfile(entryPoint, getTargetReq());
+            const auto entryPointIndicesCount = entryPointIndices.getCount();
+             
+            if (entryPointIndicesCount == 0 && compilerType == PassThroughMode::Dxc)
+            {
+                // Can support no entry points on DXC because we can build libraries
+                profile = targetReq->getTargetProfile();
+            }
+            else if (entryPointIndicesCount == 1)
+            {
```
tangent-vector, line 1213: This is reasonable for now, but I think we discussed separately how it would be good to change the `CodeGenContext::EntryPointIndices` type so that we can detect whether are are in "single entry point mode" or in "library mode" (aka "whole program" mode in the current naming). We ideally want to allow "whole program mode" compilation for DXIL even when there are one or more entry points.

(Note that at this point in compilation the automatic detection of entry points has already happened in the Slang front-end, so we might have `entryPointIndicesCount > 1` here even in "library mode," unless I'm missing something)

While the above would be the best long-term fix, it seems that in the short term this could at least be written to check `entryPointIndicesCount != 1` as a way to handle the 0 entry point and >1 entry point cases.

tangent-vector (approved):

source/slang/slang-compiler.cpp
jsmall-nvidia: I do use the 'whole program' mechanism in my later PRs.

PR #2191, opened 2022-04-15
Move public API headers

tangent-vector:
This change moves the various public API headers at the root of the repository under `include/slang/`. In order to not break (much) existing code, forwarding headers for `slang.h` and `slang-gfx.h` were kept at the root of the repository.

The Slang codebase itself made use of the `ComPtr` type and related utilities provided by headers at the root, so various changes were required to the internal code to get it to build with this change.

I know that there are outstanding changes to `slang-gfx.h` in flight, so this change may be best left until after those are completed. It is worth noting that this change does *not* alter the contents of the any of the headers, so merges should be navigable just by clobbering the files in their new locations.


tangent-vector:
That’s a good point about gfx. Should we consider moving that headed to be at `include/gfx/gfx.h`?

jsmall-nvidia:
I guess I worry `gfx` is too generic a sub folder name, `slang-gfx` I think would be preferable.

csyonghe:
We need to make sure to update our github action config to include these additional headers in our release packages.

tangent-vector:
Closing for now, since subsequent changes to the headers would cause merge pain here.

PR #2194, opened 2022-04-19
GFX renaming work part 2: slang-gfx.h renames

lucy96chen:
Changes:
- Numerous changes in `slang-gfx.h` to use the previously introduced `GfxIndex`, `GfxCount`, `Size`, and `Offset` typedefs where appropriate
- Updated a large number of additional header and implementation files to reflect these changes in the gfx header
- Updated some renames from the last pass that were missed inside implementation files
- Added casts where appropriate to avoid type conversion warnings
- Renamed `ITextureResource::Size` (which contains width/height/depth values) to `ITextureResource::Extents` to avoid confusion with the new `Size` typedef

Lots of renames in lots of files just to get everything working again, definitely will want checks to make sure there isn't anything that's been misnamed. (I'm less concerned about missed renames at the moment.)

@csyonghe @tangent-vector

### DIFF
--- a/slang-gfx.h
+++ b/slang-gfx.h
@@ -664,7 +664,7 @@ public:
         int quality = 0;                        ///< The quality measure for the samples
     };
 
-    struct Size
+    struct Extents
     {
         GfxCount width = 0;              ///< Width in pixels
         GfxCount height = 0;             ///< Height in pixels (if 2d or 3d)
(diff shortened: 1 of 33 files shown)


slang-gfx.h
```diff
@@ -46,8 +47,8 @@ typedef SlangResult Result;
 typedef SlangInt Int;
```
csyonghe, line 47: Do we still use these? If not we can probably just get rid of it?

csyonghe (approved):

slang-gfx.h
lucy96chen: Unsure, but no real harm in leaving it in for now, I think

PR #2195, opened 2022-04-19
Make artifact an interface

jsmall-nvidia:
On top of #2193 

* Added IArtifact interface
* Added IArtifactInstance interface

Use interfaces such that artifacts can used across ABI boundaries. Added IArtifactInstance to provide a mechanism for associating type/s that are not interfaces, including the ability to blob/serialize them.

A small change was made around ComObject, such that assignment *didn't* do anything with ref counts, and the super classes constructor is chained.


source/compiler-core/slang-artifact-info.h
```diff
@@ -0,0 +1,108 @@
+// slang-artifact.h
```
csyonghe, line 1: slang-artifact-info.h

source/compiler-core/slang-visual-studio-compiler-util.cpp
```diff
@@ -250,16 +252,34 @@ namespace Slang
...
+        auto desc = artifact->getDesc();
+
+        if (ArtifactInfoUtil::isCpuBinary(desc) && desc.kind == ArtifactKind::Library)
+        {
+            // Get the libray name and path
+            SLANG_RETURN_ON_FAIL(artifact->requireFileLike(ArtifactKeep::No));
+
+            libPathPool.add(ArtifactInfoUtil::getParentPath(artifact));
+            // We need the extension for windows
+            cmdLine.addArg(ArtifactInfoUtil::getBaseName(artifact) + ".lib");
```
csyonghe, line 274: The last time I played around msvc command line here, splitting the path out in `/libpath:` section didn't not work. I'd make sure this code works for the heterogeneous-hello-world example.

csyonghe (commented):
Could you make sure this change doesn't break heterogeneous-hello-world by running
```
slangc .\examples\heterogeneous-hello-world\main.slang -o main.exe
```

source/compiler-core/slang-visual-studio-compiler-util.cpp
jsmall-nvidia: Can confirm the heterogeneous-hello-world example works.

csyonghe (approved):

Issue #2087, opened 2022-01-20
Operator overloading doesn't work in user space


jsmall-nvidia:
Slang has support for operator overloading and this is used in the standard library extensively.

Unfortunately in user code (ie outside stdlib) there is a problem, in that any operator overloading in effect replaces what's in the stdlib, making that operation on a built in types no longer visible.

As a (somewhat silly) example 

```HLSL
struct Vec2d
{   
    double x, y;
};

Vec2d operator+(Vec2d a, Vec2d b)
{
    return {a.x + b.x, a.y + b.y};
}
``` 

Fails to compile with

> .slang(35): error 30019: expected an expression of type 'Vec2d', got 'double'
>    return {a.x + b.x, a.y + b.y};

PR #2199, opened 2022-04-21
Overloaded name lookup fix

jsmall-nvidia:
This addresses two recent issues

* operator overloading not working in user space ( #2087)
* a function not being able to access a function outside with the same name, but different signature

The problem occurred in lookup in `_lookUpInScopes` where as soon as a match was found, lookup would terminate. This is not unreasonable for some declarations - like variables, but meant that declarations that could be overloaded, the other overloads would never be seen.

The fix here is to allow lookup in other scopes if the decl variety is overloadable.

Whilst this appears to fix the immediate problem, it raises a variety of questions which I'll discuss in a subsequent issue.


source/slang/slang-lookup.cpp
```diff
@@ -752,6 +752,31 @@ static void _lookUpMembersInValue(
...
+// definitions with the same name)
+// 
+// For example functions are overloadable, but variables are not.
+static bool _isDeclOverloadable(Decl* decl)
+{
+    // TODO(JS): Do we need to special case around ConstructorDecl? or AccessorDecl?
+    // It seems not as they are both function-like and potentially overloadable
+
+    // If it's callable, it's a function and so overloadable 
+    if (auto callableDecl = as<CallableDecl>(decl))
```
csyonghe, line 765: Would be best to turn this into a switch so the code can be easily extended to handle other cases in the future.

csyonghe (approved):

jsmall-nvidia:
On doing some more testing around name lookup, this does change slang name lookup in perhaps 'surprising` ways.

Take 

```HLSL
float a(float x) { return x +1.0; }

namespace N
{
    int a = 3;
    struct Foo
    {
        float a(int x) { return x * x; }
        float test(int v) { return a(v + 3.0f); }
    };
}


[numthreads(4, 1, 1)]
{    
    int index = dispatchThreadID.x;
    
    N::Foo f;
    int v = f.test(index);
}
```

Previously slang would terminate name lookup as soon as a match was found. With the PR if the thing looked up is overloadable (the method a is), then search would continue into other scopes.

Previous version of Slang would therefore call `Foo::a`. Whereas with the change, it will call the free function `a` because it's a closer parameter match. 

I think a case can be made that this is probably not the desirable long term behavior. It may be acceptable for now whilst the correct behavior is determined.

tangent-vector:
We finished out conversation about the desired behavior before getting through that case, so I'm not surprised. :)

We can talk about this more 1:1, or we can have the discussion on GitHub (probably want to open an issue rather than just have it here on the PR).

jsmall-nvidia:
Sounds good. I found that specific scenario whilst writing up an issue, but since it *might* change user code behavior I thought it a good idea to write into this PR too.

PR #2215, opened 2022-04-27
Fix errors when building with the latest Xcode

vincentisambart:
Hello,
As I tried to build this project with the default settings on my Mac with the latest version of Xcode (13.3.1), I got errors about unused variables and copy assignment operator needing to be explicit. Here are a excerpt:
```
../../../source/slang/slang-check-shader.cpp:1183:16: error: variable 'result' set but not used [-Werror,-Wunused-but-set-variable]
        Result result = SLANG_OK;
               ^
In file included from ../../../source/slang/slang-compiler.cpp:13:
In file included from ../../../source/slang/slang-compiler.h:11:
../../../source/slang/../compiler-core/slang-include-system.h:14:5: error: definition of implicit copy assignment operator for 'SearchDirectory' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated-copy]
    SearchDirectory(SearchDirectory const& other) = default;
    ^
../../../source/slang/../core/slang-list.h:242:42: note: in implicit copy assignment operator for 'Slang::SearchDirectory' first required here
                            newBuffer[i] = m_buffer[i];
                                         ^
../../../source/slang/../core/slang-list.h:276:46: note: in instantiation of member function 'Slang::List<Slang::SearchDirectory>::insertRange' requested here
        void addRange(const List<T>& list) { insertRange(m_count, list.m_buffer, list.m_count); }
                                             ^
../../../source/slang/../core/slang-list.h:70:13: note: in instantiation of member function 'Slang::List<Slang::SearchDirectory>::addRange' requested here
            addRange(list);
            ^
../../../source/slang/../core/slang-list.h:48:19: note: in instantiation of member function 'Slang::List<Slang::SearchDirectory>::operator=' requested here
            this->operator=(list);
                  ^
../../../source/slang/../compiler-core/slang-include-system.h:23:8: note: in instantiation of member function 'Slang::List<Slang::SearchDirectory>::List' requested here
struct SearchDirectoryList
       ^
```

This PR fixes all those errors, though I am not sure about what to do about unused variables. I just went the simplest route, and removed them, but they were probably left for a reason (especially the `slang-parser.cpp` one).
Another solution would be to use `[[maybe_unused]]`. It's C++17 but in theory old older compilers that should be simply ignored. Though with your high level of warnings and warnings treated as errors it might cause an error, so a conditional `#define` might be needed.

### DIFF
--- a/source/compiler-core/slang-include-system.h
+++ b/source/compiler-core/slang-include-system.h
@@ -15,6 +15,7 @@ struct SearchDirectory
     SearchDirectory(String const& path)
         : path(path)
     {}
+    SearchDirectory& operator=(SearchDirectory const& other) = default;
 
     String path;
 };
--- a/source/core/slang-smart-pointer.h
+++ b/source/core/slang-smart-pointer.h
@@ -285,6 +285,7 @@ namespace Slang
         TransformablePtr(T* ptr) { *this = ptr; }
         TransformablePtr(RefPtr<T> ptr) { *this = ptr; }
         TransformablePtr(const TransformablePtr<T>& ptr) = default;
+        TransformablePtr<T>& operator=(const TransformablePtr<T>& ptr) = default;
 
         void promoteToStrongReference() { m_strongPtr = m_weakPtr; }
         void demoteToWeakReference() { m_strongPtr = nullptr; }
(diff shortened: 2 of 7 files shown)


csyonghe:
This looks good. I think getting rid of the unused local variable assignments is fine. Or alternatively we can use `SLANG_UNUSED(x)` macro that is originally intended for function parameters.

csyonghe (dismissed):

vincentisambart:
I rebased and used `SLANG_UNUSED` on `source/slang/slang-parser.cpp` to better match comments.

PR #2220, opened 2022-05-04
Turn off warnings-as-errors for gfx

tangent-vector:
Other than on Windows, we will go ahead and turn off warnings-as-errors
for the gfx project, to help clean up some of our CI build mess.


lucy96chen (commented):
Sounds good to me, but you should probably check with Yong since I turned them on initially at his request.

PR #2222, opened 2022-05-05
Various vulkan/glsl fixes.

csyonghe:
1. Include type constraint in a generic's mangled name.
2. define `dot` for integer vector types for glsl target.
3. Automatically use a format-specific default aspect mask when the user does not specify the aspect in `copyTexture`.
4. Specify reduction mode when creating a `VkSampler`.


source/slang/slang-check-decl.cpp
```diff
@@ -5342,4 +5344,49 @@ namespace Slang
...
+        OrderedDictionary<GenericTypeParamDecl*, List<Type*>> result;
+        for (auto& constraints : genericConstraints)
+        {
+            List<Type*> typeList;
+            for (auto type : constraints.Value)
+            {
+                _getCanonicalConstraintTypes(typeList, type);
+            }
+            // TODO: we also need to sort the types within the list for each generic type param.
+            result[constraints.Key] = typeList;
```
jsmall-nvidia, line 5387: Seems pretty important to be able to be canonical.

PR #2223, opened 2022-05-05
Support for HLSL `export`

jsmall-nvidia:
Allows use of `export` keyword in slang. The usage is somewhat limited to be something similar to HLSL usage. As touched on in some of the comments this is distinct concept but has some overlap to `public`. 

As  discussed with @tangent-vector we may need to refine how this works for the language more long term.


source/slang/slang-lower-to-ir.cpp
```diff
@@ -1047,6 +1047,11 @@ static void addLinkageDecoration(
         builder->addPublicDecoration(inst);
         builder->addKeepAliveDecoration(inst);
     }
+    if (decl->findModifier<HLSLExportModifier>())
+    {
+        builder->addHLSLExportDecoration(inst);
+        builder->addKeepAliveDecoration(inst);
+    }
```
tangent-vector, line 1054: It isn't specific to this PR, but I definitely get worried seeing the sheer number of cases here that are adding decorations that appear to be related to linkage, but that aren't in any way tied to our current `[import(...)]` and `[export(...)]` decorations.

The `PublicModifier`/`PublicDecoration` stuff is just plain wrong, and we will need to get rid of them sooner or later in order to be able to make the `public` keyword in the language actually do anything useful.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5408,6 +5413,10 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
 
     bool isPublicType(Type* type)
     {
+        // TODO(JS):
+        // Not clear how should handle HLSLExportModifier here. 
+        // In the HLSL spec 'export' is only applicable to functions. So for now we ignore.
+
```
tangent-vector, line 5419: If a type can't be marked `export`, then it seems like there is nothing to be done here?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5595,6 +5604,11 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         irWitnessTable->setOperand(0, irSubType);
 
         addLinkageDecoration(context, irWitnessTable, inheritanceDecl, mangledName.getUnownedSlice());
+
+        // TODO(JS):
+        // Not clear what to do here around HLSLExportModifier. 
+        // In HLSL it only (currently) applies to functions, so perhaps do nothing is reasonable.
+
```
tangent-vector, line 5611: Similarstatement here: I'm not sure why we'd need to process `export` modifiers on inheritance clauses...

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6340,7 +6354,11 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
             SLANG_UNREACHABLE("associatedtype should have been handled by visitAssocTypeDecl.");
         }
 
-        bool isPublicType = decl->findModifier<PublicModifier>() != nullptr;
+        // TODO(JS): 
+        // Not clear what to do around HLSLExportModifier. 
+        // The HLSL spec says it only applies to functions, so we ignore for now.
+
+        const bool isPublicType = decl->findModifier<PublicModifier>() != nullptr;
```
tangent-vector, line 6361: I  think the long/short here is that the presence/absence of `export` shouldn't affect whether something is considered "public" from the Slang perspective *at all* (at least that's a reasonable starting point...).

There's a bunch of messy stuff confounding the discussion here because "public" is currently getting used to mean a few different things inside of the Slang codebase, and the current `public` keyword, `PublicModifier`, and `PublicDecoration` all mean something pretty much categorically *wrong*.

tests/library/export-test.slang
```diff
@@ -0,0 +1,27 @@
...
+// This didn't work for lib_6_2 when compiling via DXC (!). Even though it's stated elsewhere the feature is available from 6.1
+
+//TEST:COMPILE: tests/library/export-library.slang -profile lib_6_3 -target dxil -o tests/library/export-library.dxil
+//TEST:COMPILE: tests/library/export-test.slang -entry computeMain -profile cs_6_3 -target dxil -r tests/library/export-library.dxil -o tests/library/export-test.dxil 
+
+// Test the produced kernel.
+
+//TEST:COMPARE_COMPUTE_EX:-slang -compute -profile cs_6_3 -dx12 -use-dxil -shaderobj -Xslang... -r tests/library/export-library.dxil -X.
+
+extern int foo(int a);
```
tangent-vector, line 15: Just to confirm: the rules for the HLSL/dxc `export` are that the function needs to be marked as `export` in the translation unit that *defines* it, but doesn't need to be marked `export` in a translation unit that simply declares and then imports it?

tangent-vector (approved):
This seems good. Most of the messy parts are ultimately about the `public` keyword and the relation modifiers/decorations being completely wrong in Slang today. Once we sort those things out, most of the questions here should get easier.

tests/library/export-test.slang
jsmall-nvidia: I believe that is the case.

PR #2226, opened 2022-05-07
Use IR pass to eliminate phi nodes

tangent-vector:
"Phi nodes" are one of the key contrivances that makes SSA (Static
Single Assignment) form work. Because SSA is so great for compiler
IRs, we kind of need to deal with phi nodes, but they also get in
the way because they don't have a direct analog in most lower-level
machine ISAs or execution models, nor in most of the high-level
languages a transpiler wants to emit. As a result a compiler like
ours needs to be able to eliminate the phi nodes from a program as
part of generating output code.

(For any clever people noting that SPIR-V supports phi nodes
directly: yes, it does. It doesn't need to and it probably *shouldn't*.
Anybody involved in the decision-making knows my reasoning, and
anybody else should feel free to ask me if they want the lecture.
Anyway...)

The basic idea of elimiating phi nodes is simple enough. We replace
each phi node with a temporary variable. Uses of the phi use values
loaded from the temporary. The operation of the phi itself
(assigning a value based on the branch taken) amounts to an assignment
into the temporary.

Previously, the Slang compiler dealt with phi nodes very late in
the process of generating code: in the middle of emitting strings
of source code in a high-level language like HLSL or GLSL. Doing the
work that late in compilation has two big drawbacks:

1. Our ability to emit clean and/or optimal code is limited because
we may not be able to make certain changes to the IR, or because we
cannot make use of additional information like a dominator tree that
might be available at other points in compilation.

2. Any other IR passes that relate to temporary variables won't be
able to see the variables that we generate for phi nodes. This could
raise issues with correctness (e.g., if we want to compute live-range
information for *all* temporary variables), or performance (we have no
way to run additional IR optimization passes after phis are eliminated).

This change addresses these problems by making the elimination of
phi nodes an explicit IR pass. Additional optimizations can easily be
run after this pass (although we'd need to be careful not to run
passes that could end up introducing new phis). The pass makes use
of the information available to it to try to produce code that will
emit to "clean" HLSL/GLSL.

The core of the pass is in `slang-ir-eliminate-phis.cpp`, and is
heavily commented, so I won't describe the approach in detail here.
There are two related issues that came up, though:

First, it turned out that our emit logic for local variables (`IRVar`
instructions) wasn't using the function we'd defined named `emitVar()`.
One worrying consequence of that oversight was that the `precise`
modifier would impact generated HLSL/GLSL for variables that turned
into SSA values (including phi nodes), but *not* for local variables
that had not been SSA'd (or that had been SSA'd and then de-SSA'd).
This change also fixes that bug; it is unclear how widespread the
impact of the original issue might be.

Second, generating explicit IR temporaries for phi nodes exposed a
pre-existing bug in the `slang-ir-restructure-scoping` pass. That pass
basically detects cases where we have an instruction `I` with a use
`U` such that the use follows the rules of SSA form ("def dominates
use," meaning `I` dominations `U`), but does not follow the more
restrictive scoping rules of high-level-language output (where a value
computed "inside" a loop is not automatically visible to code outside
the loop just because it dominates that code). That pass did not
correctly account for the case where `I` was a temporary variable.
It seems that case could not arise before now because we didn't have
any passes that would move `var`, `load`, or `store` operations out
of the basic block they started in. The fix for that pass was relatively
simple, and will make the whole thing more robust in case we add more
aggressive optimizations later.

### DIFF
--- a/source/slang/slang-ir-eliminate-phis.cpp
+++ b/source/slang/slang-ir-eliminate-phis.cpp
@@ -1,0 +1,894 @@
+// slang-ir-eliminate-phis.cpp
+#include "slang-ir-eliminate-phis.h"
+
+// This file implements a pass to take code in the Slang IR out out SSA form
+// by eliminating all "phi nodes."
+//
+// The Slang IR does not represent phi operations in the "textbook" way, and
+// it is important to understand how it *does* represent those operations in
+// order to understand this pass.
+//
+// In a textbook encoding of SSA, a basic block begins with zero or more `phi`
+// instructions:
+//
+//      block B:
+//          let x = phi(a, b);
+//          let y = phi(c, d);
+//          ...
+//
+// Each `phi` operation has a number of operands equal to the number of
+// predecessors of `B`, with one operand corresponding to each predecessor.
+// Semantically, we usually think of a `phi` as auto-magically selecting between
+// its operands beased on the control-flow path that was used to arrive at `B`.
+// Note, however, that all of the `phi` operations at the start of a block must
+// be understood as executing *simultaneously*, so that if `x` or `y` above was
+// used as a `phi` operand, they would yield their value from *before* the binding
+// not after. Oh, and also the operands of `phi` instructions are the *one* place
+// where the rule about how "defs must dominate uses" isn't upheld.
+//
+// Our encoding of the same thing is equivalent, more consistent, and in most ways
+// easier to understand. A basic block in the Slang IR can have *parameters*,
+// and a branch to such a block must pass *arguments* for those parameters:
+//
+//      block B(x,y):
+//          ...
+//
+//      block M:
+//          ...
+//          br B(a, c);
+//
+//      block N:
+//          ...
+//          br B(b, d);
+//
+// Note how in this formulation there is no auto-magical behavior. The relationship
+// between a predecessor block and the values that `x, y` take on is clear. The
+// way that `x, y` take on their new values simulataneously is also more apparent
+// (since it intuitively matches a recursive function call). Further, our IR is
+// able to follow the "def dominates use" rule more completely.
+//
+// With that out of the way, let's dive into the pass itself.
+
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+#include "slang-ir-dominators.h"
+
+namespace Slang {
+
+struct PhiEliminationContext
+{
+    // We are going to make some effort to re-use intermediate structures across
+    // different functions that we process, so that we don't do more allocation
+    // than necessary.
+    //
+    // At the top level, our pas needs to have access to the IR module, and needs
+    // a builder it can use to generate code.
+    //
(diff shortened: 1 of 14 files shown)


source/slang/slang-ir-eliminate-phis.cpp
```diff
@@ -0,0 +1,894 @@
...
+        }
+
+        auto newBranch = m_builder.emitIntrinsicInst(
+            oldBranch->getFullType(),
+            oldBranch->getOp(),
+            newOperandCount,
+            newOperands);
+        oldBranch->transferDecorationsTo(newBranch);
+
+        // TODO: We could consider just modifying `branch` in-place by clearing
```
jsmall-nvidia, line 754: Seems like a good plan, from simplicity and space/time reasons..

jsmall-nvidia (approved):
Looks good.

source/slang/slang-ir-eliminate-phis.cpp
```diff
@@ -0,0 +1,894 @@
...
+    // which is handy but also requires careful thought at use sites
+    // about which version of the information is relevant.
+    //
+    struct PhiInfo
+    {
+        ParamInfo param;
+        ArgInfo arg;
+        Count state = kState_Ready;
+    };
+    List<PhiInfo> phiInfos;
```
jsmall-nvidia, line 509: It might be better to make these member variables prefixed. Doing so would make clear at usage sites that they belong to the context. If the class could be accessed externally, and they were publicly accessible (ie 'they are part of the interface') then that might override this consideration.

source/slang/slang-ir-eliminate-phis.cpp
```diff
@@ -0,0 +1,894 @@
...
+            {
+            default:
+                continue;
+
+            case kIROp_Func:
+            case kIROp_GlobalVar:
+                break;
+            }
+
+            auto code = (IRGlobalValueWithCode*)inst;
```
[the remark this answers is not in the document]
tangent-vector: Possibly even better to use the `cast<T>(...)` that we defined for `IRInst`, since that even checks the correctness in debug/assert builds.

source/slang/slang-ir-eliminate-phis.cpp
tangent-vector: Yeah, I bounce back and forth on this stuff, and I know this particular type is actually inconsistent (some stuff using `m_` prefix and some not). My experience is that for context types like this I often go through multiple refactoring steps where things evolve between local variable <-> function parameter <-> member variable of context type, and it is frustrating if exactly one of those cases requires touching up all the use sites.

If C++ supported local functions more nicely this all might be moot.

tangent-vector:
My inclination is to land this work and then do the cleanups later (if we end up touching this code again). Getting this code checked in will give us the necessary starting point if we end up needing live-range information on the temporary variables we create for phi nodes.

PR #2227, opened 2022-05-09
Add support for `spirv_literal` 

jsmall-nvidia:
When outputting liveness ranges in SPIR-V we need to generate instrinsics that map to SPIR-V ops. These declarations have no body.  We want the GLSL output for these intrinsics about to output `spirv_literal` where appropriate. Since these intrinsics don't have a body, we cannot use a decoration on a param.

The approach here is to add a SPIRVLiteralType. This is used in the GLSL liveness pass to indicate to the GLSL emit that the parameters type is `spirv_literal`. 

As it stands the use happens at a very late stage and only for GLSL, so there is little impact on any other parts of the compiler.


source/slang/slang-ir-inst-defs.h
```diff
@@ -195,6 +195,9 @@ INST(RTTIType, rtti_type, 0, 0)
 INST(RTTIHandleType, rtti_handle_type, 0, 0)
 INST(TupleType, tuple_type, 0, 0)
 
+// A type that identifies it's contained type as being emittable as `spirv_literal.
+INST(SPIRVLiteralType, spirvLiteralType, 1, 0)
+
```
tangent-vector, line 200: From my own attempts to look into it, the semantics of `spirv_literal` in GLSL/glslang appear to be something like:

> "When encoding SPIR-V words for a call to the `spirv_instruction` function this parameter belongs to, encode the argument for this parameter as a literal value in the SPIR-V binary form, rather than as the `<id>` of another instruction."

Given that understanding, it seems that neither this type-based approach (`SPIRVLiteral<int>`) nor an encoding as a rate (`@SPIRVLiteral int`) are semantically correct.

I believe the semantically-correct encoding would be to make this be `SPIRVLiteralDecoration`, a subtype of `Decoration`, and then attach that decoration to the `IRParam` in question directly (so *not* part of the type).

This PR works fine, so there is no immediate reason to change anything, but I want to note that in case we need/want to revisit this code.

tangent-vector (approved):

source/slang/slang-ir-inst-defs.h
jsmall-nvidia: I did look at adding as a decoration originally. That such a decoration going in an IRParam would work fine, assuming there was an IRParam instruction. In this case we are adding a function forward declaration, so it has no body/blocks, so has no IRParam? 

Is the idea here to add a block/IRParams, and use some other mechanism to determine it has no body? Or am I missing something?

PR #2230, opened 2022-05-10
Initial support for COM interface in host code.

csyonghe:
This PR adds initial support for defining and using COM interfaces in host Slang code.

It is now valid to write the following Slang code:
```
[COM]
interface IObject
{
    int getValue(int value);
}

[DllImport("test-com")]
IObject createObject();

{
    int rs = object.getValue(2);
    return 0;
}
```

Which will be compiled to:
```
{
    virtual SLANG_NO_THROW int32_t SLANG_MCALL getValue(int32_t param1) = 0;
};
Slang_FuncType<ComPtr<IObject_0>>  _S1_init()
{
    return nullptr;
}
Slang_FuncType<ComPtr<IObject_0>>  _S1 = _S1_init();

ComPtr<IObject_0> createObject_0()
{
    if(*(&_S1) == nullptr)
    {
        void * _S3 = (_slang_rt_load_dll((String("test-com"))));
        void * _S4 = (_slang_rt_load_dll_func((_S3), (String("createObject"))));
        *(&_S1) = (slang_bit_cast<Slang_FuncType<ComPtr<IObject_0>> >(_S4));
    }
    else
    {
    }
    ComPtr<IObject_0> _S5 = (*(&_S1))();
    return _S5;
}

int32_t main()
{
    int32_t _S6 = object_0->getValue(int(2));
    return int(0);
}
```

This change adds a `[COM]` builtin attribute for marking an interface type as a `COM` interface.
All `[COM]` interface values will be translated into a `ComPtr<T>` in the resulting program.
In the future, we should introduce explicit IR operands to increase and decrease the reference counter instead of relying on C++ for this.


source/slang/slang-check-decl.cpp
```diff
@@ -3232,6 +3232,16 @@ namespace Slang
...
         }
+
+        if (decl->findModifier<ComInterfaceAttribute>())
+        {
+            // `associatedtype` declaration is not allowed in a COM interface declaration.
+            for (auto associatedType : decl->getMembersOfType<AssocTypeDecl>())
+            {
+                getSink()->diagnose(
+                    associatedType, Diagnostics::associatedTypeNotAllowInComInterface);
+            }
```
tangent-vector, line 3243: I'm going to push a design proposal for the `[COM] interface` feature shortly, but it is worth noting that we could potentially lift this restriction, and doing so could be beneficial in some cases. For now, there is no reason not to have the restriction, though.

It is also worth noting that there are several other kinds of member declarations we probably want to exclude. For example, neither `property` nor `__init` declarations will make a ton of sense inside a `[COM]` interface until we take the time to define rules for them.

source/slang/slang-ir-com-interface.cpp
```diff
@@ -0,0 +1,98 @@
...
+            case kIROp_WitnessTableIDType:
+            case kIROp_WitnessTableType:
+            case kIROp_ThisType:
+            case kIROp_RTTIPointerType:
+            case kIROp_RTTIHandleType:
+            case kIROp_ComPtrType:
+                continue;
+            default:
+                break;
+            }
```
tangent-vector, line 39: This logic seems a little fiddly. It seems like we are getting burned by the fact that we use the `IRInterfaceType` directly to represent an existential, instead of having something like `IRExistentialType<Interface>` for that case, and then having an `IRInterface` not even *be* an `IRType`. That approach would let us be much more nuanced in a step like this, since really what you are trying to do here is replace *existentials* for a `[COM]` interface with the corresponding `ComPtr<...>` type.

(Ideally, at that point the AST->IR lowering logic would select between different subtypes of `IRExistentialType<...>`, with `ComPtr<...>` being just one case, so that we wouldn't even have to do the logic that is going on here at all.)

source/slang/slang-ir-lower-existential.cpp
```diff
@@ -105,7 +105,16 @@ namespace Slang
             auto builder = &builderStorage;
             builder->setInsertBefore(extractInst);
 
-            auto element = extractTupleElement(builder, extractInst->getOperand(0), elementId);
+            IRInst* element = nullptr;
+            if (extractInst->getOperand(0)->getDataType()->findDecoration<IRComInterfaceDecoration>())
+            {
+                // If this is an COM interface, the elements (witness table/rtti) are just the interface value itself.
+                element = extractInst->getOperand(0);
```
tangent-vector, line 112: This *maybe* warrants a `TODO:` comment about how eventually we might define that the witness table is found by treating the object as a pointer to the witness table (so that the Slang "witness table" is identical to the COM/C++ "virtual function table").

source/slang/slang-lower-to-ir.cpp
```diff
@@ -1056,6 +1056,11 @@ static void addLinkageDecoration(
     {
         builder->addExternCppDecoration(inst, mangledName);
     }
+    if (as<InterfaceDecl>(decl->parentDecl) &&
+        decl->parentDecl->hasModifier<ComInterfaceAttribute>())
+    {
+        builder->addExternCppDecoration(inst, decl->getName()->text.getUnownedSlice());
+    }
```
tangent-vector, line 1063: I am not sure I understand this logic. It looks like this is saying that any IR inst we create for a child of a `[COM] interface` is being given the IR-level `__extern_cpp` decoration, and that is using the un-mangled original declaration name of the member.

AFAICT, the main effect that would have is to ensure that the emitted C++ declarations for a COM interface use the same member function names as the original, but that doesn't seem to be required for binary compatibility. If we emit C++ that calls a method `foo_0` or `_S3foo` instead of `foo`, will it actually break anything?

tangent-vector (approved):
It is great to see this feature taking shape.

source/slang/slang-lower-to-ir.cpp
csyonghe: Right, names shouldn't matter for binary compatibility. But if we want to generate header files from a COM interface exported by Slang, we probably need to use unmangled names?

tangent-vector: I continue to believe that generation of C++ headers based on Slang code will be best approached with something AST-based rather than IR-based.

PR #2231, opened 2022-05-11
Split render-d3d12.h/cpp into a set of smaller files

lucy96chen:
Changes:
`render-d3d12.h` and `render-d3d12.cpp` have both been split out to a set of smaller files each containing a set of related classes and structs from the original header and named based on what the classes and structs they contain are for. For example, `BufferResourceImpl` was moved out to `d3d12-buffer.h` whereas `CommandEncoderImpl` and all its derived classes (render, resource, etc) were moved out to `d3d12-command-encoder.h`. This PR also adds `d3d12-base.h`, which contains forward declarations and includes required across multiple header files (required to deal with potential circular includes), as well as `d3d12-helper-functions.h`, which contains all the helper functions formerly located in `render-d3d12.cpp`. The current division of structs/classes is as follows:

Buffer resource, command buffer, command encoder, command queue, device, fence, framebuffer, query, ray tracing, render pass, resource views, sampler, shader object, shader object layout, shader table, submitter, swap chain, textures, transient heap, vertex layout

@csyonghe


tools/gfx/d3d12/d3d12-base.h
```diff
@@ -0,0 +1,73 @@
...
+#include "../command-encoder-com-forward.h"
+#include "../d3d/d3d-swapchain.h"
+#include "../mutable-shader-object.h"
+#include "../renderer-shared.h"
+#include "../simple-render-pass-layout.h"
+#include "../transient-resource-heap-base.h"
+#include "core/slang-basic.h"
+#include "core/slang-blob.h"
+#include "core/slang-chunked-list.h"
+#include "descriptor-heap-d3d12.h"
```
csyonghe, line 14: Maybe good to rename this to d3d12-descriptor-heap. We should check other files under d3d12 directory and make their naming consistent.

csyonghe (approved):

tools/gfx/d3d12/d3d12-base.h
lucy96chen: Done for both descriptor heap and resource. I'm presuming swapchain and util are used for both D3D11 and D3D12 based on their current names so I left those alone.

csyonghe (dismissed):

PR #2232, opened 2022-05-11
More proposals

tangent-vector:
Checking in more in-progress proposals in the hopes of sparking discussion and/or guiding future implementation work.


csyonghe (approved):

jsmall-nvidia:
Your proposal text doesn't compile without errors ;)

On error handling
===========

The point is brought up about different types of errors - but the rest of the discussion seems to be about the `user handling` scenario. The other scenarios are perhaps worth discussing. 

For example something akin to rusts `panic`. Both as a language construct, a way of reporting kinds of failures. Being able to `handle` in some limited way. What happens to defer? What happens with multiple failures. What data can be carried around. What guarantees does the runtime supply (like closing handles etc).

There is also the kind of markup we see within Slang code. `assert` or marking something as `unexpected`. All these could lead to panics and in debug builds producing some useful information/breaking into debugger etc.

> Typically there is neither a reason nor a desire to do anything about nearly-unrecoverable errors; the program has well and truly crashed.

It seems to me the panic mechanism is reasonable and useful idea over *just* crashing.

## Exceptions

Whilst exceptions introduce a variety of issues, there are a few things going for them

* They have some kind of debugging environment support 
   * None of what is discussed precludes this, but it's actually a pretty important feature
   * We have support in Slang codebase by changing SLANG_ON_... macros
* Exceptions are close to zero cost in most modern implementations if an exception isn't thrown. If an exception is thrown it's cost is considerably more than a return value
 
We may want to support actual exceptions from an ABI point of view to support interop.

The point about ABI issues and exceptions is sound, but like some other recent issues, you can change the behavior 'at the boundary' such that inside the boundary does something different (faster etc) than what is available at the boundary. It is common in C++ code to do this at COM boundary for example.

## Return codes/things

One nice thing about exceptions is that a payload can be varied, and it trafficed to a catch site as is.

A problem(?) with these kinds of return mechanisms, is that because different functions have different return types typically, there are mechanisms that translate the return type as the stack is unwound. This requires more programmer effort (to do the conversions), information can be lost, and requires more code (to traffic between types) as the stack unwinds.

Secondly it implies returning multiple things - so returning a struct for example. In Rust there is the `nullptr` optimization such that null can mean false on Optional. But that relies on pointers not being able to be null. If more than a pointer is needed, then the return type becomes bigger than a register, and this typically leads to a function returning a struct. This leads to much less efficient code as returning this way typically means through memory.

A work around might be returning an HRESULT like thing, and making the rest of the return value an out parameter. If the HRESULT indicates failure it can imply features of the out structure (such as it's invalid).

## Go

Go has a by convention error return "system" through tuples. Having not developed much go, not sure what the pros and cons are around that choice. It is a far simpler language feature than what is being described here.

## Other approaches

* Mapping all 'error cases' to int type, or say a pointer to static memory 
* Using "thread local" memory as a side channel
 
On the last point, note that thread local memory doesn't actually have to be through thread local accesses (which are comparatively slow), if a context is threaded through an added function parameter. So as an example (with bad syntax to illustrate)..

```
float smallAndCheap(float f)
{
   // I can still *access* the context to write an error/panic/go to a handler etc
   // as can read the context from thread local memory (slow - but this is exceptional)
   // It can't 'throw' as in cause myFunc to unwind though
}

[can_throw] int myFunc(...)
{
   auto smallAndCheap(...);
}
```

The function with [can_throw] by it's signature is saying this from ABI point of view, threads a 'context'.

How does this context work across an ABI boundary - because thread localness is something that has ABI issues? Perhaps a context has to always be passed. 

A thread local context can be handy in lots of ways...

How does this work with multithreading? The context belongs to the thread. If an op happens that can change threads, post that point any state tied to the context would have to be reloaded.

jsmall-nvidia:
As a quick extra thought it is probably worth at least considering how 'error' codes could map down to something like a HRESULT. 

Having a single type, with a well defined structure of course has some significant limitations (and no or very limited 'payload'). But it does mean that trafficing between error types across boundaries isn't required. If the code is 64 bits (rather than COMs 32), that provides quite a lot of extra bits to make error codes work across large code bases, in a modular fashion. One of the issues with 32 bit HRESULT is the 'facility' has relatively few bits. You don't want facilities to clash across modules. 

Another problem, that could be somewhat addressed with language support would be mapping from 'HRESULT's back to user understandable strings.  

HRESULT like type also encourages the defining and using of commonly used error code values.

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+
+Proposed Approach
+-----------------
+
+We propose to allow COM interfaces to be declared using the Slang `interface` construct, with an appropriate attribute or modifier:
+
+```
+[COM] interface IDevice
+{
+    ITexture createTexture(__read TextureDesc desc) throws HRESULT;
```
jsmall-nvidia, line 43: I don't know about this. Seems more COM if it's more like...

```
Result createTexture(__read TextureDesc desc, out Ptr<ITexture>outTex);
```

By making it return the interface avoids some issue, but what happens if I want to return multiple interfaces? How do I indicate ownership for these interfaces? The ownership is an issue if you want to be able to generate code to use the interface. 

Or how do I output lots of textures?

```
Result createTextures(const TextureDesc* desc, Count textureCount, ITexture** outTexs);
```

Taking a look at the microsoft IDL compiler syntax might be useful in bringing up some of the more obscure edge cases around communicating the interface.

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+{
+    virtual HRESULT SLANG_MCALL createTexture(TextureDesc const& desc, ITexture** _result) = 0;
+
+    virtual void SLANG_MCALL setTexture(int index, ITexture* texture) = 0;
+};
+```
+
+Key things to note:
+
+* The `[COM] interface` becomes a C++ `struct` that inherits from `IUnknown`
```
jsmall-nvidia, line 62: Pedantically why not a `class`?

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+* Methods defined in the `[COM] interface` become pure-virtual `SLANG_MCALL` methods in C++
```
jsmall-nvidia, line 63: Potentially need 'no throw' macro too.

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+
+```
+class MyTexture : public slang::Object, public ITexture
+{
+    // ...
+};
+
+class MyDevice : public slang::Object, public  IDevice
+{
+    virtual HRESULT QueryInterface(REFIID riid, void **ppvObject) { ... }
```
jsmall-nvidia, line 99: public:

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+    virtual ULONG AddRef() { ... }
```
jsmall-nvidia, line 100: In IDL, there is markup to say `ppvObject` (aka `outObject` ;0) is of the type specified in riid.

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+
+All Slang `class`es translate to C++ classes that inherit from `slang::Object` (equivalent to the `RefObject` type within the current Slang implementation).
+A `class` that inherits any `[COM]` interfaces includes an implementation of `IUnknown` plus the methods to override all the interface requirements.
+
+In ordinary code that makes use of `[COM] interface` types:
+
+```
+struct Stuff
+{
+    ITexture t;
```
jsmall-nvidia, line 124: Hmmm. I guess we are saying [COM] types are *reference* types implicitly?
In Stuff does it own the pointer or not?

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Should we emit COM code that works at the plain C level, or idiomatic C++?
+
+I honestly don't know. Emitting idomatic C++ (and using things like smart pointers) certainly makes the output code easier to understand.
+
+### Can we make this work with more advanced features of Slang interfaces?
+
+Some Slang features don't have perfect analogues. For example, given that `[COM] interface`s can only be conformed to by `class` types, the use of `[mutating]` isn't especially meaningful.
+
+There is no reason why a `[COM] interface` couldn't make use of `static` methods, but there would be no way to call those from C++ without an instance of the interface type.
```
jsmall-nvidia, line 202: You could have a static member of the interface, but that is problematic. You could bundle up all of the static functions into an interface, that you can access via some GetStaticInterface.. like function. That could return an array of function pointers. Having it return an interface is probably reasonable.

jsmall-nvidia (approved):

docs/proposals/004-com-support.md
```diff
@@ -0,0 +1,240 @@
...
+The .NET platform includes some support for allowing .NET `interface`s and COM interfaces to interoperate.
+TODO: Need to study this and learn how it works.
+
+Proposed Approach
+-----------------
+
+We propose to allow COM interfaces to be declared using the Slang `interface` construct, with an appropriate attribute or modifier:
+
+```
+[COM] interface IDevice
```
jsmall-nvidia, line 41: Shouldn't this specify the guid? 

```
[COM(guid)] 
interface IDevice ...
```

If guid isn't specified, it could generate a guid from a hash of the fully scoped name.

docs/proposals/004-com-support.md
tangent-vector: I suspect that we should have a plan within this `[COM] interface` stuff for how the `REFIID riid, void** ppvObject` idiom can be expressed in terms of generics. *If* we were to define `QueryInterface` at the Slang level, I would like it if it could look something like:

```
T QueryInterface< interface T : IUnknown >() throws HRESULT;
```

The basic idea being that `QueryInterface` either signals a failure through the `HRESULT` or it returns a valid instance of `T`. The pitch here would be that `<interface T>` would be a way to define a generic parameter that only accepted `interface` types as arguments (and the `IUnknown` constraint ensures that `T` is a sub-interface of `IUnknown` and thus a `[COM]` interface).

There are certainly various pieces of the D3D and DXGI APIs that rely on the `REFIID, void**` idiom, so being able to translate those over to something idiomatic in Slang would be great.

docs/proposals/004-com-support.md
tangent-vector: > I guess we are saying [COM] types are reference types implicitly?

It is more accurate to say that only reference types can conform to a `[COM]` interface.

> In `Stuff` does it own the pointer or not?

In terms of memory management, `Stuff` retains a reference to whatever `t` refers to. Not that the answer there doesn't care whether `ITexture` is a `[COM]` interface or not. An `IWhatever` existential that holds a value of `class` type will always retain a reference.

> Presumably the [COM] interfaces don't/can't work on GPUs?

I think that is a fair assumption *for now*. I wouldn't rule anything out in the long run.

> The default for interface seems (?) to be it's a value type.

Correct, insofar as that is more or less the only behavior. An existential in Slang always behaves like the value of concrete type that it holds. If the underlying concrete type is a value type, the existential behaves like a value. If the underlying concrete type is a reference, the existential behaves like a reference.

Really, the simplest way to think of it is that *everything* is a value type, and "reference types" are just those types where the *value* is a reference/pointer to some storage.

> The class type seems like that would slang reference type. Having a modifier change the types behavior so dramatically seems 'surprising'.

It is important to note that the `[COM]` attribute does **not** cause the change in behavior. The difference you are observing is around whether the concrete type is a `struct` type or a `class` type (a distinction that people have accepted easily enough in C#).

> Something that made more clear the reference nature might be beneficial.

I don't think that can be made to work, because the same distinction would hold for *any* interface when using `class` types. In fact, the same basic thing is *also* true with generics:

```
struct Stuff<T>
{
    T t;
}
```

The programmer cannot tell at the time they declare `Stuff` whether `T` will be a value type (`struct`) or reference type (`class`). The important thing is that in 99% of cases it turns out they don't need to care.

docs/proposals/004-com-support.md
tangent-vector: Those are both good ideas. We obviously need support for manually specified GUIDs, and a system to auto-generate one from the scoped (and/or mangled) name sounds like a great default.

PR #2234, opened 2022-05-13
Special handling around return and liveness

jsmall-nvidia:
On top of #2233 

If a variable is live and it or some alias of it is returned, we don't need to emit a live end. 

The added analysis is somewhat basic in that it only looks in the block the last access is in, and if the return returns a value derived from the root. If so emitting the end can be avoided. 

Adds fix around only adding starts for the current root to block runs.


source/slang/slang-ir-liveness.cpp
```diff
@@ -256,6 +265,10 @@ struct LivenessContext
         /// Get the block from the index
     IRBlock* _getBlock(BlockIndex blockIndex) const { return m_functionBlockInfos[Index(blockIndex)].block; }
 
+        /// True if the terminator can be considered an access
+        /// This allows us to elide a scope end if the root is returned 
+    bool _isAccessTerminator(IRTerminatorInst* terminator);
+
```
tangent-vector, line 271: We should double-check the semantic rules for the relevant SPIR-V ops. The LLVM live-range instructions say that something like:

When a thread "executes" a live-range start, the contents of the relevant memory become undefined (so it needs to be emitted before any stores to initialize it), and the memory is considered live until a matching live-range-end instruction is executed **or** the function returns.

That second bit about function return is important, because it would imply that we never need to emit *any* live-range-end operations before a `return` (whether or not the live value is being returned).

source/slang/slang-ir-liveness.cpp
```diff
@@ -875,10 +945,30 @@ void LivenessContext::_findInstRunsForBlocks()
...
+        // If we don't add the terminator, everything else will behave correctly with regard
+        // adding live range end markers.
+        {
+            auto terminator = block->getTerminator();
+            if (_isAccessTerminator(terminator))
+            {
+                m_instRuns.add(terminator);
+                blockInfo->runCount++;
+            }
         }
```
tangent-vector, line 965: Thinking more, I'm not clear how this scenario would arise in a way that matters. Wouldn't this only come up if a function was trying to return a pointer into stack-allocated memory?

tangent-vector (approved):

source/slang/slang-ir-liveness.cpp
jsmall-nvidia: You can see it in the output. Without it you end up with a load into a temporary and end, and then a return of that temporary.

PR #2235, opened 2022-05-13
Fixed FXC diagnostics

apanteleev:
When the FXC downstream compiler is used with invalid arguments, such as unsupported profile, Slang would silently ignore that error and report a successful compilation. The code blob would be empty, which is confusing.


jsmall-nvidia (approved):
Does the failure you see emit anything useful in the raw diagnostics? I'm just wondering if there is a problem in the raw diagnostic parsing, such that it isn't picking up an error it should. 

Having the guard is good addition either way.

PR #2236, opened 2022-05-16
Refactor prelude emit

jsmall-nvidia:
Motivation is to make glsl prelude work correctly, as previously it was emitted before `#version` meaning little could work.

Idea here is to split output stages into

* front matter
* prelude
* preModule
* module

The front matter being everything needed before prelude. The preModule is typically the generated declarations to emulate built in types/functions, but I've used a more vague name as it can have other uses (it opens an anonymous namespace for example).


source/slang/slang-emit-c-like.h
```diff
@@ -398,17 +395,30 @@ class CLikeSourceEmitter: public SourceEmitterBase
...
+        // Note: we do not rely on the defaults for the target language,
+        // because a user could take the HLSL/GLSL generated by Slang and pass
+        // it to another compiler with non-default options specified on
+        // the command line, leading to all kinds of trouble.
+        //
+        // TODO: We need an approach to "global" layout directives that will work
+        // in the presence of multiple modules. If modules A and B were each
+        // compiled with different assumptions about how layout is performed,
+        // then types/variables defined in those modules should be emitted in
+        // a way that is consistent with that layout...
```
tangent-vector, line 414: I'm not sure the pre-existing comment is as useful here on the declaration, just because it is diving down into the guts of the particulars of what some of the front-matter *is*.

tangent-vector (dismissed):

Issue #2213, opened 2022-04-26
Support querying whether shader parameter is used post-optimization


tangent-vector:
Because of the way shader compilers have traditionally operated, many applications that rely on reflection of shader parameters need to be able to query whether or not a shader parameter is "used." Slang does not currently support such queries, but it should.

# Background

A few key choices in the design of the Slang compiler make this kind of query more subtle/challenging than it might appear.

Many other shader compilation systems assign registers/bindings to shader parameters only *after* performing optimizations, which can include the elimination of shader parameters that are unused by a particular entry point. Slang instead assigns registers/bindings *before* all non-trivial optimization steps, in order to ensure that we generate a deterministic result (will not change when lines inside function bodies get `#if`ed in/out). In practice, the Slang compiler assigns registers/bindings based on its AST structures (before IR code generation), and has no facilities at present for performing that task after optimization (where it would need to operate on the Slang IR).

The Slang API for reflection and runtime compilation is carefully designed to support scenarios where a program (or type) is loaded and compiled through the Slang front-end, and then reflection/layout information is queried *before* any particular variants of that code are linked, specialized, and emitted. This again means that the point in compilation where reflection currently takes place doesn't have access to the kinds of information that would be required to answer a query like "is this parameter used?".

# Insight

The key insight that unblocks a path forward is that we don't need to be able to use the Slang IR to answer the full query "is parameter `foo` used?". Instead, we can break that query into two distinct steps:

1. What `register`/`space` (or `binding`/`set`, etc.) was assigned to `foo`?
2. Is that `register`/`space` (etc.) used in some particular optimized kernel, or not?

The first step is something the current reflection API can answer just fine, and the second step is something that the IR is equipped to be able to answer (it just requires the plumbing work to surface it).

# API

The key thing to get right for now is the user-facing API for the feature. At the simplest, we could imagine an operation like:

```
SLANG_API bool spIsParameterLocationUsed(
    SlangCompileRequest* request,
    int entryPointIndex,
    int targetIndex,
    SlangParameterCategory category, // is this a `t` register? `s` register?
    int spaceIndex,      // `space` for D3D12, `set` for Vulkan
    int registerIndex);  // `register` for D3D12, `binding` for Vulkan
```

This would check if the kernel code that gets compiled for the given entry point and target ends up including (after optimization) a parameter that overlaps the given category, `space`, and `register`. A matching operation on `IComponentType` would probably also be needed:

```
[...]

PR #2239, opened 2022-05-16
Support for querying which parameters are used in emitted code

apanteleev:
This is for https://github.com/shader-slang/slang/issues/2213

A new post-emit pass is added, `collectMetadata`, that scans the IR and creates a set of ranges representing the binding indices of resources that were referenced in the IR. Then, a new function `spIsParameterLocationUsed` and its C++ version `ICompileRequest::isParameterLocationUsed` is added to query this information after code generation. 

A new test, `tests/reflection/used-parameters.slang`, is added for this feature. The `-no-codegen` flag was added to most of the reflection tests because the reflection test tool no longer specifies that implicitly, as codegen is needed to test the new used bindings feature.

The `spGetCompileFlags` function was added to have a non-hacky way of supporting this new feature in the reflection test.


source/slang/slang-ir-metadata.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // See if the new range is adjacent to any of the existing ranges, merge with that.
+    for (auto& range : ranges)
+    {
+        if (range.adjacentTo(newRange))
+        {
+            range.mergeWith(newRange);
+            return;
+        }
+    }
```
tangent-vector, line 33: Does this logic handle the case where two previously non-adjacent ranges become adjacent as a result of the range between them being added? That is, if we add `[0,A)` then `[B,C)` and finally `[A, B)` (with `A < B < C`), will the result have the single range `[0, C)`? It doesn't seem like it from this merge logic.

tests/cross-compile/cpp-resource-reflection.slang
```diff
@@ -1,4 +1,4 @@
-//TEST:CPU_REFLECTION: -profile cs_5_0 -entry computeMain -target cpp 
+//TEST:CPU_REFLECTION: -profile cs_5_0 -entry computeMain -target cpp -no-codegen
```
tangent-vector, line 1: I may have missed it in the PR description, but I wonder why some of these compiles need `-no-codegen` now where they were okay without it before.

tangent-vector (approved):
Looks really good.

slang.h
```diff
@@ -2127,6 +2131,15 @@ extern "C"
...
     SLANG_API unsigned spReflectionParameter_GetBindingSpace(SlangReflectionParameter* parameter);
 
+    SLANG_API SlangResult spIsParameterLocationUsed(
+        SlangCompileRequest* request,
+        SlangInt entryPointIndex,
+        SlangInt targetIndex,
+        SlangParameterCategory category, // is this a `t` register? `s` register?
+        SlangUInt spaceIndex,      // `space` for D3D12, `set` for Vulkan
+        SlangUInt registerIndex,   // `register` for D3D12, `binding` for Vulkan
+        bool& outUsed);
```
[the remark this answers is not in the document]
apanteleev: D3D also uses this signature type, but mostly when creating objects. Like, `HRESULT ID3D12Device::CreateCommandQueue(pDesc, riid, ppCommandQueue)` - that returns whether the operation was successful, and if not, why not. The actual result is returned through the out parameter. I think it's fine to use the same approach here.

slang.h
```diff
@@ -3872,6 +3890,14 @@ namespace slang
         virtual SLANG_NO_THROW SlangResult SLANG_MCALL getProgramWithEntryPoints(
             slang::IComponentType** outProgram) = 0;
 
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL isParameterLocationUsed(
+            SlangInt entryPointIndex,
+            SlangInt targetIndex,
+            SlangParameterCategory category,
+            SlangUInt spaceIndex,
+            SlangUInt registerIndex,
+            bool& outUsed) = 0;
```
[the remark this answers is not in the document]
apanteleev: How would it work though, before linking and optimization (most importantly DCE)?

source/slang/slang-ir-metadata.cpp
apanteleev: No, there will be two ranges. We can of course run a compaction pass on the ranges after all of them have been collected, but is that necessary?

tests/cross-compile/cpp-resource-reflection.slang
apanteleev: Previously, `-no-codegen` was added for all tests implicitly in the reflection test program. Now it's not, as I need to do code generation for the used bindings test. I guess I could drop this option on all/most of the other tests too, and update the expected outputs, which would result in better coverage.

PR #2252, opened 2022-05-26
Added NativeStringType

jsmall-nvidia:
The motivation here is to be able to distinguish between Slang::String and const char* for C++/C like targets.

* Added NativeString that can be used instead of String in Slang source. It will output as const char* 
* Fixed some small corner case issues around C++ string encoding/decoding
* Added support for C++11 U/u decoding
* Expanded UnownedTerminatedStringSlice
  * Made convertible to const char* 
* Made slang C++ output use UnownedTerminatedStringSlice. Doing so avoids memory allocation when not needed, and allows conversion to const char* (and so NativeString) type
* Made 'host' C++ targets available, I used for testing 

I do have a test, but need to make it work without comparing source output, which is huge.


source/slang/slang-check-conversion.cpp
```diff
@@ -639,6 +639,16 @@ namespace Slang
...
 
+        // If both are string types we assume they are convertable in both directions
+        if (as<StringTypeBase>(fromType) && as<StringTypeBase>(toType))
+        {
+            if (outToExpr)
+                *outToExpr = fromExpr;
+            if (outCost)
+                *outCost = kConversionCost_None;
+            return true;
+        }
```
tangent-vector, line 650: Have zero-cost conversion in both directions seems like it would make it impossible for the compiler to pick between overloads that take the two different string types:

```
void f(String s);
void f(NativeString s);

...
NativeString n = "...";
f(n); // which one does it call?
```

We probably need to define one of the directions as being a "widening" conversion (probably from `NativeString` to `String`) and the other as a "narrowing" conversion (probably from `String` to `NativeString`). That would allow both to be implicit but not zero-cost.

Another note here is that we *should* be able to implement the two conversion directions this implies by adding appropriately annotated `__init` declarations to the stdlib types, rather than making them be built-in compiler magic.

tangent-vector (approved):

PR #2253, opened 2022-05-27
New language feature: basic error handling.

csyonghe:
This PR adds language support for error handling.

Adds `throws E` syntax to mark a function that can throw an error code of type `E`:
```
float f() throws int { }
func f() throws int : float { }
```
A throwing function must be called with a `try` clause:
```
float a = try f();
```
A `try` clause is only allowed in a throwing function.

A throwing function will be lowered to a function that has `IRFuncType(returnType, paramTypes... ,IRFuncThrowTypeAttr(errorType))`.
A try call will be lowered into a `tryCall(callee, successBlock, failureBlock, errorVar, args...)` where `errorVar` is an `IRVar` of type `errrorType` used to receive error code. The failureBlock will contain an `IRThrow(load(errorVar))`, the successBlock can consume the `tryCall` inst itself as the result of the call.

A mandatory IR pass will execute immediately after IR lowering to turn throwing function types (`IRFuncType` with `IRFuncThrowTypeAttr`) into a normal function type that returns `IRResultType<valueType, errorType>`. The `tryCall` insts will be replaced with a normal `call` + `ifElse` that branches into the failure block upon failure. The rest of the code will consume `IRGetResultValue(call(...)) : resultValueType` as usual.

Another post-linking IR pass will rewrite `IRResultType` and their operations into normal struct types. `IRResultType<void, E>` will be converted to just `E`.

With this PR, the following example code will compile:
```
[COM]
interface IObject
{
    int getValue(int value) throws int;
}

[DllImport("test-com")]
IObject createObject();

public __extern_cpp void main() throws int
{
    //writeln("hello world");
    //MessageBoxA(nullptr, "hello world!", "example", 0);

    int rs = try object.getValue(2);
}
```
Compiles as:
```
struct ResultType_0
{
    int32_t value_0;
    int32_t error_0;
};

{
    virtual SLANG_NO_THROW ResultType_0 SLANG_MCALL getValue(int32_t param1) = 0;
};

Slang_FuncType<ComPtr<IObject_0>>  _S1_init()
{
    return nullptr;
}
Slang_FuncType<ComPtr<IObject_0>>  _S1 = _S1_init();

ComPtr<IObject_0> createObject_0()
{
    if(*(&_S1) == nullptr)
    {
        void * _S3 = (_slang_rt_load_dll((String("test-com"))));
        void * _S4 = (_slang_rt_load_dll_func((_S3), (String("createObject"))));
        *(&_S1) = (slang_bit_cast<Slang_FuncType<ComPtr<IObject_0>> >(_S4));
    }
    else
    {
    }
    ComPtr<IObject_0> _S5 = (*(&_S1))();
    return _S5;
}

int32_t main()
{
    ResultType_0 _S6 = object_0->getValue(int(2));
    if(_S6.error_0 != int(0))
    {
        return _S6.error_0;
    }
    return int(0);
}
```


source/slang/slang-check-expr.cpp
```diff
@@ -1529,6 +1538,51 @@ namespace Slang
...
+        
+        auto parentFunc = this->m_parentFunc;
+        // TODO: check if the try clause is catched.
+        // For now we assume all `try`s are not catched (because we don't have catch yet).
+        if (!parentFunc)
+        {
+            getSink()->diagnose(expr, Diagnostics::uncachedTryCallInNonThrowFunc);
+            return expr;
+        }
+        if (parentFunc->errorType->equals(m_astBuilder->getVoidType()))
```
tangent-vector, line 1559: Semantically, we probably shouldn't use the `void` type to represent the "doesn't throw" case, but for now this is probably fine.

In practice `void` is our "unit" type, and thus has a single representable value. It could thus be reasonable for a function to declare that it can `throw` an error of type `void` (in which case there is only one possible error value it can throw).

Eventually we will need/want to have a "bottom" type (probably called `Never`), and *that* would be the appropriate error type to use for a function that cannot `throw`.

source/slang/slang-check-expr.cpp
```diff
@@ -1529,6 +1538,51 @@ namespace Slang
...
+        {
+            getSink()->diagnose(expr, Diagnostics::uncachedTryCallInNonThrowFunc);
```
tangent-vector, line 1561: I think the diagnostic names here are using "cached" and the comments are using "catched" when the word should be "caught".

source/slang/slang-check-expr.cpp
```diff
@@ -1529,6 +1538,51 @@ namespace Slang
...
+            return expr;
+        }
+
+        if (!as<InvokeExpr>(expr->base))
+        {
+            getSink()->diagnose(expr, Diagnostics::tryClauseMustApplyToInvokeExpr);
+            return expr;
+        }
```
tangent-vector, line 1569: A future change might allow a `try` clause to serve as an "umbrella" such that any expressions under it are allowed to be calls to `throw`ing functions. There's no need to worry about that for now.

source/slang/slang-check-expr.cpp
```diff
@@ -1529,6 +1538,51 @@ namespace Slang
...
+        auto base = as<InvokeExpr>(expr->base);
+        if (auto callee = as<DeclRefExpr>(base->functionExpr))
+        {
+            if (auto funcCallee = as<FuncDecl>(callee->declRef.getDecl()))
+            {
+                if (funcCallee->errorType->equals(m_astBuilder->getVoidType()))
+                {
+                    getSink()->diagnose(expr, Diagnostics::tryInvokeCalleeShouldThrow, callee->declRef);
+                }
+                return expr;
```
tangent-vector, line 1579: There seems to be a missing check here to ensure that the `errorType` of the callee is compatible with the `errorType` of the caller (unless the error is caught in the current scope).

For now I suppose we'd want a type equality test here (both error types must match exactly), but eventually it would be nice to have a type conversion check (although we'd need to figure out where in the AST to stash the conversion operation(s)).

source/slang/slang-ir-inst-defs.h
```diff
@@ -271,6 +272,12 @@ INST(makeArray, makeArray, 0, 0)
 INST(makeStruct, makeStruct, 0, 0)
 INST(MakeTuple, makeTuple, 0, 0)
 INST(GetTupleElement, getTupleElement, 2, 0)
+INST(MakeResultValue, makeResultValue, 1, 0)
+INST(MakeResultValueVoid, makeResultValueVoid, 0, 0)
```
tangent-vector, line 276: We are very much at the point now where we should go ahead and consistently work it through the IR that *all* functions return a value, and it is just that IR functions that have a `Void` return type return the single global literal of that type, called `void`. Under that model, the `makeResultValueVoid` case is also not needed.

source/slang/slang-ir-inst-defs.h
```diff
@@ -432,6 +439,9 @@ INST(SwizzledStore, swizzledStore, 2, 0)
         INST(ifElse, ifElse, 4, 0)
     INST_RANGE(ConditionalBranch, conditionalBranch, ifElse)
 
+    INST(Throw, throw, 1, 0)
+    // tryCall <successBlock> <failBlock> <callee> <err> <args>...
+    INST(TryCall, tryCall, 4, 0)
```
tangent-vector, line 444: I don't follow why this instruction takes an `<err>` operand. Shouldn't the error/value output be expressed as a block parameter of `successBlock` or `failBlock`, respectively?

source/slang/slang-ir-lower-error-handling.cpp
```diff
@@ -0,0 +1,239 @@
...
+    void processTryCall(IRTryCall* tryCall)
+    {
+        // If we see:
+        // ```
+        //      value = tryCall(callee, errorVar, successBlock, failBlock, args) : returnType
+        //  successBlock:
+        //      ... (uses value) ...
+        //  failBlock:
+        //      errorVal = load(errorVar);
+        //      (uses errorVal)
```
tangent-vector, line 59: This can work for now, but it isn't really an ideal encoding because we end up needing a `var` instruction for the error, and the `value` ends up visible to all blocks dominated by the `tryCall` even if they are on the failure path.

I'd much rather have this code block look like:

```
        //  tryCall(successBlock, failBlock, callee args)
        //
        //  successBlock(value : returnType):
        //      ... (uses value) ...
        //
        //  failBlock(errorVal : errorType):
        //      ... (uses errorVal) ...
```

With that encoding, we never need any `var`s and the result value/error are only visible (per SSA rules) strictly along the control-flow path where they are well-defined.

tangent-vector (approved):

tangent-vector (approved):

PR #2255, opened 2022-05-28
Work around MacOS compilation issue with embed stlib

cek:
- The embed-stdlib-generator project is created with
  `kind = "StaticLib"` to allow the build to work, even though
  the project doesn't actually create a library.

- Unlike some other platforms, MacOs `ar` emits an error if no
  object files are listed to be added to an archive. This causes
  enable-stdlib-generator to fail on MacOS.

- Changing the project's kind to `"SharedLib"` works around the issue.
  Other values for `kind` do not seem to work around the issue.

- Add an optional flag to `generatorProject` to indicate that
  `kind = "SharedLib"` should be used, rather than `"StaticLib"`

- Add MacOS fix for `SharedLibraryUtils::getSharedLibraryFileName()`.

- Enable embed stdlib in `github_macos_build.sh`

### DIFF
--- a/github_macos_build.sh
+++ b/github_macos_build.sh
@@ -8,12 +8,7 @@ git describe --tags | sed -e "s/\(.*\)/\#define SLANG_TAG_VERSION \"\1\"/" > sla
 cat slang-tag-version.h 
 
 # Create the makefile
-#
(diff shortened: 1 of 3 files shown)


jsmall-nvidia (approved):
Looks good.

PR #2256, opened 2022-05-28
Add  Slang::String compatibility with std::ostream

cek:
Add `<<` operator to allow `Slang::String` to inerop with `std::ostream`


cek:
> Looks good.
> 
> Would be nice for description to say it's on top of #2255

Oops, sorry about that!

PR #2258, opened 2022-06-01
COM interfaces with host callable

jsmall-nvidia:
Original intention here was to add (the awkwardly named) HostHostCallable. First host means it's not kernel, and the second is just part of it's `HostCallable`. This PR adds support for this but the implementation uses HostCallable in the end. 

* Support type for COM calls
* Add a COM host callable example. 
* Added the minimum needed to make COM interfaces work in the `regular` prelude
* Tried to use ArtifactDesc/Info where appropriate to try and limit large switches/centralize code
* Supported exporting/public/extern_cpp in C++ emit
* Removed anonymous namespace for C++ output non entry points (to make accessible)

The HostHostCallable target in the end wasn't used in the end here for the example, although is still useful if slang-rt was desired and host-callable was with a regular downstream compiler. The problem here being that for slang-llvm it would need to make the slang-rt library available which isn't currently supported in that path.

Using the `kernel` style is a little surprising, but did mean that far less code needed to be compiled to make things work. It also allowed for a 'closer the the metal' implementation - using regular pointers, and native strings. If there are no entry points there also isn't any extra output around that.

Future: 

* Either migrate and/or add a unit test around the functionality
* Does not support writing to globals 
  * Using a 'regular' global doesn't work because in normal HLSL this isn't possible
  * Making static, also doesn't work because the that global is part of the KernelContext which isn't available.
  * Globals are grouped together into a GlobalContext so that doesn't work either


examples/cpu-com-example/shader.slang
```diff
@@ -0,0 +1,21 @@
...
+{
+    return in;
+}
+
+[COM]
+interface IDoThings
+{
+    int doThing(int a, int b);
+    int calcHash(NativeString in);
+}
```
tangent-vector, line 15: I really like seeing this construct Just Work in an example like this.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2526,6 +2547,45 @@ void CPPSourceEmitter::emitPreModuleImpl()
     }
 }
 
+/* virtual */void CPPSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
+{
+    // Specially handle export, as we don't want to emit it multiple times
```
tangent-vector, line 2552: We should probably have logic elsewhere to ensure that we don't redundantly add decorations.

source/slang/slang-ir-com-interface.cpp
```diff
@@ -41,19 +42,17 @@ struct ComInterfaceLoweringContext
...
-            return result;
-
+     
         IRBuilder builder(sharedBuilder);
         builder.setInsertInto(module->getModuleInst());
-        result = builder.getComPtrType(type);
+
+        IRType* result = (artifactStyle == ArtifactStyle::Kernel) ?
+            static_cast<IRType*>(builder.getPtrType(type)) :
+            static_cast<IRType*>(builder.getComPtrType(type));
```
tangent-vector, line 55: Help me make sure I understand this, because from what I'm seeing this looks worrying. It *seems* like this is saying taht for "kernel" code we are deciding that COM interface types map to ordinary "raw" pointers, while for other cases of they stay as reference-counted `ComPtr<T>`?

I'm not sure I understand the logic here. The semantics of a `[COM] interface` type should always be those of `ComPtr<T>`. Even in a CPU kernel, wouldn't we want reference-counting to ensure things don't get released prematurely?

I was honestly expecting to see a special carve-out just for uses of interface types in function parameter lists.

tangent-vector (approved):
I'm not sure I fully understand the code involving raw pointers, and I wouldn't be surprised if we need to revisit it eventually. For now, though, this seems fine.

source/slang/slang-ir-com-interface.cpp
csyonghe: Yeah, I feel like it needs to be an error if a COM object is referenced in kernel code.

source/slang/slang-ir-com-interface.h
```diff
@@ -10,6 +12,6 @@ class DiagnosticSink;
 /// Lower com interface types.
 /// A use of `IRInterfaceType` with `IRComInterfaceDecoration` will be translated into a `IRComPtr` type.
 /// A use of `IRThisType` with a COM interface will also be translated into a `IRComPtr` type.
-void lowerComInterfaces(IRModule* module, DiagnosticSink* sink);
+void lowerComInterfaces(IRModule* module, ArtifactStyle artifactStyle, DiagnosticSink* sink);
```
csyonghe, line 15: I am not sure if we want to emit ComPtr types differently on kernel code, but if we need that info, It is probably better to just pass in a `TargetReq` instead.

source/slang/slang-ir-com-interface.cpp
tangent-vector: Thanks for the context. I have no problem with this change in the short term, and I just wanted to make sure I understood the motivation.

In the long run I think the following are true:

1. We need a Slang `IFoo` for a `[COM] interface IFoo` to be 100% binary compatible/equivalent to a C++ `IFoo*`, in all contexts (function parameter/return, data structures, etc.). There is no wiggle room on that.
2. As a result of (1), we are going to need to start emitting the reference-count increment/decrement operations directly into Slang IR (which allows us to perform the relevant optimizations on them...).

Once we have (2), the surface area that we have to include in the prelude for reference-counting support is a lot smaller. In the worst case we are just talking about a `static inline` function that wraps the call to `AddRef` or `Release`, but maybe we don't even need that (and could emit them directly).

jsmall-nvidia: Sounds good. 

Although it may not be easy (see ARC) doing reference counting in IR I think will be a significant overall win. 
You can't implement a 'perfect' ComPtr in C++. It always leads to unnecessary counting/busy work, and requires extra user code to improve perf which can introduce bugs.

PR #2261, opened 2022-06-02
Add simple ray tracing test

arquelion:
Changes:
- Added a simple ray tracing test based heavily on the `ray-tracing-pipeline` example - This test uses a 2x2 texture with three differently colored triangles in three of the four pixels and nothing in the fourth and traces four total rays, one for each pixel.
- Added several new helper functions to `gfx-test-texture-util` for dumping images to file and removing padding from data read from `readTextureResource` (sufficiently small textures like the one in this test will read back with padding included, which messes with dumping images to check correctness)

### DIFF
--- a/tools/gfx-unit-test/ray-tracing-tests.cpp
+++ b/tools/gfx-unit-test/ray-tracing-tests.cpp
@@ -72,6 +72,9 @@ namespace gfx_test
 
         void init(IDevice* device, UnitTestContext* context)
         {
+            if (!device->hasFeature("ray-tracing"))
+                SLANG_IGNORE_TEST;
+
             this->device = device;
             this->context = context;
         }


csyonghe (dismissed):

tools/gfx-unit-test/ray-tracing-tests.cpp
```diff
@@ -0,0 +1,426 @@
...
+    void simpleRayTracingTestImpl(IDevice* device, UnitTestContext* context)
+    {
+        BaseRayTracingTest test;
+        test.init(device, context);
+        test.run();
+    }
+
+    SLANG_UNIT_TEST(simpleRayTracingD3D12)
+    {
+        runTestImpl(simpleRayTracingTestImpl, unitTestContext, Slang::RenderApiFlag::D3D12);
```
csyonghe, line 418: Need to check for ray-tracing support and IGNORE the test when the capability is not supported by the device.

csyonghe (approved):

PR #2264, opened 2022-06-05
Major language server features.

csyonghe:
This PR implements the following language server features:
- Hover information
- Diagnostics
- Member completion
- Go to definition
- Call signature information
- Semantic Highlighting (coloring types)

Fixing several front-end stability bugs (e.g. infinite parsing loops, nullptr dereferencing) during testing.

Details:

We currently request the client to send in full text with each document change. Semantic highlighting is also done in full text only. On every text edit, we create a new `WorkspaceVersion`, which contains a new `Linkage` for parsed ASTs. Any prexisting ASTs in older `WorkspaceVersion` will be discarded. Initially the new Linkage will not contain any ASTs, and they will be parsed on demand upon a concrete request (e.g. completion or hover information request). Doing a full reparse after every edit seems slow, but in practice this works OK and the response time is very good even for large shader codebases.

Our current AST definition drops some useful token source info that will be needed by the language server, such as the location of "(" and ")" tokens in an invoke expression. The current solution is to just add them to the fields of the AST nodes. But there are also certain AST nodes that are created during checking, e.g. `let` expression from `openExistential`, `DeclRefExpr` from overload resolution, etc that prevents us to find the original tokens from the AST. Again we implemented a short term solution to just store a reference to the original expr in checked AST nodes.

In the long term we should develop a separate representation of parsed syntax trees, or concrete syntax trees that are result of the first parsing pass. The semantics checking will then produce another AST that references back to the original tree. This will allow us to implement additional features such as code formatting.

The language server is currently also using the `DocAST` infrastructure to extract documentation. But this implementation currently just `ASSERT` fails in some places that leads to crashes. I have to remove some of the asserts. In the long run we should extend our `Token` representation to carry additional trivia info so we can access to any type of comments in the code (not just doxygen comments).

image
image


.github/workflows/linux.yml
```diff
@@ -26,6 +26,15 @@ jobs:
         CONFIGURATION=${{matrix.configuration}}
         ARCH=${{matrix.platform}}
         source ./github_build.sh
+    - uses: actions/upload-artifact@v3
```
jsmall-nvidia, line 29: Are you sure you want to do uploads on each CI build? I mean it's handy - but we have run github space a while back requiring deletions. Guess we can see if that happens again.

source/slang/slang-ast-print.h
```diff
@@ -118,7 +120,7 @@ class ASTPrinter
 
         /// Add just the parameters from a declaration.
         /// Will output the generic parameters (if it's a generic) in <> before the parameters ()
-    void addDeclParams(const DeclRef<Decl>& declRef);
+    void addDeclParams(const DeclRef<Decl>& declRef, List<Array<Index, 2>>* outParamRange = nullptr);
```
jsmall-nvidia, line 123: Might be slightly nicer to have a type like `IndexRange` that hold an index range as start/end. No biggie.

.github/workflows/linux.yml
csyonghe: Yeah, this allows me to release an extension without creating a Slang release first. The default policy is that it will stay for 90 days. Perhaps we can change that to something like 1 day and see if it works.

source/slang/slang-workspace-version.cpp
```diff
@@ -0,0 +1,358 @@
...
+            break;
+        }
+    }
+    return result;
+}
+
+void parseDiagnostics(Dictionary<String, DocumentDiagnostics>& diagnostics, String compilerOutput)
+{
+    List<UnownedStringSlice> lines;
+    StringUtil::split(compilerOutput.getUnownedSlice(), '\n', lines);
```
jsmall-nvidia, line 103: Might be better to use StringUtil::calcLines as takes into account line ending styles.

source/slang/slang-workspace-version.cpp
```diff
@@ -0,0 +1,358 @@
...
+    for (Index lineIndex = 0; lineIndex < lines.getCount(); lineIndex++)
+    {
+        auto line = lines[lineIndex];
+        Index colonIndex = line.indexOf(UnownedStringSlice("):"));
+        if (colonIndex == -1)
+            continue;
+        Index lparentIndex = line.indexOf('(');
+        if (lparentIndex > colonIndex)
+            continue;
+        String fileName = line.subString(0, lparentIndex);
```
jsmall-nvidia, line 113: UnownedStringSlice::head and tail make this a little more convenient.

source/slang/slang-workspace-version.cpp
```diff
@@ -0,0 +1,358 @@
...
+                &context);
+        }
+        searchPaths = _Move(context.paths);
+    }
+    slangGlobalSession = globalSession;
+}
+
+void Workspace::invalidate() { currentVersion = nullptr; }
+
+int parseInt(UnownedStringSlice text, Index& pos)
```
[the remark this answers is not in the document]
csyonghe: Yes, but I need to parse the int from hex.

source/slang/slang-workspace-version.cpp
```diff
@@ -0,0 +1,358 @@
...
+void* Workspace::getInterface(const Guid& uuid)
+{
+    if (uuid == ISlangUnknown::getTypeGuid() || uuid == ISlangFileSystem::getTypeGuid())
+    {
+        return static_cast<ISlangFileSystem*>(this);
+    }
+    return nullptr;
+}
+
+Int convertHexDigit(char c)
```
jsmall-nvidia, line 218: Has function in CharUtil::getHexDigitValue

source/slang/slang-workspace-version.cpp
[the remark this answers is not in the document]
csyonghe: Ah, nevermind, this is the place where I can just use StringUtil.

source/slang/slang-workspace-version.h
```diff
@@ -0,0 +1,135 @@
...
+        {
+            return uri == other.uri;
+        }
+        bool operator!=(const URI& other) const { return uri != other.uri; }
+
+        HashCode getHashCode() const { return uri.getHashCode(); }
+
+        bool isLocalFile() { return uri.startsWith("file://"); };
+        String getPath() const;
+        StringSlice getProtocol() const { return uri.subString(0, uri.indexOf("://")); }
```
jsmall-nvidia, line 27: Assumes has "://"

source/slang/slang-language-server-semantic-tokens.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+List<SemanticToken> getSemanticTokens(Linkage* linkage, Module* module, UnownedStringSlice fileName)
+{
+    auto manager = linkage->getSourceManager();
+
+    List<SemanticToken> result;
+    auto maybeInsertToken = [&](const SemanticToken& token)
+    {
+        if (token.line >= 0 && token.col >= 0 && token.length > 0 &&
+            token.type != SemanticTokenType::_NormalText)
```
jsmall-nvidia, line 439: I know it's elsewhere in our code but any identifier prefixed with _C where C is a capital letter is compiler reserved.

jsmall-nvidia (approved):
Looks good. Only pedantic comments.

source/slang/slang-workspace-version.cpp
csyonghe: I didn't see `UnownedStringSlice::head` being defined, can you explain?

source/slang/slang-language-server-semantic-tokens.cpp
csyonghe: Fixed.

csyonghe (commented):
Thanks for the review. Will commit the fixes in a follow-up PR.

source/slang/slang-ast-print.h
csyonghe: Fixed. Added `Range<T>` to core.

PR #2266, opened 2022-06-08
Add smoke test for language server.

csyonghe:
This PR adds a new `LANG_SERVER` test type to our testing infrastructure that can test the language server.
We can now create a `.slang` file with the following comment to test the language server:

```
//TEST:LANG_SERVER:
//COMPLETE:line,col           // request completion info at specified location
//SIGNATURE:line,col          // request signature info at specified location
//HOVER:line,col                 // request hover info at specified location
//DIAGNOSTIC                    // if used, this should be placed as the last request to
                                           // print all the diagnostics received from the server.
```

The responses of each request are concatenated as the output of the test, which is then compared to the expected output.
All filenames that appear in the output will be redacted so the expected output is agnostic to the running environment.

### DIFF
--- a/tests/language-server/smoke.slang.expected.txt
+++ b/tests/language-server/smoke.slang.expected.txt
@@ -1,0 +1,25 @@
+--------
+first: 6  ().;:,<>[]{}-*/%+=&|! 
+second: 6  ().;:,<>[]{}-*/%+=&|! 
+getSum: 2  ([ 
+--------
+range: 24,26 - 24,31
+content:
+```
+Pair.T Pair<Pair.T, Pair.U>.first
+```
+
+
+
+{REDACTED}.slang(20)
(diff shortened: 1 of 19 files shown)


tests/language-server/smoke.slang.expected.txt
```diff
@@ -0,0 +1,25 @@
+--------
+first: 6  ().;:,<>[]{}-*/%+=&|! 
+second: 6  ().;:,<>[]{}-*/%+=&|! 
+getSum: 2  ([ 
+--------
+range: 24,26 - 24,31
```
jsmall-nvidia, line 6: The range outputs are zero indexed, as opposed to the inputs?

jsmall-nvidia (approved):
Nice.

PR #2269, opened 2022-06-09
Language Server improvements.

csyonghe:
- Improve parser robustness around `attribute_syntax`.
- Exclude instance members in a static query.
- Coloring property accessors and constructor calls.
- Provide auto complete suggestions for HLSL semantics.
- Improved signature help cursor range check.
- Support configuring predefined preprocessor macros in extension settings.
- Improved performance by supporting incremental text change notifications.
- Implemented a serialized AST cache for language server (improves performance by 30%, but currently disabled)
- Add translation to / from UTF16 source locations.
- Include literal expr initial value in variable hover text.


source/slang/slang-workspace-version.h
```diff
@@ -86,7 +123,12 @@ namespace Slang
 
         Module* getOrLoadModule(String path);
     };
-    
+
+    struct OnwedPreprocessorMacroDefinition
```
jsmall-nvidia, line 127: Owned?

jsmall-nvidia (approved):

PR #2281, opened 2022-06-15
Extend swapchain resize test

lucy96chen:
Changes:
- Added five frames of draw calls before and after the `swapchain->resize` operation to ensure the swapchain is usable both before and after resizing

@csyonghe


csyonghe (approved):
Looks good to me.

PR #2282, opened 2022-06-15
Lower throwing COM interface method.

csyonghe:
This change implements lowering of COM interfaces that throws an error into methods that returns the error code through return value and returns the result through the last argument.

For example, the following code
```
[COM]
interface IObject
{
    int getValue(int value) throws int;
}

[DllImport("test-com")]
IObject createObject();

public __extern_cpp void main() throws int
{
    //writeln("hello world");
    //MessageBoxA(nullptr, "hello world!", "example", 0);

    int rs = try object.getValue(2);
}
```

Now compiles to

```
{
    virtual SLANG_NO_THROW int32_t SLANG_MCALL getValue(int32_t param1, int32_t * param2) = 0;
};

struct ResultType_0
{
    int32_t value_0;
    int32_t error_0;
};

int32_t main()
{
    int32_t _S6;
    int32_t _S7 = object_0->getValue(int(2), &_S6);
    int32_t _S8 = _S6;
    ResultType_0 _S9;
    if(_S7 < int(0))
    {
        ResultType_0 _S10 = { int(0), _S7 };
        _S9 = _S10;
    }
    else
    {
        ResultType_0 _S11 = { _S8, int(0) };
        _S9 = _S11;
    }
    if(_S9.error_0 != int(0))
    {
        return _S9.error_0;
    }
    return int(0);
}
```

This change contains a refactor that pulls the marshalling logic in `slang-ir-dll-import` pass to a standalone `slang-ir-marshal-native-call` module so they can be reused by both dll import pass and the lower-com-method pass.

Also included a peephole optimization pass to remove trivial inst pairs such as `getResultError(makeResultError(e))` -> `e`.

### DIFF
--- a/source/slang/slang-ir-inst-pass-base.h
+++ b/source/slang/slang-ir-inst-pass-base.h
@@ -1,0 +1,83 @@
+// slang-ir-inst-pass-base.h
+#pragma once
+
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+    struct IRModule;
+
+    class InstPassBase
+    {
+    protected:
+        IRModule* module;
+        SharedIRBuilder sharedBuilderStorage;
+
+        List<IRInst*> workList;
+        HashSet<IRInst*> workListSet;
+        void addToWorkList(IRInst* inst)
+        {
+            if (workListSet.Contains(inst))
+                return;
+
+            workList.add(inst);
+            workListSet.Add(inst);
+        }
+
+    public:
+        InstPassBase(IRModule* inModule)
+            : module(inModule)
+        {}
+
+        template <typename InstType, typename Func>
+        void processInstsOfType(IROp instOp, const Func& f)
+        {
+            workList.clear();
+            workListSet.Clear();
+
+            addToWorkList(module->getModuleInst());
+
+            while (workList.getCount() != 0)
+            {
+                IRInst* inst = workList.getLast();
+
+                workList.removeLast();
+                workListSet.Remove(inst);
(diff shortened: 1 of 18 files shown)


source/slang/slang-ir-inst-pass-base.h
```diff
@@ -0,0 +1,83 @@
...
+            workList.add(inst);
+            workListSet.Add(inst);
+        }
+
+    public:
+        InstPassBase(IRModule* inModule)
+            : module(inModule)
+        {}
+
+        template <typename InstType, typename Func>
```
jsmall-nvidia, line 33: I guess this could pull the casting 'isAImpl' function from the InstType, and so could work for non leaf types. Would also not need to pass in the inst Op or assume it matched. Could also just use static_cast

csyonghe: Good point.

source/slang/slang-ir-peephole.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    void processInst(IRInst* inst)
+    {
+        switch (inst->getOp())
+        {
+        case kIROp_GetResultError:
+            if (inst->getOperand(0)->getOp() == kIROp_MakeResultError)
+            {
+                inst->replaceUsesWith(inst->getOperand(0)->getOperand(0));
+                changed = true;
```
jsmall-nvidia, line 22: This case doesn't have removeAndDeallocate - is that intended?

csyonghe: Thanks for the catch. Will fix.

PR #2291, opened 2022-06-21
Fill out support for calls to non-decl values

tangent-vector:
The existing compiler logic has a few places (semantic checking plus AST-to-IR lowering) where it assumes that function calls (`InvokeExpr`) are only ever made to expressions that resolve to a specific `Decl` (`DeclRefExpr`). This assumption allows semantic checking and lowering code to inspect things like the parameter list of an actual declaration, rather than just the type signature of the callee, and that infrastructure is used to support various features (e.g., default argument values on parameters).

The AST and IR representations themselves have no matching requirement, and the places where the more general case of call expressions would need to be supported were relatively clear in the code. This change attempts to add suitable logic into each of those places.

Note that this change does *not* surface any valid way to form input code that would cause these new code paths to be executed, so it is entirely possible that there are bugs in the logic as written here. The primary goal of this change is simply to get a sketch of the correct code checked in so that we have something to build on once we have language features that will require this support.

### DIFF
--- a/source/slang/slang-ast-support-types.h
+++ b/source/slang/slang-ast-support-types.h
@@ -1474,6 +1474,15 @@ namespace Slang
         List<ExtensionDecl*> candidateExtensions;
     };
 
+        /// Represents the "direction" that a parameter is being passed (e.g., `in` or `out`
+    enum ParameterDirection
+    {
+        kParameterDirection_In,     ///< Copy in
+        kParameterDirection_Out,    ///< Copy out
+        kParameterDirection_InOut,  ///< Copy in, copy out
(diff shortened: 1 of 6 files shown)


csyonghe (approved):
Looks good. Thanks for implementing this so quickly!

saipraveenb25 (dismissed):
Looks good to me! 
(Although some windows tests aren't passing because of unreferenced variable?)

saipraveenb25 (approved):
I think it's fixed now

csyonghe:
Closing as this is subsumed by #2292

PR #2292, opened 2022-06-21
Added __jvp(<fn-name>) syntax intended to invoke the derivative version of functions.

saipraveenb25:
Note that the pass/logic to compute the derivative (or even the derivative type) is not there yet. __jvp doesn't actually modify the function yet, it just yields the inner function when lowered to IR.

(The AST/Checking part seems to work correctly, but the IR part hasn't been tested yet)


source/slang/slang-check-expr.cpp
```diff
@@ -1487,6 +1487,27 @@ namespace Slang
...
+        if(auto funcType = as<FuncType>(expr->baseFn->type))
+        {
+            // Resolve JVP type.
+            // expr->type = primalToJVPType(expr->type);
+            expr->type = expr->baseFn->type; // Temporary.. 
+        }
+        else
+        {
+            // Error on diagnosis.
+            // getSink()->diagnose(expr, Diagnostics::expectedFunction, expr->baseFn->name);
```
csyonghe, line 1504: Why is this commented out?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -2947,6 +2947,16 @@ struct ExprLoweringVisitorBase : ExprVisitor<Derived, LoweredValInfo>
         return info;
     }
 
+    LoweredValInfo visitJVPDerivativeOfExpr(JVPDerivativeOfExpr* expr)
+    {
+        // Dummy expr.. just emit the decl-ref of the inner function.
+        LoweredValInfo info = emitDeclRef(
```
csyonghe, line 2943: I assume you will modify this later to generate an `IRJVPDerivative` inst that has the result of `emitDeclRef` as its operand and return that inst here.

csyonghe (approved):

saipraveenb25:
Making a new pull request with fixes and more progress

PR #2293, opened 2022-06-21
Preserve specialization cache in IR for specialization pass.

csyonghe:
Our specialization pass maintains several dictionaries that caches the already specialized functions so that the next time we see a generic function being specialized with same args we can reuse a previously generated specialization instead of producing a duplicate.

The intention of this PR is to make these dictionaries in the generic specialization pass preserved in the IR module beyond the pass, so when the pass gets invoked again, the dictionaries can be recovered from the IR to prevent generating duplicated specializations.

This is necessary to get this pass ready to be invoked in an interactive style so we can handle more complicated specialization cases.


source/slang/slang-ir-specialize.cpp
```diff
@@ -623,6 +623,76 @@ struct SpecializationContext
...
+    {
+        for (auto child : dictInst->getChildren())
+        {
+            auto item = as<IRSpecializationDictionaryItem>(child);
+            if (!item) continue;
+            IRSimpleSpecializationKey key;
+            bool shouldSkip = false;
+            for (UInt i = 1; i < item->getOperandCount(); i++)
+            {
+                if (item->getOperand(i) == nullptr)
```
jsmall-nvidia, line 637: Why's the test for nullptr needed?

jsmall-nvidia (approved):

PR #2297, opened 2022-06-23
Added basic syntax to mark and request function derivatives, as well as the framework for passes to process them.

saipraveenb25:
Two IR decorators: one to mark a function for auto-diff (`IRJVPDerivativeMarker`) and one used by the auto-diff pass to hold a pointer to the generated function `IRJVPDerivativeReference`.
One IR instruction to request the derivative `IRJVPDerivativeOf` in a function call.
Two new passes: `processDerivativeMarkers()` to process the markers and generate derivative functions and `processDerivativeCalls()` to replace instances of `IRJVPDerivativeOf` with a reference to the corresponding generated function.


source/slang/slang-ir-diff-call.cpp
```diff
@@ -0,0 +1,90 @@
...
+        // Check for the 'JVPDerivativeReference' decorator on the
+        // base function.
+        if (auto jvpRefDecorator = derivOfInst->base.get()->findDecoration<IRJVPDerivativeReferenceDecoration>())
+        {
+            jvpFunc = jvpRefDecorator->getJVPFunc();
+        }
+        
+        // Substitute all uses of the 'derivativeOf' operation 
+        // with the resolved derivative function.
+        while (auto use = derivOfInst->firstUse)
```
csyonghe, line 68: derivOfInst->replaceAllUsesWith

source/slang/slang-ir-diff-call.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // derivative calls with the processed derivative 
+    // function.
+    //
+    IRModule*                       module;
+
+    bool processModule()
+    {
+        // Run through all the global-level instructions, 
+        // looking for callable blocks.
+        for (auto inst : module->getGlobalInsts())
```
csyonghe, line 22: I recommend doing a general recursive walk instead of being specific onglobal functions. This will allow you to look into the bodies of generic functions as well. You can use the code in slang-ir-generics-lowering-context.h

source/slang/slang-lower-to-ir.cpp
```diff
@@ -8417,6 +8473,16 @@ RefPtr<IRModule> generateIRForTranslationUnit(
 #endif
 
     validateIRModuleIfEnabled(compileRequest, module);
+    
+    // Process higher-order-function calls before any optimization passes
+    // to allow the optimizations to affect the generated funcitons.
+    // 1. Process JVP derivative functions.
+    processJVPDerivativeMarkers(module);
```
csyonghe, line 8480: A better way is to inspect all calls first, then generate a diff func for the callee if it hasn't been generated and replace the callee inst with the generated func. This way we don't do work for funcs that are not actually called, and we don't need to store a reference in the marker decoration.

csyonghe (approved):
We should stay consistent with the naming conventions of the codebase: Marker->Decoration, IRDerivativeOf->IRDerive(Inst names should use verb form).

Feel free to merge this in and fix in a follow up PR.

csyonghe:
baseFn->baseFunc or baseFunction

saipraveenb25:
Excellent feedback. I'll add the fixes in a follow-up PR.

PR #2303, opened 2022-06-24
Added basic auto-diff capabilities for local load/store and simple arithmetic. Also added type-checking during the semantic stage.

saipraveenb25:
Renamed `JVPDerivativeOf` to `JVPDifferentiate` to be more in line with the naming convention.
The logic to select which functions are differentiated is still based on `getGlobalInsts`. Have not been able to transition to recursive walk yet.


source/slang/slang-check-expr.cpp
```diff
@@ -1509,16 +1509,46 @@ namespace Slang
...
+            // The JVP return type is float if primal return type is float
+            // void otherwise.
+            //
+            if (primalType->resultType->equals(astBuilder->getFloatType()))
+                    jvpType->resultType = astBuilder->getFloatType();
+                else
+                    jvpType->resultType = astBuilder->getVoidType();
+            
+            // No support for differentiating function that throw errors, for now.
+            SLANG_ASSERT(primalType->errorType->equals(astBuilder->getBottomType()));
```
csyonghe, line 1537: Let's turn this into a proper diagnostic. I don't think we are going to handle this in anytime soon.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
 
 namespace Slang
 {
 
+struct JVPTranscriber
+{
+
+    // Stores the mapping of arbitrary 'R-value' instructions to instructions that represent
+    // their differential values.
+    Dictionary<IRInst*, IRInst*> instMapD;
```
csyonghe, line 16: Naming: we tend to name dictionaries that conveys its intention more clearly: e.g. mapInstToDifferentiatedValue would be proper here.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
+    }
+
+    IRType* differentiateType(IRBuilder* builder, IRType* typeP)
+    {
+        switch (typeP->getOp())
+        {
+            case kIROp_HalfType:
+            case kIROp_FloatType:
+            case kIROp_DoubleType:
+                return builder->getType(typeP->getOp());
```
csyonghe, line 67: Can just `return typeP`?

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
+        {
+            newParamListP.add(as<IRParam>(cloneInst(&cloneEnv, builder, paramP)));
+        }
+
+        // Now emit differentials.
+        List<IRParam*> newParamListD;
+        for (auto paramP : newParamListP)
+        {
+            IRParam* paramD = as<IRParam>(differentiateParam(builder, paramP));
+            mapDifferentialInst(paramP, paramD);
```
csyonghe, line 99: We can add a IRNameHintDecoration here with `IRBuilder::addNameHintDecoration` on the param if we want to see a more explanatory parameter name in the generated code.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
+        if (auto varP = as<IRVar>(loadP->getPtr()))
+        {   
+            // If the loaded parameter has a differential version, 
+            // emit a load instruction for the differential parameter.
+            // Otherwise, emit nothing since there's nothing to load.
+            // 
+            if (auto varD = as<IRVar>(getDifferentialInst(varP)))
+            {
+                IRLoad* loadD = as<IRLoad>(builder->emitLoad(varD));
+                SLANG_ASSERT(loadD);
```
csyonghe, line 177: I'd prefer to turn all `ASSERT`, `UNEXPECTED` etc. into diagnostics since that will inform the user what exactly is causing the compiler to fail.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
+            {
+                IRStore* storeD = as<IRStore>(
+                    builder->emitStore(storeLocationD, storeValD));
+                SLANG_ASSERT(storeD);
+                return storeD;
+            }
+            return nullptr;
+        }
+        
+        SLANG_UNEXPECTED("Attempting to differentiate an unsupported store instruction");
```
csyonghe, line 208: We realistically should be checking this in the front-end to make sure that doesn't happen here. But for now we can just issue a diagnostic for unsupported case in the IR pass. Let's not use SLANG_ASSERT or SLANG_UNEXPECTED in our IR passes if possible.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -3,10 +3,303 @@
...
+        case kIROp_Add:
+        case kIROp_Mul:
+        case kIROp_Sub:
+            return differentiateBinaryArith(builder, instP);
+
+        case kIROp_Construct:
+            return differentiateConstruct(builder, instP);
+
+        default:
+            SLANG_UNEXPECTED("Attempting to differentiate unrecognized instruction");
```
csyonghe, line 298: Let's use proper diagonostic here.

source/slang/slang-ir-insts.h
```diff
@@ -529,17 +529,17 @@ struct IRJVPDerivativeReferenceDecoration : IRDecoration
...
+struct IRJVPDifferentiate : IRInst
 {
     enum
     {
-        kOp = kIROp_JVPDerivativeOf
+        kOp = kIROp_JVPDifferentiate
     };
     // The base function for the call.
     IRUse base;
     IRInst* getBaseFn() { return getOperand(0); }
```
csyonghe, line 540: getBaseFunc

csyonghe (approved):
Looks good. I am really excited to see this!
Minor naming and diagnostic handling comments. We can fix them in a follow up PR.

PR #2304, opened 2022-06-25
Language server fixes and improvements

csyonghe:
- Add inlay hint feature that shows the parameter names for each call argument and auto deduced types for each `var` or `let` declaration.
- Fix type checking of JVPDiff operator to prevent crashing the language server
- Add signature help for higher order functions.

Result of inlay hints:
image

### DIFF
--- a/source/slang/slang-check-expr.cpp
+++ b/source/slang/slang-check-expr.cpp
@@ -1514,13 +1514,14 @@ namespace Slang
         // Check/Resolve inner function declaration.
         expr->baseFunction = CheckTerm(expr->baseFunction);
         
+        auto astBuilder = this->getASTBuilder();
+
         if(auto primalType = as<FuncType>(expr->baseFunction->type))
         {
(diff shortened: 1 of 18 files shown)


source/slang/slang-check-expr.cpp
```diff
@@ -1553,7 +1557,11 @@ namespace Slang
         else
         {
             // Error
-            UNREACHABLE_RETURN(nullptr);
+            expr->type = astBuilder->getErrorType();
```
csyonghe, line 1560: @saipraveenb25 We always need to assign a type to an expr after checking it to prevent crashing the language server. If an expr fails checking, we assign ErrorType to it.

saipraveenb25 (commented):
The differentiation changes look good to me, and I'll make sure to add diagnostics instead of asserts wherever possible.

jsmall-nvidia (dismissed):
Nice.

PR #2309, opened 2022-06-30
Added `[__custom_jvp(func)]` attribute, and modified the derivative pass to only process referenced functions.

saipraveenb25:
Added attribute `[__custom_jvp(func)]` to link a primal function declaration to it's corresponding hand-coded derivative.

Modified derivative pass to only produce derivatives of functions that are accessed in the code.

Added a test to check that using `[__custom_jvp(func)]` on a redeclaration correctly links the derivative function.

Moved derivative related tests to the new `tests/autodiff` folder


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -324,14 +392,81 @@ struct JVPDerivativeContext
...
+                    {
+                        IRFunc* jvpFunction = emitJVPFunction(builder, as<IRFunc>(baseFunction));
+                        builder->addJVPDerivativeReferenceDecoration(baseFunction, jvpFunction);
+                        workQueue->push(jvpFunction);
+                    }
+                    else
+                    {
+                        // TODO(Sai): This would probably be better with a more specific
+                        // error code.
+                        getSink()->diagnose(jvpDiffInst->sourceLoc,
```
csyonghe, line 452: If we implement the checking in the front end then we should not reach the situation here. Whenever possible we should be doing semantic checks instead of reporting errors in an IR pass.

csyonghe (approved):
Looks good.

PR #2312, opened 2022-06-30
Language server: auto completion of `import` file and directories.

csyonghe:


source/core/slang-string.cpp
```diff
@@ -142,6 +142,17 @@ namespace Slang
...
 
+    bool UnownedStringSlice::startsWithCaseInsensitive(UnownedStringSlice const& other) const
+    {
+        UInt thisSize = getLength();
+        UInt otherSize = other.getLength();
+
+        if (otherSize > thisSize)
+            return false;
+
+        return UnownedStringSlice(begin(), begin() + otherSize).caseInsensitiveEquals(other);
```
jsmall-nvidia, line 153: Could use head(otherSize).caseInsensitiveEquals(...);

jsmall-nvidia (dismissed):
Looks good.

PR #2320, opened 2022-07-11
Added support for differentiating out and inout parameters.

saipraveenb25:
Added type checking for out/inout parameters.
Added differentiation rules, and special case handling to avoid cloning 'side-effect' instructions.
Added two simple test cases. Will expand them to include more complex examples.


csyonghe:
You may want to do an `git submodule update` to get rid of the submodule changes.

PR #2323, opened 2022-07-12
Added support for differentiating out and inout parameters.

saipraveenb25:
Added type checking for out/inout parameters.
Added differentiation rules, and special case handling to avoid cloning 'side-effect' instructions.
Added two simple test cases. Will expand them to include more complex examples.


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -212,14 +244,14 @@ struct JVPTranscriber
...
-        if (auto destParam = as<IRVar>(storeLocation))
+        if (as<IRVar>(storeLocation) || as<IRParam>(storeLocation))
         {   
             // If the stored value has a differential version, 
             // emit a store instruction for the differential parameter.
             // Otherwise, emit nothing since there's nothing to load.
             // 
             IRInst* storeValD = getDifferentialInst(storeVal);
-            IRVar*  storeLocationD = as<IRVar>(getDifferentialInst(destParam));
+            IRInst* storeLocationD = getDifferentialInst(storeLocation);
```
csyonghe, line 254: Does `getDifferentialInst` always returns a pointer-type value? I mention this because there is a difference between an `IRVar` and an `IRParam`. An `IRVar` is always a pointer, but an `IRParam` is just an SSA temporary value, so you should never need to load an `IRParam`.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -352,16 +389,38 @@ struct JVPTranscriber
...
+    // in the differential function. We detect and avoid replicating 
+    // side-effect instructions.
+    // 
     bool requiresPrimalClone(IRBuilder*, IRInst* instP)
     {
         if (as<IRReturn>(instP))
             return false;
-        else
-            return true;
+        else if (auto paramP = as<IRParam>(instP))
```
csyonghe, line 399: For naming conventions, I don't know what does the `P` mean in `paramP`, or `instD` etc. in your code. We normally don't use single-letter abbrevations since they are confusing. I think just `param` here would be sufficient.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -352,16 +389,38 @@ struct JVPTranscriber
...
+        {
+            // Out-type parameters are discarded from the parameter list,
+            // since pure JVP functions to not write to primal outputs.
+            // 
+            if (as<IROutType>(paramP->getDataType()))
+                return false;
+        }
+        else if (auto storeP = as<IRStore>(instP))
```
csyonghe, line 407: `storeInst` would be a better variable name here.

source/slang/slang-check-expr.cpp
```diff
@@ -1509,18 +1509,43 @@ namespace Slang
         return expr;
     }
 
+    // This function proceses primal arguments that need to be carried over to the function
```
csyonghe, line 1512: What is the definition for "primal" arguments? Would be good to define it more clearly in the comment.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -374,6 +433,19 @@ struct JVPTranscriber
...
+        {
+            // We replace the operands of the old instruction with their clones,
+            // if available.
+            // 
+            for(UInt ii = 0; ii < oldInstP->getOperandCount(); ++ii)
+            {
+                auto oldOperand = oldInstP->getOperand(ii);
+                auto newOperand = findCloneForOperand(&cloneEnv, oldOperand);
+
+                instP->getOperands()[ii].init(instP, newOperand);
```
csyonghe, line 446: `instP->setOperand(ii, newOperand)`

csyonghe (approved):

source/slang/slang-ir-diff-jvp.cpp
saipraveenb25: I'm using 'P' to represent the _primal_ (or 'original') instructions that are computing `f(x)`. 'D' postfix is for instructions that are computing the corresponding differential (i.e. for df(x, dx)) for that instruction.
Yes, this could be confusing. I could switch to full postfixes: `instPrimal` and `instDiff`/`instDifferential` but it might cause the code to look bloated. Alternatively `inst` and `diffInst` maybe?

csyonghe: Yeah that looks good, or `originalInst` and `diffInst` to be more explicit.

source/slang/slang-check-expr.cpp
```diff
@@ -1509,18 +1509,43 @@ namespace Slang
...
+    // signature for the JVP function. (eg. out types can be discarded)
+    //
+    Type* primalToInputType(ASTBuilder*, Type* primalType)
+    {
+        if (auto primalOutType = as<OutType>(primalType))
+        {
+            // TODO(sai): are we sure it's not possible to run into OutType<InOutType<T>>?
```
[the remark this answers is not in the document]
saipraveenb25: Removed comment.

source/slang/slang-ir-diff-jvp.cpp
csyonghe: Yeah, you shouldn't load a `IRParam`.
For example, if I have func(int p);
Then `p` is just `p = IRParam : int`.
If I have `int i;` as an variable, then the IR is
`i = IRVar:Ptr<int>`, and you need to load it to get an `int`.

source/slang/slang-check-expr.cpp
saipraveenb25: Adjusted the comment

csyonghe (approved):

PR #2331, opened 2022-07-16
Added forward-mode autodiff support for more instructions

saipraveenb25:
Added name-hints for differential variables
Added support for swizzles and some vector constructors
Added support for multi-block functions (with only unconditional branches)
Moved derivative passes to after forced early inline pass to support operators like '+='


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -366,6 +390,77 @@ struct JVPTranscriber
         return nullptr;
     }
 
+    IRInst* differentiateSwizzle(IRBuilder* builder, IRSwizzle* swizzleP)
+    {
+        if (auto baseD = getDifferentialInst(swizzleP->getBase(), nullptr))
+        {
+            List<IRInst*> swizzleIndices;
+
+            // TODO(sai): Is this copy really needed?
```
csyonghe, line 399: You can just reuse the array in the original inst if it is not changed.
The indices you passed to IRBuilder::emitSwizzle() will be copied into the generated inst anyways.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -392,7 +487,7 @@ struct JVPTranscriber
     // in the differential function. We detect and avoid replicating 
     // side-effect instructions.
     // 
-    bool requiresPrimalClone(IRBuilder*, IRInst* instP)
+    bool hasNoSideEffects(IRBuilder*, IRInst* instP)
```
csyonghe, line 490: there is already a `hasSideEffect` function defined in slang-ir.h

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -418,45 +513,47 @@ struct JVPTranscriber
             // 
             if(!lookUp(&cloneEnv, storeLocation))
                 return false;
+        } else if (as<IRTerminatorInst>(instP))
```
csyonghe, line 516: `else` in its own line.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -474,20 +571,54 @@ struct JVPTranscriber
...
+        case kIROp_swizzle:
+            return differentiateSwizzle(builder, as<IRSwizzle>(instP));
+        
+        case kIROp_constructVectorFromScalar:
+            return differentiateByPassthrough(builder, instP);
+        
+        case kIROp_Specialize:
+            // The implementation is still incomplete, but we sometimes encounter 
+            // *unused* specialize instructions. So, we're only going to handle that case.
+            if (!oldInstP->hasUses())
```
csyonghe, line 592: Our DCE pass should be able to clear out these insts. I wonder if it is because specialize isn't marked as side-effect-free.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -8555,7 +8546,16 @@ RefPtr<IRModule> generateIRForTranslationUnit(
     // marked for mandatory "early" inlining.
     //
     performMandatoryEarlyInlining(module);
-
+
```
csyonghe, line 8549: We should run `simplifyIR()` before running the jpv passes to clean up deadcode.

csyonghe: We can also move this pass after the `constructSSA` pass.

source/slang/slang-ir-diff-jvp.cpp
saipraveenb25: Hmm.. I don't think it has the same definition of side-effect that I'm using though. For one, stores into local variables are ok, and certain calls too. Maybe I should rename this to `isPurelyFunctional()` ?

csyonghe: I see, that seems a better name. Normally you shouldn't be seeing local variable stores after SSA pass.

csyonghe (approved):

PR #2336, opened 2022-07-21
Allow dynamic dispatch to handle nested interface-typed fields.

csyonghe:
This change includes fixes to allow dynamic dispatch code gen to handle types that contain interface-typed fields.
A new test case has been added to cover the scenario.


source/slang/slang-ir-any-value-marshalling.cpp
```diff
@@ -719,10 +733,34 @@ namespace Slang
...
+        {
+            return alignUp(offset, 4) + 8;
+        }
         case kIROp_InterfaceType:
         {
-            // TODO: implement anyValue packing for interface types.
-            return -1;
+            auto interfaceType = cast<IRInterfaceType>(type);
+            auto size = SharedGenericsLoweringContext::getInterfaceAnyValueSize(interfaceType, interfaceType->sourceLoc);
+            size += 16;
```
jsmall-nvidia, line 762: Why 16? The 4's I'm presuming because it's an int/uint. Might we want to make these magic numbers an enum or constant value, such that can have a name saying what it is, and allowing for easier future changes.

source/slang/slang-ir-layout.cpp
```diff
@@ -194,7 +196,52 @@ static Result _calcNaturalSizeAndAlignment(
...
+            }
+            *outSizeAndAlignment = resultLayout;
+            return SLANG_OK;
+        }
+        break;
+    case kIROp_WitnessTableType:
+    case kIROp_WitnessTableIDType:
+    case kIROp_RTTIHandleType:
+        {
+            outSizeAndAlignment->size = 8;
```
jsmall-nvidia, line 227: 8 is another magic number.

jsmall-nvidia (approved):
Looks good.

PR #2338, opened 2022-07-21
Allow `class` to implement COM interface, [DLLExport]

csyonghe:
This PR enables a `class` to implement a `[COM]` interface, and also implements `[DLLExport]` to causes the marked function to be exported to DLL.


prelude/slang-cpp-prelude.h
```diff
@@ -272,6 +274,13 @@ struct ISlangUnknown
     virtual SLANG_NO_THROW uint32_t SLANG_MCALL release() = 0;
 };
 
+#define SLANG_COM_INTERFACE(a, b, c, d0, d1, d2, d3, d4, d5, d6, d7) \
```
jsmall-nvidia, line 277: Do we need to guard, with #ifndef? Might be problem if prelude is included as well as "slang.h", or give it a prelude exclusive name like SLANG_PRELUDE_COM_INTERFACE

source/slang/slang-emit-c-like.cpp
```diff
@@ -2790,17 +2838,59 @@ void CLikeSourceEmitter::emitClass(IRClassType* classType)
...
     m_writer->emit("public:\n");
     m_writer->indent();
 
+    if (comWitnessTables.getCount())
+    {
+        m_writer->emit("SLANG_COM_OBJECT_IUNKNOWN_ALL\n");
+        m_writer->emit("void* getInterface(const Guid & uuid)\n{\n");
+        m_writer->indent();
+        m_writer->emit("if (uuid == ISlangUnknown::getTypeGuid()) return static_cast<ISlangUnknown*>(this);\n");
+        for (auto wt : comWitnessTables)
```
jsmall-nvidia, line 2877: Does this include derived types interfaces?

source/slang/slang-emit-cpp.cpp
```diff
@@ -1779,6 +1794,23 @@ void CPPSourceEmitter::emitComInterface(IRInterfaceType* interfaceType)
     // Emit methods.
     m_writer->emit("\n{\n");
     m_writer->indent();
+    // Emit GUID.
+    m_writer->emit("SLANG_COM_INTERFACE(0x");
```
jsmall-nvidia, line 1798: Small point: I wonder if storing the guid as the native type rather than the string and then having functions that output/parse appropriately might slightly better. No ambiguity around validity, can compare quickly (no case issues etc), no additional allocation. Having functions to traffic between guids and strings is useful elsewhere.

source/slang/slang-emit-cpp.cpp
```diff
@@ -2476,6 +2508,33 @@ bool CPPSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOut
             }
             return true;
         }
+        case kIROp_MakeExistential:
+        case kIROp_MakeExistentialWithRTTI:
+        {
+            auto rsType = cast<IRComPtrType>(inst->getDataType());
```
jsmall-nvidia, line 2514: There is the irriting issue around COM interfaces where if I have 

```
interface IA : IUnknown {}
interface IB : IUnknown {}
class C : IA, IB {}

C c;

// Fails, because it's not clear which IUnknown is wanted, IAs, or IBs. 
IUnknown* ptr = &c;

// So in practice you have to static cast to a specific derived interface that implements your interface. For example
IUnknown* ptr = static_cast<IA*>(&c);

```

source/slang/slang-emit-cpp.cpp
```diff
@@ -2476,6 +2508,33 @@ bool CPPSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOut
...
+            m_writer->emit("ComPtr<");
+            m_writer->emit(getName(rsType->getOperand(0)));
```
jsmall-nvidia, line 2516: Don't know if it's covered in this PR, but the other irritating issue is wiring between implementations, when multiply derived. It's not enough to derive with a new interface if there are shared functions in the super interface shared by both, they need to be 'wired' to call the existing implementation.

tools/slang-unit-test/unit-test-com-host-callable.slang
```diff
@@ -19,14 +19,14 @@ public __extern_cpp int getGlobal()
     return intGlobal;
 }
 
-[COM]
+[COM("9E0FCAF0-DE40-4CF1-A6A4-FF24814D32F2")]
```
jsmall-nvidia, line 22: Having this have the option to work work without requiring an explicit guid could be handy. Perhaps the guid is internally generated as a hash on the name.

jsmall-nvidia (approved):

source/slang/slang-emit-c-like.cpp
csyonghe: This PR does not handle derived interfaces specifically. The intention is to leave that into a separate front-end pass that flattens the inheritance list by transitively including all derived types, so any future passes will not have to worry about this.

source/slang/slang-emit-cpp.cpp
csyonghe: This will be one of the issues when we get to deal with transitvely inherited interfaces.

PR #2340, opened 2022-07-26
Split render-cpu into smaller files

lucy96chen:
Changes:
- `render-cpu.cpp` has been split into smaller files each containing a single class and its related structs and classes, in the same vein as what was done with D3D, Vulkan, and CUDA before


tools/gfx/cpu/cpu-helper-functions.cpp
```diff
@@ -0,0 +1,18 @@
+// cpu-helper-functions.cpp
```
csyonghe, line 1: This file can just be in cpu-device.cpp

tools/gfx/cpu/cpu-base.h
```diff
@@ -0,0 +1,42 @@
+// cpu-base.h
+// Shared header file for CPU implementation
+#pragma once
+
+#include <chrono>
```
csyonghe, line 5: We probably don't need <chrono> included for all files here.

csyonghe (approved):

tools/gfx/cpu/cpu-helper-functions.cpp
lucy96chen: I'll merge this now and include these changes with the debug-layer split.

PR #2343, opened 2022-07-29
Added a new differential type system and various improvements

saipraveenb25:
Added `IDifferentiable` interface and logic to check that types conform to this interface
Added custom IR to make and access differential pairs along with a struct definition that the user can use to pass inputs.
Changed some variable naming
Moved derivative related definitions into `diff.meta.slang` and set up appropriate commands to include it into the stdlib.
Rewrote tests to conform to new typing system.

There's still quite some cleanup to do and some hacky code to rewrite, but the tests now pass again, and I wanted to make a PR before the changes get too large.


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -9,373 +9,741 @@
 namespace Slang
 {
 
+template<typename P, typename D>
+struct Pair
```
csyonghe, line 13: Let's not use generic names like `Pair` here. Consider names like `DiffInstPair` to be more specific.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -9,373 +9,741 @@
...
+
+    private:
+
+    IRInst* findDifferentiableInterface()
+    {
+        if (auto module = as<IRModuleInst>(inst))
+        {
+            for (auto globalInst : module->getGlobalInsts())
+            {
+                // TODO: This seems like a particularly dangerous way to look for an interface.
```
csyonghe, line 137: We should turn `IDifferentiable` as another `__intrinsic_type` so we can map it directly  into an IR type.
We should also add something like `__intrinsic_op` on the `associatedtype` declaration in `IDifferentiable` interface to turn all `lookupWitnessTable` insts on the associatedtype into an `IRGetDifferentiableType` inst during lower-to-ir.

csyonghe (approved):

PR #2350, opened 2022-08-05
Language server pointer type support + add `DLLImport` test

csyonghe:
Also with Natvis improvements for AST.

Make executable tests able to load and use `slang-rt`, which is needed to test `DLLImport` feature.


source/core/slang-platform.cpp
```diff
@@ -170,10 +177,11 @@ SLANG_COMPILE_TIME_ASSERT(E_OUTOFMEMORY == SLANG_E_OUT_OF_MEMORY);
 /* static */SlangResult SharedLibrary::loadWithPlatformPath(char const* platformFileName, Handle& handleOut)
 {
     handleOut = nullptr;
-
-	void* h = dlopen(platformFileName, RTLD_NOW | RTLD_GLOBAL);
-	if(!h)
-	{
+    if (strlen(platformFileName) == 0)
```
jsmall-nvidia, line 180: What does passing null or "" now mean?

source/slang/slang.natvis
```diff
@@ -15,12 +15,15 @@
     <DisplayString Condition="decl != 0">{(*(*(Slang::DeclRefBase*)this).decl).nameAndLoc}</DisplayString>
     <Expand>
       <ExpandedItem>decl ? ($T1*)(decl) : ($T1*)0</ExpandedItem>
-      <Item Name="[Substitutions]:">"========================="</Item>
-      <LinkedListItems>
-        <HeadPointer>substitutions.substitutions.pointer</HeadPointer>
-        <NextPointer>outer.pointer</NextPointer>
-        <ValueNode>this</ValueNode>
-      </LinkedListItems>
+      <Synthetic Name="[Substitutions]">
```
jsmall-nvidia, line 18: Nice to have better natvis.

Guess I worry about amount of 'stuff' needed to be kept in sync. This is to work around the AST derived types not having a virtual function, such that vs knows the type?

jsmall-nvidia (approved):

source/core/slang-platform.cpp
csyonghe: It means use RTLD_DEFAULT or load the function from the current process/module.

source/slang/slang.natvis
csyonghe: Right. It will also save one expansion click since it will just directly display the derived type.

PR #2355, opened 2022-08-10
`is` and `as` operator and `Optional<T>`.

csyonghe:
This changes implements `is` `as` operator for up- and down-casting within a type hierarchy, and an `Optional<T>` type that is used as the result of `as`.

These operators are supported in both specialization and dynamic dispatch paths.

Currently, the value in an is or as expression must be of an interface type.


source/slang/slang-parser.cpp
```diff
@@ -4790,6 +4790,10 @@ namespace Slang
         case TokenType::OpMod:
             return Precedence::Multiplicative;
         default:
+            if (token.getContent() == "is" || token.getContent() == "as")
```
jsmall-nvidia, line 4793: Seems a little unfortunate - but I guess the idea is not wanting to make 'is' or 'as' TokenType?

jsmall-nvidia (approved):
Looks good.

source/slang/slang-parser.cpp
csyonghe: Yeah, we probably want to allow using these as identifier names.

Issue #2360, opened 2022-08-15
fails to compile in release mode on GCC 12 (error: array subscript is outside array bounds)


1ace:
The release build has been broken for a while now; I *think* this is since GCC was updated to version 12 back in May, adding new checks as every new version does.
This is preventing the archlinux `shader-slang` package from being updated.
```
In file included from ../../../source/core/slang-basic.h:6,
                 from ../../../source/core/slang-token-reader.h:4,
                 from ../../../source/core/slang-token-reader.cpp:1:
In member function ‘char Slang::String::operator[](Slang::Index) const’,
    inlined from ‘Slang::List<Slang::Misc::Token> Slang::Misc::TokenizeText(const Slang::String&, const Slang::String&)’ at ../../../source/core/slang-token-reader.cpp:371:62:
../../../source/core/slang-string.h:591:30: error: array subscript [-9223372036854775808, -1] is outside array bounds of ‘const char [1]’ [-Werror=array-bounds]
  591 |             return begin()[id];
      |                              ^
```

the function where this line is:
```c++
        char operator[](Index id) const
        {
            SLANG_ASSERT(id >= 0 && id < getLength());
            return begin()[id];
        }
```

Clearly this assert needs to be converted into something that still exists in release builds.

Taking inspiration from the project I work for, you could use `__builtin_assume(cond)` (with fallback on `if (!cond) __builtin_unreachable()` for greater compiler coverage) instead of an assert that only exists in debug builds:
https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/util/macros.h#L118-136

As a side note, from experience `-Werror` is something you want in your CI so that **you** can catch issues, but should not be shipped to customers so that they're not prevented from building your product just because they have a more up-to-date system than you :)

1ace:
Actually, I read up a bit on premake (which I've never used) and I discovered that you are adding `flags { "FatalWarnings" }` there, which is causing all these problems.

I'm now adding `sed /FatalWarnings/d -i premake5.lua` to the packaging script to drop this which unblocks me, but I still suggest you drop this yourself and replace it with `CXXFLAGS=-Werror` in your CI config.

(And this is obviously orthogonal to upgrading some of your debug-only asserts into release-build compilers hints which you might want to do regardless.)

PR #2362, opened 2022-08-15
Move -Werror to CI

1ace:
As explained in #2360, you want `-Werror` in your CI to block yourself from introducing warnings in your current build setup, but you don't want to ship that as you don't want to prevent every customer on a more up-to-date system than you from being able to build your product.

---
Note: as you can see, this is catching more bugs than your old method; I am not interested in fixing those (especially since it looks like some might be in premake which I don't want to learn), so take this as a first draft, which you can build upon by adding the other fixes needed, and merge it once you're done :slightly_smiling_face:


csyonghe:
Thanks! We'll merge this PR once we landed the other ongoing ones. PR #2364 should have fixed the additional warnings.

PR #2364, opened 2022-08-16
Add gfx interface definition in Slang.

csyonghe:
- add gfx interface definitons in Slang.
- fix slang compiler to correctly type-check `out` interface argument.
- modify gfx interface to be fully COM compatible
- add convenient ShaderProgram creation methods to gfx.


slang-com-ptr.h
```diff
@@ -111,7 +111,9 @@ class ComPtr
 protected:
 	/// Gets the address of the dumb pointer.
     // Disabled: use writeRef and readRef to get a reference based on usage.
```
jsmall-nvidia, line 113: This was disabled because doing & could then overwrite the contained pointer with a new value without altering the ref. Is that not possible with this code generation path?

csyonghe: You are right. In this case I really need a `ComPtr<T>*` from the & operator so it can be passed to an `out` parameter. Perhaps I can just do `(ComPtr<T>*)(ptr.writeRef())`

tools/gfx/slang.slang
```diff
@@ -0,0 +1,451 @@
...
+};
+
+enum SlangOptimizationLevel
+{
+    SLANG_OPTIMIZATION_LEVEL_NONE = 0, /**< Don't optimize at all. */
+    SLANG_OPTIMIZATION_LEVEL_DEFAULT,  /**< Default optimization level: balance code quality and compilation time. */
+    SLANG_OPTIMIZATION_LEVEL_HIGH,     /**< Optimize aggressively. */
+    SLANG_OPTIMIZATION_LEVEL_MAXIMAL,  /**< Include optimizations that may take a very long time, or may involve severe space-vs-speed tradeoffs */
+};
+    enum SlangTypeKind
```
jsmall-nvidia, line 237: Odd tabbing.

slang-com-ptr.h
csyonghe: Actually, the code generation does the following:
```
void createFoo(out IFoo f) { f = ...; }
void main()
{
     IFoo f;
     createFoo(f);
}
```

Lowers to 
```
void createFoo(ComPtr<IFoo> *f)
{
   *f = ...;
}
void main()
{
     ComPtr<IFoo> f;
     createFoo(&f);
}
```

Which should be fine regarding to write refs. Because without that `=delete` line, operator & returns a `ComPtr<T>*` instead of `T**` so it should be fool proof against mishandling ref counts.

tools/gfx/slang.slang
```diff
@@ -0,0 +1,451 @@
...
+typedef int64_t Int;
+typedef uint64_t UInt;
+
+/*!
+@brief Severity of a diagnostic generated by the compiler.
+Values come from the enum below, with higher values representing more severe
+conditions, and all values >= SLANG_SEVERITY_ERROR indicating compilation
+failure.
+*/
+enum SlangSeverity
```
jsmall-nvidia, line 15: As long as kept in sync with slang.h, we could use names that are move `enum class`- like. 

```
enum SlangSeverity
{
    Disables = 0, 
    ///...
};
```

tools/gfx/vulkan/vk-command-encoder.h
```diff
@@ -54,6 +54,16 @@ class ResourceCommandEncoder
     , public PipelineCommandEncoder
 {
 public:
+    virtual void* getInterface(SlangUUID const&) { return this; }
```
jsmall-nvidia, line 57: Shouldn't this be checking the uuid? It's going to return as a valid interface whatever guid is passed in(!)

tools/gfx/vulkan/vk-command-encoder.h
```diff
@@ -54,6 +54,16 @@ class ResourceCommandEncoder
...
+    virtual SLANG_NO_THROW SlangResult SLANG_MCALL
+        queryInterface(SlangUUID const& uuid, void** outObject)
+    {
```
jsmall-nvidia, line 60: I guess along with that this should really be
```
if (auto ptr  = getInterface(uuid))
{   
   *outObject = ptr;
   return SLANG_OK
}
return SLANG_E_NO_INTERFACE;
```

tools/gfx/vulkan/vk-command-encoder.h
csyonghe: These exists mainly for compatibility of CPU Slang and isn't really intended to be called by the user, so I took the simplified path here.

tools/gfx/command-encoder-com-forward.h
```diff
@@ -1,124 +1,142 @@
...
-    virtual SLANG_NO_THROW void SLANG_MCALL beginDebugEvent(const char* name, float rgbColor[3])        \
-        override                                                                                        \
-    {                                                                                                   \
-        ResourceCommandEncoderBase::beginDebugEvent(name, rgbColor);                                    \
-    }                                                                                                   \
-    virtual SLANG_NO_THROW void SLANG_MCALL endDebugEvent() override                                    \
-    {                                                                                                   \
-        ResourceCommandEncoderBase::endDebugEvent();                                                    \
+#define SLANG_GFX_FORWARD_RESOURCE_COMMAND_ENCODER_IMPL(ResourceCommandEncoderBase)                \
+    virtual SLANG_NO_THROW SlangResult SLANG_MCALL queryInterface(                                 \
```
jsmall-nvidia, line 4: We might want to consider some x macro magic to generate this kind of code. That code can simplify the definition to a x macro that has the return type, the name and the param types and generates all the other stuff from that.

jsmall-nvidia (changes requested):
The getInterface/queryInterface seems problematic.

tools/gfx/vulkan/vk-command-encoder.h
jsmall-nvidia: Ok. I think would be better to do something more sensible, but can approve.

jsmall-nvidia (approved):

tools/gfx/command-encoder-com-forward.h
csyonghe: Yeah that would be a nice clean up.

PR #2365, opened 2022-08-16
Compute canonical path for modules loaded from source.

csyonghe:
Computing canonical path in `loadModuleFromSource` so that the module can be deduplicated from future imports.


source/slang/slang.cpp
```diff
@@ -1032,9 +1032,17 @@ SLANG_NO_THROW slang::IModule* SLANG_MCALL Linkage::loadModuleFromSource(
         {
             return loadedModule;
         }
+        String pathStr = path;
+        auto pathInfo = PathInfo::makeFromString(pathStr);
+        if (File::exists(pathStr))
+        {
+            String cannonicalPath;
+            Path::getCanonical(pathStr, cannonicalPath);
```
jsmall-nvidia, line 1040: Path::getCanonical, can return an error. If that happens this could fall back to the old behavior?

jsmall-nvidia (approved):

jsmall-nvidia (approved):

PR #2370, opened 2022-08-19
Call `gfx` in slang program.

csyonghe:
This PR enables a smoking test that attempts to create an `gfx` device in a CPU slang program.

Several fixes to emit logic is necessary to correct output the gfx COM interface definition from `gfx.slang`.

Another fix is to support `DLLImport` a stdcall mangled function in x86. This is done by computing the argument size in the compiler and pass it to `_slang_rt_load_dll_func`, which attempts both the unmangled name and the mangled name.

Also changes the build script to copy `gfx.slang` and `slang.slang` to the output path, and add the `slangc.exe`'s path to search directory so `import gfx` in user code can work.

With this PR, this test now runs:
```
import gfx;

{
    gfx.DeviceDesc deviceDesc = {};
    gfx.IDevice obj;
    if (gfx.succeeded(gfx.gfxCreateDevice(&deviceDesc, obj)))
        writeln("succ");
    else
        writeln("fail");
}
```

### DIFF
--- a/source/slang/slang-ir-layout.cpp
+++ b/source/slang/slang-ir-layout.cpp
@@ -256,6 +256,18 @@ static Result _calcNaturalSizeAndAlignment(
                 outSizeAndAlignment);
         }
         break;
+    case kIROp_OutType:
+    case kIROp_InOutType:
+    case kIROp_RefType:
+    case kIROp_RawPointerType:
+    case kIROp_PtrType:
+    case kIROp_NativePtrType:
+    case kIROp_ComPtrType:
+        {
+            *outSizeAndAlignment = IRSizeAndAlignment(sizeof(void*), sizeof(void*));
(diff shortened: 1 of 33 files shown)


source/slang/slang-ir-layout.cpp
```diff
@@ -256,6 +256,18 @@ static Result _calcNaturalSizeAndAlignment(
                 outSizeAndAlignment);
         }
         break;
+    case kIROp_OutType:
+    case kIROp_InOutType:
+    case kIROp_RefType:
+    case kIROp_RawPointerType:
+    case kIROp_PtrType:
+    case kIROp_NativePtrType:
```
jsmall-nvidia, line 264: NativeString?

jsmall-nvidia (approved):

PR #2372, opened 2022-08-21
Support compile-time constants formed by polynomials of generic value params

csyonghe:
With this PR, the type checker now recognizes polynomials as compile-time constants so expressions formed by generic int params can be used to specialize other types or used as array sizes.

For example,
```
struct <let u : int, let v : int>
{
    int arr[u*v];
}
```
is now a valid program, and the size of `arr` is evaluated to a `PolynomialIntVal`.
`PolynomialIntVal`s are canonicalized by the compiler, so `N+M+1-1` and `M+N` is recognized as the same value.

Polynomial canonicalization:
image

Specializing types with polynomial ints:
image


jsmall-nvidia (approved):
Nice

PR #2376, opened 2022-08-24
Compiler time evaluation of all int and bool operators.

csyonghe:
This PR implements compile time evaluation of all int and bool operators and the "?:" operator.

Fix code gen bug that causes `static const int` values defined in a struct type to emit as `integer_constant` field in target code.
The cause for this issue is that `static const` normally would lower as global insts, but if its value is dependent on a generic value parameter, then they will still be emitted inside the `IRStructType`. The solution is to introduce a hoisting pass to move these to global scope after IR specialization pass (when these constants no longer depend on any generic parameters).

Fixed a parser disambiguation issue when parsing things like `StructuredBuffer<MyType<1,2>>`.

Fixed a bug that causes COM interface definition not emitted in some situations. The fix makes sure that if an interface type is used from a `lookup_witness_table` inst, its definition will be emitted, even if it is accessed through a `NativePtr`. (By default, types referenced by `NativePtr` only need a forward declaration).


source/slang/slang-ast-type.cpp
```diff
@@ -387,15 +387,19 @@ Val* ArrayExpressionType::_substituteImplOverride(ASTBuilder* astBuilder, Substi
...
-    SLANG_ASSERT(arrlen);
+    IntVal* newArrayLength = nullptr;
+    if (arrayLength)
+    {
+        newArrayLength = as<IntVal>(arrayLength->substituteImpl(astBuilder, subst, &diff));
+        SLANG_ASSERT(newArrayLength);
+    }
     if (diff)
     {
         *ioDiff = 1;
```
jsmall-nvidia, line 398: Is there a reason to not increment, or add the diff?

source/slang/slang-ir-hoist-constants.cpp
```diff
@@ -0,0 +1,97 @@
+// slang-ir-hoist-constants.cpp
+#pragma once
```
jsmall-nvidia, line 2: #pragma once ?

jsmall-nvidia (approved):
Nice.

source/slang/slang-ast-type.cpp
csyonghe: Shouldn't matter. This is just used as a `bool`.

source/slang/slang-ir-hoist-constants.cpp
csyonghe: Fixed.

PR #2378, opened 2022-08-24
Interface static const

csyonghe:
This change allows compile time constant propagation through interfaces.

It is now possible to define a "compile time constant" requirement on an interface:
```
interface IFoo
{
    static const int value;
}
```
Which requires all implementations of the interface to provide a compile time constant named `value`.

Then references to `T.value` where `T:IFoo` can be used anywhere a static compile time constant is expected.

Here's an example:
image

Given the existing evaluation framework, the implementation is as easy as adding another `WitnessLookupIntVal` case. The previous `SomeIntVal` has been renamed to `FuncCallIntVal`.


source/slang/slang-check-decl.cpp
```diff
@@ -1053,6 +1053,33 @@ namespace Slang
                 addModifier(varDecl, m_astBuilder->create<NVAPIMagicModifier>());
             }
         }
+
+        if (auto interfaceDecl = as<InterfaceDecl>(varDecl->parentDecl))
+        {
+            if (auto basicType = as<BasicExpressionType>(varDecl->getType()))
+            {
+                switch (basicType->baseType)
+                {
```
jsmall-nvidia, line 1062: Could consider using `const BaseTypeInfo& info = BaseTypeInfo::getInfo(baseType);` will indicate intness, and will be kept hopefully in sync if any new int types are added.

jsmall-nvidia (approved):

PR #2380, opened 2022-08-24
Fix for Vulkan ray tracing test using a non-zero raygen shader index

lucy96chen:
Changes:
- Fixed math and alignment errors in Vulkan SBT creation resulting in all raygen shaders after the first being copied into the incorrect locations

### DIFF
--- a/tools/gfx/vulkan/vk-shader-table.cpp
+++ b/tools/gfx/vulkan/vk-shader-table.cpp
@@ -83,7 +83,8 @@ RefPtr<BufferResource> ShaderTableImpl::createDeviceBuffer(
 
         auto shaderGroupIndex = *shaderGroupIndexPtr;
         auto srcHandlePtr = handles.getBuffer() + shaderGroupIndex * handleSize;
-        memcpy(dstHandlePtr, srcHandlePtr, rtProps.shaderGroupBaseAlignment);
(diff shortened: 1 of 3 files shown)


tools/gfx/vulkan/vk-shader-table.cpp
```diff
@@ -84,7 +83,7 @@ RefPtr<BufferResource> ShaderTableImpl::createDeviceBuffer(
 
         auto shaderGroupIndex = *shaderGroupIndexPtr;
         auto srcHandlePtr = handles.getBuffer() + shaderGroupIndex * handleSize;
-        memcpy(dstHandlePtr, srcHandlePtr, handleSize);
+        memcpy(dstHandlePtr, srcHandlePtr, rtProps.shaderGroupBaseAlignment);
```
csyonghe, line 86: I think it should still be handleSize here.

csyonghe: Also maybe good to set remaining bytes between habdleSize and alignment to 0.

lucy96chen: dstHandlePtr consistently shows up already zeroed out in the debugger for me, but done and done just to be safe.

csyonghe (approved):

Issue #2353, opened 2022-08-08
Khronos Vulkan ValidationLayers Debug Printf not properly supported 


Qubaef:
I want to use Debug Printf extension available in Vulkan Validation Layers, described [here](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/debug_printf.md).

GL_EXT_debug_printf seems to be unsupported by slangc.exe.

In HLSL with glslangValidator.exe, following should work:
```
void main() {
float myfloat = 3.1415;
printf("My float is %f", myfloat);
}
```
but in slang it is not supported.

I am aware that such feature will take time to implement and I am not expecting it to appear anytime soon.
Are there any possible workarounds to enable the vulkan extension (capability) manually? Maybe something like inlining glsl/SPIR-V, or importing glsl/SPIR-V module in the slang file?

tangent-vector:
This is a good request. I see no reason why we can't map the built-in HLSL `printf` over to this extension.

This seems like a reasonable feature for a first-time contributor to work on. The approach for adding new intrinsic functions to Slang, or mapping existing functions to GLSL/SPIR-V extensions, is fairly easy to do (although not well documented).

The one big thing that can make a `printf`-like function tricky is the variable number of arguments. Slang doesn't currently support varargs, so `printf()` would need to be defined with multiple overloads.

Qubaef:
Support for the native printf hlsl capability got merged to master. It can now be used with GL_EXT_debug_printf Vulkan extension.

PR #2399, opened 2022-09-14
Add support for GL_EXT_debug_printf extension to slang

Qubaef:


Qubaef:
@tangent-vector I spent some time playing with the slang and trying to implement a POC for GL_EXT_debug_printf support mentioned in this [issue](https://github.com/shader-slang/slang/issues/2353).

I figured some things out, but some are still unclear to me and I wanted to ask for help:
1. In current version, printf function gets correctly mapped to glsl's debugPrintfEXT. However I don't understand how I can properly forward the string value.

This is how it looks right now in (slang to glsl mapping):
`printf("%d\n", outColor.x);` -> `debugPrintfEXT("(312235595)", (outColor_0.x));`

How can I get rid of the '(' ')' characters? How can I extract the proper value of the string?

2. I think that the best way of implementing this feature would be to allow conditional enabling/disabling just like for ray tracing extensions. One could want to implement own printf functionality (like in the examples, if I understand it correctly) or some custom definition - especially for cpp, where it is easy and quite common looking at examples). Considering that current implementation won't be perfect because of the limited amount of overloads (so the limited amount of parameters) it would be  good to be able to disable it. The only problem is, that I don't know how I can achieve such result (conditional enabling when capability was specified). I tried doing it basing on Ray Tracing capabilities, but it just too convoluted and I couldn't get it too work. Is there some easy way of doing it?

source/slang/hlsl.meta.slang
```diff
@@ -2737,7 +2737,14 @@ matrix<T,N,M> pow(matrix<T,N,M> x, matrix<T,N,M> y)
...
 
-// void printf( string format, ... );
+// __target_intrinsic(cpp, "printf(\"$0\", $1)")
+// __target_intrinsic(GL_EXT_debug_printf, "printf(\"$0\", $1)")
+// __target_intrinsic(GL_EXT_debug_printf)
+
+__target_intrinsic(hlsl)
+__glsl_extension(GL_EXT_debug_printf)
+__target_intrinsic(glsl, "debugPrintfEXT(\"$0\", $1)")
+void printf(String format, float arg0);
```
csyonghe, line 2747: Should use a generic, and perhaps use `NativeString`. No escape is necessary for $0.

```
__target_intrinsic(hlsl)
__glsl_extension(GL_EXT_debug_printf)
__target_intrinsic(glsl, "debugPrintfEXT($0)")
void printf(NativeString format);

__target_intrinsic(hlsl)
__glsl_extension(GL_EXT_debug_printf)
__target_intrinsic(glsl, "debugPrintfEXT($0, $1)")
void printf<T>(NativeString format, T arg0);

__target_intrinsic(hlsl)
__glsl_extension(GL_EXT_debug_printf)
__target_intrinsic(glsl, "debugPrintfEXT($0, $1, $2)")
void printf<T1, T2>(NativeString format, T1 arg0, T2 arg1);

...
```

Qubaef: I already tried with NatvieString and it gives similar results.
Your code above produces:
- in glsl output:  `debugPrintfEXT((312235595), (outColor_0.x));`
- spv-asm output: `ExtInst 35(NonSemantic.DebugPrintf) 1(DebugPrintf) 29 34` where `29:     25(int) Constant 312235595`

The value of NativeString gets mapped to an integer (probably an address or some kind of identifier?). I am looking for a way to map a string literal.

csyonghe: Then that must be the emit logic for a string literal is missing in `slang-emit-c-like.cpp` and somehow the default path just treats it as a int.

csyonghe: slang-emit-glsl.cpp line 1630 is where you need to modify.

Qubaef: I tried with what you proposed and it helped. Thank you.

Qubaef:
@csyonghe 

The POC is complete. I reviewed the whole process of generating the .glsl and .spv and then tried using produced shader in practice - it looks good and works as expected with the enabled GL_EXT_debug_printf Vulkan extension.

I also run slang-test.exe to check for possible regressions - no changes in the results.

One more thing left about this feature is it's dependency on GL_EXT_debug_printf capability (analogical to ray tracing capabilities). I tried adding it in the first commit, but it doesn't work and causes some random problems after enabling.
Should I try to make it work only when this capability is enabled? Or just leave as it is right now (always enabled) and remove the capability from my changes?

source/slang/slang-capability-defs.h
```diff
@@ -86,6 +86,7 @@ SLANG_CAPABILITY_ATOM1(SPIRV_1_5,   spirv_1_5,  Concrete,None,0,   SPIRV_1_4)
 SLANG_CAPABILITY_ATOM1(GLSLRayTracing,      __glslRayTracing,   Abstract,None,0,   GLSL)
 SLANG_CAPABILITY_ATOM1(GL_NV_ray_tracing,   GL_NV_ray_tracing,  Concrete,RayTracingExtension,0,   GLSLRayTracing)
 SLANG_CAPABILITY_ATOM2(GL_EXT_ray_tracing,  GL_EXT_ray_tracing, Concrete,RayTracingExtension,1,   GLSLRayTracing, SPIRV_1_4)
+SLANG_CAPABILITY_ATOM1(GL_EXT_debug_printf, GL_EXT_debug_printf, Concrete,None,0,   HLSL)
```
csyonghe, line 89: Should the last argument be GLSL rather than HLSL?

csyonghe:
I don't think we are tracking capabilities of various glsl extensions very well, so for now just having a `__glsl_extension` attribute on the `printf` declaration should be enough.

source/slang/slang-emit-glsl.cpp
```diff
@@ -1629,9 +1629,14 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+
+            StringBuilder buf;
+            const auto slice = as<IRStringLit>(inst)->getStringSlice();
+            StringEscapeUtil::appendQuoted(handler, slice, buf);
+
+            m_writer->emit(buf);
+
             return true;
         }
         case kIROp_GetStringHash:
```
csyonghe, line 1642: We need to update the implementation for `kIROp_GetStringHash` to do what `kIROp_StringLit` used to do on `inst->getOperand(0)` here.

Qubaef: I did what you said and tested it. Looks like it's working properly now.

csyonghe (approved):

PR #2402, opened 2022-09-19
Fix d3d12 implementation of `uploadBufferData`.

csyonghe:
Also added `ICommandBufferD3D12::ensureInternalDescriptorHeapsBound()` for raw d3d api interop.


lucy96chen (commented):
I don't remember doing more than moving code into this file, so looks good to me.

PR #2404, opened 2022-09-20
Support partial inference of generic arguments

tangent-vector:
A commonly requested feature is to be able to supply only some of the arguments to a generic explicitly, while allowing the rest to be inferred. A common example is a function that performs some kind of conversion:

    To convert<To, From>( From fromValue ) { .... }

A user would like to be able to call this operation like:

    int i = convert<int>( 1.0f );

but the current Slang type checker requires all or none of the generic arguments be supplied. Supplying all of the arguments is tedious:

    int i = convert<int, float>( 1.0f );

In this case, the `float` type argument is redundant and could be inferred from context. However, if the user tries to omit the generic argument list:

    int i = convert( 1.0f );

The current type-checker cannot infer the `int` type argument (even if one might claim it *should* infer based on the desired result type).

This change adds support for the `convert<int>(...)` case, by allowing a generic to be applied to a prefix of its explicit arguments, and then inferring the remaining arguments from contextual information when that "partially applied" generic is applied to value-level arguments.

Most of the changes are just plumbing: adding the notion of a partially applied generic and then supporting them during overload resolution.

A single test case is included that covers the `convert`-style use case. It is likely that more testing is needed to cover failure modes of this feature.


csyonghe (approved):
Looks good. We may need to extend the visitors in language server to be able to provide highlighting and hints for an PartiallySpecializedGenericExpr.

PR #2411, opened 2022-09-26
Use d3d12/vk debug layer when gfx debug layer is enabled.

csyonghe:


tools/gfx/vulkan/vk-device.cpp
```diff
@@ -662,7 +662,7 @@ SlangResult DeviceImpl::initialize(const Desc& desc)
             continue;
         descriptorSetAllocator.m_api = &m_api;
         initDeviceResult = initVulkanInstanceAndDevice(
-            desc.existingDeviceHandles.handles, ENABLE_VALIDATION_LAYER != 0);
+            desc.existingDeviceHandles.handles, ENABLE_VALIDATION_LAYER != 0 || isGfxDebugLayerEnabled());
```
jsmall-nvidia, line 665: Not related to this PR, but don't we want macros like ENABLE_VALIDATION_LAYER  to be prefixed?

jsmall-nvidia (approved):

PR #2413, opened 2022-09-27
Modified the new type system to support generic differentiable types …

saipraveenb25:
…and added support for differentiating overloaded functions.

 - Added AST and IR dictionary nodes to track differentiable types.
 - Added overload and generic function resolution.
 - Adjusted JVP logic to transcribe blocks, generic and functions in an 'on-demand' manner.


source/slang/slang-check-expr.cpp
```diff
@@ -1715,48 +1715,38 @@ namespace Slang
             return primalType;
     }
 
-    Expr* SemanticsExprVisitor::visitJVPDifferentiateExpr(JVPDifferentiateExpr* expr)
+    Type* SemanticsVisitor::processJVPFuncType(ASTBuilder* builder, FuncType* originalType)
     {
-        // Check/Resolve inner function declaration.
-        expr->baseFunction = CheckTerm(expr->baseFunction);
+        // Resolve JVP type here.
```
csyonghe, line 1720: We could simply find the witnesses and register them here with the function.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -221,102 +222,290 @@ struct DifferentiableTypeConformanceContext
...
+
+    void relocateNewTypes(IRBuilder* builder)
+    {
+        for (auto typeInst : generatedTypeList)
+        {
+            moveGlobalToBeforeUses(builder, typeInst);
+        }
+    }
+
+    void _createGenericDiffPairType(IRBuilder* builder)
```
csyonghe, line 398: If we get rid of generic type generation, we could just remove this, and remove the `moveGlobalToBeforeUses` logic as well.

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -890,18 +1203,169 @@ struct JVPTranscriber
...
+            this->transcribe(builder, block);
+        
+        // Reset builder position
+        builder->setInsertLoc(oldLoc);
+
+        return InstPair(primalFunc, diffFunc);
+    }
+
+    // Transcribe a generic definition
+    InstPair transcribeGeneric(IRBuilder* builder, IRGeneric* origGeneric)
```
csyonghe, line 1301: Is this still needed if we are running the pass post specialization?

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -953,15 +1418,56 @@ struct JVPTranscriber
...
+        // type. Only differentiate types if they appear inside a block.
+        // 
         if (auto origType = as<IRType>(origInst))
-            return InstPair(cloneInst(&cloneEnv, builder, origType), nullptr);
+        {   
+            // If this is a generic type, transcibe the parent 
+            // generic and derive the type from the transcribed generic's
+            // return value.
+            // 
+            if (as<IRGeneric>(origType->getParent()->getParent()) &&
```
csyonghe, line 1436: Do we still need this?

source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -1041,6 +1547,12 @@ struct JVPDerivativeContext
         // IRMakeDifferentialPair with an IRMakeStruct.
         // 
         modified |= processPairTypes(builder, module->getModuleInst(), (&diffConformanceContextStorage));
+        
+        // Temporary fix: Move generated types, if any, to before their use locations.
+        (&pairBuilderStorage)->relocateNewTypes(builder);
```
csyonghe, line 1552: We should remove this if possible.

csyonghe (approved):
I am OK with this going in, but we really should be cleaning up all the witness registration dictionary across the front-end and the IR.

csyonghe:
Merging this now.

Note that this is a 3500 line PR. Next time please try to create more frequent PRs and keep each one to be under 1000 lines, otherwise it is really difficult to review.

PR #2431, opened 2022-10-04
Add syntax for multi-level break.

csyonghe:
This change adds the syntax for multi-level `break` statement.

A future IR pass will be implemented to lower the multi-level breaks out to ordinary breaks.

### DIFF
--- a/source/slang/slang-language-server-ast-lookup.cpp
+++ b/source/slang/slang-language-server-ast-lookup.cpp
@@ -451,6 +451,13 @@ struct ASTLookupStmtVisitor : public StmtVisitor<ASTLookupStmtVisitor, bool>
         return false;
     }
 
+    bool visitLabelStmt(LabelStmt* stmt)
+    {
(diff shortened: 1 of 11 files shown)


source/slang/slang-language-server-ast-lookup.cpp
```diff
@@ -451,6 +451,13 @@ struct ASTLookupStmtVisitor : public StmtVisitor<ASTLookupStmtVisitor, bool>
         return false;
     }
 
+    bool visitLabelStmt(LabelStmt* stmt)
+    {
+        if (_isLocInRange(context, stmt->label.loc, stmt->label.getContent().getLength()))
```
jsmall-nvidia, line 456: Does this imply that preprocessor usage for the label, or doing a `#include` that holds the label won't work?

Both seem sort of iffy so perhaps that's ok. It might be a little surprising tho, with preprocessor/`#includes` notionally producing the same thing as having a source file contained the expanded output.

jsmall-nvidia (approved):

PR #2432, opened 2022-10-05
Shader caching

lucy96chen:
Changes:
- Added shader caching - This is disabled by default with both `shaderCachePath` and `shaderCacheFileSystem` under `IDevice::Desc` set to `nullptr`. Users wanting shader cache functionality will need to enable this manually by supplying at least one of these two fields on device creation. I've also added `RendererBase::getEntryPointCodeFromShaderCache`, which handles checking for existing entries in and adding new entries to the shader cache as needed, and all old call sites to `RendererBase::getEntryPointCode` have been redirected to this function instead.
- Added MD5 hashing functionality - This is contained within `slang-md5.h/cpp`. I've left the large block comment from the original source at the top of both files, but both files have had some changes to try to better conform them to Slang's coding style. All hashes used for the shader cache (AST-based and dependency-based) are MD5 hashes produced using this code.
- Added `IComponentType::computeASTBasedHash`, `IComponentType::computeDependencyBasedHash`, and `Linkage::computeDependencyBasedHash` - These functions handle producing the MD5 hashes used in the shader cache keys (dependency-based) and as checks for shader code changes (AST-based). The are implemented under their derived classes as necessary.
- Added `IShaderCacheStatistics` - This is largely used for debugging and testing purposes to ensure the cache behaves as expected between entries outright not existing, entries being dirty, and entries being up-to-date. This is class is currently implemented by `RendererBase`.
- Added unit tests for `Checksum` and the MD5 implementation as well as several tests for the shader cache itself that check all three statistics recorded in `IShaderCacheStatistics` for both a single entry, multiple entries, a single shader file with multiple entry points, and shader files which contain `import` and `#include`
- Fixes to prevent AST hash inconsistencies caused by uninitialized memory in Release mode only

TODOs for future PRs:
- Cache eviction policy and size control with tests - Next on the docket. This version of the cache has no size limitations.
- Some way to handle checking for out-of-date modules instead of using them verbatim even if the shader code associated with that module has changed on disk - This is less directly related to shader caching than it is to Slang as a whole but is listed here nonetheless as out-of-date modules being loaded for the same shader file was a major problem I ran into during preliminary testing. The test currently implements a workaround by re-creating the device between calls to `loadComputeProgram`.


tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +328,126 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
+    {
+        filename.append(String(shaderHash[i], 16).getBuffer());
+    }
+
+    *outFilename = filename;
+}
+
+void updateCacheEntry(ISlangMutableFileSystem* fileSystem, slang::IBlob* compiledCode, String shaderFilename, uint32_t* ASTHash)
+{
+    auto hashSize = 4 * sizeof(uint32_t); // MD5 hashes are 128 bits, represented here as an array of four uint32_t.
```
csyonghe, line 345: No hard code here, just `sizeof(Checksum)`

source/slang/slang.cpp
```diff
@@ -3178,6 +3187,107 @@ ISlangUnknown* Module::getInterface(const Guid& guid)
...
+    auto targetReq = linkage->targets[targetIndex];
+    uint32_t targetInfo[7];
+    targetInfo[0] = (uint32_t)targetReq->getTarget();
+    targetInfo[1] = (uint32_t)targetReq->getTargetFlags();
+    targetInfo[2] = (uint32_t)targetReq->getFloatingPointMode();
+    targetInfo[3] = (uint32_t)targetReq->getLineDirectiveMode();
+    targetInfo[4] = (uint32_t)targetReq->shouldDumpIntermediates();
+    targetInfo[5] = (uint32_t)targetReq->getForceGLSLScalarBufferLayout();
+    targetInfo[6] = (uint32_t)targetReq->shouldTrackLiveness();
+    hashGen.update(&context, targetInfo, 7 * sizeof(uint32_t));
```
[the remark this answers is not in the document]
lucy96chen: `update` wants the data to be passed in as a `const void*` though

source/slang/slang.cpp
```diff
@@ -1618,10 +1620,19 @@ void TranslationUnitRequest::_addSourceFile(SourceFile* sourceFile)
...
 
     auto pathInfo = sourceFile->getPathInfo();
-    if (pathInfo.hasFileFoundPath())
+    if (pathInfo.hasFoundPath())
     {
         getModule()->addFilePathDependency(pathInfo.foundPath);
     }
+    else
+    {
+        // No path exists for this source, so we generate a new string to use as a
```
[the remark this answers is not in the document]
lucy96chen: Why the source of the module (I'm presuming you mean `getModule()->getASTBasedHash()`) and not the contents of the source file being added?

slang.h
```diff
@@ -4423,15 +4443,20 @@ namespace slang
             IBlob**     outCode,
             IBlob**     outDiagnostics = nullptr) = 0;
 
-            /** Get the name for the specified shader file containing code for the entry point at
-                'entryPointIndex' for the chosen 'targetIndex'
+            /** Compute the hash code of all compilation arguments for the specified target. This includes
+                all dependent source file names, preprocessor defines, target options and other compiler options.
```
csyonghe, line 4447: The comment isn't accurate, for the linkage, we no longer compute hash for preprocessor defines and other compiler options here.

source/compiler-core/slang-md5.cpp
```diff
@@ -216,11 +216,26 @@ namespace Slang
...
+    {
+        update(ctx, string.begin(), string.getLength());
+    }
+
+    void MD5HashGen::update(MD5Context* ctx, String str)
+    {
+        update(ctx, str.getBuffer(), str.getLength());
+    }
+
+    void MD5HashGen::update(MD5Context* ctx, slang::Checksum checksum)
```
csyonghe, line 230: `const slang::Checksum&`

source/compiler-core/slang-md5.h
```diff
@@ -45,11 +49,31 @@ namespace Slang
...
+
+        // Helper update function for raw values (e.g. ints, uints)
+        template<typename T>
+        void update(MD5Context* ctx, T val)
+        {
+            update(ctx, &val, sizeof(T));
+        }
+        // Helper update function for Slang::List
+        template<typename T>
+        void update(MD5Context* ctx, List<T> list)
```
csyonghe, line 61: `const List<T>` to avoid copy.

source/compiler-core/slang-md5.h
```diff
@@ -45,11 +49,31 @@ namespace Slang
...
+        {
+            update(ctx, list.getBuffer(), list.getCount());
+        }
+        // Helper update function for UnownedStringSlice
+        void update(MD5Context* ctx, UnownedStringSlice string);
+        // Helper update function for Slang::String
+        void update(MD5Context* ctx, String str);
+        // Helper update function for Checksums
+        void update(MD5Context* ctx, slang::Checksum checksum);
```
csyonghe, line 70: `const slang::Checksum&`

source/slang/slang-checksum-utils.h
```diff
@@ -0,0 +1,28 @@
+// slang-checksum-utils.h
+#include "../../slang.h"
+#include "../core/slang-basic.h"
+
+namespace Slang
+{
+    // Returns the stored hash in checksum as a String.
+    inline StringBuilder checksumToString(slang::Checksum checksum)
```
csyonghe, line 8: const checksum&

source/slang/slang.cpp
```diff
@@ -1320,6 +1321,37 @@ SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::createCompileRequest(
...
+    for (auto& searchDir : searchDirectoryList)
+    {
+        auto searchPath = searchDir.path;
+        hashGen.update(&context, searchPath);
+    }
+
+    // Add the preprocessor definitions to the hash
+    for (auto& key : preprocessorDefinitions)
+    {
+        hashGen.update(&context, key.Key);
```
csyonghe, line 1347: Need to hash `key.Value` as well.

source/slang/slang.cpp
```diff
@@ -1320,6 +1321,37 @@ SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::createCompileRequest(
...
+    }
+
+    hashGen.finalize(&context, &hash);
```
csyonghe, line 1350: TargetRequest should be here too. Since we don't have `targetIndex` here, just hash all the targets.

source/slang/slang.cpp
```diff
@@ -1631,7 +1663,14 @@ void TranslationUnitRequest::_addSourceFile(SourceFile* sourceFile)
         // for non-file-based dependencies later when shader files are being hashed for
         // the shader cache.
 
-        // TODO: Add a file path dependency using a fake path string for this source
+        slang::Checksum sourceHash;
+        MD5Context context;
+        MD5HashGen hashGen;
```
csyonghe, line 1668: These five lines should just become another helper function.
How about `slang::Checksum computeChecksumForString(UnownedStringSlice text)`?

source/slang/slang.cpp
```diff
@@ -3205,86 +3244,53 @@ SlangResult Module::getDependencyBasedHashCode(
...
+        hashGen.update(&context, file);
     }
 
-    // Add the linkage to the hash
-    auto linkage = getLinkage();
-
     // Add the corresponding TargetRequest based on targetIndex to the hash (except
-    // for the linkage because we've already included it)
+    // for the linkage because the linkage's hash will be computed separately)
+    auto linkage = getLinkage();
```
csyonghe, line 3252: Move these to the linkage::computeDependencyBasedHash.

tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +331,118 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
+    ComPtr<slang::ISession> session;
+    getSlangSession(session.writeRef());
+
+    slang::Checksum linkageHash;
+    slang::Checksum programHash;
+    session->computeDependencyBasedHash(&linkageHash);
+    program->computeDependencyBasedHash(entryPointIndex, targetIndex, &programHash);
+
+    StringBuilder shaderFilename = checksumToString(linkageHash);
+    shaderFilename.append(checksumToString(programHash));
```
csyonghe, line 374: This is not correct. Should combine `programHash` and `linkageHash` for the final fileName. Just run MD5 again on these two hashes. Consider adding a new helper function: `Checksum combineChecksum(const Checksum& v0, const Checksum& v1)

source/slang/slang.cpp
lucy96chen: I already made an `update` helper for UnownedStringSlice, and it might look a bit weird to turn all five of these lines into a single helper when none of the other hashes are structured like that

source/slang/slang.cpp
lucy96chen: See Slack. Just like how we're only hashing the relevant entry point, I think it makes more sense to pass in `targetIndex` here and only hash the specified `TargetRequest` (which is also more in line with how `getEntryPointCode()` handles fetching the compiled code, a specific entry point on a specific target).

csyonghe: For gfx, `ISession` should only have one target. However, we should still move these out of `Module` into `Linkage`. We have have `Linkage::computeDependencyBasedHash` take in `targetIndex` as argument, and pass in `0` in gfx when calling it.

source/slang/slang.cpp
csyonghe: We could have a lot of other places where we just want to compute MD5 for a string and nothing else, so having this helper will be useful in many ways.

tools/gfx/renderer-shared.cpp
csyonghe: Concatenating the hashes makes the file name 64 characters long, which isn't necessary. Instead write a `Checksum combineChecksum(const Checksum& c0, const Checksum c1)`, and then do:
```
auto combinedChecksum = combineChecksum(linkageHash, programHash);
auto shaderFilename = checksumToString(combinedChecksum);
```

slang.h
```diff
@@ -4134,6 +4134,26 @@ namespace slang
         None, UnsizedArray, StructuredBuffer, ConstantBuffer, ParameterBlock
     };
 
+    // A struct for storing checksums designed to be used with the MD5 hashing implementation
+    struct Checksum
+    {
+        uint32_t checksum[4] = { 0 };
+
+        Checksum& operator=(const Checksum& rhs)
```
csyonghe, line 4142: Why is this necessary? We should be able to assign directly without this overload.

source/compiler-core/slang-md5.h
csyonghe: Use reference in stead by value.

source/slang/slang-checksum-utils.h
csyonghe: by reference, please.

slang.h
lucy96chen: Fixed in new push

source/compiler-core/slang-md5.cpp
lucy96chen: Fixed in new push

source/compiler-core/slang-md5.h
lucy96chen: Fixed in new push

source/slang/slang-checksum-utils.h
lucy96chen: Fixed in new push

source/core/slang-md5.h
```diff
@@ -0,0 +1,81 @@
...
+
+        // Helper update function for raw values (e.g. ints, uints)
+        template<typename T>
+        void update(MD5Context* ctx, T val)
+        {
+            update(ctx, &val, sizeof(T));
+        }
+        // Helper update function for Slang::List
+        template<typename T>
+        void update(MD5Context* ctx, const List<T> list)
```
csyonghe, line 62: This `List<T>` needs to be a reference as well.

tools/gfx/renderer-shared.cpp
lucy96chen: Done in the most recent push after moving `slang-md5` back to `source/core` because it seems `renderer-shared.cpp` throws unresolved external errors at me if it includes that file when it's located in `source/compiler-core`

source/core/slang-md5.h
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    class MD5HashGen
+    {
+    public:
+        void init(MD5Context* ctx);
+
+        // Helper update function for raw values (e.g. ints, uints)
+        template<typename T>
+        void update(MD5Context* ctx, T val)
```
csyonghe, line 56: Probably better to use `const T&` here, since that allows efficiently pass in big objects and not just `int`s.

slang.h
lucy96chen: I added it because you mentioned it on Slack yesterday, mostly. I'll just remove it if direct assignment works fine

tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +331,117 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
 
+void updateCacheEntry(ISlangMutableFileSystem* fileSystem, slang::IBlob* compiledCode, String shaderFilename, slang::Checksum ASTHash)
+{
+    auto hashSize = sizeof(slang::Checksum);
+
+    auto bufferSize = hashSize + compiledCode->getBufferSize();
+    List<uint8_t> contents;
+    contents.setCount(bufferSize);
+    uint8_t* buffer = contents.begin();
+    memcpy(buffer, (void*)ASTHash.checksum, hashSize);
```
csyonghe, line 342: Can just do `memcpy(buffer, &ASTHash, hashSize)`.

tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +331,117 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
+    }
+
+    // Produce a string which we can use to query the shader cache by combining two separate hashes which
+    // together comprise all the compilation arguments for this program.
+    ComPtr<slang::ISession> session;
+    getSlangSession(session.writeRef());
+
+    slang::Checksum linkageHash;
+    slang::Checksum programHash;
+    session->computeDependencyBasedHash(targetIndex, &linkageHash);
```
csyonghe, line 370: We could just pass `targetIndex` to `session->computeDependencyBasedHash()` here.

slang.h
csyonghe: Sorry, I meant `operator ==` yesterday.

lucy96chen: yeah looks fine, getting rid of this

tools/gfx/renderer-shared.cpp
csyonghe: let's rename it to `astHash` to be consistent with the naming convention.

tools/gfx/renderer-shared.cpp
lucy96chen: I did though

csyonghe: Ah, that's great. Sorry I didn't see the latest version for some reason...

tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +331,117 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
+    // Query shaderCacheFileSystem for an entry whose key matches shaderFilename
+    //    - If we find it, then copy the file contents into memory and return in outCode
+    auto result = shaderCacheFileSystem->loadFile(shaderFilename.getBuffer(), codeBlob.writeRef());
+    
+    if (SLANG_FAILED(result))
+    {
+        // If we didn't find it, call program->getEntryPointCode() to get and return the code. We also
+        // make sure to save a new entry in the shader cache.
+        program->getEntryPointCode(entryPointIndex, targetIndex, codeBlob.writeRef(), outDiagnostics);
+        ISlangMutableFileSystem* fileSystem;
```
csyonghe, line 391: This needs to be a `ComPtr<ISlangMutableFileSystem>` and then use `writeRef()` in the call to `queryInterface` to prevent memory leak.

tools/gfx/renderer-shared.cpp
```diff
@@ -325,6 +331,117 @@ void PipelineStateBase::initializeBase(const PipelineStateDesc& inDesc)
...
+        // the AST hash with the compiled code, we can determine this by comparing the stored hash with the
+        // AST hash generated earlier.
+        auto entryContents = codeBlob->getBufferPointer();
+        auto hashSize = 4 * sizeof(uint32_t);
+        if (memcmp(ASTHash.checksum, entryContents, hashSize) != 0)
+        {
+            // The AST hash stored in the entry does not match the AST hash generated earlier, indicating
+            // that the shader code has changed and the entry needs to be updated.
+            program->getEntryPointCode(entryPointIndex, targetIndex, codeBlob.writeRef(), outDiagnostics);
+            ISlangMutableFileSystem* fileSystem;
```
csyonghe, line 411: Same here, ComPtr.

tools/gfx/renderer-shared.cpp
lucy96chen: done, added `ComPtr<ISlangMutableFileSystem> mutableShaderCacheFileSystem` to `RendererBase` and moved `queryInterface` inside `RendererBase::initialize()`.

tools/gfx/renderer-shared.cpp
lucy96chen: See above

source/slang/slang-compiler.cpp
```diff
@@ -222,6 +224,34 @@ namespace Slang
...
+        slang::Checksum hashCode;
+        MD5HashGen hashGen;
+        MD5Context context;
+        hashGen.init(&context);
+
+        assert(getName());
+        if (getName())
+        {
+            hashGen.update(&context, getName()->text);
+        }
```
tangent-vector, line 239: Code style: this is code is screaming for the result of `getName()` to be put into a local variable:

```
auto name = getName();
SLANG_ASSERT(name); // Note: we are supposed to be using `SLANG_ASSERT` instead of plain `assert`
if (name)
{
    hashGen.update(&context, name->text);
}
```

tangent-vector: It is unclear to me why this code has both an `assert` that `getName()` is non-null, but then also uses an `if` to defend against it being null.

source/slang/slang-compiler.cpp
```diff
@@ -290,6 +320,32 @@ namespace Slang
         return Super::getInterface(guid);
     }
 
+    SlangResult TypeConformance::computeDependencyBasedHash(
+        slang::Checksum* outHashCode)
+    {
+        auto subtypeWitness = m_subtypeWitness->toString();
```
tangent-vector, line 326: It seems important for somebody to confirm that the `toString()` operation on these witnesses will return a "sufficiently unique" string.

source/slang/slang-serialize.cpp
```diff
@@ -210,10 +210,11 @@ SerialClasses::SerialClasses():
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! SerialWriter  !!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-SerialWriter::SerialWriter(SerialClasses* classes, SerialFilter* filter) :
+SerialWriter::SerialWriter(SerialClasses* classes, SerialFilter* filter, Flags flags) :
     m_arena(2048),
     m_classes(classes),
-    m_filter(filter)
+    m_filter(filter),
+    m_flags(flags)
```
tangent-vector, line 217: Coding style: the diffs here are an example of why it is beneficial to rearrange the punctuation for C++ constructor initializer lists, so that the punctuation comes before each initializer:

```
SerialWriter::SerialWriter(...)
    : m_arena(2048)
    , m_classes(classes)
    , m_filter(filter)
    , m_flags(flags)
{ ... }
```

Note that when the initializer list is written this way, the addition of `m_flags(...)` is a one-line diff that does *not* affect the line for `m_filter(...)` (which is unrelated to the change).

source/slang/slang.cpp
```diff
@@ -1318,6 +1321,63 @@ SLANG_NO_THROW SlangResult SLANG_MCALL Linkage::createCompileRequest(
...
+    hashGen.update(&context, targetProfile.getVersion());
+    hashGen.update(&context, targetProfile.getFamily());
+
+    auto targetProfileName = String(targetProfile.getName());
+    hashGen.update(&context, targetProfileName);
+
+    auto cookedCapabilities = targetReq->getTargetCaps().getExpandedAtoms();
+    for (auto& capability : cookedCapabilities)
+    {
+        hashGen.update(&context, capability);
```
tangent-vector, line 1373: Thinking out loud: there is a fair bit of subtlety here, where some of the fields we are including in the hash are stored as an `enum` that is *not* guaranteed to use the exact same encoding across different builds of the compiler. That situation sounds worrying ("what if a new version of the compiler uses the same integer value to represent a different capability?"), but it is actually okay in practice for the very specific reason that we are also including the compiler version into the hash, so different compiler versions would already be hashing different data at an earlier point, so that overlapping capability encodings (for example) would not create a problem.

source/slang/slang.cpp
```diff
@@ -3178,6 +3246,43 @@ ISlangUnknown* Module::getInterface(const Guid& guid)
     return Super::getInterface(guid);
 }
 
+// One implementation under ComponentType that takes entryPointIndex/targetIndex
+// One per subtype without
```
tangent-vector, line 3250: These comments were developer notes during the process of implementing the feature, but aren't really useful to anybody reading the code now.

source/slang/slang.cpp
```diff
@@ -3178,6 +3246,43 @@ ISlangUnknown* Module::getInterface(const Guid& guid)
...
+    MD5HashGen hashGen;
+    MD5Context context;
+    hashGen.init(&context);
+
+    // Add file path dependencies to the hash
+    auto fileDeps = getFilePathDependencyList();
+    for (auto& file : fileDeps)
+    {
+        hashGen.update(&context, file);
+    }
```
tangent-vector, line 3263: Every `ComponentType` has the ability to iterate over its file-path dependencies. As such, the logic right here could seemingly be folded into a base `ComponentType::computeDependencyBasedHash()` function, and not get involved in any of this recursive logic.

It seems to be that there are simplifications here that could be achieved.

This code could *really* use some comments along the way that explain the implementation strategy being used in a top-down fashion so that a developer walking through the code can understand *why* it is doing things the way it does (remember: the "why" is typically more important than the "what").

source/slang/slang.cpp
```diff
@@ -3376,6 +3481,18 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
+{
+    SLANG_UNUSED(outHashCode);
+    return SLANG_E_NOT_AVAILABLE;
+}
+
+SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::computeASTBasedHash(slang::Checksum* outHashCode)
+{
+    SLANG_UNUSED(outHashCode);
+    return SLANG_E_NOT_AVAILABLE;
+}
```
tangent-vector, line 3494: I'm not sure why these need to be implemented here at all, as opposed to being pure-virtual methods in `ComponentType`.

I'm also not clear on why these are `COM`-ish methods (with `SLANG_MCALL`, etc.) when they aren't part of any COM interface that `ComponentType` is implementing (at least AFAICT).

source/slang/slang.cpp
```diff
@@ -3708,6 +3825,48 @@ CompositeComponentType::CompositeComponentType(
...
+    slang::Checksum hashCode;
+    MD5HashGen hashGen;
+    MD5Context context;
+    hashGen.init(&context);
+
+    for (Index i = 0; i < componentCount; ++i)
+    {
+        slang::Checksum tempHash;
+        getChildComponent(i)->computeDependencyBasedHash(&tempHash);
+        hashGen.update(&context, tempHash);
```
tangent-vector, line 3841: There is a "code smell" here that is making me realize the signatures of these `compute...Hash(...)` functions aren't quite what we want. The functions should probably *not* be yielding a `Checksum` as their output, and should instead be functions that take an `MD5HashGen` and `MD5Context` as input (or a convenience type that bundles the two...):

```
SlangResult CompositeComponentType::computeDependencyBasedHash(MD5HashGen& hashGen, MD5Context const* context)
{ ... }
```

Using that as the signature for the core recursive routine would yield two big improvements:

1. Most of the implementation routines would no longer need to declare their own local `hashCode`, `hashGen`, and `context` variables, nor call `hashGen.init()`.

2. When making a recursive call like the one tagged for this comment, a function would just be able to pass `hashGen` and `context` down to the callee, rather than have to declare a distinct `tempHash` and then use `hashGen.update(..., tempHash)`.

Each of these leads to less code overall (so that's a nice win already), but item (2) there could actually lead to a better quality hashed output for hashing algorithms where the internal state of a hash generator is larger than the size of the hash code / digest produced at the end.

tangent-vector: I'm realizing just now that the call to compute the hash for the child component isn't doing anything on a failure `slang::Result`. That seems incorrect to me.

Either these routines should be using some kind of `SLANG_RETURN_ON_FAIL` or similar to handle the recursive calls, so that errors get propagated, or if none of these routines can actually fail, we shouldn't be defining them to return a `Result` in the first place.

source/slang/slang.cpp
```diff
@@ -4187,6 +4346,37 @@ SpecializedComponentType::SpecializedComponentType(
...
+    slang::Checksum hashCode;
+    MD5HashGen hashGen;
+    MD5Context context;
+    hashGen.init(&context);
+
+    auto fileDependencies = getFilePathDependencies();
+    for (auto& file : fileDependencies)
+    {
+        hashGen.update(&context, file);
+    }
```
tangent-vector, line 4360: As per an earlier comment, we should only need *one* site in the code where we loop over the file-path dependencies for the root `ComponentType` for which a hash is being requested. This logic shouldn't need to be part of the recursive worker routine.

source/slang/slang.cpp
```diff
@@ -4231,6 +4421,26 @@ void RenamedEntryPointComponentType::acceptVisitor(
...
+
+    slang::Checksum baseHash;
+    getBase()->computeDependencyBasedHash(&baseHash);
+    hashGen.update(&context, baseHash);
+
+    hashGen.finalize(&context, &hashCode);
+
+    *outHashCode = hashCode;
+    return SLANG_OK;
+}
```
tangent-vector, line 4442: My sense is also that we shouldn't bother implementing this logic around `EntryPoint` or `RenamedEntryPointComponentType`. This is again a case where a single non-recursive routine written for any `ComponentType` can iterate over the entry points of that component (without needing to know which sub-type it is), and ask for their names (whether or not they've been renamed). That should result in simpler and more general code.

tools/gfx-unit-test/imported.slang
```diff
@@ -0,0 +1,8 @@
+struct TestFunction
+            {
+                void simpleElementAdd(RWStructuredBuffer<float> buffer, uint index)
+                {
+                    var input = buffer[index];
+                    buffer[index] = input + 1.0f;
+                }
+            };
```
tangent-vector, line 8: The indentation in these files is a little weird.

tools/gfx-unit-test/shader-cache-tests.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+        ComPtr<IDevice> device;
+        ComPtr<IPipelineState> pipelineState;
+        ComPtr<IResourceView> bufferView;
+
+        // The test shader needs to physically exist on disk for loadComputeProgram to
+        // correctly load it while the shader cache should exist in memory to avoid needing
+        // to clean up old cache files from previous test runs.
+        ComPtr<ISlangMutableFileSystem> diskFileSystem;
+        ComPtr<ISlangMutableFileSystem> cacheFileSystem;
```
tangent-vector, line 29: I'm a bit confused why the `cacheFileSystem` cannot be used for loading the actual shaders as well, instead of just for storing the cache.

tangent-vector (commented):
This all seems reasonable to me, but there are also several items here around fit-and-finish that should be cleaned up. Several of those are coding style things. The main functional thing that I'd like to see addressed is that the core hash computation routines seems to be more complicated than necessary, putting logic in the "leaf" routines that could be handled in the base `ComponentType` case.

I would like to either see simplification applied to the implementation of the hash computation, *or* a sufficient amount of comments to help explain to a reviewer (or future maintainer of this code) why things are (and need to be) the way they are.

tools/gfx-unit-test/imported.slang
lucy96chen: Side effect of all the shader file writes directly from the test. Probably the extra indentation in the test to align those lines.

tools/gfx-unit-test/shader-cache-tests.cpp
lucy96chen: As I brought up in the gfx channel last week, using `cacheFileSystem` (which is a `MemoryFileSystem` and not a `RelativeFileSystem` like `diskFileSystem` is) to save the shaders results in `loadComputeProgram()` being unable to find them. I was told the shaders need to be saved to disk in order for `loadComputeProgram()` to find them.

slang.h
```diff
@@ -4134,6 +4134,12 @@ namespace slang
         None, UnsizedArray, StructuredBuffer, ConstantBuffer, ParameterBlock
     };
 
+    // A struct for storing checksums designed to be used with the MD5 hashing implementation
+    struct Checksum
+    {
+        uint32_t checksum[4] = { 0 };
+    };
+
```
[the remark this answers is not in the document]
lucy96chen: This was mostly added to encapsulate the MD5 hash representation of a 4 element uint32_t array, otherwise all the hashing code was creating and passing around arrays everywhere. I'm fine with a rename to MD5Digest and changing it in the future if need be.

source/slang/slang-compiler.cpp
lucy96chen: Leftover code that I missed going back through, done and removed

source/slang/slang.cpp
csyonghe: Let's create a class `MD5DigestBuilder` that owns both a `MD5HashGen` and `MD5Context`, then 
define the `computeDependencyBasedHashImpl(MD5DigestBuilder& builder)` as an abstract method in `ComponentTypeBase`, and implement the existing `IComponentType::computeDependencyBasedHash()` with a direct call to `computeDependencyBasedHashImpl` with a fresh  `MD5DigestBuilder`, there we have the opportunity to call `Linkage::computeDependencyBasedHash()` once to initialize the `MD5DigestBuilder`.

The rest of implementations of `computeDependencyBasedHashImpl` should only recursively call `computeDependencyBasedHashImpl` and not the public `computeDependencyBasedHash`.

slang.h
```diff
@@ -4301,6 +4307,14 @@ namespace slang
...
             SlangInt conformanceIdOverride,
             ISlangBlob** outDiagnostics) = 0;
+
+            /** Computes the hash for the linkage, which includes preprocessor defines, the compiler version,
+                and other compiler options. This is then merged with the hash produced for the program to
+                produce a key that can be used with the shader cache.
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL computeDependencyBasedHash(
+            SlangInt targetIndex,
+            Checksum* outHash) = 0;
```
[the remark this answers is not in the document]
lucy96chen: Probably doesn't, `RendererBase::getEntryPointCodeFromShaderCache()` should be the only location it's being used. Removed.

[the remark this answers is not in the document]
lucy96chen: Scratch that, this needs to be all the way up under `ISession` because `RendererBase` only has access to `getSlangSession` as a means of getting the linkage for hashing.

slang.h
```diff
@@ -4421,6 +4435,19 @@ namespace slang
...
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL computeDependencyBasedHash(
+            Checksum* outHashCode) = 0;
+
+            /** Get the MD5 hash generated from this component type's AST. Not all component types
+                will store an AST, and consequently, not all component types will have a
+                meaningful implementation for this function.
+            */
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL computeASTBasedHash(Checksum* outHashCode) = 0;
+
```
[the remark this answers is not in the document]
lucy96chen: Added, but I'm really unsure how to correctly describe these.

slang.h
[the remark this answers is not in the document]
csyonghe: Thinking of it now, this can be removed from public interface, and all its logic can be implemented in the new `ComponentType::computeDependencyBasedHash`.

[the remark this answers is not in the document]
csyonghe: The structure of methods should be:
```
 // public interface
IComponentType::computeDependencyBasedHash(MD5Digest** outDigest, int targetIndex, int entryPointIndex);

 // internal abstract method that every subtype of `ComponentType` should implement.
virtual ComponentType::computeDependencyBasedHashImpl(MD5DigestBuilder& builder) = 0;

// implementation of public `IComponentType` interface by simply calling into 
// `ComponentType::computeDependencyBasedHashImpl` + other one-time global things,
// e.g. linkage, target options, entry-point names, etc.
ComponentType::computeDependencyBasedHash(MD5Digest** outDigest, int targetIndex, int entryPointIndex); 
```

source/slang/slang.cpp
lucy96chen: Moved up to `CompositeComponentType`, which is the base that `RendererBase::getEntryPointCodeFromShaderCache` calls into.

source/slang/slang.cpp
lucy96chen: COM-ish methods because they're defined under `IComponentType`, which is necessary for `RendererBase::getEntryPointCodeFromShaderCache` to call them on `program` as it's passed in as an `IComponentType`.

source/slang/slang.cpp
lucy96chen: Moved up to `CompositeComponentType`.

tools/gfx-unit-test/shader-cache-tests.cpp
lucy96chen: I've edited this comment to better reflect that both file systems are needed due to needing to reset the cache between consecutive test runs and to save shaders to disk because that's where `loadComputeProgram()` looks when loading a shader.

tools/gfx-unit-test/imported.slang
lucy96chen: All the files with weird indentation are the ones that are constantly being overwritten at various points in the shader cache tests, so I'm going to opt to not bother in favor of having the piped code in the test look nicer.

source/slang/slang.cpp
lucy96chen: Logic to handle entry point name and name override as specified by `entryPointIndex` moved up to `CompositeComponentType`. Both `EntryPoint` and `RenamedEntryPointComponentType` immediately return now.

csyonghe: They should be moved into ComponentType class instead of CompositeComponentType. See the other comments on how to restructure these things.

source/slang/slang.cpp
csyonghe: Should be in ComponentType.

source/slang/slang.cpp
csyonghe: ComponentType.

slang.h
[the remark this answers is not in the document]
csyonghe: I am not seeing this being resolved.

source/slang/slang.cpp
lucy96chen: Fixed locally, will be in next push

source/slang/slang.cpp
lucy96chen: Fixed locally, will be in next push

source/slang/slang.cpp
lucy96chen: Restructuring done and should be working based on shader cache tests, will be in next push

source/slang/slang.cpp
lucy96chen: Fixed locally, will be in next push

source/slang/slang.cpp
lucy96chen: Fixed locally, will be in next push

slang.h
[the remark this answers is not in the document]
lucy96chen: Fixed locally, will be in next push

source/slang/slang.cpp
lucy96chen: I've added a comment about this in the new `ComponentType::computeDependencyBasedHash` implementation

source/slang/slang-compiler.cpp
lucy96chen: Deferring to a separate cleanup PR

source/slang/slang-compiler.h
```diff
@@ -267,30 +274,37 @@ namespace Slang
...
+            slang::IBlob** outCode,
             slang::IBlob** outDiagnostics) SLANG_OVERRIDE;
-        SLANG_NO_THROW SlangResult SLANG_MCALL computeDependencyBasedHash(slang::Checksum* outHashCode) SLANG_OVERRIDE;
-        SLANG_NO_THROW SlangResult SLANG_MCALL computeASTBasedHash(slang::Checksum* outHashCode) SLANG_OVERRIDE;
         SLANG_NO_THROW SlangResult SLANG_MCALL specialize(
             slang::SpecializationArg const* specializationArgs,
             SlangInt                        specializationArgCount,
-            slang::IComponentType**         outSpecializedComponentType,
-            ISlangBlob**                    outDiagnostics) SLANG_OVERRIDE;
+            slang::IComponentType** outSpecializedComponentType,
```
csyonghe, line 286: irrelavent whitespace change.

source/slang/slang-compiler.h
```diff
@@ -267,30 +274,37 @@ namespace Slang
         SLANG_NO_THROW slang::ISession* SLANG_MCALL getSession() SLANG_OVERRIDE;
         SLANG_NO_THROW slang::ProgramLayout* SLANG_MCALL getLayout(
             SlangInt        targetIndex,
-            slang::IBlob**  outDiagnostics) SLANG_OVERRIDE;
+            slang::IBlob** outDiagnostics) SLANG_OVERRIDE;
         SLANG_NO_THROW SlangResult SLANG_MCALL getEntryPointCode(
             SlangInt        entryPointIndex,
             SlangInt        targetIndex,
-            slang::IBlob**  outCode,
+            slang::IBlob** outCode,
```
csyonghe, line 281: irrelavent whitespace change.

source/slang/slang-compiler.h
```diff
@@ -267,30 +274,37 @@ namespace Slang
         SLANG_NO_THROW slang::ISession* SLANG_MCALL getSession() SLANG_OVERRIDE;
         SLANG_NO_THROW slang::ProgramLayout* SLANG_MCALL getLayout(
             SlangInt        targetIndex,
-            slang::IBlob**  outDiagnostics) SLANG_OVERRIDE;
+            slang::IBlob** outDiagnostics) SLANG_OVERRIDE;
```
csyonghe, line 277: irrelavent whitespace change.

source/slang/slang-compiler.h
```diff
@@ -267,30 +274,37 @@ namespace Slang
...
-            ISlangBlob**                    outDiagnostics) SLANG_OVERRIDE;
+            slang::IComponentType** outSpecializedComponentType,
+            ISlangBlob** outDiagnostics) SLANG_OVERRIDE;
         SLANG_NO_THROW SlangResult SLANG_MCALL renameEntryPoint(
             const char* newName,
             slang::IComponentType** outEntryPoint) SLANG_OVERRIDE;
         SLANG_NO_THROW SlangResult SLANG_MCALL link(
-            slang::IComponentType**         outLinkedComponentType,
-            ISlangBlob**                    outDiagnostics) SLANG_OVERRIDE;
+            slang::IComponentType** outLinkedComponentType,
```
csyonghe, line 292: irrelavent whitespace change.

source/slang/slang-compiler.h
```diff
@@ -267,30 +274,37 @@ namespace Slang
...
+            ISlangBlob** outDiagnostics) SLANG_OVERRIDE;
         SLANG_NO_THROW SlangResult SLANG_MCALL getEntryPointHostCallable(
             int                     entryPointIndex,
             int                     targetIndex,
-            ISlangSharedLibrary**   outSharedLibrary,
-            slang::IBlob**          outDiagnostics) SLANG_OVERRIDE;
+            ISlangSharedLibrary** outSharedLibrary,
```
csyonghe, line 297: irrelavent whitespace change.

source/slang/slang-compiler.h
```diff
@@ -762,7 +796,12 @@ namespace Slang
         void acceptVisitor(ComponentTypeVisitor* visitor, SpecializationInfo* specializationInfo)
             SLANG_OVERRIDE;
 
-        SLANG_NO_THROW SlangResult SLANG_MCALL computeDependencyBasedHash(slang::Checksum* outHashCode) SLANG_OVERRIDE;
+        virtual void updateDependencyBasedHash(
+            HashBuilder& hashBuilder,
+            SlangInt entryPointIndex) override;
+
+        virtual void updateASTBasedHash(
+            HashBuilder& hashBuilder) override { SLANG_UNUSED(hashBuilder); }
```
csyonghe, line 804: Implementation in its own line.

source/slang/slang.cpp
```diff
@@ -3481,16 +3457,39 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
-SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::computeDependencyBasedHash(slang::Checksum* outHashCode)
+SLANG_NO_THROW void SLANG_MCALL ComponentType::computeDependencyBasedHash(
+    SlangInt entryPointIndex,
+    SlangInt targetIndex,
+    slang::Hash* outHash)
 {
-    SLANG_UNUSED(outHashCode);
-    return SLANG_E_NOT_AVAILABLE;
+    HashBuilder builder;
+    builder.hashGen.init(&builder.context);
```
csyonghe, line 3466: again, this should just be in the constructor of `HashBuilder`.

source/slang/slang.cpp
```diff
@@ -3481,16 +3457,39 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
+    // 
+    // While enums are not guaranteed to be encoded the same way across all versions of
+    // the compiler, part of hashing the linkage is hashing in the compiler version.
+    // Consequently, any encoding differences as a result of different compiler versions
+    // will already be reflected in the resulting hash.
+    getLinkage()->computeDependencyBasedHash(builder, targetIndex);
+    updateDependencyBasedHash(builder, entryPointIndex);
+
+    slang::Hash hash;
+    builder.hashGen.finalize(&builder.context, &hash);
```
csyonghe, line 3478: wrap this as well.

source/slang/slang.cpp
```diff
@@ -3481,16 +3457,39 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
+    HashBuilder builder;
+    builder.hashGen.init(&builder.context);
+
+    // A note on enums that may be hashed in as part of the following two function calls:
+    // 
+    // While enums are not guaranteed to be encoded the same way across all versions of
+    // the compiler, part of hashing the linkage is hashing in the compiler version.
+    // Consequently, any encoding differences as a result of different compiler versions
+    // will already be reflected in the resulting hash.
+    getLinkage()->computeDependencyBasedHash(builder, targetIndex);
```
csyonghe, line 3474: should be `getLinkage()->updateDependencyBasedHash()`, this method in the linkage should just be internal. All internal methods are named `update*` according to the implicit convention here.

source/slang/slang.cpp
```diff
@@ -3481,16 +3457,39 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
+    *outHash = hash;
 }
 
-SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::computeASTBasedHash(slang::Checksum* outHashCode)
+SLANG_NO_THROW void SLANG_MCALL ComponentType::computeASTBasedHash(slang::Hash* outHash)
 {
-    SLANG_UNUSED(outHashCode);
-    return SLANG_E_NOT_AVAILABLE;
+    HashBuilder builder;
+    auto hashGen = builder.hashGen;
```
csyonghe, line 3485: use wrappers.

source/slang/slang.cpp
```diff
@@ -3825,46 +3824,44 @@ CompositeComponentType::CompositeComponentType(
...
+    // Add the name and name override for the specified entry point
+    // to the hash.
+    auto entryPointName = getEntryPoint(entryPointIndex)->getName()->text;
+    hashGen.update(context, entryPointName);
+    auto entryPointNameOverride = getEntryPointNameOverride(entryPointIndex);
+    hashGen.update(context, entryPointNameOverride);
 
-    *outHashCode = hashCode;
-    return SLANG_OK;
+    for (Index i = 0; i < componentCount; ++i)
```
csyonghe, line 3851: Actually, this loop may be the only thing that is needed in this method.

source/slang/slang.cpp
```diff
@@ -3481,16 +3457,39 @@ SLANG_NO_THROW SlangResult SLANG_MCALL ComponentType::getEntryPointCode(
...
+    builder.hashGen.init(&builder.context);
+
+    // A note on enums that may be hashed in as part of the following two function calls:
+    // 
+    // While enums are not guaranteed to be encoded the same way across all versions of
+    // the compiler, part of hashing the linkage is hashing in the compiler version.
+    // Consequently, any encoding differences as a result of different compiler versions
+    // will already be reflected in the resulting hash.
+    getLinkage()->computeDependencyBasedHash(builder, targetIndex);
+    updateDependencyBasedHash(builder, entryPointIndex);
```
csyonghe, line 3475: Should include all entry points here (most things in CompositeComponentType should just be moved here).

source/slang/slang.cpp
```diff
@@ -3825,46 +3824,44 @@ CompositeComponentType::CompositeComponentType(
...
+    auto context = &builder.context;
 
-    slang::Checksum hashCode;
-    MD5HashGen hashGen;
-    MD5Context context;
-    hashGen.init(&context);
+    auto componentCount = getChildComponentCount();
 
-    for (Index i = 0; i < componentCount; ++i)
+    // Add file path dependencies to the hash - all child components
```
csyonghe, line 3836: `fileDeps`, and entry point names should be hashed in `ComponentType`, not here.

slang.h
```diff
@@ -4134,10 +4134,11 @@ namespace slang
         None, UnsizedArray, StructuredBuffer, ConstantBuffer, ParameterBlock
     };
 
-    // A struct for storing checksums designed to be used with the MD5 hashing implementation
-    struct Checksum
+    // A struct storing a single hash represented as a four element uint32_t array.
+    // This is intended to be used with the current MD5 hashing implementation.
+    struct Hash
```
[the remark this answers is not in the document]
lucy96chen: Struct rename sure, but I think I'd rather keep the member name as-is since the entire array represents a single hash.

source/core/slang-digest.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include "slang-md5.h"
+#include "../../slang.h"
+
+namespace Slang
+{
+    // Wrapper struct that holds objects necessary for hashing.
+    struct DigestBuilder
+    {
+        MD5HashGen hashGen;
```
csyonghe, line 10: A minor improvement is to make these two things `private`.

source/slang/slang-hash-utils.h
```diff
@@ -0,0 +1,58 @@
...
+#include "../../slang.h"
+#include "../core/slang-basic.h"
+#include "../core/slang-md5.h"
+
+namespace Slang
+{
+    // Compute the hash for an UnownedStringSlice
+    inline slang::Digest computeHashForStringSlice(UnownedStringSlice text)
+    {
+        MD5Context context;
```
csyonghe, line 12: Let's use `DiggestBuilder` here.

source/slang/slang-hash-utils.h
```diff
@@ -0,0 +1,58 @@
...
+        slang::Digest textHash;
+        hashGen.finalize(&context, &textHash);
+
+        return textHash;
+    }
+
+    // Combines the two provided hashes to produce the final shader cache entry key.
+    inline slang::Digest combineHashes(const slang::Digest& linkageHash, const slang::Digest& programHash)
+    {
+        MD5Context context;
```
csyonghe, line 26: Let's use `DiggestBuilder` here.

csyonghe (commented):
Looks good.

source/core/slang-digest.h
```diff
@@ -25,5 +23,9 @@ namespace Slang
         {
```
csyonghe, line 23: You can do it in the next PR, but I think just write `slang::Digest finalize() {} ` would be more convenient.

Also since `Digest` is defined in `namespace slang`, I guess it doesn't hurt to have a `using slang::Digest;` in this header.

csyonghe (approved):

source/core/slang-digest.h
lucy96chen: Change will be in the new specialization test PR

PR #2435, opened 2022-10-06
Prevent out of bounds access by unaligned access in decodeLiteUInt32

sdawson-nvidia:
This prevents memory tools like gflags causing an exception on the out of bound access.

Given every value is at least one byte, only up to the 3rd-last value is safe, in the case of numBytesRemaining == 2 (the minimum), since it must be followed by at least 2 more bytes (1 for each remaining item).


jsmall-nvidia (approved):
Thanks for finding this issue and making this fix. 

On looking at the fix and other usages it seems like, the single decode function `/* static */int ByteEncodeUtil::decodeLiteUInt32(const uint8_t* in, uint32_t* out)` should probably be changed such that it doesn't have the unaligned path. As it could lead to an unaligned read into an out of bounds address.

PR #2439, opened 2022-10-10
Ignore d3d12 tests on x86.

csyonghe:


expipiplus1 (approved):
:+1:

worth opening an issue to remind us to re-enable these tests once dxc x86 gains 6.6 support.

PR #2441, opened 2022-10-11
Small IR cleanups.

csyonghe:
- Simplify one-time loops into normal branching.
- Don't emit `store(xxx. undefined)`.
- DCE after final Phi elimination pass.

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -830,6 +830,9 @@ Result linkAndOptimizeIR(
         }
     }
 
+    // Run a final round of DCE to clean up unused things after phi-elimination.
+    eliminateDeadCode(irModule);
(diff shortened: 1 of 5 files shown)


expipiplus1:
How does this affect optimization of the following:

```
let x = 0;
x <= undefined
if x == 0 then expensive;
```

Would it stop the expensive operation being elided?

expipiplus1 (approved):
:+1: assuming my missed optimization opportunity worry is unfounded.

csyonghe:
You are right, there are many uses of `undef` that can be optimized, and we should consider add them in.

The intention of this change is just that when we transform the code out of SSA, we never emit a `store(ptrToVar, undef)` since this is obviously unnecessary.

PR #2442, opened 2022-10-11
Allow setting shader model upon d3d device creation.

csyonghe:


expipiplus1 (dismissed):
Are there new intrinsics to add to hlsl.meta.slang? I guess in a subsequent change.

Probably ought to bump the profile version here (I'm assuming that this should be the maximal version) https://github.com/shader-slang/slang/blob/8bed58bd32fcea448f45cfa68d90bf5cf40c3055/source/slang/slang-workspace-version.cpp#L285

There were some places where 6.7 probably should have been mentioned, I've made those changes in [`expipiplus1/gfx_sm`](https://github.com/expipiplus1/slang/commit/681665aaf03d80dfa104ede8873d74207b6bb845) for you to pull.

...
It would be really nice to not have to touch so many places when adding a new version here...

Otherwise LGTM

PR #2443, opened 2022-10-12
Shader cache test for specialization args

lucy96chen:
Changes:
- Added a shader cache test for specialization args - This uses `compute-smoke.slang` and simply creates one of each transformer, binds it to the shader, and checks that changing the transformer type results in a cache miss.

The segment of the test for `CompositeTransformer` is currently commented out, something not quite right about how I'm binding to the fields `func0` and `func1`.

NOTE: THIS PR IS BUILT ON TOP OF THE CURRENTLY OPEN SHADER CACHE PR (2432). Once that PR is merged, I will be rebasing this PR to fix the commit history.


csyonghe (approved):

lucy96chen:
Something goofy happened with the rebase, didn't actually rebase away what should've been all but the commit containing the new test. Going to close this PR and redo the branch.

PR #2444, opened 2022-10-12
Add specialization args test

lucy96chen:
Changes:
- Added a shader cache test for specialization args - This uses compute-smoke.slang and simply creates an add and a mul transformer, binds a transformer to the shader, and checks that changing the transformer type results in a cache miss.
- Small leftover cleanup change in `slang-digest.h`

Redo of the specialization args test PR. Rebase was reintroducing lines that were removed and not producing the clean commit history I was looking for.

### DIFF
--- a/source/core/slang-digest.h
+++ b/source/core/slang-digest.h
@@ -2,10 +2,10 @@
 #include "slang-md5.h"
 #include "../../slang.h"
 
-using slang::Digest;
-
 namespace Slang
 {
+    using slang::Digest;
+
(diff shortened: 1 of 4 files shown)


source/core/slang-digest.h
```diff
@@ -2,6 +2,8 @@
 #include "slang-md5.h"
 #include "../../slang.h"
 
+using slang::Digest;
```
csyonghe, line 5: Move it to inside the `Slang` namespace.

csyonghe (approved):

PR #2446, opened 2022-10-12
Add getVersionString() to IDownstreamCompiler

lucy96chen:
Changes:
- Added `getVersionString()` to `IDownstreamCompiler` and provided implementations under `DownstreamCompilerBase`, `GlslangDownstreamCompiler`, and `DXCDownstreamCompiler` - Glslang and DXC now calculate and store the hashes for the contents of their respective DLLs which are returned by `getVersionString()` to be hashed in.
- Small change to `hashToString` to form the string one byte at a time, instead of four, to match existing MD5 output convention
- Updated `dxcapi.h`

I've manually inspected the values generated for `getVersionString()` calls for both dxc and glslang and confirmed their correctness. Also confirmed that the `ComPtr<ISlangBlob>` version of `update` calls into the correct `update` helper.


source/slang/slang-ast-type.cpp
```diff
@@ -77,6 +77,11 @@ Type* Type::getCanonicalType()
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! OverloadGroupType !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+void OverloadGroupType::_hashIntoOverride(DigestBuilder& builder)
+{
+    builder.addToDigest(_getHashCodeOverride());
```
csyonghe, line 82: Can't just use the existing hash code. This hash code has high chance of collision so they are not safe to use. This implementation should recurse down the same way as `equalsValOverride`.

source/slang/slang-ast-type.cpp
```diff
@@ -99,6 +104,11 @@ HashCode OverloadGroupType::_getHashCodeOverride()
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! InitializerListType !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+void InitializerListType::_hashIntoOverride(DigestBuilder& builder)
+{
+    builder.addToDigest(_getHashCodeOverride());
```
csyonghe, line 109: Cannot use existing hash code.

source/slang/slang-ast-type.cpp
```diff
@@ -121,6 +131,11 @@ HashCode InitializerListType::_getHashCodeOverride()
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! ErrorType !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+void ErrorType::_hashIntoOverride(DigestBuilder& builder)
+{
+    builder.addToDigest(_getHashCodeOverride());
```
csyonghe, line 136: no existing hash code

source/slang/slang-ast-type.cpp
```diff
@@ -150,6 +165,11 @@ HashCode ErrorType::_getHashCodeOverride()
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! BottomType !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+void BottomType::_hashIntoOverride(DigestBuilder& builder)
+{
+    builder.addToDigest(_getHashCodeOverride());
```
csyonghe, line 170: no existing hash code

csyonghe:
Please separate out the `hashInto` implementations from this PR. They are not correct. They can't just call into exsiting getHashCode since that hash code has high chance of collision. Need to implement the same recursive call structure like `equals` function.

source/compiler-core/slang-dxc-compiler.cpp
```diff
@@ -572,6 +573,20 @@ SlangResult DXCDownstreamCompiler::convert(IArtifact* from, const ArtifactDesc&
     return SLANG_OK;
 }
 
+SlangResult DXCDownstreamCompiler::getVersionString(slang::IBlob** outVersionString)
+{
+    if (m_desc.hasVersion())
+    {
+        auto version = m_desc.getVersionValue();
```
csyonghe, line 580: The version field in `m_desc` is never filled in. 
The right way to get version is to create an instance of dxc by calling `m_createInstance`, then query the `IDxcVersionInfo` interface on the dxc instance, this interface provides a `GetVersion` method that returns the verison number.

A better way is just to hash the dxcompiler.dll and use that digest as the version string. This is more robust against out-of-band builds of dxc.

source/compiler-core/slang-glslang-compiler.cpp
```diff
@@ -277,6 +278,20 @@ SlangResult GlslangDownstreamCompiler::convert(IArtifact* from, const ArtifactDe
     return SLANG_OK;
 }
 
+SlangResult GlslangDownstreamCompiler::getVersionString(slang::IBlob** outVersionString)
+{
+    if (m_desc.hasVersion())
+    {
+        auto version = m_desc.getVersionValue();
+        ComPtr<slang::IBlob> versionBlob = Slang::StringUtil::createStringBlob(Slang::String(version));
```
csyonghe, line 286: Again, can't use `m_desc`. Just hash the downstream dll.

csyonghe (commented):
Please remove the hashInto functions. We also need to test this. Just check the hash is not zero.

jsmall-nvidia:
I'm not sure about this as an approach, in particular...

* Hashing of a dll/shared library
* Not using the version the downstream compiler identifies itself as (in the desc)

Why?

* Hashing a shared library seems like overkill and somewhat inappropriate
  * Are we relying on a *shader cache* to be able to have read access the bytes of some other binary to work?
  * There are quite a lot of bytes
  * They may not even be all the bytes - it may (and dxc does) load other libraries
* The downstream compiler should 'know' what it is 
  * On DXC there are interfaces that can return the version number and even a hash (we don't currently do this)
  * On glsl we build it (!) - could just include the github/gitlab hash
* The name of the method should probably be getBuildHash, and arguably optional 
  * The name seems to imply something to human readable - when really it's something that's optionally tacked on to the Desc 'version' string
* Seems tailored to a specific kind of downstream compiler
  * Doesn't make much sense if the dll is statically linked.
  * What does it mean for say C++ downstream compilers? Are you going to hash visual studio or gcc?
  * What does it mean for a shared library compiler that uses other shared libraries?
  
On the last point - since it is a method on the downstream compiler interface other implementations can of course implement it in some other way (ie not a hash of the binary). I just don't think we want our implementations to rely on binary file hashing.

So I think the approach should perhaps be more along the lines of... 

* The hash would work off the desc primarily
* The desc quality should be improved in downstream compiler impls 
* There is an optionally implemented function to return a 'hash'. 

I'd add that this change breaks binary compatibility with `slang-llvm`. That may not be apparent because CPU binaries are not being held in the cache as currently tested. But the issue is there.

I suppose I also wonder if a change that is not represented within a version number change really warrants recompiling everything by default. You might want that, but you can always (presumably) force that, if you are going around changing libraries that report themselves as the same but are not the same in a way you care about.

csyonghe:
@jsmall-nvidia Hashing dxcompiler.dll is the only way I can think of that is robust against out-of-band dxc builds. We have users depending on them, and different out-of-band dxc will report the same version number through their `IDxcVersionInfo` interface. So unless you have better ideas, I'd say we use the file hash implementation at least for DXC.

I agree that we can just modify slang-glslang to report the git tag as version number.

For gcc/clang and msvc we can just run `--version`, but it is not implemented here.

For slang-llvm we should also report its git tag in our custom build.

jsmall-nvidia:
There is IDxcVersionInfo2 that does report a build hash. I don't know how well supported/how well that works in practice. 
There is a flag that DXC returns that indicates 'internal' build. Again I don't know if that is used appropriately. It it reports as an official release, then presumably the version is enough.

We do support looking up the compiler versions for downstream C++ compilers. 

It's not clear to me the scenario/s that leads to requiring a hash of the dxc binary. At least in scenarios where this is a runtime/local cache.

source/core/slang-md5.h
```diff
@@ -70,6 +70,8 @@ namespace Slang
         void update(MD5Context* ctx, String str);
```
csyonghe, line 70: I would agree with @jsmall-nvidia that we should rename the templated `update` to something else, like `updateSimpleVal` to prevent it from being picked up by a call to `update` with a complex type, e.g. `ComPtr<ISlangBlob>` unintentionally.

source/slang/slang-hash-utils.h
```diff
@@ -29,11 +29,13 @@ namespace Slang
...
-        for (Index i = 0; i < 4; ++i)
+        uint8_t* uint8Hash = (uint8_t*)hash.values;
+
+        for (Index i = 0; i < 16; ++i)
         {
-            auto hashSegmentString = String(hash.values[i], 16);
+            auto hashSegmentString = String(uint8Hash[i], 16);
 
-            auto leadingZeroCount = 8 - hashSegmentString.getLength();
+            auto leadingZeroCount = 2 - hashSegmentString.getLength();
```
csyonghe, line 38: Since the loop is only going to be executed at most once, we should just turn it into an `if`.

source/core/slang-md5.h
lucy96chen: Have enable_if guarding the general one on line 57 (for enums/arithmetic values), the second template is for Lists which shouldn't get picked up by anything that isn't a List.

source/compiler-core/slang-dxc-compiler.cpp
```diff
@@ -572,6 +575,57 @@ SlangResult DXCDownstreamCompiler::convert(IArtifact* from, const ArtifactDesc&
...
+            *outVersionString = version.detach();
+            return SLANG_OK;
+        }
+    }
+
+    // If either of the QueryInterface calls fails, we return the shared library timestamp
+    // as the version instead.
+    auto timestamp = SharedLibraryUtils::getSharedLibraryTimestamp(m_createInstance);
+    auto timestampString = String(timestamp);
+    version = RawBlob::create(timestampString.getBuffer(), timestampString.getLength());
```
csyonghe, line 624: Can just use `StringBlob::create`

source/slang/slang.cpp
```diff
@@ -1367,6 +1367,18 @@ void Linkage::updateDependencyBasedHash(
...
+
+    // Add the downstream compiler version (if it exists) to the hash
+    auto passThroughMode = getDownstreamCompilerRequiredForTarget(targetReq->getTarget());
+    auto downstreamCompiler = getSessionImpl()->getOrLoadDownstreamCompiler(passThroughMode, nullptr);
+    if (downstreamCompiler)
+    {
+        ComPtr<ISlangBlob> versionString;
+        if (SLANG_SUCCEEDED(downstreamCompiler->getVersionString(versionString.writeRef())))
+        {
+            builder.addToDigest(versionString);
```
csyonghe, line 1379: Can you verify the version/hash value looks correct by setting a breakpoint and manually inspect the value here? Also double check that this call doesn't calls into the templated version of `update`.

csyonghe (approved):
Assuming you have verified the version number and hash returned from dxc is correct, this looks good to me.

source/compiler-core/slang-dxc-compiler.cpp
```diff
@@ -611,6 +598,14 @@ SlangResult DXCDownstreamCompiler::getVersionString(slang::IBlob** outVersionStr
             versionString.append(major);
             versionString.append(".");
             versionString.append(minor);
+
+            char* commitHash;
```
csyonghe, line 602: It is safer to initialize this to nullptr, in case `getCommitInfo` call failed.

csyonghe (approved):
Looks good.

source/compiler-core/slang-dxc-compiler.cpp
lucy96chen: I'll include this change with the shader cache index PR

PR #2451, opened 2022-10-13
Allow multi-level breaks to break out of `switch` statements.

csyonghe:
Generalizes the mult-level break pass to work on `switch` statements in addition to loops.

This completes multi-break and general inlining support, so we are exposing a `[ForceInline]` attribute that normal users can use.


csyonghe:
Merging now in order to create a release.

PR #2452, opened 2022-10-14
Shader cache index implementation

lucy96chen:
Changes:
- Renamed `ShaderCacheIndex` to `PersistentShaderCache` and added `persistent-shader-cache.cpp` containing implementations
- Renamed and moved some files to be under more appropriate projects


source/core/slang-linked-list.h
```diff
@@ -253,6 +253,30 @@ template <typename T> class LinkedList
         count++;
         return n;
     };
+    void MoveToFirst(LinkedNode<T>* n)
```
csyonghe, line 256: It is prbably more useful to decompose into RemoveFromList + AddFirst.

source/slang/slang-hash-utils.h
```diff
@@ -45,5 +47,8 @@ namespace Slang
         return filename;
     }
 
-    // TODO: fromString implementation?
+    inline Digest stringToHash(String hashString)
```
csyonghe, line 50: This is worth unit-testing as well.

source/slang/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,123 @@
...
+        if (mutableShaderCacheFileSystem)
+        {
+            SLANG_RETURN_ON_FAIL(mutableShaderCacheFileSystem->saveFile(filename.getBuffer(), nullptr, 0));
+            indexFilename = filename;
+            return SLANG_OK;
+        }
+    }
+
+    String indexString;
+    File::readAllText(filename, indexString);
```
csyonghe, line 26: You can use `StringUtil::calcLines` to get your a `List<UnownedStringSlice> lines` where each element represents a line.

Then you can just iterate through the lines:
```
for (auto line : lines)
{
    List<UnownedStringSlice> words;
    StringUtil::split(line, ' ', words); // words will contain two elements, one for each hash.
    auto dependencyDiggest = stringToDigest(words[0]);
    auto astDigest = stringToDigest(words[1]);
    ...
}
```

source/slang/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,123 @@
...
+    // the list.
+    entries.MoveToFirst(entryNode);
+    return &entryNode->Value;
+}
+
+void ShaderCacheIndex::addEntry(const slang::Digest& dependencyDigest, const slang::Digest& astDigest, ISlangBlob* compiledCode)
+{
+    // Check that we do not exceed the cache's size limit by adding another entry. If so,
+    // remove the least recently used entry first.
+    if (entries.Count() == entryCountLimit)
```
csyonghe, line 60: It is better to write
`while (entryCountLimit >0 && entries.Count() >= entryCountLimit)`

source/slang/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    saveCacheIndexToFile();
+}
+
+void ShaderCacheIndex::updateEntry(const slang::Digest& dependencyDigest, const slang::Digest& astDigest, ISlangBlob* updatedCode)
+{
+    LinkedNode<ShaderCacheEntry>* entryNode;
+    keyToEntry.TryGetValue(dependencyDigest, entryNode);
+    entryNode->Value.astBasedDigest = astDigest;
+
```
csyonghe, line 82: delete the old file if astDigest is different.

source/slang/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,123 @@
...
+        mutableShaderCacheFileSystem->saveFileBlob(hashToString(dependencyDigest).getBuffer(), compiledCode);
+    }
+
+    saveCacheIndexToFile();
+}
+
+void ShaderCacheIndex::updateEntry(const slang::Digest& dependencyDigest, const slang::Digest& astDigest, ISlangBlob* updatedCode)
+{
+    LinkedNode<ShaderCacheEntry>* entryNode;
+    keyToEntry.TryGetValue(dependencyDigest, entryNode);
```
csyonghe, line 80: What if the entry does not exist in `keyToEntry`? wrap the following code in a `if`.
```
if (keyToEntry.TryGetValue(dependencyDigest, entryNode))
{
...
}
```

csyonghe: Can just call `findEntry` to find it.

source/core/slang-linked-list.h
```diff
@@ -253,6 +253,30 @@ template <typename T> class LinkedList
         count++;
         return n;
     };
+    void AddFirst(LinkedNode<T>* n)
```
csyonghe, line 256: We should just change `AddFirst(const T& nData)` to call this instead.

source/slang/slang-shader-cache-index.h
```diff
@@ -1,60 +1,75 @@
...
+    // Dictionary mapping each shader's key to its corresponding node (entry) in the list
+    // of entries.
+    Dictionary<slang::Digest, LinkedNode<ShaderCacheEntry>*> keyToEntry;
+
+    // Linked list containing the entries stored in the shader cache in order
+    // of most to least recently used.
+    LinkedList<ShaderCacheEntry> entries;
+
+    // The underlying file system used for the shader cache.
+    ISlangFileSystem* shaderCacheFileSystem = nullptr;
```
csyonghe, line 65: This should be a ComPtr.

tools/slang-unit-test/unit-test-digest-utils.cpp
```diff
@@ -29,4 +29,24 @@ SLANG_UNIT_TEST(checksum)
         String testCString = hashToString(testC);
         SLANG_CHECK(testCString.equals(String("11111111222222223333333344444444")));
     }
+
+    {
+        String digestString = String("5D6CC58E1824A4DFD0CF57395B603316");
+        slang::Digest digest = stringToHash(digestString);
+
+        SLANG_CHECK(digest.values[0] == 2395303005);
```
csyonghe, line 37: Use hex constants here to make it more clear.

source/slang/slang-shader-cache-index.h
```diff
@@ -1,60 +1,75 @@
...
+{
+    slang::Digest dependencyBasedDigest;
+    slang::Digest astBasedDigest;
+};
+
+class ShaderCacheIndex
+{
+public:
+
+    ShaderCacheIndex(SlangInt size, const char* path, ISlangFileSystem* fileSystem);
```
csyonghe, line 23: `String path` or `UnownedStringSlice path`. We don't use raw pointers in internal implementation.

source/slang/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "../core/slang-file-system.h"
+
+namespace Slang
+{
+
+ShaderCacheIndex::ShaderCacheIndex(SlangInt size, const char* path, ISlangFileSystem* fileSystem)
+{
+    entryCountLimit = size;
+    shaderCacheFileSystem = fileSystem;
+
```
csyonghe, line 16: Did not initialize `indexFileName`. Not sure if you plan to use it.

source/slang/slang-shader-cache-index.h
```diff
@@ -1,60 +1,75 @@
...
+
+    // Load a previous cache index saved to disk. If not found, create a new cache index
+    // and save it to disk as filename.
+    SlangResult loadCacheIndexFromFile(String filename);
```
csyonghe, line 27: This does not need to be a public method. Just load the file upon constructing. It probably does not need `filename` parameter either. Just use `indexFileName`.

tools/slang-unit-test/unit-test-digest-utils.cpp
```diff
@@ -29,4 +29,24 @@ SLANG_UNIT_TEST(checksum)
...
+        SLANG_CHECK(digest.values[1] == 3752076312);
+        SLANG_CHECK(digest.values[2] == 962056144);
+        SLANG_CHECK(digest.values[3] == 372465755);
```
csyonghe, line 40: It is better to just convert digest back to string here and check the string is equal to `digestString`.

source/slang/slang-hash-utils.h
```diff
@@ -45,5 +47,41 @@ namespace Slang
...
+    {
+        uint8_t uint8Hash[16];
+
+        // When the hash is converted to a String, ReverseInternalAscii is called
+        // at the very end. Since there is no way to get a char* for hashString to pass
+        // to ReverseInternalAscii to flip the string back, we instead loop starting from
+        // the end and work backwards towards the beginning.
+        for (Index i = 15; i >= 0; --i)
```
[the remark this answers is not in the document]
lucy96chen: I need to go see what other utility classes are floating around. This is probably the same idea, I just reverse-engineered the `IntToASCII` function in `slang-string.h`

source/slang/slang-shader-cache-index.cpp
lucy96chen: indexFilename is assigned in `loadCacheIndexFromFile`, not initializing it in the constructor is by design.

source/slang/slang-shader-cache-index.h
lucy96chen: But if we're going to just use a set filename for now instead of allowing users to pass one in, then sure, I can do that and initialize indexFilename in the constructor

lucy96chen:
Noting this here: There is likely a bug already in `loadCacheIndexFromFile` as it uses `AddFirst` and not `AddLast`, even though we output the index in the order of the linked list. This means the index is being loaded in reverse order currently. This is fixed in the PR containing cache index integration.

source/core/slang-linked-list.h
```diff
@@ -243,16 +243,34 @@ template <typename T> class LinkedList
...
+        else
+            head = n2;
+        if (n2)
+            n2->prev = n1;
+        else
+            tail = n1;
+    }
     void Delete(LinkedNode<T>* n, int Count = 1)
     {
         LinkedNode<T>*n1, *n2 = 0, *tn;
```
csyonghe, line 276: Call `RemoveFromFirst` in `Delete`.

source/slang/slang-hash-utils.h
```diff
@@ -4,19 +4,22 @@
...
 #include "../core/slang-digest.h"
+#include "../core/slang-char-util.h"
 
 namespace Slang
 {
+    using slang::Digest;
+
     // Compute the hash for an UnownedStringSlice
-    inline slang::Digest computeHashForStringSlice(UnownedStringSlice text)
+    inline Digest computeHashForStringSlice(UnownedStringSlice text)
```
csyonghe, line 14: `computeDigestForStringSlice`

source/slang/slang-hash-utils.h
```diff
@@ -25,9 +28,9 @@ namespace Slang
     }
 
     // Returns the stored hash in checksum as a String.
-    inline StringBuilder hashToString(const slang::Digest& hash)
+    inline String hashToString(const Digest& hash)
```
csyonghe, line 31: `digestToString`

source/slang/slang-hash-utils.h
```diff
@@ -37,13 +40,30 @@ namespace Slang
...
-            filename.append(hashSegmentString.getBuffer());
+            hashString.append(hashSegmentString.getBuffer());
         }
 
-        return filename;
+        return hashString;
     }
 
-    // TODO: fromString implementation?
+    inline Digest stringToHash(UnownedStringSlice hashString)
```
csyonghe, line 51: stringToDigest

source/slang/slang-hash-utils.h
csyonghe: I think we should just have a `DigestUtil` class here and move all these functions into it as static members. That way we can clean the naming up a lot, similar to what `StringUtil` does.
```
class DigestUtil
{
     static Digest fromString(UnownedStringSlice str);
     static String toString(const Digest& digest);
     static Digest combine(const Digest& d0, const Digest& d1);
     static Digest computeDigestForString(UnowendStringSlice text);
}

```

source/core/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,173 @@
...
+        // The key was not found in the cache, so we return nullptr.
+        *outCompiledCode = nullptr;
+        return nullptr;
+    }
+
+    // If the key is found, we need to move its corresponding entry to the front of
+    // the list and load the stored contents from disk.
+    entries.RemoveFromList(entryNode);
+    entries.AddFirst(entryNode);
+    shaderCacheFileSystem->loadFile(hashToString(key).getBuffer(), outCompiledCode);
```
csyonghe, line 88: need to call `saveCacheIndexToFile()` here. We can detect if `entryNode` is already in the front of the list, and if so skip the saving step.

source/core/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (!mutableShaderCacheFileSystem)
+    {
+        // Cannot save the index to disk if the underlying file system isn't mutable.
+        return;
+    }
+
+    StringBuilder indexString;
+    for (auto& entry : entries)
+    {
+        indexString.append(hashToString(entry.dependencyBasedDigest));
```
csyonghe, line 146: To speed things up we should probably have an operator override in `digest-util` for `StringBuilder`:
```
StringBuilder& operator <<(StringBuilder& sb, const Digest& d);
```
Then just call `indexSb << entry.dependencyBasedDigest;` here to avoid the allocation involved in creating the temporary `String`s.

source/core/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+void ShaderCacheIndex::saveCacheIndexToFile()
+{
+    if (!mutableShaderCacheFileSystem)
+    {
+        // Cannot save the index to disk if the underlying file system isn't mutable.
+        return;
+    }
+
+    StringBuilder indexString;
```
csyonghe, line 143: `indexSb`.

source/core/slang-shader-cache-index.h
```diff
@@ -0,0 +1,74 @@
...
+struct ShaderCacheEntry
+{
+    slang::Digest dependencyBasedDigest;
+    slang::Digest astBasedDigest;
+};
+
+class ShaderCacheIndex
+{
+public:
+    ShaderCacheIndex(SlangInt size, String path, ISlangFileSystem* fileSystem, String filename);
```
csyonghe, line 22: Use more descriptive parameter names. `size`->`entryCountLimit`, `filename`->`indexFileName`.
You can wrap these parameters in a `struct ShaderCacheIndexDesc`, and make it the only parameter of the constructor. The class can hold an member of `ShaderCacheIndexDesc desc` so you don't have to assign them manually one by one.

source/core/slang-shader-cache-index.h
```diff
@@ -0,0 +1,74 @@
+// slang-shader-cache-index.h
```
csyonghe, line 1: Why is this in core? It should be in `gfx` for now. `core` is for basic C++ classes like `List`, `RefPtr` etc. and not for any actual compiler/gfx logic.

source/slang/slang-hash-utils.h
```diff
@@ -4,19 +4,22 @@
 #include "../core/slang-basic.h"
 #include "../core/slang-md5.h"
 #include "../core/slang-digest.h"
+#include "../core/slang-char-util.h"
```
csyonghe, line 7: `slang-hash-util` should be in `core`. Also rename it to `slang-digest-util.h`

source/slang/slang-hash-utils.h
lucy96chen: Changed to be a static class using the method naming scheme above

source/core/slang-shader-cache-index.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "../slang/slang-hash-utils.h"
+#include "slang-io.h"
+#include "slang-string-util.h"
+#include "slang-file-system.h"
+
+namespace Slang
+{
+
+ShaderCacheIndex::ShaderCacheIndex(SlangInt size, String path, ISlangFileSystem* fileSystem, String filename)
+{
```
csyonghe, line 13: Just rename this class to PersistentShaderCache since it also handle the actual shader code.

source/core/slang-shader-cache-index.h
csyonghe: There is already a shader cache Desc in slang.h so we can just reuse that.

lucy96chen: I'll change this again in the integration PR where the ShaderCacheDesc struct is introduced

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -0,0 +1,180 @@
...
+    LinkedNode<ShaderCacheEntry>* entryNode;
+    if (!keyToEntry.TryGetValue(key, entryNode))
+    {
+        // The key was not found in the cache, so we return nullptr.
+        *outCompiledCode = nullptr;
+        return nullptr;
+    }
+
+    // If the key is found, we need to move its corresponding entry to the front of
+    // the list and load the stored contents from disk.
```
csyonghe, line 85: comment should be "save the index to disk".

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+    if (!mutableShaderCacheFileSystem)
+    {
+        // Should not save new entries if the underlying file system isn't mutable.
+        return;
+    }
+    
+    // Check that we do not exceed the cache's size limit by adding another entry. If so,
+    // remove the least recently used entry first.
+    while (entryCountLimit > 0 && entries.Count() >= entryCountLimit)
```
csyonghe, line 109: Need to embed this while loop in an `if (!mutableShaderCacheFileSystem)` otherwise you could risk entering infinite loop here. When writing a `while`, always think "could this lead to infinite loop under some situations?"

csyonghe: Ah, nevermind, I see you have an early return above, everything is good then.

source/core/slang-digest-util.h
```diff
@@ -0,0 +1,35 @@
...
+        // Returns the hash represented by hashString as a Digest.
+        static Digest fromString(UnownedStringSlice hashString);
+    };
+
+    inline StringBuilder& operator<<(StringBuilder& sb, const Digest& d)
+    {
+        for (Index i = 0; i < 4; ++i)
+        {
+            char temp[8];
+            IntToAscii(temp, d.values[i], 16);
```
csyonghe, line 30: You will need to call `sb.append(uint8Values[i[)`, otherwise it is going to produce results in a different endian order.

csyonghe: Probably need a test for this in the similar way as `toString` and `fromString`, or you can simply change the implementation of `toString` to call this.

csyonghe: Yeah, the implementation here should just be exactly what is existing in `toString`, and then make `toString` call this.

tools/gfx/persistent-shader-cache.cpp
lucy96chen: I don't think this will ever occur because of the `if (!mutableShaderCacheFileSystem)` check at the very start. The only way this infinitely loops without crashing is if `deleteLRUEntry` immediately returns, but that only occurs when the file system isn't mutable. We should never make it to this loop if that's the case.

source/core/slang-digest-util.h
csyonghe: I mean, if you make toString call `operator <<` then you don't need a dedicated test for it.

source/core/slang-digest-util.h
```diff
@@ -24,10 +24,20 @@ namespace Slang
...
-            char temp[8];
-            IntToAscii(temp, d.values[i], 16);
+            char temp[2];
+            int hashSegment = (int)uint8Hash[i];
+            // Check if we need to append a leading zero.
+            if (hashSegment < 16)
+            {
+                sb << "0";
+            }
+            IntToAscii(temp, hashSegment, 16);
```
csyonghe, line 40: just `sb.append(hashSegment, 16)`, no need to use `temp` and `IntToAscii`.

source/core/slang-digest-util.h
```diff
@@ -24,10 +24,20 @@ namespace Slang
...
+        // Must cast to uint8_t* first in order to correctly account for
+        // endianness.
+        uint8_t* uint8Hash = (uint8_t*)d.values;
+
+        for (Index i = 0; i < 16; ++i)
         {
-            char temp[8];
-            IntToAscii(temp, d.values[i], 16);
+            char temp[2];
+            int hashSegment = (int)uint8Hash[i];
```
csyonghe, line 34: The type cast isn't necessary here. just `int hashSegment = uint8Hash[i]`.

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -13,7 +13,7 @@ PersistentShaderCache::PersistentShaderCache(const Desc& desc)
 {
     entryCountLimit = desc.entryCountLimit;
```
csyonghe, line 14: Just make `desc` a member and then initialize that member directly with the parameter `desc` (rename the parameter to `inDesc` here to avoid name conflicts).

source/core/slang-digest-util.h
csyonghe: IntToAscii gives you the result in reverse order. For example IntToAscii(16) gives you "61", and you need to reverse the bytes to get the correct result. Just call `sb.append` which already does this for you.

tools/gfx/persistent-shader-cache.cpp
lucy96chen: Will move the default values all the way up to the ShaderCacheDesc in slang-gfx.h in the other PR and have them be passed in from there.

source/core/slang-linked-list.h
```diff
@@ -276,6 +270,22 @@ template <typename T> class LinkedList
...
             tail = n1;
+    }
+    void Delete(LinkedNode<T>* n, int Count = 1)
+    {
+        LinkedNode<T>*cur, *next;
+        cur = n;
+        int numDeleted = 0;
+        for (int i = 0; i < Count; i++)
+        {
+            RemoveFromList(cur);
```
csyonghe, line 281: This and line 282 should swap in order.

source/core/slang-linked-list.h
```diff
@@ -276,6 +270,22 @@ template <typename T> class LinkedList
             n2->prev = n1;
         else
             tail = n1;
```
csyonghe, line 272: should set both n->prev and n->next to null since it is already removed from the list.

csyonghe (commented):
Looks good other than the comment on LinkedList.

lucy96chen:
I just realized this should also be moved under the gfx namespace from Slang. If there's nothing else to change for this PR, it'll be included in the next push to the integration PR (2453).

csyonghe (approved):

PR #2453, opened 2022-10-15
PersistentShaderCache integration

lucy96chen:
Changes:
- Integrates `PersistentShaderCache` into `RendererBase`, replacing much of the old shader caching code located in `RendererBase::getEntryPointCodeFromShaderCache` with the new `PersistentShaderCache` methods (which are responsible for the functionality of the old code)
- Adds a test that enables cache eviction - This performs both the same hit/miss/dirty checks that all existing shader cache tests perform as well as a check on the contents of the cache file itself to confirm that entries are correctly ordered and contain the correct hashes.
- .gitignore and premake5.lua updated to ignore the shader files auto-generated by several of the shader cache tests - Users should never need to touch these files directly as the code piped into them is located in the shader test file itself.


tools/gfx/renderer-shared.cpp
```diff
@@ -365,33 +362,25 @@ Result RendererBase::getEntryPointCodeFromShaderCache(
...
     ComPtr<ISlangBlob> codeBlob;
 
-    // Query shaderCacheFileSystem for an entry whose key matches shaderFilename
-    //    - If we find it, then copy the file contents into memory and return in outCode
-    auto result = shaderCacheFileSystem->loadFile(shaderKey.getBuffer(), codeBlob.writeRef());
-    
-    if (SLANG_FAILED(result))
+    // Query the shader cache index for an entry with shaderKeyHash as its key. 
+    auto entry = cacheIndex->findEntry(shaderKey, codeBlob.writeRef());
+    if (!entry)
```
csyonghe, line 377: Restructure this so you only call getEntryPointCode once from this point.

tools/gfx/renderer-shared.cpp
```diff
@@ -403,22 +392,16 @@ Result RendererBase::getEntryPointCodeFromShaderCache(
...
-        // If the entry exists, we need to check that the entry isn't effectively dirty. Since we stored
-        // the AST hash with the compiled code, we can determine this by comparing the stored hash with the
-        // AST hash generated earlier.
-        auto entryContents = codeBlob->getBufferPointer();
-        auto hashSize = sizeof(slang::Digest);
-        if (memcmp(ASTHash.values, entryContents, hashSize) != 0)
+        // If the entry exists, we need to check that the entry hasn't changed changed since it was last accessed.
+        // Since we also stored the AST hash in the shader cache index, we can determine this by comparing the stored hash
+        // with the AST hash we generated earlier.
+        if (contentsHash == entry->Value.astBasedDigest)
```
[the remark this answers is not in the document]
lucy96chen: yes, should be !=, woops

tools/gfx/renderer-shared.h
```diff
@@ -1372,8 +1374,7 @@ class RendererBase : public IDevice, public IShaderCacheStatistics, public Slang
     SlangContext slangContext;
     ShaderCache shaderCache;
 
-    ISlangFileSystem* shaderCacheFileSystem = nullptr;
-    ComPtr<ISlangMutableFileSystem> mutableShaderCacheFileSystem = nullptr;
+    Slang::ShaderCacheIndex* cacheIndex = nullptr;
```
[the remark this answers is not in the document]
lucy96chen: changed locally, will be in next push

tools/gfx/renderer-shared.cpp
lucy96chen: Changed locally, will be in next push

slang-gfx.h
```diff
@@ -2153,11 +2166,8 @@ class IDevice : public ISlangUnknown
         ISlangUnknown* apiCommandDispatcher = nullptr;
         // The slot (typically UAV) used to identify NVAPI intrinsics. If >=0 NVAPI is required.
         GfxIndex nvapiExtnSlot = -1;
-        // The root directory for the shader cache.
-        const char* shaderCachePath = nullptr;
-        // The file system for loading cached shader kernels. The layer does not maintain a strong reference to the object,
-        // instead the user is responsible for holding the object alive during the lifetime of an `IDevice`.
-        ISlangFileSystem* shaderCacheFileSystem = nullptr;
+        // Configurations for the shader cache.
```
csyonghe, line 2169: update gfx.slang as well.

lucy96chen: Changed locally, will be in next push

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -39,25 +40,26 @@ PersistentShaderCache::PersistentShaderCache(const Desc& inDesc)
...
+            mutableShaderCacheFileSystem->saveFile(desc.cacheFilename, nullptr, 0);
             return;
         }
         // Cache index not found and we can't save a new one due to the file system being immutable.
         return;
     }
 
-    String indexString;
-    File::readAllText(desc.cacheFilename, indexString);
+    String indexString = String((char*)indexBlob->getBufferPointer());
```
csyonghe, line 55: Use `UnownedStringSlice` to avoid copy.

tools/gfx/renderer-shared.h
```diff
@@ -1372,10 +1374,7 @@ class RendererBase : public IDevice, public IShaderCacheStatistics, public Slang
     SlangContext slangContext;
     ShaderCache shaderCache;
 
-    // TODO: These should be removed when ShaderCacheIndex is ready to be integrated. ShaderCacheIndex
-    // will be responsible for keeping track of the underlying filesystem for the cache.
-    ISlangFileSystem* shaderCacheFileSystem = nullptr;
-    ComPtr<ISlangMutableFileSystem> mutableShaderCacheFileSystem = nullptr;
+    RefPtr<PersistentShaderCache> cacheIndex = nullptr;
```
csyonghe, line 1377: Just name it to `persistentShaderCache`.

tools/gfx-unit-test/shader-cache-tests.cpp
```diff
@@ -689,6 +699,148 @@ namespace gfx_test
...
+    struct CacheWithMaxEntryLimit : MultipleEntryShaderCache
+    {
+        // Check the correctness of the cache's entries by comparing the order of entries in the
+        // current state of the cache with what we expect.
+        void checkCacheFile(GfxIndex testIndex)
+        {
+            ComPtr<ISlangBlob> cacheContents;
+            cacheFileSystem->loadFile(shaderCache.cacheFilename, cacheContents.writeRef());
+
+            String apiString;
```
csyonghe, line 713: Should move this into a util function. Or can just use `RenderApiUtil::getApiName()`.
The function should return `UnowendStringSlice` instead of `String` to avoid copy.

tools/gfx-unit-test/shader-cache-tests.cpp
```diff
@@ -434,7 +444,7 @@ namespace gfx_test
     //    4. #include w/ changes in the included file
     struct ShaderFileImportsShaderCache : BaseShaderCacheTest
     {
-        Slang::String importedContentsA = Slang::String(
+        String importedContentsA = String(
             R"(struct TestFunction
```
csyonghe, line 448: If you insert a line break after "(" the indentation in the shader string will be consistent.

tools/gfx-unit-test/cache-eviction-policy-expected-d3d12-1.txt
```diff
@@ -0,0 +1,2 @@
+DA799C6427056E9E291844A469A7049D 5C5B27A848BE6D2F974B65DFD3E548E9
```
csyonghe, line 1: Can we just define this directly as constants in test code? That way we don't need these separate files.

tools/gfx/gfx.slang
```diff
@@ -1710,6 +1710,19 @@ struct SlangDesc
     slang::SlangLineDirectiveMode lineDirectiveMode = slang::SlangLineDirectiveMode::SLANG_LINE_DIRECTIVE_MODE_DEFAULT;
 };
 
+struct ShaderCacheDesc
+{
+    // The filename for the file the cache's state should be saved to or loaded from.
+    NativeString cacheFilename = "cache.txt";
```
csyonghe, line 1716: Default should be nullptr, so no caching is enabled.

tools/gfx/gfx.slang
```diff
@@ -1710,6 +1710,19 @@ struct SlangDesc
...
+    // The root directory for the shader cache.
+    NativeString shaderCachePath = nullptr;
+    // The file system for loading cached shader kernels. The layer does not maintain a strong reference to the object,
+    // instead the user is responsible for holding the object alive during the lifetime of an `IDevice`.
+    void* shaderCacheFileSystem = nullptr;
+    // The maximum number of entries stored in the cache.
+    GfxCount entryCountLimit = 1000;
```
csyonghe, line 1723: Now that I think about it, a 0 default here might be more appropriate, since caching is off by default.

tools/gfx-unit-test/cache-eviction-policy-expected-d3d12-1.txt
lucy96chen: Yeah I can do this the same way as the shader code since Git's auto-convert of LF to CRLF is likely breaking the `memcmp()` right now.

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -39,25 +40,26 @@ PersistentShaderCache::PersistentShaderCache(const Desc& inDesc)
...
 
     List<UnownedStringSlice> lines;
     StringUtil::calcLines(indexString.getUnownedSlice(), lines);
     for (auto line : lines)
     {
+        if (line == "")
```
csyonghe, line 61: I don't think we need this. Just `continue` if size of `digests` after `split` is not 2.

tools/gfx/gfx.slang
lucy96chen: The check to enable caching in `RendererBase::initialize()` checks that either `shaderCacheFileSystem` or `shaderCachePath` is non-nullptr, which they are by default, so `cacheFilename`'s default value shouldn't matter here.

tools/gfx-unit-test/cache-eviction-policy-expected-d3d12-1.txt
csyonghe: The problem with this kind of baked in digest is that different downstream compiler versions on different systems will produce different digest. Instead of hard coding, we need a better way to tell if the eviction policy is correct that adapts to different digests.

tools/gfx/gfx.slang
lucy96chen: Okay, I suppose this (and the comment above) also makes sense from a consistency perspective. Are we expecting the user to provide both a filename and entry count limit then if they're choosing to turn caching on?

csyonghe: A 0 limit should mean no limit, which is also valid. The user should specify a limit if they wish to limit.

tools/gfx/persistent-shader-cache.cpp
lucy96chen: Changed locally, will be in next push

tools/gfx/renderer-shared.h
lucy96chen: Changed locally, will be in next push

tools/gfx/persistent-shader-cache.cpp
lucy96chen: Changed locally, will be in next push

tools/gfx-unit-test/shader-cache-tests.cpp
lucy96chen: Changed locally, will be in next push.

But honestly, I'm starting to think the shader files that are automatically written by the tests probably ought to be added to .gitignore...

lucy96chen: Works for me. I've deleted the old shaders locally and updated .gitignore so it doesn't pick up the renamed version of the automatically written shaders. Will be in next push

tools/gfx/gfx.slang
lucy96chen: Changed this to 0, but I'm opting to keep the default on the cache filename since the cache is enabled by providing at least one of `shaderCachePath` and `shaderCacheFileSystem`. If someone wants to enable the cache but doesn't care about limiting entry count or exactly what the filename is, then these defaults allow them to do that.

Will be in next push

tools/gfx/gfx.slang
lucy96chen: See comment below

tools/gfx-unit-test/cache-eviction-policy-expected-d3d12-1.txt
lucy96chen: Unrelated change, but I've removed all the shader modifications. Other tests in this file already do plenty of shader modifying, and the central focus of this test is to confirm that the entries are being added in the correct order and are being removed correctly when the cache reaches size limits. The shader modifications make this more complex than it needs to be.

Instead of the hard-coded values, I've changed the test to save the contents of the cache file after each step. Since no shaders are being modified, a shader will correspond to the same line in all versions of the file, which allows us to check that the files themselves are in the right orders by checking that certain positions in the lists are equal to each other. Maybe not as straightforward as fetching the hashes first before anything else and doing straight comparisons, but that idea sounded a bit terrible in my head.

tools/gfx-unit-test/shader-cache-tests.cpp
lucy96chen: Structure of this test has changed, see comment somewhere below

lucy96chen:
There's a weird indent in premake5.lua that I have fixed locally. Will include in the next push I make after this next round of review

tools/gfx-unit-test/shader-cache-tests.cpp
```diff
@@ -689,6 +706,129 @@ namespace gfx_test
...
+            {
+                lines.add(line);
+            }
+        }
+
+        // Check the correctness of the cache's entries by comparing the order of entries in the
+        // current state of the cache with what we expect.
+        void checkCacheFiles()
+        {
+            // Check that shader A appears where we expect it to.
```
csyonghe, line 738: Should also check the number of lines are the expected. 

Should check the corresponding files exist. 

Should check the supposed to be deleted file does not exist.

csyonghe (approved):
This looks good. You can improve the test in a follow up PR.

tools/gfx-unit-test/shader-cache-tests.cpp
lucy96chen: Putting up a PR with these changes shortly

PR #2459, opened 2022-10-20
Add cache test using a graphics shader

lucy96chen:
Changes:
- Added a new shader cache test that utilizes a graphics instead of a compute shader - This is to check that the shader is correctly split into multiple entry points in the cache, each with their own distinct entry.


lucy96chen:
Closing as I have several new changes that also need to be merged before I can open an MR for integration into Falcor. The branch containing those changes was built on top of this one.

PR #2460, opened 2022-10-21
Rework differentiation of member access through `[DerivativeMember(DiffType.field)]`

csyonghe:
This change removes the need to write `_getDifferentialFor_x` methods to support auto-diff member access operations (field-extract, field-address).

Instead, we add a `[DerivativeMember(DifferentialType.memberName)]` attribute that users can put on member definitions to inform the compiler what the corresponding member in `DifferentialType` is for the original member.

To allow introducing additional attributes on original types in an extension, this change also implements `extern` variables in extensions. If a variable is defined in an extension and it is marked `extern`, that will not introduce new member to the type being extended, instead it is used for extending the original definition with new attributes. This way we can specify how original members correspond to the differential fields.

With this change, the `generic-jvp` test is now refactored so an extension can be defined as follows:
```
extension myfloat4 : IDifferentiable
{
    typedef myfloat4 Differential;

    [DerivativeMember(Differential.val)]
    extern vector<Real, 4> val;

    static Differential zero()
    {
        return myfloat4(0);
    }

    static __differentiate_jvp Differential dadd(Differential a, Differential b)
    {
        return a + b;
    }

    static __differentiate_jvp Differential dmul(Differential a, Differential b)
    {
        return a * b;
    }
};
```


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -144,6 +144,14 @@ struct DifferentiableTypeConformanceContext
     // 
     IRInst* getDifferentialForType(IRBuilder* builder, IRType* origType)
     {
+        switch (origType->getOp())
```
saipraveenb25, line 147: Should this switch-case be here explicitly? I think the extensions defined in diff.meta.slang can handle these types?

saipraveenb25 (approved):

source/slang/slang-ir-diff-jvp.cpp
csyonghe: Yeah, I added this thinking that it might be good to have a fast path for builtin types, but this is not necessary.

PR #2464, opened 2022-10-25
Mesh shader support

expipiplus1:
~~There are a heap of `TODO: Ellie` in the code where I consider things in
need of touching up.~~

~~Also need to touch up the commit history~~

A miscellaneous list of other TODOs:

- [x] test swizzled store
- [x] HLSL design is misleading, why are vertices passed as a parameter instead of being some global buffer, we should consider allowing meshoutput params to be global
- [ ] Multiple "Vertices" parameters
- [x] Test (or at least think about) generic functions dealing with Vertices, specifically how the
- specialization interacts with the SOA transformation if we cant' specialize early.
- [x] enforce that mesh outputs are always `out`
- [x] test nested structs
- [x] test using the vertex struct elsewhere in a non-output fashion, it
- shouldn't be SOAd
- [ ] tests structs with fixed array members
- [x] test passing the same parameter in differently named

    ```
    foo(out Vertices xs, out Vertices ys){
        xs[0] = 1;
        ys[0] = 2;
    }
    ... foo(vs, vs);
    ```

- [x] test passing indices and vertices in

    ```
    foo(out Vertices xs, out Indices ys){
        xs[0] = 1;
        ys[0] = 2;
    }
    ```

- [x] test diagnostic is thrown for failure/nonconformance conditions 

- [ ] should `Main` want to take more than one `Vertices`?

- [ ] test

    ```
    struct V {
        float4 wib[3] : Foo;
    };
    ...
        vs[2].wib[0] = float4(0);
    ```

- [x] What should the behaviour be for this, copy in/out, pass by ref
  (unspecified)

    ```
    foo(out V v){
        v.member1 = 0;
    }
    main(out Vertices<V> vs){
        v.member2 = 0;
        foo(vs[0]);
    }
    ```


tests/pipeline/rasterization/mesh/component-write.slang
```diff
@@ -0,0 +1,49 @@
+// component-write.slang
+
+// This tests that writing to individual components of the output struct works
+
+//TEST:CROSS_COMPILE:-target spirv -profile glsl_450+spirv_1_4 -entry main -stage mesh
```
csyonghe, line 5: We generally prefer render/compute tests that is runnable.
Admittedly we don't have mesh support in gfx so that is not possible yet.

The problem with cross compile tests is that they are very brittle to changes in IR optimization passes or downstream compiler updates.
If we are going to utilize more cross compile tests, we really need to start integrating llvm file-check and start writing more robust regex rules to check the resulting code.

expipiplus1: Yeah, it wasn't possible trivially

https://github.com/shader-slang/slang/issues/2520

PR #2465, opened 2022-10-25
Adding a differentiable standard library

saipraveenb25:


source/slang/slang-ir-specialize.cpp
```diff
@@ -373,7 +373,76 @@ struct SpecializationContext
         // represents a definition rather than a declaration.
         //
         if(!canSpecializeGeneric(genericVal))
+        {
```
csyonghe, line 376: This will cause the specialization pass to specialize all custom derivative functions even when not using auto-diff.

Instead of preemptively specialize the custom func here, we should consider run specialize pass and auto-diff pass in a loop until no more changes are possible.

source/slang/diff.meta.slang
```diff
@@ -151,3 +151,62 @@ struct __DifferentialPair
         return p();
     }
 };
+
+typealias IDFloat = IFloat & IDifferentiable;
+
+namespace dstd
```
csyonghe, line 157: We should make a decision to either define these on the original stdlib functions or duplicate them in dstd namespace.

source/slang/slang-check-modifier.cpp
```diff
@@ -628,13 +628,102 @@ namespace Slang
...
 
-            customJVPAttr->funcDeclRef = funcExpr;
+            auto diffDeclRef = as<DeclRefExpr>(diffExpr)->declRef;
+
+            UCount genericLevels = 0;
+            // If we've grabbed the outer generic for some reason,
+            // recursively construct GenericAppExpr<...>(generic)
+            // and check that to get a specialized func.
+            // 
+            while (diffDeclRef.as<GenericDecl>().getDecl() != nullptr)
```
csyonghe, line 657: I think this is not necessary. The `diffDeclRef` already contains subsitution info for the generic args. Just create a `DeclRefExpr` with `diffDeclRef` and use it  I guess the missing part is that we can't easily fill in the `type` field for the declref expr without invoking `CheckTerm`, so perhaps consider making `customJVPAttr->funcDeclRef` just a declref?

csyonghe (approved):

source/slang/diff.meta.slang
saipraveenb25: One reason why I couldn't place the decorations on the original declarations in hlsl.meta.slang is that those are defined with __BuiltinArithmeticType. This is fine for the intrinsic op since it gets handled internally, but the differential function has a body and cannot do much with __BuiltinArithmeticType.

I think once those functions are moved to IFloat, we could think about declaring IFloat as extending IDifferentiable, and move the modifiers on to the original declarations.

source/slang/slang-check-modifier.cpp
saipraveenb25: Yes, I realized about 80% through the implementation that I could use `__custom_jvp(diff_fn<T>)` instead of using this logic, but I decided to commit it anyway since we can also use just `__custom_jvp(diff_fn)` and have the substitutions made automatically.

We can remove it though, it might make things break later on.

PR #2467, opened 2022-10-26
Shader cache bugfixes and test additions

lucy96chen:
Changes:
- Fixed a bug exposed by a local integration of the shader cache into Falcor - `ComponentType::computeDependencyBasedHash` has been updated to also hash an entry point's mangled name.
- Performance improvements - AST hashing has been completely overhauled to hash the contents of a `Module` (now pre-computed in `Linkage::loadModule`) instead of the serialized AST due to the significant amount of time AST serialization requires. All AST hashing functions have been renamed to contents hashing.
- Added a new test that uses a graphics shader instead of a compute shader to check shaders which produce `ComponentType`s with multiple entry points
- Added a new test that guards against regression on computed shader cache keys where the only thing differentiating two different entry points is their mangled name - This is the bug that came up during Falcor integration.
- `RequirementDictionary` for `WitnessTable` is now an `OrderedDictionary` and a new `SerialTypeInfo<OrderedDictionary>` has been added - This is to ensure consistent Dictionary order whenever AST serialization occurs. No longer relevant for this PR but could be useful for future serialization uses.

No changes for directory creation if the shader cache path provided does not exist. This change will be made Falcor-side.

### DIFF
--- a/source/slang/slang.cpp
+++ b/source/slang/slang.cpp
@@ -1898,9 +1898,10 @@ protected:
     // by applications to decide when they need to "hot reload"
     // their shader code.
     //
-    void handleFileDependency(String const& path) SLANG_OVERRIDE
+    void handleFileDependency(String const& path, ISlangBlob* sourceBlob) SLANG_OVERRIDE
     {
         m_module->addFilePathDependency(path);
+        m_module->getContentsDigestBuilder().addToDigest(sourceBlob);
     }
 
     // The second task that this handler deals with is detecting
@@ -2283,6 +2284,8 @@ void FrontEndCompileRequest::parseTranslationUnit(
             tokens,
             getSink(),
             languageScope);
+
+        module->getContentsDigestBuilder().addToDigest(sourceFile->getContent());
 
         // Let's try dumping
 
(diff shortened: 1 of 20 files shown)


source/slang/slang.cpp
```diff
@@ -1898,9 +1898,10 @@ struct FrontEndPreprocessorHandler : PreprocessorHandler
     // by applications to decide when they need to "hot reload"
     // their shader code.
     //
-    void handleFileDependency(String const& path) SLANG_OVERRIDE
+    void handleFileDependency(String const& path, ISlangBlob* sourceBlob) SLANG_OVERRIDE
     {
         m_module->addFilePathDependency(path);
+        m_module->getContentsDigestBuilder().addToDigest(sourceBlob);
```
csyonghe, line 1904: What happens if sourceBlob is nullptr?

csyonghe (approved):
Looks good.

source/slang/slang.cpp
lucy96chen: That check slipped my mind when I went to add this, added.

csyonghe (approved):

PR #2471, opened 2022-10-27
Rename `__jvp`-->`__fwd_diff`.

csyonghe:
Rename `__jvp` operator to `__fwd_diff` and related class names from `JVPDifferentiateXXX` to `ForwardDifferentiateXXX`.

@saipraveenb25 I suggest you to merge with these renaming changes as soon as they land to prevent merge conflicts down the road.


saipraveenb25 (approved):
We might have to change some of the internal comments that use 'JVP' to refer to forward-mode auto-diff so it's less confusing for later, but it's not a priority just yet.

PR #2472, opened 2022-10-27
Rename `JVPDerivativeModifier` -> `ForwardDifferentiableAttribute`.

csyonghe:


saipraveenb25 (approved):
We should also rename the test suffixes form `-jvp` to `-fwd` (not priority)

PR #2477, opened 2022-10-31
Make `DifferentialPair` able to nest.

csyonghe:
This change makes the `DifferentialPair` type nest-able, so we can form `DifferentialPair<DifferentialPair<T>>` types.
This is a prerequisite to support nested/higher-order differentiation.

The challenge here is to define the `DifferentialPair` type so that the Differential of Differential should always evaluates to `DifferentialBottom`, which is a builtin type for representing non-existent/not-well-defined differential. This means that we need to make `DifferentialPair<DifferentialPair<T>>` to effectively form a tripple `(p.p.p, p.p.d p.d.p)`, leaving out the `p.d.d` term that has no mathematical meaning.

To allow such definition, this work starts by modifying the existing `DifferentialPair<T:IDifferentiable>` .
The new definition of `DifferentialPair` type is as follows:
```
__generic<T : IDifferentiable>
__magic_type(DifferentialPairType)
__intrinsic_type($(kIROp_DifferentialPairType))
struct DifferentialPair : IDifferentiable
{
    typedef DifferentialPair<T.Differential> Differential;
    typedef T.Differential DifferentialElementType;

    __intrinsic_op($(kIROp_MakeDifferentialPair))
    __init(T _primal, T.Differential _differential);

    __intrinsic_op($(kIROp_DifferentialPairGetDifferential))
    T.Differential d();

    T.Differential getDifferential()
    {
        return d();
    }

    __intrinsic_op($(kIROp_DifferentialPairGetPrimal))
    T p();

    T getPrimal()
    {
        return p();
    }

    [__unsafeForceInlineEarly]
    static Differential dzero()
    {
        return Differential(T.dzero(), Differential.DifferentialElementType.dzero());
    }

    [__unsafeForceInlineEarly]
    static Differential dadd(Differential a, Differential b)
    {
        return Differential(
            T.dadd(
                a.p(),
                b.p()
            ),
            Differential.DifferentialElementType.dzero());
    }

    [__unsafeForceInlineEarly]
    static Differential dmul(This a, Differential b)
    {
        return Differential(
             T.dmul(a.p(), b.p()),
            Differential.DifferentialElementType.dzero());
    }
};
```

There is one tricky bit when defining `DifferentialPair` this way. Notice the definition of `Differential`:
```
 typedef DifferentialPair<T.Differential> Differential;
```

This line has a type mismatch issue because `T.Differential` is unconstrained and using it to fill in type parameter that requires `IDifferential` conformance is not valid.

To solve this issue, we introduce a new sub-typing rule that if a generic type `T` is unconstraint, we treat it as a subtype of `DifferentialBottom`. This rule allowed us to form the `Differential` type for `DifferentialPair`, and that the `Differential.Differential` will resolve to `DifferentialPair<T.Differential, DifferentialBottom>`, which is exactly what we want.

The rest of this PR is all about dealing with the consequences of changing the stdlib definition of `DifferentialPair`.

Most of the work done here is to get `generic-impl-jvp` test to pass. First, we need to modify `dvector` type to conform to `IDifferentiable`  since it is used as `Differential` type for `myvector`. We will just rely on automatic synthesis to implement the conformances. However, `dvector` is a generic type, so the first part of the work is to extend our synthesis logic to work with generic parent types. The next obstacle is that `values` field is an array, and we need our synthesis logic to be smart enough to synthesize a loop to compute each element of the array. This required us to synthesize significantly more complex code, so I've added the `ASTSynthesizer` class to simplify the synthesis code.

Another major challenge addressed in this PR is to register type conformances only within a differentiable function at each expression site instead of doing it globally at `coerce`. This allows us to avoid running the differentiation logic for programs that don't need it.

### DIFF
--- a/source/slang/slang-ast-synthesis.cpp
+++ b/source/slang/slang-ast-synthesis.cpp
@@ -1,0 +1,175 @@
+#include "slang-ast-synthesis.h"
+
+namespace Slang
+{
+Expr* ASTSynthesizer::emitBinaryExpr(UnownedStringSlice operatorToken, Expr* left, Expr* right)
+{
+    auto infixExpr = m_builder->create<InfixExpr>();
+    infixExpr->functionExpr = emitVarExpr(m_namePool->getName(operatorToken));;
+    infixExpr->arguments.add(left);
+    infixExpr->arguments.add(right);
+    return infixExpr;
+}
+
+Expr* ASTSynthesizer::emitPrefixExpr(UnownedStringSlice operatorToken, Expr* base)
+{
+    auto prefixExpr = m_builder->create<PrefixExpr>();
+    prefixExpr->functionExpr = emitVarExpr(m_namePool->getName(operatorToken));;
+    prefixExpr->arguments.add(base);
+    return prefixExpr;
+}
+
+Expr* ASTSynthesizer::emitPostfixExpr(UnownedStringSlice operatorToken, Expr* base)
+{
+    auto postfixExpr = m_builder->create<PostfixExpr>();
+    postfixExpr->functionExpr = emitVarExpr(m_namePool->getName(operatorToken));;
+    postfixExpr->arguments.add(base);
+    return postfixExpr;
+}
+
+ForStmt* ASTSynthesizer::emitFor(Expr* initVal, Expr* finalVal, VarDecl* &outIndexVar)
+{
+    auto scopeDecl = pushVarScope()->containerDecl;
+    auto stmt = m_builder->create<ForStmt>();
+    stmt->scopeDecl = (ScopeDecl*)scopeDecl;
+    auto declStmt = emitVarDeclStmt(nullptr, m_namePool->getName("S_synth_loop_index"), initVal);
+    stmt->initialStatement = declStmt;
+    outIndexVar = (VarDecl*)declStmt->decl;
+    auto predicateExpr = emitBinaryExpr(UnownedStringSlice("<"), emitVarExpr(outIndexVar), finalVal);
+    stmt->predicateExpression = predicateExpr;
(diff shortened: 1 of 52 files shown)


tests/autodiff/generic-impl-jvp.slang
```diff
@@ -8,8 +8,8 @@ typedef float Real;
 
 typealias IDFloat = IFloat & IDifferentiable;
 
-__generic<T, let N : int>
-struct dvector
+__generic<T : IDifferentiable, let N : int>
+struct dvector : IDifferentiable
```
saipraveenb25, line 12: I'm not totally sure why `dvector` needs to be `IDifferentiable`? This test doesn't do higher order differentiation, and without `IDifferentiable`, wouldn't `dvector` be treated as conforming to `IDifferentiable` via `DifferentialBottom`?

csyonghe: I think I just forgot to remove it.

source/slang/slang-ast-synthesis.cpp
```diff
@@ -0,0 +1,175 @@
+#include "slang-ast-synthesis.h"
+
+namespace Slang
```
saipraveenb25, line 3: It looks like we're doing two separate code synthesis passes for auto-diff as a whole, one on the AST-side and one on the IR-side. 
I'm wondering if it's feasible to synthesize just the function headers (and store any required witness tables), and synthesize the method on the IR side?

saipraveenb25 (approved):
This is great! We're a lot closer to higher-order auto-diff, with these changes. 

The DifferentiableTypeDictionary might be the only janky part left we need to cleanup.

csyonghe:
I think the dictionary clean up would be trivial after this PR, because we now only collect witness for expressions inside a differentiable func. We could just stick them to a function-local decoration instead of the global dictionary.

PR #2485, opened 2022-11-02
FileStream-based implementation for updating cache index file

lucy96chen:
Changes:
- Changed updates to the cache index file on disk to go through a `FileStream` to do modifications in place instead of needing to constantly loop through entries and re-write the cache index file - The old implementation is now a fall back in case physical file paths cannot be obtained (i.e. an in-memory file system).
- Modified shader cache tests to test both on-disk and in-memory file systems for the cache where possible - Any on-disk test should now automatically clean up generated files after running.

As far as I can tell by observing load times in Falcor, this runs more or less about as quickly as the previous implementation for small caches but scales significantly better as the number of cache entries grows.


tools/gfx/persistent-shader-cache.cpp
```diff
@@ -42,39 +48,69 @@ void PersistentShaderCache::loadCacheFromFile()
...
-        ShaderCacheEntry entry = { dependencyDigest, contentsDigest };
-        auto entryNode = entries.AddLast(entry);
-        keyToEntry.Add(dependencyDigest, entryNode);
-
-        // If there are more entries present in the cache file than the entry count limit allows,
-        // ignore the rest. Since we output the lines in the cache file in order of most to least
-        // recently used, the cache's behavior will still be correct.
-        if (desc.entryCountLimit > 0 && entries.Count() == desc.entryCountLimit)
-            break;
+        auto indexString = UnownedStringSlice((char*)indexBlob->getBufferPointer());
```
csyonghe, line 61: I suggest we do binary file read write to skip the whole parsing/toString conversion altogether.

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -121,13 +161,24 @@ void PersistentShaderCache::addEntry(const slang::Digest& dependencyDigest, cons
...
+
+    ShaderCacheEntry entry = { dependencyDigest, contentsDigest, lastAccessedTime };
     auto entryNode = entries.AddFirst(entry);
     keyToEntry.Add(dependencyDigest, entryNode);
 
     mutableShaderCacheFileSystem->saveFileBlob(DigestUtil::toString(dependencyDigest).getBuffer(), compiledCode);
 
-    saveCacheToFile();
+    indexStream.seek(SeekOrigin::End, 0);
+    entryToOffset.Add(entry, indexStream.getPosition());
```
csyonghe, line 173: For binary files, just seek and write the raw bytes.

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -121,13 +161,24 @@ void PersistentShaderCache::addEntry(const slang::Digest& dependencyDigest, cons
         deleteLRUEntry();
     }
 
-    ShaderCacheEntry entry = { dependencyDigest, astDigest };
+    auto lastAccessedTime = duration_cast<duration<double>>(high_resolution_clock::now().time_since_epoch()).count();
```
csyonghe, line 164: I don't think the duration_cast here is necessary.

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -96,13 +132,17 @@ LinkedNode<ShaderCacheEntry>* PersistentShaderCache::findEntry(const slang::Dige
         entries.AddFirst(entryNode);
         if (mutableShaderCacheFileSystem)
         {
-            saveCacheToFile();
+            auto lastAccessedString = String(duration_cast<duration<double>>(high_resolution_clock::now().time_since_epoch()).count());
```
csyonghe, line 135: You can store a `List<ShaderCacheEntry> ` that is exactly the same as the whole file in memory.
Your linked list can just be `LinkedList<ShaderCacheEntry*>` or `LinkedList<Index>` that points to locations in the `List`. The offset can then be calculated directly without needing the `entryToOffset` dictionary.

tools/gfx/persistent-shader-cache.cpp
lucy96chen: done

tools/gfx/persistent-shader-cache.cpp
lucy96chen: done

tools/gfx/persistent-shader-cache.cpp
lucy96chen: done

tools/gfx/persistent-shader-cache.cpp
```diff
@@ -35,52 +41,109 @@ PersistentShaderCache::PersistentShaderCache(const IDevice::ShaderCacheDesc& inD
...
+            // While this has no impact on cache operation, it could be problematic for debugging purposes, etc.
+            indexStream.close();
+            indexStream.init(filePath, FileMode::Create, FileAccess::ReadWrite, FileShare::ReadWrite);
+            return;
+        }
+        else
+        {
+            // The cache index is not guaranteed to be ordered by most recent access, so we need a temporary list to store
+            // all the entries in order to sort them before filling in our linked list.
+            List<ShaderCacheEntry> tempEntries;
```
csyonghe, line 112: No need to use a loop to read.
```
tempEntries.setCount(numEntries);
indexStream.read(tempEntries.getBuffer(), sizeof(ShaderCacheEntry)*numEntries, ..);
```

lucy96chen: done

tools/gfx/persistent-shader-cache.h
```diff
@@ -36,38 +51,50 @@ class PersistentShaderCache : public RefObject
...
+    // of the removed entry so addEntry() can overwrite the corresponding entry in 'entries'
+    // with the new entry. This should only be called by addEntry() when the cache reaches maximum capacity.
+    Index deleteLRUEntry();
+
+    // These functions perform the same tasks as their counterparts defined above. However, without access
+    // to a physical file path, in-memory file systems cannot leverage file streams and need to fall back
+    // on a different implementation.
+    void loadCacheFromMemory();
+    ShaderCacheEntry* findEntryInMemory(const slang::Digest& key, ISlangBlob** outCompiledCode);
+    void addEntryToMemory(const slang::Digest& dependencyDigest, const slang::Digest& contentsDigest, ISlangBlob* compiledCode);
```
csyonghe, line 72: We should find a way to unify most of the logic between os file system path and memory file system path.
I think we define wrappers like  `seekInFile()` and `writeBytesToFile` to intercept the `fseek` `fwrite` calls that modifies a real file for the OS path and just do in-place modification in memory for the memory path.

lucy96chen: Looking at this again, only loading the cache from file has significant differences in logic due to their different sources. The memory path just needs the extra save call on exit and otherwise shares basically all of its logic with the OS path, which has the extra file stream read/write. Think just wrapping the file stream I/O in an if check for an on-disk file system is sufficient.

source/slang/slang.cpp
```diff
@@ -3269,7 +3263,33 @@ void Module::updateDependencyBasedHash(
...
 {
-    builder.addToDigest(getContentsDigest());
+    auto filePathDependencies = getFilePathDependencies();
+
+    DigestBuilder lastModifiedBuilder;
+    for (auto file : filePathDependencies)
+    {
+        struct stat fileStatus;
+        stat(file.getBuffer(), &fileStatus);
+        lastModifiedBuilder.addToDigest(fileStatus.st_mtime);
```
csyonghe, line 3273: What if `stat` failed?

lucy96chen: hmm, safer to change this back to returning a `Result` then to help catch `stat` returning non-zero. I was thinking that the process of building up the list of file dependencies should catch any problems, but I suppose a file could be inaccessible even if it exists.

lucy96chen: I'm going to change the return type for `compute/updateDependencyBasedHash` to `Result` as well to maintain parity between the two functions (and in case any potential future additions to the hashing can fail).

lucy96chen: Changed to handle file dependencies which do not physically reside on disk.

csyonghe (approved):

PR #2486, opened 2022-11-02
Add gfxGetAdapters function (currently implemented for D3D12/Vulkan)

skallweitNV:


tools/gfx/render.cpp
```diff
@@ -233,6 +236,40 @@ extern "C"
         return SLANG_OK;
     }
 
+    SLANG_GFX_API SlangResult SLANG_MCALL gfxGetAdapters(DeviceType type, AdapterInfo* outAdapters, Size bufferSize, GfxCount* outAdapterCount)
```
csyonghe, line 239: This two-call idiom is OK, but we've been moving to just return an allocated ISlangBlob to return buffers.

csyonghe (approved):

csyonghe (approved):

csyonghe (approved):

PR #2490, opened 2022-11-03
Add adapterIndex to IDevice::Desc for selecting adapter

skallweitNV:
This allows selecting a specific adapter if multiple adapters match the given name.


csyonghe (approved):
Looks good. Shall we consider remove the adapter name field so we don't need the user to specify both of them, which seems redundant?

skallweitNV:
I'm closing this and opening a new PR that does the identification through LUID.

PR #2491, opened 2022-11-03
Shader Execution Reordering for VK

jsmall-nvidia:
Currently outputs GLSL to test.

Changes:

* Added LocationTracker - for tracking types of index/locations as seen on Vulkan/GLSL 
* Fixed issues around GetAttribute
* Added HitObjectAttribute and related types

Getting GetAttributes to work required a 'trick'/hack. With other ray tracing functions the attributes/payload were stored as a static variable within the function body. Since the function bodies were generic on the payload/attributes this would mean that there would be a location and a global variable for each of the types.

With HitRay the .GetAttributes method needs to be able to get the attributes that are set on the launch. It's not possible to store the static variable in the MakeHit function then, because it wouldn't be available to .GetAttributes. In C++ you could write

```C++
template <typename T>
T& _accessGlobal()
{
   static T t;
   return t;
}
```

Now it's possible to read/write the global T for any type by. 

```
_accessGlobal<float>() = 1.0f;
float value = _accessGlobal<float>();
```

In Slang if we try and do something notionally similar

```
__ref T __hitObjectAttributes<T>()
{
    [__vulkanHitObjectAttributes]   
    static T t;
    return t;
}
```

Doesn't actually work, but produces a function that returns a value. Fortunately `[ForceInline]` comes to the rescue, in output the function is replaced with the global, which is what was desired. Whilst that would seem ok and reasonable perhaps, there's something fishy in that if we remove `_ref` it still works(!) Which means I have *write* access to global variable when I shouldn't as I should be receiving a copy. For now this works for our purposes, but relies on behavior that seems incorrect. If __ref were honored in a similar way to C++ it would perhaps be reasonable. 

If the desired behavior in the original code was to have locations setup based on the used types, then using something like above would improve things in that it would only output global variables for each unique type/type whereas that isn't the case right now.

To properly test GLSL output will be necessary to have a suitable build on GLSLANG.

The GLSL extension has special behavior around `hitObjectNV`, where it cannot be assigned. That produces a problem because the HLSL API uses assignment! 

```
HitObject hitObject = HitObject::MakeXXXX(..).;
```

Needs to be output as something like

```
hitObjectNV hitObject;
extensionFunctionXXX(hitObject, ... );
```

It turns out using `[ForceInline]` on the GLSL `__specialized_for_target(glsl)`. This works for this example but does not work in general. For example

```
HitObject hitObjs[10];

// This will produce GLSL code with assignment
hitObjs[0] = HitObject::MakeXX(...);

HitObject doThing() {}
// If doThing isn't ForceInline will produce an assignment
HitObject hitObj = doThing();

// Won't work because will use assignment (which can't be SSA'd away)
HitObject hitObj;

if (x & 1)
{
   hitObj = HitObject::MakeXX(...);
}
else
{
  hitObj = HitObject::MakeXX(...);
}
```

Assignment can be in source code, if SSA means no assignment is produced (this is confirmed in `hit-object-assign.slang` test.

To do:

* GLSLANG compiler testing
* Have tests use SPIR-V to test not GLSL
* Motion support isn't currently available on NVAPI
* The issues VK described above need to be improved upon - at a minimum giving an error if not possible


tests/hlsl-intrinsic/shader-execution-reordering/hit-object-trace-motion-ray.slang.1.expected
```diff
@@ -0,0 +1,96 @@
...
+    float currentTime_0 = float(_S5);
+    SomeValues_0 someValues_0;
+    SomeValues_0 _S6 = { idx_0, float(idx_0) * 2.00000000000000000000 };
+    someValues_0 = _S6;
+    RayDesc_0 ray_0;
+    ray_0.Origin_0 = vec3(float(idx_0), 0.00000000000000000000, 0.00000000000000000000);
+    ray_0.TMin_0 = 0.00999999977648258209;
+    ray_0.Direction_0 = vec3(0.00000000000000000000, 1.00000000000000000000, 0.00000000000000000000);
+    ray_0.TMax_0 = 10000.00000000000000000000;
+    hitObjectNV hit_1 = HitObject_TraceMotionRay_0(scene_0, 20U, 255U, 0U, 4U, 0U, ray_0, currentTime_0, someValues_0);
```
alelenv, line 89: Generated GLSL is illegal.
`hitObjectNV` is an opaque type, you cannot assign/initialize it.

This is exactly similar in spirit to `rayQueryEXT` (with all its syntactic limitations)
So whatever codegen slang does for translating `RayQuery`, same should be done for `hitObjectNV`

jsmall-nvidia: Ok I see. That makes some parts of the API make more sense, as it didn't seem to map similarly to the NVAPI version of things. Will look to make suitable changes.

jsmall-nvidia: @alelenv There remain some issues, but it looks like the output for these tests appear appropriate. Would you mind taking a quick look?

csyonghe (dismissed):

tests/hlsl-intrinsic/shader-execution-reordering/hit-object-trace-motion-ray.slang.1.expected
```diff
@@ -0,0 +1,88 @@
...
+layout(binding = 0)
+uniform accelerationStructureEXT scene_0;
+
+layout(location = 0)
+hitObjectAttributeNV
+SomeValues_0 t_0;
+
+SomeValues_0 HitObject_GetAttributes_0(hitObjectNV this_0)
+{
+    return t_0;
```
alelenv, line 30: This should be extracting attributes from hitObjectNV using
`hitObjectGetAttributesNV(this_0, 0)`
But I think you have mentioned this as TODO/WIP

I am seeing sensible codegen from our stack (before entering rtcore) for most of the examples(havent checked all) so change looks good to me

PR #2493, opened 2022-11-04
Use property for DifferentialPair accessors.

csyonghe:


source/slang/diff.meta.slang
```diff
@@ -70,22 +70,34 @@ struct DifferentialPair : IDifferentiable
...
 
-    __intrinsic_op($(kIROp_DifferentialPairGetDifferential))
-    T.Differential d();
+    property p : T
+    {
+        __intrinsic_op($(kIROp_DifferentialPairGetPrimal))
+        get;
+    }
+
+    property v : T
```
saipraveenb25, line 79: Is `.v` used anywhere? It seems to be identical to `.p`

csyonghe: They are not, I was at a point think about using `v` instead of `p` since `p` could mean `ptr` in many naming conventions.

saipraveenb25 (approved):

PR #2502, opened 2022-11-09
Initial plumbing of backward autodiff in the frontend.

behindthepixels:
This is a WIP PR of backward autodiff implementation.


source/slang/slang-ir-diff-jvp.cpp
```diff
@@ -2518,4 +2518,1954 @@ void stripAutoDiffDecorations(IRModule* module)
 }
 
 
+
```
csyonghe, line 2521: We probably want to move this to its own file.

csyonghe (commented):
The plumbing changes looks good! We probably want to move the backward pass into its own file and pull things that can be shared between the two passes into a separate `slang-ir-auto-diff-shared.h/cpp`.

csyonghe (approved):

source/slang/slang-ir-diff-jvp.cpp
behindthepixels: Yeah I can do some refactor after getting some basic functionality working.

csyonghe:
Superceded by PR #2514.

PR #2503, opened 2022-11-09
Fix supported shader model check in D3D12 device

skallweitNV:
Previously, `CheckFeatureSupport` was called with the highest SM first (6.7) but the call can fail and then `shaderModelData.HighestShaderModel` was just left at an unsupported SM. The fix iterates through all known SMs and checks for the one actually supported by the driver.


tools/gfx/d3d12/d3d12-device.cpp
```diff
@@ -601,21 +601,30 @@ Result DeviceImpl::initialize(const Desc& desc)
...
         SLANG_COMPILE_TIME_ASSERT(D3D_SHADER_MODEL_6_0 == 0x60);
 
         {
+            // CheckFeatureSupport(D3D12_FEATURE_SHADER_MODEL) can fail if the runtime/driver does not yet know the
+            // specified highest shader model. Therefore we assemble a list of shader models to check and
+            // walk it from highest to lowest to find the supported shader model.
+            Slang::ShortList<D3D_SHADER_MODEL> shaderModels;
+            if (m_extendedDesc.highestShaderModel != 0)
+                shaderModels.add((D3D_SHADER_MODEL)m_extendedDesc.highestShaderModel);
+            for (int i = SLANG_COUNT_OF(kKnownShaderModels) - 1; i >= 0; --i)
```
jsmall-nvidia, line 617: Assumes that kKnownShaderModels is in lowest/highest order (which it is), might benefit for a comment on kKnownShaderModels that other code assumes this.

jsmall-nvidia (approved):

PR #2504, opened 2022-11-09
Improvements to NVRTC diagnostic parsing

jsmall-nvidia:
Motivation here was around getting f32tof16 and f16tof32 working on CUDA. This doesn't *quite* get that, for mundane reasons. Trying to get it work revealed that 

* -X wasn't working for NVRTC 
* Some kinds of diagnostic errors were being ignored. 
  * When this happened the test wouldn't fail but would be ignored(!)

This improves diagnostic parsing and adds -X support. It also fixes the CUDA versions of those intrinsics. 

The problem with ignoring is around the use of SLANG_E_NOT_AVAILABLE, and perhaps how 'diagnostics' can hold the compilation error. The problem here is that SLANG_E_NOT_AVAILABLE is being used for two distinct scenarios

* Something was requested, but isn't available (say a blob can't be created on an artifact)
* As an indicator that a test should be *ignored*

Normal Result propagation means errors, are typically handed up to a caller. 

We probably want to add a new error result which means something distinct for `this test can't be executed because suitable stuff isn't available`. Then vet the uses of SLANG_E_NOT_AVAILABLE and use the new result where appropriate. The new result can then map to 'ignore' and the SLANG_E_NOT_AVAILABLE is a normal error.

But they still don't work because CUDA backend uses the use of half type to detect the need for including the `cuda_fp16.h` header. But the intrinsics work *don't* work on half type. But rely on the header. So there remains is a question on how to best fix that.

Can return to these issues once SER VK is working.


source/compiler-core/slang-nvrtc-compiler.cpp
```diff
@@ -228,24 +228,68 @@ static SlangResult _parseNVRTCLine(SliceAllocator& allocator, const UnownedStrin
...
+        if (split1 == toSlice("error") || 
+            split1 == toSlice("catastrophic error"))
+        {
+            severity = Severity::Error;
+        }
+        else if (split1 == toSlice("warning"))
+        {
+            severity = Severity::Warning;
+        }
+        else if (split1.indexOf(toSlice("error")) >= 0)
```
csyonghe, line 244: might be better to use "error:" since to avoid triggering this logic in situations like `warning: variable "errorCode" is not used`.

csyonghe (approved):

source/compiler-core/slang-nvrtc-compiler.cpp
jsmall-nvidia: Fair point, I'll add a more constrained test.

PR #2506, opened 2022-11-10
Fix inlining pass.

csyonghe:
- Perform inline without modifying original func.
- Defer multi-level-break elimination until very late in the process.
- Strengthen the condition to apply trivial loop simplification to prevent generating ill formed control flow that leads to code duplication at emit.

Also in this change:
- Improved natvis in release config.
- Disable calls to IRUse debug validation checks that impacts most heavily to performance in debug. Add a separate build flag to turn that on and off. Default to off.


source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -2,16 +2,105 @@
...
+}
+
+// Test if a loop is trivial: a trivial loop runs for a single iteration without any back edges, and
+// there is only one break out of the loop at the very end. The function generates `regionTree` if
+// it is needed and hasn't been generated yet.
+bool isTrivialSingleIterationLoop(
+    IRGlobalValueWithCode* func,
+    IRLoop* loop,
+    RefPtr<IRDominatorTree>& dom,
+    RefPtr<RegionTree>& regionTree)
```
jsmall-nvidia, line 30: Nitpik: Could make output parameters clearer with out prefix.

source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -2,16 +2,105 @@
 
 #include "slang-ir-insts.h"
 #include "slang-ir.h"
+#include "slang-ir-dominators.h"
+#include "slang-ir-restructure.h"
 
 namespace Slang
 {
 
+BreakableRegion* findBreakableRegion(Region* region)
```
jsmall-nvidia, line 11: Might want to make the functions that are private to this file static.

jsmall-nvidia (dismissed):

PR #2523, opened 2022-11-18
Data flow validation pass for diagnosing derivative loss.

csyonghe:
- Add an IR validation pass to check for unintended call to non-differential functions (and losing derivatives).
- Add `no_diff` keyword to allow the user to clarify the intention and silence the diagnostic.


saipraveenb25 (approved):
This looks good!

I'm wondering if no_diff on a non-differentiable call should just zero out the differential instead of throwing an error.

That way if calls wrapped with no_diff are eventually made differentiable, we won't have to edit the no_diff (since the callee anyway doesn't need the differential)

csyonghe:
I think at least we need a warning for that. For now I am just going to make everything an error and see how it goes. It is always easier to turn errors into warnings than the other way around.

PR #2526, opened 2022-11-22
Refactor Auto-diff passes

saipraveenb25:
Split autodiff into 4 headers (and 4 cpp impls):
- `slang-ir-autodiff.h` holds common structs: `AutoDiffSharedContext`, `DifferentiableTypeConformanceContext`, `DifferentialPairTypeBuilder` and the top-level autodiff pass `processAutodiffCalls` which calls the forward-diff, reverse-diff, pair-lowering and cleanup passes.
- `slang-ir-autodiff-pairs.h` holds the pair type lowering pass `DiffPairLoweringPass`
- `slang-ir-autodiff-fwd.h` holds the `ForwardDerivativePass`
- `slang-ir-autodiff-rev.h` holds the `ReverseDerivativePass`

This refactor makes the next steps for reverse-mode easier: we'll add three new passes, each with it's own file: `DiffTypePropagationPass`, `DiffInstUnzipPass` and `DiffInstTranspositionPass` which will be called from `ReverseDerivativePass`


source/slang/slang-ir-autodiff.h
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    IRInst* getAddMethodForType(IRBuilder* builder, IRType* origType)
+    {
+        return lookUpInterfaceMethod(builder, origType, sharedContext->addMethodStructKey);
+    }
+
+};
+
+struct DifferentialPairTypeBuilder
```
csyonghe, line 148: Should this be named `DifferentialPairLoweringContext`?

source/slang/slang-ir-autodiff-rev.cpp
```diff
@@ -0,0 +1,832 @@
...
+    // processing instructions while maintaining state.
+    //
+    BackwardDiffTranscriber         backwardTranscriberStorage;
+
+    // Diagnostic object from the compile request for
+    // error messages.
+    DiagnosticSink*                 sink;
+
+    // Builder for dealing with differential pair types.
+    DifferentialPairTypeBuilder     pairBuilderStorage;
```
csyonghe, line 816: Should the diff pair lowering step be shared by both forward and reverse pass? It doesn't seem to me that this pass should own diff pair lowering.

csyonghe (approved):

csyonghe:
Not needed to include in this PR, but we should also remove all the inst, stdlib definitions and the type-checking logic for DiffBottom related stuff.

PR #2532, opened 2022-11-25
Fix output type detection for multi-part extensions

expipiplus1:
For example .spv.asm or .dxil.asm


jsmall-nvidia (commented):
This is all looks good in principal.

I wonder though if really the problem is have an 'extension' that has a dot in it. The motivating extensions don't seem standard, just ones we have used. We use a `-` in some of our extensions. Other usage seems okay to just smash words together.

My gut is unless there are examples of 'standard' extensions that are multi-part (and thus can justify the requirement), we should avoid its usage. I think the underlying worry is that supporting multipart in this way requires more information (the filename) and not just the first removed extension.

PR #2546, opened 2022-12-01
Disable end of line trim in .editorconfig

jsmall-nvidia:
Based on an observation from @skallweitNV. 

This change is perhaps somewhat arguable, as it could be a requirement for PRs to honor the setting (or perhaps more complex source format such as clang format).

In the meantime it seems easiest to just disable.


skallweitNV:
Yes please, I did the same on my local copy for now ..

csyonghe (approved):

PR #2547, opened 2022-12-02
Add helper functions to update DifferentialPair

winmad:


csyonghe (approved):
Perhaps prefix them with __ to indicate it is intended for internal use only.

winmad:
> Perhaps prefix them with __ to indicate it is intended for internal use only.

Not sure if it is true. I have used these helper functions in DiffFalcor shaders.

PR #2549, opened 2022-12-02
Cleanup crypto utilities

skallweitNV:
This is initial cleanup for the shader cache refactor.

- Move all crypto hash functions to `slang-crypto` and consolidate the unit tests.
- Add implementation of SHA1 hash.
- Use binary blobs for returning the hash in the main `IComponentType` interface, getting rid of the `Digest` type in the main header file.


source/core/slang-crypto.cpp
```diff
@@ -0,0 +1,603 @@
...
+    SlangInt used, available;
+
+    saved_lo = m_lo;
+    if ((m_lo = (saved_lo + size) & 0x1fffffff) < saved_lo)
+        m_hi++;
+    m_hi += (uint32_t)size >> 29;
+
+    used = saved_lo & 0x3f;
+
+    if (used) {
```
jsmall-nvidia, line 90: Nitpik - Allman is the preferred indenting style.

skallweitNV: Yeah, that code was copied over. I'll do a cleanup pass with the next PR.

PR #2555, opened 2022-12-07
Remove `construct` IR op.

csyonghe:
Follow up on #2553, this PR completely removes `kIROp_Construct` from IR.

Modified all IR passes to use explicit casting/constructing op codes instead of `kIROp_Construct`.


source/slang/slang-ir.cpp
```diff
@@ -3300,19 +3301,244 @@ namespace Slang
...
+            break;
+        }
+        if (fallback)
+        {
+            return emitIntrinsicInst(type, kIROp_DefaultConstruct, 0, nullptr);
+        }
+        return nullptr;
+    }
+
+    static int _getTypeStyleId(IRType* type)
```
jsmall-nvidia, line 3430: Would perhaps benefit from an enum instead of int literals.

csyonghe: This Id is used to lookup into an array (map) for op codes.

source/slang/slang-ir.cpp
```diff
@@ -3300,19 +3301,244 @@ namespace Slang
...
+            {
+                op0 = op; op1 = kIROp_Nop;
+            }
+            OpSeq(IROp op, IROp inOp1)
+            {
+                op0 = op; op1 = inOp1;
+            }
+        };
+
+        static const OpSeq opMap[4][5] =
```
jsmall-nvidia, line 3492: Some of these conversions seem undesirable. Like a bool to a pointer, or a float to a pointer. Perhaps the point is some earlier checking means this doesn't happen in practice.

csyonghe: Right, the type-system already have checks against that. The code here is just being defensive so that it doesn't just crash if the situations occur.

jsmall-nvidia (approved):
Looks good.

PR #2558, opened 2022-12-08
Refactor shader cache

skallweitNV:
- Fix a bug in `Path::find` when using on an empty directory.
- Cleanup source formatting in crypto module.
- Fixes for file lock and adding `FileLockGuard`.
- Remove old persistent shader cache code and add a new simple persistent cache that is multi-thread/process.
- Add initial set of unit tests for persistent cache.
- Replace dependency path list in `ComponentType` with a list of `SourceFile` and rename `FilePathDependencyList` into `FileDependencyList`. This allows to compute hashes over the original data passed to the compiler front end.
- Simplify hash computation in `ComponentType`. Also add some missing pieces in `Linkage`. Move to `SHA1` hash.
- Refactor `IShaderCacheStatistics` interface into `IShaderCache` for easier extendibility. Move stats logic into `PersistentCache`.
- Make existing gfx shader cache tests work (temporarily disabled some tests, this will need more work in follwup PR).


skallweitNV:
@csyonghe the changes with regards to `FilePathDependencyList` need careful review.

source/slang/slang.cpp
```diff
@@ -3247,75 +3237,25 @@ ISlangUnknown* Module::getInterface(const Guid& guid)
...
-        {
-            statFailed = true;
-            break;
-        }
-        lastModifiedBuilder.append(fileStatus.st_mtime);
-    }
-
-    MD5::Digest temp = lastModifiedBuilder.finalize();
-    if (statFailed || temp != lastModifiedDigest)
+    for (SourceFile* sourceFile : getFileDependencies())
```
csyonghe, line 3242: I think this source file enumeration should not happen in `Module` and should be in `ComponentType::getEntryPointHash()` instead because it only need to happen once at root node. The root node (e.g. a CompositeComponentType) should contain transitively all the files in its children components in its file dependency list.

csyonghe: The problem to enumerate here is that a same file can get hashed multiple times.

skallweitNV: good point!

tools/slang-unit-test/unit-test-persistent-cache.cpp
```diff
@@ -0,0 +1,598 @@
...
+
+        LOG("Total time: %.3fs\n", seconds);
+        LOG("Total bytes written: %d\n", bytesWritten.load());
+        LOG("Write througput: %.3fMB/s\n", (bytesWritten.load() / (1024.0 * 1024.0)) / seconds);
+        LOG("Total bytes read: %d\n", bytesRead.load());
+    }
+};
+
+// skip tests
+#if 0
```
csyonghe, line 568: Why are these skipped?

tools/gfx-unit-test/shader-cache-tests.cpp
```diff
@@ -1255,92 +1252,93 @@ namespace gfx_test
...
 
-    SLANG_UNIT_TEST(specializationArgsShaderCacheVulkan)
+    SLANG_UNIT_TEST(shaderCacheSpecializationArgsVulkan)
     {
         runTestImpl(shaderCacheTestImpl<SpecializationArgsEntries>, unitTestContext, Slang::RenderApiFlag::Vulkan);
     }
-
-    SLANG_UNIT_TEST(cacheEvictionPolicyD3D12)
+#if 0
+    SLANG_UNIT_TEST(shaderCacheEvictionPolicyD3D12)
```
csyonghe, line 1305: These tests may be out of date but we need to make sure the eviction policy is tested.

csyonghe (dismissed):
Looks good to me.

csyonghe:
There are still failures on aarch64. We can merge this right now, but would be good to come to a conclusion on what we want to do on the failures. The answer should either be disabling the test on aarch64 with a reason, or trying to fix the issue.

skallweitNV:
Yeah, the aarch64 situation is as follows:
- Test properly works when executed with `slang-test` directly.
- Test fails when executed with `slang-test -use-test-server`.

I wasn't able to figure out why (the same runs on x64). It does not seem to be a permission issue. Maybe someone with more insights into the testing infrastructure should take a closer look at this. So one thing that is kinda tricky is that I can't even get stdout output when running with the test server, which led me to not invest any more time into debugging this. I think it's worth tracking down what's happening, but I might not be the most efficient in doing that as I don't know much of slang's code base.

csyonghe (dismissed):

csyonghe (approved):

PR #2562, opened 2022-12-09
Added support for nested calls

saipraveenb25:
IRCall insts are split by unzip into two insts (one in the primal block, and one in the diff block)
The same logic is used for all mixed instructions (like IRMakeDifferentialPair) etc.. the unzip pass converts `IRMixedDifferentialInstDecoration` into `IRDifferentialInstDecoration` by determining the appropriate primal type to use for mixed insts.

IRCall insts are then transposed by converting the callees from `IRForwardDerivative(IRFunc)` to `IRBackwardDerivative(IRFunc)` and converting inputs to pointer typed inputs where necessary.

Also removed `GetDifferential` as a gradient flavor. These can just be treated as simple gradients now with the unzip pass splitting pair-typed instructions.


source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -1578,8 +1587,15 @@ IRInst* ForwardDerivativeTranscriber::transcribe(IRBuilder* builder, IRInst* ori
                     builder->addNameHintDecoration(pair.differential, sb.getUnownedSlice());
                 }
 
-                // Tag the differential inst using a decoration.
-                builder->markInstAsDifferential(pair.differential, as<IRType>(pair.primal->getDataType()));
+                // Tag the differential inst using a decoration (if it doesn't have one)
+                if (!pair.differential->findDecoration<IRDifferentialInstDecoration>() &&
+                    !pair.differential->findDecoration<IRMixedDifferentialInstDecoration>())
+                {
```
csyonghe, line 1593: I don't think it makes sense to decorate types. Why do we need them on types? A type can be used anywhere, including in primal code so marking them as differential/mixed doesn't seem to make sense.

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -278,6 +266,116 @@ struct DiffTransposePass
...
+                        builder, 
+                        as<IRDifferentialPairType>(
+                            as<IRPtrTypeBase>(argTypes[ii])->getValueType())->getValueType());
+                    auto diffArgPtrType = builder->getPtrType(kIROp_PtrType, diffArgType);
+                    
+                    gradients.add(RevGradient(
+                        RevGradient::Flavor::Simple,
+                        fwdCall->getArg(ii),
+                        builder->emitLoad(
+                            builder->emitDifferentialPairAddressDifferential(
```
csyonghe, line 369: Seeing this code I just realized that anything of pointer type ends up as an intermediary that gets reused between primal and reverse functions, we can't really store and reuse them, and will have to do recompute. Because pointers can't be materialized in HLSL/GLSL.

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -278,6 +266,116 @@ struct DiffTransposePass
...
+                                diffArgPtrType,
+                                args[ii])), 
+                        nullptr));
+                }
+            }
+        }
+
```
csyonghe, line 376: I guess if we are going to have `primal_func(..., out Interemdiaries)`, we are going to replace `fwdCall` with a call to `primal_func` here.

source/slang/slang-ir-autodiff-transpose.h
csyonghe: The logic for generating a reverse derivative function should still be copying everything from primal, but don't use the value in the differential part of the function if there is already an intermediate available from the stored value, and then rely on DCE to remove the computations that are no longer needed due to the storage. This way we can safely preserve the computations for all intermediaries that we can't store or didn't choose to store.

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -856,6 +937,8 @@ struct DiffTransposePass
 
     IRInst* emitAggregateDifferentialPair(IRBuilder* builder, IRType* aggPrimalType, List<RevGradient> pairGradients)
     {
+        SLANG_UNEXPECTED("Should not run.");
```
csyonghe, line 940: Why this?

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -66,6 +86,185 @@ struct DiffUnzipPass
...
+
+        return false;
+    }
+
+    InstPair splitCall(IRBuilder* primalBuilder, IRBuilder* diffBuilder, IRCall* mixedCall)
+    {
+        IRBuilder globalBuilder;
+        globalBuilder.init(autodiffContext->sharedBuilder);
+
+        auto fwdCallee = as<IRForwardDifferentiate>(mixedCall->getCallee());
```
csyonghe, line 111: `callee` could have already been generated and therefore in a form of `IRFunc`, or  is there an assumption that this will always happen before generation of the derivative and replacement of the `IRForwardDifferentiate`?

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -66,6 +86,185 @@ struct DiffUnzipPass
...
+    InstPair splitVar(IRBuilder* primalBuilder, IRBuilder* diffBuilder, IRVar* mixedVar)
+    {
+        auto pairType = as<IRDifferentialPairType>(mixedVar->getDataType());
+        auto primalType = pairType->getValueType();
+        auto diffType = (IRType*) diffTypeContext.getDifferentialForType(primalBuilder, primalType);
+        
+        return InstPair(primalBuilder->emitVar(primalType), diffBuilder->emitVar(diffType));
+    }
+
+    InstPair splitReturn(IRBuilder*, IRBuilder* diffBuilder, IRReturn* mixedReturn)
```
csyonghe, line 219: Could probably use some comments to explain why splitting a return will result in returning a pair that is marked as differential here.

csyonghe (approved):

source/slang/slang-ir-autodiff-fwd.cpp
saipraveenb25: Ah, I don't think we decorate types? We decorate the instruction based on the type (this isn't implemented yet though)
but you're right that we should never be tagging types this way.

source/slang/slang-ir-autodiff-unzip.h
saipraveenb25: This assumption should hold if the function is being handled in reverse-mode (and methods called form this function do not have pre-defined fwd-mode/rev-mode functions) since we transpose the function (and the call inst) before going for the next set of functions to be transcribed.

We do need to additional support here for other cases (like pre-existing user-defined reverse-mode definitions). I'm hoping to add this in the next PR (along with additional testing)

source/slang/slang-ir-autodiff-unzip.h
saipraveenb25: Noted! I'll add more comments to all `splitXXX()` to make the intent clearer.

PR #2566, opened 2022-12-14
Avoid dots in auto-detected filename extensions

expipiplus1:
Supersedes #2532


jsmall-nvidia (commented):
Looks like needs a change in slang-options.cpp too.

tangent-vector:
FWIW there *is* precedent for file names that include multiple extensions, including good old `.tar.gz`. Another more recent precedent is `.meta` files in Unity, where for an asset file like `thing.jpg` there will also be a `thing.jpg.meta` file.

The general rule is such cases is that a `.foo.bar` file should be in the format of a `.bar` file, but encodes the data of (or information related to) a `.foo` file.

The usage we had of extensions like `.spv.asm` makes clear that a file is a text file of assembly code (`.txt`), but that it encodes the information of a SPIR-V module (extension `.spv`).

I'm fine with this behavior being removed for now, but I really do think it was the right approach to filename extensions. A file like `thing.a.b.c.` should be detected as a `.c` file, a `.b.c` file, *and* a `.a.b.c.` file. (It should *not* be detected as a `.a`, `.b`, or `.a.b` file)

One important use case I had in mind for this support was to allow for "literate mode" Slang to be encoded as `.slang.md` files, which would be properly-formatted Markdown with Slang code in the top-level code blocks. A good implementation of such a format basically *requires* a multi-part dotted extension. We need a filename to end with `.md` so that platforms like GitHub can auto-detect that the file is in Markdown, and thus render it properly when people browse the code. But we *also* need to be able to detect that something is a Slang-flavor Markdown file (`.slang.md`) in order to know that it should be processed using the special rules we give to that format.

PR #2567, opened 2022-12-14
s/TRACTING/TRACING/

### LINKED ISSUE #2561: Ray tracing misprint
https://github.com/shader-slang/slang/blob/92ae4949fe1af28ef31331fd4116c8111c057420/slang.h#L2059
https://github.com/shader-slang/slang/blob/92ae4949fe1af28ef31331fd4116c8111c057420/slang.h#L2575

Is SLANG_BINDING_TYPE_RAY_**TRACTING**\_ACCELERATION_STRUCTURE not a mistake for SLANG_BINDING_TYPE_RAY_**TRACING**\_ACCELERATION_STRUCTURE?

expipiplus1:
Closes #2561

Ordinarily it's probably just git noise to correct these things, however this
is part of the public API...

Is it necessary to add an enum link with the old name for back-compat?


jsmall-nvidia (approved):
Oops. 

Don't think we need to add backwards compat definition.

PR #2568, opened 2022-12-14
Fix code generation for matrix reshape.

csyonghe:
This fixes a regression introduced in the `construct` opcode cleanup PR.
Also adds the implementation for C++/CUDA path for matrix reshape.


source/slang/slang-emit-cpp.cpp
```diff
@@ -1178,6 +1178,7 @@ void CPPSourceEmitter::_emitInitDefinition(const UnownedStringSlice& funcName, c
     writer->emit("{ ");
 
     const Index paramCount = Index(funcType->getParamCount());
+    bool handled = false;
```
jsmall-nvidia, line 1181: Could perhaps avoid the bool/goto by having a method that implemented the output of the initialization contents. A call in that could could then return when it output, and then end with the default.

jsmall-nvidia (approved):

PR #2570, opened 2022-12-19
Add format checking attributes on printf-like functions

expipiplus1:


source/core/slang-secure-crt.h
```diff
@@ -47,6 +47,7 @@ inline size_t strnlen_s(const char * str, size_t numberOfElements)
 #endif
 }
 
+__attribute__((format(printf, 3, 4)))
```
csyonghe, line 50: Does msvc support this?

expipiplus1: haha, of course not. I'll hide it behind a macro...

csyonghe (dismissed):

PR #2571, opened 2022-12-19
Use SPIR-V opcode names rather than numbers

expipiplus1:
WIP: I think there are still a few in *.meta.slang that my awk script didn't catch.

Besides the noise the changes are in:

- source/slang/slang-emit-spirv.cpp
- tools/slang-lookup-generator/lookup-generator-main.cpp (This is a small program which extracts the spirv (or extinst) opnames from the json files in spirv-headers and generates a lookup table for them).
- premake5.lua


source/slang/slang-lookup-glslstd450.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+#include "../core/slang-common.h"
+#include "../core/slang-string.h"
+#include "spirv/unified1/GLSL.std.450.h"
+
+
+namespace Slang
+{
+
+static const unsigned tableSalt[81] ={
```
csyonghe, line 11: Maybe worth having a comment here that this is using a perfect hashing algorithm and the hash function is generated by lookup-generator, to prevent people from trying to modify this file directly.

source/slang/slang-ir-spirv-snippet.cpp
```diff
@@ -60,8 +61,58 @@ RefPtr<SpvSnippet> SpvSnippet::parse(UnownedStringSlice definition)
...
+                            throw Misc::TextFormatException(
+                                "Text parsing error: Unrecognized SPIR-V GLSLstd450 opcode: " + opName);
+                        }
+                        printf("BNBBB: %d\n", glslOpcode);
+                        return (SpvWord)glslOpcode;
+                    }
+                }
+                // fallthrough
+                default:
+                    throw Misc::TextFormatException(
```
csyonghe, line 112: We need to turn this kind of exception throwing into an proper diagnostic at some point.

csyonghe (dismissed):

PR #2587, opened 2023-01-12
Use gmake2 as a premake target over gmake

expipiplus1:
According to the documentation, `gmake` has been deprecated by `gmake2` https://premake.github.io/docs/Using-Premake/#using-premake-to-generate-project-files

gmake2 has better dependency handling around our custom rules leading in fewer runs of slang-generate etc...

I've been using gmake2 without issue.


csyonghe:
It seems that our teamcity agents are not configured to run gmake2. We need to fix the CI servers before landing this PR.

expipiplus1:
Right, MR is open for that

PR #2588, opened 2023-01-12
Fix issue around linking/obfuscation

jsmall-nvidia:
When compiling with obfuscation enabled, care is needed around stdlib linkage names. The stdlib is only compiled once and without obfuscation. The bug occurred because stdlib linkage decorations were assumed to also be obfuscated.

On searching through code to find other places where this issue occurs there is `visitInheritanceDecl` which bypasses the stdlib obfuscation mechansism. It's not clear if that is reasonable or not, so for now the issue is just commented.


expipiplus1:
Is this as general as it can be, shouldn't obfuscation be a property of the linked symbol rather than context || isStdLib?

I guess this isn't a concern right at the moment.

(what's up with the git history here also, a reused old branch name?)

jsmall-nvidia:
> Is this as general as it can be, shouldn't obfuscation be a property of the linked symbol rather than context || isStdLib?

The fix such that it is is conservative in trying to just resolve the exhibited problem. So yeah, there's some holes there.

Is obfuscation a property of the linked symbol? I mean you could take an approach where the symbol was marked as obfuscated, such that during lookup it would lookup obfuscated names. That would imply during linkage all modules which don't have obfuscation would have their symbols also available obfuscated.

That does have the advantage of being able to combine modules with only some obfuscated. It could work at a finer granularity too. This could be handled lazily, where if a lookup for an obfuscated item, would cause the obfuscated symbols for non obfuscated items added to the table.

>(what's up with the git history here also, a reused old branch name?)

Err yeah :| It's probably due to my process around using source tree/git, so should do something about that...

csyonghe (approved):
The problem is general and not just related to stdlib: if the user compiles one module that defines an interface without obfuscation, and another module that implements the interface with obfuscation, we will have problems. The simplest solution I can think of is to not obfuscate anything inside an interface.

PR #2596, opened 2023-01-16
Added switch-case support; fixed non-diff parameter transposition

saipraveenb25:
Added basic switch-case support for both forward and reverse-mode. For now, we require that every case label has an explicit break (no fall-throughs, but grouped labels are okay). We also do not have a proper warning/error for this situation right now.

Also fixed an issue where non-differentiable parameters were moved to the end of the parameter block, 
causing the generated function to not be consistent with it's type.

Added test case for simple switch, will expand on this with more complicated 'break' styles in a follow-up PR


source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -784,6 +784,47 @@ InstPair ForwardDiffTranscriber::transcribeLoop(IRBuilder* builder, IRLoop* orig
...
+    // Transcribe condition (primal only, conditions do not produce differentials)
+    auto primalCondition = findOrTranscribePrimalInst(builder, origSwitch->getCondition());
+    SLANG_ASSERT(primalCondition);
+    
+    // Transcribe 'default' block
+    IRBlock* diffDefaultBlock = as<IRBlock>(
+        findOrTranscribeDiffInst(builder, origSwitch->getDefaultLabel()));
+    SLANG_ASSERT(diffDefaultBlock);
+
+    // Transcribe 'default' block
```
csyonghe, line 798: break block

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -784,6 +784,47 @@ InstPair ForwardDiffTranscriber::transcribeLoop(IRBuilder* builder, IRLoop* orig
...
+
+    auto diffSwitchInst = builder->emitSwitch(
+        primalCondition,
+        diffBreakBlock,
+        diffDefaultBlock, 
+        diffCaseValuesAndLabels.getCount(),
+        diffCaseValuesAndLabels.getBuffer());
+    builder->markInstAsMixedDifferential(diffSwitchInst);
+
+    return InstPair(diffSwitchInst, diffSwitchInst);
```
csyonghe, line 825: We are returning the diff inst as primal inst for all control flow insts. Should this be changed to original inst?

csyonghe (approved):

PR #2608, opened 2023-01-25
Overhauled reverse-mode control flow handling

saipraveenb25:
Added pre-autodiff CFG Normalization pass that brings the graph into 'reversible' form, by eliminating multi-exit regions in loops and switch-case statements through intermediate control variables.

Added CFG reversal pass that is run after transposition. Traverses the fwd-mode graph and reverses all control-flow nodes 1:1 (unconditional with unconditional, if-else with if-else, etc..)

Single-iteration loops (all branches lead to a break) are now supported

For full loop support (not tested yet):

Added loop unzipping step that splits loops into primal and differential loops, adds appropriate counter variables, then replaces primal insts with a store/load from an array.

Added `IRLoopCounterDecoration` to mark insts that are to be ignored during the transposition step.

Full loop support is close, but not done yet. What's left:

A mechanism for the transposition pass to ignore loop counters, and intermediate array loads (reverse only their position but not the instruction itself)


source/slang/slang-ir-autodiff-cfg-norm.cpp
```diff
@@ -0,0 +1,624 @@
...
+struct CFGNormalizationContext
+{
+    SharedIRBuilder* sharedBuilder;
+    DiagnosticSink*  sink;
+};
+
+
+IRBlock* getOrCreateTopLevelCondition(IRLoop* loopInst)
+{
+    // For now, we're going to naively assume the next block is the condition block.
```
csyonghe, line 55: This assumption won't hold for do-while loops. Let's add a test for that later.

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -327,6 +664,188 @@ struct DiffUnzipPass
...
+        indexRegionMap[targetLoop->getTargetBlock()] = region;
+        indexRegions.add(region);
+    }
+
+    // Deallocate regions
+    void cleanupIndexRegionInfo()
+    {
+        for (auto region : indexRegions)
+        {
+            delete region;
```
csyonghe, line 693: Can we use RefPtr<IndexRegion> instead of doing new and deletes?

source/slang/slang-ir.cpp
```diff
@@ -6426,7 +6452,14 @@ namespace Slang
...
 #if _DEBUG
-        validateIRInstOperands(this);
+        // Disabling this for now. There's a case in 
+        // constructSSA that clones decorations on a param
+        // that hasn't yet been inserted into a block.
+        // This causes validation to fail because the decoration
+        // type is in IRModule, but the param is not nested under the 
+        // module yet.
+        // 
+        // validateIRInstOperands(this);
```
csyonghe, line 6462: This validation is here to make sure we don't introduce invalid IR during the autodiff process. Is there a way to keep it here? e.g. inserting the types referenced by the decoration first.

tests/autodiff/reverse-single-iter-loop.slang
```diff
@@ -0,0 +1,99 @@
...
+            {
+                x = y * 10.0f;
+                break;
+            }
+            else
+            {
+                x = y * 4.0f;
+                break;
+            }
+        }
```
csyonghe, line 51: We need to also test for statements that appear after the converge point of a loop, like here. This is where we should be inserting guarding `if`s.

source/slang/slang-ir.cpp
csyonghe: If not possible, use `disableIRValidationAtInsert()` before inserting the violating inst and `enableIRValidationAtInsert()` after inserting.

csyonghe (changes requested):

csyonghe:
Another test we should now that single iteration loops are supported is to test for functions with multiple return statements (some of which are nested in `if`s or `for`s). The single-return transformation should kick in to turn the func into a single iteration loop, followed by a multi-level-break elimination pass to remove any multi-level breaks.

source/slang/slang-ir-autodiff-rev.cpp
```diff
@@ -568,7 +569,15 @@ namespace Slang
             tempBuilder.setInsertBefore(diffPropagateFunc);
         }
 
-        auto fwdDiffFunc = generateNewForwardDerivativeForFunc(&tempBuilder, primalFunc, diffPropagateFunc);
+        // Normalize the CFG of the primal func before starting the reverse-mode pass.
+        IRCloneEnv cfgNormCloneEnv;
```
csyonghe, line 573: After merging with top of tree, you will see that `generateNewForwardDerivativeFunc` calls `prepareForBackwardAutoDiff` which makes a clone of the primal func and then makes some preparational changes to the clone of primal func before applying forward diff to it, including single-return transform, multi-level-break elimination, and the array/struct type SSA transform. The `normalizeCFG` step should be inserted there instead, and we need to make sure it is called before the array/struct type SSA step to avoid undoing the SSA transforms since normalizeCFG calls `elminatePhi` to get out of SSA first.

source/slang/slang-ir.cpp
saipraveenb25: Removed this change, and surrounded the `constructSSA()` with `disableIRValidationAtInsert()`  and `enableIRValidationAtInsert()`

source/slang/slang-ir-validate.cpp
```diff
@@ -199,6 +199,13 @@ namespace Slang
         if(inst->getFullType())
             validateIRInstOperand(context, inst, &inst->typeUse);
 
+        // Avoid validating decoration operands
```
csyonghe, line 202: Can we remove this now that validation has been supressed at higher level?

csyonghe (approved):

PR #2614, opened 2023-01-27
Add ASAN support + fixes

skallweitNV:
- Add support for ASAN (MSVC for now)
- Enable using `--enable-asan` option in premake
- Fix ASAN false positives when deleting `StringRepresentation` objects.
- Fix deep recursion in `slang-generate`.
- Fix `hello-world` and `gpu-printing` examples.

With this, all the examples pass when running debug build with ASAN enabled. Unfortunately when switching examples to use Vulkan, ASAN seems to break the loader. Needs more investigation.

### DIFF
--- a/source/core/slang-string.h
+++ b/source/core/slang-string.h
@@ -321,7 +321,9 @@ namespace Slang
             /// does not match deallocation size (due variable sized string payload).
         void operator delete(void* p)
         {
-            delete(p);
+            if (!p) return;
+            StringRepresentation* str = (StringRepresentation*) p;
+            ::operator delete(p, sizeof(StringRepresentation) + str->capacity + 1);
         }        
     };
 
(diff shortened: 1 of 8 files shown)


source/core/slang-string.h
```diff
@@ -315,6 +315,15 @@ namespace Slang
 
             return cloneWithCapacity(newCapacity);
         }
+
+            /// Overload delete to silence ASAN new-delete-type-mismatch errors.
+            /// These occur because the allocation size of StringRepresentation
+            /// does not match deallocation size (due variable sized string payload).
+        void operator delete(void* p)
```
skallweitNV, line 322: We could wrap this in a `#ifdef __SANITIZE_ADDRESS__` to only have it when ASAN is enabled.

csyonghe (approved):

PR #2633, opened 2023-02-08
Add backward derivatives for functions in diff.meta.slang

winmad:
* Add backward derivatives for a few stdlib functions.
* Add test cases for backward differentiation.
* Update the autodiff doc accordingly.


csyonghe:
The documentation changes are not included in the branch, did you forget to check them in?

source/slang/diff.meta.slang
```diff
@@ -184,18 +205,68 @@ DifferentialPair<vector<T,N>> mul(DifferentialPair<matrix<T,N,M>> left, Differen
...
+    for (int r = 0; r < R; ++r)
+        for (int n = 0; n < N; ++n)
+            left_d_result[r][n] = T(0.0);
+    
+    matrix<T, N, C>.Differential right_d_result;
+    for (int n = 0; n < N; ++n)
+        for (int c = 0; c < C; ++c)
+            right_d_result[n][c] = T(0.0);
+    
+    for (int r = 0; r < R; ++r)
```
csyonghe, line 255: Can this be expressed as a single matrix multiply?

source/slang/diff.meta.slang
```diff
@@ -494,3 +713,17 @@ DifferentialPair<T> __d_dot(DifferentialPair<vector<T, N>> dpx, DifferentialPair
     }
     return DifferentialPair<T>(result, d_result);
 }
+
+__generic<T : __BuiltinFloatingPointType, let N : int>
+[BackwardDerivativeOf(dot)]
+void __d_dot(inout DifferentialPair<vector<T, N>> dpx, inout DifferentialPair<vector<T, N>> dpy, T.Differential dOut)
+{
+    vector<T, N>.Differential x_d_result, y_d_result;
+    for (int i = 0; i < N; ++i)
```
csyonghe, line 722: Can this be expressed as a single (vector * scalar) operation instead of using an explicit loop?

csyonghe (commented):
Looks good. It seems some of the matrix/vector implementation can be simplified by calling into existing functions.

We should also consider adding in the other vector/math functions, like `log2`, `log10`, `cross`, `length`, `distance`, `reflect`, `refract`, `normalize`, `transpose`, `saturate`, `lerp`, `fma`, `mad` and the remaining trignometry functions.

csyonghe (approved):

PR #2635, opened 2023-02-08
Reverse-mode Loop Support

saipraveenb25:
Basic `for` loops now work in reverse-mode. 

The unzip pass splits `for` loops into two and inserts counter variables into each to keep track of iterations.

Arrays are allocated to store/retrieve intermediate primal insts.

Added a `[maxiters(count)]` decoration to control the size of the allocated arrays.

CFG reversal pass then flips the control flow for the differential loop

A major issue right now is that the unzip pass inserts 'inverted' instructions into seemingly random locations, which don't make sense until the flow reversal actually happens. Follow-up PR will restructure this to avoid this mixing of passes.

(We also need more complicated loop tests)


csyonghe (commented):
Let's exclude the changes from slang-lookup-glslstd450.cpp and slang-lookup-spvop.cpp in this PR.

After refactoring the loop implementation, we should add more extensive tests to cover things like `break` `continue` and nested if/loops/etc.

csyonghe:
Would you also update the documentation to talk about limitations or implications on loops, and the requirement on the `[MaxIters]` attribute?

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -262,19 +280,41 @@ struct DiffUnzipPass
 
     IRBlock* getUpdateBlock(IndexedRegion* region)
     {
+        // TODO: What if the 'continue' region has multiple
```
csyonghe, line 283: Add a release assert here to make sure the continue block ends with a unconditional branch back to loop head. If this is not the case we should just report failure/crash, so it is easy to fix in the future.

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -262,19 +280,41 @@ struct DiffUnzipPass
...
+        // 
         return region->continueBlock;
     }
+
+    IRBlock* getFirstLoopBodyBlock(IndexedRegion* region)
+    {
+        // Grab the 'condition' block.
+        auto condBlock = region->firstBlock;
+
+        SLANG_ASSERT(as<IRIfElse>(condBlock->getTerminator()));
```
csyonghe, line 295: Make it a release assert. This is not true for do-while loops.

csyonghe (approved):

source/slang/core.meta.slang
```diff
@@ -2823,6 +2823,9 @@ attribute_syntax [fastopt]              : FastOptAttribute;
 __attributeTarget(LoopStmt)
 attribute_syntax [allow_uav_condition]  : AllowUAVConditionAttribute;
 
+__attributeTarget(LoopStmt)
+attribute_syntax [maxiters(count)]      : MaxItersAttribute;
```
csyonghe, line 2827: The naming convention for new attributes is BigCamel. The existing all lower case attribute names are for HLSL compatibility.

csyonghe (approved):

docs/user-guide/07-autodiff.md
```diff
@@ -567,6 +567,6 @@ The compiler can generate forward derivative and backward propagation implementa
...
 - All operations to global resources, including texture reads or atomic writes, are treating as a non-differentiable operation.
 - If a differentiable function contains calls that cause side-effects such as updates to global memory, there will not be a guarantee on how many times the side-effect will occur during the resulting derivative function or back-propagation function.
-- All loops in a backward differentiable function must end within a statically known number of iterations. If the maximum number of iterations is not trivially deductible by the type system as a compile-time constant, a manually attribute is needed at the loop to provide the number. If the number of actually executed iterations exceeds what is being specified, the resulting runtime behavior is undefined.
+- `for` loops: In a backward differentiable function, loops currently cannot have `continue` statements although `break` statements are supported. Loops must use the attribute `[maxiters(<count>)]` to specify a maximum number of iterations. This will be used by compiler to allocate space to store intermediate data. If the actual number of iterations exceeds the provided maximum, the behaviour is undefined.
```
csyonghe, line 570: [MaxIters]

csyonghe: behavior (assuming american english?)

csyonghe (approved):

csyonghe (approved):

Issue #2324, opened 2022-07-13
Optimization level ignored in TargetDesc


jsmall-nvidia:
The optimization level set in `TargetDesc`s defined in slang.h appears to be ignored. 

`TargetDesc` can be passed in as part of the `SessionDesc` as part of global session construction. Some fields are honored from TargetDesc such as the floating point mode. But the optimization level appears to not be used (it's taken from the linkage in slang-compiler.cpp), which can be set by another API call (EndToEndCompileRequest::setOptimizationLevel)

PR #2649, opened 2023-02-14
Take into account the optimization level requested in TargetDesc

expipiplus1:
Combines conflicting optimization levels by taking the maximum between them; as such they now each represent the lower bound on the eventual optimization level.

Most of the changes are making this clear by s/optimizationLevel/minimumOptimizationLevel

Closes https://github.com/shader-slang/slang/issues/2324

It seems like there's a mixed usage of optimization level, in
TargetDesc to ensure that some optimization features are enabled,
and in Linkage/ICompileRequest to represent the user's request.
Might it be simpler just to remove the optimization level specified
in TargetDesc and implement what that's driving with a smaller
hammer?

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -3544,9 +3544,12 @@ namespace slang
             SlangDebugInfoLevel     level) = 0;
 
             /*!
-            @brief Set the level of optimization to perform.
+            @brief Set the minimum level of optimization to perform.
+            Other compiler settings such as the optimization level requested in
+            TargetDesc might increase the eventual optimization above the value
+            specified here.
             */
-        virtual SLANG_NO_THROW void SLANG_MCALL setOptimizationLevel(
+        virtual SLANG_NO_THROW void SLANG_MCALL setMinimumOptimizationLevel(
             SlangOptimizationLevel  level) = 0;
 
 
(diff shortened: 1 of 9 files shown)


jsmall-nvidia:
@expipiplus1 I share your concern that keeping both optimization levels seems somewhat confusing. Your change helps in making the API more clear that it isn't directly setting the actual optimization level. It will still be somewhat confusing as now API user will have to learn about how different methods combine to produce the actual optimization level.

My immediate reaction is that we should just have one optimization level setting and that's set on the request. That said I'm actually not entirely sure *why* there is an optimization level set on the TargetDesc. It has caused problems, and confusion, so unless there is a compelling reason to keep it (which I'm not aware of) I think removing it is the clearest path.

tangent-vector:
I believe there is (or at least *was*) a good rationale for having an optimization level on the `TargetDesc`. In cases where one is compiling for multiple targets (e.g., DXIL and SPIR-V in a single pass), it is possible that the desired optimization level/settings might vary per-target (e.g., you might need to apply a higher optimization to your SPIR-V path to match the trade-off between compile-time and runtime performance that you get on the other path).

Ideally we will change the way compilation options are stored so that each relevant object in the user/API mental model (targets, sessions, etc.) has a key/value collection of compilation options (easily allowing us to detect options that haven't been set). When evaluating the effective value of a compilation option, we would then have a notion of "inheritance" of options where, e.g., the target inherits options from the linkage/session (but options set on the target override those in the linkage).

expipiplus1:
OK, in that case I think we're more or less in agreement. I'll remove the optimization level from targetdesc for now (it's obviously not being relied upon). https://github.com/shader-slang/slang/issues/2655 So we don't mistake this as the end of the line for target driven flags.

PR #2653, opened 2023-02-15
SPIR-V SER tests

jsmall-nvidia:
* Enable Shader Execution Reordering (SER) tests for output of SPIR-V
* Update SER doc to say SPIR-V is supported.


csyonghe:
Is there a reason we need to turn the expected result into SPIRV?
The text comparison is already very annoying to maintain every time we make changes to the compiler so that the generated result is different. We have to manually inspect the test to know whether or not the change in result is actually OK.
Turning this into SPIRV makes it even harder to tell if a different result is actually correct or not.

expipiplus1:
I wonder if we could make do with some less stringent matching for these. In lots of these sorts of tests we only care about the structure of a few salient lines, the tests could look something like

```
              <a>:   (fvec3) Load XXX
              <b>:   (float) Load XXX
                              HitObjectRecordMissNV <c>(hitObj_1) <a> <b> XXX XXX XXX
              XXX:   (bool) HitObjectIsMissNV <c>(hitObj_1)

```

csyonghe:
Yes, that is why I am been arguing that we should include FileCheck.exe in our llvm build and use it to do pattern based checks instead of plain string comlarisons. Maintaining these tests is becoming an increasingly expensive burden whenever we change our IR generation logic.

PR #2659, opened 2023-02-17
Add static for loop iteration inference.

csyonghe:
This PR adds a inference step when checking `for` loops to try to statically determine the maximum number of iterations for the loop and insert a `[MaxIters]` attribute on the loop if inference suceeds.

It also reports warnings along the way if it detects suspicious code like `for (int i = 0; i < 5; j++)`.

### DIFF
--- a/source/slang/slang-check-stmt.cpp
+++ b/source/slang/slang-check-stmt.cpp
@@ -1,5 +1,6 @@
 // slang-check-stmt.cpp
 #include "slang-check-impl.h"
+#include "slang-ir-util.h"
 
 // This file implements semantic checking logic related to statements.
 
@@ -158,17 +159,20 @@ namespace Slang
 
     void SemanticsStmtVisitor::visitDoWhileStmt(DoWhileStmt *stmt)
     {
+        checkModifiers(stmt);
         WithOuterStmt subContext(this, stmt);
 
         stmt->predicate = checkPredicateExpr(stmt->predicate);
         subContext.checkStmt(stmt->statement);
+        checkLoopInDifferentiableFunc(stmt);
     }
 
     void SemanticsStmtVisitor::visitForStmt(ForStmt *stmt)
     {
         WithOuterStmt subContext(this, stmt);
-
+        checkModifiers(stmt);
         checkStmt(stmt->initialStatement);
+
         if (stmt->predicateExpression)
         {
             stmt->predicateExpression = checkPredicateExpr(stmt->predicateExpression);
@@ -178,6 +182,10 @@ namespace Slang
             stmt->sideEffectExpression = CheckExpr(stmt->sideEffectExpression);
         }
         subContext.checkStmt(stmt->statement);
+        
+        tryInferLoopMaxIterations(stmt);
+
+        checkLoopInDifferentiableFunc(stmt);
     }
 
     Expr* SemanticsVisitor::checkExpressionAndExpectIntegerConstant(Expr* expr, IntVal** outIntVal)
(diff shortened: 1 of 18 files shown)


source/slang/slang-check-stmt.cpp
```diff
@@ -327,6 +329,260 @@ namespace Slang
...
+            initialVal = assignExpr->right;
+        }
+        else
+            return;
+
+        auto initialLitVal =
+            as<ConstantIntVal>(tryFoldIntegerConstantExpression(initialVal, nullptr));
+
+
+        ConstantIntVal* finalVal = 0;
```
jsmall-nvidia, line 363: nullptr?

source/slang/slang-check-stmt.cpp
```diff
@@ -327,6 +329,260 @@ namespace Slang
         stmt->expression = CheckExpr(stmt->expression);
     }
 
+    void SemanticsStmtVisitor::tryInferLoopMaxIterations(ForStmt* stmt)
+    {
+        DeclRef<Decl> initialVar = {};
+        DeclRef<Decl> predicateVar = {};
+        Expr* initialVal = nullptr;
+        if (auto varStmt = as<DeclStmt>(stmt->initialStatement))
```
jsmall-nvidia, line 337: Might benefit with some comments saying what this does and doesn't handle. 
I'm assuming doing (say) ```for (int i = 1, j = 2;... )``` is something it just doesn't handle, but I'm not 100% sure what that maps down to in AST.

source/slang/slang-check-stmt.cpp
```diff
@@ -327,6 +329,260 @@ namespace Slang
...
+        }
+        else
+        {
+            return;
+        }
+        if (binaryExpr->arguments.getCount() != 2)
+            return;
+        auto leftCompareOperand = binaryExpr->arguments[0];
+        auto rightCompareOperand = binaryExpr->arguments[1];
+        while (auto typeCast = as<TypeCastExpr>(leftCompareOperand))
```
jsmall-nvidia, line 390: Stripping the cast/s seems as if it could be a little dangerous? Say in a rather contrived example
```
for (int i = 0x101; uint8_t(i) < uint8_t(0xffffff); ++i) { ... }
```

source/slang/slang-check-stmt.cpp
csyonghe: Fixed.

source/slang/slang-check-stmt.cpp
csyonghe: Added a comment at the top of the function.

source/slang/slang-check-stmt.cpp
csyonghe: You are right, removed.

source/slang/slang-check-stmt.cpp
```diff
@@ -327,6 +329,260 @@ namespace Slang
...
+                return;
+            }
+        }
+        else
+        {
+            getSink()->diagnose(stmt->sideEffectExpression, Diagnostics::forLoopNotModifyingIterationVariable, initialVar);
+            return;
+        }
+
+        if (!initialLitVal || !finalVal)
```
jsmall-nvidia, line 536: Don't we need to worry about side effects in the body?

csyonghe: Yeah, for that we will need to track the use of the induction variable when we check the body. Will implement it next.

csyonghe: This is now implemented.

saipraveenb25 (approved):

source/slang/slang-check-stmt.cpp
jsmall-nvidia: I didn't see a test that checked for break in the body?

csyonghe: Break is fine because it only makes the loop terminate earlier. As long as the result we derive is conservative things are good.

PR #2667, opened 2023-02-20
Added support for simple while loops

saipraveenb25:
Improved the CFG normalization pass to add a continue block to loops if one is not already available.

Added a simple test case that is equivalent to the for loop case

More complicated test cases coming soon


source/slang/slang-ir-autodiff-cfg-norm.cpp
```diff
@@ -473,14 +473,18 @@ struct CFGNormalizationPass
                     loopEndPoint = falseEndPoint;
                     isLoopOnTrueSide = false;
                 }
+
```
csyonghe, line 476: I am little worried about the assumptions being made here regarding the structures of the loop body here. If a user writes something like `for(;;) {}`, the loop body can be in any form so we won't even be able to map blocks into condition blocks or continue blocks. I wonder if the code is handling thoses cases.

csyonghe (approved):

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -277,14 +277,26 @@ struct DiffUnzipPass
...
-        // We ideally want the _last_ block before control loops back.
+        auto initBlock = getInitializerBlock(region);
+
+        auto condBlock = region->firstBlock;
+
+        IRBlock* lastLoopBlock = nullptr;
+
+        for (auto predecessor : condBlock->getPredecessors())
+        {
+            if (predecessor != initBlock)
```
csyonghe, line 288: Maybe assert when there are more than one predecessor other than the loop header?

csyonghe: We should also add a call to `eliminateContinueBlocks` in `prepareForBackwardAutoDiff`.

saipraveenb25: Yup, I'm prepping some tests for continue, I'll make this addition in that PR

source/slang/slang-ir-autodiff-cfg-norm.cpp
saipraveenb25: I'm working on removing the assumption of having a condition block (by inserting a dummy condition like `if(true)`).

After that the only requirement should be that the first block after block with the loop terminator is part of the loop itself

PR #2680, opened 2023-02-27
Detect and deduplicate read-only resource access.

csyonghe:
Mark all read-only resource intrinsics as `[__readNone]`.

Deduplicate `load`s of `ConstantBuffer` and `ParameterBlock`.


source/slang/core.meta.slang
```diff
@@ -1189,6 +1189,7 @@ for (int tt = 0; tt < kBaseTextureTypeCount; ++tt)
                     continue;
                 }
 
+                bool isReadOnly = (access == SLANG_RESOURCE_ACCESS_READ);
```
jsmall-nvidia, line 1192: Is there a possible issue with aliasing? The [__readNone] attribute in the context of a resource method would mean the methods result is entirely determined by its inputs, with the not unreasonable assumption that if it's a read only resource the resources contents can't change.

But if it is possible to mutate the contents of a read only resource, via a writable alias on some aspect of the resource then that seems problematic.

csyonghe: The HLSL compiler assumes no aliasing by default, and would require the user to pass in a specific flag to inform the compiler that there is aliasing and should not perform any optimizations.

We don't have such a flag yet in Slang, and we should probably introduce it.

csyonghe: Given that our dxc backend never generates the dxc flag to enable aliasing, I guess it is safe for us to assume there is no such uses out there in our user codebase. But we should definitely inform people about this change.

jsmall-nvidia: Ok sounds good.

jsmall-nvidia (dismissed):

PR #2681, opened 2023-02-27
Diagnose on storing differentiable value into non-differentiable location

csyonghe:
Add diagnostics to prevent unintended loss of derivative when assigning a value with derivative into a location that can't take in derivatives. The user is expected to use `detach` to clarify the intention.

For example,
```
struct DiffT : IDifferentiable
{
    float f;
}
struct NoDiffField
{
    DiffT fp;
}

[BackwardDifferentiable]
float g(float x)
{
    NoDiffField obj;
    obj.fp.f = x * x; // Error, this location cannot hold derivative.
    return obj.fp.f;
}
```


saipraveenb25 (approved):
Looks good! 
Should make it significantly easier to use DiffSlang with this diagnostic

PR #2682, opened 2023-02-28
Add `SourceBlobWithPathInfoArtifactRepresentation`.

csyonghe:
If the user has provided both a path and a source blob, we should use the user provided source blob instead of loading it from file. In this case we still need to preserve the path info, so we use the `SourceBlobWithPathInfoArtifactRepresentation` class to do that.

### DIFF
--- a/source/core/slang-io.cpp
+++ b/source/core/slang-io.cpp
@@ -286,7 +286,23 @@ namespace Slang
             // If ioBuilder doesn't end in a delimiter, add one
             if (!isDelimiter(ioBuilder[ioBuilder.getLength() - 1]))
             {
-                ioBuilder.append(kPathDelimiter);
+                // Determine the preferred delimiter to use based on existing path.
+                char preferedDelimiter = kOSCanonicalPathDelimiter;
+                if (kOSAlternativePathDelimiter != preferedDelimiter)
+                {
+                    // If we found the existing path uses the alternative delimiter, we will
+                    // use that instead of the canonical one.
+                    constexpr Index kMaxDelimiterSearchRange = 32;
+                    for (Index i = 0; i < Math::Min(kMaxDelimiterSearchRange, ioBuilder.getLength()); i++)
+                    {
+                        if (ioBuilder[i] == kOSAlternativePathDelimiter)
+                        {
+                            preferedDelimiter = kOSAlternativePathDelimiter;
+                            break;
+                        }
+                    }
+                }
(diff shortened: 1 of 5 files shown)


source/core/slang-io.cpp
```diff
@@ -286,7 +286,23 @@ namespace Slang
             // If ioBuilder doesn't end in a delimiter, add one
             if (!isDelimiter(ioBuilder[ioBuilder.getLength() - 1]))
             {
-                ioBuilder.append(kPathDelimiter);
```
jsmall-nvidia, line 289: Perhaps we should have a function that given a builder (and/or a slice) returns the delimiter type. Then we add a parameter to `append` that takes the delimiter, with the default being the usual.

That would move the logic of determining the used delimiter outside of this function, such that it is used when it's needed - which I'm assuming (?) is not common.

source/core/slang-io.cpp
```diff
@@ -286,7 +286,23 @@ namespace Slang
...
+                    // If we found the existing path uses the alternative delimiter, we will
+                    // use that instead of the canonical one.
+                    constexpr Index kMaxDelimiterSearchRange = 32;
+                    for (Index i = 0; i < Math::Min(kMaxDelimiterSearchRange, ioBuilder.getLength()); i++)
+                    {
+                        if (ioBuilder[i] == kOSAlternativePathDelimiter)
+                        {
+                            preferedDelimiter = kOSAlternativePathDelimiter;
+                            break;
+                        }
```
jsmall-nvidia, line 302: I suppose I'm not 100% clear on the desire to use an alternate delimiter if it's already seen? It would be a little odd to see a path with different delimiters. In any kind of lookup/hashing paths should be in one style (presumably the normal delimiter).

csyonghe: This is to preserve the current compiler behavior as much as possible. We can always revise this and perform canonicalization upon combine, but I think it is safer to just preserve the existing style.

csyonghe: For example, if the combine("https://a", "b") should result "https://a/b") even on windows. The easiest way to make sure such case to work is to detect and use existing delimiters.

source/core/slang-io.cpp
csyonghe: We can clean this up. For now I am going to merge this PR since it fixes a language server bug that impacts many users.

jsmall-nvidia: Ok no worries.

PR #2683, opened 2023-02-28
AD: Fixed do-while loops

saipraveenb25:
This fix is somewhat hacky (and hopefully temporary).

The actual issue might be with `cloneInst()` which seems to have issues with out-of-order blocks that access insts that haven't been cloned yet.


source/slang/slang-ir-autodiff-cfg-norm.cpp
```diff
@@ -357,6 +365,36 @@ struct CFGNormalizationPass
...
+                        // want to make sure the afterBlock is
+                        // firmly _inside_ the current region.
+                        // If it's part of the parent, add a 
+                        // dummy block.
+                        // 
+                        if (afterBlocks.contains(afterBlock))
+                        {
+                            auto newAfterBlock = builder.emitBlock();
+
+                            // TODO: This is a hack. Ideally we should be putting
```
csyonghe, line 378: Can we just add the block to cloneEnv before cloning so cloneInst will be able to find the right operand?

csyonghe (approved):

source/slang/slang-ir-autodiff-cfg-norm.cpp
saipraveenb25: So the cloning happens much later (during the intermediates splitting pass). we would need to be able to detect the problem blocks at that stage and add them in.
The issue is that if the block accessing the parameter appears before the block with the parameter (in the child-list order), then cloneInst() assumes that there is no clone for that parameter (since it hasn't seen it yet).

I think it might make sense to augment cloneInst by keeping a list of 'incomplete' uses, and whenever we make a clone, we can go through the list and update previously seen insts.

csyonghe: I see. That would work. An alternative without changing cloneInst is to go through all the IRParams and register them into cloneEnv and map them to some placeholders, the later replace those replace holders if we can't actually clone those params early.

source/slang/slang-ir-autodiff-cfg-norm.cpp
```diff
@@ -232,7 +240,7 @@ struct CFGNormalizationPass
                 breakFlagValue,
                 block,
                 afterSplitAfterBlock,
-                afterSplitAfterBlock);
+                afterSplitAfterBlock);
```
csyonghe, line 243: Not needed for this PR, but in the future we should all form the habit of self reviewing the PR and remove spurious changes like this to keep the change set clean.

csyonghe (approved):

PR #2689, opened 2023-03-07
Bottleneck DeclRef creation through ASTBuilder.

csyonghe:
This is a preliminary step to refactor the representation of `DeclRef`.

With this change all creation of non-trivial DeclRef are done through the ASTBuilder. This allows us to use ASTBuilder to create `Val`s to represent non-trivial cases of decl refs in the future.


source/slang/slang-ast-builder.h
```diff
@@ -114,16 +114,21 @@ class ASTBuilder : public RefObject
...
+            values.nodeOperand[0] = nullptr;
+            values.nodeOperand[1] = nullptr;
+        }
+        NodeOperand(NodeBase* node) { values.nodeOperand[0] = node; values.nodeOperand[1] = nullptr; }
         template<typename EnumType>
         NodeOperand(EnumType intVal)
         {
             static_assert(sizeof(EnumType) <= sizeof(values), "size of operand must be less than pointer size.");
-            values.intOperand = 0;
+            values.intOperand[0] = 0;
```
jsmall-nvidia, line 130: Hm. Doesn't this assume little endian? If enum type is less than the value int64_t. Wouldn't it be preferable to just use a cast? Care might be needed with sign extension tho.

source/slang/slang-ast-builder.h
```diff
@@ -114,16 +114,21 @@ class ASTBuilder : public RefObject
     {
         union
         {
-            NodeBase* nodeOperand;
-            int64_t intOperand;
+            NodeBase* nodeOperand[2];
```
jsmall-nvidia, line 117: Why are there 2 of these now?

source/slang/slang-check-decl.cpp
```diff
@@ -4837,7 +4838,7 @@ namespace Slang
         derivativeAttr->loc = derivativeOfAttr->loc;
         auto outterGeneric = visitor->GetOuterGeneric(funcDecl);
         auto declRef =
-            DeclRef<Decl>((outterGeneric ? (Decl*)outterGeneric : funcDecl), nullptr);
+            visitor->getASTBuilder()->getSpecializedDeclRef<Decl>((outterGeneric ? (Decl*)outterGeneric : funcDecl), nullptr);
```
jsmall-nvidia, line 4841: Not a problem with PR but presumably outerGeneric?

jsmall-nvidia (dismissed):
Looks good.

source/slang/slang-ast-builder.h
csyonghe: This isn't ideal, but we need these to be able to encode a `DeclRef`, which is currently containing two pointers.

After the DeclRef representation cleanup, it will be shrinked down to just one pointer so we can scale this back.

PR #2691, opened 2023-03-09
Add support for `[PrimalSubstitute]` and `[PrimalSubstituteOf]`.

csyonghe:
This change adds two new attributes for autodiff:
- `[PrimalSubstitute]`: associates a function with an alternate function that will be used to replace the original function for autodiff purpose. If the alternate function is differentiable, this will make the compiler to treat the original function as differentiable. If the alternate function is not differentiable, it will still be called as the primal function in generated derivative propagation code.
- `[PrimalSubstituteOf]`: associates an alternate function for autodiff purpose with an original function.

This allows us to implement differentiable stdlib even easier. Before this change, we implement `sincos` as:
```
[BackwardDifferentiable]
void __sincos_impl(T x, out T s, out T c)
{
    s = sin(x);
    c = cos(x);
}
__generic<T: __BuiltinFloatingPointType>
[ForwardDerivativeOf(sincos)]
[ForceInline]
void __d_sincos(DifferentialPair<T> x, out DifferentialPair<T> s, out DifferentialPair<T> c)
{
    __fwd_diff(__sincos_impl)(x, s, c);
}
__generic<T: __BuiltinFloatingPointType>
[BackwardDerivativeOf(sincos)]
[ForceInline]
void __d_sincos(inout DifferentialPair<T> x, T.Differential dS, T.Differential dC)
{
    __bwd_diff(__sincos_impl)(x, dS, dC);
}
```

With `[PrimalSubstituteOf]`, we can simplify it down to:
```
[BackwardDifferentiable]
[PrimalSubstituteOf(sincos)]
void __sincos_impl(T x, out T s, out T c)
{
    s = sin(x);
    c = cos(x);
}
```


saipraveenb25 (approved):
This looks good!
We might need to be careful about explaining primal substitutes that are also differentiable in the User Guide

PR #2704, opened 2023-03-16
Support GL_EXT_fragment_shader_barycentric

skogler:
Hey,

this is my first PR in here. I tried to follow conventions and change as little as possible. 

Using the capability system, you can keep the old behavior, but the default is to use the vendor-neutral GL_EXT_fragment_shader_barycentric extension.

I had to forward the `CodeGenContext` to the GLSL emitting functions in glsl-legalize, which is why there are more changes than I necessarily wanted.

Let me know what you think!

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2274,10 +2274,9 @@ matrix<T, N, M> fwidth(matrix<T, N, M> x)
 __generic<T : __BuiltinType>
 [__readNone]
 __target_intrinsic(hlsl)
-__target_intrinsic(glsl, "$0[$1]")
+__target_intrinsic(GL_NV_fragment_shader_barycentric, "$0[$1]")
+__target_intrinsic(GL_EXT_fragment_shader_barycentric, "$0[$1]")
 __glsl_version(450)
-__glsl_extension(GL_NV_fragment_shader_barycentric)
-__glsl_extension(GL_EXT_fragment_shader_barycentric)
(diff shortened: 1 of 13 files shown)


csyonghe (dismissed):
Thanks for your contribution! This looks good.

skogler:
There is an issue with the HLSL definitions, where slang generates glsl requiring both extensions.

I'll take a look later today and see if I can fix it.

skogler:
Is there a way to conditionally require extensions with the defs in `hlsl.meta.slang`?

source/slang/hlsl.meta.slang
```diff
@@ -2320,10 +2318,9 @@ vector<T,N> GetAttributeAtVertex(vector<T,N> attribute, uint vertexIndex);
 __generic<T : __BuiltinType, let N : int, let M : int>
 [__readNone]
 __target_intrinsic(hlsl)
-__target_intrinsic(glsl, "$0[$1]")
+__target_intrinsic(GL_NV_fragment_shader_barycentric, "$0[$1]")
+__target_intrinsic(GL_EXT_fragment_shader_barycentric, "$0[$1]")
```
skogler, line 2322: Is this the correct way to conditionally require a GL extension for a HLSL function?

I was not sure, but it seems to work.

csyonghe: Yes, this look correct to me.

csyonghe (approved):

PR #2707, opened 2023-03-16
Support for producing SourceMap on emit

jsmall-nvidia:
On top of #2706 

* Added "-source-map" option
* Generates source map on emit, can be used with line-declaration-mode or not
* Adds the source map (as JSON) to the artifact if produced

Whilst this is added to the artifact and so is "available", there isn't currently support for accessing or writing out.


csyonghe (approved):
So we are not emitting name mapping yet?

Issue #2512, opened 2022-11-11
Matrix swizzle writes don't work


jsmall-nvidia:
Related to #580

Take the example 

```
    float4x4 mat = {};
    
    mat._m00 = 1.f;
```

```
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unimplemented: matrix swizzle lvalue case
```

This can be worked around with

```
mat[0].x = 1.0f;
mat[0][0] = 1.0f;
```

Which are equivalent. The _m (or 1 indexed _ style) are seen as a swizzle, even if the result is the same as stetting an element with the other styles. Some enhancements...

* Have parity with the other styles at a minimum (such that you can write an element)
* If a vector write swizzle write is not supported have a more clear error
* Support fully the _ and _m styles with swizzled writes

On checking with DXC it does support matrix swizzle writes. To avoid surprise and have more complete HLSL support seems like it is worth fully supporting.

PR #2713, opened 2023-03-20
Matrix swizzle writes

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/2512

Depends on https://github.com/shader-slang/slang/pull/2686


source/slang/slang-ir.cpp
```diff
@@ -2809,6 +2809,13 @@ namespace Slang
             operands);
     }
 
+    IRVectorType* IRBuilder::getVectorType(
+        IRType* elementType,
+        UInt    elementCount)
+    {
+        return getVectorType(elementType, getIntValue(getUIntType(), elementCount));
```
csyonghe, line 2816: I think in everywhere else we are using getIntType for the index operand.
We should keep things consistent by emiting an IntType here to allow redundancy removal.

csyonghe: Maybe use `IRIntegerType` for the `elementCount`  parameter.

expipiplus1: good idea, thanks

csyonghe (approved):

PR #2729, opened 2023-03-24
AD: Fix type checking for higher-order custom derivatives definitions

saipraveenb25:


source/slang/slang-syntax.cpp
```diff
@@ -1263,8 +1259,41 @@ Index getFilterCountImpl(const ReflectClassInfo& clsInfo, MemberFilterStyle filt
                 }
             }
         }
+
         return nullptr;
     }
+    
+    Val* _tryLookupConcreteAssociatedTypeFromThisTypeSubst(ASTBuilder* builder, DeclRef<Decl> declRef)
```
csyonghe, line 1266: I wonder if this loop is not needed anymore?

tests/autodiff/high-order-user-defined-derivative.slang.expected.txt
```diff
@@ -0,0 +1,10 @@
+type: float
+0.000000
+1.000000
+0.000000
+0.000000
+type: float
```
csyonghe, line 6: delete?

source/slang/slang-syntax.cpp
saipraveenb25: Reverted!

csyonghe (approved):

PR #2734, opened 2023-03-26
Add PyTorch C++ binding generation.

csyonghe:
This changes adds code emit logic for PyTorch kernel binding.

Use `[TorchEntryPoint]` to mark functions for pytorch export.
A torch entry point can use `__dispatch_kernel` to launch cuda kernels.
Entry point function can use `TorchTensor` type to allocate tensors and pass them to kernel functions that accepts a `TensorView<T>`.

Example:
```
[CudaKernel]
void myKernel(TensorView<float> inValues, TensorView<float> outValues)
{
    if (cudaThreadIdx().x > 0)
        return;
    outValues.store(cudaThreadIdx().x, sin(inValues.load(cudaThreadIdx().x)));
}

[TorchEntryPoint]
public __extern_cpp TorchTensor<float> runCompute(TorchTensor<float> inValues)
{
    var outValues = TorchTensor<float>.alloc(1);
    __dispatch_kernel(myKernel, uint3(1, 1, 1), uint3(32, 1, 1))(inValues, outValues);
    return outValues;
}
```

Commands to compile:
```
slangc kernel.slang -o kernel.cu
slangc kernel.slang -o kernel_bind.cpp -target torch-binding
```
This will generate both `kernel.cu` and `kernel_bind.cpp` that can be used to implement a PyTorch C++ extension.

Then the user need to provide their setup script (according to pytorch documentation):
```
from setuptools import setup
from torch.utils.cpp_extension import BuildExtension, CUDAExtension

setup(
    name='kernel',
    ext_modules=[
        CUDAExtension('kernel', [
            'kernel_bind.cpp',
            'kernel.cu',
        ])
    ],
    cmdclass={
        'build_ext': BuildExtension
    })
```

With this, run setup to install extension:
```
python setup.py install
```

Now we can run a test with:

```
import math
import torch

# Our module!
import kernel

class KernelFunction(torch.autograd.Function):
    @staticmethod
    def forward(ctx, input):
        outputs = kernel.runCompute(input)
        return outputs

    @staticmethod
    def backward(ctx, grad_i):
        return grad_i
    

X = torch.randn(2, 2)
Y = KernelFunction.apply(X)
print("X=\n")
print(X)
print("Y=\n")
r=Y[0].data.cpu().numpy()
print(r)
```


csyonghe:
Test fail because slang-llvm is not updated to use latest IR opcode.
Merge now, and will rerun the tests + fixes issues later.

PR #2739, opened 2023-03-27
Apply IR simplifcation immediately after specialization to avoid duplicates.

csyonghe:
Perform IR simplications to fold constants in a specialized function, so further specializations from the specialized function will have as simple specialization arguments as possible to avoid creating specializations that eventually simplified into the same thing.


source/slang/slang-ir-specialize.cpp
```diff
@@ -2439,6 +2439,13 @@ IRInst* specializeGenericImpl(
                 // value.
                 cloneInstDecorationsAndChildren(
                     &env, module, specializeInst, specializedVal);
+
+                // If specializedVal is a function, optimize it.
```
expipiplus1, line 2443:
```suggestion
                // Perform IR simplifications to fold constants in this specialized value if it is a function, so 
                // further specializations from the specialized function will have as simple specialization 
                // arguments as possible to avoid creating specializations that eventually simplified into 
                // the same thing.
```

PR #2740, opened 2023-03-27
Create and cache flattened inheritance lists

tangent-vector:
The basic change here is to have a cached lookup that can map a `Type`, or a `DeclRef` that might refer to a type or extension, to a list of the *facets* that comprise it.

The notion of a *facet* here is similar to what the C++ standard calls "sub-objects".
A declared type like a `struct` has a facet for its own direct members, as well as a facet for each of its (transitive) base `struct` types and each `interface` it conforms to.
A type also has a facet corresponding to each `extension` that applies to it.

The set of facets for a type is de-duplicated (so that "diamond" inheritance patterns don't cause issues) and deterministically ordered, using a variation of the C3 linearization algorithm.

The creation of a linearized list of facets should help the compiler implementation in two key places:

* Testing if a type implements an interface (or inherits from a base type) should now only take time linear in the number of (transitive) bases of that type. We can simply scan the linearized facet list to see if it contains a facet corresponding to the given base.

* Looking up the members of a type (or a value of a given type) should be greatly simplified, since all of the members can be found in a single linear scan of the facet list. In addition, those facets will be ordered so that facets for "more derived" types will precede those for "less derived" types, so that shadowing in the case of overrides should be easier to implement.

Some caveats to be aware of:

* The handling of conjunction types (e.g., `IFoo & IBar`) complicates the implementation greatly, because once a conjunction is involved the simple approach to subtype testing above falls apart. Conjunctions also complicate the problem of making sure that type witnesses always have a unique representation.

* We don't implement the full C3 algorithm here because it has some failure cases that we'd still like to support. In particular if we have both `IX : IA, IB` and `IY : IB, IA`, the C3 algorithm says it is illegal to have `IZ : IX, IY` because the two bases it inherits from disagree on the relative ordering of `IA` and `IB` in their own linearizations.

* The implementation of lookup still needs to be revamped to properly handle all the nuances of shadowing and overides. It is basically impossible for a single linear list to encode when two members of the same name should or should not be considered to create ambiguity. Some amount of refinement is always going to be needed.


source/slang/slang-check-conformance.cpp
```diff
@@ -452,67 +252,16 @@ namespace Slang
             // witness for each of the case types, and that is
             // enough to build a witness for the tagged union.
             //
-            if(outWitness)
-            {
-                TaggedUnionSubtypeWitness* taggedUnionWitness = m_astBuilder->create<TaggedUnionSubtypeWitness>();
-                taggedUnionWitness->sub = taggedUnionType;
-                taggedUnionWitness->sup = DeclRefType::create(m_astBuilder, superTypeDeclRef);
-                taggedUnionWitness->caseWitnesses.swapWith(caseWitnesses);
+            TaggedUnionSubtypeWitness* taggedUnionWitness = m_astBuilder->create<TaggedUnionSubtypeWitness>();
```
csyonghe, line 255: should this be getOrCreate?

source/slang/slang-check-inheritance.cpp
```diff
@@ -0,0 +1,945 @@
...
+// information for types and decalrations.
+
+//#include "slang-lookup.h"
+//#include "slang-syntax.h"
+//#include "slang-ast-synthesis.h"
+//#include <limits>
+
+namespace Slang
+{
+    InheritanceInfo SharedSemanticsContext::getInheritanceInfo(Type* type)
```
csyonghe, line 15: Does it make sense to store the InheritanceInfo directly inside a `Type` object?

source/slang/slang-check-inheritance.cpp
```diff
@@ -0,0 +1,945 @@
...
+            auto extendedType = getTargetType(astBuilder, extensionDeclRef);
+            selfType = extendedType;
+            selfFacetKind = Facet::Kind::Extension;
+        }
+
+        // Because we are dealing with entities that have declarations, the
+        // first item in our linearization will always be a facet for
+        // the declaration itself.
+        //
+        TypeEqualityWitness* selfIsSelf = selfType ? visitor.createTypeEqualityWitness(selfType) : nullptr;
```
csyonghe, line 140: Is it valid to skip the TypeEqualityWitness for interface types?

source/slang/slang-check-overload.cpp
```diff
@@ -1415,6 +1416,14 @@ namespace Slang
             candidate.flavor = OverloadCandidate::Flavor::UnspecializedGeneric;
             candidate.status = OverloadCandidate::Status::GenericArgumentInferenceFailed;
 
+            if(candidate.item.declRef.getName()->text == "__shl")
```
csyonghe, line 1419: left over debugging code?

csyonghe (approved):

source/slang/slang-check-conformance.cpp
```diff
@@ -386,17 +192,21 @@ namespace Slang
...
             {
-                if (outWitness)
-                {
-                    *outWitness = extractExistentialType->getSubtypeWitness();
-                }
-                return true;
+                auto witness = extractExistentialType->getSubtypeWitness();
+                return witness;
             }
             return false;
```
csyonghe, line 204: `return nullptr`

csyonghe: Fixed.

source/slang/slang-check-conformance.cpp
tangent-vector: Probably, but at this point I think we need to go ahead and remove tagged union types from the codebase. The dynamic dispatch work does everything the tagged unions did, better (at least as far as I can remember).

csyonghe (approved):

source/slang/slang-check-inheritance.cpp
tangent-vector: There's a really subtle thing here, that the `InheritanceInfo` for a type does *not* depend only on the `Type`.

We create a distinct `SharedSemanticsContext` for each module that we check, and one of the reasons for that is that the set of `import`ed modules has an effect on what `extension`s are visible during checking. When different sets of `extension`s are visible, it is possible for the same `Type` to yield different linearizations (which makes perfect sense; it has different (transitive) bases and members).

csyonghe (approved):

PR #2742, opened 2023-03-28
AD: Warped-Area-Sampling test works now.

saipraveenb25:
Pending: I have to work on de-duplicating the witness synthesis code..


tests/autodiff/was/warped-sampling-1d.slang
```diff
@@ -0,0 +1,296 @@
...
+    PRNG prng = PRNG(seed);
+    
+    float d_color = 1.0 / 1000.0;
+    __bwd_diff(renderSample)(prng, d_color);
+
+    AllMemoryBarrierWithGroupSync();
+
+    if (threadIdx.x < 10)
+        endpointDifferentialBuffer[threadIdx.x] =
+            reinterpret<float, uint>(endpointDifferentialBufferUInt[threadIdx.x]);
```
csyonghe, line 287: Don't do this, floating point results are not determinstic between builds and execution environments.

saipraveenb25: I'm not sure how else to do `atomicAdd` on floats? Should I compare the uint value directly?

PR #2750, opened 2023-03-29
Enable SLANG_ENABLE_DXIL_SUPPORT on non-Windows platforms

expipiplus1:
This currently grabs the DXC headers from the system, rather than from external/dxc

We should make this consistent by either pulling in the Windows adapter from DXC into external/dxc or by making the Windows builds use <dxc/dxcapi.h>

### DIFF
--- a/external/dxc/WinAdapter.h
+++ b/external/dxc/WinAdapter.h
@@ -1,0 +1,1046 @@
+//===- WinAdapter.h - Windows Adapter for non-Windows platforms -*- C++ -*-===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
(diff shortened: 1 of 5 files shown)


expipiplus1:
While this works and I think is an overall improvement, I did notice that occasionally slang-test fails to terminate, waiting for a lock in `dxcompiler.so`'s `llvm::llvm_shutdown`, but only when being debugged under gdb... I had a cursory glance at this under valgrind, but nothing super obvious stuck out.

expipiplus1:
OK, that was a use after free bug in DXC. Probably not a great idea to add this to the test matrix until that's fixed.

expipiplus1:
Should be good to go, we dlopen with `RTLD_NODELETE` for now, so as not to call the global destructors in dxcompiler.so. I'm not installing dxcompiler on the linux builders for now, as there's actually nothing to test as all the dxil tests currently also rely on dx12.

PR #2751, opened 2023-03-29
Detect when calls to execv fail

expipiplus1:
We create a pipe which the child closes on execv. If execv fails then the child
is able to tell us why through the pipe. If it closes before we hear anything,
then we know execv was successful.

also `s/exit/_exit`

- [x] I think that maybe over time the cleanup on failure has got a bit wonky here, lots of code paths leave unclosed fds.


expipiplus1:
As much fun as it is writing these intricate functions, we should consider just using posix_spawn where all the hard work has already been done.

csyonghe (approved):

PR #2757, opened 2023-03-30
Enable CUDA render api on unix

expipiplus1:


source/core/slang-render-api-util.cpp
```diff
@@ -277,7 +277,7 @@ static bool _canLoadSharedLibrary(const char* libName)
         default: break; 
     }
 #elif SLANG_UNIX_FAMILY
-    // Assume on unix target we have Opengl and Vulkan for now
+    // Assume on unix target we never have D3D
```
csyonghe, line 280: D3D is now available on WSL, so we may want to support it in the future too.

expipiplus1: Vkd3d too

PR #2747, opened 2023-03-28
Implement FileCheck tests for several test commands

expipiplus1:
- SIMPLE, SIMPLE_EX
- SIMPLE_LINE
- REFLECTION, CPU_REFLECTION
- CROSS_COMPILE

It does not currently support the render tests or the COMPARE_COMPUTE commands

It is invoked by adding `(filecheck=MY_FILECHECK_PREFIX)` to the test command, for example

TEST:CROSS_COMPILE(filecheck=SPIRV): -target spirv-assembly

Need to go through and neaten things up, and finish support for the
compute tests (which have both data and compiler diagnostic output to
confuse things).

depends on https://github.com/shader-slang/slang-llvm/pull/64

PR #2767, opened 2023-04-02
Fix several silently failing tests

expipiplus1:
- Failing in debug builds
- Misconfigured diagnostic output
- Missing expected.txt (https://github.com/shader-slang/slang/pull/2747 should also make this not a silent success)


csyonghe (approved):
Looks great, thank you!

PR #2776, opened 2023-04-05
Update git version tag on local builds.

csyonghe:
We use git tag for release builds on github but not local builds.
This enables using the result of git describe as version string for local builds.

### DIFF
--- a/slang-tag-version.h
+++ b/slang-tag-version.h
@@ -1,1 +1,0 @@
-#define SLANG_TAG_VERSION "unknown"
--- a/slang-tag-version.h
+++ b/slang-tag-version.h
@@ -1,0 +1,1 @@
+#define SLANG_TAG_VERSION "unknown"
--- a/slang-tag-version.h
+++ b/slang-tag-version.h
@@ -1,1 +1,1 @@
-#define SLANG_TAG_VERSION "unknown"
+#define SLANG_TAG_VERSION "v0.27.6-9-ga05e12fc"
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,4 @@ build/**/*.log
 *.dll
 *.dxil
 /tests/serialization/*.map
+/slang-tag-version.h
--- a/.gitignore
+++ b/.gitignore
@@ -77,4 +77,4 @@ build/**/*.log
 *.dll
 *.dxil
 /tests/serialization/*.map
-/slang-tag-version.h
+slang-tag-version.h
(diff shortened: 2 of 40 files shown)


slang-tag-version.h
```diff
@@ -1 +1 @@
-#define SLANG_TAG_VERSION "unknown"
+#define SLANG_TAG_VERSION "v0.27.6-9-ga05e12fc"
```
jsmall-nvidia, line 1: Shouldn't this stay as unknown, or be removed from the project?

jsmall-nvidia (changes requested):
The slang-tag-version.h seems to be in .gitignore, but is still part of the project?

csyonghe:
The problem is that slang-tag-version.h is still needed for people who pull and build on linux because we haven't integrated generation of this file as a part of linux build. I am not sure if premake's prebuild event will work there.

expipiplus1:
For other shells, it could be something like this (bananas that the powershell version is so incredibly verbose)

```lua
    filter { "not system:windows" } 
        prebuildcommands {
            'printf "#define SLANG_TAG_VERSION \\"$$(git describe --tags)\\"\\n" > ' .. path.getabsolute("slang-tag-version.h")
        }
```

The prebuild event seems to work fine, although it's extra noise on every build :/

It is however displeasing that the format of this header is duplicated here and in the powershell script. Premake doesn't seem to have any support for template files, but if you could get the powershell equivalent of the (untested) following then I think that'd be neater.

```lua    
        prebuildcommands {
            'sed -e s/@VCS_TAG@/$$(git describe --tags)/' .. 
				        '<' .. path.getabsolute("slang-tag-version.h.in")
				        '>' .. path.getabsolute("slang-tag-version.h")
        }
```

(with slang-tag-version.h.in)
```c
#define SLANG_TAG_VERSION "@VCS_TAG@"
```

expipiplus1:
Also note that, as implemented without a fallback, this makes `git` a build dependency and prevents building from a source tarball.

csyonghe:
I kind of want to use C++17 has_include feature to just allow things to compile without the header to exist. On windows the power shell script is set to not return failure if any of the command fails, so strictly there is not a dependency on git.

expipiplus1:
```lua
    filter { "not system:windows" } 
        prebuildcommands {
            'printf "#define SLANG_TAG_VERSION \\"$$(git describe --tags || printf unknown)\\"\\n" > ' .. path.getabsolute("slang-tag-version.h")
        }
```

This should be tolerant of missing git

PR #2797, opened 2023-04-13
Some small fixes with Windows/DX usage

expipiplus1:


tools/gfx/d3d/d3d-util.cpp
```diff
@@ -498,15 +500,23 @@ bool D3DUtil::isTypeless(DXGI_FORMAT format)
...
-    static HMODULE s_dxgiModule = LoadLibraryA("dxgi.dll");
-    if (!s_dxgiModule)
-    {
-        fprintf(stderr, "error: failed load 'dxgi.dll'\n");
-        return nullptr;
-    }
-
+    static std::once_flag once;
+    static SharedLibrary::Handle s_dxgiModule = nullptr;
+    std::call_once(once, [](){
```
expipiplus1, line 507: We should probably catch std::system_error here and deal with it promptly

tools/gfx/d3d11/d3d11-device.cpp
```diff
@@ -46,15 +46,20 @@ SlangResult DeviceImpl::initialize(const Desc& desc)
...
     // Rather than statically link against D3D, we load it dynamically.
-    HMODULE d3dModule = LoadLibraryA("d3d11.dll");
-    if (!d3dModule)
+    SharedLibrary::Handle d3dModule;
+#if SLANG_WINDOWS_FAMILY
+    const char* libName = "d3d11"
+#else
+    const char* libName = "dxvk_d3d11";
+#endif
+    if (SLANG_FAILED(SharedLibrary::load(libName, d3dModule)))
```
expipiplus1, line 55: This leaks a library reference, but no more than before... Same for d3d12

tools/gfx-unit-test/gfx-test-texture-util.cpp
```diff
@@ -22,7 +22,7 @@
 #if GFX_ENABLE_RENDERDOC_INTEGRATION
 #    include "external/renderdoc_app.h"
 #    define WIN32_LEAN_AND_MEAN
-#    include <Windows.h>
+#    include <windows.h>
```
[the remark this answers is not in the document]
expipiplus1: According to this answer, `Windows.h` is indeed the file installed on modern systems, however that's a case-insensitive fs, and this does make a difference for linux and mingw builds (from case-sensitive filesystems).

https://stackoverflow.com/questions/15466613/lowercase-windows-h-and-uppercase-windows-h-difference

csyonghe (approved):

PR #2802, opened 2023-04-13
Warn on float-to-double coercion for arguments.

### LINKED ISSUE #2200: `fma` silently coerces to and from double on HLSL targets
In HLSL `fma` is available only on the double type. 

https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-fma

Using fma in Slang will silently promote to double do fma, and then silently convert back to float. The conversions and the double fma are likely *much* slower than what would be reasonably expected by calling with float parameters. 

There is also `mad` which doesn't have this problem.

https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/mad

Take for example

```

[numthreads(4, 1, 1)]
{

    float a = tid + 1;
    float b = tid * 2.5;
    float c = tid / 2;
    
    float d = fma(a, b, c);
    
    outputBuffer[tid] = d;
}
```

Compiles, but the fma becomes an operation on *double* silently. 

Note that if we compile the same code on DXC we get an error. 

```
float d = fma(a, b, c);
              ^~~
note: candidate function not viable: no known conversion from 'float' to 'double' for 1st argument
```

It seems clear that if you want to write for performance (rather than precision) then `mad` is the right option. Reading the docs a `mad` may or may not be fused. In HLSL there is no clear way to get a performant fused multiply add generally speaking. 

Another reason that this is in issue is when porting from GLSL `fma` is available for float and there is no `mad` option.

Coercing from double to float seems like it should be an error, because of the loss of precision.
Coercing from a float to a double should probably by default be at minimum a warning.

csyonghe:
Fixes #2200.

Note that this change only warns on argument coercion.

Writing `double a = floatVal;` will not trigger the warning despite an implicit cast is done.

The main work involved here is 
1. extend the `__implciit_conversion` modifier to carry an identifier so we can recognize the `float-double` conversion is being picked by overload resolution.
2. added a `CoercionContext` parameter to `coerce` function so we can tell the intention of this cocersion, and not issue a warning for obvious uses (such as initialization).


source/slang/slang-ast-support-types.h
```diff
@@ -128,6 +128,13 @@ namespace Slang
         kConversionCost_Impossible = 0xFFFFFFFF,
     };
 
+    typedef unsigned int BuiltinConversionName;
+    enum : BuiltinConversionName
+    {
```
jsmall-nvidia, line 133: Nitpik: Why not just use enum class, then theres no need for the kSomeName_ prefixing?

source/slang/slang-ast-support-types.h
```diff
@@ -128,6 +128,13 @@ namespace Slang
...
+        kBuiltinConversion_Unknown = 0,
+        kBuiltinConversion_FloatToDouble = 1,
```
jsmall-nvidia, line 135: Again nitpicky... 

The use of `Name` and `Context` seem like they are a little confusing as both have more normal programming meanings, and in Slang specifically. A context being a bunch of state, and Name being a uniquified string or at least text related.

CoercionContext  -> CoercionSite/CoercionLocation?
BuiltinConversionName -> Has prefix of just BuiltinConversion, but Kind, Style etc also work.

PR #2803, opened 2023-04-14
Diagnose on using uninitialized `out` param.

csyonghe:
This change adds a conservative check on using `out` parameters that are not yet initialized.

It is conservative in the sense that if a load is reachable from any stores, we will not issue an error.


source/slang/slang-diagnostic-defs.h
```diff
@@ -593,6 +593,7 @@ DIAGNOSTIC(40020, Error, cannotUnrollLoop, "loop does not terminate within the l
 DIAGNOSTIC(41000, Warning, unreachableCode, "unreachable code detected")
 
 DIAGNOSTIC(41010, Warning, missingReturn, "control flow may reach end of non-'void' function")
+DIAGNOSTIC(41015, Error, usingUninitializedValue, "use of uninitialized value.")
```
expipiplus1, line 596: I think for clarity this should name the value being used

PR #2807, opened 2023-04-14
Add warning for returning without initializing out parameter

expipiplus1:
Also name uninitialized uses of out parameters

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2649,6 +2649,15 @@ __generic<T, U>
 __intrinsic_op($(kIROp_Reinterpret))
 T reinterpret(U value);
 
+// Use an otherwise unused value
+//
+// This can be used to silence the warning about returning before initializing
+// an out paramter.
+__generic<T>
+[__readNone]
+[ForceInline]
+void unused(inout T){}
+
 // Specialized function
 
 /// Given a string returns an integer hash of that string.
--- a/source/slang/slang-ir-use-uninitialized-out-param.cpp
+++ b/source/slang/slang-ir-use-uninitialized-out-param.cpp
@@ -75,27 +75,33 @@ namespace Slang
                 }
             }
             // Check all address loads.
-            List<IRLoad*> loads;
+            List<IRInst*> loadsAndReturns;
             for (auto addr : addresses)
             {
                 for (auto use = addr->firstUse; use; use = use->nextUse)
                 {
                     if (auto load = as<IRLoad>(use->getUser()))
-                        loads.add(load);
+                        loadsAndReturns.add(load);
                 }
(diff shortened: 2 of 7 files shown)


source/slang/slang-ir-use-uninitialized-out-param.cpp
```diff
@@ -75,27 +75,33 @@ namespace Slang
                 }
             }
             // Check all address loads.
-            List<IRLoad*> loads;
+            List<IRInst*> loadsAndReturns;
```
csyonghe, line 78: Now that I think of it, we should expand this pass to cover `Var`s too. Since this pass take place after SSA pass, normal scalar vars are already lowered into SSA values. The only vars remaining are arrays and struct typed variables. The check for them is the same for `out` params.

In a separate pass we should also check all uses of `kIROp_Undef` and warn/error on those. Perhaps there are a couple valid use of `undefined` values and we need to skip those.

tests/diagnostics/uninitialized-out.slang
```diff
@@ -2,6 +2,56 @@
...
+        // Shouldn't warn
+        unused(y);
+        return;
+    }
+    y = 0;
+    // Shouldn't warn
+    return;
+}
+
+// TODO: This should warn that n is potentially uninitialized
```
csyonghe, line 45: The tricky case is something like
```
void f(out float v)
{
     for (int i = 0; i < 5; i++)
     {
          if (i == 2) v = 5;
     }
    return; // should not warn.
}
```

csyonghe (approved):

source/slang/core.meta.slang
```diff
@@ -2649,6 +2649,13 @@ __generic<T, U>
 __intrinsic_op($(kIROp_Reinterpret))
 T reinterpret(U value);
 
+// Use an otherwise unused value
+//
+// This can be used to silence the warning about returning before initializing
+// an out paramter.
+__generic<T>
+void unused(inout T){}
```
csyonghe, line 2657: [__readNone] ?

csyonghe: Or [ForceInline]. basically we need to get rid of it with optimization pass.

tests/diagnostics/uninitialized-out.slang
expipiplus1: indeed, there are only so many cases which we can reasonably catch. I think this is definitely one of them though

source/slang/slang-ir-use-uninitialized-out-param.cpp
expipiplus1: https://github.com/shader-slang/slang/issues/2817

PR #2812, opened 2023-04-18
Make SourceMap a value type

jsmall-nvidia:
On top of #2811 

* Make SourceMap a value type
* Added IBoxValue and BoxValue types
* Added swapWith for more types
* Added asBoxValue

Original motivation around this was around the wrapping around RefObject to make SourceMap available as an artifact representation. The change means the SourceMap can be treated as a `ICastable/IUnknown` from a reference counting/casting/externably accessible point of view, whilst used as SourceMap to provide simple use/access. The use of IBoxValue<T> interface allows type safety and COM conformance.

PR #2815, opened 2023-04-19
Improvements outputting containers

jsmall-nvidia:
On top of #2812 

* Split out IR from container
  * When setting a target as .slang-module behavior is same as before
  * When outputting to "container", does *not* by default output IR
* Enable output of IR (in a container) via `-emit-ir`
* Simplified some of the processing/output 
* Remove the `-source-map` option
  * Make source-map a "line directive" option
* Copy associated artifacts from source to subsequent artifacts (ie a binary artifact will contain metadata/source maps if appropriate)
* Added artifact "filtering", which simplifies output as well as writing logic

The `-source-map` directive is no longer needed for obfuscated source maps, because when obfuscation is enabled a source map is always produced. 

Having `-source-map` for emit, seems a little odd as you'd end up with a source-map and typically `#line`. The other idea in the change to making a `line-directive` is to try and make more clear that the source map enabling is on emit.


source/compiler-core/slang-artifact-util.cpp
```diff
@@ -83,6 +83,12 @@ static bool _checkRecursive(ArtifactUtil::FindStyle findStyle)
         return false;
     }
 
+    /* We currently can't write out diagnostics, so for now we'll say they are insignificant */
+    if (isDerivedFrom(desc.payload, ArtifactPayload::Diagnostics))
+    {
+        return false;
```
expipiplus1, line 89: Are we silently dropping diagnostics here? If so, can we alert the user by some other means?

jsmall-nvidia: This means when converting the artifact hierarchy to a file system, we drop diagnostics associated with any artifacts *if* we can't serialize them. If we enable serialization (as say .txt or .json), then they will by default be included.

This doesn't change regular diagnostic reporting or anything like that. The diagnostics is still available in the "source of truth" artifact hierarchy.

Not everything associated with a artifact will be writable, and so the "isSignificant" provides a bottle neck around what we consider an error if we didn't serialize it.

It may make sense in the filtering (or in the writing to file system) to have a DiagnosticSink to report issues more broadly.

PR #2822, opened 2023-04-21
Correct vulkan extension name for inline uniforms

expipiplus1:


expipiplus1:
We should probably generate this portion of `vk-device.cpp` from the vulkan spec.

PR #2824, opened 2023-04-21
Add support for `kIROp_FloatCast`

saipraveenb25:


csyonghe (approved):
We should just use the same logic to handle reinterpret, or bitCast if birth the from type and the to type are differentiable.

tests/autodiff/float-cast.slang
```diff
@@ -0,0 +1,66 @@
+//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type
```
csyonghe, line 1: May want to add -render-features here as well.

csyonghe (approved):

PR #2833, opened 2023-04-24
WIP: Add dataflow driven uninitialized out parameter detection

expipiplus1:
TODO

It should be more conservative in cases such as these
```
void foo(out int x)
{
    for (int i = 0; i < 8; i++)
        x = 0;
}
```

(this particular case has an easy rewrite to:

```
do
{
    x = 0;
    i++;
} while(i < 8)
```

because the comparison has no side effects)


csyonghe:
I wonder if it is easier to implement this pass by first introducing a local variable `t` for each `out` parameter, then assign the out param with `load(t)` right before every `return`, and then super-SSA the func. After SSA, if we see any use of `undef`, we just warn that.

By super-SSA, I mean running both our `constructSSA` and `eliminateAddressInsts` pass in a *clone* of the original func to get rid of pointer use even if the variable has struct or array type. This way we simplify everything down to just checking the existance of `undef` inst.

source/slang/slang-dataflow.h
```diff
@@ -0,0 +1,148 @@
...
+
+    // If we can't evaluate our transfer function, return the only possible value
+    if(maxNumUpdates == 0)
+        return List<InOutPair<Domain>>::makeRepeated(InOutPair<Domain>{initialInput, Domain::bottom()}, ns.getCount());
+
+    // Our algorithm state, `infos` tracks the input output of each node as
+    // well as if it should be recomputed
+    // TODO: We could prune known-stable nodes from this dictionary as we go...
+    Dictionary<Node, Info> infos;
+    std::priority_queue<Node, std::vector<Node>, P> workQueue(strictlyDependsOn);
```
csyonghe, line 62: Is there any reason to use priority queue? It seems like a normal work list is sufficient.

csyonghe: We've trying to avoid using std containers in this codebase, unless it is strictly necessary.

source/slang/slang-dataflow.h
```diff
@@ -0,0 +1,148 @@
...
+// - A list of eventual inputs and outputs of the transfer functions of the
+//   nodes specified in `ns`
+//
+template<typename T, typename P>
+List<InOutPair<typename T::Domain>> dataFlow(
+    T& context,
+    List<typename T::Node> ns,
+    P strictlyDependsOn,
+    typename T::Domain initialInput = T::Domain::bottom(),
+    UInt maxNumUpdates = std::numeric_limits<UInt>::max())
```
csyonghe, line 42: `kMaxUInt`, the rest of the codebase uses std very sparingly, so let's keep it that way.

csyonghe:
Closing since this is too old. Please reopen a PR if we decide to work on this again.

PR #2839, opened 2023-04-25
Fix most of the disabled warnings on gcc/clang

expipiplus1:
This should bring things more in line with what MSVC does (particularly with regards to unused variables)

- not -Wswitch as it was just too much

We tend to do `if(auto unused = as<Foo>(bar))` a lot, where `unused` is a
descriptive name. In most of these cases this changes these to be `const auto`
which gcc does not warn about.

Fixed one bug where we weren't passing IRCFGNormalizationPass to normalizeCFG
in prepareFuncForBackwardDiff (for the sink).


csyonghe (dismissed):
Nice cleanup!
I find that gcc warning on parenthesis around && and || a bit annoying (msvc doesn't warn against this). Do you think it is good to have?

source/compiler-core/slang-artifact-handler-impl.cpp
```diff
@@ -29,8 +29,9 @@ namespace Slang {
 
 SlangResult DefaultArtifactHandler::queryInterface(SlangUUID const& uuid, void** outObject)
 {
-	if (auto ptr = getInterface(uuid))
+	if ([[maybe_unused]] auto ptr = getInterface(uuid))
```
jsmall-nvidia, line 32: This looks like a bug in the original code. *outObject should just return ptr. 
ptr doesn't have to be equal to this, so that should be removed.

source/compiler-core/slang-doc-extractor.cpp
```diff
@@ -717,8 +717,6 @@ SlangResult DocMarkupExtractor::extract(const SearchItemInput* inputs, Index inp
 {
     struct Entry
     {
-        typedef Entry ThisType;
```
jsmall-nvidia, line 720: I dunno, just because it's not used doesn't mean it should be removed, despite warnings from gcc or elsewhere.

source/slang/slang-ast-dump.cpp
```diff
@@ -253,7 +249,7 @@ struct ASTDumpContext
         buf.appendChar('\"');
         for (const char c : slice)
         {
-            if (_isPrintableChar(c))
+            if (std::isprint(c))
```
jsmall-nvidia, line 252: Err. std::isprint behavior is different than _isPrintableCharacter (it uses locale for starters). It's not clear to me that we want this change. https://en.cppreference.com/w/cpp/string/byte/isprint

csyonghe:
For parity with msvc, I'd keep `-Wno-unused-but-set-variable`, -Wno-unused-local-typedefs` and `-Wno-parentheses`.
The problem with enabling warnings that are not issued by msvc is that it increases the chances we have to do additional fixup rounds after pushing things to CI.

jsmall-nvidia:
> The problem with enabling warnings that are not issued by msvc is that it increases the chances we have to do additional fixup rounds after pushing things to CI.

For me this is also a concern.

source/compiler-core/slang-artifact-handler-impl.cpp
expipiplus1: Yeah, overall this bit was a little suspicious. The reason I did it this way was that getInterface doesn't increment the reference count (as it should according to the COM documentation. So the only correct and behavior preserving fix was to make sure that this was the same pointer.

source/compiler-core/slang-doc-extractor.cpp
expipiplus1: I think that could be true elsewhere, but this particular case `Entry` is local to a function so doesn't take part in any interface, and within this function `Entry` is always going to be shorter (and clearer) than `Entry::ThisType`

source/slang/slang-ast-dump.cpp
expipiplus1: ah yes, that was lazy on my part not checking the context properly (I assumed that if we're "dumping" something we'll want it locale compatible). I'll change that one back

source/slang/slang-check-decl.cpp
```diff
@@ -4232,7 +4231,7 @@ namespace Slang
             // It is possible that there was an error in checking the base type
             // expression, and in such a case we shouldn't emit a cascading error.
             //
-            if( auto baseErrorType = as<ErrorType>(baseType) )
+            if( const auto baseErrorType = as<ErrorType>(baseType) )
```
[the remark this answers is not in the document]
expipiplus1: There were far too many of these to go through manually, but I didn't want to remove potentially descriptive names by removing everything. being able to add `const` was a convenient middle ground.

source/compiler-core/slang-gcc-compiler-util.cpp
```diff
@@ -117,7 +117,7 @@ SlangResult GCCDownstreamCompilerUtil::calcVersion(const ExecutableLocation& exe
 
     SLANG_COMPILE_TIME_ASSERT(SLANG_COUNT_OF(prefixes) == SLANG_COUNT_OF(types));
 
-    for (Index i = 0; i < SLANG_COUNT_OF(prefixes); ++i)
+    for (UIndex i = 0; i < SLANG_COUNT_OF(prefixes); ++i)
```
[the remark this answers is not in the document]
expipiplus1: :+1: I'll switch SLANG_COUNT_OF

expipiplus1:
> The problem with enabling warnings that are not issued by msvc is that it increases the chances we have to do additional fixup rounds after pushing things to CI.

It was actually precisely this pain-point (having CI MSVC complain about things that gcc ought to have caught) which prompted me starting on this. At the very least I think we should merge the uncontroversial changes but leave the `-Wno` flag, then I can just enable these locally.

source/slang/slang-ast-dump.cpp
expipiplus1: pissed that gcc warns on this:

```
../../../source/slang/slang-ast-dump.cpp:247:56: warning: comparison is always true due to limited range of data type [-Wtype-limits]
  247 |         static_assert(std::numeric_limits<char>::max() <= 0x80);
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```

expipiplus1: However, I think this is a potential, albeit unlikely bug

Consider the case of dumping a string "\DEL". One could imagine a handling of \DEL here which leads to the open quote symbol being deleted.

I've changed things to be a predicate over needing escaping instead, which conveniently makes this an important check.

expipiplus1:
> For parity with msvc, I'd keep `-Wno-unused-but-set-variable`, `-Wno-unused-local-typedefs` and `-Wno-parentheses`. The problem with enabling warnings that are not issued by msvc is that it increases the chances we have to do additional fixup rounds after pushing things to CI.

I've readded -Wno-parentheses. 

I'd argue that unused-but-set variables are most likely to be an error rather than something done for clarity, so I think even if MSVC can't warn on this we should keep it. I feel similarly but less strongly about unused local typedefs.

expipiplus1:
Thank you both for the tedious work of going through these changes.

I did the last couple of changes this morning, `-Wimplicit-fallthrough` caught several lurking bugs, so I feel strongly we should keep this one:

```
- Fixes JSON comment parsing
- Fixes escaped characters in CLikeSourceEmitter::emitStringLiteral
- Fixes SPIR-V Float constant emitting
- Guards against fallthrough for param ops in AutoDiffTranscriberBase and DifferentiableTypeConformanceContext
```

source/core/slang-type-text-util.cpp
```diff
@@ -197,7 +197,7 @@ struct DebugInfoFormatTable
 
     switch (type)
     {
-        default:    /* fall-through to none */
```
csyonghe, line 200: Should we use `[[fallthrough]]` or  `; // fall through` ?
`[[fallthrough]]` may not be supported by older compilers.

tools/slang-test/slang-test-main.cpp
```diff
@@ -222,7 +222,7 @@ void skipToEndOfLine(char const** ioCursor)
                     cursor++;
                 }
             }
-            ; // fall through to:
+            [[fallthrough]];
```
csyonghe, line 225: Let's pick a consistent style.

source/core/slang-type-text-util.cpp
expipiplus1: We should use `[[fallthough]]` as a c++17 feature. In this particular case I kept the comment as a reminder to fall through to the specific `None` case.

tools/slang-test/slang-test-main.cpp
expipiplus1: I think I've used `[[fallthrough]]` everywhere except for headers compiled by downstream compilers (for which we only require c++14. In those cases I've used `// fall through` which quiets gcc.

PR #2843, opened 2023-04-26
For C-like targets, emit resource declarations before other globals

saipraveenb25:
This is to fix a weird edge case in the DXC compiler, where using a array of `SamplerState` as a function parameter type before declaring any resources of the same type leads to an 'incomplete type' error.

To fix this, we emit resource declarations & bindings before all other globals (except for types that the resource declarations depend on)


tests/autodiff/intellisense-test.slang
```diff
@@ -0,0 +1,39 @@
+RWStructuredBuffer<float> outputBuffer;
```
csyonghe, line 1: If this is intended to test language server, use `//TEST:LANG_SERVER:` and place it under `tests/language-server`, you can refer to other tests there to see how to test the language server features.

tests/autodiff/multi-differential.slang
```diff
@@ -0,0 +1,96 @@
+//TEST_IGNORE_FILE:
```
csyonghe: Why ignored?

tests/autodiff/higher-order-custom-derivative.slang
```diff
@@ -0,0 +1,39 @@
+//TEST_IGNORE_FILE:
```
csyonghe: Why ignored?

tests/autodiff/multi-differential.slang
saipraveenb25: Whoops, all these tests were experimental.. (accidentally committed)
I've removed them all now

tests/autodiff/swizzled-store.slang
```diff
@@ -0,0 +1,27 @@
+//TEST_IGNORE_FILE:
```
csyonghe: delete if it is not testing anything?

tests/autodiff/higher-order-custom-derivative.slang
saipraveenb25: Whoops, all these tests were experimental.. (accidentally committed)
I've removed them all now

tests/autodiff/swizzled-store.slang
saipraveenb25: Accidentally committed test

PR #2849, opened 2023-04-27
MVP for higher order functions

expipiplus1:
The larger remaining pieces are:

- [x] Type inference (in our limited one-direction way)
- [ ] Some more semantic checking (making sure parameters are always const for instance, is it worth doing this even on a best effort case)
- [x] Parser fixes (arrows associate the wrong way for example)


tests/language-feature/saturated-cooperation/simple.slang
```diff
@@ -0,0 +1,74 @@
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX():-vk -compute -shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE_EX():-slang -compute -shaderobj -output-using-type
+
+//TEST_INPUT:ubuffer(data=[0 3 2 2], stride=4):out,name=outputBuffer
+RWStructuredBuffer<uint> outputBuffer;
+
+func saturated_cooperation<A : __BuiltinType, B>(cooperate : A -> B, fallback : A -> B, A input) -> B
```
expipiplus1, line 7: I think it looks pretty neat

expipiplus1:
I'm happy with this for now, the fusing part can come in a later PR.

There's still plenty of scope for more hof improvements however.

expipiplus1:
hmm, the saturated cooperation tests are failing on linux release, however I can't replicate this locally

csyonghe:
There are still two failed tests on TeamCity that seems legit:
```
tests/language-feature/saturated-cooperation/simple.slang (dx12)

(1): error 1004: unknown command-line option '-dxil'
```

Same for tests/language-feature/saturated-cooperation/simple2.slang (dx12)

source/slang/slang-check-overload.cpp
```diff
@@ -1432,6 +1447,18 @@ namespace Slang
...
+        {
+            // We could probably be broader than just parameters here
+            // eventually.
+            // Limit it for now though to make the specialization easier
+            ensureDecl(localDeclRef, DeclCheckState::CanUseFuncSignature);
+            const auto type = localDeclRef.getDecl()->getType();
+            if(const auto funType = as<FuncType>(type))
+                AddFuncExprOverloadCandidate(funType, context, context.originalExpr->functionExpr);
+            else
+                SLANG_UNREACHABLE("Trying to find overload for non-function typed parameter");
```
csyonghe, line 1460: This will cause intellisense to completely stop working for the current file since it will throw an exception. Consider more graceful handling of errors like this in front end, issueing an unimplemmted diagnostic maybe the best way.

source/slang/slang-mangle.cpp
```diff
@@ -238,9 +238,23 @@ namespace Slang
         {
             emitRaw(context, "E");
         }
+        else if (const auto bottomType = dynamicCast<BottomType>(type))
+        {
+            emitRaw(context, "B");
+        }
+        else if (auto funcType = dynamicCast<FuncType>(type))
```
csyonghe, line 245: Do we need to have a case for TupleType here?

source/slang/slang-parser.cpp
```diff
@@ -2243,6 +2243,20 @@ namespace Slang
         return parseThisTypeExpr(parser);
     }
 
+    static Expr* parseTupleTypeExpr(Parser* parser)
+    {
+        parser->ReadToken(TokenType::LParent);
```
csyonghe, line 2248: I am worried that this is going to introduce more ambiguity. I suggest we remove the parenthesis syntax for now, and just parse it as a Tuple<x,y> syntax.

source/slang/slang-parser.cpp
```diff
@@ -2438,6 +2452,11 @@ namespace Slang
             typeSpec.expr = parseThisTypeExpr(parser);
             return typeSpec;
         }
+        else if(parser->LookAheadToken(TokenType::LParent))
```
csyonghe, line 2455: Again, I think it will be safer to look for "Tuple" here.

csyonghe (commented):
How does this handle generic higher order functions? I want us to be more cautious at adding new syntax for types. I suggest we do the tradition generic like syntax for now, and add in better syntax at a later time.

source/slang/slang-check-overload.cpp
expipiplus1: I think the intention here was that it could never happen. would `SLANG_ASSERT(funType)` be clearer?

source/slang/slang-mangle.cpp
expipiplus1: Not yet, they're used as a utility for parsing function types only at the moment so shouldn't ever get this far. It definitely couldn't hurt however.

source/slang/slang-parser.cpp
```diff
@@ -4829,25 +4850,54 @@ namespace Slang
...
+            if(AdvanceIf(parser, TokenType::OpBitAnd))
+            {
+                auto rightExpr = _parsePostfixTypeExpr(parser);
+
+                auto andExpr = parser->astBuilder->create<AndTypeExpr>();
+                andExpr->loc = loc;
+                andExpr->left = TypeExp(leftExpr);
+                andExpr->right = TypeExp(rightExpr);
+                leftExpr = andExpr;
+            }
```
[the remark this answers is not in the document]
expipiplus1: how about `func (a,b) -> c`. This mirrors nicely the syntax we have already for function declarations.

expipiplus1:
> Syntax wise I think the tuple case will make it very hard to tell from a var decl from a C-style type cast expr.
> 
> Function type wise the use of -> makes me worry about potential issues if we ever need to add the C++ -> operator in the future for various reasons.

ok, I can drop the tuple type stuff without compromising the rest of the PR.

For functions, how about `func (a,b) -> c` (so requiring the `func` keyword). This mirrors nicely the syntax we have already for function declarations.

source/slang/slang-parser.cpp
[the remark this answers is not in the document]
csyonghe: This may cause issues if we decide to use the func keyword to define lambdas. I think Func<R, T> is safest for now and we can decide a more convenient syntax later.

source/slang/slang-check-overload.cpp
csyonghe: Hmm, but if assert turns into an assume in release build, it may cause problems.

expipiplus1: ok, seems like it was unnecessary anyway as we handle this function not adding any overload candidates gracefully.

I added an explicit return on the else branchs to make this control flow clear.

source/slang/slang-mangle.cpp
expipiplus1: done

expipiplus1:
If we require that function arguments always have to be parenthesized, then `func` as a function type keyword can be disambiguated from `func` as a declaration keyword by the next token being a open paren.

```
func foo(f : float) -> int;
let bar : func (float) -> int;
```

csyonghe:
That would make it hard if we later decide to use func to start a lambda. For example:

func (x) {return x*x;}(5);

csyonghe:
I am not saying that we will use this syntax for lambda, but it is good to be conservative when it comes to introducing ambiguity to the language when the syntax we add is not for compatibility. We don't need to make parsing hard for non compatibility reasons.

expipiplus1:
ok, I added the `functype` introducing keyword you suggested.  The legacy C declaration syntax is a real killer....

csyonghe (approved):

PR #2853, opened 2023-04-28
Use Index for FuncType param count

expipiplus1:


csyonghe (approved):
Right now all counts in the IR are `UInt`. We should do a cleanup pass to convert all of them to Index.

PR #2857, opened 2023-04-29
SSA Register Allocation improvements.

csyonghe:
This PR improves SSA register allocation pass so that we don't treat a phi arg and phi param as interfering when we see a use of phi param after the phi arg's live range, because that that point we know phi arg and phi param will have the same value.
This gets rid of unnecessary variables introduced in the eliminate-phi pass.

Added a file-check test to make sure it is behaving as intended.

Also fixes the spirv direct emit bug that causes OpVariable to appear after other ordinary insts.

Fixes a small issue in autodiff availability pass that causes a recomputed call inst to reference operands that are not pulled into recompute blocks.


source/slang/slang-ir-autodiff-primal-hoist.h
```diff
@@ -236,6 +236,34 @@ namespace Slang
         Dictionary<IRInst*, InversionInfo> invInfoMap;
     };
 
+    struct Use
```
expipiplus1, line 239: Maybe the name here could be made clearer? `UseOrPseudoUse` perhaps?

expipiplus1 (dismissed):

source/slang/slang-ir-autodiff-primal-hoist.h
csyonghe: Done.

Issue #2858, opened 2023-04-29
Make asserts more consistent


expipiplus1:
Our use of SLANG_ASSERT seems a little inconsistent; sometimes we're treating it as just a debug print and making efforts (returning `SlangResult::Fail` and handling that later for example), while other times we're treating it as "the program has gotten into an inconsistent or unreasonable state, so abort".

We should do a cleanup and split these into 

- SLANG_ASSERT, which aborts in debug builds and is `std::unreachable` in release builds
- `outputDebugMessage` for the case where we feel we can continue

expipiplus1:
See also https://github.com/shader-slang/slang/issues/2858

natduca:
I understand some actions have been taken. Probably lets close this for now, reopen when we have a forcing function to do this.

PR #2859, opened 2023-04-29
Add SLANG_ASSUME and use it in release asserts

expipiplus1:


expipiplus1:
See also https://github.com/shader-slang/slang/issues/2858

source/core/slang-common.h
```diff
@@ -76,28 +76,43 @@ namespace Slang
...
-#define SLANG_EXPECT(VALUE, MSG) if(VALUE) {} else SLANG_ASSERT_FAILURE(MSG)
-#define SLANG_ASSERT(VALUE) SLANG_EXPECT(VALUE, #VALUE)
+#define SLANG_ASSERT(VALUE) do{if(!(VALUE)) SLANG_ASSERT_FAILURE(#VALUE);} while(0)
 #else
-#define SLANG_EXPECT(VALUE, MSG) do {} while(0)
-#define SLANG_ASSERT(VALUE) do {} while(0)
+#define SLANG_ASSERT(VALUE) SLANG_ASSUME(VALUE)
 #endif
 
 #define SLANG_RELEASE_ASSERT(VALUE) if(VALUE) {} else SLANG_ASSERT_FAILURE(#VALUE)
```
csyonghe, line 115: This is not using SLANG_ASSUME?

expipiplus1: deliberately not using ASSUME. A release assert is one we specifically want to still message/abort on, even in release builds.

expipiplus1: TBH, I'm inclined to just replace it with SLANG_ASSERT as it's not really clear what it's saying: "I am certain that this condition is true, HOWEVER I'm not sure enough that I'm comfortable not checking it in release builds...". If that's the case then we should either reason further about the code to make ourselves more certain that it can't happen, or we should replace these with explicit calls to `abort()` (or whatever end of days error handling function)

csyonghe: We should be careful about that because we have many places that uses release_assert because the expr inside the assert is intended to run.

expipiplus1: Indeed, FWIW I think that clang can warn when such expressions can't be proven side effect free.

source/core/slang-common.h
```diff
@@ -76,28 +76,43 @@ namespace Slang
...
+// Do not rely on side effects of the condition being performed.
+//
+#if defined(__cpp_assume)
+#    define SLANG_ASSUME(X) [[assume(X)]]
+#elif SLANG_GCC
+#    define SLANG_ASSUME(X) do{if(!(X)) __builtin_unreachable();} while(0)
+#elif SLANG_CLANG
+#    define SLANG_ASSUME(X) __builtin_assume(X)
+#elif SLANG_VC
+#    define SLANG_ASSUME(X) __assume(X)
```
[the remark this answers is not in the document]
expipiplus1: This was exactly the intention. Even if we did have `assert(cond); assume(!cond)` the compiler would be free to optimize away the assert because it wouldn't ever trigger under the assumption of `!cond`. It's not just a hint, it's a promise which if broken means the compiler may do what it pleases.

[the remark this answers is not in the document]
csyonghe: Oh, I see it is used in release only, then that makes sense.

PR #2862, opened 2023-05-02
Refactor around options parsing/initial vk shift

jsmall-nvidia:
* Use OptionsParser to hold state
* Split out large bits of functionality into methods
* Use template to make enum extraction/conversion more convenient
* Fix small bug introduced around -Wno-/-W handling

Has preliminary work around VulkanShiftOptions, parsing. It is not currently wired to layout.

The refactor motivation was around trying to simplify integration.


csyonghe (approved):

jsmall-nvidia:
Covered in #2865

PR #2863, opened 2023-05-02
Various dxc/fxc compatibility fixes.

csyonghe:


source/slang/slang-type-layout.cpp
```diff
@@ -3183,6 +3183,35 @@ RefPtr<VarLayout> StructTypeLayoutBuilder::addField(
...
+    {
+        fieldLayout->AddResourceInfo(LayoutResourceKind::Uniform)->index = uniformOffset;
+    }
+    UniformLayoutInfo fieldInfo = fieldResult.info.getUniformLayout();
+    auto uniformInfo = m_info;
+    m_rules->AddStructField(&uniformInfo, fieldInfo);
+    m_info.alignment = uniformInfo.alignment;
+    m_info.size.raw = Math::Max(
+        m_info.size.getFiniteValue(),
+        (size_t)(uniformOffset + fieldResult.layout->FindResourceInfo(LayoutResourceKind::Uniform)->count.getFiniteValue()));
```
expipiplus1, line 3211: Should this be rounded to a multiple of `alignment`?

source/slang/slang-parser.cpp
```diff
@@ -1750,6 +1758,51 @@ namespace Slang
...
+                            }
+                            else if (parser->LookAheadToken(":", 2))
+                            {
+                                // If we see a "< xxx :", we can also parse it as a generic arg list.
+                            }
+                            else
+                            {
+                                // Otherwise, we may be looking at an effect annotation.
+                                // For now we just skip tokens until we see `>`, if we see any `"` in between,
+                                // we can conclude that this is an annocation.
```
expipiplus1, line 1785: s/annocation/annotation/

source/slang/slang-type-layout.cpp
csyonghe: I think we/dxc don't round sizes up to alignment.

source/slang/slang-parser.cpp
```diff
@@ -1750,6 +1758,51 @@ namespace Slang
...
+                                // If we see `<let` then we are looking at a generic arg list.
+                            }
+                            else if (parser->LookAheadToken(":", 2))
+                            {
+                                // If we see a "< xxx :", we can also parse it as a generic arg list.
+                            }
+                            else
+                            {
+                                // Otherwise, we may be looking at an effect annotation.
+                                // For now we just skip tokens until we see `>`, if we see any `"` in between,
```
expipiplus1, line 1784: This mentions looking for `"`, but the code below looks like it's searching for `;`

source/slang/slang-ir-use-uninitialized-out-param.cpp
```diff
@@ -63,6 +63,7 @@ namespace Slang
                         addresses.add(use->getUser());
                         break;
                     case kIROp_Store:
+                    case kIROp_SwizzledStore:
```
expipiplus1, line 66: I wonder if we should either drop swizzledstore altogether, or make a superclass for store and swizzledstore...

expipiplus1: (this is just musing, not for this PR)

source/slang/slang-emit-c-like.h
```diff
@@ -405,15 +407,15 @@ class CLikeSourceEmitter: public SourceEmitterBase
     void emitStruct(IRStructType* structType);
     // This is used independently of `emitStruct` by some GLSL parameter group
     // output functionality
-    void emitStructDeclarationsBlock(IRStructType* structType);
+    void emitStructDeclarationsBlock(IRStructType* structType, bool allowOffsetLayout);
```
expipiplus1, line 410: Would be nice to have an enum for this, otherwise the calling code has a contextless true/false

expipiplus1: likewise above

source/slang/slang-emit-c-like.h
```diff
@@ -363,7 +365,7 @@ class CLikeSourceEmitter: public SourceEmitterBase
     void diagnoseUnhandledInst(IRInst* inst);
     void emitInst(IRInst* inst);
 
-    void emitSemantics(IRInst* inst);
+    void emitSemantics(IRInst* inst, bool allowOffsets = false);
```
expipiplus1, line 368: we have a default argument here, would it be clearer to always specify, or are there a lot of cases when we don't care?

source/slang/slang-diagnostic-defs.h
```diff
@@ -544,8 +544,8 @@ DIAGNOSTIC(39007, Error, unknownRegisterClass, "unknown register class: '$0'")
 DIAGNOSTIC(39008, Error, expectedARegisterIndex, "expected a register index after '$0'")
 DIAGNOSTIC(39009, Error, expectedSpace, "expected 'space', got '$0'")
 DIAGNOSTIC(39010, Error, expectedSpaceIndex, "expected a register space index after 'space'")
-DIAGNOSTIC(39011, Error, componentMaskNotSupported, "explicit register component masks are not yet supported in Slang")
-DIAGNOSTIC(39012, Error, packOffsetNotSupported, "explicit 'packoffset' bindings are not yet supported in Slang")
+DIAGNOSTIC(39011, Error, invalidComponentMask, "invalid register component mask '$0'.")
```
expipiplus1, line 547: Do we have a policy on reusing retired diagnostic numbers? In this case obviously nobody is going to be relying on it for component mask support, so just musing.

source/slang/slang-check-modifier.cpp
```diff
@@ -918,6 +918,40 @@ namespace Slang
...
+                    break;
+                case 'y':
+                    uniformOffset += 4;
+                    break;
+                case 'z':
+                    uniformOffset += 8;
+                    break;
+                case 'w':
+                    uniformOffset += 12;
+                    break;
```
expipiplus1, line 945: Do these depend on the component type being 4 bytes?

expipiplus1 (approved):

source/slang/slang-parser.cpp
csyonghe: Yeah, we intend to look for `;` here.

source/slang/slang-diagnostic-defs.h
csyonghe: We have aboslutely no story on backward compatibility in this front.

source/slang/slang-check-modifier.cpp
csyonghe: I think so, it is specialized to xyzw according to the HLSL doc.

PR #2866, opened 2023-05-04
Fix issue with out-of-order insts during type promotion when transposing code.

saipraveenb25:
- Fixed insertion point of type promotion in transposition pass
 - Added `[BackwardDifferentiable]` to dot and cross builtin methods


winmad (commented):
I guess a few matrix-related derivative functions should be also marked as `[BackwardDifferentiable]` to enable higher-order derivatives: L370, L379, L389, L399, L422, L432, L455, L465.

source/slang/diff.meta.slang
winmad, line 1: Shall we rename `mul` to `__d_mul` at L389, L422, L455, and L465 to avoid ambiguity?

PR #2867, opened 2023-05-04
Improvements around HLSLToVulkanLayout

jsmall-nvidia:
* Add warnings in binding if remapping is enabled but a mapping isn't found
* Improved diagnostics in binding
* Added more tests
* Changed API a little so can support mappings other than "shifts"


tests/bindings/hlsl-to-vulkan-global-binding.hlsl
```diff
@@ -0,0 +1,19 @@
+//TEST:SIMPLE(filecheck=CHECK):-target glsl -profile ps_4_0 -entry main -fvk-bind-globals 5 9 -line-directive-mode none
+
+// CHECK: layout(binding = 0)
+// CHECK: uniform texture2D t_0;
```
csyonghe, line 4: We can make it more robust by:
`// CHECK: uniform texture2D t{{[_0-9]*}}`

csyonghe: Same for all other number post fixes.

csyonghe (approved):

PR #2874, opened 2023-05-08
Fusion pass for saturated_cooperation

expipiplus1:
Also includes https://github.com/shader-slang/slang/pull/2849


source/slang/hlsl.meta.slang
```diff
@@ -6885,3 +6894,115 @@ uint3 cudaBlockIdx();
...
+// waveMatch: a function to return a mask of lanes with the same input as this one
+// broadcast: a function which returns the value passed into it on the specified lane
+//
+func saturated_cooperation_using<A, B, C>(
+    cooperate : functype (A, B) -> C,
+    fallback : functype (A, B) -> C,
+    waveMatch : functype (A) -> uint4,
+    broadcast : functype (A, int) -> A,
+    A input,
+    B otherArg)
```
csyonghe, line 6968: What if we have more than one B? For example, matrix multiplication will have two arguments instead of 1.
We need to provide overloads of this function with different numbers of arguments.

source/slang/slang-check-expr.cpp
```diff
@@ -3677,4 +3677,43 @@ namespace Slang
         }
     }
 
+    Expr* SemanticsExprVisitor::visitFuncTypeExpr(FuncTypeExpr* expr)
+    {
+        // The input and output to a function type must both be types
+        for(auto& t : expr->parameters)
+            t = CheckProperType(t);
+        expr->result = CheckProperType(expr->result);
+
```
csyonghe, line 3686: May need to return error expr here if expr->result is null.

source/slang/slang-check-expr.cpp
```diff
@@ -3677,4 +3677,43 @@ namespace Slang
...
+        expr->type = m_astBuilder->getTypeType(funcType);
+
+        return expr;
+    }
+
+    Expr* SemanticsExprVisitor::visitTupleTypeExpr(TupleTypeExpr* expr)
+    {
+        // All tuple members must be types
+        for(auto& t : expr->members)
+            t = CheckProperType(t);
```
csyonghe, line 3707: I think we either need to have `CheckPropertyType` return ErrorType if `t` is null, or handle the case of null here.

source/slang/slang-check-overload.cpp
```diff
@@ -1432,9 +1449,23 @@ namespace Slang
             auto type = DeclRefType::create(m_astBuilder, genericTypeParamDeclRef);
             AddTypeOverloadCandidates(type, context);
         }
+        else if( auto localDeclRef = item.declRef.as<ParamDecl>() )
+        {
+            // We could probably be broader than just parameters here
+            // eventually.
+            // Limit it for now though to make the specialization easier
+            ensureDecl(localDeclRef, DeclCheckState::CanUseFuncSignature);
```
csyonghe, line 1457: I have a feeling that this might need to be `TypesFullyResolved`.
But that is mostly because our current checking of `ThisType` is broken and can't be fixed without a total revamp of DeclRef.

source/slang/slang-check-overload.cpp
```diff
@@ -1671,6 +1702,16 @@ namespace Slang
         {
             if (IsErrorExpr(arg))
                 return CreateErrorExpr(expr);
+
+            // If this argument is itself an overloaded value then we can't
+            // sensibly continue
+            if(as<OverloadedExpr>(arg) || as<OverloadedExpr2>(arg))
```
csyonghe, line 1708: This is unfortunate, but you won't have this limitation if __tc_call is implemented as a high-order expr like __bwd_diff.

source/slang/slang-ir-fuse-satcoop.cpp
```diff
@@ -0,0 +1,513 @@
...
+    // then we remove the first call and replace the second with a fused
+    // call.
+    IRCall* lastCall = nullptr;
+    for(auto inst = block->getFirstInst(); inst != block->getTerminator(); inst = inst->getNextInst())
+    {
+        if(auto call = isCallNamed("saturated_cooperation_using", inst))
+        {
+            if(lastCall)
+            {
+                auto fused = tryFuseCalls(builder, lastCall, call);
```
csyonghe, line 488: If I have 
```
a()
b()
c()
```
This logic seemed to be generated a cascaded call graph of:
```
f1()
{
      f2();
      c();
}
f2()
{
      a();
      b();
}
```
Which seems to be a lot of boilerplate.

In general we want to minimize the size of code we emit since dxc is very slow.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -31,6 +31,9 @@ bool FunctionCallSpecializeCondition::isParamSuitableForSpecialization(IRParam*
         //
         if (as<IRGlobalParam>(arg)) return true;
 
+        // Similarly for these global values
+        if( as<IRGlobalValueWithCode>(arg) ) return true;
```
csyonghe, line 35: An `IRGlobalValueWithCode` can be a `IRGeneric`, `IRFunc`, or simply a constant with init expr.
Do you intend to include all of them here?

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -505,6 +508,11 @@ struct FunctionParameterSpecializationContext
             //
             ioInfo.key.vals.add(oldGlobalParam);
         }
+        else if( auto globalConstant = as<IRGlobalValueWithCode>(oldArg) )
```
csyonghe, line 511: `IRGlobalValueWithCode` is not just `constant`.

tests/language-feature/higher-order-functions/inference.slang
```diff
@@ -0,0 +1,21 @@
+//TEST:SIMPLE:
```
csyonghe, line 1: We should avoid creating just `SIMPLE` tests. Whenever possible, prefer creating a COMPUTE test. 
With `SIMPLE` tests we at least need some file-check rules.

source/slang/slang-ir-fuse-satcoop.cpp
```diff
@@ -0,0 +1,513 @@
...
+    // Walk over the instructions in this block
+    // If we see a call to sat_coop then remember where it is and keep
+    // walking, if we reach another call without first encountering any
+    // instructions with which our first call can't be safely reordered
+    // then we remove the first call and replace the second with a fused
+    // call.
+    IRCall* lastCall = nullptr;
+    for(auto inst = block->getFirstInst(); inst != block->getTerminator(); inst = inst->getNextInst())
+    {
+        if(auto call = isCallNamed("saturated_cooperation_using", inst))
```
csyonghe, line 484: This name based matching does not seem to be very robust. I think it maybe better to use some special decoration to identify it.

source/slang/slang-ir-fuse-satcoop.cpp
csyonghe: I think we can turn this into a non-recursive process by scanning through the basic block, and identify all calls to `coop` functions, move code between calls as much as possible to form a contiguous sequence of calls, then generate a single function that calls the sequence.

csyonghe: A simple strategy like just move all no side effect insts to as early as possible may be sufficient for now.

expipiplus1: We fold 

```
sc(ca, fa)
sc(cb, fb)
sc(cc, fc)
```

to

```
sc(cacbcc, fafbfc)
```

where cacbcc (and fafbfc) look like

```
cacbcc(){
  cacb();
  cc();
}

cacb(){
  ca();
  cb();
}
```

These are inlined shortly after and the generated code is exactly what you'd expect: it's the body of `sat_coop` except that the original call to `cooperate` is replaced by three calls to `ca, cb, cc`.

A left fold as implemented is much simpler than having to deal with lists of functions and parameters I think.

source/slang/hlsl.meta.slang
expipiplus1: I *think* karthik said that a single parameter would be ok for their use case. Either way, shouldn't be hard to add an overload I guess, although I would prefer if people wrapped their multiple parameters in a struct (or eventually, anonymous tuple).

source/slang/slang-check-expr.cpp
expipiplus1: I can't see a similar case in the other visit functions, do you mean something like setting the whole type to `ErrorType` if one of the inputs or outputs is errortype?

source/slang/slang-check-overload.cpp
expipiplus1: Thank :)

Making this change doesn't break any of the tests, so I'll include it. Do you have a test in mind which would fail without this change?

source/slang/slang-ir-fuse-satcoop.cpp
csyonghe: Ah, I see. Yes then that will work. Do we have a test to check the inlining works as intended?

source/slang/slang-check-overload.cpp
expipiplus1: perhaps. This check is also overly conservative. If this parameter is a monotype, then we can try to unify each overload with it: i.e. resolve this overload with imaginary arguments conjured from the type of the parameter it's being passed as. I'm reluctant to implement this though.

source/slang/slang-ir-specialize-function-call.cpp
expipiplus1: Yes, although it's not technically necessary for the hof work.

We could even expand it to every global as they're all *suitable* for specialization. It could be that I've misinterpreted this function, but I thought it was more or less "if by specializing the function at this parameter, would it escape its scope?", i.e. is it globally scoped

source/slang/slang-ir-specialize-function-call.cpp
expipiplus1: mm, perhaps that was poor wording. I meant it in the sense that the identity isn't variable which is what matters for specialization.

source/slang/slang-ir-fuse-satcoop.cpp
expipiplus1: :100:

tests/language-feature/higher-order-functions/inference.slang
expipiplus1: done, thanks

source/slang/slang-ir-fuse-satcoop.cpp
expipiplus1: no, but should be an easy one for filecheck. good idea.

expipiplus1: ok done

csyonghe (approved):
Looks good!

source/slang/slang-ir-fuse-satcoop.cpp
```diff
@@ -0,0 +1,513 @@
...
+    auto spec = as<IRSpecialize>(call->getCallee());
+    if(!spec)
+        return nullptr;
+    auto func = findSpecializeReturnVal(spec);
+    if(!func)
+        return nullptr;
+
+    // TODO: Is there nothing better for finding a specific declaration in
+    // core.slang?
+    auto h = func->findDecoration<IRNameHintDecoration>();
```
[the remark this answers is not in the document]
expipiplus1: done

tests/language-feature/saturated-cooperation/fuse-product.slang
```diff
@@ -0,0 +1,85 @@
...
+    let x = saturated_cooperation(coopMul, fallback2, j, m);
+    // Put some calculation between the calls to check that the fusion still takes place
+    let e = min(m*100, 2);
+    let z = saturated_cooperation(coopExp, fallback1, k, int(e));
+
+    // check that all the piping functions we invent during fusing is inlined away
+    // CHECK:      = coopAdd
+    // CHECK-NEXT: = coopMul
+    // there is a tuple pack here
+    // CHECK-NEXT: Tuple
```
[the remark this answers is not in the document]
expipiplus1: As a further optimization, we could detect when s2 is an injection of s1 (or vice versa) and only pass one in, but that's orthogonal to this fusing (and not necessarily worth it)

expipiplus1:
What should we do about these tests which require sm 6_5 (which I don't think is universally supported on our testing infra)?

tests/language-feature/saturated-cooperation/fuse-product.slang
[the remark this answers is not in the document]
expipiplus1: if we were to eventually inline calls to `sat_coop` then we could certainly get rid of it, but I think that's a decision best made elsewhere in the compiler.

csyonghe:
You can use the `-render-features XXX` flag in the `Test:` line.

csyonghe:
I guess in this case, `-render-features wave-ops` ?

csyonghe (approved):

PR #2876, opened 2023-05-09
Various fixes for autodiff and slangpy.

csyonghe:


prelude/slang-torch-prelude.h
```diff
@@ -38,6 +38,23 @@
 #   include <stdint.h>
 #endif // SLANG_LLVM
 
+#if defined(_MSC_VER)
+#   define SLANG_PRELUDE_SHARED_LIB_EXPORT __declspec(dllexport)
+#else
+#   define SLANG_PRELUDE_SHARED_LIB_EXPORT __attribute__((__visibility__("default")))
```
jsmall-nvidia, line 44: What's the reason to remove attribute(dllexport) ? It's there for windows compatibility it seems, but that seems reasonable if you built windows binary with clang/gcc. Or if you were building under cygwin etc.

jsmall-nvidia (approved):

prelude/slang-torch-prelude.h
csyonghe: Yeah, this is added here to get things to compile for now. I don't know what the perfect definition would be here.

PR #2883, opened 2023-05-14
Add [Differentiable(n)] syntax to specify max order.

csyonghe:


expipiplus1:
Out of interest, what's the use case for this?

csyonghe:
This is added in preparation of supporting higher order differentiation through interfaces. We need to pre-generate all levels of derivatives in the interface.

Issue #2884, opened 2023-05-15
Missing (all) stencil formats


kopaka1822:
I noticed that all formats that provide a stencil have been removed.
I saw the comments, that D24_S8 is only supported on NVIDIA devices.
However, there is no reason to exclude D32_S8?

I think these formats are still relevant since the provide a good speedup for selective per-pixel operations when compared to using the discard intrinsic in the pixel shader.

kopaka1822:
I also noticed that the stencilRef value appears to be unused in DirectX12, since this is not part of the PipelineStateObject, but instead needs to be set via the command list (ID3D12GraphicsCommandList::OMSetStencilRef)

csyonghe:
You are right. The formats are missing mainly because we haven't run into a use case that needs it.
Your format PR looks great. Thank you for your contribution!

natduca:
I think this is patched, closing. If mistake feel free to reopen.

PR #2885, opened 2023-05-15
added D32_S8 format

kopaka1822:
related to #2884 

I prepared a pull request that includes the D32_S8 format.
It appears to be working for my program without breaking anything.
However, I am unsure about the change in the render.cpp where I set the channel count to 2 and kept the type at Float32 (since it is technically Float32_Uint8_X24).


csyonghe (approved):
Looks good to me. The `FormatInfoMap` thing may need some cleanup anyway.

PR #2897, opened 2023-05-21
fixed incorrect handling of D3D12_RENDER_TARGET_VIEW_DESC.ViewDimensi…

### LINKED ISSUE #2896: Problems when using RenderTargetViews with Texture2DArrays in DirectX
I could not bind the second slice of a texture 2D array as a render target view.
The debug profile complained because the view was created as Texture2D (instead of Texture2DArray) and the PlainSlice member was set to 1.
I tracked down the issue and have prepared a pull request.

kopaka1822:
fixed #2896 

It looks like the ViewDimension member was not set properly for the RTV and also UAV.
The documentation states that: "ViewDimension: This member also determines which _RTV to use in the following union" (https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_render_target_view_desc)
Therefore the type should be set if the underlying resource is an array and NOT if the requested view is an array. Otherwise, members like the first array slice would be ignored (Which is relevant if one wants to bind a specific slice).

An alternative fix would be to always use the TEXTUREXDARRAY enum, since appears to work fine with array sizes of 1 regardless. This would result in less nesting statements and overall be more readable. Let me know if you prefer that instead.

### DIFF
--- a/tools/gfx/d3d12/d3d12-device.cpp
+++ b/tools/gfx/d3d12/d3d12-device.cpp
@@ -1308,7 +1308,7 @@ Result DeviceImpl::createTextureView(
     RefPtr<ResourceViewImpl> viewImpl = new ResourceViewImpl();
     viewImpl->m_resource = resourceImpl;
     viewImpl->m_desc = desc;
-    bool isArray = resourceImpl ? resourceImpl->getDesc()->arraySize != 0 : false;
+    bool isArray = resourceImpl ? resourceImpl->getDesc()->arraySize > 1 : false;
     bool isMultiSample = resourceImpl ? resourceImpl->getDesc()->sampleDesc.numSamples > 1 : false;
     switch (desc.type)
     {
@@ -1396,7 +1396,6 @@ Result DeviceImpl::createTextureView(
         viewImpl->m_allocator = m_dsvAllocator;
         D3D12_DEPTH_STENCIL_VIEW_DESC dsvDesc = {};
         dsvDesc.Format = D3DUtil::getMapFormat(desc.format);
-        isArray = desc.subresourceRange.layerCount > 1;
         switch (desc.renderTarget.shape)
         {
         case IResource::Type::Texture1D:
--- a/tools/gfx/d3d12/d3d12-device.cpp
+++ b/tools/gfx/d3d12/d3d12-device.cpp
@@ -1321,13 +1321,21 @@ Result DeviceImpl::createTextureView(
         viewImpl->m_allocator = m_rtvAllocator;
         D3D12_RENDER_TARGET_VIEW_DESC rtvDesc = {};
         rtvDesc.Format = D3DUtil::getMapFormat(desc.format);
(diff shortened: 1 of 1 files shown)


tools/gfx/d3d12/d3d12-device.cpp
```diff
@@ -1321,13 +1321,21 @@ Result DeviceImpl::createTextureView(
         viewImpl->m_allocator = m_rtvAllocator;
         D3D12_RENDER_TARGET_VIEW_DESC rtvDesc = {};
         rtvDesc.Format = D3DUtil::getMapFormat(desc.format);
-        isArray = desc.subresourceRange.layerCount > 1;
```
kopaka1822, line 1324: This was the cause of the error. The isArray from line 1311 is the correct value (from the underlying resource)

tools/gfx/d3d12/d3d12-device.cpp
```diff
@@ -1341,14 +1349,26 @@ Result DeviceImpl::createTextureView(
...
+                        : 0;
+                }
+                else
+                {
+                    rtvDesc.Texture2D.MipSlice = desc.subresourceRange.mipLevel;
+                    rtvDesc.Texture2D.PlaneSlice =
+                        resourceImpl ? D3DUtil::getPlaneSlice(
+                            D3DUtil::getMapFormat(resourceImpl->getDesc()->format),
+                            desc.subresourceRange.aspectMask)
+                        : 0;
```
kopaka1822, line 1370: Here, it is important to handle Texture2D.PlaneSlice and Texture2DArray.PlaneSlice differently, since they are at different memory locations in the union.

csyonghe:
This change causes the `gfx-unit-test-tool/renderTargetTests` test to fail. You can run it locally with 
```
slang-test.exe gfx-unit-test-tool/renderTargetTests
```

kopaka1822:
@csyonghe thank you for pointing that out!
I fixed the issue in the last commit and all units tests are passing now.

The problem was that some unit tests create textures with 'arraySize == 0' and some with 'arraySize == 1'. In both cases a normal (not array) target is meant.

My code now works for both, however, I saw that some other parts in this project check for 0, while others check for 1.

csyonghe:
Thanks for the fix! We should unify the convention throughout the library on `arraySize`. I think we should regulate that `arraySize==0` means no array, `arraySize=1` means an array that has 1 layer.

kopaka1822:
Yes, I think you should add comments to the respective struct fields that indicate how special values are handled (arraySize = 0 is no array. arraySize > 0 is handled as array view)
Also, the same should be done for the `SubresourceRange` struct. The problem was that some unit tests use arraySize = 1 together with SubresourceRange.layerCount = 0, which resulted in an error, because only positive values or -1 are valid in the DX12 descriptor. It should be clarified if 0 means all layers, or if this is an illegal value.

csyonghe (approved):

PR #2899, opened 2023-05-23
Add loop inversion pass

expipiplus1:
At the moment this applies indiscriminately wherever it can, and while I don't
think that's a terrible default it could maybe do with some more nuance.
Perhaps either:

- Only perform when the condition expression is "small" for some definition of
  small
- Add a loop-uninvert pass, which will uninvert any loops where sccp was unable
  to eliminate the first condition check

Still left to do is check and fix all the cross compile tests where the output
has changed, and check that copying over the loop decorations is always correct.


expipiplus1:
hmm, some later pass is generating invalid Ir, note how in the second snippet 39 is being read before it's written (in the first snippet this value is being passed back using a parameter, which has been removed). Is there some requirement/assumption which the first snippet is violating (and which isn't picked up by the validator?)

```
[backwardDifferentiable]
[DifferentiableTypeDictionaryDecoration]
[export("_S3tu023test_loop_with_continuep1pff")]
[nameHint("test_loop_with_continue")]
func %testx5Floopx5Fwithx5Fcontinue	: Func(Float, Float)
{
block %68(
		[nameHint("y")]
		param %y	: Float):
	unconditionalBranch(%69, %y)

block %69(
		[nameHint("t")]
		param %t	: Float):
	unconditionalBranch(%70)

block %70:
	[loopMaxIters(3 : Int)]
	loop(%71, %72, %71, %t, 0 : Int)

block %73(
		[nameHint("t")]
		param %t1	: Float,
		[nameHint("i")]
		param %i	: Int):
	let  %74	: Bool	= cmpLT(%i, 3 : Int)
	ifElse(%74, %75, %72, %71)

block %71(
		param %76	: Float,
		param %77	: Int):
	let  %78	: Bool	= cmpGT(%76, 4 : Float)
	ifElse(%78, %79, %80, %80)

block %79:
	unconditionalBranch(%81, %76)

block %80:
	let  %82	: Float	= mul(%76, %76)
	unconditionalBranch(%81, %82)

block %81(
		[nameHint("t")]
		param %t2	: Float):
	let  %83	: Int	= call %10(1 : Int)
	let  %84	: Int	= add(%77, %83)
	unconditionalBranch(%73, %t2, %84)

block %75:
	unconditionalBranch(%71, %t1, %i)

block %72:
	return_val(%76)
}
```

```
### BEFORE-SPECIALIZE:
[backwardDifferentiable]
[DifferentiableTypeDictionaryDecoration]
[export("_S3tu023test_loop_with_continuep1pff")]
[nameHint("test_loop_with_continue")]
func %testx5Floopx5Fwithx5Fcontinue	: Func(Float, Float)
{
block %34(
		[nameHint("y")]
		param %y	: Float):
	[loopMaxIters(3 : Int)]
	loop(%35, %36, %35, %y, 0 : Int)

block %37:
	let  %38	: Bool	= cmpLT(%39, 3 : Int)
	ifElse(%38, %40, %36, %35)

block %35(
		param %41	: Float,
		param %42	: Int):
	let  %43	: Bool	= cmpGT(%41, 4 : Float)
	ifElse(%43, %44, %45, %45)

block %44:
	unconditionalBranch(%46, %41)

block %45:
	let  %47	: Float	= mul(%41, %41)
	unconditionalBranch(%46, %47)

block %46(
		[nameHint("t")]
		param %t	: Float):
	let  %48	: Int	= call %intx5Fx24init(1 : Int)
	let  %39	: Int	= add(%42, %48)
	unconditionalBranch(%37)

block %40:
	unconditionalBranch(%35, %t, %39)

block %36:
	return_val(%41)
}

```

csyonghe:
It feels like there really should be a PHI node inserted for that jump.
What's been changed in this PR?

csyonghe:
There is the requirement that all uses of a value must be dominated by its def, I wonder why validation isn't complaining on that.

csyonghe (approved):
I have not carefully examined the logic here, but if all tests are passing, I am assuming it is correct.

source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -444,6 +444,30 @@ static bool trySimplifyIfElse(IRBuilder& builder, IRIfElse* ifElseInst)
...
+                trueBlock->replaceUsesWith(afterBlock);
+                trueBlock->removeAndDeallocate();
+            }
+            else if(isFalseBranchTrivial && falseBlock != afterBlock && !falseBlock->hasMoreThanOneUse())
+            {
+                falseBlock->replaceUsesWith(afterBlock);
+                falseBlock->removeAndDeallocate();
+            }
+        }
+    }
```
expipiplus1, line 470: @csyonghe This was quite simple, but wasn't done before; is there a reason for that? Making this change did necessitate a fix here: https://github.com/shader-slang/slang/pull/2899/files#diff-88028dd102858f855a13f83c516f5bc3d10d434d43e0c9ecb5b12305b7194e60R196-R203

source/slang/slang-ir-loop-inversion.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    // Do we break on the 'true' side?
+    if(isSameBlockOrTrivialBranch(breakBlock, t) && f == a)
+    {
+        if(t != breakBlock)
+        {
+            branch->trueBlock.set(breakBlock);
+            t->removeAndDeallocate();
+        }
+        return true;
```
expipiplus1, line 74: Similarly to below, removing blocks like this can be subtle. Perhaps this could move to cfg simplification.

It changes

```
loop:
if cond goto d;
d: goto loopBreak;
loopBreak:
```
to
```
loop:
if cond goto loopBreak;
loopBreak:
```

When `d` is only used once with no arguments.

It's an enhancement in order to handle things like 

```
do{
	if(foo) break;
} while(x);
```

expipiplus1:
> I have not carefully examined the logic here, but if all tests are passing, I am assuming it is correct.

Looking forward to giving this comment a chuckle down the line :)

source/slang/slang-ir-simplify-cfg.cpp
csyonghe: I think I removed something similar to this to get some autodiff test to pass. Maybe that associated fix is the reason.

source/slang/slang-ir-loop-inversion.cpp
csyonghe: Yeah, it might be beneficial to move this to cfg simplification. However we need to be careful not to remove critical edges or skipping merge points.

Issue #2895, opened 2023-05-19
Enum can have same label defined multiple times 


jsmall-nvidia:
The following compiles on slang (as tested on shader playground)...

```
enum class Check2 { V, V = 2 };
```

It fails reporting the repeated name in DXC.

Usage of V, will report an error saying V is ambiguous.

PR #2904, opened 2023-05-30
Disallow duplicate enumerator names in the same enum

expipiplus1:
Fixes https://github.com/shader-slang/slang/issues/2895


expipiplus1:
Overloaded enums are much less defensible than overloaded functions.

jsmall-nvidia (approved):

PR #2911, opened 2023-05-31
Fix div-by-zero error during sccp.

### LINKED ISSUE #2892: Divide by 0 causes crash in SCCP
Take 

```
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-cpu -compute -shaderobj -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-cuda -compute -shaderobj -output-using-type

//TEST_INPUT:ubuffer(data=[0], stride=4):out,name outputBuffer

uint check<let b : bool>()
{
    return 1 / int(b);
}

[numthreads(4, 1, 1)]
{
    
    uint a = check<false>();
    uint b = check<true>();
    
	outputBuffer[tid] = a + b; 
}
```

Crashes in 

LatticeVal evalDiv(IRType* type, LatticeVal v0, LatticeVal v1)

When dividing v0 by 0.

Perhaps there should be an Error or DivBy0 or some other Flavor to represent this result. This could then be detected and cause termination, with a suitable diagnostic. 

I had assumed that this issue was in part due to being generic, and therefore would appear in IR passes because that's where generic specialization takes place. But will also fail with 

```
int doThing(int v) { return v; }
doThing(1 / 0);
```

csyonghe:
Fixes #2892.

### DIFF
--- a/source/slang/slang-ir-ssa-simplification.h
+++ b/source/slang/slang-ir-ssa-simplification.h
@@ -5,13 +5,14 @@ namespace Slang
 {
     struct IRModule;
     struct IRGlobalValueWithCode;
+    class DiagnosticSink;
 
     // Run a combination of SSA, SCCP, SimplifyCFG, and DeadCodeElimination pass
     // until no more changes are possible.
-    void simplifyIR(IRModule* module);
+    void simplifyIR(IRModule* module, DiagnosticSink* sink = nullptr);
 
     // Run simplifications on IR that is out of SSA form.
     void simplifyNonSSAIR(IRModule* module);
 
-    void simplifyFunc(IRGlobalValueWithCode* func);
+    void simplifyFunc(IRGlobalValueWithCode* func, DiagnosticSink* sink = nullptr);
 }
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -324,7 +324,7 @@ Result linkAndOptimizeIR(
     }
 
     lowerOptionalType(irModule, sink);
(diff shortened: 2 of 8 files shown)


source/slang/slang-ir-ssa-simplification.h
```diff
@@ -5,13 +5,14 @@ namespace Slang
 {
     struct IRModule;
     struct IRGlobalValueWithCode;
+    class DiagnosticSink;
 
     // Run a combination of SSA, SCCP, SimplifyCFG, and DeadCodeElimination pass
     // until no more changes are possible.
-    void simplifyIR(IRModule* module);
+    void simplifyIR(IRModule* module, DiagnosticSink* sink = nullptr);
```
jsmall-nvidia, line 12: The sink seems to be the mechanism to report the error to caller. Should we require a sink to be passed?

jsmall-nvidia (approved):

source/slang/slang-ir-ssa-simplification.h
csyonghe: There are several calls to `simplifyIR` from a pass that currently don't have access to `sink`. Plumbing `sink` through every pass feels messy. We should probably have a base context type for a pass that carries the sink over to simplify things. But for now this is fine since all errors will eventually be captured by a `simplifyIR` call in `linkAndOptimizeIR`, where a sink is provided there. When `sink` is not provided, the compilation can still continue normally upon a div-by-zero error (we simply don't fold).

PR #2913, opened 2023-05-31
Be lenient on same-size unsigend->signed conversion.

csyonghe:
With this change, uint->int will have cost `kConversionCost_SameSizeUnsignedToSignedConversion` (250) instead of `kConversionCost_General` (900) so we no longer report warnings on them. This has been more annoying then helpful.

Our current logic does not report warnings on int->uint, which is even worse. So there is no reason to report warnings uint->int.

If we have
```
int f(int, int);
int f(uint, uint);
```
After this change, the compiler now prefers the `f(int, int)` overload if the argument types are `(uint, int)`, previously we will choose `f(uint, uint)`.

We need to make sure that a call to `f(uint, 3)` still picks the `uint` overload. To do this, we special case the coercion cost computation if the argument is an `IntLiterialExpr`. We also need to fix the cache for overload resolution and candidate costs to factor in the int constant info in the cache key.

### DIFF
--- a/source/slang/slang-check-impl.h
+++ b/source/slang/slang-check-impl.h
@@ -55,11 +55,24 @@ namespace Slang
         return BasicTypeKey{ uint8_t(baseType), uint8_t(dim1), uint8_t(dim2) };
     }
 
-    inline BasicTypeKey makeBasicTypeKey(Type* typeIn)
+    inline BasicTypeKey makeBasicTypeKey(Type* typeIn, Expr* exprIn = nullptr)
     {
         if (auto basicType = as<BasicExpressionType>(typeIn))
         {
-            return makeBasicTypeKey(basicType->baseType);
+            auto rs = makeBasicTypeKey(basicType->baseType);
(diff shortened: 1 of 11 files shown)


source/slang/slang-check-impl.h
```diff
@@ -28,22 +28,51 @@ namespace Slang
...
+            if (auto constInt = as<IntegerLiteralExpr>(exprIn))
+            {
+                auto value = constInt->value;
+                if (value < 0)
+                {
+                    rs.knownNegative = 1;
+                    value = -value;
+                }
+                rs.knownConstantBitCount = 1;
+                while (value >>= 1)
```
expipiplus1, line 72: ... and then sign extension will cause this to not terminate.

expipiplus1 (approved):

source/slang/slang-check-impl.h
csyonghe: Fixed.

PR #2915, opened 2023-06-01
WIP: Fix do-while loop handling in higher order differentiation.

### LINKED ISSUE #2914: Today's installment of weird loops and autodiff
If we replace the for loop in `outerF` in `tests/autodiff/high-order-backward-diff-3.slang` with the equivalent do/while loop then we get bogus results from the backward diff pass:

before

```hlsl
    for (int i = 0; i < 2; ++i)
    {
        nx.x = nx.x * x.x;
    }
```

after

```
    int i = 0;
    [MaxIters(2)]
    do
    {
        nx.x = nx.x * x.x;
        i++;
    } while (i < 2);
```

The first thing I notice when looking at the diff in generated sources is the additional array `int[4]` in `s_bwd_outerF_Intermediates_0`. 

Adding `[ForceUnroll]` to the loop allows getting the correct result.

csyonghe:
In progress work to fix #2914.


saipraveenb25:
Closing. Fixed by: #2918

PR #2930, opened 2023-06-13
Pointer layout support

jsmall-nvidia:
Most of the effort here is around trying to work around issues with Ptr types producing infinite recursive loops. 

* In ir lowering, the fix is to add agg types to the map before they are completed
* In IR Layout IRPointerLayout is added. 
  * Ideally this type would contain a pointer to the value type, but making this work with the other mechanisms (hoisting/Builder) is significantly more difficult. For now this is just avoided, which seems ok because for layout the value type isn't needed to be known.
* For reflection tool, on pointer types it only outputs the *name* of the thing being pointed to
  * This isn't great. If the pointer is to a generic the name returned is just the generic name (not it's parameterization)
  * The layout for the thing being pointed to isn't necessarily emitted
* Has a fix for legalization
  * Assumes a pointer must be to something that resolves to itself
  * Can detect if that isn't the case, and asserts.  

The pointer mechanism used here assumes that a pointer is to some representation in a single address space. For some targets this may not actually be possible, for example when resources are in effect in their own address space.  

An additional problem is in emit. In essence the problem is not knowing if a use of a Ptr type - not knowing if it is being used for say a variable, or a load/store/element access, where the thing pointed to definitely needs the definition, or it is part of a type. The work around used here is when a pointer is hit to look to see if the thing being pointed to is already "open", if it is we assume all we need is a forward reference. Assuming that is reasonable there is an argument that something similar is appropriate for other pointer like types.

Future work:

* Making IRPointerTypeLayout able to hold the value type layout
* Layout adds GetPointerLayout, but for CPU targets it is *assumed* that the pointer is 64 bits
  * That isn't right for 32 bit targets, or with CHERI for 128 bit pointers
  * We probably want a command line mechanism to specify the pointer width for the target
* Reflection tool should probably use JSON system
  * We should reflect the layout for pointed to types


source/slang/slang-lower-to-ir.cpp
```diff
@@ -136,11 +138,18 @@ struct LoweredValInfo
...
-        IRInst*            val;
+        IRInst*             val;
         ExtendedValueInfo*  ext;
+        // We can compare any of the pointers above by comparing this pointer. If the union
+        // ever becomes something other than a union of pointers, this would no longer be applicable.
+        void*               aliasPtr;
     };
     Flavor flavor;
 
+        // NOTE! This relies on
```
csyonghe, line 149: Incomplete comment?

csyonghe (commented):
Shall we add a test to make sure code gen also works? e.g. no infinite recursion when lowering the struct type to IR. Should we register the struct type before recursively processing the members in `visitAggTypeDecl`?

csyonghe:
I guess it must be correct given the test passes, but I didn't find the place where the struct type is being registered before it is fully lowered.

jsmall-nvidia:
I found a fix for the emit issue, so now can run the previously disabled test.

There is a section in `visitAggTypeDecl` just after the "empty" type is created that registers is such that there isn't infinite recursion.

```
// We add the decl now such that if there are Ptr or other references 
        // to this type they can still complete
        context->setValue(decl, LoweredValInfo::simple(finishedVal));
```

csyonghe:
This PR doesn't seem to include the change to `visitAggTypeDecl`?

PR #2943, opened 2023-06-26
Fix DCE on mutable calls in a loop.

csyonghe:
This change fixes a DCE bug that mistakenly removes a call to a mutable function inside a loop.
A test case that reproduces the bug is included and fixed with this change.


source/slang/slang-ir-propagate-func-properties.cpp
```diff
@@ -238,34 +241,37 @@ class NoSideEffectFuncPropertyPropagationContext : public FuncPropertyPropagatio
...
+                case kIROp_unconditionalBranch:
+                case kIROp_Switch:
+                case kIROp_Return:
+                case kIROp_loop:
+                case kIROp_Call:
+                case kIROp_Param:
+                case kIROp_Unreachable:
+                case kIROp_Store:
+                case kIROp_SwizzledStore:
+                    break;
```
tangent-vector, line 256: The duplication here with the logic above seems unfortunate. Is there any way to factor this out into a subroutine?

source/slang/slang-ir-util.cpp
```diff
@@ -688,9 +693,45 @@ bool areCallArgumentsSideEffectFree(IRCall* call)
...
+                        {
+                            callBlock = dom->getImmediateDominator(callBlock);
+                            if (!callBlock)
+                                return false; // If we are here, var does not dominate the call, which should never happen.
+                            if (as<IRLoop>(callBlock->getTerminator()))
+                                return false; // The var is used in a loop, must return false.
+                        }
+                        // If we reach here, the var is used as an inout parameter for the call, but the call
+                        // is not nested in a loop at an higher nesting level than where the var is defined,
+                        // so we can treat the use as side effect free.
```
tangent-vector, line 736: I like this analysis, although it starts to feel like our definition of "side effect free" is getting a bit loose. What we really mean here is DCE-able and/or CSE-able instructions, right?

source/slang/slang-ir.cpp
```diff
@@ -7082,6 +7098,8 @@ namespace Slang
                 module->getDeduplicationContext()->getConstantMap().remove(IRConstantKey{ constInst });
             }
             module->getDeduplicationContext()->getInstReplacementMap().remove(this);
+            if (auto func = as<IRGlobalValueWithCode>(this))
+                module->invalidateDominatorTree(func);
```
tangent-vector, line 7102: Could this be a more general `module->invalidateAnalysis(this)`?

source/slang/slang-ir.h
```diff
@@ -2072,6 +2080,17 @@ struct IRModule : RefObject
...
 
+    IRDominatorTree* findDominatorTree(IRGlobalValueWithCode* func)
+    {
+        IRAnalysis* analysis = m_mapInstToAnalysis.tryGetValue(func);
+        if (analysis)
+            return analysis->getDominatorTree();
+        return nullptr;
+    }
+    IRDominatorTree* findOrCreateDominatorTree(IRGlobalValueWithCode* func);
+    void invalidateDominatorTree(IRGlobalValueWithCode* func) { m_mapInstToAnalysis.remove(func); }
```
tangent-vector, line 2091: Yeah, it seems like this one could be `invalidateAnalysisForInst(...)`.

tangent-vector (dismissed):

source/slang/slang-ir-propagate-func-properties.cpp
csyonghe: Done.

source/slang/slang-ir-util.cpp
csyonghe: Done.

source/slang/slang-ir.cpp
csyonghe: Done.

source/slang/slang-ir.h
csyonghe: Done.

PR #2947, opened 2023-06-28
Issue diagnostic for incorrect parameter types & directionality when defining custom derivatives

saipraveenb25:


source/slang/slang-check-decl.cpp
```diff
@@ -6956,6 +6956,23 @@ namespace Slang
...
+                //
+                auto funcType = as<FuncType>(calleeDeclRef->type);
+                for (Index ii = 0; ii < imaginaryArguments.getCount(); ++ii)
+                {
+                    if (imaginaryArguments[ii]->type.isLeftValue)
+                    {
+                        if (!as<OutTypeBase>(funcType->getParamType(ii)))
+                        {
+                            visitor->getSink()->diagnose(
+                                attr, Diagnostics::invalidCustomDerivative);
```
csyonghe, line 6970: We should make the diagnostic more informative.

saipraveenb25: Yeah, all the diagnostics around custom derivatives are a bit opaque at the moment (we only get yes/no feedback)
Let me try to add some information about expected signature.

source/slang/slang-check-decl.cpp
```diff
@@ -6985,9 +7045,41 @@ namespace Slang
         return "PrimalSubstitute";
     }
 
-    List<Expr*> getImaginaryArgsToFunc(ASTBuilder* astBuilder, FunctionDeclBase* func, SourceLoc loc)
+    // TODO: Can probably use the existing code in lower-to-ir.cpp by moving it somewhere
+    // common.. 
+    //
+    ParameterDirection getParamDirectionFromDecl(ParamDecl* paramDecl)
```
csyonghe, line 7051: Let's not duplicate code, at the very least just declare the function and use it here.

PR #2948, opened 2023-06-28
Warn on semicolon after `if`.

### LINKED ISSUE #2935: Warning for if followed by semicolon
Take the example code

```
int v = 0;
if (isOk());
   ++v;
```

It is *likely* the intent of the programmer is to have ++v executed only if `isOk()` is true, but the trailing `;` means that doesn't happen, and the indentation also helps to make it appear as if the `if` does control execution.

It would be useful had a warning for such situations.

csyonghe:
Fixes #2935.


jsmall-nvidia (approved):
Nice.

expipiplus1:
FWIW Clang has a more general "misleading indentation" warning, when an indented statement is within the same block as a control flow operator. This also catches cases such as 

```hlsl
if(b)
  a;
  c; // misleading indentation
```

PR #2951, opened 2023-06-29
Fix for operator assignment issue

jsmall-nvidia:
The motivation for this PR is around problems like

```
int a = 1;
uint b = 2;

a += b;
```

Would fail to compile. `+=` is implemented in the slang stdlib as a generic, of the form

```
void someFunc<T>(inout T lhs, T rhs); 
```

The problem stems from when the compiler tries to infer T. It could be `uint` or `int`, but in such a situation slang will go for the "larger" type in this case uint. This means the generic used is

```
void someFunc(inout uint lhs, uint rhs);
```

This fails to compile, because previously the out parameter had to match exactly, and if it didn't an error was produced.

To work around the problem generally, we can fix by adding a temporary, and changing the call into

```
uint tmpA = uint(a):
someFunc(tmpA, b);
a = int(tmpA);
```

This idea is fairly general and could work for any types we can convert between.

This is implemented by 

1) When there is a situation where previously we would fail with an error, if we allow the conversion, we insert a InOutImplicitCast/OutImplicitCast AST node replacing the ImplicitCast that was there previously
2) In lowering we convert to equivalent IR ops
3) A pass is introduced that performs the appropriate transformation depending on the types/target

On 2, it's perhaps worth saying this cast is more like a kind of "reinterpret" cast, or at least it can be handled that way on some targets, so it accepts a ptr input and produces a pointer output. The operand must be a LValue for some of the other logic to make sense.

There are the following scenarios in terms of what to do 

* Nothing - some kinds of casts are implicit for some targets such as HLSL on out parameters for same sized integer types
* A reinterpret cast. On targets with pointers, such as C++/CUDA we can fix the problem by just casting to the appropriate pointer (for some kinds of conversions)
* Creating a temporary of the right type and calling the function, and *converting* to the target (say an out parameter)
* Creating a temporary, converting the value into the temporary, calling the function, and converting back to the source

Finally this logic is currently only enabled for conversions between uint/int types of the same size, and scalar/vector/matrix forms thereof. It is more general though and could be used with floating point or other types potentially.


source/slang/slang-ir-lower-l-value-cast.cpp
```diff
@@ -0,0 +1,300 @@
...
+            case CodeGenTarget::GLSL_Vulkan:
+            case CodeGenTarget::GLSL_Vulkan_OneDesc:
+            // If we aren't emitting directly we are going to output GLSL to feed to GLSLANG
+            case CodeGenTarget::SPIRV:
+            case CodeGenTarget::SPIRVAssembly:
+            {
+                return SourceLanguage::GLSL;
+            }
+            case CodeGenTarget::HLSL:
+            case CodeGenTarget::DXBytecode:
```
csyonghe, line 258: We already have `isKhronosTarget` `isD3DTarget` `isCUDATarget` and `isCPUTarget` utility functions defined. We can just use those, or move this function to sit alongside with those?

source/slang/slang-ir-lower-l-value-cast.cpp
```diff
@@ -0,0 +1,300 @@
...
+#include "slang-ir-clone.h"
+
+#include "slang-ir-util.h"
+
+namespace Slang
+{
+
+struct LValueCastLoweringContext
+{    
+    static bool _isGeneric(IRInst* inst)
```
csyonghe, line 15: Can just use `findOuterGeneric(inst)`.

source/slang/slang-ir-lower-l-value-cast.cpp
jsmall-nvidia: Moved to same location as isKronos etc.

csyonghe (approved):

Issue #2634, opened 2023-02-08
Support `sizeof(T)`


cshenton:
It would be great to have `sizeof(T)` which behaves like [the one in DXC](https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/HLSL/sizeof.hlsl). This would make writing some generic code a bit nicer on the user end.

tangent-vector:
I agree that it would be good to support `sizeof(...)`.

There are some subtleties that would need to be worked out in adding that operator to Slang, some of which would have already come up as design choices when it was added to DXC:

* An HLSL type may not have a single well-defined size, since its size may depend on how it gets laid out (e.g., in a constant buffer, structured buffer, or byte-addressed buffer). I would assume that the `sizeof(...)` operator would be expected to yield the size of a type in its "natural" layout, which typically matches how it is stored in a byte-addressed buffer.

* In C/C++ (and I'm assuming also DXC) the `sizeof(...)` operator yields an integer constant expression, such that it can be used as the element count of an array, etc. This works because all of those languages/compilers know the single compilation target they are generating code for when they run the compiler front-end (so that, e.g., a C compiler knows whether `sizeof(int)` is 4 or 8 for the particular target). In contrast, the Slang compiler is architected so that the front-end can be run once even when generating code for multiple targets. As a result, it would not always be the case that `sizeof(T)` in Slang could be a constant integer during front-end compilation.

* Slang supports more compilation targets than DXC, and on some of those compilation targets types have well-defined sizes, even when they don't have sizes on other targets. For example, `sizeof(Texture2D)` would be well defined for Slang's CUDA back-end, even if it isn't for D3D11/12 and Vulkan. We'd have to make some subtle design choices about which types are and are not valid to use with `sizeof()`, and under what conditions. I'd rather not have the validity of a `sizeof()` expression depend on the compilation target.

With all of the above in consideration, it would probably be reasonable for somebody to implement a `sizeof()` operator in Slang that yields an integer constant expression in the front-end for types that have a well-defined size independent of the compilation target, and that issues a diagnostic in all other cases.

cshenton:
[...]

PR #2954, opened 2023-06-30
Initial sizeof/alignof implementation.

jsmall-nvidia:
Initial work for #2634

The issue describes some of the nuance around how to implement `sizeof`. The implementation adds a AST version of "natural" layout calculation. This was used over using the regular front end layout, because the requirements of only supporting one type of layout without worrying about the complicated cases of resources and so forth made it more straightforward.

Having layout calculated in the front end, also allows for constant folding.

If it isn't possible to calculate the sizeof/alignment in the front end they are lowered to IR instructions. A sizeOfLike IR pass will then attempt to turn them into values once generics have been specialized. This works fine for `sizeof(somegeneric<T>)` but the front end doesn't like something of the form of

```
int getSizeOf<T>() { return sizeof(T); }
```

So that will be addressed as future PR.


csyonghe:
To support `sizeof` expressions on generic types, you will need another type, e.g. `SizeOfIntVal` similar to the existing `FuncCallIntVal` class to allow `sizeof` being part of an unfold constexpr that can be folded later once the generic argument is known during AST time. For example, iot will allow us to evaluate `a`  as in:
```
struct GetSizeOf<T> { static const int value = sizeof(T); }
const int a = GetSizeOf<float>.value;
```

PR #2963, opened 2023-07-05
Work around for NonUniformResourceIndex with non integral types.

jsmall-nvidia:
Ignores NonUniformResourceIndex for non uint/int types. Was tested with float, to make sure generic was taken over (say) coercing into integral, and that appeared to work as expected.

Should probably issue a warning when it is elided in this manner. Look to add in future PR.


csyonghe (approved):

csyonghe:
I guess we can just mark the method as deprecated so it will automatically trigger a warning.

csyonghe (approved):

PR #2965, opened 2023-07-05
Add support for `length` for scalar floating point types.

jsmall-nvidia:


csyonghe (approved):
We should probably update these old tests to use -output-using-type.

PR #2967, opened 2023-07-06
Make DeclRefBase a Val, and DeclRef<T> a helper class.

csyonghe:
The intention of this change is to make DeclRef itself a syntax node so it can participate in deduplication, and can later be extended to have different subclasses to represent different cases of a DeclRef.

The main body of work in this change is that since DeclRef is a syntax node itself, we need to take extra care to make sure dynamic casts around `DeclRef` are still having its original meaning, that is casting the Decl being pointed to instead of the DeclRef object itself.

### DIFF
--- a/source/slang/slang-ast-base.h
+++ b/source/slang/slang-ast-base.h
@@ -356,12 +356,9 @@ class DeclRefBase : public Val
 {
     SLANG_AST_CLASS(DeclRefBase)
 
-    // The underlying declaration
-    Decl* decl = nullptr;
     Decl* getDecl() const { return decl; }
 
-    // Optionally, a chain of substitutions to perform
-    Substitutions* substitutions;
+    Substitutions* getSubst() const { return substitutions; }
 
     DeclRefBase(Decl* decl)
         :decl(decl)
@@ -412,6 +409,14 @@ class DeclRefBase : public Val
     // Debugging:
     String toString() const;
     void toText(StringBuilder& out) const;
+
+private:
+
+    // The underlying declaration
+    Decl* decl = nullptr;
+    // Optionally, a chain of substitutions to perform
+    Substitutions* substitutions;
+
 };
 
 SLANG_FORCE_INLINE StringBuilder& operator<<(StringBuilder& io, const DeclRefBase* declRef) { declRef->toText(io); return io; }
@@ -567,7 +572,7 @@ T* DeclRef<T>::getDecl() const
 template<typename T>
 Substitutions* DeclRef<T>::getSubst() const
 {
-    return declRefBase ? declRefBase->substitutions : nullptr;
+    return declRefBase ? declRefBase->getSubst() : nullptr;
 }
 
 template<typename T>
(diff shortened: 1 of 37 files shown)


source/slang/slang-ast-base.h
```diff
@@ -405,20 +489,22 @@ class Decl : public DeclBase
 
     ContainerDecl* parentDecl = nullptr;
 
+    DeclRefBase* defaultDeclRef = nullptr;
```
tangent-vector, line 492: Could use a comment indicating what this is/means.

source/slang/slang-ast-base.h
```diff
@@ -446,5 +532,115 @@ class Stmt : public ModifiableSyntaxNode
...
+    DeclRefBase* declRef = nullptr;
+    if (decl)
+    {
+        SLANG_ASSERT(decl->defaultDeclRef);
+        declRef = decl->defaultDeclRef;
+    }
+    else
+    {
+        declRef = nullptr;
+    }
```
tangent-vector, line 556: This side is a no-op, given that `declRef` is already initialized to null above, right?

source/slang/slang-ast-base.h
```diff
@@ -446,5 +532,115 @@ class Stmt : public ModifiableSyntaxNode
...
+HashCode DeclRef<T>::getHashCode() const
+{
+    if (declRefBase) return declRefBase->getHashCode();
+    return 0;
+}
+
+template<typename T>
+Type* DeclRef<T>::substitute(ASTBuilder* astBuilder, Type* type) const
+{
+    SLANG_ASSERT(declRefBase);
```
tangent-vector, line 611: I suppose that a null `DeclRef` could act as a no-op set of substitutions, but making it an invariant that the decl ref is non-null (as this code does) certainly seems safer, if there isn't existing code that violates that assumption.

source/slang/slang-ast-builder.h
```diff
@@ -254,34 +252,39 @@ class ASTBuilder : public RefObject
...
     {
-        return DeclRef<T>(this, decl, subst);
+        // We never create an actual DeclRefBase node to point to a null decl.
+        if (!decl)
+            return DeclRef<T>();
+
+        // If we don't have substitutions, use the default decl ref if it is created.
+        if (!subst)
+        {
+            auto& defaultDeclRef = static_cast<Decl*>(decl)->defaultDeclRef;
```
tangent-vector, line 266: Why is a reference used here?

tangent-vector (dismissed):
There's a lot of churn here, but most of it is straightforward.

There's a couple places where there seem to be larger code changes, and maybe there's subtle stuff there I'm missing, but in most cases it seems to be adapting code that was mutating `DeclRef`s to instead construct fresh ones.

source/slang/slang-ast-base.h
csyonghe: Fixed.

source/slang/slang-ast-base.h
csyonghe: Commented.

source/slang/slang-ast-builder.h
csyonghe: Fixed.

source/slang/slang-ast-base.h
csyonghe: Turned it into a no-op case.

PR #2970, opened 2023-07-07
Add support for texture footprint queries

tangent-vector:
This change adds operations to the Slang standard library that allow applications to query the texels that would be accessed by a texture sampling operation. These new operations are implementated for Vulkan via the `GL_NV_shader_texure_footprint` extension, and for D3D via NVAPI.

The main challenge in implementing this is that the NVAPI functions require the space/register of the texture and sampler being used for the query to be passed in as integers. In order to support this, the change adds new IR instructions that query the binding information of a resource, and a specialization pass that lowers those instructions by modifying the program to pass space/register information through the call chain.

The tests that have been added confirm that valid SPIR-V is being generated for this new API, but only validates the HLSL output for the D3D/NVAPI case.

### DIFF
--- a/source/slang/slang-ir-lower-binding-query.cpp
+++ b/source/slang/slang-ir-lower-binding-query.cpp
@@ -1,0 +1,610 @@
+// slang-ir-lower-tuple-types.cpp
+
+#include "slang-ir-lower-tuple-types.h"
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+
+// The pass in this file lowers the `getRegisterIndex()` and
+// `getSpaceIndex()` intrinsics, by replacing them with literal
+// values derived from the binding information on shader parameters.
+//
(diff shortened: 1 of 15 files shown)


source/slang/slang-ir-lower-binding-query.cpp
```diff
@@ -0,0 +1,610 @@
...
+namespace Slang
+{
+    // There are a ton of passes we've implemented now that use
+    // some basic work-list structures, and it seems a bit silly
+    // to be writing that code intermixed with the actual algorithm.
+    //
+    // For this file, we break the common work-list functionality
+    // out into a base type that we can re-use in specific passes.
+    //
+    struct WorkListPass
```
csyonghe, line 61: There is an `InstPassBase` class that seemed to be doing similar things here.

csyonghe (approved):

csyonghe (approved):

PR #2978, opened 2023-07-11
Use scratchData on `IRInst` to replace HashSets.

csyonghe:
Add a `scratchData` field to an `IRInst` for use by each IR pass for fast book keeping.

Used it to replace some `HashSet<IRInst*>`s in certain performance critical passes to speed up backend by approx. 10%.


source/slang/slang-ir-util.cpp
```diff
@@ -645,7 +645,7 @@ void setInsertAfterOrdinaryInst(IRBuilder* builder, IRInst* inst)
     }
 }
 
-bool areCallArgumentsSideEffectFree(IRCall* call)
+bool areCallArgumentsSideEffectFree(IRCall* call, bool useDominanceTree)
```
jsmall-nvidia, line 648: Might want to use an enum to indicate the useDominanceTree ness, such that it's clearer at call sites what the meaning is.

source/slang/slang-ir-util.cpp
```diff
@@ -964,6 +967,34 @@ bool isOne(IRInst* inst)
     }
 }
 
+void initializeScratchData(IRInst* inst)
+{
+    List<IRInst*> workList;
```
jsmall-nvidia, line 972: I'm assuming the use of the work list here is to avoid recursive traversal?

csyonghe: Right, it is used to implement stackless DFS on the IR module.

source/slang/slang-ir.h
```diff
@@ -733,6 +733,11 @@ struct IRInst
         getOperands()[index].init(this, value);
     }
 
+    // Reserved memory space for use by individual IR passes.
```
jsmall-nvidia, line 736: 64 bits is quite a lot of bits.
If you wanted a boolean test, you wouldn't need a reset if a counter was used. Each new used an increment of the counter, and a inst is reset if the scratch isn't equal to the counter. 
If you want multiple bits that doesn't work.

jsmall-nvidia (dismissed):

source/slang/slang-ir.h
csyonghe: This is to allow potential case where a pass may want to stick a pointer here for general data.

csyonghe: The current design is that an IR pass should not make  any assumptions on the intiial value of scratchData, and no conventions should be created to coordinate the access of the field across passes to prevent hard to debug issues. It is therefore safer to not use the counter strategy for now.

source/slang/slang-ir-util.cpp
csyonghe: Fixed.

PR #2994, opened 2023-07-15
Simplify lookup.

csyonghe:
This change simplifies the lookup logic in a type by scanning through the linearized inheritance list of the type instead of doing on-demand analysis of what base types and extensions can apply and recursively look into those scopes.

This avoids the repetitive work of testing sub type relationships or checking if an extension applies to a type, and therefore reduces the overall complexity of the type checking process, and drastically improves front-end performance.

With this change, the type checking time for a shader goes from 1.2s down to 0.33s. This should greatly improve application load time and speedup overall shader compilation.

### DIFF
--- a/source/slang/slang-ast-builder.h
+++ b/source/slang/slang-ast-builder.h
@@ -297,11 +297,6 @@ public:
         return getOrCreate<ConstantIntVal>(type, value);
     }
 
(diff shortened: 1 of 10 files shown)


csyonghe:
Superseded  by #2996.

Issue #2986, opened 2023-07-14
Binding shifts not applied on HLSL variables with no explicit register


danginsburg:
Using slang v2023.0.6 I am finding that binding shifts are not applied to samplers/textures/etc that do not have an explicit register syntax.

For example, using the command-line:

`slangc.exe test.hlsl -o test.hlsl.spv -profile glsl_450 -entry MainPs -enable-effect-annotations -stage fragment  -fvk-b-shift 0 0 -fvk-s-shift 14 0 -fvk-t-shift 30 0 -fvk-u-shift 158 0`

with the attached shader, g_sColor and g_tTexture will be assigned unshifted bindings:
```
               OpDecorate %g_tColor_0 DescriptorSet 0
               OpDecorate %g_tColor_0 Binding 3
               OpDecorate %g_sColor_0 DescriptorSet 0
               OpDecorate %g_sColor_0 Binding 4
```

However, if I change for example:

```
SamplerState g_sColor;
Texture2D g_tColor; 
```
to:
```
SamplerState g_sColor : register(s0);
Texture2D g_tColor : register(t0); 
```

The correct binding is applied:
```
               OpDecorate %g_tColor_0 DescriptorSet 0
               OpDecorate %g_tColor_0 Binding 30
               OpDecorate %g_sColor_0 DescriptorSet 0
               OpDecorate %g_sColor_0 Binding 14
```
This works with DXC and unfortunately is required for a lot of our shaders.

[test.hlsl.zip](https://github.com/shader-slang/slang/files/12050721/test.hlsl.zip)

csyonghe:
Fixed by #3000 and #3004.

PR #2998, opened 2023-07-17
WIP Support for vk-shift-* without explicit bindings

jsmall-nvidia:
For #2986

Previously the Vulkan binding inference only worked with explicit register bindings. If an HLSL (or vulkan) binding was not present, then a warning was issued, and regular Vulkan layout was used.

This change if vk-shift-* options are enabled, produces a HLSL layout and uses that result to infer Vulkan bindings.

Previously, an HLSL binding was produced *if* outputting to Vulkan *and* there is vk-shift- inference. This caused a problem on the "-combined" test (ie with Sampler2D), because HLSL isn't able to produce such a binding and so causes a crash. The work around, was to only produce HLSL binding lazily.  

That doesn't 100% resolve the issue, because ideally if there were SamplerXD entries, and they had suitable bindings, and others which need to be inferred, this would cause the exception for "not being handled", even though it is handled (through explicit binding).


jsmall-nvidia:
Tess has come up with a better way to do this, so will create a new PR.

PR #2999, opened 2023-07-18
Optimize specialization, and remove unnecessary calls to `simplifyIR`.

csyonghe:
- Optimize specialization pass to reduce the amount of status tracking and avoid over adding insts to work lists.
- Remove some calls to simplifyIR that doesn't affect the final result.
- Delete `hoistConst` pass and use SCCP to serve its purpose (SCCP the consts inside a specialized struct).
- Fix a bug in lowering default value of a enum.
- Allow running peephole optimization only on global scope.
- Make sure a type is ready for lookup before constructing its inheritance list.
- Fix for loop max iteration inference fails inside a generic.


source/slang/slang-check-conformance.cpp
```diff
@@ -104,6 +104,16 @@ namespace Slang
         // For now we are continuing to conflate all the subtype-ish relationships but not
         // tangling convertibility into it.
 
+        // First, make sure both sub type and super type decl are ready for lookup.
+        if (auto subDeclRefType = as<DeclRefType>(subType))
+        {
+            ensureDecl(subDeclRefType->declRef.getDecl(), DeclCheckState::ReadyForLookup);
+        }
```
tangent-vector, line 111: The call to `getInheritanceInfo()` below should end up doing an `ensureDecl()` in the case where the type is a `DeclRefType`, so I'm not 100% sure why this bit would be needed.

tangent-vector (approved):

PR #3004, opened 2023-07-20
Fix for vk-shift-* GLSL emit issue

jsmall-nvidia:
In #3000 led to situations where DescriptorSlots were not being used for binding. Specifically when a shift is enabled for HLSL resource kind, the binding remains on the HLSL kind. This leads to HLSL like LayoutResourceKind in IR being HLSL even when doing GLSL emit.

Parts of the GLSL emit assumed DescriptorSlot is the resource kind used. 

The fix here was to in cases where the distinction could happen, enable looking up for the HLSL kind also.

### DIFF
--- a/source/slang/slang-emit-glsl.cpp
+++ b/source/slang/slang-emit-glsl.cpp
@@ -145,13 +145,13 @@ void GLSLSourceEmitter::_requireGLSLVersion(int version)
     }
 }
 
-// In the description in `usesResourceKind`, it appears it does correctly handle cases where there are 0 offsets. Ie we don't have to traverse through 
-// to know it is used.
-//
+
 // This function was needed because the vk-shift-* change allows kinds which are "in effect" DescriptorSlots but appear as HLSL
 // kinds. 
-// This function should be used passing the "primary"
-static LayoutResourceKind _findUsedResourceKind(IRVarLayout* layout, LayoutResourceKind kind)
+// 
+// This seems safe as a test, because in the description of `usesResourceKind`, it appears that VarLayout require offsets even if
+// it's zero. 
+static LayoutResourceKind _findUsedDescriptorSlotLikeKind(IRVarLayout* layout, LayoutResourceKind kind)
 {
     // DescriptorTableSlot is the most likely, so look for that first
     if (layout->usesResourceKind(LayoutResourceKind::DescriptorTableSlot))
@@ -180,7 +180,7 @@ void GLSLSourceEmitter::_emitGLSLStructuredBuffer(IRGlobalParam* varDecl, IRHLSL
     auto layout = getVarLayout(varDecl);
     if (layout)
     {
-        const LayoutResourceKind usedKind = _findUsedResourceKind(layout, LayoutResourceKind::ShaderResource);
+        const LayoutResourceKind usedKind = _findUsedDescriptorSlotLikeKind(layout, LayoutResourceKind::ShaderResource);
         EmitVarChain chain(layout);
 
         const UInt index = getBindingOffset(&chain, usedKind);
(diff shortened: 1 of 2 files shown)


csyonghe (approved):

source/slang/slang-emit-glsl.cpp
```diff
@@ -310,7 +334,16 @@ void GLSLSourceEmitter::_emitGLSLParameterGroup(IRGlobalParam* varDecl, IRUnifor
     or IRGLSLShaderStorageBufferType which is read write.
     */
 
-    _emitGLSLLayoutQualifier(LayoutResourceKind::DescriptorTableSlot, &containerChain);
+    {
+        // TODO(JS):
+        // The *assumption* here is that we only need to detect the LayoutResourceKind as used on the initial containerChain.varLayout
+        // rather than search up the chain.
```
tangent-vector, line 340: I believe this is correct, yes. The kinds of resources consumed by the leaf variable will only depend on its type, and its type is determined by the layout of the last `varLayout` in the chain.

(The rest of the "chain" is used to summing up all the other relevant offsets)

source/slang/slang-emit-glsl.cpp
```diff
@@ -310,7 +334,16 @@ void GLSLSourceEmitter::_emitGLSLParameterGroup(IRGlobalParam* varDecl, IRUnifor
...
     */
 
-    _emitGLSLLayoutQualifier(LayoutResourceKind::DescriptorTableSlot, &containerChain);
+    {
+        // TODO(JS):
+        // The *assumption* here is that we only need to detect the LayoutResourceKind as used on the initial containerChain.varLayout
+        // rather than search up the chain. 
+        //
+        // Perhaps there is still an issue here, because perhaps it's possible (?) to have a mixture of ConstantBuffer/DescriptorSlot
+        // and in that case the usedKind would just restrict to one and so produce the wrong answer.
```
tangent-vector, line 343: I believe that it would *not* be possible to have a mix of multiple `LayoutResourceKind`s at this point, since a single leaf shader variable of resource type should only have a single `LayoutResourceKind`, and type legalization should have split up any types that would violate that rule (since they'd be illegal in GLSL).

(It might be good to validate this belief by looking at the state inside the debugger, before making any changes based on my assumption...)

source/slang/slang-emit-glsl.cpp
```diff
@@ -310,7 +334,16 @@ void GLSLSourceEmitter::_emitGLSLParameterGroup(IRGlobalParam* varDecl, IRUnifor
...
+        const auto usedKind = _findUsedDescriptorSlotLikeKind(containerChain.varLayout, LayoutResourceKind::ConstantBuffer);
+        _emitGLSLLayoutQualifier(usedKind, &containerChain);
+    }
     _emitGLSLLayoutQualifier(LayoutResourceKind::PushConstantBuffer, &containerChain);
```
tangent-vector, line 347: IIRC, the two different calls to `_emitGLSLLayoutQualifier` in the original code were basically there to address the same kind of concern as this PR: a `ConstantBuffer<...>` might be allocated to use either a `DescriptorTableSlot` *or* a `PushConstantBuffer` for Vulkan.

It might be possible to further clean up this code by making the core `_findUsedDescriptorSlotLikeKind` operation take an array of `LayoutResourceKind`s to iterate over and check for (in order), so that this call site could pass in the three kinds it cares about: `ConstantBuffer`, `PushConstantBuffer`, and `DescriptorTableSlot`.

(A wrapper around that core version of `_findUsedDescriptorSlotLikeKind` could then cover the other call sites from this PR...)

tangent-vector (approved):

PR #3010, opened 2023-07-21
Fix data-flow analysis not propagating diff property through differentiable calls

saipraveenb25:
- Add logic & new decorations to disambiguate calls to differentiable and non-differentiable calls
         (Made `TreatAsDifferentiableDecoration` a function-definiton decoration and introduced `DifferentiableCallDecoration` & `TreatCallAsDifferentiableDecoration` for call-sites.
- Add test

### DIFF
--- a/source/slang/slang-ast-expr.h
+++ b/source/slang/slang-ast-expr.h
@@ -544,7 +544,16 @@ class TreatAsDifferentiableExpr : public Expr
     
     enum Flavor 
     {
+        /// Represents a no_diff wrapper over
+        /// a non-differentiable method.
+        /// i.e. no_diff(fn(...))
+        /// 
         NoDiff,
+
+        /// Represents a call to a method that
+        /// is either marked differentiable, or has
+        /// a user-defined derivative in scope.
+        /// 
         Differentiable
     };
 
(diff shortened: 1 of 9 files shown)


source/slang/slang-ast-expr.h
```diff
@@ -541,6 +541,14 @@ class TreatAsDifferentiableExpr : public Expr
 
     Expr* innerExpr;
     Scope* scope;
+    
+    enum Flavor 
+    {
```
csyonghe, line 546: May need some comment here explaining the meaning of each value.

source/slang/slang-ir-check-differentiability.cpp
```diff
@@ -490,10 +500,10 @@ struct CheckDifferentiabilityPassContext : public InstPassBase
             case kIROp_Call:
             {
                 auto callInst = as<IRCall>(inst);
-                if (callInst->findDecoration<IRTreatAsDifferentiableDecoration>())
-                    continue;
-                if (!isDifferentiableFunc(callInst->getCallee(), DifferentiableLevel::Forward))
+                if (callInst->findDecoration<IRTreatCallAsDifferentiableDecoration>())
                     continue;
+                //if (!isDifferentiableFunc(callInst->getCallee(), DifferentiableLevel::Forward))
```
csyonghe, line 505: commented out code, delete if no longer needed.

csyonghe (dismissed):

source/slang/slang-ast-expr.h
saipraveenb25: Added some comments 👍

csyonghe (approved):

PR #3029, opened 2023-07-26
Refactor `dmul(This, Differential)` to `dmul<T:Real>(T, Differential)`

saipraveenb25:
- Add AST synthesis support for generic containers
 - Refactor relevant tests
 - Add supporting inst: `__realCast`
 - Fix parsing issue with generics that use `__generic<...>` causing them to mismatch with generics that use the other form `fn<...>`


source/slang/diff.meta.slang
```diff
@@ -797,69 +798,95 @@ void __d_cross(inout DifferentialPair<vector<T, 3>> a, inout DifferentialPair<ve
...
     __generic<T : __BuiltinFloatingPointType>                                                \
-    [BackwardDifferentiable][PreferRecompute]                                                \
+    [BackwardDifferentiable] [PreferRecompute]                                               \
     [ForwardDerivativeOf(NAME)]                                                              \
     DifferentialPair<T> __d_##NAME(DifferentialPair<T> dpx)                                  \
-    {                                                                                        \
-        typealias ReturnType = T;                                                            \
-        return DifferentialPair<T>(NAME(dpx.p), FWD_DIFF_FUNC);                              \
-    }                                                                                        \
+    {                                                                                                                                    \
```
csyonghe, line 804: trailing spaces seem to be messed up here.

source/slang/slang-check-decl.cpp
```diff
@@ -2574,6 +2574,135 @@ namespace Slang
         return false;
     }
 
+    GenericDecl* SemanticsVisitor::synthesizeGenericSignatureForRequirementWitness(
```
csyonghe, line 2577: Are there any actual changes made to this function?

saipraveenb25: This is the generic version of the signature synthesis method.
Creates the generic wrapper, clones the parameters & constraints, then calls method synthesis with substituted parameters in place

csyonghe: Ah, I see. Seems like there are code copied from an existing function, is there a way to bottleneck them through the same function?

csyonghe (approved):

source/slang/slang-check-decl.cpp
saipraveenb25: I'm not sure it's possible.. the flow is similar (create empty decl, then clone parameters), but the rest is different (type params need to be explicitly substituted into the constraints, as well as into the decl-ref so that the inner function uses the new type params for its declarations, etc..)

PR #3031, opened 2023-07-27
Generalize collectInductionValues

expipiplus1:
This now handles more complicated cases than what is generated from simple for loops. Also extend the implementation to handle slightly more complicated functions of the loop counter.


tests/autodiff/long-loop-multiple.slang
```diff
@@ -0,0 +1,41 @@
...
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<float> outputBuffer;
+
+[BackwardDifferentiable]
+float sin_series(float x, int iterations)
+{
+    float result = x;
+    float term = x;
+    [MaxIters(30)]
+    for (int i = 1; i < iterations * 10; i += 10)
```
csyonghe, line 14: We can use more tests to ensure the check is at least conservative. Such as what if `i` is modified inside a loop from a branch? Can we handle the case where there are multiple `i+=k` statements in the loop, can we fold them into the same affine transform?

expipiplus1: > Dataflow by any other name would smell as sweet

shouldn't be a problem

expipiplus1:
Ok, I've updated things so that we detect chains of additions/subtractions as well as additions and subtractions in branches

Annoyingly the cfg normalized code from something like `for(...; i += 1){ i+= 2; }` hides the `i+=2` underneath some control flow which is more involved to associate with the break flag, so we don't detect in those cases (and certainly weren't before). We do detect `while(c){ i += 1; i += 1; }` however.

PR #3035, opened 2023-07-28
WIP: Add a CLI option to emit a hash and exit

saipraveenb25:
This is to support persistent-cache stategies in `slangpy` which uses the Slang CLI


expipiplus1 (commented):
I think that throughout this change we could be more specific about what the hash is a hash of.

hash.txt
expipiplus1, line 1: committed by mistake?

source/slang/slang.cpp
```diff
@@ -2629,6 +2635,33 @@ void EndToEndCompileRequest::init()
     m_frontEndReq = new FrontEndCompileRequest(getLinkage(), m_writers, getSink());
 }
 
+SlangResult EndToEndCompileRequest::_writeHash(ComponentType* component)
+{
+    DigestBuilder<SHA1> builder;
+
+    getLinkage()->buildHash(builder, 0);
+
+    // Enumerate all file dependencies and add them to the hash.
```
expipiplus1, line 2644: Ought we include:

- Compiler options
- Compiler version
- Any environment variables

source/slang/slang-options.cpp
```diff
@@ -1879,6 +1881,10 @@ SlangResult OptionsParser::_parse(
                 m_compileRequest->setReportPerfBenchmark(true);
                 break;
             }
+            case OptionKind::ReportHashOnly:
+            {
+                m_compileRequest->setReportHashOnly(true);
```
jsmall-nvidia, line 1886: missing `break;`?

hash.txt
saipraveenb25: Whoops! Removed.

source/slang/slang-options.cpp
saipraveenb25: Yup! Fixed.

saipraveenb25:
I'm actually not sure if we should go ahead with this PR. 
As of now `slangpy` has switched to a depfile-based caching system instead of using shader hashes (https://github.com/shader-slang/slang-python/tree/875d2c5ae520bfec5814162dbf02b452ae06724c), so we don't need the CLI hash option anymore.

That said, if any of you think this could be useful, I can keep working on it.

expipiplus1:
I think if you were to continue then a tool to output just the dependencies of a shader (as a list of file names) would be more generally useful to any build system.

saipraveenb25:
Yes, that makes more sense.

This functionality appears to already exist using the `-depfile` option, and for now `slangpy` seems to work well using only the dependency timestamps to invalidate the cache. 

I'll close this PR for now.

PR #3036, opened 2023-07-31
Use ankerl/unordered_dense as a hashmap implementation

expipiplus1:
Notable changes to Dictionary:
- The subscript operator returns a reference directly to the value,
rather than a lazy ItemProxy (pair of dict pointer and key). The most important consequence is that the idiom `myMap[foo] = myMap.at(bar)` is no longer safe if `foo` doesn't exist in the map already. Consider that the construction of the value at `foo` caused the map to invalidate references, then the reference returned by `bar` is now no longer value; these can be safely replaced with the (more efficient) `myMap.set(foo, myMap.at(bar))`.

Notable changes to Hash:
- We no longer specialize `getHashCode` for `const char*` and `char*`, they are treated as regular pointers now (we were't using this specialization anyway. If desired we can add a static assert in slang-hash.h for this case).
- If we implement a "uniform hash" for a class's `getHashCode` member, then we can set `kHasUniformHash = true` to avoid the map library rehashing to regain uniformity.
- Default to 64 bit hash, which is what the map implementation expects (and hence doesn't require an expansion to 64 bits)

Other:
- The comments for obfuscated module names indicated that a stable hash was desirable, however the non-stable hashing functions were being used in the computation. I've corrected this to use the stable hashing functions, however this has necessarily introduced a change to the output module names. I've made the stable hashing functions operate on distinct types to the regular hashing functions to help prevent this happening again.

We still hash `0.f` and `-0.f` to distinct values, however this warrants reconsideration as it means our hash function is not equality preserving!

slang-profile time (95% over 10 runs):

- Before: 6.3913906 (±0.0746)
- After:  5.9276123 (±0.0964)

~~For a bunch of convenience features also bump to c++20~~

### DIFF
--- a/source/core/slang-dictionary.h
+++ b/source/core/slang-dictionary.h
@@ -76,14 +76,15 @@ namespace Slang
 
     const float kMaxLoadFactor = 0.7f;
 
-    template<typename TKey, typename TValue>
+    template<typename TKey, typename TValue, typename Hash = Slang::Hash<TKey>, typename KeyEqual = std::equal_to<TKey>>
     class Dictionary
     {
         using InnerMap = ankerl::unordered_dense::map<
             TKey,
             TValue,
-            Hash<TKey>>;
-        using ThisType = Dictionary<TKey, TValue>;
+            Hash,
+            KeyEqual>;
+        using ThisType = Dictionary<TKey, TValue, Hash, KeyEqual>;
         InnerMap map;
     public:
         //
--- a/source/core/slang-dictionary.h
+++ b/source/core/slang-dictionary.h
@@ -8,6 +8,7 @@
 #include "slang-exception.h"
 #include "slang-math.h"
 #include "slang-hash.h"
+#include "ankerl/unordered_dense.h"
 
 namespace Slang
 {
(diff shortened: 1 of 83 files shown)


expipiplus1:
Oh, and https://github.com/shader-slang/slang/commit/621863558a3da0507e55cb9b82afcf18c9784651 this is a functional change, previously the map was being updated while it was being iterated over, but that won't do any more so a copy of the keys to traverse is made first.

jsmall-nvidia:
The performance improvement is attractive. 

It does make the compiler broadly rely on a 3rd party container/code. This is in contrast to other external requirements, which either aren't part of the compiler, or are relatively easily separable or optional. 

That container implementation also uses a bunch of additional STL support, including vector. 

Which is to say, I personally don't think we've had the discussion to determine if that's ok.

source/core/slang-dictionary.h
```diff
@@ -75,495 +76,188 @@ namespace Slang
...
         }
-        ItemProxy operator[](TKey&& key) const
-        {
-            return ItemProxy(_Move(key), this);
-        }
-        int getCount() const
+
+        // Returns a reference to the value at the given key. Asserts if the
+        // value doesn't exist
+        TValue& at(const TKey& key)
```
csyonghe, line 177: Can it be `getValue` or `get` to stay consistent with `tryGetValue`?

source/core/slang-file-system.cpp
```diff
@@ -319,11 +319,8 @@ CacheFileSystem::CacheFileSystem(ISlangFileSystem* fileSystem, UniqueIdentityMod
 
 CacheFileSystem::~CacheFileSystem()
 {
-    for (const auto& pair : m_uniqueIdentityMap)
-    {
-        PathInfo* pathInfo = pair.value;
+    for (const auto& [_, pathInfo] : m_uniqueIdentityMap)
```
csyonghe, line 322: Why do we need to modify every iteration to this syntax? Is `KeyValuePair` no longer  a thing we can use explicitly?

csyonghe (approved):

source/core/slang-file-system.cpp
expipiplus1: Not this syntax specifically, but we would have needed to change `key`/`value` to `fst`/`snd`. This was not only a compatible change, but I think makes the code marginally clearer in several places.

PR #3045, opened 2023-08-02
Fix checking of `Differential.Differential == Differential` for user-defined Differential types

saipraveenb25:
- Adjust tests
 - Add diagnostic test


csyonghe:
Can you rebase this on top of tree? Only the new test case is needed.

PR #3049, opened 2023-08-03
Redesign `DeclRef` and systematic `Val` deduplication

csyonghe:
This PR redesigned the `DeclRef` representation, and implemented a centralized deduplication infrastructure for the front-end `Val` class hierarchy similar to how `IRInst`s are deduplicated.

## DeclRef Redesign

Previously, a `DeclRef` contains a pointer to the referenced Decl, and a linked list of `Substitution`s that applies to the referenced `Decl` to provide generic arguments or the concrete `This` type for an existential type. This works for most scenarios that we support, but the representation does not map very well to how the decl is being accessed and leads to confusion for complex scenarios. In this change, we use different subtypes of `DeclRefBase` to represent different kinds of `DeclRef`s:
- `DirectDeclRef` represents a simple direct reference to a `Decl`, it simply stores a `Decl*` pointer, e.g. `A` is represented as `DirectDeclRef(A)`.
- `MemberDeclRef` represents a static lookup of member of a parent container `Decl`, e.g. `A.M` is represented as `MemberDeclRef(DirectDeclRef(A), M)`
- `GenericAppDeclRef` represents a specailization of a generic. For example, `A<T1, T2>` is represented as `GenericAppDeclRef(A.inner, DirectDeclRef(A), [T1,T2])`.
- `LookupDeclRef` represents a dynamic lookup from a witness. For example, `T.method()` where `method` is provided by `IFoo`, and `T:IFoo`, is represented as `LookupDeclRef(DeclRefType(T), witness(T:IFoo), DirectDeclRef(IFoo::method))`.

We then implement several simplification/canonicalization rules so that DeclRefs has unique representations. These rules include:
- `Member(Direct(A), B)` ==> `B` 
- `Member(Lookup(w, This), x)` ==> `Lookup(w, x)`
- `Member(Lookup(w, assocType), assocTypeConstraint)` ==> `Lookup(w, assocTypeConstraint)`
- `Member(Lookup(w, extensionOfSomeInterface), extendedMethod)` ==> `Lookup(w, extendedMethod)`

The first rule is simple, since static member lookup does not need to store any info along the path, so we can just fold it into a single `DirectDeclRef`.

The second and third rule is a result of how `This` type and associated type constraint are represented in the AST. Basically if we are trying to obtain a member from `This` or associated type, we are really looking up through the witness behind it.

The last rule is a bit adhoc due to the hand-wavy way that interface extensions are represented in the AST. They really should be expressed in the form of
```
generic<T>
extension T : IFoo { void newMethod(); }
```
Instead of the current adhoc syntax of `extension IFoo{...}`. So we have to use a `Lookup(w, newMethod)` to represent a declref to the `newMethod`, where the `ThisType` info is encoded in the `LookupDeclRef`.

Following this DeclRef representation, we also made `This` an actual Decl in an `interface` decl that can be discovered by our normal lookup mechanism. The `This` type is defined as an `AggTypeBase`  that has a `ThisTypeConstraint` member that allows further lookups through `This`, because our inheritance list flattening logic will see the `ThisTypeConstraint` and treat the interface type as the super type of `This`. This also get rid of special handling logic around `ThisType`, and `ThisTypeSubtypeWitness` simply becomes a `DeclaredSubtypeWitness` that references the `ThisTypeConstraint` decl.

## `Val` Deduplication

Because `DeclRef`s are `Val`s, we benefit greatly from the being able to deduplicate them. This change hardens the handling of `Val` deduplication by making all `Val`s like `IRInst`s, where a `Val` is just an opcode (`ASTNodeType`) + a list of operands which are themselves either constants or `Val`s. No subtypes of `Val` should retain their own fields, and instead just provides accessors to the operands array defined in `Val`. This allows us to use a centralized implementation for canonicalization (through the `resolve` method as well as `equals` and `getHashCode` implementation. Because `Val`s are deduplicated, `equals` implementation is simply comparing the pointer of two resolved `Val`s.

Extra care is taken to ensure no duplicate `Val`s exists within a single deserialized AST. This is done by hijacking the pointer serialization logic for anything that derives from `Val*`, and write the resolved `Val` instead. The tricky part is when module A imports module B, the types defined in B will be included in both the serialized A and the serialized B. When we deserialize A, we want to avoid recreating the duplicate Vals. This is done by implementing a "fixup" step after AST deserialization that replaces all `Val*` fields in the deserialized AST with the resolved `Val`. In order to track all the addresses of a `Val*` field in the deserialized AST, we hijack the `deserialize` method of `Val` derived pointer types and store the address out into `SerialReader::m_fixups`. For this method to work, we need to make sure the address that hold the `Val*` in the final AST is the same as the address where deserialize is first called to obtain the `Val*` value. This is true for most value types, arrays and `List`, but not true for `OrderedDictionary`. We therefore removed all `OrderedDictionary` whose key or value type is a `Val*` from the AST definitions and replaced them with a `List`. We instead will build a runtime-only unreflected `OrderedDictionary` from the `List` on-demand.

## Other Changes
- Delete everything related to `TaggedUnion`s.
- Make ASTBuilder available as a `thread_local` global so they don't need to be passed around explicitly. The global is set and reset upon entry/exit of API entrypoints where a linkage is available.
- Unify various type-specific cases of `DeclRefType::create` and make them all follow the same logic.
- `Type::getCanonicalType` now just calls into `Val::resolve` and no longer takes `SemanticVisitor*` argument, since it never should.
- Only one `ASTBuilder` is created for a `Linkage`, instead of creating one `ASTBuilder` per module to allow `Val` deduplication across the whole linkage.


source/slang/core.meta.slang
```diff
@@ -957,12 +957,12 @@ for (int tt = 0; tt < kTypeCount; ++tt)
 
 __generic<T>
 __intrinsic_type($(kIROp_ConstantBufferType))
-__magic_type(ConstantBuffer)
+__magic_type(ConstantBufferType)
 struct ConstantBuffer {}
```
tangent-vector, line 961: One nice cleanup might be to allow for the `__magic_type` modifier to be used without an operand, in which case it would derive the corresponding C++ class name as `<nameOfDecl> + Type`.

source/slang/core.meta.slang
```diff
@@ -1238,14 +1238,14 @@ extension matrix<T, R, C> : IDifferentiable
...
-__magic_type(SamplerState, $(int(SamplerStateFlavor::SamplerState)))
+__magic_type(SamplerStateType, $(int(SamplerStateFlavor::SamplerState)))
 __intrinsic_type($(kIROp_SamplerStateType))
 struct SamplerState
 {
 }
 
     /// Sampling state for filtered texture fetches that include a comparison operation before filtering.
-__magic_type(SamplerState, $(int(SamplerStateFlavor::SamplerComparisonState)))
+__magic_type(SamplerStateType, $(int(SamplerStateFlavor::SamplerComparisonState)))
```
tangent-vector, line 1248: Unrelated to this PR: Passing a "flavor" through on these seems like overkill. We might as well just have the comparison and non-comparison cases be distinct AST node classes (with a common base).

source/slang/core.meta.slang
```diff
@@ -1347,12 +1347,12 @@ struct TextureTypeInfo
...
         {
-            sb << "__magic_type(TextureSampler," << int(flavor) << ")\n";
+            sb << "__magic_type(TextureSamplerType," << int(flavor) << ")\n";
             sb << "__intrinsic_type(" << (kIROp_TextureSamplerType + (int(flavor) << kIROpMeta_OtherShift)) << ")\n";
         }
         else
         {
-            sb << "__magic_type(Texture," << int(flavor) << ")\n";
+            sb << "__magic_type(TextureType," << int(flavor) << ")\n";
             sb << "__intrinsic_type(" << (kIROp_TextureType + (int(flavor) << kIROpMeta_OtherShift)) << ")\n";
```
tangent-vector, line 1356: Similarly, having the "flavor" bits here for texture types is increasingly seeming like an unnecessary complication.

Eventually I think we will want to shift to having a single generic type that all textures bottleneck through, such that the things currently encoded in the "flavor" bits can become explicit generic parameters/arguments.

Something like:
```
enum TextureShape { 2D, 3D, ... 2DArray, ... }
enum TextureAccess { ReadOnly, WriteOnly, ReadWrite, Append, Consume, ... }
enum Format { /* DXGI_FORMAT-style stuff... */ }
interface ITexelType { ... } // conformances for floatN, intN, etc.
interface ITextureFormat
{
    associatedtype Texel : ITexelType;
    associatedconst format : TextureFormat;
}

struct __Texture< let A : TextureAccess, let S : TextureShape, F : ITextureFormat >
{ ... }
```

source/slang/slang-ast-base.h
```diff
@@ -116,6 +124,177 @@ class SyntaxNodeBase : public NodeBase
     SourceLoc loc;
 };
 
+enum class ValNodeOperandKind
+{
+    ConstantValue,
+    ValNode,
+    ASTNode,
```
tangent-vector, line 131: It's interesting, because we really do have the makings of a strong distinction between the *syntax* nodes (inherently have source locations, etc.) and the `Val`s (which don't).

Since syntax nodes represent actual fragments of source code (or at least can), they don't really *want* deduplication (or we could have them hold a deduplicated "body" along with a single starting source location...).

In contrast, it should always be reasonable to deduplicate `Val`s, even if we don't end up relying on the deduplication for identity testing.

source/slang/slang-ast-base.h
```diff
@@ -140,25 +367,84 @@ class Val : public NodeBase
...
+    {
+        addOrAppendToNodeList(m_operands, args...);
+    }
+    template<typename T>
+    void addOperands(OperandView<T> operands)
+    {
+        for (auto v : operands)
+            m_operands.add(ValNodeOperand(v));
+    }
+    List<ValNodeOperand> m_operands;
```
tangent-vector, line 432: This is great. You've made the whole `Val` hierarchy much more uniform. We could theoretically extend this to the entire AST with a bit (okay... a lot) more work.

source/slang/slang-ast-base.h
```diff
@@ -514,6 +703,10 @@ class Decl : public DeclBase
         SLANG_RELEASE_ASSERT(state >= checkState.getState());
         checkState.setState(state);
     }
+
+private:
+    SLANG_UNREFLECTED DeclRefBase* m_defaultDeclRef = nullptr;
+    SLANG_UNREFLECTED Index m_defaultDeclRefEpoch = -1;
```
tangent-vector, line 709: I may have scanned past it all, but are there comments that would help somebody new to the code understand how these "epochs" work?

source/slang/slang-ast-builder.h
```diff
@@ -435,17 +434,17 @@ class ASTBuilder : public RefObject
...
+    FuncType* getFuncType(ArrayView<Type*> parameters, Type* result, Type* errorType = nullptr);
 
     TypeType* getTypeType(Type* type);
 
         /// Produce a witness that `T : T` for any type `T`
     TypeEqualityWitness* getTypeEqualityWitness(
         Type* type);
 
-    SubtypeWitness* getDeclaredSubtypeWitness(
+    DeclaredSubtypeWitness* getDeclaredSubtypeWitness(
```
tangent-vector, line 447: I think I left this return type as `SubtypeWitness` because of the fact that if there is a constraint like `T : IFoo & IBar` we might want to decompose that ahead of time, by extracting the two sub-witnesses for `T : IFoo` and `T : IBar` and then re-composing them into a conjunction.

The approach I ended up using for canonicalization/simplification of subtype witnesses ends up not needing/wanting that step, but I was hoping that we could avoid downstream code accidentally depending on the exact class of AST node that gets returned by these various factory functions.

(All that said, I'm assuming the signature changed here because some other code needed the specificity...)

source/slang/slang-ast-builder.h
```diff
@@ -529,9 +526,30 @@ class ASTBuilder : public RefObject
     SharedASTBuilder* m_sharedASTBuilder;
 
     MemoryArena m_arena;
+};
+
+// Retrieves the ASTBuilder for the current compilation session.
+ASTBuilder* getCurrentASTBuilder();
```
tangent-vector, line 532: I'll be interested to see if this is handled via a thread-local, or what...

source/slang/slang-ast-decl.cpp
```diff
@@ -118,4 +119,21 @@ bool isLocalVar(const Decl* decl)
...
+        }
+    }
+    SLANG_UNREACHABLE("InterfaceDecl does not have a ThisType decl.");
+}
+
+InterfaceDecl* ThisTypeConstraintDecl::getInterfaceDecl()
+{
+    return as<InterfaceDecl>(parentDecl->parentDecl);
+}
+
```
tangent-vector, line 138: I'm happy to see this cleanup be put in place. I think this is a much more sound way for us to be modeling the `This` type inside of an `interface`, and I believe it sets the stage for us to be better able to handle proper polymorphism in the context of `class` types (or even `struct`s, if we want to allow them to `override` inherited method implementations).

source/slang/slang-ast-expr.h
```diff
@@ -560,17 +560,6 @@ class TreatAsDifferentiableExpr : public Expr
...
-    ///
-    /// An expression of this form will resolve to a `TaggedUnionType`
-    /// when checked.
-    ///
-class TaggedUnionTypeExpr: public Expr
-{
-    SLANG_AST_CLASS(TaggedUnionTypeExpr)
-    List<TypeExp> caseTypes;
-};
-
```
tangent-vector, line 573: Good riddance!

source/slang/slang-ast-type.cpp
```diff
@@ -199,41 +95,54 @@ Val* DeclRefType::_substituteImplOverride(ASTBuilder* astBuilder, SubstitutionSe
...
-    // the case we especially care about is when this type references a declaration
-    // of a generic parameter, since that is what we might be substituting...
-    if (auto genericTypeParamDecl = as<GenericTypeParamDecl>(declRef.getDecl()))
+    int diff = 0;
+    DeclRef<Decl> substDeclRef = getDeclRef().substituteImpl(astBuilder, subst, &diff);
+
+    // If this declref type is a direct reference to ThisType or a Generic parameter,
+    // and `subst` provides an argument for it, then we should just return that argument.
+    //
+    if (as<DirectDeclRef>(substDeclRef.declRefBase))
```
tangent-vector, line 104: Why is it necessary to check that the decl-ref is a `DirectDeclRef` here?

I know it isn't an issue right now, but if we eventually allowed for generics to take parameters of higher kinds, I could imagine non-direct references to parameters:

```
struct Array<T> { ... }
struct List<T> { ... }

func map< X, Y, C<T> : IContainer>(
    input: C<X>,
    f: X->Y)
    -> C<Y>
{ ... }
```

In that case, we have references to `C<X>` and `C<Y>` that are referring to the parameter as part of a decl-ref that has generic arguments.

I guess if we start to go down that road we'd probably need to fold the substitution that is happening here into the core `substituteImpl` for `DeclRefBase`, at which point we might have to deal with cases where substitution on a decl-ref yields something that isn't a decl-ref (but is still a `Val`)...

tangent-vector (approved):
I cannot and will not pretend to have read every line, but what I've seen looks like what I'd expect.

This is a massive cleanup, and I hope it pays off for us going forward.

source/slang/slang-ast-builder.h
csyonghe: It's a thread_local.

PR #3050, opened 2023-08-07
Fix `Val` deduplication bug.

csyonghe:
- Use SFINAE to override `SerializeTypeInfo<T*>` for certain derived types that works more robustly for value type members.
- Fix several missing calls to `resolve`.
- Concatenate all stdlib sources into a single module before compiling and serializing it so we don't have to deduplicate Vals within stdlibs upon deserialization.


source/slang/slang-serialize-ast-type-info.h
```diff
@@ -66,6 +64,25 @@ inline void deserializePointerValue(SerialReader* reader, const SerialIndex* inS
     }
 }
 
+template<typename T>
+struct PtrSerialTypeInfo<T, std::enable_if_t<std::is_base_of_v<Val, T>>>
```
expipiplus1, line 68: This will be true if it's a private inheritance (although there's on clean way around that until c++20 :/). tbh I don't think it matters though

PR #3051, opened 2023-08-07
Correct return type for emitFieldAddress

expipiplus1:
This should not be returning UInt, it should be a pointer type


csyonghe (approved):
I remember fixing this before, maybe it is just another copy pasted location.

PR #3054, opened 2023-08-07
Default target intrinsic should not apply to SPIR-V

expipiplus1:
They're never appropriate as they're not SPIR-V Opcodes

We could perhaps do with a "Textual" superclass for the existing ones?


expipiplus1:
mm, I think that because these now have explicit capabilities, they're being ranked higher than they were causing some tests to fail.

csyonghe (approved):

PR #3064, opened 2023-08-07
SPIR-V WIP

expipiplus1:
About half of the tests in tests/compute are passing

This also includes the slew of other PRs opened today, the diff might be a
little more manageable once they're merged.

Before merging this (at least)

- [ ] Need to make the 'spirv direct by default' mode a compile time switch
- [ ] Need to add spirv/old-glsl versions of existing tests, so that both remain checked
- [ ] Fix the few non spir-v tests which have regressed
- [ ] Transcribe various TODOs to issues


tests/compute/loop-unroll.slang
```diff
@@ -4,7 +4,7 @@
 //TEST(compute):COMPARE_COMPUTE:-cpu -shaderobj
 //TEST(compute):COMPARE_COMPUTE:-cuda -shaderobj
 // Note VK output is not loop unrolled
-//TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
```
csyonghe, line 7: This will also disable the test for glslang path. Can we figure out a way to just disable the spirv path?

expipiplus1: yeah, there are a few oddities like this at the moment. I suggest disregard them for now until this is in a mergeable state.

source/slang/slang-emit-spirv.cpp
```diff
@@ -807,101 +918,104 @@ struct SPIRVEmitContext
...
-        InstConstructScope scopeInst(this, opcode, irInst);
-        SpvInst* spvInst = scopeInst;
-        emitOperand(a);
-        emitOperand(b);
-        emitOperand(c);
-        parent->addInst(spvInst);
-        return spvInst;
-    }
+        // Hash the whole global stack and opcode
+        SpvTypeInstKey key;
```
csyonghe, line 990: For better performance, we might want to use the spvInst itself as the storage for the opcode+operands instead of using a List as dictionary key, similar to how `IRInstKey` and `ValKey` are implemented.
```
struct SpvTypeInstKey
{
      SpvInst* inst;
      bool operator == (SpvTypeInstDesc desc) { ... }
}

struct SpvTypeInstDesc
{
       SpvOp opCode;
       ArrayView<SpvWord> operands;
}
```

source/slang/slang-emit-spirv.cpp
```diff
@@ -1745,9 +1884,22 @@ struct SPIRVEmitContext
                 });
             }
         case kIROp_Unreachable:
-            return emitInst(parent, inst, SpvOpUnreachable);
+            return emitOpUnreachable(parent, inst);
         case kIROp_conditionalBranch:
             SLANG_UNEXPECTED("Unstructured branching is not supported by SPIRV.");
+        case kIROp_MakeVector:
```
csyonghe, line 1890: For constructors, we also need:
- VectorReshape
- MatrixReshape
- MakeMatrix
- MakeMatrixFromScalar
- MakeArrayFromElement

We also need these type casts:
- IntCast
- FloatCast
- CastFloatToInt
- CastIntToFloat
Currently these casts can all have vector and matrix operands, so we may need to lower that first.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -182,6 +196,30 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
         }
     }
 
+    // Replace getElement(x, i) with, y = store(x); p = getElementPtr(y, i); load(p)
+    // SPIR-V has no support for dynamic indexing into values like we do.
+    // It may be advantageous however to do this further up the pipeline
+    void processGetElement(IRGetElement* inst)
```
csyonghe, line 202: This is unfortunate, we could probably use an earlier pass to de-SSA composite typed values.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -232,13 +270,34 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
 
     void processStructuredBufferType(IRHLSLStructuredBufferTypeBase* inst)
     {
+        // const auto elementTypeLayout = inst->getElementType()->findDecoration<IRTypeLayout>();
```
csyonghe, line 273: commented code.

source/slang/slang-ir-spirv-snippet.h
```diff
@@ -125,7 +128,7 @@ struct SpvSnippet : public RefObject
     };
 
     List<ASMInst> instructions;
-    List<SpvStorageClass> usedResultTypeStorageClasses;
+    List<SpvStorageClass> usedPtrResultTypeStorageClasses;
```
csyonghe, line 131: Should this be deduplicated?

tests/gpu-feature/texture/query/footprint/nv-shader-texture-footprint-simple.slang
```diff
@@ -1,7 +1,9 @@
-//TEST:SIMPLE(filecheck=SPIRV):-target spirv-assembly -entry fragmentMain -stage fragment
+//DISABLED_TEST:SIMPLE(filecheck=SPIRV):-target spirv-assembly -entry fragmentMain -stage fragment
 //DISABLED_TEST:SIMPLE(filecheck=DXIL):-target dxil-assembly -entry fragmentMain -stage fragment
 //TEST:SIMPLE(filecheck=HLSL):-target hlsl -entry fragmentMain -stage fragment
 
+// TODO: The SPIRV test is fine locally, but seems to lead to an impossible to debug hang in CI
```
csyonghe, line 5: We will need to get to the bottom of this.

tools/gfx/vulkan/vk-shader-program.cpp
```diff
@@ -74,9 +74,10 @@ Result ShaderProgramImpl::createShaderModule(
     // to control the entry point naming behavior in SPIRV-direct path
     // so we can remove the ad-hoc logic here.
     auto realEntryPointName = entryPointInfo->getNameOverride();
-    const char* spirvBinaryEntryPointName = "main";
-    if (m_device->m_desc.slang.targetFlags & SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY)
-        spirvBinaryEntryPointName = realEntryPointName;
+    const char* spirvBinaryEntryPointName =
```
csyonghe, line 77: Our spirv backend should respect the new `-fvk-use-entrypoint-name` option and use "main" if the option is not set.

csyonghe (approved):

source/slang/slang-emit-spirv.cpp
```diff
@@ -807,101 +918,104 @@ struct SPIRVEmitContext
...
-        SpvInst* spvInst = scopeInst;
-        emitOperand(a);
-        emitOperand(b);
-        parent->addInst(spvInst);
-        return spvInst;
-    }
+        List<SpvWord> ourOperands;
+        {
+            auto scopePeek = OperandMemoizeScope(this);
+            (emitOperand(ops), ...);
```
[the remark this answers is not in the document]
expipiplus1: Case number 1 here https://en.cppreference.com/w/cpp/language/fold, with the comma operator. It saves having to write some template recursion.

source/slang/slang-emit-spirv.cpp
expipiplus1: I think the casts are currently handled elsewhere in this module (although we will need to lower the matrix ops)

source/slang/slang-ir-spirv-legalize.cpp
expipiplus1: definitely, a more holistic pass could generate much better code in the case of several indexed reads.

source/slang/slang-emit-spirv.cpp
expipiplus1: Good idea, I wonder if we could also use our IRInst as an initial guess, I don't think we can generate a different SPIR-V instruction for the same IRInst ever

PR #3069, opened 2023-08-08
Clean up and improve Val deduplication performance.

csyonghe:


source/slang/slang-ast-builder.h
```diff
@@ -112,13 +112,55 @@ class SharedASTBuilder : public RefObject
...
+    Val* val;
+    HashCode hashCode;
+    ValKey() = default;
+    ValKey(Val* v)
+    {
+        val = v;
+        Hasher hasher;
+        hasher.hashValue(v->astNodeType);
+        for (auto& operand : v->m_operands)
+            hasher.hashValue(operand.values.intOperand);
```
jsmall-nvidia, line 126: I'm assuming this is okay, in a scenario where ptr is smaller than int64_t, because it is a requirement the other bits are zeroed.

jsmall-nvidia (approved):

PR #3089, opened 2023-08-10
Add support for ConstBufferPointer on Vulkan.

csyonghe:
This change adds a `ConstBufferPointer` type that maps to a `buffer_reference` on Vulkan.

Usage:
```
struct MyStruct
{
    float4 position;
    float4x4 transform;
}

struct Globals
{
    ConstBufferPointer<MyStruct> pStruct;
}

ConstantBuffer<Globals> gGlobals;

[numthreads(1,1,1)]
void main(int3 tid: SV_DispatchThreadID)
{
    MyStruct s = gGlobals.pStruct.get();
    outputBuffer[tid.x] = uint(s.position.x);
}
```

### DIFF
--- a/source/slang/slang-ir.h
+++ b/source/slang/slang-ir.h
@@ -1641,6 +1641,13 @@ struct IRRTTIPointerType : IRRawPointerTypeBase
     IR_LEAF_ISA(RTTIPointerType)
 };
 
+struct IRHLSLConstBufferPointerType : IRPtrTypeBase
+{
+    IR_LEAF_ISA(HLSLConstBufferPointerType)
+    IRInst* getElementType() { return getOperand(0); }
+    IRInst* getAlignment() { return getOperand(1); }
+};
+
 struct IRGlobalHashedStringLiterals : IRInst
 {
     IR_LEAF_ISA(GlobalHashedStringLiterals)
--- a/source/slang/slang-ir.h
+++ b/source/slang/slang-ir.h
@@ -1644,8 +1644,7 @@ struct IRRTTIPointerType : IRRawPointerTypeBase
 struct IRHLSLConstBufferPointerType : IRPtrTypeBase
 {
     IR_LEAF_ISA(HLSLConstBufferPointerType)
-    IRInst* getElementType() { return getOperand(0); }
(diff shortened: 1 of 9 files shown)


source/slang/slang-ir.h
```diff
@@ -1641,6 +1641,13 @@ struct IRRTTIPointerType : IRRawPointerTypeBase
     IR_LEAF_ISA(RTTIPointerType)
 };
 
+struct IRHLSLConstBufferPointerType : IRPtrTypeBase
+{
+    IR_LEAF_ISA(HLSLConstBufferPointerType)
+    IRInst* getElementType() { return getOperand(0); }
```
jsmall-nvidia, line 1647: We already have `getValueType` on IRPtrTypeBase, is the use of element here to in some way distinguish this use case? (Say value might imply a single value, where element implies some amount of?)

csyonghe: Ah, you are right, I'll remove it.

source/slang/slang-ir.h
```diff
@@ -1641,6 +1641,13 @@ struct IRRTTIPointerType : IRRawPointerTypeBase
...
+    IRInst* getAlignment() { return getOperand(1); }
+};
```
jsmall-nvidia, line 1649: The alignment here is presumably the "element"/"value" alignment? Assuming that's right having it in the name might add to clarity.

jsmall-nvidia: Okay I see ```All buffer reference addresses used for a particular buffer
      reference type are assumed to be aligned to this alignment value. That
      is, the base of the block must be aligned, and members of the block can
      be aligned within the block using standard layouts and offset layout
      qualifiers.```

So perhaps it should have name like Base alignment or something like that.

jsmall-nvidia (dismissed):

source/slang/slang-ir.h
csyonghe: Renaming done.

source/slang/slang-ir.h
csyonghe: Fixed.

jsmall-nvidia (approved):

PR #3102, opened 2023-08-14
Add auto-diff support for `IRSwizzleStore`

saipraveenb25:
- Lower IRSwizzleStore to multiple stores during AD preprocess.
 - Fix typo in `transcribeNonDiffInst`


out.glsl
```diff
@@ -0,0 +1,49 @@
+#version 450
```
csyonghe, line 1: unrelated?

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -1617,6 +1617,54 @@ void insertTempVarForMutableParams(IRModule* module, IRFunc* func)
     }
 }
 
+bool isResourcePointer(IRInst* ptrInst)
+{
+    if (auto getElementPtr = as<IRGetElementPtr>(ptrInst))
```
csyonghe, line 1622: You can use `getRootAddr` here.

csyonghe (approved):

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -1617,6 +1617,50 @@ void insertTempVarForMutableParams(IRModule* module, IRFunc* func)
     }
 }
 
+bool isLocalPointer(IRInst* ptrInst)
+{
+    // If it's not a local var or a function parameter, then it's probably 
+    // referencing something outside the function scope.
+    // 
+    return as<IRVar>(getRootAddr(ptrInst)) || as<IRParam>(getRootAddr(ptrInst));
```
csyonghe, line 1625: You may want to call getRootAddr once.

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -1626,6 +1670,8 @@ SlangResult ForwardDiffTranscriber::prepareFuncForForwardDiff(IRFunc* func)
     
     initializeLocalVariables(autoDiffSharedContext->moduleInst->getModule(), func);
 
+    lowerSwizzledStores(autoDiffSharedContext->moduleInst->getModule(), func);
```
csyonghe, line 1674: It will be beneficial to move this to cover all functions and not just the AD ones. We can run the pass again at the end of linkAndOptimizeIR, before hitting emit. A good place to call it might be after `finalizeSpecialization`.

csyonghe (dismissed):

source/slang/slang-ir-autodiff-fwd.cpp
csyonghe: A pitfall though is if I have a function taking an `__ref` parameter and `swizzleStore` it, the function will get inlined later, and we probably don't want to lower the swizzled store on that parameter just yet.

saipraveenb25: Hmm.. should we run the pass again after those methods are inlined?

csyonghe: That sort of defeats the purpose of the pass, which is removing the concern of `swizzledStore` throughout the codebase...

csyonghe: I guess its fine to just run the pass for AD then...

csyonghe: It looks like the right choice is to implement a targeted optimization in image load/store lowering pass to detect and combine multiple stores, instead of trying to preserve swizzledStore all over the place.

csyonghe (approved):

PR #3104, opened 2023-08-15
Remove test for removed option

expipiplus1:


csyonghe (changes requested):
Actually, this test is not just testing the option.
We should remove the `-use-glsl-matrix-layout-modifier` option from the //TEST line and change `row_major` to `column_major` in `MyStruct` declaration.

csyonghe (approved):

PR #3106, opened 2023-08-15
Fix ByteAddressBuffer array legalization.

csyonghe:


source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -74,6 +74,8 @@ struct ByteAddressBufferLegalizationContext
 
     void processGetEquivalentStructuredBuffer(IRInst* inst)
     {
+        m_builder.setInsertBefore(inst);
```
jsmall-nvidia, line 77: The reason for this builder change is otherwise it will point to a destroyed inst?

jsmall-nvidia (approved):

source/slang/slang-ir-byte-address-legalize.cpp
csyonghe: Yes.

PR #3111, opened 2023-08-15
Initial support for differentiating existential types

saipraveenb25:
We can now declare interfaces as conforming to `IDifferentiable` and use existential parameters and variables in differentiable functions. 

Supporting changes in this PR:
- Any-value-size inference pass that determines a suitable maximum size for interfaces that don't have a user-defined any-value-size. This is the primary method in which we determine the size for `IDifferentiable`.
- Mark `ExtractExistentialType` and `ExtractExistentialWitnessTable` as hoistable
- Handle 'un-storable' insts during the primal-availability step by treating them as 'transparent' and cloning them as necessary.
- Bugfixes in `cloneInst` and `constructSSA` related to hoistable insts.
- Bugfixes related to lowering differentiable this types

For example, we can create a differentiable interface `IInterface`

```
interface IInterface : IDifferentiable
{
    [Differentiable]
    float calc(float x);
}
```
and declare different implementations that can carry different data layouts and method implementations

```

struct A : IInterface
{
    float data1;

    [Differentiable]
    float calc(float x) { return x * x * x * data1; }
};

struct B : IInterface
{
    float data1;
    float data2;

    [Differentiable]
    float calc(float x) { return x * x * data1 * data2; }
};
```

We can now use `IInterface` as a dynamic parameter whose type is a run-time value:

```
[Differentiable]
float doThing(IInterface obj, float x)
{
    return obj.calc(x);
}
```

What's left (things that cannot be done currently):
 - We cannot currently create a differentiable struct that contains a differentiable interface member
 For example:
 ```
 interface IInterface : IDifferentiable 
 {
    /*...*/
 };
 struct MyStruct : IDifferentiable
{
      IInterface obj;
}
```
This will create a circular dependency where `MyStruct.Differential` simultaneously implements `IDifferentiable` and **contains** a field that implements `IDifferentiable` making it impossible to determine an any-value-size.

### CODE REVIEW

File: source/slang/diff.meta.slang [hunk 1]
```diff
@@ -25,6 +25,30 @@ attribute_syntax [DerivativeMember(memberName)] : DerivativeMemberAttribute;
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [NoDiffThis] : NoDiffThisAttribute;
 
+// A 'none-type' that acts as a run-time sentinel for zero differentials.
+public struct NullDifferential : IDifferentiable
+{ 
+    // for now, we'll use at least one field to make sure the type is non-empty
+    uint dummy;
```
csyonghe, line 32: Why is it necessary to make the type non-empty?

source/slang/slang-compiler.h
```diff
@@ -2671,7 +2671,7 @@ namespace Slang
         virtual SLANG_NO_THROW void SLANG_MCALL setDebugInfoFormat(SlangDebugInfoFormat format) SLANG_OVERRIDE;
         virtual SLANG_NO_THROW void SLANG_MCALL setReportDownstreamTime(bool value) SLANG_OVERRIDE;
         virtual SLANG_NO_THROW void SLANG_MCALL setReportPerfBenchmark(bool value) SLANG_OVERRIDE;
-
+
```
csyonghe, line 2674: delete?

source/slang/slang-ir-any-value-inference.cpp
```diff
@@ -0,0 +1,231 @@
...
+                    return;
+                }
+            }
+            break;
+        case kIROp_StructType:
+            {
+                auto structType = cast<IRStructType>(type);
+                for (auto field : structType->getFields())
+                {
+                    _findDependenciesOfTypeInSet(field->getFieldType(), targetSet, result);
```
csyonghe, line 32: While recursion is convenient, we should try to use worklists whenever possible to avoid stack related issues.

source/slang/slang-ir-any-value-inference.cpp
```diff
@@ -0,0 +1,231 @@
...
+        _findDependenciesOfTypeInSet(type, targetSet, result);
+
+        return result;
+    }
+
+    void _sortTopologically(
+        IRInterfaceType* interfaceType,
+        HashSet<IRInterfaceType*>& visited,
+        List<IRInterfaceType*>& sortedInterfaceTypes,
+        const Func<HashSet<IRInterfaceType*>, IRInterfaceType*>& getDependencies)
```
csyonghe, line 60: `Func` comes with a virtual call, if possible prefer templates.

source/slang/slang-ir-any-value-inference.cpp
```diff
@@ -0,0 +1,231 @@
...
+        //
+
+        HashSet<IRInst*> implementedInterfaces;
+        // Add all interface type that are implemented by at least one type to a set.
+        for (auto inst : module->getGlobalInsts())
+        {
+            if (inst->getOp() == kIROp_WitnessTable)
+            {
+                auto interfaceType = cast<IRWitnessTableType>(inst->getDataType())->getConformanceType();
+                implementedInterfaces.add(interfaceType);
```
csyonghe, line 108: It doesn't appear to me that this step is very beneficial since we are going to go over the uses of an interface type later to collect all the implementations. If we found no implementation there we can just skip.

source/slang/slang-ir-any-value-marshalling.cpp
```diff
@@ -782,6 +783,46 @@ namespace Slang
...
+            auto key = type->getOperand(1);
+            IRType* assocType = nullptr;
+            if (auto witnessTableType = as<IRWitnessTableTypeBase>(witnessTableVal->getDataType()))
+            {
+                auto interfaceType = as<IRInterfaceType>(witnessTableType->getConformanceType());
+
+                // Walk through interface operands to find a match, the result should be an 
+                // associated type entry.
+                //
+                for (UIndex ii = 0; ii < interfaceType->getOperandCount(); ii++)
```
csyonghe, line 798: UInt

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -1867,17 +1892,26 @@ InstPair ForwardDiffTranscriber::transcribeInstImpl(IRBuilder* builder, IRInst*
 
     case kIROp_ExtractExistentialValue:
         return transcribeSingleOperandInst(builder, origInst);
+
+    case kIROp_PackAnyValue:
```
csyonghe, line 1896: Do we need to handle UnpackAnyValue?

source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -1242,23 +1247,112 @@ static int getInstRegionNestLevel(
...
+        if (!passthroughInst(inst))
+        {
+            return List<UseChain>();
+        }
+
+        // Recurse.
+        List<UseChain> result;
+        for (auto use = inst->firstUse; use; use = use->nextUse)
+        {
+            List<UseChain> innerChain = from(use, isRelevantUse, passthroughInst);
```
csyonghe, line 1281: What if there is a cyclic use chain?

source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -1242,23 +1247,112 @@ static int getInstRegionNestLevel(
     return (int)result;
 }
 
+
+struct UseChain
+{
+    List<IRUse*> chain;
+    static List<UseChain> from(
+        IRUse* baseUse,
+        Func<bool, IRUse*> isRelevantUse,
```
csyonghe, line 1256: Use functor templates.

source/slang/slang-ir-autodiff.cpp
```diff
@@ -362,6 +382,47 @@ IRInst* AutoDiffSharedContext::findDifferentiableInterface()
     return nullptr;
 }
 
+IRStructType* AutoDiffSharedContext::findNullDifferentialStructType()
```
csyonghe, line 385: This is the point where I think we should just make NullDifferentialType an IR opcode, and use `__intrinsic_type($(kIROp_NullDifferentialType))` on the stdlib  declaration.

csyonghe: We can find anything via namehint decoration, it can be removed when obfuscation is on.

source/slang/slang-ir-autodiff.cpp
```diff
@@ -1761,6 +2014,71 @@ void removeDetachInsts(IRModule* module)
...
+
+void lowerNullCheckInsts(IRModule* module, AutoDiffSharedContext* context)
+{
+    LowerNullCheckPass pass(module, context);
+    pass.processModule();
+}
+
+void releaseNullDifferentialType(AutoDiffSharedContext* context)
+{
+    if (auto nullStruct = context->nullDifferentialStructType)
```
csyonghe, line 2064: If NullDifferentialType is just an opcode, then we don't need to do this.

source/slang/slang-ir-insts.h
```diff
@@ -1018,6 +1019,17 @@ struct IRBackwardDifferentiate : IRInst
     IR_LEAF_ISA(BackwardDifferentiate)
 };
 
+struct IRIsDifferentialNull : IRInst
```
csyonghe, line 1022: You might be able to use `IRIsType` inst here, but I see this is also a simpler use case.

csyonghe (approved):

PR #3116, opened 2023-08-16
Create storage types of different layouts for SPIRV emit.

csyonghe:
This change uses the buffer-element-lowering pass to create different storage types for struct/arrays that are used in a structured buffer or constant buffer, and wrap them with pack/unpack operations. This allows us to get rid of the dependency on front-end type layouts, which is no longer up-to-date when we are at SPIRV emit stage.

`OpTypeStruct`, `OpTypeArray` and `OpTypeRuntimeArray` cannot be deduplicated based on operands, because we will need different types that are decorated with different offsets even if their field types are the same.

Also refactored IR level type layout calculation to reuse code for different types of layouts.


expipiplus1:
Do we need to do some deeper deduplication for OpTypeStruct etc...?

csyonghe:
> Do we need to do some deeper deduplication for OpTypeStruct etc...?

I don't think we should do any deduplication on struct types. Different struct types have different semantic meanings even if their type operands are the same. If we do want that behavior, it shall be done as an IR pass before we hit spirv leglaization. For spirv emit, we cannot run any deduplication because two struct types with the same operands might be used in different context that require different layouts.

PR #3134, opened 2023-08-17
Add failing nested associated-type test

saipraveenb25:


csyonghe:
@saipraveenb25 We need to convert this into an issue and close this PR if we don't plan to get to it now.

saipraveenb25:
Will raise an issue for this. Fixing this problem needs some refactoring of the existential-types code. It is likely to take a while.

PR #3136, opened 2023-08-17
Be more careful about merge blocks during loop inversion

expipiplus1:


source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -372,6 +373,123 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+
+        // Our IR allows multiple back-edges to a loop header if this is also
+        // the loop continue block. SPIR-V does not so replace them with a
+        // single intermediate block
+        if(c == t)
+        {
+            // Subtract one predecessor for the loop entry
+            const auto numBackEdges = c->getPredecessors().getCount() - 1;
+
+            // If we have multiple back-edges, make a new block at the end of
```
csyonghe, line 424: If we do this, could the jump to the newBlock violate structured rules (e.g. if the original continue is in some nested if statement, then the new jump could be treated like a normal jump but crosses the boundary of its parent merge points).
Does SPIRV allow that?

csyonghe (dismissed):

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: It seems that the safest way to avoid violating those restrictions is to just call eliminateContinue and then elimianteMultiLevelBreak.

expipiplus1: > jump to the newBlock violate structured rules

I think this transformation is safe, at the very least it's no less safe than it was before.

csyonghe (approved):

PR #3139, opened 2023-08-18
Allow loop counters to be used as constexpr arguments.

### LINKED ISSUE #2937: DXC handles compile time constant scenario that Slang does not
Take the following code

```
Texture2D g_texture;
SamplerState g_sampler;

float4 sample(float2 uv, float ofsX, float ofsY ) 
{ 
    return g_texture.SampleLevel(g_sampler, uv, 0.0, int2(ofsX, ofsY)); 
}

[numthreads(4, 1, 1)]
{
        float4 v = sample(float2(tid, tid + 1), 1, -1);
	outputBuffer[tid] = v.x;
}
```

This will compile on DXC, even though `SampleLevel` requires the offsets (ofsX, ofsY) to be compile time constant. This doesn't compile on slang as is, reporting that ofsX and ofsY have the requirement. If we change those params to be `constexpr` then compilation does work on DXC. If compilation is to SPIR-V though the GLSL output outputs a similar error about requiring the offsets to be compile time constants.

This issue can be somewhat worked around by making the function `[ForceInline]`.

csyonghe:
Fixes #2937.


source/slang/slang-ir-constexpr.cpp
```diff
@@ -160,6 +161,143 @@ void markConstExpr(
...
+            if (canInstBeConstExpr)
+            {
+                tentativeConstExprInsts.add(inst);
+                continue;
+            }
+        }
+        // We found an inst that is not constexpr even if all insts in
+        // `tentativeConstExprInsts` are treated as constexpr.
+        // This means that the inst as well as other insts in tentativeConstExpr
+        // should be be treated as constexpr.
```
jsmall-nvidia, line 201: not?

source/slang/slang-ir-constexpr.cpp
```diff
@@ -160,6 +161,143 @@ void markConstExpr(
...
+        return false;
+    }
+    return true;
+}
+
+bool isLoopPhi(IRParam* param, IRLoop*& loopInst)
```
jsmall-nvidia, line 207: Could make static, and return IRLoop*. It it's non null then it's true.

jsmall-nvidia (dismissed):

PR #3148, opened 2023-08-24
Allow requiring glsl language extensions on structs

expipiplus1:


expipiplus1:
This isn't actually useful

PR #3149, opened 2023-08-24
Fix various issues with trivial loops

saipraveenb25:
- Fix detection of trivial loops with multi-level breaks
- Fix detection of unreachable blocks for dom-tree queries
- Fix handling of unreachable blocks during multi-level break elimination
- Fix minor issue with loop unrolling that creates unreachable blocks with branches into it.


source/slang/slang-ir-dominators.cpp
```diff
@@ -427,6 +439,7 @@ struct DominatorTreeComputationContext
     {
         // First we compute the postorder traversal order for the blocks in the CFG.
         computePostorder(code, postorder);
+        computeReachableSet(code, reachableSet);
```
csyonghe, line 442: Can this be fused with `computePostorder`, so that we only need to traverse the graph once?

source/slang/slang-ir-eliminate-multilevel-break.cpp
```diff
@@ -8,6 +8,11 @@
 
 namespace Slang
 {
+    
+bool isUnreachable(IRBlock* block)
```
csyonghe, line 12: maybe `isUnreachableRootBlock`?

source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -102,19 +102,38 @@ static bool isTrivialSingleIterationLoop(
...
+            if (terminator == loop)
+                break;
+            
+            // Check if the break originated from an inner breakable region.
+            // If so, the outer loop cannot be trivially removed.
+            // 
+            switch (terminator->getOp())
+            {
+            case kIROp_loop:
+                if (collectBlocksInRegion(context.domTree, as<IRLoop>(terminator)).contains(breakOriginBlock))
```
csyonghe, line 127: I am worried that the complexity here is unnecessarily high.

source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -123,7 +142,13 @@ static bool isTrivialSingleIterationLoop(
 
 static bool doesLoopHasSideEffect(IRGlobalValueWithCode* func, IRLoop* loopInst)
 {
-    auto blocks = collectBlocksInLoop(func, loopInst);
+    List<IRBlock*> multiLevelBreakBlocks;
+    auto blocks = collectBlocksInRegion(func, loopInst, multiLevelBreakBlocks);
```
csyonghe, line 146: Are there any other places that require the actual break blocks? If not then we should just return a number instead of the whole list.

csyonghe (approved):

csyonghe (approved):

PR #3151, opened 2023-08-25
Initial version of spirv_asm block

expipiplus1:


source/slang/slang-emit-spirv.cpp
```diff
@@ -3789,6 +3798,116 @@ struct SPIRVEmitContext
...
+
+        // This keeps track of the named IDs used in the asm block
+        Dictionary<UnownedStringSlice, SpvWord> idMap;
+
+        for(const auto spvInst : inst->getInsts())
+        {
+            const bool isLast = spvInst == inst->getLastChild();
+            const auto opcodeString = spvInst->getOpcodeString();
+            SpvOp opcode;
+            const bool foundOpCode = lookupSpvOp(opcodeString, opcode)
```
csyonghe, line 3813: Should we do this lookup in the parser and store an SpvWord on our ast?

csyonghe (approved):
This is very nice!

source/slang/slang-emit-spirv.cpp
expipiplus1: Probably, but then we'd have to look them up the other way for ast and ir dumping. Overall a good thing to do though I think

PR #3161, opened 2023-08-29
Correct parsing spirv with no rhs operands

expipiplus1:


source/slang/slang-parser.cpp
```diff
@@ -6411,8 +6411,26 @@ namespace Slang
         // At this point we can also parse bitwise or expressions
         //
         while(!(parser->LookAheadToken(TokenType::RBrace)
-            || parser->LookAheadToken(TokenType::Semicolon)))
+            || parser->LookAheadToken(TokenType::Semicolon))
```
csyonghe, line 6414: Whenever writing a loop like this, we need to consider the end of file case to avoid looping infinitely.

csyonghe: Also consider following the pattern in `ParseBlockStatement` to setup skipping/error recovering.

csyonghe (dismissed):

source/slang/slang-parser.cpp
expipiplus1: Ah, I thought this loop was ok as if we don't read a token we return from the function anyway

PR #3163, opened 2023-08-29
SPIR-V image operations

expipiplus1:


source/slang/slang-stdlib-textures.cpp
```diff
@@ -0,0 +1,1213 @@
...
+            sb << spirv << "\n";
+        }
+    }
+}
+
+void TextureTypeInfo::writeFuncDecorations(
+    const String& glsl,
+    const String& cuda)
+{
+    if(glsl.getLength())
```
csyonghe, line 99: We no longer need to write target_intrinsics, right?

source/slang/slang-stdlib-textures.cpp
```diff
@@ -0,0 +1,1213 @@
...
+    spirvBuilder << " None";
+    if(bias)
+        spirvBuilder << "|Bias";
+    if(lodLevel)
+        spirvBuilder << "|Lod";
+    if(gradX)
+        spirvBuilder << "|Grad";
+    if(constOffset)
+        spirvBuilder << "|ConstOffset";
+    if(minLod)
```
csyonghe, line 801: Also need `Sample` before `minlod`.

source/slang/slang-stdlib-textures.cpp
expipiplus1: Yeah, I had it this way to minimize a massive diff from the beginning while working on this, just to verify that the refactoring was correct. I also wonder if it's slightly faster to process the decorations version, but haven't confirmed.

source/slang/slang-stdlib-textures.cpp
expipiplus1: Yeah, I think this bit will get cleaned up, it was just implemented lazily so far.

PR #3166, opened 2023-08-30
Fix memory barrier intrinsics.

### LINKED ISSUE #3152: Improve barrier lowering to GLSL
https://github.com/shader-slang/slang/blob/ef4c9f1f1c297f1a33be95795a7a7561e0cc3bde/source/slang/hlsl.meta.slang#L2504

https://github.com/shader-slang/slang/blob/ef4c9f1f1c297f1a33be95795a7a7561e0cc3bde/source/slang/hlsl.meta.slang#L1036

I believe these can be consolidated into one statement each with [GL_KHR_memory_scope_semantics](https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_memory_scope_semantics.txt).

So ` __target_intrinsic(glsl, "memoryBarrier(), groupMemoryBarrier(), memoryBarrierImage(), memoryBarrierBuffer()") ` would become ` __target_intrinsic(glsl, "memoryBarrier()") `, which is equivalent to the following, written more explicitly:

```
__target_intrinsic(glsls, "memoryBarrier(gl_ScopeQueueFamily,
                                         gl_StorageSemanticsBuffer | gl_StorageSemanticsShared | gl_StorageSemanticsImage,
                                         gl_SemanticsAcquireRelease")
```

I'm compiling slang -> glsl -> spirv, and I end up with a combo of `OpMemoryBarrier` and `OpControlBarrier` when using `GroupMemoryBarrierWithGroupSync()`, which produces two stalls in nsight - MEMBAR & THDBAR. Seems suboptimal.

csyonghe:
Makes them produce the same spirv code as dxc.

Fixes #3152.


source/slang/hlsl.meta.slang
```diff
@@ -1076,14 +1076,34 @@ bool all(matrix<T,N,M> x)
...
 // Thread-group sync and barrier for writes to all memory spaces (HLSL SM 5.0)
-__target_intrinsic(glsl, "memoryBarrier(), groupMemoryBarrier(), memoryBarrierImage(), memoryBarrierBuffer(), barrier()")
-__target_intrinsic(cuda, "__syncthreads()")
-void AllMemoryBarrierWithGroupSync();
+void AllMemoryBarrierWithGroupSync()
+{
+    __target_switch
+    {
+    case hlsl: __intrinsic_asm "AllMemoryBarrierWithGroupSync";
+    case glsl: __intrinsic_asm "barrier()";
```
farnoy, line 1099: I don't think this is right. `barrier()` corresponds to `GroupMemoryBarrierWithGroupSync()`, but not to `AllMemoryBarrierWithGroupSync()`

https://github.com/KhronosGroup/GLSL/blob/ad383324612d8493c78d52d0c903969fe1dddd0a/extensions/khr/GL_KHR_memory_scope_semantics.txt#L461-L465
https://github.com/KhronosGroup/GLSL/blob/ad383324612d8493c78d52d0c903969fe1dddd0a/extensions/khr/GL_KHR_vulkan_glsl.txt#L678-L681

I don't think GLSL has a single built in function to cover this case like HLSL does. I'd recommend just using GL_KHR_memory_scope_semantics explicitly. The list of semantics would be very close to the SPIR-V version below as well, except the storage semantics don't map exactly from what I can tell.

csyonghe: Fixed and verified that the slang->glslang->spirv output is the same with dxc's output.

PR #3178, opened 2023-09-01
Loop inversion: Handle case where loop can have additional inner breaks

saipraveenb25:
- We now have another critical-edge-breaking block `e4` that is the target of inner breaks.
- Both `e4` and `e1` (the break branch from the loop condition) branch to the loop's new break block `b2`.
- `b2` is a clone of the old break block `b`, and it simply branches to the old break block.


csyonghe (approved):

csyonghe:
Can you modify the test with -validate-ir flag turned on so we can regress test the failure fixed with this PR?

expipiplus1 (approved):
Great work,

Would it be possible to update the comment here too, https://github.com/shader-slang/slang/pull/3178/files#diff-6cbaaba3789e42a289f45f4c8aea78a0b37757956d9cf011817e9690fd4994b7R111-R133

PR #3181, opened 2023-09-04
Add support for getting external semaphore handles

skallweitNV:
Add support for exporting semaphore handles on Vulkan.


tools/gfx/vulkan/vk-fence.cpp
```diff
@@ -40,6 +44,30 @@ Result FenceImpl::init(const IFence::Desc& desc)
...
 
+#if SLANG_WINDOWS_FAMILY
+    VkExportSemaphoreWin32HandleInfoKHR exportSemaphoreWin32HandleInfoKHR;
+#endif
+    VkExportSemaphoreCreateInfoKHR exportSemaphoreCreateInfo;
+    if (desc.isShared)
+    {
+#if SLANG_WINDOWS_FAMILY
+        exportSemaphoreWin32HandleInfoKHR.sType = VK_STRUCTURE_TYPE_EXPORT_SEMAPHORE_WIN32_HANDLE_INFO_KHR;
+        exportSemaphoreWin32HandleInfoKHR.pNext = NULL;
```
expipiplus1, line 55: I would be tempted to make this `pNext` equal to `timelineCreateInfo.pNext` (currently also `nullptr`). This will make it still work should we add any other next structs above. (same for the one below)

expipiplus1 (dismissed):
Would it be possible to `s/NULL/nullptr` also, just for consistency.

tools/gfx/vulkan/vk-fence.cpp
westlicht: Done

skallweitNV:
@expipiplus1 should be good to merge now.

csyonghe (approved):

PR #3190, opened 2023-09-07
Add Mesh and Task shader support to GFX

expipiplus1:
Additionally add a functional test for these

TODO
- [x] Correctly lower DispatchMesh to glsl/spirv


csyonghe:
Can we avoid using d3d12x.h? It is not part of core d3d12 library and we shouldn't depend on it.

slang-gfx.h
```diff
@@ -1339,6 +1339,7 @@ struct GraphicsPipelineStateDesc
     DepthStencilDesc    depthStencil;
     RasterizerDesc      rasterizer;
     BlendDesc           blend;
+    bool                meshShader;
```
csyonghe, line 1342: Why is this needed? We can tell from `program` by checking whether or not it includes a mesh shader stage. It's best to avoid this change because it increases the chance of bugs due to forgetting to set this bit and issues regarding binary compatibility.

premake5.lua
```diff
@@ -1051,7 +1051,7 @@ toolSharedLibrary "slang-reflection-test"
 toolSharedLibrary "render-test"
     uuid "61F7EB00-7281-4BF3-9470-7C2EA92620C3"
 
-    includedirs { ".", "external", "source", "tools/gfx", "tools/platform" }
+    includedirs { ".", "external", "external/vulkan/include", "source", "tools/gfx", "tools/platform" }
```
csyonghe, line 1054: Why is this needed? I am not seeing any additional changes of #include lines in this PR.

expipiplus1: The updated vulkan headers require this now (because they `#include <vulkan/vk_video.h>`)

slang-gfx.h
expipiplus1: It wasn't immediately clear how to extract this information from `ShaderProgramBase`, I've done it now in [`1c00cab` (#3190)](https://github.com/shader-slang/slang/pull/3190/commits/1c00cab1fea3472c788e19b3e98c8bd3dc118861)

csyonghe (approved):

PR #3196, opened 2023-09-08
Don't inline callees with custom derivative before autodiff.

csyonghe:


saipraveenb25 (commented):
This looks good!

We should probably do some decorator renaming at some point: `kIROp_UserDefinedBackwardDerivativeDecoration` & `kIROp_ForwardDerivativeDecoration` do not seem to be similar at first glance..

saipraveenb25 (approved):

csyonghe:
It's not just renaming, we reuse kIROp_ForwardDerivative for both user defined and compiler generated derivatives...

PR #3200, opened 2023-09-11
Correctly identify the number of operands to image sampling operands in SPIR-V

expipiplus1:


csyonghe (approved):

expipiplus1:
👀 One of these failing in only one config on CI days...

csyonghe:
I think it is probably just swift-shader doesn't support mesh shaders. You need to add a -render-feature flag in the test to disable it when mesh shader isn't reported as supported.

csyonghe:
> I think it is probably just swift-shader doesn't support mesh shaders. You need to add a -render-feature flag in the test to disable it when mesh shader isn't reported as supported.

Oh wait, this isn't even the mesh shader PR...

csyonghe:
Due to recent github image and tool chain updates, I can no longer build a working swift shader binary on linux. It would be good to figure out how to re-enable it on linux.

expipiplus1:
I can reproduce the failure on Linux with swiftshader 2020-11-06 and 2023-09-11.

texture-simpler does work, however and isn't doing anything meaningfully different.

PR #3203, opened 2023-09-12
Fix build with --enable-xlib=false

expipiplus1:
Since Jan 2022 Swiftshader no longer supports xlib (it suggests using xcb instead)


csyonghe:
I believe our examples still use xlib to display a window on Linux.

expipiplus1:
They should still work, but prior to this slang just didn't compile with enable-xlib=false

csyonghe (approved):

PR #3211, opened 2023-09-19
Fix loop inversion issue that caused ordinary blocks with multiple predecessors

saipraveenb25:
The original version can leave the CFG in an invalid state with `e4` not being a break block or merge point, but having multiple predecessors.

The updated version creates a separate jump block for each break instruction to avoid this issue.


source/slang/slang-ir-loop-inversion.cpp
```diff
@@ -118,19 +118,22 @@ static IRParam* duplicateToParamWithDecorations(IRBuilder& builder, IRCloneEnv&
...
 // c1, c2 are the Condition blocks
 // e1, e2, e3 are the critical Edge breakers
-// l is the Loop entering block
-// d is the loop boDy
-// b is the Break block
+// l is the loop entering block
+// d is the loop body
+// b is the merge point for the outer condition
+// b2 is the new break block for the loop
+//
```
expipiplus1, line 136: Is it possible to update the comment on the function to reflect the changes with e4, please?

csyonghe (approved):

PR #3215, opened 2023-09-19
Memoize types in spirv asm blocks

expipiplus1:
Still TODO: emit a diagnostic when a use of a memoized value precedes its definition.

### DIFF
--- a/source/slang/slang-emit-spirv.cpp
+++ b/source/slang/slang-emit-spirv.cpp
@@ -978,11 +978,12 @@ struct SPIRVEmitContext
     template<typename... Operands>
     SpvInst* emitInst(SpvInstParent* parent, IRInst* irInst, SpvOp opcode, const Operands& ...ops)
     {
-        InstConstructScope scopeInst(this, opcode, irInst);
-        SpvInst* spvInst = scopeInst;
-        (emitOperand(ops), ...);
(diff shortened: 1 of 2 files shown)


source/slang/slang-emit-spirv.cpp
```diff
@@ -4187,21 +4208,63 @@ struct SPIRVEmitContext
...
+                {
+                    last = emitInstMemoizedCustomOperandFunc(
+                        opParent,
+                        assignedInst,
+                        opcode,
+                        kResultID,
+                        [&](){
+                            Index i = 0;
+                            for(const auto operand : spvInst->getSPIRVOperands()) {
+                                if(i++ != 0)
```
csyonghe, line 4235: for readability, let's not use ++ and -- as an expression.

source/slang/slang-emit-spirv.cpp
```diff
@@ -4187,21 +4208,63 @@ struct SPIRVEmitContext
                 default:
                     break;
                 }
+                const auto opParent = parentForOpCode(opcode, parent);
```
csyonghe, line 4211: Not related to this IR, but the lambdas in this function is getting long enough that we should move them into standalone functions to make the code more readable.

csyonghe (approved):

PR #3222, opened 2023-09-21
Update user-guide with new slangpy features

saipraveenb25:


docs/user-guide/a1-02-slangpy.md
```diff
@@ -14,191 +14,190 @@ In addition, using a per-thread programming model also results in more optimized
...
-float square(float x)
-{
-    return x * x;
-}
-```
-This is done by simply adding a `[Differentiable]` attribute to our `square`function.
+One key point is that the basic `TensorView<T>` objects are not differentiable. They can be used as buffers for data that does not require derivatives, or even as buffers for the manual accumulation of derivatives.
 
-With that, we can now define `square_bwd_kernel` that performs backward propagation as:
+Instead, use the `DiffTensorView` type for when you need differentiable tensors. Currently, `DiffTensorView` only supports the `float` dtype variety, and requires the use of `.load(offset)` and `.store(offset, val)` instead of `[]`, although
```
csyonghe, line 93: Should we just add support for subscript operator?

docs/user-guide/a1-02-slangpy.md
```diff
@@ -446,6 +445,169 @@ back propagated derivative values.
...
+long as the kernel code is not changed, future runs will not rebuild the CUDA kernel.
+
+Because the PyTorch JIT system requires `ninja`, you need to make sure `ninja` is installed on your system
+and is discoverable from the current environment, you also need to have a C++ compiler available on the system.
+On Windows, this means that Visual Studio need to be installed.
+
+### Manual binding for kernel derivatives
+
+The above example demonstrates how to write a simple kernel function in Slang and call it from Python.
+Another major benefit of using Slang is that the Slang compiler support generating backward derivative
```
csyonghe, line 544: Should revise these sections and remove the redundant text such as the intro to slang's auto diff features.

csyonghe (approved):

PR #3229, opened 2023-09-21
Revert inlining change in #3217.

csyonghe:
#3217 causes a performance regression.

This change reverts the change to do pervasive inlining, instead it checks if the callee has fwd_diff and bwd_diff insts and don't inline if they have those insts during `performPreAutoDiffInlining` to avoid the problem that the autodiff pass cannot handle a function  that has unprocessed fwd_diff or bwd_diff insts.


saipraveenb25 (approved):
Looks good!

PR #3240, opened 2023-09-26
Fix for epoch/ASTBuilder* nullptr issue

jsmall-nvidia:
Previously these tests were failing...

tests/serialization/serialized-module-test.slang
tests/serialization/extern/extern-test.slang

The issue is around the thread_local mechanism holding the current ASTBuilder. This wasn't being setup for a couple of scenarios that were being hit. The fix is to setup the thread_local for these scenarios using `SetASTBuilderContextRAII `

The other problem was around embed-stdlib. There the problem was that the sharedASTBuilder was being destroyed before the ASTBuilder on the Session. This led to a dangling pointer being accessed by the dtor of ASTBuilder. The fix was just to destroy the ASTBuilder before anything else.

### DIFF
--- a/source/slang/slang-serialize-ast.cpp
+++ b/source/slang/slang-serialize-ast.cpp
@@ -86,6 +86,8 @@ struct ASTFieldAccess
 
         ASTBuilder builder(sharedASTBuilder, "Serialize Check");
 
+        SetASTBuilderContextRAII astBuilderRAII(&builder);
+
         DefaultSerialObjectFactory objectFactory(&builder);
 
         // We could now check that the loaded data matches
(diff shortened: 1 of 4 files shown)


source/slang/slang-serialize-ast.cpp
```diff
@@ -86,6 +86,8 @@ struct ASTFieldAccess
 
         ASTBuilder builder(sharedASTBuilder, "Serialize Check");
 
+        SetASTBuilderContextRAII astBuilderRAII(&builder);
```
csyonghe, line 89: There is a macro for this:
```
SLANG_AST_BUILDER_RAII(&builder);
```

csyonghe (approved):

PR #3249, opened 2023-09-28
Support `constref` parameters passing.

csyonghe:
This PR adds support for `__constref` parameter modifier to specify that a parameter should be passed by const reference.
A `constref` parameter is treated as an immutable value.
Adds `[constref]` attribute that can be used to mark a function to make `this` parameter passed by const reference.

Make NonCopyable type passed by constref by default.
Also makes sure constref parameters doesn't lead to redundant load/stores in auto-diffed functions.

Add tests for mixing use of `__constref` and `no_diff` in a differentiable function.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -679,6 +679,12 @@ struct Ref
 {};
 
 __generic<T>
+__magic_type(ConstRefType)
+__intrinsic_type($(kIROp_ConstRefType))
+struct ConstRef
+{};
+
+__generic<T>
 __magic_type(OptionalType)
 __intrinsic_type($(kIROp_OptionalType))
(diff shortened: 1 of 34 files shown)


saipraveenb25 (dismissed):
Looks good! 

Just to clarify, `constref` isn't going to mix with differentiable types?

csyonghe:
I haven't tested, but it is likely not going to work to have a differentiable `constref` parameter. We need to handle that carefully if we decide to support it.

csyonghe:
One way to support it is to convert it to a normal `in` parameter.

csyonghe:
OK, I added checks to prevent forming a constref differentiable parameter, including `this` and normal parameters.

PR #3256, opened 2023-10-02
Add test for failing makeStruct derivative when diff-member associations are missing

saipraveenb25:


csyonghe:
@saipraveenb25 We need to convert this into an issue and close this PR if we don't plan to get to it now.

saipraveenb25:
Will raise an issue for this. Should be simple enough to fix quickly.

PR #3258, opened 2023-10-04
SPIRV compiler performance fixes.

csyonghe:
Fixes complexity explosion in `simplifyIR` where we repetitively traverse the entire cfg to determine reachability/dominance.

Fixes a bug in constant folding of `bitcast`.

Implement a pass to turn composite SSA values into memory objects.

Fix a bug in `IRBlock::addParam`.

Make eliminatePhi pass only process phi nodes of composite types and leave scalar phi nodes as is. 

Expose more options for `simplifyIR` to avoid performing expensive optimizations where unnecessary.

Revise stdlib and get rid of `__spirv_capability` modifiers.


source/slang/slang-emit-spirv.cpp
```diff
@@ -4830,6 +4814,7 @@ SlangResult emitSPIRVFromIR(
     // feature so we can skip
     if(SLANG_FAILED(validationResult) && validationResult != SLANG_E_NOT_AVAILABLE)
     {
+        File::writeAllBytes("d:\\bug.spv", spirvOut.getBuffer(), spirvOut.getCount());
```
expipiplus1, line 4817: ^

source/slang/slang-emit.cpp
```diff
@@ -1263,15 +1278,18 @@ SlangResult emitSPIRVForEntryPointsDirectly(
     List<uint8_t> spirv, outSpirv;
     emitSPIRVFromIR(codeGenContext, irModule, irEntryPoints, spirv);
 
+#if 0
     String optErr;
     if (SLANG_FAILED(optimizeSPIRV(spirv, optErr, outSpirv)))
     {
+        File::writeAllBytes("D:\\problem-spirv.spv", spirv.getBuffer(), spirv.getCount());
```
expipiplus1, line 1285: ^

expipiplus1 (dismissed):

PR #3265, opened 2023-10-06
Small type system fixes.

csyonghe:
- Fix emitElementAddr() on attributed/rate types.
- Add diagnostic on using a read-only property as l-value.
- Properly recognize differentiablity of a custom provided derivative function referenced via `fwd_diff(f)` or `bwd_diff(f)`.
- Perform `${workspaceFolder}` macro replacement in clang-format-location settings in language server.


saipraveenb25 (approved):
Looks good!

PR #3274, opened 2023-10-12
Make SlangResult its own type

expipiplus1:
TODO:
- [ ] move the existing definitions to be enumerators
- [ ] add [[nodiscard]]

### DIFF
--- a/source/compiler-core/slang-dxc-compiler.cpp
+++ b/source/compiler-core/slang-dxc-compiler.cpp
@@ -114,9 +114,9 @@ public:
         if (intf)
         {
             *out = intf;
-            return SLANG_OK;
+            return S_OK;
         }
-        return SLANG_E_NO_INTERFACE;
+        return E_NOINTERFACE;
     }
     SLANG_NO_THROW ULONG SLANG_MCALL AddRef() SLANG_OVERRIDE { return 1; }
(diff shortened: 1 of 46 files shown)


source/compiler-core/slang-dxc-compiler.cpp
```diff
@@ -567,7 +567,7 @@ SlangResult DXCDownstreamCompiler::compile(const CompileOptions& inOptions, IArt
     DxcIncludeHandler includeHandler(&searchDirectories, options.fileSystemExt, options.sourceManager);
 
     ComPtr<IDxcOperationResult> dxcOperationResult;
-    SLANG_RETURN_ON_FAIL(dxcCompiler->Compile(dxcSourceBlob,
+    SLANG_RETURN_ON_FAIL(SlangResult(dxcCompiler->Compile(dxcSourceBlob,
```
csyonghe, line 570: Instead of changing every use to cast to SlangResult, can we change SLANG_RETURN_ON_FAIL so it handles both SlangResult and other integer return code types?

expipiplus1: I've added SLANG_RETURN_ON_FAIL_HRESULT, to retain some type safety.

(This is actually obviated by the other change making SlangResult not an enum class)

csyonghe:
I am going to close this PR since this is too old.

Issue #3282, opened 2023-10-23
`frexp` in CUDA accepts `int *` as the second argument


chenzhekl:
Unlike `frexp` in HLSL where both parameters are of the same type. `frexp` in CUDA accepts only `int *` as the second argument.

```
__device__​ float frexpf(float x, int* nptr)
__device__​ double frexp(double x, int* nptr)
```

This is the incorrect CUDA code slang generated:

```
float exp_0;
float _S1 = frexp(12.00000000000000000000, &exp_0);
```

for

```
float exp;
frexp(12.0f, exp);
```

csyonghe:
@chenzhekl Slang v2023.4.3 is now available, which includes the fix.

chenzhekl:
@csyonghe Thank you so much for you work! I can confirm that it works flawlessly.

csyonghe:
Thanks for confirming!

PR #3284, opened 2023-10-24
Make the exponent return value from frexp int

expipiplus1:
Fixes https://github.com/shader-slang/slang/issues/3282

### DIFF
--- a/deps/target-deps.json
+++ b/deps/target-deps.json
@@ -4,14 +4,14 @@
         "dependencies" : [
             {
                 "name" : "slang-llvm",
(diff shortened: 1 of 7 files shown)


expipiplus1:
Need to figure out why the CPU tests are returning the exponent without shifting it down.

csyonghe (dismissed):

csyonghe:
The CPU test failure is due to slang-llvm not in-sync. We have to update the `frexp` implementation in slang-llvm as well.

csyonghe (approved):

PR #3304, opened 2023-11-02
Add OpDot for spirv

expipiplus1:


tangent-vector (approved):

expipiplus1:
Todo for me, why didn't this get unrolled

csyonghe:
We only unroll a loop if it has `[ForceUnroll]` attribute. We don't try to make any decisions to automatically unroll a loop.

PR #3307, opened 2023-11-02
Several spirv intrinsics and small fix

expipiplus1:
Will add some tests tomorrow

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2919,11 +2919,15 @@ matrix<T, N, M> frexp(matrix<T, N, M> x, out matrix<int, N, M, L> exp)
 // Texture filter width
 __generic<T : __BuiltinFloatingPointType>
 [__readNone]
+__target_intrinsic(hlsl)
+__target_intrinsic(glsl)
+__target_intrinsic(spirv, "OpFWidth resultType resultId _0")
 T fwidth(T x);
 
 __generic<T : __BuiltinFloatingPointType, let N : int>
 __target_intrinsic(hlsl)
 __target_intrinsic(glsl)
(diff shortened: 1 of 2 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -4352,22 +4360,48 @@ __generic<let N : int> float noise(vector<float, N> x)
...
+    {
+    case hlsl:
+        __intrinsic_asm "NonUniformResourceIndex";
+    case glsl:
+        __intrinsic_asm "nonuniformEXT";
+    case spirv:
+        spirv_asm
+        {
+            OpCapability ShaderNonUniform;
+            OpDecorate $index NonUniform;
```
csyonghe, line 4377: Should this be `[ForceInline]`d to make sure the index we returned is the same SSA value as the one we just decorated?

source/slang/hlsl.meta.slang
```diff
@@ -5062,6 +5097,7 @@ matrix<T, M, N> transpose(matrix<T, N, M> x)
 __generic<T : __BuiltinIntegerType, let N : int, let M : int>
 __target_intrinsic(hlsl)
 __target_intrinsic(glsl)
+__target_intrinsic(spirv, "OpTranspose resultType resultId _0")
```
csyonghe, line 5100: It seems OpTranspose is not defined integer matrices in spirv, same for other matrix op codes.

csyonghe (approved):

source/slang/hlsl.meta.slang
expipiplus1: It generates the correct code already, but you're right, forceinline would be more correct.

PR #3314, opened 2023-11-07
CUDA: Fixes for NVRTC 12.x and warp mask ambiguity; adds CC 8.x warp reduction intrinsics.

NeilBickford-NV:
Hi Slang team! This merge request contains a few attempts to fix things I noticed while looking at the generated CUDA code for a `WaveActiveSum()` operation.

* Fixes mask ambiguity between the mask used for getting the lane index from `threadIdx.x` and the fast-path mask for a full mask of threads.

`SLANG_CUDA_WARP_MASK` (31) was used in two places: as the mask used in `_getLaneId()`:

```cuda
 __forceinline__ __device__ uint32_t _getLaneId()
{
    // ...
    return threadIdx.x & SLANG_CUDA_WARP_MASK;
}
```

and as the test for the fast-path in `_waveCalcPow2Offset()`:

```cuda
__inline__ __device__ int _waveCalcPow2Offset(WarpMask mask)
{
    // This should be the most common case, so fast path it
    if (mask == SLANG_CUDA_WARP_MASK)
    {
        return SLANG_CUDA_WARP_SIZE; // nbickford note: This is 32
    }
    ...
}
```

This latter case was probably incorrect: because `SLANG_CUDA_WARP_MASK` is `0x1f` instead of `0xffffffff`, `_waveCalcPow2Offset()` would return `32` when the last 5 threads were active, and the most common case (where all threads were active) would go through the slow-path logic.

I added `SLANG_CUDA_WARP_BITMASK` to try to resolve this ambiguity between two different sizes of mask.

* Adds full specializations for `_wave*()` when compute capability 8.x warp reduction intrinsics.

The fast path above uses 5 shuffles and syncs; on CC 8.x hardware, we can access the `__reduce_*_sync()` intrinsics, which can compile to a single [redux.sync](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-redux-sync) PTX instruction.

(This could probably use some testing, since it turns out NVRTC was compiling the tests for CC 5.0; more on that in the next issue! I can verify that compiling a .slang file to a .cu file and then compiling that with `nvcc -arch=sm_80` generates the intended intrinsics, but I haven't found a way to set the required `compute` version in a `.slang` file yet or to have the tests run on sm_80 other than by adding an option to the source code.)

* Fixes running NVRTC on CUDA 12

By default, Slang invokes NVRTC with an architecture of `compute_35` when the CUDA SDK version is 11.0+. However, CUDA 12 removed support for `compute_35`, so NVRTC produces an error unless the architecture version is `compute_50` or higher.

`slang-test -api cuda -category wave` seems to work now, but the full set of `slang-test -api cuda` tests still produces some errors (which appear to be unrelated to wave intrinsics, so far as I can tell).

Thank you!

### DIFF
--- a/prelude/slang-cuda-prelude.h
+++ b/prelude/slang-cuda-prelude.h
@@ -53,7 +53,8 @@
 #   define SLANG_CUDA_WARP_SIZE 32
 #endif
 
-#define SLANG_CUDA_WARP_MASK (SLANG_CUDA_WARP_SIZE - 1)
+#define SLANG_CUDA_WARP_MASK (SLANG_CUDA_WARP_SIZE - 1) // Used for masking threadIdx.x to the warp lane index
+#define SLANG_CUDA_WARP_BITMASK (~int(0))
 
 //
 #define SLANG_FORCE_INLINE inline
@@ -1418,10 +1419,16 @@ __inline__ __device__ bool _waveIsSingleLane(WarpMask mask)
 }
 
 // Returns the power of 2 size of run of set bits. Returns 0 if not a suitable run.
+// Examples:
+// 0b00000000'00000000'00000000'11111111 -> 8
+// 0b11111111'11111111'11111111'11111111 -> 32
+// 0b00000000'00000000'00000000'00011111 -> 0 (since 5 is not a power of 2)
+// 0b00000000'00000000'00000000'11110000 -> 0 (since the run of bits does not start at the LSB)
+// 0b00000000'00000000'00000000'00100111 -> 0 (since it is not a single contiguous run)
 __inline__ __device__ int _waveCalcPow2Offset(WarpMask mask)
 {
     // This should be the most common case, so fast path it
-    if (mask == SLANG_CUDA_WARP_MASK)
+    if (mask == SLANG_CUDA_WARP_BITMASK)
     {
         return SLANG_CUDA_WARP_SIZE;
     }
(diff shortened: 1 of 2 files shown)


prelude/slang-cuda-prelude.h
```diff
@@ -1418,10 +1419,16 @@ __inline__ __device__ bool _waveIsSingleLane(WarpMask mask)
...
+// 0b00000000'00000000'00000000'11111111 -> 8
+// 0b11111111'11111111'11111111'11111111 -> 32
+// 0b00000000'00000000'00000000'00011111 -> 0 (since 5 is not a power of 2)
+// 0b00000000'00000000'00000000'11110000 -> 0 (since the run of bits does not start at the LSB)
+// 0b00000000'00000000'00000000'00100111 -> 0 (since it is not a single contiguous run)
 __inline__ __device__ int _waveCalcPow2Offset(WarpMask mask)
 {
     // This should be the most common case, so fast path it
-    if (mask == SLANG_CUDA_WARP_MASK)
+    if (mask == SLANG_CUDA_WARP_BITMASK)
```
jsmall-nvidia, line 1431: Could you explain a little more about this change? Why would the mask be ~int(0)? 

Would it be better to do `if (mask &  SLANG_CUDA_WARP_MASK ) == SLANG_CUDA_WARP_MASK`, which would cover scenarios where user code set the mask to more bits than lanes?

jsmall-nvidia (commented):
Thanks for adding these improvements! 

I would like to understand the reasoning a little more around the masking before approving. 

I think there are probably more efficient ways to doing some of the reductions, that didn't get around to being implemented previously.

prelude/slang-cuda-prelude.h
NeilBickford-NV: Absolutely! This code isn't totally portable at the moment, because it relies on `typedef int WarpMask` being exactly `#define SLANG_CUDA_WARP_SIZE 32` bits long. (Because each bit in `WarpMask` corresponds to an active thread, we want every bit in WarpMask to be set; this is why `~int(0)` should work for the 32-bit, 32-thread case.)

If we can assume `SLANG_CUDA_WARP_SIZE` is always 32, then maybe I could change the type of `WarpMask` to `int32_t` (which could protect against the case where `int` is a length other than 32 bits on some platforms)?

In the most general case, the tricky thing is that trying to set only the least significant `SLANG_CUDA_WARP_SIZE` bits using

```cpp
#define SLANG_CUDA_WARP_BITMASK (1 << SLANG_CUDA_WARP_SIZE) - 1
```

invokes undefined behavior, because it shifts left by the number of bits in the type: from the very end of "built-in bitwise shift operators" in https://en.cppreference.com/w/cpp/language/operator_arithmetic,

> If the value of rhs is negative or is not less than the number of bits in lhs, the behavior is undefined.

If this winds up in host code, it turns out `clang -O2` recognizes this UB and totally optimizes it out, which leads to surprising behavior (this Godbolt sample shows it for unsigned ints, but the same UB happens for signed types): https://godbolt.org/z/s4TG5Wzdj

This would probably work, but it's pretty long and it could be nice if there's a nice way to do it at the preprocessor level -- what do you think about this?

```cpp
#define SLANG_CUDA_WARP_BITMASK ((SLANG_CUDA_WARP_SIZE >= sizeof(WarpMask)) ? ~WarpMask(0) : (WarpMask(1) << SLANG_CUDA_WARP_SIZE) - 1)
```

NeilBickford-NV: Oh, I should probably describe more how `_waveCalcPow2Offset(mask)` gets called with a mask of active threads, rather than the number of active threads. For a scalar type, `WaveActiveSum<T>(x)` compiles to `_waveSum(_getActiveMask(), x)`, and then the return value of  `_getActiveMask()` gets passed to `_waveCalcPow2Offset(mask)` via the following call stack:

```cpp
_waveSum<T>(_getActiveMask(), x):
  _waveReduceScalar<WaveOpAdd<T>, T>(mask, val):
    const int offsetSize = _waveCalcPow2Offset(mask)
```

Thanks!

jsmall-nvidia: Ok I see, that makes a lot of sense. I think your change as it seems reasonable and agree

```
#define SLANG_CUDA_WARP_BITMASK ((SLANG_CUDA_WARP_SIZE >= sizeof(WarpMask)) ? ~WarpMask(0) : (WarpMask(1) << SLANG_CUDA_WARP_SIZE) - 1)
```

is perhaps overkill.

jsmall-nvidia (approved):

Issue #3283, opened 2023-10-24
CMake support


fknfilewalker:
I am surprised that there is no CMake support, It would be really nice to have CMake scripts for slang.

LesleyLai:
Agreed. Lack of CMake support (and thus hard to build it on multiple platforms) is the main reason I haven't tried this project yet.

csyonghe:
@fknfilewalker @LesleyLai We do provide binary releases for both Windows and Linux. Which platforms are you on? 
We can certainly consider providing CMake build files if that is blocking adoption.

LesleyLai:
Windows and Linux cover my needs, though I generally find building from sources much easier to deal with compared to prebuilt binary.

I see that there is also a [vcpkg port of slang](https://vcpkg.roundtrip.dev/ports/shader-slang) though I am not sure how well maintained it is

natevm:
+1.

CMake is the industry standard when it comes to C++ development. It has its faults for sure, but still, it’s ubiquitous, especially in computer graphics.

If slang were to have some official CMake support, even if it were just a guideline on how to incorporate slang into an existing CMake project, that would probably help adoption quite a bit.

natduca:
Ellie is on this so setting assignee, and since its in progres putting it onto Q1 but also putting the priority to low to reflect that this might get preempted if something more urgent for this quarter comes up.

gdianaty:
+1

Regarding @LesleyLai's comment about vcpkg, the current port downloads binaries from GitHub and unzips them into the installed packages directory. Obviously this prevents project-specific dependency wrangling (e.g. slang-gfx library is unconditionally imported). If dependency wrangling is handled via CMake, much of the current Premake complexity in Slang can be reduced as well.

PR #3326, opened 2023-11-14
WIP: CMake

expipiplus1:
Todo:
- [x] Test on Windows/OS X
- [x] slang-llvm binary fetching
- [x] complete option handling
- [x] document usage (also add presets as done by Falcor)
- [ ] test integrating slang into a parent cmake project
- [x] cross build (works at the moment if you can run the host code on the build machine, but this should be a staged build)

The main difference between this and the premake build is that all generated files go into the build directory, instead of being generated alongside the source. Also, build products are split appropriately between `lib` and `bin`.

It shouldn't step on the toes of the premake build, but I'm not sure about the other way around.

This also integrates slang-glslang and slang-llvm fully into this repo, just build them as normal targets. One can either `SLANG_USE_SYSTEM_LLVM` to use whatever llvm the OS has installed for us, or have CMake build an appropriate static LLVM for us (suitable for creating a redistributable slang-llvm). I'm planning for the downloaded binary version to be taken from the most recent release in this repo (according to `git describe --abbrev=0 --tags --match v*`).

Closes https://github.com/shader-slang/slang/issues/3283


csyonghe (dismissed):
Looks good to me and feel free to merge when it is not breaking the CI.

fknfilewalker:
Can we move all the test, example and gfx targets behind some optional flag in cmake, so it is possible to include slang in another project without having to deal with all these targets? 

Additional maybe it would be nice to detect if a cmake project already included glslang (or other deps) and use these includes instead of slang complaining about double includes (not sure if this could lead to version mismatch problems)

PR #3375, opened 2023-12-02
`cbrt()` family

Cons-Cat:
I've implemented a free (CC BY 4.0) cbrt algorithm for floats (Moroz L, Samotyy V, Walczyk CJ, Cieśliński JL.). Cube roots are required for many color space conversions, so in my opinion, this is an important feature for a standard library. I personally needed this for working with Oklab colors.

In the future, double and half precision or integer cube roots might be good additions, although I expect this to cover the most common cases.

I'm happy to make any changes to the code.


source/slang/hlsl.meta.slang
```diff
@@ -6573,6 +6573,94 @@ matrix<T, N, M> sqrt(matrix<T, N, M> x)
...
+// Moroz L, Samotyy V, Walczyk CJ, Cieśliński JL.
+// Fast Calculation of Cube and Inverse Cube Roots Using a Magic Constant and Its Implementation on Microcontrollers.
+// Energies. 2021; 14(4):1058.
+// https://doi.org/10.3390/en14041058
+//
+// Published under the CC BY 4.0 license.
+//
+// They use no lookup tables, which makes them vectorizable in theory.
+
+[__readNone]
```
csyonghe, line 6587: Is this faster than calling `pow(x, 0.3333333333f)`? It is a lot of instructions.

Cons-Cat: The linked paper claims it is 20x faster than that using an unspecified C standard library, but I can benchmark this under Vulkan, which may have different results.

source/slang/hlsl.meta.slang
```diff
@@ -6573,6 +6573,108 @@ matrix<T, N, M> sqrt(matrix<T, N, M> x)
...
+// Energies. 2021; 14(4):1058.
+// https://doi.org/10.3390/en14041058
+//
+// Published under the CC BY 4.0 license.
+//
+// They use no lookup tables, which makes them vectorizable in theory.
+
+__target_intrinsic(cuda, "$P_cbrt($0)")
+[__readNone]
+float cbrt(float x)
```
csyonghe, line 6589: We also need to provide a derivative of this in diff.meta.slang to make it differentiable, like other math functions.

source/slang/hlsl.meta.slang
Cons-Cat: `pow()` appears to run about 2 to 3 times faster in a Vulkan compute shader under a loop of `1_000_000ll` iterations. The only suspicious thing to me is that `pow<float>(buff[0], 0.3333333333f);` compiles in GLSL as `pow(buff_0._data[0U], 0.3333333432674407959);` and I'm not sure where that number comes from. Maybe the compiler is reading the text `0.3333333333f` and parsing it to a float with slight error, then writing that back into the output shader?

source/slang/hlsl.meta.slang
csyonghe: That will work, but the compiler generated derivatives won't be as efficient as hand coded ones.
My suggestion is to implement a custom derivative with `[ForwardDerivativeOf(cbrt)]` and `[BackwardDerivativeOf(cbrt)]`. You can refer to other existing functions.

csyonghe: I see that the definition is just `pow(x, 0.33333)` now, in that case the perf difference should be small. It is still nicer to have custom derivatives defined, but the PR is good to merge.

csyonghe (approved):

source/slang/hlsl.meta.slang
```diff
@@ -6573,6 +6573,58 @@ matrix<T, N, M> sqrt(matrix<T, N, M> x)
...
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(cuda, "$P_rcbrt($0)")
+[__readNone, Differentiable]
+T rcbrt(T x)
+{
+    // 0 is inaccurate unless special-cased.
+    if (x == T(0))
+        return T(0);
+
+    return T(1.0) / pow(x, T(1) / T(3));
```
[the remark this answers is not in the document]
Cons-Cat: Oops! That was a silly mistake, thanks.

source/slang/hlsl.meta.slang
```diff
@@ -6597,6 +6597,67 @@ matrix<T, N, M> sqrt(matrix<T, N, M> x)
...
+[__readNone]
+T cbrt(T x)
+{
+    return pow(x, T(1) / T(3));
+}
+
+// Derivative of cube root
+__generic<T : __BuiltinFloatingPointType>
+[__readNone, Differentiable, ForwardDerivativeOf(cbrt)]
+DifferentialPair<T> cbrtFwdDiff(DifferentialPair<T> x)
```
csyonghe, line 6612: This should be implemented in diff.meta.slang, similar to how `sqrt` is implemented.

source/slang/diff.meta.slang
```diff
@@ -1364,6 +1364,12 @@ SIMPLE_UNARY_DERIVATIVE_IMPL(sqrt, T(0.5) / sqrt(max(ReturnType(T(1e-7)), dpx.p)
 SIMPLE_UNARY_DERIVATIVE_IMPL(rcp, T(-1.0) / max(ReturnType(T(1e-7)), dpx.p * dpx.p))
 // rsqrt
 SIMPLE_UNARY_DERIVATIVE_IMPL(rsqrt, T(-0.5) / (dpx.p * sqrt(dpx.p)))
+// Cube root => (0.333...) / (x ** -2/3)
```
csyonghe, line 1367: Isn't it `0.333 * x**(-2/3)`?

csyonghe:
Closing since this appears to be no longer worked on. Feel free to reopen another PR if needed.

PR #3391, opened 2023-12-08
`quaternion<T>` library

Cons-Cat:
I have a basic `quaternion<T>` struct and helpers here. It has abbreviated spellings `hquat`, `fquat`, and `dquat`, which come from GLM. The names `.imag()` and `.real()` come from `std::complex<T>` in C++. It does not support autodiff yet, mainly because I haven't looked much into autodiff on types just yet. In the future, dual-quaternions might be a good addition as well.

There is still work to be done that I would appreciate feedback on:

* I have no test cases yet. Should new test files be written for quaternions, similar to the separate test files for scalars, vectors, and matrices?
* There are only a few functions provided so far. Are there any more that would be considered particularly important? I think I should add rotations along a specified axis/axes.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -4046,6 +4046,156 @@ T dot(vector<T, N> x, vector<T, N> y)
     return result;
 }
 
+struct quaternion<T : __BuiltinFloatingPointType>
+{
+    __init(quaternion<T> q)
+    {
+        data = q.data;
+    }
+
+    __init(vector<T, 4> q)
+    {
+        data.xyzw = q;
+    }
+
+    __init(vector<T, 3> imaginary, T real)
+    {
+        data.xyz = imaginary;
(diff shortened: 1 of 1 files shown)


csyonghe:
I also think that things like this should now be wrapped in an `experimental` namespace so we don't need to worry about breaking compatibility if there are any user code that is using them.

chloekek:
This implementation assumes a representation where the `xyz` components store the imaginary parts and the `w` component stores the real part, but there are applications (e.g. OGRE) which store the real part in the `x` component and the imaginary parts in the `yzw` components. If quaternions become part of the language or are otherwise shipped with the compiler the API should be generic enough to support both representations, lest the user has to convert when working with data from/into buffers.

csyonghe:
Closing since we need a plan for where libraries like this should go first. They may not belong to the core compiler repo and can exist as a standalone library repo.

PR #3400, opened 2023-12-11
GLSL SSBO Support

expipiplus1:
One small todo is to move the lowering later in the pipeline, at the moment this hits a validation bug for the non-array case.

Documentation too


expipiplus1:
Hmm, I think that this failing CI build is spurious as it passes locally and in other CI builds...

expipiplus1:
mm, looks as though the modifier checks will have to be a little more sophisticated

Issue #3338, opened 2023-11-16
`NonUniformResourceIndex` not working in direct-to-spirv backend.


csyonghe:
The logic to attempt to emit the `OpDecoration` is executed but the resulting SPIRV still does not have the decoration inst. To repro, use the existing test `vk-texture-indexing.slang` with the following setting:

```
// vk-texture-indexing.slang

//TEST:SIMPLE(filecheck=CHECK):-target spirv -entry main -stage fragment
//TEST:SIMPLE(filecheck=CHECK):-target spirv -entry main -stage fragment -emit-spirv-directly
// CHECK: OpDecorate %{{.*}} NonUniform
```

csyonghe:
The current fully user-space implementation also risks losing the decoration with implicit type casts.
For example, if the user have something like:
```
uint8_t index = NonUniformResourceIndex(...);
array[index] = ...
```
Then we could be generating the following code:
```
a = NonUniformResourceIndex(i)
b = IntCast<uint> a
resPtr = GetElementPtr arr b
```
Where the value that is used to index the array isn't properly marked as `NonUniformResourceIndex`.
To address this issue, the safest solution for now is to introduce a dedicated inst for `NonUniformResourceIndex`, and have a pass to promote `NonUniformResourceIndex` to the top of the operator[] arguments, so we can transform the above code into:
```
b' = IntCast<uint> i
c = NonUniformResourceIndex(b')
resPtr = GetElementPtr arr c
```

pmistryNV:
Looking into the bug.

pmistryNV:
Bug fixed with https://github.com/shader-slang/slang/pull/3417 closing the bug

PR #3417, opened 2023-12-16
Fix nonuniform decoration on direct-to-spirv backend path. (#3338)

pmistryNV:
By updating the hlsl meta intrinsic for spirv path. Decorating the index that is input to NonUniformResourceIndex was getting removed by spirv-opt, when inlining as the decoration was applied to the function parameter. Thus the implementation was modified by introducing OpCopyObject that copies the object inside the function body of the meta builtin "NonUniformResourceIndex" and the decoration is applied to the result of OpCopyObject. With this approach SPIRV-OPt Inliner propagates decoration correctly.


csyonghe (approved):
Looks good!

Issue #3359, opened 2023-11-28
Issue with `-emit-spirv-directly` and ray-trace shader


mklefrancois:
When using the `-emit-spirv-directly` option with the attached shader, there will be an issue when creating the shader module.

The error that is returned is: 
```
vkCreateShaderModule(): pCreateInfo->pCode is not valid SPIR-V: End of input reached while decoding OpLoad starting at word 5381: expected more operands after 3 words.
```

But when this options isn't there, there is a new error showing up, only for the raygeneration and closesthit, but not for the miss shader:

```
vkCreateShaderModule(): pCreateInfo->pCode is not valid SPIR-V: Id is 0.
```
Interestingly, the shader work despite the error.

And if we are using the in combination the option `-fvk-use-entrypoint-name` instead of  `-entry ...`, the error is similar, but differs in the OpLoad position :

``` 
vkCreateShaderModule(): pCreateInfo->pCode is not valid SPIR-V: End of input reached while decoding OpLoad starting at word 3277: expected more operands after 3 words. 
```` 

Command line 
```
slangc.exe -emit-spirv-directly -entry rgenMain -target spirv -g3 -line-directive-mode glsl -profile glsl_460 -D__slang -force-glsl-scalar-layout -source-embed-style text -source-embed-name raytrace_rgenMain -o raytrace_rgenMain.spirv raytrace.slang
```
[slang-raytrace.zip](https://github.com/shader-slang/slang/files/13489807/slang-raytrace.zip)

**Comment by csyonghe** [comment]
The first step here is to try to reproduce the issue with the attached shader and get a sense of what exactly is going wrong.

**Comment by mklefrancois** [comment]
I have updated [nvpro-core](https://github.com/nvpro-samples/nvpro_core) and [vk_mini_samples](https://github.com/nvpro-samples/vk_mini_samples) which have what it needs to reproduce the issue.

**Steps**
* Clone both nvpro-core and vk_mini_samples. 
* With CMake GUI, check the option USE_SLANG and hit Generate
* Under the ray_trace project, open `ray_trace.cpp` and set USE_SEPARATE_SLANG to 0
```
#define USE_SEPARATE_SLANG 0
```

The Separate version compiles each entrypoint separately and is not using the `-emit-spirv-directly` option and this is working. When the define is set to 0, it is using `"_autogen/raytrace_slang.h"`, which is the result of the compilation using `-emit-spirv-directly` and `-fvk-use-entrypoint-name`.

[...]

PR #3428, opened 2023-12-30
Fix the intrinsic expansion of ObjectToWorld3x4 in spirv_asm. Data type

pmistryNV:
of result of OpLoad instruction was missing. This resulted in an incomplete OpLoad instruction getting generated causing spirv-val to report failure.
Also added dumping of spirv disassembly right after unoptimized spirv emission.

Fixes bug #3359

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -91,6 +91,7 @@
 #include "slang-emit-cuda.h"
 #include "slang-emit-torch.h"
 
+#include "slang-spirv-val.h"
 #include "../compiler-core/slang-artifact-desc-util.h"
 #include "../compiler-core/slang-artifact-util.h"
 #include "../compiler-core/slang-artifact-impl.h"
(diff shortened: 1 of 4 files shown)


pmistryNV:
@csyonghe please review it

source/slang/slang-emit.cpp
```diff
@@ -1292,6 +1293,13 @@ SlangResult emitSPIRVForEntryPointsDirectly(
     auto artifact = ArtifactUtil::createArtifactForCompileTarget(asExternal(codeGenContext->getTargetFormat()));
     artifact->addRepresentationUnknown(ListBlob::moveCreate(spirv));
 
+#if 0
```
csyonghe, line 1296: Leftover debug code?

csyonghe (approved):

source/slang/slang-emit.cpp
pmistryNV: I just wanted to keep that code disabled to print disassembled code right after code generation before spirv-opt is run.

PR #3446, opened 2024-01-12
GLSL Passthrough support for SSBO types

expipiplus1:


csyonghe (commented):
Can we turn this into an executable test?

expipiplus1:
Mm, that'll mean adding type layout support for this buffer type so that the test runner can interface with them; definitely something to do eventually though.

Would it be acceptable to just expose them the same as byte address buffers for the time being do you think?

csyonghe:
Is there a reason why the `buffer` syntax isn't parsed the same way as `cbuffer`?
The current implementation does not seem to support the case where the name of the buffer object is omitted, such as:
```
layout(std430, binding = 1) buffer ssbo
{
    vec4 data[];
};
```

I think we should be able to use the exact same logic of `ParseHLSLBufferDecl` to parse glsl `buffer` syntax.

expipiplus1:
I think this is good to go otherwise

expipiplus1:
@csyonghe good to merge?

@jkwak-work the ssbo tests here show how to do functional tests for glsl input now also.

tools/render-test/options.cpp
```diff
@@ -223,6 +223,14 @@ static gfx::DeviceType _toRenderType(Slang::RenderApiType apiType)
             //
             // TODO: At some point we could warn/error and deprecate this option.
         }
+        else if (argValue == "-allow-glsl")
```
csyonghe, line 226: In my latest change merged to top of tree, we may be able to use -lang glsl instead of this flag.

csyonghe (approved):

csyonghe:
Looks good, thank you!

Will you clean up the ssbo parsing logic to unify it with cbuffer logic next?

PR #3459, opened 2024-01-18
Docs for local vs global uniform parameters

natevm:
These changes to the docs add some clarity to folks using uniform parameters and mixed shader stages including both RT and non-RT types. 

Might want to go over the text to double check correctness


docs/user-guide/02-conventional-features.md
```diff
@@ -581,3 +581,73 @@ Entry-point uniform parameters are semantically similar to global-scope shader p
...
+> #### Note ####
+> GLSL does not support multiple entry-points; however, SPIR-V does. Vulkan users wanting to take advantage of Slang mixed entry points must pass `-fvk-use-entrypoint-name` and `-emit-spirv-directly` as compiler arguments.
+
+### Mixed Entry-Point Uniform Parameters
+
+Like with the previous `vertexMain` example, mixed entry point setups also support _entry-point uniform parameters_.
+
+However, because of certain systematic differences between entry point types, a uniform being _global_ or _local_ will have very important consequences on the underlying layout and behavior.
+
+For most all entry point types, D3D12 and Vulkan use one common root signature to define both global and local uniform parameters. For these cases, Slang maps uniforms to this common root signature.
```
csyonghe, line 623: Vulkan doesn't have a "root signature". The closest thing there is "pipeline layout".

natevm: Yes, pipeline layout would work.

docs/user-guide/02-conventional-features.md
```diff
@@ -581,3 +581,73 @@ Entry-point uniform parameters are semantically similar to global-scope shader p
...
+
+### Mixed Entry-Point Uniform Parameters
+
+Like with the previous `vertexMain` example, mixed entry point setups also support _entry-point uniform parameters_.
+
+However, because of certain systematic differences between entry point types, a uniform being _global_ or _local_ will have very important consequences on the underlying layout and behavior.
+
+For most all entry point types, D3D12 and Vulkan use one common root signature to define both global and local uniform parameters. For these cases, Slang maps uniforms to this common root signature. 
+
+However, for ray tracing entry points, these parameters map to either _global_ root signatures or to _local_ root signatures, with the latter being stored in the shader binding table.
```
csyonghe, line 625: A parameter may not always be visible in a root signature. An ordinary data field will typically be allocated in the global default constant buffer, which does not correspond directly to any entries in the root signature.

This should be reworded to something like:

> The scope where a uniform parameter is defined has a consequence on the parameter binding in the target code. For ray tracing shaders, uniform parameters defined in the global scope will determine the shape of global root signature / pipeline layout, while uniform parameters defined as entry point function parameters will be allocated in Shader Binding Tables. On D3D12, uniform entry-point parameters will affect the local root signature of the entry-point.

docs/user-guide/02-conventional-features.md
```diff
@@ -581,3 +581,73 @@ Entry-point uniform parameters are semantically similar to global-scope shader p
...
+
+[shader("closesthit")]
+void closestHitMain(uniform float3 localUniform4) 
+{ /* ... */ }
+```
+
+In this example, `globalUniform` is appended to the global root signature for _both_ compute _and_ ray generation stages for all four entry points. 
+Compute entry points lack "local root signatures", so `localUniform1` is "pushed" to the stack of reserved global uniform parameters for use in `computeMain1`. 
+Leaving that entry point scope "pops" that global uniform parameter such that `localUniform2` can reuse the same binding location for `computeMain2`.
+However, local uniforms for ray tracing shaders map to the corresponding local root signatures found in the shader binding table, and so no "push" or "pop" to the global root signature occurs for these parameters.
```
csyonghe, line 653: global root signature --> global binding locations

docs/user-guide/02-conventional-features.md
csyonghe: This paragraph can be moved to the "EntryPoint Shader Parameters" section above.

docs/user-guide/02-conventional-features.md
```diff
@@ -620,10 +620,13 @@ Like with the previous `vertexMain` example, mixed entry point setups also suppo
 
 However, because of certain systematic differences between entry point types, a uniform being _global_ or _local_ will have very important consequences on the underlying layout and behavior.
 
-For most all entry point types, D3D12 and Vulkan use one common root signature to define both global and local uniform parameters. For these cases, Slang maps uniforms to this common root signature. 
+For most all entry point types, D3D12 will use one common root signature to define both global and local uniform parameters. 
+Likewise, Vulkan descriptors will bind to a common pipeline layout. For both of these cases, Slang maps uniforms to the common root signature / pipeline layout.
```
csyonghe, line 624: I would like us to avoid overusing the term "bind" in the documentation because it is very ambiguous.

Here, I would say "Likewise in Vulkan, the binding locations of uniform parameters are typically defined by a common pipeline layout. For both targets, Slang allocates a binding location for uniform parameters that will constitute the common root signature or pipeline layout."

docs/user-guide/02-conventional-features.md
```diff
@@ -620,10 +620,13 @@ Like with the previous `vertexMain` example, mixed entry point setups also suppo
...
 
-However, for ray tracing entry points, these parameters map to either _global_ root signatures or to _local_ root signatures, with the latter being stored in the shader binding table.
-When entry points match a "ray tracing" type, we bind uniforms which are in the _global_ scope to the _global_ root signature, while uniforms which are _local_ are bound to the _local_ root signature.
+However, for ray tracing entry points and D3D12, these parameters map to either _global_ root signatures or to _local_ root signatures, with the latter being stored in the shader binding table.
+In Vulkan, D3D12's global root signatures translate to a shared ray tracing pipeline layout, while local root signatures map again to shader binding table records. 
+
+When entry points match a "ray tracing" type, we bind uniforms which are in the _global_ scope to the _global_ root signature (or ray tracing pipeline layout), while uniforms which are _local_ are bound to shader binding table records, which depend on the underlying runtime record indexing.
```
csyonghe, line 629: When compiling a ray tracing entry-point, Slang allocates uniforms defined in the global scope from the global root signature (or ray tracing pipeline layout), and allocates uniform entry-point parameters from local root signature or shader binding table.

docs/user-guide/02-conventional-features.md
natevm: > global default constant buffer, which does not correspond directly to any entries in the root signature

I did not know this, but that makes sense. I'm less familiar with D3D12, but in Vulkan, the same would be true for push constants. The above revision seems good to me.

docs/user-guide/02-conventional-features.md
```diff
@@ -581,3 +581,73 @@ Entry-point uniform parameters are semantically similar to global-scope shader p
...
+Like with the previous `vertexMain` example, mixed entry point setups also support _entry-point uniform parameters_.
+
+However, because of certain systematic differences between entry point types, a uniform being _global_ or _local_ will have very important consequences on the underlying layout and behavior.
+
+For most all entry point types, D3D12 and Vulkan use one common root signature to define both global and local uniform parameters. For these cases, Slang maps uniforms to this common root signature. 
+
+However, for ray tracing entry points, these parameters map to either _global_ root signatures or to _local_ root signatures, with the latter being stored in the shader binding table.
+When entry points match a "ray tracing" type, we bind uniforms which are in the _global_ scope to the _global_ root signature, while uniforms which are _local_ are bound to the _local_ root signature.
+
+Consider the following:
```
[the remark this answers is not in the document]
natevm: I'd prefer we do not exclude Vulkan, but can include "compiling" sure

docs/user-guide/02-conventional-features.md
natevm: >  local root signature or shader binding table.

There is no distinction between the local root signature and the shader binding table. D3D12 uses both interchangeably, while Vulkan just uses "shader binding table".

PR #3463, opened 2024-01-19
Add support for rayQueryGetIntersectionTriangleVertexPositionsEXT

kaizhangNV:
Add a member function to RayQuery<RAY_FLAG> class
GetIntersectionTriangleVertexPositions to support the GLSL built-int rayQueryGetIntersectionTriangleVertexPositionsEXT.

Also add new test file under tests/vkray/rayquery-closesthit.slang to test this functionality.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -10208,6 +10208,42 @@ ${{{{
         auto ccTF = candidateOrCommitted == 0 ? "false" : "true";
 }}}}
 
+    __glsl_extension(GL_EXT_ray_query)
+    __glsl_extension(GL_EXT_ray_tracing_position_fetch)
+    __glsl_version(460)
+    [__NoSideEffect]
+    void __glslGetIntersectionTriangleVertexPositions$(ccName)(out float3 arr[3])
+    {
+        __intrinsic_asm "rayQueryGetIntersectionTriangleVertexPositionsEXT($0, $(ccTF), $1)";
+    };
+
+    __glsl_extension(GL_EXT_ray_query)
+    __glsl_extension(GL_EXT_ray_tracing_position_fetch)
+    __glsl_version(460)
+    [__NoSideEffect]
+    float3[3] $(ccName)GetIntersectionTriangleVertexPositions()
+    {
+        typedef float3[3] float3Arr3;
+        __target_switch
+        {
+        case glsl:
+            float3 output[3];
+            __glslGetIntersectionTriangleVertexPositions$(ccName)(output);
+            return output;
+        case spirv:
+            uint iCandidateOrCommitted = $(candidateOrCommitted);
+            return spirv_asm
+            {
+                OpCapability RayQueryKHR;
+                OpCapability RayTracingPositionFetchKHR;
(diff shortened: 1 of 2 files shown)


csyonghe (commented):
Looks good, but it seems it is missing a OpCapability declaration in the resulting SPIRV.

source/slang/hlsl.meta.slang
```diff
@@ -10125,6 +10125,27 @@ struct RayQuery <let rayFlagsGeneric : RAY_FLAG = RAY_FLAG_NONE>
         }
     }
 
+    __glsl_extension(GL_EXT_ray_query)
+    __glsl_extension(GL_EXT_ray_tracing_position_fetch)
+    __glsl_version(460)
+    [__NoSideEffect]
+    void GetIntersectionTriangleVertexPositions(out float3 positions[3])
```
csyonghe, line 10132: I suggest we just expose as `float3 GetIntersectionTriangleVertexPosition(int index)` to keep consistent with the existing HitTriangleVertexPosition function.

kaizhangNV: There is a little problem by using this signature.
The glsl signature is 
```
void rayQueryGetIntersectionTriangleVertexPositionsEXT(rayQueryEXT q,
                                                               bool committed,
                                                               out vec3 positions[3]);
```

so if we define the signature as `float3 GetIntersectionTriangleVertexPosition(int index)`, we will call this function 3 times to get all vertices.

kaizhangNV:
Yea, I think we can do that.
Let me try to update another change.

kaizhangNV:
There is an issue about whether we're enable to add multiple `__intrinsic_asm` under one glsl target, like:
```
case glsl:
__intrinsic_asm "xxx";
__intrinsic_asm "yyy";
```

I guess, so the other option will be:
case glsl:
__intrinsic_asm "xxx;\nyyy";

But this form will bring us a weird indent in the generated glsl target.

kaizhangNV:
Right, so there is a little problem here.
You mentioned to change the function signature as:
```float3[3] GetIntersectionTriangleVertexPositions()```
please correct me if I misunderstood.

In that case, I have to insert few lines of code for glsl target:

```
float3[3] out;
rayQueryGetIntersectionTriangleVertexPositionsEXT(out);
return out;
```

The only way I know to do this is `__intrinsic_asm"ilne1;\nline2\nline3"`.
But in this case, the indents of line2 and line3 will not be aligned with line1.

**Comment by csyonghe** [comment]
I think that's fine. If you want it to look nice you can define:
```
void __glslGetIntersectionTriangleVertexPositions(out float3 arr[3])
{
      __intrinsic_asm("rayQueryGetIntersectionTriangleVertexPositionsEXT");
}

[ForceInline]
float3 GetIntersectionTriangleVertexPositions()
{
     __target_switch
    {
    case glsl:
          float3 output;
          __glslGetIntersectionTriangleVertexPositions(output);
         return output;
    }
}
```

kaizhangNV:
@csyonghe can you review and approve?
The latest change: https://github.com/shader-slang/slang/commit/ff9da2f291742a67b80dbb5300ec843a9d3a6986.

csyonghe (approved):

PR #3470, opened 2024-01-22
Fix incorrect behavior of operator%

### LINKED ISSUE #1059: Operator % semantics, including floating-point
The `%` operator in C-family languages has a long and complicated history. The big decision point is whether it should map to a "modulus" operation or a "remainder" operation, with the difference having to do with what happens on negative operands.

Both C99 and C++11 appear to specify that `%` should be a remainder such that `(a/b)*b + (a%b) == a` (where that expression doesn't overflow/underflow or divide-by-zero), *and* that the `a/b` operation always rounds toward zero (rather than allowing round-toward-negative-infinity).

HLSL appears to follow the modern C/C++ behavior. DXC always maps `%` to the  `*Rem` operations (not the `*Mod` operations), in both DXIL and SPIR-V.

As an extended feature on top of C/C++, HLSL supports `%` on floating-point inputs, where it performs an LLVM/SPIR-V `FRem` operation.

For our HLSL back-end this is all easy enough: we just need to map `%` through to the output and we are done. On the C/C++ back-end we can either rely on C99/C++11 semantics (which realistically are what most compilers/processors always did), or switch to `std::div()` for integers, and use the `remainder()` function from `math.h` for the floating-point case.

The GLSL target is the one where this gets interesting, because GLSL has no standard function that maps to a remainder operation, and thus could satisfy the semantics of HLSL input. For unsigned types this is a non-issue of course. For floating-point types there is a `mod()` function, but no equivalent for remainder.

jkwak-work:
Fixes #1059

This change fixes the incorrect translation of "operator%" from HLSL to SPIRV.

The issue stems from the fact that the behavior of "operator%" in GLSL differs from that in HLSL. In HLSL it behaves as "remainder" where as it behaves as "modulus" in GLSL.

We have been using SpvOpFMod for operator% when Slang compiles from HLSL to SPRIV, which is incorrect. This change switches it to SpvOpFRem.

The tests are slightly modified to reveal any potential issues.


tests/compute/frem.slang
```diff
@@ -2,12 +2,12 @@
 
 //TEST(compute):COMPARE_COMPUTE: -shaderobj
 //TEST(compute):COMPARE_COMPUTE:-cpu -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-vk -shaderobj
+//TEST(compute):COMPARE_COMPUTE:-vk -shaderobj -emit-spirv-directly
```
csyonghe, line 5: You can add `-output-using-type` to this `//TEST` lines so that floating point values will be written as floating points instead of binaries. You can search for it in the repo to see how it is used in other tests.

csyonghe (approved):

csyonghe (approved):

Issue #3432, opened 2024-01-03
slangc when compiled with option "-validate-ir" and "-emit-spirv-directly" causes validation error


pmistryNV:
Running any slang file that emits spirv causes below validation error 

hlsl.meta.slang(3001): internal error 40007: IR validation failed: def must be ancestor of use

Bug can be replicated for example by running below command:
slangc.exe -target spirv -entry main -stage fragment -emit-spirv-directly tests\cross-compile\vk-sampler-getdimension.slang -validate-ir

pmistryNV:
@csyonghe filed as discussed. Replicated it on  spirv path.

pmistryNV:
Kindly refer to the updated changes

csyonghe:
Fixed.

PR #3486, opened 2024-01-24
IRSPIRVAsmOperandInst instructions may not have IRBlock as the immedi…

pmistryNV:
…ate parent. Previously special case was added to handle IRDecoration similarly. Replace this with a common method getBlock that traverses the parent chain till it gets to the Block

Fixes bug #3432

### DIFF
--- a/source/slang/slang-ir-util.cpp
+++ b/source/slang/slang-ir-util.cpp
@@ -1020,21 +1020,14 @@ IRInst* getVulkanPayloadLocation(IRInst* payloadGlobalVar)
     return location;
 }
 
-void moveParams(IRBlock* dest, IRBlock* src)
+IRInst* getInstInBlock(IRInst* inst)
 {
-    for (auto param = src->getFirstChild(); param;)
-    {
-        auto nextInst = param->getNextInst();
-        if (as<IRDecoration>(param) || as<IRParam, IRDynamicCastBehavior::NoUnwrap>(param))
-        {
-            param->insertAtEnd(dest);
-        }
-        else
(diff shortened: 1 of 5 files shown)


pmistryNV:
@csyonghe kindly review

source/slang/slang-ir-util.cpp
```diff
@@ -1338,6 +1351,22 @@ List<IRBlock*> collectBlocksInRegion(IRGlobalValueWithCode* func, IRLoop* loopIn
     return collectBlocksInRegion(dom, loopInst, &hasMultiLevelBreaks);
 }
 
+IRBlock* getBlock(IRInst* inst)
+{
+    if (!inst)
+        return nullptr;
+
+    do {
```
csyonghe, line 1359: We usually put braces in its own line.

csyonghe (approved):
Looks good!

source/slang/slang-ir-util.cpp
pmistryNV: ok i can update it

pmistryNV: do you want me to have it like
do
{
...
}
while(inst);

csyonghe: That is fine, although I think the existing format is
```
do
{
} while (inst);
```
We don't use do-whiles that much in the codebase, so might be easier to just use `while(inst){}` or
```
for(;;)
{
       if (!inst) break;
}
```

csyonghe (approved):

PR #3488, opened 2024-01-24
[SPIRV] Support `globallycoherent` and `[vk::index()]`.

### LINKED ISSUE #3482: Support `globallycoherent` storage class in SPIR-V
It appears that we're not currently supporting `globallycoherent` in our direct-to-spirv backend. 

As a result, atomic operations currently cannot be seen outside a workgroup. 

If we can get this feature added, I can bring these slangified decoupled lookback prefix sum and one sweep algorithms from CUB into Falcor. I'd like to use these for some SW tree construction tests. 

@jeffbolznv is familiar with globallycoherent. When I asked what to expect in the spirv from globallycoherent, he replied:

> There are two ways to represent it. If it's not using the vulkan memory model, then there's a coherent decoration on the variable. If it is using vulkan memory model, then there are operands on the load and store instructions (available/visible)

>[but currently, slang is] using GLSL450 memory model and doesn't have the coherent decoration.

Attached is a slang file containing the "one sweep" radix sorter, which I'd like to get working for Vulkan. [onesweep.zip](https://github.com/shader-slang/slang/files/14034899/onesweep.zip)

csyonghe:
Fixes #3482.

Also implements `[earlydepthstencil]`.

Fixing `StructuredBuffer<bool>`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,7 @@ tools/gfx-unit-test/test-tmp-*
 
 # Intermediate source files generated during build process
 /source/slang/slang-generated-*.h
-/source/slang/slang-lookup-capability-defs.cpp
+/source/slang/slang-lookup-*.cpp
 /source/slang/*.meta.slang.h
 prelude/*.h.cpp
 /source/slang/cpp.hint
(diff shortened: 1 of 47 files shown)


natevm:
Latest changes appear to work on my end now! Thanks Yong!

Issue #3490, opened 2024-01-24
Floats Zeroed out with no warning


half-potato:
The following code will zero out TINY_VAL to 0, resulting in broken code.
```
#define TINY_VAL 1.1754944e-37

[TorchEntryPoint]
[Differentiable]
float safe_div(in float a, in float b) {
  if (abs(b) < TINY_VAL) {
    return a / TINY_VAL;
  } else {
    return a / b;
  }
}
```
```
>>> safe_math = slangpy.loadModule("safe-math.slang")
>>> safe_math.safe_div(5,1)
5.0
>>> safe_math.safe_div(5,0)
inf
>>> safe_math.safe_div(5,1e-37)
4.999999840142846e+37
>>> safe_math.safe_div(5,1.175e-37)
4.255319002099013e+37
```
Here is the generated code:
```
SLANG_PRELUDE_EXPORT
float safe_div(float a_0, float b_0)
{
    if((F32_abs((b_0))) < 0.0f)
    {
        return a_0 / 0.0f;
    }
    else
    {
        return a_0 / b_0;
    }
}
```

jkwak-work:
I think there are two problems related to this issue.
First is on the fact that "precision" setting for the ostringstream is set too high.
It is currently set to 20 and I believe it should be set to 16 like following.
std::cout.precision(std::numeric_limits<double>::max_digits10 - 1); // 17 -1

Second problem is that there is a precision lose when a value is read as a "float" type.
I found this little surprising but the following codes returns different values.
```
std::cout.precision(std::numeric_limits<double>::max_digits10 - 1);

double d = 1.7e-37; // without "f" postfix
std::cout << d << '\n'; // 1.7e-37

double f = 1.7e-37f; // with "f" postfix.
std::cout << f << '\n'; // 1.700000029707435e-37
```

The binary representation of 1.7e-37 stored in double type is "0x384cec8b57852073"
The binary representation of 1.7e-37f stored in double type is "0x384cec8b60000000".

Since our Slang assumes all floating point values in the source file as a single precision float, I think there is a lose of the precision while reading it.

jkwak-work:
I guess a simplest solution is to lower the number of digits down to how much float can store.
`std::cout.precision(std::numeric_limits<float>::max_digits10 - 1);`

swoods-nv:
Moving this from Jay's plate to Kai's, since Jai has a couple of large Q1 things on his plate. Kai, can you take this one over when you are back?

[...]

PR #3518, opened 2024-01-26
Preserve the precision of a small value near zero

jkwak-work:
Fixes an issue #3490

When a very small value is emitted, it was snapping to a zero value. This change preserves the precision more to keep the value from being a zero value.

When "ostringstream" was printing a floating point value, we set "presicion" that indicates the number of digits in decimal. We have been using a fixed value, 20, which limits the available value near zero to be 0.00000000000000000001. When a given value is smaller, the mantissa part was truncated and it became a zero.

This change calculates how much precision is needed based on the given value and the type. The mantissa bits of "float" type can represent upto 8 digits and that of "double" type can represent upto 16 digits in decimal. The exponential part of "float" can represent upto 38 digits and that of "double" can represent upto 308 digits.


csyonghe:
This will work, but emitting 100+digits for a float lit seems silly. We should switch to scientific notation when necessary.

source/slang/slang-emit-source-writer.cpp
```diff
@@ -251,7 +252,13 @@ void SourceWriter::emit(double value)
     std::ostringstream stream;
     stream.imbue(std::locale::classic());
     stream.setf(std::ios::fixed, std::ios::floatfield);
-    stream.precision(20);
+
```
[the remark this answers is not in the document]
jkwak-work: The default behavior of ostringstream is what you described.
Because we set it to be "fixed" at the line above, the output value is always in the fixed format and never in the scientific format.
I assumed that we are using "fixed" format for a reason but I guess we can remove it judging by your comment.

I will make some changes in a way not to use "std::ios::fixed" flag.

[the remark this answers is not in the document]
csyonghe: The default setting won't work either because it doesn't emit enough digits when in scientific notation. We still need at least 20 digits when in scientific notation.

jkwak-work:
I removed the use of ostringstream and implemented a function to emit float value manually.
I also updated the test case to check some edge cases.

source/slang/slang-emit-source-writer.cpp
```diff
@@ -239,35 +239,197 @@ void SourceWriter::emit(double value)
...
-    auto slice = UnownedStringSlice(str.c_str());
-    // Remove redundant trailing 0s.
-    if (slice.end() > slice.begin())
+
+    char result[30];
+    char *next = result;
+    double valueAbs = value;
+
+    // Handle negative sign
+    const long long* d2l = reinterpret_cast<long long*>(&value);
```
csyonghe, line 248: Type punning like this is C++ UB, consider:
```
uint64_t uintVal = 0;
memcpy(&uintVal, &value, sizeof(uint64_t);
```

source/slang/slang-emit-source-writer.cpp
```diff
@@ -224,7 +224,7 @@ void SourceWriter::emit(UInt64 value)
     emit(buffer);
 }
 
-void SourceWriter::emit(double value)
+inline static void SourceWriter_emit(SourceWriter *This, double value, int maxFracDigits)
```
csyonghe, line 227: naming convention: we don't use underscores as decl names. This can be
`static void _emitFloatingPointLiteral()`

source/slang/slang-emit-source-writer.cpp
```diff
@@ -239,35 +239,197 @@ void SourceWriter::emit(double value)
...
-    // Remove redundant trailing 0s.
-    if (slice.end() > slice.begin())
+
+    char result[30];
+    char *next = result;
+    double valueAbs = value;
+
+    // Handle negative sign
+    const long long* d2l = reinterpret_cast<long long*>(&value);
+    const bool negative = (*d2l < 0);
```
csyonghe, line 249: Can this just be `sign(value)`?

source/slang/slang-emit-source-writer.cpp
```diff
@@ -239,35 +239,197 @@ void SourceWriter::emit(double value)
...
+    else if (value == 0)
+    {
+        *next++ = '0';
+        *next++ = '.';
+        *next++ = '0';
+    }
+    else
+    {
+        int exponent = 0;
+        double mantissa = valueAbs;
```
csyonghe, line 278: Can we use `frexp` to extract the mantissa and exponent?

source/slang/slang-emit-source-writer.cpp
```diff
@@ -239,35 +239,197 @@ void SourceWriter::emit(double value)
...
+        const double rounding = std::pow(10, -maxFracDigits) / 2;
+
+        // Normalize the mantissa
+        while (mantissa >= 10.0) {
```
csyonghe, line 282: Instead of implementing all these logic ourselves, can we call `ostream << std::scientific << std::setprecision(precision) << value` to emit a float with specific number of digits in scientific notation?

source/slang/slang-emit-source-writer.cpp
jkwak-work: The tricky part is most of attempt to get the sign-ness will fail if the value is NaN.
I see that there is a C++11 function for this, signbit.
I will use it https://en.cppreference.com/w/cpp/numeric/math/signbit

source/slang/slang-emit-source-writer.cpp
jkwak-work: Currently the implementation uses "fixed" format and if I switch to "scientific" format, it breaks a few existing tests.
And simple values will become harder to read. As an example, "1.0" with precision value 20 will become, "1.00000000000000000000e+00".
To void the problem, I will have to check the exponent value and decide if the value is better with scientific format or fixed format. In other words, half of my current implementation will still be there.

Note that there was a previous comment saying,
"// Eventually, the right move here would be to implement proper floating-point number formatting ourselves, "
And I agree to this statement in three reasons.
1. we will have more explicit control over how floating values are emitted.
2. our coding guide line stated not to use iostream.
3. the performance will be better because each call to the member function of ostringstream will internally take a mutex lock to prevent the system wide locale change in the middle of text parsing. I have seen from my previous project that this mutex locking cost almost 30% of the time for processing the text messages.

csyonghe: How about this:
1. use isnan / isinf to check for nan and infs, and early out those cases.
2. use frexp to extract exponent bits.
3. if exponent is less than a small value, use fixed output. truncate 0s in the end.
4. otherwise, use scientific output, truncate 0s before "e".

I want to avoid implementing our own logic as much as possible, because unless we also roll an extensive test suite on this, we can't guarantee its correctness.

jkwak-work: I will change to your suggestions.

expipiplus1:
It's a shame that not all of our targets support hex float notation...

PR #3525, opened 2024-01-30
Implement basic GLSL built-in functions

### LINKED ISSUE #3362: Bring intrinsic coverage in glsl.meta.slang to parity with hlsl intrinsics.

jkwak-work:
Partially resolves #3362

This change implemented GLSL build-in functions described in the following sections of "OpenGL Spec" document.
  8.1. Angle and Trigonometry Functions
  8.2. Exponential Functions
  8.3. Common Functions
  8.5. Geometric Functions
  8.7. Vector Relational Functions
  8.8. Integer Functions

About 40 functions are newly implemented and about 150 functions were preexisted on HLSL side implementation.

The implementation of new functions hasn't been tested yet.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -501,6 +501,186 @@ public vector<float, N> uintBitsToFloat(highp vector<uint, N> x)
 }
 
 //
+// Section 8.4. Floating-Point Pack and Unpack Functions
+//
+
+[__readNone]
+[ForceInline]
+uint packUnorm1x16(float c)
+{
+    return uint(round(clamp(c, 0.0, 1.0) * 65535.0));
+}
+
+[__readNone]
+[ForceInline]
+uint packSnorm1x16(float v)
+{
+    return uint(round(clamp(v ,-1.0, 1.0) * 32767.0) + 32767.0);
+}
+
+[__readNone]
+[ForceInline]
+uint packUnorm1x8(float c)
+{
+    return uint(round(clamp(c, 0.0, 1.0) * 255.0));
+}
+
+[__readNone]
+[ForceInline]
+uint packSnorm1x8(float c)
+{
+    return uint(round(clamp(c, -1.0, 1.0) * 127.0) + 127.0);
+}
+
+[__readNone]
+[ForceInline]
+float unpackUnorm1x16(uint p)
+{
+    return float(p) / 65535.0;
+}
+
+[__readNone]
+[ForceInline]
+float unpackSnorm1x16(uint p)
+{
+    return clamp((float(p) - 32767.0) / 32767.0, -1.0, 1.0);
+}
+
+[__readNone]
+[ForceInline]
+float unpackUnorm1x8(uint p)
+{
+    return float(p) / 255.0;
+}
+
+[__readNone]
+[ForceInline]
+float unpackSnorm1x8(uint p)
+{
+    return clamp((float(p) - 127.0) / 127.0, -1.0, 1.0);
+}
+
+[__readNone]
+[ForceInline]
+uint float2half(float f)
+{
+    uint u = floatBitsToUint(f);
+    uint s = ((u >> uint(16)) & uint(0x8000));
+    uint e = ((((u & uint(0x7f800000)) - uint(0x38000000)) >> uint(13)) & uint(0x7c00));
+    uint m = ((u >> uint(13)) & uint(0x03ff));
+    return (s | e | m);
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uint packUnorm2x16(vec2 v)
+{
+    return packUnorm1x16(v.x) | (packUnorm1x16(v.y) << uint(16));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uint packSnorm2x16(vec2 v)
+{
+    return packSnorm1x16(v.x) | (packSnorm1x16(v.y) << uint(16));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uint packUnorm4x8(vec4 v)
+{
+    return packUnorm1x8(v.x) | (packUnorm1x8(v.y) << uint(8)) | (packUnorm1x8(v.z) << uint(16)) | (packUnorm1x8(v.w) << uint(24));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uint packSnorm4x8(vec4 v)
+{
+    return packSnorm1x8(v.x) | (packSnorm1x8(v.y) << uint(8)) | (packSnorm1x8(v.z) << uint(16)) | (packSnorm1x8(v.w) << uint(24));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public vec2 unpackUnorm2x16(uint p)
+{
+    return vec2(unpackUnorm1x16(p & uint(0xffff)), unpackUnorm1x16(p >> uint(16)));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public vec2 unpackSnorm2x16(uint p)
+{
+    return vec2(unpackSnorm1x16(p & uint(0xffff)), unpackSnorm1x16(p >> uint(16)));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public vec4 unpackUnorm4x8(highp uint p)
+{
+    return vec4(unpackUnorm1x8(p & uint(0xffff)), unpackUnorm1x8(p >> uint(8)), unpackUnorm1x8(p >> uint(16)), unpackUnorm1x8(p >> uint(24)));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public vec4 unpackSnorm4x8(highp uint p)
+{
+    return vec4(unpackSnorm1x8(p & uint(0xffff)), unpackSnorm1x8(p >> uint(8)), unpackSnorm1x8(p >> uint(16)), unpackSnorm1x8(p >> uint(24)));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uint packHalf2x16(vec2 v)
+{
+    return float2half(v.x) | (float2half(v.y) << uint(16));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public float half2float(uint h)
+{
+    uint s = ((h & uint(0x8000)) << uint(16));
+    uint e = (((h & uint(0x7c00)) + uint(0x1c000)) << uint(13));
+    uint m = ((h & uint(0x03ff)) << uint(13));
+    return uintBitsToFloat(s | e | m); 
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public vec2 unpackHalf2x16(uint p)
+{
+    return vec2(half2float(p & uint(0xffff)), half2float(p >> uint(16)));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public double packDouble2x32(uvec2 v)
+{
+    return 0.0;
+    //return asdouble(uint64_t(v.x) | (uint64_t(v.y) << 32));
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public uvec2 unpackDouble2x32(double v)
+{
+    uint64_t u = 0; // asuint64(v);
+    return uvec2(uint(u & 0xFFFFFFFF), uint(u >> 32));
+}
+
+//
 // Section 8.5. Geometric Functions
 //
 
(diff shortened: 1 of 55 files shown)


source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+    return rsqrt(x);
+}
+
+//
+// Section 8.3. Common Functions
+//
+
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(glsl)
+[__readNone]
```
csyonghe, line 525: These functions should probably be marked [ForceInline] so slang can produce clean inlined code.

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+{
+    return (a ? y : x);
+}
+
+__generic<T, let N : int>
+__target_intrinsic(glsl)
+[__readNone]
+public vector<T, N> mix(vector<T, N> x, vector<T, N> y, vector<bool, N> a)
+{
+    vector<T, N> result;
```
csyonghe, line 600: Can be lerp(x, y, vector<T, N>(a));

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+__target_intrinsic(glsl)
+[__readNone]
+public vector<float, N> uintBitsToFloat(highp vector<uint, N> x)
+{
+    return asfloat(x);
+}
+
+__target_intrinsic(glsl)
+[__readNone]
+public float fma(float a, float b, float c)
```
csyonghe, line 670: This should probably be added to HLSL.meta.slang so it is available everywhere.

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+
+//
+// Section 8.5. Geometric Functions
+//
+
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public T dot(T x, T y)
```
csyonghe, line 691: This should already exist in HLSL.meta.slang. If not let's move it there.

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+[__readNone]
+[ForceInline]
+public T faceforward(T n, T i, T ng)
+{
+    return dot(ng, i) < T(0.0f) ? n : -n;
+}
+
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(glsl)
+[__readNone]
```
csyonghe, line 707: Same here. In general if HLSL.meta.slang is missing such an overload, let's just add the function there for completeness.

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+public vector<uint,N> uaddCarry(highp vector<uint,N> x, highp vector<uint,N> y, out lowp vector<uint,N> carry)
+{
+    VECTOR_MAP_TRINARY(uint, N, uaddCarry, x, y, carry);
+}
+
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(glsl)
+[ForceInline]
+public T normalize(T x)
```
csyonghe, line 752: Move to HLSL.meta.slang

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+public vector<uint,N> bitfieldReverse(highp vector<uint,N> value)
+{
+    VECTOR_MAP_UNARY(int, N, bitfieldReverse, value);
+}
+
+
+__target_intrinsic(glsl)
+public uint bitCount(uint value)
+{
+    value = value - ((value >> 1) & 0x55555555);
```
csyonghe, line 967: Should map to countbits

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,707 @@ ${{{{
...
+{
+    return firstbithigh(value);
+}
+
+__generic<let N : int>
+__target_intrinsic(glsl)
+[ForceInline]
+public vector<bool,N> not(vector<bool,N> x)
+{
+    vector<bool,N> result;
```
csyonghe, line 1060: Can just call logicalNot

csyonghe (approved):

jkwak-work:
I pushed a new change based on Yong's suggestions.

source/slang/hlsl.meta.slang
```diff
@@ -6041,6 +6051,16 @@ vector<T,N> normalize(vector<T,N> x)
...
 }
 
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(hlsl)
+__target_intrinsic(glsl)
+__target_intrinsic(spirv, "OpExtInst resultType resultId glsl450 Normalize _0")
+[__readNone]
+T normalize(T x)
+{
+    return x;
```
csyonghe, line 6061: What exactly should this do? Should it always return 1?

csyonghe (approved):

source/slang/glsl.meta.slang
```diff
@@ -407,13 +422,723 @@ ${{{{
...
+public T faceforward(T n, T i, T ng)
+{
+    return dot(ng, i) < T(0.0f) ? n : -n;
+}
+
+__generic<T : __BuiltinFloatingPointType>
+__target_intrinsic(glsl)
+[__readNone]
+[ForceInline]
+public T refract(T i, T n, T eta)
```
csyonghe, line 713: Should this be moved to hlsl.meta.slang as well?

source/slang/glsl.meta.slang
csyonghe: I see.

source/slang/hlsl.meta.slang
jkwak-work: Oh! You are right. I guess it should be always 1.

jkwak-work:
My 3rd commit moved refact from glsl.meta.slang to hlsl.meta.slang and it also fixed the mistake on "normalize" function.

source/slang/hlsl.meta.slang
csyonghe: Well, should technically be just sign(x), ignoring the 0 case...

tests/glsl-intrinsic/intrinsic-basic.slang
```diff
@@ -0,0 +1,474 @@
...
+
+    // 8.8. Integer Functions
+    umulExtended(genUType(0), genUType(0), outGenUType, outGenUType2);
+    voidResults = voidResults && (genUType(0) == outGenUType && genUType(0) == outGenUType2);
+
+    imulExtended(genIType(0), genIType(0), outGenIType, outGenIType);
+    voidResults = voidResults && (genIType(0) == outGenIType && genIType(0) == outGenIType2);
+
+    return voidResults
+        // 8.1. Angle and Trigonometry Functions
```
csyonghe, line 260: These tests can be improved by using non zero values.

tests/glsl-intrinsic/intrinsic-basic.slang
```diff
@@ -0,0 +1,474 @@
+//TEST:SIMPLE(filecheck=CHECK_GLSL): -allow-glsl -stage fragment -entry computeMain -target glsl
```
csyonghe, line 1: Any reason why we can't use executable test?

source/slang/hlsl.meta.slang
```diff
@@ -4328,7 +4338,7 @@ matrix<T, N, M> exp(matrix<T, N, M> x)
 
 __generic<T : __BuiltinFloatingPointType>
 __target_intrinsic(hlsl)
-__target_intrinsic(glsl)
+__target_intrinsic(glsl, "exp2(float($0))")
```
csyonghe, line 4341: I don't think this is right, especially if T is `float16_t`. Should be kept as `__target_intrinsic(glsl)`.

source/slang/hlsl.meta.slang
```diff
@@ -4553,31 +4562,32 @@ matrix<T, N, M> floor(matrix<T, N, M> x)
...
+// Fused multiply-add
+__generic<T : __BuiltinFloatingPointType>
 __target_intrinsic(hlsl)
 __target_intrinsic(glsl)
 __target_intrinsic(cuda, "$P_fma($0, $1, $2)")
 __target_intrinsic(cpp, "$P_fma($0, $1, $2)")
 __target_intrinsic(spirv, "OpExtInst resultType resultId glsl450 Fma _0 _1 _2")
 [__readNone]
-double fma(double a, double b, double c);
+T fma(T a, T b, T c);
```
csyonghe, line 4573: hlsl doesn't have `float`/`half` version of `fma`, should probably call `mad` instead.

csyonghe: Only call `fma` when T is double.

source/slang/hlsl.meta.slang
jkwak-work: The problem with exp2 on GLSL is that there is no exp2 for double-type.
What I am trying here is to use float version of it when it uses double-type.
But I didn't realized that __BuiltinFloatingPointType includes float16_t.

Hmmm... I will think about how to fix this.

tests/glsl-intrinsic/intrinsic-basic.slang
jkwak-work: The intention is to make sure the function signatures exist.
We will have to add more detailed functional tests especially for the new functions I added.
I was thinking that our new intern can do that.

tests/glsl-intrinsic/intrinsic-basic.slang
jkwak-work: I couldn't figure how to read back the resulting value and check it for all targets.
At the time of writing this code, if I use RWStructuredBuffer, it was giving me error when the target is glsl or SPIRV.
Please let me know if there is any ways to check if the resulting value is "true".

source/slang/hlsl.meta.slang
jkwak-work: I see.. I didn't know that fma is only for double on HLSL.
I guess I didn't need to make the changes in diff.meta.slang after all.

source/slang/hlsl.meta.slang
csyonghe: You can convert all the __target_intrinsic into a __target_switch inside the body, there you can do:
```
__target_switch
{
   ...
case glsl:
      if (__isHalf<T>() || __isFloat<T>())
            __intrinsic_asm "exp2($0)";
      else
           __intrinsic_asm "exp2(float($0))`;
}
```

tests/glsl-intrinsic/intrinsic-basic.slang
csyonghe: `RWStructuredBuffer` should work, and if it is crashing we should debug it together.

If you merge with top of tree, it should also work if you write SSBO syntax like:

```
buffer MyBlockName2
{
    float4 result;
} outputBuffer;
```

csyonghe: Then you can define the test line as:
```
//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl
```
With that, add these comments to the test file to check the buffer content:
```
// BUF: 1.0
// BUF: 0.0
// BUF: 0.0
// BUF: 0.0
```
This will ensure that the buffer content is `[1.0, 0.0, 0.0, 0.0]`.

csyonghe:
The Falcor test failure seems legit. We need to figure out why.

jkwak-work:
Yes, I will see what is going on.

jkwak-work:
I locally tested it with Farcor and I didn't have any problem on my Windows.
I am triggering the tests again with my latest changes.

csyonghe:
Falcor is now working, but tests/glsl-intrinsic/intrinsic-basic.slang seems to be failing?

jkwak-work:
The tests are triggered again after I disabled some of tests for exp2(double). I will figure out how to handle it on spirv later.

csyonghe:
The spirv intrinsics needs to implement the same casting it seems. Can be something like

let xf = float(x);
return double(spirv_asm
{
    result:$$float = OpExtInst ... $xf
});

jkwak-work:
I fixed exp2(double) function for SPIR-V.
The tests are trigged with my fix.

source/slang/hlsl.meta.slang
```diff
@@ -4561,16 +4571,34 @@ matrix<T, N, M> exp(matrix<T, N, M> x)
...
+            __intrinsic_asm "exp2($0)";
+        __intrinsic_asm "exp2(float($0))";
+    case spirv:
+        if (__isHalf<T>())
+        {
+            return spirv_asm { OpExtInst $$T result glsl450 Exp2 $x };
+        }
+        else
+        {
+            float xf = __slang_noop_cast<float>(x);
```
csyonghe, line 4592: `__slang_noop_cast` seems dangerous since it doesn't actually do any casting. We could define:
```
__intrinsic_op($(kIROp_FloatCast))
T __floatCast<T, U>(U v);
```
And then use
```
float xf = __floatCast<float>(x);
```

jkwak-work:
I replaced __slang_noop_cast to a new function __floatCast.
The tests are re-triggered but I expect them to all pass because the new change is trivial.

source/slang/hlsl.meta.slang
```diff
@@ -4589,7 +4592,7 @@ T exp2(T x)
         }
         else
         {
-            float xf = __slang_noop_cast<float>(x);
+            float xf = __floatCast<float>(x);
```
csyonghe, line 4595: There is already a `__realCast` defined in core.meta.slang that does exactly this, so we can just use that. Sorry I didn't realize that earlier.

csyonghe (approved):

Issue #3523, opened 2024-01-29
GLSL input compute shaders missing some layout information


expipiplus1:
Specifically `local_size_x` etc are being lost during translation when using GLSL as an input language.

PR #3528, opened 2024-01-30
Correctly apply glsl local size layout to entry points during lowering

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/3523

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -6938,7 +6938,6 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
     IGNORED_CASE(IncludeDecl)
     IGNORED_CASE(ImplementingDecl)
     IGNORED_CASE(UsingDecl)
-    IGNORED_CASE(EmptyDecl)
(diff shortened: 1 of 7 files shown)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -6947,6 +6946,29 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
 
 #undef IGNORED_CASE
 
+    LoweredValInfo visitEmptyDecl(EmptyDecl* decl)
+    {
+        for(const auto modifier : decl->modifiers)
+        {
+            if(const auto layoutLocalSizeAttr = as<GLSLLayoutLocalSizeAttribute>(modifier))
+            {
+                for(const auto d : context->irBuilder->getModule()->getModuleInst()->getGlobalInsts())
```
csyonghe, line 6955: How do we make sure the entry point has already been lowered at this point?

csyonghe: Instead of searching, we probably just keep a reference to the current entrypoint in the context.

csyonghe (approved):

PR #3554, opened 2024-02-06
deploy slang-glslang with macos releases

### LINKED ISSUE #3469: Failed to load downstream compiler 'spirv-opt'
Hi,

On macos (arm64), the latest slang release gives the following error when compiling anything
```
(0): error 100: failed to load downstream compiler 'spirv-opt'
(0): note 99999: failed to load dynamic library 'slang-glslang'
```

An example command is
```
slangc ./vk_helloworld_device.slang -profile sm_6_7 -target spirv -emit-spirv-directly -force-glsl-scalar-layout -fvk-use-entrypoint-name -matrix-layout-row-major -I ./projects/gprt/gprt -o ./vk_helloworld_device.spv
```

It seems `libslang-glslang.dylib` is not present in the release for macos. This leads to errors in finding `spirv-opt` based on this line: https://github.com/shader-slang/slang/blob/master/source/compiler-core/slang-glslang-compiler.cpp#L376
```c++
static SlangResult locateGlslangSpirvDownstreamCompiler(const String& path, ISlangSharedLibraryLoader* loader, DownstreamCompilerSet* set, SlangPassThrough compilerType)
{
    ...
    SLANG_RETURN_ON_FAIL(DownstreamCompilerUtil::loadSharedLibrary(path, loader, nullptr, "slang-glslang", library));

    SLANG_ASSERT(library);
    if (!library)
    {
        return SLANG_FAIL;
    }

    auto compiler = new GlslangDownstreamCompiler(compilerType);
    ComPtr<IDownstreamCompiler> compilerIntf(compiler);
    SLANG_RETURN_ON_FAIL(compiler->init(library));

    set->addCompiler(compilerIntf);
    return SLANG_OK;
}
```

This error can be fixed if I manually compile slang with
```
./external/slang-binaries/premake/premake-5.0.0-alpha16/bin/osx/premake5 gmake2 --cc=clang --enable-xlib=false --enable-embed-stdlib=true --arch=aarch64 --deps=true --build-glslang=true
make config=release_aarch64 -j8
```

The `--build-glslang=true` flag seems to be important, but it is not present in `github_macos_build.sh`.

westlicht:
Fixes #3469.

### DIFF
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -66,6 +66,7 @@ jobs:
           IDENTITY_ID: d6ada82a113e4204aaad914e1013e9548ffd30d0
         run: |
           /usr/bin/codesign --force --options runtime -s ${IDENTITY_ID} ./bin/macosx-${{matrix.targetPlatform}}/release/libslang.dylib -v
+          /usr/bin/codesign --force --options runtime -s ${IDENTITY_ID} ./bin/macosx-${{matrix.targetPlatform}}/release/libslang-glslang.dylib -v
           /usr/bin/codesign --force --options runtime -s ${IDENTITY_ID} ./bin/macosx-${{matrix.targetPlatform}}/release/libgfx.dylib -v
           /usr/bin/codesign --force --options runtime -s ${IDENTITY_ID} ./bin/macosx-${{matrix.targetPlatform}}/release/slangd -v
           /usr/bin/codesign --force --options runtime -s ${IDENTITY_ID} ./bin/macosx-${{matrix.targetPlatform}}/release/slangc -v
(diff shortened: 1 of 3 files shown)


.github/workflows/release-macos.yml
```diff
@@ -87,6 +87,7 @@ jobs:
           7z a ${SLANG_BINARY_ARCHIVE} slang-gfx.h
           7z a ${SLANG_BINARY_ARCHIVE} prelude/*.h
           7z a ${SLANG_BINARY_ARCHIVE} bin/*/*/libslang.dylib
+          7z a ${SLANG_BINARY_ARCHIVE} bin/*/*/libslang-glslang.dylib
```
csyonghe, line 90: Need to codesign and notarize this dylib too.

westlicht: Done. Also added `libslang-glslang.dylib` to the dist package.

csyonghe (approved):

PR #3572, opened 2024-02-12
cmake: add options to disable gfx, test, example, ... targets

fknfilewalker:
This PR allows to disable targets (gfx, tests, examples) that are not really necessary for projects that want to include slang as a cmake dependency for enduser usecases.

Not sure if these options should be on or off by default or if there are more targets that could also be optionally disabled.


csyonghe:
They should all be default ON, or it will break our CI setup. Meanwhile we also have users who depends on `gfx` and look for examples.

csyonghe:
Other optional projects are: slangd, slang-rt,, and vk-piipeline-create.

fknfilewalker:
> Other optional projects are: slangd, slang-rt,, and vk-piipeline-create.

I guess llvm bins are required for slang-rt, so fetching llvm bins could also be disabled when slang-rt is disabled?

Do tests or examples depend on slang-rt?

csyonghe:
slang-llvm is required whenever the user is requesting Slang to generate CPU code from their shaders. Both slang-llvm and slang-glslang are optional. But when they are missing Slang won't be able to produce code for those targets.

fknfilewalker:
> slang-llvm is required whenever the user is requesting Slang to generate CPU code from their shaders. Both slang-llvm and slang-glslang are optional. But when they are missing Slang won't be able to produce code for those targets.

ok did that
for tools/slang-test I had to make the glslang dep dynamic, I did run the test afterwards and it seems fine

fknfilewalker:
Left side was before on my system, and right now with everything turned off (just pure slang remains)
image

csyonghe:
We have direct spirv gen implemented and about to switch that as the default. However, we still pass our generated spirv to spirv-opt, otherwise we seem to hit driver crashes despite the slang generated SPIRV is valid.

csyonghe:
And we build spirv-opt in slang-glslang, so that dll needs to continue to be around for a while.

csyonghe (dismissed):

expipiplus1 (changes requested):
This duplicates the functionality of `SLANG_SLANG_LLVM_FLAVOR=DISABLE`.

Is it possible to also update `docs/building.md` please?

expipiplus1:
The naming is fine as it is.

csyonghe (approved):

expipiplus1:
Thank you!

PR #3578, opened 2024-02-13
cmake: option to build a static library version of slang

fknfilewalker:
This adds an option to cmake that allows to build and link a static version of the slang library.
The resulting `slang.lib` is about 1 GB in size in debug, which is a bit on the larger side but I don't know why.


expipiplus1:
ok looks good, I think that the name would be better as `SLANG_LIB_TYPE` (or maybe `SLANG_SLANG_LIB_TYPE`)

expipiplus1:
@fknfilewalker just to make sure, you've tested this and it's useful to you right? If so then it's good to merge.

csyonghe (approved):

fknfilewalker:
Yes I did run it. The only downside is the quite large lib size.
There is a whole blog about making dxc better just by building it as a static library so I guess this could be useful to more people.
https://devlog.hexops.com/2024/building-the-directx-shader-compiler-better-than-microsoft/

PR #3580, opened 2024-02-13
Partially implement shader_subgroup extension(s); Partially resolves #3548

### LINKED ISSUE #3548: Implement "subgroup" extension for GLSL in glsl.meta.slang and bring the parity to HLSL
This task is a subtask from the other issue #3362 , which is a big task.
This issue is to cover a smaller portion of it so that we can work with other people together.

We have "subgroup" functionality implemented for HLSL but it is not implemented for GLSL.
The goal of this task is to bring the functionality for GLSL as much as slangc already does for HLSL.

The description of the "KHR_shader_subgroup" extension can be found from [here](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_shader_subgroup.txt).

I found this document very useful for the task.
https://github.com/shader-slang/slang/blob/master/docs/user-guide/a1-04-interop.md

ArielG-NV:
The following PR partially implements shader_subgroup extension(s) (GL_KHR_shader_subgroup & GLSL_EXT_shader_subgroup_extended_types); Fully implement all functions, stub's all built-in variables; Partially resolves #3548

GL_KHR_shader_subgroup implemented based on [Khronos specification links]:
    * https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_shader_subgroup.txt
    * https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_shader_subgroup_extended_types.txt

This change implements following functions described in the KHR_shader_subgroup. The extension consists of the following extensions:
    GL_KHR_shader_subgroup_basic
    GL_KHR_shader_subgroup_vote
    GL_KHR_shader_subgroup_arithmetic
    GL_KHR_shader_subgroup_ballot
    GL_KHR_shader_subgroup_shuffle
    GL_KHR_shader_subgroup_shuffle_relative
    GL_KHR_shader_subgroup_clustered
    GL_KHR_shader_subgroup_quad

Notes of worthy{

    * added preamble function and macros for implementing subgroup functionality (and tests) to make it possible to iterate on the functionality with reasonable effort in the future

    * __requireGLSLExtension does not work as intended inside glsl.meta; as a result half, int16, int64 int8, all are ommited from testing for glsl shader_subgroup type extension support
    
    * CUDA, CPP, HLSL implementations were largly out of scope and not implemented. This is due to the implementation being non trivial for many functions

    * doubles cause a silent crash on most subgroup functions tested (silent shader hang) on Intel Iris due to missing F64 support 

    * built-in variables are only **stubbed** (not implemented) due to being a non trivial implementation (seperate task)

    * due to spec stating that and/or/xor are logical ops under bool and have different spir-v functions for this, subgroupAnd/Or/Xor and related functions were reimplementated under glsl/spir-v to accept different types than the hlsl implementation, and work with them logically differently

    * commits should describe all major changes and deviations from hlsl.meta 
}

Random fixes{

    * hlsl.meta incorrectly sets `OpCapability` as `GroupNonUniformBallot` when the `OpCapability` should be `GroupNonUniformVote`; this is as per SPIR-V spec for all SPIR-V calls used in `GL_KHR_shader_subgroup_vote`: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGroupNonUniformAll

    * hlsl.meta incorrectly uses for WaveMaskPrefixBitOr (SPIR-V) OpGroupNonUniformBitwiseAnd intead of OpGroupNonUniformBitwiseOr
}


source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+
+//macros used to make implementation and modification reasonable under regex
+#define macroAnyVecGenericBuiltIn __generic<T : __BuiltinType, let N : int>
+#define macroAnyVecGenericLogical __generic<T : __BuiltinLogicalType, let N : int>
+#define macroAnyVecGenericArithmetic __generic<T : __BuiltinArithmeticType, let N : int>
+#define macroAnyVec vector<T, N>
+
+__generic<T : __BuiltinType>
+void shader_subgroup_preamble() {
+    // checks needed for shader_subgroup functions; __requireGLSLExtension does not work, the call fails to find the OP
```
jkwak-work, line 2843: I am not sure what this comment means.
It says __requireGLSLExtension does not work, but you are using it at 4 lines below.
And shouldn't you be using "__glsl_extension", instead?

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+// 'system' variables: https://github.com/shader-slang/slang/issues/411
+
+#define glsl_caps_shader_subgroup(param1) __glsl_extension(param1) [require(glsl)] 
+#define spirv_caps_shader_subgroup(param1) __spirv_version(param1) [require(spirv)] 
+
+//macros used to make implementation and modification reasonable under regex
+#define macroAnyVecGenericBuiltIn __generic<T : __BuiltinType, let N : int>
+#define macroAnyVecGenericLogical __generic<T : __BuiltinLogicalType, let N : int>
+#define macroAnyVecGenericArithmetic __generic<T : __BuiltinArithmeticType, let N : int>
+#define macroAnyVec vector<T, N>
```
jkwak-work, line 2839: Let's not use define macro as much as we can.
It may seem verbose but we don't want to scroll up and down to see what the macro define might be doing something wrong.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
     return 0;
 }
 
+// TODO: implementation of built-in variables; proper tests; these are stubs
+// likley related to the following issue since GLSL adds new 
+// 'system' variables: https://github.com/shader-slang/slang/issues/411
+
+#define glsl_caps_shader_subgroup(param1) __glsl_extension(param1) [require(glsl)] 
+#define spirv_caps_shader_subgroup(param1) __spirv_version(param1) [require(spirv)]
```
jkwak-work, line 2833: Let's not use a macro defines as much as possible.
This forces the reader to scroll up and down to check what the macro is doing.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    case spirv:
+        spirv_asm {
+            OpCapability Shader;
+            OpMemoryBarrier Subgroup AcquireRelease|UniformMemory
+        };
+    case cpp:
+        // TODO: cpp
+        return;
+    }
+}
```
jkwak-work, line 2959: Can this be like following?
```
[ForceInline] public void subgroupMemoryBarrierBuffer()
{
    return DeviceMemoryBarrier();
}
```

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    case spirv:
+        if (__isBool<T>()) return spirv_asm {OpCapability GroupNonUniformArithmetic; OpGroupNonUniformLogicalOr $$T result Subgroup ExclusiveScan $value};
+        else return spirv_asm {OpCapability GroupNonUniformArithmetic; OpGroupNonUniformBitwiseOr $$T result Subgroup ExclusiveScan $value};
+    case hlsl:
+    case cuda:
+    case cpp:
+        // TODO: implementation
+        return value;
+    }
+}
```
jkwak-work, line 3499: Is seems like there is WaveMaskPrefixBitOr in hlsl.meta.slang.
Can we just call that?

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
```diff
@@ -0,0 +1,210 @@
...
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl -emit-spirv-directly
+
+#version 430
+
+// breaks on all gpu's tested (Nvidia & Intel) -- GLSL does not work, SPIR-V does work
+//#define TEST_when_glsl_subgroupInclusiveXor_is_not_bugged
+//#define TEST_when_glsl_subgroupInclusiveAdd_is_not_bugged
+
+// __requireGLSLExtension does not work with GLSL, as a result half/i8/i16/i64 all don't run
+//#define TEST_when_requireGLSLExtension_works
```
jkwak-work, line 19: Can you describe more details of what you meant by "not work"?
Have you debugged and see what is happening?

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
```diff
@@ -0,0 +1,129 @@
...
+
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl -emit-spirv-directly
+#version 430
+
+// __requireGLSLExtension does not work with GLSL, as a result half/i8/i16/i64 all don't run
+//#define TEST_when_requireGLSLExtension_works
+
+precision highp float;
+precision highp int;
```
jkwak-work, line 19: Does there actually make any difference?
If not, can we remove them?

jkwak-work (commented):
I have some comments about how glsl.meta.slang is implemented

source/slang/glsl.meta.slang
ArielG-NV: as done in hlsl.meta.slang (`WaveMaskPrefixSum`), 
``` if (__isHalf<T>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_float16"); ``` 
is used to enable an extension which is required based on type used with an ext.

despite this, __requireGLSLExtension does not work with glsl targeted files (glsl.meta.slang)

File: source/slang/glsl.meta.slang (hunk 4 above)
Comment (ArielG-NV) [reply source/slang/glsl.meta.slang:2959]: subgroupMemoryBarrierBuffer targets syncing within the subgroup, the defined hlsl.meta.slang function targets a larger granularity of barrier. HLSL lacks an equivalent to GLSL & SPIR-V.

File: source/slang/glsl.meta.slang (hunk 1 above)
Comment (jkwak-work) [reply source/slang/glsl.meta.slang:2843]: Again, what do you mean by "not work".
Does it causes a compile error?
Or is the requirement applied to all types?
Or is the line doing nothing?

File: source/slang/glsl.meta.slang (hunk 4 above)
Comment (jkwak-work) [reply source/slang/glsl.meta.slang:2959]: Probably it will be good idea to leave a comment with the information.
Because any programmer who will look at the function will have the same question as I had.

File: source/slang/glsl.meta.slang (hunk 5 above)
Comment (ArielG-NV) [reply source/slang/glsl.meta.slang:3499]: we cannot since as per spec:
```
genIType subgroupExclusiveOr(genIType value);
genUType subgroupExclusiveOr(genUType value);
genBType subgroupExclusiveOr(genBType value);
```
where bool types do logical operations rather than bitwise; this is different from HLSL which does bitwise operations (bool(0x10) | bool(0x10) will work differently across HLSL & GLSL)

there is another issue with hlsl.meta using __BuiltinArithmeticType, although I believe hlsl.meta should be using __BuiltinLogicalType (waveallbitor is defined with __int_type__, but this was not looked into)

File: tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang (hunk 6 above)
Comment (ArielG-NV) [reply tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang:19]: __requireGLSLExtension does not enable the extension specified as it should be for glsl targets (the error is a missing extension for enabling extended features for shader_subgroup type used)

Comment (ArielG-NV) [reply tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang:19]: (elaboration added under comment)

File: source/slang/glsl.meta.slang (hunk 4 above)
Comment (ArielG-NV) [reply source/slang/glsl.meta.slang:2959]: (elaboration added in code comment)

File: source/slang/glsl.meta.slang [hunk 8]
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    __target_switch
+    {
+    case glsl:
+        if (__type_equals<T, half>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_float16");
+        if (__type_equals<T, uint8_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int8");
+        if (__type_equals<T, uint16_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int16");
+        if (__type_equals<T, uint64_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int64");
+        return;
+    case spirv:
+    case cpp:
```
csyonghe, line 2853: Don't provide a case on target that we don't support this intrinsic.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    // checks needed for shader_subgroup functions; __requireGLSLExtension does not work, the call fails to find the OP
+    __target_switch
+    {
+    case glsl:
+        if (__type_equals<T, half>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_float16");
+        if (__type_equals<T, uint8_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int8");
+        if (__type_equals<T, uint16_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int16");
+        if (__type_equals<T, uint64_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int64");
+        return;
+    case spirv:
```
csyonghe, line 2852: There needs to be an inline spirv block here for the spirv backend.

source/slang/glsl.meta.slang
ArielG-NV: does this imply for all not yet implemented intrinsics?
example of what I am asking about:
```
__glsl_extension(GL_KHR_shader_subgroup_vote) [require(glsl)] 
__spirv_version(1.3) [require(spirv)]
[ForceInline] public bool subgroupAll(bool value)
{
    __target_switch
    {
    case cuda: 
    case glsl:
    case spirv:
    case hlsl:
        return WaveActiveAllTrue(value);
    case cpp:
        // TODO: cpp
        return false;
    }
}
```  

to:
```
__glsl_extension(GL_KHR_shader_subgroup_vote) [require(glsl)] 
__spirv_version(1.3) [require(spirv)]
[ForceInline] public bool subgroupAll(bool value)
{
    __target_switch
    {
    case cuda: 
    case glsl:
    case spirv:
    case hlsl:
        return WaveActiveAllTrue(value);
    }
}
```

csyonghe: That's right.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+
+// GL_KHR_shader_subgroup_arithmetic
+
+__generic<T : __BuiltinArithmeticType>
+glsl_caps_shader_subgroup(GL_KHR_shader_subgroup_arithmetic)
+spirv_caps_shader_subgroup(1.3)
+[ForceInline] public T subgroupAdd(T value)
+{
+    shader_subgroup_preamble<T>();
+    return WaveMaskSum(WaveGetActiveMask(), value);
```
csyonghe, line 3078: Should just call `WaveActiveSum`.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1988 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+}
+
+__generic<T : __BuiltinArithmeticType>
+glsl_caps_shader_subgroup(GL_KHR_shader_subgroup_arithmetic)
+spirv_caps_shader_subgroup(1.3)
+[ForceInline] public T subgroupMul(T value)
+{
+    shader_subgroup_preamble<T>();
+    return WaveMaskProduct(WaveGetActiveMask(), value);
```
csyonghe, line 3087: Same here, call WaveActive* functions instead.

source/slang/glsl.meta.slang
ArielG-NV: __requireGLSLExtension is doing nothing, but it should be doing something (adding the extension to the final compiled glsl code)

source/slang/glsl.meta.slang
csyonghe: Is this still needed?

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1762 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+__generic<T : __BuiltinType>
+[require(glsl)]
+void typeRequireChecks_shader_subgroup_GLSL() {
+    // the following is a seperate function call, since else the `__requireGLSLExtension` and associated __intrinsic_asm is ignored if the calling function also calls an __intrinsic_asm
+    __target_switch
+    {
+    case glsl:
+        if (__type_equals<T, half>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_float16");
+        else if (__type_equals<T, float16_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_float16");
+        else if (__type_equals<T, uint8_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int8");
```
jkwak-work, line 2841: Is there a reason why you don't use a logical or here?
```
else if (__type_equals<T, uint8_t>() || __type_equals<T, int8_t>()) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int8");

File: tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang [hunk 13]
```diff
@@ -0,0 +1,186 @@
+//TEST:SIMPLE(filecheck=CHECK_GLSL):  -allow-glsl -stage compute -entry computeMain -target glsl
+//TEST:SIMPLE(filecheck=CHECK_SPV):  -allow-glsl -stage compute -entry computeMain -target spirv
+//TEST:SIMPLE(filecheck=CHECK_HLSL): -allow-glsl -stage compute -entry computeMain -target hlsl -DTARGET_HLSL
+//TEST:SIMPLE(filecheck=CHECK_CUDA): -allow-glsl -stage compute -entry computeMain -target cuda -DTARGET_CUDA 
+
+// not testing cpp due to missing impl
+//T-EST:SIMPLE(filecheck=CHECK_CPP):  -allow-glsl -stage compute -entry computeMain -target cpp
+
```
jkwak-work, line 8: Let's remove this.
We can add this when CPP side gets implemented.

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
```diff
@@ -0,0 +1,186 @@
...
+        && subgroupAnd(T(1)) == T(1)
+        && subgroupOr(T(1)) == T(1)
+        && subgroupXor(T(1)) == T(0) 
+        && subgroupInclusiveAnd(T(1)) == T(1)
+        && subgroupInclusiveOr(T(1)) == T(1)
+        && subgroupInclusiveXor(T(1)) == T(0)
+        && subgroupExclusiveAnd(T(1)) == T(1)
+        && subgroupExclusiveOr(T(1)) == T(1)
+        && subgroupExclusiveXor(T(1)) == T(1)
+#endif // #if !defined(TARGET_HLSL) && !defined(TARGET_CUDA)
```
jkwak-work, line 36: I think we will need a comment or more meaningfull macro define about why this block is commented out for HLSL and CUDA.
Something like
```
#if !defined(TARGET_HLSL) && !defined(TARGET_CUDA)
// HLSL doesn't have logical operators for subgroup equivalent.
// CUDA side support is not implemented yet.
#define TEST_when_subgroup_logical_operators_are_implemented
#endif
...
#if defined(TEST_when_subgroup_logical_operators_are_implemented)
...
#endif
```

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-ballot.slang
```diff
@@ -0,0 +1,134 @@
...
+        && testVBroadcastX<int16_t, 3>()
+        && testVBroadcastX<int16_t, 4>()
+        && test1BroadcastX<int64_t>() 
+        && testVBroadcastX<int64_t, 2>()
+        && testVBroadcastX<int64_t, 3>()
+        && testVBroadcastX<int64_t, 4>()
+        && test1BroadcastX<uint>()
+        && testVBroadcastX<uint, 2>()
+        && testVBroadcastX<uint, 3>()
+        && testVBroadcastX<uint, 4>()
```
jkwak-work, line 95: Maybe we can add uint8_t and uint16_t as well?

jkwak-work (commented):
I leave some minor comments.

source/slang/glsl.meta.slang
ArielG-NV: for `case glsl`: __requireGLSLExtension is required

for `case: spirv`: we are required to output some spirv_asm (including empty block) to avoid crash (tested just now to be sure). The OpCapability code added can be removed since Slang handles the Int64/16/8/float16 capability under the hood

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-ballot.slang
```diff
@@ -0,0 +1,134 @@
...
+}
+
+bool testBallot() {
+    return true 
+        && (subgroupBallot(true).x == 0xFFFFFFFF)
+        && (subgroupInverseBallot(uvec4(0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF)) == true)
+        && (subgroupBallotBitExtract(uvec4(0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF), 0) == true)
+        && (subgroupBallotBitCount(uvec4(0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF)) == 32)
+        && (subgroupBallotInclusiveBitCount(uvec4(0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF)) != 0)
+#ifdef TEST_when_glsl_subgroupBallotExclusiveBitCount_is_not_bugged
```
jkwak-work, line 110: Can you fix this?
It seems like the error message is following:
```
glsl.meta.slang(2841): warning 41000: unreachable code detected
        return;
        ^~~~~~
```
When the issue is reproduced, you can see the error message in a file, "shader-subgroup-ballot.slang.3.actual"

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
ArielG-NV: we could use DISABLE_TEST so when running tests on a file it is shared that a test is ignored (we need to one day add cpp/cuda support)

jkwak-work (commented):
Can we remove TEST_when_glsl_subgroupBallotExclusiveBitCount_is_not_bugged ?

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-ballot.slang
ArielG-NV: I can't reproduce the issue despite anything I try, may be due to the new branch having changes

source/slang/glsl.meta.slang
```diff
@@ -2836,58 +2836,20 @@ void typeRequireChecks_shader_subgroup_GLSL() {
...
-        else if (__type_equals<T, int16_t>()) {
-
-            spirv_asm {OpCapability Int16};
-        }
-        else if (__type_equals<T, int64_t>()) {
-
-            spirv_asm {OpCapability Int64};
-        }
-        else spirv_asm {};
+        __intrinsic_asm "";
```
csyonghe, line 2852: This can just be `return`.

source/slang/glsl.meta.slang
csyonghe: Can we just `return`?

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1724 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
     return 0;
 }
 
+// TODO: implementation of built-in variables; proper tests; these are stubs
+// likley related to the following issue since GLSL adds new 
+// 'system' variables: https://github.com/shader-slang/slang/issues/411
+
+__generic<T : __BuiltinType>
+[require(glsl)]
+void typeRequireChecks_shader_subgroup_GLSL() {
```
csyonghe, line 2834: Add `[ForceInline]` here.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1724 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+        else if (__type_equals<T, uint64_t>()
+            || __type_equals<T, int64_t>()
+            ) __requireGLSLExtension("GL_EXT_shader_subgroup_extended_types_int64");
+
+        __intrinsic_asm "";
+    }
+}
+
+__generic<T : __BuiltinType>
+[require(spirv)] [require(glsl)]
```
csyonghe, line 2857: These `[require]` declarations can be omitted here since 1) it is not public, and 2) Slang can derive that from the `__target_switch`.

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1724 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    case spirv:
+        spirv_asm {};
+    }
+
+} 
+
+// GL_KHR_shader_subgroup_basic Built-in Variables 
+
+__glsl_extension(GL_KHR_shader_subgroup_basic) [require(glsl)] 
+public uint gl_NumSubgroups;
```
csyonghe, line 2875: We need to turn these into properties and define the SPIRV intrinsics for them.

```
[require(glsl)] [require(spirv)]
public property uint gl_NumSubgroups
{
     get
     {
           __target_switch
          {
          case glsl:
                  __requireGLSLExtension("GL_KHR_shader_subgroup_basic");
                  __intrinsic_asm "gl_NumSubgroups";
           case spirv:
                  return spirv_asm {
                       result:$$uint = OpLoad builtin(NumSubgroups:uint);
                    };
          }
     }
}
```

source/slang/glsl.meta.slang
ArielG-NV: we can use 'return' without any asm output

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1724 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
+    }
+}
+
+// GL_KHR_shader_subgroup_vote
+
+__glsl_extension(GL_KHR_shader_subgroup_vote) [require(glsl)] 
+__spirv_version(1.3) [require(spirv)]
+[ForceInline] public bool subgroupAll(bool value)
+{
+    __target_switch
```
csyonghe, line 3032: __target_swtich isn't needed here.

source/slang/glsl.meta.slang
ArielG-NV: without outputting at a minimum intrinsic_asm (just `return` does not work) there is a problem with `internal error 99999: unimplemented feature in Slang compiler`

source/slang/glsl.meta.slang
```diff
@@ -2825,21 +2825,1724 @@ public uint rayQueryGetIntersectionTypeEXT(rayQueryEXT q, bool committed)
...
 }
 
+// TODO: implementation of built-in variables; proper tests; these are stubs
+// likley related to the following issue since GLSL adds new 
+// 'system' variables: https://github.com/shader-slang/slang/issues/411
+
+__generic<T : __BuiltinType>
+[require(glsl)]
+void typeRequireChecks_shader_subgroup_GLSL() {
+    // the following is a seperate function call, since else the `__requireGLSLExtension` and associated __intrinsic_asm is ignored if the calling function also calls an __intrinsic_asm
```
jkwak-work, line 2835: This seems like something we need to debug and fix; not for this PR tho.

jkwak-work (commented):
I left a trivial comment you can ignore.

source/slang/glsl.meta.slang
ArielG-NV: has the following feature been tested outside of defining in a struct/related-type?

I've been finding that `public property uint gl_NumSubgroups` as a standalone does not really work?

csyonghe: You probably need to modify `isDeclAllowed` to allow a PropertyDecl to be nested under ModuleDecl at slang-parser.cpp:4215.

ArielG-NV: propertiesDecl does have permission; 

the error which was provided seems to be an internal compiler related subject(?): `error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unimplemented: addInArg case`

source/slang/glsl.meta.slang
jkwak-work: The best lines im source code is "no lines". They are error free and we don't need to maintain. It is always best to minimize your changes.

source/slang/glsl.meta.slang
jkwak-work: Let me know if you need help on debugging the error. You can search the souce files with the given error messages and set a break point there. And when you reproduce the issue the program will stop at the breakpoints, and you can trace backward to figure out what condition made the program to branch into the error message printing code. That condition will be the real reason why you were getting the error message.

Don't be scared by words like "internal error" or "unimplemented". Those words actually mean "developers need to debug this with a debugger attached".

ArielG-NV:
fixing pushed code which includes raytracing changes*

csyonghe (approved):

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
```diff
@@ -0,0 +1,219 @@
...
+#version 430
+
+#define DEF_BLOCK
+#if defined(DEF_BLOCK)\
+    && !defined(TARGET_HLSL) \
+    && !defined(TARGET_CUDA)
+// hlsl does not treat boolean types with subgroup.* as a logical operator
+// cuda is missing an implementation
+#define TEST_when_logical_operators_are_implemented
+#endif
```
jkwak-work, line 20: I am not sure what "DEF_BLOCK" is here for.
If it is for debugging purpose, you can do something like this but remove when you submit.
```
#if 0 && !defined(TARGET_HLSL) && !defined(TARGET_CUDA)
```
or
```
#if 1 || !defined(TARGET_HLSL) && !defined(TARGET_CUDA)
```

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
```diff
@@ -0,0 +1,219 @@
...
+#endif // #if !defined(TEST_when_logical_operators_are_implemented)
+        ;
+}
+
+__generic<T : __BuiltinLogicalType, let N : int>
+bool testVLogical() {
+    typealias gvec = vector<T, N>;
+
+    return true
+#if !defined(TEST_when_logical_operators_are_implemented)
```
jkwak-work, line 53: Shouldn't this be without "!" like this?
```
#if defined(TEST_when_logical_operators_are_implemented)
```

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
```diff
@@ -0,0 +1,131 @@
...
+    uint data[];
+} outputBuffer;
+
+layout(local_size_x = 32) in;
+
+__generic<T : __BuiltinLogicalType>
+bool test1ShuffleX() {
+    return true
+        && subgroupShuffle(T(1), 1) == T(1)
+#if !defined(TARGET_CUDA) && !defined(TARGET_HLSL)
```
jkwak-work, line 27: It seems that you can now remove this.
I tested on my local machine and HLSL and CUDA works fine for "subgroupShuffleXor()".

Although the implementation for CUDA and HLSL is missing, I don't see why we want to explicitly remove from the testing with the ugly #if  when they are not needed.

jkwak-work (commented):
I left some comments.

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
ArielG-NV: (I forgot to share a comment on this) -- I agree, the if'defs should be using the cleaner syntax desired as in the arithmetic test. 

for the subgroupShuffleXor(), it should not compile for HLSL/CUDA, I don't know how it did on your local machine, but there is no implementation -- at a minimum slang will complain about a warning on returning values (and therefore not compile stdlib)

ArielG-NV:
note: the last 3 testing ci commits did not change anything, the test which failed (but passed locally) just passes now

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic.slang
```diff
@@ -0,0 +1,219 @@
+//TEST:SIMPLE(filecheck=CHECK_GLSL):  -allow-glsl -stage compute -entry computeMain -target glsl -DTARGET_GLSL
+//TEST:SIMPLE(filecheck=CHECK_SPV):  -allow-glsl -stage compute -entry computeMain -target spirv -emit-spirv-directly -DTARGET_SPIRV
```
csyonghe, line 2: I don't think -D here will work as expected. You need to wrap it with -xslang to pass it through to the slang compiler. Here you are just specifying command line args to slang-test.

should be:
```
-xslang -DTARGET_SPIRV
```

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
```diff
@@ -0,0 +1,139 @@
...
+
+void computeMain()
+{
+
+    outputBuffer.data[0] = true
+        && testShuffleX()
+        ;
+
+    // CHECK_GLSL: void main(
+    // CHECK_SPV: OpEntryPoint
```
csyonghe, line 134: These checks are too simplistic. At least verify the expected SPIRV opcodes are existent.

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
```diff
@@ -0,0 +1,139 @@
...
+//TEST:SIMPLE(filecheck=CHECK_SPV):  -allow-glsl -stage compute -entry computeMain -target spirv -emit-spirv-directly
+
+// not testing hlsl due to missing impl
+//DISABLE_TEST:SIMPLE(filecheck=CHECK_HLSL): -allow-glsl -stage compute -entry computeMain -target hlsl -DTARGET_HLSL
+// not testing cuda due to missing impl
+//DISABLE_TEST:SIMPLE(filecheck=CHECK_CUDA): -allow-glsl -stage compute -entry computeMain -target cuda -DTARGET_CUDA 
+// not testing cpp due to missing impl
+//DISABLE_TEST:SIMPLE(filecheck=CHECK_CPP):  -allow-glsl -stage compute -entry computeMain -target cpp -DTARGET_CPP
+
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl
```
csyonghe, line 11: These runnable tests seem to have inconsistency on the CI server across runs. We either need to get to the bottom of that, or simply disable runnable tests and rely on more thorough SIMPLE test pattern checking.

ArielG-NV: I will look into the CI consistency issue (I believe it is likely related to subgroup size & partitioning)

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
ArielG-NV: with generics the string checking is quite unreliable; for the larger testing code base, most code does not check for every op code (especially for bigger templated tests).

I can probably still add, but it will take a while to add to all of the files

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
ArielG-NV: looking at the test logs more and the probably launch order of tests, arithmetic.5 & arithmetic.6 on full capability tests finish quite a bit after the other arithmetic tests when pass/fail; if they run only a partial test they finish very close to the other arithmetic tests; I believe this is a sign that the tests were taking too long due to testing too many types & permutations 

(likely need to split up the test)

csyonghe: Yeah, it could be a timeout issue.

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang
jkwak-work: I agree that we should check the expected SPIRV opcodes.
We can probably do it as a new task later.

For now, the following test will cover the spir-v test.
```
//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl -emit-spirv-directly
```

jkwak-work:
Looks good to me.

csyonghe (approved):

PR #3583, opened 2024-02-14
HLSL texture intrinsic test first draft.

tgrimesnv:
First draft of texture intrinsic hlsl test.

Some things to note (possible bugs found):
- Some intrinsics are only available in pixel shaders and this is a compute shader so those can't be tested here. But most can be.

- The offset variant of SampleLevel causes a fail/crash in the "cpu" profile. So I had to disable cpu.

- SampleCmpLevelZero was causing glslang errors or Vulkan validation layer issues, so I had to disable Vulkan for now.

- GetDimensions for CubemapArray and float parameters were failing/crashing. Those are commented out for now.

- SampleGrad for Cubemap seems to be defined in slang with the wrong parameters. The DDX/DDY parameters should be of type float3, but slang seems to have them as float2. Using either one will result in a failure/crash. These are currently commented out. (BTW, I know in the commented out function I have a mismatch where it says float3 but there's only two parameters, ignore that)


csyonghe:
Looks good. We should fix the issues uncovered in this test.
The d3d11 test seems to fail on github hosted CI machines, so maybe add a -dx11 test line with `-render-feature hardware-device` to disable it on CI machines that does not have a  GPU.

csyonghe:
Can you create separate issues for the bugs?

tgrimesnv:
> Looks good. We should fix the issues uncovered in this test. The d3d11 test seems to fail on github hosted CI machines, so maybe add a -dx11 test line with `-render-feature hardware-device` to disable it on CI machines that does not have a GPU.

I'll add that line. That seems like it would likely explain the dx11 issue which I do not see locally.

> Can you create separate issues for the bugs?

Will do.

csyonghe (approved):

csyonghe:
Can you merge with top of tree and see if the GetDimension issues are still there? There are some recent fixes on those intrinsics.

tgrimesnv:
Since SampleGrad for Cubemaps now works, I went ahead and uncommented it and pushed.

tgrimesnv:
I will now try the GetDimensions issues.

csyonghe (approved):

Issue #3493, opened 2024-01-25
[Slang Unit Tests] Intrinsics coverage SLANG/HLSL (HLSL.meta.slang and core.meta.slang files)


akolheNV:

tgrimesnv:
Added RT pipeline test from PR 3793

tgrimesnv:
I think this one can be closed. If there's anything else that comes up a new issue can be made for that.

PR #3623, opened 2024-02-24
atomic intrinsic test

tgrimesnv:
Adding atomic intrinsic tests on uint/int RWStructuredBuffer's and groupshared memory.

I was seeing Vulkan validation errors when using RW textures, so that will need an issue and a follow-up test later on.

This is for: https://github.com/shader-slang/slang/issues/3493 - Intrinsics coverage SLANG/HLSL (HLSL.meta.slang and core.meta.slang files)


csyonghe (approved):
Thanks for adding this test. Have you verified that the result is correct?

tgrimesnv:
Yes, as I added each function I watched the outputbuffer's values that were reported in the console window.

Issue #3628, opened 2024-02-26
__init() is not found unless decorated with __unsafeForceInlineEarly


jkwak-work:
It appears that __init() function doesn't work unless it is decorated with "[__unsafeForceInlineEarly]".
The error message looks like this when __init() has "[ForceInline]" or no decoration.
```
tests/hlsl/coopvec.slang(884): note 99999: an internal error threw an exception while working on code near this location
    CoopVector<T, COOP_VEC_SIZE> cv = CoopVector<T, COOP_VEC_SIZE>(src);
                                 ^~
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: hit unreachable code: stdlib error: __init() has no definition.
```

When "[__unsafeForceInlineEarly]" is used, I found that it generates an infinite loop for the following function,
```
    [__unsafeForceInlineEarly]
    __init(T value)
    {
        for (int i = 0; i < N; ++i)
            WriteToIndex(i, value);
    }
```
The generated code evaluates the condition, "i<N", outside of a for-loop only once and the loop became infinite.

csyonghe:
Let's just remove that assert where it is complaining the function does not have a definition.

jkwak-work:
That worked.
I will prepare a PR for the fix.

PR #3630, opened 2024-02-26
WAR for ForceInline not working issue in stdlib

jkwak-work:
Work-around for an issue #3628

This commit allows __init() functions in stdlib to be decleared without "[__unsafeForceInlineEarly]" or "[ForceInline]".

We need to find a proper solution for the issue later.


csyonghe (approved):

csyonghe:
This check is not needed in my opinion and we can just delete the code.

jkwak-work:
The second commit removed the code block as suggested.

PR #3635, opened 2024-02-27
Implement short-circuit logic operator

kaizhangNV:
Implement short-circuit evaluation for logic && and || operator.

The short-circuit behavior is only used when the operands involved are scalar and the parent function is non-differentiable.

In implementation, we define a new class 'LogicOperatorShortCircuitExpr' derived from 'OperatorExpr'. In the visitInvoke() call, we will create a new expression object 'LogicOperatorShortCircuitExpr' if the expression is logic && or ||. So that we can generate new IR code in the new visit function 'visitLogicOperatorShortCircuitExpr' to implement the short-circuit behavior.

Add new test to test the short-circuit behavior.


csyonghe:
The failure in Falcor seems legit. We may want to investigate that assertion error.

kaizhangNV:
> The failure in Falcor seems legit. We may want to investigate that assertion error.

Sure, I'm working on the GPU test failures for the GLSL intrinsic now. After that, I will look into the Falcor failure.

source/slang/slang-check-decl.cpp
```diff
@@ -1767,11 +1768,22 @@ namespace Slang
...
         if (auto initExpr = varDecl->initExpr)
         {
-            // If the variable has an explicit initial-value expression,
-            // then we simply need to check that expression and coerce
-            // it to the type of the variable.
-            //
-            initExpr = CheckTerm(initExpr);
+            if (std::getenv("RUN") &&
+                varDecl->hasModifier<HLSLStaticModifier>() &&
+                this->m_outerDecl == nullptr)
```
csyonghe, line 1773: why m_outerDecl==nullptr is needed?

csyonghe: For now let's check for both `static` and `const` modifier.

csyonghe: The setup for m_outerDecl can be done independently to whether or not it is static const.
Consider renaming it to m_outerVarDecl and always set the state.

source/slang/slang-check-expr.cpp
```diff
@@ -2376,6 +2376,13 @@ namespace Slang
         LogicOperatorShortCircuitExpr* newExpr = nullptr;
         isArgumentsChecked = false;
 
+        // If the logic expression is inside the generic parameter list, it cannot support short-circuit
+        // which will generate the ifelse branch.
+        if (m_outerGenericApp != nullptr || m_outerDecl != nullptr)
```
csyonghe, line 2381: Another case we need to detect is `int array[a&&b]`.

csyonghe: Probably instead of keeping m_outerGenericApp or m_outerDecl as state, just keep a `bool shouldShortCircuitLogicalExpr` as the state.

csyonghe (commented):
The mechanism looks correct to me.

source/slang/slang-check-expr.cpp
kaizhangNV: Yea, I'm also considering this. Because in those circumstances, there should not be nested?

source/slang/slang-check-decl.cpp
kaizhangNV: > why m_outerDecl==nullptr is needed?

No need, will remove.

kaizhangNV: > For now let's check for both `static` and `const` modifier.

I didn't find static modifer in slang-ast-modifier.h, only see ConstModifier. Should we define one?

source/slang/slang-check-expr.cpp
kaizhangNV: Update using `bool shouldShortCircuitLogicalExpr` : https://github.com/shader-slang/slang/pull/3635/commits/4759154ba75b2d59ab77232bb3471bf409143d08

Disable short-circuit in index expression: https://github.com/shader-slang/slang/pull/3635/commits/2684bc5812fc217653afe3bf3d19fadb72555b02

source/slang/slang-check-decl.cpp
kaizhangNV: Never mind, I thought we have a modifier for StaticConstModifer.
I changed to check HLSLStaticModifier and ConstModifier.: https://github.com/shader-slang/slang/pull/3635/commits/9897ff42b1d9536006cf68b0097f4044210d01a4

kaizhangNV:
Resolved most of the comments except this one:

> The setup for m_outerDecl can be done independently to whether or not it is static const. Consider renaming it to m_outerVarDecl and always set the state.

source/slang/slang-check-decl.cpp
```diff
@@ -1767,11 +1768,26 @@ namespace Slang
...
+            {
+                disableShortCircuitLogicalExpr();
+            }
+
             // If the variable has an explicit initial-value expression,
             // then we simply need to check that expression and coerce
             // it to the type of the variable.
             //
             initExpr = CheckTerm(initExpr);
+
```
csyonghe, line 1783: This logic is problematic because we could be comming into this function with shortcircuiting disabled.
Instead of manually enable and disable here, why can't we create a new Context with `m_shouldShortCircuit` set to false and do `dispatchExpr` with that context?

source/slang/slang-check-expr.cpp
```diff
@@ -1906,11 +1906,27 @@ namespace Slang
     {
         auto baseExpr = checkBaseForMemberExpr(subscriptExpr->baseExpression);
 
+        // If the base expression is a type, it means that this is an array declaration,
+        // then we should disable short-circuit in case there is logical expression in
+        // the subscript
+        auto baseType = baseExpr->type.Ptr();
```
csyonghe, line 1912: Same here, why not create a subcontext?

source/slang/slang-check-expr.cpp
```diff
@@ -2371,19 +2386,105 @@ namespace Slang
...
+
+        const ReflectClassInfo& classInfo = expr->functionExpr->getClassInfo();
+        const ASTNodeType astType = ASTNodeType(classInfo.m_classId);
+        if (ASTNodeType::VarExpr == astType)
+        {
+            VarExpr const* varExpr = as<VarExpr const>(expr->functionExpr);
+            if ((varExpr->name->text == "&&") || (varExpr->name->text == "||"))
+            {
+                // Don't use short-circuiting in differentiable function
+                if (getParentDifferentiableAttribute())
```
csyonghe, line 2409: Why can't we use short-circuiting in differentiable function?

source/slang/slang-check-impl.h
```diff
@@ -943,6 +955,13 @@ namespace Slang
         ASTBuilder* m_astBuilder = nullptr;
 
         Scope* m_outerScope = nullptr;
+
+        // By default, we will support short-circuit evaluation for the logic expression.
+        // However, there are few exceptions where we will disable it:
+        // 1. the logic expression is inside the generic parameter list.
+        // 2. the logic expression is in the init expression of a static const variable.
+        // 3. the logic expression is in the index of array.
```
csyonghe, line 963: the expression is a size of an array type, or we can just remove the comment because I assume no addiitional work is required to ensure this is the case?

source/slang/slang-check-overload.cpp
```diff
@@ -2113,6 +2113,11 @@ namespace Slang
     Expr* SemanticsExprVisitor::visitGenericAppExpr(GenericAppExpr* genericAppExpr)
     {
         // Start by checking the base expression and arguments.
+
+        // Disable the short-circuiting logic expression when the experssion is in
+        // the generic parameter.
+        disableShortCircuitLogicalExpr();
```
csyonghe, line 2119: use subcontext.

source/slang/slang-check-decl.cpp
kaizhangNV: There is nothing preventing us from using subContext and dispatchExpr actually. I'm doing the manually enable and disable because it seems simpler without introducing more level of recursion.

> because we could be coming into this function with shortcircuiting disabled
Isn't it our goal here? We do want to have the shortcircuiting disabled here right?

I can change to subContext anyway. But I'm not quite understand the disadvantage of surrounding the `CheckTerm` by enable/disable pair. Can you please explain me more? Thanks.

source/slang/slang-check-impl.h
kaizhangNV: update the comment.

source/slang/slang-check-overload.cpp
kaizhangNV: Done

source/slang/slang-check-decl.cpp
csyonghe: Because when you set a state such as m_shouldShortCircuit, you need to save its current value to a stack, and restore it after you've done with the sub-tree. The current value of `m_shortCircuit` is not necessarily `true` or `false` and we want to be robust. Instead of using an explicit stack data-structure, we just use the callstack to hold the previous state value. The subcontext mechanism accomplishes this goal.

source/slang/slang-check-expr.cpp
```diff
@@ -2371,19 +2384,99 @@ namespace Slang
...
+        isArgumentsChecked = false;
+
+        // If the logic expression is inside the generic parameter list, it cannot support short-circuit
+        // which will generate the ifelse branch.
+        if (!m_shouldShortCircuitLogicExpr)
+        {
+            return nullptr;
+        }
+
+        const ReflectClassInfo& classInfo = expr->functionExpr->getClassInfo();
```
csyonghe, line 2399: `expr->functionExpr` might be nullptr. Always check for nullptr in semantics visitor.
Also it should be `auto astType = expr->functionExpr->astNodeType;`.

csyonghe: Usually we don't directly compare astType, and we just write:
`if (as<VarExpr>(expr->functionExpr)`
This automatically takes care of the case of expr->functionExpr being null.

source/slang/slang-check-expr.cpp
```diff
@@ -2371,19 +2384,99 @@ namespace Slang
...
+                newExpr = m_astBuilder->create<LogicOperatorShortCircuitExpr>();
+                if (varExpr->name->text == "&&")
+                {
+                   newExpr->flavor = LogicOperatorShortCircuitExpr::Flavor::And;
+                }
+                else
+                {
+                   newExpr->flavor = LogicOperatorShortCircuitExpr::Flavor::Or;
+                }
+                newExpr->loc = expr->loc;
```
csyonghe, line 2441: also need:
newExpr->scope = expr->scope;

source/slang/slang-check-expr.cpp
```diff
@@ -2371,19 +2384,99 @@ namespace Slang
...
     {
         // check the base expression first
         if (!expr->originalFunctionExpr)
             expr->originalFunctionExpr = expr->functionExpr;
+
+        // if the expression is '&&' or '||', we will convert it
+        // to use short-circuit evaluation.
+        bool isArgumentsChecked = false;
+        Expr* newExpr = convertToLogicOperatorExpr(expr, isArgumentsChecked);
+        if (newExpr != nullptr)
```
csyonghe, line 2461: Can simplify into:
```
if (auto newExpr = convertToLogicOperatorExpr(expr, isArgumentsChecked))
    return newExpr;
```

source/slang/slang-check-expr.cpp
kaizhangNV: Done removing this.

source/slang/slang-check-expr.cpp
```diff
@@ -1906,6 +1906,20 @@ namespace Slang
     {
         auto baseExpr = checkBaseForMemberExpr(subscriptExpr->baseExpression);
 
+        // If the base expression is a type, it means that this is an array declaration,
+        // then we should disable short-circuit in case there is logical expression in
+        // the subscript
+        auto baseType = baseExpr->type.Ptr();
+        auto baseTypeType = as<TypeType>(baseType);
```
csyonghe, line 1913: Reading this function again, now I understand why you want to avoid the extra recursion. Another simpler way to do this is something like:
```
auto subVisitor = baseTypeType?SemanticsVisitor(disableShortCircuitLogicalExpr()) : *this;

for (auto& arg : subscriptExpr->indexExprs)
   arg = subVisitor.CheckTerm(arg);
```

source/slang/slang-check-expr.cpp
kaizhangNV: Thanks for the tips, updated.

source/slang/slang-check-expr.cpp
kaizhangNV: Done

source/slang/slang-check-expr.cpp
kaizhangNV: Done

csyonghe: I see, that's fine then.

source/slang/slang-check-expr.cpp
```diff
@@ -2371,19 +2378,93 @@ namespace Slang
...
     Expr* SemanticsExprVisitor::visitInvokeExpr(InvokeExpr* expr)
     {
         // check the base expression first
         if (!expr->originalFunctionExpr)
             expr->originalFunctionExpr = expr->functionExpr;
+
+        // if the expression is '&&' or '||', we will convert it
+        // to use short-circuit evaluation.
+        bool isArgumentsChecked = false;
+        if (auto newExpr = convertToLogicOperatorExpr(expr, isArgumentsChecked))
```
csyonghe, line 2451: Can we check `expr->arguments` first, and then call `convertToLogicOperatorExpr` to see if we can convert? So we don't have to maintain `isArgumentsChecked` and handle different cases below.

kaizhangNV: updated.

csyonghe (approved):
Looks good!

Cazadorro:
I know this is a decently recent change, but I believe the documentation still says that [there's no short circuiting logic scalars](https://shader-slang.com/slang/user-guide/conventional-features.html#Expressions).  This merge means the documentation is wrong correct?

csyonghe:
You are right, the documentation should be updated.

Issue #3559, opened 2024-02-07
Add support for bitfields


savage309:
Using bit operations is a work around, but having bit fields will make it easier to pack data.

natevm:
For what it's worth, HLSL describes their 2021 bitfields at the bottom of dev blog here: 

https://devblogs.microsoft.com/directx/announcing-hlsl-2021/

One application where I'd really like to have bitfields for is to directly support constructing certain Vulkan structures on the device like `VkAccelerationStructureInstanceKHR`, since then I can fill out my top level acceleration structures much faster. 

![image](https://github.com/shader-slang/slang/assets/13104331/ee2058b1-d8be-4b93-80d4-d69dffed477f)

Thanks for looking into this!

PR #3639, opened 2024-02-27
Add support for bitfields

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/3559


source/slang/slang-check-decl.cpp
```diff
@@ -8943,6 +8980,146 @@ namespace Slang
...
+                dispatchSomeBitPackedMembers();
+
+            // If this member wouldn't fit into the current group, dispatch
+            // everything so far;
+            if(totalWidth + thisFieldWidth > std::max(thisFieldTypeWidth, backingWidth))
+                dispatchSomeBitPackedMembers();
+
+            // Add this member to the group,
+            // Grow the backing width if necessary
+            backingWidth = std::max(thisFieldTypeWidth, backingWidth);
```
csyonghe, line 9114: For consistency with C++, I think the behavior should be if the backing with of the currently open backing field is different from the type of this field, then we start a new field.

tests/language-feature/bitfield/generic.slang
```diff
@@ -0,0 +1,29 @@
...
+// CHECK-NEXT: 0
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<uint> outputBuffer;
+
+struct S<T> {
+    int foo : 8;
+    uint bar : 24;
+    T baz;
+};
```
jkwak-work, line 15: I think we should be sensitive about which buts will be used first. The first variable could use the most significant its or least significant bits. Because this behavior is not standardized in C/C++, gcc and vsc provides an option to change it. By default gcc and vsc have different behavior. This issue had been a big pain during the playstation 3 generation because Sony was using gcc while xbox was using visual studio.

tests/language-feature/bitfield/zero-width.slang
```diff
@@ -0,0 +1,28 @@
...
+// CHECK-NEXT: 4567
+// CHECK-NEXT: 1
+// CHECK-NEXT: 0
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<uint> outputBuffer;
+
+struct S {
+    int foo : 8;
+    int breaker : 0;
```
csyonghe, line 13: What's the semantics of a 0bit field?

csyonghe (approved):
Looks good!

csyonghe:
Looks like cuda doesn't like 0-wide bit fields.

tests/language-feature/bitfield/generic.slang
expipiplus1: I wonder if even this could be done as an annotation on the fields themselves..

source/slang/slang-check-decl.cpp
expipiplus1: I think that this is implementation defined for C and C++, and at least gcc and clang are happy to push multiple types together into one field

tests/language-feature/bitfield/zero-width.slang
expipiplus1: It only has one value (0), and indicates that the next bits shouldn't be part of an existing field. I dislike very much how it's this sort of magic inconsistent thing going on, but it's consistent with how my C compiler does things at least...

tests/language-feature/bitfield/generic.slang
expipiplus1: https://github.com/shader-slang/slang/issues/3646

source/slang/slang-check-decl.cpp
expipiplus1: https://github.com/shader-slang/slang/issues/3646

source/slang/slang-ir-bit-field-accessors.cpp
```diff
@@ -0,0 +1,160 @@
+#include "slang-ir-bit-field-accessors.h"
+
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+static IRInst* maybeUnwrapGeneric(IRInst* inst)
```
csyonghe, line 8: There is already a `getResolvedInstForDecoration` function that unwrapps all generic/specialize insts and returns the inner thing being specialized.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -8872,6 +8873,25 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         }
     }
 
+    void addBitFieldAccessorDecorations(IRInst* irFunc, Decl* decl)
+    {
+        // If this is an accessor under a property we can move the bitfield
+        // modifiers on the property to the accessor function.
+        if(as<AccessorDecl>(decl) && as<PropertyDecl>(decl->parentDecl))
```
csyonghe, line 8880: The `as<PropertyDecl>` test seems not needed.

csyonghe (approved):

source/slang/slang-ir-bit-field-accessors.cpp
expipiplus1: Ah I didn't see this before merging, will address tomorrow.

csyonghe: No worries, it's a very minor issue.

PR #3652, opened 2024-03-01
Fix build with SLANG_DISABLE_STDLIB_SOURCE

expipiplus1:


csyonghe (approved):

csyonghe:
I think the right thing to do here is to make glsl module as a standard builtin module just like `core`. Right now we are being very hacky by just storing a string of the glsl module and load from it on demand. What we really should be doing is to serialize and embed both core and glsl, and deserialize the embeded glsl module on demand.

csyonghe:
Closing this and we should work on https://github.com/shader-slang/slang/issues/3665 instead.

PR #3675, opened 2024-03-05
Implement raytracing extension(s); resolves #3560 for GLSL & SPIR-V targets

### LINKED ISSUE #3560: Implement ray tracing extensions for GLSL in glsl.meta.slang and bring the parity to HLSL
This task is a subtask from the other issue #3362 , which is a big task.
This issue is to cover a smaller portion of it so that we can work with other people together.

We have "ray-tracing" functionality implemented for HLSL but it is not implemented for GLSL.
The goal of this task is to bring the parity to GLSL as much as slangc already does for HLSL.

The description for the "ray_query" extension can be found from [here](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_ray_query.txt)

The description for the "shader_invocation_reoder" extension can be found from [here](https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_shader_invocation_reorder.txt)

The description for the "ray_tracing_motion_blur" extension can be found from [here](https://github.com/KhronosGroup/GLSL/blob/main/extensions/nv/GLSL_NV_ray_tracing_motion_blur.txt)

ArielG-NV:
The following PR implements raytracing extension(s) (GLSL_EXT_ray_tracing, GLSL_EXT_ray_query, GLSL_NV_shader_invocation_reorder & GLSL_NV_ray_tracing_motion_blur); for GLSL & SPIR-V targets only: fully implements all functions, fully implements all built-in variables, fully implements all syntax; partially resolves #3560

Raytracing Extensions can be found here: https://github.com/KhronosGroup/GLSL/blob/main/extensions/

notes of worth:
* #3560 is only partially solved since HLSL support was removed
* HLSL support was implemented (added and then removed in an earlier commit for future reference), but was removed at the end due to the implementation being obtuse and likely hard to maintain 
* SPIR-V required direct referencing of raytracing objects, as a result, __rayPayloadFromLocation, __rayAttributeFromLocation, and __rayCallableFromLocation, all take in a integer literal and use this to add a placeholder to fill-in with an actual layout(location) bound object.
* tests only test compiling and checks for opcodes emitted


source/slang/glsl.meta.slang
```diff
@@ -2787,47 +2787,1573 @@ public vec4 shadow2DProjLod(sampler2DShadow sampler, vec4 coord, float lod)
...
+    __target_switch
+    {
+    case glsl:
+        requireGLSLExtForRayTracingBuiltin();
+    case spirv:
+        return;
+    }
+}
+
+[require(glsl)]
```
csyonghe, line 2825: Please take a look at the capability definitions in `slang-capabilities.capdef` and properly mark these intrinsics with capaiblity requirements.

For example, `gl_LaunchIDNV` should be:
[require(spirv, raytracing)]
[require(glsl, raytracing)]

source/slang/hlsl.meta.slang
```diff
@@ -9751,9 +9752,11 @@ bool __reportIntersection(float tHit, uint hitKind)
 {
     __target_switch
     {
-    case _GL_EXT_ray_tracing: __intrinsic_asm "reportIntersectionEXT";
+    case _GL_EXT_ray_tracing:
+    case glsl: __intrinsic_asm "reportIntersectionEXT";
```
csyonghe, line 9756: You shouldn't need `case glsl` if the requirements are properly marked in glsl.meta.slang

source/slang/hlsl.meta.slang
```diff
@@ -10619,6 +10670,377 @@ struct RayQuery <let rayFlagsGeneric : RAY_FLAG = RAY_FLAG_NONE>
         }
     }
 
+///missing hlsl equivlent; only implemented for glsl & spirv
+    __glsl_extension(GL_EXT_ray_query)
+    __glsl_version(460)
+    [__NoSideEffect]
+    int CandidateRayInstanceCustomIndex()
```
csyonghe, line 10677: This is good. Let's mark them with proper capabilities too.

source/slang/slang-ir.cpp
```diff
@@ -4325,11 +4327,114 @@ namespace Slang
...
+    {
+        //instead of forcing order of code, a single pass if needed can be used
+        //likley forcing order is a faster and better idea
+
+        if(!checkedForRayVariables)
+        {
+            checkedForRayVariables = true;
+            for(auto i : getGlobalInsts())
+            {
+                if(auto rayPayload = i->findDecoration<IRVulkanRayPayloadDecoration>())
```
csyonghe, line 4341: `findDecoration` is O(n), so instead of doing a bunch of findDecorations, iterate over all the decorations and check the type:
```
for (auto decor : i->getDecorations())
{
        if (auto rayPayload = as<IRVulkanRayPayloadDecoration>(decor))
               ...
        else if (...)
              ....
}
```

source/slang/slang-ir.cpp
```diff
@@ -4325,11 +4327,114 @@ namespace Slang
...
+        }
+    }
+
+    IRInst* IRModule::getRayVariableFromLocation(IRInst* payloadVariable, Slang::IROp op, DiagnosticSink* sink)
+    {
+        IRBuilder builder(payloadVariable);
+        IRInst** varLayoutPointsTo = nullptr;
+        int intLitValue = -1;
+        IRIntLit* intLit = as<IRIntLit>(payloadVariable);
+        if(intLit){
```
csyonghe, line 4372: coding style: { in its own line.

source/slang/slang-ir.cpp
```diff
@@ -4325,11 +4327,114 @@ namespace Slang
...
+            {
+                varLayoutPointsTo = this->getRayAttributeVariableFromLocation(intLitValue);
+            }
+            else if (kIROp_SPIRVAsmOperandRayCallableFromLocation == op)
+            {
+                SLANG_ASSERT(kIROp_SPIRVAsmOperandRayCallableFromLocation == op); //final case
+                varLayoutPointsTo = this->getRayCallableVariableFromLocation(intLitValue);
+            }
+        }
+        else{
```
csyonghe, line 4388: {

source/slang/slang-ir.cpp
```diff
@@ -4325,11 +4327,114 @@ namespace Slang
...
+        }
+
+        IRInst* resultVariable;
+        if (!varLayoutPointsTo)
+        {
+            // if somehow the location tied variable is missing and an error was not thrown by the compiler
+            resultVariable = builder.getIntValue(builder.getIntType(), 0);
+            sink->diagnose(payloadVariable, Diagnostics::expectedRayTracingPayloadObjectAtLocationButMissing, intLitValue);
+        }
+        else {
```
csyonghe, line 4399: {

source/slang/slang-ir.cpp
```diff
@@ -4325,11 +4327,114 @@ namespace Slang
...
+            resultVariable = *varLayoutPointsTo;
+        }
+        return resultVariable;
+    }
+
+    void IRModule::storeLocationToRayPayloadVariable(int location, IRInst* inst)
+    {
+        // Try to remove duplicates since IR will try to remove/re-add globals 
+        // sometimes to move them around, this case needs to be handled
+        m_RayLocationToPayloads.set(location, inst);
```
csyonghe, line 4409: We shouldn't add a new field to `IRModule` since we don't need this for more than the span of a single pass.
Instead we should be able to reconstruct the dictionary from all the decorations.

source/slang/slang-ir.cpp
```diff
@@ -5854,11 +5959,46 @@ namespace Slang
...
+            inst
+        );
+        addInst(i);
+        return i;
+    }
     IRSPIRVAsmOperand* IRBuilder::emitSPIRVAsmOperandId(IRInst* inst)
     {
         SLANG_ASSERT(as<IRSPIRVAsm>(m_insertLoc.getParent()));
-        const auto i = createInst<IRSPIRVAsmOperand>(
+        auto i = createInst<IRSPIRVAsmOperand>(
```
csyonghe, line 6001: Why are these changes needed?
Let's try to minimize our changes.

source/slang/slang-ir.h
```diff
@@ -2322,6 +2323,25 @@ struct IRModule : RefObject
     {
         return m_containerPool;
     }
+
+    // Raytracing variables/attributes managment functions
+    bool checkedForRayVariables = false;
```
csyonghe, line 2328: These functions should belong to a separate pass instead of in `IRModule`.

source/slang/slang-ir.cpp
ArielG-NV: when reviewing the code, it was noticed that the ptr was assigned as const, but not used as const (value may change); then the ptr is returned, casting away the const

there is no need for const (and likely a bad idea to set it as const) if the ptr won't be treated as a const

source/slang/slang-ir.h
ArielG-NV: There is a small problem with moving the functions+var's of this pass; even though for-now it was decided to drop HLSL support for GLSL raytracing functions, if it ever is brought back (or CUDA, or other), the exact same function & pass will need to be shared between all of the API's, in which I don't exactly know where to store the data which is needed for the final pass (code gen/emit'ting) such that all can access the data.

the options are to:
1. keep it where it is for the future
or
2. GLSL will only go to SPIR-V and GLSL, therefore move the functions+var's inside the SPIR-V emitter

csyonghe: My understanding is that there is no need to store that data explicitly on the IR, since you can construct the dictionary from the decorations, no?

csyonghe: You can write a separate function, called:
```
Dictionary<int, IRInst*> createMapFromVulkanLocationToPayloadVar(IRModule*);
```
And define it in a separate header like `slang-ir-vulkan-location-to-payload.h` and implement it in a cpp.

Then anything that needs this can just include the header file and call the function to get the map constructed from IR.

ArielG-NV: 👍I will move the logic outside the Module and cache the data in the code pass structure that needs the data (SourceEmitterBase)

csyonghe: Sounds good!

csyonghe: The general rule is to do a little translation work as possible at emit. Emit should stay as direct as possible from IR to target code, because it is far easier to maintain separate small and well-defined passes than to maintain a big emit pass that does 100 different things, especially when some of those things may have consequences on each other.

Doing a pass scanning through all the opcodes is fine.

csyonghe (dismissed):

source/slang/slang-ir-early-intrinsic-simplification.cpp
```diff
@@ -0,0 +1,169 @@
+// slang-ir-ssa-simplification.cpp
+#include "slang-ir-early-intrinsic-simplification.h"
+#include "slang-ir.h"
+#include "../core/slang-performance-profiler.h"
+#include "slang-ir-util.h"
+
+namespace Slang
+{
+    // ONLY should be used in this compilation unit
+    struct cacheOfDataToReplaceOps
```
csyonghe, line 10: typenames are BigCamel.

source/slang/slang-ir-early-intrinsic-simplification.cpp
```diff
@@ -0,0 +1,169 @@
+// slang-ir-ssa-simplification.cpp
```
csyonghe, line 1: inconsistent comment.

source/slang/slang-ir-early-intrinsic-simplification.h
```diff
@@ -0,0 +1,15 @@
+// slang-ir-ssa-simplification.h
```
csyonghe, line 1: comment.

source/slang/slang-ir-early-intrinsic-simplification.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    void recurseAllOpsToReplace(cacheOfDataToReplaceOps* cache)
+    {
+        for (auto func : cache->funcsToSearch)
+        {
+            recurseInFuncForOpsToReplace(func, cache);
+        }
+    }
+
+    void simplifyIntrinsicsEarly(TargetProgram* target, IRModule* module, DiagnosticSink* sink)
```
csyonghe, line 158: This needs a more specific name, e.g. `replaceVulkanLocationWithGlobalVar`.

source/slang/slang-ir-early-intrinsic-simplification.cpp
```diff
@@ -0,0 +1,169 @@
+// slang-ir-ssa-simplification.cpp
+#include "slang-ir-early-intrinsic-simplification.h"
```
csyonghe, line 2: Let's use a more specific file name, since the file is currently doing a specific thing.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -2997,7 +2999,13 @@ void legalizeEntryPointsForGLSL(
         legalizeEntryPointForGLSL(session, module, func, context, glslExtensionTracker);
     }
 
-    assignRayPayloadHitObjectAttributeLocations(module);
+    // GLSL should NEVER mangle user declared ray locations; else user setting a location to reference  
+    // (GLSL references locations by a constexpr number) could be invalid. Furthermore, we need to fetch
+    // types and what location variable is where when we resolve locations to translate GLSL->SPIR-V/HLSL
+    if (!context->getTargetProgram()->getOptionSet().getBoolOption(CompilerOptionName::AllowGLSL))
```
csyonghe, line 3005: We will deprecate the AllowGLSL option. Maybe just remove the check.

source/slang/slang-ir-link.cpp
```diff
@@ -1606,6 +1606,40 @@ LinkedIR linkIR(
 
     context->builder->setInsertInto(context->getModule()->getModuleInst());
 
+    // Raypayload/related objects found under GLSL are likley never referenced
+    // this is because internally they are used by slang functions; these emit later in code gen
+    // as a result we emit them now
+    // we also emit these early since if these are not at the top of scope, that could be deadly to a compile
+    if (targetProgram->getOptionSet().getBoolOption(CompilerOptionName::AllowGLSL))
```
csyonghe, line 1613: Instead of doing this in linking, can we just mark the global variables of those payload as [KeepAlive] and let linking copy them in?

source/slang/hlsl.meta.slang
```diff
@@ -9725,7 +9724,8 @@ void TraceMotionRay(
         let tmax = Ray.TMax;
 
         p = Payload;
-        spirv_asm {
+        spirv_asm
```
csyonghe, line 9727: This function and others in hlsl.meta.slang, such as TraceRay and TraceMotionRay, needs [require(spirv, raytracing)]

source/slang/hlsl.meta.slang
```diff
@@ -9555,7 +9553,8 @@ void CallShader(uint shaderIndex, inout Payload payload)
             static Payload p;
 
             p = payload;
-            spirv_asm {
+            spirv_asm 
+            {
```
csyonghe, line 9557: need capability declaration for `CallShader`.

source/slang/slang-ir-link.cpp
ArielG-NV: I tested and looked around the code: 
KeepAliveDecoration allows the var to survive optimizations, but the initial linking will still never copy the raytracing object in GLSL syntax since the var will likley never be referenced.

we don't need the `addInst` though since once cloned the inst will be in the new context and will add later anyways; I think we should still add to keep the code working more like `insertGlobalValueSymbols`

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -2993,14 +2993,6 @@ void legalizeEntryPointsForGLSL(
     {
         legalizeEntryPointForGLSL(session, module, func, context, glslExtensionTracker);
     }
-
-    // GLSL should NEVER mangle user declared ray locations; else user setting a location to reference  
-    // (GLSL references locations by a constexpr number) could be invalid. Furthermore, we need to fetch
-    // types and what location variable is where when we resolve locations to translate GLSL->SPIR-V/HLSL
-    if (!context->getTargetProgram()->getOptionSet().getBoolOption(CompilerOptionName::AllowGLSL)) 
-    {
-        assignRayPayloadHitObjectAttributeLocations(module);
```
csyonghe: We don't need this anymore?

ArielG-NV: we need it, I was just confused on why things were working without it (turns out, visual studio was just not working properly, so I rebuilt everything); I will adjust the call though

source/slang/slang-ir-link.cpp
csyonghe: You also need a `kIROp_HLSLExportDecoration` on the varible to make it survive linking regardless of whether it has uses.

csyonghe: See slang-ir-link.cpp:1673 for reference.

ArielG-NV: I don't need to clone for the attribute to live, but I realized the emitted object will be at the bottom of the program if not emitted early

ArielG-NV: I checked, and this would be a fatal program error

ArielG-NV: we need to early emit the raytracing objects to ensure they are at top of scope above everything else (including stdlib functions)

csyonghe: The IR is designed to allow arbitrary odering at the global scope level, so I wonder why the implementation requires those objects to be on top, which seems worrying.

ArielG-NV: raytracing intrinsic's resolve at just before emitting stage (used to be during emitting stage) 

due to this since we don't reference raytracing objects normally in GLSL, they are placed at the bottom 

we can change this behavior reliably by making all raytracing objects 'dependent' on all raytracing intrinsic I suppose

csyonghe: This means that we need to either
1) update the emit logic in emit-glsl to always emit payload variables before anything else;  or 
2) add some decorations to establish that dependency, such as decorate the call to the `traceRay` function with a reference to the payload variable.

We should not write code based on assumption that things in global scope is in any particular order.

csyonghe: Btw, why is it a problem if it is declared at the bottom? Does GLSL not work in that case?

ArielG-NV: that's what the code currently does, it ensures the ray objects are emitted directly after all hashed strings, at top of scope

ArielG-NV: > Btw, why is it a problem if it is declared at the bottom? Does GLSL not work in that case?

likely does not work: https://shader-playground.timjones.io/788a40ca09027e0f1be0fddefe447dee

csyonghe: We can't assume that the order of global insts are preserved across passes, so just placing them at the beginning of the module at linking does not guarantee they will continue to be at top when we get to emit.
That's why I am saying we should not make assumptions on the ordering of global insts.

If you want to implement 1), you need to do so in emit-glsl.cpp to emit those global variables first, then emit entrypoint.

csyonghe: You example is different from what we are talking about here. The issue we are discussing is that there is no direct reference to that global variable in the glsl, and it is accessed indirectly through an index in code. So the GLSL should be if the variable itself is defined last.

ArielG-NV: I am changing the example to reflect as such

ArielG-NV: does not work, order matters

ArielG-NV: > We can't assume that the order of global insts are preserved across passes, so just placing them at the beginning of the module at linking does not guarantee they will continue to be at top when we get to emit. That's why I am saying we should not make assumptions on the ordering of global insts.
> 
> If you want to implement 1), you need to do so in emit-glsl.cpp to emit those global variables first, then emit entrypoint.

I will do this (this is what the code originally did, so I can likley just use that code)

csyonghe: OK, then we need to modify `slang-emit-c-like.cpp:: CLIkeSourceEmitter::computeEmitActions()` to visit all ray tracing globals first.

ArielG-NV: we would also need to modify the SPIR-V emitter(?)

csyonghe: I don't think the SPIRV needs to change. Should be fine to have out of order spirv. Will need to see if that passes validation.

csyonghe: But in case SPIRV does need it, the logic is the same there. You can add calls to `ensureInst` on those globals in `emitSPIRVFromIR()`.

csyonghe: We already have some logic that to ensure certain types of insts are emitted before the others.

ArielG-NV: I will implement this shortly 👍, I will just adjust slang-ir-glsl-legalize first so we can use HLSL+GLSL syntax for raytracing before hand (I think it would be quite nice to have HLSL syntax access for if adding to legacy GLSL raytracing code)

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -2135,6 +2135,8 @@ void buildEntryPointReferenceGraph(SPIRVEmitSharedContext* context, IRModule* mo
         if (globalInst->getOp() == kIROp_Func && globalInst->findDecoration<IREntryPointDecoration>())
         {
             visit(as<IRFunc>(globalInst), globalInst);
+
```
csyonghe, line 2123: Left-over comment?

source/slang/slang-ir-early-raytracing-intrinsic-simplification.cpp
```diff
@@ -132,11 +132,12 @@ namespace Slang
                 {
                     auto op = i->getOperand(0);
                     IRInst* globalVar = cache->getRayVariableFromLocation(op, i->getOp());
-                    auto builder = IRBuilder(i);
-                    builder.setInsertBefore(i);
+                    auto builder = IRBuilder(i->getParent());
+                    builder.setInsertInto(i->getParent());
```
csyonghe, line 136: You want to insert before `i`, right?,  The code after this change will insert spirvASM at the end of the parent.

source/slang/slang-ir-link.cpp
csyonghe: Continue this thread, if we make sure to also emit `HLSLExport` decoration when we lower IR for the global ray tracing object, it should just be cloned during linking without the explicit loop here. If this is not the case we need to figure out why.

source/slang/slang-ir-link.cpp
```diff
@@ -4,6 +4,7 @@
 #include "slang-capability.h"
 #include "slang-ir.h"
 #include "slang-ir-insts.h"
+#include "slang-legalize-types.h"
```
csyonghe, line 7: Not needed anymore?

source/slang/slang-ir-link.cpp
```diff
@@ -1739,6 +1740,7 @@ LinkedIR linkIR(
 
     context->builder->setInsertInto(context->getModule()->getModuleInst());
 
+
```
csyonghe, line 1743: Not needed.

source/slang/slang-legalize-types.cpp
```diff
@@ -166,6 +166,23 @@ RefPtr<TupleLegalElementWrappingObj> LegalElementWrapping::getTuple() const
 
 //
 
+bool isRaytracingObject(IRInst* inst)
```
csyonghe, line 169: Still needed?

ArielG-NV: still needed

ArielG-NV: I can remove the call though

csyonghe (approved):

Issue #3538, opened 2024-02-02
CUDA writer generates incorrect code for swizzles on the left-hand side of an assignment operator


NBickford-NV:
Hi Slang team! Here's an interesting bug I just ran into: when compiling the following code for CUDA using `slangc -target cuda -entry computeMain -stage compute repro.slang`:

```hlsl
#define T uint4

RWStructuredBuffer<T> outputBuffer;

[shader("compute")]
[numthreads(1, 1, 1)]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
  T foo = T(dispatchThreadID.xyz, 0);
  //----------------------------------------------------------------------------
  // Note that the output here assigns to the .xw field of a T type on the left-
  // hand sign in CUDA, which causes compilation to fail:
  foo.xw = foo.zw + foo.yz;
  //----------------------------------------------------------------------------
  outputBuffer[dispatchThreadID.x] = foo;
}
```

the `foo.xw = foo.zw + foo.yz;` line (where `foo` is a 4-element vector) generates code without a swizzle on the right-hand side, but not the left:

```cuda
#line 15 "repro.slang"
struct GlobalParams_0
{
    RWStructuredBuffer<uint4 > outputBuffer_0;
};


#line 15
extern "C" __constant__ GlobalParams_0 SLANG_globalParams;
#define globalParams_0 (&SLANG_globalParams)

#line 7
extern "C" __global__ void computeMain()
{

#line 7
    uint3  _S1 = blockIdx * blockDim + threadIdx;

    int3  _S2 = int3 {_S1.x, _S1.y, _S1.z};

#line 9
    uint3  _S3 = make_uint3 ((uint)_S2.x, (uint)_S2.y, (uint)_S2.z);

#line 9
    uint4  _S4 = make_uint4 (_S3.x, _S3.y, _S3.z, 0U);

#line 9
    uint4  foo_0 = _S4;



    foo_0.xw = uint2 {_S4.z, _S4.w} + uint2 {_S4.y, _S4.z}; // <- Here

    *(&(globalParams_0->outputBuffer_0[uint(_S1.x)])) = foo_0;
    return;
}
```

This causes compilation to fail, since CUDA's `uint4` doesn't have support for swizzles.

This seems to be specific to the CUDA backend, since the GLSL output uses swizzles on both sides:

```glsl
#version 450
layout(row_major) uniform;
layout(row_major) buffer;

#line 3 0
[...]

PR #3691, opened 2024-03-06
Fix the cuda left-hand swizzle issue (#3538)

kaizhangNV:


kaizhangNV:
The change basically only generate
`foo.xy = vec2 `
to

`foo.x = vec2.x`
`foo.y = vec2.y`

for C++ and CUDA target.

source/slang/slang-emit-c-like.cpp
```diff
@@ -352,6 +352,50 @@ void CLikeSourceEmitter::_emitType(IRType* type, DeclaratorInfo* declarator)
...
 
+void CLikeSourceEmitter::_emitSwizzleStorePerElement(IRInst* inst)
+{
+    auto subscriptOuter = getInfo(EmitOp::General);
+    auto subscriptPrec = getInfo(EmitOp::Postfix);
+
+    // store the right-handside expression into a temporary variable first, if
+    // the right-handside expression is already a variable, we can skip this.
+    auto ii = cast<IRSwizzledStore>(inst);
+    String tempIdentifierName = "tmp";
```
csyonghe, line 363: Don't do this yourself, instead, just make sure we don't fold `ii->getSource()` by handling a case for IRSwizzledStore use after line 1496 in slang-emit-c-like.cpp (inside CLikeSourceEmitter::shouldFoldInstIntoUseSites):

```
if (auto swizzledStore = as<IRSwizzledStore>(inst))
{
      if (isCPUTarget(getTargetReq()) || isCUDATarget(getTargetReq()))
           return false;
}
```

csyonghe: This should make sure that we will always emit the source operand as a standalone assignment statement before we reach the swizzled store inst. So we can simply do a loop and call emitOperand(ii->getSource()) in the loop without duplicating the logic that computes `source`.

source/slang/slang-emit-c-like.cpp
```diff
@@ -2760,32 +2804,44 @@ void CLikeSourceEmitter::_emitInst(IRInst* inst)
...
+            switch (this->getTarget())
+            {
+                case CodeGenTarget::CPPSource:
+                case CodeGenTarget::HostCPPSource:
+                case CodeGenTarget::CUDASource:
+                    // cpp and cuda target don't support swizzle on the left handside, so we
+                    // have to assign the element one by one.
+                    _emitSwizzleStorePerElement(inst);
+                    break;
+                default:
```
csyonghe, line 2816: You can `return` for the three cases and keep the remaining logic outside of the `switch` to reduce nesting levels.
You can also just call `isCPUTarget(getTargetReq()) || isCUDATarget(getTargetReq())`.

csyonghe (approved):

PR #3703, opened 2024-03-07
WIP Drop Premake

expipiplus1:
TODO:

- Replace the release action workflows
- Falcor tests
- Test as a submodule
- Archive slang-glslang and slang-llvm repos
- Remove premake generated files and the cmake code to clean them
- Update build instructions
- Document CI setup for CMake


csyonghe:
We need to update the build instructions document along with this change.

kaizhangNV:
looks like nothing change on : release-*.yml files. How is that going to work? Because those workflows also use premake to build the release binary?

PR #3717, opened 2024-03-08
Add nightly vkcts workflow

dzysk:
Adding a workflow to run vkcts nightly in slang mode using the expected list of passing tests.


csyonghe:
Is it possible to run CTS in more than one thread? We should use something like 12 threads.
Need to change back to schedule before merging.

dzysk:
There seems to be a limitation in slang right now that is preventing multi-threaded execution of the tests. However, I opened an issue to address that limitation: https://github.com/shader-slang/slang/issues/3739 When the limitation is resolved then the workflow can be updated to run the tests multi-threaded.

.github/workflows/vk-gl-cts-nightly.yml
```diff
@@ -0,0 +1,56 @@
...
+        .\premake.bat vs2019 --arch=x64 --deps=true --no-progress=true --enable-cuda=true
+        
+        .\make-slang-tag-version.bat
+        
+        MSBuild.exe slang.sln -v:m -m -property:Configuration=Release -property:Platform=x64 -property:WindowsTargetPlatformVersion=10.0.19041.0 -maxcpucount:12
+    - uses: robinraju/release-downloader@v1.7
+      with:
+        latest: true
+        repository: "shader-slang/VK-GL-CTS"
+        fileName: "VK-GL-CTS_WithSlang-0.0.3-win64.zip"
```
csyonghe, line 34: The filename shouldn't include a version number of it since it will never be accurate?

csyonghe (approved):

PR #3750, opened 2024-03-12
Fix derivative implementation of `clamp`.

### LINKED ISSUE #3749: Incorrect forward derivatives of built-in clamp function
Issue encountered and tested in slang **v2024.1.1** and **v2023.5.1**.

Default clamp implementation returns zero derivatives for this kind of code:
```hlsl

[ForwardDifferentiable]
float3 func(...) 
{
    float3 color(...);
    // return clamp_(result, 0.0f, 1.0f); // This works
    return clamp(result, 0.0f, 1.0f); // This does not work
}
```

A workaround is a simple clamp implementation.
```hlsl
[Differentiable]
[PreferRecompute]
float clamp_(float val, float val_min, float val_max)
{
    if (val > val_max) {
        return val_max;
    } else if (val < val_min) {
        return val_min;
    } else {
        return val;
    }
}
[Differentiable]
[PreferRecompute]
float3 clamp_(float3 val, float val_min, float val_max)
{
    return float3(
        clamp_(val.x, val_min, val_max),
        clamp_(val.y, val_min, val_max),
        clamp_(val.z, val_min, val_max)
    );
}
```

---

Tested on modified soft 2d rasterizer. 
- https://github.com/aroidzap/slang-python/blob/clamp_fwd_diff_bug/examples/soft-rasterizer-example/rasterizer2d.py
- https://github.com/aroidzap/slang-python/blob/clamp_fwd_diff_bug/examples/soft-rasterizer-example/soft-rasterizer2d.slang#L153

With `clamp_` workaround:
image

With built-in `clamp`:
image

csyonghe:
Closes #3749.

### DIFF
--- a/source/slang/diff.meta.slang
+++ b/source/slang/diff.meta.slang
@@ -1548,7 +1548,7 @@ DifferentialPair<T> __d_clamp(DifferentialPair<T> dpx, DifferentialPair<T> dpMin
 {
     return DifferentialPair<T>(
         clamp(dpx.p, dpMin.p, dpMax.p),
-        dpx.p < dpMin.p ? (dpx.p > dpMax.p ? dpMax.d : dpx.d) : dpMin.d);
+        dpx.p < dpMin.p ? dpMin.d : (dpx.p > dpMax.p ? dpMax.d : dpx.d));
 }
 __generic<T : __BuiltinFloatingPointType>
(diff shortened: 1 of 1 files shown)


aroidzap:
I tested it and the issue is now fixed in the Slang `v2024.1.2`

PR #3755, opened 2024-03-13
Implement glsl atomic's [non image or memory scope] with optional extension(s); resolves #3587 for GLSL & SPIR-V targets

### LINKED ISSUE #3587: Implement "Atomic" GLSL functions to match to how much HLSL side works
This task is to implement "atomic" functions that [OpenGL 4.6 spec](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf) requires.

This is also a part of the other umbrella task #3362 .

There are 32 function names related to this task and they are described in the section 8.10 and 8.11 of OpenGL 4.6 spec.
The function names are listed below:

uint atomicCounterIncrement(atomic_uint c)
uint atomicCounterDecrement(atomic_uint c)
uint atomicCounter(atomic_uint c)
uint atomicCounterAdd(atomic_uint c, uint data)
uint atomicCounterSubtract(atomic_uint c, uint data)
uint atomicCounterMin(atomic_uint c, uint data)
uint atomicCounterMax(atomic_uint c, uint data)
uint atomicCounterAnd(atomic_uint c, uint data)
uint atomicCounterOr(atomic_uint c, uint data) A
uint atomicCounterXor(atomic_uint c, uint data)
uint atomicCounterExchange(atomic_uint c, uint data)
uint atomicCounterCompSwap(atomic_uint c, uint compare, uint data)
uint atomicAdd(inout uint mem, uint data)
int atomicAdd(inout int mem, int data)
uint atomicMin(inout uint mem, uint data)
int atomicMin(inout int mem, int data)
uint atomicMax(inout uint mem, uint data)
int atomicMax(inout int mem, int data)
uint atomicAnd(inout uint mem, uint data)
int atomicAnd(inout int mem, int data)
uint atomicOr(inout uint mem, uint data)
int atomicOr(inout int mem, int data)
uint atomicXor(inout uint mem, uint data)
int atomicXor(inout int mem, int data)
uint atomicExchange(inout uint mem, uint data)
int atomicExchange(inout int mem, int data)
uint atomicCompSwap(inout uint mem, uint compare, uint data)
int atomicCompSwap(inout int mem, int compare, int data)

ArielG-NV:
The following PR implements atomic operations & types associated with OpenGL 4.6, GL_EXT_vulkan_glsl_relaxed, GLSL_EXT_shader_atomic_float, GLSL_EXT_shader_atomic_float2, for GLSL & SPIR-V targets.

Fully implements all functions, and built-in type's, resolves https://github.com/shader-slang/slang/issues/3560 for GLSL & SPRI-V targets.

[Atomic extensions for GLSL can be found here](https://github.com/KhronosGroup/GLSL/tree/main)

Notes of worth:

* atomic_uint is well defined in GLSL->OpenGL, although was removed in GLSL->VK unless a compiler extension is supported (GL_EXT_vulkan_glsl_relaxed). This support entails transforming all atomic_uint operations and references into a storage buffer. SPIR-V has AtomicCounter+AtomicStorage (atomic_uint parallel) but does not implement these capabilities for SPIR-V->VK in any scenario. Due to the case we transform atomic_uint ourselves (GLSL_Syntax->Slang_IR) to accommodate transforming atomic_uint into valid syntax.
* GLSL_EXT_shader_atomic_float2 (all float16_t & some float/double operations) support is minimal and worth watching out for if enabling the tests.


source/slang/slang-ir-link.cpp
```diff
@@ -1520,6 +1521,138 @@ static void diagnoseUnresolvedSymbols(TargetRequest* req, DiagnosticSink* sink,
     }
 }
 
+void convertAtomicToStorageBuffer(
```
csyonghe, line 1524: Can we just expose the type as a pure user space thing? E.g. this in stdlib:

struct atomic_uint
{
RWStructuredBuffer<uint> buffer;
}

uint atomicAdd(atomic_uint v, uint x)
{
    uint oldV;
    InterlockedAdd(v.buffer[0], x, oldV);
    return oldV;
}

ArielG-NV: we cannot due to the very unique glsl layout syntax

ArielG-NV: in a regular structured buffer we can have an unbound int array, which is fine and works for the data storage

when fetching and indexing we would need an intrinsic much like what is being done with raytracing, this time with offset of an atomic uint; although this time todo this we would need a pass to convert all same binding atomic_uint's and field access  into 1 atomic_uint and fetch proper field access;

we would still need some magic type to work with and its still going to mangle output, we just will do it differently

csyonghe: I see. Yeah this is the right thing to do.

ArielG-NV: this as in change to the other implementation choice or keep as is (for now)?

csyonghe: The current implementation is good as is.

source/slang/slang-diagnostic-defs.h
```diff
@@ -243,6 +243,8 @@ DIAGNOSTIC(20101, Warning, unintendedEmptyStatement, "potentially unintended emp
 
 DIAGNOSTIC(30102, Error, declNotAllowed, "$0 is not allowed here.")
 
+DIAGNOSTIC(20001, Error, missingLayoutBindingModifier, "Expecting 'binding' modifier in the layout qualifier here")
```
csyonghe, line 246: 20001 is already used, need to use a different id.

csyonghe (approved):

PR #3798, opened 2024-03-19
Change representation of float literal in language translation

kaizhangNV:
Fix the issue (#3490).

Previous implementation could zero out the very small float literal.

We now use scientific notation instead to represent the float numbers whose exponential part are larger than a threshold. In other cases, keep using fixed notation.


csyonghe:
There are small floating point errors on the Falcor unit tests. These errors may be fine, but could also be legit issues. How can we verify if we are more correct or less correct with this change?

csyonghe (approved):

Issue #3631, opened 2024-02-26
Implement "Image" GLSL functions to match to how much HLSL side works


jkwak-work:
This task is to implement "image" functions that [OpenGL 4.6 spec](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf) requires.

This is also a part of the other umbrella task https://github.com/shader-slang/slang/issues/3362 .

There are 33 function names related to this task and they are described in the section 8.12 of OpenGL 4.6 spec.
The function names are listed below:
int imageSize(readonly writeonly gimage1D image)
ivec2 imageSize(readonly writeonly gimage2D image)
ivec3 imageSize(readonly writeonly gimage3D image)
ivec2 imageSize(readonly writeonly gimageCube image)
ivec3 imageSize(readonly writeonly gimageCubeArray image)
ivec3 imageSize(readonly writeonly gimage2DArray image)
ivec2 imageSize(readonly writeonly gimage2DRect image)
ivec2 imageSize(readonly writeonly gimage1DArray image)
ivec2 imageSize(readonly writeonly gimage2DMS image)
ivec3 imageSize(readonly writeonly gimage2DMSArray image)
int imageSize(readonly writeonly gimageBuffer image)
int imageSamples(readonly writeonly gimage2DMS image)
int imageSamples(readonly writeonly gimage2DMSArray image)
gvec4 imageLoad(readonly IMAGE_PARAMS)
void imageStore(writeonly IMAGE_PARAMS, gvec4 data)
uint imageAtomicAdd(IMAGE_PARAMS, uint data)
int imageAtomicAdd(IMAGE_PARAMS, int data)
uint imageAtomicMin(IMAGE_PARAMS, uint data)
int imageAtomicMin(IMAGE_PARAMS, int data)
uint imageAtomicMax(IMAGE_PARAMS, uint data)
int imageAtomicMax(IMAGE_PARAMS, int data)
uint imageAtomicAnd(IMAGE_PARAMS, uint data)
int imageAtomicAnd(IMAGE_PARAMS, int data)
uint imageAtomicOr(IMAGE_PARAMS, uint data)
int imageAtomicOr(IMAGE_PARAMS, int data)
uint imageAtomicXor(IMAGE_PARAMS, uint data)
int imageAtomicXor(IMAGE_PARAMS, int data)
uint imageAtomicExchange(IMAGE_PARAMS, uint data)
int imageAtomicExchange(IMAGE_PARAMS, int data)
[...]

PR #3810, opened 2024-03-21
Implement GLSL gimageDim & memory qualifiers with optional extension(s); resolves #3587 for GLSL & SPIR-V targets #3631

### LINKED ISSUE #3587: Implement "Atomic" GLSL functions to match to how much HLSL side works

ArielG-NV:
The following PR implements for GLSL syntax image operations, memory qualifier support & types associated using OpenGL 4.6, GLSL_EXT_shader_image_int64, GLSL_EXT_shader_atomic_float, GLSL_EXT_shader_atomic_float2, for GLSL & SPIR-V targets.

Fully implements all functions, and built-in type's, resolves https://github.com/shader-slang/slang/issues/3631 for GLSL & SPRI-V targets.

[Image extensions for GLSL can be found here](https://github.com/KhronosGroup/GLSL/tree/main).

*Notes:*
testing tools:

- R64_.* images only support Storage buffers on most hardware, as a result the format is not tested inside gfx, but is implemented for use

- Testing tools intentionally handle input texture data with uint8_t->int8_t, expect for {128,255} to get {-128,-1}

target implementation:
- All 64bit types do not support sampled images or buffer views on hardware tested, these tests are partially disabled.
- Atomic_float2 support is not complete on hardware tested and the following tests were similarly, partially disabled.
- Added memory qualifiers due to the heavy use with gimageDim operations; errors for most methods of using these types were added to ssbo and image tests.
- gimageRect support is 'omitted' (compiled but as 2D shape) since it is legacy and not used in Vulkan.


source/slang/glsl.meta.slang
```diff
@@ -2821,6 +2821,453 @@ public vec4 shadow2DProjLod(sampler2DShadow sampler, vec4 coord, float lod)
     return textureProjLod(sampler, coord, lod);
 }
 
+//// gimageDim Implementation
+
+public const int GL_TEXTURE_CUBE_MAP_POSITIVE_X = 0;
```
csyonghe, line 2826: `public static const int`.

Due to the legacy of HLSL design, global `const`s are treated as uniform parameters. `static const` makes them actual constant enums.

source/slang/glsl.meta.slang
```diff
@@ -2821,6 +2821,453 @@ public vec4 shadow2DProjLod(sampler2DShadow sampler, vec4 coord, float lod)
...
+        >;
+
+${{{{
+        if(targetShape.isRect)
+            continue;
+}}}}
+
+        [require(spirv)]
+        [require(glsl)]
+        [ForceInline] public $(imageSizeIntOrIVec) imageSize(readonly writeonly $(fullTypeName) image)
```
csyonghe, line 3021: `readonly writeonly` doesn't make much sense?

source/slang/slang-ast-expr.h
```diff
@@ -677,6 +677,7 @@ class SPIRVAsmOperand
         SampledType, // __sampledType(T), this becomes a 4 vector of the component type of T
         ImageType, // __imageType(texture), returns the equivalaent OpTypeImage of a given texture typed value.
         SampledImageType, // __sampledImageType(texture), returns the equivalent OpTypeSampledImage of a given texture typed value.
+        ImagePointer, // __imagePointer(texture), returns the global variable pointer of the given texture unmangled by loads/stores
```
csyonghe, line 680: I believe you can just use:
```
spirv_asm {
    OpBlah &globalVar
}
```
To reference the `globalVar` without `Load`.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3537,6 +3537,10 @@ void CLikeSourceEmitter::emitStructDeclarationsBlock(IRStructType* structType, b
                 emitPackOffsetModifier(fieldKey, fieldType, packOffsetDecoration);
             }
         }
+        if(this->getTargetProgram()->getOptionSet().getBoolOption(CompilerOptionName::AllowGLSL))
```
csyonghe, line 3540: Let's try to avoid wrapping things in `if (CompilerOptionName::AllowGLSL)`, because I want to remove this option going forward.

source/slang/slang-emit-spirv.cpp
```diff
@@ -5553,6 +5611,15 @@ struct SPIRVEmitContext
                     emitOperand(ensureInst(sampledType));
                     break;
                 }
+                case kIROp_SPIRVAsmOperandImagePointer:
```
csyonghe, line 5614: As commented above, I don't think we need this case, since we can just do `&var`.

source/slang/slang-parser.cpp
```diff
@@ -7779,13 +7786,95 @@ namespace Slang
...
+    static NodeBase* parseVolatileModifier(Parser* parser, void* /*userData*/)
+    {
+        ModifierListBuilder listBuilder;
+
+        auto hlslMod = parser->astBuilder->create<HLSLVolatileModifier>();
+        hlslMod->keywordName = getName(parser, "volatile");
+        hlslMod->loc = parser->tokenReader.peekLoc();
+        listBuilder.add(hlslMod);
+
+        if (!parser->options.allowGLSLInput)
```
csyonghe, line 7799: There isn't much point checking for this option here, since by the time we reach this location, the token is already treated as if it is a modifier. If we are truly keen to make the keyword conditional, then we should not be registering the modifier parser function if we are not in glsl mode. But I think we don't need to worry that much at the moment.

tests/glsl-intrinsic/image/i64imageDim.slang
```diff
@@ -0,0 +1,476 @@
...
+{
+    uint data[1];
+} outputBuffer;
+
+layout(local_size_x = 1) in;
+
+//TEST_INPUT: RWTexture1D(format=R64_SINT, size=4, content=one, mipMaps = 1):name image_1d
+uniform layout(binding=0,r64i) i64image1D image_1d;
+#ifdef TEST_when_hardware_supports_i64_and_u64_buffers
+//COM:TEST_INPUT: RWTextureBuffer(format=R64_INT, stride=8, data=[0 1 0 1 0 1 0 1]):name image_buffer
```
csyonghe, line 23: Let's use the new syntax for this:
```
//TEST_INPUT: SET IMAGE_BUFFER = RWTextureBuffer(format=R64_INT, stride=8, data=[0 1 0 1 0 1 0 1])
```

source/slang/glsl.meta.slang
ArielG-NV: its glsl syntax for "no side effect, no reading, glsl is doing something with the variable, not with it's data"

ArielG-NV:
note: I will address the review of code after fixing falcor regressions

source/slang/glsl.meta.slang
csyonghe: Ah, that reminds me we should also decorate these functions with `[__readNone]` to indicate the function has no side effects and can be removed if its result is not used.

source/slang/slang-ast-expr.h
ArielG-NV: the code generated loads a pointer to the texture object if we use &, we instead need a direct call to the texture object and to skip this load, __imagePointer skips it

slang-gfx.h
```diff
@@ -221,6 +221,8 @@ class IShaderProgram: public ISlangUnknown
     x(R16G16_FLOAT, 4, 1) \
     x(R16_FLOAT, 2, 1) \
     \
+    x(R64_UINT, 8, 1) \
```
csyonghe, line 224: This is going to break binary compatibility, that's causing the failures in Falcor.

ArielG-NV: does our testing system not replace the header?

ArielG-NV: so it seems we can't add R64.* support unless the compile process is changed for falcor (to not rely on pacman)?

source/slang/slang-emit-spirv.cpp
ArielG-NV: commented why the workaround is required, an exact pattern of what happens if we do not will be attached in a comment here (with why it does not work)

source/slang/glsl.meta.slang
```diff
@@ -2821,6 +2821,455 @@ public vec4 shadow2DProjLod(sampler2DShadow sampler, vec4 coord, float lod)
...
+                return spirv_asm
+                {
+                    OpCapability ImageQuery;
+                    result:$$$(imageSizeIntOrIVec) = OpImageQuerySize $image;
+                };
+            }
+        }
+        [require(spirv)]
+        [require(glsl)]
+        [ForceInline] public $(targetType.prefix)vec4 imageLoad(readonly $(IMAGE_PARAMS))
```
csyonghe, line 3037: Should be [NoSideEffect]

source/slang/slang-emit-spirv.cpp
csyonghe: I think it will work if the image parameter is __ref.

ArielG-NV: Image objects are not an L value, I don't believe __ref will work.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -254,8 +254,30 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
             }
             else
             {
-                if (!as<IRDecoration>(use->getUser()))
+                auto user = use->getUser();
+                if (!as<IRDecoration>(user))
                 {
+                    // do not emit load if we have a `store(Ptr<T>, globalParam<T>)`, instead
```
csyonghe, line 260: This should be done in a separate pass. I believe we already have a `removeRedundantLoadStore` pass that tries to fold loads and stores, and it has more condition checks to ensure the operation can be done safely.

We shouldn't need to implement this logic here. If this is to workaround some problems, we need to tackle that issue in a different way.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -254,8 +254,30 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                    // replace the store(Ptr<T>, globalParam<T>) with a globalParam<T> since 
+                    // this is the end product anyways (save on: 1 load, 1 store op). 
+                    // This also stops TextureType from breaking
+                    if(auto opStore = as<IRStore>(user))
+                    {
+                        auto dstType = opStore->getOperand(0)->getDataType();
+                        auto srcType = opStore->getOperand(1)->getDataType();
+                        // Cover the case where we have a Ptr<Ptr<T>>
+                        while (dstType->getOp() == kIROp_PtrType && srcType->getOp() == kIROp_PtrType)
```
csyonghe, line 269: I don't quite understand the logic here. Why do we need to handle the case of ptr<ptr<T>?

source/slang/slang-ir-spirv-legalize.cpp
ArielG-NV: this is to solve the previous issue with textures. I will check if the logic can be moved to `removeRedundantLoadStore`

I noticed if we use a constref/ref we still have an issue with `insertLoadAtLatestLocation` inserting a Load it should not insert. This also was why interpolation was breaking.

For textures specifically we have the additional issue that we cannot just store(dst,load(src)) since we have a special way of constructing the type.

source/slang/slang-ir-spirv-legalize.cpp
ArielG-NV: I don't know if slang supports Ptr<Ptr< or plans to, it can be removed if not important.

csyonghe: This currently will occur with pointer-typed user code, such as:
```
void main () {
   int* ptr;
}
```

will be lowered as:
```
%ptr = IRVar : Ptr<Ptr<int>>
```

But I don't know why this will come up in spirv legalization pass.

ArielG-NV: I suppose I will leave out this change for now then (won't be hard to add if important anyways)

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: I think the general logic here is to not insert a load when the pointer is being used in a call, and that call takes an `inout` or `ref` parameter at the argument index.

ArielG-NV: this would work for most cases, interpolate does exactly this (its exactly the solution I used)

for TextureType since it is not an L value I must use a `__constref` which does not appear as a ref in SPIR-V legalization

adding a new step to eliminateRedundantLoadStore does fix the issue from what it looks like (and adding the step to run after SPIR-V simplification)

csyonghe:
I think I know the problem here. When spirv legalization handles the global image parameter, it assumes that all existing uses of the image parameter are by-value, although the atomic image operation needs to access it by reference.

Given that the image used to be just a GlobalParam of `Image` type, trying to access its address isn't legal since virtual SSA registers does not have an address, so the assumption is that the existing code can't be accessing it by reference, so we insert loads at all use sites after turning the Image parameter into an Image* parameter. There isn't a way to know that some uses needs to be translated into a pointer, omitting the load. I guess your previous `__imagePointer` operand is trying to get over this hurdle, which can't be solely replaced by using `__ref` and `&`...

This is a big mess, but I think the cleanest way to hack this is to introduce a separate inst:
`GetLegalizedSPIRVParamAddr(v)` that returns the address of `v` after its legalization.

Then we can write our spirv as:
```
__intrinsic_op($(kIROp_GetLegalizedSPIRVParamAddr))
Ptr<T, addrSpace> __getLegalizedSPIRVParamAddr<let addrSpace:int, T>(T val);

spirv_asm {
    OpImageAtomicAdd $__getLegalizedSPIRVParamAddr<$(SpvStorageClassUniformConstant)>(image) ...
}
```

This is similar to your `__imagePointer` idea, but in a slightly more general form so that this same pattern can be used in other places if they ever come up.

ArielG-NV:
to give full scope of the issue in a comment since I do know what is wrong:
* In SPIR-V ASM, `processGlobalParam` applies a `load` whenever a `global` is used since these `global`'s (like a `Image`) are pointers to the actual data. 
* Most `Image` operations work on the `Image` inside the pointer meaning `processGlobalParam` works as intended. 
* The issue is that in some senarios we don't want this `load` that `processGlobalParam` produces, and want to work on a pointer to the `global` . The normal way todo this is `&`. The problem is when using a global a `load` is emitted first, then when trying to use `&`, Slang tries to resolve the not `load`'ed global into a pointer through a `store` which Slang fails todo, and even if it worked the code gen would be very inefficent. Currently Slang has no way to distinct these situations. `__intrinsic_op($(kIROp_ImageSubscript)) ref` solves the issue through a hack (as the comments describe) and major assumptions are made which we cannot do in our senario.

solutions attempted:
* intrinsic which just skips any single wrapping load when emitting (`__imagePointer`)
* removing store(dst,load(src)) redundant pattern
    * moving this to a redundant code optimization pass

source/slang/slang-emit-spirv.cpp
```diff
@@ -2724,6 +2758,13 @@ struct SPIRVEmitContext
                 const auto memorySemantics =  emitIntConstant(IRIntegerValue{SpvMemorySemanticsMaskNone}, builder.getUIntType());
                 return emitOpAtomicIDecrement(parent, inst, inst->getFullType(), inst->getOperand(0), memoryScope, memorySemantics);
             }
+        case kIROp_GetLegalizedSPIRVGlobalParamAddr:
```
csyonghe, line 2761: Is this needed? My assumption is that this inst should be legalized out during spirv legalization.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -2573,6 +2611,74 @@ void addArg(
     }
 }
 
+void compareParamToArgumentMemoryQualifier(
```
csyonghe, line 2614: Doing type check at IR lowering time is usually too late. You should be able to check this during visitInvokeExpr.

csyonghe (dismissed):

source/slang/slang-emit-spirv.cpp
ArielG-NV: `processGlobalParam` replaces all `globalParam` with a `load(globalParam)`, it walks the use's list.

adding a intrinsic meant the intrinsic will get a `load(globalParam)` after legalization from a `globalParam`.

csyonghe: Yeah, I think you still need that logic in `insertLoadAtLatestLocation` to handle the case where a user is `GetLegalizedParamAddr`, in which case you replace that getAddr inst with the pointer itself and remove the inst.

ArielG-NV: I understand what you are saying, will do 👍

csyonghe:
> * what would the SPRI-V storage class do?

During spirv legalization, we are assigning all pointer types with a spirv storage class (as the "address space" operand in the pointer type). This is required by the SPIRV spec, and different types of parameters needs to be declared with different storage classes.
See SPIRV spec here: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_storage_class

ArielG-NV:
> > * what would the SPRI-V storage class do?
> 
> During spirv legalization, we are assigning all pointer types with a spirv storage class (as the "address space" operand in the pointer type). This is required by the SPIRV spec, and different types of parameters needs to be declared with different storage classes. See SPIRV spec here: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_storage_class

My question is more directed to: why do we need this address space as a argument, we already can resolve it from the inputted variable (and do as such when resolving globals).

csyonghe:
I wrote that code because I am not sure if the spirv legalization pass is able to resolve the address space. But it seems like it does the job fine, so maybe this is not needed at all.

source/slang/slang-check-expr.cpp
```diff
@@ -2185,6 +2185,76 @@ namespace Slang
...
+            {
+                foundOp = opThis;
+                target = UnownedStringSlice("writeonly");
+            }
+
+            if (foundOp == ASTNodeType::NodeBase)
+                continue;
+
+            // Only texture types are allowed to have memory qualifiers on parameters
+            if(paramIn->getType()->astNodeType != ASTNodeType::TextureType)
```
csyonghe, line 2238: This can be moved to `SemanticsVisitor::checkModifiers`, otherwise we are repeating this check on every call.

As part of that checking, you can fold all image modifiers into a combined flag in the form of a `CheckedImageModifiers` Modifier, whose content is a combined uint bits:

```
class CheckedGLSLImageMemoryModifiers: Modifier
{
     SLANG_AST_CLASS(CheckedGLSLImageModifiers)
     GLSLImageMemoryModifierBit combinedBits;
}
```

So if we check `coherent writeonly`, we will fold those two modifiers into one `CheckedGLSLImageMemoryModifiers` that is `GLSLImageMemoryModifierBit::Coherent | GLSLImageMemoryModifierBit::WriteOnly`

csyonghe: You can modifier `checkModifier` to return a `CheckedGLSLImageMemoryModifiers` on a `GLSL**MemoryModifier`,
and combine them in `checkModifier` into a single `CheckedGLSLImageMemoryModifiers`.

source/slang/slang-check-expr.cpp
```diff
@@ -2185,6 +2185,76 @@ namespace Slang
...
+        ASTNodeType foundOp;
+        for (auto modThis : argDeclRef.getDecl()->modifiers)
+        {
+            UnownedStringSlice target;
+            opThis = modThis->astNodeType;
+            foundOp = ASTNodeType::NodeBase;
+            if (opThis == ASTNodeType::GloballyCoherentModifier)
+            {
+                foundOp = opThis;
+                target = UnownedStringSlice("coherent");
```
csyonghe, line 2211: You don't need this, just pass the modifier AST node to `diagnose` and it should work just fine. See `void printDiagnosticArg(StringBuilder& sb, Modifier* modifier)` defined in slang-ast-modifier.cpp.

source/slang/slang-check-expr.cpp
csyonghe: After checking the modifiers and folding them  into `CheckedGLSLImageMemoryModifiers`, you can add checking logic in `SemanticsDeclBodyVisitor::visitParamDecl` to issue the `image memory modifier on non-iimage type` diagnostic.

slang-gfx.h
csyonghe: We use a prebuilt binary of Falcor to speed up CI process, and to make sure we don't break binary compatiblity.

source/slang/slang-check-modifier.cpp
```diff
@@ -1228,6 +1228,37 @@ namespace Slang
...
+            if(!memoryQualifiers)
+            {
+                newModifier = true;
+                memoryQualifiers = getASTBuilder()->create<MemoryQualifierCollectionModifier>();
+            }
+            memoryQualifiers->addQualifier(m,
+                memoryQualifierBit);
+            if (newModifier)
+            {
+                m->next = memoryQualifiers;
```
csyonghe, line 1256: `next` will be overwritten by `checkModifiers` so this assignment will have no effect. We should just delete this line.

csyonghe (approved):

source/slang/slang-check-modifier.cpp
ArielG-NV: I need to return multiple modifiers (current+next) to keep compatibility with other code. 

I will add to checkMemoryQualifiers to solve this.

csyonghe: It is safe to just return `nullptr` here, in that case the original modifier won't appear in the checked modifier list.

ArielG-NV: checked modifiers replace the modifiers list in use. Since we need the memory qualifier and container, both need to be added

ArielG-NV: edit: I don't need to handle 2 modifiers, that was a misjudgment of the code, I have another way I can do the implementation

csyonghe: I think you can just add the original modifier to the list inside the new modifier, and return the new modifier.

csyonghe: The original modifier does not need to exist in the checked modfiier list, as long as our ir-lowering logic is no longer looking for those original modifiers. We keep them in the list of the new modifier just in case we need their source loc to report diagnostics or in case language server needs them.

source/slang/glsl.meta.slang
jkwak-work: The OpenGL spec says this,
```
A variable could be qualified as both readonly and writeonly, disallowing both read and write.
Such variables can still be used with some queries, for example imageSize() and .length()
```
Yong and I asked the same question.
And probably whoever will look at this code will have the same question.
Can you leave a comment about this to be more efficient for the readers?

tests/glsl/ssboMemberMemoryQualifier.slang
```diff
@@ -0,0 +1,48 @@
+//TEST:SIMPLE(filecheck=CHECK-GLSL): -stage compute -entry computeMain -allow-glsl -target glsl
+//TEST:SIMPLE(filecheck=CHECK-SPV): -stage compute -entry computeMain -allow-glsl -target spirv -emit-spirv-directly
+
+volatile buffer Block1
+{
+    restrict uint data1;
+    readonly uint data2;
+    readonly writeonly uint data3;
```
jkwak-work, line 8: Can we test about "readonly writeonly" keyword with ImageSize and .length() as the OpenGL spec says?
```
A variable could be qualified as both readonly and writeonly, disallowing both read and write.
Such variables can still be used with some queries, for example imageSize() and .length()
```

ArielG-NV: .length is not implemented and not currently trivially implementable (due to swizzle implementation). imageSize is implemented & tested with `readonly writeonly`.

csyonghe (approved):

PR #3815, opened 2024-03-22
Changes to make UE TSR working with Slang.

jkwak-work:
Adding constructor for matrix that takes a partial matrix and a vector.
For example, "mat4x4 var = mat4x4(mat3x4, vec4)".

Adding implicit cast from vector<T,1> to T.

transpose(matrix<T,R,C>) had ambigous calls when T is int, because there
are definition for __BuiltinIntegerType and __BuiltinLogicalType. "int"
couldn't be resolved properly. In order to resolve the issue,
__BuiltinLogicalType got [OverloadRank(-1)], which gives
__BuiltinIntegerType a higher chance to be used.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1073,7 +1073,7 @@ __magic_type(MatrixExpressionType)
 struct matrix : IArray<vector<T,C>>
 {
     __intrinsic_op($(kIROp_MakeMatrixFromScalar))
-    __implicit_conversion($(kConversionCost_ScalarToVector))
+    __implicit_conversion($(kConversionCost_ScalarToMatrix))
     [TreatAsDifferentiable]
     __init(T val);
 
@@ -1158,13 +1158,13 @@ extension matrix<T,N,M,L> : IFloat
     [OverloadRank(-1)]
     [TreatAsDifferentiable]
     [__unsafeForceInlineEarly]
-    __implicit_conversion($(kConversionCost_ScalarToVector))
+    __implicit_conversion($(kConversionCost_IntegerToFloatConversion))
     __init(int v) { this = matrix<T,N,M>(T(v)); }
 
     [OverloadRank(-1)]
     [TreatAsDifferentiable]
     [__unsafeForceInlineEarly]
-    __implicit_conversion($(kConversionCost_ScalarToVector))
+    __implicit_conversion($(kConversionCost_ScalarToMatrix))
     __init(float v) { this = matrix<T,N,M>(T(v)); }
 
     // IDifferentiable.
@@ -1259,7 +1259,7 @@ for (int tt = 0; tt < kBaseTypeCount; ++tt)
 
     sb << "extension " << tname << "\n";
     sb << "{\n";
-    sb << "    __implicit_conversion($(kConversionCost_ScalarToVector))\n";
+    sb << "    __implicit_conversion($(kConversionCost_VectorToScalar))\n";
     sb << "    __init(vector<" << tname << ",1> v) { this = v[0]; }\n";
     sb << "}\n";
 }
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1259,7 +1259,7 @@ for (int tt = 0; tt < kBaseTypeCount; ++tt)
 
     sb << "extension " << tname << "\n";
     sb << "{\n";
-    sb << "    __implicit_conversion(" << kConversionCost_VectorToScalar << ")\n";
+    sb << "    __implicit_conversion(" << kConversionCost_OneVectorToScalar << ")\n";
     sb << "    __init(vector<" << tname << ",1> v) { this = v[0]; }\n";
     sb << "}\n";
 }
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1259,7 +1259,7 @@ for (int tt = 0; tt < kBaseTypeCount; ++tt)
 
     sb << "extension " << tname << "\n";
     sb << "{\n";
-    sb << "    __implicit_conversion($(kConversionCost_VectorToScalar))\n";
+    sb << "    __implicit_conversion(" << kConversionCost_VectorToScalar << ")\n";
     sb << "    __init(vector<" << tname << ",1> v) { this = v[0]; }\n";
     sb << "}\n";
 }
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1158,7 +1158,7 @@ extension matrix<T,N,M,L> : IFloat
     [OverloadRank(-1)]
     [TreatAsDifferentiable]
     [__unsafeForceInlineEarly]
-    __implicit_conversion($(kConversionCost_IntegerToFloatConversion))
+    __implicit_conversion($(kConversionCost_ScalarIntegerToFloatMatrix))
     __init(int v) { this = matrix<T,N,M>(T(v)); }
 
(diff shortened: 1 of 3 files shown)


source/slang/core.meta.slang
```diff
@@ -1185,6 +1193,13 @@ extension matrix<T,N,M,L> : IFloat
     }
 }
 
+__generic<T:__BuiltinIntegerType, let R : int, let C : int, let L : int>
+extension matrix<T,R,C,L>
+{
+    __implicit_conversion(200)
```
csyonghe, line 1199: This needs to be turned in to a `kConversionCost_*` constant.

source/slang/core.meta.slang
```diff
@@ -1232,6 +1247,22 @@ for (int tt = 0; tt < kTypeCount; ++tt)
...
+
+for (int tt = 0; tt < kBaseTypeCount; ++tt)
+{
+    if(kBaseTypes[tt].tag == BaseType::Void) continue;
+
+    const char* tname = kBaseTypes[tt].name;
+
+    sb << "extension " << tname << "\n";
+    sb << "{\n";
+    sb << "    __implicit_conversion(200)\n";
```
csyonghe, line 1261: kConversionCost_*

csyonghe: The cost can be lower since float1 should in most cases be considered identical to float.

csyonghe (dismissed):

source/slang/core.meta.slang
jkwak-work: I will make it 10.

source/slang/core.meta.slang
jkwak-work: I will add it.

jkwak-work: Hmm.. I just realized that the numbers have some meanings...
I will need to give higher value for int -> int8_t and int16_t.

source/slang/slang-ast-support-types.h
```diff
@@ -143,6 +143,9 @@ namespace Slang
         // Default case (usable for user-defined conversions)
         kConversionCost_Default = 500,
 
+	// Cost of converting an integer to int16_t
+        kConversionCost_IntegerToShort = 700,
```
csyonghe, line 147: Maybe give it more general name IntegerTruncate?

source/slang/slang-ast-support-types.h
```diff
@@ -157,6 +160,8 @@ namespace Slang
         // a vector (this will be added to the cost, if any, of converting
         // the element type of the vector)
         kConversionCost_ScalarToVector = 1,
+        kConversionCost_VectorToScalar = 1,
+        kConversionCost_ScalarToMatrix = 1,
```
csyonghe, line 164: Maybe make ScalarToMatrix higher than SclarToVector

csyonghe (approved):

source/slang/slang-ast-support-types.h
jkwak-work: Yes I will do that.

source/slang/slang-ast-support-types.h
jkwak-work: Yes, I will go with 10.

jkwak-work:
This changes may not pass CI.
I am investigating.

jkwak-work:
It seems like my first commit is causing a problem.
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241,5): warning MSB8065: Custom build for item "..\..\..\source\slang\core.meta.slang" succeeded, but specified output "d:\sbf\git\slang\missing_features_for_tsr\source\slang\core.meta.slang.h" has not been created. This may cause incremental build to work incorrectly. [D:\sbf\git\slang\missing_features_for_tsr\build\visual-studio\run-generators\run-generators.vcxproj]
```
I am gonna continue my investigation tomorrow.

source/slang/core.meta.slang
```diff
@@ -1232,6 +1248,22 @@ for (int tt = 0; tt < kTypeCount; ++tt)
...
+
+for (int tt = 0; tt < kBaseTypeCount; ++tt)
+{
+    if(kBaseTypes[tt].tag == BaseType::Void) continue;
+
+    const char* tname = kBaseTypes[tt].name;
+
+    sb << "extension " << tname << "\n";
+    sb << "{\n";
+    sb << "    __implicit_conversion($(kConversionCost_VectorToScalar))\n";
```
csyonghe, line 1262: Should be sb<<"..."<<kConversionCost_cxx<<")";

jkwak-work: That did it!!
Thank you very much.
I was worried for a while.

jkwak-work:
Hmm. It looks like some of tests start to fail.
I guess my change was not ready for PR.
I am sorry about this.

Let me investigate and fix the issues.

source/slang/core.meta.slang
```diff
@@ -1419,6 +1451,19 @@ for( int C = 2; C <= 4; ++C )
         if(rr == R && cc == C) continue;
         sb << "__intrinsic_op(" << int(kIROp_MatrixReshape) << ") __init(matrix<T," << rr << "," << cc << ", L> value);\n";
     }
+
+    // Initialize from matrix and vector
+    if (R > 2)
```
csyonghe, line 1456: If R==4, we shouldn't define anything here, right?

source/slang/slang-ast-support-types.h
```diff
@@ -157,6 +160,9 @@ namespace Slang
         // a vector (this will be added to the cost, if any, of converting
         // the element type of the vector)
         kConversionCost_ScalarToVector = 1,
+        kConversionCost_VectorToScalar = 1,
```
csyonghe, line 163: Perhaps we should make VectorToScalar smaller than ScalarToVector so we always prefer the overload that makes things simpler. Also might be more accurate to name this OneVectorToScalar.

source/slang/core.meta.slang
jkwak-work: When R is 4, we will have to have the followings:
```
__generic<T, let L:int> __extension matrix<T, 4,2, L>
{
__init(matrix<T,3,2> m, vector<T,2> row3) { this = This(m[0], m[1], m[2], row3); }
}
__generic<T, let L:int> __extension matrix<T, 4,3, L>
{
__init(matrix<T,3,3> m, vector<T,3> row3) { this = This(m[0], m[1], m[2], row3); }
}
__generic<T, let L:int> __extension matrix<T, 4,4, L>
{
__init(matrix<T,3,4> m, vector<T,4> row3) { this = This(m[0], m[1], m[2], row3); }
}
```
It is little confusing but what we shouldn't have is when R is 1 where it should just take a vector.

source/slang/slang-ast-support-types.h
jkwak-work: Make sense.
I will rename it and.... raise ScalarToVector by one.

source/slang/core.meta.slang
csyonghe: Oh, I see what's going on here, my bad. This is good then. In general HLSL might be allowing constructing a matrix with arbitrary vectors + matrix parameters and just concatenatibf them together. But we don't have to add those in this PR if the epic shader is not using them.

csyonghe (approved):

csyonghe (approved):

PR #3817, opened 2024-03-22
Make `-no-mangle` option work, add `-no-hlsl-binding`.

### LINKED ISSUE #3813: SLANG_COMPILE_FLAG_NO_MANGLING doesn't work as expected
When SLANG_COMPILE_FLAG_NO_MANGLING is used, the variable names are expected to have no mangling.
But they appear to be mangled.

csyonghe:
Potentially fixes #3813.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -1772,7 +1772,13 @@ namespace Slang
                 parentAggTypeDecl->unionTagsWith(getTypeTags(varDeclRefType));
             }
         }
-
+        if (getOptionSet().getBoolOption(CompilerOptionName::NoMangle) &&
+            (isGlobalDecl(varDecl) || as<StructDecl>(getParentDecl(varDecl))))
+        {
+            // If -no-mangle option is set, we will add `ExternCpp` modifier to all
+            // global variables and struct fields to prevent mangling.
+            addModifier(varDecl, m_astBuilder->create<ExternCppModifier>());
+        }
         checkVisibility(varDecl);
     }
 
@@ -1812,6 +1818,12 @@ namespace Slang
             member->nameAndLoc.loc = structDecl->wrappedType.exp->loc;
             member->loc = member->nameAndLoc.loc;
             structDecl->addMember(member);
+        }
+
+        if (getOptionSet().getBoolOption(CompilerOptionName::NoMangle))
+        {
+            // If -no-mangle option is set, we will add `ExternCpp` modifier to prevent mangling.
+            addModifier(structDecl, m_astBuilder->create<ExternCppModifier>());
         }
         checkVisibility(structDecl);
     }
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -1773,7 +1773,7 @@ namespace Slang
             }
         }
         if (getOptionSet().getBoolOption(CompilerOptionName::NoMangle) &&
-            (isGlobalDecl(varDecl) || as<StructDecl>(getParentDecl(varDecl))))
+            isGlobalDecl(varDecl))
         {
             // If -no-mangle option is set, we will add `ExternCpp` modifier to all
             // global variables and struct fields to prevent mangling.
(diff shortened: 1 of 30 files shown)


jkwak-work:
This works well with UE.

jkwak-work:
While reviewing the code, I realized that I need to test one more aspect.
I will leave the result as soon as my testing is done. ETA 1hour

jkwak-work:
While I was testing the code, I encountered an assertion failure in a file, "slang-ir-collect-global-uniforms.cpp".
```
auto globalParam = _getGlobalParamFromLayoutFieldKey(fieldLayoutAttr->getFieldKey());
SLANG_ASSERT(globalParam);
```
I am going to send you an email with the failing shader source file.

jkwak-work:
I think this change is good by itself although it is causing an issue with UE shaders for another reason.
I have verified that three following behaviors with the change:
1. The names of "cbuffer" are no longer mangled with a prefix "SLANG_parameterGroup_"
2. The variables don't get the register values unless the source explicitly specified.
3. Global variables, members in cbuffer and members in struct are no longer mangled in their names.
I think that if the members in the struct get name mangling, it wouldn't have the assertion failure I mentioned on my previous comment. To be clear, the members in cbuffer shouldn't get the name mangling.

jkwak-work (approved):

source/slang/slang-check-decl.cpp
```diff
@@ -1772,7 +1772,13 @@ namespace Slang
                 parentAggTypeDecl->unionTagsWith(getTypeTags(varDeclRefType));
             }
         }
-
+        if (getOptionSet().getBoolOption(CompilerOptionName::NoMangle) &&
+            (isGlobalDecl(varDecl) || as<StructDecl>(getParentDecl(varDecl))))
```
jkwak-work, line 1776: Does "as<StructDecl>" apply to both cbuffer and struct?
Can we apply this only to cbuffer not struct?
Because it looks like the compilation fails if the no-mangle is applied to both.
I think it will be fine to apply it only to cbuffer as far as UE is concerned.

jkwak-work (approved):

source/slang/slang-check-decl.cpp
csyonghe: We translate cbuffer to ConstantBuffer<StructDecl>. If we want to apply only to members of cbuffer, we need to decorate the struct members we generate from that cbuffer with __extern_cpp.

csyonghe: Ah, I see why things would fail... Because if two symbols have the same mangled name, we will crash.

The short term fix of only extern_cpp the cbuffer fields seems like the right thing to do.

jkwak-work (approved):
The code looks good.
And UE works fine with the tweak now.

jkwak-work:
I have a trivial question about the change.
In order for this change to be effective, I had to add the following lines in the UE shader compile code,
```
+       slang::CompilerOptionEntry compilerOptions[2];
+       compilerOptions[0].name = slang::CompilerOptionName::NoMangle;
+       compilerOptions[0].value.intValue0 = 1;
+       compilerOptions[1].name = slang::CompilerOptionName::NoHLSLBinding;
+       compilerOptions[1].value.intValue0 = 1;
+       sessionDesc.compilerOptionEntries = compilerOptions;
+       sessionDesc.compilerOptionEntryCount = UE_ARRAY_COUNT(compilerOptions);
```
I was wondering if it is necessary when I am already using the other compiler flag,
```
compileRequest->setCompileFlags(SLANG_COMPILE_FLAG_NO_MANGLING);
```

jkwak-work:
Actually, I am gonna test one more time if SLANG_COMPILE_FLAG_NO_MANGLING is enough without the other flag setting.

jkwak-work:
So that was my misunderstanding.
This line appears to be working,
```
compileRequest->setCompileFlags(SLANG_COMPILE_FLAG_NO_MANGLING);
```
I need to apply NoHLSLBinding tho,
```
	slang::CompilerOptionEntry compilerOptions[1];
	compilerOptions[0].name = slang::CompilerOptionName::NoHLSLBinding;
	compilerOptions[0].value.intValue0 = 1;
	sessionDesc.compilerOptionEntries = compilerOptions;
	sessionDesc.compilerOptionEntryCount = UE_ARRAY_COUNT(compilerOptions);
```

csyonghe:
Yes we will translate the SLANG_COMPILE_FLAG_NO_MANGLING flag to the NoMangle option entry. The flags are legacy API and we have to keep them for binary compatibility. All new options are exposed as CompilerOptionEntry's.

Issue #3637, opened 2024-02-27
Constant propogation can incorrectly perform sign extension


expipiplus1:
testcase:

```hlsl
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vulkan -output-using-type

// CHECK: 9223372036854775808
// CHECK: 1
// CHECK: 0
// CHECK: 18446744073709551615

// This tests exhibits a bug in constant folding in which the right shift
// operator unconditionally performs sign-extension.

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=8):out,name=outputBuffer
RWStructuredBuffer<uint64_t> outputBuffer;

[numthreads(1, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    const uint64_t topBitSet = 0x8000000000000000ull;
    const uint64_t bottomBitSet = topBitSet >> 63; // This actually ends up sign-extending to set all the bits
    const uint64_t noBitsSet = topBitSet >> 64; // This exhibits undefined behaviour, as the compiler shifts right by at least the number of bits in the first operand
    const uint64_t allBitsSet = int64_t(topBitSet) >> 63;
    outputBuffer[0] = topBitSet;
    outputBuffer[1] = bottomBitSet;
    outputBuffer[2] = noBitsSet;
    outputBuffer[3] = allBitsSet;
}
```

The code which caught my eye:

https://github.com/shader-slang/slang/blob/188482e83dc4f7dc4c940214033dd95518b9d18c/source/slang/slang-ir-sccp.cpp#L682-L686

This code should check the type of the first operand and decide to sign-extend or not.

Additionally, it and the left shift version should take care not to shift by more than 64 bits as this is UB

Additionally, the above test case fails for the CPU target

kaizhangNV:
Fix is merged, close the issue now.

PR #3820, opened 2024-03-22
Fix the sign-extending issue in right shift

kaizhangNV:
In constant folding of a right shift operation,slang always uses signed interger as the operand no matter the input source code is signed or unsigned, this could causes sign-extending issue if the input source is unsigned integer with highest bit set to 1.

Fix the issue by checking the original type of the input and use the unsigned type if the input is unsigned.

### DIFF
--- a/source/slang/slang-ir-sccp.cpp
+++ b/source/slang/slang-ir-sccp.cpp
@@ -676,11 +676,23 @@ struct SCCPContext
     }
     LatticeVal evalLsh(IRType* type, LatticeVal v0, LatticeVal v1)
     {
+        IntInfo info = getIntTypeInfo(type);
+        if (info.isSigned == false)
+        {
+            return evalBinaryIntImpl(
+               type, v0, v1, [](IRUnsignedIntegerValue c0, IRUnsignedIntegerValue c1) { return c0 << c1; });
+        }
         return evalBinaryIntImpl(
             type, v0, v1, [](IRIntegerValue c0, IRIntegerValue c1) { return c0 << c1; });
     }
     LatticeVal evalRsh(IRType* type, LatticeVal v0, LatticeVal v1)
     {
+        IntInfo info = getIntTypeInfo(type);
+        if (info.isSigned == false)
+        {
+            return evalBinaryIntImpl(
+               type, v0, v1, [](IRUnsignedIntegerValue c0, IRUnsignedIntegerValue c1) { return c0 >> c1; });
+        }
         return evalBinaryIntImpl(
             type, v0, v1, [](IRIntegerValue c0, IRIntegerValue c1) { return c0 >> c1; });
     }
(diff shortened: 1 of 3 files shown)


csyonghe (commented):
We also need to take care of this during AST-level folding in `FuncCallIntVal::tryFoldImpl`.
I believe the current test case will trigger the folding logic in AST level.

To trigger the logic in IR level, we need to make it so that the AST level folding logic can't see the folding opportunity, by wrapping things in an inline function:
```
[ForceInline] int myFunc() { return 16; }
void main()
{
    uint64_t value = 0x8000000000000000ull >> myFunc();
}
```

kaizhangNV:
> We also need to take care of this during AST-level folding in `FuncCallIntVal::tryFoldImpl`. I believe the current test case will trigger the folding logic in AST level.
> 
> To trigger the logic in IR level, we need to make it so that the AST level folding logic can't see the folding opportunity, by wrapping things in an inline function:
> 
> ```
> [ForceInline] int myFunc() { return 16; }
> void main()
> {
>     uint64_t value = 0x8000000000000000ull >> myFunc();
> }
> ```

tests/bugs/gh-3637.slang
```diff
@@ -0,0 +1,27 @@
+//TEST(compute):COMPARE_COMPUTE_EX:-cuda -compute -shaderobj
```
csyonghe, line 1: These needs to be: COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):...

source/slang/slang-ir-sccp.cpp
```diff
@@ -676,11 +676,23 @@ struct SCCPContext
     }
     LatticeVal evalLsh(IRType* type, LatticeVal v0, LatticeVal v1)
     {
+        IntInfo info = getIntTypeInfo(type);
+        if (info.isSigned == false)
+        {
+            return evalBinaryIntImpl(
+               type, v0, v1, [](IRUnsignedIntegerValue c0, IRUnsignedIntegerValue c1) { return c0 << c1; });
```
csyonghe, line 683: I wonder why this doesn't lead to warnings when `evalBinaryIntImpl` calls the lambda with signed integer arguments.

csyonghe (dismissed):

tests/bugs/gh-3637.slang.3.expected.txt
```diff
@@ -0,0 +1,4 @@
+type: uint64_t
```
csyonghe, line 1: With the `filecheck` directive, you can delete this `.expected.txt` file.

kaizhangNV: Sure, will do.
(I just jumped back from #3825), and will also take care of with the one test failure.

source/slang/slang-ir-sccp.cpp
kaizhangNV: There is no precision narrow down or overflow, so in order to turn on this kind of warning, we might need "Wcoercion" in gcc.

tests/bugs/gh-3637.slang.3.expected.txt
kaizhangNV: File is removed.

PR #3822, opened 2024-03-23
Allow anonymous struct.

### LINKED ISSUE #3821: UE uses "unnamed struct" and Slang needs to support
Unreal Engine expects the following HLSL syntax to work properly.
```
static const struct
{
    int member;
} varName = { value };
```
But Slang errors out as following,
```
c.slang(2): error 20001: unexpected '{', expected identifier
{
^
```
The following syntax works for Slang but we need to support what UE expects.
```
struct StructName
{
    int member;
};
static const StructName varName = { value };
```

csyonghe:
Closes #3821.


jkwak-work (approved):
The code looks good.
And UE works well with this change.

PR #3833, opened 2024-03-25
Fix default space assignment.

### LINKED ISSUE #3829: Parameter block binding set changes when changed from global to entry-point parameter
(Originally posted as discussion https://github.com/shader-slang/slang/discussions/3778, please close if not an issue.)

Changing a parameter block from a global to an entry-point parameter seems to change its binding set (descriptor set index) from 0 to 1.  This seems like unexpected behaviour, as there are no other bindings, but I am not clear on the specification.

To reproduce this use the following shader (tested on v2024.1.2) with `-o test.glsl -entry main`:

Input (global):

```slang
struct Params {
    Sampler2D<float4> tex;
}
ParameterBlock<Params> params;
[shader("pixel")]
float4 main(float2 uv) {
    return params.tex.Sample(uv);
}
```

Output (global), set 0 as expected:

```glsl
layout(binding = 0)
uniform sampler2D params_tex_0;
```

Input (entry-point parameter):

```slang
struct Params {
    Sampler2D<float4> tex;
}
[shader("pixel")]
float4 main(float2 uv, ParameterBlock<Params> params) {
    return params.tex.Sample(uv);
}
```

Output (entry-point parameter), set 1 unexpectedly:

```glsl
layout(binding = 0, set = 1)
uniform sampler2D entryPointParams_params_tex_0;
```

My mental model of entry-point parameters is that they should be equivalent to declaring extra globals _only for that entry-point_, so these two cases should generate identical bindings, but perhaps my mental model is wrong? In general, using entry-point parameters seems useful when a file contains multiple entry-points (so you don't want to use global variables), so it would be useful to know what the expected behaviour is here.  Thanks!

csyonghe:
Fixes #3829.

### DIFF
--- a/source/slang/slang-parameter-binding.cpp
+++ b/source/slang/slang-parameter-binding.cpp
@@ -3609,6 +3609,7 @@ static bool _calcNeedsDefaultSpace(SharedParameterBindingContext& sharedContext)
                     break;
 
                 case LayoutResourceKind::RegisterSpace:
+                case LayoutResourceKind::SubElementRegisterSpace:
                 case LayoutResourceKind::VaryingInput:
                 case LayoutResourceKind::VaryingOutput:
                 case LayoutResourceKind::HitAttributes:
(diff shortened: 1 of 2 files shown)


kaizhangNV:
The change looks good to me.
Only one suggestion: can you add something in the commit message such that people could know how the change fixes the issue? This is helpful when the fix is not very straightforward.

kaizhangNV (approved):

csyonghe:
@kaizhangNV Thank you for your suggestion. I update the PR description to reflect what was done here.

PR #3836, opened 2024-03-26
Support mutable existential parameters.

### LINKED ISSUE #3818: When an object is send to a function as its interface type, it gets assigned to a wrong temporary variable, "AnyValue16"
The issue is observed when an interface has a "[mutating]" keyword on a method.
When an object of the derived class is used as a function argument and the type of function parameter is its interface type, calling to the mutating function appears to fail.

I found a simplified example that can reproduce the issue.
Note that the issue can be reproduced only when both REPRODUCE_ISSUE and USE_INTERFACE are set to 1.
When either of them is 0, the issue is not observed.
```
#define REPRODUCE_ISSUE 1
#define USE_INTERFACE 1

interface IOperation
{
#if REPRODUCE_ISSUE
    [mutating]
#endif
    void Store(float v);

    float Load();
};

struct TOperationMax : IOperation
{
    float result;

#if REPRODUCE_ISSUE
    [mutating]
#endif
    void Store(float v)
    {}

    float Load()
    {
        return result;
    }
};

float Run(IOperation op, float val)
{
    op.Store(val);
    return op.Load();
}

[numthreads(4, 1, 1)]
[shader("compute")]
{
    TOperationMax opMax;
    opMax.result = 0.f;
    float val = 2.f;
#if USE_INTERFACE
    float result = Run(opMax, val);
#else
    opMax.Store(val);
    float result = opMax.Load();
#endif
    outputBuffer[0] = result;
}
```
You can compile the code above with the following command to reproduce the issue.
```
slangc.exe -target dxil -entry computeMain -profile cs_6_6 -o a.dxil a.slang
```
The error message appears to always have a keyword, "AnyValue16" .
```
dxc 1.6: a.slang(42): error :  cannot initialize a variable of type 'AnyValue16' with an lvalue of type 'TOperationMax_0'
dxc 1.6: note :     AnyValue16 _S3 = _S1;
dxc 1.6: note :                ^     ~~~
```
I don't know what it means but that is the error signature for this issue.

You can also change the target to "hlsl" and find code like following,
```
struct TOperationMax_0
{
    float result_0;
};
struct AnyValue16
{
    uint field0_0;
    uint field1_0;
    uint field2_0;
    uint field3_0;
};
float Run_0(TOperationMax_0 _S1, float _S2)
{
    AnyValue16 _S3 = _S1;
```

csyonghe:
Fixes #3818.

This change implements the necessary logic in existential type handling to support mutable existential parameters.

## Up-casting to base interface type is not an l-value
The first change is to make an up-cast from a concrete type to a base interface type an R-Value operation, which means the result of the up-cast is always a non-left-value. Consider the following case:
```
float Run(inout IOperation op, float val)
{
    ...
}

struct TOperationMax : IOperation { ... }

{
    TOperationMax opMax;
    float result = Run(opMax, val); // not valid!
}
```
The call to `Run` should not be valid, because by definition, `Run` can return a value of different type through `op` argument, and in that case the new value won't fit in the `opMax` variable anymore.
Instead, the user must wrap things into an existential value, and then call the `Run` function:
```
{
    TOperationMax opMax;
    IOperation op = opMax; // box opMax into op.
    float result = Run(op, val); // OK! op can be of different type after the call.
}
```

## Don't open an existential argument if it is passed to a mutable interface typed parameter.

We have logic to "open" an existential value when using it to call a function that  takes an interface typed parameter, so that the function is called with a concrete typed argument that can be picked up by the specialization pass. The "opened" existential value is considered a sub type of the interface type so the call is still valid. However since opening an existential yields a r-value, we can't pass it to a mutable parameter location. Therefore, we have logic inplace to check that if the parameter is `out`, we don't want to open it. The same should apply if the parameter is `inout`, but we are missing that case we need to add the case there.

## Correctly lowering `ExtractEixstentialType` to `AnyValueType`.

Currently we lower `ExtractExistentialType` inst to a tuple extract, which works if the type is used as a value. If the type is used as a type of some other inst, it should be lowered into `AnyValueType` instead.

## Fix `lowerWitnessLookup` pass

The `lowerWitnessLookup` pass replaces the `this` argument in all calls to the witness lookup result to the existential value object. This is incorrect if some of the calls mutates the `this` parameter. Instead we need to reconstruct the existential value for `this` for each call independently, using only the initial existential object's type info.


jkwak-work (approved):
The code looks good.
But I wouldn't pretend that I understood all.

The explanations on the PR description makes all sense.
And the explanations on the design document helps me to understand much better.

I learned a lot about existential type for the first time.
Thank for the new learning opportunity.

jkwak-work:
It looks like there is a similar issue when the interface uses "__generic".
Here is an example,
```
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-shaderobj -output-using-type

__generic<T:__BuiltinFloatingPointType>
interface IOperation
{
    [mutating]
    void Store(float v);

    float Load();
};

__generic<T:__BuiltinFloatingPointType>
struct TOperationMax : IOperation<T>
{
    float result;

    [mutating]
    void Store(float v)
    {
        result = v;
    }

    float Load()
    {
        return result;
    }
};

__generic<T:__BuiltinFloatingPointType>
float Run(inout IOperation<T> op, float val)
{
    op.Store(val);
    return op.Load();
}

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name outputBuffer

[numthreads(4, 1, 1)]
[shader("compute")]
{
    TOperationMax<float> opMax;
    opMax.result = 0.f;
    float val = 2.f;
    IOperation<float> op = opMax;
    float result = Run<float>(op, val);
    // CHECK: 2.0
    outputBuffer[0] = result;
} 
```
And when tried, the compile error message looks like following,
```
core(231): error 50100: No type conformances are found for interface 'IDifferentiable'. Code generation for current target requires at least one implementation type present in the linkage.
core(130): error 50100: No type conformances are found for interface 'IFloat'. Code generation for current target requires at least one implementation type present in the linkage.
core(327): error 50100: No type conformances are found for interface '__BuiltinFloatingPointType'. Code generation for current target requires at least one implementation type present in the linkage.
```
It sounds like capability related problem but I am not sure.

PR #3842, opened 2024-03-27
Fix parsing issue around __transparent_block.

csyonghe:
Our decl parsing func is assumed to always return a non-null ast node upon successful parse.

The parseTransparentBlockDecl directly inserts the parsed decls in to the current parent isntead of constructing a new decl, so it was returning null. The problem with that is the calling function `ParseDeclWithModifiers` will see it as failed and continue to try fallback parsing logic, resulting wrong parsing behavior.

The fix here is to have `parseTransparentBlockDecl` return an EmptyDecl instead.


kaizhangNV (approved):
LGTM.

PR #3848, opened 2024-03-27
Fix lookup to prevent finding `typedef` itself.

### LINKED ISSUE #3845: Allow typedef from global scope
Hi,

I am trying to typedef a type from a global namespace into a non-global namespace by writing:
```c++
struct Foo {};

namespace foo {
    typedef ::Foo Foo; // unexpected '::', expected identifier; works in dxc/hlsl
    //typedef Foo Foo; // cyclic reference 'Foo'; allowed in dxc/hlsl
}
```
Unfortunately it gives me this error in slang 2024.1.5:
```
shader.slang(4): error 20001: unexpected '::', expected identifier
    typedef ::Foo Foo;
            ^~

```
Is it possible to support the global scope syntax `::` for namespaces? Thanks!

csyonghe:
Fixes #3845.

The change here is to extend decl lookup functions to take a `declToExclude` parameter that excludes the specified decl from lookup result.
Before we check the type of `typedef Foo Foo`, we will set the typedef decl itself to be excluded from lookup, so that when we check the type expr of `Foo`, we won't mistakenly use the typedef itself as its own type.

### DIFF
--- a/source/slang/slang-ast-support-types.h
+++ b/source/slang/slang-ast-support-types.h
@@ -1421,6 +1421,9 @@ namespace Slang
         Scope*              scope       = nullptr;
         Scope*              endScope    = nullptr;
 
+        // A decl to exclude from the lookup, used to exclude the current decl being checked, such as in typedef Foo Foo;
+        // to avoid finding itself.
+        Decl* declToExclude = nullptr;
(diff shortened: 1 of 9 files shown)


jkwak-work (approved):
The code looks good.

PR #3849, opened 2024-03-27
nightly falcor perf run and vkcts weekly run

dzysk:
Adding a nightly run for falcor perf tests which will run the tests 10 times so we can take an average of the 10 runs to reduce any noise in the numbers. The data is then pushed to google sheets: https://docs.google.com/spreadsheets/d/17I-S9Ciy_Paz6emctPe4Iw28jX0bvL18TbaetybC80I


extras/falcor-perf-extract-results.py
```diff
@@ -0,0 +1,170 @@
...
+    frontend_execution_slang.append(float(rtn_json[results_id]["frontend-execution"]["slang"]))
+    spirv_generation_glslang.append(float(rtn_json[results_id]["spirv-generation"]["glslang"]))
+    spirv_generation_slang.append(float(rtn_json[results_id]["spirv-generation"]["slang"]))
+    spirv_compilation_glslang.append(float(rtn_json[results_id]["spirv-compilation"]["glslang"]))
+    spirv_compilation_slang.append(float(rtn_json[results_id]["spirv-compilation"]["slang"]))
+
+# In order to have charts in the spreadsheet that automatically update to the most recent
+# 30 test runs, we need to check if we already have 30 runs worth of data. Deleting and
+# re-inserting rows breaks the charts. So, instead we need to update the cells one by one.
+if len(recent_records) > 29:
```
csyonghe, line 100: Instead of deleting data, can we just preserve all the data and create charts from google sheets using only the last 30 rows? That way we can view last 30 days, last year etc. if we want.

dzysk: I couldn't find a way for google sheets to automatically use the last 30 rows. However, no data is lost. I have one sheet for the raw data from every run, one sheet that has all the averages, then there is the sheet for the last 30 runs. Only the sheet for the last 30 runs ever has data removed, but that data is still available on the other sheets.

csyonghe: I see, that sounds good!

csyonghe:
Closing and we should investigate using https://github.com/benchmark-action/github-action-benchmark for reporting.

PR #3850, opened 2024-03-27
Allow var/param names to be the same as type name.

### LINKED ISSUE #3824: Cyclic reference on member reusing namespace name
It would be valid HLSL or C++ to name a member after a namespace, like `example` of `Struct` here:
```hlsl
namespace example {
  struct Example {}
}

struct Struct {
  example::Example example;
}
```

But Slang v2024.1.5 complains:
```
shader.slang(6): fatal error 39999: cyclic reference 'example'.
  example::Example example;
                   ^~~~~~~
```

Is it supposed to be supported?

csyonghe:
Expand on #3848 to allow the variables, struct fields and function parameters to be named the same as their types.
The fix is the same: exclude the current var/param decl from the lookup when checking its type.

Fixes #3824.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -1547,7 +1547,8 @@ namespace Slang
             }
             else
             {
-                initExpr = CheckExpr(initExpr);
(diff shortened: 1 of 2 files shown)


jkwak-work (approved):
Simple enough

PR #3851, opened 2024-03-27
Implement 8.14-8.19 of OpenGL-GLSL specification; resolves #3692 for GLSL & SPIR-V targets

### LINKED ISSUE #3692: Implement miscellaneous GLSL functions to match to how much HLSL side works
There are GLSL functions described in [OpenGL Spec](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf) from a section 8.14 to 8.19.
Some of them may not have a HLSL counter part but we will need to implement as many as we can.

This is also a part of [the other umbrella task](https://github.com/shader-slang/slang/issues/3362).

Following functions are in those sections:

Section 8.14. Fragment Processing Functions
genFType dFdx(genFType p)
genFType dFdy(genFType p)
genFType dFdxFine(genFType p)
genFType dFdyFine(genFType p)
genFType dFdxCoarse(genFType p)
genFType dFdyCoarse(genFType p)
genFType fwidth(genFType p)
genFType fwidthFine(genFType p)
genFType fwidthCoarse(genFType p)
float interpolateAtCentroid(float interpolant)
vec2 interpolateAtCentroid(vec2 interpolant)
vec3 interpolateAtCentroid(vec3 interpolant)
vec4 interpolateAtCentroid(vec4 interpolant)
float interpolateAtSample(float interpolant, int sample)
vec2 interpolateAtSample(vec2 interpolant, int sample)
vec3 interpolateAtSample(vec3 interpolant, int sample)
vec4 interpolateAtSample(vec4 interpolant, int sample)
float interpolateAtOffset(float interpolant, vec2 offset)
vec2 interpolateAtOffset(vec2 interpolant, vec2 offset)
vec3 interpolateAtOffset(vec3 interpolant, vec2 offset)
vec4 interpolateAtOffset(vec4 interpolant, vec2 offset)

Section 8.15. Noise Functions
float noise1(genFType x) 
vec2 noise2(genFType x)
vec3 noise3(genFType x)
vec4 noise4(genFType x)

Section 8.16. Shader Invocation Control Functions
void barrier()

Section 8.17. Shader Memory Control Functions
void memoryBarrier()
void memoryBarrierAtomicCounter()
void memoryBarrierBuffer()
void memoryBarrierShared()
void memoryBarrierImage()
void groupMemoryBarrier()

Section 8.18. Subpass-Input Functions
gvec4 subpassLoad(gsubpassInput subpass)
gvec4 subpassLoad(gsubpassInputMS subpass, int sample)

Section 8.19. Shader Invocation Group Functions
bool anyInvocation(bool value)
bool allInvocations(bool value)
bool allInvocationsEqual(bool value)

ArielG-NV:
The following PR implements 8.14-8.19 of the [OpenGL-GLSL specification](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf).

Fully implements all functions and built-in type's, resolves https://github.com/shader-slang/slang/issues/3692 for GLSL & SPRI-V targets.

_Notes:_
Testing Tools:
* Fragment shaders cannot test computational results. Only OpCodes are checked for proper emitting.

Implementation Notes:
* SubpassInput requires an unknown image format. This error check has been added to `checkDecl` rather than `checkModifiers` since types need to be resolved for this error to be checked.
* `AtomicCounterMemory` memory-constraint requires `OpCapability AtomicStorage`, `AtomicStorage` is invalid for Vulkan targets. glslang outputs for `barrier`, `memoryBarrier`, and `groupMemoryBarrier` ``AtomicCounterMemory` as a memory constraint. This compiles as valid SPIR-V for Vulkan since `OpCapability AtomicStorage` is not declared. This behavior of glslang is undefined as per [3.31.Capability of the SPIR-V specification](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_capability). We will omit `AtomicCounterMemory` from our barrier calls.


source/slang/glsl.meta.slang
```diff
@@ -7263,3 +7263,704 @@ __glsl_version(430) [require(glsl)]
     }
     }
 }
+
+/// Section 8.14. Fragment Processing Functions
+
+[require(hlsl, sm_4_0)]
+__glsl_version(430) [require(glsl)]
+__spirv_version(1.0) [require(spirv)]
```
csyonghe, line 7271: need to require `fragment` capability here, since this is only available for fragment shaders?

source/slang/glsl.meta.slang
```diff
@@ -7263,3 +7263,704 @@ __glsl_version(430) [require(glsl)]
...
+    }
+    }
+}
+
+[require(hlsl, sm_5_0)]
+__glsl_version(430) [require(glsl)]
+__spirv_version(1.0) [require(spirv)]
+[__NoSideEffect]
+[GLSLRequireShaderInputParameter(0)]
+public float interpolateAtSample(__ref float interpolant, int sample)
```
csyonghe, line 7544: This is defined for fragment, geometry, and other graphics stages, but not vertex, compute, raygen, miss etc. stages, so we need to reflect this in capability requirements.

ArielG-NV: I did not add any capabilities for stages, I did not know these were in Slang, I will do this right-now

source/slang/hlsl.meta.slang
```diff
@@ -195,6 +195,14 @@ struct __ShapeBuffer : __ITextureShape
     static const int dimensions = 1;
     static const int planeDimensions = 1;
 }
+__magic_type(TextureShapeSubpassType)
```
csyonghe, line 198: Does this have to be in hlsl.meta.slang?

Any shape we add here is going to contirbute to a gigantic cross-product space of all texture type combinations and it is going to increase the size of the stdlib, decreasing the startup performance by noticeable amount, so we need to decide if it is worth it.

ArielG-NV: I will look to see if this type needs to be here

source/slang/hlsl.meta.slang
```diff
@@ -14030,3 +14038,289 @@ struct ConstBufferPointer
         }
     }
 }
+
+/// HLSL Fragment Processing Derivatives
+
+[__NoSideEffect]
+[ForceInline]
```
csyonghe, line 14045: This is already defined in hlsl.meta.slang:4215.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -3546,6 +3556,27 @@ struct ExprLoweringContext
         // TODO: also need to handle this-type substitution here?
     }
 
+    void validateInvokeExprArgsWithFunctionModifiers(
```
csyonghe, line 3559: If this modifier is only intend to be used inside stdlib and we know it is never going to be wrong, then we don't need to validate it. This will slow down the compiler unnecessarily.

source/slang/slang-parameter-binding.cpp
```diff
@@ -329,11 +329,17 @@ struct UsedRanges
...
+{
+    const constexpr size_t kInputAttachmentIndexUnused = 0xFFFFFFFF;
+};
+
 struct ParameterBindingInfo
 {
     size_t              space = 0;
     size_t              index = 0;
     LayoutSize          count = 0;
+    size_t              inputAttachmentIndex = ParameterAndLayoutConstants::kInputAttachmentIndexUnused;
```
csyonghe, line 342: We shouldn't be extending `ParameterBindingInfo` type here for something that is only used for a specific type of global var. What is this needed for? Can we just add a parameter cateogry for InputAttachment instead?

source/slang/slang-parameter-binding.cpp
```diff
@@ -1071,15 +1088,24 @@ static void addExplicitParameterBindings_GLSL(
...
     if( (foundResInfo = typeLayout->FindResourceInfo(LayoutResourceKind::DescriptorTableSlot)) != nullptr )
     {
-        // Try to find `binding` and `set`
-        if (auto attr = varDecl.getDecl()->findModifier<GLSLBindingAttribute>())
+        for (auto dec : varDecl.getDecl()->modifiers)
         {
-            resInfo = foundResInfo;
-            semanticInfo.index = attr->binding;
-            semanticInfo.space = attr->set;
+            // Try to find `binding` and `set`
```
csyonghe, line 1097: I think inputattachment index should exist as a ResourceInfo entry rather than as an augmentation to DescriptorTableSlot. 

InputAttachment comes from framebuffer and not a descriptor table, IIRC.

source/slang/slang-parameter-binding.cpp
ArielG-NV: input_attachment_index is part of the same conflict group as any other descriptor object (image binding 0, set 0, cannot be with a subpass binding 0, set 0). The difference is input_attachment_index is an additional conflict that has to be watched for.

source/slang/slang-parameter-binding.cpp
ArielG-NV: [input_attachment's are part of the renderpass, but they also need to be bound to a descriptor set and can have conflicts with other objects from what I see in LunarG's samples](https://github.com/LunarG/VulkanSamples/blob/master/API-Samples/input_attachment/input_attachment.cpp)

source/slang/slang-parameter-binding.cpp
csyonghe: I still don't understand why we can't juse use `binding` for `inputAttachmentIndex`, in what situation can they diverge?

ArielG-NV: _what each qualifier means:_

1. input_attachment_index references the index set when setting up renderpass's:
 ```
VkAttachmentReference renderPassAttachment{};
colorReference.attachment = 0;
```

2. binding & set is referencing the descriptor set and where we bound the renderpass's image view.

_why we need them both for error reporting:_
1. if we bound to `descriptor set 1, binding 0, attachment 0` and `descriptor set 0, binding 0, attachment 0`, we should warn since this is a conflict which [spec warns is illegal for input_attachment_index](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_vulkan_glsl.txt).
```
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor;
layout (input_attachment_index = 0, set = 1, binding = 0) uniform subpassInput inputColor2;
```

2. similarly, the same code here should conflict, but due to binding being the same, which is a problem since this is an impossible scenario to have (your descriptor is either for `input_attachment_index = 1` or `input_attachment_index = 0`, not both).
```
layout (input_attachment_index = 1, set = 0, binding = 0) uniform subpassInput inputColor;
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor2;
```

3. finally, since this is a regular descriptor, the following code will also conflict: due to an impossible scenario of having two resources which are different using the same descriptor bindings.
```
layout (set = 0, binding = 0) uniform image1D inputTexture1D;
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor;
```

_from what I understand on why specification defines things this way:_
1. Framebuffers are sometimes inside special memory regions. In this scenario, a input_attachment_index can 'directly' fetch a pointer to the memory.
2. Sometimes hardware stores a framebuffer as a regular image, this means to figure out where the image is in memory a descriptor can be used.

Comment (csyonghe) [reply source/slang/slang-parameter-binding.cpp:342]: It starts to sound to me like the variable `inputColor` should have two `ParameterBindingInfo` entries, one for DescriptorTableSlot, and one for InputAttachment.

Comment (csyonghe) [reply source/slang/slang-parameter-binding.cpp:342]: In `addExplicitParameterBindings_GLSL`, we have the opportunity to insert as many `ParameterBindingInfo` as we want based on the modifiers on the variable.

File: source/slang/hlsl.meta.slang (hunk 3 above)
Comment (ArielG-NV) [reply source/slang/hlsl.meta.slang:198]: in order to combat this issue I changed the texture inheritance hierarchy to minimize the scope of each generic (these changes reflect on the `__ITextureImpl` as well

`__ITextureShape` ->
    -> `__ITextureShapeUnique `
         -> `Subpass`
    -> `__ITextureShapeTraditional`
         -> `__ITextureShape1D2D3D`
              -> 1D, 2D etc...
         -> Buffer, Cube, etc....

File: source/slang/hlsl.meta.slang [hunk 8]
```diff
@@ -150,9 +150,36 @@ interface __ITextureShape
...
+__intrinsic_type($(kIROp_TextureShapeSubpassType))
+struct __ShapeSubpass : __ITextureShapeUnique
+{
+    static const int flavor = $(SLANG_TEXTURE_SUBPASS);
+    static const int dimensions = 4;
+    static const int planeDimensions = 4;
+}
+__magic_type(TextureType)
+__intrinsic_type($(kIROp_TextureType))
+struct __TextureImplUnique<T, Shape: __ITextureShapeUnique, let isArray:int, let isMS:int, let sampleCount:int, let access:int, let isShadow:int, let isCombined:int, let format:int>
```
csyonghe, line 170: I don't think we need to separate out __TextureImpl type into different types. Introducing the hierarchy for TextureShaoe should be sufficient.

source/slang/hlsl.meta.slang
```diff
@@ -150,9 +150,36 @@ interface __ITextureShape
     static const int dimensions;
     static const int planeDimensions;
 }
+
+//
 [sealed]
 [builtin]
-interface __ITextureShape1D2D3D : __ITextureShape
+interface __ITextureShapeUnique : __ITextureShape
```
csyonghe, line 157: Can we find a better name than Traditional and Unique? I don't understand what it means by "unique". If the only thing under IUnique is Subpass, then we don't need IUnique either.

csyonghe: I also don't understand why Subpass is modeled as a texture shape. How many intrinsics are we sharing with texture types? Can SubpassInput be a standalone type independent of textures?

source/slang/hlsl.meta.slang
```diff
@@ -13298,8 +13310,8 @@ ${
 // further clutter the original type declarations.
 }
 
-__generic<T, Shape: __ITextureShape, let sampleCount:int, let isShadow:int, let format:int>
-extension __TextureImpl<T,Shape,0,0,sampleCount,0,isShadow,0,format>
+__generic<T, Shape: __ITextureShapeTraditional, let sampleCount:int, let isShadow:int, let format:int>
```
csyonghe, line 13313: Looking here I think we can keep our existing TextureImpl unchanged. The :__ITextureShapeTraditional constraint will already exclude the Subpass shape.

source/slang/slang-check-modifier.cpp
```diff
@@ -387,6 +387,23 @@ namespace Slang
 
             anyValueSizeAttr->size = int32_t(value->getValue());
         }
+        else if (auto GLSLRequireShaderInputParameter = as<GLSLRequireShaderInputParameterAttribute>(attr))
```
csyonghe, line 390: glslRequireShader...

source/slang/slang-parameter-binding.cpp
```diff
@@ -1072,14 +1072,29 @@ static void addExplicitParameterBindings_GLSL(
...
+            if (auto glslBindingAttr = as<GLSLBindingAttribute>(dec))
+            {
+                resInfo = foundResInfo;
+                semanticInfo.index = glslBindingAttr->binding;
+                semanticInfo.space = glslBindingAttr->set;
+            }
+            // Try to find `input_attachment_index`
+            if (auto glslAttachmentIndexAttr = as<GLSLInputAttachmentIndexLayoutModifier>(dec))
```
csyonghe, line 1090: else if

source/slang/slang-parameter-binding.cpp
```diff
@@ -1072,14 +1072,29 @@ static void addExplicitParameterBindings_GLSL(
...
+            {
+                // Subpass fills semantic info of a descriptor & subpass
+                subpassSemanticInfo.index = stringToInt(glslAttachmentIndexAttr->valToken.getContent());
```
csyonghe, line 1093: We already converted this to int when we check the modifier, right?

source/slang/slang-parser.cpp
```diff
@@ -8285,7 +8292,7 @@ namespace Slang
         _makeParseModifier("sample",        HLSLSampleModifier::kReflectClassInfo),
         _makeParseModifier("centroid",      HLSLCentroidModifier::kReflectClassInfo),
         _makeParseModifier("precise",       PreciseModifier::kReflectClassInfo),
-        _makeParseModifier("shared",        HLSLEffectSharedModifier::kReflectClassInfo),
+        _makeParseModifier("shared",        parseSharedModifier),
```
csyonghe, line 8295: Is this change still needed?

source/slang/slang-parameter-binding.cpp
```diff
@@ -1113,6 +1128,9 @@ static void addExplicitParameterBindings_GLSL(
         semanticInfo.kind = kind;
 
         addExplicitParameterBinding(context, parameterInfo, varDecl.getDecl(), semanticInfo, count);
+        if (foundSubpass)
```
csyonghe, line 1131: Do we actually consume this info?

source/slang/hlsl.meta.slang
ArielG-NV: I can separate it out; it works as a TextureType logically, but I suppose there is no need to keep it as one.

source/slang/slang-parameter-binding.cpp
ArielG-NV: it is stored as a int literal token (all GLSLLayoutModifier's are)

I wonder if this should be changed though, its significantly slower to call stringToInt 3+ times in a program than it is to cast a int type to another int

source/slang/slang-parser.cpp
ArielG-NV: it is not needed

source/slang/slang-parameter-binding.cpp
ArielG-NV: we do use foundSubpass, it is how the overlapping system now functions

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -363,9 +368,15 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
         {
             format = decor->getFormat();
         }
+        if (textureType->GetBaseShape() == SLANG_TEXTURE_SUBPASS)
```
csyonghe, line 371: shouldn't be needed anymore

source/slang/hlsl.meta.slang
```diff
@@ -11406,15 +11406,22 @@ ${{{{
...
-
-struct VkSubpassInputMS<T>
+__magic_type(TextureShapeSubpassType)
+__intrinsic_type($(kIROp_TextureShapeSubpassType))
+struct __ShapeSubpass : __ITextureSubpass
+{
+    static const int flavor = $(SLANG_TEXTURE_SUBPASS);
+}
+__magic_type(TextureType)
+__intrinsic_type($(kIROp_TextureType))
```
csyonghe, line 11422: Why does it need to be TextureType?

Can we use a separate type for ast and IR, so we can get rid of this unnecessary generic arguments that are irrelevant to this type.

That will also allow us to get rid of the ITextureShapeSubpass interface.

csyonghe:
Let's implement the SubpassInput<T>​ type as dxc does, and represent it as a separate magic type and IR type that happens to also emit as OpTypeImage. This should clean up the stdlib a lot.
________________________________

source/slang/slang-parameter-binding.cpp
jkwak-work: Can you be more specific about the descriptor set? You can not expect us to understand what you meant by giving us a link to a source file. And please give us more explanation about the "conflict" you were talking about.

ArielG-NV: The following comment was addressed in connection with other suggestions, a new ResourceInfo entry was added. A deeper explanation was provided to answer another comment, and will be provided (added to slightly) below:

_what each qualifier means:_

1. input_attachment_index references the index set when setting up renderpass's:
 ```
VkAttachmentReference renderPassAttachment{};
colorReference.attachment = 0;
```

2. binding & set is referencing the descriptor set and where we bound the renderpass's image view.
```
    VkDescriptorSetLayoutBinding layout_bindings[1];
    layout_bindings[0].binding = 0;
    layout_bindings[0].descriptorType = VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT;
    layout_bindings[0].descriptorCount = 1;
    layout_bindings[0].stageFlags = VK_SHADER_STAGE_FRAGMENT_BIT;
    layout_bindings[0].pImmutableSamplers = NULL;
```

_why we need them both for error reporting:_
1. if we bound to `descriptor set 1, binding 0, attachment 0` and `descriptor set 0, binding 0, attachment 0`, we should warn since this is a conflict which [spec warns is illegal for input_attachment_index](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_vulkan_glsl.txt).
```
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor;
layout (input_attachment_index = 0, set = 1, binding = 0) uniform subpassInput inputColor2;
```

2. similarly, the same code here should conflict, but due to binding being the same, which is a problem since this is an impossible scenario to have (your descriptor is either for `input_attachment_index = 1` or `input_attachment_index = 0`, not both).
```
layout (input_attachment_index = 1, set = 0, binding = 0) uniform subpassInput inputColor;
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor2;
```

3. finally, since this is a regular descriptor, the following code will also conflict: due to an impossible scenario of having two resources which are different using the same descriptor bindings.
```
layout (set = 0, binding = 0) uniform image1D inputTexture1D;
layout (input_attachment_index = 0, set = 0, binding = 0) uniform subpassInput inputColor;
```

File: slang.h [hunk 18]
```diff
@@ -2251,13 +2253,17 @@ extern "C"
         // The register space offset for the sub-elements that occupies register spaces.
         SLANG_PARAMETER_CATEGORY_SUB_ELEMENT_REGISTER_SPACE,
 
+        // The input_attachment_index subpass occupancy tracker
+        SLANG_PARAMETER_CATEGORY_SUBPASS,
+
         //
-        SLANG_PARAMETER_CATEGORY_COUNT,
+        SLANG_PARAMETER_CATEGORY_COUNT_REV2,
```
csyonghe, line 2260: Is this necessary? What happens if we just update the value of `SLANG_PARAMETER_CATEGORY_COUNT`?

source/slang/glsl.meta.slang
```diff
@@ -7263,3 +7263,700 @@ __glsl_version(430) [require(glsl)]
     }
     }
 }
+
+/// Section 8.14. Fragment Processing Functions
+
+__glsl_version(430)
+__spirv_version(1.0)
+[require(any_gfx_target, fragmentprocessing)]
```
csyonghe, line 7271: Can this be just `[require(any_gfx_target, fragment)]`

source/slang/glsl.meta.slang
```diff
@@ -7263,3 +7263,700 @@ __glsl_version(430) [require(glsl)]
...
+    __target_switch
+    {
+    case hlsl: __intrinsic_asm "AllMemoryBarrier()";
+    case glsl: __intrinsic_asm "memoryBarrier()";
+    case spirv:
+    {
+        spirv_asm
+        {
+            OpCapability Shader;
+            OpMemoryBarrier Invocation AcquireRelease|WorkgroupMemory|ImageMemory|UniformMemory
```
csyonghe, line 7744: Should this be `Invocation` or `Device`? What's does glslang use?

source/slang/glsl.meta.slang
```diff
@@ -7263,3 +7263,700 @@ __glsl_version(430) [require(glsl)]
...
+    [require(glsl_spirv, subpass)]
+    public typealias $(subpassLoadEntry.prefix)subpassInput = __SubpassImpl<
+        $(subpassLoadEntry.typeName),
+        0 // isMS
+    >;
+
+    [require(glsl_spirv, subpass)]
+    [__NoSideEffect]
+    [ForceInline]
+    public $(subpassLoadEntry.typeName) subpassLoad($(subpassLoadEntry.prefix)subpassInput subpass)
```
csyonghe, line 7883: Can we define this as a single generic function, instead of duplicating it in the meta `for` loop?

source/slang/hlsl.meta.slang
```diff
@@ -11404,19 +11404,64 @@ ${{{{
...
+struct __SubpassImpl<T, let isMS:int>
 {
-    T SubpassLoad();
 }
 
-struct VkSubpassInputMS<T>
+__generic<T, let isMS:int=0>
+typealias SubpassInputMS = __SubpassImpl<T, isMS>;
+
+__generic<T, let isMS:int=0>
```
csyonghe, line 11419: Default generic argument value in an extension have no meaning.

csyonghe: I believe things in this extension can be defined directly in the original `struct` type. If you intend to provide `SubpassLoad` for `isMS = 0` only, you should write:
```
__generic<T>
extension __SubpassImpl<T, 0>
{
    // ...
}
```

source/slang/hlsl.meta.slang
```diff
@@ -11404,19 +11404,64 @@ ${{{{
...
+        }
+        }
+    }
 }
 
+__generic<T, let isMS:int=1>
+typealias SubpassInput = __SubpassImpl<T, isMS>;
+
+__generic<T, let isMS:int=1>
+extension __SubpassImpl<T, isMS>
```
csyonghe, line 11445: Again, I don't know if the intention is to provide the function for `isMS=1` only, if so this needs to be changed too.

source/slang/slang-capabilities.capdef
```diff
@@ -385,4 +395,4 @@ alias DX_6_3 = sm_6_3;
 alias DX_6_4 = sm_6_4;
 alias DX_6_5 = sm_6_5;
 alias DX_6_6 = sm_6_6;
-alias DX_6_7 = sm_6_7;
+alias DX_6_7 = sm_6_7;
```
csyonghe, line 398: This change is unnecessary.

source/slang/slang-ir-any-value-marshalling.cpp
```diff
@@ -223,7 +223,9 @@ namespace Slang
                 break;
             }
             default:
-                if (as<IRTextureTypeBase>(dataType) || as<IRSamplerStateTypeBase>(dataType))
+                if (as<IRTextureTypeBase>(dataType)
```
csyonghe, line 226: Can this be `isResourceType(dataType`?

source/slang/slang-ir-any-value-marshalling.cpp
```diff
@@ -855,7 +857,9 @@ namespace Slang
             return alignUp(offset, 4) + alignUp((SlangInt)anyValueSize, 4);
         }
         default:
-            if (as<IRTextureTypeBase>(type) || as<IRSamplerStateTypeBase>(type))
+            if (as<IRTextureTypeBase>(type)
```
csyonghe, line 860: `isResourceType`

source/slang/slang-ir-specialize-resources.cpp
```diff
@@ -1250,6 +1252,8 @@ bool isIllegalSPIRVParameterType(IRType* type, bool isArray)
     // all Texture types.
     if (as<IRTextureType>(type))
         return true;
+    if(as<IRSubpassInputType>(type))
```
csyonghe, line 1255: Not needed because it is already covered by `isIllegalGLSLParameterType`.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -247,6 +247,11 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
             }
             else if(const auto spirvAsmOperand = as<IRSPIRVAsmOperandInst>(user))
             {
+                // Skip load's for referenced `Input` variables since a ref implies
+                // passing as is, which needs to be a pointer (pass as is).
+                if (user->getDataType()->getOp() == kIROp_RefType
```
csyonghe, line 252: user->getDataType() might be null, probably better to be defensive about it.

slang.h
csyonghe: My preference is we don't introduce this complexity unless absolutely necessary.

ArielG-NV: `SLANG_PARAMETER_CATEGORY_COUNT` changing value broke a Falcor test. 

I have no preference on solution if another also solves the issue.

source/slang/glsl.meta.slang
ArielG-NV: device should be used as memory scope👍

source/slang/glsl.meta.slang
ArielG-NV: I would assume not since we need a way to enforce glsl/hlsl versioning requirements?

slang.h
csyonghe: Maybe keep the actual count named as count, and the old one named as COUNT_V1.

ArielG-NV: I will check if this works.

source/slang/slang-ir-validate.cpp
```diff
@@ -223,6 +223,15 @@ namespace Slang
         {
         case kIROp_DifferentiableTypeDictionaryItem:
             return;
+        default:
+
+            // It is possible that a global parameter will not emit their ancestor until after
```
csyonghe, line 228: I don't quite understand why we are hitting a new exception scenario that is not covered previously, and I am not sure this is the correct workaround. Can you describe in detail what are the issue we are working around here?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -2204,6 +2204,16 @@ void addVarDecorations(
...
                 builder->getIntValue(builder->getIntType(), stringToInt(glslLocationMod->valToken.getContent())));
         }
+        else if (auto glslInputAttachmentMod = as<GLSLInputAttachmentIndexLayoutModifier>(mod))
+        {
+            auto subpassType = as<IRSubpassInputType>(inst->getDataType());
+
+            if (!subpassType)
+                context->getSink()->diagnose(inst, Diagnostics::InputAttachmentIndexOnlyAllowedOnSubpass);
+
+            builder->addDecoration(inst, kIROp_GLSLInputAttachmentIndexDecoration,
```
csyonghe, line 2214: We should fold the arg expr into a Val and lower that Val here. This will allow users to use an expr instead of just a literal. We need to allow other modifiers to take exprs instead of literals eventually.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -3546,6 +3556,21 @@ struct ExprLoweringContext
         // TODO: also need to handle this-type substitution here?
     }
 
+    void validateInvokeExprArgsWithFunctionModifiers(
+        InvokeExpr* expr,
+        FunctionDeclBase* decl,
+        List<IRInst*>& irArgs)
+    {
+        if (auto glslRequireShaderInputParameter = decl->findModifier<GLSLRequireShaderInputParameterAttribute>())
```
csyonghe, line 3564: Is this attribute something that the user will write? If not we can skip the validation.

source/slang/core.meta.slang
```diff
@@ -2604,3 +2604,6 @@ attribute_syntax [KnownBuiltin(name : String)] : KnownBuiltinAttribute;
 
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [NonUniformReturn] : NonDynamicUniformAttribute;
+
+__attributeTarget(FunctionDeclBase)
```
csyonghe, line 2608: Prefix with __ if not intended to be used directly by user.

source/slang/slang-lower-to-ir.cpp
ArielG-NV: Invoke arguments may need to be validated that `in` objects are being used. Some functions requires `in` objects as a parameter

source/slang/slang-lower-to-ir.cpp
ArielG-NV: Should this be done now or in a separate commit where the entirety of `static NodeBase* parseLayoutModifier(Parser* parser, void* /*userData*/)` (and related) is modified to allow expr input?

Currently `GLSLInputAttachmentIndexDecoration` parses through use of the `GLSLLayoutModifier` which most layout attributes resolve from.

csyonghe: Let's do it in another PR then.

source/slang/slang-ir-validate.cpp
ArielG-NV: The workaround to emit `in` parameters directly where called as arguments relies on the following steps:
```
                // Add an entry point parameter for all the inputs.
                auto firstBlock = entryPointFunc->getFirstBlock();
                builder.setInsertInto(firstBlock);
                auto inputParam = builder.emitParam(inputStructType);
                builder.addLayoutDecoration(inputParam, paramLayout);
```
emits a kIROp_Param for each eventual `GlobalParam`. 

since these kIROp_Param are the actual GlobalParam and are treated as such in the code, we can do the same here
```
                ...
                // Replace all global variable references with
                // actual global parameter for all Call's
                for (Index i = 0; i < inputVars.getCount(); i++)
                {
                            ...
                            auto field = builder.emitFieldExtract(inputType, inputParam, inputKeys[i]);
                            ...
```

The downside of doing this is that during `validateIRInstOperand` the kIROp_Param will not be in scope since it is a entryPoint parameter (different function). Only until later in compilation are all entryPoint's transformed into a GlobalParam in global scope.

csyonghe: I think we should avoid changing the general IR validation logic just to allow us violating it, even if we eventually are correcting it, because this is not a scalable design. If somebody adds in a pass between where we violate the rule and where we correct the IR, that pass can fail/crash for reasons suprising to the author. Is there another way to get things to work without this exception?

csyonghe: It seems that this replacement step kicks in too early. We can consider moving it to after the parameter is turned into a global parameter again.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -2732,6 +2732,49 @@ void legalizeEntryPointParameterForGLSL(
...
+                    if (user->getOp() != kIROp_Call) continue;
+                    for (Slang::UInt operandIndex = 0; operandIndex < user->getOperandCount(); 
+                        operandIndex++)
+                    {
+                        auto operand = user->getOperand(operandIndex);
+                        auto operandUse = user->getOperands() + operandIndex;
+                        if (operand != globalVarToReplace)
+                            continue;
+                        replaceBuilder.setInsertBefore(user);
+                        auto structMade = materializeValue(builder, globalValue);
```
csyonghe, line 2770: isn't this `materialized`?

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -2732,6 +2732,49 @@ void legalizeEntryPointParameterForGLSL(
         IRInst* materialized = materializeValue(builder, globalValue);
 
         pp->replaceUsesWith(materialized);
+
+        // We finally need to replace all global variable references of a global 
+        // parameter with the actual global parameter for all function calls. 
+        // Global parameters are used with a OpStore to copy its data into a global 
+        // variable intermediary. We will follow the uses of a global parameter until 
+        // we find this OpStore, then we will replace uses of the intermediary object. 
+        IRBuilder replaceBuilder(materialized);
```
csyonghe, line 2741: Reading the code here, it seems the logic is:
1. Find the associating global var of `globalValue`, by searching for a Store into the var where the value is `materialized`.
2. Replace  use of globalVarToReplace with `materialized`.

I suggest getting rid of the shacky step 1 with a decoration put on the global var that references back the global parameter when we created that global var initially. Then this logic will be clean and robust.

source/slang/slang-ir-glsl-legalize.cpp
ArielG-NV: `materialized` is local to a function block (specifically the entry point), the code will work if we use it, but I think this is by pure chance since this would imply no visible ancestor?

source/slang/slang-ir-glsl-legalize.cpp
ArielG-NV: I think I will put the decoration on the kIROp_Param that becomes a kIROp_GlobalParam since it means we don't need some search through all GlobalInst's, same principal though.

the life time of variables to explain what I am talking about:
1. we use a `in` variable somewhere in early parsing and expression generation --> make `global variable`.

2. `GlobalVarTranslationContext` --> we make kIROp_Param & we relate kIROp_Param to a `global variable`.

3. `legalizeRayTracingEntryPointParameterForGLSL` --> we make kIROp_GlobalParam from a kRIOp_Param. We also replace uses of a `global variable` here. We know which kIROp_Param we have a connection of the variable from step `2`'.

csyonghe: Yes, this is a good solution. We should be decorating IRParam to avoid violating the rules.

csyonghe (approved):

PR #3852, opened 2024-03-28
Fix type union logic in generic type inference.

csyonghe:


jkwak-work (approved):
This looks perfect and it doesn't have any hacks we talked about.

However, kConversionCost_IntegerToHalfConversion is too high that a following example wouldn't work as expected. It should be below 550.
```
    half val = 1.h;
    half result = 1 - val;
```
When the conversion-cost from int to half is exactly at 550, I get the following error,
```
core.meta.slang(1834): note 39999: candidate: func -(double, double) -> double
core.meta.slang(1826): note 39999: candidate: func -(float, float) -> float
core.meta.slang(1818): note 39999: candidate: func -(half, half) -> half
```
In order for it to work as half, the conversion-cost from int to half should be below 550.

Note that the conversion-cost from int to float/double is 400.
The conversion-cost from half to float/double is 150 as kConversionCost_RankPromotion.

jkwak-work (approved):
Looks good!

PR #3857, opened 2024-03-29
Improve compile time performance.

csyonghe:
1. Add caching for implicit conversion overload resolution results.
2. Handle invalidation of type checking cache upon extension registration more gracefully.
3. Add a setting to `simplifyIR` to disable `redundancyRemoval` pass (still enabled by default).

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2153,7 +2153,7 @@ ${{{{
     {
 }}}}
     __intrinsic_op($(kIROp_CombinedTextureSamplerGetTexture))
-    __TextureImpl<T, Shape, isArray, isMS, sampleCount, 0, isShadow, 0, format> __getTexture();
+    __TextureImpl<T, Shape, isArray, 0, sampleCount, 0, isShadow, 0, format> __getTexture();
 
     __intrinsic_op($(kIROp_CombinedTextureSamplerGetSampler))
     SamplerState __getSampler();
--- a/source/slang/slang-check-impl.h
+++ b/source/slang/slang-check-impl.h
@@ -554,6 +554,7 @@ namespace Slang
     {
         OverloadCandidate conversionFuncOverloadCandidate;
         ConversionCost cost = kConversionCost_Impossible;
+        bool isAmbiguous = false;
     };
 
     struct ImplicitCastMethodKey
@@ -577,7 +578,7 @@ namespace Slang
             , toType(toType)
             , constantVal(0)
             , isConstant(false)
-            , isLValue(fromExpr?fromExpr->type.isLeftValue:false)
+            , isLValue(fromType.isLeftValue)
         {
             if (auto constInt = as<IntegerLiteralExpr>(fromExpr))
             {
--- a/source/slang/slang-check-impl.h
+++ b/source/slang/slang-check-impl.h
@@ -553,6 +553,7 @@ namespace Slang
     struct ImplicitCastMethod
     {
         OverloadCandidate conversionFuncOverloadCandidate;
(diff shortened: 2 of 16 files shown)


tests/language-feature/extensions/extension-method-simple.slang
```diff
@@ -0,0 +1,30 @@
+// interface-extension.slang
+
+// Test that an `extension` applied to an interface type works as users expect
```
jkwak-work, line 3: These comments are little confusing.
I assume that they are copied from another file?
If yes, you may need to modify these.

source/slang/slang-check-impl.h
```diff
@@ -541,6 +541,45 @@ namespace Slang
...
+        {
+            return fromType == other.fromType && toType == other.toType && isConstant == other.isConstant && constantVal == other.constantVal && isLValue == other.isLValue;
+        }
+        ImplicitCastMethodKey() = default;
+        ImplicitCastMethodKey(QualType fromType, Type* toType, Expr* fromExpr)
+            : fromType(fromType)
+            , toType(toType)
+            , constantVal(0)
+            , isConstant(false)
+            , isLValue(fromType.isLeftValue)
```
jkwak-work, line 573: This is a nitpick.
It will be better if "isLValue" is initialized in between "fromType" and "toType" because that's the order they are declared.

tests/language-feature/extensions/extension-method-simple.slang
csyonghe: Good catch, thanks!

jkwak-work (approved):
Looks good

jkwak-work:
I have some timing numbers here.

Without this change, slangc takes 47.19s ~ 48.18s for compiling 40 TSR shaders.
With this change, slangc takes slightly less time, 46.19s ~ 47.71s for the same TSR shaders.

Without this change, FXC took 90.28s ~ 91.69s for compiling 6 TSR shaders.
With this change, FXC took little longer, 94.42s ~ 95.10s for the same TSR shaders.

I noticed that with this changes, the number of HLSL lines generated from Slangc is about 3% less.
I am not sure if it is expected.

csyonghe:
What about the perf numbers from dxc?

jkwak-work:
I am sorry. I thought I was using FXC but it was DXC.
All of the data I shared here are based on DXC.

jkwak-work (approved):
Code looks good

csyonghe:
Given your measurements, I am not sure if we should be merging these changes, since it doesn't seem to be a net win.

csyonghe:
Update: the changes here resulted 20% speedup in frontend when processing typical shader workloads, so I am going to merge it.

source/slang/hlsl.meta.slang
```diff
@@ -2153,7 +2153,7 @@ ${{{{
     {
 }}}}
     __intrinsic_op($(kIROp_CombinedTextureSamplerGetTexture))
-    __TextureImpl<T, Shape, isArray, isMS, sampleCount, 0, isShadow, 0, format> __getTexture();
+    __TextureImpl<T, Shape, isArray, 0, sampleCount, 0, isShadow, 0, format> __getTexture();
```
jkwak-work, line 2156: It seems like I made a mistake here.
I wonder how this has been working?
I don't see where `isMS` is declared.

jkwak-work (approved):

source/slang/hlsl.meta.slang
csyonghe: I was confused too, not sure why it just compiled prior to this change, must be some other bugs that I didn't have time to track down.

PR #3858, opened 2024-03-29
Fix assertions due to malformed switch statements

sriramm-nv:
The change looks for duplicate default cases in a switch statement, and flag them as invalid

### DIFF
--- a/source/slang/slang-check-stmt.cpp
+++ b/source/slang/slang-check-stmt.cpp
@@ -239,6 +239,48 @@ namespace Slang
         subContext.checkStmt(stmt->body);
     }
 
+    void SemanticsStmtVisitor::validateCaseStmts(SwitchStmt* stmt, DiagnosticSink* sink)
+    {
+        auto blockStmt = as<BlockStmt>(stmt->body);
+        if (!blockStmt)
+            return;
+
+        auto seqStmt = as<SeqStmt>(blockStmt->body);
+        if (!seqStmt)
+            return;
+
+        bool hasDefaultStmt = false;
+        HashSet<Val*> caseStmtVals;
+        for (auto& sStmt : seqStmt->stmts)
+        {
+            // check that all case tags are unique
+            if (auto caseStmt = as<CaseStmt>(sStmt))
+            {
+                if (caseStmt->exprVal)
+                {
+                    // exprVal contains the constant folded expr, that is checked for
+                    // uniqueness within the scope of the switch statement.
+                    if (!caseStmtVals.add(caseStmt->exprVal))
+                    {
+                        sink->diagnose(sStmt, Diagnostics::duplicateCasesInASwitch);
+                        return;
+                    }
+                }
+            }
+
+            // check that there is at most one `default` clause
+            else if (auto defaultStmt = as<DefaultStmt>(sStmt))
+            {
+                if (hasDefaultStmt)
+                {
+                    sink->diagnose(sStmt, Diagnostics::multipleDefaultsInASwitch);
+                    return;
+                }
+                hasDefaultStmt = true;
+            }
+        }
+    }
+
     void SemanticsStmtVisitor::visitSwitchStmt(SwitchStmt* stmt)
     {
         WithOuterStmt subContext(this, stmt);
@@ -247,16 +289,16 @@ namespace Slang
         stmt->condition = CheckExpr(stmt->condition);
         subContext.checkStmt(stmt->body);
 
-        // TODO(tfoley): need to check that all case tags are unique
-
-        // TODO(tfoley): check that there is at most one `default` clause
+        // check the case value exits within the switch
+        validateCaseStmts(stmt, getSink());
     }
 
     void SemanticsStmtVisitor::visitCaseStmt(CaseStmt* stmt)
     {
-        // TODO(tfoley): Need to coerce to type being switch on,
-        // and ensure that value is a compile-time constant
         auto expr = CheckExpr(stmt->expr);
+
+        // coerce to type being switch on, and ensure that value is a compile-time constant
+        auto exprVal = tryConstantFoldExpr(expr, ConstantFoldingKind::CompileTime, nullptr);
         auto switchStmt = FindOuterStmt<SwitchStmt>();
 
         if (!switchStmt)
@@ -270,6 +312,7 @@ namespace Slang
         }
 
         stmt->expr = expr;
+        stmt->exprVal = exprVal;
         stmt->parentStmt = switchStmt;
     }
 
--- a/source/slang/slang-check-stmt.cpp
+++ b/source/slang/slang-check-stmt.cpp
@@ -253,27 +253,26 @@ namespace Slang
         HashSet<Val*> caseStmtVals;
         for (auto& sStmt : seqStmt->stmts)
         {
-            // check that all case tags are unique
             if (auto caseStmt = as<CaseStmt>(sStmt))
             {
+                // check that all case tags are unique
                 if (caseStmt->exprVal)
                 {
                     // exprVal contains the constant folded expr, that is checked for
                     // uniqueness within the scope of the switch statement.
                     if (!caseStmtVals.add(caseStmt->exprVal))
                     {
-                        sink->diagnose(sStmt, Diagnostics::duplicateCasesInASwitch);
+                        sink->diagnose(sStmt, Diagnostics::switchDuplicateCases);
                         return;
                     }
                 }
             }
-
-            // check that there is at most one `default` clause
             else if (auto defaultStmt = as<DefaultStmt>(sStmt))
             {
+                // check that there is at most one `default` clause
                 if (hasDefaultStmt)
                 {
-                    sink->diagnose(sStmt, Diagnostics::multipleDefaultsInASwitch);
+                    sink->diagnose(sStmt, Diagnostics::switchMultipleDefault);
                     return;
                 }
                 hasDefaultStmt = true;
@@ -298,6 +297,8 @@ namespace Slang
         auto expr = CheckExpr(stmt->expr);
 
         // coerce to type being switch on, and ensure that value is a compile-time constant
+        // The Vals in the AST are pointer-unique, making them easy to check for duplicates
+        // by addeing them to a HashSet.
         auto exprVal = tryConstantFoldExpr(expr, ConstantFoldingKind::CompileTime, nullptr);
         auto switchStmt = FindOuterStmt<SwitchStmt>();
 
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -476,6 +476,9 @@ DIAGNOSTIC(30504, Warning, forLoopTerminatesInFewerIterationsThanMaxIters, "the 
 DIAGNOSTIC(30505, Warning, loopRunsForZeroIterations, "the loop runs for 0 iterations and will be removed.")
 DIAGNOSTIC(30510, Error, loopInDiffFuncRequireUnrollOrMaxIters, "loops inside a differentiable function need to provide either '[MaxIters(n)]' or '[ForceUnroll]' attribute.")
 
+DIAGNOSTIC(30600, Error, switchMultipleDefault, "multiple 'default' cases not allowed within a 'switch' statement")
+DIAGNOSTIC(30601, Error, switchDuplicateCases, "duplicate cases not allowed within a 'switch' statement")
+
 // TODO: need to assign numbers to all these extra diagnostics...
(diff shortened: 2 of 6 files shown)


source/slang/slang-check-stmt.cpp
```diff
@@ -239,17 +239,57 @@ namespace Slang
...
         WithOuterStmt subContext(this, stmt);
 
         // TODO(tfoley): need to coerce condition to an integral type...
         stmt->condition = CheckExpr(stmt->condition);
-        subContext.checkStmt(stmt->body);
-
-        // TODO(tfoley): need to check that all case tags are unique
 
-        // TODO(tfoley): check that there is at most one `default` clause
+        auto seqStmt = validateCaseStmts(stmt, getSink());
```
csyonghe, line 291: validate should happen after checking the stmt->body, so when you validate the case value exits, they are already type checked.

csyonghe (changes requested):

source/slang/slang-check-stmt.cpp
```diff
@@ -239,17 +239,57 @@ namespace Slang
...
+            return nullptr;
+
+        bool hasDefaultStmt = false;
+        HashSet<Val*> caseStmtVals;
+        for (auto& sStmt : seqStmt->stmts)
+        {
+            // check that all case tags are unique
+            if (auto caseStmt = as<CaseStmt>(sStmt))
+            {
+                if (auto caseVal = tryConstantFoldExpr(caseStmt->expr, ConstantFoldingKind::CompileTime, nullptr))
```
csyonghe, line 259: When we visit case stmt, we can fold the constant int to a Val and store that Val on the case stmt. then when we are here, we can simply grab the folded Val from the checked caseStmt.

Currently our compiler cannot fold a enum value as a IntVal, but we will plan to support that soon. When that is supported, we should also be diagnosing an error if a case value cannot be folded into Val.

source/slang/slang-check-stmt.cpp
sriramm-nv: ah i see. in that case, i see no reason to return the seqStmt, so i changed the validate to a void function.

source/slang/slang-check-stmt.cpp
sriramm-nv: could you please elaborate on the "store that Val on the case stmt" ?
Case stmt has a member expr, which has a value, but it is private.

csyonghe: image

I mean add a `Val* exprVal` field in `CaseStmt` class.
Then in `SemanticsStmtVisitor::visitCaseStmt`, try fold `stmt->expr` into `Val` and store it in `stmt->exprVal`.

Note the TODO in `SemanticsStmtVisitor::visitCaseStmt`:
```
        // TODO(tfoley): Need to coerce to type being switch on,
        // and ensure that value is a compile-time constant
```
Here we will have the opportunity to issue an error when folding into IntVal failed.

If you do so then in `SemanticsStmtVisitor::visitSwitchStmt`, the case stmt would have been checked after `subContext.checkStmt(stmt->body);`, and you should be able to grab the caseStmt's `exprVal` and add it to the hash set.

sriramm-nv: thanks Yong. Was unsure about adding a new member  to the CaseStmt.

source/slang/slang-check-stmt.cpp
```diff
@@ -239,6 +239,48 @@ namespace Slang
...
+        HashSet<Val*> caseStmtVals;
+        for (auto& sStmt : seqStmt->stmts)
+        {
+            // check that all case tags are unique
+            if (auto caseStmt = as<CaseStmt>(sStmt))
+            {
+                // all constant folded exprs are checked for uniqueness
+                if (caseStmt->exprVal)
+                {
+                    if (caseStmtVals.contains(caseStmt->exprVal))
```
csyonghe, line 262: if (!caseStmtVd.add(...))


This avoids double dictionary lookups.

tests/diagnostics/switch-mutable-cases.slang
```diff
@@ -0,0 +1,16 @@
...
+
+// Tests to evaluate the behavior of code blocks within a switch statement. A switch statement with duplicate cases with same values is not allowed and should throw an error
+
+void test()
+{
+    int a = 1;
+    int b = 2;
+    switch (0) {
+        // CHECK: ([[# @LINE+1]]): error 39999: expression {{.*}}
+        case a: break;
```
csyonghe, line 11: case a shouldn't be a valid input code.

Should be case 0.

In the future we should add a diagnostic if Folding to Val failed.

sriramm-nv: this was a contrived test case that is to catch the case when const-folding to int val would fail. This is automatically caught by `slang-check-expr.cpp:CheckIntegerConstantExpression`, which is why I didn't want to add another diagnostic in `visitSwitchStmt`.

If I have const int for cases, then it wouldn't fail at `tryFoldIntegerConstantExpression` within that function.

do you want me to remove this test case, or can we leave it as a negative test?

csyonghe: OK, let's leave it here.

csyonghe:
You may want to sync this branch with top of tree to see if that fixes some errors we are seeing here.

source/slang/slang-check-stmt.cpp
```diff
@@ -247,16 +288,17 @@ namespace Slang
...
     }
 
     void SemanticsStmtVisitor::visitCaseStmt(CaseStmt* stmt)
     {
-        // TODO(tfoley): Need to coerce to type being switch on,
-        // and ensure that value is a compile-time constant
         auto expr = CheckExpr(stmt->expr);
+
+        // coerce to type being switch on, and ensure that value is a compile-time constant
+        auto exprVal = CheckIntegerConstantExpression(expr, IntegerConstantExpressionCoercionType::AnyInteger, nullptr, ConstantFoldingKind::CompileTime);
```
csyonghe, line 300: Should call `tryConstantFoldExpr` to allow enum typed IntVals.

sriramm-nv: `CheckIntegerConstantExpression` does call `tryConstantFoldExpr` as well as attempting to coerce when we use `const int` values for cases. So, I feel that this is a better choice here.

csyonghe: The difference is that CheckIntegetConstantExpr is trying to coerce the Val into an int after folding, which may fail if the enum type has an underlying type that is not int.

sriramm-nv: Actually you are right. A falcor test with enum cases failed due to cost analysis:

```
C:\actions-runner\_work\slang\slang\FalcorBin\build\windows-vs2022\bin\Release\shaders\Scene/Scene.slang(662): error 30019: expected an expression of type 'int', got 'HitType'
```
The cost was >= 90000, and caused the folding to fail at this point at `slang-check-conversion.cpp`.
```
if (cost >= kConversionCost_Explicit)
```

I changed this to use `tryConstantFoldExpr`. And I'm not able to get the positive test case with int to pass, it fails with an internal error.
```
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: assert failure: !inst || as<T>(inst)
```
I'm considering removing the extra test case.

source/slang/slang-check-stmt.cpp
```diff
@@ -247,16 +289,16 @@ namespace Slang
...
     }
 
     void SemanticsStmtVisitor::visitCaseStmt(CaseStmt* stmt)
     {
-        // TODO(tfoley): Need to coerce to type being switch on,
-        // and ensure that value is a compile-time constant
         auto expr = CheckExpr(stmt->expr);
+
+        // coerce to type being switch on, and ensure that value is a compile-time constant
+        auto exprVal = tryConstantFoldExpr(expr, ConstantFoldingKind::CompileTime, nullptr);
```
csyonghe, line 301: tryConstantFoldExpr does not produce diagnostic if folding failed. We could try to test if exprVal is null and issue error if so. Hopefully this won't break the Falcor test this time.

sriramm-nv: Well, unfortunately it did fail Falcor image test. So I pulled that check out of the PR :(

csyonghe (dismissed):

source/slang/slang-diagnostic-defs.h
```diff
@@ -556,6 +556,8 @@ DIAGNOSTIC(39999, Note, moreOverloadCandidates, "$0 more overload candidates")
 
 DIAGNOSTIC(39999, Error, caseOutsideSwitch, "'case' not allowed outside of a 'switch' statement")
 DIAGNOSTIC(39999, Error, defaultOutsideSwitch, "'default' not allowed outside of a 'switch' statement")
+DIAGNOSTIC(39999, Error, multipleDefaultsInASwitch, "multiple 'default' cases not allowed within a 'switch' statement")
+DIAGNOSTIC(39999, Error, duplicateCasesInASwitch, "duplicate cases not allowed within a 'switch' statement")
```
jkwak-work, line 560: If you look at the other errors from "30500", the names starts with "forLoopXXX".
I think we can keep the pattern here for switch.
```
switchMultipleDefaults
switchDuplicatedCases
```

source/slang/slang-check-stmt.cpp
```diff
@@ -239,6 +239,48 @@ namespace Slang
...
+            if (auto caseStmt = as<CaseStmt>(sStmt))
+            {
+                if (caseStmt->exprVal)
+                {
+                    // exprVal contains the constant folded expr, that is checked for
+                    // uniqueness within the scope of the switch statement.
+                    if (!caseStmtVals.add(caseStmt->exprVal))
+                    {
+                        sink->diagnose(sStmt, Diagnostics::duplicateCasesInASwitch);
+                        return;
```
jkwak-work, line 266: I am just thinking loud here.
I am wondering if it is possible for two different memory pointers to point to a same value.

source/slang/slang-check-stmt.cpp
```diff
@@ -239,6 +239,48 @@ namespace Slang
...
+                    }
+                }
+            }
+
+            // check that there is at most one `default` clause
+            else if (auto defaultStmt = as<DefaultStmt>(sStmt))
+            {
```
jkwak-work, line 273: This is a nitpick and feel free to ignore.
I think the empty line here can cause a mistake.
It would be better if it was like following,
```
            }
            else if (auto defaultStmt = as<DefaultStmt>(sStmt))
            {
                // check that there is at most one `default` clause
```

source/slang/slang-check-stmt.cpp
csyonghe: In our ASTs, everything inherits from `Val` is globally deduplicated, meaning that Vals will be pointer-unique.
The same applies for all IRInst* that are `HOISTABLE`.

source/slang/slang-diagnostic-defs.h
sriramm-nv: done

source/slang/slang-check-stmt.cpp
sriramm-nv: fixed it

source/slang/slang-check-stmt.cpp
sriramm-nv: thanks for the comment. I added some more explanation where the const folding is done.

source/slang/slang-diagnostic-defs.h
```diff
@@ -476,6 +476,9 @@ DIAGNOSTIC(30504, Warning, forLoopTerminatesInFewerIterationsThanMaxIters, "the
 DIAGNOSTIC(30505, Warning, loopRunsForZeroIterations, "the loop runs for 0 iterations and will be removed.")
 DIAGNOSTIC(30510, Error, loopInDiffFuncRequireUnrollOrMaxIters, "loops inside a differentiable function need to provide either '[MaxIters(n)]' or '[ForceUnroll]' attribute.")
 
+DIAGNOSTIC(30600, Error, switchMultipleDefault, "multiple 'default' cases not allowed within a 'switch' statement")
```
jkwak-work, line 479: You should add a comment following the pattern on the other values.
```
// Switch
```

jkwak-work (approved):
Looks good to me except one missing comment as I mentioned.

source/slang/slang-diagnostic-defs.h
sriramm-nv: ah, every time i push a patch, it requires the review to be approved again.

jkwak-work (approved):
Looks good to me

PR #3865, opened 2024-04-01
Fix peephole optimization of `TypeEquals`.

### LINKED ISSUE #3861: the "is" statement does not work properly in generics with spirv
I have this code:
```hlsl
interface I
{

}

struct A : I
{
    uint i;
};

struct B : I
{
    float2 f2;
};

func test<T : I>(T t)
{
    if (T is A)
    {
        printf("A\n");
    }
    if (T is B)
    {
        printf("B\n");
    }
}

[shader("compute")]
[numthreads(1,1, 1)]
void main()
{
    B b;
    test(b);
}
```

And i expect this to print "B\n", as i call the generic function test with the type B. But i both paths are taken, "A" and "B" are printed.

I compiled under the Config:
```
pc platform: x86-64msvc windows11
slang::TargetDesc::format: SlangCompileTarget::SLANG_SPIRV
slang::TargetDesc::profile: glsl_460+spirv_1_4
slang::TargetDesc::flags: SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY
slang::SessionDesc::defaultMatrixLayoutMode: SLANG_MATRIX_LAYOUT_COLUMN_MAJOR
```

csyonghe:
Closes #3861.

The cause is that we are using both `kIROp_TypeType` and `kIROp_TypeKind` to represent the type of an inst that represents a type. This needs to be clean up separately so that we unify them into a single opcode.

### DIFF
--- a/tests/language-feature/types/is-on-type.slang
+++ b/tests/language-feature/types/is-on-type.slang
@@ -1,0 +1,45 @@
+
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-slang -compute
+
+// Test that `is` operator works on generic type param.
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+interface I
+{
+
+}
+
+struct A : I
+{
+    uint i;
+};
+
+struct B : I
+{
+    float2 f2;
+};
+
+func test<T : I>(T t) -> int
+{
+    int rs = 0;
+    if (T is A)
+    {
+        rs++;
+    }
+    if (T is B)
+    {
+        rs+=2;
+    }
+    return rs;
+}
+
+[shader("compute")]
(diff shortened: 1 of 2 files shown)


tests/language-feature/types/is-on-type.slang
```diff
@@ -0,0 +1,45 @@
...
+{
+    uint i;
+};
+
+struct B : I
+{
+    float2 f2;
+};
+
+func test<T : I>(T t) -> int
```
jkwak-work, line 24: Just out of curiosity, what is this syntax starting with "func"?
This is not HLSL nor GLSL, is it?
I don't remember reading this syntax in the Slang user guide either.

jkwak-work (approved):
The code looks good and simple.

tests/language-feature/types/is-on-type.slang
csyonghe: No, this is the "modern" syntax supported in Slang. We should add to the documentation that we support this.

Issue #3671, opened 2024-03-04
__init shouldn't be able to return a value


jkwak-work:
It appears that slangc currently allows "__init()" functions to return a value.
We should have a compiler error for this case.

Currently we have two bugs in core.meta.slang,
```
__generic<T:__BuiltinFloatingPointType, let N : int>
extension vector<T,N> : IFloat
{
    [OverloadRank(-1)]
    [__unsafeForceInlineEarly] __init(int v) { return vector<T,N>(T(v)); }
    [OverloadRank(-1)]
    [__unsafeForceInlineEarly] __init(float v) { return vector<T,N>(T(v)); }
```
The correct code should be assigning to "this",
```
__init(...) { this = vector<T,N>(T(v)); }
```
We may find more cases like this once we have a compiler error implemented.

jkwak-work:
When I try to use "return;" inside of "__init()", like this,
```
__init()
{
    return;
}
```
it hits an error in a following condition,
```
    void SemanticsStmtVisitor::visitReturnStmt(ReturnStmt *stmt)
    {
        auto function = getParentFunc();
        if (!stmt->expression)
        {
            if (function && !function->returnType.equals(m_astBuilder->getVoidType()))
            {
                getSink()->diagnose(stmt, Diagnostics::returnNeedsExpression);
            }
        }
```
The error message is
```
'return' should have an expression.
```
, which is not the case for __init.

The related code is in the following function,
```
    void SemanticsDeclHeaderVisitor::visitConstructorDecl(ConstructorDecl* decl)
    {
        // We need to compute the result tyep for this declaration,
        // since it wasn't filled in for us.
        decl->returnType.type = findResultTypeForConstructorDecl(decl);

        checkCallableDeclCommon(decl);
    }
```
I tried to replace it to
```
        decl->returnType.type = m_astBuilder->getVoidType();
```
And it breaks bunch.
The error message looks like following,
```
core.meta.slang(986): error 30019: expected an expression of type 'half', got 'void'
    static const half maxValue = half(65504);
```
It seems that __init is expected to return a type of its own type.

We may need to make an exception for the constructor when we check the return statement, "SemanticsStmtVisitor::visitReturnStmt()".

PR #3866, opened 2024-04-01
Fix __init() functions that returns an existing value

sriramm-nv:
Fixes the issue https://github.com/shader-slang/slang/issues/3671

* The slang constructors are not expected to return a value like other member
  functions, but must constructed a new value and return void.

* This patch fixes this in the IR lowering without complaining about
  this illegal behavior.
** Translate    CopyableType return ...; --> this = ...; return;
** Translate NonCopyableType return ...; --> this = ...; return *this;

* This patch also fixes the issue with type checking when __init()
  returns a void that mismatches the base type of the struct/ class


tests/diagnostics/non-void-return-constructor.slang
```diff
@@ -0,0 +1,35 @@
...
+        // A constructor from the inside the body is treated as a function that modifies `this`.
+        // because we are expected to write:
+        // this.v = 5;
+        // here.
+ 
+        MyType t;
+        t.v = 5;
+        // CHECK: ([[# @LINE+1]]): error 39999: constructor cannot return a value
+        return t;
+    }
```
jkwak-work, line 20: Can we have one more test case that a constructor can call "return" without any value?

sriramm-nv: can I put that in the same file, or create a new one? 
my concern is that no errors will be reported when we have a constructor with a return with no value. so, i can't add a CHECK line.

jkwak-work: You will need to have a new file for a good test case.

sriramm-nv: Thanks Jay. But based on an offline chat, i ended up adding an extra case in the same file.

jkwak-work:
Please leave a more descriptive commit message; see [an example](https://github.com/jkwak-work/slang/blob/master/CONTRIBUTION.md#commit-to-the-branch)

tests/diagnostics/non-void-return-constructor.slang
```diff
@@ -0,0 +1,47 @@
...
+    // this.v = 5;
+
+    // A test case where the constructor returns a value.
+    // This automatically gets resolved as this = t; return;
+    __init()
+    {
+        MyType t;
+        t.v = 5;
+        return t;
+    }
```
jkwak-work, line 20: Shouldn't this be an error?
Why do we allow this?

sriramm-nv: there has been test cases where something like this was done, and as a result, we wanted to fix it within the IR lowering, and not complain.
@csyonghe , should we just fix the Falcor test, and have the error thrown, or proceed with the silent fix?

csyonghe: Let's try to support this without throwing an error, because this is semantically valid.

`__init` is treated at the IR level something that returns a newly constructed value, instead of a function that modifies an existing value. However, this is not true for `[NonCopyableType]`, in which case we are only modifying `this`. See `slang-lower-to-ir.cpp, maybeAddReturnDestinationParam`.

Therefore, the right fix should be:
If we see `return ...`, translate it into `this = ...; return` if we are lowering to NonCopyableType, or translate it into `this = ...; return *this; ` if we are lowering for other types.

We can tell if we are lowering a constructor for a `NonCopyableType` by checking `context->returnDestination.flavor != LoweredValInfo::Flavor::None`. (We already do this in `visitReturnStmt`.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6006,7 +6010,7 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
         // a (non-`void`) value, and one for returning "no value"
         // (which effectively returns a value of type `void`).
         //
-        if( auto expr = stmt->expression )
+        if (auto expr = stmt->expression)
```
csyonghe, line 6013: We also need to handle the case where `expr==nullptr`, and we are inside a ConstructorDecl. In this case, we need to:
1. just `return` if for `NonCopyableType`.
2. emit `return *this;` for ordinary case.

Make sure to comment well.

sriramm-nv: does this solve 2.
`getBuilder()->emitReturn(getSimpleVal(context, context->thisVal));`

csyonghe: yes. but we can't do this for 1).

sriramm-nv: right. for 1, I'm simply doing `getBuilder()->emitReturn();`

sriramm-nv:
updated the PR with all the corner case handling. though it makes the code look confusing. i'll think about refactoring once the Falcor test issue is fixed.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6014,13 +6017,25 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
         {
             if (context->returnDestination.flavor != LoweredValInfo::Flavor::None)
             {
-                // If this function should return via a __ref parameter, do that and return void.
-                lowerRValueExprWithDestination(context, context->returnDestination, expr);
-                getBuilder()->emitReturn();
-                return;
+                if (constructorDecl)
```
csyonghe, line 6020: In this branch we should not do the `if(constructorDecl)`, Noncopyable types's constructor should be handled the same as ordinary functions as in 6032-6035.

sriramm-nv: thanks for clarifying this. i got confused with the different scenarios, without trying an actual test case for each.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6016,6 +6023,16 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
                 return;
             }
 
+            if (constructorDecl)
+            {
+                // If this function is a constructor, but returns a value, rewrite it as
+                // this = val;
+                // return;
+                lowerRValueExprWithDestination(context, context->thisVal, expr);
+                getBuilder()->emitReturn();
```
csyonghe, line 6032: You want to return thisVal as well.

sriramm-nv: done

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6050,6 +6067,21 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
             // with no value, which can only occur in a function with
             // a `void` result type.
             //
+            // Special handling when `return` is inside a constructor.
+            if (constructorDecl)
+            {
+                if (context->returnDestination.flavor != LoweredValInfo::Flavor::None)
+                {
+                    // Simply emit `return` for NonCopyableType.
```
csyonghe, line 6075: When writing comments, focus on why rather than what. It is clear that the code is emitting just a return. But we should try to explain why:

A constructor for non copyable type is in the form of 
void ctor(inout this),
Where a constructor for ordinary type is in the form of
Type ctor()

csyonghe (approved):

csyonghe (dismissed):

tests/diagnostics/non-void-return-constructor.slang
```diff
@@ -0,0 +1,47 @@
...
+    // A valid use-case when the constructor returns a void
+    __init(int v)
+    {
+        if (v == 0)
+        {
+            this.v = 5;
+            return this;
+        }
+        this.v = v;
+    }
```
jkwak-work, line 31: This code looks little confusing.
Should we allow the function to return a value at the same time it doesn't need to have a return statement at the end?

jkwak-work (commented):
I have two questions.

tests/diagnostics/non-void-return-constructor.slang
sriramm-nv: this is a positive test that confirms that it is ok to have a return from the constructor. i will change `return this;` to `return;` and possibly add another `return;` at the end if it helps. or i can just leave `return this;` as it is, and only add another `return;` at the end.

sriramm-nv:
@jkwak-work @csyonghe I am noticing from today that these test cases or passing on master. The fixes that are in this patch are no longer needed. Do you see anything else that went in to master that might have affected this behavior?

csyonghe:
I believe the test cases are required. The fact that the tests pass on master means the coverage isn't enough.

Consider tests that include:
- early returns in an `if`, with or without expr value.
- non copyable types
and the combination of the things above.

tests/diagnostics/ctor-noncopyable-retval-legal.slang
```diff
@@ -0,0 +1,103 @@
...
+
+// Test to catch compilation time errors with return values in constructors within a NonCopyable struct.
+// A constructor from the callsite's point of view is a function that returns the struct type.
+// A constructor from the inside the body is treated as a function that modifies `this` and not return.
+
+// In the following test-cases where the constructor either returns an expr or no value,
+// within a condition for early exit or not, it automatically gets resolved as
+// return *this; or return;
+
+[__NonCopyableType] struct S
```
csyonghe, line 11: We have opportunity to turn this test into an executable test. If we actually run the shader and write something to a buffer, then getting the correct execution result will imply there are no compile errors.

sriramm-nv: definitely. do you recommend any example tests that I can look up ?

tests/diagnostics/ctor-retval-legal.slang
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+
+void test<T : IFoo>(int id)
+{
+    T t0;
+    t0.compute(id);
+    T t1 = T(1);
+    t1.compute(id + 4);
```
csyonghe, line 60: why + 4? Can just be +1 so we don't need that many //CHECK: 0.0 lines in between.

csyonghe (dismissed):

tests/diagnostics/ctor-retval-legal.slang
```diff
@@ -0,0 +1,82 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj -output-using-type
```
jkwak-work, line 1: My understanding is that "diagnostics" directory is for testing the error messages.
I think this file should be relocated to elsewhere because it is not for error test.

tests/diagnostics/ctor-retval-legal.slang
```diff
@@ -0,0 +1,82 @@
...
+        }
+        else {
+            this.x = 3 * v;
+            return;
+        }
+    }
+
+    __init(int ival) {
+        float val = ival;
+        Impl v = Impl(val);
```
jkwak-work, line 40: What is Impl?
Is it same as "This"?

tests/diagnostics/ctor-retval-legal.slang
```diff
@@ -0,0 +1,82 @@
...
+    t1.compute(id + 4);
+    T t2 = T(-1);
+    t2.compute(id + 8);
+}
+
+
+[numthreads(1, 1, 1)]
+void computeMain(uint id : SV_DispatchThreadID)
+{
+    // CHECK: 4.000000
```
jkwak-work, line 69: Shouldn't this be 2?

tests/diagnostics/ctor-retval-legal.slang
sriramm-nv: done

tests/diagnostics/ctor-retval-legal.slang
sriramm-nv: done

tests/diagnostics/ctor-retval-legal.slang
```diff
@@ -0,0 +1,82 @@
...
+    __init(int);
+    __init(float);
+    void compute(int);
+}
+
+struct Impl : IFoo
+{
+    float x;
+
+    __init(float v) {
```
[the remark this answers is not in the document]
sriramm-nv: done

tests/diagnostics/ctor-retval-legal.slang
csyonghe: Yes. `This` is a typedef for the owning type. So `Impl` and `This` should mean the same thing in this context.

tests/diagnostics/ctor-noncopyable-retval-legal.slang
```diff
@@ -0,0 +1,103 @@
...
+        // CHECK-NOT: ([[# @LINE+1]]): error
+        return;
+    }
+}
+
+[__NonCopyableType] struct S3 : S
+{
+    __init()
+    {
+        S3 t;
```
jkwak-work, line 72: Doesn't this cause an infinite loop?
When __init() tries to create a variable "t", it will keep getting into the same __init() recursively.

sriramm-nv: correct. this will cause a runtime failure, but from compilation perspective this is acceptable by the typecheker

tests/diagnostics/ctor-noncopyable-retval-legal.slang
```diff
@@ -0,0 +1,103 @@
+//TEST:SIMPLE(filecheck=CHECK):
```
csyonghe, line 1: Is this still necessary? seems superseded by the runnable tests.

tests/language-feature/interfaces/constructor-noncopyable-return.slang
```diff
@@ -0,0 +1,59 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj -output-using-type
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+// Note: spirv_by_reference is only supported for passing opaque types, so this test won't produce
```
csyonghe, line 3: make sure the comments are up to date.

tests/language-feature/interfaces/constructor-return.slang
```diff
@@ -0,0 +1,57 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj -output-using-type
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -vk -shaderobj -output-using-type
+
+// Test to catch errors with return values in constructors within an ordinary struct.
```
csyonghe, line 5: comment not up to date.

tests/diagnostics/ctor-noncopyable-retval-legal.slang
sriramm-nv: i left them behind because it explicitly listed all the test cases, including the infinite loop case. shall i remove them?

csyonghe: I see, it is fine to leave them here.

tests/language-feature/interfaces/constructor-noncopyable-return.slang
sriramm-nv: I was not able to run this test on vk. Do you any suitable comment for adding to that?

tests/language-feature/interfaces/constructor-return.slang
sriramm-nv: updated

tests/language-feature/interfaces/constructor-noncopyable-return.slang
csyonghe: That's fine, let's not worry about it for now.

csyonghe (approved):

PR #3867, opened 2024-04-01
Support `[RequirePrelude]` attribute on types.

### LINKED ISSUE #3846: Allow __requirePrelude for a type or globally for files.
Hi,

currently the `__requirePrelude` seems to only be supported in functions. This breaks for type definitions when generating C++ code. This slang code:
```c++
// 2024.1.5: slangc.exe -entry computeMain -stage compute -target cpp

__target_intrinsic(cpp, "Foo")
//__requirePrelude("#include <Foo.hpp>"); // not allowed
//__requirePrelude(cpp, "#include <Foo.hpp>"); // also not allowed
struct Foo {
    __init() {
        __requirePrelude("#include <Foo.hpp>"); // only works when actually initializing Foo
        __intrinsic_asm ";";
    }
};

struct Bar {
    int bar;
    Foo foo;
};

uniform StructuredBuffer<Bar> bar;
uniform RWStructuredBuffer<int> out;

void workaround() {
    __requirePrelude("#include <Foo.hpp>");
    __intrinsic_asm "";
}

[shader("compute")]
[numthreads(1, 1, 1)]
void computeMain() {
    //workaround(); //<- fix
    out[0] = bar[0].bar;
}
```
generates C++ code containing this struct:
```c++
struct Bar_0
{
    int32_t bar_0;
    Foo foo_0; // <- unknown type
};
```
but is missing the definition for `Foo` which would come from `Foo.hpp`. The only workaround I can currently see is to call dummy functions from every slang entry point to make sure my includes are generated (see `workaround()` function). I think supporting the `__requirePrelude` for type definitions or globally in slang files would fix the issue more elegantly.

Thanks for taking this into consideration!

csyonghe:
Closes #3846.


source/slang/slang-emit-c-like.cpp
```diff
@@ -129,6 +129,16 @@ void CLikeSourceEmitter::emitPreModuleImpl()
 // Types
 //
 
+void CLikeSourceEmitter::ensureTypePrelude(IRType* type)
+{
+    if (auto requirePreludeDecor = as<IRRequirePreludeDecoration>(findBestTargetDecoration<IRRequirePreludeDecoration>(type)))
```
kaizhangNV, line 134: Why not just call _findBestTargetIntrinsicDecoration() ?

kaizhangNV: Never mind, I see the reason. The template parameter is different.
But where _findBestTargetIntrinsicDecoration() is actually used?

csyonghe: They are different. We only want to look at all RequirePrelude decorations, and not TargetIngrinsic decorations.

kaizhangNV (approved):
It looks good to me.

PR #3869, opened 2024-04-01
Fix static member lookup in pointer types.

### LINKED ISSUE #3868: Extending Ptr<T> fails
The following code fails to compile
```hlsl
extension Ptr<int> {
  static func FromHandle(uint64_t handle) -> Ptr<int> {
    return (Ptr<int>)handle;
  }
}

[shader("compute")]
[numthreads(1, 1, 1)]
func Main() -> void {
  let x = Ptr<int>::FromHandle(0ull);
  *x = 0;
}
```
With the error:
```
error 300
27: 'FromHandle' is not a member of 'typeof(Ptr<int>)'.
  let x = Ptr<int>::FromHandle(0);
                    ^~~~~~~~~~
```

Furthermore, a generic version doesn't work either, with the same error.
```hlsl
__generic<T> extension Ptr<T> {
  // ...
}
```

This was compiled with the following settings:

- Host: Windows 11
- Triple: x86_64-windows-msvc
- slang::TargetDesc::format: SLANG_SPIRV
- slang::TargetDesc::flags: SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY
- slang::TargetDesc::profile: "spirv_1_5"

csyonghe:
Closes #3868.

When looking up members of a pointer type, we will automatically dereference the pointer and lookup in the pointee type instead.

However when the lookup is static, such as when the user is doing (Ptr<T>::member), we shouldn't be doing this implicit deref step, and instead need to lookup in the Ptr type itself.

### DIFF
--- a/source/slang/slang-check-expr.cpp
+++ b/source/slang/slang-check-expr.cpp
@@ -3728,7 +3728,9 @@ namespace Slang
                         this,
                         expr->name,
                         type,
-                        m_outerScope);
+                        m_outerScope,
+                        LookupMask::Default,
+                        LookupOptions::NoDeref);
 
                     // We need to confirm that whatever member we
                     // are trying to refer to is usable via static reference.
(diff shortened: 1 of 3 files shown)


jkwak-work (approved):
Looks good

Issue #3385, opened 2023-12-07
Support SM6.6 WaveSize


csyonghe:
The `WaveSize` keyword introduced in SM6.6 does not yet seem to be supported in Slang. 

```
[WaveSize(<numLanes>)]
void main() ...
```

csyonghe:
Assigning to @jkwak-work as it seems this is critical to get the UE shader's performance right.

jkwak-work:
Fix is merged.
Closing.

PR #3871, opened 2024-04-02
Support SM6.6 keyword "WaveSize"

jkwak-work:
Resolves an issue #3385

Shader Model 6.6 added a new keyowrd, "WaveSize". See the following link for more details:
  https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_WaveSize.html

This change is required for measuring the compile time of UE TSR shaders.


csyonghe (approved):
Looks good!

source/slang/slang-check-modifier.cpp
```diff
@@ -362,6 +362,47 @@ namespace Slang
...
+                auto intValue = checkLinkTimeConstantIntVal(arg);
+                if (!intValue)
+                {
+                    return false;
+                }
+                if (auto constIntVal = as<ConstantIntVal>(intValue))
+                {
+                    bool isValidWaveSize = false;
+                    const IntegerLiteralValue waveSize = constIntVal->getValue();
+                    for (int validWaveSize : { 4, 8, 16, 32, 64, 128 })
```
ArielG-NV, line 383: I believe the warp sizes should include `1` and `2` if Slang is to fully support ARM GPU's ([Mali specifically](https://tvm.apache.org/2018/01/16/opt-mali-gpu)) since some ARM paired GPU's support independent thread launching:  
```suggestion
                    for (int validWaveSize : { 1, 2, 4, 8, 16, 32, 64, 128 })
```

Otherwise everything looks good.

ArielG-NV (approved):

source/slang/slang-check-modifier.cpp
jkwak-work: This change is mainly for supporting what SM6.6 spec supports.
The spec document explicitly says the following,
```
The allowed wave sizes that an HLSL shader may specify are the powers of 2 between 4 and 128, inclusive. In other words, the set: [4, 8, 16, 32, 64, 128].
```
We can easily add 1, and 2 if we have a use case later.
I am not going to add anything more than the spec says.

PR #3874, opened 2024-04-02
Allow enum values to be used as generic arguments.

csyonghe:
The main change here is to allow constant folding to preserve the enum types from enum case values, and to allow constant propagation logic to identify the binary operator overloads.


jkwak-work (approved):
Looks good.

PR #3880, opened 2024-04-02
Legalization of non-struct when function expects struct, resolves #3840

### LINKED ISSUE #3840: Feature Request: Legalize Non-struct TraceRay payload types in HLSL backend
To optimize performance and minimize continuation state, I'll often trace rays using very simple and sometimes even compressed payload types, eg a float3 for a color, a float for the hit distance, perhaps a small array of floats for maintaining a list of hits... 

In the SPIR-V backend of slang, this works great. Eg, the following code to determine a hit distance "just works":

```
int temp;
HitObject hit = HitObject::TraceRay( rtAccel, flags, 0xff, 1, 1, 1, ray,
        temp // not actually used... 
);
if (hit.IsHit()) return hit.GetRayDesc().TMax; // <= note, no miss shader called, improving perf
```

But in the HLSL backend to Slang, the underlying DXR backend throws an error, which goes uncaught by Slang:
`User defined type intrinsic arg must be struct`

Personally I'm not a fan of requiring the following code:
```
struct EmptyStruct{int temp;};
EmptyStruct temp;
HitObject hit = HitObject::TraceRay( rtAccel, flags, 0xff, 1, 1, 1, ray,
        temp // not actually used... 
);
...
```

Would it be possible to instead do some hlsl-legalization of "non-struct" types like scalars, vectors, matrices, and arrays by wrapping them in an anonymous struct before passing them to a TraceRay call?

ArielG-NV:
resolves https://github.com/shader-slang/slang/issues/3840

Introduce `__forceVarIntoStructTemporarily()` to solve the issue of passing "non-struct type's" into a parameter that only accepts "struct type's".

The code generation transforms the following code:
```
int dataIn = 1;
HitObject hit = HitObject::TraceRay( rtAccel, flags, 0xff, 1, 1, 1, ray, dataIn);
```
effectively into:
```
struct SomeStruct
{
 int data;
}
...
int dataIn = 1;
SomeStruct dataInStruct; 
dataInStruct.data = dataIn; 
HitObject hit = HitObject::TraceRay( as, flags, 0xff, 1, 1, 1, ray, dataInStruct);
dataIn = dataInStruct.data;
```


source/slang/hlsl.meta.slang
```diff
@@ -9614,7 +9634,17 @@ void TraceRay(
 {
     __target_switch
     {
-    case hlsl: __intrinsic_asm "TraceRay";
+    case hlsl:
```
csyonghe, line 9637: This function should now be [ForceInline] to keep resulting code clean.

source/slang/slang-emit.cpp
```diff
@@ -839,6 +840,19 @@ Result linkAndOptimizeIR(
         break;
     }
 
+    // Legalize non struct parameters that are expected to be structs for HLSL. 
+    switch (target)
```
csyonghe, line 844: Can be `isD3DTarget(target)`

source/slang/slang-ir-hlsl-legalize.cpp
```diff
@@ -0,0 +1,94 @@
...
+            for(UInt i = 0; i < call->getArgCount(); i++)
+            {
+                auto arg = call->getArg(i);
+                if(arg->getOp() != kIROp_ForceVarIntoStructTemporarily)
+                    continue;
+                auto forceStructArg = arg->getOperand(0);
+                auto forceStructBaseType = as<IRType>(forceStructArg->getDataType()->getOperand(0));
+                if(forceStructBaseType->getOp() == kIROp_StructType)
+                {
+                    arg->replaceUsesWith(arg->getOperand(0));
```
csyonghe, line 38: should be `call->setOperand(i, arg->getOperand(0)` to avoid unexpected changes in other places.

csyonghe (approved):

source/slang/hlsl.meta.slang
ArielG-NV: The reason ForceInline was not added is because it causes us to generate different code than glslang (cross-compile-test will fail). 

Understanding these tests are depreciated (and my new tests cover the code better) I will remove the cross-compile tests. I am bringing this up inside the PR for future reference and to explain why some tests will be modified

tests/vkray/raygen-trace-ray-param-struct.slang
```diff
@@ -9,9 +9,6 @@ struct testStruct
 
 RWStructuredBuffer<int> outputBuffer1;
 
-// CHECK-DAG: TraceRay(
-// CHECK-DAG: TraceMotionRay(
-
 // CHECK-DAG: void main(
```
jkwak-work, line 12: Can you remove "-DAG"?

jkwak-work (commented):
I don't know what this means,
```
adjust how we replace operands in senarios to avoid unexpected side effects of replacing operands without any explicit checks
```
Please write commit messages more human readable next time.

tests/vkray/raygen-trace-ray-param-struct.slang
ArielG-NV: I can remove them now (since we use ForceInline the code is predictable now)

csyonghe (dismissed):

csyonghe (approved):

PR #3881, opened 2024-04-02
Refactor memory qualifier decorators to be a bit-flag set, resolves #3841

### LINKED ISSUE #3841: Change memory qualifiers to be fully centered around memory qualifier collections
Currently Slang still uses memory qualifier modifiers on their own. This should be removed in favor for memory qualifier collection checks for faster qualifier operations & to remove legacy usage of these modifiers.

This will be faster through the need to only find a memoryQualifierCollection, rather than 1-5 individual modifiers (which all are legal to combine together)

ArielG-NV:
resolves #3841 & #3883

Replaces GloballyCoherent, ReadOnly, WriteOnly, Volatile, and Restrict memory modifiers and decorations with a bit flag set (`MemoryQualifierSetModifier`) to more efficiently manage memory qualifiers.

Added `restrict` modifier to test to ensure the code works when dropping a `restrict` memory qualifier.


source/slang/slang-emit-spirv.cpp
```diff
@@ -3319,36 +3319,47 @@ struct SPIRVEmitContext
...
-        case kIROp_GLSLWriteOnlyDecoration:
-            emitOpDecorate(getSection(SpvLogicalSectionID::Annotations),
-                decoration,
-                dstID,
-                SpvDecorationNonReadable);
+        case kIROp_MemoryQualifierSetDecoration:
+        {
+            auto collection = as<IRMemoryQualifierSetDecoration>(decoration);
+            IRIntegerValue flags = collection->getMemoryQualifierBit();
+            if (flags | MemoryQualifierSetModifier::Flags::kCoherent)
```
csyonghe, line 3326: flags & MemoryQualifierSetModifier::Flags::kCoherent

source/slang/slang-emit-spirv.cpp
```diff
@@ -3441,15 +3452,16 @@ struct SPIRVEmitContext
...
+                else if (auto collection = as<IRMemoryQualifierSetDecoration>(decor))
                 {
-                    emitOpMemberDecorate(
-                        getSection(SpvLogicalSectionID::Annotations),
-                        decor,
-                        spvStructID,
-                        SpvLiteralInteger::from32(id),
-                        SpvDecorationCoherent
-                    );
+                    if(collection->getMemoryQualifierBit() | MemoryQualifierSetModifier::Flags::kCoherent)
```
csyonghe, line 3457: &

csyonghe (changes requested):

source/slang/slang-emit-spirv.cpp
ArielG-NV: Thank you for catching that big mistake which I included everywhere...
I will fix it in the other areas as well.

csyonghe: No problems, it is the kind of mistake I make all the time. Wonder why the test case didn't catch it.

csyonghe: Maybe good to have some CHECK lines and CHECK-NOT lines to make sure we have all the right decorations.

ArielG-NV: The tests should be checking for the modifier syntax outputted, I realized I did not add it, which I will add.

csyonghe (approved):

csyonghe (approved):

source/slang/slang-emit-hlsl.cpp
```diff
@@ -1274,6 +1273,13 @@ void HLSLSourceEmitter::emitVarDecorationsImpl(IRInst* varDecl)
             m_writer->emit(")]]\n");
             continue;
         }
+        if (auto collection = as<IRMemoryQualifierSetDecoration>(decoration))
```
jkwak-work, line 1276: Feel free to ignore this comment.
But I will try to stay away from the term "collection", because the meaning is too vague.
I like to suggest `qualifierSet` for this case.

PR #3885, opened 2024-04-02
Fix the erroneous logic of determining whether or not to emit DepthReplacing.

### LINKED ISSUE #3884: Do not use `DepthReplacing` execution mode when the fragment shader is not modifying depth.
Helps improves performance.

csyonghe:
Closes #3884.

### DIFF
--- a/source/slang/slang-emit-spirv.cpp
+++ b/source/slang/slang-emit-spirv.cpp
@@ -2894,7 +2894,7 @@ struct SPIRVEmitContext
     }
 
 
-    SpvExecutionMode isDepthOutput(IRInst* builtinVar)
+    SpvExecutionMode getDepthOutputExecutionMode(IRInst* builtinVar)
     {
         SpvExecutionMode result = SpvExecutionModeMax;
         bool isDepthVar = false;
@@ -2943,6 +2943,7 @@ struct SPIRVEmitContext
             {
             case kIROp_SwizzledStore:
             case kIROp_Store:
+            case kIROp_Call:
                 return result;
             }
         }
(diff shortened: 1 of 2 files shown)


source/slang/slang-emit-spirv.cpp
```diff
@@ -2960,24 +2961,19 @@ struct SPIRVEmitContext
...
+                mode = thisMode;
+            else if (mode != thisMode)
             {
                 needDepthReplacingMode = true;
-                if (mode == SpvExecutionModeMax)
-                    mode = thisMode;
-                else if (mode != thisMode)
-                    mode = SpvExecutionModeDepthReplacing;
+                mode = SpvExecutionModeDepthReplacing;
                 break;
```
ArielG-NV, line 2971: Does the code here intentionally not `break;` at `if (mode == SpvExecutionModeMax)`?

csyonghe: No we can't break because the goal here is to find an evidence that we need DepthReplacing. And we don't know we don't need it unless we have examined every globalInst.

source/slang/slang-emit-spirv.cpp
```diff
@@ -2960,24 +2961,19 @@ struct SPIRVEmitContext
...
         }
         if (!needDepthReplacingMode)
             return;
-        emitOpExecutionMode(getSection(SpvLogicalSectionID::ExecutionModes),
-            nullptr,
-            entryPoint,
-            SpvExecutionModeDepthReplacing);
-        if (mode != SpvExecutionModeDepthReplacing &&
-            mode != SpvExecutionModeMax)
+        if (mode != SpvExecutionModeMax)
```
ArielG-NV, line 2976:
```suggestion
        if (mode == SpvExecutionModeMax)
            return;
```
purely a syntax candy suggestion: removes need for the indentation & braces below

ArielG-NV (dismissed):
Everything looks good 👍

ArielG-NV (approved):

csyonghe:
Thanks for the review!

Issue #3838, opened 2024-03-26
variable created inside scope, then used outside of it


GDBobby:
!flag is set to a variable inside the loop, which is fine, but then it uses that variable outside of the scope it was created in, which causes an error
`
glslang:  OceanFFT.slang(70): error :  '_S14' : undeclared identifier
`
`
glslang:  OceanFFT.slang(70): error :  '' : compilation terminated 
`
Line 150 and 183 in the generated glsl file
Error happens when compiling to spirv directly from slangc, in both optimization level 0 and 3
[OceanFFT_slang.txt](https://github.com/shader-slang/slang/files/14752995/OceanFFT_slang.txt)

[OceanFFT_slang_glsl.txt](https://github.com/shader-slang/slang/files/14752996/OceanFFT_slang_glsl.txt)

csyonghe:
Does it work if you compile with
-emit-spirv-directly?

GDBobby:
yes, it compiles, thank you

PR #3892, opened 2024-04-03
WIP: Fix the variable scope issue (#3838)

kaizhangNV:
In the IR optimization pass, we turn all the loop to do-while loop form. But in the do-while loop form, the loop body block is dominating the blocks after the loop break block. This assumption is fine for SPIRV and IR code, however, it's incorrect for all the other language target (e.g. c/c++/cuda/glsl/hlsl) because the instructions defined in the loop body is invisible from outside of the loop. Therefore, when translating to other textual language, there could be issue for the variables scope.

To fix this issue, we first detect the instructions that are defined inside the loop block, then check if these instructions are used after the break block. If so, we duplicate these instructions right before their users such that we can make those instructions available globally.

### DIFF
--- a/source/slang/slang-ir-variable-scope-correction.cpp
+++ b/source/slang/slang-ir-variable-scope-correction.cpp
@@ -26,7 +26,7 @@ struct VariableScopeCorrectionContext
     void _processFunction(IRFunc* funcInst);
     void _processInstruction(IRDominatorTree* dominatorTree, IRInst* instAfterParam,
         IRInst* originInst, const List<IRLoop*>& loopHeaderList, List<IRInst*>& workList);
-    void _processStorableInst(IRInst* insertLoc, IRInst* inst, const List<IRUse*>& outOfScopeUser);
+    void _processStorableInst(IRInst* insertLoc, IRInst* inst, const List<IRUse*>& outOfScopeUses);
     void _processUnstorableInst(IRInst* inst, const List<IRUse*>& outOfScopeUser);
 
     bool _isStorableType(IRType* inst);
@@ -108,11 +108,7 @@ void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
         }
     }
 
-    auto instAfterParam = funcInst->getFirstBlock()->getFirstChild();
-    while(instAfterParam->getOp() == kIROp_Param)
-    {
-        instAfterParam = instAfterParam->getNextInst();
-    }
+    auto instAfterParam = funcInst->getFirstBlock()->getFirstOrdinaryInst();
 
     for(auto inst = workList.begin(); inst != workList.end(); inst++)
     {
@@ -151,25 +147,25 @@ bool VariableScopeCorrectionContext::_isOutOfScopeUse(IRInst * userInst, IRDomin
 void VariableScopeCorrectionContext::_processInstruction(IRDominatorTree* dominatorTree, IRInst* instAfterParam,
         IRInst* originInst, const List<IRLoop*>& loopHeaderList, List<IRInst*>& workList)
 {
-    List<IRUse*> outOfScopeUsers;
+    List<IRUse*> outOfScopeUses;
     for (auto use = originInst->firstUse; use; use=use->nextUse)
     {
         if(_isOutOfScopeUse(use->getUser(), dominatorTree, loopHeaderList))
         {
-            outOfScopeUsers.add(use);
+            outOfScopeUses.add(use);
         }
     }
 
-    if (outOfScopeUsers.getCount() == 0)
+    if (outOfScopeUses.getCount() == 0)
         return;
 
     if(_isStorableType(originInst->getDataType()))
     {
-        _processStorableInst(instAfterParam, originInst, outOfScopeUsers);
+        _processStorableInst(instAfterParam, originInst, outOfScopeUses);
     }
     else
     {
-        _processUnstorableInst(originInst, outOfScopeUsers);
+        _processUnstorableInst(originInst, outOfScopeUses);
         // After processing the user, we need to add operands of the instruction to the worklist
         // for later processing.
         for(UInt idx = 0; idx < originInst->getOperandCount(); idx++)
@@ -179,7 +175,7 @@ void VariableScopeCorrectionContext::_processInstruction(IRDominatorTree* domina
     }
 }
 
-void VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, IRInst* inst, const List<IRUse*>& outOfScopeUser)
+void VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, IRInst* inst, const List<IRUse*>& outOfScopeUses)
 {
     auto type = inst->getDataType();
     // store instruction must have a result type
@@ -196,11 +192,11 @@ void VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, IRI
     // last, replace operands in the use site instruction with the new variable
     // Note, because "dstPtr" is a pointer type, we have to insert a load(dstPtr) instruction before use it.
     // Simply replace any operand with pointer could generate error code.
-    for (auto user : outOfScopeUser)
+    for (auto use : outOfScopeUses)
     {
-        m_builder.setInsertBefore(user->getUser());
+        m_builder.setInsertBefore(use->getUser());
         auto loadInst = m_builder.emitLoad(type, dstPtr);
-        m_builder.replaceOperand(user, loadInst);
+        m_builder.replaceOperand(use, loadInst);
     }
 }
 
--- a/source/slang/slang-ir-variable-scope-correction.cpp
+++ b/source/slang/slang-ir-variable-scope-correction.cpp
@@ -27,7 +27,7 @@ struct VariableScopeCorrectionContext
     IRInst* _processStorableInst(IRInst* insertLoc, IRInst* inst, IRInst* user);
 
     void _replaceOperand(IRInst* inst, IRInst* oldOperand, IRInst* newOperand);
-    bool _isStorableInst(IRInst* inst);
+    bool _isStorableType(IRType* inst);
     bool _isOutOfScopeUse(IRInst* inst, IRDominatorTree* domTree, const List<IRLoop*>& loopHeaderList);
 
     IRModule* m_module;
@@ -203,7 +203,7 @@ IRInst* VariableScopeCorrectionContext::_processInstruction(IRDominatorTree* dom
     // Check if the user of this instruction is out of the scope of the loop
     if(_isOutOfScopeUse(useInst, dominatorTree, loopHeaderList))
     {
-        bool isStoable = _isStorableInst(originInst);
+        bool isStoable = _isStorableType(originInst->getDataType());
         IRInst* newInst = nullptr;
         if(isStoable)
         {
@@ -281,7 +281,7 @@ IRInst* VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, 
     // Note, because "dstPtr" is a pointer type, we have to insert a load(dstPtr) instruction before use it.
     // Simply replace any operand with pointer could generate error code.
     m_builder.setInsertBefore(user);
-    auto loadInst = m_builder.emitLoad(dstPtr->getDataType(), dstPtr);
+    auto loadInst = m_builder.emitLoad(type, dstPtr);
 
     return loadInst;
 }
@@ -303,21 +303,29 @@ void  VariableScopeCorrectionContext::_replaceOperand(IRInst* inst, IRInst* oldO
     }
 }
 
-bool VariableScopeCorrectionContext::_isStorableInst(IRInst* inst)
+bool VariableScopeCorrectionContext::_isStorableType(IRType* type)
 {
-    auto type = inst->getDataType();
+    if (!type)
+        return false;
 
-    // If the instruction has a result, and the result is not a void type, we consider it as a storable instruction
-    if (type)
-    {
-        // Take care of pointer type, because we can't store a pointer type, so pointer type is regarded as a non-storable instruction
-        if (type->getOp() == kIROp_PtrType)
-        {
-            return false;
-        }
+    if (as<IRBasicType>(type))
         return true;
+
+    switch(type->getOp())
+    {
+        case kIROp_VectorType:
+        case kIROp_MatrixType:
+        case kIROp_StructType:
+            return true;
+        case kIROp_ArrayType:
+        case kIROp_UnsizedArrayType:
+            if (auto arrayType = as<IRArrayTypeBase>(type))
+                return _isStorableType(arrayType->getElementType());
+            else
+                return false;
+        default:
+            return false;
     }
-    return false;
 }
 
 } // anonymous
--- a/source/slang/slang-ir-variable-scope-correction.cpp
+++ b/source/slang/slang-ir-variable-scope-correction.cpp
@@ -246,7 +246,11 @@ void VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, IRI
     m_builder.emitStore(dstPtr, inst);
 
     // last, replace operands in the use site instruction with the new variable
-    _replaceOperand(user, inst,  dstPtr);
+    // Note, because "dstPtr" is a pointer type, we have to insert a load(dstPtr) instruction before use it.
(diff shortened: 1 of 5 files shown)


source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,139 @@
...
+                    // instructions in this block are considered to be defined in the loop. We need to search
+                    // the instruction's use sites to see if there is any uses are out of the loop.
+                    if (!dominatorTree->dominates(breakBlock, block))
+                    {
+                        debugPrint(indent++, block, "block is not dominated by break block");
+                        // traverse all uses of this instruction
+                        for (auto use = inst->firstUse; use; use=use->nextUse)
+                        {
+                            debugPrint(indent++, use->getUser()->getParent(), "inst's use block is");
+                            if (auto userBlock = as<IRBlock>(use->getUser()->getParent()))
```
csyonghe, line 89: getBlock(use->getUser())

because some insts (e.g. insts inside a spirv block) has parent that is not directly a block.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,139 @@
...
+                            {
+                                // If the use site of this instruction is dominated by the break block, it means that the
+                                // instruction is used after the break block, so we need to make that instruction available globally.
+                                // By doing so, we record all the users of this instructions.
+                                if (dominatorTree->dominates(breakBlock, userBlock))
+                                {
+                                    debugPrint(indent, inst, "inst is defined in loop but used outside of loop");
+                                    debugPrint(indent++, use->getUser(), "The user is");
+
+                                    auto list = workListMap.tryGetValue(inst);
```
csyonghe, line 99: workListMap.add(inst) returns true if insts does not exist and was added by the call. Use this to avoid querying the hash map twice.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // traverse each instruction in the function
+    for (auto block = funcInst->getFirstBlock(); block; block = block->getNextBlock())
+    {
+        for (auto inst = block->getFirstChild(); inst; inst = inst->getNextInst())
+        {
+            uint32_t indent = 1;
+            printf("process inst = %d\n", inst->getOp());
+            // traverse the dominator tree, theoretically, the variables in each of dominator should be accessible in the current block,
+            // except the loop block, so we have to find out if there is a loop block in the dominator chain.
+            auto dominatorBlock = dominatorTree->getImmediateDominator(block);
```
csyonghe, line 68: The logic here is correct, but I encourage you to think about how to reduce the complexity of these loops a bit. For example, we can do a lot of these dominator chasing work once per block instead of once per inst.

kaizhangNV: Yea, the loop order can definitely be adjusted to reduce the iterations.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "slang-ir.h"
+
+#include "slang-ir-dominators.h"
+#include "slang-ir-variable-scope-correction.h"
+
+namespace Slang
+{
+
+namespace { // anonymous
+struct VariableScopeCorrectionContext
```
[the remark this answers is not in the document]
kaizhangNV: Add detailed description.

source/slang/slang-ir-variable-scope-correction.cpp
kaizhangNV: Reduce the complexity by using 2 passes.

source/slang/slang-ir-variable-scope-correction.cpp
kaizhangNV: Done.

source/slang/slang-ir-variable-scope-correction.cpp
kaizhangNV: Done

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -54,65 +54,78 @@ static void debugPrint(uint32_t indent, IRInst* inst, const char* message)
...
                     }
                 }
-            }
-
-            if (instList.getCount() > 0)
-            {
-                workListMap.add(inst, instList);
+                if (instList.getCount() > 0)
+                {
+                    workListMap.add(inst, instList);
```
csyonghe, line 122: You added it to the worklist, but we didn't process the worklist?

csyonghe:
I think the flow should be:

1. compute Dictionary<IRBlock*, List<IRLoop*>> for all blocks in the func the list of loop regions they are in.
2. add all insts in the func to work list.
3. Loop until work list is dry. For all insts in the work list, do the checking and duplication/load/store when needed.

source/slang/slang-ir-variable-scope-correction.cpp
kaizhangNV: The process is in line 130.

kaizhangNV:
> compute Dictionary<IRBlock*, List<IRLoop*>> for all blocks in the func the list of loop regions they are in.
Why does a block have more loop regions? Do you mean we need to consider about multiple level loops?

> Loop until work list is dry. For all insts in the work list, do the checking and duplication/load/store when needed.
I'm still thinking about our previous discussions.
What kinds of instructions are needed to duplicate outside loop?

Why do we need to check the load/store? I know the current process is not exactly correct, but seems like just duplicating the the instructions right before the use site resolves the issue. I can't come up with a corner case that can fail this for now.

csyonghe:
There are certain things that you can't declare a variable and store it in.

For example, the result of `getElementPtr()` inst is a pointer.
You can't define a variable whose type is a pointer in HLSL.
The only way to make it available outside the loop is to duplicate it.

csyonghe:
On the other side, there are certain things that you don't want to duplicate.
For example:
```
int someGlobalVar;
void doSomeSideEffect() { someGlobalVar++; }
...

do
{
       int a = doSomeSideEffect();
}while();
int b = a + 1;
```

You cannot duplicate the call to `doSomeSideEffect` because that changes the semantics of the program.

kaizhangNV:
Those are really good examples.
I need to think about how to categorize which instructions need to be duplicated, and which are not.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+bool VariableScopeCorrectionContext::_isStorableInst(IRInst* inst)
+{
+    auto type = inst->getDataType();
+
+    // If the instruction has a result, and the result is not a void type, we consider it as a storable instruction
+    if (type)
+    {
+        // Take care of pointer type, because we can't store a pointer type, so pointer type is regarded as a non-storable instruction
+        if (type->getOp() == kIROp_PtrType)
```
csyonghe, line 298: Instead of checking what can't be stored, just check for what can be stored.
There are way more things we can't store than the things we can.

We can only store these:
IRBasicType
IRVectorType
IRMatrixType
IRArrayTypeBase, if element type is storable.
IRStructType

You can define a `isStorableType(IRType* type)` function to compute this.

csyonghe: Also, everything is storable if we are on cpp/cuda target.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,317 @@
...
+    if (debugid != 0)
+        printf("%s, id: %d\n", message, debugid);
+    else
+        printf("\n");
+}
+
+void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
+{
+    IRDominatorTree* dominatorTree = m_module->findOrCreateDominatorTree(funcInst);
+    Dictionary<IRInst*, List<IRInst*>> workListMap;
```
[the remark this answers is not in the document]
kaizhangNV: yeah, consider the recursive processing, I think I should do this.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,317 @@
...
+    Dictionary<IRBlock*, List<IRLoop*>> loopHeaderMap;
+
+    bool enableLog = false;
+
+    // traverse all blocks in the function
+    for (auto block = funcInst->getFirstBlock(); block; block = block->getNextBlock())
```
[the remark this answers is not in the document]
kaizhangNV: Done this update in the new change.

source/slang/slang-ir-variable-scope-correction.cpp
kaizhangNV: Yea, I did the update in my latest change.
I will add the cpp/cuda check then. But the good news is that after I fix the last issue in the `_processStorableInst` that fails the Falcor image test, now all the CIs passed.

kaizhangNV: The cpp/cuda check is done.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    for(auto it = workList.begin(); it != workList.end(); it++)
+    {
+        auto inst = *it;
+        if (auto loopHeaderList = loopHeaderMap.tryGetValue(getBlock(inst)))
+        {
+            for (auto use = inst->firstUse; use; use=use->nextUse)
+            {
+                _processInstruction(dominatorTree, instAfterParam, inst, use->getUser(), *loopHeaderList, workList);
```
csyonghe, line 126: Should pass use instead of use->getUser so we can legalize this use later.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,307 @@
...
+        }
+        else
+        {
+            // duplicate the instruction right before the use site.
+            newInst = _processUnstorableInst(originInst, useInst);
+            // When we duplicate the out-of-scope instruction, the side-effect is that the operands of the instruction
+            // could also be out of the scope. So we have to recursively check all the operands of the instruction. Such
+            // issue dooesn't happen for the storable instruction because we store the result of the instruction in a new
+            // variable visible in the whole function.
+            UInt operandCount = newInst->getOperandCount();
```
csyonghe, line 182: This can be irBuilder.replaceOperand(use, newInst)

The remaining of the code here is not necessary. Basically you want processInst to make a single use valid at a time.

csyonghe:
The reason to use workList is to avoid any recursive calls. I think the structure here can be streamlined so that we are processing one use at at a time instead of one instruction at a time. An instruction may need to be duplicated multiple times if the use sites are in different loop regions.

For storable inst, because we already created the variable at global scope, we only need to process once and insert a load at each out of scope use site.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,307 @@
...
+    auto dstPtr = m_builder.emitVar(type);
+
+    // insert a store instruction after the instruction
+    m_builder.setInsertAfter(inst);
+    m_builder.emitStore(dstPtr, inst);
+
+    // last, replace operands in the use site instruction with the new variable
+    // Note, because "dstPtr" is a pointer type, we have to insert a load(dstPtr) instruction before use it.
+    // Simply replace any operand with pointer could generate error code.
+    m_builder.setInsertBefore(user);
```
csyonghe, line 245: Instead of handling just one user here, simply loop through all uses and insert loads at all out of scope use sites.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,341 @@
...
+        return;
+    }
+
+    debugPrint(0, nullptr, "", enableLog);
+    // Traverse all the instructions in function.
+    for (auto block : funcInst->getBlocks())
+    {
+        if(loopHeaderMap.containsKey(block))
+        {
+            for (auto inst = block->getFirstChild(); inst; inst = inst->getNextInst())
```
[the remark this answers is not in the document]
kaizhangNV: Done

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,307 @@
...
+{
+    // For a given instruction, we need to check all the users of this instruction to see
+    // if the user is out of the scope of the loop. If so, we need to make the instruction available globally.
+    // In addition, we have to recursively check all the operands of this instruction to see if they are also
+    // out of the scope of the loop.
+
+    // Check if the user of this instruction is out of the scope of the loop
+    if(_isOutOfScopeUse(useInst, dominatorTree, loopHeaderList))
+    {
+        bool isStoable = _isStorableType(originInst->getDataType());
```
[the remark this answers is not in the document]
kaizhangNV: Done

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,263 @@
...
+        _processStorableInst(instAfterParam, originInst, outOfScopeUsers);
+    }
+    else
+    {
+        _processUnstorableInst(originInst, outOfScopeUsers);
+        // After processing the user, we need to add operands of the instruction to the worklist
+        // for later processing.
+        for(UInt idx = 0; idx < originInst->getOperandCount(); idx++)
+        {
+            workList.add(originInst->getOperand(idx));
```
csyonghe, line 177: We need to use a hash set for the workList to make sure the same inst doesn't get inserted twice into the worklist.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+    }
+    return false;
+}
+
+void VariableScopeCorrectionContext::_processInstruction(IRDominatorTree* dominatorTree, IRInst* instAfterParam,
+        IRInst* originInst, const List<IRLoop*>& loopHeaderList, List<IRInst*>& workList)
+{
+    List<IRUse*> outOfScopeUsers;
```
csyonghe, line 154: -> outOfScopeUses

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,263 @@
...
+        // After processing the user, we need to add operands of the instruction to the worklist
+        // for later processing.
+        for(UInt idx = 0; idx < originInst->getOperandCount(); idx++)
+        {
+            workList.add(originInst->getOperand(idx));
+        }
+    }
+}
+
+void VariableScopeCorrectionContext::_processStorableInst(IRInst* insertLoc, IRInst* inst, const List<IRUse*>& outOfScopeUser)
```
csyonghe, line 182: outOfScopeUses

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,263 @@
...
+    auto dstPtr = m_builder.emitVar(type);
+
+    // insert a store instruction after the instruction
+    m_builder.setInsertAfter(inst);
+    m_builder.emitStore(dstPtr, inst);
+
+    // last, replace operands in the use site instruction with the new variable
+    // Note, because "dstPtr" is a pointer type, we have to insert a load(dstPtr) instruction before use it.
+    // Simply replace any operand with pointer could generate error code.
+    for (auto user : outOfScopeUser)
```
csyonghe, line 199: for (auto use : ...)

csyonghe (commented):
This is now looking clean. Need to rename some `user` to `use`.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -0,0 +1,263 @@
...
+                if (inst->getOp() == kIROp_Var)
+                {
+                    continue;
+                }
+                workList.add(inst);
+            }
+        }
+    }
+
+    auto instAfterParam = funcInst->getFirstBlock()->getFirstChild();
```
csyonghe, line 111: This can just be:
funcInst->getFirstBlock()->getFirstOrdinaryInst().

csyonghe (approved):
There is a worrying sign that this PR used to cause some Falcor failures.

The thing is none of these logic added here should trigger when compiling the Falcor codebase. We probably need to make sure that we are not making unnecessary modifications.

PR #3893, opened 2024-04-03
Not emit DepthReplacing when frag shader uses SV_Position.

csyonghe:
#3884 was not completely fixed by #3885. This enhances the emit logic further.


jkwak-work (approved):
Code looks good to me.

PR #3895, opened 2024-04-04
Add decoration PerPrimitiveEXT when a mesh output variable is decorated with PrimitiveID

### LINKED ISSUE #3872: Device Lost when writing `SV_PrimitiveID` to `OutputPrimitives<T, N>`
The following shader produces a DEVICE_LOST error 
```hlsl
const static let color: float3[] = {
  float3(1.0, 0.0, 0.0),
  float3(0.0, 1.0, 0.0),
  float3(0.0, 0.0, 1.0),
};

const static let position: float3[] = {
  float3( 0.5,  0.5, 0.0),
  float3( 0.0, -0.5, 0.0),
  float3(-0.5,  0.5, 0.0),
};

struct SVertexOutput {
  float4 Position: SV_Position;
  float3 Color;
};

struct SPrimitiveOutput {
  uint PrimitiveId: SV_PrimitiveID;
};

[shader("mesh")]
[numthreads(1, 1, 1)]
[outputtopology("triangle")]
func MainMesh(
  OutputVertices<SVertexOutput, 3> outputVertices,
  OutputPrimitives<SPrimitiveOutput, 1> outputPrimitives,
  OutputIndices<uint3, 1> outputIndices,
) -> void {
  SetMeshOutputCounts(3, 1);
  outputVertices[0] = { float4(position[0], 1.0), color[0] };
  outputVertices[1] = { float4(position[1], 1.0), color[1] };
  outputVertices[2] = { float4(position[2], 1.0), color[2] };
  outputIndices[0] = uint3(0, 1, 2);

  // This next line causes the DEVICE_LOST error
  outputPrimitives[0] = { 0 };
}

[shader("pixel")]
func MainPixel() -> void {}
```

I have attached a sample application for further testing, building steps:
```
cd extracted_folder
mkdir build
cd build
cmake ..
cmake --build .
```
Building requires a C++ compiler that supports at least C++14

If untouched, executing the application will produce `VK_ERROR_DEVICE_LOST`, this was tested on the following configurations:
- Windows 11
- NVIDIA RTX 3070 - 551.86, NVIDIA RTX 4070 Super - 551.86

[bug.zip](https://github.com/shader-slang/slang/files/14838685/bug.zip)

pmistryNV:
Fixes #3872

### DIFF
--- a/source/slang/slang-emit-spirv.cpp
+++ b/source/slang/slang-emit-spirv.cpp
@@ -3827,6 +3827,7 @@ struct SPIRVEmitContext
                     // non-tessellation shader stages, and if so include a declaration of
                     // Geometry capability.
                     bool needGeometryCapability = true;
+                    bool needPerPrimitiveEXTDecoration = false;
                     if (entryPoints)
                     {
                         for (auto entryPoint : *entryPoints)
@@ -3835,9 +3836,13 @@ struct SPIRVEmitContext
                             {
                                 switch (entryPointDecor->getProfile().getStage())
                                 {
+                                case Stage::Mesh:
+                                    // For outputs decorated with PrimitiveID it must be decorated
+                                    // with PerPrimitiveEXT to comply with Vulkan validation rules
+                                    if (m_capabilities.contains(SpvCapabilityMeshShadingEXT))
+                                        needPerPrimitiveEXTDecoration = true;
                                 case Stage::Geometry:
                                 case Stage::Intersection:
-                                case Stage::Mesh:
                                 case Stage::Amplification:
                                 case Stage::AnyHit:
                                 case Stage::ClosestHit:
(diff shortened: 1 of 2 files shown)


pmistryNV:
@csyonghe kindly review

source/slang/slang-emit-spirv.cpp
```diff
@@ -3835,9 +3836,13 @@ struct SPIRVEmitContext
                             {
                                 switch (entryPointDecor->getProfile().getStage())
                                 {
+                                case Stage::Mesh:
+                                    // For outputs decorated with PrimitiveID it must be decorated
+                                    // with PerPrimitiveEXT to comply with Vulkan validation rules
+                                    if (m_capabilities.contains(SpvCapabilityMeshShadingEXT))
```
csyonghe, line 3842: This `if` seems unnecessary because if the builtin is used by a mesh shader then that capability must have been defined.

csyonghe:
You can write 
```
Fixes #3872
```
In the PR description so that github will automatically close that issue upon merge of the PR.

source/slang/slang-emit-spirv.cpp
csyonghe: I think we should get rid of use of NV_mesh_shader in the compiler since there is no reason to use it anymore.

pmistryNV: Ok do you want me to create a new bug for removing support for NV_mesh_shader? In that case I can remove this check and just add a comment.

csyonghe:
@pmistryNV Is the plan to implement the more robust solution that covers all other builtins in this PR, or do you prefer to merge this first and do the complete fix in a separate PR?

pmistryNV:
> @pmistryNV Is the plan to implement the more robust solution that covers all other builtins in this PR, or do you prefer to merge this first and do the complete fix in a separate PR?

Practically this fix is sufficient for this bug. I think we can have a follow up PR to clean up the implementation.

csyonghe (approved):

PR #3899, opened 2024-04-05
Enable NonUniformResourceIndex support for glsl, hlsl and spirv

sriramm-nv:


source/slang/slang-emit-glsl.cpp
```diff
@@ -2068,6 +2068,21 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
             maybeCloseParens(assignNeedsClose);
             return true;
         }
+        case kIROp_NonUniformResourceIndex:
+        {
+            // Need to emit as a Function call for HLSL
+            m_writer->emit("NonUniformResourceIndex");
```
nanokatze, line 2074:
```suggestion
            // Need to emit as a Function call for GLSL
            m_writer->emit("nonuniformEXT");
```

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,44 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        IRInst* args[] = { sb, index };
+        auto addrInst = builder.emitIntrinsicInst(
+            builder.getPtrType(kIROp_PtrType, value->getFullType(), SpvStorageClassStorageBuffer),
+            kIROp_RWStructuredBufferGetElementPtr,
+            2,
+            args);
+        auto textureInst = builder.emitRWStructuredBufferGetElementPtr(value, index);
+        nonUniformResourceIndexInst->replaceUsesWith(textureInst);
+        nonUniformResourceIndexInst->removeAndDeallocate();      
+        addUsersToWorkList(textureInst);
```
nanokatze, line 1205: Somewhat confused as to what's going on here. I assume this is still WIP?

nanokatze: I'm guessing slang would prefer to do this by propagating `NonUniformResource` down to the users, so probably something like the following?
```
auto value = nonUniformResourceIndexInst->GetOperand(0);
traverseUses(nonUniformResourceIndexInst, [&](IRUse* use) 
{
 auto user = use->getUser();
 // Optionally:
 if (user->getOp() != /* ops for which NonUniformResourceIndex makes sense */)
  return;
 IRBuilder builder(user);
 auto decoratedUser = builder.emitNonUniformResourceIndex(user);
 user->replaceUsesWith(decoratedUser);
});
// Optionally also 
// nonUniformResourceIndexInst->replaceUsesWith(value);
```

sriramm-nv: Yes it is still a WIP. I’m hoping to get it ready in a few days. I ll update the title until then. Sorry for the confusion.

source/slang/hlsl.meta.slang
```diff
@@ -6396,55 +6396,10 @@ T __copyObject<T>(T v)
...
-        return indexCopy;
-    default:
-        return index;
-    }
-}
-
-__glsl_extension(GL_EXT_nonuniform_qualifier)
-[__readNone]
-[ForceInline]
 [NonUniformReturn]
```
csyonghe, line 6399: I think this function should not be decorated with [NonUniformReturn].

source/slang/slang-ir-insts.h
```diff
@@ -327,6 +327,19 @@ struct IRRequireGLSLVersionDecoration : IRDecoration
     }
 };
 
+// Do We Need This?
```
csyonghe, line 330: Yes, it is good to have it.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,166 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                case kIROp_NonUniformResourceIndex:
+                    printf("!! kIROp_NonUniformResourceIndex !!\n");
+                    break;
+                default:
+                    printf("!! Unknown !!\n");
+                    return;
+                };
+                for (auto use = inst->firstUse; use; use = use->nextUse)
+                {
+                    visit(use->getUser());
```
csyonghe, line 1252: Should avoid recursive calls. Use work lists instead of recursion.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,166 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
         addUsersToWorkList(newStore);
     }
 
+    void processNonUniformResourceIndex(IRInst* nonUniformResourceIndexInst)
+    {
+        // implement the translation to spirv by waking up the use-def chain
```
csyonghe, line 1172: I think the logic here should be: if we see `gep(pArray, nonUniformRes(x))`, then replace it with
`nonUniformRes(gep(pArray, x))` then add this new nonUniformRes inst to work list.
If we see `instCast(nonUniformRes(x))`, do the same and replace it with `nonUniformRes(intCast(x))`.

If we see `nonUniformRes(load(x))` or any other unknown opcodes, stop.

Then after that, replace all `nonUniformRes(x)` with `x`, and add a nonuniform res decoration to `x`.

Peudo code:
```
workList.add(nonUniformResourceIndexInst);
for (Index i = 0; i < workList.getCount(); i++)
{
      traverseUsers(inst)
      {
            auto user = use->getUser();
            IRBuilder builder(user);
            builder.setInsertBefore(user);
            IRInst* newUser = nullptr;
            switch(user->getOp())
            {
            case IntCast:
                  newUser = builder.emitIntCast(user->getFullType(), inst->getOperand(0);
                  break;
            case GEP:
                    if (use is operand 1)
                        newUser = builder.emitGEP(user->getFullType(), inst->getOperand(0);
                   break;
            case Load:
                    newUser = builder.emitLoad(user->getFullType(), inst->getOperand(0);
                    break;
            }
           if (!newUser) continue;
           auto nonUniformUser = builder.emitNonUniformResourceIndex(newUser);
           user->replaceAllUsesWith(nonUniformUser);
           switch(user->getOp())
           {
           case IntCast:
           case GEP:
                  workList.add(nonUniformUser);
            }
      }
}

// Now we have "upstream"ed all `NonUniformResourceIndex` insts, 
// the next step is to add a decoration the operands of all nonuniformresourceindex insts
for (auto nonUniformResInst : workList)
      auto operand =  nonUniformResInst->getOperand(0);
      if (isResourceType(operand->getDataType()) || operand's type is a pointer to resource type)
          builder.addNonUnifomResourceDecoration(operand);
      nonUniformResInst->replaceUsesWith(operand);

```

sriramm-nv: Seems like this is going in the right direction. Instead of using a for loop, I used a recursive visitor 
```
 std::function<void(IRInst*)> visit = [&](IRInst* inst)
 {
     traverseUses(inst, [&](IRUse* use)
     {
         auto user = use->getUser();
         IRBuilder builder(user);
         builder.setInsertBefore(user);
         IRInst* newUser = nullptr;
         switch (user->getOp())
         {
         case kIROp_IntCast:
             newUser = builder.emitCast(user->getFullType(), inst->getOperand(0));
             break;
         case kIROp_GetElementPtr:
             newUser = builder.emitElementAddress(user->getFullType(), user->getOperand(0), inst->getOperand(0));
             break;
         case kIROp_Load:
             newUser = builder.emitLoad(user->getFullType(), inst->getOperand(0));
             break;
         default:
             break;
         };
         if (newUser)
         {
             auto nonUniformUser = builder.emitNonUniformResourceIndex(newUser->getFullType(), newUser);
             user->replaceUsesWith(nonUniformUser);
         }
     });
 };
 visit(nonUniformResourceIndexInst);
```
The logic is almost the same, except for the fact that we don't have a worklist at hand to deal with the next steps.
What is interesting in this approach is that the, processWorkList() function that invokes this code within `processNonUniformResourceIndex`, actually keeps track of all the instructions, and the first time it calls the function it processes the GEP, and the next time it processes the Load. It is intelligent to keep checking the IR until all `NonUniform`s are processed.

After applying the first part of the transform, this is what I am getting. 
```
	{1710}	let  %21	: Int	= nonUniformResourceIndex(0 : Int)
	{1767}	let  %22	: TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %9)	= getElement(%textures, %21)
	{2714}	let  %23	: Vec(Int, 2 : Int)	= makeVector(0 : Int, 0 : Int)
	{2688}	SPIRVAsm %24	: Float
	{
		{2692}	SPIRVAsmInst(95 : UInt, __sampledType(Float), %"sampled", getElement(%textures, %21), makeVector(0 : Int, 0 : Int), 2 : UInt, 0 : Int)
		{2694}	SPIRVAsmInst(__truncate, Float, result, __sampledType(Float), %"sampled")
	}
```
to
```
	{1710}	let  %21	: Int	= nonUniformResourceIndex(0 : Int)
	{2775}	let  %22	: Ptr(TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18), 0 : Int)	= getElementPtr(%textures, 0 : Int)
	{2776}	let  %23	: Ptr(TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18), 0 : Int)	= nonUniformResourceIndex(%22)
	{2780}	let  %24	: Ptr(TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18), 0 : Int)	= nonUniformResourceIndex(%22)
	{2781}	let  %25	: TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18)	= load(%22)
	{2782}	let  %26	: TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18)	= nonUniformResourceIndex(%25)
	{2790}	let  %27	: Vec(Int, 2 : Int)	= globalValueRef(%28)
	{2688}	SPIRVAsm %29	: Float
	{
		{2692}	SPIRVAsmInst(95 : UInt, __sampledType(Float), %"sampled", nonUniformResourceIndex(%25), globalValueRef(%28), 2 : UInt, 0 : Int)
		{2694}	SPIRVAsmInst(__truncate, Float, result, __sampledType(Float), %"sampled")
	}
```

If this is the case, can I perform the inst -> decoration conversion in the spirv emitter?

csyonghe: The second step shall be done inside the function processing function.

The philosophy is never do any actual IR transformation during emit. Emit should just be straight forward conversion from slang ir to spirv.

csyonghe: The work list mechanism is to speed up the process so we don't need to rediscover all nonUniformInsts. We also don't want to involve any recursion because it is slow and risks overflowing the stack.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,80 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                IRInst* newUser = nullptr;
+                switch (user->getOp())
+                {
+                case kIROp_IntCast:
+                    printf("INTCAST!! %d\n", user->_debugUID);
+                    newUser = builder.emitCast(user->getFullType(), inst->getOperand(0));
+                    break;
+                case kIROp_GetElementPtr:
+                    // Replace gep(pArray, nonUniformRes(x)), into nonUniformRes(gep(pArray, x))
+                    printf("GETELEMENTPTR!! %d\n", user->_debugUID);
```
csyonghe, line 1208: You need to make sure the nonUniformResourxeIndex is on the index instead of the array. If it is on the array we don't want to push any further up.

gep(nonUniform(array), index) => no transform.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,80 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        // an extract element address instruction
+        // if the first use is an extract element ptr, find all the loads that uses the element,
+        // and the decorate them
+        //
+        // if the first use is a getelementptr, we may want to decorate the get element ptr itself,
+        // because user might
+        // call an image atomic operation, that operates on the ptr instead of the texture / resource
+        // todo rename the decoration itself nonuniformresourceindex -> nonuniformresource
+        List<IRInst*> resWorkList;
+        std::function<void(IRInst*)> visit;
```
csyonghe, line 1191: Don't use std::function if you are not recursing, it is slow.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,80 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        };
+        visit(nonUniformResourceIndexInst);
+
+        // Now we have "upstream"ed all `NonUniformResourceIndex` insts, 
+        // the next step is to add a decoration the operands of all nonuniformresourceindex insts
+        for (auto nonUniformResInst : resWorkList)
+        {
+            auto operand = nonUniformResInst->getOperand(0);
+            IRBuilder builder(nonUniformResInst);
+            //if (isResourceType(operand->getDataType())) // || operand's type is a pointer to resource type)
```
csyonghe, line 1236: I think we still need to guard the type so we don't add the decoration on unrelated things. isResourceType probably didn't include the buffer types, need to double check.

source/slang/slang-ir.cpp
```diff
@@ -6132,6 +6132,18 @@ namespace Slang
         return i;
     }
 
+    // IR emitter for a dedicated instruction to represent NonUniformResourceIndex qualifier.
+    IRInst* IRBuilder::emitNonUniformResourceIndex(IRType* type, IRInst* val)
```
csyonghe, line 6136: Can simply things by deriving type from inst instead of asking the caller to provide the type.

source/slang/slang-emit-glsl.cpp
sriramm-nv: Thanks!

source/slang/hlsl.meta.slang
sriramm-nv: done

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: understood!

source/slang/slang-ir.cpp
sriramm-nv: done

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: i'm using a simple `if (user->getOperand(0)->getOp() != kIROp_NonUniformResourceIndex)` to check this. Is there a better way?

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: This is the final IR I'm seeing after all the fixups:
```
	[{2796}	NonUniformResource]
	[{2784}	NonUniformResource]
	{2775}	let  %21	: Ptr(TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18), 0 : Int)	= getElementPtr(%textures, 0 : Int)
	[{2798}	NonUniformResource]
	{2778}	let  %22	: TextureType(Float, TextureShape1DType, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, 0 : Int, %18)	= load(%21)
	{2806}	let  %23	: Vec(Int, 2 : Int)	= globalValueRef(%24)
	{2688}	SPIRVAsm %25	: Float
	{
		{2692}	SPIRVAsmInst(95 : UInt, __sampledType(Float), %"sampled", load(%21), globalValueRef(%24), 2 : UInt, 0 : Int)
		{2694}	SPIRVAsmInst(__truncate, Float, result, __sampledType(Float), %"sampled")
	}
```
 Does it look like its in the right step.

source/slang/hlsl.meta.slang
```diff
@@ -6396,55 +6396,9 @@ T __copyObject<T>(T v)
...
-            OpDecorate $indexCopy NonUniform;
-        };
-        return indexCopy;
-    default:
-        return index;
-    }
-}
+__generic<T:__BuiltinArithmeticType>
+__intrinsic_op($(kIROp_NonUniformResourceIndex))
+   T NonUniformResourceIndex(T index);
```
jkwak-work, line 6401: I think you can keep "[__readNone]" as it was.

sriramm-nv:
The latest code changes now goes all the way to spirv and it decorates both the GEP and Load Inst.
for the input
```

    float2 tmp1 = textures[NonUniformResourceIndex(0)].Load(int3(0, 0, 0)); // get element ptr
    float2 tmp2 = textures[NonUniformResourceIndex(uint(0))].Load(int3(0, 0, 0)); // int cast from uint // try float
```
it generates
```
OpDecorate %13 NonUniform
OpDecorate %15 NonUniform
OpDecorate %23 NonUniform
OpDecorate %26 NonUniform

%13 = OpAccessChain %_ptr_UniformConstant_7 %textures %int_0
%15 = OpLoad %7 %13

%23 = OpAccessChain %_ptr_UniformConstant_7 %textures %uint_0
%26 = OpLoad %7 %23
```
May I know if this is still ok ,or we should only decorate the `OpLoad`. If so, any suggestions on the code changes?

source/slang/slang-ir-insts.h
```diff
@@ -327,6 +327,18 @@ struct IRRequireGLSLVersionDecoration : IRDecoration
     }
 };
 
+struct IRSPIRVNonUniformResourceDecoration : IRDecoration
+{
+    enum { kOp = kIROp_SPIRVNonUniformResourceDecoration };
+    IR_LEAF_ISA(RequireGLSLVersionDecoration)
+
+        IRConstant* getSPIRVVersionOperand() { return cast<IRConstant>(getOperand(0)); }
```
ArielG-NV, line 335:
```suggestion
    IRConstant* getSPIRVVersionOperand() { return cast<IRConstant>(getOperand(0)); }
```

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,103 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+    void processNonUniformResourceIndex(IRInst* nonUniformResourceIndexInst)
+    {
+        // implement the translation to spirv by waking up the use-def chain
+        // when we reach the leaf node (texture
+        // trace the use-def chain from nonuniformresource index def all the way to the
+        // textures / resources
+        // Decorate the following:
+        //   go through GEP and see if it calls an intrinsic function, decorate the address itself (GetElementPtr)
+        //   go through GEP to identify the pointer access and the Loads that it accesses (GetElementPtr -> Load)
+        //   go throuhg ExtractElement instruction for array access (GetElement)
```
ArielG-NV, line 1179:
```suggestion
        //   go through ExtractElement instruction for array access (GetElement)
```

source/slang/slang-ir-insts.h
```diff
@@ -327,6 +327,18 @@ struct IRRequireGLSLVersionDecoration : IRDecoration
...
+    IntegerLiteralValue getSPIRVVersion()
+    {
+        return getSPIRVVersionOperand()->value.intVal;
+    }
```
ArielG-NV, line 339: How is getSPIRVVersion used?
If it is just to get a SPIR-V version, this can be acquired from `linkage->targets[i]->getTargetCaps()`, where getTargetCaps() returns a capability set with the SPIR-V version being targeted (or at-least is supposed to at some point)

ArielG-NV:
The following code is legal hlsl. Something similar should probably should be added to test cases:
(referring to nested ``NonUniformResourceIndex(NonUniformResourceIndex(.*))`
```
RWStructuredBuffer<int> buffer[];

[numthreads(8, 1, 1)]
void main(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    RWStructuredBuffer<int> buffer2 = buffer[NonUniformResourceIndex(NonUniformResourceIndex(dispatchThreadID.x))];
    buffer2[0] = 1;
}
```

source/slang/slang-ir-insts.h
sriramm-nv: I updated this struct to return the nonuniform resource value as it is typically an integer index.

source/slang/hlsl.meta.slang
sriramm-nv: done

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: isResourceType is not available in spirv-legalize. could you please help me understand the meaning of this requirement?

source/slang/slang-emit-spirv.cpp
```diff
@@ -2774,59 +2809,79 @@ struct SPIRVEmitContext
...
+            result = emitImageLoad(parent, as<IRImageLoad>(inst));
+            break;
         case kIROp_ImageStore:
-            return emitImageStore(parent, as<IRImageStore>(inst));
+            result = emitImageStore(parent, as<IRImageStore>(inst));
+            break;
         case kIROp_ImageSubscript:
-            return emitImageSubscript(parent, as<IRImageSubscript>(inst));
+            result = emitImageSubscript(parent, as<IRImageSubscript>(inst)); break;
+            break;
```
nanokatze, line 2865: redundant break here

source/core/slang-string.h
```diff
@@ -510,6 +510,19 @@ namespace Slang
             return getData() + getLength();
         }
 
+        // Function to convert to std::string
+        std::string toStdString() const {
```
csyonghe, line 514: We don't use std::string in this codebase. Use String.

Unless there is any specific reasons, I don't think the basic things like string need to be changed at this stage of the project.

source/core/slang-string.h
```diff
@@ -510,6 +510,19 @@ namespace Slang
...
+            return std::string(begin(), end());
+        }
+
+        // Function to convert to C-style char* string
+        char* toCString() const {
```
csyonghe, line 519: This is just produceString().getBuffer()

source/slang/hlsl.meta.slang
```diff
@@ -6401,55 +6401,10 @@ T __copyObject<T>(T v)
...
```
csyonghe, line 6407: No indentation needed here.

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: Find the function and maybe declare it in slang-ir-util.h so you can use it here. The purpose is to avoid decorating on things that are irrelevant. For example no need to declare an int value as non uniform.

csyonghe (commented):
Need to clean up all debug dumps.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,116 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                    break;
+                case kIROp_RWStructuredBufferGetElementPtr:
+                    // Replace buffer[nonUniformRes(nonUniformRes(x))], into nonUniformRes(buffer[x])
+                    // This fails when converting rwstructbuf(nonuniformres(gep(...), int) -> rwstructbuf(get(...), int).
+                    // newUser = builder.emitRWStructuredBufferGetElementPtr(inst->getOperand(0), user->getOperand(1));
+                    break;
+                case kIROp_Load:
+                    newUser = builder.emitLoad(user->getFullType(), inst->getOperand(0));
+                    break;
+                case kIROp_Call:
```
csyonghe, line 1219: This seems dangerous, why is it needed?

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1822,6 +1932,13 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
             case kIROp_RWStructuredBufferStore:
                 processRWStructuredBufferStore(inst);
                 break;
+            case kIROp_NonUniformResourceIndex:
+                processNonUniformResourceIndex(inst);
+                {
```
csyonghe, line 1937: Put this inside processNonUniformRes.

source/slang/slang-legalize-types.cpp
```diff
@@ -203,6 +203,25 @@ bool isResourceType(IRType* type)
...
+{
+    if (isResourceType(type))
+        return true;
+    while (auto ptrType = as<IRPtrTypeBase>(type))
+    {
+        type = ptrType->getValueType();
+    }
+
+    // Special handling for RWStructuredBuf
+    if (const auto rwStructBufferType = as<IRStructType>(type))
```
csyonghe, line 218: Struct type is not rwstructured buffer. You should be looking for IRHLSLStructuredBufferTypeBase.

tests/compute/nonuniformres-array-of-textures.slang
```diff
@@ -0,0 +1,40 @@
...
+[numthreads(1, 1, 1)]
+void main(uint2 pixelIndex : SV_DispatchThreadID)
+{
+
+    // CHECK1-DAG: OpDecorate %[[VAR1:[a-zA-Z0-9_]+]] NonUniform
+    // CHECK1-DAG: OpDecorate %[[VAR2:[a-zA-Z0-9_]+]] NonUniform
+    // CHECK1-DAG: OpDecorate %[[VAR3:[a-zA-Z0-9_]+]] NonUniform
+    // CHECK1-DAG: OpDecorate %[[VAR4:[a-zA-Z0-9_]+]] NonUniform
+    // CHECK1-DAG: OpDecorate %[[VAR5:[a-zA-Z0-9_]+]] NonUniform
+    // CHECK1-DAG: OpDecorate %[[VAR6:[a-zA-Z0-9_]+]] NonUniform
```
jkwak-work, line 17: I don't think you need "-DAG" here when the result is deterministic in a certain order.

tests/compute/nonuniformres-nested-rwstructuredbuf.slang
```diff
@@ -0,0 +1,31 @@
+// vk-texture-indexing.slang
+
```
jkwak-work, line 2: Probably you want to delete these lines.

source/slang/slang-emit-c-like.cpp
```diff
@@ -2367,6 +2370,10 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
         m_writer->emit("GroupMemoryBarrierWithGroupSync()");
         break;
 
+    case kIROp_NonUniformResourceIndex:
+        emitOperand(inst->getOperand(0), getInfo(EmitOp::General)); // Directly emit NonUniformResourceIndex Operand0;
+        break;
+
```
jkwak-work, line 2376: Shouldn't you emit with the keyword "NonUniformResourceIndex" like following?
```
m_writer->emit("NonUniformResourceIndex(");
emitOperand(inst->getOperand(0), getInfo(EmitOp::General));
m_writer->emit(")");

```
It seems like we don't have a test targeting HLSL for "NonUniformResourceIndex".
Can you add one and make sure we don't regress?

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: It is for the case of 
```
buffer[NonUniformResourceIndex(0)].InterlockedAdd(0, 1);
```
which results in: 
```
kIROp_Call (171) "RWByteAddressBuffer.InterlockedAdd", kIROp_NonUniformResourceIndex, ... )
```
So, I added the above to deal with this. if I ignore it, then the first param `nonUniformResourceIndex(%1)` will result in a compilation failure.

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: done

tests/compute/nonuniformres-array-of-textures.slang
sriramm-nv: let me try to remove it

tests/compute/nonuniformres-nested-rwstructuredbuf.slang
sriramm-nv: done

source/slang/slang-legalize-types.cpp
sriramm-nv: The trace from debugger shows that RWStructBuf is saved as a `kIROp_StructType`
```
| type | 0x000001b38ef58150 {kIROp_StructType (102) 0x8ef58150} | Slang::IRType *
   | [op] | kIROp_StructType (102) | Slang::IROp
   | [UID] | 1093 | unsigned int
     | [type] | 0x000001b38ef580e0 {kIROp_TypeKind (48) 0x8ef580e0} | Slang::IRInst *
     | [name] | "RWStructuredBuffer" | char[18]
```

source/slang/slang-emit-c-like.cpp
sriramm-nv: this is just the base case, half expecting to fail, instead of silently working. hlsl, glsl has their own overloads of this. i'll add suitable tests.

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: But this doesn't seem right? I would say that in this case the IR:

%gep = GetElementAddr ...
%ld = load gep
%buffer = NonUniformIndex ld
IRCall %RWStructuredBuffer_InterlockedAdd buffer ...

Is the right form you want, right?

source/slang/slang-legalize-types.cpp
csyonghe: It is eventually lowered into a struct during legalize spirv or some legalize buffer pass. We need to make sure this is handled before they are lowered into structs to avoid doing incorrect handling with other ordinary structs.

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: This doesn't seem to be the pattern we get. The rwStructuredBufferGep is on the output (store), the input path is pretty simple:
```
        {674}   let  %32        : Int   = nonUniformResourceIndex(2 : Int)
        {1360}  call %RWByteAddressBufferx5FInterlockedAdd(%32, 0 : UInt, 1 : UInt)
        {1345}  let  %35        : UInt  = call %RWByteAddressBufferx5FLoad(0 : Int, 0 : Int)
        {910}   let  %36        : Float = castIntToFloat(%35)
```
The output path is little more tedious:
```
        {1486}  let  %29        : Vec(UInt, 3 : Int)    = load(%27)
        {1400}  let  %30        : UInt  = swizzle(%29, 0 : Int)
        {1401}  let  %31        : Int   = intCast(%30)
        {863}   let  %33        : UInt  = intCast(%31)
        {882}   let  %34        : Ptr(Float)    = rwstructuredBufferGetElementPtr(%output, %33)
        {911}   store(%34, %36)
```


Sharing the IR that we see in this example for context:
```
{666}   block %28:
        {1486}  let  %29        : Vec(UInt, 3 : Int)    = load(%27)
        {1400}  let  %30        : UInt  = swizzle(%29, 0 : Int)
        {1401}  let  %31        : Int   = intCast(%30)
        {674}   let  %32        : Int   = nonUniformResourceIndex(2 : Int)
        {1360}  call %RWByteAddressBufferx5FInterlockedAdd(%32, 0 : UInt, 1 : UInt)
        {1408}  SPIRVAsm _      : Void
        {
                {1409}  SPIRVAsmInst(225 : UInt, 1 : UInt, 2376 : UInt)
        }

        {863}   let  %33        : UInt  = intCast(%31)
        {882}   let  %34        : Ptr(Float)    = rwstructuredBufferGetElementPtr(%output, %33)
        {1345}  let  %35        : UInt  = call %RWByteAddressBufferx5FLoad(0 : Int, 0 : Int)
        {910}   let  %36        : Float = castIntToFloat(%35)
        {911}   store(%34, %36)
        {913}   return_val(void_constant)
}
```

This is quite different from the other example that you showed where the nonUniform indexing is done within the InterlockedAdd
```
    InterlockedAdd(texArray[NonUniformResourceIndex(dispatchThreadID.x)][uint2(0)], 2);
```

I feel like this is the only way we can deal with this scenario. It is used by the `tests/spirv/coherent-2.slang` example.

source/slang/slang-legalize-types.cpp
sriramm-nv: It is converted from `IRHLSLStructuredBufferTypeBase` into `IRStructType` in `processGlobalParam:lowerStructuredBufferType` within `legalizeSPIRV:processModue`
right before `processWorkList` where we deal with the `nonUniformResourceInfo`.

Is it ok to not do the struct conversion, if it's user (GEP) performs a NonUniformResource access? That way, we still have a rwStructBuffer type

For clarity, this is the sequence I'm talking about:
```
        {502}   let  %24        : UInt  = nonUniformResourceIndex(%23)
        [{541}  nameHint("buffer1")]
        {538}   let  %buffer1_  : RWStructuredBuffer(Int, DefaultLayout, %13)   = getElement(%buffer, %24)
        {544}   let  %25        : Ptr(Int)      = rwstructuredBufferGetElementPtr(%buffer1_, 0 : UInt)
```

After `lowerStructuredBufferType`, this gets converted into:
```
        {502}   let  %26        : UInt  = nonUniformResourceIndex(%25)
        {1002}  let  %27        : Ptr(%RWStructuredBuffer, 12 : Int)    = getElementPtr(%buffer, %26)
        {544}   let  %28        : Ptr(Int)      = rwstructuredBufferGetElementPtr(%27, 0 : UInt)
}
```

csyonghe: Type legalization happens way before spirv legalization so you should never need to worry about this in slang-legalize-types.cpp. When in this file, you can assume all structured buffer parameters are having the IRHLSLStructuredBufferBase type.

sriramm-nv: Ah I see your point. Even though the file is type-legalize, it is actually hijacked and being called from spirv-legalize. That is why, by that time, the original type is lost.

csyonghe (approved):
Looks good! Great job!

csyonghe (approved):

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: The change https://github.com/shader-slang/slang/pull/3965 makes this irrelevant issue. The problem of having to decorate a Call to an atomic function will be unnecessary if the call is inlined.

source/slang/slang-legalize-types.cpp
sriramm-nv: by making a more explicit check for struct buf, like `ptrType->getAddressSpace() == SpvStorageClassStorageBuffer` we can move past this problem.

sriramm-nv:
I updated a few comments to explain the logic better, and fixed some typos, and update the coherent-2 test case to check for NonUniform decoration

source/slang/hlsl.meta.slang
jkwak-work: This hasn't been addressed.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,125 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        // texture def all the way to the leaf operations. To be precise:
+        // - go through GEP and see if it calls an intrinsic function,
+        //   then decorate the address itself (GetElementPtr)
+        // - go through GEP to identify the pointer access and the Loads that it
+        //   accesses (GetElementPtr -> Load), then decorate the load instruction.
+        // - go through IntCasts to deal with u32 -> i32 / vice-versa (IntCast)
+        List<IRInst*> resWorkList;
+
+        // Handle cases when `nonUniformResourceIndexInst` inst is wrapped around
+        // an index in a nested fashion, i.e. nonUniform(nonUniform((index)) by
```
jkwak-work, line 1183: The typo on your comment got me confused for a few seconds.
Please remove one openning parenthesis.
```
nonUniform(nonUniform(index))
```

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1167,6 +1167,125 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        // only adding the inner-most inst in the worklist, and work our way out.
+        auto insti = nonUniformResourceIndexInst;
+        while (insti->getOp() == kIROp_NonUniformResourceIndex)
+        {
+            if (resWorkList.getCount() != 0)
+                resWorkList.removeLast();
+            resWorkList.add(insti);
+            insti = insti->getOperand(0);
+        }
```
jkwak-work, line 1192: Would this handle a cases like this?
```
nonUniform(nonUniform(index)+1)
```
I am not sure if it is a valid expression or not.

source/slang/hlsl.meta.slang
sriramm-nv: done

source/slang/slang-ir-inst-defs.h
```diff
@@ -1003,14 +1006,22 @@ INST(HighLevelDeclDecoration,               highLevelDecl,          1, 0)
         /// Recognized by SPIRV-emit pass so we can emit a SPIRV `Block` decoration.
     INST(SPIRVBlockDecoration, spvBlock, 0, 0)
 
+        /// Decorates a SPIRV-inst as [NonUniformResource] to guarantee non-uniform index lookup of
```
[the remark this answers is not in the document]
sriramm-nv: we just added a [NonUniformResource] decorator. but in this case, as a comment, I updated it with back tick

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: good catch

source/slang/slang-ir-spirv-legalize.cpp
sriramm-nv: seems legal to me, but it is not currently handled by the change.
this will look something like a 
```
GEP(Buffer, NonUniform(Add(NonUniform(Swizzle(Load, Int)), Int)) )
```

In the current state of this patch, the inner most nonUniform is not handled. @csyonghe do we need to address this?

jkwak-work:
I see that "DO NOT COMMIT" on one of your commits. Is it committed by a mistake? Probably not, I guess.

msriram:
> I see that "DO NOT COMMIT" on one of your commits. Is it committed by a mistake? Probably not, I guess.

ah, it is harmless, because there are other changes on top of this one that reverted the suspicious code. once the changes are squashed, this will no longer be a problem.

jkwak-work:
> Looks good. We really need a systematic pass over all texture/buffer intrinsic member functions to make sure they are [ForceInline]d for this to work.
> 
> @ArielG-NV @jkwak-work for viz.

I agree.
And as a forward-looking perspective, I think we may need to refactor to be more data-oriented.
If we have XML files that have information of which functions should have force-inline, as an example, it will improve a long term maintainability.
Similarly to the capability system, if we have XML files for which functions require what extensions or system, we will be apply them more systemically rather than hard-coding decorators to each functions.

Issue #3738, opened 2024-03-11
Support for struct fields with default values


natduca:
Came up in discussion to support struct fields with default values... 
struct Foo {
int a = 1;
};

PR #3907, opened 2024-04-08
Init expressions for struct fields support, #3738

ArielG-NV:
Following commit handles init expressions of struct's.

The general implementation follows C++ init expression rules for derived classes.

The logic was implemented after type resolution (`SemanticsDeclAttributesVisitor`):
1. Create a default constructor if missing.
2. Check all member variables (`this` and `super`) for if a member has an init expression, continue to *3* if found.
3. For each constructor, insert a member variable's init expression at the beginning of a constructor. This is to follow how C++ does construction of objects.

Some important notes about implementation:
* We must handle the scenario that there is inheritance. To handle the inheritance information processing `findLevelsOfInheritance` was created.
* If a user manually sets overload rank's of constructor expression's we have no way to assume new default constructor overload ranks.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -9308,8 +9308,161 @@ namespace Slang
             this, funcDecl, attr, DeclAssociationKind::PrimalSubstituteFunc);
     }
 
+    static int32_t findLevelsOfInheritance(AggTypeDecl* decl, int32_t inheritanceLevel, List<AggTypeDecl*>& linearInheritanceHierarchy)
+    {
+        // We need to assume multiple `InheritanceDecl` are possible per `decl` member list.
+        int32_t inheritanceDepth = inheritanceLevel;
+        for (auto maybeInheritance : decl->members)
+        {
+            auto inheritance = as<InheritanceDecl>(maybeInheritance);
+            if (!inheritance)
+                continue;
+            auto isValidInheritance = as<DeclRefType>(inheritance->base.type);
+            if (!isValidInheritance || !isValidInheritance->getDeclRef())
+                continue;
+            auto isInherited = as<AggTypeDecl>(isValidInheritance->getDeclRef().getDecl());
+            if (!isInherited)
+                continue;
+            linearInheritanceHierarchy.add(isInherited);
+            int32_t depthFound = findLevelsOfInheritance(isInherited, inheritanceLevel-1, linearInheritanceHierarchy);
+            if (depthFound < inheritanceDepth)
+                inheritanceDepth = depthFound;
+        }
+        return inheritanceDepth;
+    }
+
     void SemanticsDeclAttributesVisitor::visitStructDecl(StructDecl* structDecl)
     {
+        List<AggTypeDecl*> linearInheritanceHierarchy;
+        linearInheritanceHierarchy.add(structDecl);
+        int32_t levelsOfInheritance = findLevelsOfInheritance(structDecl, 0, linearInheritanceHierarchy);
+
+        List<ConstructorDecl*> ctorList;
+        bool fillCtorList = false;
+
+        auto findAllCtor = [&]() {
+            ConstructorDecl* ctorDefault = nullptr;
+            fillCtorList = true;
+            for (auto maybeCtor : structDecl->members)
+            {
+                auto ctor = as<ConstructorDecl>(maybeCtor);
+                if (!ctor)
+                    continue;
+                ctorList.add(ctor);
+
+                // It is possible BlockStmt has a child with the type of
+                // `ExpressionStmt` if an existing constructor has only 1
+                // expression. This would be a senario we need to
+                // put the `ExpressionStmt` inside a `SeqStmt`.
+                // Since we found an init expression, we need to ensure
+                // we are adding to a constructor which can handle multiple
+                // statments since we will be adding `init` expressions to
+                // the constructor.
+                auto block = as<BlockStmt>(ctor->body);
+                if (!as<SeqStmt>(block->body))
+                {
+                    auto tmpExpr = block->body;
+                    auto seqStmt = m_astBuilder->create<SeqStmt>();
+                    seqStmt->stmts.add(tmpExpr);
+                    block->body = seqStmt;
+                }
+                if (ctor->getParameters().getCount() != 0)
+                    continue;
+                ctorDefault = ctor;
+            }
+            if (!ctorDefault)
+            {
+                ctorDefault = m_astBuilder->create<ConstructorDecl>();
+                auto ctorName = getName("$init");
+                ctorDefault->ownedScope = m_astBuilder->create<Scope>();
+                ctorDefault->ownedScope->containerDecl = ctorDefault;
+                ctorDefault->ownedScope->parent = getScope(structDecl);
+                ctorDefault->parentDecl = structDecl;
+                ctorDefault->loc = structDecl->loc;
+                ctorDefault->closingSourceLoc = ctorDefault->loc;
+                ctorDefault->nameAndLoc.name = ctorName;
+                ctorDefault->nameAndLoc.loc = ctorDefault->loc;
+                ctorDefault->returnType.type = calcThisType(makeDeclRef(structDecl));
+                auto body = m_astBuilder->create<BlockStmt>();
+                body->scopeDecl = m_astBuilder->create<ScopeDecl>();
+                body->scopeDecl->ownedScope = m_astBuilder->create<Scope>();
+                body->scopeDecl->ownedScope->parent = getScope(ctorDefault);
+                body->scopeDecl->parentDecl = ctorDefault;
+                body->scopeDecl->loc = ctorDefault->loc;
+                body->scopeDecl->closingSourceLoc = ctorDefault->loc;
+                body->closingSourceLoc = ctorDefault->closingSourceLoc;
+                ctorDefault->body = body;
+                body->body = m_astBuilder->create<SeqStmt>();
+                structDecl->addMember(ctorDefault);
+                ctorList.add(ctorDefault);
+            }
+
+            // It is possible a struct is a child of another struct/related-type.
+            // Since we are creating a new constructor function, it is possible
+            // a parent and a child to both have `__init()` constructors.
+            // As a result, this can cause a compile error when relying on the default
+            // constructor of an object because a child can use a parent object's
+            // constructor.
+            // To solve this issue and produce C++ like behavior we need to add overload
+            // ranks if not present already to prioritize the default constructor of a child.
+            // If an overload rank is manually set, we cannot assume any preference of
+            // constructor by the user and should ignore any special logic.
+            if (ctorDefault->findModifier<OverloadRankAttribute>())
+                return;
+            auto overloadRank = m_astBuilder->create<OverloadRankAttribute>();
+            overloadRank->rank = -levelsOfInheritance;
+            addModifier(ctorDefault, overloadRank);
+        };
+
+        // To follow C++ constructor behavior when adding a init expression to a member
+        // variable we must insert init expressions to the start of all constructors.
+        // Furthermore, to follow C++ behavior we must also ensure we use init expressions
+        // of all accessable parent member variables.
+        Dictionary<Decl*, Expr*> cachedDeclToCheckedVar;
+        for (auto aggTypeDecl : linearInheritanceHierarchy)
+        for (auto m : aggTypeDecl->members)
+        {
+            auto varDeclBase = as<VarDeclBase>(m);
+            if (varDeclBase
+                && varDeclBase->initExpr)
+            {
+                if (!fillCtorList)
+                    findAllCtor();
+
+                for (auto ctor : ctorList)
+                {
+                    auto body = as<BlockStmt>(ctor->body);
+                    auto seqStmt = as<SeqStmt>(body->body);
+
+                    VarExpr* memberVarExpr = m_astBuilder->create<VarExpr>();
+                    memberVarExpr->scope = ctor->ownedScope;
+                    memberVarExpr->name = m->getName();
+                    memberVarExpr->type = nullptr;
+                    auto assign = m_astBuilder->create<AssignExpr>();
+                    assign->left = memberVarExpr;
+                    assign->right = varDeclBase->initExpr;
+                    auto stmt = m_astBuilder->create<ExpressionStmt>();
+                    stmt->expression = assign;
+
+                    Expr* checkedMemberVarExpr;
+                    if (cachedDeclToCheckedVar.containsKey(m))
+                        checkedMemberVarExpr = cachedDeclToCheckedVar[m];
+                    else
+                    {
+                        checkedMemberVarExpr = CheckTerm(memberVarExpr);
+                        cachedDeclToCheckedVar.add({ m, checkedMemberVarExpr });
+                    }
+                    if (!checkedMemberVarExpr->type.isLeftValue)
+                        continue;
+
+                    if (seqStmt->stmts.getCount() == 0)
+                        seqStmt->stmts.add(stmt);
+                    else
+                        seqStmt->stmts.insert(0, stmt);
+                }
+            }
+        }
+
         int backingWidth = 0;
         [[maybe_unused]]
         int totalWidth = 0;
--- a/tests/language-feature/struct-field-initializers/struct-field-initializer.slang
+++ b/tests/language-feature/struct-field-initializers/struct-field-initializer.slang
@@ -1,0 +1,60 @@
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain 
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -emit-spirv-directly
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-cpu -compute -entry computeMain
+//TEST(smoke,compute):COMPARE_COMPUTE(filecheck-buffer=BUF):-dx12 -use-dxil -compute -entry computeMain
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+static int myTwo = 2;
+static int myThree = 1+2;
+
+struct DefaultStructNoInit
+{
+    int data0;
+    int data1 = myTwo;
+    int data2 = 2;
+};
+struct DefaultStructWithInit
+{
+    int data0 = 3;
+    int data1 = myThree;
+    int data2;
+    __init()
+    {
+        data2 = 3;
+    }
+};
+struct DefaultStructWithInit2
+{
+    int data0 = 4;
+    int data1 = 1;
+    int data2 = 1;
+    __init()
+    {
+        data1 = 4;
+        data2 = 4;
+    }
+};
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    DefaultStructNoInit noInit = DefaultStructNoInit();
+    noInit.data0 = 2;
+    DefaultStructWithInit withInit = DefaultStructWithInit();
+    DefaultStructWithInit2 withInit2 = DefaultStructWithInit2();
+    // BUF: 1
+    outputBuffer[0] = true
+        && noInit.data0 == 2
+        && noInit.data1 == 2
+        && noInit.data2 == 2
+
+        && withInit.data0 == 3
+        && withInit.data1 == 3
+        && withInit.data2 == 3
+
+        && withInit2.data0 == 4
+        && withInit2.data1 == 4
+        && withInit2.data2 == 4
+        ;
+}
--- a/tests/language-feature/struct-field-initializers/struct-field-initializer-inherited.slang
+++ b/tests/language-feature/struct-field-initializers/struct-field-initializer-inherited.slang
@@ -1,0 +1,51 @@
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain 
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-vk -compute -entry computeMain -emit-spirv-directly
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=BUF):-cpu -compute -entry computeMain
+//TEST(smoke,compute):COMPARE_COMPUTE(filecheck-buffer=BUF):-dx12 -use-dxil -compute -entry computeMain
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+static int myOne = 1;
+
+public struct DefaultStruct_base
+{
+    int data0;
+    int data1 = myOne;
+    int data2 = 1;
+    int data3 = 1;
+};
+struct DefaultStruct : DefaultStruct_base
+{
+    int data4 = 3;
+    int data5 = 0;
+    __init()
+    {
+        data3 = data3 == 1;
+        data5 = 1;
+        data4 = data3;
+    }
+};
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    DefaultStruct_base initStructBase = DefaultStruct_base();
+    initStructBase.data0 = 1;
+
+    DefaultStruct initStruct = DefaultStruct();
+    initStruct.data0 = 1;
+    // BUF: 1
+    outputBuffer[0] = true
+        && initStructBase.data0 == 1
+        && initStructBase.data1 == 1
+        && initStructBase.data2 == 1
+        && initStructBase.data3 == 1
+
+        && initStruct.data0 == 1
+        && initStruct.data1 == 1
+        && initStruct.data2 == 1
+        && initStruct.data3 == 1
+        && initStruct.data4 == 1
+        && initStruct.data5 == 1;
+        ;
+}


source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+    static int32_t findLevelsOfInheritance(AggTypeDecl* decl, int32_t inheritanceLevel, List<AggTypeDecl*>& linearInheritanceHierarchy)
+    {
+        // We need to assume multiple `InheritanceDecl` are possible per `decl` member list.
+        int32_t inheritanceDepth = inheritanceLevel;
+        for (auto maybeInheritance : decl->members)
+        {
+            auto inheritance = as<InheritanceDecl>(maybeInheritance);
+            if (!inheritance)
+                continue;
+            auto isValidInheritance = as<DeclRefType>(inheritance->base.type);
```
kaizhangNV, line 9320: suggest not to use isXXX as the name for object, "is" prefix is usually used for bool variable.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+            if (!isValidInheritance || !isValidInheritance->getDeclRef())
+                continue;
+            auto isInherited = as<AggTypeDecl>(isValidInheritance->getDeclRef().getDecl());
+            if (!isInherited)
+                continue;
+            linearInheritanceHierarchy.add(isInherited);
+            int32_t depthFound = findLevelsOfInheritance(isInherited, inheritanceLevel-1, linearInheritanceHierarchy);
```
kaizhangNV, line 9327: I'm curious why you use negative value for the level?
Can't the logic become?
```
 int32_t depthFound = findLevelsOfInheritance(isInherited, inheritanceLevel + 1, linearInheritanceHierarchy);
inheritanceDepth = max(depthFound, inheritanceDepth)
```

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+        List<ConstructorDecl*> ctorList;
+        bool fillCtorList = false;
+
+        auto findAllCtor = [&]() {
+            ConstructorDecl* ctorDefault = nullptr;
+            fillCtorList = true;
+            for (auto maybeCtor : structDecl->members)
+            {
+                auto ctor = as<ConstructorDecl>(maybeCtor);
+                if (!ctor)
```
kaizhangNV, line 9349: we seems prefer to use like this
```
if (auto ctor = as<ConstructorDecl>(maybeCtor))
{
               ...
}
```

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+                // It is possible BlockStmt has a child with the type of
+                // `ExpressionStmt` if an existing constructor has only 1
+                // expression. This would be a senario we need to
+                // put the `ExpressionStmt` inside a `SeqStmt`.
+                // Since we found an init expression, we need to ensure
+                // we are adding to a constructor which can handle multiple
+                // statments since we will be adding `init` expressions to
+                // the constructor.
+                auto block = as<BlockStmt>(ctor->body);
+                if (!as<SeqStmt>(block->body))
```
kaizhangNV, line 9362: similarly, this can be simplified as

if (tmpExpr = as<SeqStmt>(block->body))
{
                    auto seqStmt = m_astBuilder->create<SeqStmt>();
                    seqStmt->stmts.add(tmpExpr);
                    block->body = seqStmt;
}

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+            addModifier(ctorDefault, overloadRank);
+        };
+
+        // To follow C++ constructor behavior when adding a init expression to a member
+        // variable we must insert init expressions to the start of all constructors.
+        // Furthermore, to follow C++ behavior we must also ensure we use init expressions
+        // of all accessable parent member variables.
+        Dictionary<Decl*, Expr*> cachedDeclToCheckedVar;
+        for (auto aggTypeDecl : linearInheritanceHierarchy)
+        for (auto m : aggTypeDecl->members)
```
kaizhangNV, line 9423: indent.

source/slang/slang-check-decl.cpp
ArielG-NV: I thought negative made sense (since we use this logic for getting a constructor overload ranking), but in hindsight positive is better and easier to read, I will change the code accordingly.

source/slang/slang-check-decl.cpp
ArielG-NV: The following code structure was done to avoid having 2-3 extra indentations in-favor for an early leave

source/slang/slang-check-decl.cpp
ArielG-NV: I can move the block scope definition into the if statement, I cannot remove the `!as<SeqStmt>(block->body)` check. The check ensures the as<BlockStmt>(ctor->body)->body is a `SeqStmt` to allow us to add more than 1 statement to the existing constructor for struct field init expression support.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,161 @@ namespace Slang
...
+                        checkedMemberVarExpr = cachedDeclToCheckedVar[m];
+                    else
+                    {
+                        checkedMemberVarExpr = CheckTerm(memberVarExpr);
+                        cachedDeclToCheckedVar.add({ m, checkedMemberVarExpr });
+                    }
+                    if (!checkedMemberVarExpr->type.isLeftValue)
+                        continue;
+
+                    if (seqStmt->stmts.getCount() == 0)
```
kaizhangNV: Do you have to check the getCount()?
Can't we just call `seqStmt->stmts.insert(0, stmt);`?

kaizhangNV:
I went through the change, and left some comments. But I'd like @csyonghe to give it another review.

kaizhangNV:
For the Falcor error: you should go ahead to check this file mentioned in the error report:
```
    (Error) GFX Error: dxc 1.6: C:/actions-runner/_work/slang/slang/FalcorBin/build/windows-vs2022/bin/Release/shaders/Scene/Displacement/DisplacementData.slang(31): error :  non-void function 'DisplacementData_x24init_0' should return a value [-Wreturn-type]
```

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,160 @@ namespace Slang
...
+            ctorList.add(ctor);
+
+            // It is possible BlockStmt has a child with the type of
+            // `ExpressionStmt` if an existing constructor has only 1
+            // expression. This would be a senario we need to
+            // put the `ExpressionStmt` inside a `SeqStmt`.
+            // Since we found an init expression, we need to ensure
+            // we are adding to a constructor which can handle multiple
+            // statments since we will be adding `init` expressions to
+            // the constructor.
```
jkwak-work, line 9349: Please make this sentence more readable.
I don't think I understood fully but it could be something like,
```
If a constructor can handle multiple statements, "init" is added to the constructor.
```

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,160 @@ namespace Slang
...
+            body->scopeDecl->loc = ctorDefault->loc;
+            body->scopeDecl->closingSourceLoc = ctorDefault->loc;
+            body->closingSourceLoc = ctorDefault->closingSourceLoc;
+            ctorDefault->body = body;
+            body->body = m_astBuilder->create<SeqStmt>();
+            structDecl->addMember(ctorDefault);
+            ctorList.add(ctorDefault);
+        }
+
+        // It is possible a struct is a child of another struct/related-type.
```
jkwak-work, line 9388: I think you meant "derived".

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,160 @@ namespace Slang
...
+        // Since we are creating a new constructor function, it is possible
+        // a parent and a child to both have `__init()` constructors.
+        // As a result, this can cause a compile error when relying on the default
+        // constructor of an object because a child can use a parent object's
+        // constructor.
+        // To solve this issue and produce C++ like behavior we need to add overload
+        // ranks if not present already to prioritize the default constructor of a child.
+        // If an overload rank is manually set, we cannot assume any preference of
+        // constructor by the user and should ignore any special logic.
```
jkwak-work, line 9397: I am not sure what this means.
Judging by the implementation, it sounds like,
```
If an overload rank is manually set, we need to respect the overload-rank from the user.
```
There is no point of having a comment if I have to deduce the meaning from the implementation.
The comment is to help the reader to understand the implementation; not the other way around.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,279 @@ namespace Slang
             this, funcDecl, attr, DeclAssociationKind::PrimalSubstituteFunc);
     }
 
+    struct LinearInheritanceGraphWithInfo
```
csyonghe, line 9311: There is already a getInheritanceInfo or similarly named function that return you a linear use list of type facets of a given type, including all transitive bases and extensions. We should just use that.

ArielG-NV: I will switch the "getting inheritance" part of the logic over, the auxiliary processing + data-fetching will still need to be a separate pass.

ArielG-NV:
note for debugging: 
test_BSDFViewer_d3d12 fails sometimes.
test_SceneDebugger_d3d12 fails sometimes.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,272 @@ namespace Slang
...
+        body->scopeDecl->loc = ctor->loc;
+        body->scopeDecl->closingSourceLoc = ctor->loc;
+        body->closingSourceLoc = ctor->closingSourceLoc;
+        ctor->body = body;
+        body->body = m_astBuilder->create<SeqStmt>();
+        decl->addMember(ctor);
+        return ctor;
+    }
+
+    static SeqStmt* legalizeCtorBlockToASeqStmt(ASTBuilder* m_astBuilder, BlockStmt* stmt)
```
csyonghe, line 9406: Rename to ensureCtorBodyIsSeqStmt

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,272 @@ namespace Slang
...
+                aggTypeInfo->ctorList.add(ctor);
+                if (ctor->members.getCount() != 0)
+                    continue;
+                aggTypeInfo->defCtor = ctor;
+            }
+        }
+        return inheritanceDepth;
+    }
+
+    static ConstructorDecl* _createCtor(SemanticsDeclAttributesVisitor* visitor, ASTBuilder* m_astBuilder, AggTypeDecl* decl)
```
csyonghe, line 9379: Why is this done in SemanticsAttributeVisitor? I feel the ctor signature synthesis should be done earlier in SemanticsDeclHeaderVisitor, and the ctor body synthesis can be done later in SemanticsDeclBodyVisitor.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,272 @@ namespace Slang
...
+            auto seqStmt = m_astBuilder->create<SeqStmt>();
+            seqStmt->stmts.add(tmpExpr);
+            stmt->body = seqStmt;
+            return seqStmt;
+        }
+        return as<SeqStmt>(stmt->body);
+    }
+    static void legalizeCtorOverloadRank(ASTBuilder* m_astBuilder, ConstructorDecl* ctor, Facet::DirectnessVal rank)
+    {
+        // It is possible a "this" has a "super". Since we are creating a new constructor
```
csyonghe, line 9427: Instead of hacking this with overloadRank, we should implement the proper logic in resolveOverload to prefer the method from a more derived type. Note that this is necessary not only for constructors but also other member methods.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,272 @@ namespace Slang
...
+        }
+
+        auto overloadRankMod = m_astBuilder->create<OverloadRankAttribute>();
+        overloadRankMod->rank = rank;
+        addModifier(ctor, overloadRankMod);
+    }
     void SemanticsDeclAttributesVisitor::visitStructDecl(StructDecl* structDecl)
     {
+        // We need the data of our (maybe) derived "this" and (if exists) "super".
+        // We also need the total inheritance depth to properly set overload ranks
```
csyonghe, line 9446: Why can't we just look at all InheritanceDecl members and call the base types constructor? That way we don't need to go all the way down to every leaf super type.

source/slang/slang-check-decl.cpp
```diff
@@ -9308,8 +9308,272 @@ namespace Slang
...
+                        auto seqStmt = legalizeCtorBlockToASeqStmt(m_astBuilder, as<BlockStmt>(ctor->body));
+
+                        VarExpr* memberVarExpr = m_astBuilder->create<VarExpr>();
+                        memberVarExpr->scope = ctor->ownedScope;
+                        memberVarExpr->name = m->getName();
+                        memberVarExpr->type = nullptr;
+                        auto assign = m_astBuilder->create<AssignExpr>();
+                        assign->left = memberVarExpr;
+                        assign->right = varDeclBase->initExpr;
+                        auto stmt = m_astBuilder->create<ExpressionStmt>();
```
csyonghe, line 9548: Stmt->loc = m->loc

source/slang/slang-check-impl.h
```diff
@@ -633,10 +633,10 @@ namespace Slang
         FunctionDifferentiableLevel getFuncDifferentiableLevel(FunctionDeclBase* func);
 
             /// Get the processed inheritance information for `type`, including all its facets
-        InheritanceInfo getInheritanceInfo(Type* type);
+        InheritanceInfo getInheritanceInfo(Type* type, Facet::DirectnessVal inheritanceDepth);
```
csyonghe, line 636: I am not sure it is necessary to track inheritanceDepth here.

ArielG-NV: > Instead of hacking this with overloadRank, we should implement the proper logic in resolveOverload to prefer the method from a more derived type. Note that this is necessary not only for constructors but also other member methods.

Once I address the `resolveOverload` logic directly, we should be able to forgo any changes to ` getInheritanceInfo` and related logic

source/slang/slang-check-decl.cpp
ArielG-NV: The code was done in `SemanticsAttributeVisitor` to ensure we created CTor the earliest possible after type resolution. 

Splitting should work, all we need is the ability to delete a ctor if unused + access to parent member list.

I suppose we can:
1. Create a default ctor if missing in `SemanticsDeclHeaderVisitor`.
2. During `SemanticsDeclBodyVisitor` we can fill the body, if it won't be filled we delete.

source/slang/slang-check-overload.cpp
```diff
@@ -1013,6 +1013,29 @@ namespace Slang
...
+        // at parameters, which is resolved later.
+        if(left.declRef.getParent() == right.declRef.getParent())
+            return 0;
+
+        // If both are members of an inherited decl, prefer the more derived inheritance member
+        auto leftStruct = left.declRef.getParent().as<StructDecl>();
+        auto rightStruct = right.declRef.getParent().as<StructDecl>();
+        if (leftStruct && rightStruct)
+        {
+            auto leftType = DeclRefType::create(m_astBuilder, leftStruct);
```
csyonghe, line 1026: This step drops the generic substitution arguments from the parent type. DeclRef::getParent should give you a decl ref to the parent that has the generic arguments in the case if a generic specialization, we need to create a DeclRef type from that full DeclRef instead of just a decl.

source/slang/slang-check-overload.cpp
```diff
@@ -1013,6 +1013,29 @@ namespace Slang
...
+            auto inheritanceInfo = getShared()->getInheritanceInfo(leftType);
+            for (auto facet : inheritanceInfo.facets)
+                if (facet.getImpl()->getType()->equals(rightType))
+                    return -1;
+            inheritanceInfo = getShared()->getInheritanceInfo(rightType);
+            for (auto facet : inheritanceInfo.facets)
+                if (facet.getImpl()->getType()->equals(rightType))
+                    return 1;
+        }
+
```
csyonghe, line 1038: Can remove the following comment.

source/slang/slang-lookup.cpp
```diff
@@ -449,6 +449,12 @@ static void _lookUpMembersInSuperTypeDeclImpl(
             if (int(request.options) & int(LookupOptions::IgnoreBaseInterfaces))
                 continue;
         }
+        // If we are looking up only immediate members, ignore non "Self" facets
+        else if (int(request.options) & int(LookupOptions::IgnoreInheritance)
```
csyonghe, line 453: What about extensions?

struct A : B
{ int member1; }

extension A { int member2; }

Do we want to consider member2?

tests/language-feature/higher-order-functions/overload-different-param-count.slang
```diff
@@ -0,0 +1,59 @@
+//TEST:SIMPLE(filecheck=CHECK): -target glsl -entry computeMain -stage compute
```
csyonghe, line 1: What is this testing?

tests/language-feature/struct-field-initializers/struct-field-initializer.slang
```diff
@@ -0,0 +1,60 @@
...
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    DefaultStructNoInit noInit = DefaultStructNoInit();
+    noInit.data0 = 2;
+    DefaultStructWithInit withInit = DefaultStructWithInit();
+    DefaultStructWithInit2 withInit2 = DefaultStructWithInit2();
+    // BUF: 1
+    outputBuffer[0] = true
+        && noInit.data0 == 2
```
csyonghe, line 48: Why? Should be 0?

source/slang/slang-check-decl.cpp
```diff
@@ -7261,12 +7307,172 @@ namespace Slang
...
+        for (auto& declInfo : inheritanceDefaultCtorList)
+        {
+            if (!declInfo.defaultCtor)
+                continue;
+            for (auto ctor : structDeclInfo.ctorList)
+            {
+                auto seqStmt = _ensureCtorBodyIsSeqStmt(m_astBuilder, ctor);
+                auto ctorName = declInfo.defaultCtor->getName();
+
+                auto lookupResult = lookUpMember(
```
csyonghe, line 7399: Isn't this just going to return ctor? In that case why not just create a VarRxpr whose DeclRef is ctor?

Seems like the two loops here should swap? Loop on every ctor, and inside it loop on every base?

source/slang/slang-check-decl.cpp
```diff
@@ -7261,12 +7307,172 @@ namespace Slang
...
+                thisExpr->type = ctor->returnType.type;
+
+                auto assign = m_astBuilder->create<AssignExpr>();
+                assign->left = coerce(CoercionSite::Initializer, declInfo.defaultCtor->returnType.type, thisExpr);
+                assign->right = invoke;
+                auto stmt = m_astBuilder->create<ExpressionStmt>();
+                stmt->expression = assign;
+                stmt->loc = ctor->loc;
+
+                seqStmt->stmts.insert(insertOffset, stmt);
```
csyonghe, line 7422: Avoid insert this many times. Create a single SeqStmt to hold all new stmts and simply insert that seqStmt at top.

source/slang/slang-check-decl.cpp
```diff
@@ -7261,12 +7307,172 @@ namespace Slang
...
+                    checkedMemberVarExpr = cachedDeclToCheckedVar[m];
+                else
+                {
+                    checkedMemberVarExpr = CheckTerm(memberVarExpr);
+                    cachedDeclToCheckedVar.add({ m, checkedMemberVarExpr });
+                }
+                if (!checkedMemberVarExpr->type.isLeftValue)
+                    continue;
+
+                seqStmt->stmts.insert(insertOffset, stmt);
```
csyonghe, line 7461: Same thing, don't insert for each member.

tests/language-feature/higher-order-functions/overload-different-param-count.slang
ArielG-NV: overloads for when parameter count is different. It was one of the failures I encountered, so I added a test for it.

tests/language-feature/struct-field-initializers/struct-field-initializer.slang
ArielG-NV: `noInit.data0 = 2;` was places earlier in call scope to test that the init logic does not break order of initialization. `noInit.data0 == 2` returns true

source/slang/slang-check-decl.cpp
ArielG-NV: If we swap the loop, we will do more `loop`s despite computing the same work.
The current layout allows `if (!declInfo.defaultCtor) continue;` to skip an inner loop.

source/slang/slang-check-decl.cpp
ArielG-NV: I will do the following for the members since it will reduce inserts by N, where N is number of init expression members)

I think we should not do the following for init expressions of parents since we only insert 0-1 statements per function. We are going to have to allocate one SeqStmt+N-loop-of-elements regardless.

source/slang/slang-check-decl.cpp
ArielG-NV: I understand why the following was suggested, it was in reference to: 
> Avoid insert this many times. Create a single SeqStmt to hold all new stmts and simply insert that seqStmt at top.

ArielG-NV: I will change the order, it will be faster to create 1 seqStmt then insert twice probably

source/slang/slang-check-overload.cpp
ArielG-NV: I will add a test for this as well (override across generics)

source/slang/slang-lookup.cpp
ArielG-NV: I will add a test for this and implement accordingly

ArielG-NV: I was looking into how to implement the following, I believe the least "hacked" code would be to have `IgnoreInheritance`, and use lookup to find a default constructor.

csyonghe:
I think the right behavior is to always synthesize a default constructor if it is not defined directly in the struct. We should disallow defining default ctor in an extension, because this can create inconsistent behavior for modules that sees the extension and the modules without the extension, which is not desired.

If a struct does not have a default ctor, we always synthesize one. If the user attempts to extend a type with default ctor, it shall be an error.

ArielG-NV:
> I think the right behavior is to always synthesize a default constructor if it is not defined directly in the struct. We should disallow defining default ctor in an extension, because this can create inconsistent behavior for modules that sees the extension and the modules without the extension, which is not desired.
> 
> If a struct does not have a default ctor, we always synthesize one. If the user attempts to extend a type with default ctor, it shall be an error.

```
__magic_type(NativeStringType)
__intrinsic_type($(kIROp_NativeStringType))
struct NativeString
{
    __target_intrinsic(cpp, "int(strlen($0))")
    int getLength();

    __target_intrinsic(cpp, "(void*)((const char*)($0))")
    Ptr<void> getBuffer();

    property int length { [__unsafeForceInlineEarly] get{return getLength();} }

    __intrinsic_op($(kIROp_getNativeStr))
    __init(String value);
};

extension Ptr<void>
{
    __implicit_conversion($(kConversionCost_PtrToVoidPtr))
    [__unsafeForceInlineEarly]
    __init(NativeString nativeStr) { this = nativeStr.getBuffer(); }

    __generic<T>
    __intrinsic_op(0)
    __implicit_conversion($(kConversionCost_PtrToVoidPtr))
    __init(Ptr<T> ptr);

    __generic<T>
    __intrinsic_op(0)
    __implicit_conversion($(kConversionCost_PtrToVoidPtr))
    __init(NativeRef<T> ptr);
}
```
we currently use extensions of structs to define constructors in our code.

source/slang/slang-ast-support-types.h
```diff
@@ -1181,6 +1181,7 @@ namespace Slang
         /// "under-construction" and not being checked, then it's safe to
         /// consider all names we've inserted so far. This is used when
         /// checking to see if a keyword is shadowed.
+        IgnoreInheritance = 1 << 4, //< Lookup only non inheritance children of a struct (including `extension`)
```
jkwak-work, line 1184: You need three slashes for the Doxygen comment.
See the document of [Doxygen](https://www.doxygen.nl/manual/docblocks.html) "Putting documentation after members" section.

tests/language-feature/struct-field-initializers/struct-field-initializer-import.slang
```diff
@@ -0,0 +1,26 @@
...
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    DefaultStructNoInit noInit;
+    DefaultStructNoInit2 noInit2;
+    modifyOut(noInit2);
+    // BUF: 1
+    outputBuffer[0] = true
+        && noInit.data0 == 2
+        && noInit.data0 == 2
+        ;
```
jkwak-work, line 25: Did you miss "2" on the second one?
```
    outputBuffer[0] = true
        && noInit.data0 == 2
        && noInit2.data0 == 2
        ;
```

tests/language-feature/struct-field-initializers/struct-field-initializer-extension.slang
```diff
@@ -0,0 +1,56 @@
...
+        data0 = 3;
+        data2 = 3;
+    }
+}
+
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    DefaultStructNoInit noInit = DefaultStructNoInit();
+    DefaultStructWithInit withInit = DefaultStructWithInit();
```
jkwak-work, line 43: I am little worried that there is no testing for the constructor that takes an argument.
Maybe you can modify this test as following,
```
extension DefaultStructWithInit
{
    __init(int indata)
    {
        data0 = indata; // <== Changes here
        data2 = 0; // <== Changes here
    }
    __init()
    {
        data0 = 3;
        data2 = 3;
    }
}

[numthreads(1, 1, 1)]
{
    DefaultStructNoInit noInit = DefaultStructNoInit();
    DefaultStructWithInit withInit0 = DefaultStructWithInit();
    DefaultStructWithInit withInit1 = DefaultStructWithInit(4); // <== Changes here
```

tests/language-feature/struct-field-initializers/struct-field-initializer.slang
```diff
@@ -0,0 +1,60 @@
...
+{
+    int data0 = 4;
+    int data1 = 1;
+    int data2 = 1;
+    __init()
+    {
+        data1 = 4;
+        data2 = 4;
+    }
+};
```
jkwak-work, line 38: I think you should check more of the "conflicting" cases like this one.
This test shows the conflicts of the default value and the values set in the constructor.
I think you can find more conflicting cases like this.
One of such cases i can think of is when the deriving class try to set the value when the value is already set by its ancenstor.
```
struct Parent
{
    bool watchTVWhileEating = false;
};
struct BadChild : Parent
{
    __init()
    {
        watchTVWhileEating = true;
    }
};
```

csyonghe:
> > I think the right behavior is to always synthesize a default constructor if it is not defined directly in the struct. We should disallow defining default ctor in an extension, because this can create inconsistent behavior for modules that sees the extension and the modules without the extension, which is not desired.
> 
> > 
> 
> > If a struct does not have a default ctor, we always synthesize one. If the user attempts to extend a type with default ctor, it shall be an error.
> 
> 
> 
> ```
> 
> __magic_type(NativeStringType)
> 
> __intrinsic_type($(kIROp_NativeStringType))
> 
> struct NativeString
> 
> {
> 
>     __target_intrinsic(cpp, "int(strlen($0))")
> 
>     int getLength();
> 
> 
> 
>     __target_intrinsic(cpp, "(void*)((const char*)($0))")
> 
>     Ptr<void> getBuffer();
> 
> 
> 
>     property int length { [__unsafeForceInlineEarly] get{return getLength();} }
> 
> 
> 
>     __intrinsic_op($(kIROp_getNativeStr))
> 
>     __init(String value);
> 
> };
> 
> 
> 
> extension Ptr<void>
> 
> {
> 
>     __implicit_conversion($(kConversionCost_PtrToVoidPtr))
> 
>     [__unsafeForceInlineEarly]
> 
>     __init(NativeString nativeStr) { this = nativeStr.getBuffer(); }
> 
> 
> 
>     __generic<T>
> 
>     __intrinsic_op(0)
> 
>     __implicit_conversion($(kConversionCost_PtrToVoidPtr))
> 
>     __init(Ptr<T> ptr);
> 
> 
> 
>     __generic<T>
> 
>     __intrinsic_op(0)
> 
>     __implicit_conversion($(kConversionCost_PtrToVoidPtr))
> 
>     __init(NativeRef<T> ptr);
> 
> }
> 
> ```
> 
> we currently use extensions of structs to define constructors in our stdlib code?

I mean just default ctors.

tests/language-feature/struct-field-initializers/struct-field-initializer-extension.slang
ArielG-NV: This test, tests a failure of using default ctor as an extension and to be sure Slang does not crash (for language server). The goal is to fail if using a default ctor in `extension`.

I will add a separate test for parameter functions in `extension`s which are designed to run. I will also add a test for constructor with parameter to ensure it also works as intended.

source/slang/slang-check-decl.cpp
```diff
@@ -1806,6 +1807,112 @@ namespace Slang
...
+        body->closingSourceLoc = ctor->closingSourceLoc;
+        ctor->body = body;
+        body->body = m_astBuilder->create<SeqStmt>();
+        decl->addMember(ctor);
+        return ctor;
+    }
+
+    static ConstructorDecl* _getDefaultCtor(ASTBuilder* m_astBuilder, SemanticsVisitor* visitor, StructDecl* structDecl)
+    {
+        // we may have a default ctor inside an extension,
```
csyonghe, line 1839: I think we need to disregard default ctors came from extensions, so this should just be a member search within the structDecl.

source/slang/slang-check-decl.cpp
```diff
@@ -7261,12 +7368,163 @@ namespace Slang
...
+        {
+            auto seqStmt = _ensureCtorBodyIsSeqStmt(m_astBuilder, ctor);
+            auto seqStmtChild = m_astBuilder->create<SeqStmt>();
+            seqStmtChild->stmts.reserve(inheritanceDefaultCtorList.getCount());
+            for (auto& declInfo : inheritanceDefaultCtorList)
+            {
+                if (!declInfo.defaultCtor)
+                    continue;
+
+                auto ctorToInvoke = m_astBuilder->create<VarExpr>();
```
csyonghe, line 7445: We we are forming a VarExpr like this, we need to make sure to fill in its type as well so it is the same as any other checked expr.

source/slang/slang-check-decl.cpp
```diff
@@ -1816,6 +1817,86 @@ namespace Slang
...
+        }
+        return nullptr;
+    }
+
+
+    static List<ConstructorDecl*> _getCtorList(ASTBuilder* m_astBuilder, SemanticsVisitor* visitor, StructDecl* structDecl, ConstructorDecl** defaultCtorOut)
+    {
+        List<ConstructorDecl*> ctorList;
+
+        auto lookupDefaultConstructor = lookUpMember(
```
csyonghe, line 1864: rename to ctorLookupResult.

This is not just about default ctor.

source/slang/slang-check-decl.cpp
```diff
@@ -1816,6 +1817,86 @@ namespace Slang
...
+            visitor->getName("$init"),
+            DeclRefType::create(m_astBuilder, structDecl),
+            structDecl->ownedScope,
+            LookupMask::Function,
+            LookupOptions::IgnoreInheritance);
+
+        if (!lookupDefaultConstructor.isValid())
+            return ctorList;
+
+        auto parseLookupItemLogic = [&](LookupResultItem& item)
```
csyonghe, line 1876: parse has a specific meaning that is concerting from string to structural representation.

Here should use handle/process.

source/slang/slang-check-decl.cpp
```diff
@@ -7271,12 +7352,163 @@ namespace Slang
...
+            DeclAndCtorInfo(ASTBuilder* m_astBuilder, SemanticsVisitor* visitor, StructDecl* parent, const bool getOnlyDefault)
+            {
+                if (getOnlyDefault)
+                    defaultCtor = _getDefaultCtor(parent);
+                else
+                    ctorList = _getCtorList(m_astBuilder, visitor, parent, &defaultCtor);
+            }
+        };
+        List<DeclAndCtorInfo> inheritanceDefaultCtorList{};
+        for (auto member : structDecl->members)
```
csyonghe, line 7402: struct decl->getMembersOfType<InheritanceDecl>()

source/slang/slang-check-decl.cpp
```diff
@@ -7271,12 +7352,163 @@ namespace Slang
...
+            if (!declRefType)
+                continue;
+            auto structOfInheritance = as<StructDecl>(declRefType->getDeclRef().getDecl());
+            if (!structOfInheritance)
+                continue;
+            inheritanceDefaultCtorList.add(DeclAndCtorInfo(m_astBuilder, this, structOfInheritance, true));
+        }
+        DeclAndCtorInfo structDeclInfo = DeclAndCtorInfo(m_astBuilder, this, structDecl, false);
+
+        std::size_t insertOffset = 0;
```
csyonghe, line 7417: Size

csyonghe (approved):

csyonghe:
Looks good, only minor suggestions.

It is great to see this coming together. Amazing job!

csyonghe (approved):

csyonghe:
Remove wip from pr before submitting.

PR #3912, opened 2024-04-09
Initial pass to add capability declarations to stdlib intrinsics.

### LINKED ISSUE #3978: Add compiler option to disable capability-related diagnostic errors
(PR already in progress)

ArielG-NV:
Initial implementation of capabilities

- Added capabilities as first in all attribute lists of a functions/types. This has been done mostly because capabilities may apply on a attribute instead of the target function if not first. The other reason of this approach is to keep the coding style consistent.

- Tagged all functions appropriately. Some code has not been tagged because the code is not using any GLSL/sm version (calls 0 intrinsic's)

- Added `default` support for `__target_switch` with capabilities.

- Fixed bug inside intrinsic-basic.slang. Shader attempts to compile itself as fragment when it is meant to compile as compute.

- Added to tests to diagnose and fix compile errors the new capabilities introduced.

- Did not remove `__target_intrinsic` code yet, some placeholder code was placed to allow the code to compile. There is a future issue planned to deal with the removal of these functions where possible.

- Did not account for 'double' operations universally requiring GLSL400 for stdlib. Also did not account for some "uint"/"int" operations requiring GLSL450 for stdlib. This likely will need to be done compiler side or with a more advanced capability system. 

Closes #3978


source/slang/slang-capabilities.capdef
```diff
@@ -397,3 +448,83 @@ alias DX_6_4 = sm_6_4;
...
+alias appendstructuredbuffer = sm_5_0 + shader_stages_fra_com;
+alias atomic_hlsl = _sm_4_0 | GLSL_430_SPIRV_1_0;
+alias atomic_hlsl_nvapi = _sm_4_0 + hlsl_nvapi;
+alias atomic_hlsl_sm_6_6 = _sm_6_6;
+alias byteaddressbuffer = sm_4_0;
+alias byteaddressbuffer_rw = sm_4_0 + shader_stages_fra_com;
+alias consumestructuredbuffer = sm_5_0 + shader_stages_fra_com;
+alias fragmentprocessing = fragment + sm_5_0;
+alias fragmentprocessing_derivativecontrol = fragment + sm_5_0 + GL_ARB_derivative_control;
+alias getattributeatvertex = fragment + _sm_6_1 | fragment + GL_EXT_fragment_shader_barycentric;
```
jkwak-work, line 463: Wouldn't it be better with underscore characters in the names?
```
append_structured_buffer
consume_structured_buffer
fragment_processing
get_attribute_at_vertex
```

source/slang/slang-capabilities.capdef
```diff
@@ -397,3 +448,83 @@ alias DX_6_4 = sm_6_4;
...
+alias memorybarrier_com = compute + sm_5_0;
```
jkwak-work, line 464: what is "com"? What does it stand for? compute?

source/slang/slang-capabilities.capdef
```diff
@@ -397,3 +448,83 @@ alias DX_6_4 = sm_6_4;
...
+alias image_samples = GLSL_450 | spirv_1_0 + spvImageQuery | sm_4_1;
+alias image_query = GLSL_430 | GLSL_430 + spvImageQuery;
+alias image_loadstore = GL_EXT_shader_image_load_store + GLSL_420;
+alias nonuniformqualifier = sm_5_1;
+alias printf = GL_EXT_debug_printf | _sm_4_0 | _cuda_sm_2_0 | cpp;
+alias texturefootprint = GL_NV_shader_texture_footprint + GLSL_450 | hlsl_nvapi + _sm_4_0;
+alias texturefootprintclamp = texturefootprint + GL_ARB_sparse_texture_clamp;
+alias texture_shadowlod = GL_EXT_texture_shadow_lod + GLSL_140;
+alias texture_cube = GLSL_400;
+alias shader5_sm_4_0 = GL_ARB_gpu_shader5 | _sm_4_0 | _cuda_sm_2_0 | cpp;
```
jkwak-work, line 487: Probably this is correct but I am wondering if this uses GL_ARB_gpu_shader5 not 4?

jkwak-work (commented):
It is difficult to review on the webpage when there are too many lines modified.
My web-browser freezes.
I will try to review on my local repo later.

source/slang/slang-check-decl.cpp
```diff
@@ -9564,6 +9564,13 @@ namespace Slang
             CapabilitySet set;
             for (auto targetCase : stmt->targetCases)
             {
+                // We may revieve a `default:` case for a `__target_switch`. If this is the case,
+                // we must resolve the target capability as `any_target` to hint to the capability 
+                // system that the following function must be defined on all codegen targets. If
+                // we do not handle `default:`, the codegen will fail when trying to find a specific
+                // codegen target not handled explicitly by a `case` statment.
+                if (targetCase->capability == 0)
```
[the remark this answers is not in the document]
ArielG-NV: yes. Default is the only 'null' capability

tests/glsl-intrinsic/intrinsic-basic.slang
```diff
@@ -1,8 +1,8 @@
-//TEST:SIMPLE(filecheck=CHECK_GLSL): -allow-glsl -stage fragment -entry computeMain -target glsl
-//TEST:SIMPLE(filecheck=CHECK_SPV):  -allow-glsl -stage fragment -entry computeMain -target spirv
-//TEST:SIMPLE(filecheck=CHECK_HLSL): -allow-glsl -stage fragment -entry computeMain -target hlsl
-//TEST:SIMPLE(filecheck=CHECK_CUDA): -allow-glsl -stage fragment -entry computeMain -target cuda -DTARGET_CUDA
-//TEST:SIMPLE(filecheck=CHECK_CPP):  -allow-glsl -stage fragment -entry computeMain -target cpp
+//TEST:SIMPLE(filecheck=CHECK_GLSL): -allow-glsl -stage compute -entry computeMain -target glsl
+//TEST:SIMPLE(filecheck=CHECK_SPV):  -allow-glsl -stage compute -entry computeMain -target spirv
+//TEST:SIMPLE(filecheck=CHECK_HLSL): -allow-glsl -stage compute -entry computeMain -target hlsl
+//TEST:SIMPLE(filecheck=CHECK_CUDA): -allow-glsl -stage compute -entry computeMain -target cuda -DTARGET_CUDA
+//TEST:SIMPLE(filecheck=CHECK_CPP):  -allow-glsl -stage compute -entry computeMain -target cpp
```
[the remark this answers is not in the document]
ArielG-NV: The test is a compute shader.
```
[numthreads(4, 1, 1)]
{
```
This test will not compile with capabilities since we require compute shaders for `numthreads`, if compiling as fragment shader this is invalid and compiling will now fail with the capabilities system.

source/slang/slang-capabilities.capdef
ArielG-NV: I am open for any suggestions on naming and welcome them. the existing naming scheme avoided `_`,  so I decided to just have a placeholder style which can be changed later and made something which could easily be regex'ed for modification.

source/slang/slang-capabilities.capdef
ArielG-NV: for shader stages since we can have very long names, I for-now used _com for compute, _fra for fragment, _ray for raygen, _any for anyhit, etc.

I welcome any suggestions to improve the naming to keep it short and concise.

source/slang/slang-capabilities.capdef
ArielG-NV: glsl 130 implements shader4 (mostly just adds Uint and Int support).

Slang only supports Vulkan officially for now. [As per Vulkan GLSL spec](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_vulkan_glsl.txt) anything at or below GLSL 140 is assumed to be supported if using Vulkan.

For old OpenGL support (if Slang is to support GLSL 130), it would mostly rely on the compiler enabling shader4 if Uint/Int types are found for some special annotated functions.

source/slang/slang-capabilities.capdef
jkwak-work: "com" and "any" could mean too many things. I don't see why we cannot use the full words like "_compute", "_fragment" and "_anyhit".

Another aspect is that if the name is distinctive, it is easy to replace the words with "vim" or any text replacing features. If the words are too short and common like "com" or "any", it will be difficult to replace them later.

jkwak-work: One of my rules for naming, for your information, is that when the scope of the variable is short, I go with short names. When the scope is larger, I use more descriptive and longer names.

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
jkwak-work: Is there a way to avoid using a magic number?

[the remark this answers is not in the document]
ArielG-NV: We would need to add a bool/enum flag for `case` type.

To elaborate, the way the target intrinsic system works internally is like a if-else tree, not like a c++ switch statment. The first satisfied `case` is where we specialize our function. This includes `default`, which is essentially just a `if(true)`.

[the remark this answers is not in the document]
ArielG-NV: there should be a None capability I can use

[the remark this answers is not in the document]
ArielG-NV: I could also make a new capability which is set for all `default` statement's.

csyonghe:
@jkwak-work You can modify the URL of this page from `github.com` to `github.dev`, preserving all other parts of the URL, that will lead you into an online version of vscode that shows the diff of the PR in an editor, so it won't freeze the browser.

source/slang/slang-check-decl.cpp
```diff
@@ -9562,8 +9562,31 @@ namespace Slang
...
+                // we must resolve the target capability as `any_target` to hint to the capability 
+                // system that the following function must be defined on all codegen targets. If
+                // we do not handle `default:`, the codegen will fail when trying to find a specific
+                // codegen target not handled explicitly by a `case` statment.
+                
+                // We must also ensure the `default` case is last so we have priority to hit `case` statments
+                // without complex logic. We must shift all case statments when doing this logic instead of 
+                // swapping with the final element since some capabilities have overlapping targets.
+                // As a result order matters when resolving `case` statments
+                if (CapabilityName(stmt->targetCases[targetCaseIndex]->capability) == CapabilityName::__default_case)
```
csyonghe, line 9578: Can we avoid introducing this __default_case as a capability?
We should keep capability definitions clean and only about actual targets and capabilities.

It seems the goal here is just to tell if a case is a default case. You can simply check if `capability` is `0` or something like that.

csyonghe:
We should add a couple more tests such as using compute-only intrinsics from a vertex shader and verify that we can see diagnostic output.

Also add a test to compile a shader that uses spirv only intrinsics to hlsl and verify diagnostic output.

source/slang/slang-check-decl.cpp
ArielG-NV: That was the original implementation, I will revert to the original implementation.

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
csyonghe: Let's just use Invalid here.

source/slang/hlsl.meta.slang
```diff
@@ -1028,6 +1083,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
     }
 
     [__readNone]
+    [require(cpp_hlsl_glsl_spirv, texture_sm_4_1)]
```
jkwak-work, line 1086: Shouldn't this have "cuda" as well?

source/slang/slang-capabilities.capdef
```diff
@@ -397,3 +451,80 @@ alias DX_6_4 = sm_6_4;
...
+alias atomic_glsl_hlsl_cuda5_int64 = atomic_glsl_int64 | atomic_hlsl_nvapi | _cuda_sm_6_0;
+alias atomic_glsl_hlsl_cuda6_int64 = atomic_glsl_int64 | atomic_hlsl_nvapi | _cuda_sm_6_0;
+alias atomic_glsl_hlsl_cuda9_int64 = atomic_glsl_int64 | atomic_hlsl_nvapi | _cuda_sm_9_0;
+
+alias breakpoint = GL_EXT_debug_printf | hlsl | _cuda_sm_8_0 | cpp;
+
+alias rayobject = raytracing | rayquery;
+alias raytracing_allstages = raytracing_stages + raytracing;
+alias raytracing_anyhit = anyhit + raytracing;
+alias raytracing_int = raytracing_stages_intersection + raytracing;
```
jkwak-work, line 505: Can you rename this to something more readable? like
raytracing_intersection or something?

jkwak-work (commented):
> @jkwak-work You can modify the URL of this page from `github.com` to `github.dev`, preserving all other parts of the URL, that will lead you into an online version of vscode that shows the diff of the PR in an editor, so it won't freeze the browser. 

This was crazy good.
I didn't know about this...
Probably we should mention this on the Contribution guide.

source/slang/slang-capabilities.capdef
```diff
@@ -65,7 +68,22 @@ alias any_target = hlsl | glsl | c | cpp | cuda | spirv;
...
+alias cpp_hlsl_glsl_spirv = cpp | hlsl | glsl | spirv_1_0;
+alias hlsl_cuda = hlsl | cuda;
+alias hlsl_cuda_spirv = hlsl | cuda | spirv;
+alias hlsl_glsl_cuda = hlsl | glsl | cuda;
+alias hlsl_glsl_spirv = hlsl | glsl | spirv;
+alias hlsl_glsl_cuda_spirv = hlsl | glsl | cuda | spirv_1_0;
+alias hlsl_spirv = hlsl | spirv;
+alias glsl_cuda_spirv = glsl | cuda | spirv;
 alias glsl_spirv = glsl | spirv;
+
```
jkwak-work, line 86: What is the rule of the ordering the keywords?
Can we sort them by alphabet order like this?
```
alias cpp_cuda = cpp | cuda;
alias cpp_cuda_glsl_hlsl = cpp | cuda | glsl | hlsl;
alias cpp_cuda_glsl_hlsl_spirv = cpp | cuda | glsl | hlsl | spirv_1_0;
alias cpp_cuda_glsl_spirv = cpp | cuda | glsl | spirv;
alias cpp_cuda_hlsl = cpp | cuda | hlsl;
alias cpp_glsl_hlsl_spirv = cpp | glsl | hlsl | spirv_1_0;
alias cpp_hlsl = cpp | hlsl;
alias cuda_glsl_hlsl = cuda | glsl | hlsl;
alias cuda_glsl_hlsl_spirv = cuda | glsl | hlsl | spirv_1_0;
alias cuda_glsl_spirv = cuda | glsl | spirv;
alias cuda_hlsl = cuda | hlsl;
alias cuda_hlsl_spirv = cuda | hlsl | spirv;
alias glsl_hlsl_spirv = glsl | hlsl | spirv;
alias glsl_spirv = glsl | spirv;
alias hlsl_spirv = hlsl | spirv;
```

source/slang/hlsl.meta.slang
```diff
@@ -2775,73 +2962,90 @@ ${{{{
...
     uint64_t InterlockedExchangeU64(uint byteAddress, uint64_t value)
     {
         let buf = __getEquivalentStructuredBuffer<uint64_t>(this);
         return __atomicExchange(buf[byteAddress / 8], value);
     }
 
     // SM6.6 6 64bit atomics.
-    __specialized_for_target(hlsl)
+    [ForceInline]
     void InterlockedAdd64(uint byteAddress, int64_t valueToAdd, out int64_t outOriginalValue)
```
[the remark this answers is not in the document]
ArielG-NV: capabilities propagate from the children.

a [require(hlsl_glsl_spirv)] is not needed since the `__target_switch` implies the requirement.

source/slang/hlsl.meta.slang
```diff
@@ -11467,6 +12042,7 @@ ${{{{
 
 __magic_type(SubpassInputType)
 __intrinsic_type($(kIROp_SubpassInputType))
+[require(hlsl_glsl_spirv, subpass)]
```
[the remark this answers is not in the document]
ArielG-NV: https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf
gsubpassInput

source/slang/slang-capabilities.capdef
ArielG-NV: The rule was [term1]_[term2]... where [term1] and length have priority over term2 for ordering.

I can sort alphabetically.

source/slang/slang-check-decl.cpp
```diff
@@ -9555,16 +9561,54 @@ namespace Slang
...
+                if (CapabilityName(stmt->targetCases[targetCaseIndex]->capability) == CapabilityName::Invalid)
+                {
+                    if (targetCaseCount - 1 != targetCaseIndex)
+                    {
+                        for (Index i = targetCaseIndex; i < targetCaseCount - 1; i++)
+                            std::swap(stmt->targetCases[i], stmt->targetCases[i + 1]);
+                        continue;
+                    }
+
+                    if (!maybeRequireCapability)
```
csyonghe, line 9602: This is going to create a situation where a target switch case has a capbility set that specifies more than one code gen target (e.g. glsl|hlsl) if we have:
```
[requre(glsl)] [require(hlsl)] [require(cuda)] func()
{
       __target_swtich
       {
       case cuda: ...
       default: ... // the cap for default is now glsl|hlsl
       }
}
```

Our existing logic may be handling this just fine, but I am not sure.

source/slang/slang-capability.cpp
```diff
@@ -988,6 +988,104 @@ void CapabilitySet::canonicalize()
     m_conjunctions.sort();
 }
 
+CapabilitySet CapabilitySet::getTextualTargetsThisIsMissingFromOther(const CapabilitySet& other)
```
csyonghe, line 991: Why `textualTarget`? Shouldn't it handle non textual targets just fine?

source/slang/slang-check-decl.cpp
```diff
@@ -9753,6 +9804,8 @@ namespace Slang
                     diagnoseUndeclaredCapability(funcDecl, Diagnostics::useOfUndeclaredCapability, failedAvailableCapabilityConjunction);
                     funcDecl->inferredCapabilityRequirements = declaredCaps;
                 }
+                else
+                    funcDecl->inferredCapabilityRequirements.simpleJoinWithSetMask(declaredCaps, CapabilityName::stage);
```
csyonghe, line 9808: Can you explain why  this is needed with a concrete example?

source/slang/slang-check-decl.cpp
```diff
@@ -9481,8 +9484,9 @@ namespace Slang
             ensureDecl(visitor, referencedDecl, DeclCheckState::CapabilityChecked);
         }
         
-        if (resultCaps.implies(nodeCaps))
-            return;
+        // Always join capabilities to ensure only the lowest denominator will be apart of the final requirments.
```
csyonghe, line 9487: There may be a subtlety here, but the intention here is to reduce the amount of calls to `join` which can be expensive.
Generally, we will if resultCaps = (compute|fragment) and nodeCaps = (fragment), then resultCaps shouldn't imply nodeCaps, so this early return should never hit?

source/slang/slang-capability.cpp
ArielG-NV: You are correct, my naming would be wrong here (intended behavior is for targets*).

source/slang/slang-check-decl.cpp
ArielG-NV: The existing logic does not compress all require's into 1
I will add this (should be very trivial) -- just a join of every-single require

source/slang/slang-check-decl.cpp
ArielG-NV: the general problem here will be described with a concrete example below: 
1. I assign capability to a function, this resolves as a `declaredCaps`. 
2. 2. We assign a shader stage to `declaredCaps`. 
3. 3. Now when we call this function, since we likely never join `declaredCaps` with `inferredCapability`, future checks of `inferredCapability` will not have data of any shader stage requirement.

concrete example:
```
[ForceInline]
[require(glsl_hlsl_spirv, shadermemorycontrol_compute)]
public void memoryBarrierAtomicCounter()
{
    __target_switch
    {
    case hlsl: __intrinsic_asm "AllMemoryBarrier()";
    case glsl: __intrinsic_asm "memoryBarrierBuffer()";
    case spirv:
    {
        spirv_asm
        {
            OpCapability Shader;
            OpMemoryBarrier Device AcquireRelease|UniformMemory
        };
    }
    }
}
```
If we try to use this function in a vertex shader without the change provided, the code will not error on the Slang side of things since the shader does not propagate `compute` into the "inferredCapabilityRequirements".

Technically speaking, "join" is more correct that joinWithSetMask, but there are issues with doing this on the 
design & performance side of things with the current state of the capabilities system and how we plan to auto-magically manage code based on capabilities.

source/slang/slang-check-decl.cpp
ArielG-NV: the problem is as follows:
lets assume we have the following function body
```
{
Store(...);
... = Sample(...);
}
```
What will happen is, Store sets the inferredCaps to {Fragment, Compute, RayGen, Anyhit, Callable....}.
Then we will get the function Sample, which has the nodeCaps of {Fragment}. 
By logic of Implies, Store does imply Sample, Store has Fragment, it is a super set, this is what `Implies` checks for. 

The problem is that we are not checking for a super set because this means that we will not join the caps to cull for the lowest common denominator. As a result the function's inferred caps will require {Fragment, Compute, RayGen, Anyhit, Callable....} instead of {Fragment}, this is a problem because calling/current functions will allow greater leniency in what can be inside the body. 

For an extended example, we can with the old logic before the change we allow "Fragment only" + "Compute only" functions in a body as long as the first propagated requirement is "Fragment | Compute", this is incorrect behavior.

ArielG-NV:
note: due to regressions, logic had to be amended to not affect all inheritanceDecl since generics would break

ArielG-NV:
NOTE: this is not complete code if it passes the test

csyonghe:
Now that all tests are passing, let's add a compiler option to allow people to opt out of the errors, then we can merge this PR.

source/slang/slang-check-decl.cpp
csyonghe: I think it is an error to say (Fragment | Compute) implies (Fragment)?

csyonghe: I wonder if the right fix is to fix `implies` instead? That may have other cascading issues, but is it the right thing to do?

ArielG-NV: > I think it is an error to say (Fragment | Compute) implies (Fragment)?
This does correctly return false, that is why we needed to change the logic.

implies works as intended, _I think_.

ArielG-NV: This means the solution is we then do:
join if implies is false & join if implies is true?
if this is correct, then we should just omit implies.

source/slang/slang-check-decl.cpp
```diff
@@ -9942,14 +9988,20 @@ namespace Slang
         decl->inferredCapabilityRequirements = getDeclaredCapabilitySet(decl);
     }
 
+    static inline UnownedStringSlice breakAtThisDeclNameToCheckCapability = UnownedStringSlice("");
+
     void SemanticsDeclCapabilityVisitor::visitFunctionDeclBase(FunctionDeclBase* funcDecl)
     {
+#if defined(_DEBUG) && _WIN32 && defined(_MSC_VER)
```
csyonghe, line 9995: Remove

source/slang/slang-check-decl.cpp
```diff
@@ -9974,7 +10026,7 @@ namespace Slang
                 auto stageCaps = CapabilitySet(Profile(entryPointAttr->stage).getCapabilityName());
                 if (declaredCaps.isIncompatibleWith(stageCaps))
                 {
-                    getSink()->diagnose(funcDecl->loc, Diagnostics::stageIsInCompatibleWithCapabilityDefinition, funcDecl, stageCaps, declaredCaps);
+                    getSink()->diagnoseCapabilityErrors(this->getOptionSet(), funcDecl->loc, Diagnostics::stageIsInCompatibleWithCapabilityDefinition, funcDecl, stageCaps, declaredCaps);
```
csyonghe, line 10029: Just make this a global function instead of extending DiagnosticSink.

This class is shared with things outside of the slang compiler so it shouldn't have any slang specific logic.

source/slang/slang-check-decl.cpp
```diff
@@ -9745,13 +9748,14 @@ namespace Slang
             ensureDecl(visitor, referencedDecl, DeclCheckState::CapabilityChecked);
         }
         
-        if (resultCaps.implies(nodeCaps))
```
csyonghe, line 9748: So in your example, after processing Store, the resultCaps is (Fragment|Compute). 

When we process Sample, resultCaps shouldn't be implying Compute? So this implies call should return false.

In that case the two capabilities will be joined to form the new resultCaps that is just (Fragment).

So why do we need to remove the implies call here?

My guess is that it is returning true in this case and that is questionable.

ArielG-NV: I actually we can add back the implies now (I think partially what forcing join solved was something another fix I added closer to the end). CI of course will check to be sure.

csyonghe (approved):

PR #3914, opened 2024-04-09
Allow COM based API to discover and check entrypoints without [shader] attribute.

### LINKED ISSUE #3882: Allow `IModule::findEntryPointByName` to work with entrypoints without `[shader]` attribute.
If the user writes something like:
```
float4 vsMain() { return 0; }
```

And calls `module->findEntryPointByName("vsMain")`,
we will return `nullptr` because there is nothing in the module defined with `shader` attribute.

We should allow this function to proceed to find other functions with the name and just return that.

This will enable our users to get rid of the dependence on `ICompileRequest` API, which has this feature to compile entrypoints not marked as `[shader("...")]`.

Once we have this we can deprecate all the legacy APIs, including `ICompileRequest` interface.

csyonghe:
Our COM based API has provided a `IModule::findEntryPointByName` method that can return
an `IEntryPoint` from an `IModule`.

However, this function only works when the entrypoint is decorated with the `[shader]` attribute.
If not, the function returns nullptr.

This changes adds `IModule::findAndCheckEntryPoint` so that if the function is not decorated
with [shader], we will call `findAndValidateEntryPoint` to find and validate the entrypoint function
on demand.

The work here is just to set up the FrontEndCompileRequest and FrontEndEntryPointRequests
so that the checking logic and be invoked.

Closes #3882.


jkwak-work (approved):
Looks good to me

PR #3924, opened 2024-04-10
Fix typelayout for append/consume structured buffers.

### LINKED ISSUE #3923: Explicit binding for vulkan is not correctly assigned for AppendStructuredBuffer
Hello,

This code does not compile:
https://shader-playground.timjones.io/c077cc3bbce720f5f53e1da2feb7a542

The automatic binding counters are also not correct, for example here:
https://shader-playground.timjones.io/80ba8fb6b7d9c412db0aba00c9f97786

-> Why is inBuffer assigned binding = 12?

Thanks for looking into it.

csyonghe:
The current logic double counts the counter resource.

Fixes #3923.


jkwak-work (approved):
Looks good to me.

PR #3925, opened 2024-04-10
Properly compile `gl_WorkgroupSize`.

### LINKED ISSUE #3910: gl_WorkGroupSize must be a constant
(Thanks for [glsl.meta.slang](https://github.com/shader-slang/slang/blob/master/source/slang/glsl.meta.slang) it's quite helpful to port over GLSL code.)

This shader:
```hlsl
import "glsl";

[shader("compute")] [numthreads(256, 1, 1)] void compute() {
  const uint x = gl_WorkGroupSize.x;
}
```

When compiled like this (v2024.1.6) and passed through `spirv-val` will complain that gl_WorkGroupSize isn't decorated as it should:
```
slangc compute.slang -target spirv -o compute.spv -emit-spirv-directly -fvk-use-entrypoint-name
spirv-val compute.spv 
error: line 7: BuiltIn decoration on target <id> '9[%gl_WorkGroupSize]' must be a constant for WorkgroupSize
  OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
```

csyonghe:
Instead of relying on the WorkgroupSize SPIRV built-in, which has been deprecated and superseded with a dedicated execution mode in spirv 1.6 (which we are not targeting), we simply resolve the workgroup size from the `[numthreads]` decoration and substitute it into the user code.

Fixes #3910.

### DIFF
--- a/source/slang/slang-ir-translate-glsl-global-var.cpp
+++ b/source/slang/slang-ir-translate-glsl-global-var.cpp
@@ -244,7 +244,8 @@ namespace Slang
                     if (auto parentFunc = getParentFunc(use->getUser()))
                     {
                         auto referenceSet = referenceGraph.tryGetValue(parentFunc);
-                        if (!referenceSet) return;
+                        if (!referenceSet) 
+                        	return;
                         if (referenceSet->getCount() == 1)
                         {
                             // If the function that uses the workgroup size is only used by one entry point,
--- a/source/slang/slang-ir-translate-glsl-global-var.cpp
+++ b/source/slang/slang-ir-translate-glsl-global-var.cpp
@@ -18,10 +18,15 @@ namespace Slang
             buildEntryPointReferenceGraph(referencingEntryPoints, module);
 
             List<IRInst*> entryPoints;
+            // Traverse the module to find all entry points.
+            // If we see a `GetWorkGroupSize` instruction, we will materialize it.
+            // 
             for (auto inst : module->getGlobalInsts())
             {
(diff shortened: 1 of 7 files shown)


source/slang/slang-ir-translate-glsl-global-var.cpp
```diff
@@ -223,6 +228,83 @@ namespace Slang
...
+        //
+        void materializeGetWorkGroupSize(IRModule* module, Dictionary<IRInst*, HashSet<IRFunc*>>& referenceGraph, IRInst* workgroupSizeInst)
+        {
+            IRBuilder builder(workgroupSizeInst);
+            traverseUses(workgroupSizeInst, [&](IRUse* use)
+                {
+                    if (auto parentFunc = getParentFunc(use->getUser()))
+                    {
+                        auto referenceSet = referenceGraph.tryGetValue(parentFunc);
+                        if (!referenceSet) return;
```
ArielG-NV, line 247:
```suggestion
                        if (!referenceSet) 
                        	return;
```
To allow code stepping

ArielG-NV (dismissed):
just the 1 comment, code looks good 👍.

ArielG-NV (approved):

tests/spirv/subgroup-size.slang
```diff
@@ -0,0 +1,16 @@
+import "glsl";
```
jkwak-work, line 1: Is this necessary?

Issue #3927, opened 2024-04-10
Investigate why `sp*` API does not produce any result when compiling to binary module.


csyonghe:
When using the legacy sp* api to compile a slang module into a binary module, the spGetContainerCode function did not return any code.

swoods-nv:
Kai, do you have an estimate for this one?

kaizhangNV:
> Kai, do you have an estimate for this one?

I have the fix for this one. The fix change actually fixes more issues than this one.

PR #3936, opened 2024-04-12
Fix the issues when compiling slang to library

kaizhangNV:
Fix issue #3927

This change fixes multiple issues encountered when compiling slang source into library by legacy APIs.

* 'emit-ir' compile flag has to specified after the call to specify the container format to SLANG_CONTAINER_FORMAT_SLANG_MODULE. This is not supposed to happen. So we change how we enable the creation of the container to avoid the matters of the calls order.

* ArtifactContainerUtil::filter could success with output as nullptr, we need to check the output to avoid segment fault.

* In library compile, when serializing the container we should not add the IRModule for the target because there is no code generation for library. Therefore, check the 'SkipCodeGen' option to decide whether we need to serialize the target's IRModule.


csyonghe (dismissed):
Looks good!

csyonghe (approved):

jkwak-work:
I think you could have also fixed another potential problem in "ComponentType::getResultAsFileSystem()".
```
    // Filter the containerArtifact into things that can be written
    ComPtr<IArtifact> writeArtifact;
    SLANG_RETURN_ON_FAIL(ArtifactContainerUtil::filter(artifact, writeArtifact));
    SLANG_RETURN_ON_FAIL(ArtifactContainerUtil::writeContainer(writeArtifact, "", fileSystem));
```

PR #3938, opened 2024-04-12
WIP: Emit header files for CPP and CUDA targets

aroidzap:
It is quite cumbersome to work without headers in languages like C++ or CUDA.

I tried to add an experimental `-emit-header` option for `slangc` and Slang attribute `[HeaderExport]`, to be able to emit only marked structs and function definitions into header file.

This is just a draft, I wanted to get some feedback - what would be best way to approach this. Maybe a new target can be created for header export, but it would be nice to re-use specific target implementations if possible.

---

For the C++ headers i would swap `slang-cpp-prelude.h` with `slang-cpp-types.h`, but for CUDA there is only `slang-cuda-prelude.h` - so as a workaround maybe just mark everything as inline to be able to include it multiple times?


source/slang/slang-emit-c-like.cpp
```diff
@@ -1105,7 +1105,23 @@ String CLikeSourceEmitter::getName(IRInst* inst)
     String name;
     if(!m_mapInstToName.tryGetValue(inst, name))
     {
-        name = generateName(inst);
+        // unmangle names, when emitting header
+        if (isEmmitingHeader()) {
```
csyonghe, line 1109: Our convention is for braces to be in their own lines.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3513,6 +3539,13 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
 
 void CLikeSourceEmitter::emitStruct(IRStructType* structType)
 {
+    // TODO: add similar functionality like markTypeForPyExport for [HeaderExport]
+
+    // When emmiting header, skip if should not be emmited.
+    if (isEmmitingHeader() && !structType->findDecoration<IRHeaderExportDecoration>()) {
```
csyonghe, line 3545: There may be an issue here if an exported struct has a member whose type is not exported.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1105,7 +1105,23 @@ String CLikeSourceEmitter::getName(IRInst* inst)
...
+            {
+                StringBuilder sb;
+                for (auto c : nameHintDecor->getName()) {
+                    if (c == '.')
+                        sb.append('_');
+                    else {
+                        sb.append(c);
+                    }
+                }
+                name = sb.produceString();
```
jkwak-work, line 1120: Wouldn't this conflict with another name that actually uses '_' as a part of its name?
You may want to check how "__extern_cpp" avoids the name mangling.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3450,6 +3466,11 @@ bool CLikeSourceEmitter::isTargetIntrinsic(IRInst* inst)
     return findTargetIntrinsicDefinition(inst, intrinsicDef);
 }
 
+bool CLikeSourceEmitter::isEmmitingHeader() {
```
jkwak-work, line 3469: Did you mean `isEmittingHeader()'?

jkwak-work: I think the name will be better with "shouldXXX()" and probably something like "shouldEmitOnlyHeader()".
Because the compiler option is a global state that isn't gonna change once set.
"isEmitting" sounds like a transitional state where it can be in a emitting state only temporarily.

Also note that the other compiler options are queried with functions whose name starts with "should".
```
        bool shouldEmitSPIRVDirectly()
        {
            return getBoolOption(CompilerOptionName::EmitSpirvDirectly);
        }
        bool shouldUseScalarLayout()
        {
            return getBoolOption(CompilerOptionName::GLSLForceScalarLayout);
        }
```

source/slang/slang-emit-c-like.cpp
aroidzap: But I would like to keep namespaces somehow, if possible.

aroidzap: I changed this to use `appendScrubbedName`, but it still does basically the same - replaces dot with underscore.

aroidzap:
> I think the changes here are fine.
> 
> We do need to expose these as new target, CppHeader and CudaHeader.
> 
> You can keep the isEmittingHeader option in the emitter, just route the header target to the same source Emitter and properly set that option.
> 
> This way users can just do slangc a.slang -o a.h

I added `hpp,h++,hxx` for C++ headers and `cuh` for CUDA headers, to follow the C++ and CUDA sources extensions already defined in Slang.

source/slang/slang-emit-c-like.cpp
csyonghe: This shouldn't be necessary because `generateName(inst)` will already call `appendScrubbedName` for you to generate a valid name for the target. Why do we need a separate branch here?

csyonghe: I suggest using __extenr_cpp in user code and remove this special case logic.

source/slang/slang-emit-c-like.cpp
csyonghe: This is not resolved.

source/slang/slang-ir-header-export.cpp
```diff
@@ -0,0 +1,110 @@
...
+        else if (auto arrayType = as<IRArrayType>(type))
+        {
+            // TODO: not sure about this
+            IRBuilder builder(arrayType->getModule());
+            if (!arrayType->findDecoration<IRHeaderExportDecoration>())
+                builder.addHeaderExportDecoration(arrayType);
+
+            markForHeaderExport_(builder, arrayType->getElementType());
+            return;
+        }
```
jkwak-work, line 55: I like to suggest to write some test cases for this part, then it will become more clear of what needs to happen.

source/slang/slang-ir-inst-defs.h
```diff
@@ -771,6 +771,7 @@ INST(HighLevelDeclDecoration,               highLevelDecl,          1, 0)
     INST(CudaKernelBackwardDerivativeDecoration,         CudaKernelBwdDiffRef,         0, 0)
     INST(AutoPyBindExportInfoDecoration,    PyBindExportFuncInfo,   0, 0)
     INST(PyExportDecoration,    PyExportDecoration,   0, 0)
+    INST(HeaderExportDecoration, HeaderExportDecoration, 0, 0)
```
jkwak-work, line 774: I don't think you need this.
Couldn't you make it working with "__extern_cpp"?

jkwak-work (commented):
You will need to add some tests.

source/slang/slang-emit-c-like.cpp
aroidzap: I'm sorry, I cannot se where the problem is - code in this thread was already updated and I tried to fix it everywhere in my changes. Is there some official `.clang-format` or other file so I can get my changes auto-formatted? Thanks.

source/core/slang-type-text-util.cpp
```diff
@@ -51,11 +51,13 @@ static const TypeTextUtil::CompileTargetInfo s_compileTargetInfos[] =
     { SLANG_SPIRV_ASM,      "spv-asm",                                          "spirv-asm,spirv-assembly", "SPIR-V assembly" },
     { SLANG_C_SOURCE,       "c",                                                "c",                        "C source code" },
     { SLANG_CPP_SOURCE,     "cpp,c++,cxx",                                      "cpp,c++,cxx",              "C++ source code" },
+    { SLANG_CPP_HEADER,     "hpp,h++,hxx",                                      "hpp,h++,hxx",              "C++ source header" },
```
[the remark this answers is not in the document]
aroidzap: I just don't want to confuse C `.h, .inc` vs C++ headers. 
I also think that `.c++` and `.h++` is really not common, but maybe `.cc` and `.hh` can be added.

source/slang/slang-emit-c-like.cpp
aroidzap: `generateName(inst)` generates names like `_S1,  _S2` or `something_0`, `something_1` I wanted to get original names from Slang without any suffixes or changes.

csyonghe:
The PR is showing changes from other recently merged PRs. Make sure you rebase your changes on top of tree, and exclude all the commits that are not part of this change so we can have a clean diff here.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1111,7 +1113,20 @@ String CLikeSourceEmitter::getName(IRInst* inst)
     String name;
     if(!m_mapInstToName.tryGetValue(inst, name))
     {
-        name = generateName(inst);
+        // unmangle names, when emitting header
+        if (shouldEmitOnlyHeader())
```
csyonghe, line 1117: I think we should bottleneck name generating logic through the same branch regardless of `shouldEmitOnlyHeader` setting so our cpp code can actually match the header code. Is there a reason why we need different behavior then just calling `generateName` here?

source/slang/slang-emit-c-like.cpp
```diff
@@ -3526,6 +3546,11 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
 
 void CLikeSourceEmitter::emitStruct(IRStructType* structType)
 {
+    if (shouldEmitOnlyHeader() && !structType->findDecoration<IRExternCppDecoration>())
```
csyonghe, line 3549: Instead of skipping here, we should just make sure we never generate the work to emit the struct from the beginning. See `CLikeSourceEmitter::computeEmitActions` where we are trying to figure out what insts needs to be emitted. We should just make sure the irrelavent structs are not added. Also if you have:
```
struct MyType {}

[HeaderExport]
void myFunc(MyType t);
```

That should result `MyType` to be emitted automatically. By carefully handling `computeEmitActions` we should be able to handle this gracefully.

aroidzap:
I have an issue with the https://github.com/shader-slang/slang/pull/4042 MR and its `lowerBuiltinTypesForKernelEntryPoints`.

Even though I have something like
```
__extern_cpp struct externCppStruct 
{
    uint2 xy;
};

[CudaKernel]
__extern_cpp void externCppKernel(externCppStruct s) { }
```

The generated code is now
```cuda
struct _VectorStorage_uint2_0
{
    FixedArray<uint, 2>  data_0;
};

struct U_StructStorage_externCppStruct_0
{
    _VectorStorage_uint2_0 xy_0;
};

__global__ void externCppKernel(U_StructStorage_externCppStruct_0 s_0)
{
    return;
}
```

previously it was simply
```cuda
struct externCppStruct
{
    uint2  xy_0;
};

__global__ void externCppKernel(externCppStruct s_0)
{
    return;
}
```

Is this expected behavior, even when the struct is marked as `__extern_cpp`?

source/slang/slang-emit-c-like.cpp
aroidzap: Main downside of using `__extern_cpp` is that it does not support namespaces. Unfortunately I think I would really need to drop this, otherwise there is a name mismatch between header and source code - for some reason it worked, but it is not a clear solution. Any ideas how to have `__extern_cpp` that supports namespaces? Because for example, if I want to have CPU/CUDA code (that switches at runtime if cuda is available), I would have name clashes - and namespaces could solve this for me (if they are different for CPU and CUDA sources).

csyonghe: How about this:
we allow `__extern_cpp` on `namespace`s as well. And `__extern_cpp` on a decl means that the decl gets a mangled name that is in the form of: `parentNamespaceNameThatHasExternCppModifier_DeclName`.

So:
```
__extern_cpp namespace A
{
         namespace B {
              __extern_cpp struct C;
        }
}
Will have the mangled name `A_C`.
```

csyonghe:
@saipraveenb25 I think the fix here is that if we are lowering a `struct` type that has `__extern_cpp` modifier, we want to transfer the modifier to the lowered type.

csyonghe:
@aroidzap If instead we generate:
```
struct externCppStruct
{
    FixedArray<uint,2> xy;
};

__global__ void externCppKernel(externCppStruct s_0)
{
    return;
}
```
will that work for you?

csyonghe:
The reason here is that we have to lower all types that appear in the parameter list of a `CudaKernel` to not use the builtin vector types because CUDA builtin vector types has a wider alignment than all other targets, causing mismatch in data layout with C++/HLSL/SPIRV. Our solution is to always lower them into raw arrays.

aroidzap:
Yes itn would help, thanks. 

Regarding to the automatic transitive application of __extern_cpp to function arguments and struct members, do you think i should drop it and just put __extern_cpp everywhere in user code? I would need to apply this also in the source (not only header) but this would require taking back the slangc option or slang attribute. 

And regarding the namespaces missing in the __extern_cpp, what do you think is the best idea to approach that?

Thank you very much for your help!

csyonghe:
I think transitive application to struct members makes sense. I am not sure why we need to apply to argument names, they shouldn't matter, right?

For name mangling, you can check the mangleName function and see how __extern_cpp modifier is handled there.

aroidzap:
I thought that it would be nice to have `__extern_cpp` automatically propagated to it's argument types, so the user don't have to care at all. But having this just for the structs could be nice compromise. 

What about this namespace support for __extern_cpp? Could it be achieved or I should somehow add prefix by some macro (so in can have different prefix for cuda, cpp and possibly metal code).

source/slang/slang-emit-c-like.cpp
aroidzap: Yes this would be nice, thanks.

Unfortunately it probably won't fully solve my problem with linking together CPU and CUDA code.

csyonghe:
That looks fine. I don't think there is an existing PR that changes the externCpp behavior for struct fields.

source/core/slang-type-text-util.cpp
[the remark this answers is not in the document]
csyonghe: I am fine with treating .h as cpp header because we won't be supporting c as target anytime soon. Treating cpp as the default is likely the desired behavior anyways.

source/slang/slang-emit-c-like.cpp
```diff
@@ -1150,7 +1153,20 @@ String CLikeSourceEmitter::getName(IRInst* inst)
     String name;
     if(!m_mapInstToName.tryGetValue(inst, name))
     {
-        name = generateName(inst);
+        // unmangle names, when emitting header
+        if (shouldEmitOnlyHeader())
+        {
+            if (auto nameHintDecor = inst->findDecoration<IRNameHintDecoration>())
```
csyonghe, line 1159: This feels hacky. We should use the same name emit logic for header and cpp files. For things that needs to be emitted without name mangling, they should be marked as extern_cpp.

source/slang/slang-emit-metal.cpp
```diff
@@ -196,6 +196,12 @@ void MetalSourceEmitter::emitParameterGroupImpl(IRGlobalParam* varDecl, IRUnifor
 
 void MetalSourceEmitter::emitEntryPointAttributesImpl(IRFunc* irFunc, IREntryPointDecoration* entryPointDecor)
 {
+    if (shouldEmitOnlyHeader())
```
csyonghe, line 199: We shouldn't use hacks like this. If entry points need to be omitted from header emit, they should be removed from the ir before emitting. We should use a clean up pass before emit to remove entry point functions and all function bodies, then add export decoration to all remaining functions or types so they are not dead code eliminated or ignored by emit.

source/slang/slang-ir-header-export.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "slang-ir-header-export.h"
+
+#include "slang-ir.h"
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+
+static void addExternCppDecorationTransitively_(IRBuilder & builder, IRInst * inst)
+{
```
csyonghe, line 11: We should use a work list instead of recursive calls to prevent stack overflow.

source/slang/slang-ir-header-export.cpp
```diff
@@ -0,0 +1,112 @@
...
+                structBuilder.addExternCppDecoration(structType, UnownedStringSlice(""));
+            }
+
+            for (auto field : structType->getFields())
+            {
+                addExternCppDecorationTransitively_(structBuilder, field->getFieldType());
+            }
+            return;
+        }
+        else if (auto arrayType = as<IRArrayType>(type))
```
csyonghe, line 48: This should be IRArrayTypeBase.

Should also handle IRPtrTypeBase, IRHLSLStructuredBufferTyoeBase etc.

source/slang/slang-ir-header-export.cpp
```diff
@@ -0,0 +1,112 @@
...
+    DiagnosticSink* diagnosticSink;
+
+    void processInst(IRInst* inst, IRExternCppDecoration* /*externCppDecoration*/)
+    {
+        IRBuilder builder(module);
+        addExternCppDecorationTransitively_(builder, inst);
+        // TODO: required? builder.addNameHintDecoration(inst, externCppDecoration->getFunctionName());
+        // TODO: required? builder.addExternCppDecoration(inst, externCppDecoration->getFunctionName());
+        // TODO: required? builder.addPublicDecoration(inst);
+        // TODO: required? builder.addKeepAliveDecoration(inst);
```
csyonghe, line 73: Need keep alive and HLSL export.

source/slang/slang-ir-header-export.cpp
```diff
@@ -0,0 +1,112 @@
...
+            default:
+                break;
+            }
+        }
+
+        for (auto candidate : candidates)
+        {
+            processInst(candidate.inst, candidate.exportDecoration);
+        }
+    }
```
csyonghe, line 101: Should delete all function modules after marking everything.

aroidzap:
Hello, I would like to push this forward, I would update to the latest master and try to add some tests. 

Are there any updates about this issue - how to handle header files? Is it still feasible to be added and maintained? Or should I 
rather stick to some other approach? I would still like to use Slang as compile-time only tool, without any runtime dependencies though.

If this is feasible, should I add just C++ and CUDA headers and add Metal headers etc. in separate MR? Or do all targets at once?

csyonghe:
I think the approach of this PR is fine. We just need to address the comments.

jhelferty-nv:
@aroidzap Ping. Are you anticipating getting back to this?

aroidzap:
I had another priorities unfortunately, but yes.

source/core/slang-type-text-util.cpp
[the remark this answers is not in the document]
aroidzap: I'll make it only hpp.

aroidzap:
I tried to remove as much unnecessary code as possible and added two tests. Now I'm looking into how to make it a less "hacky" solution. But I'm not sure how to proceed exactly. Any more specific hints would be welcome.

mkeshavaNV:
@aroidzap Ping. Are you anticipating getting back to this? If you don't plan to work on this soon, can you kindly mark it as draft?

aroidzap:
Hi, I'll try to resolve it soon, I'm just not sure exactly how yet. Any help/hints would be welcome.

csyonghe:
What I was suggesting with my comment is that we should add a pass before emit, that loops through all global functions in the IR, and just delete its body (all IRBlock typed children of the IRFunc). We run this pass when in header output mode.

aroidzap:
Hello, @csyonghe please is there some part of the code, where can I see removal of IRFunc? I was not able to make it work.

aroidzap:
Hi, I tried to resolve all discussions, I hope it is now merge-able. If you need some further tests or something, let me know. Thanks!

source/slang/slang-emit-c-like.cpp
```diff
@@ -3937,6 +3943,8 @@ void CLikeSourceEmitter::emitFuncDecl(IRFunc* func, const String& name)
     auto funcType = func->getDataType();
     auto resultType = func->getResultType();
 
+    emitFunctionPreambleImpl(func);
+
```
aroidzap, line 3947: Not sure about this change, but it seemed to be missing, as otherwise `__global__` for CUDA kernels was missing in the declaration.

source/slang/slang-emit-c-like.cpp
```diff
@@ -5294,6 +5302,29 @@ void CLikeSourceEmitter::computeEmitActions(IRModule* module, List<EmitAction>&
...
+                    if (as<IRBlock>(child))
+                    {
+                        child->removeAndDeallocate();
+                    }
+                }
+                if (inst->findDecoration<IRCudaKernelDecoration>() ||
+                    inst->findDecoration<IREntryPointDecoration>())
+                {
+                    ensureGlobalInst(&ctx, inst, EmitAction::Level::Definition);
+                }
```
aroidzap: I'll revert this, it should not be necessary.

include/slang.h
```diff
@@ -619,12 +619,14 @@ typedef uint32_t SlangSizeT;
         SLANG_DXIL_ASM,
         SLANG_C_SOURCE,              ///< The C language
         SLANG_CPP_SOURCE,            ///< C++ code for shader kernels.
+        SLANG_CPP_HEADER,            ///< C++ header for shader kernels.
```
jkwak-work: You cannot add a new enum value in the middle when you modify `slang.h`, because it will break the backward compatibility.
Please move to the end of the enum values.

source/slang/slang-emit-c-like.cpp
jkwak-work: This seems sensible to me.

jkwak-work (commented):
Looks good to me.

I have minor comments.

CI seems to be failing but I am not sure why,
The error message I saw are:
```
[810/1369] Building CXX object external\slang-rhi\CMakeFiles\slang-rhi.dir\Debug\src\cuda\cuda-shader-table.cpp.obj
[811/1369] Generating slang-core-module-generated.h, ../slang-glsl-module/slang-glsl-module-generated.h
FAILED: source/slang-core-module/slang-core-module-generated.h source/slang-glsl-module/slang-glsl-module-generated.h C:/actions-runner/_work/slang/slang/build/source/slang-core-module/slang-core-module-generated.h C:/actions-runner/_work/slang/slang/build/source/slang-glsl-module/slang-glsl-module-generated.h 
C:\Windows\system32\cmd.exe /C "cd /D C:\actions-runner\_work\slang\slang\build\source\slang-core-module && C:\actions-runner\_work\slang\slang\build\generators\Debug\bin\slang-bootstrap.exe -archive-type riff-lz4 -save-core-module-bin-source C:/actions-runner/_work/slang/slang/build/source/slang-core-module/slang-core-module-generated.h -save-glsl-module-bin-source C:/actions-runner/_work/slang/slang/build/source/slang-core-module/../slang-glsl-module/slang-glsl-module-generated.h"
[812/1369] Building CXX object external\slang-rhi\CMakeFiles\slang-rhi.dir\Debug\src\cuda\cuda-texture.cpp.obj
[813/1369] Building CXX object external\slang-rhi\CMakeFiles\slang-rhi.dir\Debug\src\cuda\cuda-utils.cpp.obj
[814/1369] Building CXX object external\slang-rhi\CMakeFiles\slang-rhi.dir\Debug\src\cuda\optix-api.cpp.obj
[815/1369] Building CXX object external\slang-rhi\CMakeFiles\slang-rhi.dir\Debug\src\cuda\cuda-surface.cpp.obj
```
This PR doesn't seem to be related to the error.
Maybe you should rebase from ToT to resolve it?

jhelferty-nv:
It looks like this is still causing the same failure across a number of debug builds (and a couple of release builds):
```
[525/1252] Generating slang-core-module-generated.h, ../slang-glsl-module/slang-glsl-module-generated.h
FAILED: [code=134] source/slang-core-module/slang-core-module-generated.h source/slang-glsl-module/slang-glsl-module-generated.h /home/runner/work/slang/slang/build/source/slang-core-module/slang-core-module-generated.h /home/runner/work/slang/slang/build/source/slang-glsl-module/slang-glsl-module-generated.h 
cd /home/runner/work/slang/slang/build/source/slang-core-module && /home/runner/work/slang/slang/build/generators/Debug/bin/slang-bootstrap -archive-type riff-lz4 -save-core-module-bin-source /home/runner/work/slang/slang/build/source/slang-core-module/slang-core-module-generated.h -save-glsl-module-bin-source /home/runner/work/slang/slang/build/source/slang-core-module/../slang-glsl-module/slang-glsl-module-generated.h
terminate called after throwing an instance of 'Slang::InternalError'
Aborted (core dumped)
```

If you're building for release, try building for debug to see if you can reproduce the error (or vice versa)?

aroidzap:
I think it should be ok now.

aroidzap:
Thanks for running the pipeline, I'll look into the failing tests.

aroidzap:
I fixed my own test, but I'm not exactly sure why others are failing. I'll wrap more header related generation code of KernelContext in `if (m_target == CodeGenTarget::CPPHeader)` - that might help.

jhelferty-nv:
@jkwak-work Looks like CI is passing now. Any remaining concerns?

aroidzap:
No worries, thanks for the review and approval! :)

tests/headers/generate-hpp-header.slang
```diff
@@ -0,0 +1,41 @@
...
+
+//CHECK: void kernel(Vector<uint32_t, 3>  {{[[:alnum:]_]+}}, Struct * {{[[:alnum:]_]+}}, KernelContext * {{[[:alnum:]_]+}});
+export __extern_cpp void kernel(uint3 globalIdx, Struct s)
+{
+    sharedMemory[globalIdx.x] = s.vec[globalIdx.x] + s.x;
+    s.result[0] += sharedMemory[globalIdx.x];
+}
+
+// needs dummy entrypoint as no output is generated to stdout
+// otherwise it only works when slangc -o output.hpp is specified
```
kaizhangNV, line 39: I found this is not the case.
If we don't specify the `__extern_cpp`, nothing will be generated, even I use `slangc -o output.hpp`

PR #3941, opened 2024-04-12
Fix IR lowering bug of do-while loops.

### LINKED ISSUE #3930: Compiler crash
Not quite sure which part slangc (v2024.1.7) doesn't like exactly, but it only happens if the `do while` is present in this minimal repro:
```hlsl
struct Constants {
  uint *buffer;
}
[[vk::push_constant]] Constants push_constants;

[shader("compute")] [numthreads(1, 1, 1)] void compute(void) {
  // The following assignment can be commented out, what appears to matter is
  // that index can not be statically determined by the compiler (so it could be
  // left uninitialized for the sake of the example).
  uint index = push_constants.buffer[0];

  // There must be at least two iterations.
  [[unroll]] for (uint i = 0; i < 2; i++) {
    // Commenting this line out would work.
    // The barrier is necessary in the real code (groupshared sync).
    GroupMemoryBarrierWithWaveSync();

    if (index > 0 && i < push_constants.buffer[index - 1]) {
      // This would work fine (but the condition is retested):
      // for (; index > 0 && i < push_constants.buffer[index - 1]; index--);
      do {
        // Commenting this line out would work.
        index--;
      } while (index > 0 && i < push_constants.buffer[index - 1]);

      // More stuff here, hence the do while.
    }
  }
}
```

```
slangc compute.slang -target spirv -o compute.spv -emit-spirv-directly -force-glsl-scalar-layout -fvk-use-entrypoint-name
Segmentation fault (core dumped)
```

csyonghe:
The lowering logic for do-while loops is making an incorrect assumption that the current insert location is still the same block after lowering an expression. However this is not true with the recent short circuiting operator work. Lowering exprs can lead to new block being created, and the insert loc can be changed to a different loc.

This change fixes the logic to not make this assumption, and adds an IR validation logic to guard against this type of IR error.

This fixes a reported crash.

Closes #3930.

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -5890,6 +5890,12 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
         {
             auto irCondition = getSimpleVal(context,
                 lowerRValueExpr(context, condExpr));
+
+            // One thing to be careful here is that lowering irCondition
+            // may create additional blocks due to short circuiting, so
+            // the block we are current inserting into is not necessarily
+            // the same as `testLabel`.
+            // 
             auto invCondition = builder->emitNot(irCondition->getDataType(), irCondition);
 
             // Now we want to `break` if the loop condition is false,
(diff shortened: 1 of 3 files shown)


ArielG-NV (approved):
Everything looks good.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -5890,6 +5890,12 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
         {
             auto irCondition = getSimpleVal(context,
                 lowerRValueExpr(context, condExpr));
+
+            // One thing to be careful here is that lowering irCondition
+            // may create additional blocks due to short circuiting, so
+            // the block we are current inserting into is not necessarily
+            // the same as `testLabel`.
+            // 
             auto invCondition = builder->emitNot(irCondition->getDataType(), irCondition);
```
jkwak-work, line 5899: Is there a reason why we use an inverted value?

csyonghe: I don't know for sure. I remember this might be to work around an issue in autodiff passes.

PR #3942, opened 2024-04-12
Fix micro expansion issue for __LINE__.

### LINKED ISSUE #2134: Macro Expansion Bug
image
image

```MESSY_PAD(3)``` is expanded to ```uint3 padVar11``` but it should be ```uint3 padVar14```.

### LINKED ISSUE #3940: Support __LINE__ macro for the caller of a macro
[Posted on behalf of a user without GH account]

We are using the `__LINE__` macro in combination with GPU printing to implement an assert in shader code. The assert roughly looks like this:

```
#define STR2(X) # X
#define STR(X) STR2(X)
#define CARB_ASSERT(X) \                     //<-- let's say this is line 100 in our file
if (!(X)) { \
        println_assert("Assertion triggered in " __FILE__ ":" STR(__LINE__) ", expression : " #X); \
}
```

this assert is used just as expected

```
void main(dispatchIdx...)
{
    float3 radiance = compute();
    CARB_ASSERT(!isNanOrInf(radiance)); ///< will show line 100 from different file, not this line
}
```

I browsed tests\preprocessor\special-macro-complex.slang but didn't find a matching test for it. I did find this comment which shows the usage of __LINE__ in a macro but probably expects a different result:
```
// LINE number will be for the macro invocation, not the expansion
```

csyonghe:
Fixes #3940.
Closes #2134.

### DIFF
--- a/source/slang/slang-preprocessor.cpp
+++ b/source/slang/slang-preprocessor.cpp
@@ -857,6 +857,11 @@ struct ExpansionInputStream : InputStream
 
     TokenType peekRawTokenType() { return peekRawToken().type; }
 
+    void setInitiatingMacroSourceLoc(SourceLoc loc)
+    {
+        m_initiatingMacroInvocationLoc = loc;
+        m_isInExpansion = true;
+    }
 private:
         /// The base stream that macro expansion is being applied to
     InputStream*   m_base = nullptr;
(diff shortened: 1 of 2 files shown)


jkwak-work (approved):
Looks good to me.

PR #3954, opened 2024-04-15
[GFX] Fix d3d12 buffer view creation logic for StructuredBuffers.

### LINKED ISSUE #3952: Allow GFX to automatically create D3D SRV based on the element type of a StructuredBuffer.

csyonghe:
Fixes #3952.

When binding a buffer resource to a StructuredBuffer parameter in D3D12, we need to specify an element stride. This stride can be derived from Slang's reflection info.

This change changes the logic in resource buffer view creation and binding, so that when binding a buffer to a buffer entry, we will create a specialized descriptor with the stride info, and cache the descriptor on the buffer view object.

### DIFF
--- a/tools/gfx/d3d12/d3d12-shader-object-layout.h
+++ b/tools/gfx/d3d12/d3d12-shader-object-layout.h
@@ -66,6 +66,9 @@ public:
         /// as a sub-object.
         uint32_t subObjectIndex;
 
+        /// The stride of a structured buffer.
+        uint32_t bufferElementStride;
+
         bool isRootParameter;
 
         /// Is this binding range represent a specialization point, such as an existential value, or a `ParameterBlock<IFoo>`.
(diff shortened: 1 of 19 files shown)


tools/gfx/d3d12/d3d12-shader-object-layout.h
```diff
@@ -66,6 +66,9 @@ class ShaderObjectLayoutImpl : public ShaderObjectLayoutBase
         /// as a sub-object.
         uint32_t subObjectIndex;
 
+        /// The stride of a structured buffer.
+        uint32_t bufferElementStride;
```
jkwak-work, line 70: This variable may be used with an uninitialized value.

jkwak-work (dismissed):

tools/gfx/d3d12/d3d12-shader-object-layout.h
csyonghe: We actually default initialize the whole struct whenever it is initialized, so this is fine.

jkwak-work (approved):

jkwak-work (approved):

Issue #3673, opened 2024-03-05
Support `if (let f = ...)` syntax


csyonghe:
We already have `Optional<T>` type and an `as` operator that produces an `Optional<T>`. It will be nice to have a `if (let` syntax so people can write:
```
if (let a = expr as Type) {...}
```
For such expressions, we should lower it into:
```
{
     let a = expr as Type;
     if (a.hasValue)
     {
     }
}
```

PR #3958, opened 2024-04-16
Implement if(let ...) syntax (#3673)

kaizhangNV:
Implement the parser to parse the if (let ....) statement.

Add "struct IfLetStmt: IfStmt" to define the
if "(let xxx = xxxxx)" syntax.

In parsing the predicate expression, we check if the identifier is "let", then parse the "let" expression and return it to predicate.

In parseIfStatement(), we will check if the predicate expression is a 'let' expression. If so, we will create an IfLetStmt.


source/slang/slang-ast-stmt.h
```diff
@@ -79,6 +79,13 @@ class IfStmt : public Stmt
     Stmt* negativeStatement = nullptr;
 };
 
+class IfLetStmt: public IfStmt
```
csyonghe, line 82: I am not sure if we want to inherit from IfStmt as they might be different enough.

source/slang/slang-parser.cpp
```diff
@@ -5581,11 +5589,23 @@ namespace Slang
 
     IfStmt* Parser::parseIfStatement()
```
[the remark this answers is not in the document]
kaizhangNV: The reason I made the check inside `parseIfStatement` is because if there are couple of left parentheses, peeking next token will not be enough.

source/slang/slang-parser.cpp
```diff
@@ -3882,6 +3882,14 @@ namespace Slang
         return decl;
     }
 
+    static NodeBase* parseIfLetDecl(
```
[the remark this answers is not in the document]
kaizhangNV: The reason I create a new function is because in `parseLetDecl`, there is a check for semicolon
```
    static void parseModernVarDeclCommon(
        Parser*         parser,
        VarDecl* decl)
    {
        parseModernVarDeclBaseCommon(parser, decl);
        expect(parser, TokenType::Semicolon);  -> here
    }

```
but there won't be `;` in the `if(let ...)` case.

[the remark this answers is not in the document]
csyonghe: Let's not make that an expr because it shouldn't be treated as one. Just write all the parsing logic in the parseIfLetStmt function.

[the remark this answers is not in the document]
kaizhangNV: Sounds like we have to create a `parseIfLetStmt` function, so we do want to peek next couple tokens until we find a `let` keyword?

source/slang/slang-ast-stmt.h
csyonghe: Right, in this case we are desiguring a single IfLet into a DeclStmt + IfStmt. There is no inheritance relationship between the two.

source/slang/slang-parser.cpp
```diff
@@ -5579,20 +5587,75 @@ namespace Slang
         return varDeclrStatement;
     }
 
+    static void desugarIfLetStatement(IfLetStmt* ifletStatement, Parser* parser)
```
csyonghe, line 5590: Here we should just create and return a new SeqStmt consisting of a DeclStmt and a IfStmt.

After lowering, there should not be any IfLet left.

source/slang/slang-parser.cpp
```diff
@@ -7469,6 +7568,10 @@ namespace Slang
                 }
                 return newExpr;
             }
+            else if (identifierTokenContent == toSlice("let"))
```
csyonghe, line 7571: I don't think we need to make let an expr, since that could introduce ambiguities in places we don't want.

source/slang/slang-parser.cpp
```diff
@@ -7437,6 +7500,42 @@ namespace Slang
         return asmExpr;
     }
 
+    static Expr* parseIfLetExpr(Parser* parser)
```
csyonghe, line 7503: Just fold all these in parseIfLetStmt.

source/slang/slang-ast-stmt.h
kaizhangNV: done

source/slang/slang-parser.cpp
kaizhangNV: Done

source/slang/slang-parser.cpp
kaizhangNV: done

source/slang/slang-parser.cpp
kaizhangNV: Done

source/slang/slang-parser.cpp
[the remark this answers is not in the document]
kaizhangNV: Done

source/slang/slang-parser.cpp
```diff
@@ -3882,6 +3883,14 @@ namespace Slang
         return decl;
     }
 
+    static NodeBase* parseIfLetDecl(
+        Parser* parser, void* /*userData*/)
```
csyonghe, line 3887: This is not needed, you can just parse the fields directly into IfLetStmt.

source/slang/slang-parser.cpp
```diff
@@ -5579,6 +5597,144 @@ namespace Slang
...
+        infixExpr->arguments.add(boolLiteralExpr);
+        return infixExpr;
+    }
+
+    static VarDecl* parseIfLetExpr(Parser* parser)
+    {
+        Decl* parsedDecl = nullptr;
+        auto identifierToken = peekToken(parser);
+        auto name = identifierToken.getName();
+        auto syntaxDecl = tryLookUpSyntaxDecl(parser, name);
```
csyonghe, line 5632: Don't need any of these.

Just:

ReadToken(TokenType::LParent);
ReadToken("let");
ifLetStmt->varName = readToken(TokenType::Identifier);
ReadToken(TokenType::OpAssign);
ifLetStmt->varExpr = parseExpr();
ReadToken(TokenType::RParent);
ifLetStmt->trueStmt = parseStatement();
if (advanceIf("else"))
    ifLetStmt->falseStmt = parseStatement();

source/slang/slang-parser.cpp
kaizhangNV: done

source/slang/slang-parser.cpp
kaizhangNV: done.

jkwak-work:
Can you mention the issue number on your commit message?
In this case it seems to be `#3673 `

kaizhangNV:
> Can you mention the issue number on your commit message? In this case it seems to be `#3673 `

done

source/slang/slang-parser.cpp
```diff
@@ -5579,6 +5589,119 @@ namespace Slang
...
+        FillPosition(tmpVarDeclStmt);
+        tmpVarDeclStmt->decl = tempVarDecl;
+        newBody->stmts.add(tmpVarDeclStmt);
+
+        // construct 'if (tempVarDecl.hasValue == true)'
+        VarExpr* tempVarExpr = astBuilder->create<VarExpr>();
+        tempVarExpr->scope = currentScope;
+        FillPosition(tempVarExpr);
+        tempVarExpr->name = tempVarDecl->getName();
+        ifStatement->predicate = constructIfLetPredicate(this, tempVarExpr);
```
jkwak-work, line 5654: We are not strictly mimicking C++ syntax, but in C++, this is possible,
```
#include <cstdio>
int main()
{
        if (int a = 0; a > 0)
        {
                printf("a\n");
        }
        else if (a == 0) // <== available on the else clause.
        {
                printf("b\n");
        }
        return 0;
}
```
It seems like your implementation would work for this scoping.
Can we have some test cases for it?

jkwak-work (commented):
We need some test cases.

source/slang/slang-parser.cpp
kaizhangNV: this implementation actually doesn't work in else scope. The variable only works in `if` branch.
line 5692 guarantees that.

But this is still a design decision, the variable will be `none` in else branch.

kaizhangNV: The tests cases will be uploaded.

source/slang/slang-ast-stmt.h
```diff
@@ -79,6 +79,7 @@ class IfStmt : public Stmt
     Stmt* negativeStatement = nullptr;
 };
 
+
```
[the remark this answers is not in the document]
kaizhangNV: removed.

source/slang/slang-parser.cpp
```diff
@@ -5579,6 +5589,119 @@ namespace Slang
...
+        auto opExpr = parser->astBuilder->create<VarExpr>();
+        opExpr->name = getName(parser, "==");
+        opExpr->scope = parser->currentScope;
+        parser->FillPosition(opExpr);
+
+        // create a 'true' literal
+        BoolLiteralExpr* boolLiteralExpr = parser->astBuilder->create<BoolLiteralExpr>();
+        boolLiteralExpr->value = true;
+
+        // create a "var.hasValue == true" expression
```
csyonghe, line 5610: Don't need == true,
Just if(x.hasValue)

source/slang/slang-parser.cpp
```diff
@@ -5579,6 +5589,119 @@ namespace Slang
...
+
+        // parse 'let var = X as Y'
+        ReadToken("let");
+        auto identifierToken = ReadToken(TokenType::Identifier);
+        ReadToken(TokenType::OpAssign);
+        auto initExpr = ParseInitExpr();
+
+        // insert 'let tempVarDecl = X as Y;'
+        auto tempVarDecl = astBuilder->create<LetDecl>();
+        tempVarDecl->nameAndLoc = NameLoc(getName(this, "OptVar"), identifierToken.loc);
```
csyonghe, line 5640: Maybe "$optVar" to avoid conflicting with names in user code.

source/slang/slang-parser.cpp
csyonghe: The if following else will be another statement handled by the parse statement so that should be fine.

csyonghe (commented):
Need a test case.

csyonghe:
Please also update the user guide documentation under the convenience language features section after as-is operator.

Can use another test that does something like

if (let a = x as Foo)
   ...
else if (let a = x as Bar)
   ...

source/slang/slang-parser.cpp
kaizhangNV: done.

source/slang/slang-parser.cpp
kaizhangNV: done

csyonghe:
Also make sure to test intellisense locally.

kaizhangNV:
> Also make sure to test intellisense locally.

intellisense is working as expected on Linux vscode.

Screenshot from 2024-04-17 16-49-24

csyonghe:
Looking good, thanks for testing!

csyonghe (approved):

csyonghe:
Implementation looks very clean here. Great job!

docs/user-guide/06-interfaces-generics.md
```diff
@@ -677,6 +677,54 @@ T compute<T>(T a1, T a2)
...
+}
+
+struct MyImpl3 : IFoo
+{
+    void foo() { printf("MyImpl3");}
+}
+
+void test(IFoo foo)
+{
+    if (t == foo as MyImpl1)
```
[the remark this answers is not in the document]
kaizhangNV: Done adding the comments.

docs/user-guide/06-interfaces-generics.md
```diff
@@ -677,6 +677,54 @@ T compute<T>(T a1, T a2)
 // compute(3, 1) == 2
 ```
 
+`as` operator can also be used in the `if` predicate to test if an object can be downcasted to a specific type, once the downcast test is successful,
```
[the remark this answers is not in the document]
kaizhangNV: Done adding more test cases for just casting.

kaizhangNV:
> Implementation looks very clean here. Great job!

Thanks. I updated the user-guide, and unit tests. Also move the unit tests to a new folder.

docs/user-guide/06-interfaces-generics.md
```diff
@@ -677,6 +677,64 @@ T compute<T>(T a1, T a2)
...
+    // {
+    //      Optional<MyImpl1> $OptVar = foo as MyImpl1;
+    //      if ($OptVar.hasValue)
+    //      {
+    //          MyImpl1 t = $OptVar.value;
+    //          t.foo();
+    //      }
+    //      else if ...
+    // }
+    if (let t == foo as MyImpl1)
```
csyonghe, line 715: Should be let t = foo as MyImpl1

kaizhangNV: ah, typo.

docs/user-guide/06-interfaces-generics.md
```diff
@@ -677,6 +677,64 @@ T compute<T>(T a1, T a2)
...
+    // {
+    //      Optional<MyImpl1> $OptVar = foo as MyImpl1;
+    //      if ($OptVar.hasValue)
+    //      {
+    //          MyImpl1 t = $OptVar.value;
+    //          t.foo();
+    //      }
+    //      else if ...
+    // }
+    if (let t = foo as MyImpl1)
```
csyonghe, line 715: Can we add a comment at the end of this line 
// `t` has type MyImpl1

csyonghe (approved):

docs/user-guide/06-interfaces-generics.md
kaizhangNV: sure, added

csyonghe (approved):

Issue #3959, opened 2024-04-16
Error when specifying explicit underlying type on unscoped enums


pborsutzki:
Hi,

the following code produces an error in Slang 2024.1.7, when compiling it via `slangc.exe -entry computeMain -stage compute -target spirv-asm -emit-spirv-directly`

```c++
[UnscopedEnum]
enum Foo : uint32_t {
    Bar = 42
};

[shader("compute")]
[numthreads(1, 1, 1)]
void computeMain()
{}
```

The reported error is:
```
shader.slang(6): fatal error 39999: cyclic reference 'Foo'.
enum Foo : uint32_t {
     ^~~
```

The error is gone, when removing the explicit type specification or the `[UnscopedEnum]` attribute.

Thanks for having a look!

ArielG-NV:
to note what seems to be the issue for future debugging of issue:

Assume enum called `ourEnum`.

moduleScope --> _lookUpDirectAndTransparentMembers --> visitEnum [`ourEnum`] --> ensureDecl [uint32_t] ReadyForLookupResolve --> declToExclude is empty when resolving (likely the issue, should exclude `ourEnum` by this point?)  --> _lookUpInScopes [uint32_t is moduleScope, so it is in same scope as 'ourEnum'?] --> _lookUpDirectAndTransparentMembers --> visitEnum ['ourEnum'], causes failure

csyonghe:
Fixed in #3962

PR #3962, opened 2024-04-16
Fix for unscoped enums circular reference causing an error, #3959

ArielG-NV:
If we have already started checking this transparent member, there is no need to send another check. We just need to ensure a transparent member is ready for reference, which is already being handled if checking started. Transparent members should only be a circular dependency if a direct member which is handled in different code.


source/slang/slang-lookup.cpp
```diff
@@ -224,14 +224,18 @@ static void _lookUpDirectAndTransparentMembers(
...
 
-    // TODO(tfoley): should we look up in the transparent decls
-    // if we already has a hit in the current container?
     for(auto transparentInfo : containerDecl->getTransparentMembers())
     {
         // The reference to the transparent member should use the same
         // path as we used in referring to its parent.
         DeclRef<Decl> transparentMemberDeclRef = astBuilder->getMemberDeclRef(parentDeclRef, transparentInfo.decl);
 
+        // If we have already started checking this member, there is no need to send another check.
```
csyonghe, line 233: Why do we need to change the lookup logic to prevent the diagnostics? It seems like lookup shouldn't be responsible for this?

csyonghe (dismissed):

tests/language-feature/enums/unscoped-enum-explicit-type copy.slang
```diff
@@ -0,0 +1,27 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj
+
+//TEST_INPUT:ubuffer(data=[0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<uint32_t> outputBuffer;
+
+interface myType
```
jkwak-work, line 6: Can you leave a comment somewhere here to describe what problem this is testing?
Please provide a summary so that we don't need to read the whole code to understand the purpose of it.

tests/language-feature/enums/unscoped-enum-explicit-type copy.slang
```diff
@@ -0,0 +1,27 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj
```
jkwak-work, line 1: Can you rename the "file name" properly?

ArielG-NV: I accidently added the file, there was only supposed to be 1

ArielG-NV:
note: due to regressions, logic had to be amended to not affect all inheritanceDecl since generics would break

csyonghe (approved):

PR #3963, opened 2024-04-16
Support combined texture sampler when targeting HLSL.

### LINKED ISSUE #3917: Are combined sampler/texture objects supported?
I see there is a `Sampler2D` type, but attempting to use it:

```hlsl
Sampler2D bar;

void frag()
{
}
```
With
```
slangc test.slang -stage fragment -entry frag -o test.hlsl
```
Results in:
```
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unexpected: unhandled shader parameter kind
```
Version v2024.1.6

csyonghe:
Closes #3917.


source/slang/slang-ir-lower-combined-texture-sampler.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        LowerCombinedSamplerContext context;
+
+        // Lower combined texture sampler type into a struct type.
+        for (auto globalInst : module->getGlobalInsts())
+        {
+            if (auto textureType = as<IRTextureTypeBase>(globalInst))
+            {
+                if (getIntVal(textureType->getIsCombinedInst()) != 0)
+                {
```
ArielG-NV, line 108:
```suggestion
            auto textureType = as<IRTextureTypeBase>(globalInst);
            if (!textureType  || getIntVal(textureType->getIsCombinedInst()) != 0)
            {
```
reduce indentation

source/slang/slang-ir-lower-combined-texture-sampler.cpp
```diff
@@ -0,0 +1,208 @@
...
+                    {
+                        auto typeUser = use->getUser();
+                        if (use != &typeUser->typeUse)
+                            continue;
+
+                        if (auto layoutDecor = typeUser->findDecoration<IRLayoutDecoration>())
+                        {
+                            // Replace the original VarLayout with the new StructTypeVarLayout.
+                            if (auto varLayout = as<IRVarLayout>(layoutDecor->getLayout()))
+                            {
```
ArielG-NV, line 121:
```suggestion
                        auto layoutDecor = typeUser->findDecoration<IRLayoutDecoration>();
                        if (!layoutDecor)
                            continue;
                        // Replace the original VarLayout with the new StructTypeVarLayout.
                        auto varLayout = as<IRVarLayout>(layoutDecor->getLayout());
                        if (!varLayout)
                            continue;
```

source/slang/slang-ir-lower-combined-texture-sampler.cpp
```diff
@@ -0,0 +1,208 @@
...
+                {
+                    nextInst = inst->getNextInst();
+                    switch (inst->getOp())
+                    {
+                    case kIROp_CombinedTextureSamplerGetTexture:
+                    case kIROp_CombinedTextureSamplerGetSampler:
+                        {
+                            auto combinedSamplerType = inst->getOperand(0)->getDataType();
+                            auto loweredInfo = context.mapTypeToLoweredInfo.tryGetValue(combinedSamplerType);
+                            if (!loweredInfo) continue;
```
ArielG-NV, line 186:
```suggestion
                            if (!loweredInfo) 
                                continue;
```
new line continue to allow debugging (code step-ing)

ArielG-NV:
1 suggestion, otherwise everything else looks good to me.
Since combined texture sampler fills a texture and sampler binding slot, I think we should have a test where we trigger a `Diagnostics::parameterBindingsOverlap` warning to be sure the locations emit as expected without side effects

csyonghe:
Yeah, the test is trying to cover that with the `//CHECK` lines that include the bindings.

Probably worth adding to the test some more texture and sampler type parameters to make sure they are correctly laid out after the combined sampler.

pema99:
Quick question: Capitalization. In HLSL, to my knowledge, only `sampler2D` is a valid construct, and not `Sampler2D`. Does the Slang parser accept the lowercase variant? When I tried with the last release, I was getting errors when I tried it. I could work around it with a `#define`, but I'm wondering if its intentional.

Just to illustrate what I mean:

test_good.hlsl:
```glsl
sampler2D foo;

float4 main() : SV_Target {
	return tex2D(foo, 0);
}
```
test_bad.hlsl:
```glsl
Sampler2D foo;

float4 main() : SV_Target {
	return tex2D(foo, 0);
}
```

```
> fxc test_good.hlsl /Tps_5_0 /Gec
... Valid output

> fxc test_bad.hlsl /Tps_5_0 /Gec
test_bad.hlsl(1,1): error X3000: unrecognized identifier 'Sampler2D'
```

csyonghe:
We do not support `sampler2D`.
`Sampler2D` is a slang-specific type that is not intended to provide backward compatibility with the legacy `sampler2D` in old HLSL. We also don't have intrinsic support for things like `tex2D`.

pema99:
> We do not support `sampler2D`. `Sampler2D` is a slang-specific type that is not intended to provide backward compatibility with the legacy `sampler2D` in old HLSL. We also don't have intrinsic support for things like `tex2D`.

That makes sense. Just adding support for the Slang-specific Sampler2D type, as you've done in this PR, is sufficient to cover my use case. Thanks again for the quick answer :)

csyonghe:
Thank you! Please don't hesitate to let us know if you run into any further issues.

source/slang/slang-ir-lower-combined-texture-sampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+        LowerCombinedSamplerContext context;
+
+        // Lower combined texture sampler type into a struct type.
+        for (auto globalInst : module->getGlobalInsts())
+        {
+            auto textureType = as<IRTextureTypeBase>(globalInst);
+            if (!textureType || getIntVal(textureType->getIsCombinedInst()) == 0)
+                continue;
+                    auto typeInfo = context.lowerCombinedTextureSamplerType(textureType);
```
jkwak-work, line 108: indentation seems off.

source/slang/slang-ir-lower-combined-texture-sampler.cpp
```diff
@@ -0,0 +1,206 @@
...
+        IRStructType* type;
+        IRType* samplerType;
+        IRType* textureType;
+        IRTypeLayout* typeLayout;
+    };
+
+    struct LowerCombinedSamplerContext
+    {
+        Dictionary<IRType*, LoweredCombinedSamplerStructInfo> mapTypeToLoweredInfo;
+        TargetRequest* target;
```
jkwak-work, line 22: It seems like "target" is unused. You may want to remove it?

jkwak-work (approved):
It looks good to me.

Issue #3935, opened 2024-04-11
Two same "struct" are generated with different names when an arithmetic math is applied to the generic-argument


jkwak-work:
**Problem Description**
While integrating Slang to UE 5.3, the following issue was encountered.

It appears that there is a bug on the deduplication of the generic arguments when the generic argument got an arithmetic math.

**Goal**
The generic arguments should be uniquely identified even when there are any arithmetic math applied to it.

**Repro-steps***
It is easier to show the problem than explaining it.
```
RWStructuredBuffer<float> output;

__generic<let ArraySize:int>
struct MyStruct
{
  int elems[ArraySize];
};

__generic<let ArraySize:int>
MyStruct<ArraySize / 2> Reduce(MyStruct<ArraySize> o)
{
  MyStruct<ArraySize / 2> result;
  result.elems[0] = o.elems[0];
  return result; // <==== Error happens here
}

[numthreads(4, 1, 1)]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
  MyStruct<4> a;
  output[0] = Reduce(a).elems[0];
}
```
When HLSL is generated from the Slang shader above, the function "Reduce" looks like following.
```
MyStruct_0 Reduce_0(MyStruct_1 o_0)
{
    MyStruct_2 result_0;
    result_0.elems_0[int(0)] = o_0.elems_0[int(0)];
    return result_0;
}
```
Note that the return type is "MyStruct_0" and it is trying to return a different type, "MyStruct_2".
Their implementations are same as you can see below,
```
struct MyStruct_0
{
    int  elems_0[int(2)];
};
struct MyStruct_2
{
    int  elems_0[int(2)];
};
```
This causes the compilation to fail.

swoods-nv:
Jay, is there an estimate for this one?

jkwak-work:
This is probably an easy one but I don't have much confidence.
I put 3 days for now.

csyonghe:
The issue is usually because we failed to identify the two specializations are the same thing.

In IR, we usually deduplicate `specialize` insts. When we specialize the `Reduce` function, we want to make sure we didn't create duplicated specialized structs. To trace this down, you can set a breakpoint after specialization pass, and inspect the IR for duplicate struct types. Find the duplicate struct type with the bigger UID, and try to track down why it gets created. There is a lookup process before creating the specialized struct type to see if a specialization is already created. Check why that lookup failed to determine the case.

csyonghe:
[...]

PR #3968, opened 2024-04-16
Support arithmetics on generic arguments

jkwak-work:
Resovles an issue #3935

Slang had to fold the generic arguments after specialization.

### DIFF
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -7,6 +7,7 @@
 #include "slang-ir-ssa-simplification.h"
 #include "slang-ir-lower-witness-lookup.h"
 #include "slang-ir-dce.h"
+#include "slang-ir-sccp.h"
 #include "../core/slang-performance-profiler.h"
 
 namespace Slang
@@ -211,7 +212,29 @@ struct SpecializationContext
         IRGeneric* genericVal,
         IRSpecialize* specializeInst)
     {
-        // First, we want to see if an existing specialization
+        // We need to fold the generic arguments here in order to uniquely identify
+        // which specializations need to be generated.
+        // The folding of the generic-arguments are deferred until the specialization
+        // step here, because the exact value of the generic-arguments can be unknown
+        // until the specialization. The exact values of the generic-arguments may come
+        // from the other modules, and they will be unknown until linking the modules.
+        //
+        UInt argCount = specializeInst->getArgCount();
+        {
+            IRBuilder builder(module);
+            for (UInt ii = 0; ii < argCount; ++ii)
+            {
+                IRUse* argUse = specializeInst->getArgOperand(ii);
+                auto originalArg = argUse->get();
+                IRInst* foldedArg = tryConstantFoldInst(module, originalArg);
+                if (foldedArg == originalArg)
+                    continue;
+
+                specializeInst = as<IRSpecialize>(builder.replaceOperand(argUse, foldedArg));
+            }
+        }
+
+        // We want to see if an existing specialization
         // has already been made. To do that we will construct a key
         // for lookup in the generic specialization context.
         //
@@ -223,7 +246,6 @@ struct SpecializationContext
         //
         Key key;
         key.vals.add(specializeInst->getBase());
-        UInt argCount = specializeInst->getArgCount();
         for (UInt ii = 0; ii < argCount; ++ii)
         {
             key.vals.add(specializeInst->getArg(ii));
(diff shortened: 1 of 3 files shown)


source/slang/slang-ir-specialize.cpp
```diff
@@ -226,7 +227,9 @@ struct SpecializationContext
         UInt argCount = specializeInst->getArgCount();
         for (UInt ii = 0; ii < argCount; ++ii)
         {
-            key.vals.add(specializeInst->getArg(ii));
+            IRInst* genericArg = specializeInst->getArg(ii);
```
csyonghe, line 230: It would be better if we can also need to specialize the generic using the folded args.

csyonghe (approved):

source/slang/slang-ir-specialize.cpp
jkwak-work: I will find a better place to do the folding.

jkwak-work: I was trying to find a better place to do the folding; preferably earlier location than here.
But it seems like this is the best place to fold.
Because here is the first place that branches to either find one or generate one.

I look back on your comment and if I take "also" as a keyword, I can find another location to do the folding at a location earlier than "BEFORE-SPECIALIZE" point.
The last part of a function, "cloneGlobalValueWithCodeCommon()", is where generic functions are copied.
I am not sure why copies are happening, but it seems to make copies two copies for the given test case below.

Maybe I need to look for a place where the initial one gets built up and do the folding there.

I can use some help based on what I have learned so far. @csyonghe

csyonghe: Sorry about the confusion.

This is the best place to try folding.

My previous comment is that it is better if we can refactor specializeGenericImpl function called down below to take in the newly folded args and use them to perform the actual specialization, so we generate more compact specialized code.

source/slang/slang-ir-specialize.cpp
```diff
@@ -211,7 +212,28 @@ struct SpecializationContext
...
+        for (UInt ii = 0; ii < argCount; ++ii)
+        {
+            IRInst* originalArg = specializeInst->getArg(ii);
+            IRInst* foldedArg = tryConstantFoldInst(module, originalArg);
+            if (foldedArg == originalArg)
+                continue;
+
+            IRBuilder builder(module);
+            builder.setInsertBefore(originalArg);
+            originalArg->replaceUsesWith(foldedArg);
```
csyonghe, line 232: Note that `replaceUsesWith` may delete and replace `specailizeInst` with some existing inst.

The safest way to do this is:
```
specailizeInst = builder.replaceOperand(specailizeInst->getArgOperand(ii), foldedArg)
```

You probably need to add that `getAgOperand()` function to return an `IRuse` instead of an IRInst*.

jkwak-work: Thank you for the information.
I modified it to use IRUse.

source/slang/slang-ir-specialize.cpp
```diff
@@ -211,7 +212,37 @@ struct SpecializationContext
...
+        // from the other modules, and they will be unknown until linking the modules.
+        //
+        UInt argCount = specializeInst->getArgCount();
+        {
+            IRBuilder builder(module);
+            for (UInt ii = 0; ii < argCount; ++ii)
+            {
+                IRUse* nextUse = nullptr;
+                IRInst* argInst = specializeInst->getArg(ii);
+                for (auto iuse = argInst->firstUse; iuse; iuse = nextUse)
```
csyonghe, line 229: There is not need to iterate over the uses of argInst.

```
IRUse* argUse = specializeInst->getArgOperand(ii);
auto originalArg = argUse->get();
IRInst* foldedArg = tryConstantFoldInst(module, originalArg);
if (foldedArg == originalArg)
    continue;

specializeInst = builder.replaceOperand(argUse, foldedArg);
```

jkwak-work: I don't see a function called "getArgOperand()" anywhere.
Instead, I can do something like this if I don't need to iterate for uses.
```
            IRBuilder builder(module);
            for (UInt ii = 0; ii < argCount; ++ii)
            {
                IRInst* argInst = specializeInst->getArg(ii);
                IRUse* argUse = argInst->firstUse;
                auto originalArg = argUse->get();
                IRInst* foldedArg = tryConstantFoldInst(module, originalArg);
                if (foldedArg == originalArg)
                    continue;

                specializeInst = as<IRSpecialize>(builder.replaceOperand(argUse, foldedArg));
            }
```

csyonghe: Just add it to that class:
```
IRUse* getArgOperand(Index i) { return getOperands() + 1 + i; }
```

csyonghe: Note that argInst->firstUse is not necessarily the operand of the IRCall.

jkwak-work: I see.
That makes sense.

csyonghe (approved):

PR #3972, opened 2024-04-17
Incorrect argument parsing by slangc -- partially fix #3931 

### LINKED ISSUE #3931: Alignment of float4x4 seems to change in spirv backend depending on surrounding code
Hi,

using Slang v2024.1.6 the following code behaves quite strange when targeting spirv. It looks like the alignment of the 4x4 matrix `FooBar::c` changes depending on the surrounding code:

```c++
//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-vk -emit-spirv-directly -compute -shaderobj -xslang -matrix-layout-row-major -xslang -O0
//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-slang -compute -cpu -shaderobj -xslang -matrix-layout-row-major -xslang -O0

//#define WORKAROUND_A // manual padding
//#define WORKAROUND_B // manual loop unroll
//#define WORKAROUND_C // do not use pointers to load memory

struct FooBar {
    uint64_t a;
    uint32_t b;
#ifdef WORKAROUND_A
    uint32_t pad;
#endif
    float4x4 c;

    int load() {
        return *((int *)a);
    }
};

//TEST_INPUT: ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer

//TEST_INPUT:ubuffer(data=iota(0,2048,1),stride=4):name=sb
uniform StructuredBuffer<FooBar, ScalarDataLayout> sb;

int test(int val)
{
    uint32_t res = 0;
#ifndef WORKAROUND_B
    for (int i = 0; i < 4; ++i) {
        res |= bit_cast<uint32_t>(sb[val].c[0][i]) << i * 8;
    }
#else
    res |= bit_cast<uint32_t>(sb[val].c[0][0]) << 0;
    res |= bit_cast<uint32_t>(sb[val].c[0][1]) << 8;
    res |= bit_cast<uint32_t>(sb[val].c[0][2]) << 16;
    res |= bit_cast<uint32_t>(sb[val].c[0][3]) << 24;
#endif

#ifndef WORKAROUND_C
    if (val == 42) { // is never true when running this test
        res |= sb[val].load();
    }
#endif
    return bit_cast<int>(res);
}

[numthreads(4, 1, 1)]
[shader("compute")]
void computeMain(
    int3 dispatchThreadID : SV_DispatchThreadID)
{
    int tid = dispatchThreadID.x;

    int inVal = tid;
    int outVal = test(inVal);

    // CHECK: 6050403
    // CHECK: 1A191817
    // CHECK: 2E2D2C2B
    // CHECK: 4241403F
    outputBuffer[tid] = outVal;
}
```
To run this code with your testing framwork, you need this [patch that adds iota data initialization](https://github.com/shader-slang/slang/files/14944806/slang-cbdc7d6-Added.an.iota.generator.for.test.buffer.data.initialization.patch). Feel free to put it into your code base in case you find it useful.

If any of the `WORKAROUND_*` defines are active, the SPIRV backend behaves as expected, if all of these are off, it seems to ignore the alignment of the 4x4 matrix and returns this result:
```
3020100
17161514
2B2A2928
3F3E3D3C
```

This one really took me a while to find a working reproducer for as it requires some pointer access code into memory _anywhere_ in the program to reproduce. This is why I think it might be caused by the changing the spirv memory model globally `Logical` from `PhysicalStorageBuffer64`, as soon as such code exists. It is the same problem I already was trying to reproduce with #3889, but failed due to wrong usage of your testing framework. I hope this one is an actual reproducer.

Thanks for having a look!

ArielG-NV:
Problems:
1. Incorrect slangc arg parsing (possible to have row major & column major in option_set) -- addressed in pr.
2. spirv-opt 2023-6 & upstream incorrectly optimizes code and removes member decorations in some odd cases, works with spirv-opt 2024-1 -- not solved in this pr since we use upstream of spirv-opt currently to solve different issues.


csyonghe (approved):

tests/bugs/gh-3931.slang
```diff
@@ -0,0 +1,24 @@
+//TEST:SIMPLE(filecheck=CHECK): -O0 -target spirv -emit-spirv-directly -stage compute -entry computeMain -matrix-layout-row-major
+//COM:TEST:SIMPLE(filecheck=CHECK): -O2 -target spirv -emit-spirv-directly -stage compute -entry computeMain -matrix-layout-row-major
```
jkwak-work, line 2: Is the second test meant to be commented out with "//COM:"?
I thought this test is to test if the row-major-ness works properly or not when "-O2" is used, isn't it?

ArielG-NV: there were 2 problems:
> Incorrect slangc arg parsing (possible to have row major & column major in option_set) -- addressed in pr.

> spirv-opt 2023-6 & upstream incorrectly optimizes code and removes member decorations in some odd cases, works with spirv-opt 2024-1 -- not solved in this pr since we use upstream of spirv-opt currently to solve different issues.

Of these, we can only fix problem one since this is a problem inside Slang. This allows all code to work every-run with row major'ness set.
Problem 2 cannot be addressed since the spirv-opt version we are using cannot be downgraded. using O0 implies no optimization, O2 implies using spirv-opt, which we cannot do.

PR #3977, opened 2024-04-18
ForceInline SampleLevel to allow decorations to be applied

### LINKED ISSUE #3969: ForceInline atomic operations on Byte Address Buffers
When using atomic operations such as `InterlockedAdd` on Byte Address Buffers, the resultant IR shows as Call operation, as follows:
```
639}   func %main      : Func(Void)
{
...
        {1541}  let  %26        : UInt  = swizzle(%25, 0 : Int)
        {1542}  let  %27        : Int   = intCast(%26)
        {1438}  SPIRVAsm %28    : Int
        {
                {1441}  SPIRVAsmInst(83 : UInt, Int, result, 0 : Int)
        }
        {1445} let %29: Int = nonUniformResourceInfo(%28)  
        {1501}  call %RWByteAddressBufferx5FInterlockedAdd(%29, 0 : UInt, 1 : UInt)
        
        {973}   let  %30        : UInt  = intCast(%28)
        {992}   let  %31        : Ptr(Float)    = rwstructuredBufferGetElementPtr(%output, %30)
        {1486}  let  %32        : UInt  = call %RWByteAddressBufferx5FLoad(0 : Int, 0 : Int)
        ...
}
```

This prevents any decorations applied on the Atomic operation, since it is encapsulated by the call.

The goal is to ForceInline the atomic operations so that the decorations can be applied to the appropriate instruction.
For example, after inlining, the above IR would look like
```
 [{1480} NonUniformResource]
{1476} let  %31 : Ptr(%RWStructuredBuffer, 12 : Int) = getElementPtr(%buffer, 2 : Int)
{1483} let  %32 : Ptr(UInt, 12 : Int) = rwstructuredBufferGetElementPtr(%31, 0 : UInt)
{1397} SPIRVAsm %33 : UInt
{
 {1401} SPIRVAsmInst(234 : UInt, UInt, result, rwstructuredBufferGetElementPtr(%31, 0 : UInt), 1 : UInt, 0 : UInt, 1 : UInt)
}
```

sriramm-nv:
Fixes #3969 

NonUniformResourceInfo instruction is applied as a Decoration on the backing resource. With the following shader, this is applied to the Function Call.
```
 vRadiance.rgb *= g_bindless_Texture2D [NonUniformResourceIndex (light.m_nCookie_nLuminaireType_nPathTraceFlags.x)].SampleLevel(g_Sampler, vUv, 0.0).rgb;
```
as shown below:
```
	{145371}	let  %1826	: Int	= nonUniformResourceIndex(%1789)
	{177146}	let  %1828	: Vec(Float, 4 : Int)	= call %SampleLevel(%1826, %gx5FCookieSampler, %1827, 0 : Float)
```

This patch ForceInlines SampleLevel intrinsic function call so that the Decoration is correctly applied on the resource.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -618,6 +618,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,1,format>
     }
 
     [__readNone]
+    [ForceInline]
     T SampleLevel(vector<float, Shape.dimensions+isArray> location, float level)
     {
         __target_switch
@@ -668,6 +669,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,1,format>
     }
 
     [__readNone]
+    [ForceInline]
     T SampleLevel(vector<float, Shape.dimensions+isArray> location, float level, constexpr vector<int, Shape.planeDimensions> offset)
     {
         __target_switch
@@ -1045,6 +1047,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
     }
 
     [__readNone]
+    [ForceInline]
     T SampleLevel(SamplerState s, vector<float, Shape.dimensions+isArray> location, float level)
     {
         __target_switch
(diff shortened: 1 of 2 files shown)


csyonghe (approved):
We need to do a systematic pass over all intrinsics on resource types and make sure they are all inlined.

sriramm-nv:
Got it. I'll submit that as a follow-up

Issue #3516, opened 2024-01-25
Option to zero-initialize all variables


tangent-vector:
All GPU stages typically care about minimizing live state (aka "register pressure"), but this is especially acute for ray tracing pipelines, where thread state may need to be evicted/stored as part of scheduling or rebalancing work. At points where a thread might be suspended, the state that needs to be stored out includes all of the live variables.

Downstream compilers can infer the liveness of variables by looking at loads and stores, but there can be cases where the Slang compiler has more complete information about live ranges that a downstream compiler cannot infer.

A concrete example is when there is a variable declared inside a loop, that is conditionally assigned on each iteration:

```
for(...)
{
    SomeType tmp;
    ...
    // code that may or may not fully initialize `tmp`
    ...
    someFunctionThatUses( tmp );
    ...
    someFunctionThatMaySuspendThisThread();
}
```

Because of constraints on the SPIR-V and DXIL representations, compilation for those targets will effectively move the declaration of `tmp` outside of the loop:

```
SomeType tmp;
for(...)
{ ... }
```

As a result, it can be difficult or impossible for a downstream compiler to know that the value of `tmp` from one loop iteration cannot be observed by the next iteration. The downstream compiler may have to be conservative and save/restore the state of `tmp` whenever the thread suspends in the loop.

While there are more narrow ways to inform a downstream compiler about liveness, and we can/should support those in Slang, there is also a relatively simple fix that can help in many scenarios: guarantee that all variables are fully initialized at their point of declaration.

Effectively, that means that in the scenario above, even though `tmp` will be hoisted out of the loop, we would emit a complete initialization of it on each loop iteration:

```
SomeType tmp;
for(...)
{
    tmp = { 0, ... }; // how ever many `0`s are needed
    ...
}
```

With that representation, a downstream compiler can easily see that the assignment at the top of each loop iteration effectively "kills" the value of `tmp` from the previous iteration, so that it is no longer live at the potential thread suspend point.

Because this feature might impact the performance of generated code, it should probably be enabled under a switch at first.

csyonghe:
Is there a work load where we can verify the potential gain of this change?

csyonghe:
I doubt this matters to downstream today since there is pervasive inlining and SROA.

nsubtil:
RTX Remix with SER disabled shows strong benefits from a change like this.

csyonghe:
That's good to know.

When implementing this, we need to be careful about the case where `SomeType` has resource type fields.

PR #3987, opened 2024-04-19
Add slangc flag to `-zero-initialize` all variables

ArielG-NV:
Adds `-zero-initialize` flag to set values to a __default() expression if they are missing an initExpr.
Fixes [#3516](https://github.com/shader-slang/slang/issues/3516)


source/slang/slang-check-decl.cpp
```diff
@@ -1636,6 +1636,23 @@ namespace Slang
             }
         }
 
+        if (!varDecl->initExpr &&
```
csyonghe, line 1639: I think this logic should be in `SemanticsDeclBodyVisitor::checkVarDeclCommon` instead.
There we are checking if a default ctor exists and calls it if so.

We should extend that logic by saying if default ctor does not exist, then initialize it with `__default<T>()`.

tests/language-feature/enums/zero-initialize/generic.slang
```diff
@@ -0,0 +1,88 @@
...
+{
+// BUF: 1
+    outputBuffer[0] = true
+    && getValue<int>()
+#ifndef DX12
+    && getValue<uint8_t>()
+    && getValue<int8_t>()
+    && getValue<uint16_t>()
+    && getValue<int16_t>()
+#endif
```
jkwak-work, line 34: 8bit is not supported by dx12 but 16bit is.
Didn't 16bit work?

tests/language-feature/enums/zero-initialize/regular.slang
```diff
@@ -0,0 +1,64 @@
...
+    vector<int, 4> vectorVal3 = getDefaultIntVec4();
+// BUF: 1
+    outputBuffer[0] = true
+    && reg1 == 0
+    && reg2 == 0
+    && reg3 == 0
+    && vectorVal1 == vector<int, 2>(0)
+    && vectorVal2 == vector<int, 3>(0)
+    && vectorVal3 == vector<int, 4>(0)
+    && MyFlags::One == MyFlags(1)
```
jkwak-work, line 62: I am not sure why this enum case is a part of the test.
Can we do something more like this?
```
double getDefaultEnum()
{
    MyFlags val;
    return val;
}
```

tests/language-feature/enums/zero-initialize/generic.slang
```diff
@@ -0,0 +1,88 @@
...
+    typealias gvec = vector<T, N>;
+
+    gvec genericOut;
+    return genericOut == gvec(T(0));
+}
+[numthreads(1, 1, 1)]
+void computeMain(int3 dispatchThreadID: SV_DispatchThreadID)
+{
+// BUF: 1
+    outputBuffer[0] = true
```
jkwak-work, line 27: Shouldn't we also test boolean?

tests/language-feature/enums/zero-initialize/regular.slang
```diff
@@ -0,0 +1,64 @@
...
+//TEST(smoke,compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-dx12 -use-dxil -compute -entry computeMain -allow-glsl -xslang -zero-initialize
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+enum MyFlags
+{
+    Zero = 0,
+    One,
+}
```
jkwak-work, line 13: I am wondering if we should also test cases of enum that inherited from integer types.
Something like,
```
enum MyEnum : uint16_t // <=== all integer types like int8_t and int64_t
{ Whatever };
MyEnum var;
Test if var is zero.
```

csyonghe:
You should write 
Fixes #xxxx
Or
Closes #xxxx
For GitHub to automatically link the issue.

tests/language-feature/enums/zero-initialize/regular.slang
ArielG-NV: The goal is to ensure we do not break enum's by giving an init expression.

tests/language-feature/enums/zero-initialize/regular.slang
ArielG-NV: I can add it, never a problem to add more.

tests/language-feature/enums/zero-initialize/generic.slang
ArielG-NV: the types work, I forgotten to set the profile to sm_6_2.

source/slang/slang-check-decl.cpp
ArielG-NV: We will not be-able todo this logic in BodyVisitor since then when setting init's to struct fields we would have already parsed+setup the default ctor body logic.

source/slang/slang-ast-modifier.h
```diff
@@ -1585,6 +1585,24 @@ class DynamicUniformModifier : public Modifier
     SLANG_AST_CLASS(DynamicUniformModifier)
 };
 
+// Sharing IRInst data of AggType (Struct/Interface) with members.
```
csyonghe, line 1588: This shouldn't be needed. In general we should never be creating AST nodes that references IR constructs.

source/slang/slang-ir.h
```diff
@@ -1856,8 +1856,10 @@ struct IRStructField : IRInst
 struct IRStructType : IRType
 {
     IRFilteredInstList<IRStructField> getFields() { return IRFilteredInstList<IRStructField>(getChildren()); }
-
+    
     IR_LEAF_ISA(StructType)
+
+    IRFunc* defaultCtor = nullptr;
```
csyonghe, line 1862: We shouldn't be adding fields to IR insts like this. All references must be encoded as either a child or an operand.

And IRStruct should not be referencing ctors directly. You can use a decoration to associate a default ctor if necessary. But i don't think we should need this, because we should be generating all default ctor calls during IR lowering, and we should never need to have the concept of a ctor in the IR.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -8549,6 +8551,18 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
             }
         }
 
+        // Add modifier to constructors so they are aware of the parent IRType.
+        // This follows the lowering logic of ensureAllDeclsRec->{visitAggTypeDecl, visitFunctionDeclBase}
+        if(isStruct)
+        {
+            for (auto ctor : decl->getMembersOfType<ConstructorDecl>())
+            {
+                auto parentStructTypeModifier = this->context->astBuilder->create<ParentStructTypeModifier>();
```
csyonghe, line 8560: You can simply maintain a side band dictionary in the ir lowering context to keep track of struct types and their ctors.

tests/language-feature/zero-initialize/generic.slang
```diff
@@ -0,0 +1,107 @@
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl -xslang -zero-initialize
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -emit-spirv-directly -allow-glsl -xslang -zero-initialize
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-cpu -compute -entry computeMain -allow-glsl -xslang -zero-initialize
+//TEST(smoke,compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-dx12 -use-dxil -compute -entry computeMain -allow-glsl -profile sm_6_2 -xslang -zero-initialize -xslang -DDX12
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+__generic<T : __BuiltinArithmeticType>
```
csyonghe, line 9: For generics, we will need a IDefaultInitializable builtin interface that has a __init() requirement.

Then for any generic type that conforms to IDefaultInitializable, we can just call the interface method to initialize it. By default, we make all builtin types conform to this interface, and we can make all user defined struct types to automatically conform to the interface as we synthesize the default constructor.

source/slang/slang-ir.h
csyonghe: I think this is a hole in our generics system that we don't have to address in this PR. 

What we probably should do now is that if T doesn't have a default initializer from any constraints, then we don't call any default ctor even after specialization. However, we still want to make sure we zero initialize whatever type it became after specialization with defaultConstruct inst.


So when we lower the generic function to IR, we check if T has default ctor, if it has we call it. If not we emit a defaultConstruct inst.

csyonghe: If we do this, we don't need to keep track of the default ctor of each struct type in the IR in any explicit way.

tests/language-feature/zero-initialize/generic.slang
```diff
@@ -0,0 +1,107 @@
...
+bool getValue()
+{
+    T genericOut;
+    return genericOut == T(0);
+}
+bool getValueB()
```
jkwak-work, line 15: Can you rename to `getValueBoolean()`

tests/language-feature/zero-initialize/generic.slang
```diff
@@ -0,0 +1,107 @@
...
+{
+    bool genericOut;
+    return genericOut == bool(0);
+}
+
+__generic<T : __BuiltinArithmeticType, let N : int>
+bool getValueV()
```
jkwak-work, line 22: Please rename this to `getValueVector()`

tests/language-feature/zero-initialize/regular.slang
```diff
@@ -0,0 +1,92 @@
...
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+enum MyFlags1
+{
+    Zero = 0,
+    One,
+}
+
+enum MyFlags2 : uint8_t
```
jkwak-work, line 15: Please rename this to reflect the nature of the type.
Something like `MyFlag_uint8`

jkwak-work: I am curious if and how this 8 bit integer works for DX12.

tests/language-feature/zero-initialize/regular.slang
```diff
@@ -0,0 +1,92 @@
...
+    float reg2 = getDefaultFloat();
+    double reg3 = getDefaultDouble();
+    vector<int, 2> vectorVal1 = getDefaultIntVec2();
+    vector<int, 3> vectorVal2 = getDefaultIntVec3();
+    vector<int, 4> vectorVal3 = getDefaultIntVec4();
+// BUF: 1
+    outputBuffer[0] = true
+    && reg1 == 0
+    && reg2 == 0
+    && reg3 == 0
```
jkwak-work, line 82: Please change this to be more readable.
Something like
```
    && defaultInt == int(0)
    && defaultFloat == float(0)
    && defaultDouble == double(0)
```

jkwak-work:
I am wondering if we can have tests for cases when "-zero-initialize" is not used.
In game development, the uninitialized state is preferred over implicit initialization with zero values for the runtime performance reason.
I am little worried that we may accidently initialize variables to zero even when "-zero-initialize" is not used.

csyonghe:
@jkwak-work Agreed that we need a test case to cover this.

Basically, when -zero-initialize is not specified, then we should *not* be generating default ctors if a type does not have any member initializers, when we should *not* be generating assignments in default ctors for members that are not initialized.

Then, if we don't synthesis such a default ctor, we won't be calling the default ctor at variable declaration sites.

ArielG-NV:
I will add the test case, I agree that there should be one for each possible senario: 
1. default init generated due to field init expression
2. no init should be generated due to 0 fields with init expressions

source/slang/slang-ir.h
ArielG-NV: The workaround added was not to handle generics, generics work as intended with __default.

Passing a struct<->Ctor was done because of the cases like `vector<...struct...>`. We don't exactly know what is inside the container, so we cannot simply call an 'invoke' to the ctor. Only __default has emitting logic to manage these code edge cases which is why a ctor was passed through.

csyonghe: Well,if you have:
```
struct MyContainer<T>
{
     T obj;
}
```
Then the right answer is always do `obj = __default<T>()`, which will eventually lower into `defaultConstruct`.

If you have:
```
struct MyContainer<T:IDefaultInitializable> // or any other interface that leads to `__init()` to be found
{
    T obj;
}
```
Then you lower it into:
```
obj = T();
```

tests/language-feature/zero-initialize/generic.slang
ArielG-NV: This would solve the problem we have with vectors and remove the need for a workaround

we could define the vector IDefaultInitializable as the base type default, which would resolve into a ctor

source/slang/slang-ir.h
csyonghe: In either case, you shouldn't need to keep track of this additional ctor info.

tests/language-feature/zero-initialize/generic.slang
csyonghe: Yes, all builtin interfaces should inherit from IDefaultInitializable.

tests/language-feature/zero-initialize/struct-no-zero-init.slang
```diff
@@ -0,0 +1,37 @@
...
+    MyStruct myStruct;
+    return myStruct;
+}
+[numthreads(1, 1, 1)]
+void computeMain(int3 dispatchThreadID: SV_DispatchThreadID)
+{
+    MyStruct myStruct = getStruct();
+#ifdef fail
+    myStruct.
+#endif
```
jkwak-work, line 29: Not sure what these is for.
It might be related to the line 3 where "FAIL" is defined?

source/slang/slang-check-decl.cpp
```diff
@@ -1639,7 +1639,8 @@ namespace Slang
             }
         }
 
-        // We must keep zero-initialize in SemanticsDeclHeaderVisitor.
+        // We must keep zero-initialize for struct fields in
```
csyonghe, line 1642: I am not sure I understand the comment here.

source/slang/slang-check-decl.cpp
csyonghe: We should be synthesizing ctor signature in header visitor, and synthesize the ctor body in body visitor, there we call the ctors of the field. Why isn't this possible? We shouln't be checking the init exprs in body visitor IMO.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -498,6 +498,9 @@ struct SharedIRGenContext
     Dictionary<Stmt*, IRBlock*> breakLabels;
     Dictionary<Stmt*, IRBlock*> continueLabels;
 
+    // A reference manager to a ctor's parent
+    Dictionary<FunctionDeclBase*, IRType*> ctorBaseToParentStruct;
```
csyonghe, line 502: We should be able to get this with `findResultTypeForConstructorDecl`?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -9968,9 +9969,9 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         auto subContext = nestedContextFunc.getContext();
         auto lowered = lowerFuncDeclInContext(subContext, subBuilder, decl);
 
-        auto parentStructTypeModifier = decl->findModifier<ParentStructTypeModifier>();
-        if (parentStructTypeModifier && decl->members.getCount() == 0)
-            as<IRStructType>(parentStructTypeModifier->getArg())->defaultCtor = as<IRFunc>(lowered.val);
+        IRType** maybeParentStruct = this->context->shared->ctorBaseToParentStruct.tryGetValue(decl);
+        if (maybeParentStruct && decl->members.getCount() == 0)
```
csyonghe, line 9973: Don't think this is needed.

tests/language-feature/zero-initialize/struct-no-zero-init.slang
ArielG-NV: I was testing the language server and was going to add a language server test, but realized it just works as is.

source/slang/slang-lower-to-ir.cpp
ArielG-NV: SemanticsVisitor logic is disjoint from IRGenContext and is not something which can work-without a fully visited SemanticsVisitor?

source/slang/slang-lower-to-ir.cpp
ArielG-NV: (reference to previous comment - https://github.com/shader-slang/slang/pull/3987#discussion_r1576752102)

source/slang/core.meta.slang
```diff
@@ -1081,6 +1117,15 @@ struct matrix : IArray<vector<T,C>>
     int getCount() { return R; }
 
     __subscript(int index) -> vector<T,C> { __intrinsic_op($(kIROp_GetElement)) get; }
+
+    [ForceInline]
+    [TreatAsDifferentiable]
+    __init()
```
csyonghe, line 1123: This can just map to `kIROp_MakeMatrixFromScalar(T())`.

source/slang/core.meta.slang
```diff
@@ -1056,6 +1083,14 @@ struct vector : IArray<T>
     int getCount() { return N; }
 
     __subscript(int index) -> T { __intrinsic_op($(kIROp_GetElement)) get; }
+
+    [ForceInline]
+    [TreatAsDifferentiable]
+    __init()
```
csyonghe, line 1089: klIROp_MakeVectorFromScalar(T())

source/slang/slang-ast-type.h
```diff
@@ -97,6 +97,11 @@ class BasicExpressionType : public ArithmeticExpressionType
     }
 };
 
+class IDefaultInitializable : public DeclRefType
```
csyonghe, line 100: `IDefaultInitializableType`

csyonghe: It doesn't seem like this is actually used by any front end logic? Can it be removed?

source/slang/core.meta.slang
```diff
@@ -74,6 +74,12 @@ syntax snorm : SNormModifier;
 ///
 syntax __extern_cpp : ExternCppModifier;
 
+__magic_type(IDefaultInitializable)
```
csyonghe, line 77: Can be done separately, but when we have `struct MyType : IDefaultInitializable { int f; }`
we should be able to automatically synthesize a default initializer to make it conform to the interface.

csyonghe: This can be done by extending the logic in `trySynthesizeConstructorRequirementWitness`.

csyonghe (approved):
I think this works fine, but I wonder if the following makes sense:

When `--zero-initialize` is specified, we make all user defined structs and interfaces conform to `IDefaultInitializable`.

This means that if I have a `struct MyType {}`, it implicitly becomes `struct MyType:IDefaultInitializable{}` and then the interface requirement synthesis logic will kick in to automatically generate that the default ctor.

If have a `interface IFoo`, we will automatiaclly make it `interface IFoo:IDefaultInitializable`, so that any `T:IFoo` will also be default initializable and we will invoke the default ctor  for variable definitions like `T obj;`.

ArielG-NV:
> I think this works fine, but I wonder if the following makes sense:
> 
> When `--zero-initialize` is specified, we make all user defined structs and interfaces conform to `IDefaultInitializable`.
> 
> This means that if I have a `struct MyType {}`, it implicitly becomes `struct MyType:IDefaultInitializable{}` and then the interface requirement synthesis logic will kick in to automatically generate that the default ctor.
> 
> If have a `interface IFoo`, we will automatiaclly make it `interface IFoo:IDefaultInitializable`, so that any `T:IFoo` will also be default initializable and we will invoke the default ctor for variable definitions like `T obj;`.
> 
> We probably want to exclude stdlib from this behaivor, though.

logic{
```
AttributeVisitor::visitStruct()
{
    if (zero init is toggled we do)
    {
       --> addInterface(IDefaultInitializable).
    }
}
```
//...
```
in SemanticsDeclBodyVisitor::varDeclCommon --> if(IDefaultInitializableType), set all members as we do currently. This will result in a implicit generation of defaultCtor.
```

}

csyonghe:
If we add `IDefaultInitializable` to `struct`s if they don't already (transitively) have them, then when we are at `ConformanceVisitor` we will try to check if the type is indeed conform to `IDefaultInitializable` ( in `SemanticsVisitor::checkConformance`).

Then inside `checkConformance`, we will try to find a default ctor that will satisfy the requirement in `IDefaultInitializable`.
If we find it, we are good, and we can add it to the witness table.
If not, we will try to synthesize it in `trySynthesizeConstructorRequirementWitness`. There we should extend the logic and implement the ctor synthesis.

>in SemanticsDeclBodyVisitor::varDeclCommon --> if(IDefaultInitializableType), set all members as we do currently. This will result in a implic ...

This logic is not necessary given the above flow.

ArielG-NV:
> If we add `IDefaultInitializable` to `struct`s if they don't already (transitively) have them, then when we are at `ConformanceVisitor` we will try to check if the type is indeed conform to `IDefaultInitializable` ( in `SemanticsVisitor::checkConformance`).
> 
> Then inside `checkConformance`, we will try to find a default ctor that will satisfy the requirement in `IDefaultInitializable`. If we find it, we are good, and we can add it to the witness table. If not, we will try to synthesize it in `trySynthesizeConstructorRequirementWitness`. There we should extend the logic and implement the ctor synthesis.
> 
> > in SemanticsDeclBodyVisitor::varDeclCommon --> if(IDefaultInitializableType), set all members as we do currently. This will result in a implic ...
> 
> This logic is not necessary given the above flow.

👍 Sounds good, I will go for an implementation around this.

csyonghe:
@ArielG-NV Is that a question, or a finding? It seems that the new design should handle the situation fine?

jkwak-work:
> While implementing I found an issue with inheritance and `IDefaultInitalizable` making sense to use.
> 
> If we have the scenario of inheritance where the parent does not have `IDefaultInitalizable` but the child has `IDefaultInitalizable`, calling the parent should not invoke `IDefaultInitalizable` logic. On the contrary, the child would be expected to exhibit behavior of `IDefaultInitalizable` for all inherited variables.

I don't why it is a problem if your description was to point out a problem. In Java or c++, when a child is XXX-able, it doesn't make their parent XXX-able.

ArielG-NV:
I think my original question about design was not worded well.

My question is more of a question on how we plan to handle inheritance on the initialization side of member variables.

1.
Do we follow existing C++ behavior:
```
struct myStruct_p 
{
     int t1[3];
};

struct myStruct : myStruct_p
{
     int t2[3];
     myStruct() = default;
};
int main()
{ 
    myStruct s = myStruct(); // all members are default init'd
    myStruct_p s2 = myStruct_p(); // members are uninitialized
    return 0;
}
```

2.
Or do we do something functionally equivalent to this [leave parents uninitialized]:
```
struct myStruct_p 
{
     int t1[3];
};

struct myStruct : myStruct_p
{
     int t2[3];
     myStruct() = default;
};
int main()
{ 
    myStruct s = myStruct(); // only myStruct members (not inherited) are default init'd
    myStruct_p s2 = myStruct_p(); // members are uninitialized
    return 0;
}
```

I ask for the specifics since reducing register pressure on a GPU is significantly more important than the cost of initialization on a CPU. These approaches would also require significantly different code changes, it would be better to be sure of `1.` or `2.` before implementing.

csyonghe:
If base is not default initializable, then we cannot automatically synthesize a default ctor. The same applies to normal members as well. If a struct has a member that is not default initializable, then we cannot automatically synthesize the default ctor for the type either.

csyonghe:
Marking the title as WIP, please remove it once it is ready to review again.

source/slang/slang-check-decl.cpp
```diff
@@ -1948,8 +1948,57 @@ namespace Slang
...
+        if (defaultCtor)
+        {
+            auto* invoke = visitor->getASTBuilder()->create<InvokeExpr>();
+            auto member = visitor->getASTBuilder()->getMemberDeclRef(declRefType->getDeclRef(), defaultCtor);
+            invoke->functionExpr = visitor->ConstructDeclRefExpr(member, NULL, defaultCtor->loc, nullptr);
+            return invoke;
+        }
+        else
+        {
+            auto* defaultCall = visitor->getASTBuilder()->create<VarExpr>();
```
csyonghe, line 1975: This will work, but it will be faster to make `__default` a syntax instead of a function call right now, so we can do `__default(T)` that maps to `DefaultValExpr`, which lowers to the DefaultConstruct opcode. This saves us from running lookup and overload resolution logic for every variable initialization.

source/slang/slang-check-decl.cpp
```diff
@@ -1948,8 +1948,57 @@ namespace Slang
         checkVisibility(classDecl);
     }
 
+    static InvokeExpr* _getDefaultInitOfVar(SemanticsVisitor* visitor, VarDeclBase* varDecl)
```
csyonghe, line 1951: =>  `constructDefaultInitExprForVar`

source/slang/slang-check-decl.cpp
```diff
@@ -4106,59 +4156,63 @@ namespace Slang
...
+                    auto stmt = m_astBuilder->create<ExpressionStmt>();
+                    stmt->expression = assign;
+                    seqStmt->stmts.add(stmt);
+                    break;
+                }
             }
         }
-
-        _addMethodWitness(witnessTable, requiredMemberDeclRef, makeDeclRef(ctorDecl));
+        if(!isDefaultInitializableType)
```
csyonghe, line 4214: Why do we need to guard this?

source/slang/slang-check-decl.cpp
```diff
@@ -5340,15 +5394,23 @@ namespace Slang
...
-                // Otherwise, report diagnostic now.
-                if (!requiredMemberDeclRef.getDecl()->hasModifier<BuiltinRequirementModifier>() &&
-                    !(requiredMemberDeclRef.as<GenericDecl>() &&
-                        getInner(requiredMemberDeclRef.as<GenericDecl>())->hasModifier<BuiltinRequirementModifier>()))
+            // If we failed to look up a member with the name of the
+            // requirement, it may be possible that we can still synthesis the
+            // implementation if this is one of the known builtin requirements.
+            // Otherwise, report diagnostic now.
+            if (!lookupResult.isValid() 
+                && !requiredMemberDeclRef.getDecl()->hasModifier<BuiltinRequirementModifier>()
```
csyonghe, line 5402: Should we make the default ctor requirement a BuiltinRequirement, like `IComparable::equals`?

source/slang/slang-check-decl.cpp
```diff
@@ -7512,6 +7591,21 @@ namespace Slang
         }
         DeclAndCtorInfo structDeclInfo = DeclAndCtorInfo(m_astBuilder, this, structDecl, false);
 
+        // ensure all varDecl members are processed up to SemanticsBodyVisitor so we can be sure that if init expressions 
+        // of members are to be synthisised, they are.
+        bool isDefaultInitializableType = this->getShared()->findInheritance<InterfaceDecl>(structDecl, m_astBuilder->getDefaultInitializableType());
```
csyonghe, line 7596: The logic here means that we will not consider the type as default initializable iff there is no `:IDefaultInitializable` declared on the type. `extension MyType : IDefaultInitializable` won't work. If this is intended we probably need the logic to prevent extending a type that is not default initializable with some interface  that is default initializable.

csyonghe: Are there any reasons why we can't allow `extension MyType : IDefaultIinitializable`?

source/slang/slang-check-decl.cpp
```diff
@@ -7512,6 +7591,21 @@ namespace Slang
...
+        for (auto m : structDecl->members)
+        {
+            auto varDeclBase = as<VarDeclBase>(m);
+            if (!varDeclBase)
+                continue;
+            ensureDecl(m->getDefaultDeclRef(), DeclCheckState::DefaultConstructorReadyForUse);
+            if (!isDefaultInitializableType
```
csyonghe, line 7603: If !isDefaultInitializableType, we shouldn't enter this loop at all?

csyonghe:
This is good, but I want to test this with some compile performance benchmarks.

source/slang/slang-check-decl.cpp
ArielG-NV: fixed 👍

source/slang/slang-check-decl.cpp
ArielG-NV: we still need `ensureDecl(m->getDefaultDeclRef(), DeclCheckState::DefaultConstructorReadyForUse);` loop to handle member variables and setting their default inits (a struct without `isDefaultInitializableType` for example). This was a problem with the code the testing did not check for.

There shouldn't be a cost doing ensureDecl regardless -- we just do it earlier.

source/slang/slang-check-decl.cpp
ArielG-NV: This is a bug with findInheritance. We should be checking for it. I will address this.

ArielG-NV:
note: did not address all comments yet (https://github.com/shader-slang/slang/pull/3987#discussion_r1605529224)

source/slang/slang-check-decl.cpp
csyonghe: In that case you can just call isSubtype or tryGetSubtypeWitness.

ArielG-NV: I reread the plan for `IDefaultIinitializable`, I originally misunderstood what was the plan
`isSubtype` works well.

source/slang/slang-check-decl.cpp
ArielG-NV: note: added `DefaultConstructExpr` (to follow naming of `kIROp_DefaultConstruct`), I did not change the `__intrinsic_op` out (in slang-parser.cpp) since I was unclear on if we are to swap out this use of `__default` as well.

csyonghe: Sure, that is fine.

source/slang/slang-check-conformance.cpp
```diff
@@ -51,19 +51,23 @@ namespace Slang
 
     SubtypeWitness* SemanticsVisitor::isSubtype(
         Type* subType,
-        Type* superType)
+        Type* superType,
+        bool subtypeInheritanceIsNotFullyResolved
```
csyonghe, line 55: I would like to understand why this additional parameter is needed. When this parameter is added, how does any incoming code know when to set it to true and when to set it to false?

Instead of introducing that question to every callsite, we should try to avoid requiring it.

It seems that the intent here is for some situations we don't want the cache the result. But in that case it seems the right approach is to invalidate the cache entry when an inheritance decl is checked. Will there be a bigger problem when that inheritance decl is resolved after this function is called?

source/slang/slang-check-decl.cpp
```diff
@@ -6074,11 +6074,9 @@ namespace Slang
 
             if (this->getOptionSet().getBoolOption(CompilerOptionName::ZeroInitialize))
             {
-                // Force add IDefaultInitializableType to any struct missing `IDefaultInitializableType`.
-                // This requires that we check for transitive relationships of all struct->interface, 
-                // struct->interface->interface..., relationship 
+                // Force add IDefaultInitializableType to any struct missing (transitively) `IDefaultInitializableType`.
                 auto* defaultInitializableType = m_astBuilder->getDefaultInitializableType();
-                if(isSubtype(DeclRefType::create(m_astBuilder, decl), defaultInitializableType))
+                if(isSubtype(DeclRefType::create(m_astBuilder, decl), defaultInitializableType), true)
```
csyonghe, line 6079: I am confused here. It seems that this code is trying to test if `this` is a sub type of `IDefaultIinitializable`, and if it is, and we set `-zero-initialize` flag, then we create an inheritance decl on the struct to make it conform to `IDefaultInitializable`.

Why is this needed? If a type transitively inherits `IDefaultInitializable`, then it should also be default initializable without a direct inheritance decl.

ArielG-NV: `                if(isSubtype(DeclRefType::create(m_astBuilder, decl), defaultInitializableType), true)` -->                 if(!isSubtype(DeclRefType::create(m_astBuilder, decl), defaultInitializableType), true)

source/slang/slang-check-conformance.cpp
ArielG-NV: what the param does:
1. no caching --> if we did not 'finish' the lookup stage and still are adding structs, we must not cache, we will have incorrect cache.
2. We do not 'ensure' the subtype decl, we assume it is safe to lookup. This is required for when we call the function at a time during/before the lookup stage. If we do not stop the ensureDecl, we will have a infinite recursive call.

source/slang/slang-check-decl.cpp
ArielG-NV: I will also add tests to hopefully not let this type of regression go unnoticed in the future

source/slang/slang-check-decl.cpp
```diff
@@ -4090,8 +4129,9 @@ namespace Slang
             }
         }
 
+        bool isDefaultInitializableType = requiredMemberDeclRef.getParent() == getASTBuilder()->getDefaultInitializableTypeInterfaceDecl();
         bool isInWrapperType = isWrapperTypeDecl(context->parentDecl);
-        if (!isInWrapperType)
+        if (!isInWrapperType && !isDefaultInitializableType)
```
jkwak-work, line 4134: Can we have a test case for the wrapper-type specifically for this change?
I am not familiar with the wrapper-type and I am not sure if this change will cause any trouble.

source/slang/slang-check-decl.cpp
```diff
@@ -4090,8 +4129,9 @@ namespace Slang
...
         {
             return false;
```
jkwak-work, line 4136: Before your change, when `isInWrapperType` was true, this function was returning `false`.
With your change, this is changed to return `true`.
Is this change intentional? or should it still return `false`?

source/slang/core.meta.slang
```diff
@@ -480,6 +475,16 @@ ${{{{
...
 {
+
+    typedef $(kBaseTypes[tt].name) ThisType;
+
+    // Declare initializer of `This` type for default initialization
+    [ForceInline]
+    __init()
+    { 
+        this = __default<ThisType>();
+    }
```
jkwak-work, line 486: I see that cpp files checks the option like this,
````
this->getOptionSet().getBoolOption(CompilerOptionName::ZeroInitialize)
```
and change the behavior based on the command-line argument.
But this line doesn't check the option and it will always initialize the variables with zero, isn't it?
If it is the case, I am worried that Slang generated code will pay the cost of initialization even when they shouldn't be initialized.

If my understanding is incorrect, please let me know how it works.

source/slang/slang-check-impl.h
```diff
@@ -2057,9 +2057,14 @@ namespace Slang
             ///
         SubtypeWitness* isSubtype(
             Type*                   subType,
-            Type*                   superType);
+            Type*                   superType,
+            bool subtypeInheritanceIsNotFullyResolved = false //resolving members may not be required in `SemanticsDeclBasesVisitor`
```
jkwak-work, line 2061: Please avoid unnecessary default function values.
Also move the comment to the doxygen comments placed above the function name.

source/slang/slang-check-conformance.cpp
jkwak-work: I like to suggest to refactor the new boolean variable.
It appears that the behavior around the new boolean variable is that when it is true, "isSubtype()" behaves same as before.
Only when the boolean variable is false, "isSubtype()" behaves little differently; I am not clear on how different tho.

If my observation is correct, you could make a new function like following without changing the existing function, "isSubtype()", 
```
SubtypeWitness * XXX:isSubtypeWhenInheritanceIsNotFullyResolved(Type* subType, Type* superType)
{
    auto result = isSubtype(subType, superType);
    if (result == null_ptr)
    {
       do something here...
    }
    return result;
}
```
The reason why I suggest this is that "boolean" parameter is not a good thing to have in general.
It can be only true or false, which can be same as having two distinct functions instead of having a single function with a boolean parameter.
For more information, I like to suggest a very good book called, "Clean code", written by Martin C Robert.
His another book, "Agile Software Development, Principles, Patterns, and Practices", is also a fantastic book to checkout.

source/slang/slang-check-decl.cpp
```diff
@@ -7566,8 +7649,11 @@ namespace Slang
             for (auto& m : structDecl->members)
             {
                 auto varDeclBase = as<VarDeclBase>(m);
+
+                // Static variables are initialized at start of runtime, not inside a constructor
                 if (!varDeclBase
-                    || !varDeclBase->initExpr)
+                    || !varDeclBase->initExpr
+                    || varDeclBase->hasModifier<HLSLStaticModifier>())
```
jkwak-work, line 7656: It looks like this covers static member variables.
But I am not seeing a code that covers static variables declared inside of functions like,
```
void MyFunc()
{
    static int i;
}
```
Is it covered elsewhere? If not, you may need to.

source/slang/slang-options.cpp
```diff
@@ -289,6 +289,10 @@ void initCommandOptions(CommandOptions& options)
         { OptionKind::Language,     "-lang", "-lang <language>", "Set the language for the following input files."},
         { OptionKind::MatrixLayoutColumn, "-matrix-layout-column-major", nullptr, "Set the default matrix layout to column-major."},
         { OptionKind::MatrixLayoutRow,"-matrix-layout-row-major", nullptr, "Set the default matrix layout to row-major."},
+        { OptionKind::ZeroInitialize, "-zero-initialize", nullptr, 
+        "Set all variables to zero."
+        "Structs will set all struct-fields without an init expression to 0."
+        "All variables will call their default constructor if not explicitly initialized as usual."},
```
jkwak-work, line 295:
```
Set all variables to zero
```
is not the correct description for the implementation, is it?
It will be better with "Initialize all variables to zero".

source/slang/slang-check-decl.cpp
```diff
@@ -1948,8 +1948,47 @@ namespace Slang
         checkVisibility(classDecl);
     }
 
+    static Expr* constructDefaultInitExprForVar(SemanticsVisitor* visitor, VarDeclBase* varDecl)
```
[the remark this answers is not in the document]
ArielG-NV: I believe the following name was suggested instead of `maybe...` because if this function doesn't construct it is due to invalid types to avoid a language server crash.
We generally do not add `maybe` if a language server crash is being avoided.

source/slang/core.meta.slang
ArielG-NV: stdlib structs do not set init expressions if they have an empty `initExpr`.
This is due to the following code having `!isFromStdLib` to ensure we don't default init `(.*).meta.slang` types unless `CompilerOptionName::ZeroInitialize` is enabled.

`SemanticsDeclBodyVisitor::checkVarDeclCommon`
`else if(!varDeclBaseType || !varDeclBaseType->getDeclRef() || !isFromStdLib(varDeclBaseType->getDeclRef().getDecl()))`

I don't think this affects raytracing types, but it seems it does not (falcor works)

ArielG-NV:
> I like to suggest to refactor the new boolean variable.

Likely best for future refactor, agreed.

source/slang/slang-check-decl.cpp
ArielG-NV: The logic we want is:
1. **do not** `return false` if (`isInWrapperType == true` || `isDefaultInitializableType` == true)
alternatively,
2. `return false` if ( !(`isInWrapperType == true` || `isDefaultInitializableType` == true) )

following De'morgons law:
!(x || y) == (!x && !y)

source/slang/slang-check-decl.cpp
ArielG-NV: > // Used if this type declaration is a wrapper, i.e. struct FooWrapper:IFoo = Foo;

^ what we define a wrapperType as in the code

I can add a test case

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
jkwak-work: I am not sure what you meant here. Why is function name related to a crash. What is language server?

source/slang/slang-check-decl.cpp
jkwak-work: Sorry I had a typo. I was talking about the case where `isInWrapperTypr` is false. It was always returning false regardless other values. Now with your change, it has a chance to return true. It is fine if it is an expected behavior.

source/slang/core.meta.slang
jkwak-work: I don't think I am following your explanation but it raised another question.
When slangc runs for the first time, it compiles stdlib and it is cached. And next time we run slangc, we don't recompile stdlib. 

From this point, the run time compiler option will have no impact on how stdlib gets compiled, because the compilation doesn't happen anymore.

Is the implementation consider this case?

jkwak-work: The part I don't understand is that you said, 
```
stdlib structs do not set init expressions if they have an empty initExpr.
```
But it looks like you are setting initExpr here.
```
this = __default<ThisType>();
```
I think "__default<ThisType>()" is initExpr, isn't it?

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
ArielG-NV: 1. I will add `maybe...` for simplicity. If someone else disagrees it will be an easy change

2. To answer the other questions

> What is language server?

Having no type is an error with Slang. Slangc will not lead to a position like this.

Slangc though is not the only way to use Slang, Slang has a language server. This is how the "Slang vscode extension" works.

The language server does not mirror how a regular compile works. Due to this, Slang code needs to be robust enough to avoid a crash if invalid data is given to a function. This is possible because Slang's language server works on incomplete, likely invalid input. A crash is an issue because then the code will not be able to send the 'errors' to the calling application.

> Why is function name related to a crash

All code to avoid a language server failure are not addressed in function name from what I have seen.

source/slang/slang-check-decl.cpp
ArielG-NV: That is expected behavior.
There is a second `if` statment using `isInWrapper` to handle the optional unique code of `isInWrapper`'s later on in the function.

```
        if (isInWrapperType)
        {
            SemanticsDeclBodyVisitor bodyVisitor(withParentFunc(ctorDecl));
```

tests/language-feature/zero-initialize/IDefaultExplicit-wrapper-type.slang
```diff
@@ -0,0 +1,52 @@
...
+
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -emit-spirv-directly -allow-glsl
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-cpu -compute -entry computeMain -allow-glsl
+//TEST(smoke,compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-dx12 -use-dxil -compute -entry computeMain -allow-glsl -profile sm_6_2 -xslang -DDX12
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+struct base : IDefaultInitializableType
```
jkwak-work, line 13: Can you rename with a capital letter 'B' for types?
```
Base
```

source/slang/core.meta.slang
ArielG-NV: > Does the implementation consider this case?

This has 2 parts:
1. does IDefaultInitialize get added to meta.slang

stdlib will not automatically add `IDefaultInitialize` if compiled without.
This cannot be worked around since a user does not have the ability to compile (.).meta.slang

2. does it still work as expected

Yes. To simplify the answer I will use vec2 as an example: if we write `vec2 myVar` with the `-zero-initialize flag`, the following code will compile `vec2 myVar = vec2()`.

ArielG-NV: `int a = __default<ThisType>();` => `int a = 0;`
`float a = __default<ThisType>();` => `float a = 0.0f;`

jkwak-work: So it will initialized to zero even when `-zero-initialize` is not set, is it the case?

ArielG-NV: > So it will initialized to zero even when -zero-initialize is not set, is it the case?

no, the init is not automatically called. It just exists for us to use.
Image if I make an `interface thing{ __init(); }` --> this interface would not work with `int`, which is incorrect.

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
jkwak-work: I have never heard the term, "language server".
Do we have a better description or document about it somewhere?

ArielG-NV:
Update: jkwak-work found a flaw with how std types were treated. This is being addressed.

source/slang/core.meta.slang
jkwak-work: After talking on Slack for a bit, it turned out that @ArielG-NV  was on an impression that we shouldn't be calling the constructors if the struc/class is defined in stdlib.

He thought that it is our design choice because he misinterpreted what Yong said on [one of comments](https://github.com/shader-slang/slang/pull/3987#pullrequestreview-2032213162),
> We probably want to exclude stdlib from this behaivor, though.

I don't believe that is what we want at all but I like to hear from @csyonghe if my understanding is correct.

jkwak-work:
> Update: jkwak-work found a flaw with how std types were treated. This is being addressed.

Can you push the change so that I can see how it is changed?

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
ArielG-NV: MSVC has a good description:
1. https://learn.microsoft.com/en-us/visualstudio/extensibility/language-server-protocol?view=vs-2022

JetBrains has a simpler but more lacking explanation:
1. https://blog.jetbrains.com/platform/2023/07/lsp-for-plugin-developers/

ArielG-NV:
> > Update: jkwak-work found a flaw with how std types were treated. This is being addressed.
> 
> Can you push the change so that I can see how it is changed?

pushed the changes

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
jkwak-work: When you said about the language server, you were talking about a crash with slang extension for visual studio?

When you said, "Slang has a language server", i got an impression that we implemented something called a language server. But it seems like you meant to say, "slang uses a language server," is it the case?

source/slang/slang-check-decl.cpp
```diff
@@ -4106,59 +4142,67 @@ namespace Slang
...
             }
         }
-
+        if (isDefaultInitializableType)
+            context->parentDecl->addMember(ctorDecl);
+        else
+        
         _addMethodWitness(witnessTable, requiredMemberDeclRef, makeDeclRef(ctorDecl));
+        
         return true;
```
jkwak-work, line 4206: Indentation seems off.
And it is better not to have an empty line after "else" because it can cause mistakes.

jkwak-work:
When I generated GLSL of the test, I see that "data1" is incorrectly unused with your code.
It seems like there is a bug in your implementation.

ArielG-NV:
> With this change, if I don't use "-zero-initialize" option, the results are expected to be same as before. And I tested it locally and saw that one of tests we have shows difference.
> 
> ```
> tests/language-feature/struct-field-initializers/struct-field-initializer-static.slang
> ```
> 
> Can you see why the result is changed with your code, and make sure everything works as expected?
> 
> The test is not failing but SPIR-V code is different when I compared the outcome with the following command.
> 
> ```
> slangc -target spirv-asm -o out.spirv tests/language-feature/struct-field-initializers/struct-field-initializer-static.slang
> ```

Took a bit to check because I wanted to get a clean master first*

There are differences. These of which I can explain, I will share the shaders for ease of reading:

```glsl
### --> zero-init branch
#version 450
layout(row_major) uniform;
layout(row_major) buffer;

#line 7 0
layout(std430, binding = 0) buffer StructuredBuffer_int_t_0 {
    int _data[];
} outputBuffer_0;

#line 9
struct DefaultStructNoInit_base_0
{
    int data2_0;
};


#line 9
DefaultStructNoInit_base_0 DefaultStructNoInit_base_x24init_0()
{

#line 9
    DefaultStructNoInit_base_0 _S1;



    _S1.data2_0 = 2;

#line 9
    return _S1;
}


#line 16
struct DefaultStructNoInit_0
{
    DefaultStructNoInit_base_0 base_0;
    int data3_0;
};


#line 16
DefaultStructNoInit_0 DefaultStructNoInit_x24init_0()
{

#line 16
    DefaultStructNoInit_0 _S2;

#line 16
    _S2.base_0 = DefaultStructNoInit_base_x24init_0();

    _S2.data3_0 = 2;

#line 16
    return _S2;
}


#line 12
int data1_0;


#line 22
void main()
{

#line 22
    data1_0 = 2;

    DefaultStructNoInit_0 noInit_0 = DefaultStructNoInit_x24init_0();

#line 24
    bool _S3;

#line 30
    if(noInit_0.base_0.data2_0 == 2)
    {

#line 30
        _S3 = noInit_0.data3_0 == 2;

#line 30
    }
    else
    {

#line 30
        _S3 = false;

#line 30
    }

#line 26
    outputBuffer_0._data[0U] = int(_S3);

#line 32
    return;
}

```

------------------------

```glsl
### --> master
#version 450
layout(row_major) uniform;
layout(row_major) buffer;

#line 7 0
layout(std430, binding = 0) buffer StructuredBuffer_int_t_0 {
    int _data[];
} outputBuffer_0;


int data1_0;


#line 9
struct DefaultStructNoInit_base_0
{
    int data2_0;
};


#line 9
DefaultStructNoInit_base_0 DefaultStructNoInit_base_x24init_0()
{

    data1_0 = 2;

#line 9
    DefaultStructNoInit_base_0 _S1;



    _S1.data2_0 = 2;

#line 9
    return _S1;
}


#line 16
struct DefaultStructNoInit_0
{
    DefaultStructNoInit_base_0 base_0;
    int data3_0;
};


#line 16
DefaultStructNoInit_0 DefaultStructNoInit_x24init_0()
{

#line 16
    DefaultStructNoInit_0 _S2;

#line 16
    DefaultStructNoInit_base_0 _S3 = DefaultStructNoInit_base_x24init_0();

#line 16
    _S2.base_0 = _S3;

    _S2.data3_0 = 2;

#line 16
    return _S2;
}




void main()
{

#line 22
    data1_0 = 2;

    DefaultStructNoInit_0 noInit_0 = DefaultStructNoInit_x24init_0();

#line 24
    bool _S4;

#line 29
    if(data1_0 == 2)
    {

#line 29
        _S4 = noInit_0.base_0.data2_0 == 2;

#line 29
    }
    else
    {

#line 29
        _S4 = false;

#line 29
    }
    if(_S4)
    {

#line 30
        _S4 = noInit_0.data3_0 == 2;

#line 30
    }
    else
    {

#line 30
        _S4 = false;

#line 30
    }

#line 26
    outputBuffer_0._data[0U] = int(_S4);

#line 32
    return;
}
```

ArielG-NV:
differences; 
1.   master:  `static int data1 = 2;` init's in the `__init()` of a function and at entrypoint. Now, only entry point sets the variable
2. data1 is now places lower in program code (since referenced first at a later date)

ArielG-NV:
why this happens:
since we run `ensureDecl(struct.member, DeclCheckState::DefaultConstructorReadyForUse)`, now the code sets up all modifiers+checks+inits correctly before we run bodyVisitor->visitAggTypeDecl()

ArielG-NV:
> When I generated GLSL of the test, I see that "data1" is incorrectly unused with your code. It seems like there is a bug in your implementation.

This is incorrect, master branch does things incorrectly.

source/slang/slang-check-decl.cpp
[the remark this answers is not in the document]
jkwak-work: Interesting...
How do I test whether or not my change didn't break the language server?
Does slang-test have a test coverage on that?
Or do I need to install something more and run other tests?

jkwak-work (approved):

source/slang/slang-check-decl.cpp
```diff
@@ -1948,8 +1948,46 @@ namespace Slang
...
+            {
+                defaultCtor = _getDefaultCtor(structDecl);
+            }
+        }
+
+        if (defaultCtor)
+        {
+            auto* invoke = visitor->getASTBuilder()->create<InvokeExpr>();
+            auto member = visitor->getASTBuilder()->getMemberDeclRef(declRefType->getDeclRef(), defaultCtor);
+            invoke->functionExpr = visitor->ConstructDeclRefExpr(member, NULL, defaultCtor->loc, nullptr);
```
csyonghe, line 1970: NULL->nullptr

csyonghe (dismissed):

csyonghe (approved):

Issue #4001, opened 2024-04-22
Improve direct spirv backend test coverage.


csyonghe:
These tests are currently running for glsl path but not direct to spirv:
```
tests\hlsl-intrinsic\ray-tracing\rt-pipeline-intrinsics-ahit.slang
tests/hlsl-intrinsic/ray-tracing/rt-pipeline-intrinsics-rgen.slang
tests\gpu-feature\texture\query\footprint\nv-shader-texture-footprint.slang
tests/gpu-feature/texture/query/footprint/nv-shader-texture-footprint-simple.slang
tests/vkray/callable.slang
tests/vkray/anyhit.slang
```

PR #4002, opened 2024-04-22
Switch to direct-to-spirv backend as default.

### LINKED ISSUE #3477: Make OptionKind::EmitSpirvDirectly turn ON by default for target CodeGenTarget target SPIRV
Blocked by:
- #3547
- Fix the perf issue exposed by #3547.
- #3440
- #3438 
- #3437

csyonghe:
The change is mostly about making all existing tests to pass.

Identified a couple tests where we were missing the coverage for direct to spirv backend.
They are listed in issue #4001.

Closes #3477.

Another big fix included in this PR is that our spirv emit logic for texture load/sample/store operations where the element type is `half` or `vector<half>` is incorrect. SPIRV does not allow `half` to be used as texture element type and we can only define it as `float` and convert them into `half`.


ArielG-NV:
should we be adding to all tests such as `rt-pipeline-intrinsics-rgen.slang` a direct to spirv test instead of the should-be-avoided spirvViaGLSL path?

csyonghe:
I am tracking all that in a separate issue here: https://github.com/shader-slang/slang/issues/4001

For this PR, the goal is not to lose test coverage.

ArielG-NV:
PR looks good to me then.

tools/slang-test/options.cpp
```diff
@@ -290,9 +290,9 @@ static bool _isSubCommand(const char* arg)
         {
             optionsOut->skipApiDetection = true;
         }
-        else if (strcmp(arg, "-emit-spirv-directly") == 0)
+        else if (strcmp(arg, "-emit-spirv-via-glsl") == 0)
         {
-            optionsOut->emitSPIRVDirectly = true;
+            optionsOut->emitSPIRVDirectly = false;
         }
```
jkwak-work, line 296: Shouldn't we still support "-emit-spirv-directly"?
"tests/compute/half_texture.slang", as an example, still uses the option.

jkwak-work (approved):
Looks good to me.

tools/slang-test/options.cpp
csyonghe: We do, it just no longer has an effect because emit spirv directly is the default.

PR #4004, opened 2024-04-22
Disable DXC warnings for missing bitwise op parantheses

### LINKED ISSUE #3980: Snooze DXC warning "-Wbitwise-op-parentheses"
**Problem description**
When compiled UE TSR shaders, all DXC warnings were treated as errors by default. As a workaround, the warnings from DXC were ignored.

The first one is about the precedence of the bitwise operators.
```
/Engine/Public/WaveBroadcastIntrinsics.ush:90:42: warning: '&' within '|' [-Wbitwise-op-parentheses]
                return ((DestLaneIndex_0 & Settings_1.SwizzleAnd_0 | Settings_1.SwizzleOr_0) ^ Settings_1.SwizzleXor_0) % 32U;
                         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~ ~
/Engine/Public/WaveBroadcastIntrinsics.ush:90:42: note: place parentheses around the '&' expression to silence this warning
                return ((DestLaneIndex_0 & Settings_1.SwizzleAnd_0 | Settings_1.SwizzleOr_0) ^ Settings_1.SwizzleXor_0) % 32U;
                                         ^
                         (                                        )
```

**Goal**
It will be nice to silence certain warnings selectively.

**Repro steps**
This can be reproduced with a following Slang shader,
```
uint BitwiseParentheses(uint rotate, uint mask, uint index)
{
    return (mask & index) | ((~mask) & (index + rotate));
}
```
It gets translated to HLSL as following,
```
uint BitwiseParentheses_0(uint rotate_0, uint mask_0, uint index_0)
{
   return mask_0 & index_0 | ~mask_0 & index_0 + rotate_0;
}
```
Note that all of parentheses are removed.
According to the C operator precedence, it is still correct but DXC prints warnings.
```
gh.hlsl:10:9: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma warning(disable: 3557)
        ^
gh.slang:16:19: warning: '&' within '|' [-Wbitwise-op-parentheses]
    return mask_0 & index_0 | ~mask_0 & index_0 + rotate_0;
           ~~~~~~~^~~~~~~~~ ~
gh.slang:16:19: note: place parentheses around the '&' expression to silence this warning
    return mask_0 & index_0 | ~mask_0 & index_0 + rotate_0;
                  ^
           (               )
gh.slang:16:39: warning: '&' within '|' [-Wbitwise-op-parentheses]
    return mask_0 & index_0 | ~mask_0 & index_0 + rotate_0;
                            ~ ~~~~~~~~^~~~~~~~~~~~~~~~~~~~
gh.slang:16:39: note: place parentheses around the '&' expression to silence this warning
    return mask_0 & index_0 | ~mask_0 & index_0 + rotate_0;
                                      ^
                              (                           )
```

**Proposed solution**
An easy solution is to add the following [pragma for DXC](https://github.com/tex3d/DirectXShaderCompiler/blob/main/tools/clang/test/HLSLFileCheck/hlsl/preprocessor/pragma/pragma-diagnostic.hlsl) in slang-hlsl-prelude.h
```
#ifdef __DXC_VERSION_MAJOR
    // warning: '&' within '|' [-Wbitwise-op-parentheses]
    #pragma dxc diagnostic ignored "-Wbitwise-op-parentheses"
#endif
```
We will need to have slang-test for checking the behavior around the bitwise operators.
The parentheses shouldn't be removed if they are needed.

jkwak-work:
Resolves #3980

Based on the operator precedence, Slang may omits the parentheses if they are not needed. DXC prints warnings for such cases and some applications may treat the warnings as errors.

This commit disables such warnings and adds more through tests to prevent any regressions.


tests/bugs/gh-3980.slang
```diff
@@ -0,0 +1,224 @@
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-slang -compute -output-using-type
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-slang -compute -output-using-type -dx12 -profile cs_6_6 -use-dxil
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -shaderobj -output-using-type -emit-spirv-directly
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-cpu -compute -output-using-type
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-cuda -compute -output-using-type
+
+// Slang removes parentheses characters for the bitwise operators when they are not needed.
+// DXC pritns warning messages even when the expression is correct.
```
ArielG-NV, line 9:
```suggestion
// DXC prints warning messages even when the expression is correct.
```

prelude/slang-hlsl-prelude.h
```diff
@@ -1,4 +1,13 @@
...
 #endif
-#pragma warning(disable: 3557)
+
+#ifdef __DXC_VERSION_MAJOR
+    // warning: '&' within '|' [-Wbitwise-op-parentheses]
+    #pragma dxc diagnostic ignored "-Wbitwise-op-parentheses"
+    // warning: operator '>>' has lower precedence than '+'; '+' will be evaluated first [-Wshift-op-parentheses]
+    #pragma dxc diagnostic ignored "-Wshift-op-parentheses"
+#else
+    // warning X3557: loop doesn't seem to do anything, forcing loop to unroll
```
ArielG-NV, line 11: not 100% related to this issue, but worth bringing up:

I wonder if we should add a similar warning in the front end:  "The loop only executes for a limited number of iterations or doesn't seem to do anything so consider removing it or forcing it to unroll."

and we also add a second warning: "unrolling may reduce performance in large shaders due to register pressure" if a loop is large and running for many iterations.

ArielG-NV (approved):
1 question, 1 minor suggestion.

Otherwise, code looks good.

tgrimesnv (dismissed):
Looks good to me.

prelude/slang-hlsl-prelude.h
csyonghe: Instead of silencing the warning, we should consider changing our emit logic to always emit a parenthesis if the operator is &, | or << >>.

csyonghe (dismissed):

prelude/slang-hlsl-prelude.h
jkwak-work: Interesting. I thought we were intentionally not emitting the parentheses. I will see if I can emit them.

jkwak-work: > not 100% related to this pull request, but worth bringing up since we are silencing warnings:
> 
> I wonder if we should add a similar warning in the front end: "The loop only executes for a limited number of iterations or doesn't seem to do anything so consider removing it or forcing it to unroll."
> 
> and we also add a second warning: "unrolling may reduce performance in large shaders due to register pressure" if a loop is large and running for many iterations.

I believe we have the warning already; or similar

jkwak-work:
I need a help from @csyonghe regarding a function "rightSide()".
When I look at a function, "CLikeSourceEmitter::defaultEmitInstExpr()", I see a pattern like this, which makes sense.
```
        needClose = maybeEmitParens(outerPrec, info);
        emitOperand(inst->getOperand(0), leftSide(outerPrec, info));    
        m_writer->emit(" ");
        m_writer->emit(info.op);
        m_writer->emit(" ");                                                                  
        emitOperand(inst->getOperand(1), rightSide(outerPrec, info));   // <== "rightSide" is confusing.
```
And when I look at the function signature of "rightSide()", it looks like this,
```
SLANG_INLINE EmitOpInfo rightSide(EmitOpInfo const& prec, EmitOpInfo const& outerPrec)
```
Note that the first parameter of "rightSide()" is "prec" not "outerPrec".
But when it was called, "outerPrec" is used as a first argument.
When I searched for "rightSide()", two different ways are used half and half.

I am not sure if the arguments are intentionally swapped or it is a bug.

This is giving me a trouble on how things work.
When the case is `a | b & c | d` and it is processing `&`, I expect to get
```
outerPrec.leftPrecedence = bitOr_right
outerPrec.rightPrecedence = bitOr_left
```
But I am getting the following
```
outerPrec.leftPrecedence = bitOr_left
outerPrec.rightPrecedence = bitOr_right
```

jkwak-work:
While waiting for an answer from Yong, I am pushing my new change that has a "fix" for "rightSide()".
When I tried locally, all tests are still passing.
I am curious if it would cause a trouble on CI.

csyonghe:
The whole thing is confusing and people maybe commiting wrong code all the time without realizing or being captured by any tests.

The `rightSide()` call has outerPrec and prec reversed compared to the `leftSide()` call, and the callsites should respect this order.

We should do a pass through the use of `leftSide` and `rightSide` and to make sure to add more tests to ensure correctness.

csyonghe:
In your example of `a | b & c | d`, when we are processing `&` and is about to emit `a|b`, because `a|b` appears at the left side of `&`, the new outer prec to call `emitOperand` for `a|b` will be `leftSide(outerPrecOfBitAnd, precOfBitAnd)`. This means that it will be
{
    leftPrec = outPrecOfBitAnd.leftPrec
    rightPrec = prefOfBitAnd.leftPrec
}
Which is what you want, because the `outPrecOfBitAnd.leftPrec` describes the precedence to the left end of the entire `a | b & c | d` expr, which is indeed what is to the left end of `a|b`, and the right end of the expr `a|b`  has the precedence `prefOfBitAnd.leftPrec` which is the precedence of the left side of `&`.

source/slang/slang-emit-c-like.cpp
```diff
@@ -2289,7 +2318,7 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
             }
         }
 
-        emitOperand(operand, rightSide(prec, outerPrec));
+        emitOperand(operand, rightSide(outerPrec, prec));
```
csyonghe, line 2321: I think `rightSide` should be called with `rightSide(prec, outerPrec)`.

jkwak-work:
It is very confusing but an easiest way to think about the bug is following.

When the function signature is like this,
```
EmitOpInfo rightSide(EmitOpInfo const& prec, EmitOpInfo const& outerPrec);
```
Calling the function as following seems wrong, because "outerPrec" is not used as "outerPrec".
```
rightSide(outerPrec, info); // outerPrec is used as prec and info is used as outerPrec.
```
To avoid the future mistakes, I like to suggest to change the function signature.
```
EmitOpInfo leftSide(EmitOpInfo const& outerPrec, EmitOpInfo const& prec); // UNCHANGED
EmitOpInfo rightSide(EmitOpInfo const& outerPrec, EmitOpInfo const& prec); // PARAMETERS ARE SWITCHED
```
And when we call `rightSide()`, we should call as `rightSide(outerPrec, info)`.

jkwak-work:
Here is my understanding for the example, "a | b & c | d".
If I draw a binary tree of AST, it will looks like this,
```
       "|"
     /     \
   "|"    "d"
  /   \ 
"a"   "&"
     /    \
   "b"   "c"
```
The processing order is "b&c", "a|(b&c)" and "(a|(b&c))|d".

What I like to emit is "a | **(** b & c ) | d".
When deciding the opening parentheses before "b", `needClose = maybeEmitParens(outerPrec, info);` should be true.
`maybeEmitParens()` does the following,
```
    bool needParens = (prec.leftPrecedence <= outerPrec.leftPrecedence)
        || (prec.rightPrecedence <= outerPrec.rightPrecedence);
```
For "(b&c)", it is comparing the precedence of "b" from the current operator perspective and the outer operator perspective,
```
(prec.leftPrecedence <= outerPrec.leftPrecedence)
```
And it also compares the precedence of "c" from the current operator perspective and the outer operator perspective,
```
(prec.rightPrecedence <= outerPrec.rightPrecedence);
```
If either of them is true, it will emit the parentheses.

The following should be true in order to make it working.
```
outerPrec.leftPrecedence = BitOr_Right // which is for "b" from "a|b" perspective.
prec.leftPrecedence = BitAnd_Left // which is for "b" from "b&c" perspective.
prec.rightPrecedence = BitAnd_Right // which is for "c" from "b&c".
outerPrec.rightPrecedence = BitOr_Left // which is for "c" from "c|d" perspective.
```
However I am getting the following,
```
outerPrec.leftPrecedence = BitOr_Left // which is for "a" from "a|b" perspective. INCORRECT
prec.leftPrecedence = BitAnd_Left // which is for "b" from "b&c" perspective.
prec.rightPrecedence = BitAnd_Right // which is for "c" from "b&c".
outerPrec.rightPrecedence = BitOr_Right // which is for "d" from "c|d" perspective. INCORRECT
```
When I debugged and traced back, the incorrect values are coming from "rightSide" because the order of arguments are wrong.

I believe my fix is a proper way to resolve the mistakes.

csyonghe:
Yes, I didn't realize you also swapped the order of the parameters in rightSide. Your changes are correct.

And my previous example is completely wrong, it is assuming the expr is (a|b)&(c|d).

The explanation is correct if the example is changed to (a&b|c&d). Sorry about it, my brain is in a total mess.

csyonghe (approved):
Looks good to me, and thank you for adding the test!

jkwak-work:
MacOS Build CI is failing due to a missing test script.
I am going to ignore it and merge the change because I don't think it is related to my change.

PR #4005, opened 2024-04-23
WIP: Force Inline If RefType

### LINKED ISSUE #3997: Spirv validation and __ref with atomics
Hi,

I am trying to implement atomics for the C++ backend, while wrapping the ones for spirv using this code:

```c++
float atomicAdd(__ref float value, float amount)
{
    __target_switch
    {
    case cpp:
        __requirePrelude("#include <atomic>");
        __intrinsic_asm "std::atomic_ref(*$0).fetch_add($1)";
    case spirv:
        return __atomicAdd(value, amount);
    }
}

[numthreads(4, 1, 1)]
[shader("compute")]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID) {
    atomicAdd(outputBuffer[0], 1);
}
```
The code seems to work "somewhat" but it fails spirv validation (1.3.280.0) when compiling it with `-g` and `-O0` (using slangc 2024.1.8 and `slangc.exe -target spirv -emit-spirv-directly -g -O0` as parameters):
```
error: line 95: OpFunctionCall Argument <id> '49[%49]'s type does not match Function <id> '56[%_ptr_Function_float]'s parameter type.
  %54 = OpFunctionCall %float %atomicAdd %49 %float_1
```

Not sure if I am doing something wrong, if slang generates bad spirv or if spirv val is wrong. I still find it interesting that the problem is "resolved" when omitting `-g` or using at least `-O1`.

Can you have a look into this?

ArielG-NV:
Adjusting the type based inlining pass of performStringInlining to be a more generic TypeInlining.

Fixes #3997


csyonghe (dismissed):

source/slang/slang-ir-inline.h
```diff
@@ -11,7 +11,7 @@ namespace Slang
     class DiagnosticSink;
 
         /// Any call to a function that takes or returns a string parameter is inlined
-    Result performStringInlining(IRModule* module, DiagnosticSink* sink);
+    Result performTypeInlining(IRModule* module, DiagnosticSink* sink);
```
jkwak-work, line 14: Maybe you can update this comment to reflect that it works also for ref type now?

jkwak-work (commented):
Please leave an issue number on one of your commit messages next time.
It will help us to track where we can find more details later.

source/slang/hlsl.meta.slang
```diff
@@ -11580,9 +11580,12 @@ void SetMeshOutputCounts(uint vertexCount, uint primitiveCount)
 //
 // This function doesn't return.
 //
+// This function cannot be inlined due to a legalization pass happening mid-way through processing
+// and later more processing happening to the function which requires eventual inlining.
 [KnownBuiltin("DispatchMesh")]
 [require(glsl_hlsl_spirv, meshshading)]
-void DispatchMesh<P>(uint threadGroupCountX, uint threadGroupCountY, uint threadGroupCountZ, __constref P meshPayload)
+[noRefInline]
```
csyonghe, line 11587: This feels like a really hacky way to get things working, particularly when this is the only case that we are trying to workaround here. The glsl legalization logic probably needs a revisit so that it is more conformant with how rest of the compiler is written. We should adopt something similar to your ray payload technique here to replace `meshPayload` with a global index.

csyonghe (approved):

source/slang/hlsl.meta.slang
ArielG-NV: do we merge this pr, make a seperate issue, and redesign it later?

csyonghe: Yes, let's merge the pr first.

PR #4009, opened 2024-04-23
Force Inline all stdlib functions

sriramm-nv:
Fixes #4003


csyonghe:
I don't think we need to ForceInline all stdlib functions, only need to do so for the functions that can take resource parameters.

jkwak-work:
The description says it fixes 4003 but 4003 is not an issue but PR.
I cannot find any information of what this change is trying to achieve.

csyonghe:
The original intend here is that if the user has:
`resourceArray[NonUniformResourceIndex(index)].someMethodOnTheResource();`
Then `someMethodOnTheResource` must be [ForceInline]d for the non uniform resource pass to work for the spirv backend. If we see any resource memeber methods that are not forceinlined, we should make them so.

We should not be changing every stdlib function.

csyonghe:
I am going to close this PR now, because I think there is a result of me not clearly communicating what is required.

Issue #3999, opened 2024-04-22
Don't diagnose an error when a symbols is defined as `extern` and `export`.


csyonghe:
Right now we diagnose an error for the following code:
```
extern int test() { return 0; }

// error: test already has a body.
export int test() { return 1; }

RWStructuredBuffer<int> output;

[numthreads(1,1,1)]
void main()
{
    output[0] = test();    
}
```

This error is not necessary. In this case we should allow the second `test` to override the first one during linking and overload resolution.

PR #4010, opened 2024-04-23
Do not diagnose error when a symbols is defined as 'extern' and 'export'

kaizhangNV:
Fix the issue (#3999).

For a function is defined as extern and export at the same time, don't report error, we can use the 'export' function to overload the 'extern' function.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -1141,6 +1141,15 @@ namespace Slang
         return 0;
     }
 
+    int getExportRank(DeclRef<Decl> left, DeclRef<Decl> right)
+    {
+        if (left.getDecl() && left.getDecl()->hasModifier<ExternModifier>())
+        {
+            return (right.getDecl() && right.getDecl()->hasModifier<HLSLExportModifier>()) ? -1 : 0;
+        }
+        return 0;
+    }
+
     int SemanticsVisitor::CompareOverloadCandidates(
         OverloadCandidate*	left,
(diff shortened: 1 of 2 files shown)


source/slang/slang-check-overload.cpp
```diff
@@ -1348,6 +1348,13 @@ namespace Slang
                 return;
             }
         }
+        // If this function is marked as 'extern' and in the overload candidate,
+        // it indicates that there is a 'export' function in this module that will
+        // overload this function, so we don't add this function to the overload candidate.
+        if (funcDecl->hasModifier<ExternModifier>())
```
csyonghe, line 1354: We should be adding the function to a candidate.
We need to extend `CompareOverloadCandidates` to favor `export` function instead of the `extern` function.

source/slang/slang-check-overload.cpp
```diff
@@ -1218,6 +1229,11 @@ namespace Slang
             auto overloadRankDiff = getOverloadRank(right->item.declRef) - getOverloadRank(left->item.declRef);
             if (overloadRankDiff)
                 return overloadRankDiff;
+
+            // `export` function is more flavored than `extern` function. But other modifiers are not considered.
+            auto externExportDiff = getExportRank(left->item.declRef, right->item.declRef);
```
csyonghe, line 1234: This check should be before overload rank. We use overload rank as the last resort to disambiguate and it is really just a hack and should not take precedence over anything.

csyonghe (approved):

PR #4013, opened 2024-04-24
WIP Updating CONTRIBUTION guide to use CMake

jkwak-work:
Releated to #3703

Removing the build instruction with Premake and replacing it with an instruction with CMake. It is because we are going to move over to CMake anytime soon.

Bumping the required CMake version to 3.25.0. When CMakePresets.json has "version:6" it requires CMake version to be 3.25 or above. See the URL below for more information,
  https://cmake.org/cmake/help/latest/release/3.25.html

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -73,28 +73,35 @@ Please follow the instructions of how to [Building Slang from Source](docs/build
 For a quick reference, follow the instructions below.
 
 #### Windows
-Find where the premake executable is. The location may change when we upgrade Premake.
-```
-C:\git\slang> where /r external\slang-binaries\premake premake*.exe
-```
+Download and install CMake from [CMake.org/download](https://cmake.org/download)
 
-Run the premake with the following command line options after replacing "PREMAKE" with the result from the previous command.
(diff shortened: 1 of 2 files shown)


CONTRIBUTION.md
```diff
@@ -73,28 +73,35 @@ Please follow the instructions of how to [Building Slang from Source](docs/build
...
+Or you can build with a following command:
+```
+C:\git\slang> cmake.exe --build --preset release
+```
+
 #### Linux
-Find where the premake is and run it with gmake2 option.
+Install CMake and Ninja.
+```
+$ sudo apt-get install cmake ninja-build
```
ArielG-NV, line 94: I wonder if it would be better if we would suggest to install cmake over `pip install`. `pip` always has the latest version of CMake.

The problem with `apt` is that most repos have a old versions of CMake. For example, Ubuntu 22 and older do not have cmake 3.25 in their default apt repos list.

This would probably stop the whole "why is cmake not working" issue.

jkwak-work:
Closing for now because the workflow with CMake doesn't seem to be working on Windows yet.

PR #4014, opened 2024-04-24
Support derivative functions in compute & capabilities adjustments

### LINKED ISSUE #4000: Compute shader with derivative intrinsic's do not compile correctly on direct-to-spirv backend.

ArielG-NV:
fixes #4000

PR implements derivative functions in compute shaders properly so we have the functionality for SPIR-V & GLSL. Tests reflect fragment and compute usage of derivative functions .

PR also adjusts capabilities to correct wrong SPRI-V target capabilities for when using textures.

Remarks:
1. `__requireComputeDerivative()` is a intrinsic_op and not modifier since Inlining will destroy the IR-lowered modifier->decoration.
2. Frontend has 2 options added, `-compute-derivative-quad` and `-compute-derivative-linear` for selecting what a "derivative in compute shader` is defined as. The code follows how `-matrix-layout-column-major` works in majority of senarios.


source/slang/slang-emit-spirv.cpp
```diff
@@ -2730,6 +2739,40 @@ struct SPIRVEmitContext
...
+                    break;
+
+                HashSet<IRFunc*>* entryPointsUsingInst = EntryPointsUsingInst(*this, parentFunc);
+                for (IRFunc* entryPoint : *entryPointsUsingInst)
+                {
+                    auto entryPointDecor = entryPoint->findDecoration<IREntryPointDecoration>();
+                    if (!entryPointDecor || entryPointDecor->getProfile().getStage() != Stage::Compute)
+                        continue;
+
+                    ensureExtensionDeclaration(UnownedStringSlice("SPV_NV_compute_shader_derivatives"));
```
csyonghe, line 2757: Does this mean that if I have a module with two compute shaders, the execution mode will only get declared for the first compute shader?

csyonghe (commented):
I wonder how is this handled in dxc, do they have similar compiler flags as well?

csyonghe:
It doesn't seem like the execution mode info should be provided by a compiler flag. Instead it should come as a decoration to place on the compute shader. We can introduce [DerivativeGroupQuad] and [DerivativeGroupLinear] decorations that can apply to compute shader entry points, and by default we use the linear mode.

We need to have a documentation page for all spirv specific additions in our compiler. Start a md file in docs folder called spirv-features.md and document the additional decorations there.

ArielG-NV:
> It doesn't seem like the execution mode info should be provided by a compiler flag. Instead it should come as a decoration to place on the compute shader. We can introduce [DerivativeGroupQuad] and [DerivativeGroupLinear] decorations that can apply to compute shader entry points, and by default we use the linear mode.
> 
> We need to have a documentation page for all spirv specific additions in our compiler. Start a md file in docs folder called spirv-features.md and document the additional decorations there.

I think we should stick with QUAD derivatives as default since HLSL uses QUAD derivatives.

csyonghe:
> I think we should stick with QUAD derivatives as default since HLSL uses QUAD derivatives.

Sure. We can even omit the work for specifying the mode for now until someone requests it.

ArielG-NV:
The functional code (+ the many errors) is 100% done now, just adding tests.

source/slang/slang-emit-spirv.cpp
```diff
@@ -6083,6 +6130,14 @@ struct SPIRVEmitContext
     }
 };
 
+HashSet<IRFunc*>* EntryPointsUsingInst(SPIRVEmitContext& context, IRInst* inst)
```
csyonghe, line 6133: Naming: getReferencingEntryPoints()

source/slang/slang-ir-inst-defs.h
```diff
@@ -594,6 +594,7 @@ INST_RANGE(TerminatorInst, Return, Unreachable)
 
 INST(RequirePrelude, RequirePrelude, 1, 0)
 INST(RequireGLSLExtension, RequireGLSLExtension, 1, 0)
+INST(RequireComputeDerivative, RequireComputeDerivative, 0, 0)
```
csyonghe, line 597: Maybe this should be a HOISTABLE inst so that we will have one instance of it in the global scope. Then we can simply use rhe existing reference graph to find referencingEntryPoints.

ArielG-NV: If this is for C-like-targets, we don't currently build a entry-point reference graph, I changed the logic to build one.

If this is just about not emitting multiple instructions for `RequireComputeDerivative`, I believe we still need to associate an instruction<->function in case we have multiple entry points (some use, some do not compute derivatives).

csyonghe: Well, once it is hoistable, we can simply check all entry points referencing the only instance of the requireDerivative inst, like how we are handling all the other builtin variables.

csyonghe: Ah, I see what you mean, we still need to have something in the function body that references the global thing so this isn't going to do much.

source/slang/slang-emit-c-like.cpp
```diff
@@ -4341,6 +4351,9 @@ void CLikeSourceEmitter::computeEmitActions(IRModule* module, List<EmitAction>&
     ctx.actions = &ioActions;
     ctx.openInsts = InstHashSet(module);
 
+    
+    buildEntryPointReferenceGraph(this->m_referencingEntryPoints, module);
```
csyonghe, line 4355: We should only need this for glsl backend, right?

source/slang/slang-ir-util.h
```diff
@@ -330,8 +325,23 @@ inline bool isCompositeType(IRType* type)
...
 IRParam* getParamAt(IRBlock* block, UIndex ii);
 
+void verifyComputeDerivativeGroupModifiers(
+    DiagnosticSink* sink,
+    SourceLoc errorLoc,
+    bool quadAttr,
+    bool linearAttr,
+    IRNumThreadsDecoration* numThreadsDecor);
+
+void buildEntryPointReferenceGraph(Dictionary<IRInst*, HashSet<IRFunc*>>& referencingEntryPoints, IRModule* module);
```
csyonghe, line 341: Consider move this to its own header/cpp file. slang-ir-call-graph.cpp

csyonghe (approved):

source/slang/slang-emit-c-like.cpp
ArielG-NV: yes

ArielG-NV: I can move the logic to init for GLSL specifically for now.

csyonghe (approved):

csyonghe (approved):

PR #4017, opened 2024-04-24
Updating CONTRIBUTION guide to use CMake

jkwak-work:
Releated to #3703

Removing the build instruction with Premake and replacing it with an instruction with CMake. It is because we are going to move over to CMake anytime soon.

Bumping the required CMake version to 3.25.0. When CMakePresets.json has "version:6", it requires CMake version to be 3.25 or above. See the URL below for more information,
  https://cmake.org/cmake/help/latest/release/3.25.html

CMakeLists.txt copies the prebuilt binary files from
  external/slang-binaries/bin/windows-x64


jkwak-work:
It looks like slang-llvm.dll is copied to a wrong directory and it make slang-test ignore FILECHECK tests.
I will push a fix for it soon.

csyonghe (approved):

PR #4028, opened 2024-04-25
Keep const-ness in generic functions

### LINKED ISSUE #3834: "const" variable loses const-ness inside of a generic function
I found that "const"-ness is removed inside of a generic function.
See the following example,
```
//#define USE_GENERIC 1

uniform sampler2D uniform_sampler2D;

#if USE_GENERIC
__generic<T:__BuiltinArithmeticType>
#endif
vec4 MyFunc()
{
    const ivec2[4] offsets = { ivec2(0,0), ivec2(0,0), ivec2(1,1), ivec2(1,1) };
    return textureGatherOffsets(uniform_sampler2D, vec2(0), offsets);
}

[numthreads(4, 1, 1)]
{
#if USE_GENERIC
    vec4 result = MyFunc<float>();
#else
    vec4 result = MyFunc();
#endif

    outputBuffer[0] = result[0];
}
```
When the file is saved as "const_with_generic.slang", you can generate two GLSL source files with following commands:
```
slangc.exe -target glsl -lang glsl -entry computeMain -allow-glsl -stage compute -DUSE_GENERIC=0 -o const_without_generic.glsl const_with_generic.slang
slangc.exe -target glsl -lang glsl -entry computeMain -allow-glsl -stage compute -DUSE_GENERIC=1 -o const_with_generic.glsl const_with_generic.slang
```
When "USE_GENERIC=0", the "const" is preserved; see the following result code:
```
vec4 MyFunc_0()
{
    const ivec2 _S1 = ivec2(0, 0);

#line 11
    const ivec2 _S2 = ivec2(1, 1);
    const vec2 _S3 = vec2(0.0);

#line 12
    vec4 _S4 = (textureGatherOffsets((uniform_sampler2D_0), (_S3), ivec2[]((_S1), (_S1), (_S2), (_S2)), (0)));
```
But when "USE_GENERIC=1", the "const" modifier is lost; see the following result code:
```
vec4 MyFunc_0()
{
    ivec2 _S1 = ivec2(0, 0);

#line 11
    ivec2 _S2 = ivec2(1, 1);
    vec2 _S3 = vec2(0.0);

#line 12
    vec4 _S4 = (textureGatherOffsets((uniform_sampler2D_0), (_S3), ivec2[]((_S1), (_S1), (_S2), (_S2)), (0)));
```
Because of the difference, when textureGatherOffsets() is called from a generic function, it always results in a compile error.
It is because the offset value to "textureGatherOffsets()" is required to be a compile-time constant.
```
$ slangc.exe -target spirv -lang glsl -entry computeMain -allow-glsl -stage compute -o const_without_generic.spirv -DUSE_GENERIC=1 const_with_generic.slang
glslang:  const_with_generic.slang(12): error :  'textureGatherOffsets(...)' : must be a compile-time constant: offsets argument
glslang:  const_with_generic.slang(12): error :  '' : compilation terminated
glslang: note : ERROR: 2 compilation errors.  No code generated.
```

jkwak-work:
Closes #3834

The issue was that "const" variables inside of generic functions became non-const variables.  This issue prevented some of GLSL texture functions from being called inside of generic functions.

`propagateConstExpr()` was iterating global functions excepts generic functions. This commit includes generic functions and resolves the issue.

### DIFF
--- a/source/slang/slang-ir-constexpr.cpp
+++ b/source/slang/slang-ir-constexpr.cpp
@@ -650,6 +650,14 @@ void propagateConstExpr(
         default:
             break;
 
+        case kIROp_Generic:
+            {
+                auto gen = as<IRGeneric>(gv);
+                gv = as<IRFunc>(findGenericReturnVal(gen));
+                if (nullptr == gv)
+                    break;
+            }
+            [[fallthrough]];
         case kIROp_Func:
         case kIROp_GlobalVar:
             {
--- a/tests/bugs/gh-3834.slang
+++ b/tests/bugs/gh-3834.slang
@@ -1,0 +1,32 @@
+//TEST:SIMPLE(filecheck=CHECK): -stage compute -entry computeMain -target glsl -allow-glsl
+//TEST:SIMPLE(filecheck=CHECK): -stage compute -entry computeMain -target glsl -allow-glsl -DUSE_GENERIC
+
+//#define USE_GENERIC
+
+uniform sampler2D uniform_sampler2D;
+RWStructuredBuffer<float> outputBuffer;
+
+#ifdef USE_GENERIC
+__generic<T:__BuiltinArithmeticType>
+#endif
+vec4 MyFunc()
+{
+    //CHECK: const ivec2 offset0
+    //CHECK: const ivec2 offset1
+    const ivec2 offset0 = ivec2(0,0);
+    const ivec2 offset1 = ivec2(1,1);
+    const ivec2[4] offsets = { offset0, offset0, offset1, offset1 };
+    return textureGatherOffsets(uniform_sampler2D, vec2(0), offsets);
+}
+
+[numthreads(4, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+#ifdef USE_GENERIC
+    vec4 result = MyFunc<float>();
+#else
+    vec4 result = MyFunc();
+#endif
+
+    outputBuffer[0] = result[0];
+}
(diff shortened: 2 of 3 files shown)


source/slang/slang-ir-constexpr.cpp
```diff
@@ -195,6 +195,14 @@ void maybeAddToWorkList(
     {
         context->workList.add(gv);
         context->onWorkList.add(gv);
+
+        // Iterating uses to include generic functions
```
csyonghe, line 199: Can you explain with an example what issue we are having and why this solves the problem?

csyonghe (dismissed):

source/slang/slang-ir-constexpr.cpp
jkwak-work: I can leave a more descriptive comment as requested.
But before that, I like to check it with you if I am doing the right thing.
The issue I am trying to address is described in [my comment](https://github.com/shader-slang/slang/issues/3834#issuecomment-2076069811).

Basically, it `module->getGlobalInsts()` doesn't include generic functions.
Only way I can iterate the generic functions is to iterate all of uses like how it is done in `SpecializationContext::addToWorkList()`.
Is this a right way to iterate generic functions?

csyonghe: getGlobalInsts() should return all insts, including IRGeneric. However, the function body of a generic function is an IRFunc in the body of the generic.

You need to do:

if (auto gen=as<IRGeneric>(globalInst))
   auto innerFunc = as<IRFunc>(findGenericReturnVal(gen));

jkwak-work: Oh. I see.
I will try that.

jkwak-work:
I pushed a new change and it works more nicely.

csyonghe (dismissed):

Issue #3695, opened 2024-03-06
Error when compiling Slang code for PyTorch Kernel


half-potato:
Slang code:
```
[Differentiable]
float3 project(in float3 xyz, in float4x4 wct) {
  float4 xyzw = float4(xyz, 1.f);
  let p_view = mul(xyzw, wct);
  float2 pix2d = {p_view.x / p_view.z, p_view.y / p_view.z};
  return {pix2d.x, pix2d.y, p_view.z};
}

[Differentiable]
float3 inv_project(in float2 xy, in float dist, in float4x4 inv_wvt) {
  let p_hom = float4(xy * dist, dist, 1.f);
  let out = mul(p_hom, inv_wvt);
  return {out.x, out.y, out.z};
}


[AutoPyBindCUDA]
[CUDAKernel]
float3 inv_project_th(TensorView<float> xy, float dist, TensorView<float> inv_wvt) {
  
}
```

When built:
```
Traceback (most recent call last):
  File "/home/amai/micromamba/envs/gaussian_splatting/lib/python3.11/site-packages/torch/utils/cpp_extension.py", line 2100, in _run_ninja_build
    subprocess.run(
  File "/home/amai/micromamba/envs/gaussian_splatting/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['ninja', '-v']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/amai/test/test.py", line 3, in <module>
    slangpy.loadModule("test.slang")
  File "/home/amai/micromamba/envs/gaussian_splatting/lib/python3.11/site-packages/slangpy/slangpy.py", line 616, in loadModule
    rawModule = _loadModule(fileName, moduleName, buildDir, options, sourceDir=outputFolder, verbose=verbose, includePaths=includePaths, dryRun=False)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/amai/micromamba/envs/gaussian_splatting/lib/python3.11/site-packages/slangpy/slangpy.py", line 535, in _loadModule
[...]

PR #4042, opened 2024-04-26
Added diagnostics & built-in type lowering for `[CUDAKernel]` functions

saipraveenb25:
This PR adds
- Diagnostics for non-void return from a cuda kernel entry point
- Diagnostics for using differentiable types in a differentiable cuda kernel entry point
- Logic for converting built-in types (float3, float3x3, etc..) to portable struct types and unpacks the parameter back into a built-in type on the CUDA side. This is because built-in types have different implementations in CUDA & CPP targets, which causes signature mis-match when linking.

PyTorch-side logic for wrapping the storage types is in a separate PR in slang-torch: https://github.com/shader-slang/slang-torch/pull/4/

Fixes https://github.com/shader-slang/slang/issues/3695


source/slang/slang-ir-pytorch-cpp-binding.cpp
```diff
@@ -858,6 +859,99 @@ IRFunc* generateCUDAWrapperForFunc(IRFunc* func, DiagnosticSink* sink)
...
+                }
+                
+                changed = true;
+            }
+        }
+
+        if (changed)
+        {
+            // Rebuild func-type
+            List<IRType*> paramTypes;
```
csyonghe, line 944: Can call fixupFuncType()

source/slang/slang-ir-pytorch-cpp-binding.cpp
```diff
@@ -858,6 +859,99 @@ IRFunc* generateCUDAWrapperForFunc(IRFunc* func, DiagnosticSink* sink)
...
+        List<IRParam*> params;
+        for (auto param : func->getFirstBlock()->getParams())
+        {
+            params.add(param);
+        }
+
+        bool changed = false;
+        for (auto param : params)
+        {
+            LoweredBuiltinTypeInfo info = LoweredBuiltinTypeInfo();
```
csyonghe, line 890: This is not handling struct types?

Can we extend the existing lower-buffer-element-type pass to make it work for CUDA?

saipraveenb25: I tried to adapt the lower-buffer-element pass to also process this, but there were too many thing that had to be done differentl,y. For one, the buffer lowering pass needs to care about matrix and array types, and expresses them in terms of vectors. For builtin-type lowering we need to lower matrix and vector types into structs of arrays, and we don't necessarily care about alignment (since this pass is `[CUDAKernel]` specific). 
We also don't need to care about rewriting accesses, since we can construct the built-in type back on the CUDA side & retain all the other instructions.

But yes, we do need to handle built-in types that are nested inside structs.

source/slang/slang-ir-pytorch-cpp-binding.cpp
saipraveenb25: Will fix this up.

source/slang/slang-ir-pytorch-cpp-binding.cpp
csyonghe: That's what I mean by extending the pass. We need to implement the logic to selectively do certain things based on the target we are generating code for, but share the overall structure/flow of the pass so we don't have to duplicate the logic for e.g. handling arrays and structures.

saipraveenb25: I can give it another try, but even the structure of the rewriting pass needs to be a bit different. 

The current rewriting pass finds buffer types and works backwards from there (affecting all uses of the type). We need a separate, more narrowly scoped pathway which only affects the parameter type of a kernel function (and not other uses of the type).  

Technically, we can group the two passes together, but I felt that since both the traversal logic and rewrite logic was different, it might be more confusing to have them in the same step. 
My intention was to instead group this rewrite pass with the other type rewriting passes in the auto-binding logic (we already have a device-to-host type translation step, and a host-to-tuple-type translation step). 

Would that work better, or should I try again to unify this logic with the buffer lowering pass?

csyonghe: I see. I'll trust your judgement on how much overlap there is to worth the unification. I think what we need is a pass that can rewrite the type of a global variable, a struct field, or a function parameter. This pass can then be used to drive both rewriting we are dealing with here. For AutoPyBind, you only enable the worklist discovery logic for entry point parameters, and not anything else.

saipraveenb25: It is true that we need to unify type rewriting steps though.. I feel like I've written 3 or 4 different passes that almost work the same way, but not quite. 
I'll raise a separate issue and try to merge these passes together (starting with the auto-bind related rewrites) as a part of the cleanup later this week.

csyonghe (dismissed):

csyonghe (approved):

27474210:
....

PR #4050, opened 2024-04-28
Replace __target_intrinsics and __specialize_for_target, part 1

### LINKED ISSUE #3906: Remove most uses of `__target_intrinsic` and `__specialize_for_target`
There are a few cases where `__target_intrinsic` and `__specialize_for_target` is required: 1. edge cases 2. types 3. `[[vk::.*]]`. 

Other than these scenarios (and possibly a few others), we should be removing all uses of these attributes since they are incompatible with the capabilities system and are not needed anymore in most cases.

jkwak-work:
Partially resolves #3906

Most but not all __target_intrinsics are replaced with __target_switch. All __specialize_for_target are replaced with __target_switch.

This change is mostly processed by a temporary c++ program mechanically. Because the change is already too big, the remaining __target_intrinsics will be replaced later in another commit.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -322,14 +322,18 @@ public vector<T,N> atan(vector<T,N> y, vector<T,N> x)
 }
 
 __generic<T : __BuiltinFloatingPointType>
-__target_intrinsic(cuda, "$P_asinh($0)")
-__target_intrinsic(cpp, "$P_asinh($0)")
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_spirv, GLSL_130)]
 public T asinh(T x)
 {
-    return log(x + sqrt(x * x + T(1)));
+    __target_switch
+    {
+    case cpp: __intrinsic_asm "$P_asinh($0)";
+    case cuda: __intrinsic_asm "$P_asinh($0)";
+    default:
+        return log(x + sqrt(x * x + T(1)));
+    }
 }
 
 __generic<T : __BuiltinFloatingPointType, let N:int>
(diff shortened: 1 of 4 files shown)


source/slang/glsl.meta.slang
```diff
@@ -322,14 +322,18 @@ public vector<T,N> atan(vector<T,N> y, vector<T,N> x)
...
 [require(cpp_cuda_glsl_hlsl_spirv, GLSL_130)]
 public T asinh(T x)
 {
-    return log(x + sqrt(x * x + T(1)));
+    __target_switch
+    {
+    case cpp: __intrinsic_asm "$P_asinh($0)";
+    case cuda: __intrinsic_asm "$P_asinh($0)";
+    default:
+        return log(x + sqrt(x * x + T(1)));
```
ArielG-NV, line 335: This is not a problem with the PR, although I did notice we implement many GLSL functions without a GLSL/SPIRV intrinsic path when such options exist (asinh/pack functions are an example)

An issue was made to address this (for SPIR-V). A comment is being added to link the PR to the issue for future reference: #4051

source/slang/glsl.meta.slang
```diff
@@ -833,29 +895,39 @@ public T faceforward(T n, T i, T ng)
...
 [require(cpp_cuda_glsl_hlsl_spirv, GLSL_400)]
-matrix<T,N,N> inverse(matrix<T,N,N> m);
+matrix<T,N,N> inverse(matrix<T,N,N> m)
+{
+    __target_switch
+    {
+    case glsl: __intrinsic_asm "inverse";
+    case hlsl: __intrinsic_asm "inverse";
+    }
+}
```
ArielG-NV, line 930: does this function compile?
it targets cpp, cuda, glsl, hlsl, spirv, but only implements glsl & hlsl.

jkwak-work: CI is all passing and it compile fine on my local machine, which seems strange.

source/slang/glsl.meta.slang
jkwak-work: I noticed it too.
Thank you for filing the issue.
I assigned it to myself.

source/slang/glsl.meta.slang
ArielG-NV: I see in `intrinsic-basic.slang` the test for `inverse` is never ran (`#ifdef`)
```
#if defined(TEST_when_inverse_works)
        && equals(mat2(0), inverse(mat2(0)))
        && equals(mat3(0), inverse(mat3(0)))
        && equals(mat4(0), inverse(mat4(0)))
#endif // #if defined(TEST_when_inverse_works)
```
I believe this may be why no error appears?

ArielG-NV (approved):
Minor comments have been shared, otherwise code looks good to me.

ArielG-NV (dismissed):

csyonghe (approved):
Looks good. I can't go through every single line of change here, and I am assuming things are fine given all tests are passing.

PR #4054, opened 2024-04-30
Add option -disable-short-circuit

kaizhangNV:
Add option -disable-short-circuit to disable short circuit for logic operators && and ||. Also, disable the short circuit by default in the stdlib.

### DIFF
--- a/source/slang/slang.cpp
+++ b/source/slang/slang.cpp
@@ -170,6 +170,7 @@ void Session::init()
     // Built in linkage uses the built in builder
     m_builtinLinkage = new Linkage(this, builtinAstBuilder, nullptr);
     m_builtinLinkage->m_optionSet.set(CompilerOptionName::DebugInformation, DebugInfoLevel::None);
+    m_builtinLinkage->m_optionSet.set(CompilerOptionName::DisableShortCircuit, true);
 
     // Because the `Session` retains the builtin `Linkage`,
     // we need to make sure that the parent pointer inside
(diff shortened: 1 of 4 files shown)


source/slang/slang.cpp
```diff
@@ -170,6 +170,7 @@ void Session::init()
     // Built in linkage uses the built in builder
     m_builtinLinkage = new Linkage(this, builtinAstBuilder, nullptr);
     m_builtinLinkage->m_optionSet.set(CompilerOptionName::DebugInformation, DebugInfoLevel::None);
+    m_builtinLinkage->m_optionSet.set(CompilerOptionName::DisableShortCircuit, true);
```
csyonghe, line 173: Let's keep the default to not disable.

kaizhangNV: I noticed that if we don't disable the short-circuit in stdlib, then the user code could be overload by the builtin  "&&".
So something like
`bool = int && int`
will still be short-circuiting.
How can we not allow such overload?

csyonghe: Why choosing && would lead to short circuiting?

kaizhangNV: Because the stdlib defines "&&" operator which is compiled by enabling short-circuit.

csyonghe (approved):

source/slang/slang.cpp
kaizhangNV: Won't this cause issue? Is `and(T1, T2)` defined somewhere else?

kaizhangNV: https://github.com/shader-slang/slang/pull/4056
New PR is trying to fix this issue.

Issue #3425, opened 2023-12-28
Differential types do not match principle type despite all members being differentiable


natevm:
If I have a structure like this, which implements IDifferentiable:

```
struct Ray: IDifferentiable {
    float3 Origin;
    float3 Direction;
    float TMin, TMax;
}
```

Then I would expect the Differential type for "Ray" to also be "Ray". However, the differential type is instead a "Ray.Differential". 

There appears to be some missing logic in the synthesis to use the existing type as differential if all its members' differential type are the same as their original type.

pmistryNV:
As part of the bug will add test cases to turn on the IR validation in spirv only path.

natevm:
> As part of the bug will add test cases to turn on the IR validation in spirv only path.

That would be very helpful. I had a similar issue related to slang SPIRV not being validated. 

For compiler perf reasons, it might be worth considering disabling validation though (or conditionally including via a flag?) I’m not sure how long validation takes, but I could see it being a problem if we’re always doing it.

PR #4057, opened 2024-04-30
Handle case where types can be used as their own `Differential` type.

saipraveenb25:
Detect and skip synthesis wherever it is possible to use the type as its own differential type. 
This case is fairly common, and this PR allows assignments like `T.Differential = T`, while generating less code.

Note: This PR also contains a fix for an issue with out-of-date type resolution cache. Witness tables created during the conformance checking phase not being taken into account during the decl type resolution phase because the epoch is not updated after conformance checking.
This leads to certain complex associated-type lookup chains (such as the one in tests/compute/assoctype-nested-lookup.slang) not resolving properly and causing errors.

Fixes https://github.com/shader-slang/slang/issues/3425


source/slang/slang-check-expr.cpp
```diff
@@ -1145,6 +1177,34 @@ namespace Slang
...
+    {
+        // A struct can be used as its own differential type if all its members are differentiable
+        // and their differential types are the same as the original types. 
+        //
+        bool canBeUsed = true;
+        for (auto member : aggTypeDecl->members)
+        {
+            if (auto varDecl = as<VarDecl>(member))
+            {
+                if (!isTypeDifferentiable(varDecl->getType()))
```
csyonghe, line 1190: This shouldn't be needed, because calling `tryGetDifferentialType()` on a non diffiernetiable type will return nullptr and the logic below should just work?

source/slang/slang-check-decl.cpp
```diff
@@ -2144,7 +2144,69 @@ namespace Slang
...
+            if (auto thisWitness = findThisTypeWitness(
+                SubstitutionSet(requirementDeclRef),
+                as<InterfaceDecl>(requirementDeclRef.getParent()).getDecl()))
+            {
+                if (auto declRefType = as<DeclRefType>(thisWitness->getSub()))
+                {
+                    substSet = SubstitutionSet(declRefType->getDeclRef());
+                }
+            }
+            auto satisfyingType = DeclRefType::create(m_astBuilder, m_astBuilder->getMemberDeclRef(substSet.declRef, context->parentDecl));
```
csyonghe, line 2168: Isn't this `context->conformingType`?

source/slang/slang-check-decl.cpp
```diff
@@ -2144,7 +2144,69 @@ namespace Slang
...
+            
+            auto assocTypeDef = m_astBuilder->create<TypeDefDecl>();
+            assocTypeDef->nameAndLoc.name = getName("Differential");
+            assocTypeDef->type.type = satisfyingType;
+            assocTypeDef->parentDecl = context->parentDecl;
+            assocTypeDef->setCheckState(DeclCheckState::DefinitionChecked);
+            context->parentDecl->members.add(assocTypeDef);
+            
+            // Add derivative member attributes to all the fields pointing to themselves.
+            for (auto member : context->parentDecl->getMembersOfType<VarDeclBase>())
```
csyonghe, line 2178: May need to leave a TODO comment here that we also need to look at all base types from `InheritanceDecl`s, they are essentially the same as members.

source/slang/slang-check-decl.cpp
```diff
@@ -2144,7 +2144,51 @@ namespace Slang
...
+        // 
+        if (!aggTypeDecl
+            && as<AggTypeDecl>(context->parentDecl)
+            && canStructBeUsedAsSelfDifferentialType(as<AggTypeDecl>(context->parentDecl)))
+        {
+            // If the parent type can be used as its own differential type, we will create a typealias
+            // to itself as the differential type.
+            //
+            SubstitutionSet substSet;
+            if (auto thisWitness = findThisTypeWitness(
```
csyonghe, line 2159: I still don't understand what the purpose of this logic is. `substSet` is no longer used?

saipraveenb25: Whoops.. I think I copied it in from the regular synthesis logic and forgot to clean it up.

csyonghe (approved):

PR #4064, opened 2024-05-01
SPIRV: Fix performance issue when handling large arrays.

### LINKED ISSUE #4038: Accessing pointer of structs containing large arrays leads to long compilation times
Compiling the following snippet takes around 15 seconds and the process allocates over 2GB of memory, growing proportionally to the array sizes. This seems to be specific to the SPIRV target and does not happen with HLSL (although I'm aware it doesn't support pointers).

```hlsl
// slangc -target spirv -emit-spirv-directly -profile glsl_460 -lang slang -O3 test2.slang
struct WorkData {
    float A[2048 * 2048];
    float B[2048 * 2048];
};
struct PushData {
    WorkData* Input;
    float* Dest;
};

[vk::push_constant] ConstantBuffer<PushData> cb;

[numthreads(64, 1, 1)]
void ComputeMain(uint tid: SV_DispatchThreadID)
{
    cb.Dest[tid] = cb.Input->A[tid] * cb.Input->B[tid];
}
```

csyonghe:
The problem of long compile time is due to our `lower-buffer-element-type` pass is emitting expanded/unrolled instructions to pack/unpack each element in an array.

We fix this problem by emitting a loop to copy the array when the array is bigger than a threshold, instead of just emit raw insts that processes each element.

Fixes #4038.

### DIFF
--- a/source/slang/slang-ir-lower-buffer-element-type.cpp
+++ b/source/slang/slang-ir-lower-buffer-element-type.cpp
@@ -9,6 +9,8 @@ namespace Slang
 {
     struct LoweredElementTypeContext
     {
+        static const IRIntegerValue kMaxArraySizeToUnroll = 32;
+
         struct LoweredElementTypeInfo
         {
             IRType* originalType;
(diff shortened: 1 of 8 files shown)


jkwak-work (approved):
Looks good to me.
It was interesting for me to see how a loop statement gets created.

PR #4066, opened 2024-05-01
Add LoadAligned and StoreAligned methods to ByteAddressBuffers

### LINKED ISSUE #4062: Add alignment information to the ByteAddressBuffer Inst
Follow-up to https://github.com/shader-slang/slang/issues/3533

Vector load/store operations on ByteAddressBuffer requires the compiler to check if the memory is aligned at least to the size of the data type of the resource. This check is costly, and can be avoided by simply adding the alignment information to the instruction itself, and check if it is set, during lowering from ByteAddressBuffer load/store to StructuredBuffer load/store.

Use the test `tests/compute/byte-address-buffer-aligned.slang` as reference.
```
[vk::binding(2, 3)] RWByteAddressBuffer buffer0;

[shader("compute")]
[numthreads(1,1,1)]
void computeMain(uint3 threadId : SV_DispatchThreadID)
{
	buffer0.Store(64, buffer0.Load<uint4>(32));
}
```

sriramm-nv:
Fixes #4062

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2824,23 +2824,23 @@ uint64_t __asuint64(uint2 i)
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
-T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset, int aligned);
+T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset, int alignment);
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-T __byteAddressBufferLoad<T>(RWByteAddressBuffer buffer, int offset, int aligned);
+T __byteAddressBufferLoad<T>(RWByteAddressBuffer buffer, int offset, int alignment);
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-T __byteAddressBufferLoad<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int aligned);
+T __byteAddressBufferLoad<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int alignment);
 
 __intrinsic_op($(kIROp_ByteAddressBufferStore))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-void __byteAddressBufferStore<T>(RWByteAddressBuffer buffer, int offset, int aligned, T value);
+void __byteAddressBufferStore<T>(RWByteAddressBuffer buffer, int offset, int alignment, T value);
 
 __intrinsic_op($(kIROp_ByteAddressBufferStore))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-void __byteAddressBufferStore<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int aligned, T value);
+void __byteAddressBufferStore<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int alignment, T value);
 
 __generic<T, L:IBufferDataLayout=DefaultDataLayout>
 __magic_type(HLSLStructuredBufferType)
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -108,7 +108,20 @@ struct ByteAddressBuffer
         {
         case hlsl: __intrinsic_asm ".Load";
         default:
-            return __byteAddressBufferLoad<uint>(this, location);
+            return __byteAddressBufferLoad<uint>(this, location, 0);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint LoadAligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load";
+        default:
+            return __byteAddressBufferLoad<uint>(this, location, 1);
         }
     }
 
@@ -124,7 +137,20 @@ struct ByteAddressBuffer
         {
         case hlsl: __intrinsic_asm ".Load2";
         default:
-            return __byteAddressBufferLoad<uint2>(this, location);
+            return __byteAddressBufferLoad<uint2>(this, location, 0);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint2 Load2Aligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load2";
+        default:
+            return __byteAddressBufferLoad<uint2>(this, location, 1);
         }
     }
 
@@ -140,7 +166,20 @@ struct ByteAddressBuffer
         {
         case hlsl: __intrinsic_asm ".Load3";
         default:
-            return __byteAddressBufferLoad<uint3>(this, location);
+            return __byteAddressBufferLoad<uint3>(this, location, 0);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint3 Load3Aligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load3";
+        default:
+            return __byteAddressBufferLoad<uint3>(this, location, 1);
         }
     }
 
@@ -156,7 +195,20 @@ struct ByteAddressBuffer
         {
         case hlsl: __intrinsic_asm ".Load4";
         default:
-            return __byteAddressBufferLoad<uint4>(this, location);
+            return __byteAddressBufferLoad<uint4>(this, location, 0);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint4 Load4Aligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load4";
+        default:
+            return __byteAddressBufferLoad<uint4>(this, location, 1);
         }
     }
 
@@ -167,7 +219,14 @@ struct ByteAddressBuffer
     [ForceInline]
     T Load<T>(int location)
     {
-        return __byteAddressBufferLoad<T>(this, location);
+        return __byteAddressBufferLoad<T>(this, location, 0);
+    }
+
+    [__readNone]
+    [ForceInline]
+    T LoadAligned<T>(int location)
+    {
+        return __byteAddressBufferLoad<T>(this, location, 1);
     }
 };
 
@@ -2765,23 +2824,23 @@ uint64_t __asuint64(uint2 i)
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
-T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset);
+T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset, int aligned);
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-T __byteAddressBufferLoad<T>(RWByteAddressBuffer buffer, int offset);
+T __byteAddressBufferLoad<T>(RWByteAddressBuffer buffer, int offset, int aligned);
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-T __byteAddressBufferLoad<T>(RasterizerOrderedByteAddressBuffer buffer, int offset);
+T __byteAddressBufferLoad<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int aligned);
 
 __intrinsic_op($(kIROp_ByteAddressBufferStore))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-void __byteAddressBufferStore<T>(RWByteAddressBuffer buffer, int offset, T value);
+void __byteAddressBufferStore<T>(RWByteAddressBuffer buffer, int offset, int aligned, T value);
 
 __intrinsic_op($(kIROp_ByteAddressBufferStore))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer_rw)]
-void __byteAddressBufferStore<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, T value);
+void __byteAddressBufferStore<T>(RasterizerOrderedByteAddressBuffer buffer, int offset, int aligned, T value);
 
 __generic<T, L:IBufferDataLayout=DefaultDataLayout>
 __magic_type(HLSLStructuredBufferType)
@@ -2898,7 +2957,20 @@ struct $(item.name)
         {
         case hlsl: __intrinsic_asm ".Load";
         default:
-            return __byteAddressBufferLoad<uint>(this, location);
+            return __byteAddressBufferLoad<uint>(this, location, 0);
+        }
+    }
+
+    [__NoSideEffect]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
+    uint LoadAligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load";
+        default:
+            return __byteAddressBufferLoad<uint>(this, location, 1);
         }
     }
 
@@ -2914,7 +2986,20 @@ struct $(item.name)
         {
         case hlsl: __intrinsic_asm ".Load2";
         default:
-            return __byteAddressBufferLoad<uint2>(this, location);
+            return __byteAddressBufferLoad<uint2>(this, location, 0);
+        }
+    }
+
+    [__NoSideEffect]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
+    uint2 Load2Aligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load2";
+        default:
+            return __byteAddressBufferLoad<uint2>(this, location, 1);
         }
     }
 
@@ -2930,7 +3015,20 @@ struct $(item.name)
         {
         case hlsl: __intrinsic_asm ".Load3";
         default:
-            return __byteAddressBufferLoad<uint3>(this, location);
+            return __byteAddressBufferLoad<uint3>(this, location, 0);
+        }
+    }
+
+    [__NoSideEffect]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
+    uint3 Load3Aligned(int location)
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load3";
+        default:
+            return __byteAddressBufferLoad<uint3>(this, location, 1);
         }
     }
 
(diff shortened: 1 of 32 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -105,13 +105,13 @@ struct ByteAddressBuffer
     [__readNone]
     [ForceInline]
     [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
-    uint Load(int location)
+    uint Load(int location, int aligned)
```
csyonghe, line 108: You can't change the signature of public facing method here. Instead, just pass 0 to `__byteAddressBufferLoad` call in the body. Same applies to other methods.

source/slang/hlsl.meta.slang
```diff
@@ -121,13 +121,13 @@ struct ByteAddressBuffer
     [__readNone]
     [ForceInline]
     [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
-    uint2 Load2(int location)
+    uint2 Load2(int location, int aligned)
```
csyonghe, line 124: For Load2/Load3/Load4/Load<T>, we want to also add:
Load2Aligned/Load3Aligned/Load4Aligned/LoadAligned<T> in parallel to this methods and pass `1` to `aligned` parameter when calling `__byteAddressBufferLoad`.

source/slang/hlsl.meta.slang
```diff
@@ -2676,23 +2747,23 @@ uint64_t __asuint64(uint2 i)
 
 __intrinsic_op($(kIROp_ByteAddressBufferLoad))
 [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
-T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset);
+T __byteAddressBufferLoad<T>(ByteAddressBuffer buffer, int offset, int aligned);
```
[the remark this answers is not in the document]
sriramm-nv: does this require a new Inst to be generated as well? kIROp_ByteADdressBufferLoadAligned ?

I thought of using bool, but for some reason, that didn't work. I will try again to change it to bool. But if you really insist on a new instruction, I can go ahead and do that.

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -205,17 +206,13 @@ struct ByteAddressBufferLegalizationContext
         return false;
     }
 
-    bool checkUnaligned(IRInst* baseOffset, IRIntegerValue immediateOffset, IRType* elementType, IRIntegerValue elementCount)
+    bool checkUnaligned(IRInst* isAligned)
     {
         // Check whether the given composite resource type is aligned to the baseOffset
```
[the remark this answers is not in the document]
sriramm-nv: will fix.

source/slang/hlsl.meta.slang
sriramm-nv: done

source/slang/hlsl.meta.slang
sriramm-nv: done

source/slang/hlsl.meta.slang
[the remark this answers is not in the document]
csyonghe: Our IR doesn't really have a strong distinction between int and bool in that a bool literal is also an `IRIntLit` whose type is `IRBoolType`. Given that this is just an internal intrinsic that maps 1:1 to an inst, this is fine as long as we document the meaning of the operand.

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -324,9 +322,19 @@ struct ByteAddressBufferLegalizationContext
             // legalization if the array type isn't in the right form
             // for us to proceed.
             //
+
             if (auto elementCountInst = as<IRIntLit>(arrayType->getElementCount()))
             {
-                return emitLegalSequenceLoad(type, buffer, baseOffset, immediateOffset, kIROp_MakeArray, arrayType->getElementType(), elementCountInst->getValue());
+                // Emit an aligned load operation on an array when using a LoadAligned inst.
+                // Else, fallback to scalarizing the loads.
+                if (!isValid(aligned))
```
csyonghe, line 330: I am not sure this is correct.

If I have:
```
struct MyType
{
    float4 arr[4];
};

buffer.LoadAligned<MyType>(offset = 16, alignment = 16)
```
The current logic will cause this to be translated into a load from `[0]`, right?

In general, we need to have:
```
ByteAddressBuffer::Load<T>(uint offset);
ByteAddressBuffer::Load<T>(uint offset, uint alignment);

ByteAddressBuffer::LoadAligned<T>(uint offset)
{
    return Load<T>(offset, __alignOf<T>());
}

```

We almost always want to load the data from an equivalent `StructuredBuffer<uint4>` or `StructuredBuffer<uint2>` if we know `offset + immediateOffset` is aligned to 16 or 8 bytes. The `alignment` parameter if provided by the user, tells us what `offset` is aligned to if it is not a known compile time value.

csyonghe: Note that in this case, sizeof(MyType) is 64, but `Load<MyType>(offset =16,  alignment=16)` should still result in aligned vectorized reads, despite `offset=16` is not aligned to the boudary of `MyType`'s size (64).

sriramm-nv: I think I understand now. Instead of having align as a true / false value, we can have it as an integer value that the user can pass, and use that information to get the exact alignment check.
So, within the stdlib, all the references of Load/ LoadAligned would be modified to something like below.
```
    [__NoSideEffect]
    [ForceInline]
    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
    T Load<T>(int location)
    {
        return __byteAddressBufferLoad<T>(this, location, 0);
    }

    [__NoSideEffect]
    [ForceInline]
    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
    T Load<T>(int location, int align)
    {
        return __byteAddressBufferLoad<T>(this, location, align);
    }

    [__NoSideEffect]
    [ForceInline]
    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
    T LoadAligned<T>(int location)
    {
        return Load<T>(location, __alignOf<T>());
    }
```
Does this look ok?, we may want to default align = -1, because, align = 0, now also means that the resource is aligned.

csyonghe: I think `alignment = 1` as default makes sense because 1 byte is the minmum alignment you can get (which means no alignments).

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -205,17 +206,17 @@ struct ByteAddressBufferLegalizationContext
...
-        IRIntegerValue elementStride = elementLayout.getStride();
-        bool isUnaligned = true;
-        if (auto baseOffsetVal = as<IRIntLit>(baseOffset)) {
-            isUnaligned = ((baseOffsetVal->getValue() + immediateOffset) % (elementStride * elementCount)) != 0;
+        if (auto alignInst = as<IRIntLit>(alignment)) {
+            if (auto baseOffsetVal = as<IRIntLit>(offset)) {
+                return ((baseOffsetVal->getValue() + immediateOffset) % (alignInst->getValue())) == 0;
+            }
         }
-        return isUnaligned;
```
jkwak-work, line 218: I have a question.
Wasn't the code before your change doing the same thing that you are doing in this change?
It was checking the alignment based on the alignment of the given element type.
Is the difference on whether the alignment is based on the element type of the vector type?

source/slang/slang-ir-inst-defs.h
```diff
@@ -428,25 +428,27 @@ INST(ImageStore, imageStore, 3, 0)
...
 
 // Store (almost) arbitrary-type data to a byte-address buffer
 //
-// byteAddressBufferLoad(%buffer, %offset, %src)
+// byteAddressBufferLoad(%buffer, %offset, %aligned, %src)
 //
 // where
 // - `buffer` is a value of some `ByteAddressBufferTypeBase` type
 // - `offset` is an `int`
+// - `aligned` is a `bool`
```
jkwak-work, line 448: This is no longer true, is it?

sriramm-nv:
before cleaning up the code, I want to understand if this is generally a  right approach for this problem.
before this change we had:
```
    float4 res = buffer0.Load<float4>(0);
    buffer0.Store(16, res);
```
with this change now we are forcing the user to:
```
    float4 res = buffer0.LoadAligned<float4>(0);
    buffer0.StoreAligned(16, res);
```

and if we use the old code, we will not be generating vectorized load /store.

csyonghe:
I don't understand why this is the case. All loads are now bottlenecked through the ByteAddressBufferLoad(offset, alignment) instruction, and alignment simply means that `offset` is guaranteed to aligned to that if it is not known at compile time.

If it is 1, then it means offset does not have any alignment guarantee. However if offset is known to be a constant, then we will still use that constant for any alignment checks, so optimizations for Load<float4>(0) should still apply.

source/slang/slang-ir-byte-address-legalize.cpp
sriramm-nv: I had some misunderstanding with the expectation. now it should be fixed.

source/slang/slang-emit.cpp
```diff
@@ -542,6 +542,10 @@ Result linkAndOptimizeIR(
 
     validateIRModuleIfEnabled(codeGenContext, irModule);
 
+    // Fold all calls to builtin functions that results in simple integer values.
+    // When postInlining flag is set, this happens for alignOf<T>() function, which
+    // infers the size of T and replaces the use with the IntLiteral.
+    peepholeOptimize(targetProgram, irModule, PeepholeOptimizationOptions::getPostInlining());
```
csyonghe, line 548: I don't think this is needed because we just run it inside the simplifyIR call above.

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -124,22 +124,23 @@ struct ByteAddressBufferLegalizationContext
...
+        auto legalLoad = emitLegalLoad(type, buffer, offset, 0, alignment);
 
         // If it currently possible for the legalization
         // to fail (perhaps because of something else that
         // is invalid in the IR), so we will defensively
         // leave the code along in that case.
         //
-        if(!legalLoad)
-            return;
+if (!legalLoad)
```
csyonghe, line 135: What happened with indentation here?

source/slang/slang-diagnostic-defs.h
```diff
@@ -753,7 +753,7 @@ DIAGNOSTIC(41201, Warning, expectDynamicUniformValue, "value stored at this loca
 DIAGNOSTIC(41202, Error, notEqualBitCastSize, "invalid to bit_cast differently sized types: '$0' with size '$1' casted into '$2' with size '$3'")
 DIAGNOSTIC(41203, Warning, notEqualReinterpretCastSize, "reinterpret<> into not equally sized types: '$0' with size '$1' casted into '$2' with size '$3'")
 
-DIAGNOSTIC(41300, Error, byteAddressBufferUnaligned, "Base offset '$0' of the byteAddressBuffer access is not divisible by the explicitly defined alignment val `$1`")
+DIAGNOSTIC(41300, Error, byteAddressBufferUnaligned, "Unknown Base offset prevents us from performing an aligned byteAddressBuffer access since the explicit alignment '$0' is not divisible by the size of the base type `$1`")
```
csyonghe, line 756: I don't understand this error message. Why is `offset` unknown? We generate the error because it is known.

source/slang/slang-emit.cpp
sriramm-nv: got it, i updated the peephole optimize call within SimplyIR with the getPostInlining() flag set.

source/slang/slang-ir-byte-address-legalize.cpp
sriramm-nv: good catch. probably changed as a mistake

source/slang/slang-diagnostic-defs.h
sriramm-nv: I meant to say, the value IRInst* buffer can be equated to a IntLiteral and requires the user to specify alignment explicitly.
I modified the error message to this. Any better?
```
invalid alignment `$0` specified for the byte address buffer resource with the element size of `$1`
```

source/slang/slang-ir-peephole.cpp
```diff
@@ -250,6 +251,27 @@ struct PeepholeContext : InstPassBase
 
         switch (inst->getOp())
         {
+        case kIROp_AlignOf:
+            // Fold all calls to builtin functions that results in simple integer values.
+            // When postInlining flag is set, this happens for alignOf<T>() function, which
+            // infers the size of T and replaces the use with the IntLiteral.
+            if (isPostInlining && inst->getDataType()->getOp() == kIROp_IntType)
```
csyonghe, line 258: Why do we need to protect this with `isPostInlining`? I see no problems with just running them regardless of pre or post linking. If any of the operands isn't of the right type or isn't a constant, we just skip any changes.

source/slang/slang-ir-peephole.cpp
```diff
@@ -250,6 +251,27 @@ struct PeepholeContext : InstPassBase
...
+            {
+                auto alignOfInst = as<IRAlignOf>(inst);
+                auto baseType = alignOfInst->getBaseOp()->getDataType();
+                if (as<IRParam>(baseType))
```
csyonghe, line 262: Instead of doing this, we should extend `getNaturalSizeAndAlignment` to `tryGetNaturalSizeAndAlignment` that returns SLANG_FAIL if it encounters things it does not understand. And when `tryGetNaturalSizeAndAlignment` fails, we don't apply any changes here.

sriramm-nv: i didn't do this, because  I'm not clear on how to prevent getNaturalSizeAndAlignment from crashing.

sriramm-nv: worked with Jay on modifying this, to not explicitly check for Op param.

sriramm-nv:
@csyonghe , @jkwak-work do you have any more recommendations that I can address?

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -382,11 +416,15 @@ struct ByteAddressBufferLegalizationContext
             //
             if (auto elementCountInst = as<IRIntLit>(vecType->getElementCount()))
             {
-                // Emit an aligned vector load operation when the data (elementCount * elementSize) is divisible
-                // by the offset. Else, fallback to scalarizing the loads.
-                if (m_options.scalarizeVectorLoadStore || checkUnaligned(baseOffset, immediateOffset, vecType->getElementType(), elementCountInst->getValue()))
+                // Emit an aligned vector load operation when using a LoadAligned inst.
+                // Else, fallback to scalarizing the loads.
+                IRSizeAndAlignment elementLayout;
+                SLANG_RETURN_NULL_ON_FAIL(getNaturalSizeAndAlignment(m_targetProgram->getOptionSet(), vecType->getElementType(), &elementLayout));
```
csyonghe, line 422: This call should never fail, right? Because we must have reached here from emitLegalLoad() where the type does have a size.

If so this needs to be a SLANG_RELEASE_ASSERT.

source/slang/slang-ir-peephole.cpp
```diff
@@ -250,6 +250,24 @@ struct PeepholeContext : InstPassBase
...
         switch (inst->getOp())
         {
+        case kIROp_AlignOf:
+            // Fold all calls to alignOf<T>() that returns a simple integer value.
+            if (targetProgram && inst->getDataType()->getOp() == kIROp_IntType)
+            {
+                IRSizeAndAlignment sizeAlignment;
+                auto alignOfInst = as<IRAlignOf>(inst);
+                auto baseType = alignOfInst->getBaseOp()->getDataType();
+                getNaturalSizeAndAlignment(targetProgram->getOptionSet(), baseType, &sizeAlignment);
```
csyonghe, line 260: If this failed, should break.

source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -124,22 +125,23 @@ struct ByteAddressBufferLegalizationContext
...
+        if (!legalLoad)
             return;
 
-        // If we were able to generate a legal load operation,
-        // then the value it yields can be used to fully
-        // replace the previous illegal load.
-        //
-        load->replaceUsesWith(legalLoad);
-        load->removeAndDeallocate();
+// If we were able to generate a legal load operation,
```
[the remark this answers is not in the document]
sriramm-nv: fixed it

source/slang/slang-ir-byte-address-legalize.cpp
sriramm-nv: updated the exit code

source/slang/slang-ir-peephole.cpp
sriramm-nv: the subsequent check helps with that I guess, but I added a conditional check around getNaturalSizeAndAlignment

source/slang/slang-ir-peephole.cpp
```diff
@@ -257,12 +257,14 @@ struct PeepholeContext : InstPassBase
                 if (!targetProgram)
                     break;
 
+                // Save the alignment information and exit early if it is invalid
                 IRSizeAndAlignment sizeAlignment;
                 auto alignOfInst = as<IRAlignOf>(inst);
                 auto baseType = alignOfInst->getBaseOp()->getDataType();
-                getNaturalSizeAndAlignment(targetProgram->getOptionSet(), baseType, &sizeAlignment);
-                if (sizeAlignment.size == 0)
+                if (getNaturalSizeAndAlignment(targetProgram->getOptionSet(), baseType, &sizeAlignment) ||
```
csyonghe, line 264: should be:
```
if (SLANG_FAILED(getNaturalSizeAndAlignment(targetProgram->getOptionSet(), baseType, &sizeAlignment))
    break;
if (sizeAlignment.size == 0)
    break;
```

source/slang/hlsl.meta.slang
```diff
@@ -108,7 +108,33 @@ struct ByteAddressBuffer
...
         default:
-            return __byteAddressBufferLoad<uint>(this, location);
+            return __byteAddressBufferLoad<uint>(this, location, 0);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint Load(int location, int alignment)
```
csyonghe, line 118: This overload is not necessary. No alignment needed for loading a single uint value.
We only need Load(int location, int alignment) for the generic version `Load<T>`.

source/slang/hlsl.meta.slang
```diff
@@ -108,7 +108,33 @@ struct ByteAddressBuffer
...
+        case hlsl: __intrinsic_asm ".Load";
+        default:
+            return __byteAddressBufferLoad<uint>(this, location, alignment);
+        }
+    }
+
+    [__readNone]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, byteaddressbuffer)]
+    uint LoadAligned(int location)
```
csyonghe, line 131: Again, this overload is not necessary. Alignment has no meaning if we are just loading a single uint.

source/slang/hlsl.meta.slang
```diff
@@ -2898,7 +3016,33 @@ struct $(item.name)
...
         default:
-            return __byteAddressBufferLoad<uint>(this, location);
+            return __byteAddressBufferLoad<uint>(this, location, 0);
+        }
+    }
+
+    [__NoSideEffect]
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
+    uint Load(int location, int alignment)
```
csyonghe, line 3026: Not needed.

source/slang/hlsl.meta.slang
```diff
@@ -2898,7 +3016,33 @@ struct $(item.name)
...
+    {
+        __target_switch
+        {
+        case hlsl: __intrinsic_asm ".Load";
+        default:
+            return __byteAddressBufferLoad<uint>(this, location, alignment);
+        }
+    }
+
+    [__NoSideEffect]
```
csyonghe, line 3036: Not needed.

source/slang/hlsl.meta.slang
```diff
@@ -3806,15 +4045,38 @@ ${{{{
 
     [ForceInline]
     [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
-    void Store(
-        uint address,
-        uint value)
+    void Store(uint address, uint value)
```
csyonghe, line 4048: Not needed.

source/slang/hlsl.meta.slang
```diff
@@ -3806,15 +4045,38 @@ ${{{{
...
+        {
+        case hlsl: __intrinsic_asm ".Store";
+        default:
+            __byteAddressBufferStore(this, address, 0, value);
+        }
+    }
+
+
+    [ForceInline]
+    [require(cpp_cuda_glsl_hlsl_spirv, byteaddressbuffer_rw)]
```
csyonghe, line 4060: Not needed.

csyonghe (approved):
Looks good except for the unnecessary single word overloads.

sriramm-nv:
@csyonghe thanks for making a thorough pass on this PR, and iron out several issues. This definitely helps in understanding best practices for developing in slang codebase.

csyonghe (approved):
Looks good!

PR #4069, opened 2024-05-01
Fix compile failures when using debug symbol.

### LINKED ISSUE #4070: Enabling debug symbol can cause compilation failure.
We should enable debug symbol in our CI to capture these issues.

csyonghe:
This change is to prevent generating debugValue inst on types that may cause complications during future lowering passes to avoid crashing the compiler.

Currently we are avoiding generating debug inst for existential typed values, because the existence of the debug value inst can lead to generation of dynamic dispatch logic. We don't want debug symbols to trigger that because it will result in slower code and triggers crashes due to corner cases that cannot be handled by our dynamic dispatch passes.

Fixes #4070.


jkwak-work (approved):
It looks straightforward.
But somehow github diff view is failing to show what is actually changed.

PR #4073, opened 2024-05-01
Copy default target's optionSet to code-gen target's optionSet

kaizhangNV:
In current implementation, the some options will be to added to the target that is only specified by command line "-target". But if user specifies the target by just using slang API, e.g. 'spAddCodeGenTarget', those options will be missed.

To fix the problem, we copy the default target's options to the code-gen target's option set. The default target will only be useful when there is no target specified in the command line.

### DIFF
--- a/source/slang/slang-compiler.h
+++ b/source/slang/slang-compiler.h
@@ -2835,6 +2835,8 @@ namespace Slang
         };
         Dictionary<TargetRequest*, RefPtr<TargetInfo>> m_targetInfos;
 
+        CompilerOptionSet m_optionSetForDefaultTarget;
+
         CompilerOptionSet& getTargetOptionSet(TargetRequest* req);
 
         CompilerOptionSet& getTargetOptionSet(Index targetIndex);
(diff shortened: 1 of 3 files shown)


csyonghe (approved):

source/slang/slang-compiler.h
```diff
@@ -2835,6 +2835,8 @@ namespace Slang
         };
         Dictionary<TargetRequest*, RefPtr<TargetInfo>> m_targetInfos;
 
+        CompilerOptionSet m_optionSetForDefaultTarget;
```
csyonghe, line 2838: Why can't we use m_defaultTarget.optionSet?

kaizhangNV: We can use it, but still we can only copy it. It's life cycle will be end after parseOptions() call.

kaizhangNV: Update the change by using copy the optionSet instead of call inheritFrom().

csyonghe: I see.

csyonghe (approved):

PR #4080, opened 2024-05-02
Fix fmod behavior targetting GLSL and SPIR-V

jkwak-work:
The default implementation of fmod was doing "Modulo" operation when "fmod" in HLSL should do "remainder" operation.

### DIFF
--- a/tests/cross-compile/fmod.slang
+++ b/tests/cross-compile/fmod.slang
@@ -4,7 +4,8 @@
 // expected output on Vulkan/GLSL.
 
 //TEST(compute):COMPARE_COMPUTE:-dx11 -compute -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj
+//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj -emit-spirv-via-glsl
+//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj -emit-spirv-directly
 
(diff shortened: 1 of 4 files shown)


csyonghe:
Looks good to me but I don't know why the test is failing.

jkwak-work:
I made a mistake on the following lines and it is addressed.
```
    case glsl:
        // GLSL doesn't have a function for remainder.
        if (x < T(0))
            __intrinsic_asm "-mod(-$0,abs($1))";
        else
            __intrinsic_asm "mod($0,abs($1))";
```
When using `__intrinsic_asm`, the other part of implementations are not emitted, which ignored "if (x < T(0))".
I am not sure if it is an intentional design or a bug.
I am using "operator?" to workaround the limitation.

tests/cross-compile/fmod.slang
```diff
@@ -4,7 +4,8 @@
 // expected output on Vulkan/GLSL.
 
 //TEST(compute):COMPARE_COMPUTE:-dx11 -compute -shaderobj
-//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj
+//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj -emit-spirv-via-glsl
```
csyonghe, line 7: Just `//TEST(compute):COMPARE_COMPUTE:-vk -compute -shaderobj`.
CI will run with both config.

jkwak-work: Right. But I was little confused when my local tests are all passing and only CI was failing.

PR #4088, opened 2024-05-02
Slang: update pointer related documentation

### LINKED ISSUE #3970: Pointer support requires clarification in user-guide
https://shader-slang.com/slang/user-guide/conventional-features.html indicates that pointers are not supported.

> Note
> Slang does not currently support pointer types as in C/C++. Pointers cannot be implemented robustly and completely on many of the target platforms Slang currently supports.

Though there is also a section describing support for pointers in SPIRV and Cuda:
https://shader-slang.com/slang/user-guide/convenience-features.html#pointers

The documentation should be updated, probably removing the statement about pointers not being supported.

I believe the section in convenience-features should be marked "(limited)" similarly to the inheritance section below it, along with a "Limitations" subsection explaining the limitations of using pointers.  As I understand, pointers work for global memory, but not shared or local memory.  That should be clarified.

cheneym2:
Slang does have some support for pointers.  Remove an outdated comment stating the contratry, and update the section that describes pointer support to also list some relevant limitations.

Fixes #3970


docs/user-guide/03-convenience-features.md
```diff
@@ -375,8 +375,11 @@ int test(MyType* pObj)
 
 Pointer types can also be specified using the generic syntax: `Ptr<MyType>` is equivalent to `MyType*`.
 
-> #### Note
-> Slang currently does not support pointers to immutable values, i.e. `const T*`.
+### Limitations
+
+- Slang supports pointers to global memory, but not shared or local memory.
```
csyonghe, line 380: We also don't support producing `coherent` load/stores, or specifying custom alignment.

csyonghe: We can add that "For example, it is invalid to take the pointer of a local variable."

docs/user-guide/02-conventional-features.md
```diff
@@ -17,10 +17,6 @@ Types
 
 Slang supports conventional shading language types including scalars, vectors, matrices, arrays, structures, enumerations, and resources.
 
-> #### Note ####
```
csyonghe, line 20: We can add a paragraph to mention that Slang has limited support for pointers when targeting platforms with native pointer support.

cheneym2: ack

docs/user-guide/03-convenience-features.md
cheneym2: Just noticed this.  I added an extra bit explaining the bullet.

csyonghe (approved):

docs/user-guide/03-convenience-features.md
```diff
@@ -361,7 +361,7 @@ float4 myPackedVector = reinterpret<float4>(myVal);
 
 `reinterpret` can pack any type into any other type as long as the target type is no smaller than the source type.
 
-## Pointers
+## Pointers (limited)
```
jkwak-work, line 364: @cheneym2 , I think the link on the left side is broken when you added " (limited)".
"doc/user-guide/toc.html" had to be modified too.

docs/user-guide/03-convenience-features.md
```diff
@@ -361,7 +361,7 @@ float4 myPackedVector = reinterpret<float4>(myVal);
...
 
 Slang supports pointers when generating code for SPIRV, C++ and CUDA targets. The syntax for pointers is similar to C, with the exception that operator `.` can also be used to dereference a member from a pointer. For example:
 ```csharp
```
jkwak-work, line 367: The example code below doesn't work.
The variable name `pNext` is declared twice.

docs/user-guide/03-convenience-features.md
cheneym2: Is toc.html generated or manually edited?

Issue #4075, opened 2024-05-01
Fix autodiff test failures after enabling debug symbol.


csyonghe:
The following tests are failing with PR #4070:

tests/autodiff/struct-this-parameter.slang
tests/autodiff/was/warped-sampling-1d.slang

These two tests are failing after enabling debug symbol, and is currenlty worked around by add `-g0` to test options.
We need to investigate why the tests is producing HLSL with out of bound access when debug symbol is enabled.

When we enable debug symbols, we will insert `IRDebugValue` isnts that are considered to have side effects, thus prevent optimization of loops. This for some reason leads to us generating backward propagation functions that access out-of-bounds in arrays.

PR #4090, opened 2024-05-02
Fix unzipping logic for inout non-diff parameters and adjust tests

saipraveenb25:
+ Removed `-g0` from `struct-this-parameter.slang` test. Works correctly with the new unzipping logic.
+ Removed `-g0` from `was/warped-sampling-1d.slang` test. Works correctly with DX12 & CS_5_1. CS_5_0 appears to run into an FXC compiler bug with detecting infinite loops where there don't appear to be any.

Fixes issue #4075


source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -372,10 +372,30 @@ struct DiffUnzipPass
             }
             else
             {
-                // For non differentiable arguments, we can simply pass the argument as is
-                // if this isn't a `out` parameter, in which case it is removed from propagate call.
-                if (!as<IROutType>(arg->getDataType()))
-                    diffArgs.add(arg);
+                // Ignore `out` parameters since these are removed from propagate call.
+                if (!as<IROutType>(resolvedPrimalFuncType->getParamType(ii)))
```
csyonghe, line 376: General code style: avoid unnecessary nesting levels if possible. In this case, we can use three flat if-else cases.

csyonghe (approved):

source/slang/slang-ir-autodiff-unzip.h
```diff
@@ -372,29 +372,31 @@ struct DiffUnzipPass
...
-                    else
-                    {
-                        // Pure 'in' type. Simply re-use the original argument inst.
-                        diffArgs.add(arg);
-                    }
+                    diffArgs.add(tempPrimalVar);
+                }
+                else
+                {
+                    // For pure 'in' type. Simply re-use the original argument inst.
```
csyonghe, line 393: what if it is `out` type?

saipraveenb25: In this case, it looks like the old out-type detection case was never hit (because it was testing the arg-type and not the param type), and instead we had logic in transposition to discard the arguments corresponding to `out` types. 

If we discard the out type args at the unzipping step, it seems to trigger some argument count equality checks. I think the right place to be doing this is at the transposition step anyway.

csyonghe: Sounds good.

csyonghe (approved):

PR #4094, opened 2024-05-02
Remove target_intrinsic from `min(int,int)`

### LINKED ISSUE #3906: Remove most uses of `__target_intrinsic` and `__specialize_for_target`

### LINKED ISSUE #4093: Use SMin/UMin for SPIR-V target
Falcor unit test is failing when "min(int,int)" function uses SMin/UMin for SPIR-V target.
```
[__readNone]
[require(cpp_cuda_glsl_hlsl_spirv, sm_2_0_GLSL_140)]
T min(T x, T y)
{
    __target_switch
    {
    case hlsl: __intrinsic_asm "min";
    case glsl: __intrinsic_asm "min";
    case cuda: __intrinsic_asm "$P_min($0, $1)";
    case cpp: __intrinsic_asm "$P_min($0, $1)";
#if 0 // <==== THIS NEEDS TO BE ENABLED.
    case spirv:
        if (__isSignedInt<T>())
            return spirv_asm {
                OpExtInst $$T result glsl450 SMin $x $y
            };
        else
            return spirv_asm {
                OpExtInst $$T result glsl450 UMin $x $y
            };
#endif
    default:
        return (x < y) ? x : y;
    }
}
```
When the code above is enabled, CI fails on the following Falcor test, which appears to be an application crash.
```
[----------] 12 tests from SampleGeneratorTests.cpp
[ RUN      ] SampleGeneratorTests.cpp:SampleGenerator_HeitzBelcour (D3D12)
[       OK ] SampleGeneratorTests.cpp:SampleGenerator_HeitzBelcour (D3D12) (3494 ms)
[ RUN      ] SampleGeneratorTests.cpp:SampleGenerator_HeitzBelcour (Vulkan)
[       OK ] SampleGeneratorTests.cpp:SampleGenerator_HeitzBelcour (Vulkan) (3566 ms)
[ RUN      ] SampleGeneratorTests.cpp:SampleGenerator_LeapedHalton (D3D12)
[       OK ] SampleGeneratorTests.cpp:SampleGenerator_LeapedHalton (D3D12) (1067 ms)
[ RUN      ] SampleGeneratorTests.cpp:SampleGenerator_LeapedHalton (Vulkan)
```
It can be locally reproduced only when a debug build of slang.dll is used.

jkwak-work:
Closes #3906
Closes #4093

The last remaining function that uses `__target_intrinsic` is `min(int,int)`.

For some reason, Falcor-unit-test fails on CI when this change is applied. It can be reproduced on a local machine if you use a debug build of slang.dll

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -8389,14 +8389,32 @@ matrix<T, N, M> max(matrix<T, N, M> x, matrix<T, N, M> y)
 
 // minimum
 __generic<T : __BuiltinIntegerType>
-__target_intrinsic(hlsl)
-__target_intrinsic(glsl)
-__target_intrinsic(cuda, "$P_min($0, $1)")
-__target_intrinsic(cpp, "$P_min($0, $1)")
-__target_intrinsic(spirv, "OpExtInst resultType resultId glsl450 fus(FMin, UMin, SMin) _0 _1")
 [__readNone]
 [require(cpp_cuda_glsl_hlsl_spirv, sm_2_0_GLSL_140)]
-T min(T x, T y);
+T min(T x, T y)
+{
+    __target_switch
+    {
+    case hlsl: __intrinsic_asm "min";
+    case glsl: __intrinsic_asm "min";
+    case cuda: __intrinsic_asm "$P_min($0, $1)";
+    case cpp: __intrinsic_asm "$P_min($0, $1)";
+#if 0
+    case spirv:
+        if (__isSignedInt<T>())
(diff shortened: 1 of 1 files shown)


jkwak-work:
A proper fix is submitted. Closing this PR

PR #4099, opened 2024-05-03
Don't bottleneck Wave intrinsics through `WaveMask*` for spirv.

### LINKED ISSUE #4097: Incorrect WaveMatch result with SPIR/V
According to the "spec" the `WaveMatch` intrinsic should return `0` for any lane that is inactive or beyond `WaveGetLaneCount`. This is what I see when running a test shader compiled to HLSL. When generating SPIR/V, at least on NVIDIA the returned mask `uint4` repeats the value of `.x` across all other components instead of them being `0`.

from https://microsoft.github.io/DirectX-Specs/d3d/HLSL_ShaderModel6_5.html :

> The return value is a uint4 representing a 128b bit mask which identifies lanes in the current wave matching current lane’s value of val. Bits in the mask corresponding to inactive lanes, or at positions beyond current implementation’s wave width, will contribute 0’s. Bits in the mask corresponding to active lanes which match the value of val in the current lane will be set to 1. The bit in the mask corresponding to the current lane will always be set to 1.

csyonghe:
Fixes #4097.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -12359,8 +12359,7 @@ uint4 WaveMatch(T value)
     __target_switch
     {
     case hlsl: __intrinsic_asm "WaveMatch";
-    case glsl: __intrinsic_asm "subgroupPartitionNV($1)";
-    case cuda: __intrinsic_asm "_waveMatchScalar($0, $1)";
+    case glsl: __intrinsic_asm "subgroupPartitionNV($0)";
     case spirv:
         return spirv_asm
         {
@@ -12380,8 +12379,7 @@ uint4 WaveMatch(vector<T,N> value)
     __target_switch
     {
     case hlsl: __intrinsic_asm "WaveMatch";
-    case glsl: __intrinsic_asm "subgroupPartitionNV($1)";
-    case cuda: __intrinsic_asm "_waveMatchScalar($0, $1)";
+    case glsl: __intrinsic_asm "subgroupPartitionNV($0)";
     case spirv:
         return spirv_asm
         {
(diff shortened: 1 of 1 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -12305,6 +12359,14 @@ uint4 WaveMatch(T value)
     __target_switch
     {
     case hlsl: __intrinsic_asm "WaveMatch";
+    case glsl: __intrinsic_asm "subgroupPartitionNV($0)";
```
sriramm-nv, line 12362: is subgroupPartitionNV part of HLSL as well?

csyonghe: This is implementing the intrinsic for the `case glsl` case, so it is for GLSL.

sriramm-nv: oh i thought HLSL was pass through case.

sriramm-nv:
looks good to me except for that one line.

source/slang/hlsl.meta.slang
csyonghe: __intrinsic_asm is like a `return`, and there is no passthrough for __target_switch.

source/slang/hlsl.meta.slang
```diff
@@ -12329,6 +12399,14 @@ uint4 WaveMatch(matrix<T,N,M> value)
     __target_switch
     {
     case hlsl: __intrinsic_asm "WaveMatch";
+    case glsl:
+    case cuda:
+    case spirv:
+        uint4 result = uint4(0xFFFFFFFF);
+        [ForceUnroll]
+        for (int i = 0; i < N; i++)
+            result &= WaveMatch(value[0]);
```
jkwak-work, line 12408: Shouldn't this be like this?
```
result &= WaveMatch(value[i]);
```

csyonghe: You are right. Thanks for catching this, will fix in a new PR.

PR #4100, opened 2024-05-03
Fix `Ptr::__subscript` to accept any integer index.

### LINKED ISSUE #4029: uint64_t invalid pointer indexer?
I've just started with slang not too long ago, and ran this:

```
struct ElementWiseInputData<T : IArithmetic>{
    T* lhs; 
    T* rhs; 
    T* out; 
    uint64_t size; 
}

void add_elements<T : IArithmetic>(ElementWiseInputData<T> input_data, uint64_t index){
    input_data.out[index] = input_data.lhs[index] + input_data.rhs[index];
}


[[vk::push_constant]] ElementWiseInputData<uint32_t> array_info; 

[shader("compute")]
[numthreads(1024,1,1)]
void add_main(uint3 threadId: SV_DispatchThreadID) {
    // test<uint> t; 
    // t.x = 3; 
     ElementWiseInputData<uint32_t> temp = array_info;  // for some reason becomes a ConstantBuffer, and can't be directly used with out copying? 
    if (threadId.x < array_info.size) {
        let index = threadId.x; 
        array_info.out[index] = array_info.lhs[index] + array_info.rhs[index];
        add_elements(temp, uint64_t(index)); //duplicated for testing reasons. 
    }    
}
```

and everything ran fine. But i'm getting some worrying warnings: 

```
Path\cmake-build-debug\examples.exe
Pathl/examples/shaders/test_shader.slang(19): warning 30081: implicit conversion from 'uint64_t' to 'int' is not recommended
    input_data.out[index] = input_data.lhs[index] + input_data.rhs[index];
                   ^~~~~
Path/examples/shaders/test_shader.slang(19): warning 30081: implicit conversion from 'uint64_t' to 'int' is not recommended
    input_data.out[index] = input_data.lhs[index] + input_data.rhs[index];
                                           ^~~~~
Path/examples/shaders/test_shader.slang(19): warning 30081: implicit conversion from 'uint64_t' to 'int' is not recommended
    input_data.out[index] = input_data.lhs[index] + input_data.rhs[index];
                                                                   ^~~~~

```
We have code we were going to transfer over to slang which most definitely needs indexes over 2^31 and 2^32. In vulkan GLSL this is not an issue, and heck, I want to say this isn't even an issue in HLSL, at least we haven't noticed it there.   It would maybe be possible to get around this by adding to the pointer itself? but I don't think we want to have to wrap pointers themselves here.

It looks like the offending line is here? 

https://github.com/shader-slang/slang/blob/52dcb5bd44aa15f07826062c53fae344d55959e9/source/slang/core.meta.slang#L697

But the underlying offsets support both int and int64:?

```
__intrinsic_op($(kIROp_GetOffsetPtr))
Ptr<T> __getOffsetPtr<T>(Ptr<T> ptr, int index);

__intrinsic_op($(kIROp_GetOffsetPtr))
Ptr<T> __getOffsetPtr<T>(Ptr<T> ptr, int64_t index); 
```

and should in theory support uint and uint64 as well with addtional functions?

csyonghe:
Fixes #4029.

The main change here is to make `__subscript` a generic on `T:__BuiltinIntegerType`, and allow parser to accept generic __subscript decls.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -694,7 +694,8 @@ struct Ptr
     __intrinsic_op($(kIROp_CastIntToPtr))
     __init(int64_t val);
 
-    __subscript(int index) -> T
+    __generic<TInt : __BuiltinIntegerType>
+    __subscript(TInt index) -> T
     {
         __intrinsic_op($(kIROp_GetOffsetPtr))
         ref;
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -10050,6 +10050,10 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         {
             return ensureDecl(context, typedefDecl);
         }
+        else if (auto subscriptDecl = as<SubscriptDecl>(genDecl->inner))
+        {
+            return ensureDecl(context, subscriptDecl);
+        }
         SLANG_RELEASE_ASSERT(false);
         UNREACHABLE_RETURN(LoweredValInfo());
     }
--- a/source/slang/slang-parser.cpp
+++ b/source/slang/slang-parser.cpp
@@ -4311,6 +4311,7 @@ namespace Slang
             case ASTNodeType::TypeAliasDecl:
             case ASTNodeType::TypeDefDecl:
             case ASTNodeType::ExtensionDecl:
+            case ASTNodeType::SubscriptDecl:
                 return true;
             default:
                 return false;
--- a/tests/spirv/ptr-subscript.slang
+++ b/tests/spirv/ptr-subscript.slang
@@ -1,0 +1,12 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv -entry main -stage compute -emit-spirv-directly
+
+// CHECK: OpEntryPoint
+
+ConstantBuffer<Ptr<int>> cbPtr;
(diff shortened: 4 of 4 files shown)


ArielG-NV (approved):
Looks good.

jkwak-work:
Looks good to me but wouldn't this allow little more than usual? I think __BuoltinIntegerType includes like a memory pointer type too.

csyonghe:
I think you are referring to IntPtr type? Similar to the samely named type in .net framework, IntPtr means pointer-sized integer type, which is int64 in most cases.

jkwak-work:
Right, that's what I was thinking.
Allowing Ptr<T> to be a type that can go to the subscribe operator feel like this in C++
```
int main()
{
        int var = 0;
        int *ptr = &var;
        int arr[4];
        arr[ptr] = 0; // error: invalid types ‘int [4][int*]’ for array subscript".
        return 0;
}
```
Correct me if I misunderstood.
But allowing a pointer as an array index should be an error if that's what this change does.

csyonghe:
Int* is not an BuiltinIntegerType.

IntPtr is. IntPtr is not int*. It is an int whose size is the same as a pointer.

On x86, IntPtr == Int32.
On x64, IntPtr == Int64.

csyonghe:
See https://manpages.ubuntu.com/manpages/lunar/man3/intptr_t.3type.html

jkwak-work:
Oh. I see. Thanks for clarifying it for me.

PR #4106, opened 2024-05-04
Fix macos release script.

csyonghe:


fknfilewalker:
macos bins still miss slang-rt

PR #4114, opened 2024-05-06
Delete `wrap-global-context` pass.

csyonghe:
The pass was added for the metal backend without realizing that the existing `explicit-global-context` does 99% of the job. Instead of duplicating the logic in a different pass for metal, we extend same explicit-global-context pass to work for metal.


jkwak-work (approved):
Looks good to me.
It is interesting that almost same implementation already existed.

PR #4116, opened 2024-05-06
Support groupshared variables for Metal.

### LINKED ISSUE #4045: Metal: groupshared variable.
In HLSL, groupshared variables are defined as global variables.
In metal, global variables are not allowed.

We currently have a pass that packs all global variables into a context. However, this is assuming all variables are threadlocal.

If we have a global variable that is groupshared, we can't just insert a field at `threadgroup` space directly into the context type, as this is disallowed by the Metal spec.

Instead, we should define all groupshared variables in entry point functions direclty as local variables, and pass a pointer to the variable in the global context. 

For example, if we have:
```
groupshared float myArray[16];
void computeMain()
{
     f();
}
void f()
{
     myArray[0] = 1;
}
```
should become:
```
struct Context
{
    array<float,16> threadgroup* myArray;
}
void computeMain()
{
     array<float,16> threadgroup myArray;
      Context ctx = {&myArray};
      f(&ctx);
}
void f(Context threadlocal* ctx)
{
    ctx->myArray[0] = 1;
}
```

csyonghe:
This change extends the explicit context pass to handle groupshared variables differently, such that they are allocated first in the entry point as a local variable, and then its pointers are passed in the global context object.

For example,
```
groupshared int data[3];
void test()
{
   int a = data[2];
}
void main()
{
    test();
}
```

Is emitted as:
```
struct KernelContext
{
    array<int,3> threadgroup* data;
};
void test(KernelContext thread* ctx)
{
    int a =  (*ctx->data)[[2];
}
void main()
{
    threadgroup array<int,3> data;
    KernelContext ctx;
    ctx.data = &data;
    test(&ctx);
}
```

Closes #4045.

### DIFF
--- a/source/slang/slang-emit-metal.cpp
+++ b/source/slang/slang-emit-metal.cpp
@@ -494,7 +494,7 @@ void MetalSourceEmitter::emitSimpleTypeImpl(IRType* type)
         case kIROp_ParameterBlockType:
         case kIROp_ConstantBufferType:
         {
-            emitType((IRType*)type->getOperand(0));
+            emitSimpleTypeImpl((IRType*)type->getOperand(0));
             m_writer->emit(" constant*");
             return;
         }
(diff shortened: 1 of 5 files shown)


source/slang/slang-emit-metal.cpp
```diff
@@ -494,7 +494,7 @@ void MetalSourceEmitter::emitSimpleTypeImpl(IRType* type)
         case kIROp_ParameterBlockType:
```
jkwak-work, line 494: I guess the answer is yes but is it correct to give "constant" keyword for parameter block type?

jkwak-work (approved):
It looks good to me.
But I have one more question.
When a threadlocal variable is declared as a function local variable scope, it isn't in the stack memory but in each thread memory?
Is it how Metal works?

csyonghe:
It means allocating the array from groupshared memory.

If you think of a normal int arr[5] as
arr = alloca(sizeof_int * 5)

Then threadgroup int arr[5] is just

arr = alloca_from_groupshared(sizeof_int*5)


If you look at the llvm ir generated from this code, it is translated back into a global group shared array by the metal compiler.

source/slang/slang-emit-metal.cpp
csyonghe: IIRC, according to the metal spec, constant memory space is what maps to a ConstantBuffer in d3d, or a uniform buffer in Vulkan.

PR #4118, opened 2024-05-07
Support Metal math functions

### LINKED ISSUE #4024: Metal: add math intrinsics.
This include vector and matrix algebra and elementary functions.

jkwak-work:
Closes #4024

Note that Metal document says Metal doesn't support "double" type;
"Metal does not support the double, long long, unsigned long long, and long double data types."

According to Metal document, math functions are not defined for integer types.

That leaves only two types to test: half and float.

As a code clean up, `__floatCast` is replaced with `__realCast`.
But I had to add a new signature that can convert from integer to float.

Some of GLSL functions are moved to hlsl.meta.slang.
For those functions, there isn't builtin functions for HLSL but there are for GLSL and Metal.

I need to revisit later why "nextafter(T,T)" is not working.
The error message was "undefined identifier" as if the functions doesn't exist at all.
The document says it is supported from Metal version 3.1. and my compiler is 3.2.
This isn't a big problem because there isn't any existing Slang function that maps to "nextafter".


source/slang/hlsl.meta.slang
```diff
@@ -9636,12 +10548,16 @@ void sincos(T x, out T s, out T c)
...
-[require(cpp_cuda_glsl_hlsl_spirv, sm_2_0_GLSL_140)]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_metal_spirv, sm_2_0_GLSL_140)]
 void sincos(vector<T,N> x, out vector<T,N> s, out vector<T,N> c)
 {
     __target_switch
     {
     case hlsl: __intrinsic_asm "sincos";
+    case metal:
+        s = __sincos_metal(x, c);
```
csyonghe, line 10559: Can be 
*($1)=sincos($0,*($2))

csyonghe (dismissed):
Looks good, I think in the future PRs we can skip the metal only intrinsics and focus on mapping existing HLSL intrinsics to metal.

source/slang/hlsl.meta.slang
```diff
@@ -8883,6 +9648,44 @@ matrix<T,R,C> mul(matrix<T,R,N> left, matrix<T,N,C> right)
...
+    __target_switch
+    {
+    case metal: __intrinsic_asm "nextafter";
+    default:
+        if (isnan(x)) return x;
+        if (isnan(y)) return y;
+        if (x == y) return y;
+        T dir = ((x < y) ? T(1) : -T(1));
+        if (T is half)
+            return x + dir * __realCast<T>(bit_cast<half>(uint16_t(1)));
```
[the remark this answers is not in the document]
jkwak-work: Good idea.
I meant to use `dir` as just sign value but I can fuse the sign and the value at the compile time.

source/slang/hlsl.meta.slang
jkwak-work: Oh! I will give a try.

source/slang/hlsl.meta.slang
```diff
@@ -6114,6 +6460,36 @@ matrix<T,N,M> exp2(matrix<T,N,M> x)
...
+__generic<T : __BuiltinFloatingPointType>
+[__readNone]
+[require(cpp_cuda_glsl_hlsl_metal_spirv, sm_2_0_GLSL_140)]
+T exp10(T x)
+{
+    __target_switch
+    {
+    case metal: __intrinsic_asm "exp10";
+    default:
+        const T ln10 = T(2.30258509299); // ln(10)
```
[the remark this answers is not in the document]
jkwak-work: Right. I will update it.

source/slang/core.meta.slang
```diff
@@ -421,6 +421,8 @@ __generic<T, let N : int> __intrinsic_op(select) vector<T,N> select(vector<bool,
 // Allow real-number types to be cast into each other
 __intrinsic_op($(kIROp_FloatCast))
     T __realCast<T : __BuiltinRealType, U : __BuiltinRealType>(U val);
+__intrinsic_op($(kIROp_FloatCast))
```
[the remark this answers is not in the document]
jkwak-work: I will change it.

csyonghe:
Note that our downstream metal compiler implementation always invokes metal.exe with std=metal_macos2.3 

We need to derive this from the profile level we are compiling instead of a hard coded setting. This is probably why nextafter failed.

source/slang/hlsl.meta.slang
jkwak-work: This should work but it is currently not working.
The error message is
```
metal 32023.155: diff.meta.slang(1979): error :  unknown type name 'out'
void _sincos_impl_0(float x_0, out float s_0, out float c_0)
```
It sounds like we cannot use `out` in Metal.
I guess we should emit `&` istead of `out`.

csyonghe: Yeah it needs to be 
T thread&

All pointers and references need an address space. We can use thread now but we will need a pass that derives and specializes functions with address spaces.

source/slang/slang-emit-metal.cpp
```diff
@@ -298,24 +298,26 @@ bool MetalSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inO
...
-    case kIROp_UInt16Type: m_writer->emit("ushort"); break;
-    case kIROp_IntType:    m_writer->emit("int");    break;
-    case kIROp_UIntType:   m_writer->emit("uint");   break;
-    case kIROp_Int64Type:  m_writer->emit("long");   break;
-    case kIROp_UInt64Type: m_writer->emit("ulong");  break;
-    }
+    emitSimpleTypeImpl(elementType);
 
-    if (elementCount > 1)
+    switch (elementType->getOp())
```
csyonghe, line 303: I think we can get rid of this `switch` because it is not helping us achieving anything more?

jkwak-work: I agree to that.
I will try to remove  it next time when I need to modify this function.

Today while I was reading the document, I learned that Metal supports an expression of `vec<T,N>`. I think that will be more preferable syntax. There is also matrix<T,C,R> as well. I guess Metal is column major.

csyonghe: If they are mat<T,A, B> and are A major, the we map Slang::row to A and Slang::col to B. This means that similar to Vulkan, a slang row is a metal column.

csyonghe: I need to double check their definition of matrix vector multiplication and matrix constructor to make sure all these makes sense. But I expect this is likely going to be the case.

PR #4122, opened 2024-05-07
capture/replay: interface implementation 1

kaizhangNV:
- Add global session, filesystem, and session capture interface classes: GlobalSessionCapture for IGlobalSession FileSystemCapture for ISlangFileSystemExt SessionCapture for ISession

- Add environment variables to enable it The 2 variables are SLANG_CAPTURE_LAYER and SLANG_CAPTURE_LOG_LEVEL

    SLANG_CAPTURE_LAYER:

    In slang_createGlobalSession(), after the compiling/loading stdlib,
    we will check the capture environment variable, if it's set to 1,
    we will create a GlobalSessionCapture object and return to user
    code.

    SLANG_CAPTURE_LOG_LEVEL: This is to set the log level, user can
    choose the loglevel to debug. (We can remove this when the feature
    is fully implemented).

- Update premake file to add the capture/replay source folder

### DIFF
--- a/source/slang-capture-replay/capture_utility.cpp
+++ b/source/slang-capture-replay/capture_utility.cpp
@@ -1,0 +1,82 @@
+
+#include <cstring>
+#include <string>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <mutex>
+
+#include "capture_utility.h"
+
+constexpr const char* kCaptureLayerEnvVar = "SLANG_CAPTURE_LAYER";
+constexpr const char* kCaptureLayerLogLevel = "SLANG_CAPTURE_LOG_LEVEL";
+
+namespace SlangCapture
+{
+    static thread_local unsigned int g_logLevel = LogLevel::Silent;
+
+    static bool getEnvironmentVariable(const char* name, std::string& out)
+    {
+#ifdef _WIN32
+        char* envVar = nullptr;
+        size_t sz = 0;
+        if (_dupenv_s(&envVar, &sz, name) == 0 && envVar != nullptr)
+        {
+            out = envVar;
+            free(envVar);
+        }
+#else
+        if (const char* envVar = std::getenv(name))
+        {
+            out = envVar;
+        }
+#endif
+        return out.empty() == false;
+    }
+
+    bool isCaptureLayerEnabled()
+    {
+        std::string envVarStr;
+        if(getEnvironmentVariable(kCaptureLayerEnvVar, envVarStr))
+        {
+            if (envVarStr == "1")
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    void setLogLevel()
+    {
+        // We only want to set the log level once
+        if (g_logLevel != LogLevel::Silent)
+        {
+            return;
+        }
+
+        std::string envVarStr;
+        if (getEnvironmentVariable(kCaptureLayerLogLevel, envVarStr))
+        {
+            char* end = nullptr;
+            unsigned int logLevel = std::strtol(envVarStr.c_str(), &end, 10);
+            if (end && (*end == 0))
+            {
+                g_logLevel = std::min((unsigned int)(LogLevel::Verbose), logLevel);
+                return;
+            }
+        }
+    }
+
+    void slangCaptureLog(LogLevel logLevel, const char* fmt, ...)
+    {
+        if (logLevel > g_logLevel)
+        {
+            return;
+        }
+
+        va_list args;
+        va_start(args, fmt);
+        vfprintf(stdout, fmt, args);
+        va_end(args);
+    }
+}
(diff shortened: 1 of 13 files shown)


source/slang-capture-replay/slang-session.h
```diff
@@ -0,0 +1,99 @@
...
+            SlangInt conformanceIdOverride,
+            ISlangBlob** outDiagnostics) override;
+        SLANG_NO_THROW SlangResult createCompileRequest(
+            SlangCompileRequest**   outCompileRequest) override;
+        SLANG_NO_THROW SlangInt getLoadedModuleCount() override;
+        SLANG_NO_THROW slang::IModule* getLoadedModule(SlangInt index) override;
+        SLANG_NO_THROW bool isBinaryModuleUpToDate(const char* modulePath, slang::IBlob* binaryModuleBlob) override;
+
+    private:
+        SLANG_FORCE_INLINE slang::ISession* asExternal(SessionCapture* session)
```
jkwak-work, line 91: You probably want to make this `static` member method.

source/slang-capture-replay/slang-session.cpp
```diff
@@ -0,0 +1,196 @@
...
+    {
+        slangCaptureLog(LogLevel::Verbose, "%s\n", "SessionCapture::createTypeConformanceComponentType");
+        SlangResult result = m_actualSession->createTypeConformanceComponentType(type, interfaceType, outConformance, conformanceIdOverride, outDiagnostics);
+        return result;
+    }
+
+    SLANG_NO_THROW SlangResult SessionCapture::createCompileRequest(
+        SlangCompileRequest**   outCompileRequest)
+    {
+        slangCaptureLog(LogLevel::Verbose, "%s\n", "SessionCapture::createCompileRequest");
```
jkwak-work, line 170: You could do without the indirection,
```
slangCaptureLog(LogLevel::Verbose, "SessionCapture::createCompileRequest\n");
```
Or you can avoid any mistakes on the function names by using the `__FUNC__` macro.
```
slangCaptureLog(LogLevel::Verbose, "%s:%d:%s\n", __FILE__, __LINE__, __FUNC__);
```

source/slang-capture-replay/capture_utility.cpp
```diff
@@ -0,0 +1,66 @@
...
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    void setLogLevel()
+    {
+        static std::mutex logLevelMutex;
```
jkwak-work, line 31: I think a mutex locking is an overkill for setting an integer value.
Alternatively we can make the variable thread-local and let each thread have their own variants.

source/slang-capture-replay/capture_utility.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (logLevel > g_logLevel)
+        {
+            return;
+        }
+
+        va_list args;
+        va_start(args, fmt);
+        vfprintf(stdout, fmt, args);
+        va_end(args);
```
jkwak-work, line 64: Are we going to generate multiple files each of which will work for one thread?
Or are we going to generate one big log file that contains all?

jkwak-work (commented):
Looks good to me so far

source/slang-capture-replay/capture_utility.cpp
kaizhangNV: I'm going to generate one log file per thread, so we don't need to handle any thread issues.

source/slang-capture-replay/capture_utility.cpp
kaizhangNV: Yes, thinking about the same thing as well.

jkwak-work (approved):
Looks good to me.

source/slang-capture-replay/slang-session.cpp
csyonghe: Yes, we should be making use of `__FUNC__`.

source/slang-capture-replay/slang-filesystem.h
```diff
@@ -0,0 +1,73 @@
...
+            void* userData) override;
+
+        virtual SLANG_NO_THROW OSPathKind getOSPathKind() override;
+    private:
+        // We should release the actual file system, because if 1) the file
+        // system is provided by user, we will increment the ref-count;
+        // 2) if the file system is created by slang, we will transfer the ownership
+        // from Linkage to FileSystemCapture, in both cases, we should release the
+        // 'm_actualFileSystem'.
+        ISlangFileSystemExt* m_actualFileSystem = nullptr;
```
csyonghe, line 68: This can just be a `ComPtr<ISlangFileSystemExt>` so we don't need to worry about managing ref counts.

csyonghe (dismissed):

source/slang-capture-replay/slang-session.cpp
kaizhangNV: OK, this log call is actually generated by copilot, I will update that.

csyonghe (approved):

PR #4130, opened 2024-05-07
Add github action to ensure PRs are labeled.

csyonghe:


jkwak-work (approved):
Change looks good to me.
But I am not sure if this is a best way to approach.

Maybe we should 't allow any breaking changes to be merged in the first place. If must be submitted, it should have a runtime option to disable it.

I think we can also have a RELEASE_NOTE.md file like README.md that should be modified with the changes if it would have an impact on the end user experience.

csyonghe:
While we should avoid break changes when possible, there will be times where introducing the change is the best way forward, and we should allow that to happen as long as the decision has been made thoughtfully. 

Enforcing the tags is to help our release process so we don't miss out mentioning these changes in our release notes.

jkwak-work:
Perhaps "breaking" might be a strong word. How about "experimental", "internal", "release-blocker", "WIP" or "incomplete".

csyonghe:
The label is intended to convey a strong signal so we don't miss out important changes in our release notes. Anything that could potentially break existing application must be tagged with breaking change label.

csyonghe:
Note that experimental features behind a compiler flag that is not enabled by default in exisiting applications does not count as breaking change.

When we flip the default setting of some compiler option, delete or change legacy API, or change/remove some experiemental language feature, that is when we should tag the PR with breaking change label.

PR #4135, opened 2024-05-08
Slang: Support UTF-8 with Byte Order Markers

### LINKED ISSUE #4046: Slang fails to parse file with BOM
When compiling a file that has a byte order mark, Slang chokes. Happens both with `slangc` and the native API (for example when #include'ing a file with a BOM).

I've attached a simple file with a BOM (.txt because github won't let me upload a .hlsl tile)
[repro.txt](https://github.com/shader-slang/slang/files/15140008/repro.txt)

Observe:
```
>file repro.txt
repro.txt: C source, Unicode text, UTF-8 (with BOM) text, with CRLF line terminators

>slangc -lang hlsl repro.txt
repro.txt(1): error 10000: illegal character (0xEF)
∩╗┐void main()
^
repro.txt(1): error 10000: illegal character (0xBB)
∩╗┐void main()
 ^
repro.txt(1): error 10000: illegal character (0xBF)
∩╗┐void main()
  ^
```

cheneym2:
Slang: Support UTF-8 with Byte Order Markers

Slang APIs are documented as taking UTF-8 encoded shader source,
though it's not explicitly documented whether it is allowed to
include a BOM (Byte Order Marker). This change adds support for
UTF-8 BOM markers by virtue of disposing of BOM data. As a bonus,
UTF-16 input which can cleanly decode to UTF-8 is now also
accepted.

Throwing out the BOM on input is done by leveraging existing
functionality in "determineEncoding()", however a bug exists there
for null-terminated single character input, where the null byte
caused a heuristic to guess UTF-16, even though the null byte
isn't part of the string. The bug in "determineEncoding" is fixed
by removing the buggy heuristic.

The decoding of UTF-16 to UTF-8 is done through an existing
'decode' method. This change fixes a bug in UTF16-LE 'decode'
where it was decoded as if it were Big-Endian.

Adds 3 small tests to ensure the compiler doesn't choke on source
files in UTF-8 (with BOM), UTF16-LE, or UTF16-BE.

Bonus: Fixes a bug in diagnostic reporting where hex values were
incorrectly translated to text, leading to incorrect, possibly
truncated strings.

Fixes #4046


source/compiler-core/slang-source-loc.cpp
```diff
@@ -573,15 +573,38 @@ int SourceFile::calcColumnIndex(int lineIndex, int offset, int tabSize)
...
+    CharEncoding::getEncoding(type)->decode(
+        rawContentBegin + offset,
+        rawContentSize - offset,
+        decodedBuffer);
+    List<uint8_t> decodedBuffer2;
+    for(auto i : decodedBuffer)
+    {
+        decodedBuffer2.add(static_cast<uint8_t>(i));
+    }
+    m_contentBlob = ListBlob::moveCreate(decodedBuffer2);
```
jkwak-work, line 600: I think this should be `StringBlob` from `String`
```
    String decodedBuffer2;
    decodedBuffer2.prepareForAppend(decodedBuffer.getCount());
    for(auto i : decodedBuffer)
    {
        decodedBuffer2.append(i);
    }

    m_contentBlob = StringBlob::create(decodedBuffer2);
```

source/compiler-core/slang-source-loc.cpp
```diff
@@ -573,15 +573,38 @@ int SourceFile::calcColumnIndex(int lineIndex, int offset, int tabSize)
...
+    auto type = CharEncoding::determineEncoding(
+        rawContentBegin,
+        rawContentSize,
+        offset);
+    SLANG_ASSERT(rawContentSize >= offset);
+
+    List<char> decodedBuffer;
+    CharEncoding::getEncoding(type)->decode(
+        rawContentBegin + offset,
+        rawContentSize - offset,
```
jkwak-work, line 593: This line is giving me a compiler warning,
```
D:\sbf\git\slang\cheneym2\source\compiler-core\slang-source-loc.cpp(594,22): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data [D:\sbf\git\slang\cheneym2\build\compiler-core.vcxproj]
```
Please apply an appropriate type casting to avoid the warning.
```
int(rawContentSize - offset),
```

source/compiler-core/slang-source-loc.cpp
```diff
@@ -573,15 +573,38 @@ int SourceFile::calcColumnIndex(int lineIndex, int offset, int tabSize)
...
+        decodedBuffer);
+    List<uint8_t> decodedBuffer2;
```
csyonghe, line 595: It is unfortunate that we need to do a full copy just to convert List<char> to List<uint8_t>. We can probably implement some Move operator on a List to allow us to move the buffer pointer without a copy, admittedly this is likely going to be a UB.

source/core/slang-char-encode.cpp
```diff
@@ -135,25 +135,6 @@ class Utf16CharEncoding : public CharEncoding //UTF16
             return CharEncodeType::UTF16Reversed;
         }
     }
-    else
```
csyonghe, line 138: I think we still want to keep this logic, but to make sure we don't treat trailing null terminator as utf16 string.

source/core/slang-char-encode.cpp
```diff
@@ -92,17 +92,17 @@ class Utf16CharEncoding : public CharEncoding //UTF16
...
 		while (index < length)
 		{
-			const Char32 codePoint = getUnicodePointFromUTF16([&]() -> Byte
-			{
-                if (index < length)
-                    return bytes[index++];
-                else
-                    return Byte(0);
-			});
+			auto readByte = [&]() -> Byte
```
csyonghe, line 95: #Indentation seems off.

source/core/slang-char-encode.cpp
cheneym2: Are you sure?  This code was only used if bytesCount == 0 or 1.  There's not enough data to guess UTF16.  What am I missing?

source/core/slang-char-encode.cpp
cheneym2: Oops, thanks. This file is generally using tabs, and the pre-existing lines had a mixture of spaces and tabs.  I intended to have the newly added code be only tabs to match the local (though not Slang coding style compliant) style, but I missed a line or two.  The next patch revision will be all tabs in this spot.

source/compiler-core/slang-source-loc.cpp
cheneym2: Yes, that seems to have resolved the problems I was seeing. Thanks!

source/compiler-core/slang-source-loc.cpp
cheneym2: Ack

source/compiler-core/slang-source-loc.cpp
cheneym2: Yes, I agree is unfortunate and I imagined there must be some better solution, but it seems to be moot as Jay proposed using String + StringBlob.  Though it still requires the copy, since it's from a List to a String, it's doesn't seem as unnecessary.

source/core/slang-char-encode.cpp
cheneym2: If we keep the UTF16 guessing logic, it will have to be inside the bytesCount >= 2 block above because it makes no sense to only guess UTF16 on strings with length 0 and 1.

cheneym2:
The CI error is in tests/bugs/implicit-cast.slang (vk)

<pre>
error: ERROR:
EXPECTED{{{
2}}}
ACTUAL{{{
0
1
0
0
}}}
</pre>

csyonghe:
Can you reproduce this error locally? I am going to rerun the test.

cheneym2:
I will set up my windows laptop to try reproducing the test error.  I needed to do that anyway, so now's the time.

source/compiler-core/slang-source-loc.cpp
```diff
@@ -573,15 +573,38 @@ int SourceFile::calcColumnIndex(int lineIndex, int offset, int tabSize)
...
+        rawContentSize,
+        offset);
+    SLANG_ASSERT(rawContentSize >= offset);
+
+    List<char> decodedBuffer;
+    CharEncoding::getEncoding(type)->decode(
+        rawContentBegin + offset,
+        int(rawContentSize - offset),
+        decodedBuffer);
+    String decodedBuffer2;
```
csyonghe, line 595: If we are going to create a copy anyways, we can just do:
```
m_contentBlob = RawBlob::create(decidedBuffer.getBuffer(), decodedBuffer.getCount());
```

If we can live with some pointer punning UB (which we already do in many other places in this codebase anyways), we can avoid that copy with:

```
auto capacity = decodedBuffer.getCapacity();
auto count = decodedBuffer.getCount();
auto bufferPtr = (uint8_t*)decodedBuffer.detachBuffer();
List<uint8_t> u8buffer;
u8buffer.attachBuffer(bufferPtr, count, capacity);
m_contentBlob = ListBlob::moveCreate(u8buffer);
```

cheneym2: What do you think of Jay's suggestion to use String and StringBlob?

cheneym2: I don't know what the difference is in effect.

csyonghe: Let's do either option I posted above. The reason for RawBlob is that it copies faster than per byte copies.

cheneym2:
> I will set up my windows laptop to try reproducing the test error. I needed to do that anyway, so now's the time.

I reproduced the error on my windows laptop using the "slang-test" binary downloaded from the failed CI.

source/compiler-core/slang-source-loc.cpp
cheneym2: Okay

cheneym2:
The SPIRV output from compiling the slang source with slangc is identical before and after my change.  I confirmed both the "-target spirv-asm" and "-target spirv" output matches between the slang build from my failed CI and Ariel's passing CI https://github.com/shader-slang/slang/actions/runs/9037392983

Similarly, I ran a "slangc" compile of "tests/bug/implicit-cast.slang" before and after my change on Linux and saw identical spirv-asm output in that case also.

cheneym2:
I found the connection between my change and the test regression.  Previously, the bug in determineEncoding was causing this test's .expected file (which only contains the digit '2') as UTF16. The whole contents is skipped, and the empty '.expected' file causes a false match in the slang-test result checker.  The 'actual' output is cropped to a zero length and the two empty files match.  My change fixes the bug, revealing the actual mismatch.  Note that replacing implicit-cast.slang.expected with any other single character contents also causes a false pass in the TOT code.

csyonghe:
This is a good find! Seems like the right fix here is to update the expected result to 0 1 0 0.

csyonghe (approved):

source/core/slang-char-encode.cpp
csyonghe: Yes, you are right. The guessing logic should be in the >=2 block instead of in this branch.

PR #4137, opened 2024-05-08
Metal: propagate and specialize address space.

### LINKED ISSUE #4128: Metal: `out` modifier causes a compile error
**Problem description**
When `out` modifier is used for a function parameter, Metal compiler prints an error.

**Repro step**
The following slang-test case can reproduce the issue.
```
//TEST:SIMPLE(filecheck=METALLIB): -target metallib

//TEST_INPUT: ubuffer(data=[0 0 0 0], stride=4):out,name outputBuffer

void Test_out_param(out int value)
{
    value = 1;
}

[numthreads(1,1,1)]
void computeMain()
{
    // METALLIB: define void @computeMain
    int value = 0;
    Test_out_param(value);
    outputBuffer[0] = value;
}
```
When the issue is reproduced, the exact error message is following,
```
metal 32023.155: tests/metal/out.slang(7): error :  unknown type name 'out'
void Test_out_param_0(out int value_0)
                      ^
metal 32023.155: tests/metal/out.slang(19): error :  no matching function for call to 'Test_out_param_0'
    Test_out_param_0(&value_1);
    ^~~~~~~~~~~~~~~~
metal 32023.155: tests/metal/out.slang(7): note :  candidate function not viable: no known conversion from 'int *' to 'int' for 1st argument; remove &
void Test_out_param_0(out int value_0)
     ^
2 errors generated.
```

**Goal**
The modifier `out` probably needs to be emitted as `&` sign for Metal.

### LINKED ISSUE #4129: Metal: a temporary variable needs to respect the address space qualifier
**Problem description**
When a temporary variable is created by Slang, the variable needs to have a proper address space qualifier.

**Repro step**
The following slang-test case can reproduce the issue,
```
//TEST:SIMPLE(filecheck=METAL): -target metal
//TEST:SIMPLE(filecheck=METALLIB): -target metallib

uniform RWStructuredBuffer<float> outputBuffer;

[numthreads(1,1,1)]
void computeMain()
{
    outputBuffer[0]++;
}
```
When ran, it generates the following source code,
```
#include <metal_stdlib>
#include <metal_math>
using namespace metal;

[[kernel]] void computeMain(float device* outputBuffer_0 [[buffer(0)]])
{
    float* _S1 = outputBuffer_0+0U; // COMPILE ERROR HERE DUE TO A MISSING "device" KEYWORD.
   *_S1 = *_S1 + 1.0;
    return;
}
```
And the Metal compiler prints a following error message,
```
metal 32023.155: tests/metal/out.slang(9): error :  pointer type must have explicit address space qualifier
    float* _S1 = outputBuffer_0+0U;
         ^
1 error generated.
```

**Goal**
When a temporary variable is created, it has to have the address space qualifier.
It is `device` in this case.

csyonghe:
This PR adds a pass for the metal backend to propagate address space through the pointer typed values.
Initially, all `ConstantBuffer<T>` values has `Uniform` address space, all global buffer resource has `Global` address space, all local variables has `ThreadLocal` address space, and all `groupshared` variables has `ThreadGroup` address space.

The propagation pass propagates the address space through ops such as `getElementPtr` and `fieldAddress`.
If we encounter a `call`, we will collect address space from all arguments, and specialize the callee with the address space.
This means that if we have:
```
void output(out int x)
{
      x = 1;
}
void main()
{
      int p;
      output(p); // will specialize output for the `thread` address space.
      static groupshared int ps;
      output(ps); // will specialize for `threadgroup` address space.
}
```
We will generate:
```
void output_thread(int thread* x)
{
      *x = 1;
}
void output_threadgroup(int threadgroup* x)
{
      *x = 1;
}
void main()
{
      int p;
      output_thread(&p);
      threadgroup int ps;
      output_threadgroup(&ps);
}
```

Fixes #4129.
Fixes #4128.


jkwak-work (approved):
Thanks for fixing this.
It looks good to me.
But I will need to play with it to understand it better later.

PR #4139, opened 2024-05-08
Support [__ref] attribute to make this pass by reference.

### LINKED ISSUE #4110: Interlocked Ops within member function causes invalid spirv generation
For the following repro:
```hlsl

struct Buf
{
    uint t;

    [mutating]
    func tester()
    {
        uint prev;
        InterlockedAdd(t, 1, prev);
    }
};

struct Push
{
    Buf * b;
};

[[vk::push_constant]] Push push;

[shader("compute")]
[numthreads(1, 1, 1)]
void main()
{
    push.b->tester();
}
```

I get the following validation error for the generated spirv:
```
VUID-VkShaderModuleCreateInfo-pCode-08737(ERROR / SPEC): msgNum: -1520283006 - Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode is not valid SPIR-V: [VUID-StandaloneSpirv-None-04686] AtomicIAdd: Vulkan spec only allows storage classes for atomic to be: Uniform, Workgroup, Image, StorageBuffer, PhysicalStorageBuffer or TaskPayloadWorkgroupEXT.
  %original = OpAtomicIAdd %uint %34 %uint_1 %uint_0 %uint_1
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.3.268.0/windows/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
    Objects: 0
```

This only happens when tester is a member function. As soon as tester is a free function taking a pointer to Buf this error vanishes.

Config:
```
Slang Version: v2024.1.10
pc platform: x86-64msvc windows11
slang::TargetDesc::format: SlangCompileTarget::SLANG_SPIRV
slang::TargetDesc::profile: glsl_450+spirv_1_4
slang::TargetDesc::flags: SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY
slang::SessionDesc::defaultMatrixLayoutMode: SLANG_MATRIX_LAYOUT_COLUMN_MAJOR
```

csyonghe:
Fixes #4110.

## New feature
- Added [__ref] attribute that can be placed on struct member functions that causes `this` parameter to be passed by reference.


ArielG-NV (approved):
Looks good to me.

tests/spirv/ref-this.slang
```diff
@@ -0,0 +1,30 @@
...
+
+struct Push
+{
+    Buf * b;
+};
+
+[[vk::push_constant]] Push push;
+
+[shader("compute")]
+[numthreads(1, 1, 1)]
```
jkwak-work, line 26: Probably we should have a functional test with multiple threads that checks if the operations are happening atomically.

source/slang/core.meta.slang
```diff
@@ -2561,6 +2561,9 @@ attribute_syntax [nonmutating] : NonmutatingAttribute;
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [constref] : ConstRefAttribute;
 
+__attributeTarget(FunctionDeclBase)
+attribute_syntax [__ref] : RefAttribute;
```
jkwak-work, line 2565: Because this new attribute is only for `this`, how about a name like `__ref_this` to be more explicit.

jkwak-work:
If this is going to be an official way to workaround the limitations of the reference feature, it will need to be explained better on our document.

ArielG-NV:
> Because this new attribute is only for this, how about a name like __ref_this to be more explicit.
I agree with changing the syntax, it is quite confusing.

PR #4143, opened 2024-05-09
More Metal Intrinsics.

csyonghe:
- Barriers
- Vector/Matrix multiply
- vector/matrix reshape
- cross, dot, reflect, refract
- radians, degrees
- clamp, mix, lerp, saturate, step, smoothstep
- countbits, reversebits
- discard
- dfdx, dfdy, fwidth
- isfinite, isinf, isnan
- select


jkwak-work (approved):
Looks good to me.
I got confused for a while on the order of "select", but it seems correct.

source/slang/hlsl.meta.slang
```diff
@@ -6030,7 +6048,7 @@ void DeviceMemoryBarrierWithGroupSync()
 
 __generic<T : __BuiltinFloatingPointType, let N : int>
 [__readNone]
-[require(cpp_cuda_glsl_hlsl_spirv, sm_2_0_GLSL_140)]
+[require(cpp_cuda_glsl_hlsl_metal_spirv, sm_2_0_GLSL_140)]
 T distance(vector<T, N> x, vector<T, N> y)
```
jkwak-work, line 6052: There is `Ts distance(T, T)` in Metal

source/slang/hlsl.meta.slang
```diff
@@ -10453,6 +10528,7 @@ int sign(T x)
     __target_switch
     {
     case hlsl: __intrinsic_asm "sign";
+    case metal: __intrinsic_asm "sign";
```
jkwak-work, line 10531: The return type of `sign` is float.
You may need to type-cast to int like how glsl does one line below.

jkwak-work (approved):
Looks good to me

PR #4148, opened 2024-05-10
[gfx] add missing Result to IRayTracingCommandEncoder::bindPipline

skallweitNV:


tools/gfx/debug-layer/debug-command-encoder.cpp
```diff
@@ -470,16 +470,17 @@ void DebugRayTracingCommandEncoder::deserializeAccelerationStructure(
...
 {
     SLANG_GFX_API_FUNC;
     auto innerPipeline = getInnerObj(state);
     IShaderObject* innerRootObject = nullptr;
     commandBuffer->rootObject.reset();
-    baseObject->bindPipeline(innerPipeline, &innerRootObject);
+    Result result = baseObject->bindPipeline(innerPipeline, &innerRootObject);
     commandBuffer->rootObject.baseObject.attach(innerRootObject);
     *outRootObject = &commandBuffer->rootObject;
+    return result;
```
jkwak-work, line 483: Should it return earlier when the result is fail?

jkwak-work (commented):
Looks good to me.
But I will defer the approval to Yong.

csyonghe (approved):

tools/gfx/debug-layer/debug-command-encoder.cpp
skallweitNV: not sure but it's consistent with the other `bindPipeline` implementations

PR #4156, opened 2024-05-13
Propagate warning settings on `Linkage` to IR passes.

### LINKED ISSUE #4155: Cannot disable warning 41203
Falcor has many reinterpret casts from a large to a smaller struct and we therefore get many warnings like:
```
warning 41203: reinterpret<> into not equally sized types: '' with size '4' casted into 'DynamicType' with size '0'
```

Disabling the warnings with either `spOverrideDiagnosticSeverity` or through the "new" `CompilerOptionEntry` mechanism doesn't seem to work.

csyonghe:
Fixes #4155.


jkwak-work (approved):
Looks good to me.

PR #4158, opened 2024-05-13
Implement texture functions for Metal target

jkwak-work:
This commit is to implement texture functions for Metal target.

The following functions are implemented and tested.
 - GetDimensions()
 - CalculateLevelOfDetail()
 - CalculateLevelOfDetailUnclamped()
 - Sample()
 - SampleBias()
 - SampleLevel()
 - SampleCmp()
 - SampleCmpLevelZero()
 - Gather()
 - SampleGrad()
 - Load()

Metal has limited support for the texture functions compared to HLSL.
 - LOD is not supported for 1D texture,
 - Depth textures are limited to 2D, 2DArray, Cube and CubeArray textures.
 - "Offset" variants are limited to 2D, 2DArray, 2D-Depth, 2DArray-Depth and 3D textures.

The functions that cannot be implemented for Metal should properly be handled by the capability system later.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -964,14 +964,16 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,access,isShadow,0,forma
 
     [__readNone]
     [ForceInline]
-    [require(glsl_hlsl_spirv, texture_querylod)]
+    [require(glsl_hlsl_metal_spirv, texture_querylod)]
     float CalculateLevelOfDetail(SamplerState s, TextureCoord location)
     {
         __requireComputeDerivative();
         __target_switch
         {
         case hlsl:
-            __intrinsic_asm "CalculateLevelOfDetail";
+            __intrinsic_asm ".CalculateLevelOfDetail";
+        case metal:
+            __intrinsic_asm "$0.calculate_clamped_lod($1, $2)";
         case glsl:
             __intrinsic_asm "textureQueryLod($p, $2).x";
         case spirv:
@@ -984,14 +986,16 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,access,isShadow,0,forma
 
     [__readNone]
     [ForceInline]
-    [require(glsl_hlsl_spirv, texture_querylod)]
+    [require(glsl_hlsl_metal_spirv, texture_querylod)]
     float CalculateLevelOfDetailUnclamped(SamplerState s, TextureCoord location)
     {
         __requireComputeDerivative();
         __target_switch
         {
         case hlsl:
-            __intrinsic_asm "CalculateLevelOfDetailUnclamped";
+            __intrinsic_asm ".CalculateLevelOfDetailUnclamped";
+        case metal:
+            __intrinsic_asm "$0.calculate_unclamped_lod($1, $2)";
         case glsl:
             __intrinsic_asm "textureQueryLod($p, $2).y";
         case spirv:
@@ -1008,7 +1012,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
 {
     [__readNone]
     [ForceInline]
-    [require(cpp_cuda_glsl_hlsl_spirv, texture_sm_4_1_fragment)]
+    [require(cpp_cuda_glsl_hlsl_metal_spirv, texture_sm_4_1_fragment)]
     T Sample(SamplerState s, vector<float, Shape.dimensions+isArray> location)
     {
         __requireComputeDerivative();
@@ -1017,6 +1021,32 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
             case cpp:
             case hlsl:
                 __intrinsic_asm ".Sample";
+            case metal:
+                if (isArray == 1)
+                {
+                    switch (Shape.flavor)
+                    {
+                    case $(SLANG_TEXTURE_1D):
+                        __intrinsic_asm "$0.sample($1, ($2).x, uint(($2).y))";
+                    case $(SLANG_TEXTURE_2D):
+                        __intrinsic_asm "$0.sample($1, ($2).xy, uint(($2).z))";
+                    case $(SLANG_TEXTURE_CUBE):
+                        __intrinsic_asm "$0.sample($1, ($2).xyz, uint(($2).w))";
+                    }
+                }
+                else
+                {
+                    switch (Shape.flavor)
+                    {
+                    case $(SLANG_TEXTURE_1D):
+                    case $(SLANG_TEXTURE_2D):
+                    case $(SLANG_TEXTURE_3D):
+                    case $(SLANG_TEXTURE_CUBE):
+                        __intrinsic_asm "$0.sample($1, $2)";
+                    }
+                }
+                // TODO: This needs to be handled by the capability system
+                __intrinsic_asm "<invalid intrinsic>";
             case glsl:
                 __intrinsic_asm "$ctexture($p, $2)$z";
             case cuda:
@@ -1062,7 +1092,7 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
 
     [__readNone]
     [ForceInline]
-    [require(cpp_glsl_hlsl_spirv, texture_sm_4_1_fragment)]
+    [require(cpp_glsl_hlsl_metal_spirv, texture_sm_4_1_fragment)]
     T Sample(SamplerState s, vector<float, Shape.dimensions+isArray> location, constexpr vector<int, Shape.planeDimensions> offset)
     {
         __requireComputeDerivative();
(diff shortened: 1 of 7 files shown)


tests/metal/texture.slang
```diff
@@ -0,0 +1,672 @@
...
+
+        // METAL: d2D{{.*}}.sample_compare(
+        // METALLIB: call {{.*}}.sample_compare_depth_2d.f32(
+        && float(1) == d2D.SampleCmp(shadowSampler, float2(u, u), 0)
+
+#if defined(TEST_AFTER_FIXING_CAPABILITY_PROBLEM)
+        // M-ETAL: d2DArray{{.*}}.sample_compare(
+        // M-ETALLIB: call {{.*}}.sample_compare_depth_2d_array.f32(
+        && float(1) == d2DArray.SampleCmp(shadowSampler, normalize(float3(u, 1 - u, u)), 0)
+#endif
```
jkwak-work, line 488: @ArielG-NV ,can you take a look at these two cases and give me some help on why the second one doesn't work with the capability system?
Both of them use the same signature,
```
    [__readNone]
    [ForceInline]
    [require(glsl_hlsl_metal_spirv, texture_shadowlod)]
    float SampleCmp(SamplerComparisonState s, vector<float, Shape.dimensions+isArray> location, float compareValue)
```
How come one of them works and the other one doesn't?
When the issue occurs, the exact error message is following,
```
tests/metal/texture.slang(661): error 36107: entrypoint 'computeMain' requires capability 'hlsl + fragment | hlsl + compute | hlsl + raygen | hlsl + intersection | hlsl + anyhit | hlsl + closesthit | hlsl + miss | hlsl + callable | glsl | spirv_1_0', which is incompatible with the current compilation target 'metal + compute'.
void computeMain()
     ^~~~~~~~~~~
}
```

ArielG-NV: debugging what is wrong locally, I will update when I figure out what went wrong. 
There were a few issues with the old capability system skipping to check certain capabilities that I found while moving over to the new system, I will check if this is one of these scenarios or if something else happened.

csyonghe (commented):
We should also make sure `lowerCombinedTextureSamplers` is called for metal, and combined texture samplers works.

tests/metal/texture.slang
ArielG-NV: I currently only see the error:
```
standard error = {
tests/metal/texture.slang(142): note 99999: an internal error threw an exception while working on code near this location
    t1D.GetDimensions(width);
                     ^
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: assert failure: (0 <= argIndex) && (argIndex < m_argCount)
}
```
is your local code pushed to your branch (so I may test it)?

ArielG-NV: update: everything works after latest PR

source/slang/slang-stdlib-textures.cpp
```diff
@@ -184,27 +192,56 @@ void TextureTypeInfo::writeGetDimensionFunctions()
...
+
+            StringBuilder metal;
+            const char* metalMipLevel = "0";
+
             if (includeMipInfo)
+            {
+                ++paramCount;
                 params << "uint mipLevel, ";
 
+                if (baseShape != SLANG_TEXTURE_1D);
```
jkwak-work: Note: there was a bug that is fixed in the latest PR, @ArielG-NV

jkwak-work:
> We should also make sure `lowerCombinedTextureSamplers` is called for metal, and combined texture samplers works.

I see. This function looks to be creating a default sampler for the combined textures.
I will use it to make the combined textures working.

jkwak-work:
> > We should also make sure `lowerCombinedTextureSamplers` is called for metal, and combined texture samplers works.
> 
> I see. This function looks to be creating a default sampler for the combined textures. I will use it to make the combined textures working.

I got the combined textures working.
But I will prepare a separate PR for it because this change is already big.

source/slang/hlsl.meta.slang
```diff
@@ -872,14 +872,16 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,access,isShadow,0,forma
...
     float CalculateLevelOfDetail(SamplerState s, TextureCoord location)
     {
         __requireComputeDerivative();
         __target_switch
         {
         case hlsl:
-            __intrinsic_asm "CalculateLevelOfDetail";
+            __intrinsic_asm ".CalculateLevelOfDetail";
+        case metal:
+            __intrinsic_asm "$0.calculate_clamped_lod($1, $2)";
```
csyonghe, line 884: Can be simplified to ".calculate_clamp_lod" so it is consistent with the HLSL case.

source/slang/hlsl.meta.slang
```diff
@@ -1520,13 +1959,32 @@ vector<TElement,4> __glsl_gatherCmp(__TextureImpl<T, Shape, isArray, 0, sampleCo
...
+                }
+                else
+                {
+                    // Tv gather_compare(sampler s, float2 coord, float compare_value, int2 offset = int2(0)) const
+                    __intrinsic_asm "$0.gather_compare($1, $2, $3, $4)";
+                }
+            }
+        }
+        // TODO: This needs to be handled by the capability system
+        __intrinsic_asm "<Metal support gather_compare with offset only for 2D depth>";
```
csyonghe, line 1987: The capability system won't have the feature to detect things in a branch like this. Perhaps we can consider a __intrinsic_asm "<illegal: message>"; syntax and then we issue a diagnostic with source loc pointing to the call site.

csyonghe (approved):

source/slang/hlsl.meta.slang
jkwak-work: Good catch. I will make the change and check if there are more like this

source/slang/hlsl.meta.slang
jkwak-work: Is it possible to issue a diagnostic from hlsl.meta.slang? Or should it be implemented from c++ side?

source/slang/hlsl.meta.slang
```diff
@@ -1382,13 +1690,77 @@ Array<T,4> __makeArray<T>(T v0, T v1, T v2, T v3);
...
+            {
+                if (isShadow == 1)
+                {
+                    // Tv gather(sampler s, float2 coord, uint array, int2 offset = int2(0)) const
+                    __intrinsic_asm "$0.gather($1, ($2).xy, as_type<uint>(($2).z))";
+                }
+                else
+                {
+                    // Tv gather(sampler s, float2 coord, uint array, int2 offset = int2(0), component c = component::x) const
+                    __intrinsic_asm "$0.gather($1, ($2).xy, as_type<uint>(($2).z), int2(0), metal::component($3))";
```
jkwak-work, line 1713: I think I made a mistake here. `as_type` is `reinterprete_cast` and I intended to use it as `static_cast`. I was having some troubles with type casting in Metal and only thing worked was `as_type` when I tried it. I will fix this tomorrow.

tests/bindings/multi-file-shared.slang
```diff
@@ -7,6 +7,10 @@ PUBLIC float4 use(float  val) { return val; };
 PUBLIC float4 use(float2 val) { return float4(val,0.0,0.0); };
 PUBLIC float4 use(float3 val) { return float4(val,0.0); };
 PUBLIC float4 use(float4 val) { return val; };
+
+#ifdef __SLANG__
+[require(hlsl)]
+#endif
```
jkwak-work, line 13: I am not sure why I have to have this one here. Without this change, this test was failing due to the capability system warning. I suspect that there might be a bug in the capability checking logic.

csyonghe:
@jkwak-work Is this ready for merge?

jkwak-work:
Yes it is ready for review.
I wanted to sort out the capability side issue, but it can be done after the merge

source/slang/hlsl.meta.slang
csyonghe: There isn't such a construct yet. We could implement something like `kIROp_CodeGenError` inst that will result in a code-gen time error if it doesn't get dead code eliminated at emit time.

tests/bindings/multi-file-shared.slang
csyonghe: That seems problematic.
Have you tried this with Ariel's pending PR?

csyonghe (dismissed):

tests/bindings/multi-file-shared.slang
csyonghe: If this test was not failing before, it implies something is broken, and we should probably fix it before merging.

If Ariel's pending PR fixes this issue, then let's merge this PR after his.

jkwak-work:
@ArielG-NV helped me to find that the use of `normalize` was causing one of two issues.
The latest change has the fix.

I am still investigating why the change in `tests/binding/multi-file-shared.slang` is needed.

tests/bindings/multi-file-shared.slang
jkwak-work: I figured why this was causing a trouble.
I had to add `metal` to the following aliases,
```
alias any_target = hlsl | glsl | c | cpp | cuda | spirv;
alias any_textual_target = hlsl | glsl | c | cpp | cuda;
alias any_gfx_target = hlsl | glsl | spirv;
```
I am going to submit my fix shortly.

tests/metal/texture.slang
```diff
@@ -0,0 +1,642 @@
...
+    // Metal doesn't support some features, and we need a new test that
+    // can check if the capability system can error them out properly.
+    #define NEED_TO_TEST_FOR_METAL_CAPABILITY 1
+#endif
+
+#if defined(EMIT_SOURCE)
+    // It appears that Slang-test doesn't initialize the depth cube texture
+    // properly.
+    #define TEST_WHEN_DEPTH_CUBE_WORKS
+#endif
```
jkwak-work, line 19: I will create a new issue for this.
The problem is that when the type is depth-cube texture, our "TEST_INPUT" method is not able to set it up properly.

source/slang/hlsl.meta.slang
jkwak-work: I will create a new issue for this.

jkwak-work:
All issues are addressed properly.
The change is ready for review and merge.

csyonghe (approved):

PR #4162, opened 2024-05-14
Fix CFG reversal logic for loops with an inner condition.

### LINKED ISSUE #4153: reverse while loop will could cause issue when loop inversion is turned off
This is probably not a real issue. But when we  turn the loop inversion off, 
this test will crash on Vulkan:
tests/autodiff/reverse-while-loop-2.slang

technically, turning off loop inversion should not crash anything. But maybe because our loop reversion is assuming the loop inversion in default, so the unexpected behavior is expected.

Open this issue to remind us that we should fix the https://github.com/shader-slang/slang/issues/4151 as soon as possible.

saipraveenb25:
Handles a corner case where the first block after the condition on the true-side is another condition. This would currently result in an invalid reverse graph, where the reverse version of the true-block is the merge point for two different branching insts (the reverse version of the loop as well as the second condition). This patch simply adds a blank block when constructing the reverse-loop (similar to critical edge breaking) so that each branch inst in the reversed loop has a unique merge block.

Fixes #4153


jkwak-work (commented):
Looks good to me.
But I have a question.
Why do we reverse the loop logic in the first place?

csyonghe (approved):

csyonghe:
@jkwak-work In the revserse mode auto diff pass, we need to reverse all control flows in the backward propagation function,  that's why there is this logic to produce that reversed cfg.

jkwak-work:
Aha. I see.
Thank for the explanation.
It had something to do with `backward` propagation.

PR #4168, opened 2024-05-14
Add diagnostic to prevent defining unsized static variables.

csyonghe:


source/slang/slang-lower-to-ir.cpp
```diff
@@ -7924,86 +7951,66 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
+        NestedContext nested(this);
+        auto subBuilder = nested.getBuilder();
+        auto subContext = nested.getContext();
 
-        auto builder = getBuilder();
+        IRGeneric* outerGeneric = nullptr;
+
+        // If we are static, then we need to insert the declaration before the parent.
+        // This tries to match the behavior of previous `lowerFunctionStaticConstVarDecl` functionality
+        if (isFunctionStaticVarDecl(decl))
```
jkwak-work, line 7962: I am not sure if this is what I think it is.
This function name sounds like checking a static variable inside of a function like following.
```
bool MyFunc()
{
    static int s_vars[] = { 0, 1, 2 };
    static int s_unsized[];
```
I don't see this case in the test file.
Can we have a test for this case too?

jkwak-work (approved):
Looks good to me.

source/slang/slang-lower-to-ir.cpp
csyonghe: `static int s_vars[] = { 0, 1, 2 };` is probably a case we should allow. Let me update this PR to cover that.

source/slang/slang-check-decl.cpp
```diff
@@ -2040,11 +2040,22 @@ namespace Slang
             }
         }
 
-        if (auto parentDecl = as<AggTypeDecl>(getParentDecl(varDecl)))
+        std::optional<TypeTag> varTypeTags;
```
kaizhangNV, line 2043: It looks like this lambda is not necessary to be reference binding as `varTypeTags` is not used anywhere else?

csyonghe: This logic was there because `getTypeTags` isn't needed for all control paths in the past and we just wanted to call it on-demand. But the new logic means it is required for most paths, so I changed it to just get and store it in a variable.

PR #4170, opened 2024-05-14
Remove use of `G0` and `__target_intrinsic` in stdlib.

### LINKED ISSUE #4053: "$G0" keyword doesn't work when used in __target_switch
**Problem description**
"$G0" keyword works when used in __target_intrinsic.
But it doesn't work when used in __target_switch.

**Repo steps**
The following function exists in core.meta.slang, and works as it is.
```
__generic<T>
__target_intrinsic(cuda, "SLANG_ALIGN_OF($G0)")
__target_intrinsic(cpp, "SLANG_ALIGN_OF($G0)")
[__readNone]
[require(cpp_cuda)]
int __alignOf();
```
But when it is converted to use __target_switch, it fails on one of tests, "tests/cuda/cuda-layout.slang".
```
__generic<T>
[__readNone]
[require(cpp_cuda)]
int __alignOf()
{
    __target_switch
    {
    case cpp: __intrinsic_asm "SLANG_ALIGN_OF($G0)";
    case cuda: __intrinsic_asm "SLANG_ALIGN_OF($G0)";
    }
}
```

The same problem is observed with the following functions as well,

- __sizeOf()
- Ptr<T> TensorView::data_ptr();
- Ptr<T> TensorView::data_ptr_at();
- T TensorView::load(uint x);
- T TensorView::load(uint x, uint y);
- T TensorView::load(uint x, uint y, uint z);
- T TensorView::load(uint x, uint y, uint z, uint w);
- void TensorView::store(uint x, T val);
- void TensorView::store(uint x, uint y, T val);
- void TensorView::store(uint x, uint y, uint z, T val);
- void TensorView::store(uint x, uint y, uint z, uint w, T val);

When ran with a debug build, it fails on the following assertion,
```
const char* IntrinsicExpandContext::_emitSpecial(const char* cursor)
{
...omit...
        case 'G':
        {
            // Get the type/value at the index of the specialization of this generic

            Index argIndex = parseNat() + m_argIndexOffset;
            
            IRSpecialize* specialize = as<IRSpecialize>(m_callInst->getCallee());
            SLANG_ASSERT(specialize);
```
"specialzie" is supposed to be `kIROp_Specialize`, but it is `null_ptr` when the issue is reproduced.

### LINKED ISSUE #4093: Use SMin/UMin for SPIR-V target

csyonghe:
Closes #4053.
Closes #4093.

This PR extends the `__intrinsic_asm` construct to allow additional exprs as arguments, that can be referenced through the intrinsic text in the form of `"$[0]"`. For example:

```
void f<T>()
{
     __intrinsic_asm "$[0]()", T;
}
```

Will cause `f<int>()` to be emitted as `int()`.

This removes the need for `G0` in the intrinsic text, where the generic argument may no longer be accessible in the expansion context.

With this PR, we have removed all use of `__target_intrinsic` on our stdlib functions. They are still being used on types and fields, which continues to be their intended use.


jkwak-work (dismissed):
Looks good to me.
The new syntax can come in handy later.

source/slang/hlsl.meta.slang
```diff
@@ -9894,15 +9894,23 @@ vector<T,N> fmax3(vector<T,N> x, vector<T,N> y, vector<T,N> z)
...
+    case glsl:
+    case metal:
+        __intrinsic_asm "min";
+    case cuda:
+    case cpp:
+        __intrinsic_asm "$P_min($0, $1)";
+    case spirv:
+        __intrinsic_asm "OpExtInst resultType resultId glsl450 fus(FMin, UMin, SMin) _0 _1";
+    }
+}
```
jkwak-work, line 9913: This change may trigger an unexpected issue described in #4093
https://github.com/shader-slang/slang/issues/4093
If CI fails, try again without this change.

csyonghe: That's really weird, they should generate the same spirv.

csyonghe: This is fixed. The reason is that if there is a global `static const int x = min(1,2)`, we will end up with an `SPIRVAsm` inst in global scope and our spirv emit logic can't handle that.

The fix is to make sure we inline all global scope spirvasms into their use sites.

jkwak-work: Interesting that 4093 is also resolved...
I am not sure what the cause was but I am happy not to worry about it anymore.
Great job.

csyonghe:
Yes, I expect we'll find more use for the syntax down the road.
Thanks for reviewing!

PR #4175, opened 2024-05-16
RasterizerOrder resource for spirv and metal.

csyonghe:
Also fixes the byte address buffer logic for metal.

## New Feature
- RasterizerOrdered buffers and textures are now supported on the spirv and metal backend.
- [RW][RasterizerOrdered]ByteAddressBuffer types are now supported on the metal backend.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2062,8 +2062,6 @@ int __offsetOf(in T t, in F field)
 }
 
 /// Mark beginning of "interlocked" operations in a fragment shader.
-//__glsl_extension(GL_ARB_fragment_shader_interlock)
-//__glsl_version(420)
 [require(glsl_spirv, GL_ARB_fragment_shader_interlock, fragment)]
 __intrinsic_op($(kIROp_BeginFragmentShaderInterlock))
 void beginInvocationInterlock();
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2062,42 +2062,16 @@ int __offsetOf(in T t, in F field)
 }
 
 /// Mark beginning of "interlocked" operations in a fragment shader.
-__glsl_extension(GL_ARB_fragment_shader_interlock)
-__glsl_version(420)
-void beginInvocationInterlock()
-{
-    __target_switch
-    {
-    case glsl:
-        __intrinsic_asm "beginInvocationInterlockARB";
-    case spirv:
-        spirv_asm {
-            OpCapability FragmentShaderPixelInterlockEXT;
-            OpExtension "SPV_EXT_fragment_shader_interlock";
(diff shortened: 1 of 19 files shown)


source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -617,6 +619,66 @@ struct ByteAddressBufferLegalizationContext
...
+            SLANG_RETURN_NULL_ON_FAIL(getNaturalSizeAndAlignment(m_targetProgram->getOptionSet(), type, &sizeAlignment));
+            if (sizeAlignment.size == 8)
+            {
+                // We need to load the value as two 4-byte values and then combine them.
+                auto loOffset = offset;
+                auto hiOffset = emitOffsetAddIfNeeded(offset, 4);
+                IRInst* loadLoArgs[] = { buffer, loOffset };
+                IRInst* loadHiArgs[] = { buffer, hiOffset };
+                auto loLoad = m_builder.emitIntrinsicInst(m_builder.getUIntType(), kIROp_ByteAddressBufferLoad, 2, loadLoArgs);
+                auto hiLoad = m_builder.emitIntrinsicInst(m_builder.getUIntType(), kIROp_ByteAddressBufferLoad, 2, loadHiArgs);
```
jkwak-work, line 636: Do we need to worry about endian-ness?
I guess we can go with little-endian and fix if there are any problems later.

source/slang/core.meta.slang
```diff
@@ -2062,42 +2062,16 @@ int __offsetOf(in T t, in F field)
...
-            OpCapability FragmentShaderPixelInterlockEXT;
-            OpExtension "SPV_EXT_fragment_shader_interlock";
-            OpExecutionMode __entryPoint PixelInterlockOrderedEXT;
-            OpBeginInvocationInterlockEXT;
-        };
-    }
-}
+//__glsl_extension(GL_ARB_fragment_shader_interlock)
+//__glsl_version(420)
+[require(glsl_spirv, GL_ARB_fragment_shader_interlock, fragment)]
```
jkwak-work, line 2067: I think you should remove the comments.

jkwak-work (approved):
Looks good to me.
And it was a very interesting read.
I never knew about the `beginInvocationInterlock`, which looks like mutex lock on c++.

And it was also interesting to see how you can emit certain code from c++ side.

ArielG-NV:
Looks good to me

ArielG-NV (commented):
Looks good to me

source/slang/slang-ir-byte-address-legalize.cpp
csyonghe: Eventually we should but all targets we care about are little endian so we can live with the looseness here for now, I think.

PR #4177, opened 2024-05-16
capture/relay: Add capture interface classes

kaizhangNV:
Add `ModuleCapture` class for capturing `IModule`
  - The `IModule` can only be created from -- `ISession::loadModule` -- `ISession::loadModuleFromIRBlob` -- `ISession::loadModuleFromSource` -- `ISession::loadModuleFromSourceString` so, we create the `ModuleCapture` at those methods in `SessionCapture` class. We use a hash map to store a map from `IModule` to `ModuleCapture` to avoid creating new `ModuleCapture` when there is already an old one.
  - In `SessionCapture::getLoadedModule`, we will assert on not finding a `ModuleCapture` instance.

Add `EntryPointCapture` class for capturing `IEntryPoint`.
  - The `IEntryPoint` can only be created from: -- `IModule::findEntryPointByName` -- `IModule::findAndCheckEntryPoint` so, we create the `EntryPointCapture` at those methods in `ModuleCapture`. Similarly, we use a hash map to store a map from `IEntryPoint` to `EntryPointCapture`.
  - In `IModule::getDefinedEntryPoint`, we will assert on not finding a `EntryPointCapture` instance.

Add `CompositeComponentTypeCapture` class for capturing CompositeComponentType, but since user is only exposed to `IComponentType`, so `CompositeComponentTypeCapture` just inherits from `IComponentType`.
  - `CompositeComponentType` can only be created from: -- ISession::createCompositeComponentType so create it here.

Add `TypeConformanceCapture` class for capturing `ITypeConformance`.
  - The `ITypeConformance` can only be created from: -- `ISession::createTypeConformanceComponentType` so create it here.

In addition, because `EntryPointCapture` and `ModuleCapture` share a same base class `IComponentType`, we generate the COM GUID for those two classes to differentiate them.

### DIFF
--- a/source/slang-capture-replay/slang-composite-component-type.cpp
+++ b/source/slang-capture-replay/slang-composite-component-type.cpp
@@ -1,0 +1,129 @@
+#include "capture_utility.h"
+#include "slang-composite-component-type.h"
+
+namespace SlangCapture
+{
+    CompositeComponentTypeCapture::CompositeComponentTypeCapture(slang::IComponentType* componentType)
+        : m_actualCompositeComponentType(componentType)
+    {
+        assert(m_actualCompositeComponentType != nullptr);
+        slangCaptureLog(LogLevel::Verbose, "%s: %p\n", __PRETTY_FUNCTION__, m_actualCompositeComponentType);
+    }
+
+    CompositeComponentTypeCapture::~CompositeComponentTypeCapture()
+    {
+        m_actualCompositeComponentType->release();
+    }
+
+    ISlangUnknown* CompositeComponentTypeCapture::getInterface(const Guid& guid)
+    {
+        if (guid == IComponentType::getTypeGuid())
+        {
+            return static_cast<ISlangUnknown*>(this);
+        }
+        return nullptr;
+    }
+
+    SLANG_NO_THROW slang::ISession* CompositeComponentTypeCapture::getSession()
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        slang::ISession* res = m_actualCompositeComponentType->getSession();
+        return res;
+    }
+
+    SLANG_NO_THROW slang::ProgramLayout* CompositeComponentTypeCapture::getLayout(
+        SlangInt    targetIndex,
+        slang::IBlob**     outDiagnostics)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        slang::ProgramLayout* res = m_actualCompositeComponentType->getLayout(targetIndex, outDiagnostics);
+        return res;
+    }
+
+    SLANG_NO_THROW SlangInt CompositeComponentTypeCapture::getSpecializationParamCount()
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        SlangInt res = m_actualCompositeComponentType->getSpecializationParamCount();
+        return res;
+    }
+
+    SLANG_NO_THROW SlangResult CompositeComponentTypeCapture::getEntryPointCode(
+        SlangInt    entryPointIndex,
+        SlangInt    targetIndex,
+        slang::IBlob**     outCode,
+        slang::IBlob**     outDiagnostics)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        SlangResult res = m_actualCompositeComponentType->getEntryPointCode(entryPointIndex, targetIndex, outCode, outDiagnostics);
+        return res;
+    }
+
+    SLANG_NO_THROW SlangResult CompositeComponentTypeCapture::getResultAsFileSystem(
+        SlangInt    entryPointIndex,
+        SlangInt    targetIndex,
+        ISlangMutableFileSystem** outFileSystem)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        SlangResult res = m_actualCompositeComponentType->getResultAsFileSystem(entryPointIndex, targetIndex, outFileSystem);
+        return res;
+    }
+
+    SLANG_NO_THROW void CompositeComponentTypeCapture::getEntryPointHash(
+        SlangInt    entryPointIndex,
+        SlangInt    targetIndex,
+        slang::IBlob**     outHash)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        m_actualCompositeComponentType->getEntryPointHash(entryPointIndex, targetIndex, outHash);
+    }
+
+    SLANG_NO_THROW SlangResult CompositeComponentTypeCapture::specialize(
+        slang::SpecializationArg const*    specializationArgs,
+        SlangInt                    specializationArgCount,
+        slang::IComponentType**            outSpecializedComponentType,
+        ISlangBlob**                outDiagnostics)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        SlangResult res = m_actualCompositeComponentType->specialize(specializationArgs, specializationArgCount, outSpecializedComponentType, outDiagnostics);
+        return res;
+    }
+
+    SLANG_NO_THROW SlangResult CompositeComponentTypeCapture::link(
+        slang::IComponentType**            outLinkedComponentType,
+        ISlangBlob**                outDiagnostics)
+	{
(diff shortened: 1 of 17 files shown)


source/slang-capture-replay/slang-composite-component-type.cpp
```diff
@@ -0,0 +1,129 @@
...
+        : m_actualCompositeComponentType(componentType)
+    {
+        assert(m_actualCompositeComponentType != nullptr);
+        slangCaptureLog(LogLevel::Verbose, "%s: %p\n", __PRETTY_FUNCTION__, m_actualCompositeComponentType);
+    }
+
+    CompositeComponentTypeCapture::~CompositeComponentTypeCapture()
+    {
+        m_actualCompositeComponentType->release();
+    }
```
jkwak-work, line 16: I think release should happen in release not in the destructor.
I am not sure if it is possible or not tho.

The case I can think of is if the application tries to reuse CompositeComponentTypeCapture object after calling,
`CompositeComponentTypeCapture::release()`.
In this case, the memory pointer to m_actualCompositeComponentType will be overwritten and the resource will leak.

source/slang-capture-replay/slang-module.cpp
```diff
@@ -0,0 +1,234 @@
...
+    SLANG_NO_THROW SlangResult ModuleCapture::findEntryPointByName(
+        char const*     name,
+        slang::IEntryPoint**   outEntryPoint)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+
+        slang::IEntryPoint* pEntryPoint = nullptr;
+        SlangResult res = m_actualModule->findEntryPointByName(name, &pEntryPoint);
+        EntryPointCapture* entryPointCapture = getEntryPointCapture(pEntryPoint);
+        *outEntryPoint = static_cast<slang::IEntryPoint*>(entryPointCapture);
```
jkwak-work, line 35: You may need to check if `outEntryPoint` is a null_ptr, before writing a value to it.
Same for all other parameters that takes double pointer type.

source/slang-capture-replay/slang-module.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    SLANG_NO_THROW SlangResult ModuleCapture::serialize(ISlangBlob** outSerializedBlob)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        SlangResult res = m_actualModule->serialize(outSerializedBlob);
+        return res;
+    }
+
+    SLANG_NO_THROW SlangResult ModuleCapture::writeToFile(char const* fileName)
+	{
```
jkwak-work, line 77: The opening braces for the functions looks in a wrong indentation.
There are many other cases.

source/slang-capture-replay/slang-session.cpp
```diff
@@ -79,8 +87,20 @@ namespace SlangCapture
...
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+
+        Slang::List<slang::IComponentType*> componentTypeList;
+
+        // get the actual component types from our capture wrappers
+        assert(SLANG_OK == getActualComponentTypes(componentTypes, componentTypeCount, componentTypeList));
+
+        slang::IComponentType* compositeComponentType = nullptr;
+        SlangResult result = m_actualSession->createCompositeComponentType(
+                componentTypeList.getBuffer(), componentTypeCount, &compositeComponentType, outDiagnostics);
```
jkwak-work, line 99: I am not sure if this is intentional or a mistake.
The function call to `getActualComponentTypes()` will be removed in release build where `assert()` will be removed.
The behavior looks like `getActualComponentTypes()` will populate `componentTypeList` and it will be used by `createCompositeComponentType()`.
If it is the case, it may not work as expected.

source/slang-capture-replay/slang-session.cpp
```diff
@@ -159,38 +179,102 @@ namespace SlangCapture
...
         return result;
     }
 
+    ModuleCapture* SessionCapture::getModuleCapture(slang::IModule* module)
+    {
+        ModuleCapture* moduleCapture = nullptr;
+        moduleCapture = m_mapModuleToCapture.tryGetValue(module);
+        if (!moduleCapture)
+        {
+            moduleCapture = new ModuleCapture(module);
```
jkwak-work, line 240: Shouldn't we worry about the case where the memory actually ran out because the application used all up?
If that happens, `moduleCapture` will get a null_ptr, and this function will need to handle it properly.

source/slang-capture-replay/slang-session.cpp
```diff
@@ -159,38 +179,102 @@ namespace SlangCapture
...
+    }
+
+    SlangResult SessionCapture::getActualComponentTypes(
+            slang::IComponentType* const*   componentTypes,
+            SlangInt                        componentTypeCount,
+            List<slang::IComponentType*>&   outActualComponentTypes)
+    {
+        for (SlangInt i = 0; i < componentTypeCount; i++)
+        {
+            slang::IComponentType* const& componentType = componentTypes[i];
```
jkwak-work, line 254: The assertion thing I pointed out above can cause a problem here in release build because componentTypes will have zero size.

source/slang-capture-replay/slang-type-conformance.cpp
```diff
@@ -0,0 +1,131 @@
...
+        SlangInt    targetIndex,
+        slang::IBlob**     outDiagnostics)
+	{
+        slangCaptureLog(LogLevel::Verbose, "%s\n", __PRETTY_FUNCTION__);
+        slang::ProgramLayout* res = m_actualTypeConformance->getLayout(targetIndex, outDiagnostics);
+        return res;
+    }
+
+    SLANG_NO_THROW SlangInt TypeConformanceCapture::getSpecializationParamCount()
+	{
```
jkwak-work, line 46: I see that the indentation problem is when a tab-character is used.
Please replace tab-characters with 4 space-characters.

kaizhangNV: The tab thing is because I use "%s//" in vim to replace something globally. But I forgot using `:retab`, Now it's OK.

source/slang-capture-replay/slang-session.cpp
kaizhangNV: Yes, I'm considering define a release version assert actually.
Because I don't want the program proceed when hitting any of assert.

jkwak-work: We already have one.

jkwak-work: Hmmm. I think this needs to be wrapped with `do .. while(0)`
```
#define SLANG_RELEASE_ASSERT(VALUE) if(VALUE) {} else SLANG_ASSERT_FAILURE(#VALUE)
```

source/slang-capture-replay/slang-module.cpp
kaizhangNV: By looking at the actual `findEntryPointByName`, we seems expecting crash if the output pointer is nullptr.
Because we don't return any error on that case.

So if I check the nullptr, it will change the behavior, because user program will see the `SLANG_OK` as the return value, but output is nothing.

I think the correct fix will be to fix the actual API, not the capture logic. I will try to fix all the APIs along my way.

source/slang-capture-replay/slang-session.cpp
kaizhangNV: If this is the case, then slang will be crash, we cannot do anything about it.
IMO, we shouldn't worry about this extremely corner case.

source/slang-capture-replay/slang-module.cpp
kaizhangNV: Fix the nullptr check for output parameter in the actual APIs.

source/slang-capture-replay/slang-session.cpp
kaizhangNV: This is what I mentioned yesterday, it will end up to this function:
SLANG_RETURN_NEVER void handleSignal(SignalType type, char const* message);

and it will throw an exception without any useful message printed out.

source/slang-capture-replay/slang-session.cpp
kaizhangNV: fix that assert issue.

csyonghe (approved):

source/slang-capture-replay/slang-session.cpp
jkwak-work: A proper way to handle OOM is to check if the `new` returned null_ptr, and simply return null_ptr.

What could happen from my experience is that the memory is not completely out-of-memory but not enough to allocate for Slang.
From the application developer point of view, Slang is crashing when spawning a certain number of threads with no clear indication.

We should handle the out-of-memory case properly, which is not hard to do in this specific case.

PR #4180, opened 2024-05-16
Test binding index for combined and not-combined textures

jkwak-work:
This commit adds a test case that check the texture binding index for the combined textures and not-combined textures. It also tests the binding index that is explicitly specified.

### DIFF
--- a/tests/metal/texture-binding-index.slang
+++ b/tests/metal/texture-binding-index.slang
@@ -1,0 +1,24 @@
+//TEST:REFLECTION(filecheck=REFLECT): -stage compute -entry computeMain  -target metal
+
+// REFLECT: "name": "sampler"
+// REFLECT: {"kind": "shaderResource", "index": 0}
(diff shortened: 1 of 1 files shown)


tests/metal/texture-binding-index.slang
```diff
@@ -0,0 +1,24 @@
+//TEST:REFLECTION(filecheck=REFLECT): -stage compute -entry computeMain  -target metal
```
csyonghe, line 1: Would be good to include a SIMPLE test that checks for the actual binding attributes in the generated metal source.

csyonghe (dismissed):

tests/metal/texture-binding-index.slang
jkwak-work: Of course, I will made the change

jkwak-work:
I added a test that checks `[[texture(X)]]` and `[[sampler(Y)]]`.

csyonghe (approved):

PR #4183, opened 2024-05-17
`-ignore-capabilities` now skips the capability pass 100%

ArielG-NV:
This is done for compile performance if chosen 

note: this may break code down the line, currently works since nothing depends on the capabilities pass in any meaningful way.

### DIFF
--- a/tests/language-feature/capability/intrinsic-texture-ignore-capability.slang
+++ b/tests/language-feature/capability/intrinsic-texture-ignore-capability.slang
@@ -1,0 +1,498 @@
+#version 450 core
+#extension GL_EXT_texture_shadow_lod : enable
+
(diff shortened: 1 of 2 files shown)


tests/language-feature/capability/intrinsic-texture-ignore-capability.slang
```diff
@@ -0,0 +1,498 @@
+#version 450 core
```
csyonghe, line 1: What is the purpose of this test?

csyonghe (approved):

tests/language-feature/capability/intrinsic-texture-ignore-capability.slang
jkwak-work: You should write it in the test file as a comment.

PR #4186, opened 2024-05-17
Add `-minimum-slang-optimization` to favor compile time.

csyonghe:
## New Feature

- Add `-minimum-slang-optimization` compiler option to enable a minimum set of optimizations in the Slang compiler. This speeds up the compilation time in Slang for users who wish to favor faster compile times.

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -862,6 +862,7 @@ extern "C"
             SourceEmbedName,
             SourceEmbedLanguage,
             DisableShortCircuit,   // bool
+            MinimumSlangOptimization, // bool
 
             // Target
 
--- a/slang.h
+++ b/slang.h
@@ -1710,6 +1710,17 @@ extern "C"
         int targetIndex,
         bool forceScalarLayout);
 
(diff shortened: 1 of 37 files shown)


slang.h
```diff
@@ -862,6 +862,7 @@ extern "C"
             SourceEmbedName,
             SourceEmbedLanguage,
             DisableShortCircuit,   // bool
+            MinimumSlangOptimization, // bool
```
kaizhangNV, line 865: I'm wondering that if we should use -O<n> to indicate the level of optimization?

csyonghe: For now I don't want to mix these two things up just yet. This allows `-O` to continue to affect downstream compilation without being tied to slang's logic.

jkwak-work (approved):
Looks good to me.

I wonder how faster it can be with the new option.

jkwak-work:
Probably it is expected but I noticed that some of slang-test tests are failing when the option is enabled.

csyonghe:
@jkwak-work this is fine. We have tests that assumes the optimizations exist and is looking for a specific pattern that is no longer the case without the optimization.

PR #4191, opened 2024-05-19
Increase the CMake version requirement

jkwak-work:
This commit increases the minimum CMake version from 3.20 to 3.25.

I was trying to see if I can lower it to 3.16, but I found that we are currently using CMake feature that requires a version 3.25 not 3.20.

This finding is not new. I made a similar change to CMakePresets.json a few days ago. At that time, I didn't realize that the same change had to be made for CMakeList.txt as well.


ArielG-NV:
> Warning: Currently the required CMake version is 3.20 or above.

`/slang/CONTRIBUTION.md` also requires updating.

jkwak-work:
Thanks for catching that. I prepared a new PR to include both.

PR #4199, opened 2024-05-21
Support int and uint typed samplers for GLSL texture functions

### LINKED ISSUE #4196: GLSL texture functions need to support int and uint types of samplers as well as float
GLSL spec requires all texture functions to take `gsamplerXXX` types as the first input argument where `g` indicates `float`, `int` and `uint`.
It says,
> In the prototypes below, the g in the return type gvec4 is used as a placeholder for either nothing, i, or u making a return type of vec4, ivec4, or uvec4.

According to the section 8.9 in [the GLSL document](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf), the following functions take `gsamplerXXX` as the first argument except for the shadow/depth samplers.

- textureSize,
- textureQueryLod
- textureQueryLevels
- textureSamples
- texture
- textureProj
- textureLod
- textureOffset
- texelFetch
- texelFetchOffset
- textureProjOffset
- textureLodOffset
- textureProjLod
- textureProjLodOffset
- textureGrad
- textureGradOffset
- textureProjGradOffset
- textureGather
- textureGatherOffset
- textureGatherOffsets

This is a regression from one of recent changes. https://github.com/shader-slang/slang/pull/3963
And it causes out nightly CTS test to fail.

jkwak-work:
Closes #4196

The issue was introduced when we tried to support the sampler-less textures for GLSL texture functions. And some of GLSL functions could no longer take int/uint typed sampler anymore.

This commit fixes both problems that it works with the sample-less textures and it also supports int/uint typed samplers.

One of the findings is that GLSL supports only three types of samplers: `samplerXX`, `isamplerXX` and `usamplerXX`. They are not variable sized vectors and it means we need to support only three types with fixed number of components: vector<float,4>, vector<int,4> and vector<uint,4>.

Depth textures are always float type.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -1682,10 +1682,10 @@ public typealias usamplerBuffer = SamplerBuffer<uint4>;
 // textureSize
 // -------------------
 
-__generic<T:__BuiltinArithmeticType, let N:int>
+__generic<T:__BuiltinArithmeticType>
 [ForceInline]
 [require(glsl_hlsl_spirv, texture_size)]
-public int textureSize(Sampler1D<vector<T,N>> sampler, int lod)
+public int textureSize(Sampler1D<vector<T,4>> sampler, int lod)
 {
     int result;
     int numberOfLevels;
@@ -1693,10 +1693,10 @@ public int textureSize(Sampler1D<vector<T,N>> sampler, int lod)
     return result;
 }
 
-__generic<T:__BuiltinArithmeticType, let N:int>
+__generic<T:__BuiltinArithmeticType>
 [ForceInline]
 [require(glsl_hlsl_spirv, texture_size)]
-public ivec2 textureSize(Sampler2D<vector<T,N>> sampler, int lod)
+public ivec2 textureSize(Sampler2D<vector<T,4>> sampler, int lod)
 {
     vector<int,2> result;
     int numberOfLevels;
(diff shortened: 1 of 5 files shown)


jkwak-work:
This change breaks one of existing tests, "tests/diagnostics/int-texture-sample.slang".
I am not sure why we explicitly disallow integer type textures. @csyonghe 
Can we remove this test? or should we disallow only for HLSL? because we need the int/uint samplers for GLSL texture functions.

csyonghe (approved):
I understand the fix is to make the functions accept integral element types. The N -> 4 change is really just to simplify things and not for fixing anything, right?

csyonghe:
> This change breaks one of existing tests, "tests/diagnostics/int-texture-sample.slang".
> 
> I am not sure why we explicitly disallow integer type textures. @csyonghe 
> 
> Can we remove this test? or should we disallow only for HLSL? because we need support the int/uint samplers for GLSL texture functions.

The reason is that these functions are not defined in HLSL, and we cannot map them to any HLSL intrinsic. After all, if the texel is integer, you can interpolate it with any sampler other than nearest, so you may just as well use Load instead of Sample.

csyonghe:
The fix for glsl is probably just to define them separately for glsl.

csyonghe:
Alternatively, we can use the capability system to define `Sample` for integral types only for SPIRV target,

jkwak-work:
I see. I will try to limit the integer type support to GLSL only.

jkwak-work:
> I understand the fix is to make the functions accept integral element types. The N -> 4 change is really just to simplify things and not for fixing anything, right?

Right. "N -> 4" is just a clean up.
The goal is to get the integer types working for GLSL functions

jkwak-work:
Abandoning this PR.
I am preparing another PR for the issue.

PR #4203, opened 2024-05-21
Fix all Clang-14 warnings

ArielG-NV:
Fix all Clang-14 warnings


source/slang/slang-lower-to-ir.cpp
```diff
@@ -9954,7 +9954,7 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
                 getBuilder()->addRequireSPIRVVersionDecoration(irFunc, spvVersion->version);
             else if (auto cudasmVersion = as<RequiredCUDASMVersionModifier>(modifier))
                 getBuilder()->addRequireCUDASMVersionDecoration(irFunc, cudasmVersion->version);
-            else if (auto nonUniform= as<NonDynamicUniformAttribute>(modifier))
+            else if (as<NonDynamicUniformAttribute>(modifier))
```
jkwak-work, line 9957: This looks like an unused local variable warning. But Visual Studio should be able to catch these. Can you double check if it is disabled on visual studio side by a mistake?

jkwak-work (approved):

PR #4207, opened 2024-05-22
Increase MSVC warning level to 4 for Slang projects

jkwak-work:
This commit increases the MSVC warning level from 2 to 4 for Slang projects.

And this commit lowers the warning level from 4 to 2 for glslang, because we are not interested in the warnings from glslang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,7 +328,7 @@ if(SLANG_ENABLE_SLANG_GLSLANG)
     slang_add_target(
         source/slang-glslang
         MODULE
-        USE_EXTRA_WARNINGS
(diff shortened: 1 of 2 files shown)


ArielG-NV (commented):
I don't see an issue with the change, I am fine with the extra warnings.
I will let others review+approve so we have more input for if this may cause a problem.

csyonghe:
Does this change reveal any additional warnings?

csyonghe (approved):

jkwak-work:
No, I was surprised that increasing the level from 2 to 4 didn't print any additional warnings.
Actually, I am gonna double check if it is actually working as expected.

Issue #4208, opened 2024-05-22
Fix memory leak in slang.dll


jkwak-work:
When we run "slangc.exe", `_CrtDumpMemoryLeaks()` reports the memory leaks on Windows platform.
The memory leak has been reported for a while but it has been ignored.
It is because the messages are printed to the "debug output", which is visible only if you ran slangc.exe from VisualStudio.

Memory leak is never a good thing and we should fix it.
We should also setup CI to prevent it from happening in the first place.

jkwak-work:
When the change #4210 is merged, the issue can be reproduced by just running slangc.exe.
Here is an example,
```
$ build/Debug/bin/slangc.exe
Detected memory leaks!
Dumping objects ->
{1319712} normal block at 0x0000021ACE6BE010, 48 bytes long.
 Data: <`\&             > 60 5C 26 C4 1A 02 00 00 00 00 00 00 00 00 00 00
{3949} normal block at 0x0000021AC4265C60, 48 bytes long.
 Data: <          k     > 00 00 00 00 00 00 00 00 10 E0 6B CE 1A 02 00 00
{3948} normal block at 0x0000021AC425F420, 32 bytes long.
 Data: <                > 00 00 00 00 00 00 00 00 00 10 00 00 00 00 00 00
{3947} normal block at 0x0000021AC4201C50, 1024 bytes long.
 Data: <                > CD CD CD CD CD CD CD CD CD CD CD CD CD CD CD CD
{3921} normal block at 0x0000021AC420B4C0, 904 bytes long.
 Data: <X a             > 58 D1 61 F5 FA 7F 00 00 01 00 00 00 00 00 00 00
{234} normal block at 0x0000021AC41DBED0, 1024 bytes long.
 Data: < 7              > 80 37 92 F5 FA 7F 00 00 D4 A6 A7 F5 FA 7F 00 00
{215} normal block at 0x0000021AC41DBA90, 1024 bytes long.
 Data: <                > 10 AC 91 F5 FA 7F 00 00 C0 BE A7 F5 FA 7F 00 00
{200} normal block at 0x0000021AC41DB650, 1024 bytes long.
 Data: <                > FC A6 91 F5 FA 7F 00 00 B4 A6 A7 F5 FA 7F 00 00
{190} normal block at 0x0000021AC41DB210, 1024 bytes long.
 Data: <                > E0 A1 91 F5 FA 7F 00 00 D0 A6 A7 F5 FA 7F 00 00
[...]

PR #4210, opened 2024-05-22
Print memory leak info in Debug build of slangc.exe

jkwak-work:
When memory leak is detected, this commit will dump the information about the memory leak. This feature is available only in Debug build on Windows platform.

Also note that the message will not be printed on the client applications that use slang.dll, because the printing happens as a part of slangc.exe not slang.dll.

I found a bug that Slang::StdWriters was closing `stdout` and `stderr` in its destructor, which prevented Crt functions to print the messages to `stdout` and `stderr`.

### DIFF
--- a/source/core/slang-std-writers.cpp
+++ b/source/core/slang-std-writers.cpp
@@ -10,8 +10,8 @@ namespace Slang
 {
     RefPtr<StdWriters> stdWriters(new StdWriters);
 
-    RefPtr<FileWriter> stdError(new FileWriter(stderr, WriterFlag::AutoFlush));
-    RefPtr<FileWriter> stdOut(new FileWriter(stdout, WriterFlag::AutoFlush));
+    RefPtr<FileWriter> stdError(new FileWriter(stderr, WriterFlag::AutoFlush | WriterFlag::IsUnowned));
+    RefPtr<FileWriter> stdOut(new FileWriter(stdout, WriterFlag::AutoFlush | WriterFlag::IsUnowned));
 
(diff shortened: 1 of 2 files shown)


jkwak-work:
This commit is related to https://github.com/shader-slang/slang/issues/4208

ArielG-NV (commented):
I am good with the change.

I am wondering if we want to route all crt memory leak errors through the supplied callback so we can send all errors though a `DiagnosticSink`. It would be a benefit if we could only use Slang's diagnostic system to report errors.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetdumpclient?view=msvc-170

jkwak-work:
> I am good with the change.
> 
> I am wondering if we want to route all crt memory leak errors through the supplied callback `_CrtSetDumpClient` so we can send all errors though a `DiagnosticSink`. It would be a benefit if we could use Slang's diagnostic system to report errors for everywhere.
> 
> https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetdumpclient?view=msvc-170

If we want to use our own custom printing, it will have to be with _CrtSetReportHook.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetreporthook?view=msvc-170

But it will be ideal not to involve any other systems in the debugging process as much as possible.
The memory leak checking is supposed to happen very last minute before the process exit.
If there are lingering resources used by the DiagnosticSink, the resources will be reported as leak.

csyonghe (approved):

PR #4219, opened 2024-05-23
Make CTS nightly run reporting status to our slack channel.

kaizhangNV:


.github/workflows/vk-gl-cts-nightly.yml
```diff
@@ -12,7 +12,6 @@ jobs:
   build:
     runs-on: [Windows, self-hosted]
     timeout-minutes: 180
-    continue-on-error: true
```
jkwak-work, line 15: Is the default false?
Can we just have it explicitly set to false?

jkwak-work (approved):
Looks good to me

.github/workflows/vk-gl-cts-nightly.yml
kaizhangNV: Yea, it's default to false.

kaizhangNV: But it's doesn't matter now, because I set
```
    - name: result notification
      id: slack-notify
      if : always()
```

which will make sure this "slack-notify" keep running no matter CTS fail or not.

PR #4220, opened 2024-05-24
Skeletal gfx metal backend

cek:
Adds a skeletal metal backend for gfx, using metal-cpp and metal-cpp-extensions. The backend methods are almost universally stubbed out, with some small bits of initial progress. Currently only builds using CMake (i.e., no XCode/premake support).


external/metal-cpp-extensions/AppKit/AppKit.hpp
```diff
@@ -0,0 +1,36 @@
+/*
+ *
+ * Copyright 2020-2021 Apple Inc.
```
jkwak-work, line 3: I am not sure if we can just submit a licensed source file in a public repo like this.

skallweitNV: Not sure we even need the `metal-cpp-extensions` for implementing the backend.
We mostly just need `metal-cpp` but I think it would make sense to put those headers into a git submodule. Unfortunately Apple only distributes those headers in a zip file (https://developer.apple.com/metal/cpp/) but we could put them into a git repo or just fork this one: https://github.com/bkaradzic/metal-cpp

jkwak-work:
MacOS build is failing on the CI.
It needs to be fixed.

skallweitNV:
Will look into it

Issue #4167, opened 2024-05-14
Make sure performance profiler is disabled in production environment


csyonghe:

kaizhangNV:
Current related change is here: https://github.com/shader-slang/slang/pull/4221

PR #4221, opened 2024-05-24
Disable slang profiler in production build

kaizhangNV:
Fix issue #4167.

Add `--production-build=true` in our build system to indicate the production build which is ONLY used when publish a new release.

In production build, we will disable the SLANG_PROFILER.

### DIFF
--- a/.github/workflows/release-linux-arm64.yml
+++ b/.github/workflows/release-linux-arm64.yml
@@ -29,6 +29,7 @@ jobs:
           export CONFIGURATION=${{matrix.configuration}}
           export ARCH=${{matrix.platform}}
           export TARGETARCH=${{matrix.targetPlatform}}
+          export PRODUCTION_BUILD=1
           echo "building..."
           source ./github_build.sh
           echo "creating binary archieves..."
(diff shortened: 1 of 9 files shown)


jkwak-work:
Is there a performance or memory cost if we don't disable it on the release build? If not much, can we just keep it?

kaizhangNV:
> Is there a performance or memory cost if we don't disable it on the release build? If not much, can we just keep it?
> 
> Edit: Is it possible to measure how much performance difference it makes when we don't disable it in release build?

There is no such measurement yet, but I don't expect it could cause much performance cost.

kaizhangNV:
I'm thinking about that we should make the profiling in the hierarchical way.
For example:

1. No profile at all.
2. Just slang time and downstream time
3. Break down slang time to semantics check, IR generation, IR link, target emit
4. Further breakdown each phase in level 2.

Then accept a compile option to specify the profiling level.

kaizhangNV:
So let's stage this change for now, and have some discussion in next dev meeting to find out how to should proceed.

csyonghe:
I am fine with checking in the changes here, but we need to add the flag for both premake and cmake, given that we are still in this transitionary period where both build systems exist. We need a timeline to retire premake so we don't have to constantly maintain both build configs.

csyonghe (approved):

PR #4222, opened 2024-05-24
Fix clang-18 build

exdal:
Simply fixes this build error:
```
slang/source/core/slang-performance-profiler.cpp:42:59: error: use of overloaded operator '<<' is ambiguous (with operand types 'StringBuilder' and 'rep' (aka 'long long'))
```


jkwak-work (commented):
I am not sure why this fixes a warning.
`int` is same as `Int32` isn't it?
I don't see anything is different before and after.

Can you share the warning message you were dealing with?

*EDIT: My bad. I found the warning message from the description.
It looks like the change for "int -> Int32" isn't needed, is it?

source/core/slang-performance-profiler.cpp
```diff
@@ -39,7 +39,7 @@ namespace Slang
             for (auto func : data)
             {
                 out << func.key << ": \t";
-                out << func.value.invocationCount << "\t" << func.value.duration.count()/1000000 << "\n";
+                out << static_cast<Int32>(func.value.invocationCount) << "\t" << static_cast<double>(func.value.duration.count()/1000000) << "\n";
```
jkwak-work, line 42: Shouldn't this be more like
```
static_cast<double>(func.value.duration.count())/1000000
```
I think the division should happen after the type is changed to double; not before.

exdal: You are right, I've completely misread the entire message, further testing with godbolt (https://godbolt.org/z/zYG75axMq), I will be pushing the latest commit.

jkwak-work (approved):
Looks good to me.
Thank you for adjusting the change quickly.

PR #4226, opened 2024-05-24
CTS: stage some known failure tests for now

kaizhangNV:
Stage some known failure test cases, will enable them back when the fix is merged.

The failure tests can be checked in
https://github.com/shader-slang/VK-GL-CTS/blob/main/test-lists/slang-waiver-tests.xml

TODO:
Change the schedule to per-PR, will change it back when it's ready to merge.


jkwak-work (approved):
Looks good to me.
Please submit after changing the "schedule" section.

PR #4224, opened 2024-05-24
Fix pointer example

cheneym2:
Make the example shown for pointers something that would compile. Don't redefine pNext and do define MyType.

PR #4228, opened 2024-05-25
Support SPIR-V DebugTypePointer

### LINKED ISSUE #4077: No debuginfo generated for buffer references
In HLSL, using the `vk::BufferPointer` syntax/, I don't see any debug info emitted for buffer pointer struct members when using the `-g3` options. There's no `OpDebugTypePointer` for the buffer reference and no `OpDebugTypeStruct` for the struct being pointed to, and indeed I don't see any code in slang to emit the `OpDebugTypePointer` instruction. With GLSL the `layout(buffer_reference)` syntax doesn't seem to supported at all.

jkwak-work:
Closes #4077

With this commit, a memory pointer type is always emitted.
When its base type is basic types, it will emit DebugTypeBasic and DebugTypePointer.
Otherwise, it will emit "uint64" as DebugTypeBasic.


source/slang/slang-ir-util.cpp
```diff
@@ -230,8 +230,9 @@ bool isSimpleDataType(IRType* type)
         return true;
     case kIROp_ArrayType:
     case kIROp_UnsizedArrayType:
-    case kIROp_PtrType:
         return isSimpleDataType((IRType*)type->getOperand(0));
+    case kIROp_PtrType:
```
csyonghe, line 234: I am not sure we need to change `isSimpleDataType` here.

The right fix is that if the value type of a pointer type is basic type, then we want to emit OpDebugTypePointer, plus the OpDebugValue logic to update the variable with a pointer typed value.

If the value type isn't basic, then we want to emit the pointer as uint64 (OpDebugTypeBasic), plus the OpDebugValue inst to update the variable with a uint64 value.

jkwak-work:
I made some changes so that a memory pointer type is always emitted.
When its base type is basic types, it will emit DebugTypeBasic and DebugTypePointer.
Otherwise, it will emit "uintptr" as DebugTypeBasic.

I couldn't work on DebugValue because it currently doesn't seem to work.
I created [an new issue](https://github.com/shader-slang/slang/pull/4224) and it seems like the "storage class" is not set properly.

csyonghe:
If you take one of our existing pointer test cases and turn on debug symbol, you should be able to exercise the debug value logic, right?

jkwak-work:
Right. I saw one test case. I will try it.

jkwak-work:
The test code had some problem and I fixed it.
It also tests DebugValue from the emitted SPIR-V asm.
"DebugValue" was already working fine.

csyonghe (approved):

PR #4230, opened 2024-05-26
Print warning when unsigned integer is promoted to signed

### LINKED ISSUE #3944: Operator << incorrectly truncates values.
uint8_var << 15

Is resolved to call <<(uint8, uint8) but should be calling << (int, int) instead with a warning.

We should also make sure related operators such as * also get treated correctly.

jkwak-work:
** Draft: this code is still WIP.
I need to check if the approach is good before proceed.

Closes #3944

This commit is to print warning message when a given argument is unsigned and it is coerced to signed type. It applies only to the following operator overloading:
 - operator+
 - operator-
 - operator*
 - operator/
 - operator<
 - operator>
 - operator<=
 - operator>=
 - operator&
 - operator|
 - operator^

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1666,85 +1666,136 @@ ${{{{
 
 for (auto op : intrinsicBinaryOps)
 {
-    for (auto type : kBaseTypes)
+    for (auto firstType : kBaseTypes)
     {
-        if ((type.flags & op.flags) == 0)
+        if ((firstType.flags & op.flags) == 0)
             continue;
 
-        char const* leftType = type.name;
-        char const* rightType = leftType;
-        char const* resultType = leftType;
+        sb << "\n// Operator overloading: " << firstType.name << "\n";
 
-        if (op.flags & BOOL_RESULT) resultType = "bool";
+        for (auto secondType : kBaseTypes)
+        {
+            if ((secondType.flags & op.flags) == 0)
+                continue;
 
-        // TODO: We should handle a `SHIFT` flag on the op
-        // by changing `rightType` to `int` in order to
-        // account for the fact that the shift amount should
-        // always have a fixed type independent of the LHS.
-        //
-        // (It is unclear why this change hadn't been made
-        // already, so it is possible that such a change
-        // breaks overload resolution or other parts of
-        // the compiler)
+            // When an unsigned integer value is implicitly casted to a signed
+            // integer value, we need to warn.
+            const bool needToWarn = (op.flags & INT_MASK)
+                && (((firstType.conversionKind == kBaseTypeConversionKind_Signed
+                        && secondType.conversionKind == kBaseTypeConversionKind_Unsigned)
+                    || (firstType.conversionKind == kBaseTypeConversionKind_Unsigned
+                        && secondType.conversionKind == kBaseTypeConversionKind_Signed)));
 
-        // scalar version
-        sb << "__intrinsic_op(" << int(op.opCode) << ") " << resultType << " operator" << op.opName << "(" << leftType << " left, " << rightType << " right);\n";
-        sb << "__intrinsic_op(" << int(op.opCode) << ") " << resultType << " __" << op.funcName << "(" << leftType << " left, " << rightType << " right);\n";
+            const bool isSameType = (firstType.conversionKind == secondType.conversionKind
+                && firstType.conversionRank == secondType.conversionRank);
 
-        // vector version
-        sb << "__generic<let N : int> ";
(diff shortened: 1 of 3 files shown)


jkwak-work:
I like to get some feedback if I am on a right track.
While waiting for a feedback, I will implement the actual warning printing logic.

jkwak-work:
I like to explain what is happening in the change.

Currently the operator overloading takes two arguments with a same type like following,
```
__intrinsic_op(245)
int8_t operator+(int8_t left, int8_t right);
```
I am explicitly overloading for the cases where their types have different sign-ness.
```
int8_t    operator+(int8_t left, uint8_t right);
uint16_t operator+(int8_t left, uint16_t right);
uint      operator+(int8_t left, uint right);
uint64_t operator+(int8_t left, uint64_t right);
uintptr_t operator+(int8_t left, uintptr_t right);
```
For each overloading, I want to print a warning and call the pre-existing function like following,
```
uint operator+(int8_t left, uint right);
{
    typealias ReturnType = uint;
    print_warning("implicit casting from int8_t to uint is not recommended.");
    return operator+(ReturnType(left), ReturnType(right));
}
```
We currently don't have a function that can just print a warning, as far as I know.
So I may have to add one or I may need to implement it with `__intrinsic_op`.

I have been trying to implement it with `__intrinsic_op` but I am having a trouble figuring out where to convert from `kIROp_Add_warn` to the regular ones like `kIROp_Add`.

I also found that we have `[deprecated(message)]` attribute.
Maybe I can add a similar one like `[warning(message)]`.

csyonghe:
Given that the check should be done in the front end, we shouldn't need to use a separate IR op to generate the warning. Having an attribute like [warning] that behaves the same as [deprecate] should be sufficient.

We can also consider defining dedicated interfaces for signed integer types and unsigned integer types, so we can use a single generic function to cover the cases.

jkwak-work:
Thanks for the comment. That makes sense.
I will make some changes accordingly.

jkwak-work:
I like to share an update here.

The plan is to add functions that print a warning like following,
```
__intrinsic_op(245)
int8_t operator+(int8_t left, int8_t right);

// The following attribute will print a warning like,
// warning (XXX): Implicit integer type conversion between signed and unsigned types is not recommended. "int8_t" will be cast to "uint".
[warning_implicit_cast_between_sign_and_unsign(int8_t, uint)]
uint operator+(int8_t left, uint right);
{
    typealias ReturnType = uint;
    return operator+(ReturnType(left), ReturnType(right));
}
```
A problem I encountered is that this prevents the constant-folding when the new functions are used.
A simple for-loop that uses `[ForceUnroll]` stopped working.
The following code snippet shows the example.
```
[ForceUnroll]
for (int i = 0; i < m + n; i++) // operator+ doesn't return a compile-time constant anymore.
```
Adding `[ForceInline]` and `constexpr` didn't help.
I think `constexpr` should be an ideal solution but it currently doesn't seem to work for functions; I created [a bug for it](https://github.com/shader-slang/slang/issues/4236).

jkwak-work:
Here is another update.
As Yong told me, `[__unsafeForceInlineEarly]` worked well as an alternative to `constexpr`.

My local change is mostly working but there is a problem with for-loop inside of a function with [Differentiable].
According to the error messages, all for-loop inside of [Differentiable] must have `[ForceUnroll]` or `[MaxIters(x)]`.

Strangely there are cases where for-loops are allowed not to have any attributes.
One of the examples is in "tests/autodiff/loop-init.slang".
I guess Slang figures out that those for-loops can be unrolled.
Here is the code snippet,
```
    [Differentiable]
    float3 eval(const A miData, const float3 wi, const float3 wo)
    {
        float3 albedo;
        for (uint i = 0; i < 3; i++) albedo[i] = miData.data[i]; // <====== this one

        float3 result = float3(1.f);
        [ForceUnroll] for (uint i = 0; i < TBsdfCount; i++) result *= albedo;
        return result;
    }
```
If I add a following function to the test file, the test starts failing with an error message saying that all for-loops must have `[ForceUnrol]` or `[MaxIters(x)]` inside of `[Differentiable]`.
```
[__unsafeForceInlineEarly]
[__readNone]
bool operator<(uint left, int right)
{
    return left < uint(right);
}
```
The problem is that the first for-loop, which is "for (uint i = 0; i < 3; i++)", becomes not unrolled automatically.

And then, when I explicitly add `[ForceUnroll]` to the first for-loop,... three of four subtests are passing.
The failing test is still failing when the `[ForceUnroll]` is added and the `operator<` is not present.

I will investigate why for-loops are not automatically unrolled with the new operator overloading.

jkwak-work:
Abandoning this PR because it turned out that operator overloading has its limitation for the purpose of checking the implicit type promotion.

PR #4235, opened 2024-05-28
DNI: Kai/perf break down

kaizhangNV:


kaizhangNV:
@csyonghe for visibility.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -10961,7 +10961,7 @@ RefPtr<IRModule> generateIRForTranslationUnit(
         //
         IRDeadCodeEliminationOptions options;
         options.keepExportsAlive = true;
-        eliminateDeadCode(module, options);
+        // eliminateDeadCode(module, options);
```
csyonghe, line 10964: I would keep this one.

kaizhangNV:
surprisingly, this change seems not introducing lots of tests regression.

csyonghe:
Removing optimizations pre-linking should mostly be fine in terms of correctness.

source/core/slang-performance-profiler.cpp
```diff
@@ -39,11 +40,104 @@ namespace Slang
...
+        std::atomic<long> m_propagateConstExpr;
+        std::atomic<long> m_eliminateDeadCode;
+    };
+
+    // SerialPerformaceProfilerImpl profiler is intent to be shared by all threads, it's thread safe object
+    // because the only member function is re-entrant function, all the member variables are atomic.
+    SerialPerformaceProfiler* Slang::SerialPerformaceProfiler::getProfiler()
+    {
+        static SerialPerformaceProfilerImpl profiler = SerialPerformaceProfilerImpl();
+        return &profiler;
```
jkwak-work, line 138: Is it possible to make `profiler` a thread_local?

source/core/slang-performance-profiler.cpp
```diff
@@ -39,11 +40,104 @@ namespace Slang
...
+        std::atomic<long> m_synthesizeBitFieldAccessors;
+        std::atomic<long> m_constructSSA;
+        std::atomic<long> m_simplifyCFG;
+        std::atomic<long> m_peepholeOptimize;
+        std::atomic<long> m_propagateConstExpr;
+        std::atomic<long> m_eliminateDeadCode;
+    };
+
+    // SerialPerformaceProfilerImpl profiler is intent to be shared by all threads, it's thread safe object
+    // because the only member function is re-entrant function, all the member variables are atomic.
```
jkwak-work, line 134: Can we simply require the application to take a mutex lock before entering the critical section/function?
I am not sure using multiple automic objects is better than the previous code that was taking a single mutex_lock.
If the application will take a mutex_lock before entering the function, we are using atomic objects for no good reason and it comes with a certain amount of cost.

kaizhangNV: We can just close this PR now.
This PR is only for debug purpose, so I marked it as Draft. But still thanks for the review.

Issue #4176, opened 2024-05-16
Capability System: capability-generator-main.cpp runs should support inclusiveJoin and union+error separately


ArielG-NV:
`alias sm_2_0_GLSL_140 = _GLSL_140 + sm_4_0 | sm_4_0;` has the effective output of `alias sm_2_0_GLSL_140 = sm_4_0;`. This is incorrect.
capability definitions need to be reworked to address the many possibilities and use cases.

Design is as described below.

ArielG-NV:
```
[require(hlsl, sm_4_0)]
[require(glsl, a)]
[require(glsl, b)]
void f();
```
We used to have f's capability to be:
`(hlsl + sm_4_0) | (glsl + a) | (glsl + b)`
But now it is
(hlsl + sm_4_0)  | (glsl + a + b)`

ArielG-NV:
We will error in the scenario above (the union of 2 equally abstract capability sets)

We will allow an inclusive join (`alias sm_2_0_GLSL_140 = _GLSL_140 + sm_4_0 | sm_4_0;` = `alias sm_2_0_GLSL_140 = _GLSL_140 + sm_4_0 | hlsl + sm_4_0 | spirv + sm_4_0...`) with use of a special operator

ArielG-NV:
adding some examples of no-error vs error directly from .capdef for future testing framework

ArielG-NV:
```
alias noErr1 = _sm_6_5 + fragment | _sm_6_0 + vertex;
alias noErr2 = _sm_6_5 + fragment + vertex + cpp;
alias noErr3 = _sm_6_5 + fragment;
alias noErr4 = GL_NV_shader_texture_footprint + GL_NV_compute_shader_derivatives + fragment | _GL_NV_shader_texture_footprint + fragment;
alias noErr5 = _GL_NV_shader_texture_footprint | _GL_NV_shader_texture_footprint + fragment;

alias err1 = GL_NV_shader_texture_footprint + GL_NV_compute_shader_derivatives + fragment | _GL_NV_shader_texture_footprint + _GL_NV_shader_atomic_fp16_vector + fragment;
alias err2 = GL_NV_shader_texture_footprint | GL_NV_ray_tracing_motion_blur;
alias err3 = GL_ARB_shader_image_size | GL_NV_shader_texture_footprint + fragment;
```

PR #4237, opened 2024-05-28
Capabilities generator inclusive join and misc

ArielG-NV:
fixes: #4176

Capability generator currently generates capabilities in an undesirable manner. Some capabilities generate incorrectly, some generate capabilities which should be an error (or are an error). Some capabilities cannot be constructed with the current logic. This is the primary goal of this PR.

Associated uintset & capability issues are addressed in this PR.

some specific changes of importance:
1. inclusive join (more information at: https://github.com/shader-slang/slang/issues/4176).
2. removed union (`|`).
3. error if union'ing sets such that disjoint sets are possible.
4. error if join is invalid.
5. .natvis fixes: some errors were causing .natvis to provide only default debug info.
6. explain and describe abstract atom using "key atom" and "keyhole" logic.


tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -65,6 +77,86 @@ struct CapabilityDef : public RefObject
...
+                        if (abstractFound.contains(i))
+                            continue;
+                        firstTargetSet.remove(i);
+                    }
+                }
+                if (alreadySetStage)
+                {
+                    for (auto i : firstStageSet)
+                    {
+                        if (!abstractFound.contains(i))
```
jkwak-work, line 136: Is this correct to have `!`?
The logic between firstStageSet and firstTargetSet appear to be almost identical exception this line.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -220,6 +329,24 @@ struct CapabilityDefParser
 struct CapabilityConjunction
 {
     HashSet<CapabilityDef*> atoms;
+
+    String toString() const
+    {
+        bool first = true;
+        String outS = "[";
+        for (auto i : atoms)
```
jkwak-work, line 337: Please rename `i` to `atom`

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -220,6 +329,24 @@ struct CapabilityDefParser
 struct CapabilityConjunction
 {
     HashSet<CapabilityDef*> atoms;
+
+    String toString() const
+    {
+        bool first = true;
+        String outS = "[";
```
jkwak-work, line 336: Please rename `outS` to `result`.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -259,11 +421,82 @@ struct CapabilityConjunction
...
+            {
+                cImpliesThis = true;
+                conjunctions.fastRemoveAt(i);
+            }
+            else
+                i++;
+        }
+        if (cImpliesThis)
+        {
+            conjunctions.add(_Move(c));
```
jkwak-work, line 453: This `_Move` will not work as you may expect.
The parameter, `c`, must be an rvalue-reference, "&&".

source/core/slang-uint-set.h
```diff
@@ -101,6 +90,9 @@ class UIntSet
         /// Add a value
     inline void add(UInt val);
     inline void add(const UIntSet& val);
+    inline void add(const List<Element>& other);
```
jkwak-work, line 93: This reveals the internal implementation, again.
Can you avoid having this?

source/core/slang-uint-set.h
```diff
@@ -145,37 +137,37 @@ class UIntSet
...
+            m_LSB = bitscanForward(m_processedElement);
+            m_processedElement &= m_processedElement - 1;
         }
     public:
-        Iterator(const List<Element>* inContext)
+        Iterator(const List<Element>* context)
         {
-            context = inContext;
+            m_context = context;
         }
```
jkwak-work, line 154: As I pointed out previously, this constructor should be "private" not "public".

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -65,6 +77,86 @@ struct CapabilityDef : public RefObject
...
+                            firstTargetSet.add(otherAbstractAtomsPresent);
+                    }
+                    else if (CapabilityDef::ptrOfStage && base == CapabilityDef::ptrOfStage)
+                    {
+                        if (!alreadySetStage)
+                            firstStageSet.add(otherAbstractAtomsPresent);
+                    }
+
+                    abstractFound.add(otherAbstractAtomsPresent);
+                }
```
jkwak-work, line 122: I am not sure about the performance but the logic will be much simpler if you break this loop into two loops.
```
for (auto otherAbstractAtomsPresent : atom->abstractAtomsPresent)
    abstractFound.add(otherAbstractAtomsPresent);

for (auto& otherAbstractAtomsPresent : atom->abstractAtomsPresent)
{
    auto base = otherAbstractAtomsPresent->getAbstractBase();

    if (!alreadySetTarget)
    {
        if (CapabilityDef::ptrOfTarget && base == CapabilityDef::ptrOfTarget)
        {
            for (auto item : otherAbstractAtomsPresent)
                if (abstractFound.contains(item))
                {
                    firstTargetSet.add(item);
                    alreadySetTarget = true;
                }
        }
    }

    if (!alreadySetStage)
    {
        if (CapabilityDef::ptrOfStage && base == CapabilityDef::ptrOfStage)
        {
            for (auto item : otherAbstractAtomsPresent)
                if (!abstractFound.contains(item))
                {
                    firstStageSet.add(item);
                    alreadySetTarget = true;
                }
        }
    }
````

File: tools/slang-capability-generator/capability-generator-main.cpp [hunk 8]
```diff
@@ -259,11 +421,82 @@ struct CapabilityConjunction
...
+                i++;
+            }
+        }
+        conjunctions.add(_Move(c));
+    }
+    void removeImplied()
+    {
+        for (Index i = 0; i < conjunctions.getCount(); i++)
+        {
+            for (Index ii = 0; ii < conjunctions.getCount(); ii++)
```
jkwak-work, line 484: Can ii start from i+1?
I think `a.implies(b)` is same as `b.implies(a)`, isn't it?

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: It seems like having `!` is the correct one and the other one should have `!` too.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: I will change the function to be more neat.

source/core/slang-uint-set.h
ArielG-NV: This function does not reveal the internal implementation:
```
inline void UIntSet::add(const List<Element>& other)
{
    for (auto i : other)
        add((UInt)i);
}
```

it just does `add` on a list for convience.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -45,13 +52,18 @@ struct SerializedArrayView
 
 struct CapabilityDef : public RefObject
 {
-    Index enumValue;
+public:
+    static inline CapabilityDef* ptrOfTarget = nullptr;
+    static inline CapabilityDef* ptrOfStage = nullptr;
```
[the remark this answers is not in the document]
jkwak-work: You are telling me that only reason you are making the variable `static` is to use the fancy syntax ` = nullptr`. That doesn't answer my question.

My question was from a design perspective of why you have to use `inline` from a class design perspective.
`inline` will make the variable inline, which may not have an effect you want.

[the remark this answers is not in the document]
jkwak-work: You are telling me that this code is guarded by mutex locking. Can you tell me where the locking is?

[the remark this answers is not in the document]
ArielG-NV: > inline will make the variable inline, which may not have an effect you want.

Inline is a good and desired behavior here (since we are in a .cpp file the behavior of the inline will not notably affect code gen). If the variable was not `inline`, it would not make any sense to have a default anyways.

source/core/slang-uint-set.h
jkwak-work: This function reveals that the class is internally using `List<Element>`. It creates an unnecessary dependency between the calling site and `List` and `Element`.
If we want to refactor the class later, as an example, we wouldn't be able to change Element nor List because calling sites will use those types.
This is bad from a class design perspective.

tools/slang-capability-generator/capability-generator-main.cpp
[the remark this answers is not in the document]
jkwak-work: Also note that when multiple threads share a variable, there are a lot more things that can go wrong. It is not just race condition from a software perspective. The hardware may keep the data in the CPU cache forever assuming that nothing needs to be synchronized.
A proper synchronization method is required to properly share a variable between multiple threads.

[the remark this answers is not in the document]
ArielG-NV: > You are telling me that this code is guarded by mutex locking. Can you tell me where the locking is?

I think my point was missed due to how I phrased my explanation, I will rephrase it.

These 2 variables are only written during `def` parsing. These variables find the unique value "target" and "stage" to write to their respective variables. As a result, this is safe logic to run.
```
    static inline CapabilityDef* ptrOfTarget = nullptr;
    static inline CapabilityDef* ptrOfStage = nullptr;
```

To do any processing beyond `def` parsing, we require to read these 2 variables. Since to find these variables is very expensive, we need to cache them since otherwise performance will be quite bad (even though this is a preprocessing step).

These variables cannot be made thread_local since if/when we multi-thread the parser code, we will never find "target" or "stage" nodes ever during the program lifetime if we do this.

We could (once done parsing `CapabilityDef`s) assign a `const&` and disallow usage of the mutable `static`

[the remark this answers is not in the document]
jkwak-work: I guess I was not clear that `static` member variable is bad from a class design perspective. 

Your intention seems to share the value among multiple threads, and that can justify the use of `static`.

I am still not convinced about thread safety. Your explanation sounds like the application can call theses function from multiple threads but I don't see any mutex locking. You cannot avoid locking if a variable will be used by multiple threads.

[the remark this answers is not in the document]
ArielG-NV: > I am still not convinced about thread safety. Your explanation sounds like the application can call theses function from multiple threads but I don't see any mutex locking. You cannot avoid locking if a variable will be used by multiple threads.

Nothing in capability-generator-main.cpp is multithreaded. Nothing easily can be made multithreaded (dependencies are everywhere). There likely needs to be a bigger design change across .capdef file & the generator if we for-some reason need threading.
```
def A : target;
alias B = A;
....
```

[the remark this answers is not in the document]
ArielG-NV: > I guess I was not clear that static member variable is bad from a class design perspective.

I can move it elsewhere, I just put the variable in the same namespace as the data it is used with (for the reason that the objects are used together)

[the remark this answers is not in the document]
jkwak-work: > > inline will make the variable inline, which may not have an effect you want.
> 
> Inline is a good and desired behavior here (since we are in a .cpp file the behavior of the inline will not notably affect code gen). If the variable was not `inline`, it would not make any sense to have a default anyways.

Are you telling me you don't know how to initialize a static member variable without `inline`?

source/core/slang-uint-set.h
ArielG-NV: Element is wrong, you are correct, this is a type-o.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -45,13 +52,18 @@ struct SerializedArrayView
 
 struct CapabilityDef : public RefObject
```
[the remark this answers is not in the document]
ArielG-NV: the only closed variable is `abstractAtomsPresent`. Once it is properly encapsulated, things should be fine.

tools/slang-capability-generator/capability-generator-main.cpp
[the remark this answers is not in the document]
ArielG-NV: > Are you telling me you don't know how to initialize a static member variable without inline?

no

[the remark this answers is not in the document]
ArielG-NV: as per spec: The declaration of a non-inline static data member in its class definition is not a definition and may be of an incomplete type other than cv void. The definition for a static data member that is not defined inline in the class definition shall appear in a namespace scope enclosing the member's class definition.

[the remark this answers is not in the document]
jkwak-work: About the use of `inline`, the most reasonable explanation is "since we are in a .cpp file" but it is not convincing enough to justify.
For the sake of a good coding practice, I recommend not to use it here and initialize it in a traditional manner.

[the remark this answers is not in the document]
jkwak-work: I guess I misunderstood.
The use of `static` is not to share the data among threads because the code is not expected to run with multiple threads and it will be difficult to make it working with multiple threads.
That sounds like you don't need to make it a `static`.
Please refactor the code if it is possible to avoid `static`.

tools/slang-capability-generator/capability-generator-main.cpp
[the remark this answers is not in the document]
jkwak-work: Your implementation indeed encapsulated the variable `abstractAtomsPresent` but the struct/class is now in a strange state that is neither.
This is not good from the class design perspective.
I don't know what you mean by "Once".
And I don't think it will be "fine" from the design perspective.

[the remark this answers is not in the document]
ArielG-NV: I will just make it public

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -43,17 +50,48 @@ struct SerializedArrayView
...
     CapabilityFlavor flavor;
-    int rank;
+    int rank = 0; // optional, 0 is default
     List<List<CapabilityDef*>> canonicalRepresentation;
     SerializedArrayView serializedCanonicalRepresentation;
+    SourceLoc sourceLoc;
+    HashSet<const CapabilityDef*> abstractAtomsPresent;
 
-    CapabilityDef* getAbstractBase()
+    CapabilitySharedContext& sharedContext;
```
jkwak-work, line 92: `&` is not needed.
Just make a copy.

You seem to be abusing `&` too much.
Please use it more precisely.

When you make the variable `&`, it is basically a memory pointer that points to a variable outside of the class or function.
The life time of the variable can end while this class/function still uses it.
It is unnecessary to use `&` for this case and many other cases as a function parameter.

jkwak-work (commented):
Looks good to me.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: `CapabilitySharedContext& sharedContext;` is a reference because imagine the case: `this->sharedContext.var = nullptr //invalidate pointer, not the data inside`

This case would not work if we copied the context, we need a pointer/reference to it (reference cannot be null and must be initialized, so ref is preferred)

ArielG-NV: If there is a preferred method, I would be happy to change the implementation .

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -126,32 +227,39 @@ struct CapabilityDefParser
...
 
-    SlangResult parseDefs()
+    SlangResult parseDefs(CapabilitySharedContext& capabilitySharedContext)
     {
         auto tokens = m_lexer->lexAllSemanticTokens();
         m_tokenReader = TokenReader(tokens);
         for (;;)
         {
-            RefPtr<CapabilityDef> def = new CapabilityDef();
+            RefPtr<CapabilityDef> def = new CapabilityDef(capabilitySharedContext);
```
jkwak-work, line 262: Does this have to be allocated from heap memory?
Can it be a stack memory?

ArielG-NV: I think the reason heap was used here is because there is a reasonable chance someone would run out of stack memory if we used stack (dynamic allocation of unknown very large size).

jkwak-work: How big do you think it will be?

jkwak-work: I don't know how you got that number but I don't think that is a right answer.

It is important to understand where the allocated memory comes from.
In this case, if you put `CapabilityDef` in a stack memory, it will occupy the stack memory as much as `sizeof(CapabilityDef)` tells us.
When I tried locally, I see "sizeof(CapabilityDef)" returns 176 (byte).

Note that the class has member variables that allocate from the heap memory.
When "CapabilityDef" is allocated from a stack memory, it doesn't make its member variables to allocate the memory from stack when they are calling `new` internally.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -78,85 +104,68 @@ struct CapabilityDef : public RefObject
...
-        bool alreadySetTarget = false;
-        HashSet<CapabilityDef*> firstTargetSet;
-
-        bool alreadySetStage = false;
-        HashSet<CapabilityDef*> firstStageSet;
-
-        HashSet<CapabilityDef*> abstractFound;
-
-        for (auto& c : canonicalRepresentation)
+        HashSet<const CapabilityDef*> targetAbstractAtomsOfType;
```
csyonghe, line 109: I find the this name confusing, what does `targetAbstractAtomsOfType` mean? Is it a set of all concrete target capDefs?

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -64,8 +87,11 @@ struct CapabilityDef : public RefObject
     List<List<CapabilityDef*>> canonicalRepresentation;
     SerializedArrayView serializedCanonicalRepresentation;
     SourceLoc sourceLoc;
+    HashSet<const CapabilityDef*> abstractAtomsPresent;
```
csyonghe, line 90: What is the purpose of this member, can you put a comment here?

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: I will just move it to the stack, no need to speculate on memory sizes I suppose.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: I agree the name is confusing, it does not explain what is happening. 
The variable and related will be renamed to `sharedAbstractAtomsInCanonicalSet_target`, where the variable tracks the "shared abstract atoms in a canonical set for the 'target' atom type"

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -18,18 +18,27 @@ namespace Diagnostics
...
+    Union = 0,
+    InclusiveJoin = 1 << 0,
+};
 
 struct CapabilityDef;
 
 struct CapabilityConjunctionExpr
 {
     List<CapabilityDef*> atoms;
+    AddCapabilityOptions howToJoinWithNextSetFlag;
```
csyonghe, line 40: May be cleaner to use:
```
CapabilityJoinOperator joinOperator;
```

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -18,18 +18,27 @@ namespace Diagnostics
...
+
 enum class CapabilityFlavor
 {
     Normal,
     Abstract,
     Alias
 };
+enum class AddCapabilityOptions
+{
+    Union = 0,
```
csyonghe, line 31: Do we ever need `Union` in the generator? What happens if we only support inclusive join(|&) and ordinary join(+)?

In that case we can just repurpose | to mean inclusive join and delete this enum.

tools/slang-capability-generator/capability-generator-main.cpp
csyonghe: The naming here is still bit confusing, since `glsl` is not an abstract atom, but a concrete atom that inherits from an abstract atom. The set here probably refers to things like `glsl` and not `target`, right?

csyonghe (approved):

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: > The set here probably refers to things like glsl and not target, right?
yes.

I don't really know a good word to simplify "list of 'stage' and 'target' abstract typed atoms"

csyonghe:
The use of `abstractAtom` isn't quite right. There isn't such a thing called abstract atom.
We have `stage` as an abstract capability, which doesn't correspond to any atom.
Then we have `fragment` as an concrete atom that inherits from `stage`, but `fragment` itself isn't abstract.

We need a good terminology to refer to atoms that inherits from an abstract atom. But since we only have two types of abstract atoms, `stage` and `target`, we can also just refer to them as `stageOrTargetAtoms`.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: We never needed union in the generator, Union is just a restricted 'inclusive join'.

I wonder if we should keep the enum just for 'ease of reading' (easier to change/add in the future if need-be)?

ArielG-NV:
> We need a good terminology to refer to atoms that inherits from an abstract atom. But since we only have two types of abstract atoms, stage and target, we can also just refer to them as stageOrTargetAtoms.

@csyonghe we could call the elements: "abstract facet"

tools/slang-capability-generator/capability-generator-main.cpp
csyonghe: We could, but my concern is that the `howToJoinWithNextSet` isn't a standard practice of defining expression trees. If we only have one way of joining, then the current syntax tree is clean and make sense, but having all terms appear in a single list that are joined in different ways depending on `list[i-1].howToJoinWithNextSet` is very weird and hard to understand.

csyonghe:
We can name them `mutuallyExclusiveAtoms`.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: makes sense, `howToJoinWithNextSet` will be removed.

ArielG-NV:
> We can name them `mutuallyExclusiveAtoms`. But that is confusing as well because it may seem that `glsl` and `fragment` are mutually exclusive. Maybe just `keyAtoms`, with some comment on what `key` means.

/// Any atom immediately derived off of an abstract atom is a keyAtom. 
/// Each abstract atom derived from is a distinct "keyhole"
"keyAtom"

ArielG-NV:
I like the new abstract definition. I will adjust code+comments accordingly

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -65,21 +111,186 @@ struct CapabilityDef : public RefObject
...
+        , m_defs(defs)
     {
     }
     
     Lexer* m_lexer;
     DiagnosticSink* m_sink;
 
     Dictionary<String, CapabilityDef*> m_mapNameToCapability;
-    List<RefPtr<CapabilityDef>> m_defs;
+    PreallocatedStackVector<CapabilityDef, kCapabilityDefCount>& m_defs;
```
csyonghe, line 294: I don't quite understand why this optimization is necessary. This is just generator code that runs once during build. As long as we are not incredibly slow, performance doesn't matter and we should prefer simplicity over low level optimizations. List<RefPtr<>> is fine here.

tools/slang-capability-generator/capability-generator-main.cpp
csyonghe: Again, performance shouldn't matter too much here in this generator code that runs only once during build.

csyonghe (approved):

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: I do not have a preference here.
@jkwak-work, any opinion on the matter?

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -263,7 +540,71 @@ struct CapabilityDisjunction
...
+    {
+        for (Index i = 0; i < conjunctions.getCount(); i++)
+        {
+            for (Index ii = 0; ii < conjunctions.getCount(); ii++)
+            {
+                if (ii == i)
+                    continue;
+                if (conjunctions[i].implies(conjunctions[ii]))
+                {
+                    conjunctions.fastRemoveAt(ii);
```
csyonghe, line 598: fastRemoveAt swaps element at ii with the last element. Could this lead to wrong result?

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -18,6 +18,8 @@ namespace Diagnostics
 #undef DIAGNOSTIC
 }
 
+const Index kCapabilityDefCount = 3000;
```
csyonghe, line 21: This feels unsafe. Let's not use static allocation here.

csyonghe (dismissed):

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: You are correct, a incorrect result is possible.

```
i=getCount()-1;
ii=0
```

Would cause issues, regular 'remove' should be used.
* we remove as many times as overlapping sets, so normally only once, this should be fine.

1. `if [LAST_INDEX] => [0]`, we swap and delete 0.
2. now, `if [LAST_INDEX] => [1]`, we have a problem since [LAST_INDEX ] will never compare with [1]
* requires specific order

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: You need to bring it back to use `new`.
When I asked if we can avoid using `new`, your answer was `because it is too big for the stack`.
So I told you that it is not as big as you think.

But now I was looking at the code around it, it has to be allocated from the heap.
It is not because it is too big, but because the object should be alive after this function call is ended.

I don't remember telling you to use stack.
I was only asking if we can avoid unnecessary use of heap if possible.
In this case, it is not possible.

jkwak-work: You need to be aware of the different characteristics of stack and heap.
I assume your experience is mostly with Java and Java doesn't have an explicit way to use stack.
In C++, the use is always preferred but it is not always possible.

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: I meant that you can change the code to do the following,
```
for (Index ii = i+1; ii < conjunctions.getCount(); ii++)
```

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: I don't understand your explanation, Ariel.

What do you mean by "if last index is 0, we swap and delete 0"?
If last index is 0, it means the size of the container is 1 and the for-loop wouldn't even start, because there is nothing to compare to when there is only one element.

"if last index is 1, we have a problem since last index will never compare with 1".
I don't understand what this means.
If last index is 1, it means there are only two elements.
You will need to compare the first element and the second, and that's it.
Why would you want to compare the last element with the last element?

In general, "fast remove" type of remove is same as the regular remove except it changes the order of items in an unpredictable way, which is unpreferable in most cases.
Other than that, it is same as the regular remove.
In other words, if "fast remove" causes a problem, you should expect the same problem with the regular remove.

ArielG-NV: I think there is a misunderstanding in what I mean by [LAST_INDEX]

ArielG-NV: [LAST_INDEX] => [1]
means
conjunctions[LAST_INDEX].implies(conjunctions[1])

I was using `=>` as mathematical notation `implies`

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: >You need to be aware of the different characteristics of stack and heap.
 
I am very aware of `stack` vs. `heap`,  I primarily use C++/C for development, I understand the differences.

I believed "Can it be a stack memory?" was a suggestion to use stack.
* when I said I thought it was too big I was referring to how it is dynamic in size, but this remark seemed to be disregarded?

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: This is a reoccurring communication problem with you, Ariel.
You assume a certain expression by yourself and start using it without explaining what it means.
Please use a plain English or define your magic symbols so that people can read your formula.

ArielG-NV: To define the implies symbol, the symbol is defined as per the mathematical relational symbol ["implication"](https://mathvault.ca/hub/higher-math/math-symbols/logic-symbols/)

jkwak-work: How about `[` and `]`?
Is there a mathematical use of them that I don't know of?

ArielG-NV: `[]` was to represent indexing since the context of the problem is `if (conjunctions[i].implies(conjunctions[ii]))`

ArielG-NV: the entire context of this file is "relational operations with code"

source/core/slang-uint-set.h
jkwak-work: This utility function is not a good thing to have in general.
If we were able to use STL, there is for_each to do this type of things.

I am just letting you know that it is better not to have these things.
But I wouldn't object too hard for the connivence when there isn't for_each.

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: the code suggestion is incorrect: `conjunctions[x].implies(conjunctions[y]) != conjunctions[y].implies(conjunctions[x])`

source/core/slang-uint-set.h
ArielG-NV: I agree that this function is likely a bad idea from an STL design (not general enough)

maybe change this function to something like:
```
template<typename T>
inline void UIntSet::add(const T& other)
{
    for (auto i : other)
        add((UInt)i);
}
```

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: Please avoid using symbols when it can be simpler in a plan English.
In this specific case, your statement/assertion is not true always.
There surely will be a case where the following condition is true,
```
conjunctions[x].implies(conjunctions[y]) == conjunctions[y].implies(conjunctions[x])
```
Please try to avoid miscommunication more.

csyonghe: The full loops here are correct. Both i and ii needs to be from 0 to the end, because `a.implies(b)` is not equivalent to `b.implies(a)`. We need to make sure both (a,b) and (b,a) are tested.

The algebraic implication operation is not commutative.

ArielG-NV: I will reword the following:

the code suggestion is not going to function as intended: 
`conjunctions[x].implies(conjunctions[y])` is only equal to `conjunctions[y].implies(conjunctions[x])` when `y == x`. Implication is not a commutative operation, we cannot make this change.

source/core/slang-uint-set.h
csyonghe: I am OK with having dedicated `add` that allows us to use a more specialized implementation instead of relying on iterators too heavily.

If we are to follow existing container classes, the `add` here should be renamed to `addRange()` to signify that the operand is a collection.

tools/slang-capability-generator/capability-generator-main.cpp
csyonghe: Seems like there was some misunderstanding here, but I think we are on the same page now that it is OK to just heap allocate `CapabilityDef` and store them in RefPtrs.

The `CapabiltiyDef` is really an AST node here, and it is fine and convenient to just use heap allocation.

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -263,7 +440,71 @@ struct CapabilityDisjunction
...
+            {
+                if (ii == i)
+                    continue;
+                if (conjunctions[i].implies(conjunctions[ii]))
+                {
+                    conjunctions.removeAt(ii);
+                    ii--;
+                    if (i > ii)
+                        i--;
+                }
```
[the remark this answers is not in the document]
ArielG-NV: `fastRemoveAt` will work if we add a branch, yes.
The logic will be a bit different than demonstrated in the example code provided, but there should be no issues.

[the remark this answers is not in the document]
jkwak-work: What "example code" are you talking about?
Is there a documentation related to this code?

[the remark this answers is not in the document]
ArielG-NV: > What "example code" are you talking about?

Your example code. The code you wrote down and provided as an example.

tools/slang-capability-generator/capability-generator-main.cpp
jkwak-work: Yes, there was a miscommunication between me and Ariel. I had a very long conversation with Ariel about this. It was my fault that my question and instruction were not detailed enough and it lead Ariel to believe to make changes in a way that any experienced programmers wouldn't prefer.

I will try to make my questions and instructions more clear next time to avoid a miscommunications like this. For this specific case, I should have asked my question in a following, as an example.
```
Can we change the function local variable to be allocated in the stack memory in the function that the variable is created from. And keep its member variables as it is in heap. And don't make changes to the List variable that the object will be added into.
```
Maybe that's too long but basically my instruction was not clear.

csyonghe (approved):

Issue #3995, opened 2024-04-22
Metal Meshshading support


Dynamitos:
Basic metal support has already been added with #3967, it could be extended to support amplification(object) and mesh shaders. This would be immensely useful for the project I am working on right now, so I would be willing to try myself, although I have no experience with compiler programming, and worry if my contribution would even be helpful.

csyonghe:
Hi @Dynamitos, as you have seen we have began the work to implement the Metal backend. For now our plan is to get to vertex+fragment+compute shader supported first, and make sure most intrinsics are supported, and then move on to mesh shaders and ray tracing. We welcome contributions at anytime. We are also happy to walkthrough the necessary parts of the codebase with you if you would like to help.

zmarlon:
This looks amazing!

Dynamitos:
@csyonghe, I finally had some time to start working on this, starting with something basic to get a feel for the codebase as a whole, though looking at the spec for mesh shading this will be quite challenging. Perhaps you could find some time to look over what I have so far or give some tips on how the mesh output type could be implemented

csyonghe:
Mesh shading support is added with #4280, so closing this issue as completed.

PR #4238, opened 2024-05-28
Metal Task Shader payload

Dynamitos:
Starting to work on https://github.com/shader-slang/slang/issues/3995

I started implementing some basic mesh shading features, for now something basic since this is my first time working with the codebase. Before merging this I would like to at least complete the task shader part by generating the group spawning call. However I cannot figure out how to do a function call on the `mesh_grid_properties` parameter that is emitted.


jkwak-work (commented):
This looks good to me.
But I will defer the approval to Yong.

csyonghe:
@Dynamitos Can we do this for `DispatchMesh`:
In `hlsl.meta.slang`::`DispatchMesh<P>`, add a target switch case for metal:
```
__target_switch
{
...
case metal:
    __intrinsic_asm "_slang_mgp.set_threadgroups_per_grid(uint3($0, $1, $2)); _slang_mesh_payload = $3; return";
}
```
Then you just need to make sure to emit `_slang_mpg` and `_slang_mesh_payload` as additional parameters in the object shader entrypoint, you can inject those parameters during legalization before you hit emit. This will only work if `DispatchMesh` is called directly from the amplification shader and not in a different function.

Dynamitos:
@csyonghe Thanks for the tip, implemented it just now. For now a bit of a hack since I couldn't get the params to be called exactly `_slang_mgp`, and also I'm not sure if passing the payload as pointer even works, but since everything is emitted as a pointer, it looks a little strange, but it should work. Caveat of course is that the call to DispatchMesh needs to happen in the entry point, because the parameters are not passed to other functions.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -306,15 +309,110 @@ namespace Slang
...
+                // Add the MetalPayload resource info, so we can emit [[payload]]
+                varLayoutBuilder.findOrAddResourceInfo(LayoutResourceKind::MetalPayload);
+                auto paramVarLayout = varLayoutBuilder.build();
+                builder.addLayoutDecoration(packedParam, paramVarLayout);
+
+                // Now we replace the call to DispatchMesh with a call to the mesh grid properties
+                // But first we need to create the parameter
+                const auto meshGridPropertiesType = builder.getMetalMeshGridPropertiesType();
+                auto mgp = builder.emitParam(meshGridPropertiesType);
+                builder.addNameHintDecoration(mgp, toSlice("_slang_mgp"));
```
csyonghe, line 399: If you also add `ExternCppDecoration` here, it will be emitted as is without `_0` suffix.

tests/metal/simple-task.slang
```diff
@@ -0,0 +1,102 @@
...
+// TEST_INPUT: ubuffer(data=[0 0 0 0], stride=4):out,name outputBuffer
+
+uniform RWStructuredBuffer<float> outputBuffer;
+
+cbuffer Uniforms
+{
+    float4x4 modelViewProjection;
+}
+
+// CHECK-ASM: define void @taskMain
```
csyonghe, line 14: There is no //CHECK: lines in the file, so test in line 1 will fail

Dynamitos: I'm not entirely sure what those check comments do, copied them from the rasterization test file.

tests/metal/simple-task.slang
```diff
@@ -1,6 +1,11 @@
-//TEST:SIMPLE(filecheck=CHECK): -target metal
-//TEST:SIMPLE(filecheck=CHECK-ASM): -target metallib
-//TEST:REFLECTION(filecheck=REFLECT):-target metal -entry main_kernel -stage task
+// TEST:SIMPLE(filecheck=CHECK): -target metal
+// TEST:SIMPLE(filecheck=CHECK-ASM): -target metallib
+// TEST:REFLECTION(filecheck=REFLECT):-target metal -entry main_kernel -stage task
+
+// CHECK:      0
```
csyonghe, line 5: I don't quite understand these results. Can we put something more meaningful here, such as:
```
// CHECK: _slang_mgp.set_threadgroups_per_grid
```

tests/metal/simple-task.slang
```diff
@@ -1,6 +1,11 @@
-//TEST:SIMPLE(filecheck=CHECK): -target metal
-//TEST:SIMPLE(filecheck=CHECK-ASM): -target metallib
-//TEST:REFLECTION(filecheck=REFLECT):-target metal -entry main_kernel -stage task
+// TEST:SIMPLE(filecheck=CHECK): -target metal
```
csyonghe, line 1: There shouldn't be a space between `//` and `TEST`, otherwise the test won't get picked up by our test driver.

tests/metal/simple-task.slang
csyonghe: They are regular expressions used to match the result of the generated source file or buffer content.

A //CHECK line will try to match a line in the output with the regex following `CHECK:`, if such a match if found, it is successful, otherwise it will fail the test.

tests/metal/simple-task.slang
Dynamitos: I added some checks that should work, however I can't get the tests to run on my macbook, the runner keeps ignoring them.

csyonghe: Our test runner just try to run the `metal` command directly, so it needs to be made available from PATH. We should probably enhance the logic in `locateMetalCompiler` in `slang-metal-compiler.cpp` so that it can discover metal sdk on macos.

tests/metal/simple-task.slang
```diff
@@ -0,0 +1,103 @@
+//TEST:SIMPLE(filecheck=CHECK): -entry taskMain -stage amplification -target metal
+//TEST:SIMPLE(filecheck=CHECK-ASM): -entry taskMain -stage amplification -target metallib
```
csyonghe, line 2: We probably cannot run this test here because we always specify metal2.3 when we run metal compiler. We need to extend the logic to detect the use of metal 3.0 and specify the std accordingly. Before then this is going to fail.

Dynamitos: Should I leave this PR like this for now until this is fixed and start working on the mesh stage on a different branch, or continue working on this one?

csyonghe: Let's disable this asm test by removing this line, and check this PR in.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -306,15 +309,110 @@ namespace Slang
...
+                SLANG_ASSERT(payloadType);
+
+                builder.setInsertBefore(entryPoint.entryPointFunc->getFirstBlock()->getFirstOrdinaryInst());
+                const auto annotatedPayloadType =
+                    builder.getPtrType(
+                        kIROp_RefType,
+                        payloadPtrType->getValueType(),
+                        AddressSpace::MetalObjectData
+                    );
+                auto packedParam = builder.emitParam(annotatedPayloadType);
```
csyonghe, line 386: If there are more than one DispatchMesh call, this will insert duplicate parameters into the entry point. We should probably make sure we insert at most once for the referencing entry point?

csyonghe: Also there could be more than one entry points in a single module. I feel like instead of inserting into the entry point, we should try to find the parent function of the call site, assert that it is a object shader entry point, and insert a param into the EntryPoint if it is t already there.

csyonghe: `auto parentFunc = getParentFunc(call)` should give you the user func, you can then `builder.setInsertInto(parentFunc->getFirstBlock()->getFirstOrdinaryInst())` for the param.

Dynamitos: So we still assume that the DispatchMesh call is done directly from inside the entry point, but we make sure that the payload parameter is only generated once per entrypoint?

csyonghe: Yes, let's do that for now. we can add a diagnostic if oarent func isnt an entrypoint.

csyonghe:
Another way to do it, is to do the mesh EntryPoint legalization before anything else, and just declare a global variable for mgp, and just access that global variable for the call.

Then all these global variable accesses will be legalized together with other global variables (we declare them as entry point parameter and pass them around to other functions via a global context object).

csyonghe:
The call to `introduceExplicitGlobalContext` in slang-emit.cpp:1115 in linkAndOptimizeIR() does the global variable to context object transformation. So this entry point legalization needs to happen before that to take advantage of it. Consider moving it to be called at the same time as glsl mesh shader legalization.

Dynamitos:
How do we handle different entry points using different types for their payloads? Since the variable name is fixed through the intrinsic asm, having two entry points with different data types would maybe be a problem? I don't know the internals well enough to judge that, its just something that comes to mind.

Also, payload data is passed by reference, I don't know if how the interaction with the global context object would work, especially since the payload is located in its own special `object_data` address space.

The HLSL/Direct-X spec states that the [`DispatchMesh` intrinsic can only be 'called exactly once per amplification shader, must not be called from non-uniform flow control'](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#dispatchmesh-intrinsic). Metal would support calling it multiple times, so I would still like to enable that, but perhaps by simply going up the call chain all the way to the entry point and asserting that all of them need to be amplification entry points.

csyonghe:
It maybe fine to define a global variable that is a pointer to the object data address space, and just store into that pointer at DispatchMesh call. Then at the create context pass we can create a field for the context type that is also a pointer in object data address space, and copy the pointer parameter into that field at the beginning of the entry point. The handling logic there should be exactly the same as other in / out global variables. In some sense, this is exactly like a global output variable.

Dynamitos:
I tried to implement it the way you described, but to be completely honest, this might be a little to difficult for me. I understand what you mean and what the result is supposed to be like, but the globals get optimized away and it is pretty late in my timezone, will try again tomorrow.

csyonghe:
No worries. We can check in this PR as is. I can take a look at this once I got time.

Dynamitos:
Should I update the branch or will this be merged into the main automatically?

csyonghe (approved):

csyonghe:
Looks good, merged. Thank you!

Issue #4165, opened 2024-05-14
Capabilities System: If a profile is set and implicitly upgraded, we should emit a warning.


ArielG-NV:
Currently if we do: `-profile _GLSL_150` we won't error if we implicitly upgrade the capability to `_GLSL_400` (or if adding extensions). This should be a warning.

csyonghe:
To clarify the behavior: we should introduce a restrictive mode enabled with -restrictive-capability-check option. In this mode, we diagnose an error whenever an upgrade is needed.

ArielG-NV:
Question about behavior: 
* No warning or error will be thrown if a user does not set a CompilerOptionName::Capability or CompilerOptionName::Profile. 
    * Else a user clearly not trying to use capabilities beyond the convince features will be subject to hundreds of warnings/errors.

Issue #4164, opened 2024-05-14
Capability System: adjust the capabilities of all versions and profiles to include implied extensions 


ArielG-NV:
currently version/profile capabilities are not correct. Capabilities like `GLSL_450` should be defined with implied extensions.

Furthermore, `glsl450` profile should be defined with capability `GLSL_450`

csyonghe:
Currently we use _GL_EXT_FOO to represent the real glsl extension EXT_FOO,
L then we also define a user facing alias GL_EXT_FOO = _GL_EXT_FOO|_SPV_EXT_FOO to mean either the glsl extension or the equivalent spv extension depending on whether the user is targeting glsl or spirv. This setup makes it possible for our users to specify:
-profile spirv+GL_EXT_FOO and we will be able to resolve it to spirv+SPV_EXT_FOO.

We should maintain this setup. I was simply asking us to define GLSL_450 profile such that it implies all extensions for 450:
def glsl450:glsl440, _GL_EXT_FOO, _GL_EXT_BAR, ...

This means that when the user has specified -profile glsl450 they will be able to use ext_foo without having to explicitly specify it again even in restrictive mode. But if they are targeting 430 they will need to do -profile glsl430+_Gl_EXT_FOO.

What is important here is that the canonical representation of glsl450 will contain the _GL_EXT_FOO atom. But we can also know that the atom is implicitly inferred by seeing that the ext atom does not exist in the compacted set.

Issue #4163, opened 2024-05-14
Capabilities System: Capabilities Generator should generate capabilities (precompute) as serialized UIntSet buffers


ArielG-NV:
Generator should precompute relationships into serialized/ready-to-use UIntSet buffers. This should make loading capabilities faster through being more direct

-> ~2% of stdlib compile time

PR #4241, opened 2024-05-29
Capability System: Implicit capability upgrade warning/error

ArielG-NV:
fixes: #4164
fixes: #4165
fixes: #4317

Support a warning/error if capabilities are implicitly upgraded and test accordingly. 

Notable changes:
1. Implicit upgrades are only checked if **Profile/Capability** compiler options are used. Profile is automatically added with render-test in some scenarios.
2. `ImpliesReturnFlags` & `ImpliesFlags` was added to remove `bool` parameters for managing `implies` and to allow for a clearer "implies" operation usage pattern.
3.  `maybeDiagnose` replaces `diagnoseCapabilityErrors` to allow for a more general "type" of error. This was done since "maybeDiagnoseWarningOrError" was added (for togglable warning<->error), and requires a diagnosis wrapper which may not diagnose an error for use with capabilities.
4. `_spirv_*` have been added to handle 'spirv version' capabilities, `spirv_*` are now being used as 'spirv core' (extensions promoted to core are associated with `spirv_*`).
5. `_spirv_*` now translates into `glsl_spirv_*`. This was done since there is not much of a reason to manually specify `glsl_spirv_*` when `spirv` target has a subset of the capabilities of `glsl_spirv_*` (subset since glslang bugs sometimes require inflated versions). 
6. `GLSL_*` and `sm_*` specify a version and cross-platform representation of extensions of the version.


ArielG-NV:
note: profile is automatically set in render test, likely need to fix capability representation in .capdef (#4163) to avoid errors.

csyonghe (commented):
In all of our capability related diagnostic messages, we should try to simplify the atoms so that if `a` implies `b`, we don't print out `b`. For example, if we want to report `spirv_1_5`, we just report `spirv_1_5` instead of `spirv_1_0+spirv_1_1+spirv_1_2+spirv_1_3+spirv_1_4+spirv_1_5`.

ArielG-NV:
> In all of our capability related diagnostic messages, we should try to simplify the atoms so that if a implies b, we don't print out b. For example, if we want to report spirv_1_5, we just report spirv_1_5 instead of spirv_1_0+spirv_1_1+spirv_1_2+spirv_1_3+spirv_1_4+spirv_1_5.

new issue or add to this issue?

source/slang/slang-capability.h
```diff
@@ -139,8 +139,15 @@ struct CapabilitySet
     /// Is this capability set incompatible with the given `other` atomic capability.
     bool isIncompatibleWith(CapabilitySet const& other) const;
 
+    enum class ImpliesReturnFlags : int
+    {
+        NotImplied = 0,
+        Implied = 1 << 1,
```
kaizhangNV, line 145: 1<<0 or 1<<1 ?

csyonghe:
@ArielG-NV Can do a new issue to simplify capability printing.

source/slang/slang-capabilities.capdef
```diff
@@ -383,120 +405,187 @@ alias fragmentshaderinterlock = _GL_ARB_fragment_shader_interlock | hlsl_nvapi |
...
+                ;
+alias spirv_1_5 = _spirv_1_5 + GL_EXT_nonuniform_qualifier + GL_EXT_buffer_reference
+                | spirv_1_4
+                ;
+alias spirv_1_6 = _spirv_1_6 + GL_EXT_debug_printf + GL_EXT_demote_to_helper_invocation
+                | spirv_1_5
+                ;
+alias spirv_latest = _spirv_1_6;
+
+alias sm_4_0_version = _sm_4_0
```
csyonghe, line 439: Maybe add a comment here on what `x_version` means, it means `x` for the corresponding target, and the closest match of equivvalent feature set for other targets.

csyonghe: do we want to automatically translate spirv_1_0 to spirv_1_0_version when the user specifies `slangc shader.slang -profile spirv_1_0_version -target hlsl`?

source/slang/slang-capability.cpp
```diff
@@ -827,6 +848,53 @@ bool CapabilitySet::checkCapabilityRequirement(CapabilitySet const& available, C
...
+inline CapabilityName maybeConvertSpirvVersionToGlslSpirvVersion(CapabilityName& atom)
+{
+    if (atom >= CapabilityName::_spirv_1_0 && asAtom(atom) <= getLatestSpirvAtom())
+    {
+        return (CapabilityName)((Int)CapabilityName::glsl_spirv_1_0 + ((Int)atom - (Int)CapabilityName::_spirv_1_0));
+    }
+    return CapabilityName::Invalid;
+}
+
+void CapabilitySet::AddSpirvVersionFromOtherAsGlslSpirvVersion(CapabilitySet& other)
```
csyonghe, line 861: naming conversion. `Add`->`add`.

source/slang/slang-capabilities.capdef
ArielG-NV: > do we want to automatically translate spirv_1_0 to spirv_1_0_version when the user specifies slangc shader.slang -profile spirv_1_0_version -target hlsl?

1. I did not add spirv_*_version but that is a good idea. 

2. We probably also should also warn if a user uses any capability defined as internal in `slangc` (`_*`)

Should '1'/'2' be another issue or part of this PR?

csyonghe:
It seems this is breaking the `slang-unit-test-tool/getTargetCode` test I added yesterday.

ArielG-NV:
> It seems this is breaking the slang-unit-test-tool/getTargetCode test I added yesterday.

I will check it out

ArielG-NV:
`slang-unit-test-tool/getTargetCode` was requesting `-profile sm_5_0` for a spirv target, this causes a warning since sm_5_0 implies spirv_1_0 is requested

csyonghe (approved):

natevm:
+1 on the simplified compatibility printing. The compatibility system is a good idea, but with too complex of an error printed, I’m more likely to disable the feature than to address it. Programmers are very susceptible to alarm fatigue.

ArielG-NV:
#4347 for tracking the simplified compatibility printing issue

PR #4242, opened 2024-05-29
Add APIs to get profile of compile time

kaizhangNV:
Add serial time measurement

Add profiler to measure lots of stages in slang compilation, and it can accumulate the time spent in each thread in multi-threads case and finally report a serial timing info.

### DIFF
--- a/slang.h
+++ b/slang.h
@@ -1574,6 +1574,16 @@ extern "C"
     
     #define SLANG_UUID_ISlangWriter ISlangWriter::getTypeGuid()
 
+    struct ISlangProfiler : public ISlangUnknown
+    {
+        SLANG_COM_INTERFACE(0x5fb632d2, 0x979d, 0x4481, { 0x9f, 0xee, 0x66, 0x3c, 0x3f, 0x14, 0x49, 0xe1 })
+        virtual SLANG_NO_THROW size_t SLANG_MCALL getEntryCount() = 0;
+        virtual SLANG_NO_THROW const char* SLANG_MCALL getEntryName(uint32_t index) = 0;
+        virtual SLANG_NO_THROW long SLANG_MCALL getEntryTimeMS(uint32_t index) = 0;
+        virtual SLANG_NO_THROW uint32_t SLANG_MCALL getEntryInvocationTimes(uint32_t index) = 0;
+    };
+    #define SLANG_UUID_ISlangProfiler ISlangProfiler::getTypeGuid()
+
     namespace slang {
     struct IGlobalSession;
     struct ICompileRequest;
@@ -2030,7 +2040,8 @@ extern "C"
     /*! @see slang::ICompileRequest::getCompileTimeProfile */
     SLANG_API SlangResult spGetCompileTimeProfile(
         SlangCompileRequest* request,
-        ISlangBlob** compileTimeProfile);
+        ISlangProfiler** compileTimeProfile,
+        bool isClear);
 
 
     /** Extract contents of a repro.
@@ -4430,7 +4441,7 @@ namespace slang
 
         virtual SLANG_NO_THROW void SLANG_MCALL setIgnoreCapabilityCheck(bool value) = 0;
 
-        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangBlob** compileTimeProfile) = 0;
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangProfiler** compileTimeProfile, bool isClear) = 0;
 
     };
 
--- a/slang.h
+++ b/slang.h
@@ -1576,7 +1576,7 @@ extern "C"
 
     struct ISlangProfiler : public ISlangUnknown
     {
-        SLANG_COM_INTERFACE(0x5fb632d2, 0x979d, 0x4481, { 0x9f, 0xee, 0x66, 0x3c, 0x3f, 0x14, 0x49, 0xe1 })
+        SLANG_COM_INTERFACE(0x197772c7, 0x0155, 0x4b91, { 0x84, 0xe8, 0x66, 0x68, 0xba, 0xff, 0x06, 0x19 })
         virtual SLANG_NO_THROW size_t SLANG_MCALL getEntryCount() = 0;
         virtual SLANG_NO_THROW const char* SLANG_MCALL getEntryName(uint32_t index) = 0;
         virtual SLANG_NO_THROW long SLANG_MCALL getEntryTimeMS(uint32_t index) = 0;
@@ -2041,7 +2041,7 @@ extern "C"
     SLANG_API SlangResult spGetCompileTimeProfile(
         SlangCompileRequest* request,
         ISlangProfiler** compileTimeProfile,
-        bool isClear);
+        bool shouldClear);
 
 
     /** Extract contents of a repro.
@@ -4441,7 +4441,8 @@ namespace slang
 
         virtual SLANG_NO_THROW void SLANG_MCALL setIgnoreCapabilityCheck(bool value) = 0;
 
-        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangProfiler** compileTimeProfile, bool isClear) = 0;
+        // return a copy of internal profiling results, and if `shouldClear` is true, clear the internal profiling results before returning.
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangProfiler** compileTimeProfile, bool shouldClear) = 0;
 
     };
 
--- a/slang.h
+++ b/slang.h
@@ -2027,6 +2027,11 @@ extern "C"
     SLANG_API SlangResult spEnableReproCapture(
         SlangCompileRequest* request);
 
+    /*! @see slang::ICompileRequest::getCompileTimeProfile */
+    SLANG_API SlangResult spGetCompileTimeProfile(
+        SlangCompileRequest* request,
+        ISlangBlob** compileTimeProfile);
+
 
     /** Extract contents of a repro.
 
@@ -4424,6 +4429,8 @@ namespace slang
         virtual SLANG_NO_THROW void SLANG_MCALL setTargetUseMinimumSlangOptimization(int targetIndex, bool value) = 0;
 
         virtual SLANG_NO_THROW void SLANG_MCALL setIgnoreCapabilityCheck(bool value) = 0;
+
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangBlob** compileTimeProfile) = 0;
 
     };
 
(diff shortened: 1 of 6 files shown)


source/slang/slang.cpp
```diff
@@ -6203,6 +6204,21 @@ void const* EndToEndCompileRequest::getEntryPointCode(int entryPointIndex, size_
...
 
+SlangResult EndToEndCompileRequest::getCompileTimeProfile(ISlangBlob** compileTimeProfile)
+{
+    if (compileTimeProfile == nullptr)
+    {
+        return SLANG_E_INVALID_ARG;
+    }
+
+    StringBuilder outString;
+    SerialPerformaceProfiler::getProfiler()->accumulateResults(outString, PerformanceProfiler::getProfiler());
```
csyonghe, line 6215: Why do we need to accumulate in a mutex here? Can we simply return what is in PerformanceProfiler::getProfiler() ?

source/core/slang-performance-profiler.cpp
```diff
@@ -39,11 +41,53 @@ namespace Slang
...
+    public:
+       void accumulateResults(StringBuilder& out, PerformanceProfiler* profile)
+       {
+           PerformanceProfilerImpl* profilerImpl = static_cast<PerformanceProfilerImpl*>(profile);
+            for (auto func : profilerImpl->data)
+            {
+                auto milliseconds = std::chrono::duration_cast< std::chrono::milliseconds >(func.value.duration);
+                long ms = milliseconds.count();
+
+                m_mutex.lock();
```
jkwak-work, line 61: I think this mutex_lock will be better go outside of the for-loop.
It will keep locking and unlocking too many times.

source/core/slang-performance-profiler.cpp
```diff
@@ -39,11 +41,53 @@ namespace Slang
...
+                if (entry == m_compileTimeProfiler.end())
+                {
+                    m_compileTimeProfiler.emplace(std::make_pair(func.key, 0l));
+                    entry = m_compileTimeProfiler.find(func.key);
+                }
+                entry->second += ms;
+                m_mutex.unlock();
+            }
+
+            for (auto entry : m_compileTimeProfiler)
```
jkwak-work, line 72: This for-loop will need to be inside of a lock protection.
Because m_compileTimeProfiler can be modified while reading it.

source/core/slang-performance-profiler.cpp
```diff
@@ -1,5 +1,7 @@
 #include "slang-performance-profiler.h"
 #include "slang-dictionary.h"
+#include <atomic>
```
jkwak-work, line 3: Do you still need to include `atomic`?
I am not sure which one on your current code uses it.

source/slang/slang.cpp
kaizhangNV: Then how can the user get the serialized timing? They need to accumulate themselves?

csyonghe: Well, they should because only they know how many threads are created. I actually did not understand when/how things are aggregated give the current implementation.

source/core/slang-performance-profiler.cpp
kaizhangNV: done.

source/core/slang-performance-profiler.cpp
kaizhangNV: done.

source/core/slang-performance-profiler.cpp
kaizhangNV: done.

source/slang/slang.cpp
kaizhangNV: Given current implementation, each slang working thread has to call this API. 
And then the main thread will get the result by calling this API again.

This is kind of like the parallel reduction.

kaizhangNV: we don't need to return the string alternatively, but return the profiler instead. So user can just use the profiler data after `join()` for all the threads.

Basically, the usage for now is
```
void ThreadFunc()
{
     //... some slang work
     SerialPerformaceProfiler profiler;
     spGetCompileTimeProfile(&profiler);
}

void main()
{
      join(thread1);
      ....
      join(threadN);

      SerialPerformaceProfiler profiler;
      spGetCompileTimeProfile(&profiler);
       // code to print profiler
}
```

kaizhangNV: I think what you're thinking about is that each thread will hold a thread local variable, and then the main thread will accumulate them together.
```
SerialPerformaceProfiler globalProfiler;
void ThreadFunc()
{
     //... some slang work
     SerialPerformaceProfiler profiler;
     spGetCompileTimeProfile(&profiler);

     {
         // scope lock
         // accumulate(globalProfiler, profiler);
     }
}
```

It just depends on whether we want to do the accumulation or user.

source/core/slang-performance-profiler.cpp
```diff
@@ -39,11 +40,53 @@ namespace Slang
...
+
+            for (auto entry : m_compileTimeProfiler)
+            {
+                out << entry.first.c_str() << ": \t" << entry.second << " ms\n";
+            }
+       }
+    private:
+        // Those are supposed to accumulate the time spent in each thread
+        std::unordered_map<std::string, long> m_compileTimeProfiler;
+        std::mutex m_mutex;
```
jkwak-work, line 79: Shouldn't this be a static?
As it is, each thread will have their own mutex, isn't it?

kaizhangNV: This class itself is static.

kaizhangNV: I mean the instance of this class is static.

source/slang/slang.cpp
kaizhangNV: @csyonghe can you suggest that which way we want?
Let slang do the accumulation or user do the accumulation?
IMO, either way is fine, we just need to document the behavior carefully. It's just trivial work to change it in the implementation.

kaizhangNV: I also another thought to make the profile in a hierarchical way. That might be a follow-up change.

source/core/slang-performance-profiler.cpp
jkwak-work: Yes, I figured it after leaving my comment.
Thanks for clarifying it tho.

source/slang/slang.cpp
kaizhangNV: If we decide to only measure the time per thread, then we don't need to use lock at all. So the ball will be passed to user.

kaizhangNV: But if that is our decision, I don't think I will need this `SerialPerformaceProfiler` class at all.
But I cannot return a string anymore. Instead, I have to return the profiler to user, so that they can use the profiler to do the follow-up operation. But that will requires us to expose the an interface to user. So I might need to create a struct/class exposed to slang.h used to carry the result.

csyonghe: I think we should let the user aggregate. We can still return a string by serializing the result into a json. But it may be easier to just expose another com interface to represent the structural profiling result.

source/core/slang-performance-profiler.cpp
```diff
@@ -56,26 +56,27 @@ namespace Slang
...
             {
-                out << entry.first.c_str() << ": \t" << entry.second << " ms\n";
+                auto milliseconds = std::chrono::duration_cast< std::chrono::milliseconds >(entry.second.duration);
+                out << entry.first.c_str() << ": \t" << entry.second.invocationCount << "\t" << milliseconds.count() << " ms\n";
             }
        }
     private:
         // Those are supposed to accumulate the time spent in each thread
-        std::unordered_map<std::string, long> m_compileTimeProfiler;
+        std::unordered_map<std::string, FuncProfileInfo> m_compileTimeProfiler;
```
csyonghe, line 79: Any reason not to use `Dictionary<String, FuncProfileInfo>` here?

csyonghe (approved):

source/core/slang-performance-profiler.cpp
kaizhangNV: Since we decide to let user aggregate the result, I'd like to return the map to user directly.
But `Dictionary` is not a public interface to user. So maybe it's better to return the cpp standard map?

csyonghe: The c++ std data structure provides no binary compatibility guarantee either. That means a gcc compiled slang.dll won't be usable by  a clang compiled application. We should instead just serialize the data and return the serialized blob.

source/core/slang-performance-profiler.cpp
```diff
@@ -1,13 +1,9 @@
 #include "slang-performance-profiler.h"
 #include "slang-dictionary.h"
+#include <unordered_map>
```
csyonghe, line 3: I guess we can just use our own containers for consistency with the rest of the code base.

csyonghe (dismissed):
So the supposed use right now is to call isClear=true before each compile, and then call again after the compile and accumulate the values in user space?

kaizhangNV:
> So the supposed use right now is to call isClear=true before each compile, and then call again after the compile and accumulate the values in user space?

It's actually called after each compile, e.g.

```
spCompile();
bool isClear = true;
spGetCompileProfile(..., isClear);
```

in this case, the profiling data for this compile job will be returned to user, and slang will clear it.

source/core/slang-performance-profiler.cpp
kaizhangNV: yea, I can change the std::vector to List then.

slang.h
```diff
@@ -4425,6 +4441,8 @@ namespace slang
 
         virtual SLANG_NO_THROW void SLANG_MCALL setIgnoreCapabilityCheck(bool value) = 0;
 
+        virtual SLANG_NO_THROW SlangResult SLANG_MCALL getCompileTimeProfile(ISlangProfiler** compileTimeProfile, bool isClear) = 0;
```
csyonghe, line 4444: Maybe name it as "shouldClear", and add a comment on what it means -- "return a copy of internal profiling results, and if shouldClear is true, clear the profiling counters before returning."

csyonghe (approved):

slang.h
kaizhangNV: done

source/core/slang-performance-profiler.cpp
kaizhangNV: Done

csyonghe (approved):

csyonghe:
The use of `strncpy` caused a warning in visual studio. I believe you can just use `strncpy_s`, and we should have this defined for linux in `slang-secure-crt.h`.

kaizhangNV:
> strncpy_s



> The use of `strncpy` caused a warning in visual studio. I believe you can just use `strncpy_s`, and we should have this defined for linux in `slang-secure-crt.h`.



> The use of `strncpy` caused a warning in visual studio. I believe you can just use `strncpy_s`, and we should have this defined for linux in `slang-secure-crt.h`.

Yeah, updated.

csyonghe (approved):

PR #4245, opened 2024-05-30
Fix confusion in Translation Units doc

eliemichel:
I think the sentence was saying the opposite of what it meant!


jkwak-work (commented):
This makes sense to me.
But I will defer the approval to Yong

csyonghe (approved):

csyonghe:
Your understanding is right, the doc is saying the opposite thing. Thanks for fixing!

PR #4249, opened 2024-05-30
Fix a bug on default initialization of interface typed value.

### LINKED ISSUE #3831: Initializing interface type with empty initializer results in invalid HLSL code
When you have a interface type `IHitInfo` and try to empty initialize it directly as `IHitInfo hit = {};` or indirectly by storing it in `struct PathState` and then empty initialize that `PathState path = {}`, Slang generates an invalid HLSL code dxc errors out.

Ideally Slang compiler would report an error, to make it easier to address.

csyonghe:
`default(IFoo)` should lower into a `DefaultConstruct : IFoo` inst instead of `MakeStruct : IFoo` for our passes to handle it correctly to produce valid HLSL output.

Closes #3831.

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -4413,14 +4413,14 @@ struct ExprLoweringVisitorBase : public ExprVisitor<Derived, LoweredValInfo>
             {
                 return LoweredValInfo::simple(getBuilder()->getIntValue(irType, 0));
             }
+            else if (declRef.as<InterfaceDecl>())
+            {
+                return LoweredValInfo::simple(getBuilder()->emitDefaultConstruct(irType));
+            }
             else if (auto aggTypeDeclRef = declRef.as<AggTypeDecl>())
             {
-                if (aggTypeDeclRef.as<InterfaceDecl>())
-                {
-                    return LoweredValInfo::simple(getBuilder()->emitDefaultConstruct(irType));
-                }
-
                 List<IRInst*> args;
+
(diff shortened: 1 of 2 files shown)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -4415,8 +4415,12 @@ struct ExprLoweringVisitorBase : public ExprVisitor<Derived, LoweredValInfo>
             }
             else if (auto aggTypeDeclRef = declRef.as<AggTypeDecl>())
             {
-                List<IRInst*> args;
+                if (aggTypeDeclRef.as<InterfaceDecl>())
+                {
+                    return LoweredValInfo::simple(getBuilder()->emitDefaultConstruct(irType));
+                }
```
jkwak-work, line 4421: I have a question here.
Can we cast to `InterfaceDecl` before casting to `AggTypeDecl`?
Or does it have to check `AggTypeDecl` first for a reason?

In other words, can this be changed to the following?
```
            else if (aggTypeDeclRef.as<InterfaceDecl>())
            {
                return LoweredValInfo::simple(getBuilder()->emitDefaultConstruct(irType));
            }
            else if (auto aggTypeDeclRef = declRef.as<AggTypeDecl>())
            {
```

jkwak-work (dismissed):
Looks good to me.

source/slang/slang-lower-to-ir.cpp
csyonghe: You are right, this is definitely cleaner. changed.

PR #4255, opened 2024-06-01
Print warning when operator<< shifting too much

### LINKED ISSUE #3944: Operator << incorrectly truncates values.

jkwak-work:
Closes #3944

For the given type of the left side operand to `operator<<` is not big enough for the right side operand, print a warning that the result will be always zero.

### DIFF
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -740,7 +740,7 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
-DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount, `$1`, is bigger than the bit count of type, `$0`.")
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount exceeds the number of bits, (`$0` << `$1`).")
 
 DIAGNOSTIC(41901, Error, unsupportedUseOfLValueForAutoDiff, "unsupported use of L-value for auto differentiation.")
 DIAGNOSTIC(41902, Error, cannotDifferentiateDynamicallyIndexedData, "cannot auto-differentiate mixed read/write access to dynamically indexed data in '$0'.")
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -740,7 +740,7 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
-DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "the type of left hand side operand, `$0`, for operatnr<< is not big enough for the shifting amount, `$1`.")
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount, `$1`, is bigger than the bit count of type, `$0`.")
 
 DIAGNOSTIC(41901, Error, unsupportedUseOfLValueForAutoDiff, "unsupported use of L-value for auto differentiation.")
 DIAGNOSTIC(41902, Error, cannotDifferentiateDynamicallyIndexedData, "cannot auto-differentiate mixed read/write access to dynamically indexed data in '$0'.")
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -740,7 +740,7 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
-DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount exceeds the number of bits, (`$0` << `$1`).")
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount exceeds the number of bits and the result will be always zero, (`$0` << `$1`).")
 
 DIAGNOSTIC(41901, Error, unsupportedUseOfLValueForAutoDiff, "unsupported use of L-value for auto differentiation.")
 DIAGNOSTIC(41902, Error, cannotDifferentiateDynamicallyIndexedData, "cannot auto-differentiate mixed read/write access to dynamically indexed data in '$0'.")
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -740,6 +740,8 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "the type of left hand side operand, `$0`, for operatnr<< is not big enough for the shifting amount, `$1`.")
+
 DIAGNOSTIC(41901, Error, unsupportedUseOfLValueForAutoDiff, "unsupported use of L-value for auto differentiation.")
 DIAGNOSTIC(41902, Error, cannotDifferentiateDynamicallyIndexedData, "cannot auto-differentiate mixed read/write access to dynamically indexed data in '$0'.")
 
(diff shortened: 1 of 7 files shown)


source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,62 @@
...
+                    auto rhsLit = as<IRIntLit>(rhs);
+                    if (!rhsLit)
+                        continue;
+
+                    IRIntegerValue shiftAmount = rhsLit->getValue();
+
+                    IRInst* lhs = opShiftLeft->getOperand(0);
+                    IRUse lhsType = lhs->typeUse;
+
+                    // TODO: Not sure how to get the size of the type.
```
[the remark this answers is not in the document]
csyonghe:
```
IRSizeAndAlignment sizeAlignment;
if (SLANG_SUCCEED(getNaturalSizeAndAlignment(lhs->getDataType(), sizeAlignment))
    ...
```

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,62 @@
...
+                    {
+                        (void)lhsType;
+                        printf("JKWAK found 'operator<<' shifting by %" PRId64 "\n", shiftAmount);
+                        //sink->diagnose(child, Diagnostics::operatorShiftLeftOverflow);
+                    }
+                }
+            }
+        }
+
+        // TODO: Need to figure out if this can be just "inst->getChildren()"
```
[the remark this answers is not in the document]
csyonghe: GetChildren should be fine. The difference is that getDecorationAndChildren also includes all decorations. There is no code inside decorations so it is fine to skip them.

source/slang/slang-ir-operator-shift-overflow.cpp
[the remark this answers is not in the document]
jkwak-work: I couldn't use `getNaturalSizeAndAlignment()`.
The following is the funciton signature of it,
```
Result getNaturalSizeAndAlignment(CompilerOptionSet& optionSet, IRType* type, IRSizeAndAlignment* outSizeAndAlignment);
```
I couldn't provide the first argument.
This function is mostly called from "slang-emit-XXX.cpp", which has an access to "targetProgram".
But on my case, this function is called during "generateIR()", which is too early and don't have an access to "targetProgram" yet.

I had to manually get the type sizes but I am not sure about the size of "intptr_t" and "uintptr_t" cases.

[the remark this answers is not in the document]
csyonghe: You can use the option from the linkage->m_optionSet. The linkage can be obtained from codeGenContext.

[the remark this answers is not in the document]
jkwak-work: Thank you for the information. I will try it.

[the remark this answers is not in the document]
jkwak-work: That worked and now the code looks cleaner.
Thank you for the help.

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,66 @@
+// slang-ir-operator-shift-overflow.cpp
```
sriramm-nv, line 1: please add a header with some details about this file.

source/slang/slang-diagnostic-defs.h
```diff
@@ -740,6 +740,8 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount, `$1`, is bigger than the bit count of type, `$0`.")
```
sriramm-nv, line 743: how about?
"left shift amount, `$1`, exceeds the maximum number of bits supported by the data type, `$0`.")

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "slang-ir-operator-shift-overflow.h"
```
sriramm-nv, line 2: you might want to rename this into something more generic, so that additional checks can be added here. 

lets say, if you want to check for valid integer promotions for various other operators, other than shift..

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    class DiagnosticSink;
+    struct IRModule;
+
+    void checkForOperatorShiftOverflowRecursive(
+        IRInst* inst,
+        CompilerOptionSet& optionSet,
+        DiagnosticSink* sink)
+    {
+        if (auto code = as<IRGlobalValueWithCode>(inst))
```
sriramm-nv, line 19: in some code,  i noticed that we have patterns like 
auto code = as<IRGlobalValueWithCode>(inst);
if (!code)
 continue;

this is done to keep the nesting managable. but I leave this up to you.

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+                    IRInst* lhs = opShiftLeft->getOperand(0);
+                    IRType* lhsType = lhs->getDataType();
+
+                    IRSizeAndAlignment sizeAlignment;
+                    if (SLANG_FAILED(getNaturalSizeAndAlignment(optionSet, lhsType, &sizeAlignment)))
+                        continue;
+
+                    IRIntegerValue shiftAmount = rhsLit->getValue();
+                    if (sizeAlignment.size * 8 <= shiftAmount)
```
sriramm-nv, line 43: instead of using a magic number 8, you may choose to define it as a const / compile-time macro.

tests/diagnostics/warning_operator_left_shift_overflow.slang
```diff
@@ -0,0 +1,27 @@
...
+
+    uint result = (u8 << 0)
+        + (u8 << 1)
+        + (u8 << 2)
+        + (u8 << 3)
+        + (u8 << 4)
+        + (u8 << 5)
+        + (u8 << 6)
+        + (u8 << 7)
+        // CHECK-NOT: warning 41030:
```
sriramm-nv, line 19: is this `check-not` required?

sriramm-nv (approved):
Overall I don't have any complaints with the implementation. Just some minor cleanup suggested. Please feel free to merge this as you see it fit

source/slang/slang-diagnostic-defs.h
jkwak-work: I like "exceeds" and "number of" is probably more natural in English expression than "count of".
I will change the message.
But there is no "maximum" nor minimum for the number of bits, so I will skip that word.

I am little more concerned about showing the "$1" in the middle of the message tho.
Probably I should move it to the end of the line like this,
```
left shift amount exceeds the number of bits, (`$0` << `$1`).
```

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: I may need to do it to handle the other operators.
But the approach for the other operators are not decided yet so I wouldn't be preemptive about it yet.

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: Aha, you are talking about the indentation.
You are right, I can change it here.

FYI, this line was copied from another file, slang-ir-missing-return.cpp.

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: I think 8 is pretty straight forward.
There is absolutely zero chance that the number 8 will be changed to other numbers like 7 or 9.
Using a constant variable for this seems an unnecessary overkill with less readability.

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: There is a chance that this file may evolve to do something more than what it does now.
It seems too early to write a descriptive comment for it at this point, because it will stale and add more cost of code/comment maintenance for no good reason.

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: It turned out I cannot do it for this case.
There is another block down there that needs to be executed regardless code is null pointer or not.

sriramm-nv (dismissed):

sriramm-nv:
looks good to me

source/slang/slang-ir-operator-shift-overflow.cpp
```diff
@@ -0,0 +1,66 @@
...
+    void checkForOperatorShiftOverflowRecursive(
+        IRInst* inst,
+        CompilerOptionSet& optionSet,
+        DiagnosticSink* sink)
+    {
+        if (auto code = as<IRGlobalValueWithCode>(inst))
+        {
+            for (auto block : code->getBlocks())
+            {
+                for (auto opShiftLeft : block->getChildren())
```
csyonghe, line 23: In the future we mostly likely are going to extend this pass to check for other operators as well. So might be good to turn this into
```
for (auto inst : block->getChildren())
{
    switch(inst->getOp()) { ... }
}
```

csyonghe (dismissed):

source/slang/slang-ir-operator-shift-overflow.cpp
jkwak-work: `inst` is already in use.
I will go with... `opInst`

source/slang/slang-diagnostic-defs.h
```diff
@@ -740,6 +740,8 @@ DIAGNOSTIC(41021, Error, differentiableFuncMustHaveOutput, "a differentiable fun
 DIAGNOSTIC(41022, Error, differentiableFuncMustHaveInput, "a differentiable function must have at least one differentiable input.")
 DIAGNOSTIC(41023, Error, getStringHashMustBeOnStringLiteral, "getStringHash can only be called when argument is statically resolvable to a string literal")
 
+DIAGNOSTIC(41030, Warning, operatorShiftLeftOverflow, "left shift amount exceeds the number of bits, (`$0` << `$1`).")
```
csyonghe, line 743: Maybe also add that "..., the result is always 0."

csyonghe (approved):

PR #4260, opened 2024-06-03
Add an option to disable source map in obfuscation

kaizhangNV:


source/slang/slang-lower-to-ir.cpp
```diff
@@ -10980,13 +10980,13 @@ RefPtr<IRModule> generateIRForTranslationUnit(
         //
         eliminateDeadCode(module, dceOptions);
 
+        if (stripOptions.stripSourceLocs && linkage->m_optionSet.shouldHaveSourceMap())
```
csyonghe, line 10983: Isn't this `linkage->m_optionSet.shouldObfuscateCode() && linkage->m_optionSet.shouldHaveSourceMap()`?

kaizhangNV: I updated the patch.

csyonghe: what is the difference between `shouldObfuscateCode` and `shouldStripNameHints`?

PR #4261, opened 2024-06-03
Fix performance issue in obfuscate source loc

csyonghe:


source/compiler-core/slang-source-map.cpp
```diff
@@ -148,7 +148,7 @@ void SourceMap::advanceToLine(Index nextLineIndex)
     const auto lastEntryIndex = m_lineEntries.getCount();
 
     // For all the new entries they will need to point to the end 
-    m_lineStarts.setCount(nextLineIndex + 1);
+    m_lineStarts.growToCount(nextLineIndex + 1);
```
kaizhangNV, line 151: Can we just set the capacity to `uniqueLocCount+1` outside the loop such that we don't need to keep increasing the space?

kaizhangNV: Yea, maybe I'm wrong with the intention here.

I just infer the capacity from reference code there:
```
    for (Index i = 0; i < uniqueLocCount; ++i)
    {
        sourceMap->advanceToLine(i);
    }
```

so it will be `m_lineStarts.setCount(1)` to `m_lineStarts.setCount(uniqueLocCount)`

PR #4268, opened 2024-06-04
Fixes incorrect profile string in hello-word example

chaoticbob:
This CL fixes a typo in the profile string for the hello-world example. Currently `"glsl440"` is passed into `findProfile()` resulting in a return value of `SLANG_UNKNOWN_PROFILE` since format of the profile name is incorrect. The correct profile name is `glsl_440`, therefore the string passed into `findProfile()` should be `"glsl_440"` which results in valid profile value.

### DIFF
--- a/examples/hello-world/main.cpp
+++ b/examples/hello-world/main.cpp
@@ -114,7 +114,7 @@ int HelloWorldExample::createComputePipelineFromShader()
     slang::SessionDesc sessionDesc = {};
     slang::TargetDesc targetDesc = {};
     targetDesc.format = SLANG_SPIRV;
-    targetDesc.profile = slangGlobalSession->findProfile("glsl440");
+    targetDesc.profile = slangGlobalSession->findProfile("glsl_440");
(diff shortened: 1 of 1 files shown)


jkwak-work (approved):
Looks good to me

examples/hello-world/main.cpp
```diff
@@ -114,7 +114,7 @@ int HelloWorldExample::createComputePipelineFromShader()
     slang::SessionDesc sessionDesc = {};
     slang::TargetDesc targetDesc = {};
     targetDesc.format = SLANG_SPIRV;
-    targetDesc.profile = slangGlobalSession->findProfile("glsl440");
+    targetDesc.profile = slangGlobalSession->findProfile("glsl_440");
```
csyonghe, line 117: This example is compiling to Vulkan, so it seems the right profile to use here is `spirv_1_5`.

jkwak-work: @chaoticbob , can you make a new change based on Yong's suggestion?
It appears that I approved the change sooner than I should have.

chaoticbob: @jkwak-work of course. Done in pr#4297.

PR #4269, opened 2024-06-04
Precompute UIntSet from individual capabilities inside generator

ArielG-NV:
fixes: #4163

Precompute UIntSet from individual capabilities inside generator (removes intermediate form of capabilities).

generator changes:
1. OutputUIntSetGenerator
    * Added to serialize a generator function which generates a UIntSet
2. SerializedConjunction
    * Added to store serialization information of a single conjunction with the data of "generator function name" and a UIntSet cache. Storing a UIntSet cache of the 'generator function' makes finding duplicate conjunctions easier and faster.
3. Generator output has 3 objects
    * `kCapabilityArray` --> has conjunction data
    * `kCapabilityConjunctions` --> has list of pointers to `kCapabilityArray` which makeup a list of conjunctions for an atom. There is no separator between lists.
    * `kCapabilityNameInfos` --> actual conjunction info is fetched by pointing to `kCapabilityConjunctions` with an offset and array size
4. CapabilitySet::CapabilitySet
    * changed constructor to support UIntSet initialization instead of a intermediate form.
    * We still expand capabilities which are missing `target` and `stage` atoms.
5. A few other functions were changed to support a UIntSet for when fetching atom info.

### DIFF
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -19,20 +19,22 @@ constexpr Index intLog2(unsigned x)
     return x == 1 ? 0 : 1 + intLog2(x >> 1);
 }
 
+// if `in` is 0, result is undefined behavior
 static inline Index bitscanForward(uint64_t in)
 {
+    assert(in != 0);
 #if defined(_MSC_VER)
 
 #ifdef _WIN64
     uint64_t out = 0;
-    assert(_BitScanForward64((unsigned long*)&out, in));
+    _BitScanForward64((unsigned long*)&out, in);
     return Index(out);
 #else
     uint32_t out;
     // check for 0s in 0bit->31bit. If all 0's, check for 0s in 32bit->63bit
     if (_BitScanForward((unsigned long*)&out, *(((uint32_t*)&in) + 1)))
         return Index(out);
-    assert(_BitScanForward((unsigned long*)&out, *(((uint32_t*)&in))));
+    auo ret = _BitScanForward((unsigned long*)&out, *(((uint32_t*)&in)));
     return Index(out);
 #endif// #ifdef _WIN64
 
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -22,7 +22,7 @@ constexpr Index intLog2(unsigned x)
 // if `in` is 0, result is undefined behavior
 static inline Index bitscanForward(uint64_t in)
 {
-    assert(in != 0);
+    SLANG_ASSERT(in != 0);
 #if defined(_MSC_VER)
 
 #ifdef _WIN64
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -22,9 +22,8 @@ constexpr Index intLog2(unsigned x)
 // if `in` is 0, result is undefined behavior
 static inline Index bitscanForward(uint64_t in)
 {
-    SLANG_ASSERT(in != 0);
+    assert(in != 0);
 #if defined(_MSC_VER)
-
 #ifdef _WIN64
     uint64_t out = 0;
     _BitScanForward64((unsigned long*)&out, in);
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -22,8 +22,9 @@ constexpr Index intLog2(unsigned x)
 // if `in` is 0, result is undefined behavior
 static inline Index bitscanForward(uint64_t in)
 {
-    assert(in != 0);
+    SLANG_ASSERT(in != 0);
 #if defined(_MSC_VER)
+
 #ifdef _WIN64
     uint64_t out = 0;
     _BitScanForward64((unsigned long*)&out, in);
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -34,7 +34,7 @@ static inline Index bitscanForward(uint64_t in)
     // check for 0s in 0bit->31bit. If all 0's, check for 0s in 32bit->63bit
     if (_BitScanForward((unsigned long*)&out, *(((uint32_t*)&in) + 1)))
         return Index(out);
-    auo ret = _BitScanForward((unsigned long*)&out, *(((uint32_t*)&in)));
+    _BitScanForward((unsigned long*)&out, *(((uint32_t*)&in)));
     return Index(out);
 #endif// #ifdef _WIN64
 
--- a/source/core/slang-uint-set.h
+++ b/source/core/slang-uint-set.h
@@ -25,17 +25,15 @@ static inline Index bitscanForward(uint64_t in)
 
 #ifdef _WIN64
     uint64_t out = 0;
-    _BitScanForward64((unsigned long*)&out, in);
+    assert(_BitScanForward64((unsigned long*)&out, in));
     return Index(out);
 #else
-    constexpr uint32_t bitsInType = sizeof(uint32_t) * 8;
     uint32_t out;
     // check for 0s in 0bit->31bit. If all 0's, check for 0s in 32bit->63bit
-    _BitScanForward((unsigned long*)&out, *(((uint32_t*)&in) + 1));
-    if (out != bitsInType)
+    if (_BitScanForward((unsigned long*)&out, *(((uint32_t*)&in) + 1)))
         return Index(out);
-    _BitScanForward((unsigned long*)&out, *(((uint32_t*)&in)));
-    return Index(out + bitsInType);
+    assert(_BitScanForward((unsigned long*)&out, *(((uint32_t*)&in))));
+    return Index(out);
 #endif// #ifdef _WIN64
 
 #else 
--- a/source/slang/slang-capability.cpp
+++ b/source/slang/slang-capability.cpp
@@ -167,6 +167,11 @@ bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base)
 
 //// CapabiltySet
 
+// MSVC incorrectly throws warning
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable:4702)
+#endif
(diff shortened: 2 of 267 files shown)


tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -755,50 +768,56 @@ SlangResult generateDefinitions(DiagnosticSink* sink, List<RefPtr<CapabilityDef>
             anyStageAtomSet.add(def->enumValue);
         }
     }
-    outputUIntSetAsBufferValues("kAnyTargetUIntSetBuffer", anyTargetUIntSetHash, anyTargetAtomSet);
-    outputUIntSetAsBufferValues("kAnyStageUIntSetBuffer", anyStageUIntSetHash, anyStageAtomSet);
-    
+    outputUIntSetGenerator("generatorOf_kAnyTargetUIntSetBuffer", anyTargetUIntSetHash, anyTargetAtomSet);
```
csyonghe, line 771: Why is the buffer of elements called "generatorOf_" in the generated code? Maybe just "rawBuffer_`

tools/slang-capability-generator/capability-generator-main.cpp
```diff
@@ -657,22 +657,35 @@ void calcCanonicalRepresentations(DiagnosticSink* sink, List<RefPtr<CapabilityDe
...
-    // underlying type may change, bits stay the same.
-    resultBuilder << "inline static CapabilityAtomSet generate_" << nameOfBuffer << "()\n";
-    resultBuilder << "{\n";
-    resultBuilder << "    CapabilityAtomSet generatedSet;\n";
-
+    resultBuilder << "    CapabilityAtomSet " << nameOfBuffer << ";\n";
     for (Index i = 0; i < set.getBuffer().getCount(); i++)
     {
-        resultBuilder << "    generatedSet.addRawElement(UIntSet::Element(" << set.getBuffer()[i] << "), " << i << ");\n";
+        resultBuilder << "    " << nameOfBuffer << ".addRawElement(UIntSet::Element(" << set.getBuffer()[i] << "), " << i << "); \n";
```
csyonghe, line 666: Can we reserve the buffer space before adding the elements? Something like "nameOfBuffer.reserveRawBuffer(set.getBuffer().getCount())`

csyonghe (approved):

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: `kAnyTargetUIntSetBuffer` is the buffer, `generatorOf_kAnyTargetUIntSetBuffer` is the value generator for `kAnyTargetUIntSetBuffer`

usage is as follows:
`const static CapabilityAtomSet kAnyTargetUIntSetBuffer = generatorOf_kAnyTargetUIntSetBuffer();`

tools/slang-capability-generator/capability-generator-main.cpp
ArielG-NV: we can 👍

source/slang/slang-capability.cpp
```diff
@@ -170,121 +158,107 @@ bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base)
...
 }
 
 //// CapabiltySet
 
-void CapabilitySet::addToTargetCapabilityWithValidUIntSetAndTargetAndStage(CapabilityName target, CapabilityName stage, CapabilityAtomSet setToAdd)
+// MSVC incorrectly throws warning
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable:4702)
+#endif
```
jkwak-work, line 174: This seems to be level 4 warning. I guess level 4 warning is working after all

source/slang/slang-capability.cpp
```diff
@@ -170,121 +158,107 @@ bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base)
...
-    for(auto i : canonicalRepresentation)
-        setToAdd.add(asAtomUInt(i));
-
-    addToTargetCapabilityWithValidUIntSetAndTargetAndStage(target, stage, setToAdd);
+    auto stageSet = getAtomSetOfStages();
+    CapabilityAtomSet out;
+    CapabilityAtomSet::calcIntersection(out, stageSet, atomSet);
+    for(auto i : out)
+        return (CapabilityAtom)i;
+    return CapabilityAtom::Invalid;
```
jkwak-work, line 192: This looks little strange.
It will return either Invalid or the first element of `out`.
I don't see why `for` loop is used.
That might be why you were getting the unreachable code warning

source/slang/slang-capability.cpp
```diff
@@ -170,121 +158,107 @@ bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base)
...
-        for (int i = 0; i < canonicalRepresentation.getCount(); i++)
-            setToAdd.add((UInt)canonicalRepresentation[i]);
-        CapabilityName targetAtom{};
-        for (const auto& targetAtomCanonicalRep : anyTargetInfo.canonicalRepresentation)
+        CapabilityName atom = (CapabilityName)i;
+        CapabilityAtomSet conjunctionPermutation = setToPermutate;
+        auto targetInfo = _getInfo(atom);
+        conjunctionPermutation.add(*targetInfo.canonicalRepresentation[0]);
+
+        if constexpr (keyholeAtomToPermuteWith == CapabilityName::target)
```
jkwak-work, line 208: Is it necessary to use `constexpr` here?
It wouldn't break any but it doesn't look like it will do any either.

ArielG-NV: `if` statments are not ensured to resolve at compile time unless `constexpr` (even if the parameters are clearly known at compile time).

In practice this change will at the very least ensure the debug build resolves the `if` at compile time.

source/slang/slang-capability.cpp
ArielG-NV: A `for` loop is used since `out` is a UIntSet. A UIntSet is non trivial to get element's from and check it empty. Due to this, a iterator is used to check for a first element to keep the code readable.

The `for` loop is why MSVC misdiagnosis the code. This is just a MSVC compiler bug.

jkwak-work: I wouldn't say it is a bug because it is just warning to prevent common mistakes.
It is to warn that there "might" be a mistake.

jkwak-work: I recommend you not to silence the warning and change to code as following,
```
    auto iter = out.begin();
    if (iter == out.end())
        return CapabilityAtom::Invalid;
    return (CapabilityAtom )*iter;
```

source/slang/slang-capability.cpp
jkwak-work: Are you sure that the condition wouldn't be resolved at the compile time?
```
(keyholeAtomToPermuteWith == CapabilityName::target)
```
I will be very surprised if this wouldn't be resolved at a compile time.
We heavily relied on this type of compile time resolving feature to optimize the for-loop on PS3.

The keyword wouldn't hurt to have here.
But it is not a good practice to use keyword inappropriately.
It draws an unnecessary attention for the other engineers who will need to read the lines.

ArielG-NV: > Are you sure that the condition wouldn't be resolved at the compile time?

That is not the reason `constexpr` exists in C++. Anything that is compile time evaluable can be evaluated by the compiler.
`constexpr` enforces behavior (or errors) to no longer rely on undefined compiler behavior.

I agree it likely will be evaluated at least in release. Debug builds likely will not. I can remove the `constexpr` since it is not required.

ArielG-NV: turns out it is a warning to not use `constexpr` here

jkwak-work: I think you said it backward.
But anyway, if it prints a warning, let's keep it.

ArielG-NV: > I think you said it backward. 

I don't see the problem in what I wrote?

jkwak-work: Sorry. I misread it.
Thanks for fixing the indentation at the next if-condition too.

jkwak-work (approved):
Looks good to me.

ArielG-NV:
note about CI failure: some issue with code additions is causing Win32 build to hang

source/core/slang-uint-set.h
```diff
@@ -19,23 +19,23 @@ constexpr Index intLog2(unsigned x)
     return x == 1 ? 0 : 1 + intLog2(x >> 1);
 }
 
+// if `in` is 0, result is undefined behavior
 static inline Index bitscanForward(uint64_t in)
 {
+    assert(in != 0);
```
jkwak-work, line 25: Avoid using the raw `assert`.
```
SLANG_ASSERT(in != 0);
```

source/core/slang-uint-set.h
```diff
@@ -19,23 +19,23 @@ constexpr Index intLog2(unsigned x)
...
 #if defined(_MSC_VER)
 
 #ifdef _WIN64
     uint64_t out = 0;
     _BitScanForward64((unsigned long*)&out, in);
     return Index(out);
```
jkwak-work, line 31: I saw that one of your code change wrapped __BitScanForward64 with `assert`.
Anything inside of `assert` will be removed in Release build so don't do it.
If you must, you can use SLANG_RELEASE_ASSERT`.

ArielG-NV: I made a mistake when I did `assert()` with a function that has a 'out' parameter, I forgotten the function had a 'out' parameter when I made the change, this is why I adjusted the assert accordingly in the commit after.

I did not know Slang had a `SLANG_RELEASE_ASSERT`, thank you for bringing it up.

csyonghe (approved):

PR #4270, opened 2024-06-05
Use memcpy to replace strncpy_s

kaizhangNV:
Use memcpy to replace strncpy_s in SlangProfiler::SlangProfiler to fix the error in Windows.

### DIFF
--- a/source/core/slang-performance-profiler.cpp
+++ b/source/core/slang-performance-profiler.cpp
@@ -65,7 +65,7 @@ namespace Slang
 
             if (strSize > 0)
             {
-                strncpy_s(profileEntry.funcName, strSize, func.key, strSize);
+                memcpy(profileEntry.funcName, func.key, strSize);
             }
             profileEntry.invocationCount = func.value.invocationCount;
             profileEntry.duration = func.value.duration;


source/core/slang-performance-profiler.cpp
```diff
@@ -65,7 +65,7 @@ namespace Slang
 
             if (strSize > 0)
             {
-                strncpy_s(profileEntry.funcName, strSize, func.key, strSize);
+                memcpy(profileEntry.funcName, func.key, strSize);
```
jkwak-work, line 68: It looks like you will be missing the null terminator with the new code.
As an example, when strlen returns a value of 1. You need copy two characters not one because you will need to include the null terminator.
Isn't it?

kaizhangNV: I don't think It will miss the null terminator.
Because:

1. the funcName is initialize to all 0
2. the strSize is min(sizeof(funcName)-1, strlen(func,key)), so it means that I will copy at most `sizeof(funcName)-1` characters, so that will always make funcName has null terminator at end.

jkwak-work: Probably we should not initialize the member to zero and copy the null terminator explicitly.
It is easy to fix by adding 1 like following,
```
memcpy(profileEntry.funcName, func.key, strSize + 1);
```

source/core/slang-performance-profiler.cpp
```diff
@@ -65,7 +65,7 @@ namespace Slang
...
             }
```
jkwak-work, line 69: Also note that when `strlen()` returns zero, you need to copy a null terminator.
The current code will unchange the name, which is probably not what is indended.

source/core/slang-performance-profiler.cpp
jkwak-work: Actually I take this back.
The last character might not be a null terminator if it was longer than 255.
It should be more like
```
profileEntry.funcName[strSize] = 0;
```

jkwak-work:
I think a better question is whether we should copy the name of functions.`I think it will add a relatively high cost for performance, which will add more noise to the performance result.
It will be better if the user can provide a literal string and we keep the pointer to them.

PR #4276, opened 2024-06-05
Fix build warnings and treat warnings as error on CI

jkwak-work:


csyonghe:
One thing I would like to suggest here is to enable warnings as errors only on CI and not by default. This reduces the friction for users if some later compiler version is producing different set of warnings that are not captured by us.

jkwak-work:
> One thing I would like to suggest here is to enable warnings as errors only on CI and not by default. This reduces the friction for users if some later compiler version is producing different set of warnings that are not captured by us.

That's a good point. I will make sure that only CI will treat warnings as errors.

source/slang/CMakeLists.txt
```diff
@@ -271,9 +271,10 @@ slang_add_target(
     PUBLIC_HEADERS ${slang_SOURCE_DIR}/slang*.h
 )
 
-if(MSVC)
-    # Treat warnings as error
-    target_compile_options(slang PRIVATE /WX)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # a warning is disabled for a memset boundary check.
+    # everything looked find and it is unclear why the checking fails
```
csyonghe, line 276: find->fine

csyonghe (approved):

Issue #3832, opened 2024-03-25
Support warnings on uninitialized variables.


csyonghe:

sriramm-nv:
might be a breaking change for OV.

sriramm-nv:
AI Sriram: split this up into pieces that can be addressed in sprint 5.

venkataram-nv:
At this point, the bulk of this feature has been implemented. Remaining parts are more specialized, and are likely better off being smaller, specialized issues.

PR #4278, opened 2024-06-06
Add warning for uninitialized variables

sriramm-nv:
Fixes bug #3832

- Add test case for repro

### DIFF
--- a/tests/diagnostics/uninit-var.slang
+++ b/tests/diagnostics/uninit-var.slang
@@ -1,5 +1,5 @@
 // uninit-var.slang
-//TEST:SIMPLE(filecheck=CHECK): -target hlsl -entry main -profile cs_6_0
+//TEST:SIMPLE(filecheck=CHECK): -target hlsl -entry computeMain -profile cs_6_0
 
 //DISABLED_TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj
 //DISABLED_TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj
@@ -7,11 +7,20 @@
 
 struct Data
 {
-    float value;
+    float3 value;
 };
 
 RWStructuredBuffer<Data>  inputBuffer;
 RWStructuredBuffer<Data> outputBuffer;
+
+float foo(out float3 v)
+{
+    float uninitializedVariable;
+
+    // CHECK: warning
+    float r = v.x + uninitializedVariable;
+    return r;
+}
 
 [numthreads(1, 1, 1)]
 void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
@@ -19,10 +28,8 @@ void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
     uint tid = dispatchThreadID.x;
     Data inputData = inputBuffer[tid];
 
-    float uninitializedVariable;
 
-    // CHECK: warning
-    float outputData = inputData.value + uninitializedVariable;
+    float outputData = foo(inputData.value);
 
     outputBuffer[tid].value = outputData;
 }
--- a/tests/diagnostics/uninit-var.slang
+++ b/tests/diagnostics/uninit-var.slang
@@ -1,0 +1,28 @@
+// uninit-var.slang
+//TEST:SIMPLE(filecheck=CHECK): -target hlsl -entry main -profile cs_6_0
+
+//DISABLED_TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj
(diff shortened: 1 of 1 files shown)


tests/diagnostics/uninit-var.slang
```diff
@@ -0,0 +1,35 @@
...
+};
+
+RWStructuredBuffer<Data>  inputBuffer;
+RWStructuredBuffer<Data> outputBuffer;
+
+float foo(out float3 v)
+{
+    float uninitializedVariable;
+
+    // CHECK: warning
```
jkwak-work, line 20: This check will always pass and you need to make the check more specific to the warning you are expecting.
I think you can make it something like
```
// CHECK: warning 4101{{[56]}}
````

The reason why it will still pass when there was no warning about the uninitialized variable is that the output shader without any warning or error will have the following line.
```
#ifndef __DXC_VERSION_MAJOR
    #pragma warning(disable: 3557)
#endif
```

I also noticed that the warning you are getting is not about the intended variable, "uninitializedVariable" and it is about "v" not being set for the other components "y" and "z".
```
tests/temp2.slang(22): error 41015: use of uninitialized value 'v'
    float r = v.x + uninitializedVariable;
                ^
tests/temp2.slang(23): warning 41016: returning without initializing out parameter 'v'
    return r;
    ^~~~~~
}
```
If you remove the keyword "out" from "float foo(out float3 v)", the warning doesn't show up.
In other words, I don't think we are printing warnings for the uninitialized variables.

File: tests/diagnostics/uninit-var.slang [hunk 2]
```diff
@@ -0,0 +1,35 @@
+// uninit-var.slang
+//TEST:SIMPLE(filecheck=CHECK): -target hlsl -entry computeMain -profile cs_6_0
+
+//DISABLED_TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj
+//DISABLED_TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj
+//DISABLED_TEST(compute):COMPARE_COMPUTE_EX:-d3d12 -compute -shaderobj
```
ArielG-NV, line 6: any reason to include the disabled tests?

tests/diagnostics/uninit-var.slang
```diff
@@ -0,0 +1,35 @@
+// uninit-var.slang
+//TEST:SIMPLE(filecheck=CHECK): -target hlsl -entry computeMain -profile cs_6_0
```
ArielG-NV, line 2: I think there should also be a test for the opposite scenario: we do not throw a warning/error if the same scenario has an initialized variable (easier to debug and catch if something breaks)

Todo this, add something like the following:
```
//TEST:SIMPLE(filecheck=CHECK_INITIALIZED): -target hlsl -entry computeMain -profile cs_6_0 -DINITIALIZED
// ...
//CHECK_INITIALIZED-NOT: error 41015
#ifdef INITIALIZED
    float uninitializedVariable = 0;
#else 
    float uninitializedVariable;
#endif
```

tests/diagnostics/uninit-var.slang
```diff
@@ -0,0 +1,35 @@
+// uninit-var.slang
```
ArielG-NV, line 1: There likely should be isolated tests, each testing for a different uninitialized variable warning/error.

The 3 warning/errors are attached below:
```
DIAGNOSTIC(41015, Error, usingUninitializedValue, "use of uninitialized value '$0'")
DIAGNOSTIC(41016, Warning, returningWithUninitializedOut, "returning without initializing out parameter '$0'")
DIAGNOSTIC(41017, Warning, returningWithPartiallyUninitializedOut, "returning without fully initializing out parameter '$0'")
```

ArielG-NV (commented):
Some comments related to expanding the scope of tests.

tests/diagnostics/uninit-var.slang
sriramm-nv: thanks! I updated the test to add explicit checks. now I have a real issue to deal with.

tests/diagnostics/uninit-var.slang
sriramm-nv: noted. i added this case to the test.

tests/diagnostics/uninit-var.slang
sriramm-nv: Even simple tests produce more than one of these diagnostics, so I am not sure if we can isolate them into separate tests. May be I can isolate 41015 + 41016 into one, and 41015 + 40107 into one.
Also, should I also change 41015 into a Warning?

PR #4280, opened 2024-06-06
Metal: Mesh Shaders

Dynamitos:
Started working on the mesh shading stage, but this seems a lot more difficult than the task stage, because the IR is generated in a way where it references an output array and writes to it, but in Metal the output is set through a function call.


source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -329,8 +342,61 @@ namespace Slang
...
+        if(primitiveType == nullptr)
+        {
+            primitiveType = builder.getVoidType();
+        }
+        builder.setInsertBefore(entryPoint.entryPointFunc->getFirstBlock()->getFirstOrdinaryInst());
+        auto meshParam = builder.emitParam(builder.getMetalMeshType(vertexType, primitiveType, maxVertices, maxPrimitives, topology));
+        builder.addExternCppDecoration(meshParam, toSlice("_slang_mesh"));
+
+        
+        traverseUsers(verticesParam, [&](IRInst* inst){
```
csyonghe, line 384: Do we need something similar for indices and primitives?

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -329,8 +342,61 @@ namespace Slang
...
+            auto meshRef = as<IRMeshOutputRef>(inst);
+            traverseUsers(inst, [&](IRInst* user){
```
csyonghe, line 386: In general, we use traverseUses instead of traverseUsers because we can then test if the use is from the right operand we can handle. For example, the handling of store here assumes meshRef is the first operand of the store inst. If not, that isn't something we can handle here.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -319,6 +319,19 @@ namespace Slang
         }
 
         IRBuilder builder{ entryPoint.entryPointFunc->getModule() };
+        auto outputDeco = (IROutputTopologyDecoration*)func->findDecorationImpl(kIROp_OutputTopologyDecoration);
+        SLANG_ASSERT(outputDeco);
```
csyonghe, line 323: `func->findDecoration<IROutputTopologyDecoration>()`

source/slang/slang-emit-metal.cpp
```diff
@@ -653,6 +663,23 @@ void MetalSourceEmitter::emitSimpleTypeImpl(IRType* type)
...
+            emitType(meshType->getVertexType());
+            m_writer->emit(", ");
+            emitType(meshType->getPrimitiveType());
+            m_writer->emit(", ");
+            emitSimpleValue(meshType->getMaxNumVertices());
+            m_writer->emit(", ");
+            emitSimpleValue(meshType->getMaxNumPrimitives());
+            m_writer->emit(", metal::topology::");
+            // TODO: this is pretty hacky, there has to be a better way to do this, but I dont see why this wouldn't work
+            m_writer->emitRawTextSpan(meshType->getTopology()->getStringSlice().begin(), meshType->getTopology()->getStringSlice().end());
```
csyonghe, line 679: This can work, but from IR design point of view, it is better to have a enum value for the topology argument, and just use IRIntLit to represent topology. Then we can use an explicit switch statement here to emit the right text.

csyonghe (approved):
Looks good, again this is assuming all mesh outputs are done in the entry point function, but this is totally fine for now.

csyonghe:
Would be good to have a test to exercise the new logic.

source/slang/slang-ir-metal-legalize.cpp
Dynamitos: yes, this doesn't work at all yet, I just wanted to know if this kind of approach is even reasonable

source/slang/slang-ir-metal-legalize.cpp
Dynamitos: I still don't really understand the idea behind IRUse and IRUser, will have to look into that

source/slang/slang-emit-metal.cpp
Dynamitos: Will do that

Dynamitos:
This is just a basic idea for how the mesh output could be done, since the requirement to call a function instead of writing to an array is completely different. Primitives should work the same as vertices, indices could be easier since they have to be written to the array at once in HLSL, but haven't looked into it

csyonghe:
A IRUse is a struct {IRInst* user, IRInst* usedVal;}

An IRInst is a op code + a tail-allocated array of IRUse. Each element IRUse in the array will have user be the inst itself, and usedVal be the inst referenced by the operand.

IRUse also has the functionality to allow you to iterate through all the other uses of "usedVal".

So if you have an IRUse* use, you can actually find the argument index of this operand by:

```
user = use->getUser();
operandId = use - user->getOperands();
```

The design here is the same as LLVM's IR def-use chain, and there may be documentation around that to explain this as well.

source/slang/slang-emit-metal.cpp
Dynamitos: how do i create a IRIntLit? there doesnt seem to be a builder emit function, and creating something so basic seems wrong

csyonghe: You can use `IRBuilder::getIntValue(type, intVal)` to get/construct a integer literal IR inst from an int.

csyonghe:
@Dynamitos Do you still plan to work on this patch? If so please reopen it so we don't lose track of it.

Dynamitos:
I didnt get around to working on this for a while, and somehow messed up something with git, but now we should be back on track. I restarted from the master branch and integrated the changes again. I got a lot farther this time but this still needs to be done:

- Generate the MetalSetVertex, etc. IR conditionally in the core.meta.slang OR
- Replace the MeshOutputRef/MeshOutputSet with the metal versions in the legalize
- Remove the mesh array parameters after replacing the references with the mesh type
- Could probably use a bit of cleanup

Dynamitos:
Also I noticed the [[position]] semantic isnt showing up, is this related to mesh shading or will this be fixed independently?

csyonghe:
> Also I noticed the [[position]] semantic isnt showing up, is this related to mesh shading or will this be fixed independently?

Do you have a test case that reproduces this issue? We may need to fix it separately if it can be reproduced with a vertex / fragment shader.

Dynamitos:
It seems to work with the vertex shaders, but the simple-task.slang mesh stage test doesnt generate a position semantic, but user semantics are generated.

Also Im having trouble with a target_switch for the OutputVertices struct, i want to do something like this:

    [require(glsl_hlsl_metal_spirv, meshshading)]
    struct OutputVertices
    {
        __subscript(uint index) -> T
        {
            // TODO: Make sure this remains write only, we can't do this with just
            // a 'set' operation as it's legal to only write to part of the output
            // buffer, or part of the output buffer at a time.
            __target_switch
            {
            case metal:
            {
                [mutating]
                __intrinsic_op($(kIROp_MetalSetVertex))
                set;
            }
            default:
            {
                __intrinsic_op($(kIROp_MeshOutputRef))
                ref;
            }
            }
        }
    };

Since in metal the entire output needs to be written at once, this would directly generate the correct code, however it doesnt like the target_switch, do i have to escape it in some way?

csyonghe:
You cannot use __target_switch outside a function body.

It should be:
```
[require(glsl_hlsl_metal_spirv, meshshading)]
struct OutputVertices
{
    __subscript(uint index) -> T
    {
        // TODO: Make sure this remains write only, we can't do this with just
        // a 'set' operation as it's legal to only write to part of the output
        // buffer, or part of the output buffer at a time.

            [mutating]
            set
            {
                    // implement set.
             }

             [require(glsl_hlsll_spirv)]
             __intrinsic_op($(kIROp_MeshOutputRef))
            __ref { ... }
    }
};
```

csyonghe:
We should investigate why we need `kIROp_MeshOutputRef` at all. It seems like we can just use a getter and setter for both hlsl and spirv, and we make the getter require hlsl/spirv, while making the setter require hlsl/spirv/metal.

Dynamitos:
if im inside of the set function body, the __intrinsic_op doesnt work anymore, and __intrinsic_asm only evaluates the IROp enum value.

for the position semantic problem i found out that it works if i specify the vertex struct as a return type, but i dont know how to properly fix this.

Dynamitos:
the problem is that the mesh function always returns void, and the vertex type which should contain the [[position]] semantic is just a parameter which can normally not have the semantic, the easiest solution i can think of is to use check the vertex type explicitly in the mesh stage instead of the return type, but this probably wont work for primitive attributes

Dynamitos:
I haven fixed the position issue yet, and im not sure how to do that in a sane manner. Mesh outputs and payload should be correct now, however the other backends break when extracting the MeshOutputSet to a different function. I have tried to fix it, but I dont really understand the problem. But it definitely needs to be fixed before this can be merged. Alternatively the MeshOutputSets could be replaced in the legalize step, but this messes up the order of the code, the vertices get set before the values get initialized

csyonghe:
@Dynamitos I still don't understand the position problem you are running into, can you help me understanding the problem with a concrete shader example and use it to explain what gap we are having right now?

Dynamitos:
Here is an example:

    struct Vert
    {
        float4 pos : SV_Position;
    };

    struct Prim
    {
        uint prim : SV_PrimitiveID;
    }

    [outputtopology("triangle")]
    [numthreads(1, 1, 1)]
    void test(OutputVertices<Vert, 3> verts, OutputIndices<uint3, 1> ind, OutputPrimitives<Prim, 1> prims)
    {
        SetMeshOutputCounts(1, 1);
        verts[0] = { float4(0, 0, 0, 1) };
        verts[1] = { float4(1, 0, 0, 1) };
        verts[2] = { float4(0, 1, 0, 1) };
        ind[0] = uint3(0, 1, 2);
        prims[0] = { 0 };
    }


This snippet produces the following metal code:

    struct Vert_0
    {
        float4 pos_0;
    };
    struct Prim_0
    {
        uint prim_0;
    };

    [[mesh]] void test(metal::mesh<Vert_0, Prim_0, 3U, 1U, metal::topology::triangle> _slang_mesh)
    {
        _slang_mesh.set_primitive_count((1U));
        Vert_0 _S1 = { float4(0.0, 0.0, 0.0, 1.0) };
        _slang_mesh.set_vertex(0U,_S1);
        Vert_0 _S2 = { float4(1.0, 0.0, 0.0, 1.0) };
        _slang_mesh.set_vertex(1U,_S2);
        Vert_0 _S3 = { float4(0.0, 1.0, 0.0, 1.0) };
        _slang_mesh.set_vertex(2U,_S3);
        _slang_mesh.set_index(0U*3,(uint3(0U, 1U, 2U))[0]);
        _slang_mesh.set_index(0U*3,(uint3(0U, 1U, 2U))[1]);
        _slang_mesh.set_index(0U*3,(uint3(0U, 1U, 2U))[2]);
        ;
        Prim_0 _S4 = { 0U };
        _slang_mesh.set_primitive(0U,_S4);
        return;
    }

which is completely fine except for the fact that the [[position]] semantic is missing from the `Vert` struct and the [[primitive_id]] semantic is missing from the `Prim` struct.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
...
+                        );
+
+                    param->setFullType(annotatedPayloadType);
                 }
             }
+            IROutputTopologyDecoration* outputDeco = entryPoint.entryPointFunc->findDecoration<IROutputTopologyDecoration>();
+            const auto topology = outputDeco->getTopology();
+            const auto topStr = topology->getStringSlice();
+            UInt topologyEnum = 0;
+            if(topStr.caseInsensitiveEquals(toSlice("point"))){
```
csyonghe, line 1304: Style: brace on its own line.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
...
+                topologyEnum = 1;
+            } else if(topStr.caseInsensitiveEquals(toSlice("line"))) {
+                topologyEnum = 2;
+            } else if(topStr.caseInsensitiveEquals(toSlice("triangle"))) {
+                topologyEnum = 3;
+            } else {
+                // abort, idk how
```
csyonghe, line 1311: SLANG_UNEXPECTED("unknown topology");

Or you can just diagnose an error message with sink->diagnose(...)

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
...
+            builder.addExternCppDecoration(meshParam, toSlice("_slang_mesh"));
+
+
+            verticesParam->removeFromParent();
+            verticesParam->removeAndDeallocate();
+            
+            indicesParam->removeFromParent();
+            indicesParam->removeAndDeallocate();
+
+            if(primitivesParam != nullptr) {
```
csyonghe, line 1408: Brace.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
                     varLayoutBuilder.findOrAddResourceInfo(LayoutResourceKind::MetalPayload);
                     auto paramVarLayout = varLayoutBuilder.build();
                     builder.addLayoutDecoration(param, paramVarLayout);
+
+                    IRConstRefType* type = as<IRConstRefType>(param->getDataType());
```
csyonghe, line 1288: This can just be a PtrType.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
...
+                            kIROp_ConstRefType,
+                            type->getValueType(),
+                            AddressSpace::MetalObjectData
+                        );
+
+                    param->setFullType(annotatedPayloadType);
                 }
             }
+            IROutputTopologyDecoration* outputDeco = entryPoint.entryPointFunc->findDecoration<IROutputTopologyDecoration>();
+            const auto topology = outputDeco->getTopology();
```
csyonghe, line 1301: What if outputDeco doesn't exist?

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1284,9 +1284,131 @@ namespace Slang
                     varLayoutBuilder.findOrAddResourceInfo(LayoutResourceKind::MetalPayload);
                     auto paramVarLayout = varLayoutBuilder.build();
                     builder.addLayoutDecoration(param, paramVarLayout);
+
```
csyonghe, line 1287: It is cleaner to put all these logic into a separate function called legalizeMeshEntryPoint, and called from legalizeEntryPoint based on the stage in a switch statement so we can add other stage specific legalizations easily in the future.

Dynamitos:
I have noticed two more things with the metal backend:

- SV_PrimitiveID is converted to [[patch_id]] when used as fragment shader input
- texture writes need to be converted to 4-component vectors regardless of the actual texture components.

The first one i fixed by replacing the metalSystemValueType, but it is odd enough to think that maybe there was a reason for it resoving to [[patch_id]]?
For the second one it should probably be done in the legalization by inserting a MakeVector operation into the argument of the write.

Should this be handled in a separate pull request or merged with this one?

csyonghe:
What should SV_PrimitiveID be translated into? It is likely a mistake in the original implementation. In that case a separate PR for that fix will be better, since we can get that merged in very quickly.

Dynamitos:
there is a separate [[primitive_id]] semantic, [[patch_id]] is only valid after a tesselation stage

csyonghe:
Yeah, that is a mistake and we should fix it in a separate PR.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1285,16 +1341,85 @@ namespace Slang
...
+                    SLANG_ASSERT(primitiveType);
+
+                    primitivesParam = param;
+                    auto primStruct = as<IRStructType>(primitiveType);
+                    for(auto field : primStruct->getFields())
+                    {
+                        auto key = field->getKey();
+                        if(auto deco = key->findDecoration<IRSemanticDecoration>())
+                        {
+                            if(deco->getSemanticName().caseInsensitiveEquals(toSlice("sv_primitiveid")))
```
csyonghe, line 1390: Is this logic here to workaround the bug?

Dynamitos: this is for annotating mesh and primitive attributes at all. Since they are not part of the return type, they dont get processed in `wrapReturnValueInStruct`. The other issue is when taking the a SV_PrimitiveID as an input parameter in the fragment shader.

csyonghe: We probably also want to include a check to make sure PrimitiveType is a struct, but this can be done in a common verification step around slang-lower-to-ir.cpp::generateIRForTranslationUnit (line 11192) and doesn't need to be part of this PR.

csyonghe (approved):
Looks good to me. Thank you for this significant contribution!

PR #4282, opened 2024-06-06
Fix python syntax highlighting in slangpy chapter

eliemichel:
I guess it did not like the space between backtits and "Python"


jkwak-work (approved):
It looks like this is the right way to use the syntax.

There is another python code on the same page that shows up with proper highlights.
And it uses this syntax.

PR #4283, opened 2024-06-06
Remove unnecessary call to __requireComputeDerivative

jkwak-work:
When SPIR-V uses operators whose name has a keyword, "Implicit", they require calling a function "__requireComputeDerivative()".

When it uses "Explicit", the function doesn't need to be called.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -2592,7 +2592,6 @@ public vector<T,4> textureLod(__TextureImpl<
 [require(glsl_hlsl_spirv, texture_shadowlod)]
 public float textureLod(sampler2DShadow sampler, vec3 p, float lod)
 {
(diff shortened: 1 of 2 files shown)


csyonghe (dismissed):
Is our HLSL intrinsic definitions correct? SampleLevel for example shouldn't need this either, right?

ArielG-NV:
> Is our HLSL intrinsic definitions correct? SampleLevel for example shouldn't need this either, right?

SampleLevel (including other explicit LOD functions) and gradient functions do not need `__requireComputeDerivative`.

jkwak-work:
That's a good catch.
I didn't think about hlsl.meta.slang.
I spotted a few functions that uses `Implicit` but not calling `__requireComputeDerivative()`.
I am pushing the new change for a review.

PR #4285, opened 2024-06-06
make String::trim consistent with String::trimStart/trimEnd

skallweitNV:
Make `String::trim` remove carriage return and newline characters equal to what `String::trimStart` and `String::trimEnd` do.


jkwak-work (approved):
It makes sense.
And it matches to the behavior of `trimEnd()` too.

PR #4286, opened 2024-06-06
Limit the supported texture types for GLSL

jkwak-work:
GLSL supports only four types for textures: vec4, ivec4, uvec4, and float. The generic parameterization for the vector element count is not needed for the GLSL texture functions.

In other words, `N` is removed from `__generic` and it is replaced with `4`.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -1682,10 +1682,10 @@ public typealias usamplerBuffer = SamplerBuffer<uint4>;
 // textureSize
 // -------------------
 
-__generic<T:__BuiltinArithmeticType, let N:int>
+__generic<T:__BuiltinArithmeticType>
 [ForceInline]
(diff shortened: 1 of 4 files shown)


csyonghe:
I understand the change is to make the definitions more conformant to glsl, but does it have any downside to allow more texture types?

ArielG-NV (approved):
The changes look good.
Before merge, please checkout [the following question](https://github.com/shader-slang/slang/pull/4286#issuecomment-2152630526)

ArielG-NV:
I tested the question.

It seems this change is not (?) required if we prefer functionality over conformance to GLSL

jkwak-work:
Thanks for the comments and also for the testing.
Honestly I didn't think more than thinking about the GLSL conformance.
And it appears that my understanding was incorrect based on my testing that is similar to what Ariel did.
I am abandoning this change.

Issue #4291, opened 2024-06-06
Metal: Failing compute tests


skallweitNV:
The following tests in the `compute` category currently fail to compile with Metal.

All failing tests are disabled using:
```
//DISABLE_TEST(compute):COMPARE_COMPUTE:-slang -shaderobj -mtl
```

This should make it easy to find them.

Some of these are failing to compile due to #4288 

For reference and marking fixed tests, here is the full list:

- [x] `tests/autodiff-dstdlib/determinant.slang`
- [x] `tests/autodiff-dstdlib/dstdlib-mul-mat-mat.slang`
- [x] `tests/autodiff-dstdlib/dstdlib-mul-mat-vec.slang`
- [x] `tests/autodiff-dstdlib/dstdlib-mul-vec-mat.slang`
- [x] `tests/autodiff-dstdlib/dstdlib-sqrt.slang`
- [x] `tests/autodiff/custom-intrinsic.slang`
- [ ] `tests/autodiff/global-param-hoisting.slang`
- [x] `tests/autodiff/reverse-matrix-ops.slang`
- [x] `tests/bindings/nested-parameter-block-2.slang`
- [x] `tests/bugs/atomic-coerce.slang`
- [ ] `tests/bugs/buffer-swizzle-store.slang`
- [ ] `tests/bugs/dxbc-double-problem.slang`
- [ ] `tests/bugs/frexp-double.slang`
- [x] `tests/bugs/frexp.slang`
- [x] `tests/bugs/generic-default-matrix.slang`
- [x] `tests/bugs/generic-groupshared.slang`
- [x] `tests/bugs/generic-member-method.slang`
- [x] `tests/bugs/generic-type-duplication.slang`
- [x] `tests/bugs/gh-2959.slang`
- [x] `tests/bugs/gh-3980.slang`
- [x] `tests/bugs/interface-type-self-ref.slang`
- [x] `tests/bugs/mutating/buffer-write-dce.slang`
- [x] `tests/bugs/mutating/mutating-call-in-loop-dce.slang`
- [x] `tests/bugs/mutating/mutating-generic-method.slang`
- [x] `tests/bugs/mutating/resource-specialization-inout.slang`
- [x] `tests/bugs/negative-literal.slang`
- [x] `tests/bugs/nested-existential-dyndispatch.slang`
- [x] `tests/bugs/nested-switch.slang`
- [ ] `tests/bugs/op-assignment-unify-mat.slang`
[...]

PR #4292, opened 2024-06-06
Metal compute tests

skallweitNV:
This PR adds support for running compute tests with the new metal backend.

The current state is:
- 280 tests pass
- 18 tests fail (generate unexpected result)
- 155 tests fail to compile (in the metal downstream compiler)

All tests that fail to compile are disabled using
```
//DISABLE_TEST(compute):COMPARE_COMPUTE:-slang -shaderobj -mtl
```

See #4291 for more details.


skallweitNV:
I don't quite understand why this change leads to the following failing tests:

- tests/compute/matrix-layout-structured-buffer.slang.3 syn (dx12)
- tests/compute/matrix-layout-structured-buffer.slang.4 syn (dx12)
- tests/compute/buffer-layout.slang.5 syn (cuda)

Also there is one `COMPARE_HLSL_RENDER` test that fails on metal:

- tests/bugs/texture2d-gather.hlsl

@csyonghe ?

csyonghe:
The texture2d test may be using a vertex-fragment shader and we seem to be generating invalid vertex shader code for metal, causing the test to fail.

csyonghe:
It sure about the matrix layout tests. None of the changes here seemed to be relevant.

skallweitNV:
I "fixed" the tests by specifying the actual APIs they should run on instead of disabling for metal.

csyonghe (approved):

csyonghe:
Looks good! We can clean up the test discovery logic later to make this more elegant, but we don't have to do that now.

PR #4294, opened 2024-06-06
Partial implementation of static_assert

jkwak-work:
Partially implements `static_assert(C,M)` where C is a compile-time constant value and M is an error message gets printed when C is evaluated to false.

Currently it can be used like a function call and it cannot be used as a part of struct declaration or in a global scope.


source/slang/slang-emit-c-like.cpp
```diff
@@ -2774,6 +2774,26 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
     {
         break; //should already have been parsed and used.
     }
+    case kIROp_StaticAssert:
```
csyonghe, line 2777: Adding a check here is likely too late. C-like emit also does not cover spirv backend.

We need a separate pass to do the static assert checking.

The problem with this kind of post link checking is that it introduces situations where the code checked fine in the front end, but fails after linking. This usually leads to user frustrations and the general design philosophy is to avoid them when possible and try our best to capture errors as early as possible.

For this particular case the only way to allow front end check is to duplicate all the sampling functions, but that creates a lot of bloat in the stdlib, so doing the post link check may be the right trade-off.

jkwak-work: I appreciate this feedback.
Can I ask more help on where the checking should happen?
I am thinking maybe I can do it in `generateIRForTranslationUnit()` where the missing-return check and operator<< check happens.

csyonghe: This likely need to happen late because we have to inline and specialize our generics before we can resolve those constexpr values. We can just do it right before going into emit at the end of linkAndOptimizeIR.

csyonghe:
After validation, we need to remove the static assert instructions to allow dead code elim pass to clean up the code so we don't generate code for things that are only used in static assert.

source/slang/hlsl.meta.slang
```diff
@@ -1654,15 +1654,44 @@ extension __TextureImpl<T,Shape,isArray,isMS,sampleCount,0,isShadow,0,format>
...
     [ForceInline]
     [require(cpp_cuda_glsl_hlsl_metal_spirv, texture_sm_4_1)]
     T SampleLevel(SamplerState s, vector<float, Shape.dimensions+isArray> location, float level)
     {
+        __checkSupportedType();
         __target_switch
         {
             case cpp:
             case hlsl:
+                static_assert(T is IFloat, "HLSL doesn't support texture sampling for integer types.");
```
jkwak-work, line 1694: `static_assert` seems to be working as expected if it is used directly inside of `__target_switch`.
That being said, the following line works as expected,
```
static_assert(T is IFloat, "HLSL doesn't support texture sampling for integer types.");
```
But if I call it indirectly inside of another function it doesn't work.
In this case the function call to `__checkSupportedType()` itself doesn't exist in the IR-dump.
I think there is a bug when `__intrinsic_asm` is used.

Here is IR dump for more detail,
```
generic %150    : %144
{
block %151(
                [TypeConstraintDecoration(%IFloat)]
                [nameHint("T")]
                param %T1       : type_t,
                [TypeConstraintDecoration(%x5Fx5FITextureShape)]
                [nameHint("Shape")]
                param %Shape1   : type_t,
                [nameHint("isArray")]
                param %isArray1 : Int,
                [nameHint("isMS")]
                param %isMS1    : Int,
                [nameHint("sampleCount")]
                param %sampleCount1     : Int,
                [nameHint("isShadow")]
                param %isShadow1        : Int,
                [nameHint("format")]
                param %format1  : Int,
                param %152      : witness_table_t(%IFloat),
                param %153      : witness_table_t(%x5Fx5FITextureShape)):
        let  %154       : Int   = lookupWitness(%153, %58)
        let  %155       : Int   = lookupWitness(%153, %139)
        let  %156       : Int   = add(%isArray1, %154)
        [ForceInline]
        [readNone]
        [nameHint("SampleLevel")]
                func %SampleLevel       : Func(%T1, TextureType(%T1, %Shape1, %isArray1, %isMS1, %sampleCount1, 0 : Int, %isShadow1, 0 : Int, %format1, %153), SamplerState, Vec(Float, %156), Float)
        {
block %157(
                        [nameHint("this")]
                        param %this     : TextureType(%T1, %Shape1, %isArray1, %isMS1, %sampleCount1, 0 : Int, %isShadow1, 0 : Int, %format1, %153),
                        [nameHint("s")]
                        param %s        : SamplerState,
                        [nameHint("location")]
                        param %location : Vec(Float, %156),
                        [nameHint("level")]
                        param %level    : Float):
                unconditionalBranch(%158)

block %158:
                StaticAssert(true, "HLSL doesn\'t support texture sampling for integer types.")
                GenericAsm(".SampleLevel")
}
        return_val(%SampleLevel)
}
```
I expect that the function call to `__checkSupportedType()` should appear above the line, `unconditionalBranch(%158)`.
But it is not there.

Ariel and I have seen a problem like this before.
I will debug and see what is happening.

I am guessing that it could be by design that when we call `__intrinsic_asm`, we may take a short-cut to skip functions calls.

csyonghe:
This is a good place to do the check. As long as all static_assert inst are removed we are fine. No need for other verifications.

csyonghe:
If the inst is cleaned up after autodiff, we also need to make sure the autodiff pass handles the inst.

csyonghe:
A case should be added to slang-ir-autodiff-fwd.cpp:1944 for the opcode to be treated as non differentiable.

jkwak-work:
> A case should be added to slang-ir-autodiff-fwd.cpp:1944 for the opcode to be treated as non differentiable.

Thanks for the information. I haven't had a chance to look at autodiff side code. I will check out this code and make sure it works well with autodiff.

csyonghe:
Regarding the static_assert inst inside a function with a single `__intrinsic_asm` inst issue:

The reason here is that if a function has an __intrinsic_asm, it will not be inlined, so the condition expr in the static_assert expr will never be substituted and folded into a constant. The key here is to allow the compiler to inline a function with `__intrinsic_asm`. To do that, we need the following changes:

1. Extend `GenericAsm` instruction to allow it to take additional arguments, so that `GenericAsm("$0 $1", arg0, arg1)` means that `$0` is going to be emitted from `arg0`, `$1` is going to be emitted from `arg1`. 
2. Lower __intrinsic_asm "str"; into `GenericAsm("str", param0, param1, ...)` where param0, param1 are the parameters of the function.
3. Make `GenericAsm` not a terminator inst, and just a normal inst.

With these changes, we can inline a function with GenericAsm just like normal functions, and the operands to `GenericAsm` will be substituted from the callee arguments during inlining. Once a function with `GenericAsm` and `StaticAssert` insts are inlined, the checks for StaticAssert will get the folded value and will work as expected.

jkwak-work:
It appears that `StaticAssert` IR is not eliminated yet and I am not sure if it can be eliminated.

The reason why it is marked as alive is because `shouldInstBeLiveIfParentIsLive()` and `mightHaveSideEffects()` returns true for it.
When I changed `mightHaveSideEffects()` to return false, it gets eliminated, but then static_assert no longer works.

The problem is that when we compile stdlib, kIROp_StaticAssert shouldn't be removed and when we compile the user code, the static assertion check should happen.

I am not sure at what point we can remove kIROp_StaticAssert if it can be removed.

jkwak-work:
> Regarding the static_assert inst inside a function with a single `__intrinsic_asm` inst issue:
> 
> The reason here is that if a function has an __intrinsic_asm, it will not be inlined, so the condition expr in the static_assert expr will never be substituted and folded into a constant. The key here is to allow the compiler to inline a function with `__intrinsic_asm`. To do that, we need the following changes:
> 
> 1. Extend `GenericAsm` instruction to allow it to take additional arguments, so that `GenericAsm("$0 $1", arg0, arg1)` means that `$0` is going to be emitted from `arg0`, `$1` is going to be emitted from `arg1`.
> 2. Lower __intrinsic_asm "str"; into `GenericAsm("str", param0, param1, ...)` where param0, param1 are the parameters of the function.
> 3. Make `GenericAsm` not a terminator inst, and just a normal inst.
> 
> With these changes, we can inline a function with GenericAsm just like normal functions, and the operands to `GenericAsm` will be substituted from the callee arguments during inlining. Once a function with `GenericAsm` and `StaticAssert` insts are inlined, the checks for StaticAssert will get the folded value and will work as expected.

Thanks for this information.
It sounds like a limitation more than a bug; or it can be seen as a known issue.

It doesn't seem like a trivial bug fix.
And I will go with a workaround for the GLSL integer texture support task this time.

Do you think we need a new issue for this limitation? If yes, I can file one with a repro code.

source/slang/slang-emit.cpp
```diff
@@ -362,6 +362,42 @@ void calcRequiredLoweringPassSet(RequiredLoweringPassSet& result, CodeGenContext
...
+                {
+                    sink->diagnose(inst, Diagnostics::staticAssertionFailureWithoutMessage);
+                }
+            }
+        }
+        else
+        {
+            sink->diagnose(condi, Diagnostics::staticAssertionConditionNotConstant);
+        }
+        break;
```
csyonghe, line 391: You can just `inst->removeAndDeallocate()` here to remove the static assert inst since validation is already done at this point and the StaticAssert inst has no value past this point.

jkwak-work: This worked!

jkwak-work: Ok. I turned out that it causes a problem if I remove as suggested.
When I remove StaticAssert after it is applied once, it doesn't work for the next specialization.

csyonghe: The idea is to apply the check only once when we know we have done everything we can to resolve  the values, so we shouldn't be running the checking pass more than once.

jkwak-work: I am gonna need to go over one of the brownbag records and understand better how the generics are instantiated. But my understanding here is that the static assertion has to  be alive for the next specialization if the StaticAssert was in a generic function.

For the cases where the functions are not generic, I guess I can removed them.
 And I can probably remove StaticAssert from specialized functions; not from the generic functions.
I will try more on this.

csyonghe:
The place you are invoking the pass is after the loop that calls specializeModule, so at this time we have specialized all generics that we can specialize, and it should be fine to remove all static assert insts. What problem are you seeing?

jkwak-work:
> The place you are invoking the pass is after the loop that calls specializeModule, so at this time we have specialized all generics that we can specialize, and it should be fine to remove all static assert insts. What problem are you seeing?

Your explanation makes sense. The problem I saw is that the test I added in this PR fails in a way that some of the static assertion failed as expected but some of them were silent as if they were removed.

I will need to debug and investigate it more.

jkwak-work:
It turned out that I shouldn't have removed the instance while iterating the children.

csyonghe (approved):

PR #4297, opened 2024-06-07
Fixed profile string per request in pr#4268

chaoticbob:
Changed profile string from `glsl_440` to `spirv_1_5` per request in pr#4268.


jkwak-work (approved):
Thanks for the PR

PR #4306, opened 2024-06-07
Adding functional test for GLSL texture functions

jkwak-work:
This commit adds a functional test to GLSL texture functions. It appears that the testing frameworkd doesn't work for combined sampler for "shadow" and "buffer".


jkwak-work:
I am going to submit this without waiting for a review.
This change is tedious and too long to review.

PR #4308, opened 2024-06-08
Metal system value overhaul.

### LINKED ISSUE #4288: Metal: Kernel function input argument types
Most metal kernel function input arguments need to have `uint` or `ushort` types, but slang currently doesn't transform signed to unsigned types. For example `int3 tid: SV_DispatchThreadID` is emitted as `int3 tid_0 [[thread_position_in_grid]]` which leads to a error in the metal compiler.

The supported types are documented in table 5.8 in https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf

csyonghe:
Add legalization logic to translate HLSL SV semantic to metal system values attributes.

If the declared HLSL parameter has type that is different from the Metal spec-defined type, we will insert conversion logic at the beginning of the entry point.

Also fixes the handling of stage-in parameters with user defined semantics. We shouldn't flatten the members of a struct-typed stage in parameter if the parameter has explicit user semantic.

Closes #4288.

### DIFF
--- a/source/slang/slang-emit-metal.cpp
+++ b/source/slang/slang-emit-metal.cpp
@@ -170,8 +170,11 @@ void MetalSourceEmitter::emitFuncParamLayoutImpl(IRInst* param)
             break;
         }
     }
-    if (auto sysSemanticAttr = layout->findSystemValueSemanticAttr())
-       _emitSystemSemantic(sysSemanticAttr->getName(), sysSemanticAttr->getIndex());
+    if (!maybeEmitSystemSemantic(param))
+    {
+        if (auto sysSemanticAttr = layout->findSystemValueSemanticAttr())
+            _emitUserSemantic(sysSemanticAttr->getName(), sysSemanticAttr->getIndex());
+    }
 }
 
 void MetalSourceEmitter::emitParameterGroupImpl(IRGlobalParam* varDecl, IRUniformParameterGroupType* type)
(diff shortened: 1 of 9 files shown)


source/slang/slang-emit-metal.cpp
```diff
@@ -741,84 +744,24 @@ void MetalSourceEmitter::_emitType(IRType* type, DeclaratorInfo* declarator)
...
+    return false;
+}
+
+void MetalSourceEmitter::_emitUserSemantic(UnownedStringSlice semanticName, IRIntegerValue semanticIndex)
+{
+    if (!semanticName.startsWithCaseInsensitive(toSlice("SV_")))
     {
         m_writer->emit(" [[user(");
-        m_writer->emit(semanticName);
+        m_writer->emit(String(semanticName).toUpper());
```
jkwak-work, line 764: Is there a reason why to use upper cases?
Can we just use it as it is?

jkwak-work (approved):
Looks good to me

source/slang/slang-emit-metal.cpp
csyonghe: HLSL doesn't distinguish upper/lower casing for semantic names, but metal does. So the only way to ensure we generate valid code is to canonicalize the name.

jkwak-work: Oh. I see.
In that case, should we add "_0" kind of suffix to avoid conflicts?
Both "sem" and "Sem" will become "SEM", as an example.

jkwak-work: I take this back. I misunderstood what you meant.
Since Slang is case insensitive, it should be fine to make them all upper case.

PR #4309, opened 2024-06-08
Small SPIRV emit cleanup around vector element extract.

csyonghe:
If we have `GetElement(vector, index)` where `index` is a `GlobalValueRef(IntLit)`, we still want to generate OpCompositeExtract instead of `OpVectorExtractDynamic`.


jkwak-work (approved):
Looks good to me.

PR #4311, opened 2024-06-08
Support all integer typed indices in StructuredBuffer Load/Store/[].

### LINKED ISSUE #4274: Add support for RWStructuredBuffer with 64bit index
For the CUDA and CPP targets, the prelude-part is ok (uses size_t), but it is limited by slang having int/uint index only, see:
https://github.com/shader-slang/slang/blob/master/source/slang/hlsl.meta.slang#L4908
https://github.com/shader-slang/slang/blob/master/source/slang/hlsl.meta.slang#L4914

csyonghe:
Closes #4274.


jkwak-work (dismissed):
Looks good to me.

I was thinking of adding more overloading but this looks better.

jkwak-work:
It looks like some tests are failing.
Let me retrigger and see if it goes away.

jkwak-work:
It seems like you need to resolve a merge conflict.
It conflicts with Spirv Block fix.

jkwak-work (approved):

PR #4312, opened 2024-06-08
Disallow certain types of decls in `interface` to provide better diagnostic message.

### LINKED ISSUE #4266: typealias with interface does not work as expected
Static const vars can easily be defined in interfaces and then set in the implementation
```c#
interface IFoo {
    static const bool isBar;
};

struct Bar {
    static const bool isBar = true;
};
```

for typealias this does not work and the error messages are not clear to me
e.g., this for instance displays an error in struct Bar:
`member 'Scalar' does not match interface requirement.(38105)`
`see interface requirement declaration of 'Scalar'`
```csharp
interface IFoo {
    typealias Scalar = void;
};

struct Bar : IFoo {
    typealias Scalar = float;
};
```
if I don't set anything in Bar I get
`type 'Bar' does not provide required interface member 'Scalar'(38100)`
`see declaration of 'Scalar'`

Even setting Scalar to void in Bar does not work?

Slang version 2024.1.20

csyonghe:
Closes #4266.

### DIFF
--- a/source/slang/slang-parser.cpp
+++ b/source/slang/slang-parser.cpp
@@ -4213,13 +4213,8 @@ namespace Slang
             case ASTNodeType::SubscriptDecl:
             case ASTNodeType::PropertyDecl:
             case ASTNodeType::AssocTypeDecl:
-            case ASTNodeType::TypeAliasDecl:
-            case ASTNodeType::TypeDefDecl:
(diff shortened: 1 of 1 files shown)


jkwak-work (approved):
Looks good to me.

PR #4318, opened 2024-06-10
Fix SPIRV emit for `Flat` decoration and TessLevel builtin.

### LINKED ISSUE #4313: error 49999: unknown system-value semantic 'SV_InsideTessFactor'
Are tessellation shaders supported yet? 
Is there an example how to map both endpoints to the hull stage?

Also, VulkanUtil::getShaderStage: domain stage is mapped to VK control, hull to evaluation 
https://github.com/shader-slang/slang/blob/master/tools/gfx/vulkan/vk-util.cpp#L183

A simple sample I get the error with (v2024.1.21, SPIRV target):
```hlsl
struct HS_CONTROL_POINT_OUTPUT
{
    float4 Position : POSITION;
    float2 TexCoord : TEXCOORD;
};

struct HS_CONSTANT_DATA_OUTPUT
{
    float EdgeTessFactor[4] : SV_TessFactor;
    float InsideTessFactor[2] : SV_InsideTessFactor;
};


struct DS_OUTPUT
{
    float4 Position : SV_Position;
    float2 TexCoord : TEXCOORD;
};

// Domain Shader (DS)
[domain("quad")]
DS_OUTPUT main(HS_CONSTANT_DATA_OUTPUT input, const OutputPatch<HS_CONTROL_POINT_OUTPUT, 4> patch, float2 uv : SV_DomainLocation)
{
    DS_OUTPUT output;
    output.Position = 0;
    output.TexCoord = 0;
    return output;
}
```

### LINKED ISSUE #4315: SPIR-V Validation error - Fragment stage integer Input decorated Flat multiple times
I have run into an issue with slang spirv output failing spirv validation checks in newly updated source builds of slang and the vulkan validation layers:

slang commit: <https://github.com/shader-slang/slang/commit/9a23a9aab3721828526c921db1e779008e133e8f>
validation layers commit: <https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/f34c751cbffb7287101e8d8682305a9b805892e3>

The following shader
```c++
struct VertexOutput
{
    nointerpolation int a : SOME_VALUE;
    float4              b : SV_Position;
};

[shader("vertex")]
VertexOutput Vertex()
{
    VertexOutput out;
    out.a = 0;
    out.b = float4(0, 0, 0, 1);
    return out;
}

[shader("fragment")]
float4 Fragment(in VertexOutput vo)
{
    return float4(float(vo.a), 0, 0, 1);
}
```
produces the following validation error
```
Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
ID '36' decorated with Flat multiple times is not allowed.
  %vo_a = OpVariable %_ptr_Input_int Input
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```
Note: This only occurs when `a` is an integer (I tried with `int` and `uint`). No double annotation or spirv error occurs when `a` is a `float`.

Inspecting spirv-assembly generated using the same compile options led to the following output 
```
slangc 
  -lang slang 
  -matrix-layout-column-major 
  -force-glsl-scalar-layout 
  -target spirv-assembly 
  -fvk-use-entrypoint-name 
  -emit-spirv-directly
```
```
; SPIR-V
; Version: 1.5
; Generator: Khronos; 40
; Bound: 40
; Schema: 0
OpCapability Shader
OpMemoryModel Logical GLSL450
OpEntryPoint Vertex %Vertex "Vertex" %entryPointParam_Vertex_a %gl_Position
OpEntryPoint Fragment %Fragment "Fragment" %entryPointParam_Fragment %vo_a
OpExecutionMode %Fragment OriginUpperLeft

; Debug Information
OpSource Slang 1
OpName %entryPointParam_Vertex_a "entryPointParam_Vertex.a"  ; id %25
OpName %Vertex "Vertex"  ; id %2
OpName %vo_a "vo.a"  ; id %36
OpName %entryPointParam_Fragment "entryPointParam_Fragment"  ; id %39
OpName %Fragment "Fragment"  ; id %32

; Annotations
OpDecorate %entryPointParam_Vertex_a Location 0
OpDecorate %entryPointParam_Vertex_a Flat
OpDecorate %gl_Position BuiltIn Position
OpDecorate %vo_a Location 0
OpDecorate %vo_a Flat
OpDecorate %entryPointParam_Fragment Location 0
OpDecorate %vo_a Flat

; Types, variables and constants
%void = OpTypeVoid
%3 = OpTypeFunction %void
%int = OpTypeInt 32 1
%float = OpTypeFloat 32
%v4float = OpTypeVector %float 4
%int_0 = OpConstant %int 0
%float_0 = OpConstant %float 0
%float_1 = OpConstant %float 1
%18 = OpConstantComposite %v4float %float_0 %float_0 %float_0 %float_1
%_ptr_Output_int = OpTypePointer Output %int
%_ptr_Output_v4float = OpTypePointer Output %v4float
%_ptr_Input_int = OpTypePointer Input %int
%entryPointParam_Vertex_a = OpVariable %_ptr_Output_int Output
%gl_Position = OpVariable %_ptr_Output_v4float Output
%vo_a = OpVariable %_ptr_Input_int Input
%entryPointParam_Fragment = OpVariable %_ptr_Output_v4float Output

; Function Vertex
%Vertex = OpFunction %void None %3
%4 = OpLabel
OpStore %entryPointParam_Vertex_a %int_0
OpStore %gl_Position %18
OpReturn
OpFunctionEnd

; Function Fragment
%Fragment = OpFunction %void None %3
%33 = OpLabel
%34 = OpLoad %int %vo_a
%37 = OpConvertSToF %float %34
%38 = OpCompositeConstruct %v4float %37 %float_0 %float_0 %float_1
OpStore %entryPointParam_Fragment %38
OpReturn
OpFunctionEnd
```
where `Flat` is being applied twice
```
OpDecorate %vo_a Location 0
OpDecorate %vo_a Flat
OpDecorate %entryPointParam_Fragment Location 0
OpDecorate %vo_a Flat
```
The same double annotations occurs with the more minimal set of compile options building just the `Fragment` entry point
```
slangc 
  -lang slang 
  -target spirv-assembly
  -emit-spirv-directly
  -entry Fragment
```
Note: This only occurs when code is generated for the Fragment entry point. Generating Vertex results in only a single `Flat` decoration.
Note: After commit <https://github.com/shader-slang/slang/commit/f1de1817ca10e34ec6a844100f10f0de3340c9f2> this repro works without `-emit-spirv-directly` as it is now selected by default.

The first validation layers commit that started catching this was <https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/a9b3f651c1b8a6515ec876f25a570f6572dc2516> which updated the SPIRV-Tools dependency to commit <https://github.com/KhronosGroup/SPIRV-Tools/commit/02470f606fe1571de808cb773d8c521ab201aaff> in which `* Disallow duplicate decorations generally` would be relevant here.

A recent commit (latest as of this issue being made) <https://github.com/shader-slang/slang/commit/9a23a9aab3721828526c921db1e779008e133e8f> seems to address *some* duplicate decorations, but not `OpDecorate ... Flat`

From bisecting, I believe the first commit that introduced this double decoration was <https://github.com/shader-slang/slang/commit/6f8a20688e0c4b989db152b4d06aeab04fac0567> (which was made before the validation check was added to `spirv-val`)

Happy to provide any more context if necessary.

csyonghe:
Clean up the logic to emit `Flat` decoration for builtin and non-builtin shader inputs so that we don't emit duplicate `Flat` decorations.

Fix the legalization logic of array-typed system values to allow emitting correct code for `SV_TessFactor` and `SV_InsideTessFactor` for spirv.

Closes #4313.
Closes #4315.


tests/spirv/flat.slang
```diff
@@ -0,0 +1,27 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv
+
+// Check that we don't emit duplicate Flat decorations for the same variable.
+
+// CHECK: OpDecorate %vo_a Flat
+// CHECK-NOT: OpDecorate %vo_a Flat
+
```
jkwak-work, line 7: This issue can be reproduced with an existing test writting in GLSL. `tests/glsl/flat-in.slang`
Maybe we can have the same check there too.

source/slang/slang-emit-spirv.cpp
```diff
@@ -6330,28 +6356,6 @@ SlangResult emitSPIRVFromIR(
         context.ensureInst(irEntryPoint);
     }
 
-    // Declare integral input builtins as Flat if necessary.
-    for (auto globalInst : context.m_irModule->getGlobalInsts())
```
jkwak-work, line 6334: I am trying to understand why the issue was happening with the previous code.
It looks like it shouldn't have emitted the duplicated decoration.
Did this for-loop end up iterating a same instance more than once?
Can you share some information of what the problem was?

tests/spirv/tess-factor.slang
```diff
@@ -0,0 +1,32 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv -stage domain
```
ArielG-NV, line 1: should we add a glsl test for `gl_TessLevelInner`?

source/slang/slang-emit-spirv.cpp
csyonghe: The duplication is introduced because the code here didn't check if `globalInst` already has a `InterpolationModeDecoration`. The only reason to have the loop here to make sure the spirv builtin variables we added during spirv-emit has proper `flat` decorations, the logic is not required for all other user defined global variables, which is already handled by glsl-legalization pass.

I cleaned this up by moving the logic here to `getBuiltinGlobalVar` instead, so we don't double decorate ordianry user-defined global vars.

tests/spirv/tess-factor.slang
csyonghe: We should, but that can come in a separate PR.

tests/spirv/flat.slang
csyonghe: That test is written in GLSL and intended for glsl compatibility only.
So let's keep the test here to dedicate for correctness purpose.

jkwak-work (approved):

csyonghe:
Opened #4332 to track development of `InputPatch` and `OutputPatch`.

PR #4324, opened 2024-06-10
[gfx] Metal graphics support

skallweitNV:
Adds support for basic Metal graphics pipelines.

Resource binding for different shader stages is not handled correctly yet. But otherwise, this should support most of gfx API for graphics pipelines, framebuffers etc.

The swapchain is implemented using extra backbuffer images, doing an extra copy to the Metal renderable on present. This is to support the current API of gfx, which needs explicit access to the backbuffer images. Metal doesn't support this natively.


csyonghe (approved):
Looks good!

PR #4326, opened 2024-06-10
Enable more metal tests

skallweitNV:
Enable more metal tests that now work due to #4308

I updated #4291 to reflect the tests still failing.

PR #4327, opened 2024-06-10
Enable full test on macos.

### LINKED ISSUE #4328: Disable all failing metal tests and enable full testing on CI.

csyonghe:
Closes #4328.


skallweitNV:
That won't work yet. I only disabled crashing tests, not ones that generate incorrect results.

csyonghe:
I see. We need to disable them as well so we can enable this. I am also not sure whether those gfx unit tests will work.

skallweitNV:
Ok, lets first merge #4326 and then disable the list at the bottom of #4291

skallweitNV:
Actually you should probably just disable those tests as part of this PR

csyonghe:
Yeah, will do that.

PR #4331, opened 2024-06-10
[gfx] Metal texture fixues

skallweitNV:
- fix texture and texture view creation
- support texture upload from `initData`


tools/gfx/metal/metal-device.cpp
```diff
@@ -316,29 +321,29 @@ Result DeviceImpl::createTextureResource(
         break;
     }
 
-    NS::UInteger arrayLength = calcEffectiveArraySize(desc);
+    bool isArray = desc.arraySize > 0;
```
jkwak-work, line 324: Shouldn't this be 1 instead of 0?
```
bool isArray = desc.arraySize > 1;
```
The previous code was using `> 1`, and it would make more sense to be an array when the size is bigger than 1.

tools/gfx/metal/metal-device.cpp
```diff
@@ -461,51 +510,35 @@ Result DeviceImpl::createTextureView(
 {
     AUTORELEASEPOOL
 
-    auto resourceImpl = static_cast<TextureResourceImpl*>(texture);
-    RefPtr<TextureResourceViewImpl> view = new TextureResourceViewImpl(this);
-    view->m_desc = desc;
-    view->m_device = this;
-    if (texture == nullptr)
+    auto textureImpl = static_cast<TextureResourceImpl*>(texture);
+    RefPtr<TextureResourceViewImpl> viewImpl = new TextureResourceViewImpl(this);
```
jkwak-work, line 514: It is unlikely to happen but you will need to return an error code when the memory allocation fail.

jkwak-work (approved):
Looks good to me

tools/gfx/metal/metal-device.cpp
csyonghe: There is a difference between an 1-layer array texture vs a normal texture. To distinguish between them, we use size=0 to mean not an array, and 1 to mean a size-1 array.

skallweitNV: exactly

tools/gfx/metal/metal-device.cpp
skallweitNV: fixed

jkwak-work (approved):

PR #4338, opened 2024-06-11
implicit register binding for hlsl to non-hlsl targets

ArielG-NV:
private review

### DIFF
--- a/source/slang/slang-parameter-binding.cpp
+++ b/source/slang/slang-parameter-binding.cpp
@@ -1151,7 +1151,6 @@ static void addExplicitParameterBindings_GLSL(
     if (hlslToVulkanLayoutOptions == nullptr || !hlslToVulkanLayoutOptions->canInferBindings())
     {
         _maybeDiagnoseMissingVulkanLayoutModifier(context, varDecl.as<VarDeclBase>());
-        return;
     }
 
     // We need an HLSL register semantic to to infer from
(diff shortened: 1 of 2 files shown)


source/slang/slang-parameter-binding.cpp
```diff
@@ -1190,20 +1190,31 @@ static void addExplicitParameterBindings_GLSL(
...
+    //resInfo = typeLayout->findOrAddResourceInfo(hlslInfo.kind);
     resInfo = typeLayout->findOrAddResourceInfo(hlslInfo.kind);
-    
+
     semanticInfo.kind = resInfo->kind;
     semanticInfo.index = UInt(hlslInfo.index);
     semanticInfo.space = UInt(hlslInfo.space);
-
+    
+    if (resInfo->count.isFinite() && resInfo->count.getFiniteValue() == 0)
```
csyonghe, line 1206: I don't understand what the following code is doing. Why do we set DescriptorTableSlot count to 0?

ArielG-NV: I don't know if this is currently the best way todo this, I will need to look around more

From what I understand, `DescriptorTableSlot` as a representation does not make sense if we are just mapping 1:1 `register(binding,set)`.

csyonghe: `DescriptorTableSlot` is `binding`. We rely on it to report the `binding` value for vulkan targets, so the DescirptorTableSlot offset should be exactly the same as the explicit binding value.

ArielG-NV: any reason we use `DescriptorTableSlot` as a binding offset?

csyonghe: We don't have any other dedicated enum to represent `the binding offset inside a descriptor set`. We use `DescriptorTableSlot` to mean exactly that for Vulkan.

ArielG-NV: 👍we store data same to how a user would use descriptor sets, that makes sense.

source/slang/slang-parameter-binding.cpp
```diff
@@ -1190,20 +1190,31 @@ static void addExplicitParameterBindings_GLSL(
     // If inference is not enabled for this kind, we can issue a warning
     if (!hlslToVulkanLayoutOptions->canInfer(vulkanKind, hlslInfo.space))
```
jkwak-work, line 1191: Due to your change at line 1154, `hlslToVulkanLayoutOptions` can be `nullptr` here.

tests/hlsl/register-syntax.slang
```diff
@@ -0,0 +1,48 @@
...
+//CHECK_HLSL: u7
+//CHECK_HLSL: u9
+//CHECK_HLSL: u12
+
+//CHECK_GLSL: binding = 4, set = 2
+//CHECK_GLSL: binding = 5, set = 1
+//CHECK_GLSL: binding = 6
+//CHECK_GLSL: binding = 7
+//CHECK_GLSL: binding = 10
+//CHECK_GLSL: binding = 12
```
jkwak-work, line 15: The order of the variables may change.
You should include the variable name in the search pattern like,
```
//CHECK_HLSL: b0{{.*}}u4, space2
//CHECK_HLSL: b1{{.*}}u5, space1

ArielG-NV: very good point about the order, I think I need to use `CHECK-DAG` though since `CHECK` enforces ordering of statements. This is an issue since if `b0` and `b1` are swapped, `CHECK` would not find the statements.

csyonghe (approved):

PR #4339, opened 2024-06-11
Fix global value inlining for spirv_asm blocks.

csyonghe:


jkwak-work (approved):
Looks good to me

jkwak-work:
I am not sure if the regression is completed resolved by this change.
I saw the regression when I ran CTS test.
The validation error started showing up with the following commit,
```
38c0bacc Fix SPIRV emit for `Flat` decoration and TessLevel builtin. (#4318)
```
The error message looks something like following,
```
error: line 29: [VUID-StandaloneSpirv-Flat-04670] Flat decoration on target <id> '58[%vtx_out_result0]' storage class must be Input or Output
```
The same errors are still observed with this commit.

I am attaching a list of the failing CTS test cases for more information.
[ctslist.Flat.decorator.txt](https://github.com/user-attachments/files/15795802/ctslist.Flat.decorator.txt)

jkwak-work:
I created a bug for the regression. #4345

PR #4341, opened 2024-06-11
Fix crash when using optional type in a generic.

### LINKED ISSUE #4343: `Optional` type doesn't work in generic function.
The lowering logic of optional types is not robust against generics.
The following test case reproduces a crash:
```
//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-slang -compute
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUF):-slang -compute -vk

interface IFoo
{
    void foo();
}

struct S : IFoo { int x; void foo(); }

struct P
{
    IFoo f;
}
struct Tr
{
    int test<T:IArithmetic>(T t, inout P p)
    {
        const IFoo hit = p.f;
        let castResult = hit as S;
        if (!castResult.hasValue)
            return 0;
        return castResult.value.x;
    }
}

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name buffer

RWStructuredBuffer<int> buffer;

[numthreads(1,1,1)]
void computeMain()
{
    P p;
    S s;
    s.x = 2;
    p.f = s;
    Tr tt;
    // BUF: 2
    buffer[0] = tt.test(0, p);
}
```

csyonghe:
Closes #4343.

### DIFF
--- a/source/slang/slang-ir-lower-optional-type.cpp
+++ b/source/slang/slang-ir-lower-optional-type.cpp
@@ -15,6 +15,10 @@ namespace Slang
         InstWorkList workList;
         InstHashSet workListSet;
 
+        IRGeneric* genericOptionalStructType = nullptr;
+        IRStructKey* valueKey = nullptr;
+        IRStructKey* hasValueKey = nullptr;
+
         OptionalTypeLoweringContext(IRModule* inModule)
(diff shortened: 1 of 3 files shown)


jkwak-work (approved):
Looks good to me

PR #4344, opened 2024-06-12
Add option to preserve shader parameter declaration in output SPIRV.

### LINKED ISSUE #4340: There is no way to keep unused descriptor bindings when compiling to SpirV
Hi ! 

I am currently investigating porting a Vulkan codebase from glslc+spirv-opt to slang, but I am hitting a snag.

There doesn't seem to be a way to tell slang to not eliminate bindings that are not used in the shader (equivalent to `--preserve-bindings` with spirv-opt).  Dead binding elimination is still performed with `-O0`.

For example, the SpirV compiled from this:

	[[vk::binding(0)]] Sampler2D in_color; // declared, but never used
	[[vk::binding(1)]] RWTexture2D out_color;

	[shader("compute")]
	[numthreads(1, 1)]
	void main() {
		out_color[int2(0, 0)] = float4(1.0);
	}

Will have no mention of binding 0.

**Why would we ever want that?**

We use SpirV reflection to generate pipeline layouts at load time. This allows us to not have to search/create layouts per draw call. Pipeline layouts have to match descriptor sets pretty much exactly, which is fine when you can just declare the layout of your descriptor sets in a file and `#include` it where needed, but a massive problem when the compiler kills unused bindings. 

For example: A unlit shader might not read the light list from eg. descriptor set 0 binding 2. Having the binding be eliminated means that this shader will now generate a pipeline layout that isn't compatible with the descriptor set 0 (which might be frame global).

**Why not hardcode pipeline layouts on the cpp side?**

Doing this for a lot of shaders is a lot of work.

csyonghe:
Closes #4340.

This change adds a `-preserve-params` / `CompilerOptionName::PreserveParameters` option, together with `-O0`, this option can preserve the shader parameter declarations in the output SPIRV even if they are not used by the shader.

For example, given the following shader:
```
RWStructuredBuffer<float> buffer;

struct TT
{
    Texture2D tex;    
}
[numthreads(1, 1, 1)]
void f(ConstantBuffer<TT> t)
{
    return;
}
```

Compile with `slangc test.slang -target spirv -preserve-params -O0` will preserve the declarations of `buffer` and `tex` in the resulting SPIRV module.


source/slang/slang-ir-dce.h
```diff
@@ -12,6 +12,7 @@ namespace Slang
         bool keepExportsAlive = false;
         bool keepLayoutsAlive = false;
         bool useFastAnalysis = false;
+        bool keepGlobalParamsAlive = true;
```
jkwak-work, line 15: Shouldn't this be false by default. It seems little inconsistent to the lines above.

jkwak-work (approved):
Looks good to me.

source/slang/slang-ir-dce.h
csyonghe: true is more conservative than false, and it is safer not to remove something by default.

PR #4349, opened 2024-06-12
Fix crash on invalid entrypoint varying parameter.

### LINKED ISSUE #4350: Crash on invalid entry point parameter
Slang crashes when there is an entrypoint parameter contains resouce typed fields and the parameter is not decorated with `uniform`.

Reproducer:
```
RWStructuredBuffer<float> buffer;

struct TT
{
    Texture2D tex;    
}
[numthreads(1, 1, 1)]
void f(TT t)
{
    return;
}
```

This is because the layout for `t` will be invalid assuming `t` is varying input. We should diagnose an error instead of crashing.

csyonghe:
Closes #4350.


tests/bugs/invalid-entrypoint-param.slang
```diff
@@ -0,0 +1,17 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv
+
+// `TT` is not valid for defining a varying entrypoint parameter,
+// and we should diagnose an error.
+
+// CHECK: error
```
jkwak-work, line 6: This would be better if this check has an explicit error number, 39028, because it could match to an unexpected error elsewhere.

jkwak-work (dismissed):
Looks good to me.

PR #4354, opened 2024-06-12
Add compiler option to treat enum types as unscoped.

### LINKED ISSUE #4298: Add compile flag to allow for relaxed enum rules
When compiling shader source that's shared between multiple platforms, it would be useful to have a compiler flag to relax the enum rules in Slang to do the following:
1. Allow unscoped enum usage without the explicit `[UnscopedEnum]` attribute
2. Allow usage of anonymous enums

Compilers for other platforms may not be able to ignore unknown attributes resulting in a compile error. In some cases it not be desirable or infeasible for someone to modify their shader code base to adhere to Slang's requirements for scoped enums.

csyonghe:
Also add support for anonymous enum types.

Closes #4298.


jkwak-work (approved):
Looks good to me

csyonghe:
Thank you for all the prompt reviews!

PR #4359, opened 2024-06-12
Add constant folding for % operator.

csyonghe:


source/slang/slang-ir-sccp.cpp
```diff
@@ -933,6 +944,29 @@ struct SCCPContext
...
                 divisor);
         }
+        case kIROp_FRem:
+        case kIROp_IRem:
+        {
+            // Detect divide by zero error.
+            auto divisor = getLatticeVal(inst->getOperand(1));
+            if (divisor.flavor == LatticeVal::Flavor::Constant)
+            {
+                if (isIntegralType(divisor.value->getDataType()))
```
jkwak-work, line 954: Does it need to be an integral type?
We could warn when it is a floating type too, isn't it?

jkwak-work (approved):
Looks good to me.

source/slang/slang-ir-sccp.cpp
csyonghe: For float types, people may do (1.0/0.0) intentionally to get `nan`, and this is allowed in most c-like languages.

PR #4361, opened 2024-06-12
Delete glsl_vulkan and glsl_vulkan_one_desc targets.

### LINKED ISSUE #4271: Remove extraneous legacy glsl target names
Slang has three glsl targets:

`glsl`
`glsl-vulkan`
`glsl_vulkan-one-desc`

I would expect `glsl` to be OpenGL GLSL, but it spits out `gl_VertexIndex` instead of `gl_VertexId` for example, which is Vulkan GLSL. So all three targets are actually Vulkan GLSL

csyonghe:
Closes #4271.

## Breaking Change:

Previously deprecated `SLANG_GLSL_VULKAN` and `SLANG_GLSL_VULKAN_ONE_DESC` compile targets are being removed.


jkwak-work (approved):
Looks good to me.

PR #4362, opened 2024-06-12
Fix emit logic for getElementPtr.

csyonghe:
Fix the emit logic of `getElementPtr` on targets with pointer support.

Previously, if we have 
```
vec3 v;
v[1] = 2;
````
We will emit:
```
*&(&v)[int(0)] = 2
```
This is wrong.
With this change we now emit:
```
v.y = 2
```
which leads to correct result.


jkwak-work (dismissed):
Looks good to me.

I hope this can solve many problems in Metal.

PR #4364, opened 2024-06-12
Remove `IRHLSLExportDecoration` and `IRKeepAliveDecoration` for non-CUDA/Torch targets

### LINKED ISSUE #4289: Slangc crashes when using  `[AutoPyBindCUDA]` and `[shader("compute")]` in the same file
Using slangtorch's loadModule on a slang file that defines both an `[AutoPyBindCUDA]` entry point and and a `[shader("compute")]`  entry point (different functions) causes `slangc` to crash ("*Compilation failed with error 3221225477*" in Python traceback, no stderr from slangc).

Example:

```csharp
[AutoPyBindCUDA]
[CUDAKernel]
void square(TensorView<float> input, TensorView<float> output)
{}

[shader("compute")]
[numthreads(1,1,1)]
void computeMain(uint3 threadId: SV_DispatchThreadID)
{}
```

saipraveenb25:
Fixes #4289.


tests/autodiff/hlsl-torch-cross-compile.slang
```diff
@@ -0,0 +1,54 @@
...
+{
+    v[0] = func1(v[0]);
+    v[1] = func1(v[1]);
+}
+
+// Shouldn't see torchMain (or its transformations) anywhere in the HLSL output
+
+// HLSL: {{.*}}func1{{.*}}
+// HLSL-NOT: {{.*}}torchMain{{.*}}
+// HLSL: {{.*}}computeMain{{.*}}
```
jkwak-work, line 29: I don't think you need to have `{{.*}}` at the beginning and the end of the line.
And you may want to add more "HLSL-NOT" case above and beyond like this:
```
// Shouldn't see torchMain (or its transformations) anywhere in the HLSL output
// HLSL-NOT:torchMain
// HLSL:func1
// HLSL-NOT:torchMain
// HLSL:computeMain
// HLSL-NOT:torchMain
```

jkwak-work (approved):
Looks good to me.

Is there an issue number related to this PR?

tests/autodiff/hlsl-torch-cross-compile.slang
saipraveenb25: Oh yeah.. this would be cleaner.

saipraveenb25:
Yes, I've added the issue number now.

csyonghe (approved):

Issue #3087, opened 2023-08-10
HLSL SV_InstanceId not supported in pixel shader


danginsburg:
The following shader fails to compile:
```
struct PSInput
{
    uint vInstance : SV_InstanceID;
	float4 color : COLOR;
};

float4 MainPs(PSInput input) : SV_TARGET
{
	return input.color + float(input.vInstance).xxxx;
}
```

with output:

```
slangc.exe sv_instance.hlsl -o sv_instance.hlsl.spv -profile glsl_450 -entry MainPs -enable-effect-annotations -stage fragment -fvk-b-shift 0 0 -fvk-s-shift 14 0 -fvk-t-shift 30 0 -fvk-u-shift 158 0 -dump-intermediates
glslang:  sv_instance.hlsl(7): error :  'gl_InstanceIndex' : undeclared identifier
glslang:  sv_instance.hlsl(7): error :  '' : compilation terminated
glslang: note : ERROR: 2 compilation errors.  No code generated.
```

This can be worked around by passing the SV_InstanceID from VS -> PS.

csyonghe:
This is a GLSlang limitation, and we will need to wait until the SPIRV backend is ready before we can fix this issue.

natduca:
Putting in Q2 since we're trying to get spirv backend stable in Q1, then we can get to this after

csyonghe:
DXC generates this:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %MainPs "MainPs" %in_var_SV_InstanceID %in_var_COLOR %out_var_SV_TARGET
               OpExecutionMode %MainPs OriginUpperLeft
               OpSource HLSL 600
               OpName %in_var_SV_InstanceID "in.var.SV_InstanceID"
               OpName %in_var_COLOR "in.var.COLOR"
               OpName %out_var_SV_TARGET "out.var.SV_TARGET"
               OpName %MainPs "MainPs"
               OpDecorate %in_var_SV_InstanceID Flat
               OpDecorate %in_var_SV_InstanceID Location 0
               OpDecorate %in_var_COLOR Location 1
               OpDecorate %out_var_SV_TARGET Location 0
```

There is no special `Builtin` decoration applied, so my understanding is that the vertex shader still needs to write instanceID explicitly as a vertex output for this shader to work. So this is not really simplifying anything in the user code. 

@danginsburg Am I missing anything here?

danginsburg:
[...]

PR #4368, opened 2024-06-12
SPIR-V `SV_InstanceId` support in pixel shader

ArielG-NV:
fixes: #3087

Problem:
1. SV_InstanceID (HLSL) is allowed as an output semantic with HLSL, it is also allowed for input into pixel shader. We need to account for multiple entry points in 1 file using both of these scenarios at once.
2. SPIRV does not allow `SV_InstanceID` equivalent as a builtin, `SV_InstanceID` must be passed as a regular `Flat` `Input` from vertex shader.

Solution: Slang needs to treat these SV objects as non built-in's. As a result:
1. Slang needs to allocate for vertex output and fragment Input binding slots for all SV_InstanceID uses (if the target is SPIRV). This allows Slang to prepare an open slot to bind these SV objects to.
2. Slang needs to not emit built in modifier for these not actual built in variables (under the specific circumstances described).

Note: we do not need to add `Flat` ourselves to any fragment `Input` variables since the parameter being transformed is a `uint`. This means to avoid validation errors `Flat` must be added anyways, and Slang implements this behavior for us.


tests/bugs/gh-3087.slang
```diff
@@ -0,0 +1,24 @@
+//TEST:SIMPLE(filecheck=HLSL): -allow-glsl -entry main -target hlsl
+//TEST:SIMPLE(filecheck=GLSL): -allow-glsl -entry main -target glsl
+//TEST:SIMPLE(filecheck=SPIRV): -allow-glsl -entry main -target spirv
```
jkwak-work, line 3: I don't think you need `-allow-glsl`.

source/slang/slang-ir-inst-defs.h
```diff
@@ -1129,7 +1129,7 @@ INST(CudaKernelLaunch, CudaKernelLaunch, 6, 0)
 INST(GetEquivalentStructuredBuffer,     getEquivalentStructuredBuffer, 1, 0)
 
 /* Layout */
-    INST(VarLayout, varLayout, 1, HOISTABLE)
+    INST(VarLayout, varLayout, 1, 0)
```
jkwak-work, line 1132: My understanding is that HOISTABLE means it is some kind of global property.
I think VarLayout should remain HOISTABLE.
What is the reason why you "don't believe it needs to be HOISTABLE."

jkwak-work (commented):
Correct me if I am missing something.
But I think the variable needs to be marked as `Flat`.
I don't see anywhere in your change to make it `Flat`.
Also make sure you are testing that the keyword `Flat` is present in the emitted Spirv-asm

source/slang/slang-ir-inst-defs.h
ArielG-NV: > "don't believe it needs to be HOISTABLE."

I checked and I realized my assumption was incorrect, we do need `HOISTABLE`. The reason we need it is due to the optional operands which need to be shared, this was an incorrect assumption (it is why tests are failing)

ArielG-NV:
> Also make sure you are testing that the keyword Flat is present in the emitted Spirv-asm

I added to the test, this is a good suggestion (it is a VK validation error to not have `Flat`)

> But I think the variable needs to be marked as Flat.
> I don't see anywhere in your change to make it Flat.

We do not need to add `Flat` ourselves to any fragment `Input` variables since the parameter being transformed is a `uint`.
This means to avoid validation errors `Flat` must be added anyways, and Slang implements this behavior for us.
This is behavior which Slang implements is used in more than 1 instance, so it is likely correct to be using this behavior.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -1285,6 +1285,33 @@ ScalarizedVal createSimpleGLSLGlobalVarying(
         declarator,
         &systemValueInfoStorage);
 
+    {
+        uint32_t index = 0;
+        IRSystemValueSemanticAttr* systemSemantic = nullptr;
```
csyonghe, line 1290: `systemSemantic = inVarLayout->findSystemValueSemanticAttr()l`

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -1285,6 +1285,33 @@ ScalarizedVal createSimpleGLSLGlobalVarying(
...
+            {    
+                systemSemantic = systemValueSemanticAttr;
+                break;
+            }
+        }
+        // Validate the system value, convert to a regular parameter if this is not a valid system value for a given target.
+        if (systemSemantic && isSPIRV(codeGenContext->getTargetFormat()) && systemSemantic->getName().caseInsensitiveEquals(UnownedStringSlice("sv_instanceid"))
+            && ((stage == Stage::Fragment) || (stage == Stage::Vertex && inVarLayout->usesResourceKind(LayoutResourceKind::VaryingOutput))))
+        {
+            IRCloneEnv cloneEnv;
```
csyonghe, line 1303: Instead of using `IRClone`, just use a `List` to store all the operands you want to retain, and create a new inst with that list of operands. Simply call `emitIntrinsicInst` to do that. There is no need to use the clone mechanim and the `cloneInstExcludingSomeOperands` function, since the layout inst will not have any children.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -1285,6 +1285,38 @@ ScalarizedVal createSimpleGLSLGlobalVarying(
...
+    {
+
+        auto systemSemantic = inVarLayout->findAttr<IRSystemValueSemanticAttr>();
+        // Validate the system value, convert to a regular parameter if this is not a valid system value for a given target.
+        if (systemSemantic && isSPIRV(codeGenContext->getTargetFormat()) && systemSemantic->getName().caseInsensitiveEquals(UnownedStringSlice("sv_instanceid"))
+            && ((stage == Stage::Fragment) || (stage == Stage::Vertex && inVarLayout->usesResourceKind(LayoutResourceKind::VaryingOutput))))
+        {
+            ShortList<IRInst*> newOperands;
+            auto opCount = inVarLayout->getOperandCount();
+            newOperands.setCount(opCount);
```
csyonghe, line 1297: Just call `reserveOverflowBuffer` and `add` in the loop to get rid of the `opCount` here.

source/slang/slang-ir-clone.cpp
```diff
@@ -100,6 +100,47 @@ IRInst* cloneInstAndOperands(
     return newInst;
 }
 
+IRInst* cloneInstExcludingSomeOperands(
```
csyonghe, line 103: No longer required?

csyonghe (approved):

PR #4372, opened 2024-06-13
capture/replay: implement infrastructure for capture

kaizhangNV:
Add infrastructure for output stream
    This is the interface to record the method and parameter, and also
    provide functionality to write all the serialized data into file.

Add capture manager:
    Capture manager is associated to global session object, it will
    provide the functionality to actual record all the APIs.

Implement some of parameter encoder functions.

### DIFF
--- a/source/core/slang-stream.h
+++ b/source/core/slang-stream.h
@@ -138,7 +138,10 @@ public:
     void setContent(const void* contents, size_t contentsSize)
     {
         m_ownedContents.setCount(contentsSize);
-        ::memcpy(m_ownedContents.getBuffer(), contents, contentsSize);
+        if (contentsSize > 0)
+        {
+            ::memcpy(m_ownedContents.getBuffer(), contents, contentsSize);
+        }
         _setContents(m_ownedContents.getBuffer(), m_ownedContents.getCount());
     }
 
--- a/source/slang-capture-replay/capture-manager.cpp
+++ b/source/slang-capture-replay/capture-manager.cpp
@@ -1,0 +1,53 @@
+
+#include <string>
+#include <sstream>
+#include <thread>
+#include "capture-manager.h"
+
+namespace SlangCapture
+{
+    CaptureManager::CaptureManager(uint64_t globalSessionHandle)
+        : m_encoder(&m_memoryStream)
+    {
+        std::stringstream ss;
+        ss << "gs-"<< globalSessionHandle <<"t-"<<std::this_thread::get_id() << ".cap";
+        m_fileStream = std::make_unique<FileOutputStream>(ss.str());
+    }
+
+    void CaptureManager::clearWithHeader(const ApiCallId& callId, uint64_t handleId)
+    {
+        m_memoryStream.flush();
+        FunctionHeader header {};
+        header.callId = callId;
+        header.handleId = handleId;
+
+        // write header to memory stream
+        m_memoryStream.write(&header, sizeof(FunctionHeader));
+    }
+
+    ParameterEncoder* CaptureManager::beginMethodCapture(const ApiCallId& callId, uint64_t handleId)
+    {
+        clearWithHeader(callId, handleId);
+        return &m_encoder;
+    }
+
+    void CaptureManager::endMethodCapture()
+    {
+        FunctionHeader* pHeader = const_cast<FunctionHeader*>(
+                reinterpret_cast<const FunctionHeader*>(m_memoryStream.getData()));
+
+        pHeader->dataSizeInBytes = m_memoryStream.getSizeInBytes() - sizeof(FunctionHeader);
+
+        std::hash<std::thread::id> hasher;
+        pHeader->threadId = hasher(std::this_thread::get_id());
+
+        // write capture data to file
+        m_fileStream->write(m_memoryStream.getData(), m_memoryStream.getSizeInBytes());
+
+        // take effect of the write
+        m_fileStream->flush();
+
+        // clear the memory stream
+        m_memoryStream.flush();
+    }
+}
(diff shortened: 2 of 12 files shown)


source/slang-capture-replay/output-stream.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    FileOutputStream::~FileOutputStream()
+    {
+        m_fileStream.close();
+    }
+
+    void FileOutputStream::write(const void* data, size_t len)
+    {
+        SLANG_CAPTURE_ASSERT(m_fileStream.write(data, len));
```
jkwak-work, line 29: Does `write` return zero when `len` is zero?

source/slang-capture-replay/parameter-encoder.h
```diff
@@ -0,0 +1,51 @@
...
+        void encodeStruct(slang::PreprocessorMacroDesc const& desc);
+        void encodeStruct(slang::CompilerOptionEntry const& entry);
+        void encodeStruct(slang::CompilerOptionValue const& value);
+        void encodeStruct(slang::TargetDesc const& targetDesc);
+
+    private:
+        template <typename T>
+        void encodeValue(T value)
+        {
+            m_stream->write(&value, sizeof(T));
```
jkwak-work, line 45: This looks like just writing the binary fata as is
 What kind of "encode" is supposed to happen in this function given that the name of the function is "encodeValue".

source/slang-capture-replay/output-stream.cpp
kaizhangNV: Yes, it only return false when the stream cannot be written.

source/slang-capture-replay/parameter-encoder.h
kaizhangNV: Basically, all the scalar value should fall into this function.

source/slang-capture-replay/slang-global-session.cpp
```diff
@@ -1,5 +1,5 @@
-
-#include <vector>
+#include <sstream>
+#include <thread>
```
jkwak-work, line 2: Are these needed?

source/slang-capture-replay/output-stream.cpp
jkwak-work: Here is the implementation of "write()"
```
SlangResult FileStream::write(const void* buffer, size_t length)
{
    auto bytesWritten = fwrite(buffer, 1, length, m_handle);
    return (bytesWritten == length) ? SLANG_OK : SLANG_FAIL;
}
```
If the `len` is zero, it will return a value, zero, and it will trigger an assertion failure and it will crash the application unnecessarily.
I like to suggest to change to the following,
```
    void FileOutputStream::write(const void* data, size_t len)
    {
        if (len == 0) return;
        SLANG_CAPTURE_ASSERT(m_fileStream.write(data, len));
    }
```

source/slang-capture-replay/capture-manager.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "capture-manager.h"
+
+namespace SlangCapture
+{
+    CaptureManager::CaptureManager(uint64_t globalSessionHandle)
+        : m_encoder(&m_memoryStream)
+    {
+        std::stringstream ss;
+        ss << "gs-"<< globalSessionHandle <<"t-"<<std::this_thread::get_id() << ".cap";
+        m_fileStream = std::make_unique<FileOutputStream>(ss.str());
```
jkwak-work, line 14: This will create the file in a directory where the application started.
I am not sure if that will be an always available option.
We may need a way to specify the directory where the captured data gets stored.

source/core/slang-stream.h
```diff
@@ -138,7 +138,10 @@ class OwnedMemoryStream : public MemoryStreamBase
     void setContent(const void* contents, size_t contentsSize)
     {
         m_ownedContents.setCount(contentsSize);
-        ::memcpy(m_ownedContents.getBuffer(), contents, contentsSize);
+        if (contentsSize > 0)
+        {
+            ::memcpy(m_ownedContents.getBuffer(), contents, contentsSize);
+        }
```
jkwak-work, line 144: This change is unnecessary.
`memcpy` can handle when the size is zero.
However, the behavior is undefined if `contents` is nullptr.

jkwak-work (approved):
Looks good to me

source/core/slang-stream.h
kaizhangNV: I don't want us hit the undefined behavior, so made this change.

source/slang-capture-replay/slang-global-session.cpp
kaizhangNV: No need anymore. I remove some the logic into capture-manager.cpp

source/slang-capture-replay/capture-manager.cpp
kaizhangNV: This is a temporary solution, I will think of some better way for the directory.
We fill definitely create a folder to store those capture files.

source/slang-capture-replay/slang-global-session.cpp
kaizhangNV: I remove those in the latest update.

source/slang-capture-replay/capture-manager.cpp
csyonghe: That can be specified as a separate environment variable.

csyonghe (approved):
Looks good.

source/core/slang-stream.h
jkwak-work: If I wasn't clear, I was suggesting the following code,
```
        if (contents)
        {
            ::memcpy(m_ownedContents.getBuffer(), contents, contentsSize);
        }
```
Your current code can still hit the undefined behavior.
If contents is nullptr, the behavior is undefined.

Issue #4377, opened 2024-06-13
CPP/CUDA/Metal target crashes with use of GLSL shader inputs 


ArielG-NV:
`kIROp_GlobalVariableShadowingGlobalParameterDecoration` is applied on some GlobalVar's when using GLSL syntax to define objects.

As a result `introduceExplicitGlobalContext` fails due to the function never handling/expecting IRDecoration on a GlobalVar.

PR #4378, opened 2024-06-13
Fix cuda/cpp/metal crash when using GLSL shader inputs

ArielG-NV:
fixes: #4377
Decorations were not expected as an input, this causes a crash since there is no 'context pointer' for a decoration use site.

Previously `layout(location = 0) in highp vec4 a_position;` caused a crash, now it does not.

### DIFF
--- a/source/slang/slang-ir-explicit-global-context.cpp
+++ b/source/slang/slang-ir-explicit-global-context.cpp
@@ -462,11 +462,15 @@ struct IntroduceExplicitGlobalContextPass
         for( IRUse* use = globalVar->firstUse; use; use = nextUse )
         {
             nextUse = use->nextUse;
+            auto user = use->getUser();
+
+            // Ensure the use site checked actually requires a replacement
+            if (as<IRDecoration>(user))
+                continue;
 
             // At each use site, we need to look up the context
             // pointer that is appropriate for that use.
             //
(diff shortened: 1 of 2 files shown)


source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -462,11 +462,15 @@ struct IntroduceExplicitGlobalContextPass
         for( IRUse* use = globalVar->firstUse; use; use = nextUse )
         {
             nextUse = use->nextUse;
+            auto user = use->getUser();
+
+            // Ensure the use site checked actually requires a replacement
+            if (as<IRDecoration>(user))
+                continue;
```
jkwak-work, line 469: Shouldn't you limit this to `kIROp_GlobalVariableShadowingGlobalParameterDecoration`?
Is it safe to apply to all `IRDecoration`?

ArielG-NV: I thought about doing that but it looks like that would be incorrect with the function logic.

The function is trying to look at uses and crashes if a use is not a child of a function. This does not make sense since IRDecoration's are not often a child of a function (if used to store auxiliary data the Decoration should not be blindly modified).

csyonghe (approved):

PR #4382, opened 2024-06-13
fix the clang/gcc warning

ArielG-NV:


kaizhangNV (approved):

source/slang/slang-capability.cpp
```diff
@@ -201,10 +201,12 @@ void CapabilitySet::addPermutationsOfConjunctionForEachInContainer(CapabilityAto
 
         if constexpr (keyholeAtomToPermuteWith == CapabilityName::target)
         {
+            SLANG_UNUSED(knownTargetAtom);
```
jkwak-work, line 204: You can move this to line 194 at the beginning of the function and remove the other line

jkwak-work (approved):
Looks good to me.

Were you able to reproduce it tho?

ArielG-NV:
> Were you able to reproduce it tho?

I can repro the warning in some scenarios, not all.

csyonghe (approved):

jkwak-work (approved):
Looks good to me.
But I am little worried that this kind of problem will happen again if we cannot figure out how to reproduce it consistently.

Issue #4375, opened 2024-06-13
[Metal] Hoist vertex attributes.


csyonghe:
Given
```
struct AssembledVertex
{
	float3	position;
	float3	color;
        float2  uv;
};

struct CoarseVertex
{
	float3	color;
        float2  uv;
};

struct VertexStageInput
{
	AssembledVertex assembledVertex	: A;
};

float4 vertexMain(VertexStageInput input) : SV_Position
{
	VertexStageOutput output;
	return output;
}
```

We generate this:
```
struct AssembledVertex_0
{
    float3 position_0;
    float3 color_1;
    float2 uv_1;
};

struct vertexInput_0
{
    AssembledVertex_0 assembledVertex_0 [[attribute(0)]]; // metal does not allow [[attribute]] here
};

#line 49
[[vertex]] VertexStageOutput_0 vertexMain(vertexInput_0 _S1 [[stage_in]], SLANG_ParameterGroup_Uniforms_0 constant* Uniforms_0 [[buffer(1)]])
{
}
```

Metal does not allow defining `[[attribute]]` on composite types. We need to hoist all vertex attributes out to the top level of the vertex input struct.

csyonghe:
We should be able to reuse the `EntryPointVaryingParamLegalizeContext` pass to first hoist, flatten and SOA-ize all the nested attributes and fragment stage_in inputs into individual top level parameters, and then pack these parameters into a struct decorated with `[[stage_in``.

We can first translate `vertexMain` into something like
```
[attribute(0)] in float3 position;
[attribute(1)] in float3 color;
[attribute(2)] in float2 uv;
float4 vertexMain() : SV_Position
{
}
```
The `EntryPointVaryingParamLegalizeContext` should have everything we need to do this transform.

Then we pack them all into a struct, and pass that struct as `[stage_in]` parameter:
```
struct StageIn_vertexMain
{
[attribute(0)] float3 position;
[attribute(1)] float3 color;
[attribute(2)] float2 uv;
};
float4 vertexMain(StageIn_vertexMain stageIn [[stage_in]]) : SV_Position
{
...
}
```

jkwak-work:
I see. That make sense.

ArielG-NV:
Note for implementing:

from `legalizeEntryPointForMetal`:
```
        hoistEntryPointParameterFromStruct(entryPoint);
        packStageInParameters(entryPoint);
        legalizeSystemValueParameters(entryPoint, sink);
[...]

PR #4385, opened 2024-06-13
Implement for metal `SV_GroupIndex`

ArielG-NV:
fixes: #4370 

 1. If we don't have `SV_GroupThreadId` available we create one using `SV_GroupIndex`s location data since `SV_GroupIndex` is not an actual SV for metal.
 2. We emit code emulating `SV_GroupThreadId` from the same logic that CUDA/CPP uses.
 
 note: added `SV_GroupId` translation to `threadgroup_position_in_grid` so I can get a `SV_GroupIndex` test to run


source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -671,10 +681,66 @@ namespace Slang
...
+                    auto uintType = builder.getUIntType();
+                    auto newGroupIndex = svBuilder.createGlobalVar(uintType);
+                    svBuilder.addNameHintDecoration(newGroupIndex, UnownedStringSlice("sv_groupindex"));
+
+                    svBuilder.setInsertBefore(entryPoint.entryPointFunc->getFirstOrdinaryInst());
+                    auto computeExtent = emitCalcGroupExtents(svBuilder, entryPoint.entryPointFunc, builder.getVectorType(uintType, builder.getIntValue(builder.getIntType(), 3)));
+                    auto groupIndexCalc = emitCalcGroupThreadIndex(svBuilder, groupThreadId, computeExtent);
+                    svBuilder.emitStore(newGroupIndex, groupIndexCalc);
+
+                    param->replaceUsesWith(newGroupIndex);
```
csyonghe, line 742: What type does `param` have? Isn't it a `int` instead of `int*`? If so we should just replace it with `groupIndexCalc`? Why do we need to replace a `param` which is local to the entrypoint, with a global variable? We should just insert the computation of the group index within the entrypoint and use the SSA value of the index as the param?

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -177,6 +177,118 @@ void assign(IRBuilder& builder, LegalizedVaryingVal const& dest, IRInst* src)
     assign(builder, dest, LegalizedVaryingVal::makeValue(src));
 }
 
+
```
csyonghe, line 180: I didn't know we already had this logic. Good find!

csyonghe:
You may also want to enable all the tests that this PR fixes. Do you have a mac to test locally?

jkwak-work (commented):
Can you avoid using special characters for the branch name next time?
It give a hard time for my scripts to process.

source/slang/slang-ir-legalize-varying-params.cpp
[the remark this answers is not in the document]
csyonghe: Reusing `EntryPointVaryingParamLegalizeContext` here should also allow us to fix https://github.com/shader-slang/slang/issues/4375, because we can just use the logic here to SOA-ify all the varying inputs and hoist them all to top level.

ArielG-NV:
> You may also want to enable all the tests that this PR fixes. Do you have a mac to test locally?

I do not have a Mac locally setup,

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -177,6 +177,118 @@ void assign(IRBuilder& builder, LegalizedVaryingVal const& dest, IRInst* src)
...
+    // where `groupExtents` is the X,Y,Z extents of
+    // each thread group in threads (as given by
+    // `[numthreads(X,Y,Z)]`).
+
+    return builder.emitAdd(type,
+        builder.emitMul(type,
+            groupID,
+            groupExtents),
+        groupThreadID);
+}
```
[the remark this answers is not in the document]
ArielG-NV: I can't move this function back currently (may change due to other review comments and redesign of code).

I can't move it since `emitCalcGroupExtents` was originally inside a class which is not supposed to be available beyond the current compilation unit.

source/slang/slang-ir-metal-legalize.cpp
ArielG-NV: I see your point, I did not need to make a globalVar

csyonghe (approved):

csyonghe:
I am  going to merge this PR as is. It seems that a natural next step following this is to investigate how we can reuse the entrypoint param legalization logic for metal, so that we can convert all vertex attribute inputs and fragment stage_in inputs buried inside nested structs into top level entry point parameters.

jkwak-work:
Yes, we will do the refactoring as a part of #4375.

Issue #768, opened 2019-01-05
Add a Bindless<T> type


tangent-vector:
Renderers that can assume modern GPU hardware, and especially those doing ray tracing, may want to use “bindless” idioms. Slang can/should simplify this task with a carefully designed language feature.

# What is Bindless?

The key to any bindless approach is that there exists some kind of handle type that can in principle refer to (more or less) any texture/buffer/resource at the use site. This allows references to resources to be encoded as “plain old data,” which can improve the simplicity and flexibility of engine code for passing data to the GPU.

OpenGL has long supported bindless directly via extensions, with the handles for typed resources being opaque 64-bit values, and the handles for flat buffers being GPU virtual addresses. This amounts to deciding by fiat that the `Texture2D` type is 64 bits of plain old data, and subsequently allowing it to be packed in buffers however users like. This strictly increases the flexibility of GLSL by allowing types that used to be illegal (e.g., a `struct` with a texture in it).

# What’s the problem?

In contrast to OpenGL with extensions, achieving a “bindless” approach on D3D12 or Vulkan requires a lot more explicit work on the user’s part, and tends to have a large impact on shader code. A typical D3D12 approach would be to maintain a single large descriptor heap for resource views, and to represent a “handle” to a view as an index (potentially just 32 bits) into that heap.

A D3D12 application then needs to bind a monolithic descriptor table that spans the entire heap, and declare various unbounded-size arrays in shader code (e.g., `Texture2D gAllMyTextures[];`, `RWStructuredBuffer<Foo> gAllMyRWStructuredBuffersOfFoo[];`, etc.) that will be backed by that table. The existence of parameterized types like `RWStructureBuffer<_>` means that the number of such declarations required is unbounded and depends on the shader code that gets authored. Any shader code will need to explicitly use, e.g., `uint` instead of `Texture2D` in data structures, and then manually indirect through the appropriate global array when it needs to get from the handle to the resource it references.

(The above ignores a bunch of API-specific details which make it harder to write code that works seamlessly across D3D12 and Vulkan)

All of this complexity makes it hard to write a re-usable module that might include data types mixing resources and uniform data, if we want the module to work under both traditional binding models and application-emulated bindless. Basic questions like “should this `struct` field be a `uint` or a `Texture2D`?” can’t be answered without knowing how your module will be used.

## Ray Tracing Requirements

[...]

PR #4389, opened 2024-06-14
Add generic descriptor indexing intrinsic

### LINKED ISSUE #4351: Request for a generic descriptor indexing intrinsic
[Descriptor indexing](https://docs.vulkan.org/samples/latest/samples/extensions/descriptor_indexing/README.html) allows shaders to access resources in a natural and flexible way, while further eliminating the complexity of managing and updating descriptor sets. This feature can already be used today in Slang, but it requires that global bindings be declared for every accessed resource type and format beforehand, which can become a bit tedious and makes it difficult for users to implement abstractions on top.

Following from https://github.com/shader-slang/slang/issues/4150, it is not possible to define bindings inside a generic type due to design constraints in the compiler. What I propose instead, is for Slang to provide an intrinsic type that allows indexing over arbitrary resource types, similarly to `ResourceDescriptorHeap` in HLSL. This way, only one binding variable would need to be defined _per_ descriptor set and binding index:

```hlsl
// Provided by Slang
__magic_type(...)
struct __RuntimeDescriptorArray {
    __intrinsic_op(...)
    T Get<T>(uint index);
}; 

// User code
[vk::binding(0, 0)] 
__RuntimeDescriptorArray g_TextureDescriptors;  // treated as opaque / unbounded array by layout and reflection

let texture = g_TextureDescriptors.Get<Texture2D<uint2>>(...);
```

The SPIRV backend would then emit the appropriate `OpVariable` definitions as it finds the specialized Get() calls. Potentially, this could also be used to implement the HLSL global heaps in a relatively portable way, although I'm not familiar with their binding model.

If this gets approved, I would be willing to work on an implementation if that is preferred.

dubiousconst282:
Add generic descriptor indexing intrinsic

Closes #4351

---

This seems to be working nicely as far as I can tell, but I have a few questions:

- Type layout is created based on a temporary `StructuredBuffer<int>[]` type. Reflection will still report the type as the ResourceArray struct, but I'm not sure if this would cause any issues or if it should be done in some other way.

- Specialization will report an error if the ResourceArray variable is used by anything other than the get intrinsic. Should this be done elsewhere or is fine to keep it as is?

- Not sure about the naming, and whether it is better to have an enum instead of global constants like discussed in the original issue.


csyonghe:
I think you can just create an ArrayTypeLayout for the type, with the size being unsized and element type being a special "DynamicResourceEntry" type. The TypeLayout for DynamicResourceEntry type should be a SimpleTypeLayout that has one DescriptorTableSlot entry.

csyonghe:
I am not sure why this change need to interact with specialization, they should be orthogonal?

jkwak-work:
> I am not sure why this change need to interact with specialization, they should be orthogonal?

I think the specialization side code is trying to add global uniform variable for each types used on "get" method.
If I borrow the example from the proposed design, it will looks like this,
```
static const int Resource = 0;
static const int Sampler = 1;
interface IBuiltinResourceType<let kind: int>
{}

extension Texture2D : IBuiltinResourceType<Resource>
{}

extension Sampler2D : IBuiltinResourceType<Sampler>
{}

struct ResourceArray<let kind:int>
{
    T get<T:IBuiltinResourceType<kind>>(int index);
}

// Specialization code adds two following global variables,
// because "get" method was used with two different types:
// uniform Texture1D resArray_0[];
// uniform Texture2D resArray_1[];
ResourceArray<Resource> resArray;

void main()
{
    Texture1D t = resArray.get<Texture1D>(1);
    Texture2D t = resArray.get<Texture2D>(2);
}
```
I think this should happen in the glsl-legalization rather than specialization.

jkwak-work:
Here is my answers to the questions in the description

> Type layout is created based on a temporary StructuredBuffer<int>[] type. Reflection will still report the type as the ResourceArray struct, but I'm not sure if this would cause any issues or if it should be done in some other way.

I think the reflection side needs to be handled properly but I don't see any code change related to that.
We will need to start from writing a test case for it and clarify what kind of changes will be required.

> Specialization will report an error if the ResourceArray variable is used by anything other than the get intrinsic. Should this be done elsewhere or is fine to keep it as is?

As I mentioned on my previous comment, I think that should be handled in the legalization part not specialization part.

> Not sure about the naming, and whether it is better to have an enum instead of global constants like discussed in the original issue.

I think enum is better.

csyonghe:
Yes, we should just create those aliased global variables during glsl legalization pass. Basically, convert getResourceElementInst into getElement(aliasedGlobalArrayForResourceType).

dubiousconst282:
Thank you both for the review.

I have moved the aliased parameter generation code to legalization. At first I had implemented it in the SPIRV legalization pass, but then moved it to specialization so it could be shared with the GLSL target, without realizing it would do the wrong thing for the other targets.

I have not fully addressed the comments involving type layout and reflection yet, because I'm thinking that having just a "DynamicResource" type would be more general and maybe require less changes on that side (considering that it would be needed for proxying through the ArrayType anyway). Then the revised proposal would be something like:

```hlsl
struct __DynamicResource<let kind : __DynamicResourceKind = __DynamicResourceKind.General> 
{
    T as<T : __IDynamicResourceCastable<kind>>();
}
enum __DynamicResourceKind
{
    General = 0, // CBV_SRV_UAV
    Sampler = 1
}
interface __IDynamicResourceCastable<let kind : __DynamicResourceKind = __DynamicResourceKind.General> { }
```

I believe that with implicit casts, this would have full parity with the HLSL syntax in terms of usage:

```hlsl
__generic<T, Shape : __ITextureShape, let isArray : int, let isMS : int, let sampleCount : int, let access : int, let isShadow : int, let isCombined : int, let format : int>
extension __TextureImpl<T, Shape, isArray, isMS, sampleCount, access, isShadow, isCombined, format> : __IDynamicResourceCastable<__DynamicResourceKind.General> {
    __implicit_conversion(1)
    __init(__DynamicResource res);
}

[numthreads(1)]
void computeMain(int2 tid: SV_DispatchThreadID) {
    let tex1 = g_TextureHeap[0].as<RWTexture2D>();
    RWTexture2D tex2 = g_TextureHeap[0];

    tex1[tid] = 1.0;
}
```

Please let if I should proceed with this or keep the current interface.

csyonghe:
Sure, the new approach looks good to me too. In this case is the g_textureHeap just a normal unsized array of DynamicResource?

dubiousconst282:
Yes, it is user defined like before. (will edit previous comment to clarify.)

dubiousconst282:
Okay I think this is ready for another review.

The only thing that I can tell is missing is support for casting to buffer types like `StructuredBuffer<T>` and others that don't derive from ITextureImpl, although with pointers I personally feel like this isn't as critical and can be added later, maybe with the template generator thingy.

Change notes, fwiw, since I squashed everything:
- Legalization supports both a single DynamicResource and arrays
- Indirect access through functions works via specialization
- Reflection correctly reports it as `DynamicResource`. I preferred this because `Resource` reports additional info such as access and shape, which is not known.

source/slang/hlsl.meta.slang
```diff
@@ -19889,3 +19889,50 @@ struct ConstBufferPointer
...
+// For Khronos targets, `__DynamicResource` can be used to declare "untyped" global bindings as
+// usual (e.g. unsized arrays for descriptor indexing), which will then be materialized into 
+// new aliased bindings for each distinct cast type.
+//
+
+__magic_type(DynamicResourceType)
+__intrinsic_type($(kIROp_DynamicResourceType))
+struct __DynamicResource<let kind = __DynamicResourceKind.General>
+{
+    __intrinsic_op($(kIROp_CastDynamicResource))
```
csyonghe, line 19906: This is a downcast right?

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -3729,4 +3729,97 @@ void legalizeDispatchMeshPayloadForGLSL(IRModule* module)
     });
 }
 
+void legalizeDynamicResourcesForGLSL(CodeGenContext* context, IRModule* module)
+{
+    List<IRGlobalParam*> toRemove;
+    Dictionary<IRType*, IRGlobalParam*> aliasedParams;
```
csyonghe, line 3735: The dictionary is keyed on a single IRType, this seems incorrect if there are more than one DynamicResource global parameter. Two distinct dynamic resource global params should map to distinct aliased global params, so this dictionary needs to be cleared after processing each original global param, or be changed to key on a IRGlobalParam+IRType pair.

dubiousconst282: Fixed

source/slang/hlsl.meta.slang
csyonghe: I mean we are semantically treating all texture and buffer types as subtypes that conforms to IDynamicCastable, so casting from DynamicResource to a Texture is more of a downcast(casting from a generic to a more derived type) than a up cast.

source/slang/hlsl.meta.slang
```diff
@@ -19889,3 +19889,50 @@ struct ConstBufferPointer
...
+}
+
+enum __DynamicResourceKind
+{
+    General = 0, // CBV_SRV_UAV
+    Sampler = 1
+}
+
+__generic<T, Shape : __ITextureShape, let isArray : int, let isMS : int, let sampleCount : int, let access : int, let isShadow : int, let isCombined : int, let format : int>
+extension __TextureImpl<T, Shape, isArray, isMS, sampleCount, access, isShadow, isCombined, format> : __IDynamicResourceCastable<__DynamicResourceKind.General>
```
csyonghe, line 19920: I think we can simply add more extensions here to support all StructuredBuffer/ByteAddressBuffer types.

dubiousconst282: I have added a template to generate the extensions for types that HLSL allows, not sure if they are all needed.

source/slang/hlsl.meta.slang
dubiousconst282: Yes I think from a semantic perspective that makes sense.

csyonghe:
I want to bring this up explicitly here that the current implementation cannot handle cases where the global dynamic resource parameter is not referenced directly in use sites and instead are passed around with local variables or function parameters. For example, the following case will not work:


DynamicResource gRes;

```
void f(DynamicResource t)
{
    Texture2D tt = t;
    tt.Sample(...);
}

void main()
{
    f(gRes);
}
```

Which is fine as a limitation for now especially when this type will be used only internally in stdlib when we implement the full Bindless<T> support. But I want to make sure the limitation is well understood.

dubiousconst282:
I think this case is handled correctly by specialization (it is covered by the test), e.g.:

```hlsl
__DynamicResource g_SingleRes;
RWStructuredBuffer<float4> outputBuffer;

[noinline]
float4 IndirectAccess(__DynamicResource resource, int2 pos)
{
    Texture2D texture = resource;
    return texture[pos];
}

[numthreads(1)]
void computeMain(int2 tid: SV_DispatchThreadID) {
    outputBuffer[tid.x + tid.y * 128] = IndirectAccess(g_SingleRes, tid);
}
```

On the GLSL targets compiles to:
```glsl
#version 460
#extension GL_EXT_scalar_block_layout : require
#extension GL_EXT_samplerless_texture_functions : require
layout(row_major) uniform;
layout(row_major) buffer;

#line 73 0
layout(scalar, binding = 1) buffer StructuredBuffer_vectorx3Cfloatx2C4x3E_t_0 {
    vec4 _data[];
} outputBuffer_0;

#line 2364 1
layout(binding = 0)
uniform texture2D g_SingleRes_0;

#line 84 0
vec4 IndirectAccess_0(ivec2 _S1)
{

#line 79
    return (texelFetch((g_SingleRes_0), ((ivec2(uvec2(_S1)))), 0));
}

void main()
{

#line 83
    ivec2 _S2 = ivec2(gl_GlobalInvocationID.xy);
    outputBuffer_0._data[_S2.x + _S2.y * 128] = IndirectAccess_0(_S2);
    return;
}
```

Although I don't think this would be a significant limitation either.

csyonghe:
Ah, I see that you added a case for dynamic resource to be treated as specializable parameter type. That will work then.

dubiousconst282:
Can you clarify on how `__getResourceFromHandle()` would work? The thing with DynamicResource is that it allows users to specify the binding points directly, which I think maps more closely to how Vulkan's descriptor indexing works.

My intent for this type was that it would allow users to define their own Bindless type, since an earlier proposal had uncertainties on the representation and was closed (https://github.com/shader-slang/slang/issues/768).

csyonghe:
The user will need to specify the binding for the implicit heap via a compiler option, e.g. `-fvk-dynamic-descriptor-set 10`.

The reason to make the global array system defined instead of user defined is to avoid conflicts and allow reusing code written by different developers.

If library A and library B are written by two developers and they all want to use bindless model, how can one consume both A and B without worrying about conflicts?

The only way to facilitate seamless sharing is for the compiler to own this definition so there will always be one declaration of such array and there could never be conflicts.

csyonghe:
Internally, we will introduce a kIROp_GetGlobalResourceHeap inst that is hoistable so it will be globally dedupplicated after linking.

Then we make __getGlobalResourceHeap<ResourceKind>() intrinsic map to this op code.

Then we can have __getResourceFromHandle implemented by calling this intrinsic.

The binding for the global heap can be provide explicitly with a compiler option, or implicitly by the system where we just allocate the next available space after all other parameters have been assigned a binding. We will need the reflection API to be able to tell the user which space is allocated for the implicit heap.

csyonghe:
With this, the bindless type can simply be implemented in stdlib as an ordinary struct type:


```
struct Bindless<T: __BuiltinResourceType>
{
    uint64 handle;

    [ForceInline]
    T get() { return __getGlobalResourceHeap<T.ResourceKind>().get<T>(handle); }
} 
```

dubiousconst282:
From what I can tell, normal descriptors already suffer with mostly the same conflict issues, although I can see why having a more strict solution like you described is preferred.

I think having the binding points be defined via a compiler option instead instead of implicitly would be better.

Ideally the [finer resource types](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDescriptorType.html) that Vulkan defines compared to HLSL should be considered as well. Although [VK_EXT_mutable_descriptor_type](https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_mutable_descriptor_type.html) makes this less of an issue, it is still more limiting in some ways (not as widely supported as plain descriptor indexing, can't bind a texture as Image and SampledImage at the same time).

csyonghe:
I still want to have this PR merged in. No matter what we do for Bindless<T> they will be based on the changes here.

dubiousconst282:
Ah, sorry, I missed that `__getGlobalResourceHeap` will reuse what's implemented here.

dubiousconst282:
The tests `incomplete-type.slang` and `constant-buffer-unsized.slang` fail when the stdlib declares an extension for ConstantBuffer like attached below. I've had no luck debugging so far, maybe you have some hint for me to look for?

```hlsl
__generic<T>
extension ConstantBuffer<T> : __IDynamicResourceCastable<__DynamicResourceKind.General>
{
    __intrinsic_op($(kIROp_CastDynamicResource))
    __implicit_conversion($(kConversionCost_GenericParamUpcast))
    __init(__DynamicResource res);
}
```

csyonghe:
What kind of failure are you seeing? ConstantBuffer has an implicit conversion to T, adding another conversion here may be changing the diagnostic message that leads to a false failure.

dubiousconst282:
Both tests are supposed to trigger a diagnostic about invalid usage of ConstantBuffer, but this never happens when the extensions are present.

---

It turns out that this diagnostic check happens in the same method where default variable initialization is handled, and it will early return if there is no appliable initializer function for the variable type:

https://github.com/shader-slang/slang/blob/323fdff4544ac3b640bdc617d7f7d64ab56fbf5a/source/slang/slang-check-decl.cpp#L2082-L2086
https://github.com/shader-slang/slang/blob/323fdff4544ac3b640bdc617d7f7d64ab56fbf5a/source/slang/slang-check-decl.cpp#L2136-L2146

I don't think the early returns are correct, so I've changed them to a else block. I couldn't get all of the tests to run on my laptop so I'll wait to see what CI thinks

FacelessTiger:
I'd like to mention that I've seen multiple descriptor set setups for bindless. Some have each descriptor type in their own set so they can use VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT, others just put them at different bindings within the same set with a high limit, and others I've seen use the aforementioned [VK_EXT_mutable_descriptor_type](https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_mutable_descriptor_type.html) extension to put everything but samplers into one set with one binding.

If Slang decides to go with the defining binding points via compiler option route for bindless, I'd like to see it have enough flexibility to support all of these options

csyonghe:
The PR is looking good. @dubiousconst282 can you resolve the merge conflicts?

dubiousconst282:
@csyonghe Resolved

csyonghe (approved):

Ipotrick:
this looks exciting :O

PR #4392, opened 2024-06-14
[Metal] Fix global constant array emit.

### LINKED ISSUE #4373: [Metal] global static array.
We currently generate invalid metal code where there is a global static constant array.
For example, `tests/compute/static-const-array.slang` does not compile to valid metal.

Metal requires global constant arrays to be defined in `constant` address space, but we are missing address space declarations in our output.

csyonghe:
Fixes #4373.


jkwak-work (approved):
Looks good to me.

PR #4396, opened 2024-06-14
Improve Direct SPIRV Backend Test Coverage & Fix Associated Bug

ArielG-NV:
fixes: #4001

1. Added direct-spirv tests for 'raytracing' and 'texture-footprint'
2. Fixed bug with texture-footprint using direct-spirv path
3. Changed when we emit raytracing/rayquery extensions with backends (to reduce 'incorrect' extension emitting)

Note: 
1. `tests/hlsl-intrinsic/ray-tracing/rt-pipeline-intrinsics-ahit.slang` checks for the following code:
```
// SPIRV-DAG: OpTerminateRayKHR
// SPIRV-DAG: OpIgnoreIntersectionKHR
```
This code is checked using `DAG` since the used `OpBranch` has varying ordering when emitted by direct-spirv versus glslang-spirv.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -19209,8 +19209,7 @@ ${
             return spirv_asm {
                 OpCapability ImageFootprintNV;
                 OpExtension "SPV_NV_shader_image_footprint";
-                %sampledImage:__sampledImageType(this) = OpSampledImage $this $sampler;
-                %resultVal:$$__FootprintData = OpImageSampleFootprintNV %sampledImage $coords $granularity $useCoarseLevel;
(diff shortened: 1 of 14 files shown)


csyonghe (approved):

jkwak-work:
I am curious if this change can allow us to enable more CTS tests related to the ray tracing.

ArielG-NV:
>I am curious if this change can allow us to enable more CTS tests related to the ray tracing. 

I don't think the change will affect CTS raytracing tests.
I say this because the changes were to emit less 'unneeded' raytracing extensions (we don't emit SPV_EXT_ray_query when only SPV_EXT_ray_tracing is needed). I don't believe CTS checks for extra extensions being enabled.

This may allow us to enable more CTS tests in general though since texture-footprint had a bug previously.

Issue #4404, opened 2024-06-17
Implement for `kIROp_ImageSubscript` (GLSL/SPIRV/Metal) Multisampling support


ArielG-NV:
Problem:
Currently the following code does not compile correctly:
```
RWTexture2DArrayMS<int4> outputTexture4;
...
outputTexture4[0, 0].xz = int2(1,2); // fails
outputTexture4[int3(0, 0, 1), 1].xz = int2(3,4); //fail
```

Expected Behavior:
Code compiles correctly.

PR #4408, opened 2024-06-17
Expand upon existing `ImageSubscript` support (Metal, GLSL, SPIRV)

ArielG-NV:
fixes: #4394
fixes: #4404

1. Added ImageSubscript support for Metal and test case(s)
    * Merge GLSL/SPIRV/Metal `ImageSubscript` legalization pass
2. Added multisampled texture support to glsl/spirv/metal for when using ImageSubscript 
    * Metal does not support `read+write` on multisampled textures, check was added to verify target before legalization.
3. Fixed minor metal texture `Load`/`Read` bugs 
    * [HLSL methods of access do not support subscript accessor for texture cube array](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/texturecubearray) 
    * removed swizzling of uint/int/float
    * other odd bugs which were causing compile errors
    
   
Notes:
* Compute tests do not work due to what seems to be the GFX backend (causes crash without error report). These tests are disabled.
* resolves `tests/bugs/buffer-swizzle-store.slang` of #4291,


source/slang/hlsl.meta.slang
```diff
@@ -2420,7 +2422,7 @@ extension __TextureImpl<T,Shape,isArray,0,sampleCount,0,isShadow,isCombined,form
                 {
                     if (isArray == 1)
                         // T read(uint2 coord, uint face, uint array, uint lod = 0) const
-                        __intrinsic_asm "$0.read(vec<uint,2>(($1).xy), uint(($1).z), uint(($1).w))";
+                        __intrinsic_asm "<invalid intrinsics>";
```
csyonghe, line 2425: There is now a static_cast intrinsic that we can use to report an error.

csyonghe (dismissed):

source/slang/hlsl.meta.slang
ArielG-NV: Note about the solution: 
1. I have addressed the problem by adding `static_assert`.
2. I have kept `__intrinsic_asm` since I noticed else we get an stdlib compile error. The error seems to be due to the missing `__intrinsic_asm` (or an equivalent return statement).

ArielG-NV:
https://github.com/shader-slang/slang/issues/4411 seems to be related to this issue

source/slang/hlsl.meta.slang
```diff
@@ -2385,9 +2385,11 @@ extension __TextureImpl<T,Shape,isArray,0,sampleCount,0,isShadow,isCombined,form
                 // lod is not supported for 1D texture
                 if (isArray == 1)
                     // Tv read(uint coord, uint array, uint lod = 0) const
+                    //__intrinsic_asm "$0.read(uint(($1).x), uint(($1).y, uint(($1).z))";
```
kaizhangNV, line 2388: what is this comment for?

source/slang/slang-emit.cpp
```diff
@@ -1156,11 +1157,16 @@ Result linkAndOptimizeIR(
     // Legalize `ImageSubscript` and constant buffer loads for GLSL.
     switch (target)
     {
+    case CodeGenTarget::Metal:
```
kaizhangNV, line 1160: It looks like all 4 cases will call
```
legalizeImageSubscript(targetRequest, irModule, sink)
```
?

Can you combine them?

source/slang/slang-ir-legalize-image-subscript.cpp
```diff
@@ -0,0 +1,196 @@
...
+    {
+        IRBuilder builder(module);
+        for (auto globalInst : module->getModuleInst()->getChildren())
+        {
+            auto func = as<IRFunc>(globalInst);
+            if (!func)
+                continue;
+            for (auto block : func->getBlocks())
+            {
+                auto inst = block->getFirstInst();
```
kaizhangNV, line 177: can this 
```
auto inst = block->getFirstInst();
IRInst* next;
for ( ; inst; inst = next)

```

be

```
for (auto inst : block->getChildren())
```

source/slang/slang-ir-legalize-image-subscript.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "slang-parameter-binding.h"
+#include "slang-ir-legalize-varying-params.h"
+
+namespace Slang
+{
+    void legalizeImageSubscriptStore(TargetRequest* target, IRBuilder& builder, IRInst* storeInst, DiagnosticSink* sink)
+    {
+        builder.setInsertBefore(storeInst);
+        auto getElementPtr = as<IRGetElementPtr>(storeInst->getOperand(0));
+        IRImageSubscript* imageSubscript = nullptr;
```
kaizhangNV, line 17:
```
        auto getElementPtr = as<IRGetElementPtr>(storeInst->getOperand(0));
        IRImageSubscript* imageSubscript = nullptr;
        if (getElementPtr)
            imageSubscript = as<IRImageSubscript>(getElementPtr->getBase());
        else
            imageSubscript = as<IRImageSubscript>(storeInst->getOperand(0));
```

if can be simplified as
```
IRImageSubscript* imageSubscript = nullptr;
if (auto getElementPtr = as<IRGetElementPtr>(storeInst->getOperand(0)))
{}
else
{}
```

source/slang/hlsl.meta.slang
ArielG-NV: The comment was to not delete the LOD sampling code of 1D textures.
The comment should be removed since this behavior will likely not be supported by metal in the near future.

source/slang/slang-ir-legalize-image-subscript.cpp
```diff
@@ -0,0 +1,196 @@
...
+        if (getElementPtr)
+            imageSubscript = as<IRImageSubscript>(getElementPtr->getBase());
+        else
+            imageSubscript = as<IRImageSubscript>(storeInst->getOperand(0));
+        assert(imageSubscript);
```
kaizhangNV, line 22: This could cause problem in release build.
Please do not use `assert` here. Maybe use `SLANG_RELEASE_ASSERT`

source/slang/slang-ir-legalize-image-subscript.cpp
```diff
@@ -0,0 +1,196 @@
...
+        IRTextureType* textureType = as<IRTextureType>(imageSubscript->getImage()->getFullType());
```
kaizhangNV, line 23: nullptr check to avoid crash later.

source/slang/slang-ir-legalize-image-subscript.cpp
ArielG-NV: `SLANG_ASSERT` is correct behavior here.
`assert` was a mistake which was used in the old glsl version of the function.

source/slang/slang-ir-legalize-image-subscript.cpp
ArielG-NV: I don't exactly know why we need to check for a nullptr here. From what I understand:
1. if `imageSubscript` is `nullptr`, we have a very huge bug in Slang. Earlier in the call stack, the code `if(getRootAddr)` ensures we have a valid `imageSubscript` before hand.
2. if `imageSubscript->getImage()` is `nullptr` the code would have failed already and the language-server would not work with textures.

kaizhangNV:
This may not be a blocker for your change.
But I personally don't like the long function like `legalizeImageSubscriptStore` . It will always increase the cyclomatic complexity.
Especially `legalizeImageSubscriptStore` is definitely separable regarding the functionality..

ArielG-NV:
> It will always increase the cyclomatic complexity. Especially legalizeImageSubscriptStore is definitely separable regarding the functionality..

I agree, I will rename the function. Should be fine to assume the functionality is to legalize `ImageSubscript` since it is in the file `slang-ir-legalize-image-subscript.cpp`.

source/slang/slang-ir-legalize-image-subscript.cpp
ArielG-NV: `getElementPtr` is used in a later scope, I cannot move it into an inline if assignment

source/slang/slang-ir-legalize-image-subscript.cpp
ArielG-NV: This cannot be done since the order of `block->getChildren()` is not equivalent to the order of `next = inst->getNextInst();`

source/slang/slang-ir-legalize-image-subscript.cpp
kaizhangNV: I see.

source/slang/slang-ir-legalize-image-subscript.cpp
kaizhangNV: Does the order matter here?

source/slang/slang-ir-legalize-image-subscript.cpp
kaizhangNV: Well, we just don't want to see there is nothing prompted except a "segment fault" if any unexpected bad thing happens.

ArielG-NV: I can add assert's

ArielG-NV: Are these recommended checks for debug and release mode or only debug mode (developers)?

csyonghe: Debug check is fine.

source/slang/slang-ir-legalize-image-subscript.cpp
csyonghe: If you are modifying the indications while iterating them, you should call 
for (auto inst: getModifiableChildren()).

All other iterate-while-modify loops are not correct due to automatic inst hoisting.

For example, if inside a function you have:

X = call ...
y = makeStruct X, 1
z = call f y

Now when you iterate the code and during the first iteration you replaces X to 2.

That makes the y inst hoistable to global scope after the modification. And now the next iteration you set inst to y, but 
y = makeStruct 2,1
is now at global scope so you will be processing the insts at global scope from iteration 2 and all the remaining insts in the function are not processed.

The only safe way to modify insts in a loop is to first copy all children insts into a list, and then work through that list. This is what getModifiableChildren does.

csyonghe (approved):

jkwak-work:
It seems like `tests/bugs/buffer-swizzle-store.slang` is still failing for Metal.
See, https://github.com/shader-slang/slang/pull/4494

jkwak-work:
> It seems like `tests/bugs/buffer-swizzle-store.slang` is still failing for Metal. See, #4494

It seems like the test was missing a keyword `-compute`.
I am running the test again with the fix.

PR #4412, opened 2024-06-18
Add capture logic to other interfaces.

kaizhangNV:
Add capture logics to ISession, IModule, ITypeConformance, IEntryPoint ICompositeComponentType.

Add few encode methods to encode the array.

Fix some capture logic in global session. We will not need to encode the null_ptr output before actual actual of the captured APIs. Previously we used this as a place holder, but it's actually not necessary, it can be detected if the output is not captured during replay.

Dump the shader files to disk.

Also set a default directory for the capture files.


source/slang-capture-replay/slang-filesystem.cpp
```diff
@@ -27,12 +32,31 @@ namespace SlangCapture
...
         slangCaptureLog(LogLevel::Verbose, "%p: %s, :%s\n", m_actualFileSystem.get(), __PRETTY_FUNCTION__, path);
         SlangResult res = m_actualFileSystem->loadFile(path, outBlob);
+
+        // Since the loadFile method could be implemented by client, we can't guarantee the result is always as expected,
+        // we will check every thing to make sure we won't crash at writing file.
+        //
+        // We can only dump the file content after this 'loadFile' call, no matter this call crashes or file is not
+        // found, we can't save the file anyway, so we don't need to pay special care to the crash recovery. We will
+        // know something wrong with the loadFile call if we can't find the file in the capture directory.
+        if ((res == SLANG_OK) && (*outBlob != nullptr) && ((*outBlob)->getBufferSize() != 0))
```
csyonghe, line 50: If the inner load file call failed, we should also record this and make sure we return the same error code during replay.

csyonghe (approved):

Issue #4405, opened 2024-06-17
Metal: `Load` only works with `vector<T,4>` typed textures


ArielG-NV:
Affected test: `tests\bugs\buffer-swizzle-store.slang`
```
RWTexture2D<float2> g_test;
RWStructuredBuffer<float> outputBuffer;
void computeMain( uint2 dispatchThreadID : SV_DispatchThreadID )
{
    g_test[dispatchThreadID].xy = float2(0.0, 1.0);
    outputBuffer[dispatchThreadID.x] = g_test[dispatchThreadID].y;
}
```
compiles approximately into:
```
[[kernel]] void computeMain(uint3 dispatchThreadID_0 [[thread_position_in_grid]], texture2d<float, access::read_write> g_test_1 [[texture(0)]], float device* outputBuffer_1 [[buffer(0)]])
{
...
    float2 _S5 = (((&kernelContext_0)->g_test_0).read(vec<uint,2>(((int2(dispatchThreadID_1))).xy)));
```
This code is not going to compile since `.read`  returns a `vector<T,4>` in all cases

ArielG-NV:
related to [#4291](https://github.com/shader-slang/slang/issues/4291)

ArielG-NV:
This problem also seems to affect `sample`

PR #4417, opened 2024-06-18
Metal: Implement fix for non vector4 texture Load/Sample

ArielG-NV:
fixes: #4405

The change modifies stdlib to ensures all metal texture `read`/`sample` uses swizzle their `vector<T,4>` output into the same parent element type of __TextureImpl<T>

1. Added test cases to texture.slang to cover all types.


ArielG-NV:
Note: this PR relies upon changes from #4408
(made a separate PR since this PR is very simple despite changing lots of code)

csyonghe (approved):

PR #4418, opened 2024-06-18
Ray tracing validation

ccummingsNV:
Implemented support for ray tracing validation (https://developer.nvidia.com/blog/ray-tracing-validation-at-the-driver-level/) in Vulkan. Can be implemented for D3D if/when NVApi is accessible from slang.

To support it I pushed the Vulkan submodule up to the minimum required version for RT validation (1.3.280). I figure if we want to go to the very latest that should be a separate commit.


slang-gfx.h
```diff
@@ -2266,6 +2266,8 @@ class IDevice : public ISlangUnknown
         ShaderCacheDesc shaderCache = {};
         // Configurations for Slang compiler.
         SlangDesc slang = {};
+        /// Whether to enable ray tracing validation (currently only Vulkan - D3D requires app layer to use NVAPI)
```
csyonghe, line 2269: For binary compatibility, all new fields must be added as an extendedDesc instead of directly inside the struct.

ccummingsNV: Hi Yong - thanks - didn't see that mechanism for descriptors. Latest commit matches the pattern used for the D3D 12 device extended descriptor.

PR #4425, opened 2024-06-18
Disambiguate int type matrix multiply

### LINKED ISSUE #4414: Post-op multiplication results in ambiguous call
The shader below is generating a baffling error message for a post-op multiplication expression:
```
error 39999: ambiguous call to 'mul' with arguments of type (matrix<int,2,2>, vector<int,2>)
  b.zw = mul(a, b.yx);
            ^
hlsl.meta.slang(10850): note 39999: candidate: func mul<int, 2, 2>(int2x2, int2) -> int2
hlsl.meta.slang(10826): note 39999: candidate: func mul<int, 2, 2>(int2x2, int2) -> int2
```

As far as I can tell this is a `int2x2 * int2` which should work?

**CMD**
```
slangc.exe -target spirv -lang slang -D__spirv__ -emit-spirv-directly -profile ps_6_0 -entry main shader.slang
```

**Shader**
```
void main() {
  int2x2 a;
  int4 b;

  b.zw = mul(a, b.yx);
}
```

jkwak-work:
Closes #4414

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -10877,6 +10877,7 @@ vector<T,N> mul(matrix<T,N,M> left, vector<T,M> right)
 }
 __generic<T : __BuiltinLogicalType, let N : int, let M : int>
 [__readNone]
+[OverloadRank(-1)]
 [require(cpp_cuda_glsl_hlsl_metal_spirv, sm_4_0_version)]
 vector<T,N> mul(matrix<T,N,M> left, vector<T,M> right)
 {
--- a/tests/bugs/gh-4414.slang
+++ b/tests/bugs/gh-4414.slang
@@ -1,0 +1,10 @@
+//TEST:SIMPLE(filecheck=CHK):-target hlsl -entry main -stage vertex
+
+void main()
+{
+  int2x2 a;
+  int4 b;
+
+  //CHK-NOT:error 39999: ambiguous call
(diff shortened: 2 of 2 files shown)


csyonghe (approved):
The change here is fine, but I wonder why we couldn't disambiguate before the change. Is BuiltinIntegerTyoe inheriting from BuiltinLogicalType? If so it should always be able to pick the more constrained overload.


For example, if I have
interface IBase;
interface IDerived:IBase;

struct B: IBase;
struct D : IDerived;

void f<T:IBase>(T x);

void f<T:IDerived>(T x);

Then we should always pick the second overload for

f(D());

Because converting D to IDerived is cheaper than converting D to IBase.

jkwak-work:
I think this is because `int` implements both `__BuiltinIntegerType` and `__BuiltinLogicalType` at the same time.
```
struct int
    : __BuiltinType

#line 461 "core.meta.slang"
    ,  __BuiltinSignedArithmeticType
#line 470 "core.meta.slang"
    ,  __BuiltinArithmeticType
    ,  __BuiltinIntegerType
#line 476 "core.meta.slang"
    ,  __BuiltinLogicalType
#line 484 "core.meta.slang"
{
   ...omit...
}
```
I think their costs are same in this case.

csyonghe:
We can probably delete that explicit inheritance line for int types, and make sure biultinIntegerType inherits from BulltinLogicalType.

Issue #4415, opened 2024-06-18
Allow casting of 4-component vectors to 2x2 matrices and vice versa


chaoticbob:
This is for on-ramping. This is a pretty common trick to avoid having to pack and unpack a 4-component vectors to 2x2 matrices and vice versa. slang currently generates a very long error for this.

**Shader**
```
float4 main(float4 inputF : A, int4 inputI : B, uint4 inputU : C) : SV_Target {
  float2x2 matF1 = (             float2x2)inputF;
  float2x2 matF2 = (row_major    float2x2)inputF;
  float2x2 matF3 = (column_major float2x2)inputF;
  float4 f1 = (float4)matF1;
  float4 f2 = (float4)matF2;
  float4 f3 = (float4)matF3;
  
  int2x2 matI1 = (             int2x2)inputI;
  int2x2 matI2 = (row_major    int2x2)inputI;
  int2x2 matI3 = (column_major int2x2)inputI;
  int4 i1 = (int4)matI1;
  int4 i2 = (int4)matI2;
  int4 i3 = (int4)matI3;
  
  uint2x2 matU1 = (             uint2x2)inputU;
  uint2x2 matU2 = (row_major    uint2x2)inputU;
  uint2x2 matU3 = (column_major uint2x2)inputU;
  uint4 u1 = (uint4)matU1;
  uint4 u2 = (uint4)matU2;
  uint4 u3 = (uint4)matU3;

  return float4(f1.x, f2.x, f3.x, 0) +
         (float4)int4(i1.x, i2.x, i3.x, 0) +
         (float4)uint4(u1.x, u2.x, u3.x, 0);
}
```

PR #4432, opened 2024-06-19
Added float2x2 cast to float4 (and vise versa)

ArielG-NV:
fixes: #4415

Note: not adding `matrix<1,N>` and `matrix<N,1>` translations since this will bloat stdlib with many expensive to process `extension` operations. `matrix<N,1>` is already planned to be treated as vectors (which should solve this problem by proxy).


source/slang/core.meta.slang
```diff
@@ -1210,6 +1210,31 @@ extension matrix<int16_t,R,C,L>
...
+
+__generic<T>
+extension vector<T, 4>
+{
+    __generic<let L : int>
+    __implicit_conversion($(kConversionCost_MatrixToVector))
+    [__unsafeForceInlineEarly]
+    __init(matrix<T, 2, 2, L> value)
+    {
+        this.xy = value[0];
```
csyonghe, line 1222: Instead of doing this, which lowers to swizzledSet, just lower it to makeVector(elementExtract, elementExtract)

source/slang/slang-ast-support-types.h
```diff
@@ -167,6 +167,10 @@ namespace Slang
         kConversionCost_ScalarToMatrix = 10,
         kConversionCost_ScalarIntegerToFloatMatrix = kConversionCost_IntegerToFloatConversion + kConversionCost_ScalarToMatrix,
 
+        // Additional conversion cost to add when casting matrix to vector and vise versa
+        kConversionCost_MatrixToVector = 1,
```
csyonghe, line 171: This cost seems too small.

ArielG-NV: I will set it to 3:
1. more expensive than kConversionCost_OneVectorToScalar (2)
2. cheaper than a transpose on registers by a large amount (5)

csyonghe:
I really don't think we should allow implicit cast from float2x2 to float4 because they are fundamentally different types. This is not the same as 1-vector and scalar.

csyonghe:
Why do we need this?

ArielG-NV:
> Why do we need this?

https://github.com/shader-slang/slang/issues/4415

csyonghe:
I don't think we should provide everything HLSL has. And I think this is one of the bad things we shouldn't be adding to our language.

csyonghe:
We can add an explicit conversion between float2x2 and float4, but we shouldn't make them implicit.

csyonghe (approved):

Issue #4428, opened 2024-06-19
`tbuffer` or `TextureBuffer` use causes silent failure due to layout rules 


chaoticbob:
The shader below results in `NULL` pointer access since the rules for texture buffer objects are not implemented:
```
// slang-type-layout.cpp:
LayoutRulesImpl* GLSLLayoutRulesFamilyImpl::getTextureBufferRules()
{
    return nullptr;
}
```
Call chain starts in `slang-type-layout.cpp:createParameterGroupTypeLayout()`.

This is causing a silent failure, should return an error message at the very least if not implemented.

**Shader**
```
tbuffer MyTBuffer {
    float4 tb_val;
}

float4 main() : SV_Target {
    return tb_val;
}
```

ArielG-NV:
I suppose there are 2 problems to address here:
1. An error should be more clear here (`Error: unimplemented layout rules for object XYZ`)
2. add layout rules to `tbuffer`

chaoticbob:
I agree. Apologies, I was going for the least amount of work needed to address the current behavior. But you're right, would be good to have the functionality.

chaoticbob:
This also applies to `TextureBuffer` (note: it's much less used based on memory from previous experience).

ArielG-NV:
a seperate issue was made to implement spirv and metal (significant changes required): #4435

hlsl and glsl should both work with the linked PR.

Issue #4435, opened 2024-06-19
SPIRV Emit: `TextureBufferType` support


ArielG-NV:
`TextureBufferType` currently has no way to emit with the SPIRV backend.

This should be added.

csyonghe:
How does HLSL/dxc handle texture buffer types that exceeds 4 floats?

csyonghe:
Looking at the HLSL spec, it seems a tbuffer is treated as a u32 format texture, and loads from the buffer is lowered into texture fetches + bit casts. Seems like a lowering pass is in order to support this on spirv and metal. I doubt it offers any actual benefit over the use of a StructuredBuffer.

ArielG-NV:
> I doubt it offers any actual benefit over the use of a StructuredBuffer.

tbuffer predates `StructuredBuffer`, likely only kept around for legacy reasons

ArielG-NV:
> How does HLSL/dxc handle texture buffer types that exceeds 4 floats?

likely as you said here: "loads from the buffer is lowered into texture fetches + bit casts"

the logic should be reasonably simple since Slang knows the offset of every field, therefore meaning all accesses are fairly obvious in byte offset.

jkwak-work:
I never heard of `tbuffer` until today.
It appears that `tbuffer` was introduced by SM 4.0 when `cbuffer` was introduced.
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-constants?utm_source=chatgpt.com

This is super old legacy syntax.
When `tbuffer` was introduced, `cbuffer` was recommended over `tbuffer`.
Now `ConstantBuffer` is recommended over `cbuffer` so that makes `tbuffer` almost "not-recommened" level.

I don't know about the difficulty of supporting this for Metal, but I disagree that this is a "good" for slang newbie dev.
I am going to remove the "GoodFirstBug" label.

jhelferty-nv:
Deprioritizing this right now since tbuffer is a rarely-used feature, and may be deprecated at some point.

PR #4436, opened 2024-06-19
Implementing `tbuffer` layout(s)

ArielG-NV:
Fixes: #4428

This means `tbuffer` now works for hlsl and glsl targets.
SPIR-V and Metal has yet to implement logic for `TextureBufferType` emitting/legalization, separate issue to track this (#4435).

Changes:
1. Add resource layouts for 'TextureBuffer'.
2. Add on to existing logic a way to allocate appropriate registers for TextureBufferType (this was made to work with parameter block logic).
3. fix some emit logic of `TextureBufferType` for metal target so users get a useful error that points to #4435


csyonghe (approved):
Does HLSL allow trailing unsized arrays in a tbuffer?

ArielG-NV:
> Does HLSL allow trailing unsized arrays in a tbuffer?

no, tbuffer does not.

Issue #4440, opened 2024-06-20
SPIR-V: `SubpassInput` should default to `float4` if explicit type parameter is not provided


chaoticbob:
Slang currently expects an explicit type parameter for `SubpassInput`, this is contrary to the expected usage of `SubpassInput` from DXC (where it originated I'm guessing) and results in the following compile error:
```
shader.hlsl(3): error 30400: generic type '<DeclRef<GenericDecl>>' used without argument
SubpassInput gSubpassInputs[32];
```

According to [DXC's SPIR-V docs](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#subpass-inputs):
*In the above, T is a scalar or vector type. If omitted, it will defaults to float4.*

**Shader**
```
SubpassInput gSubpassInputs[32];

float4 main(uint index : A) : SV_TARGET{
	float4 v2 = gSubpassInputs[NonUniformResourceIndex(index)].SubpassLoad();
    return v2;
}
```

PR #4462, opened 2024-06-24
Implement HLSL resource bindings and default type `float4` to `SubpassInput<T>`

ArielG-NV:
fixes: [#4440](https://github.com/shader-slang/slang/issues/4440)

1. Removed `GLSLInputAttachmentIndexLayoutModifier` and the somewhat 'hacky' binding model 'input_attachment' previously relied upon. This was changed to work with the slang-type-layout rules system. This means we propagate resourceInfo rather than a modifier to store 'input_attachment' binding data. This change allows Slang automatic bindings, HLSL bindings, GLSL bindings, and translation of GLSL to and from HLSL bindings to work as expected.
2. Added default argument `float4` to SubpassInput<T>.
3. Merged glsl.meta and hlsl.meta SubpassInput logic.

Note: 
* To match how other bindings are validated (and used), validation for "`input_attachment_index` is only allowed to be used with `SubpassInput`" was removed. To validate requires full type value data. This is not ready by the point of checkModifiers/Attributes. If validation is desired, a separate check would need to be added for when adding parameter-bindings

### DIFF
--- a/.github/workflows/release-linux-glibc-2-17.yml
+++ b/.github/workflows/release-linux-glibc-2-17.yml
@@ -51,11 +51,11 @@ jobs:
             tar -czf ${SLANG_BINARY_ARCHIVE_TAR} bin/*/*/slangc bin/*/*/slangd bin/*/*/libslang.so bin/*/*/libslang-glslang.so bin/*/*/libgfx.so bin/*/*/libslang-llvm.so docs/*.md README.md LICENSE slang.h slang-com-helper.h slang-com-ptr.h slang-tag-version.h slang-gfx.h prelude/*.h
             echo "SLANG_BINARY_ARCHIVE=${SLANG_BINARY_ARCHIVE}" >> $GITHUB_OUTPUT
             echo "SLANG_BINARY_ARCHIVE_TAR=${SLANG_BINARY_ARCHIVE_TAR}" >> $GITHUB_OUTPUT
-      # - name: UploadBinary
-      #   uses: softprops/action-gh-release@v1
-      #   with:
-      #     files: |
-      #       ${{ steps.build.outputs.SLANG_BINARY_ARCHIVE }}
-      #       ${{ steps.build.outputs.SLANG_BINARY_ARCHIVE_TAR }}
-      #   env:
-      #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: UploadBinary
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            ${{ steps.build.outputs.SLANG_BINARY_ARCHIVE }}
+            ${{ steps.build.outputs.SLANG_BINARY_ARCHIVE_TAR }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/docs/design/coding-conventions.md
+++ b/docs/design/coding-conventions.md
@@ -200,15 +200,16 @@ case JUST_A:  doA();        break;
 case JUST_B:         doB(); break;
 ```
 
-Don't put space between a control-flow keyword and a following `(`.
+Put space between a control-flow keyword and a following `(`.
+This is a default setting of Visual Studio 2019 and 2022.
 Examples of how to format the major C++ control-flow constructs follow:
 
 ```c++
 void example()
 {
-    for(int ii = 0; ii < N; ++ii)
+    for (int ii = 0; ii < N; ++ii)
     {
-        if(ii == 0)
+        if (ii == 0)
         {
         }
         else
@@ -217,7 +218,7 @@ void example()
     }
 
     int x = 0;
-    while(x < 100)
+    while (x < 100)
     {
         x++;
     }
@@ -225,7 +226,7 @@ void example()
     int mode = 0;
     do
     {
-        switch(mode)
+        switch (mode)
         {
         case 0:
             x /= 2;
@@ -240,7 +241,7 @@ void example()
             mode--;
             break;
         }
-    } while(x > 0);
+    } while (x > 0);
 }
 ```
 
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -1402,15 +1402,7 @@ __generic<let N:int>
 [require(cpp_cuda_glsl_hlsl_spirv, shader5_sm_5_0)]
 public vector<uint,N> bitCount(vector<uint,N> value)
 {
-    __target_switch
-    {
-    case glsl: __intrinsic_asm "bitCount";
-    case spirv: return spirv_asm {
-        result:$$vector<uint,N> = OpBitCount $value
-    };
-    default:
-        VECTOR_MAP_UNARY(uint, N, countbits, value);
-    }
+    return countbits(value);
 }
 
 [__readNone] 
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -6917,6 +6917,7 @@ vector<T,N> cospi(vector<T,N> x)
 
 // Population count
 [__readNone]
+[ForceInline]
 [require(cpp_cuda_glsl_hlsl_metal_spirv, shader5_sm_5_0)]
 uint countbits(uint value)
 {
@@ -6933,6 +6934,27 @@ uint countbits(uint value)
         __intrinsic_asm "$P_countbits($0)";
     case spirv:
         return spirv_asm {OpBitCount $$uint result $value};
+    }
+}
+
+__generic <let N : int>
+[__readNone]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_metal_spirv, shader5_sm_5_0)]
+vector<uint, N> countbits(vector<uint, N> value)
+{
+    __target_switch
+    {
+    case hlsl:
+        __intrinsic_asm "countbits";
+    case glsl:
+        __intrinsic_asm "bitCount";
+    case metal:
+        __intrinsic_asm "popcount";
+    case spirv:
+        return spirv_asm {OpBitCount $$vector<uint, N> result $value};
+    default:
+        VECTOR_MAP_UNARY(uint, N, countbits, value);
     }
 }
 
--- a/tests/hlsl-intrinsic/countbits.slang
+++ b/tests/hlsl-intrinsic/countbits.slang
@@ -1,0 +1,27 @@
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -compute -cpu
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -compute -dx11
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -compute -dx12
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -vk -compute
+//DISABLE_TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -cuda -compute
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -mtl -compute
+
+//CHK:1
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+    uint r1 = countbits(0b1);
+    uint2 r2 = countbits(uint2(0b0, 0b1));
+    uint3 r3 = countbits(uint3(0b0, 0b1, 0b11));
+    uint4 r4 = countbits(uint4(0b0, 0b1, 0b11, 0b111));
+
+    outputBuffer[0] = true
+        && (r1 == 1)
+        && (r2.x == 0 && r2.y == 1)
+        && (r3.x == 0 && r3.y == 1 && r3.z == 2)
+        && (r4.x == 0 && r4.y == 1 && r4.z == 2 && r4.w == 3)
+        ;
+}


tests/glsl-intrinsic/subpass-input/subpass-input-hlsl.slang
```diff
@@ -0,0 +1,32 @@
...
+RWTexture2D<float4> t;
+
+layout (location = 0) out vec4 outColor;
+
+void main() {
+// CHECK_SPV-COUNT-2: OpImageRead
+// CHECK_GLSL-COUNT-2: subpassLoad
+// CHECK_HLSL-COUNT-2: SubpassLoad
+    outColor = vec4(true
+        && subpassHLSL1.SubpassLoad() == vec4(1)
```
kaizhangNV, line 28: please stop using && in all of your test code, because I'm afraid that the short-circuit could bypass something, though I know that the short-circuit is not a default behavior. But in case some day to enable it by default, and people forgot about changing the test.

You can use "&" instead.

tests/glsl-intrinsic/subpass-input/subpass-input-hlsl.slang
```diff
@@ -0,0 +1,32 @@
...
+
+RWTexture2D<float4> t;
+
+layout (location = 0) out vec4 outColor;
+
+void main() {
+// CHECK_SPV-COUNT-2: OpImageRead
+// CHECK_GLSL-COUNT-2: subpassLoad
+// CHECK_HLSL-COUNT-2: SubpassLoad
+    outColor = vec4(true
```
kaizhangNV, line 27: where is the check  for outColor value?

tests/glsl-intrinsic/subpass-input/subpass-input-hlsl.slang
ArielG-NV: Booleans are 32 bit in GLSL/HLSL/Metal/Other.
GLSL says the result of a boolean expression is 0 or 1, but I don't think this is ensured for other targets.
`0b10` is true, `0b01` is true, `0b10 & 0b01` is false, and `0b10 && 0b01` is true.

I am confused on how short-circuit would break the code, if the code returns false (short-circuits), then the test will fail as it should?

tests/glsl-intrinsic/subpass-input/subpass-input-hlsl.slang
ArielG-NV: This is a compile test that is to ensure SubpassInput compiles correctly.

Separately, There is no known outColor value, since slang-test backend does not support SubpassImage, we would have no known value to test against.

kaizhangNV: So you don't care about the result?

ArielG-NV: yes, we don't currently care about the result

tests/glsl-intrinsic/subpass-input/subpass-input-hlsl.slang
kaizhangNV: The whole point is that I want to remind you that if you want each statement to be executed,
&& is not gonna ensure that.

I just want to remind you to be careful with "&&" in your test code.

If you worry about boolean representation issue, you can also use if..else or even !!(expression).

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,87 @@ static void addExplicitParameterBindings_GLSL(
     // the index/offset/etc.
     //
 
-    TypeLayout::ResourceInfo* resInfo = nullptr;
-    TypeLayout::ResourceInfo* foundResInfo = nullptr;
-
-    LayoutSemanticInfo semanticInfo;
-    semanticInfo.index = 0;
-    semanticInfo.space = 0;
+    const int kMaxResCount = 2;
```
kaizhangNV, line 1078: can you explain a little bit about meaning of those magic numbers?

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,87 @@ static void addExplicitParameterBindings_GLSL(
...
+    {
+        TypeLayout::ResourceInfo* resInfo = nullptr;
+        LayoutSemanticInfo semanticInfo;
+        ResAndSemanticInfo()
+        {
+            semanticInfo.index = 0;
+            semanticInfo.space = 0;
+        }
+    };
+    ResAndSemanticInfo info[kMaxResCount] = {};
```
kaizhangNV, line 1093: I don't fully understand here,
you defined a array with size of 2, but looks like you only use the first element because 
const int kResInfo = 0;

so why didn't you just define a variable instead of an array?

source/slang/slang-parameter-binding.cpp
ArielG-NV: `kMaxResCount` is the size of an array.
This was used since we don't have a Slang `array` object and we are not supposed to be using stdlib.
Other parts of Slang code use c arrays, so I follow the pattern here.

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,87 @@ static void addExplicitParameterBindings_GLSL(
...
+    const int kResInfo = 0;
 
-    LayoutSemanticInfo subpassSemanticInfo;
-    bool foundBinding = false;
-    bool foundSubpass = false;
+    TypeLayout::ResourceInfo* foundResInfo = nullptr;
+    struct ResAndSemanticInfo
+    {
+        TypeLayout::ResourceInfo* resInfo = nullptr;
+        LayoutSemanticInfo semanticInfo;
```
kaizhangNV, line 1086: You can just defined it as
LayoutSemanticInfo semanticInfo {};

so you don't need the constructor.

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,87 @@ static void addExplicitParameterBindings_GLSL(
...
+        ResAndSemanticInfo()
+        {
+            semanticInfo.index = 0;
+            semanticInfo.space = 0;
+        }
+    };
+    ResAndSemanticInfo info[kMaxResCount] = {};
+    
+    if ((foundResInfo = typeLayout->FindResourceInfo(LayoutResourceKind::InputAttachmentIndex)) != nullptr)
```
kaizhangNV, line 1095: Looks like you can remove "!= nullptr"

source/slang/slang-parameter-binding.cpp
ArielG-NV: 1 object may have 2 explicit bindings simultaneously.
`ResAndSemanticInfo` only stores the explicit binding information of 1 explicit binding.
`GLSLInputAttachmentIndexLayoutAttribute` may appear along side any other binding attribute. 
There is a minimum of 1 binding, max of 2 bindings, instead of allocating a List, an array on the stack is faster.

source/slang/slang-parameter-binding.cpp
ArielG-NV: The code before my changes does not zero `LayoutSemanticInfo::kind`, `LayoutSemanticInfo semanticInfo {};` would zero everything.

kaizhangNV: Well, by looking at previous code, I think they should zero `kind` as well, though the `kind` will be assigned anyway.

source/slang/slang-parameter-binding.cpp
kaizhangNV: Yes, I understand.
But my question is where are you actually using `info[kSubpassResInfo]`?

ArielG-NV: `info[kSubpassResInfo]` is used when generating parameter bindings
```
        for (int i = 0; i < kMaxResCount; i++)
        {
            auto* resInfoItem = info[i].resInfo;
            auto& semanticInfo = info[i].semanticInfo;
```

source/slang/slang-check-modifier.cpp
```diff
@@ -474,6 +474,17 @@ namespace Slang
             }
             overloadRankAttr->rank = int32_t(rank->getValue());
         }
+        else if (auto inputAttachmentIndexLayoutAttribute = as<GLSLInputAttachmentIndexLayoutAttribute>(attr))
+        {
+            if (attr->args.getCount() != 1)
```
cheneym2, line 479: https://github.com/shader-slang/slang/blob/master/docs/design/coding-conventions.md suggests no space between if and (

ArielG-NV: I genuinely don't know if we are following that style suggestion, it seems for the last 5(6?) years of code in this file the suggestion has never been followed.

Should I start following it...?

jkwak-work: I think we need to update the document for the spacing in the control flow.
Because I don't think we are doing it.

ArielG-NV: > I think we need to update the document for the spacing in the control flow.
Because I don't think we are doing it.

I agree, likely best since this part of the styling is not being (has never been?) followed.

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,91 @@ static void addExplicitParameterBindings_GLSL(
...
+        {
+            info[kSubpassResInfo].resInfo = foundResInfo;
+            // Subpass fills semantic info of a descriptor and subpass
+            info[kSubpassResInfo].semanticInfo.index = (UInt)glslAttachmentIndexAttr->location;
+            info[kSubpassResInfo].semanticInfo.space = 0;
+        }
+    }
 
-    if( (foundResInfo = typeLayout->FindResourceInfo(LayoutResourceKind::DescriptorTableSlot)) != nullptr )
+    if(auto foundDescriptorTableSlot = typeLayout->FindResourceInfo(LayoutResourceKind::DescriptorTableSlot))
```
kaizhangNV, line 1108: Why can't this just be
`if(foundResInfo = typeLayout->FindResourceInfo(LayoutResourceKind::DescriptorTableSlot))` ?

ArielG-NV: because MSVC decided to make it a warning/error...

source/slang/slang-emit-c-like.h
```diff
@@ -463,7 +471,7 @@ class CLikeSourceEmitter: public SourceEmitterBase
 
     virtual void emitPostDeclarationAttributesForType(IRInst* type) { SLANG_UNUSED(type); }
     virtual bool doesTargetSupportPtrTypes() { return false; }
-    virtual void emitLayoutSemanticsImpl(IRInst* inst, char const* uniformSemanticSpelling = "register") { SLANG_UNUSED(inst); SLANG_UNUSED(uniformSemanticSpelling); }
+    virtual void emitLayoutSemanticsImpl(IRInst* inst, char const* uniformSemanticSpelling = "register", EmitLayoutSemanticOption layoutSemanticOption = EmitLayoutSemanticOption::kPostType) { SLANG_UNUSED(inst); SLANG_UNUSED(uniformSemanticSpelling); SLANG_UNUSED(layoutSemanticOption); }
```
jkwak-work, line 474: Can we avoid using the default value especially when the function is a virtual function.
Because the behavior can be often unexpected and I have seen a few hard bugs stemming from this.

source/slang/slang-emit-hlsl.cpp
```diff
@@ -93,6 +93,13 @@ void HLSLSourceEmitter::_emitHLSLRegisterSemantic(LayoutResourceKind kind, EmitV
         break;
 
         case LayoutResourceKind::InputAttachmentIndex:
+        {
+            m_writer->emit("[[vk::input_attachment_index("); 
+            m_writer->emit(index);
+            m_writer->emit(")]]");
```
jkwak-work, line 99: This HLSL syntax will be recognized by DXC but not by FXC.
I wonder if there is a way to emit them only for DXC.

ArielG-NV: SubpassInput only works for dxc (fxc does not support sm_6_0+?), I think if this is added to the capability system it would be fine?

source/slang/slang-emit-c-like.h
ArielG-NV: Should I also change the old behavior of `char const* uniformSemanticSpelling = "register"` to no longer use the "register" default as well?

jkwak-work: I am not sure what "behavior" you are referring to.
I assume you can remove `= "register"` and change the rest of code to make them working.

source/slang/slang-check-modifier.cpp
jkwak-work: I created a PR for the document update.
https://github.com/shader-slang/slang/pull/4498

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,91 @@ static void addExplicitParameterBindings_GLSL(
...
 
-    TypeLayout::ResourceInfo* resInfo = nullptr;
-    TypeLayout::ResourceInfo* foundResInfo = nullptr;
-
-    LayoutSemanticInfo semanticInfo;
-    semanticInfo.index = 0;
-    semanticInfo.space = 0;
+    const int kMaxResCount = 2;
+    const int kSubpassResInfo = 1;
+    const int kResInfo = 0;
```
jkwak-work, line 1080: Just FYI, a common trick for this is using enum, which can also be used as an array size because they are compile time constant.
```
enum
{
  kResInfo,
  kSubpassResInfo,
  kMaxResCount
};
```

source/slang/slang-parameter-binding.cpp
```diff
@@ -1065,81 +1075,91 @@ static void addExplicitParameterBindings_GLSL(
...
+            semanticInfo.kind = kind;
+            
             addExplicitParameterBinding(context, parameterInfo, varDeclBase.getDecl(), semanticInfo, count);
-            if (foundSubpass)
-                addExplicitParameterBinding(context, parameterInfo, varDeclBase.getDecl(), subpassSemanticInfo, count);
+            setABinding = true;
         }
-        return;
+        if(setABinding)
+            return;
```
jkwak-work, line 1162: I am not sure if it matters but the behavior here is little different from the previous code.
When `resInfo` was found, it was always returning regardless of `varDecl` can be converted to `VarDeclBase`.
But your change will return only when both conditions are true.

jkwak-work (dismissed):
It looks good to me.

source/slang/slang-parameter-binding.cpp
ArielG-NV: I think it may matter for the language server-- I will change the logic to how it was before

source/slang/slang-check-modifier.cpp
csyonghe: I wouldn't mind if somebody just run clang format through the code base and we stick to whatever that format is from that point on.

source/slang/slang-parameter-binding.cpp
csyonghe: The warning can be suppressed with double parenthesis: if ((a=b))

source/slang/slang-emit-hlsl.cpp
csyonghe: Why are we generating HLSL from this? This is not supported for d3d, and is added to dxc to produce spirv. We can already produce spirv directly. There is unlikely a need to go from slang to HLSL then to spirv.

csyonghe (approved):

Issue #4463, opened 2024-06-24
Metal: Function return type incorrectly checked when assigning address space


ArielG-NV:
`getFuncResultAddrSpace` assumes the return type is only a pointer when determining a return type address space. This logic should account for pointer-like-types as well.

fixes: `tests\optimization\func-resource-result\func-resource-result-simple.slang`, part of #4291

PR #4465, opened 2024-06-24
Metal: function return type address space checks

ArielG-NV:
fixes: #4463
fixes the following not working test: `func-resource-result-simple.slang`, related to #4291

When getting the address space of a function return type, only `Ptr<T>` returns were accounted for. The code now checks the address space of pointer-like-types.


jkwak-work (approved):
Looks good to me.

PR #4473, opened 2024-06-25
Support atomic intrinsics for Metal

jkwak-work:
This commit adds a support for the atomic intrinsics in Metal.

Metal requires the first argument for the atomic functions to be an atomic data type. This implementation rely on the fact that we can do a C-style type casting from a regular data type to an atomic data type.

### DIFF
--- a/source/slang/slang-emit-metal.cpp
+++ b/source/slang/slang-emit-metal.cpp
@@ -253,6 +253,24 @@ bool MetalSourceEmitter::tryEmitInstStmtImpl(IRInst* inst)
     case kIROp_discard:
         m_writer->emit("discard_fragment();\n");
         return true;
+    case kIROp_MetalAtomicCast:
+    {
+        auto oldValName = getName(inst);
+        auto op0 = inst->getOperand(0);
+
+        m_writer->emit("atomic_");
+        emitType(op0->getDataType());
+        m_writer->emit(" ");
+        m_writer->emit(oldValName);
+        m_writer->emit(" = ");
+
+        m_writer->emit("((atomic_");
+        emitType(op0->getDataType());
+        m_writer->emit(")(");
+        emitOperand(op0, getInfo(EmitOp::General));
+        m_writer->emit("));\n");
+        return true;
+    }
     }
     return false;
 }
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -8596,9 +8596,62 @@ void GroupMemoryBarrierWithGroupSync()
 
 // Atomics
 
+__generic<T>
+__intrinsic_op($(kIROp_MetalAtomicCast))
+[require(metal)]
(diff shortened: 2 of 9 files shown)


source/slang/slang-emit-metal.cpp
```diff
@@ -253,6 +253,24 @@ bool MetalSourceEmitter::tryEmitInstStmtImpl(IRInst* inst)
     case kIROp_discard:
         m_writer->emit("discard_fragment();\n");
         return true;
+    case kIROp_MetalAtomicCast:
+    {
+        auto oldValName = getName(inst);
+        auto op0 = inst->getOperand(0);
+
+        m_writer->emit("atomic_");
+        emitType(op0->getDataType());
```
csyonghe, line 262: This can work now, but it may subject to breakage in the future if the type has other modifiers like constexpr etc.

A more correct and future proof way is to introduce a IRMetalAtomicType(baseType) inst, and form a PtrType(IRMetalAtomicType(IntType)) as the result type of a MetalAtomicCast inst, and implement the emit logic for IRMetalAtomicType properly in the emit logic.

csyonghe (approved):

csyonghe:
Another case that has not been implemented here but we will need is to handle lowering of InterlockedAdd(texture[loc], Val).

Since metal does not support obtaining a pointer from a texture handle, we will need to extend our legalizeImageSubscript pass to translate 
`InterlockedAdd(MetalAtomicCast(imageSubscript(texture,loc)), Val)` into

```
call ".atomic_add" texture loc val 
```

We can recognize this case by checking the user of MetalAtomicCast, if the user is a call, and the callee has a [KnownBuiltin("metal_texture_atomic")] decoration (we need to mark the intrinsics in stdlib as such), then we should apply that transform in leglaizeImageSubscript.

source/slang/slang-emit-metal.cpp
jkwak-work: That sounds good. I will try it tomorrow.

jkwak-work:
> Another case that has not been implemented here but we will need is to handle lowering of InterlockedAdd(texture[loc], Val).
> 
> Since metal does not support obtaining a pointer from a texture handle, we will need to extend our legalizeImageSubscript pass to translate `InterlockedAdd(MetalAtomicCast(imageSubscript(texture,loc)), Val)` into
> 
> ```
> call ".atomic_add" texture loc val 
> ```
> 
> We can recognize this case by checking the user of MetalAtomicCast, if the user is a call, and the callee has a [KnownBuiltin("metal_texture_atomic")] decoration (we need to mark the intrinsics in stdlib as such), then we should apply that transform in leglaizeImageSubscript.

I was having this exact problem today. Thanks for the information.

I will follow the instructions tomorrow and see how it goes.

csyonghe:
Ariel is also working on the legalizeImageSubscript pass recently so you may want to coordinate with him on this as well. But this PR can be merged as is without the texture support.

Issue #4468, opened 2024-06-24
Metal: `legalizeArrayReturnType` leads to missing address space for `Out` parameter


ArielG-NV:
breaks: `tests\bugs\gh-2959.slang`, #4291 

With Metal as a target, kIROp_MakeArray incorrectly results in:
```
void _S1(array<uint, int(2)> _S2)
{

#line 14
    (*_S2)[int(0)] = 0U;

#line 14
    (*_S2)[int(1)] = 1U;

#line 14
    return;
}
```

This should be of the form:
```
void _S1(SOME_ADDRESS_SPACE array<uint, int(2)>* _S2)
{

#line 14
    (*_S2)[int(0)] = 0U;

#line 14
    (*_S2)[int(1)] = 1U;

#line 14
    return;
}
```

This is a source of a compile error

Issue #2959, opened 2023-07-05
Initialization with an array


jsmall-nvidia:
Take the code

```
RWStructuredBuffer<float> outputBuffer;

static uint g_values[2] = { 0, 1};

[numthreads(4, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
	uint tid = dispatchThreadID.x;

	outputBuffer[tid] = g_values[tid];
}
```

Produces an error 

```
dxc: x.slang(4): error :  'g_values_0_init' declared as array of functions of type 'const uint ()'
dxc: note : uint  g_values_0_init[int(2)]()
```

The HLSL emitted is... 

```
uint  g_values_0_init[int(2)]()
{
    uint  _S1[int(2)] = { 0U, 1U };

    return _S1;
}
static uint  g_values_0[int(2)] = g_values_0_init();
```

Which should presumably be `uint[2] g_values_0_init()`.

natduca:
This is a footgun bug and we should fix this. Putting in this quarter if we can manage.

PR #4478, opened 2024-06-25
Remove returned-array-legalization pass for metal

ArielG-NV:
fixes: #4468
fixes: `tests\bugs\gh-2959.slang`, part of https://github.com/shader-slang/slang/issues/4291

problem: 
`legalizeArrayReturnType` does not assign an address space to an expected pointer argument of an automatically generated/modified function.

solution:
Remove returned-array-legalization pass for metal, then we won't generate a function that requires legalization.


source/slang/slang-ir-legalize-array-return-type.cpp
```diff
@@ -89,8 +91,11 @@ void legalizeArrayReturnType(IRModule* module)
             if (func->getResultType()->getOp() == kIROp_ArrayType)
             {
                 makeFuncReturnViaOutParam(builder, func);
+                functionsToProcess.add(func);
             }
         }
     }
+    if(isMetalTarget(targetRequest))
+        specializeAddressSpace(module, functionsToProcess);
```
jkwak-work, line 99: I am not sure if this is a best place to do this.

The intention is to solve a problem for the array return type by applying the specialize-address-space.
But this doesn't look well organized and more like a hack.

Is it possible to solve the array return type in the normal flow where the `specializeAddressSpace()` function will be called?

ArielG-NV: > Is it possible to solve the array return type in the normal flow where the specializeAddressSpace() function will be called?

From what I understand the current additional logic is the only non-hack solution without disjointing metal and non metal logic.
`moveGlobalVarInitializationToEntryPoints()` and  `introduceExplicitGlobalContext()` relies upon legalization of `legalizeIRForMetal(irModule, sink);`. `legalizeArrayReturnType()` relies upon legalization of `moveGlobalVarInitializationToEntryPoints()` and `introduceExplicitGlobalContext()`.

> The intention is to solve a problem for the array return type by applying the specialize-address-space.
But this doesn't look well organized and more like a hack.

The problem being solved is that every function which is automatically generated (or modifies parameter list of a function) requires to have their address space re-legalized. This is required since else emitting code may lack the address space information needed to emit Metal code.

`specializeAddressSpace(module);` can be called instead of `specializeAddressSpace(module, functionsToProcess);` as a legalization pass, but this means we will run an IR pass over the entire program instead of just the functions which need re-legalization.

jkwak-work: The problem might be on the fact that `IRFunc* specializeFunc(const FuncSpecializationKey& key)` handles only the function arguments not the return type.
```
for (auto param : specializedFunc->getParams())
```
But I don't think we can actually handle the return type like how the parameters are handled.
So I guess it is fine as how you did.

jkwak-work (approved):
Looks good to me.
Good job identifying where the fix should be.

source/slang/slang-ir-legalize-array-return-type.cpp
ArielG-NV: > The problem might be on the fact that IRFunc* specializeFunc(const FuncSpecializationKey& key) handles only the function arguments not the return type.

`kIROp_Return` and `kIROp_Call` both manage the return type. This works properly with pointers (and pointer-like types: #4465)

tests/metal/groupshared-threadlocal-same-parameter.slang
```diff
@@ -0,0 +1,53 @@
...
+}
+
+[numthreads(2, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+    AllMemoryBarrierWithGroupSync();
+    uint tid = dispatchThreadID.x;
+    if (tid == 0)
+    {
+        g_valuesReturned[tid] = maybeGroupSharedReturn(tid)[tid];
```
jkwak-work, line 34: This new test is failing,
```
metal 32023.35: /tmp/unknown-Hhojjc.metal(47): error :  assigning 'array<uint, int(2)> *' (aka 'array<unsigned int, int(2)> *') to 'threadgroup array<uint, int(2)> *' (aka 'threadgroup array<unsigned int, int(2)> *') changes address space of pointer
    (&kernelContext_1)->g_valuesReturned_0 = &g_valuesReturned_1;
                                             ^~~~~~~~~~~~~~~~~~~
metal 32023.35: /tmp/unknown-Hhojjc.metal(49): error :  assigning 'array<uint, int(2)> *' (aka 'array<unsigned int, int(2)> *') to 'threadgroup array<uint, int(2)> *' (aka 'threadgroup array<unsigned int, int(2)> *') changes address space of pointer
    (&kernelContext_1)->g_values_0 = &g_values_1;
                                     ^~~~~~~~~~~
2 errors generated.
```
The variables in KernelContext uses "threadgroup" whereas the output from the funciton uses "thread".

ArielG-NV: 👍 This is a bug in address space legalization that also needs to be fixed

csyonghe (commented):
Can we delay the legalizeArrayReturnType until after legalizeIRForMetal is done, so we don't need to explicitly call specializeAddressSpace from legalizeArrayReturnType?

The higher level question is, do we really need legalizeArrayReturnType for metal? We emit arrays as array<T> type in metal, which should be allowed as function return types.

ArielG-NV:
> The higher level question is, do we really need legalizeArrayReturnType for metal? 

We can remove the pass for metal.

csyonghe (approved):

PR #4482, opened 2024-06-26
Include examples in default build

### LINKED ISSUE #4383: cmake build does not build examples
Our cmake build seems still unfinished for building examples. Even though there are flags to enable them

```
option(SLANG_ENABLE_GFX "Enable gfx targets" ON)
option(SLANG_ENABLE_EXAMPLES "Enable example targets, requires SLANG_ENABLE_GFX" ON)
```
it still doesn't build the examples.

We need to enable the examples build to make it a default in our CI.

venkataram-nv:
Fixes #4383

Previous cmake configuration explicitly excludes the examples from the default build; now they are included into the default build.

Also configuring the CI so that it does not unnecessarily build the examples for all actions (but at least once per platform).


.github/workflows/ci.yml
```diff
@@ -97,6 +97,7 @@ jobs:
             cmake --preset default --fresh \
               -DSLANG_SLANG_LLVM_FLAVOR=FETCH_BINARY \
               -DSLANG_SLANG_LLVM_BINARY_URL=$(pwd)/build/dist-release/slang-slang-llvm.zip \
+              -DSLANG_ENABLE_EXAMPLES=OFF \
```
jkwak-work, line 100: I think we should build examples for ci.yml.
And skip it for windows-selfhosted.yml

Can you tell us which CI runs are building examples with your change?

venkataram-nv: The ones building examples now are (seems like all):
ci.yml (new),
compile-regression-test.yml,
falcor-compiler-perf-test.yml,
windows.yml,
linux.yml,
falcor-test.yml,
linux-arm64.yml

jkwak-work: We should be building the examples only once for each platform.
I am not sure what platforms our CI covers.
Can you figure out which CI setting runs for what platform and OS?
Something like

- ci.yml : Linux_x64_release, and ??
- compile-regression-test.yml : ???
- windows.yml : Windows_x64_release
- linux.yml : ??
- linux-arm64.yml : Linux_arm64_release

I don't think we want to build the examples for falcor-XXX tests.

venkataram-nv: ci.yml:
- Linux GCC x64 **Debug**

compile-regression-test.yml:
- Windows x64 Release

windows.yml:
- Windows x64 Release
- Windows x64 Debug
- Windows Win32 Release
- Windows Win32 Debug

linux.yml:
- Linux GCC x64 Release
- Linux Clang x64 Release
- Linux GCC x64 Debug
- Linux Clang x64 Debug

linux-arm.yml:
- Linux GCC aarch64 Release

jkwak-work: Yea... it looks like we should build the example for all "windows.yml", "linux.yml" and "linux-arm.yml".
The rest of CI should skip because it will be redundant.

jkwak-work (approved):
Looks good to me.

jkwak-work:
@venkataram-nv , is this ready for merge?
If yes, please convert the PR from draft to non-draft, so that we can merge.

venkataram-nv:
> @venkataram-nv , is this ready for merge? If yes, please convert the PR from draft to non-draft, so that we can merge.

Yes, it should be good to go now.

PR #4485, opened 2024-06-26
Fix Texture2DMSArray

### LINKED ISSUE #4427: Missing support for `Texture2DMSArray`
Slang currently does not support `Texture2DMSArray` but does support `Texture2DMS`. Please add support for `Texture2DMSArray`.

```
Texture2DMS      tex;
Texture2DMSArray mstex;

float4 main()
{
    return tex.Load(int2(0, 0), 0);
    return mstex.Load(int3(0, 0, 0), 0);
}
```

jkwak-work:
Close #4427

We had the postfix order wrong for the keyword MS. This commit changes the incorrect name Texture2DArrayMS to Texture2DMSArray.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -1636,8 +1636,6 @@ public typealias sampler2DMS = Sampler2DMS<float4>;
 public typealias isampler2DMS = Sampler2DMS<int4>;
 public typealias usampler2DMS = Sampler2DMS<uint4>;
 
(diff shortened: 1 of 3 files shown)


ArielG-NV (commented):
`tests\compute\texture-subscript-multisample.slang` uses `RWTexture2DArrayMS` instead of `RWTexture2DMSArray`, I believe this should be changed as well.

ArielG-NV (approved):
The Implementation code looks good

Issue #4489, opened 2024-06-26
tests/cpu-program/gfx-smoke.slang fails on local test


jkwak-work:
**Problem description**
One of the slang-test tests started failing as a regression.
```
tests/cpu-program/gfx-smoke.slang
```
This test is disabled on CI but when you run it on a local machine, it fails.
The regression started happening from the following commit,
```
b7e82434 Wed Jun 12 12:46:24 2024 -0400 Add slangc flag to `-zero-initialize` all variables (#3987) 
```
Note that there was a build break that causes the specific commit to have a compile error.
You will need to cherry-pick the right next commit.

**Repro step**
Checkout the commit that caused the regression and cherry-pick to fix the compile error,
```
git checkout b7e82434
git cherry-pick 8bf7d11
build slang-test.exe
slang-test.exe tests/cpu-program/gfx-smoke.slang
```

When the issue is reproduced, the error looks like following,
```
tests/cpu-program/gfx-smoke.slang(7): note 99999: an internal error threw an exception while working on code near this location
    gfx.DeviceDesc deviceDesc = {};
                   ^~~~~~~~~~
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unexpected: InitializerListType
```

ArielG-NV:
```
    void SemanticsDeclBodyVisitor::visitAggTypeDecl(AggTypeDecl* aggTypeDecl)
	
	//...
        ensureDecl(m->getDefaultDeclRef(), DeclCheckState::DefaultConstructorReadyForUse); // crash is introduced indirectly from this line
	//...
```

ArielG-NV:
detailed explanation of the problem: https://github.com/shader-slang/slang/pull/4491#issuecomment-2194872824

Issue #3176, opened 2023-09-01
Investigate why `tests/cpu-program/gfx-smoke.slang` fails on Linux when using test driver.


csyonghe:

swoods-nv:
Moving to Q2-- this is lower priority than everything else testing-related being worked on at the moment.

swoods-nv:
Blocked by #3953

PR #4491, opened 2024-06-26
Nested initialization list in struct crash fix

ArielG-NV:
fixes: #4489 

Problem:
1. Previously, `TopStruct var = {}` would crash if a nested member also had a initExpr assigned to `{}`. This crash is due to a particular ordering of how `{}` resolves its own type blindly setting it as an invalid type as a place holder.

Solution:
1. only set the init list type when invalid (if we checkTerm twice on the same object, this won't break the code)

Note: removed the overhaul to default initialization since it breaks Falcor and likely needs more time dedicated to it, the code history will be kept in this branch (in the mean-time it is best to fix the current regression with a more simple fix).


source/slang/slang-check-conversion.cpp
```diff
@@ -512,8 +512,14 @@ namespace Slang
         UInt argCount = fromInitializerListExpr->args.getCount();
         UInt argIndex = 0;
 
-        // TODO: we should handle the special case of `{0}` as an initializer
-        // for arbitrary `struct` types here.
+        if (argCount == 0 && outToExpr && toType)
```
csyonghe, line 515: If !outToExpr, you still want to return true but not create the out expr.

source/slang/slang-check-impl.h
```diff
@@ -2870,4 +2870,47 @@ namespace Slang
 
     RefPtr<EntryPoint> findAndValidateEntryPoint(
         FrontEndEntryPointRequest* entryPointReq);
+
+
+    [[maybe_unused]] static ConstructorDecl* _getDefaultCtor(StructDecl* structDecl)
```
csyonghe, line 2875: Why maybe_unused is needed here?

ArielG-NV: slang-check-impl compiled to multiple compilation units, this function is only used across 2 compilation units

csyonghe (commented):
Why do we double check the expr in the first place? Can we stop doing that?

ArielG-NV:
> Why do we double check the expr in the first place?

This is not isolated behavior, If we have two `AssignExpr` that share the same `->right` member, we will run `checkTerm` on the `->right` member twice. This only breaks `visitInitializerListExpr` because `visitInitializerListExpr` is dependent on being called only once.

Why this crash happens now?

The behavior is showing up now due to minor ordering changes in AST building.
Previous Logic: 
1. Slang did not `ensureDecl(memberVariable, DeclCheckState::DefinitionChecked)` when checking a `structDecl`, we only auto generate `__init` for `structDecl`.
2. Naturally `ensureDecl()` on the auto generated `__init` for `structDecl`. This causes a check for `AssignExpr` (contains `initExpr`). This assigns place-holder invalid type to `initExpr` from call to `visitInitializerListExpr`
3. Slang naturally runs `ensureDecl(memberVariable, DeclCheckState::DefinitionChecked)`
4. Now we call `checkVarDeclCommon` to check the memberVariable. This runs `CheckTerm()` on our `initExpr` as well. This re-assigns place-holder invalid type from call to `visitInitializerListExpr`. We now run `coerce` to change the `initExpr` type to something valid.

New Logic:
1. Slang calls `ensureDecl(memberVariable, DeclCheckState::DefinitionChecked)` when checking a `structDecl`
2. Now we call `checkVarDeclCommon` to check the memberVariable. This runs `CheckTerm()` on our `initExpr`. This assigns place-holder invalid type from call to `visitInitializerListExpr`. We now run `coerce` to change the `initExpr` type to something valid.
3. we auto generate `__init` for `structDecl` now.
4. Naturally `ensureDecl()` on the auto generated `__init` for `structDecl`. This causes a check for `AssignExpr` (contains `initExpr`). This re-assigns place-holder invalid type to `initExpr` from call to `visitInitializerListExpr`

We were double checking the `initExpr` before. Code working was pure chance on how Slang was parsing `structDecl`s.

>  Can we stop doing that?

The solution is to not use `initExpr` directly ever (clone the AST node before using) since we will only have 1 instance of each `initExpr` in that case. This is a hack though, `visitInitializerListExpr` is fundamentally unstable with Slang AST.

csyonghe:
Shouldn't we exit immediately if we can determine the expr is already checked?

ArielG-NV:
> Shouldn't we exit immediately if we can determine the expr is already checked?

I don't know if this would be a problem (we may modify arg list)?

csyonghe:
An expr should never need to be checked twice. Once it is checked it should be in its final form and not be modified elsewhere.

kaizhangNV:
@ArielG-NV 
Does the gfx-smoke test pass?
If yes, can you enable the gfx-smoke test and re-run all the checks?

ArielG-NV:
> Does the gfx-smoke test pass?
If yes, can you enable the gfx-smoke test and re-run all the checks?

It does pass, I can (and will) enable the test.

kaizhangNV:
ok, can you please enable the test in this PR?

ArielG-NV:
> ok, can you please enable the test in this PR?

we never enabled gfx-smoke since `gfx` has problems with CI

kaizhangNV:
> > ok, can you please enable the test in this PR?
> 
> we never enabled gfx-smoke since `gfx` has problems with CI, here should be an issue open for this already.

Do you know what is the issue? And do we have issue tracking this?

ArielG-NV:
> Do you know what is the issue? And do we have issue tracking this?


https://github.com/shader-slang/slang/issues/3176

csyonghe (approved):

PR #4492, opened 2024-06-27
DRAFT: Implement embedded precompiled dxil slang-module feature

cheneym2:
When specified "-embed-dxil" creates an implicit DXIL target which is compiled
with DXC and stored in the slang-module as an IR operation.

When a slang-module is loaded, any precompiled binaries are loaded too
and are used to skip IR linking and optimization stages, instead passing
through to DXC for linkage.

### DIFF
--- a/source/slang/slang-ir.cpp
+++ b/source/slang/slang-ir.cpp
@@ -2059,7 +2059,7 @@ namespace Slang
 
     UnownedStringSlice IRConstant::getStringSlice()
     {
-        assert(getOp() == kIROp_StringLit);
+        assert(getOp() == kIROp_StringLit || getOp() == kIROp_BlobLit);
         // If the transitory decoration is set, then this is uses the transitoryStringVal for the text storage.
         // This is typically used when we are using a transitory IRInst held on the stack (such that it can be looked up in cached), 
         // that just points to a string elsewhere, and NOT the typical normal style, where the string is held after the instruction in memory.
@@ -2133,6 +2133,7 @@ namespace Slang
             {
                 return value.ptrVal == rhs->value.ptrVal;
             }
+            case kIROp_BlobLit:
             case kIROp_StringLit:
             {
                 return getStringSlice() == rhs->getStringSlice();
@@ -2174,6 +2175,7 @@ namespace Slang
             {
                 return combineHash(code, Slang::getHashCode(value.ptrVal));
             }
+            case kIROp_BlobLit:
             case kIROp_StringLit:
             {
                 const UnownedStringSlice slice = getStringSlice();
@@ -2264,6 +2266,7 @@ namespace Slang
                 irValue->value.ptrVal = keyInst.value.ptrVal;
                 break;
             }
+            case kIROp_BlobLit:
             case kIROp_StringLit:
             {
                 const UnownedStringSlice slice = keyInst.getStringSlice();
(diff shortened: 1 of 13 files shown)


source/slang/slang-ir.cpp
```diff
@@ -2387,6 +2390,29 @@ namespace Slang
...
+        memset(&stackDecoration, 0, sizeof(stackDecoration));
+        stackDecoration.m_op = kIROp_TransitoryDecoration;
+        stackDecoration.insertAtEnd(&keyInst);
+
+        keyInst.m_op = kIROp_BlobLit;
+        keyInst.typeUse.usedValue = nullptr; // not used
+
+        IRConstant::StringSliceValue& dstSlice = keyInst.value.transitoryStringVal;
+        dstSlice.chars = const_cast<char*>(inSlice.begin());
+        dstSlice.numChars = uint32_t(inSlice.getLength());
```
jkwak-work, line 2411: This is a question.
Is it necessary to create the temporary object, `inSlice`?
Can this be just like the following?
```
dstSlice.chars = (const char*)(blob->getBufferPointer());
dstSlice.numChars = blob->getBufferSize();
```

source/slang/slang-options.cpp
```diff
@@ -1912,6 +1934,8 @@ SlangResult OptionsParser::_parse(
                 linkage->m_optionSet.set(optionKind, compressionType);
                 break;
             }
+            case OptionKind::EmbedDXIL: SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::DXIL)); break;
+            case OptionKind::EmbedSPIRV: SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::SPIRV)); break;
```
jkwak-work, line 1938: I think it will be easier to read if these are broken into multiple lines,
```
    case OptionKind::EmbedDXIL:
        SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::DXIL));
        break;
    case OptionKind::EmbedSPIRV:
        SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::SPIRV));
        break;
```

source/slang/slang-serialize-ir.cpp
```diff
@@ -195,6 +195,14 @@ Result IRSerialWriter::write(IRModule* module, SerialSourceLocWriter* sourceLocW
                 switch (srcInst->getOp())
                 {
                     // Special handling for the ir const derived types
+                    case kIROp_BlobLit:
+                    {
+			// Blobs are serialized into string table like strings
```
jkwak-work, line 200: The indentation here is off, because it uses tab characters.
We use space characters for indentations.

jkwak-work (commented):
It looks good to me.
But I don't have any idea about the TU and TP problem.

jkwak-work:
@cheneym2 , if you look at the right side, there is a clickable text, "Convert to draft".
It can make the PR actually a draft that wouldn't be submitted by a mistake.

cheneym2:
> @cheneym2 , if you look at the right side, there is a clickable text, "Convert to draft". It can make the PR actually a draft that wouldn't be submitted by a mistake.



> @cheneym2 , if you look at the right side, there is a clickable text, "Convert to draft". It can make the PR actually a draft that wouldn't be submitted by a mistake.

Thanks, done.  In gitlab, starting the merge request with DRAFT actually makes it unmergeable, I thought maybe github worked like that too.

PR #4493, opened 2024-06-27
Add API for querying dependency files on IModule

skallweitNV:
This is required for shader hot reloading in our framework.

### DIFF
--- a/source/slang/slang-compiler.cpp
+++ b/source/slang/slang-compiler.cpp
@@ -2471,6 +2471,18 @@ namespace Slang
         return nullptr;
     }
 
+    SLANG_NO_THROW SlangInt32 SLANG_MCALL Module::getDependencyFileCount()
+    {
+        return (SlangInt32)getFileDependencies().getCount();
+    }
+
+    SLANG_NO_THROW char const* SLANG_MCALL Module::getDependencyFilePath(
+        SlangInt32 index)
+    {
+        SourceFile* sourceFile = getFileDependencies()[index];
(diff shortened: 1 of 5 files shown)


skallweitNV:
Is this change actually "non-breaking"? It just adds two functions to the `IModule` API, doesn't change any existing API.

source/slang/slang-compiler.cpp
```diff
@@ -2471,6 +2471,18 @@ namespace Slang
...
+    SLANG_NO_THROW SlangInt32 SLANG_MCALL Module::getDependencyFileCount()
+    {
+        return (SlangInt32)getFileDependencies().getCount();
+    }
+
+    SLANG_NO_THROW char const* SLANG_MCALL Module::getDependencyFilePath(
+        SlangInt32 index)
+    {
+        SourceFile* sourceFile = getFileDependencies()[index];
+        return sourceFile->getPathInfo().hasFoundPath() ? sourceFile->getPathInfo().foundPath.getBuffer() : "unknown";
```
jkwak-work, line 2483: I think there could be an actual file called "unknown" for a certain situation.
Probably it is better with an empty string, "".

jkwak-work (commented):
Looks good to me.
But I will let Yong to approve because it is an API change.

source/slang/slang-compiler.cpp
csyonghe: Yeah, it may be better to just return nullptr.

csyonghe (approved):

source/slang/slang-compiler.cpp
skallweitNV: Done.

PR #4494, opened 2024-06-27
DNI: testing all metal compute tests on CI

jkwak-work:
This PR is intented to run all CI tests for Metal compute test. It is CI testing only and not for an actual merge.


jkwak-work:
Closing because the issue #4291 is closed

PR #4496, opened 2024-06-27
switch from ubuntu-18 to centos-7 build

kaizhangNV:


jkwak-work:
Why are we switching it and what are the differences?
Is there an issue number for this?

kaizhangNV:
> Why are we switching it and what are the differences? Is there an issue number for this?

The libstdc++ version doesn't match. The issue is internal.

jkwak-work:
Shouldn't we be adding one more CI for the new OS instead of switching from one we have been using?
Do we have another CI coverage on Ubuntu-18?

Issue #4504, opened 2024-06-28
GLSL to Slang Questions


jkwak-work:
Creating an issue from a discussion.

### Discussed in https://github.com/shader-slang/slang/discussions/4454

<div type='discussions-op-text'>

<sup>Originally posted by **LouChiSoft** June 22, 2024</sup>
Hi, I'm converting some GLSL code to Slang (and them compiling it back out to GLSL for comparison). I have a very basic GLSL shader that looks like this:
```glsl
#[compute]
#version 450

layout(std430, binding = 0) buffer buffer_t {
    float _data[];
} data;

layout(local_size_x_id = 0, local_size_y_id = 1, local_size_z = 1) in;
void main()
{
    uint index = gl_GlobalInvocationID.x + (gl_WorkGroupSize.x * gl_GlobalInvocationID.y);
    data._data[index] = data._data[index] * 5;
    return;
}
```
Most of the conversion is straight forward but I have ran into two issues so far that I can't seem to find an answer on:
- Issue 1 is how do you define a specialisation constant in Slang? As you can see in the example above I use specialisation constants to decide the size of the workgroup at runtime, but so far I have only been able to get compile time sizes working in Slang.
- Issue 2 is that `WorkGroupSize()` doesn't appear to work. Whenever I try to use it I get `unimplemented feature in Slang compiler`. I assume that means that it doesn't work for the GLSL backend, is there an alternative?

Slang version for reference:
```C#
RWStructuredBuffer<float> data;

[shader("compute")]
[numthreads(5, 2, 1)]
void computeMain(uint3 threadId: SV_DispatchThreadID)
{
    int3 size = WorkgroupSize();
    uint index = threadId.x + (size.x * threadId.y);
    data[index] = data[index] * 5;
}
```</div>

ArielG-NV:
Note: problem is that `WorkgroupSize();` inst is not deleted properly, causes the kIROp to make it to emitting IR stage (causes compile to fail since unimplemented KIROp)

PR #4505, opened 2024-06-28
Fix `WorkgroupSize()` compiler failure

ArielG-NV:
fixes: #4504

Problem:
In some cases the globally hoisted `WorkgroupSize()` function when being legalized was not being destroyed after legalization. This meant the unimplemented kIROp causes errors later in a compile.

Solution:
Destroy the globally hoisted `WorkgroupSize()` Inst after legalization in all cases.

Note:
Removed capability requirement on `gl_WorkGroupSize` since no longer applies. kIROp of `WorkgroupSize()` is implemented for all targets.


jkwak-work (approved):
Looks good to me

Issue #4508, opened 2024-06-28
Capability System: `__target_switch` never compares `case:` capabilities to the parent `[require(...)]`


ArielG-NV:
Problem:
following code will compile but should at a minimum warn:
```
[require(glsl_hlsl)]
void myfunc()
{
    __target_switch
    {
    case hlsl:
    case spirv:
    }
}
```
Expected behavior:
error since `spirv` is not listed by the parent `[require(...)]`

Issue #4506, opened 2024-06-28
Capability System: `[require(....)]` never checks if the capabilities required are invalid


ArielG-NV:
Problem:
`[require(hlsl, glsl)]` is allowed despite being `Invalid`

Solution:
This should be an error.

PR #4490, opened 2024-06-26
Support status argument for GatherXXX

jkwak-work:
This commit adds an argument to all texture GatherXXX functions. The new argument is for "status" as described in SM5.0 definision.

Close #4466

PR #4510, opened 2024-06-28
Emit errors when invalid capabilities are used

ArielG-NV:
fixes: #4506
fixes: #4508

1. As per #4506, no longer allow capabilities to resolve into an `Invalid` capability when using `[require(...)]`
2. As per #4508, no longer allow mismatch between `case` and `[require(...)]` capabilities for a calling function.
3. Fixes to incorrect hlsl.meta capabilities. Related to #4490, `texture_gather` had incorrect capabilities causing issues.
4. Added a ref to the parent function/decl for when erroring with capabilities to help debug meta.slang files for when weird source locations are generated.

Note: 
This is not a breaking change. All capability errors can be turned off with the `ignore-capabilities` flag


source/slang/slang-check-decl.cpp
```diff
@@ -10178,20 +10183,22 @@ namespace Slang
...
+        typedef SemanticsDeclReferenceVisitor<CapabilityDeclReferenceVisitor<ProcessFunc, RefParentDiagnosticFunc>> Base;
 
         const ProcessFunc& handleReferenceFunc;
+        const RefParentDiagnosticFunc& handleReferenceParentDiagnosticFunc;
         RequireCapabilityAttribute* maybeRequireCapability;
         SemanticsContext& outerContext;
-        CapabilityDeclReferenceVisitor(const ProcessFunc& processFunc, RequireCapabilityAttribute* maybeRequireCapability, SemanticsContext& outer)
+        CapabilityDeclReferenceVisitor(const ProcessFunc& processFunc, const RefParentDiagnosticFunc& refParentDiagnosticFunc, RequireCapabilityAttribute* maybeRequireCapability, SemanticsContext& outer)
             : handleReferenceFunc(processFunc)
+            , handleReferenceParentDiagnosticFunc(refParentDiagnosticFunc)
```
jkwak-work, line 10198: I think the name is too long and it became harder to read.
I like to suggest to rename the existing variable "handleReferenceFunc" to just "processFunc" and the new one to "parentDiagnosticFunc".
The keywords, "Reference" and "handle", don't carry any meaning.

I also like to note that functors in C++ are normally copiable objects and passed around as a copy not as a reference.
The uses of `&` here seem unnecessary and it can be a potential problem later if the functor is originated from a stack memory and it is de-referenced after the stack became invalid.

jkwak-work (dismissed):
It looks good to me.

source/slang/slang-check-decl.cpp
ArielG-NV: I agree with renaming variables. 

> I also like to note that functors in C++ are normally copiable objects and passed around as a copy not as a reference.

A copy can be used here, I will switch the types around.
I think the reason we copy by reference here is because lambdas do not decay into a simple function pointer in most cases, lambdas can explode in size very quickly depending on the parameters captured.

jkwak-work (approved):
Looks good to me.

PR #4513, opened 2024-06-28
Set debug working directory for slangc.exe

jkwak-work:
This commit changes the working directory setting for slangc.exe in Visual Studio Debugging from "build" directory to the root of the repository.

It is often the case that the users debug slang shaders under "tests" directory, and the working directory needs to be manually changed to the root directory of the repository.


jkwak-work:
I am merging this change without a review approval.
The change is trivial enough and CI tests are all passing.

Issue #3855, opened 2024-03-28
(slangpy/autodiff) Incorrect handling of broadcasted PyTorch tensors


saipraveenb25:
PyTorch often creates broadcasted tensors during its backward pass to save on memory, even if the user does not use braodcasting in their code. An example: the backward derivative of `.sum()` is a broadcasted tensor. 

The broadcasted dimension has a stride of 0, causing our current offset calculation to map several indices to the same memory location.

This is fine for tensors used with `loadOnce()` or `storeOnce()` operations, but doesn't work for `load()`/`store()` or when using the `[]` operator because the first read of the derivative rewrites the location with 0, causing subsequent reads to have the wrong value.

We could force all tensors to be contiguous to resolve this issue, but as we saw previously, this can cause performance bottlenecks for applications that don't need to use contiguous tensors. 

An alternative, but more involved approach, is to split `DiffTensorView` into two variants: one that supports only `load()`/`store()` and the other supports only `loadOnce()`/`storeOnce()`, and check each of them separately. The former variant cannot have 0 strides (enforced by throwing errors in `make_tensor_view`). The latter variant can use 0-strides to retain the performance & memory benefits of broadcasting.

Additionally: the current example in the slangpy documentation contains this problem, so it needs to be updated asap to use `loadOnce()`/`storeOnce()`

Useful resource for understanding PyTorch stride quirks: https://martinlwx.github.io/en/how-to-reprensent-a-tensor-or-ndarray/

PR #4516, opened 2024-06-28
Error out when constructing tensor views from tensors with 0 stride.

saipraveenb25:
This avoids a problem with broadcasted tensors. Our tensor-view platform is designed to allow unrestricted access to tensor memory, while broadcasted tensors were designed for 'read-only' use-cases. Trying to write into a broadcasted tensor needs re-allocation, which Slang is not designed to do.

For now, we enforce contiguity on tensors with any 0 strides.

In the future, we will introduce a ConstTensorView object to allow such tensors to be used as an input.

This patch also propagates name-hint information through structs & arrays of tensors, to allow sensible names for the error messages (before this the error messages were temporary inst numbers, which is nearly impossible to debug)

Resolves a part of #3855 (will open a new issue for ConstTensorView if necessary)

This is a potentially breaking change because some cases where we were (incorrectly) allowing broadcasted tensors will now error out. The fix is typically quite simple: call `.contiguous()` on the tensor before passing it in.

### DIFF
--- a/prelude/slang-torch-prelude.h
+++ b/prelude/slang-torch-prelude.h
@@ -140,6 +140,9 @@ TensorView make_tensor_view(torch::Tensor val, const char* name, torch::ScalarTy
     for (int i = 0; i < val.dim(); ++i)
     {
         res.strides[i] = val.stride(i) * elementSize;
+        if (res.strides[i] == 0)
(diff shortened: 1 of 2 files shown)


jkwak-work (approved):
It looks good to me.

PR #4518, opened 2024-06-28
Resource searching for examples

### LINKED ISSUE #4314: Running examples like "triangle" fails to find corresponding "shader.slang"
Figured I'd post this here, since a new user will likely hit the same issue. 

When I run an example like "triangle.exe" within vscode (using the cmake build system), the current working directory for triangle.exe by default is "build/Debug/bin/", which is the same directory containing triangle.exe. This example expects "shader.slang" to be in the current working directory, but the actual shader is at "examples/triangle/shaders.slang"

An easy enough workaround is to just copy the shader.slang file into /build/Debug/bin, but a novice user might not know how to interpret the crash in the example. When the shader is not found, the app kinda just silently fails without any useful error messages being printed out to the terminal.

venkataram-nv:
Fixes #4314 

Extending examples so that they search for the resources needed in directories above CWD.

Now copies of the example resources are put into build for all platforms (before used to be only Windows).

Users should now be able to run the programs anywhere within the build directory (even from CLI).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -704,10 +704,7 @@ if (SLANG_ENABLE_EXAMPLES AND SLANG_ENABLE_GFX)
     )
     set_target_properties(all-examples PROPERTIES FOLDER examples)
     function(example dir)
-        set(debug_dir ${dir})
-        if (MSVC)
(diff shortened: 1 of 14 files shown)


jkwak-work (approved):
Looks good to me.

PR #4522, opened 2024-07-01
Supply SPIRV capability for textureQueryLod

### LINKED ISSUE #4357: textureQueryLod needs to be used with ImageQuery capability
**Problem Description**
A regression from CTS test is found when `textureQueryLod` is used.
The tests pass but they print a validation error.

**Goal**
We should use the function with a capability, "ImageQuery", before using the instruction, `OpImageQueryLod`.

**Repro step**
The following vertex shader code is from CTS test, "dEQP-VK.glsl.texture_functions.query.texturequerylod.isampler2d_fragment".
```
//TEST:SIMPLE:-target spirv-asm -entry main -stage fragment

#version 450 core
layout(location = 0) out mediump vec4 o_color;
layout (location = 0) in highp vec2 v_texCoord;
layout(set = 0, binding = 0) uniform highp isampler2D u_sampler;

void main()
{
        o_color = vec4(textureQueryLod(u_sampler, v_texCoord), 0.0, 1.0);
}
```
When ran, a validation error is observed as following,
```
error: line 40: Opcode ImageQueryLod requires one of these capabilities: ImageQuery
  %20 = OpImageQueryLod %v2float %16 %19
```

venkataram-nv:
Fixes #4357 

Add a test to check that the capability is indeed in the generated code.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -1890,6 +1890,7 @@ public vec2 textureQueryLod(__TextureImpl<
         __intrinsic_asm "textureQueryLod";
     case spirv:
         return spirv_asm {
+            OpCapability ImageQuery;
             result:$$float2 = OpImageQueryLod $sampler $p
         };
     default:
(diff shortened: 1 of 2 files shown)


jkwak-work (approved):
Looks good to me

PR #4523, opened 2024-07-01
Fix the type error in kIROp_RWStructuredBufferLoad

kaizhangNV:
In StructuredBuffer::Load(), we allow any type of integer as the input. However, when emitting glsl code,
StructuredBuffer::Load(index)
will be translated to the subscript index of the buffer, e.g. buffer[index], however, glsl doesn't allow 64bit integer as the subscript.

So the easiest fix is to convert the index to integer when emitting glsl.

### DIFF
--- a/source/slang/slang-emit-glsl.cpp
+++ b/source/slang/slang-emit-glsl.cpp
@@ -2068,7 +2068,9 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
 
             emitOperand(inst->getOperand(0), leftSide(outerPrec, prec));
             m_writer->emit("._data[");
+            m_writer->emit("uint(");
             emitOperand(inst->getOperand(1), getInfo(EmitOp::General));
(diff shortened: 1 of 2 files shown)


source/slang/slang-emit-glsl.cpp
```diff
@@ -2068,7 +2068,9 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
 
             emitOperand(inst->getOperand(0), leftSide(outerPrec, prec));
             m_writer->emit("._data[");
+            m_writer->emit("uint(");
```
csyonghe, line 2071: Let's add a comment here on glsl not supporting anything but 32bit int, so we don't forget why we have the logic here.

csyonghe (dismissed):

source/slang/slang-emit-glsl.cpp
kaizhangNV: done.

PR #4525, opened 2024-07-01
Various fixes around differentiable member associations `[DerivativeMember(<diff-member>)]`

### LINKED ISSUE #4103: `struct` construction using braces fails if differentiable member associations are missing.
The decorator that points a member in the differential type to its corresponding member is missing here, and that causes a crash when differentiating the `IRMakeStruct` instruction.

We likely need some way to automatically form those associations (atleast for when the differential type itself)

Example test that fails:
```

struct Foo : IDifferentiable
{
    typealias Differential = Foo;
    float x[3];
};

[Differentiable]
Foo getFoo(float x)
{
    return { { x, x, x } };
}

[Differentiable]
float foobar(float x)
{
    int i = 3 * int(floor(x));
    Foo foo = getFoo(x);
    return foo.x[i] * foo.x[i];
}

[numthreads(1, 1, 1)]
{
    {
        float a = 0.5;
        var d = fwd_diff(foobar)(diffPair(a, 1.0)).d;
        outputBuffer[0] = d;
    }
}
```

saipraveenb25:
This PR includes the following:
+ Add diagnostic for missing diff-member associations.
+ Automatically create diff member associations if differential type is the same as the primal type.
+ Move diff-member attribute checking to conformance-checking phase to avoid circularity issues.

Fixes #4103

Potentially breaking change: some cases where member associations were not explicitly added will now prompt the user to add the association explicitly with `[DerivativeMember(...)]` attribute or use a `no_diff` on the member declaration.


tests/autodiff/struct-without-diff-associations.slang
jkwak-work, line 1: This may make more sense to be in a directory, `tests/diagnostics`.

source/slang/slang-check-decl.cpp
```diff
@@ -5182,6 +5218,9 @@ namespace Slang
             if (!diffMemberType)
                 continue;
 
+            // Pull up the derivative member name from the atribute
```
jkwak-work, line 5221: Typo `attribute`

jkwak-work (approved):
Looks good to me.

I am little surprised that these types of error checking wasn't already there.

jkwak-work (approved):
Looks good to me, again.

Issue #3289, opened 2023-10-26
Missing compiler warning when attempting to autodiff code that uses shared memory


NBickford-NV:
Hi Slang team! I've been appreciating Slang's errors when I attempt to add `[BackwardDifferentiable]` to code that uses non-differentiable functions, but I think I ran into a case where a warning or error is missing.

Here's a function that broadcasts a value from thread 5 to all other threads in the group using shared memory (note: this code assumes there aren't reads of `s_shared` pending when thread 5 enters this function; otherwise it has a RaW hazard and needs an extra `GroupMemoryBarrierWithGroupSync()` at the top of the function):

```hlsl
groupshared float s_shared;

[BackwardDifferentiable]
float get_thread_5_value(float v, uint group_thread_id)
{
  if(5 == group_thread_id)
  {
    s_shared = v;
  }
  GroupMemoryBarrierWithGroupSync();
  return s_shared;
}
```

I'm calling it using a kernel that looks like this:

```hlsl
[shader("compute")]
[numthreads(128, 1, 1)]
void example_kernel(uint3 group_thread_id : SV_GroupThreadID, uint3 dispatch_thread_id : SV_DispatchThreadID)
{
  DifferentialPair<float> value = diffPair(g_in[dispatch_thread_id.x], 0.f);
  
  bwd_diff(get_thread_5_value)(value, group_thread_id.x, 1.0f);
  
  g_out[dispatch_thread_id.x] = value.d;
}
```

When compiled for CUDA (I haven't tested other targets), Slang 2023.4.1 produces code that always returns a differential of 0, without warnings or errors:

```cuda
__device__ void s_bwd_prop_get_thread_5_value_0(DiffPair_float_0 * dpv_1, uint group_thread_id_1, float _s_dOut_0)
{
    DiffPair_float_0 _S1 = *dpv_1;
    if(5U == group_thread_id_1)
    {
        *(&s_shared_0) = _S1.primal_0;
    }
    __syncthreads();
    dpv_1->primal_0 = (*dpv_1).primal_0;
    dpv_1->differential_0 = 0.0f;
    return;
}
```

I think the correct backward derivative looks something like this; since this is complex and uses more shared memory than the original function, maybe it would be good for now to produce a warning or an error if someone tries to autodiff a function that accesses shared memory?

```hlsl
// For a variable x, let x_i be the value of x in the i'th thread in the group.
// Suppose we have y_i = v_5 and dR/dy_i = dResult_i.
[...]

PR #4526, opened 2024-07-01
Treat global variables and parameters as non-differentiable when checking derivative data-flow

saipraveenb25:
Global parameters & variables are by-default not differentiable (even if they are of a differentiable type), because our auto-diff passes do not touch anything outside of function bodies (and will silently refuse to propagate gradients through them)

The solution, if derivative propagation is necessary, is to use wrapper objects with differentiable getter/setter methods (and we should provide a few such objects in the stdlib). 

This PR adds an error message to enforce this scenario and avoid the situation where the gradient is dropped quietly without warning

Fixes: #3289

This is a potentially breaking change: User code that was previously working with global variables of a differentiable type will now throw an error (previously the gradient would be dropped without warning). The solution is to use `detach()` to keep same behavior as before or rewrite the access using differentiable getter/setter methods.


jkwak-work (approved):
Looks good to me and it makes sense.

Issue #4434, opened 2024-06-19
`and` intrinsic does not support vector arguments


chaoticbob:
Currently, Slang does not support vector types for the `and` intrinsic:
```
shader.hlsl(14): error 30019: expected an expression of type 'bool', got 'vector<bool,2>'
    n = and(m, l);
            ^
shader.hlsl(14): error 30019: expected an expression of type 'bool', got 'vector<bool,2>'
    n = and(m, l);
               ^
shader.hlsl(15): error 30019: expected an expression of type 'bool', got 'vector<bool,3>'
    q = and(o, p);
            ^
shader.hlsl(15): error 30019: expected an expression of type 'bool', got 'vector<bool,3>'
    q = and(o, p);
               ^
shader.hlsl(16): error 30019: expected an expression of type 'bool', got 'vector<bool,4>'
    t = and(r, s);
            ^
shader.hlsl(16): error 30019: expected an expression of type 'bool', got 'vector<bool,4>'
    t = and(r, s);
               ^
shader.hlsl(19): error 30019: expected an expression of type 'bool', got 'vector<bool,2>'
    n = and(n, true);
            ^
shader.hlsl20): error 30019: expected an expression of type 'bool', got 'vector<bool,3>'
    q = and(q, true);
            ^
shader.hlsl(21): error 30019: expected an expression of type 'bool', got 'vector<bool,4>'
```

To ease porting, it would be nice to have support for a vectorized `and`. It's reasonable to assume that the `and` intrinsic gets a decent amount of usage in cases where clarity is desired. Note that there is a special case for when the arguments are a vector and a scalar bool. The scalar bool is implicitly expanded to a vector. However, there is not any special cases for arguments of vectors of two different dimensions, so no truncation.

**Shader***
```
float4 main() : SV_Target
{
    bool a, b, c;
    c = and(a, b);

    bool1 i, j, k;
    bool2 l, m, n;
    bool3 o, p, q;
    bool4 r, s, t;
    k = and(i, j);
    n = and(m, l);
    q = and(o, p);
    t = and(r, s);
        
    k = and(k, true);
    n = and(n, true);
    q = and(q, true);
    t = and(t, true);

    bool2 tv = bool2(and(a, b), and(k.x, and(n.x, and(q.x, t.x))));   
    return tv.x ? float4(0, 0, 0, 0) : float4(1, 1, 1, 1);
}
```

ArielG-NV:
[...]

PR #4529, opened 2024-07-02
Add vector overloads for or and and

### LINKED ISSUE #4441: Missing support for `or` intrinsic
Since Slang supports the `and` and `select` intrinsics, it only feels right to support the `or` intrinsic:
```
float4 main() : SV_Target
{
    bool a, b, c;
    c = or(a, b);

    bool1 i, j, k;
    bool2 l, m, n;
    bool3 o, p, q;
    bool4 r, s, t;
    k = or(i, j);
    n = or(m, l);
    q = or(o, p);
    t = or(r, s);
        
    k = or(k, true);
    n = or(n, true);
    q = or(q, true);
    t = or(t, true);

    bool2 tv = bool2(or(a, b), or(k.x, or(n.x, or(q.x, t.x))));   
    return tv.x ? float4(0, 0, 0, 0) : float4(1, 1, 1, 1);
}
```

expipiplus1:
Closes #4441 and #4434

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2255,6 +2255,51 @@ bool and(bool v0, bool v1)
     return __and(v0, v1);
 }
 
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+__intrinsic_op($(kIROp_And))
+vector<bool, N> and<let N : int>(vector<bool, N> v0, vector<bool, N> v1);
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+vector<bool, N> and<let N : int>(bool b, vector<bool, N> v)
+{
+    return and(vector<bool, N>(b), v);
+}
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+vector<bool, N> and<let N : int>(vector<bool, N> v, bool b)
+{
+    return and(v, vector<bool, N>(b));
+}
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+bool or(bool v0, bool v1)
+{
+    return __or(v0, v1);
+}
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+__intrinsic_op($(kIROp_Or))
+vector<bool, N> or<let N : int>(vector<bool, N> v0, vector<bool, N> v1);
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+vector<bool, N> or<let N : int>(bool b, vector<bool, N> v)
+{
+    return or(vector<bool, N>(b), v);
+}
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+vector<bool, N> or<let N : int>(vector<bool, N> v, bool b)
+{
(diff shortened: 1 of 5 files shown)


source/slang/core.meta.slang
```diff
@@ -2255,6 +2255,51 @@ bool and(bool v0, bool v1)
...
+{
+    return and(vector<bool, N>(b), v);
+}
+
+[__unsafeForceInlineEarly]
+[OverloadRank(-10)]
+vector<bool, N> and<let N : int>(vector<bool, N> v, bool b)
+{
+    return and(v, vector<bool, N>(b));
+}
```
jkwak-work, line 2275: I am not sure if we should support this function that may lead to inefficient code.
A proper implementation for this should happen at the application level in a following manner,
```
vector<bool,N> result = (b? v : vector<bool,N>(false));
```

jkwak-work: @chaosink  for an opinion.

jkwak-work: @chaoticbob for an opinion.

expipiplus1: What makes you think it would be inefficient? Probably best left to the eventual backend to determine precisely what combination of conditional move and bitand is best.  

I agree with not supporting it for different reasons (to avoid proliferation of overloads), but this ship has sailed because hlsl already supports this.

tests/bugs/gh-4434.slang
```diff
@@ -0,0 +1,37 @@
...
+[numthreads(4, 1, 1)]
+void computeMain(uint tid : SV_GroupIndex)
+{
+    bool a, b, c;
+    c = and(a, b);
+
+    bool1 i, j, k;
+    bool2 l, m, n;
+    bool3 o, p, q;
+    bool4 r, s, t;
```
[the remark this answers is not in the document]
jkwak-work: I see.
@venkataram-nv for visibility who is working on a task to error out the use of uninitialized variables.

source/slang/core.meta.slang
chaoticbob: I'm trying to catch up here, is the question here if a vectorized version of the `and` (and also `or`) intrinsic can lead to inefficient code?

jkwak-work: Hmmm. Right.
I think "?:" is probably more efficient than vectorized "and" operation.

chaoticbob: If I'm reading this correctly, the question is about the efficiency of `and`ing a vector and a scalar. For SPIR-V this compiles down an `OpCompositeConstruct` for the scalar to expand to a vector followed by `OpLogicalAnd` to perform the vectorized `and` operation. DXIL handles it with individual instructions per component.

As @expipiplus1 alluded to, it's difficult to gauge how efficient or inefficient this will be for the driver level shader compiler. Driver level shader compilers may have specialized instructions or patterns to handle these cases.

In addition to HLSL already supporting it, it's generally not good idea to be overly prescriptive about how application level code should be written, especially in cases where expression is awkwardly limited in the interest of potential efficiency.

jkwak-work: Ok. I guess the downstream compiler will be able to optimize it out.

Issue #1990, opened 2021-10-26
Stack overflow bug in type checking when using auto type deduction in a generic method.


csyonghe:
```
// Stack overflow bug in type checking when using auto type deduction in a generic method.

//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx12 -profile sm_6_0 -use-dxil -output-using-type
//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx11 -profile sm_5_0 -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute  -output-using-type

[anyValueSize(12)]
interface IBSDF
{
    uint getLobes();
}

[anyValueSize(16)]
interface IMaterial
{
    associatedtype BSDF : IBSDF;
    BSDF setupBSDF();
}

struct StandardBSDF : IBSDF
{
    uint getLobes()
    {
        return 0;
    }
}

struct StandardMaterial : IMaterial
{
    typedef StandardBSDF BSDF;
    StandardBSDF setupBSDF() { StandardBSDF g; return g; }
};

//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=gOutputBuffer
RWStructuredBuffer<float> gOutputBuffer;

//TEST_INPUT: type_conformance StandardMaterial:IMaterial = 0

interface IInterface
{
    void randomStuff();
};
struct Impl : IInterface
{
    void randomStuff(){}
};

void test<R : IInterface>(R r, int id)
{
    float result = 0.0;
    // Use of auto type deduction from a generic method leads to stack overflow during
    // type checking for `bsdf.getLobes`.
    let bsdf = createDynamicObject<IMaterial, int>(id, 0).setupBSDF();
    result = bsdf.getLobes();
    gOutputBuffer[0] = float(result);
}

[numthreads(1, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    Impl impl;
    test(impl, dispatchThreadID.x);
}
```

Expected result:
```
type: float
0.000000
```

tfoleyNV:
[...]

PR #4530, opened 2024-07-02
Warnings for uninitialized values

### LINKED ISSUE #3832: Support warnings on uninitialized variables.

venkataram-nv:
Addresses #3832 

Changes:
- Previously `slang-ir-uninitialized-out-params.*` dictated diagnostics logic for output parameters. Since this has similar behavior to checking usage of uninitialized variables, it has been refactored into `slang-ir-uninitialized-values.*` which handles both.
- Many `std` functions are meta function which operate on the type of the parameter. However, in doing so, they instantiate an undefined variable of the desired type and then pass that to some compiler intrinsic. These intrinsic have their own op codes so they can be detected and ignored.
- Likewise, differentiable functions (and functions synthesized in the process) are ignored, since requiring certain values to be initialized may conflict with the autodiff pass.

What this does not implement (could be future PRs):
- Checking uninitialized fields within constructors
- Partially uninitialized values with respect to data structure (e.g. arrays/structs/vector types)
- Partially uninitialized values with respect to control flow (e.g. if/else/loop)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -1440,14 +1440,14 @@ static void addLinkageDecoration(
...
+    //
+    // Care is needed around stdlib as it is only compiled once and *without* obfuscation,
     // so any linkage name to stdlib *shouldn't* have obfuscation applied to it.
-    if (context->shared->m_obfuscateCode && 
+    if (context->shared->m_obfuscateCode &&
         !isFromStdLib(decl))
     {
         const auto obfuscatedName = getHashedName(mangledName.getUnownedSlice());
-    
+
```
ArielG-NV, line 1450: (example of white space added

source/slang/slang-lower-to-ir.cpp
```diff
@@ -1440,14 +1440,14 @@ static void addLinkageDecoration(
     const String mangledName = getMangledName(context->astBuilder, decl);
 
     // Obfuscate the mangled names if necessary.
-    // 
-    // Care is needed around stdlib as it is only compiled once and *without* obfuscation, 
+    //
+    // Care is needed around stdlib as it is only compiled once and *without* obfuscation,
     // so any linkage name to stdlib *shouldn't* have obfuscation applied to it.
-    if (context->shared->m_obfuscateCode &&
```
ArielG-NV, line 1446: Even though this is a random space at the end of a line, we should not be removing random white space.

venkataram-nv: Noted. Some the files right now (lower-to-ir included) are modified just for debugging. I will revert them to the original state when I finish implementing and testing.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -0,0 +1,315 @@
+#include "slang-ir-use-uninitialized-values.h"
+#include "slang-ir-insts.h"
+#include "slang-ir-reachability.h"
+#include "slang-ir.h"
+
+namespace Slang
+{
+    bool metaOp(IRInst* inst) {
```
ArielG-NV, line 8: Coding style for this project is new line for braces with few exceptions.
```suggestion
    bool metaOp(IRInst* inst) 
    {
```

tests/autodiff/material2/MxLayeredMaterial.slang
```diff
@@ -55,8 +55,7 @@ public struct MxLayeredMaterial : IMaterial
     public UsedMaterialInstance setupMaterialInstance(out MaterialInstanceData miData)
     {
         float3 albedo = getAlbedo(baseColor);
-        UsedMaterialInstance mi;
         for (uint i = 0; i < 3; i++) miData.data[i] = albedo[i];
-        return mi;
+        return UsedMaterialInstance();
```
ArielG-NV, line 59: This is not needed, struct's automatically call their default constructor if available. This may need to be a case when checking for uninitialized variables.

tests/autodiff/treat-as-differentiable-1.slang
```diff
@@ -30,7 +30,7 @@ float use(IFoo o, float x)
 [numthreads(1, 1, 1)]
 void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
 {
-    B b;
+    B b = B();
```
ArielG-NV, line 33: [Same reasoning](https://github.com/shader-slang/slang/pull/4530/files#r1665861986)
```suggestion
    B b;
```

tests/bugs/gh-1990.slang
```diff
@@ -28,7 +28,10 @@ struct StandardBSDF : IBSDF
 struct StandardMaterial : IMaterial
 {
     typedef StandardBSDF BSDF;
-    StandardBSDF setupBSDF() { StandardBSDF g; return g; }
+    StandardBSDF setupBSDF() {
```
ArielG-NV, line 31:
```suggestion
    StandardBSDF setupBSDF()
    {
```

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,94 @@
...
+    float result = inputData.value;
+
+    //CHK: error 41015: use of uninitialized value 'undefVar'
+    result += undefVar;
+
+    //CHK: error 41015: use of uninitialized value 'undefData'
+    result += undefData.value;
+
+    //CHK: error 41015: use of uninitialized value 'nestedData'
+    result += nestedData.data.value;
```
ArielG-NV, line 91: Are the order of these errors/warnings fixed?

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,139 @@
...
+    int field;
+
+    __init() {
+        // TODO: specialized warning for initialized
+        // fields in the constructors
+        field = 1;
+    }
+};
+
+struct NoDefault {
```
ArielG-NV, line 111:
```suggestion
struct NoDefault 
{
```

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,139 @@
...
+    int f(int i) {
```
ArielG-NV, line 112:
```suggestion
    int f(int i) 
    {
```

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,139 @@
...
+{
+    __init() {}
+};
+
+// No warnings for empty structs even without __init()
+struct NonEmptyStruct
+{
+    int field;
+
+    __init() {
```
ArielG-NV, line 104:
```suggestion
    __init()
    {
```

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,139 @@
...
+
+// Should not warn here (unconditionalBranch)
+float3 unconditional(int mode)
+{
+    float f(float) { return 1; }
+
+    float k0;
+    float k1;
+
+    if (mode == 1) {
```
ArielG-NV, line 11:
```suggestion
    if (mode == 1)
    {
```

tests/diagnostics/uninitialized-variables.slang
```diff
@@ -0,0 +1,139 @@
...
+    result += undefVar;
+
+    //CHK-DAG: error 41015: use of uninitialized value 'undefData'
+    result += undefData.value;
+
+    //CHK-DAG: error 41015: use of uninitialized value 'nestedData'
+    result += nestedData.data.value;
+
+    return float4(result);
+}
```
ArielG-NV, line 139: A check at the end is required to ensure we only receive `error 41015` an expected number of times to avoid regressions

`//CHK-NOT: error 41015`

ArielG-NV:
This PR seems to not handle the case of `groupshared` memory (as tested).
This memory is not zero initialized from what I remember, it is undefined behavior to not initialize this memory.

For this case there are 3 scenarios:
1. for-sure Initialized
2. for-sure uninitialized
3. May be (un)initialized -- thread0 may initialize groupshared memory for thread1 under some branch.

venkataram-nv:
> This PR seems to not handle the case of `groupshared` memory (as tested). This memory is not zero initialized from what I remember, it is undefined behavior to not initialize this memory before use.
> 
> For this case there are 3 scenarios:
> 
>     1. For-sure Initialized (possible through store and atomicStore/InterlockedExchange operations)
> 
>     2. For-sure uninitialized
> 
>     3. Maybe (un)initialized -- thread_0 may initialize groupshared memory for thread_1 under some branch which runs only on some threads (likely not easily detectable).

I see, I will add that part then. Should we be emitting similar diagnostics for global variables (e.g. `static float x;`)?

ArielG-NV:
> I see, I will add that part then. Should we be emitting similar diagnostics for global variables (e.g. static float x;)?

I don't know the answer to this question since the behavior of `static` across platforms is inconsistent:
1. Metal: Unknown
2. HLSL: zero initializes, ["[if a static] does not include an initializer, the value is set to zero"](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax)
3. GLSL: uninitialized, ["Global variables without storage qualifiers that are not initialized in their declaration or by the application will not be initialized, but rather will enter main() with undefined values"](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf)
4. SPIR-V: Likely follows GLSL behavior.

venkataram-nv:
Currently the warning system does not properly handle constructors (this is the reason for the modification in [ctor-ordinary-retval-legal.slang](https://github.com/shader-slang/slang/pull/4530/files#diff-ed54864449615c09c97aa1286ed61b48b2c2f25590582c31d5b33c663782c00c)). I intend to fix this in another PR since fixing this involves modifying some other files.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        addresses.add(inst);
+        for (auto user : users)
+        {
+            if (isAliasable(user))
+            {
+                auto instAddresses = getAliasableInstructions(user);
+                addresses.addRange(instAddresses);
+            }
+        }
```
ArielG-NV, line 158: Question: 
do we need to allocate a list here, would the following work?

```suggestion
        addresses.add(inst);
        for (auto use = inst->firstUse; use; use = use->nextUse)
        {
            IRInst* user = use->getUser();
            // Meta instructions only use the argument type
            if (isMetaOp(user) || !isAliasable(user))
                continue;
            addresses.addRange(getAliasableInstructions(user));
        }
```

venkataram-nv: Yes, that works, thanks for the suggestion.

ArielG-NV:
I was testing and it seems the following code does not warn despite using uninitialized values.

Example:
```
RWStructuredBuffer<int> output;

void inoutTest(inout int badInt)
{
    output[0] = badInt;
}

void inTest(int badInt)
{
    output[1] = badInt;
}

[numthreads(1, 1, 1)]
{
    int input1;
    inoutTest(input1);

    int input2;
    inTest(input2);

    output[2] = input1;
    output[3] = input2;
}
```

1. `inoutTest(input1);`->` output[0] = badInt;` should warn
2. `inTest(input2);`->` output[1] = badInt;` should warn
3. `output[2] = input1;` should warn
4. `output[3] = input2;` should warn

venkataram-nv:
@ArielG-NV 
Currently the methods do not analyze the behavior of functions with respect to its input (e.g. if its passed as an `out` parameter). I plan to separate that into a separate smaller issue (along with another for constructors). As for buffer accesses, the assumption is that it is a parameter that is passed into the shader, so usage of it is not treated as undefined.

ArielG-NV:
> Currently the methods do not analyze the behavior of functions with respect to its input
👍 likely worth making an issue with all of the next steps.

Otherwise, code looks good to me.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -0,0 +1,381 @@
...
+        // Nothing to initialize for a pure interface
+        if (as<IRInterfaceType>(type))
+            return true;
+
+        // For pointers, check the value type (primarily for globals)
+        if (auto ptr = as<IRPtrType>(type))
+            return canIgnoreType(ptr->getValueType());
+
+        // In the case of specializations, check returned type
+        if (auto spec = as<IRSpecialize>(type)) {
```
csyonghe, line 119: brace on its own line.

csyonghe (approved):

16-Bit-Dog (approved):

ArielG-NV (approved):

PR #4532, opened 2024-07-03
Add WGSL output support

aleino-nv:
Tint is integrated to support this transformation:

- Slang (input from File)
- SPIR-V (using existing Slang transform)
- Tint program (using Tint's SPIR-V reader)
- WGSL (output, using Tint's WGSL writer)

The Vulkan headers needed to be updated to build Tint.


aleino-nv:
Here is some [output](https://aleino-nv.github.io/issues/4483/results2.html).

csyonghe:
Closing this since we have decided to no go through the tint route and work on emitting WGSL directly.

Issue #4531, opened 2024-07-02
`all` intrinsic fails codegen on cuda and spirv


expipiplus1:
blocking https://github.com/shader-slang/slang/pull/4529

expipiplus1:
Related to https://github.com/shader-slang/slang/issues/4386

expipiplus1:
Fixed for glsl by https://github.com/shader-slang/slang/commit/40a4022670c9b6a53c31ba4841eaf744d6402bf2 and cuda by 639576cbe9d9cf604c554fda8dda48485201704f

PR #4534, opened 2024-07-03
Add specializations for 1 vectors in functions using the $N glsl intrinsic placement

### LINKED ISSUE #4533: Invalid intrinsic expansion for 1 vectors
For some targets we lower 1 vectors to just the element type, this prevents some snippets applying correctly because they're expecting a vector type.

expipiplus1:
Closes #4533
Fixes part of #4531


jkwak-work (approved):
Looks good to me.

csyonghe (commented):
This doesn't seem like a systematic solution. All other intrinsics should have the same issue with 1-vectors and we need to handle them all in a unified way.

Issue #4487, opened 2024-06-26
Metal: Slang does not legalize `[[color(n)]]`/`[[attribute(n)]]` attributes correctly 


ArielG-NV:
Related to: #4291
Breaks compile of:
* tests/compute/compile-time-loop.slang
* tests/compute/constexpr.slang
* tests/compute/discard-stmt.slang
* tests/compute/texture-sampling.slang

Problem:
the following metal code is allowed:
```
struct FragmentStageOutput_0
{
    float4 fragment_0 [[color(0)]];
};

[[fragment]] FragmentStageOutput_0 fragmentMain(float device* outputBuffer_0 [[buffer(0)]])
{
    thread FragmentStageOutput_0 output_0;
    (&output_0)->fragment_0 = float4(1.0, 1.0, 1.0, 1.0);
    *(outputBuffer_0+int(0)) = 1.0;
    return output_0;
}
```

the following metal code is not allowed (Slang generates the following code if `struct` wrapping a `SV_Target`):
```
struct float4Wrapper
{
    float4 v;
}
struct FragmentStageOutput_0
{
    float4Wrapper fragment_0 [[color(0)]];
};

[[fragment]] FragmentStageOutput_0 fragmentMain(float device* outputBuffer_0 [[buffer(0)]])
{
    thread FragmentStageOutput_0 output_0;
    (&output_0)->fragment_0.v = float4(1.0, 1.0, 1.0, 1.0);
    *(outputBuffer_0+int(0)) = 1.0;
    return output_0;
}
```

Specifically, metal allows `[[color()]]` and `[[attribute()]]` to only be annotated on `PRIMITIVE` and `vector<PRIMITIVE,N>`.

ArielG-NV:
note: this is very similar to `hoistEntryPointParameterFromStruct`

ArielG-NV:
> Is this duplicate to https://github.com/shader-slang/slang/issues/4375 ?

Looks like it is

ArielG-NV:
`[[color(n)]]` is fragment shader

ArielG-NV:
Plan to fix the following issues:
*1.* Allow n number "allowed types" instead of only two "allowed types" for when legalizing `System Semantics` for Metal 
*2.* Move parts of `legalizeEntryPointForMetal` into `EntryPointVaryingParamLegalizeContext` so code compiles 
*3.* Restructure so User Input Semantics legalization works 
*4.* Restructure so User Output Semantics legalization works. 
[...]

PR #4536, opened 2024-07-03
Fixes to Metal Input parameters and Output value input/output semantics

ArielG-NV:
NOTE: still a draft, require to use a TreeMap (OrderedHashMap is an element-order preserving hashmap) to implement overlapping semantic index legalization

Fixes: #4487
Prepares code for a larger overhaul (#4375).

Note:
1. Flattening a struct is the process of making a struct have 0 struct/class members.

Changes:
1. Flatten all input(`flattenInputParameters`)/output(part of `wrapReturnValueInStruct`) structs and assign semantics accordingly.
2. Semantics logic is changed to avoid conflicts rather than to conflict: 1. avoid overlapping semantics 2. Prefer assigning explicit semantics specified by a user.
3. Added a more dynamic `legalizeSystemValue` system so more than 2 types can be targeted for legalization. This is required to legalize `output`. There is still no preference for any converted type, the first valid type will be converted to.
4. Fixed some issue with incorrect output semantics if not a fragment stage (when there are not any assigned semantics)
5. Created `convertSystemValueSemanticNameToEnum` and expanded the existing System-Value Enum system. This allows (sometimes) faster comparisons and helps prepare code for porting into `slang-ir-legalize-varying-params.cpp` (#4375)
6. Separated `legalizeSystemValueParameters`. This was done to make it easier to run `legalizeSystemValue` 1 system-value at a time to simplify logic. This change is optional and can be undone if not preferred.
7. Wrap everything inside a MetalEntryPointLegalizationContext. This was done since it simplifies a lot of logic and will be required for #4375
	* Git diff is quite ugly due to the MetalEntryPointLegalizationContext.
9. Changed some tests which had wrong launch arguments and/or invalid test content


csyonghe (commented):
Can you comment/document more details on what each legalization step is going, with simple examples, and make sure to explain each step of the functions? As of now this isn't easy to review without knowing at the high level the intent of each function.

ArielG-NV:
> Can you comment/document more details on what each legalization step is going, with simple examples, and make sure to explain each step of the functions? As of now this isn't easy to review without knowing at the high level the intent of each function.

Will do 👍

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1213 @@
...
+            //     ...
+            // }
+            // ```
+
+            auto func = entryPoint.entryPointFunc;
+            List<IRParam*> paramsToProcess;
+            for (auto param : func->getParams())
             {
-                if (!key->findDecoration<IRSemanticDecoration>() && !paramVarLayout->findAttr<IRSemanticAttr>())
+                if (as<IRStructType>(param->getDataType()))
```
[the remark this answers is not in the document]
ArielG-NV: This code was unchanged, should this be new behavior this PR modifies?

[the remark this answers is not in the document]
csyonghe: I think so, it doesn't make sense to legalize uniform parameters.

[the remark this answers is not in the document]
ArielG-NV: 👍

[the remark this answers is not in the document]
ArielG-NV: Assuming by uniform parameter means:
`float4 fragmentMain(VOut fragmentIn, uniform UniformInParam uniformIn) : SV_Target`
(specifically `uniform UniformInParam uniformIn`)

It seems that uniform parameters are already hoisted out earlier.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+                    paramsToProcess.add(param);
+                }
+            }
 
-                IRVarLayout* fieldLayout = structTypeLayout ? structTypeLayout->getFieldLayout(fieldIndex) : nullptr;
+            IRBuilder builder(func);
+            builder.setInsertBefore(func);
+            for (auto param : paramsToProcess)
+            {
+                auto structType = as<IRStructType>(param->getDataType());
```
csyonghe, line 93: Should comment here that the struct type is already flattened earlier so we don't need to account for nested struct type fields here.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+                struct Fragment
+                {
+                    float4 p1 : SV_TARGET0;
+                    NestedFragment p2 : SV_TARGET1;
+                    NestedFragment p3 : SV_TARGET2;
+                };
 
-        if (paramsToPack.getCount() == 0)
-            return;
+                // Legalized without flattening
```
csyonghe, line 201: Add "-- abandoned approach"

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+                    continue;
+                if (param->findDecorationImpl(kIROp_HLSLMeshPayloadDecoration))
+                    continue;
+                // If we find a IRParam with a IRStructType member, we need to flatten the entire IRParam
+                if (auto structType = as<IRStructType>(param->getDataType()))
                 {
-                    auto resourceKind = offsetAttr->getResourceKind();
-                    if (resourceKind == LayoutResourceKind::VaryingInput)
+                    IRBuilder builder(func);
+                    Dictionary<IRStructField*, IRStructField*> mapOldFieldToNewField;
```
csyonghe, line 241: How can this dictionary not having conflicting keys if a struct type is used in duplicate fields? After flattening a single original field can map to different flattened fields, right?

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+                    for (UInt i = 0; i < oldLayout->getOperandCount(); i++)
+                    {
+                        auto operand = oldLayout->getOperand(i);
+                        if (as<IRVarOffsetAttr>(operand) || as<IRUserSemanticAttr>(operand) || as<IRSystemValueSemanticAttr>(operand) || as<IRStageAttr>(operand))
+                            instToCopy.add(operand);
+                    }
+                    IRVarLayout* newLayout = builder.getVarLayout(instToCopy);
+                    builder.addLayoutDecoration(newKey, newLayout);
+                }
+                // step 3b
```
csyonghe, line 720: I think this will insert conflicting fields if a struct has been used in more than 1 field.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+            // 3. Store the mapping from old to new struct fields to allow copying a old-struct to new-struct.
+            builder.setInsertAfter(src);
+            auto newStruct = builder.createStructType();
+            copyNameHintAndDebugDecorations(newStruct, src);
+            return _flattenNestedStructs(builder, newStruct, src, nullptr, nullptr, mapFieldToField, varsWithSemanticInfo);
         }
-        else if (semanticName == "sv_sampleindex")
+
+        template<typename CopyLogicFunc>
+        void _replaceAllReturnInst(IRBuilder& builder, IRFunc* targetFunc, IRStructType* newType, CopyLogicFunc copyLogicFunc)
```
csyonghe, line 755: Add a comment for this function.

tests/metal/nested-struct-fragment-output.slang
```diff
@@ -0,0 +1,40 @@
...
+
+struct NestedFragment
+{
+    float2 p3;
+};
+struct Fragment
+{
+	float4 p1;
+    float3 p2;
+    NestedFragment p3;
```
csyonghe, line 19: Add one more NestedFragment field here.

tests/metal/nested-struct-fragment-output.slang
```diff
@@ -0,0 +1,40 @@
...
+//TEST:SIMPLE(filecheck=METALLIB): -target metallib -stage fragment -entry fragmentMain
+
+//METAL-DAG: color(0)
+//METAL-DAG: color(1)
+//METAL-DAG: color(2)
+//METALLIB: @fragmentMain
+
+RWStructuredBuffer<float> outputBuffer;
+
+struct NestedFragment
```
csyonghe, line 11: When writing tests, 3 is the magical number. Test with 3-layer nested structs instead of just two.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1371 @@
...
+                return {};
             auto sysValAttr = layoutDecor->findAttr<IRSystemValueSemanticAttr>();
             if (!sysValAttr)
-                continue;
+                return {};
             auto semanticName = String(sysValAttr->getName());
             auto sysAttrIndex = sysValAttr->getIndex();
-            systemValWorkItems.add({ param, semanticName, sysAttrIndex });
+
+            return std::make_optional<SystemValLegalizationWorkItem>({ var, semanticName, sysAttrIndex });
```
csyonghe, line 1314: Should be able to use the same initializer list style return here as the earlier return statements.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -731,67 +1393,105 @@ namespace Slang
...
-                }
-                else
+                varTypeIsPermitted = varTypeIsPermitted || permittedType == varType;
+            }
+
+            if (!varTypeIsPermitted)
+            {
+                // Note: we do not currently prefer vector<T,N> to vector<P,N>.
+                // vector<T,N> to vector<T,N-1> is equally preferred.
+                bool foundAConversion = false;
```
csyonghe, line 1432: why can't we convert uint3 to int3?

tests/metal/nested-struct-fragment-output.slang
csyonghe: Also need to test explicit semantics.
Does it work if we have a nested struct with explicit semantics, but no explicit semantics inside the nested struct fields?

source/slang/slang-ir-metal-legalize.cpp
ArielG-NV: The variable is not flattened by this pass: first we hoist, then pack `[StageIn]`, then reformat the `[StageIn]` to be legal

The following code demonstrates this
```
            //Input Parameter Legalize
            hoistEntryPointParameterFromStruct(entryPoint);
            packStageInParameters(entryPoint);
            flattenInputParameters(entryPoint);
```

source/slang/slang-ir-metal-legalize.cpp
ArielG-NV: We can convert uint3 to int3

The following comment was more about addressing the following example:
system value example allowed types: `float4`, `uint4`, `int4`
user used: `float2`
Slang will equally prefer `float4` to `uint4` to `int4`. This means the type may lose data if slang selects `uint4` or `int4`.

I will update the comment for clarity

source/slang/slang-ir-metal-legalize.cpp
ArielG-NV: This is an oversight with the current design. 
The hierarchy of members needs to be tracked to solve this problem.

csyonghe:
Is this still a draft or ready for review?

ArielG-NV:
> Is this still a draft or ready for review?

CI passes, the code is ready for review now.

source/slang/slang-ir-legalize-varying-params.h
```diff
@@ -21,13 +22,63 @@ void legalizeEntryPointVaryingParamsForCUDA(
...
     IRFunc* entryPoint,
     IRVectorType* type);
+
+IRInst* emitCalcGroupIndex(
+    IRBuilder& builder,
+    IRInst* groupThreadID,
+    IRInst* groupExtents);
+
+// SystemValueSemanticName member definition macro
+#define SYSTEM_VALUE_SEMANTIC_NAMES(M)              \
```
jkwak-work, line 41: It looks like this macro is used only in `source/slang/slang-ir-legalize-varying-params.cpp`.
Maybe you want to move this to the cpp file.
I see that you left a similar comment a few lines above but it is unclear which lines "the following below" refers to.

source/core/slang-tree-map.h
```diff
@@ -0,0 +1,474 @@
...
+#include "slang-list.h"
+#include "slang-common.h"
+#include <initializer_list>
+
+namespace Slang
+{
+
+    // Array-backed Red-black Tree Map.
+    template<typename T>
+    class TreeMap
```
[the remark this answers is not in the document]
ArielG-NV: > My guess is that you want to have the result to be in a sorted order

This is correct. I want the result to be in sorted over after every "add" operation.

> In practice, it is more efficient and simple to collect elements in a random order and do one time sort at the end; if there aren't duplicates.

Sorting a `List` after adding a new element will likely be slower in our use-case. This is because the code using `TreeSet` does the following 3 functionalities:
1. Distinct pass over elements where we 'search for an element', then we 'add a new element',  `TreeSet` is significantly faster to search than an unordered list, adding elements is slower.
2. [If we were using a `List` we would sort here]
3. We in a loop then search from least to greatest and insert a sorted element, `List` will likely be slower to insert a sorted element, `List` will be slightly faster to iterate.

> If we will have this container, the naming should change to Set, because Map implies a mapping from A to B.

I agree.

source/slang/slang-ir-legalize-varying-params.h
ArielG-NV: > It looks like this macro is used only in source/slang/slang-ir-legalize-varying-params.cpp.
> Maybe you want to move this to the cpp file.

This will not work. `SYSTEM_VALUE_SEMANTIC_NAMES` is a helper macro which is used to help define members of `SystemValueSemanticName`. `SystemValueSemanticName` members are used by `slang-ir-metal-legalize.cpp` and `slang-ir-legalize-varying-params.cpp`.

> I see that you left a similar comment a few lines above but it is unclear which lines "the following below" refers to.

This "similar" comment refers to functionality requested by #4375. I will adjust the comment accordingly to be clearer.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +10,1213 @@
...
             {
-                auto fieldParam = builder.emitParam(field->getFieldType());
-                
-                IRCloneEnv cloneEnv;
-                cloneInstDecorationsAndChildren(&cloneEnv, builder.getModule(), field->getKey(), fieldParam);
-
-                IRVarLayout* fieldLayout = structTypeLayout ? structTypeLayout->getFieldLayout(fieldIndex) : nullptr;
-                if (varLayout)
+                auto key = field->getKey();
+                bool foundDecoration = true; // Some decorations appear twice, destroy all
```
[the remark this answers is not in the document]
jkwak-work: do { } while(false}; works too.

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -10,709 +11,1479 @@
...
+                    varLayoutBuilder.findOrAddResourceInfo(LayoutResourceKind::MetalPayload);
+                    auto paramVarLayout = varLayoutBuilder.build();
+                    builder.addLayoutDecoration(packedParam, paramVarLayout);
+
+                    // Now we replace the call to DispatchMesh with a call to the mesh grid properties
+                    // But first we need to create the parameter
+                    const auto meshGridPropertiesType = builder.getMetalMeshGridPropertiesType();
+                    auto mgp = builder.emitParam(meshGridPropertiesType);
+                    builder.addExternCppDecoration(mgp, toSlice("_slang_mgp"));
+                    }
```
csyonghe, line 1356: indent is wrong.

tests/metal/sv_target-complex-1.slang
```diff
@@ -0,0 +1,34 @@
...
+struct NestedReturn2
+{
+    float4 debugAux1;
+    float4 debugAux2;
+};
+
+// Semantics are supposed to ignore uppercase/lowercase differences
+struct Output
+{
+    float4 Diffuse : SV_TarGet0;
```
csyonghe, line 17: The layout has overlap here, this should be an error instead of us silently adjusting things to make them non-overlapping.

csyonghe (approved):

PR #4552, opened 2024-07-06
set macosx-min-version=10.13

skallweitNV:


jkwak-work:
Three tests are failing, MacOS build CI and Verify-VSProject, and Windows build CI.
Let me restart them and see if they pass.

skallweitNV:
using `CMAKE_OSX_DEPLOYMENT_TARGET` now

skallweitNV:
`std::filesystem::path` is only used in very few places so we could just get rid of its usage.

PR #4557, opened 2024-07-08
Add user guide section on user-defined attributes.

csyonghe:


jkwak-work (approved):
Looks good to me.
It seems very interesting, but I cannot think of where this can be used.

csyonghe:
The user defined attribute system is to allow users to build tools around their shaders.

For example, an application may display an UI to tweak a shader parameter. User defined attributes can provide additional meta data on what kind of UI should be used to set its value, for example:

```
[RangedInt(0, 10)]
[DefaultValue(5)]
uniform int scale;

[Color]
uniform float3 color;
```

These attributes can the be queried from the reflection API, and the application can then display a text box/ drop down/ color picker based on the meta data.

Issue #4556, opened 2024-07-08
crashes when converting matrix date type in -emit-spirv-direct


kaizhangNV:
Slang crashes when converting a matrix's data type, e.g.

```
    float3x4 a = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 };
    double3x4 b;

    b = (double3x4)a;      // This will cause crash
```
The reason is that in  the `emitFloatCast` we didn't check if the operand is matrix type, because we assume that the operand must be scalar type.

PR #4559, opened 2024-07-08
Fix the issue in emitFloatCast

kaizhangNV:
This fixes issue: #4556
In emitFloatCast function, we only considered the input type is float scalar or float vector, so if the input type is a float matrix type, it will crash.

We should also handle the float matrix type.

Also, we add some diagnose info to point out the source location where there is error happened, so in the future it's easier to tell us what happens.

### DIFF
--- a/tests/bugs/gh-4556.slang
+++ b/tests/bugs/gh-4556.slang
@@ -1,0 +1,21 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-dx12 -compute -output-using-type -shaderobj
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -compute -output-using-type -shaderobj
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -glsl -compute -output-using-type -shaderobj
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-mtl -compute -output-using-type -shaderobj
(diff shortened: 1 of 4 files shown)


csyonghe:
We should add a regression test to cover the new logic.

csyonghe (approved):

tests/bugs/gh-4556.slang
```diff
@@ -0,0 +1,21 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-dx12 -compute -output-using-type -shaderobj
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -compute -output-using-type -shaderobj
+//TEST(compute, vulkan):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -glsl -compute -output-using-type -shaderobj
```
jkwak-work, line 3: I am getting a validation error for the 2nd and 3rd test.
The error message is following,
```
error: line 51: Expected float scalar or vector type as Result Type: FConvert
  %b = OpFConvert %mat3v4double %12
```
You may want to add the following test to catch the problem.
```
//TEST(compute):SIMPLE(filecheck=SPIRV): -entry computeMain -target spirv-asm -stage compute
//SPIRV-NOT: Validation of generated SPIR-V failed
```
```

jkwak-work: I am trying to figure out why CI is fine and my machine is not fine.
I have a higher version of Vulkan-SDK than what CI uses, but I am not sure if that's the real cause or not yet.

PR #4568, opened 2024-07-09
Fix lowering of associated types and synthesis of dispatch functions.

### LINKED ISSUE #4101: Nested associated types with multiple possible implementations will crash the compiler.
Currently, nested associated types are expected to be fully resolved rather than treated as existential types. This is a limitation of the current existential type lowering approach that needs some refactoring. 

This pattern is found in the latest differentiable path tracer in Falcor, and is currently being implemented using a temporary workaround.

Here is a full test-case that triggers this problem:

```C#
// Test calling differentiable function through dynamic dispatch.

//TEST_INPUT:ubuffer(data=[0 0 0 0 0], stride=4):out,name=outputBuffer

[anyValueSize(16)]
interface IFoo
{
    float foo();
}

[anyValueSize(16)]
interface INestedInterface
{
    associatedtype NestedAssocType : IFoo;
}

[anyValueSize(16)]
interface IInterface
{
    associatedtype MyAssocType : INestedInterface;
    MyAssocType.NestedAssocType calc(float x);
    float uncalc(MyAssocType.NestedAssocType obj);
}

// ================================

struct A_Assoc_Assoc : IFoo
{
    float a;

    float foo()
    {
        return a;
    }
}

struct A_Assoc : INestedInterface
{
    typedef A_Assoc_Assoc NestedAssocType;
}

struct A : IInterface
{
    typedef A_Assoc MyAssocType

    int data1;

    __init(int data1) { this.data1 = data1; }

    A_Assoc_Assoc calc(float x) { return { x * x * x * data1 }; }
    float uncalc(A_Assoc_Assoc obj) { return obj.a; }
};

// ================================

struct B_Assoc_Assoc : IFoo
{
    float b;

    float foo()
    {
        return b;
    }
}

struct B_Assoc : INestedInterface
{
    typedef B_Assoc_Assoc NestedAssocType;
}

struct B : IInterface
{
    typedef B_Assoc MyAssocType

    int data1;
    int data2;

    __init(int data1, int data2) { this.data1 = data1; this.data2 = data2; }

    B_Assoc_Assoc calc(float x) { return { x * x * data1 * data2 }; }
    float uncalc(B_Assoc_Assoc obj) { return obj.b; }
};

// ================================

float doThing(IInterface obj, float x)
{
    let o = obj.calc(x);
    return o.foo();
}

float f(uint id, float x)
{
    IInterface obj;

    if (id == 0)
        obj = A(2);
    else
        obj = B(2, 3);

    return doThing(obj, x);
}

//TEST_INPUT: type_conformance A:IInterface = 0
//TEST_INPUT: type_conformance B:IInterface = 1

[numthreads(1, 1, 1)]
{
    outputBuffer[0] = f(dispatchThreadID.x, 1.0); // A.calc, expect 12
    outputBuffer[1] = f(dispatchThreadID.x + 1, 1.5); // B.calc, expect 18
}
```

saipraveenb25:
We have a couple of corner cases with associated types and the synthesis of dispatch functions for existential types that were incorrectly handled.

1. When lowering types to IR, if we have a 'lookup chain' of an associated type of an associated type of a this-type, this has to be lowered directly as the inner-most associated type (no `IRLookupWitnessMethod` inst is necessary because the base is an abstract type). We add a check for `IRInterfaceRequirementEntry` when checking for abstract witness tables.

2. The type of the result inst from the witness-table-id-lookup function is incorrect. It is currently `WitnessTableID(param-interface-type)` when it should be `WitnessTableID(result-interface-type)`. This patch fixes this issue (and also fixes the 'key not found in dictionary' error)

This patch also adds a test (`nested-assoc-types.slang`), and removes some unused code (`struct IRLookupWitnessTable`)

Fixes #4101


jkwak-work (approved):
Looks good to me.
I didn't know an associated type can be another associated type.

csyonghe (approved):

Issue #4583, opened 2024-07-09
slang-torch: dtype `half` is not supported properly


saipraveenb25:
Using half-type tensors causes weird CUDA compilation errors. Likely due to torch explicitly disabling the half type in CUDA when generating the ninja build files.

Example: 

```csharp
[AutoPyBindCUDA]
[CUDAKernel]
void square(TensorView<half> input, TensorView<half> output)
{
    // Get the 'global' index of this thread.
    uint3 dispatchIdx = cudaThreadIdx() + cudaBlockIdx() * cudaBlockDim();

    // If the thread index is beyond the input size, exit early.
    if (dispatchIdx.x >= input.size(0))
        return;

    output[dispatchIdx.x] = input[dispatchIdx.x] * input[dispatchIdx.x];
}

```

```python

module = slangtorch.loadModule("square-half.slang")
self.module = module

X = torch.tensor([1., 2., 3., 4.]).cuda().half()
Z = torch.zeros_like(X).cuda().half()

self.module.square(input=X, output=Z).launchRaw(blockSize=(32, 1, 1), gridSize=(1, 1, 1))

expected = torch.tensor([1., 4., 9., 16.]).cuda().half()

assert(torch.all(torch.eq(Z, expected)))

```

PR #4584, opened 2024-07-09
Add `float16` support for the `TorchTensor` type

saipraveenb25:
This adds the corresponding `dtype` to the prelude.

This patch is one part of the fix for issue #4583 (the second part is in the slang-torch repo)


jkwak-work (approved):
Looks good to me.

PR #4587, opened 2024-07-10
Add reflection API for functions.

### LINKED ISSUE #4567: Extend the reflection API to allow querying function signatures
Our reflection API should be able to find any functions and return its signature (return type, parameter, attributes) to the user.

The existting `EntryPointReflection` type is really just a `FunctionReflection`, and we should be able to provide `Module::findFunctionByName` in addition to `Module::findEntryPointByName`.

We need to extend `EntryPointReflection` type to support query for attributes, by adding `getUserAttributeCount`, `getUserAttributeByIndex` and `findUserAttributeByName` like we did for `VariableReflection`.

csyonghe:
This change adds `SlangFunctionReflection` type in the reflection API that provides methods for querying function result type, parameters and user-defined attributes.

`ProgramLayout::findFunctionByName` can now find a function with the given name and returns a `FunctionReflection`.

`IEntryPoint` now has a `getFunctionReflection` method that returns an `FunctionReflection` for the entrypoint.

Closes #4567.


jkwak-work (approved):
Looks good to me.

I am wondering what happen if there are more than one functions found by `findFunctionByName()`.

eliemichel:
Nice! I was able to use these new features successfully :) Just a few minor notes:

The first argument of `FunctionReflection::findUserAttributeByName` is called `session` but it is actually the `globalSession` that is supposed to go in there if I understand correctly.

Little API inconsistency: Getting a parameter by index is called `FunctionReflection::getParameter` while getting a user attribute by index is called `FunctionReflection::getUserAttributeByIndex` with an explicit "ByIndex". And it is the other way around in the C API: `spReflectionFunction_GetParameterByIndex` versus `spReflectionVariable_GetUserAttribute`.

eliemichel:
Is there a way to get the `no_diff` type qualifier in the function argument reflection?

csyonghe:
@eliemichel I made the api naming more consistent.

For no_diff queries, you can now do this with `funcReflection->getParameterByIndex(0)->findModifier(slang::Modifier::NoDiff)`.

PR #4588, opened 2024-07-10
Fix the invalid spirv generation for matrix cast

kaizhangNV:
Spirv doesn't have instruction to do the float cast for the matrix type. So we have to convert the matrix row by row, and then construct them to a new matrix.

Update the unit test to make sure the cast won't miss any elements.

Also update the workflow to validate the spirv generation.


jkwak-work:
Were you able to reproduce the validation error locally without upgrading the vulkan sdk version?

jkwak-work:
This PR will close #4589.

kaizhangNV:
> sdk

Yes, I can. And now the issue is gone with this change.

.github/workflows/windows-selfhosted.yml
```diff
@@ -47,6 +47,10 @@ jobs:
       
     - name: test
       run: |
+        $ErrorActionPreference = "SilentlyContinue"
+        where.exe spirv-dis
+        spirv-dis --version
+        $env:SLANG_RUN_SPIRV_VALIDATION='1'
```
jkwak-work, line 53: I don't think we need this change anymore.
The file "windows-selfhosted.yml" is delete at ToT.
The same functionality is implemented in ci.yml, which sets SLANG_RUN_SPIRV_VALIDATION correctly.

kaizhangNV: Ok, I can remove this file from this change.

csyonghe (dismissed):
Looks good to me.

kaizhangNV:
I rebase on TOT and remove the `windows-selfhosted.yml`.

PR #4596, opened 2024-07-10
A couple of small fixes to cmake releases

expipiplus1:


.github/workflows/release.yml
```diff
@@ -154,10 +154,10 @@ jobs:
           cpack --preset "$config" -G ZIP
           cpack --preset "$config" -G TGZ
           triggering_ref=${{ github.ref_name }}
-          base=$(pwd)/slang-${triggering_ref#v}-${{matrix.os}}-${{matrix.platform}}
-          mv "$(pwd)/build/dist-${config}/slang.zip" "${base}.zip"
+          base=slang-${triggering_ref#v}-${{matrix.os}}-${{matrix.platform}}
+          mv --verbose "$(pwd)/build/dist-${config}/slang.zip" "${base}.zip"
           echo "SLANG_BINARY_ARCHIVE_ZIP=${base}.zip" >> $GITHUB_OUTPUT
-          mv "$(pwd)/build/dist-${config}/slang.tar.gz" "${base}.tar.gz"
+          mv --verbose "$(pwd)/build/dist-${config}/slang.tar.gz" "${base}.tar.gz"
```
jkwak-work, line 160: I don't think `$(pwd)` is needed.
And it is probably better to wrap the string with double quotes.
```
          base="slang-${triggering_ref#v}-${{matrix.os}}-${{matrix.platform}}"
          mv "build/dist-${config}/slang.zip" "${base}.zip"
```

Issue #4599, opened 2024-07-10
slang-torch: `inline-mlp` example fails to compile


saipraveenb25:
Trying to compile this file: https://github.com/shader-slang/slang-torch/blob/main/examples/inline-mlp-example/inline-mlp.slang
results in `(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unexpected: Generic type/value shouldn't be handled here!` with top-of-tree 

Related issue: https://github.com/shader-slang/slang-torch/issues/11

jkwak-work:
Should we have a CI to check the examples regularly?

jkwak-work:
I assume that the example was working before but it started not working.
Is this a regression?

saipraveenb25:
@jkwak-work. Yes! This is something we had discussed before but never implemented. 

We need to run the tests in the slang-torch repository regularly with the CI, but it does require a somewhat more complex setup than our current tests. They need a python 3.9 + torch 2.3 environment set up.

PR #4602, opened 2024-07-10
Fix issue with synthesizing `Differential` type for self-differential generic types

saipraveenb25:
The problem was that we were using the type that was performing the lookup for `.Differential` which can have substitutions based on the local context where the decl is being referenced.

We need to synthesize the type as if it is being referenced from the decl itself since the Differential typealias is a property provided at the declaration site.

Fixes: #4599


jkwak-work (approved):
Looks good to me.

PR #4606, opened 2024-07-11
Remove syntax error in release script

expipiplus1:


csyonghe:
I think I'll just change the release script of vscode extension and slang-torch for now.

PR #4618, opened 2024-07-11
Fix incorrect codegen when returning initializer list as existential value.

### LINKED ISSUE #4464: Incorrect initializers when using interfaces
Another interface problem.

When this slang program is compiled, it has conformances issues. 
[InitializerTestsInterface.cs.slang.txt](https://github.com/user-attachments/files/15956951/InitializerTestsInterface.cs.slang.txt)
When conformances are provided as:
    conformances.add("Test0", "ITest", 0);
    conformances.add("Test1", "ITest", 1);
    conformances.add("TestAny", "ITest", 2);
the following HLSL code is produced, which seems to have duplicate ID for both `Test0` and `Test1` (see lines 50 and 76 in `Tuple_0 makeTest0_0()` and `Tuple_0 makeTest1_0()` respectively).
[1.hlsl.txt](https://github.com/user-attachments/files/15956974/1.hlsl.txt)

When no conformances are provided, all three structs seem to have the same ID 0.
This manifests in the test itself failing, mostly in the `Test1` part of it.

When conformances are provided without explicit IDs (i.e., ID is `uint(-1)`), then it does not auto-assign IDs and will fail to compile on the dxc level, complaining that it has two `case`s with `4294967295U`, viz:
[UndefinedID.hlsl.txt](https://github.com/user-attachments/files/15957064/UndefinedID.hlsl.txt)

Trying to limit the code only to a small case, the code fails to compile when conformances aren't provided (cannot figure them out on its own), but succeeds when conformance is provided (possibly because it compiles out all the other implementations, so duplicate assignment of ID is not an issue).
[InitializerTestsFails.cs.slang.txt](https://github.com/user-attachments/files/15957097/InitializerTestsFails.cs.slang.txt)

csyonghe:
Closes #4464.

The problem here is that if we have a struct type without a ctor:
```
struct Foo : IInterface
{
}
```
And then if we return a default construct of `Foo` as an exsitential:
```
IInterface makeFoo()
{
    return Foo();
}
```

We will translate `Foo()` call into an `InitializerListExpr` because `Foo` does not have a ctor.
However, when coercing `InitializerListExpr` to the return type `IInterface`, we are just directly producing an initializerListExpr for `IInterface`, which becomes `return IInterface{};`. This is obviously wrong because the type info is lost.

Instead we need to make sure not to go into `coerceInitializerListExpr` if the expr to be coerced is already a coerced initializer list (where the type is no longer `InitializerListType`).

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -16,14 +16,6 @@
       }
     },
     {
-      "name": "dev",
-      "inherits": "default",
-      "description": "Development build with debug assisting features",
-      "cacheVariables": {
-        "SLANG_ENABLE_IR_BREAK_ALLOC": "TRUE"
-      }
-    },
-    {
       "name": "msvc-base",
       "hidden": true,
       "inherits": "default",
@@ -66,7 +58,7 @@
   "buildPresets": [
     {
       "name": "debug",
-      "configurePreset": "dev",
+      "configurePreset": "default",
       "configuration": "Debug"
     },
     {
@@ -145,7 +137,7 @@
       "steps": [
         {
           "type": "configure",
-          "name": "dev"
+          "name": "default"
         },
         {
           "type": "build",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -16,6 +16,14 @@
       }
     },
     {
+      "name": "dev",
+      "inherits": "default",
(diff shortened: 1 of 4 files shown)


jkwak-work (approved):
Looks good to me.

PR #4620, opened 2024-07-11
Change `IDefaultInitializableType` to `IDefaultInitializable`

ArielG-NV:
Change `IDefaultInitializableType` to `IDefaultInitializable` to use a uniform naming scheme for `interface` identifiers across Slang.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -75,7 +75,7 @@ syntax snorm : SNormModifier;
 syntax __extern_cpp : ExternCppModifier;
 
 __magic_type(DefaultInitializableType)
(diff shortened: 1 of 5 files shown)


jkwak-work (approved):
Looks good to me.

Just FYI, Visual Studio has a renaming feature under "refactoring" category.
A keyboard short cut for it is Ctrl+R Ctrl+R.
The search can include comments as well.

ArielG-NV:
> Just FYI, Visual Studio has a renaming feature under "refactoring" category.

Thank-you for sharing, never knew about this.

Regex was a good enough for this case, but I can 100% imagine a scenario where symbol based renaming is better.

PR #4623, opened 2024-07-12
Start autodiff pass if `IRDerivativeMemberDecoration` is found

### LINKED ISSUE #4622: `IDerivativeMember` references its parent causing a 'circular-definition' when emitting code
repro:
```
struct PowActivationEx : IDifferentiable
{
	float power;
}
RWStructuredBuffer<float> return_value;

[numthreads(256, 1, 1)]
void computeMain()
{
    PowActivationEx args;
	return_value[0] = args.power;
}
```

ArielG-NV:
Fixes #4622

Problem:
Currently, if a decoration refers to its-parent through its operands we will have a recursive dependency and fail. The reason we encounter this problem is because autodiff decorations are not stripped if the autodiff -pass never runs. This happens due to the autodiff pass not running if we only have a `IRDerivativeMemberDecoration`.

Solution:
Start autodiff pass if we found a `IRDerivativeMemberDecoration`.

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -292,6 +292,7 @@ void calcRequiredLoweringPassSet(RequiredLoweringPassSet& result, CodeGenContext
     case kIROp_BackwardDifferentiate:
     case kIROp_ForwardDifferentiate:
     case kIROp_MakeDifferentialPairUserCode:
+    case kIROp_DerivativeMemberDecoration:
         result.autodiff = true;
         break;
     case kIROp_VerticesType:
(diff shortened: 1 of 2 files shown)


csyonghe (approved):

saipraveenb25:
@csyonghe: This seems like it would result in always enabling the autodiff pass? Technically, we have a few differentiable types in stdlib, so there will always be a diff-member decoration somewhere in the IR. 

Maybe we should consider running the decorations removal step independently of the main autodiff pass?

ArielG-NV:
The change @saipraveenb25 suggested was added in the same branch (it seems after the code was merged) -- https://github.com/shader-slang/slang/commit/36508e4f02ebe96733ce6a336e0415377bf06caa

saipraveenb25:
Ah, I see.. can you open another PR with the same branch?

PR #4625, opened 2024-07-12
Feature/capture

kaizhangNV:
- Finishes all the decoder works including the dispatch and all the decoding infrastructure.
- Reorganize the file structure
    - capture component in `capture` folder
    - replay component in `replay` folder
- Create a executable placeholder under `tools/slang-replay` folder, this will the final replayer.
- Update the build process
    - Only the capture logic should be built into slang share library
    - Replay logic should not be in slang, it will be a separate executable.
- There are few modules in replayer:
    - File processor: this is used to read the capture file block by block (each block is an API call info)
    - slang-decoder: this is used to decode each block of data, and translate it into the formal data structure for each API
    - decoder-consumer: this is the consumer who consumes the decoded data, we will have 2 consumers:
        - json consumer: convert the capture file into a Json format file for debugging purpose
        - replayer consumer: replay the capture file
    - decoder-helper: this is the infrastructure for decoding the captured binary data.
   
The decoder should be strictly symmetric to encoder.


source/slang-capture-replay/replay/slang-decoder.cpp
```diff
@@ -0,0 +1,1974 @@
...
+
+    bool SlangDecoder::ICreateGlobalSession(ObjectID objectId, ParameterBlock const& parameterBlock)
+    {
+        return false;
+    }
+
+    bool SlangDecoder::IGlobalSession_createSession(ObjectID objectId, ParameterBlock const& parameterBlock)
+    {
+        StructDecoder<slang::SessionDesc> sessionDesc;
+        sessionDesc.decode(parameterBlock.parameterBuffer, parameterBlock.parameterBufferSize);
```
kaizhangNV, line 426:
```
ld: warning: ignoring duplicate libraries: 'Release/lib/libcore.a'
ld: Undefined symbols:
  SlangCapture::StructDecoder<slang::SessionDesc, void>::decode(unsigned char const*, long long), referenced from:
      SlangCapture::SlangDecoder::IGlobalSession_createSession(unsigned long long, SlangCapture::SlangDecoder::ParameterBlock const&) in slang-decoder.cpp.o
      SlangCapture::SlangDecoder::IGlobalSession_getSessionDescDigest(unsigned long long, SlangCapture::SlangDecoder::ParameterBlock const&) in slang-decoder.cpp.o
```

This is weird issue that I haven't seen in gcc and linux clang-15. It only shows up on apple clang-15.

kaizhangNV: The issue is fixed by moving the template explicit instantiation from header file to cpp file.

source/slang-capture-replay/replay/decoder-consumer.h
```diff
@@ -0,0 +1,178 @@
...
+// #include "../../slang/slang-compiler.h"
+#include "../util/capture-format.h"
+
+namespace SlangCapture
+{
+    class IDecoderConsumer
+    {
+    public:
+        virtual void ICreateGlobalSession(ObjectID objectId, const uint8_t* parameterBuffer, int64_t bufferSize) = 0;
+		virtual void IGlobalSession_createSession(ObjectID objectId, slang::SessionDesc const&  desc, ObjectID outSessionId) = 0;
```
csyonghe, line 14: indent.

source/slang-capture-replay/replay/decoder-consumer.h
```diff
@@ -0,0 +1,178 @@
...
+#include "../../../slang.h"
+// #include "../../slang/slang-compiler.h"
+#include "../util/capture-format.h"
+
+namespace SlangCapture
+{
+    class IDecoderConsumer
+    {
+    public:
+        virtual void ICreateGlobalSession(ObjectID objectId, const uint8_t* parameterBuffer, int64_t bufferSize) = 0;
```
csyonghe, line 13: This is a weird name. Should be `createGlobalSession`.

source/slang-capture-replay/replay/decoder-helper.cpp
```diff
@@ -0,0 +1,53 @@
+#include <vector>
```
csyonghe, line 1: Since we have `List`, we shouldn't use `<vector>`.

source/slang-capture-replay/replay/file-processor.h
```diff
@@ -0,0 +1,30 @@
...
+#define FILE_PROCESSOR_H
+
+#include <string>
+#include "../../core/slang-stream.h"
+#include "../util/capture-format.h"
+#include "slang-decoder.h"
+
+namespace SlangCapture
+{
+    class FileProcessor
```
csyonghe, line 11: The name is too generic, what do you mean by `process a file`?

source/slang-capture-replay/replay/parameter-decoder.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+        SLANG_CAPTURE_ASSERT(bufferSize >= (int64_t)(readByte + stringLength));
+
+        if (stringLength == 0)
+        {
+            return readByte;
+        }
+
+        uint8_t* data = (uint8_t*)typeDecoder.allocate(stringLength + 1);
+        memcpy(data, buffer + readByte, stringLength);
```
jkwak-work, line 27: The allocation can fail and return null pointer.
The behavior of memcpy will be undefined when data is nullptr.

source/slang-capture-replay/replay/file-processor.cpp
```diff
@@ -0,0 +1,132 @@
...
+        Slang::FileAccess fileAccess = Slang::FileAccess::Read;
+        Slang::FileShare fileShare = Slang::FileShare::None;
+
+        // Open the capture file with read-only access
+        SlangResult res = m_inputStream.init(path, fileMode, fileAccess, fileShare);
+
+        if (res != SLANG_OK)
+        {
+            SlangCapture::slangCaptureLog(SlangCapture::LogLevel::Error, "Failed to open file %s\n", filename.c_str());
+            std::abort();
```
jkwak-work, line 20: There are many ways to handle an error case. Killing the entire process is probably the last thing we should do.

source/slang-capture-replay/util/capture-format.h
```diff
@@ -147,5 +152,21 @@ namespace SlangCapture
...
+    struct FunctionHeader
+    {
+        uint32_t         magic {MAGIC_HEADER};
+        ApiCallId        callId {InvalidCallId};
+        ObjectID         handleId {0};
+        uint64_t         dataSizeInBytes {0};
+        uint64_t         threadId {0};
+    };
+
+    struct FunctionTailer
```
jkwak-work, line 165: `Footer` will be a better counter word to `Header`

jkwak-work (commented):
I have some comments.
And the meaning of `decode` and `encode` is unclear to me.

kaizhangNV:
> I have some comments. And the meaning of `decode` and `encode` is unclear to me.

`encode` is to encode the APIs and parameters to a specific format in binary.
`decode` is to reconstruct a binary stream with specific format to the API call.

source/slang-capture-replay/replay/decoder-helper.h
```diff
@@ -0,0 +1,108 @@
...
+    {
+    public:
+        void setPointer(void* data) { m_pointer = static_cast<T>(data); }
+        T getPointer() const             { return m_pointer; }
+        void setPointerAddress(uint64_t address) { m_pointerAddress = address; }
+        void setDataSize(size_t size) { m_dataSize = size; }
+        size_t getDataSize() const { return m_dataSize; }
+    private:
+        T m_pointer {nullptr};
+        uint64_t m_pointerAddress = 0;
```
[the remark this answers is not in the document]
kaizhangNV: This one is decoded from the binary stream, and it's used for tracking the allocation purpose, we don't want to cast this to any other type.
So 64bit integer is good enough for this variable.

Note that for this kind application, since the capture file could hand over platforms, e.g. captured on windows, and want to replay on Linux, I always try my best to use explicit non-os-dependent data type (e.g. uint{#bits}_t) to store to the file.

source/slang-capture-replay/replay/file-processor.cpp
kaizhangNV: I cannot think of we can do here if we're failing to open the file?

source/slang-capture-replay/replay/parameter-decoder.cpp
```diff
@@ -0,0 +1,215 @@
+#include <string.h>
+#include "parameter-decoder.h"
+
+namespace SlangCapture
+{
+    size_t ParameterDecoder::decodeString(const uint8_t* buffer, int64_t bufferSize, PointerDecoder<char*>& typeDecoder)
+    {
+        SLANG_CAPTURE_ASSERT((buffer != nullptr) && (bufferSize > 0));
```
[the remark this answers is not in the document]
kaizhangNV: We could, but I don't want to do it in this patch.
Maybe a later change after function complete.

source/slang-capture-replay/replay/file-processor.cpp
jkwak-work: Right, you cannot do anything in the constructor.
I like to suggest that you make a dedicated method for opening a file like
```
bool OpenFile(const std::string &filename);
```

jkwak-work:
> > I have some comments. And the meaning of `decode` and `encode` is unclear to me.
> 
> `encode` is to encode the APIs and parameters to a specific format in binary. `decode` is to reconstruct a binary stream with specific format to the API call.

Thanks for the explanation.
It sounds like "encode" could be better with "capture" and "decode" with "replay"?

kaizhangNV:
> > > I have some comments. And the meaning of `decode` and `encode` is unclear to me.
> > 
> > 
> > `encode` is to encode the APIs and parameters to a specific format in binary. `decode` is to reconstruct a binary stream with specific format to the API call.
> 
> Thanks for the explanation. It sounds like "encode" could be better with "capture" and "decode" with "replay"?

The design is layered.
"encoder" module is only responsible for encoding the API and parameters block.
"capture" module is responsible for intercept the API, hook our own interface, and hand over to actual underlying interface.

While on the replay side:

"decode" module is responsible for decoding each block of the capture data into structure data structure (API + parameter block).

"replay" module is the one of the consumers that consumes the decoded data. Currently in this change, we have 2 consumers:
replayer and Json.

Those modules are de-coupled and should not be mixed together.

In addition, I design the consumer as an abstract, so in the future we might be able to add more consumers if necessary (e.g. validation consumer such that we can validate each API call.)

I feel that I would create a detailed design doc for what I have implemented here. 
In the draft design, I didn't mention about the detailed implement diagram.

jkwak-work:
> > > > I have some comments. And the meaning of `decode` and `encode` is unclear to me.
> > > 
> > > 
> > > `encode` is to encode the APIs and parameters to a specific format in binary. `decode` is to reconstruct a binary stream with specific format to the API call.
> > 
> > 
> > Thanks for the explanation. It sounds like "encode" could be better with "capture" and "decode" with "replay"?
> 
> The design is layered. "encoder" module is only responsible for encoding the API and parameters block. "capture" module is responsible for intercept the API, hook our own interface, and hand over to actual underlying interface.
> 
> While on the replay side:
> 
> "decode" module is responsible for decoding each block of the capture data into structure data structure (API + parameter block).
> 
> "replay" module is the one of the consumers that consumes the decoded data. Currently in this change, we have 2 consumers: replayer and Json.
> 
> Those modules are de-coupled and should not be mixed together.
> 
> In addition, I design the consumer as an abstract, so in the future we might be able to add more consumers if necessary (e.g. validation consumer such that we can validate each API call.)
> 
> I feel that I would create a detailed design doc for what I have implemented here. In the draft design, I didn't mention about the detailed implement diagram.

I don't think the problem is a lack of documentation. If you use a proper naming in the code, it can be a efficient way to communicate. The words "encoding/decoding" are too generic and they don't seem right words for what you are doing.

If capture/replay are taken for other uses, I suggest marshaling or serialization for encoding.

source/slang-capture-replay/replay/decoder-helper.cpp
kaizhangNV: act

source/slang-capture-replay/replay/file-processor.h
kaizhangNV: act

source/slang-capture-replay/replay/decoder-consumer.h
kaizhangNV: act

source/slang-capture-replay/replay/decoder-consumer.h
kaizhangNV: act

kaizhangNV:
> > > > > I have some comments. And the meaning of `decode` and `encode` is unclear to me.
> > > > 
> > > > 
> > > > `encode` is to encode the APIs and parameters to a specific format in binary. `decode` is to reconstruct a binary stream with specific format to the API call.
> > > 
> > > 
> > > Thanks for the explanation. It sounds like "encode" could be better with "capture" and "decode" with "replay"?
> > 
> > 
> > The design is layered. "encoder" module is only responsible for encoding the API and parameters block. "capture" module is responsible for intercept the API, hook our own interface, and hand over to actual underlying interface.
> > While on the replay side:
> > "decode" module is responsible for decoding each block of the capture data into structure data structure (API + parameter block).
> > "replay" module is the one of the consumers that consumes the decoded data. Currently in this change, we have 2 consumers: replayer and Json.
> > Those modules are de-coupled and should not be mixed together.
> > In addition, I design the consumer as an abstract, so in the future we might be able to add more consumers if necessary (e.g. validation consumer such that we can validate each API call.)
> > I feel that I would create a detailed design doc for what I have implemented here. In the draft design, I didn't mention about the detailed implement diagram.
> 
> I don't think the problem is a lack of documentation. If you use a proper naming in the code, it can be a efficient way to communicate. The words "encoding/decoding" are too generic and they don't seem right words for what you are doing.
> 
> If capture/replay are taken for other uses, I suggest marshaling or serialization for encoding.

I don't understand. I have never used only "encode" or "decode" in the code for the class name or variable name. I gave the name for function encode and parameter encode.

I don't want to explain here for the difference between marsh/serialize and encoding, you can google it, but what I have done here is not just serialization.

csyonghe (approved):

Issue #4621, opened 2024-07-12
Slang-test crashes when initializing associated types inline


venkataram-nv:
This pertains running `slang-test` on a single file, namely `tests/compute/assoctype-nested-lookup.slang`. Specifically, this occurs when changing line 18 from:
```
    T.Bar b;
```
to:
```
    T.Bar b = T.Bar();
```

There seems to be no problems when compiling with `slangc` (both on Release and Debug). However, running `slang-test` on this crashes for (on both configs). In the debug build, the output looks like the following:
```
Supported backends: fxc dxc glslang spirv-dis visualstudio genericcpp nvrtc llvm spirv-opt
Check gl,ogl,opengl: Supported
Check vk,vulkan: Supported
Check dx12,d3d12: Supported
Check dx11,d3d11: Supported
Check cuda: Supported
error: ERROR:
EXPECTED{{{
result code = 0
standard error = {
}
standard output = {
}
}}}
ACTUAL{{{
result code = 1
standard error = {
}
standard output = {
assert failure: index < getOperandCount()}
}}}

error: ERROR:
EXPECTED{{{
type: float
3.000000
}}}
ACTUAL{{{
}}}

FAILED test: 'tests/compute/assoctype-nested-lookup.slang (dx11)'
passed test: 'tests/compute/assoctype-nested-lookup.slang.1 (vk)'
error: ERROR:
EXPECTED{{{
result code = 0
standard error = {
}
standard output = {
}
}}}
ACTUAL{{{
result code = 1
standard error = {
}
standard output = {
assert failure: index < getOperandCount()}
}}}

error: ERROR:
EXPECTED{{{
type: float
3.000000
}}}
ACTUAL{{{
}}}

FAILED test: 'tests/compute/assoctype-nested-lookup.slang.2 syn (dx12)'

===
33% of tests passed (1/3)
===

failing tests:
---
tests/compute/assoctype-nested-lookup.slang (dx11)
tests/compute/assoctype-nested-lookup.slang.2 syn (dx12)
---
```

Issue #4577, opened 2024-07-09
Warnings for using uninitialized values in functions


venkataram-nv:
Continuation of #3832

There should be an exception for functions which pass the variables as an `out/inout/__ref` parameter.

PR #4626, opened 2024-07-12
Warnings function parameters

venkataram-nv:
Addresses #4577

Handles warnings for usage of uninitialized variables through functions. The changes primarily include:
- [x] Treating IRVars as (initialy) undefined variables
- [x] Treating functions taking the value as `inout` or `out` or `__ref` as stores
- [x] Interface/generic functions by the same token
- [x] Mutable methods as stores

The previously monolithic test file `tests/diagnostics/uninitialized.slang` has now been split into multiple files, each testing separate parts of the diagnostics.

DRAFT STATE: Passes all tests except for `tests/compute/assoctype-nested-lookup.slang` which is the subject of #4621. ~~Some temporary fixes were taken to resolve some of the tests, and these will be readdressed when the implicit/default initialization interfaces have been ironed out.~~


tests/autodiff/material2/MxWeights.slang
```diff
@@ -3,6 +3,11 @@
 public struct MxWeights<let TBsdfCount : int>
 {
     public float3 weights[TBsdfCount];
+
+    public __init()
+    {
+        weights[0] = float3(0.0f);
```
jkwak-work, line 9: Shouldn't you do for-loop?
```
for (int i = 0; i < TBsdCount; i++)
weights[i] = float3(0);
```

tests/bugs/generic-param-cast.slang
```diff
@@ -10,7 +10,9 @@ struct A<let I : int>
 
 struct B<let U : uint>
 {
-    A<U> a;
+    // TODO: recursive check for non-empty fields
+    // e.g. struct of non-empty structs
+    A<U> a = A<U>();
```
jkwak-work, line 15: I am not sure why this is needed when struct A has no members.

tests/bugs/specialize-existential-in-generic.slang
```diff
@@ -20,7 +20,7 @@ struct Impl : IFoo
 
 struct GenType<T : IFoo>
 {
-    T obj;
+    T obj = T();
```
jkwak-work, line 23: Does this call the constructor of IFoo? Or the constructor of the actual type that implements IFoo?

If it is the first case, I am not sure if it will be valid to call the constructor of an interface for initialization.

If it is the second case and the actual type has an explicit constructor, should we still need to use `= T();` explicitly?

tests/autodiff/material2/MxWeights.slang
venkataram-nv: Yes, it should be. I had it like this because it was enough to silence the warnings (partial uninitialization is not checked for yet).

jkwak-work (dismissed):
Looks good to me.

jkwak-work:
Looks good to me.

jkwak-work (approved):

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -146,8 +173,54 @@ namespace Slang
...
+        IRFunc* ftn = nullptr;
+        IRFuncType* ftype = nullptr;
+        if (auto spec = as<IRSpecialize>(callee))
+            ftn = as<IRFunc>(resolveSpecialization(spec));
+        else if (auto fwd = as<IRForwardDifferentiate>(callee))
+            ftn = as<IRFunc>(fwd->getBaseFn());
+        else if (auto rev = as<IRBackwardDifferentiate>(callee))
+            ftn = as<IRFunc>(rev->getBaseFn());
+        else if (auto wit = as<IRLookupWitnessMethod>(callee))
+            ftype = as<IRFuncType>(callee->getFullType());
```
jkwak-work, line 191: Is this correct to use `callee` not `wit`?
```
ftype = as<IRFuncType>(wit->getFullType());
```

jkwak-work: The reason why I am asking is that `wit` is unused and the compiler prints a warning.

PR #4628, opened 2024-07-12
Move if_let syntax to convenience-features section

kaizhangNV:


docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
 }
 ```
 
+---
+layout: user-guide
+permalink: /user-guide/convenience-features-if-let
+---
```
jkwak-work, line 389: What is these lines for?
I thought they appear only at the beginning of the documents once.

docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
...
+}
+
+void test(IFoo foo)
+{
+    // This syntax will be desugared to the following:
+    // {
+    //      Optional<MyImpl1> $OptVar = foo as MyImpl1;
+    //      if ($OptVar.hasValue)
+    //      {
+    //          MyImpl1 t = $OptVar.value;
```
jkwak-work, line 422: I am not sure what `$` sign is for.
Shouldn't you replace `$OptVar` to `optVar`?

docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
...
+}
+
+void main()
+{
+    MyImpl1 v1;
+    test(v1);
+
+    MyImpl2 v2;
+    test(v2);
+}
```
jkwak-work, line 446: I think we have have an actual implementation of the code as a test case under `tests/language-feature/if_let.slang` to prevent a regression.

jkwak-work (dismissed):
Looks good to me.

I understand that this is just moving the existing text to another location but I have some questions for the content.

docs/user-guide/03-convenience-features.md
csyonghe: This is incorrect and we should delete these lines.

docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
 }
 ```
 
+---
+layout: user-guide
+permalink: /user-guide/convenience-features-if-let
+---
+## `if_let` syntax
+Slang supports `if (let ...)` syntax to simplify the code when checking the type of a value. The syntax is similar to Rust's
```
csyonghe, line 391: "when checking the type of a value" is not what it really is here.
should be "when working with an `Optional<T>` value".

csyonghe: should be `if (let name = expr)`

docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
...
+`if let` syntax, the variable defined by `let` must be an `Optional<T>` type, for example:
```
csyonghe, line 392: "the value expression must be an ...`

docs/user-guide/03-convenience-features.md
csyonghe: Let's just use `optVar`.

docs/user-guide/03-convenience-features.md
```diff
@@ -383,6 +383,70 @@ int caller()
...
+{
+    void foo() { printf("MyImpl2");}
+}
+
+struct MyImpl3 : IFoo
+{
+    void foo() { printf("MyImpl3");}
+}
+
+void test(IFoo foo)
```
csyonghe, line 415: The example here should not use `as` operator since we haven't talked about it at this point of the documentation.

We can simply use:
```
Optional<int> getOptInt() { ... }

void test()
{
    if (let x = getOptInt())
    {
          // if we are here, `getOptInt` returns a value `int`.
          // and `x` represents the `int` value.
    }
}
```

docs/user-guide/06-interfaces-generics.md
```diff
@@ -677,63 +677,10 @@ T compute<T>(T a1, T a2)
 // compute(3, 1) == 2
 ```
 
-`as` operator can also be used in the `if` predicate to test if an object can be casted to a specific type, once the cast test is successful,
-the object can be used in the `if` block as the casted type without the need to retrieve the `Optional<T>::value` property:
-```csharp
-interface IFoo
```
csyonghe, line 683: We can keep the example here.

docs/user-guide/03-convenience-features.md
kaizhangNV: So you say this is wrong anchor?
I want to reference it in another chapter at here:
https://github.com/shader-slang/slang/pull/4628/files#diff-eb730bcc516fb44ca0442cb2f58e4197739f09f352ab72bb1bec7a88dfe1e1daR682

kaizhangNV: I figure out the correct way to add anchor.

docs/user-guide/03-convenience-features.md
kaizhangNV: We have some unit tests (both positive and negatibe) already for this, please check `tests/language-feature/if-let`

docs/user-guide/03-convenience-features.md
kaizhangNV: done

docs/user-guide/03-convenience-features.md
kaizhangNV: act

docs/user-guide/03-convenience-features.md
kaizhangNV: act

docs/user-guide/06-interfaces-generics.md
kaizhangNV: act

docs/user-guide/03-convenience-features.md
kaizhangNV: act

csyonghe (dismissed):

docs/user-guide/06-interfaces-generics.md
```diff
@@ -704,8 +706,8 @@ void test(IFoo foo)
 {
     // This syntax will be desugared to the following:
     // {
-    //      Optional<MyImpl1> $OptVar = foo as MyImpl1;
-    //      if ($OptVar.hasValue)
+    //      Optional<MyImpl1> optVar = foo as MyImpl1;
+    //      if (optVar.hasValue)
     //      {
     //          MyImpl1 t = $OptVar.value;
```
jkwak-work, line 712: This line still have `$OptVar`

kaizhangNV: ACT.

csyonghe (approved):

PR #4632, opened 2024-07-15
Always run autodiff-decoration-stripping-pass

ArielG-NV:
1. Always run autodiff-decoration-stripping-pass (removes temp autodiff decorations that break slang backend-emitting).
2. Do not start autodiff pass if MemberDerivative is found since autodiff will always run in this case (bad for compiler performance).


jkwak-work:
Is there an issue number for this?

ArielG-NV:
> Is there an issue number for this?

(This is a follow-up)[https://github.com/shader-slang/slang/pull/4623#issuecomment-2226372478] to revise code which was merged.

csyonghe (approved):

Issue #4356, opened 2024-06-12
Use ExplicitLod instructions for texture sampling in vertex shader


jkwak-work:
**Problem Description**
Because there is no way to implicitly calculate LOD for texture sampling in the vertex shader, it causes a validation error when "ImplicitLod" variants of the texture instructions are used in vertex shader.

**Goal**
We should use "ExplicitLod" variants in vertex shader with LOD value zero.

**Repro steps**
The following shader code is from CTS test, "dEQP-VK.glsl.texture_functions.textureproj.sampler2d_vec4_fixed_vertex".
```
//TEST:SIMPLE:-target spirv-asm -stage vertex -entry main -allow-glsl

#version 450 core
layout(location = 0) in highp vec4 a_position;
layout(location = 4) in highp vec4 a_in0;
layout(location = 0) out mediump vec4 v_color;
layout(set = 0, binding = 0) uniform highp sampler2D u_sampler;
layout(set = 0, binding = 1) uniform buf0 { highp vec4 u_scale; };
layout(set = 0, binding = 2) uniform buf1 { highp vec4 u_bias; };
out gl_PerVertex {
        vec4 gl_Position;
};

void main()
{
        gl_Position = a_position;
        v_color = vec4(textureProj(u_sampler, a_in0))*u_scale + u_bias;
}
```
When ran, the following validation error message is observed,
```
error: line 60: OpEntryPoint Entry Point <id> '2[%main]'s callgraph contains function <id> '2[%main]', which cannot be used with the current execution model:
ImplicitLod instructions require Fragment or GLCompute execution model: ImageSampleProjImplicitLod
```

csyonghe:
Supporting this is not trivial and require infrastructural work.

For the intrinsic definition, we should be able to use `__target_switch` to switch to different implementation based on `stage`:
```
float4 texture(...)
{
      __target_switch
      {
      case spirv:
             __target_switch
             {
              case vertex:
                    // use explicit lod
              case fragment:
                   // use implicit lod
              ...

              }

       }
}
```

The tricky part is when handling modules that has both vertex and fragment shaders:
```
[shader("vertex")]
func vert() { texture(...); }

[shader("fragment")]
func frag() { texture(...); }
```

When compiling this file to spirv, we  need to create two specializations of `texture` and then specialize them differently based on which stage they are for.

[...]

PR #4641, opened 2024-07-16
Use slang-glslang.dll for spirv-validation

### LINKED ISSUE #4610: Use spirv-val.exe from external/slang-binaries
**Problem description**
It appears that `external/slang-binaries/spirv-tools/windows-x64/bin/spirv-val.exe` is currently not in use.
In fact, none of spirv-XXX executables under `external/slang-binaries/spirv-tools` is used anywhere.

Currently slang.dll uses "spirv-val.exe" and "spirv-dis.exe" found from `Path` setting.
This is probably not an expected behavior, and we should prefer the prebuilt executables over the executables installed on the system.

Because of this problem, the developers of slang are required to install the latest version of Vulkan SDK all the time.
And more seriously, we cannot use features that are available on the ToT of spirv-tools if the features are not available with a latest official release binary yet.

**Goal**
Slang should prefer `spirv-val.exe` and `spirv-dis.exe` from `external/slang-binaries` directory over the executables installed in the system.

We had a similar problem for [the asset files for examples](https://github.com/shader-slang/slang/pull/4518).
The resolution for the problem was to search the assets in the current directory first and check the parent directory.

Similarly we should look for spirv-tools executables in the following places in the given order,
1. "spirv-tools" directory relative to where slang.dll exists.
2. "../spirv-tools" directory relative to where slang.dll exists.
3. use `Path` environment variable.

To make this work, cmake will need to copy the relevant spirv-tools executables from `external/slang-binaries/spirv-tools` to a new directory, `build/spirv-tools`.

**More details**
The logic on Slang.dll side is pretty simple and it can be found from [here](https://github.com/shader-slang/slang/blob/6ed7e7f60b3e60be2589b2aa7e812f7315493db5/source/slang/slang-emit-spirv.cpp#L6580)

1. check if an environment variable, "LANG_RUN_SPIRV_VALIDATION", is set or not.
2. check if "--skip-spirv-validation" is set.
4. when the step 1 is true and the step 2 is false, it runs "spirv-val XXX" command. It completely relies on `Path` setting to find where "spirv-val" is.
5. if the validation fails, prints an error message, "internal error 99999: Validation of generated SPIR-V failed."

A change should be made at the step 3 and it should search for the executable in a certain way that prefers `external/slang-binaries`.

Alternatively, we could update the document and require the users to change `Path` to include `external/slang-binaries...`
But that seems unnecessarily invasive to the user environment.

jkwak-work:
This change replaces the use of "spirv-val.exe" with an API call to "spvtools::SpirvTools::Validate()".

Closes #4610


jkwak-work:
As a testing, I tried a known validation failure case (#4356), and a validation failure is observed.

jkwak-work:
I made a mistake and this PR is closed.
I had to create a new PR with the same changes. #4642

PR #4644, opened 2024-07-16
Fix gfx/metal uniform data binding and memory leaks.

### LINKED ISSUE #4545: Metal GFX backend: cbuffer/uniform related syntax for `slang-test` is not working as expected
* tests/compute/entry-point-uniform-params.slang --> returns 0 in all cases, only possible if shader fails, or if data is not uploaded correctly. Uses uniforms.
* tests/hlsl/packoffset.slang --> returns 0 in all cases, only possible if shader fails, or if data is not uploaded correctly. Uses `set OBJECT` syntax.

csyonghe:
The cause of the test correctness issue in the offset computation when binding a shader object as a constant buffer leads to double counting of index offsets when a default constant buffer needs to be created for the entrypoint object.

The problem of the memory leak issue is that PipelineState and ShaderProgram should not store a storng reference back to device, because a device contains a pipeline state cache that stores a reference to all currently created pipeline states. Having PipelineState reference back a device leads to a cyclic reference preventing destruction of all objects.

Fixes #4545.

### DIFF
--- a/tools/gfx/d3d11/d3d11-shader-object.cpp
+++ b/tools/gfx/d3d11/d3d11-shader-object.cpp
@@ -318,12 +318,6 @@ Result ShaderObjectImpl::bindAsConstantBuffer(
     // buffer for any "ordinary" data in `X`, and then bind the remaining
     // resources and sub-objects.
     //
-    // The one important detail to keep track of its that *if* we bind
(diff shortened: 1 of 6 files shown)


jkwak-work (approved):
Looks good to me.

PR #4646, opened 2024-07-16
Add slang-llvm and slang-glslang readmes

expipiplus1:
Just extraced and updated from the original repos


source/slang-llvm/README.md
```diff
@@ -0,0 +1,30 @@
...
+Limitiations
+============
+ 
+* Only supports `host-callable`
+
+Building LLVM/Clang
+===================
+
+This repo's `external/build-llvm.sh` script builds llvm with the correct
+options to be used by slang, please refer to that.
```
kaizhangNV, line 30: Any license info needs to be mentioned here?

kaizhangNV (dismissed):
LGTM.
Only a nit mentioned in comment below.

source/slang-llvm/README.md
expipiplus1: I think that's dealt with in the top level readme :)

PR #4649, opened 2024-07-16
Remove generated file from source and build at build time

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/4650


expipiplus1:
> `prelude/slang-cpp-host-prelude.h.cpp`.

I don't think these files have ever existed in the repo, are they left over from when you used to build with premake?

Issue #4654, opened 2024-07-16
Metal: Metal fails to compile `ByteAddressBuffer` correctly with use of `getEquivalentStructuredBuffer`


ArielG-NV:
Problem:
Seems that in IR `ByteAddressBuffer` duplicates after the `legalizeByteAddressBufferOps` pass. This means that we end up with 2 variables bound to the same location, SPIRV/GLSL handles this duplicate through a `eliminateDeadCode` pass deleting the duplicate. Metal does not delete the duplicate.

Issue #4481, opened 2024-06-25
Metal: atomic instructions for buffers


jkwak-work:
**Problem description**
Most of the atomic instructions are implemented for Metal by PR #4473.
But the atomic member functions for buffers are not implemented.

The current implementation piggy-back on how Spir-v does, but it doesn't actually work.
```
struct RWByteAddressBuffer
{
    void InterlockedAdd(UINT dest, UINT value)
    {
        __target_switch
        {
        case metal:
        case spirv:
            let buf = __getEquivalentStructuredBuffer<uint>(this); // <=== DON'T WORK FOR METAL
            ::InterlockedAdd(buf[dest / 4], value);
        }
    }
}
```

**Goal**
We need to make the atomic member functions working for the buffers.

**Repro steps**
We have atomic test for buffer for HLSL, "tests/hlsl-intrinsic/byte-address-buffer-atomics.slang".
It can be copied over to "tests/metal/atomic-buffer.slang" and make it working.

But for a simple repro, the following code can be used for a test.
```
//TEST:SIMPLE: -entry computeMain -stage compute -target metal
//TEST:SIMPLE: -entry computeMain -stage compute -target metallib

RWByteAddressBuffer bbuffer;

[numthreads(1, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    bbuffer.InterlockedAdd(0, 1);
}
```

jkwak-work:
The problem seems to be on the fact that Metal doesn't allow us to obtain a memory pointer access to the buffer.
https://github.com/shader-slang/slang/pull/4473#issuecomment-2190388723

A proper solution would be to extend our `legalizeImageSubscript()`.

csyonghe:
A metal buffer is already represented as a pointer. We should be able to easily emit getEquivalentStructuredBuffer as a C style cast.

PR #4655, opened 2024-07-16
Metal: `Interlocked` (atomic) member function support for buffers

ArielG-NV:
fixes: #4654
fixes: #4481

1. Add `Interlocked` (atomic) member function support for buffers to Metal
2. Fix `__getEquivalentStructuredBuffer` so it works with Metal targets


source/slang/slang-ir-byte-address-legalize.cpp
```diff
@@ -92,6 +92,11 @@ struct ByteAddressBufferLegalizationContext
         // Get the equivalent structured buffer for the buffer.
         if( auto structuredBuffer = getEquivalentStructuredBuffer(elementType, buffer) )
         {
+            if (buffer != structuredBuffer)
+            {
+                buffer->replaceUsesWith(structuredBuffer);
+                buffer->removeAndDeallocate();
+            }
```
jkwak-work, line 99: I am trying to understand why this is needed.
Can you add a comment for this block so that other people may not accidently remove it in a long term?

jkwak-work (approved):
Looks good to me.

source/slang/hlsl.meta.slang
```diff
@@ -9335,7 +9378,7 @@ for(const char* T : {"int64_t", "uint64_t"})
 {
 }}}}
 [ForceInline]
-[require(cuda_glsl_hlsl_spirv, atomic_glsl_hlsl_cuda)]
+[require(cuda_glsl_hlsl_spirv, atomic_glsl_hlsl_cuda_metal)]
```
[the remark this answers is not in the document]
ArielG-NV: It should not be a capability error.

The reason why is because of how the capability system resolves capabilities:
`atomic_glsl_hlsl_cuda_metal` has the targets `cuda` `glsl` `hlsl` `metal` `spirv`
`cuda_glsl_hlsl_spirv` has the target `cuda` `glsl` `hlsl` `spirv`

`cuda_glsl_hlsl_spirv + cuda_glsl_hlsl_metal_spirv` is equivalent to `cuda_glsl_hlsl_spirv`. As a result, the `metal` target is not a requirement of the function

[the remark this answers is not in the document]
jkwak-work: Thank you for the explanation. But I am still having a trouble understanding it.
Would you mind explain little more of how `+` sign works? I thought [require(xx,yy)] means "xx && yy" where `&&` means both conditions should be true. I was reading it as "the function requires the target to be (cuda || glsl || hlsl || spirv) && (atomic_cuda || atomic_glsl || atomic_hlsl || atomic_metal || atomic_spirv)", but I guess I am reading it wrong.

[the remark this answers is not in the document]
ArielG-NV: '+' is similar to `&&`. What has to be remembered when evaluating is that we discard invalid sets. Note that, valid sets must only have 1 key (abstract-atom-of-type) fill a keyhole (abstract-atom-type) per set.

What this means is `(cuda || glsl || hlsl || spirv) && (atomic_cuda || atomic_glsl || atomic_hlsl || atomic_metal || atomic_spirv)` is a correct representation.
The way we evaluate this expression can be thought of in the following way:
1. `(cuda || glsl || hlsl || spirv) && (atomic_cuda || atomic_glsl || atomic_hlsl || atomic_metal || atomic_spirv)` is equivalent to the expression of `(cuda && atomic_cuda || cuda + && atomic_glsl || cuda + && atomic_hlsl || cuda + && atomic_metal || cuda + && atomic_spirv || glsl + atomic_cuda ||  glsl + && atomic_glsl || glsl + && atomic_hlsl || ...)
2. Since invalid sets are discarded, we reflect this when simplifying the expression: `(cuda + atomic_cuda || glsl + atomic_glsl || hlsl + atomic_hlsl || spirv + atomic_spirv)`
3. It is important to notice that `cuda + atomic_metal`, `glsl + atomic_metal`, `hlsl + atomic_metal`, `spirv + atomic_metal` are all invalid, so we discard these sets.

source/slang/slang-emit.cpp
```diff
@@ -1039,6 +1039,7 @@ Result linkAndOptimizeIR(
         case CodeGenTarget::MetalLib:
         case CodeGenTarget::MetalLibAssembly:
             byteAddressBufferOptions.scalarizeVectorLoadStore = true;
+            byteAddressBufferOptions.treatGetEquivalentStructuredBufferAsGetThis = true;
```
csyonghe, line 1042: What does treatGetEquivalentBufferAsGetThis do?

ArielG-NV: `treatGetEquivalentStructuredBufferAsGetThis` causes all calls to `getEquivlentStructuredBuffer` to return a byteAddressBuffer (`this`) instead of a StructuredBuffer. 

This works since Metal does not care for the difference of these types (from what I understand).

The problem this solves is that the change allows legalizing and removing all calls to `getEquivlentStructuredBuffer` without introducing another duplicate buffer that DCE won't clean up.

csyonghe: Can you add a comment to explain this?

csyonghe (approved):

PR #4658, opened 2024-07-16
Fix LLVM compile error on Windows

jkwak-work:
Currently our CI doesn't cover the case of building with LLVM on windows.

The following compile error is observed when build with LLVM on Windowsanually:
```
D:/sbf/git/slang/test_wsl/source/slang-glslang/slang-glslang.cpp:39:31: error: static assertion expression is not an integral constant expression
   39 |     SLANG_COMPILE_TIME_ASSERT(SLANG_OFFSET_OF(TBuiltInResource, limits) > 0);
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

To address the problem, SLANG_OFFSET_OF is defined as `__buildint_offsetof` when `__llvm__` is defined.


jkwak-work:
Currently aarch64 CI is not working.
I will hold off this PR until the change can be tested against aarch64.

include/slang.h
```diff
@@ -254,15 +254,16 @@ convention for interface methods.
...
 // GCC Specific
 #if SLANG_GCC_FAMILY
-
 #	define SLANG_NO_INLINE __attribute__((noinline))
 #	define SLANG_FORCE_INLINE inline __attribute__((always_inline))
 #   define SLANG_BREAKPOINT(id) __builtin_trap();
 #	define SLANG_ALIGN_OF(T)	__alignof__(T)
+#endif // SLANG_GCC_FAMILY
 
+#if SLANG_GCC_FAMILY || defined(__llvm__)
```
ArielG-NV, line 263: `__builtin_offsetof` should also work with MSVC, should we be adding it as-well to avoid compiler hacks?
godbolt link: https://godbolt.org/z/cGfWfz3nP

```suggestion
#if SLANG_GCC_FAMILY || defined(__llvm__) || defined(_MSC_VER)
```

expipiplus1:
Is there a reason we don't just include `cstddef`?

(Also, is `__llvm__` the correct gate to use here rather than `__clang__`?)

include/slang.h
jkwak-work: I am reluctant to adopt the suggested change because there isn't any problem with the current implementation.
I am going to skip the suggested change for now and we can add it when it becomes a problem.
Personally I am more comfortable using the `((T*)0)->elem` workaround, because that's what I have been doing on almost all projects in the past, if not all.

jkwak-work:
> Is there a reason we don't just include `cstddef`?
> 
> (Also, is `__llvm__` the correct gate to use here rather than `__clang__`?)

I don't know why we want to avoid including stddef, but we are doing it.
`__clang__` is a right way to do it; thanks for pointing it out.

csyonghe (approved):

PR #4659, opened 2024-07-16
Enable warnings-as-error for CI

### LINKED ISSUE #4664: Enable warnings-as-error on CI tests
With recent changes with CMake, most of CI tests started running without "warnings-as-error".
We should re-enable it to prevent any warnings and potential bugs.

jkwak-work:
Closes #4664

A warning of "unused-function" is exempted from warnings-as-error; only on macOS.
The following function was causing the warning and it make sense to keep it as it.
```
static inline SlangReflectionFunction* convert(FunctionDeclBase* func)
```


jkwak-work:
Currently CI test for aarch64 is not working.
I am going to hold off this PR until it is tested on aarch64.

.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           - { os: macos, runs-on: macos-latest }
           - { os: windows, runs-on: windows-latest }
           # When to have warnings
-          - warnings-as-errors: false
+          - warnings-as-errors: true
           - { compiler: cl, warnings-as-errors true }
```
expipiplus1, line 37: this line is redundant now

expipiplus1 (changes requested):
We can just remove `warnings-as-errors` now if they're all going to be `true`

Maybe consider changing these files too

```
.github/workflows/falcor-test.yml
.github/workflows/compile-regression-test.yml
.github/workflows/falcor-compiler-perf-test.yml
.github/workflows/vk-gl-cts-nightly.yml
```

source/slang/CMakeLists.txt
```diff
@@ -274,6 +274,10 @@ slang_add_target(
     PUBLIC_HEADERS ${slang_SOURCE_DIR}/slang*.h ${CMAKE_CURRENT_BINARY_DIR}/slang-version-header/*.h
 )
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
```
expipiplus1, line 277: Is this options specific to `AppleClang` or would just `Clang` be more appropriate?

expipiplus1 (changes requested):

source/slang/CMakeLists.txt
expipiplus1: I get the warning on Linux with regular clang

image

expipiplus1 (changes requested):
> Closes #4664
> 
> A warning of "unused-function" is exempted from warnings-as-error; only on macOS. The following function was causing the warning and it make sense to keep it as it.
> 
> ```
> static inline SlangReflectionFunction* convert(FunctionDeclBase* func)
> ```

Instead of blanket excluding such warnings, we should remove this function, or at least just put `[[maybe_unused]]` like we do later on: https://github.com/shader-slang/slang/blob/f545ef30052de7b4cab389cb58ffb0ea67ffb32c/source/slang/slang-reflection-api.cpp#L103

source/slang/CMakeLists.txt
jkwak-work: Interesting.
I will make it to AppleClang and Clang.
```
if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```

jkwak-work:
> We can just remove `warnings-as-errors` now if they're all going to be `true`

We may need to occasionally turn it off in the future.
If the variable is references only one time, I will be fine to hard-code it.
But since it is referred more than once at the moment, I think it is better to keep it for now.

jkwak-work:
I was able to reproduce the used-function warning with LLVM on windows.
`[[maybe_unused]]` appears to be working.

jkwak-work:
With the ToT, `SLANG_MAYBE_UNUSED` is no longer needed but I am going to keep the definition.
Because it is a relevant macro for this commit and we may need it in future.

PR #4665, opened 2024-07-16
Separating slang-test in CI for emit-spirv-via-glsl

jkwak-work:
When CI test failes, we expect to see the error messages at the end of each section of CI testing result. But currently when "emit-spirv-via-glsl" fails, the error messages appear in the middle, which lead us to believe that there is no errors when you look only at the end of the result.

This commit separates out the "emit-spirv-via-glsl" case to make the error message more visiable.


jkwak-work:
It appears that an option "-server-count 8" for slang-test was intentionally missing from Linux/macOS platforms.
I wonder if "-use-test-server" should also be removed from those platforms.

.github/workflows/ci.yml
```diff
@@ -113,20 +113,28 @@ jobs:
...
               failed=1
-          else
+          fi
+          exit $failed
+      - name: Test Slang spirv-via-glsl
+        run: |
+          export SLANG_RUN_SPIRV_VALIDATION=1
+          export SLANG_USE_SPV_SOURCE_LANGUAGE_UNKNOWN=1
+          failed=0
+          if [[ "${{matrix.full-gpu-tests}}" == "true" ]]; then
```
expipiplus1, line 130: This conditional can be raised to be an `if` field alongside `name` and `run` now

.github/workflows/ci.yml
```diff
@@ -113,20 +113,28 @@ jobs:
...
+              -expected-failure-list tests/expected-failure-github.txt ||
               failed=1
-          else
+          fi
+          exit $failed
+      - name: Test Slang spirv-via-glsl
+        run: |
+          export SLANG_RUN_SPIRV_VALIDATION=1
+          export SLANG_USE_SPV_SOURCE_LANGUAGE_UNKNOWN=1
+          failed=0
```
expipiplus1, line 129: No need for the failed variable tracking now

.github/workflows/ci.yml
```diff
@@ -113,20 +113,28 @@ jobs:
               -category ${{ matrix.test-category }} \
               -api all-cpu ||
               failed=1
```
expipiplus1, line 115: No need for the failed variable tracking now

expipiplus1 (changes requested):
lgtm aside from comments.

Although it might be neater to just synthesize the "via glsl" tests in the same way we synthesize cuda tests, that way we can dispense with having to run slang-test twice at all

.github/workflows/ci.yml
expipiplus1: I meant in the yaml having `if: matrix.full-gpu-tests == 'true'`

PR #4666, opened 2024-07-16
Initial implementation for decl-tree reflection API

### LINKED ISSUE #4617: Slang Reflection API: Add Decl-tree refelction for Python-based metaprogramming/boiler-plate generation

saipraveenb25:
This patch adds Slang API methods for walking all the declarations in the AST.

We expose this functionality through an abstract `DeclReflection` class that can be a type, function or a variable declaration. We also provide ways to cast the decl to a `FunctionReflection`, `TypeReflection` or `VariableReflection` and traverse through the child nodes (for instance, a struct type will have component variable declarations)

Additional changes:
+ Convert `FunctionReflection` to be a `DeclRef` rather than a `Decl` without specialization args. The user-facing API remains the same, so this not a breaking change.

+ Query to look for the mutating modifier (Use `findModifier(slang::Modifier::Mutating)`)

+ Query to check if a variable has a default expression

Fixes #4617


source/slang/slang-reflection-api.cpp
```diff
@@ -2843,6 +2843,98 @@ SLANG_API SlangReflectionVariable* spReflectionFunction_GetParameter(SlangReflec
...
+    }
+    else if (as<ModuleDecl>(slangDecl))
+    {
+        return SLANG_DECL_KIND_MODULE;
+    }
+    else
+        return SLANG_DECL_KIND_UNSUPPORTED_FOR_REFLECTION;
+}
+
+SLANG_API SlangReflectionFunction* spReflectionDecl_castToFunction(SlangReflectionDecl* decl)
```
csyonghe, line 2898: We should in a separate PR make `SlangReflectionFunction` a declref, and this function should just form a default decl ref to the func decl.

source/slang/slang.cpp
```diff
@@ -546,11 +546,24 @@ SlangResult Session::_readBuiltinModule(ISlangFileSystem* fileSystem, Scope* sco
...
 
-ISlangUnknown* Session::getInterface(const Guid& guid)
+SLANG_NO_THROW SlangResult SLANG_MCALL Session::queryInterface(SlangUUID const& uuid, void** outObject)
 {
-    if(guid == ISlangUnknown::getTypeGuid() || guid == IGlobalSession::getTypeGuid())
-        return asExternal(this);
-    return nullptr;
+    if (uuid == ISlangInternal::getTypeGuid())
+    {
+        // Special case to cast directly into internal type
```
csyonghe, line 553: Why not addref?

source/slang-capture-replay/slang-global-session.cpp
```diff
@@ -28,11 +28,24 @@ namespace SlangCapture
...
+    SLANG_NO_THROW SlangResult SLANG_MCALL GlobalSessionCapture::queryInterface(SlangUUID const& uuid, void** outObject) 
     {
-        if(guid == ISlangUnknown::getTypeGuid() || guid == IGlobalSession::getTypeGuid())
-            return asExternal(this);
-        return nullptr;
+        if (uuid == ISlangInternal::getTypeGuid())
+        {
+            // Special case to cast directly into internal type
+            // NOTE! No addref(!)
+            this->m_actualGlobalSession->queryInterface(uuid, outObject);
```
csyonghe, line 37: I think it should be fine to have a addref call here so that we use a consistent convention throughout the codebase.

saipraveenb25: When I had an addRef (and a release) it keepscrashing with huge memory leak dumps. 
I'm not quite sure what was going on, then I found the same workaround in `EndToEndCompileRequest::queryInterface` here (https://github.dev/shader-slang/slang/blob/f545ef30052de7b4cab389cb58ffb0ea67ffb32c/source/slang/slang.cpp#L3040) and I assumed that using ref-counting during a simple cast was causing problems with the system.

saipraveenb25: Okay, I think the problem is that we use `asInternal` in a lot of places and we get a `Session*` as a raw pointer. None of the places take care of releasing the pointer explicitly, because thus far, Session* was a simple cast with no ref-counting

Updating `asInternal` to use `ComPtr` fixes this issue

source/slang/slang.cpp
saipraveenb25: added + changed `asInternal` to use ComPtr

csyonghe (approved):

PR #4671, opened 2024-07-18
Support parameter block in metal shader objects.

csyonghe:
Implements support for ParameterBlock parameters in gfx/metal's shader object implementation.

Fix RW texture creation.


jkwak-work (approved):
Looks good to me.

I don't think I am fully following what happens in the change. But I didn't any problems.

csyonghe:
It seems that GitHub runner doesn't support metal 3/ argument buffer tier 2 so my current implementation can't run there.

I need to see if I can retrofit a metal 2 implementation, and if not I have to disable the tests on GitHub.

PR #4677, opened 2024-07-18
Fix bug with uninitiailized variable warnings in nested scopes

venkataram-nv:
Addressed to #3832 

Fixes a bug in #4530 which results in the warning system ignoring uninitialized variables in nested scopes.


jkwak-work (approved):
Looks good to me.

Issue #4578, opened 2024-07-09
Warnings for leaving constructors without initializing fields


venkataram-nv:
Continuation of #3832

If a constructor has concrete fields and does not initialize them in a constructor, there should be warnings in the appropriate locations.

PR #4680, opened 2024-07-18
Warnings for uninitialized fields in constructors

venkataram-nv:
Addresses #4578 

Adds new tests in the file `tests/diagnostics/uninitialized-fields.slang`.

Warnings will look like the following:

```
.\tests\diagnostics\uninitialized-fields.slang(22): warning 41020: exiting constructor without initializing field 'next'
    __init(int v, Simple* ptr)
    ^~~~~~
.\tests\diagnostics\uninitialized-fields.slang(39): warning 41020: exiting constructor without initializing field 'another_field'
            return;
            ^~~~~~
.\tests\diagnostics\uninitialized-fields.slang(60): warning 41021: default initializer for 'Implicit' will not initialize field 'a'
    int a;
        ^
.\tests\diagnostics\uninitialized-fields.slang(63): warning 41021: default initializer for 'Implicit' will not initialize field 'c'
    int c;
        ^
```

New decorations:
* ConstructorDecoration: indicates whether an IRFunc is a constructor
* SynthesizedDecoration: indicates whether an IRFunc was synthesized (for improved diagnostics)

### DIFF
--- a/source/slang/slang-ast-decl.h
+++ b/source/slang/slang-ast-decl.h
@@ -371,8 +371,10 @@ class FunctionDeclBase : public CallableDecl
 class ConstructorDecl : public FunctionDeclBase
 {
     SLANG_AST_CLASS(ConstructorDecl)
-    
-    bool synthesized = false;
+   
+    // Indicates whether the declaration was synthesized by
+    // slang and not actually provided by the user
+    bool isSynthesized = false;
 };
 
 // A subscript operation used to index instances of a type
--- a/source/slang/slang-ast-decl.h
+++ b/source/slang/slang-ast-decl.h
@@ -371,6 +371,8 @@ class FunctionDeclBase : public CallableDecl
 class ConstructorDecl : public FunctionDeclBase
 {
     SLANG_AST_CLASS(ConstructorDecl)
+    
+    bool synthesized = false;
 };
 
 // A subscript operation used to index instances of a type
--- a/source/slang/slang-ir-use-uninitialized-values.cpp
+++ b/source/slang/slang-ir-use-uninitialized-values.cpp
@@ -378,7 +378,7 @@ namespace Slang
 
         for (auto store : stores)
         {
-            if (reachability.isInstReachable(inst, reference))
+            if (reachability.isInstReachable(store, reference))
                 return true;
         }
 
@@ -426,7 +426,8 @@ namespace Slang
         List<IRStructField*> uninitializedFields;
 
         auto fields = type->getFields();
-        for (auto field : fields) {
+        for (auto field : fields)
+        {
             if (!isFieldUsed(field, usedKeys))
                 uninitializedFields.add(field);
         }
@@ -440,6 +441,29 @@ namespace Slang
         if (!stype)
             return;
         
+        auto synthesized = func->findDecoration<IRSynthesizedDecoration>();
+        
+        auto emitWarnings = [&](const List<IRStructField*>& fields, IRReturn* ret)
+        {
+            for (auto field : fields)
+            {
+                if (synthesized)
+                {
+                    sink->diagnose(field->getKey(),
+                        Diagnostics::fieldNotDefaultInitialized,
+                        stype,
+                        field->getKey());
+                }
+                else
+                {
+                    sink->diagnose(ret,
+                        Diagnostics::constructorUninitializedField,
+                        field->getKey());
+                }
+            }
+
+        };
+
         // Work backwards, get exit points and find sources
         for (auto block : func->getBlocks())
         {
@@ -450,12 +474,7 @@ namespace Slang
                     continue;
 
                 auto fields = checkFieldsFromExit(reachability, ret, stype);
-                for (auto field : fields)
-                {
-                    sink->diagnose(ret,
-                        Diagnostics::constructorUninitializedField,
-                        field->getKey());
-                }
+                emitWarnings(fields, ret);
             }
         }
     }
@@ -474,7 +493,7 @@ namespace Slang
         ReachabilityContext reachability(func);
 
         // Used for a further analysis and to skip usual return checks
-        bool constructor = func->findDecoration <IRConstructorDecorartion> ();
+        auto constructor = func->findDecoration <IRConstructorDecorartion> ();
 
         // Check out parameters
         for (auto param : firstBlock->getParams())
--- a/source/slang/slang-ir-use-uninitialized-values.cpp
+++ b/source/slang/slang-ir-use-uninitialized-values.cpp
@@ -360,7 +360,7 @@ namespace Slang
         return false;
     }
 
-    static bool isInstStoredInto(IRInst* inst)
+    static bool isInstStoredInto(ReachabilityContext& reachability, IRInst* reference, IRInst* inst)
     {
         auto addresses = getAliasableInstructions(inst);
         
@@ -373,9 +373,13 @@ namespace Slang
             {
                 IRInst* user = use->getUser();
                 collectLoadStore(stores, loads, user, alias);
-                if (stores.getCount())
-                    return true;
             }
+        }
+
+        for (auto store : stores)
+        {
+            if (reachability.isInstReachable(inst, reference))
+                return true;
         }
 
         return false;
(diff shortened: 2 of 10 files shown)


tests/diagnostics/uninitialized-fields.slang
```diff
@@ -0,0 +1,92 @@
...
+
+    __init()
+    {
+        field = 0;
+        if (flag == 0)
+            //CHK-DAG: warning 41020: exiting constructor without initializing field 'another_field'
+            return;
+
+        another_field = 0;
+    }
```
jkwak-work, line 42: I am not sure if a return is inserted at the end of the constructor.
Maybe we should have another case to test it like this,
```
    __init()
    {
        field = 0;
        if (flag == 0)
        {
            another_field = 0;
            return;
        }

        //CHK-DAG: warning 41020: exiting constructor without initializing field 'another_field'
    }
```

jkwak-work: I am not sure if the following case will print a warning,
```
    __init()
    {
        field = 0;
        if (flag == 0)
        {
            another_field = 0;
        }

        return;
    }
```

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -348,8 +348,135 @@ namespace Slang
...
+            }
+
+        };
+
+        // Work backwards, get exit points and find sources
+        for (auto block : func->getBlocks())
+        {
+            for (auto inst = block->getFirstInst(); inst; inst = inst->next)
+            {
+                auto ret = as<IRReturn>(inst);
```
jkwak-work, line 466: You may want to use `TerminatorInst` like
```
auto ret = as<TerminatorInst>(inst);
```

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -348,8 +348,135 @@ namespace Slang
...
+
+    static void checkConstructor(IRFunc* func, ReachabilityContext& reachability, DiagnosticSink* sink)
+    {
+        IRStructType* stype = as<IRStructType>(func->getResultType());
+        if (!stype)
+            return;
+        
+        auto synthesized = func->findDecoration<IRSynthesizedDecoration>();
+        
+        auto emitWarnings = [&](const List<IRStructField*>& fields, IRReturn* ret)
```
jkwak-work, line 440: I don't think we want to use `emit` for printing warnings.
You may want to reword it to `printWarning`, `printDiagonostics` or something like that.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -348,8 +348,135 @@ namespace Slang
...
+    static List<IRStructField*> checkFieldsFromExit(ReachabilityContext& reachability, IRReturn* ret, IRStructType* type)
+    {
+        IRInst* origin = traceInstOrigin(ret->getVal());
+
+        // We don't want to warn on delegated construction
+        if (!isUninitializedValue(origin))
+            return {};
+
+        // Now we can look for all references to fields
+        List<IRStructKey*> usedKeys;
```
jkwak-work, line 404: You may want to use HashSet type for this variable.
If the user has a struct with hundreds members, it can make a big difference.

jkwak-work (approved):
Looks good to me.

source/slang/slang-ir-use-uninitialized-values.cpp
venkataram-nv: Thanks, I was looking for a data structure to use here instead.

source/slang/slang-ir-use-uninitialized-values.cpp
venkataram-nv: Sounds good. Is lambda use alright?

tests/diagnostics/uninitialized-fields.slang
venkataram-nv: It seems that constructors will always end with a return, so in the first case there will be a warning that is printed. The second case does not issue a warning because we do not checking for potentially uninitialized values. We only trigger warnings if it a field will *never* be initialized by the time of next use/return;

tests/compute/struct-default-init.slang
```diff
@@ -3,9 +3,9 @@
 
 struct Test
 {
-	int a;
```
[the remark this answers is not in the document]
venkataram-nv: It currently does indeed issue a warning here. The reason is that a default constructor is still created, and within that default constructor the field `a` is not initialized. I think its worth keeping the warning; even if it is not a problem in the current file, it could be a problem when using with modules.

source/slang/slang-ast-decl.h
```diff
@@ -371,6 +371,8 @@ class FunctionDeclBase : public CallableDecl
 class ConstructorDecl : public FunctionDeclBase
 {
     SLANG_AST_CLASS(ConstructorDecl)
+    
+    bool synthesized = false;
```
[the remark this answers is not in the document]
venkataram-nv: Its meant to be like a flag that indicates if a constructor has been synthesized by slangc and not provided by the user. Its used later to handle implicitly created default constructors.

source/slang/slang-ir-use-uninitialized-values.cpp
venkataram-nv: We need an explicit return so we can get the value returned (constructed) and trace is back so a `IRTerminatorInst` wouldn't be enough.

jkwak-work: There are other ways to return from the middle of the constructor.
`__intrinstic_asm` is one of them.
And they are grouped into `IRTerminatorInst`.

tests/compute/struct-default-init.slang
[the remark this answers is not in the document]
jkwak-work: I am not sure if we want to print a warning about this case,
```
struct Test
{
    int a;
};
```
C/C++ doesn't do that....

source/slang/slang-ir-use-uninitialized-values.cpp
jkwak-work: We have a rule for lambda.
```
Lambdas are allowed, but think carefully about whether just declaring a subroutine would also work.
```
https://github.com/jkwak-work/slang/blob/master/docs/design/coding-conventions.md
In general, it is better to stay away from it.
In this case, it is called only once, my preference is not to use it, but it seems fine.

source/slang/slang-ast-decl.h
```diff
@@ -371,6 +371,10 @@ class FunctionDeclBase : public CallableDecl
 class ConstructorDecl : public FunctionDeclBase
 {
     SLANG_AST_CLASS(ConstructorDecl)
+   
+    // Indicates whether the declaration was synthesized by
+    // slang and not actually provided by the user
+    bool isSynthesized = false;
```
jkwak-work, line 377: I am still little skeptical about the word "synthesize".
It is fine as it is but can you tell me how you would define the word in this context?

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -348,8 +348,139 @@ namespace Slang
...
     }
 
+    static bool isInstStoredInto(ReachabilityContext& reachability, IRInst* reference, IRInst* inst)
+    {
+        auto addresses = getAliasableInstructions(inst);
+        
+        List<IRInst*> stores;
+        List<IRInst*> loads;
+
+        for (auto alias : addresses)
```
jkwak-work, line 358: Probably easier to read if this was
```
for (auto alias : getAliasableInstructions(inst))
```

jkwak-work (approved):
Looks still good to me.

source/slang/slang-ast-decl.h
venkataram-nv: Synthesize here means a declaration generated by slang which was never actually parsed from the file.

source/slang/slang-ir-use-uninitialized-values.cpp
venkataram-nv: I see. In this case I wanted to avoid passing multiple parameters to a separate function.

tests/compute/struct-default-init.slang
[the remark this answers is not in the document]
venkataram-nv: That is true. I will see if I can make an exception for this.

jkwak-work (approved):
Looks good.

PR #4684, opened 2024-07-18
Add a proposal for new AST/IR serialization

tangent-vector:
This is intended to be a starting-point proposal/RFC document that will be discussed and iterated on to arrive at a POR that we can then use to create multiple bite-sized issues from.


docs/proposals/implementation/ast-ir-serialization.md
```diff
@@ -0,0 +1,286 @@
...
+The serialized IR for a module does *not* encode information sufficient for semantic checking of modules that `import` it.
+
+Currently, deserialization of the AST or IR for a module is an all-or-nothing operation.
+Either the entire `Decl` hierarchy of the AST is deserialized and turned into in-memory C++ objects, or none of it is.
+Similarly, we can either construct the `IRInst` hierarchy for an entire module, or none of it.
+
+Releases of the Slang compiler typically included a serialized form of the standard library ("stdlib") module, and the runtime cost of deserializing this module has proven to be a problem for users of the compiler.
+Becuse parts of the Slang compiler are not fully thread-safe/reentrant, the stdlib must be deserialized for each "global session," so that deserialization cost is incurred per-thread in scenarios with thread pools.
+Even in single-threaded scenarios, the deserialization step adds significantly to the startup time for the compiler, making single-file compiles less efficient than compiling large batches of files in a single process.
```
jkwak-work, line 23: It feels like something is missing between "Background" and "Proposed solution".
This paragraph sounds like "Problem description" or the kind.

tangent-vector: The proposed changes should be beneficial across the board for any scenarios involving linking of large compiled modules. However, the concerns are greatest around the standard library at present, and if it weren't for the stdlib we might be able to delay implementing any kind of fix for this stuff until much later.

tangent-vector: We should probably codify the expected format and sections for implementation-focused proposals like this, along the lines of what we have for design proposals.

I'd like to be able to get the raw text of this proposal checked in somewhere that the team can easily see/share it first, and then we can work to massage it into a form we are happier with over time.

csyonghe (approved):

PR #4690, opened 2024-07-18
Fix for invalid swizzle causing crash

### LINKED ISSUE #4689: AST: `CheckSwizzleExpr` can cause `Stack around the variable 'elementIndices' was corrupted` error
Problem:
`CheckSwizzleExpr` logic allows parsing a swizzle of 5+ characters, this causes an out-of-bounds array access

ArielG-NV:
Fixes #4689

If swizzle code is provided a 5+ element swizzle the `checkSwizzleExpr` code will do an out-of-bounds-array-store sometimes causing a crash (since this is undefined behavior).


source/slang/slang-check-expr.cpp
```diff
@@ -3759,6 +3759,15 @@ namespace Slang
                 continue;
             }
 
+            // If elementCount is already at 4 stop trying to assign a swizzle element and send an error,
+            // we cannot have more valid swizzle elements than 4.
+            if (elementCount == 4)
+            {
```
jkwak-work, line 3765: Shouldn't it be `>=`?

source/slang/slang-check-expr.cpp
```diff
@@ -3759,6 +3759,15 @@ namespace Slang
...
+                getSink()->diagnose(swizExpr, Diagnostics::invalidSwizzleExpr, swizzleText, baseElementType->toString());
+                anyError = true;
+                continue;
```
csyonghe, line 3768: should be `break`?

source/slang/slang-check-expr.cpp
ArielG-NV: to get to 5, we have to be 4 first.

csyonghe (approved):

PR #4691, opened 2024-07-19
Fix the issue of name mangling

kaizhangNV:
During our name mangling, we should add the direction of the parameter in the name, otherwise it could have the name collision which will result in invalid code generation:

e.g.

// in slang-module.slang
export func(float a) { ...}

// in test.slang
extern func(inout float a);

when we compile test.slang, slang will pass a pointer type to the 'func', however, in the slang-module.slang, `func` expects a value instead of pointer. This will lead the wrong spirv code.

So we should add the parameter direction into the mangle name such that above two symbols will have the different mangled names, and we will catch this during IR-link stage.

### DIFF
--- a/source/slang/slang-mangle.cpp
+++ b/source/slang/slang-mangle.cpp
@@ -554,19 +554,24 @@ namespace Slang
                 switch (paramDirection)
                 {
                 case kParameterDirection_Ref:
-                    emitRaw(context, "ref");
+                    emitRaw(context, "r_");
                     break;
                 case kParameterDirection_ConstRef:
-                    emitRaw(context, "constref");
+                    emitRaw(context, "c_");
                     break;
                 case kParameterDirection_Out:
-                    emitRaw(context, "out");
+                    emitRaw(context, "o_");
                     break;
                 case kParameterDirection_InOut:
-                    emitRaw(context, "inout");
+                    emitRaw(context, "io_");
+                    break;
+                case kParameterDirection_In:
+                    emitRaw(context, "i_");
                     break;
                 default:
-                    emitRaw(context, "in");
+                    StringBuilder errMsg;
+                    errMsg << "Unknown parameter direction: " << paramDirection;
+                    SLANG_ABORT_COMPILATION(errMsg.toString().begin());
                     break;
                 }
                 emitType(context, getType(context->astBuilder, paramDeclRef));
--- a/source/slang/slang-mangle.cpp
+++ b/source/slang/slang-mangle.cpp
@@ -547,6 +547,18 @@ namespace Slang
 
             for(auto paramDeclRef : parameters)
             {
+                // "out"/"inout" and "in" modifier makes big difference in the spirv code generation,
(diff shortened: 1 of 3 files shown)


source/slang/slang-mangle.cpp
```diff
@@ -547,6 +547,18 @@ namespace Slang
 
             for(auto paramDeclRef : parameters)
             {
+                // "out"/"inout" and "in" modifier makes big difference in the spirv code generation,
+                // because "out"/"inout" parameter will be passed by pointer. Therefore, we need to
+                // distinguish them in the mangled name to avoid name collision.
+                if(paramDeclRef.getDecl()->hasModifier<OutModifier>() ||
```
csyonghe, line 553: We need a different prefix for each different direction. There should already be a getParameterDirection function you can use. in, inout and out and ref, const ref should all get different mangled names.

kaizhangNV: done.

source/slang/slang-mangle.cpp
```diff
@@ -547,6 +547,28 @@ namespace Slang
...
             for(auto paramDeclRef : parameters)
             {
+                // parameter modifier makes big difference in the spirv code generation, for example
+                // "out"/"inout" parameter will be passed by pointer. Therefore, we need to
+                // distinguish them in the mangled name to avoid name collision.
+                ParameterDirection paramDirection = getParameterDirection(paramDeclRef.getDecl());
+                switch (paramDirection)
+                {
+                case kParameterDirection_Ref:
+                    emitRaw(context, "ref");
```
csyonghe, line 557: Let's try to make mangled names shorter by shortening those prefixes. in can be i_, inout can be io_, out can be o_, ref can be r_, const ref can be cr_.

source/slang/slang-mangle.cpp
```diff
@@ -547,6 +547,28 @@ namespace Slang
...
+                    emitRaw(context, "constref");
+                    break;
+                case kParameterDirection_Out:
+                    emitRaw(context, "out");
+                    break;
+                case kParameterDirection_InOut:
+                    emitRaw(context, "inout");
+                    break;
+                default:
+                    emitRaw(context, "in");
```
csyonghe, line 569: Should make a case for in, and default should be an abort error.

source/slang/slang-mangle.cpp
csyonghe: The purpose of mangling is to encode info in a compact identifier string. Readability is a secondary goal. We should also try to make sure the encoding scheme has a viable decoding scheme, so if needed we can parse the mangled name back to the structural representation about a function.

kaizhangNV: done

source/slang/slang-mangle.cpp
kaizhangNV: It's no harm to do this, though I don't think it will go to default according to the implementation of `getParameterDirection`.

csyonghe (approved):

Issue #4676, opened 2024-07-18
Capability System: Fully define all capabilities meant for a user with 2 representations `_Internal_atom`/`External_atom`


ArielG-NV:
To reduce the complexity of a change (for easier PR reviews) the larger https://github.com/shader-slang/slang/issues/4347 task will be split into multiple smaller issues.

Problem:
Currently when printing capabilities we have a problem, we print `_capability` atoms.
Due to this printing, users attempt to solve capability problems using these `_capability` atoms.

Using `_` has 2 problems for User Slang code written:
1. `_` atoms are normally designed to only work with 1 target 
2. `_` atoms change in definition and can break user-code.

Solution:
1. Enforce in the capability-generator during capability creation that all `_myAtom` capabilities must have a corresponding `myAtom` capability, `myAtom` should be fully defined on all targets with related features.
2. During capability printing don't print `_` if a atom is prefixed with a `_` 
3. [should be a new issue] Warn any-users using a `_` prefixed capability since only non `_` prefixed atoms should be used.

ArielG-NV:
Note: this will likely require making all `_internal` atoms isolated to a target and all `external ` atoms fully defined on all targets.

PR #4702, opened 2024-07-22
Add `_Internal`/`External` atom enforcement and validation.

ArielG-NV:
Fixes: #4676

Changes:
* Added `validateInternalAtomExternalAtomPair` to the capability generator to ensure all `_Internal` atoms have a corresponding `External` atom.
* Validation of 'RequireCapabilityAttribute' warns if a user uses an '_Internal' atom.
* Added 'External' atoms to atoms with an already existing '_Internal' atom.
* Printing an atom removes '_'.
* Changed atoms to store all names as `UnownedStringSlice`'s since these are preferred over `const char*`.
* Fixed some incorrect which were checking for the incorrect warning/error (capability4.slang, capability5.slang, capability6.slang).


source/slang/slang-capability.h
```diff
@@ -322,6 +322,9 @@ bool isCapabilityDerivedFrom(CapabilityAtom atom, CapabilityAtom base);
     /// Find a capability atom with the given `name`, or return CapabilityAtom::Invalid.
 CapabilityName findCapabilityName(UnownedStringSlice const& name);
 
+    /// Check if 'name' is an '_Internal' or 'External' capability.
+bool isInternalCapabilityName(CapabilityName name);
```
jkwak-work, line 326: I am not sure if we want to call it an "internal".
How about "terminal" to indicate that the capability doesn't include the lower versions of it.

ArielG-NV: From what I understand the point of having a distinction of '_Internal' versus `External' is that we want the ability to have certain capabilities we can safely change/remove/reformat for the ease of making stdlib capabilities without introducing breaking changes into user code.

These naturally will tend to be a "terminal" like atom but I do not know if this will always be the case.

source/slang/slang-capability.cpp
```diff
@@ -897,6 +903,17 @@ void CapabilitySet::addSpirvVersionFromOtherAsGlslSpirvVersion(CapabilitySet& ot
...
 
+void printDiagnosticArg(StringBuilder& sb, CapabilityAtom atom)
+{
+    printDiagnosticArg(sb, (CapabilityName)atom);
+}
+
+void printDiagnosticArg(StringBuilder& sb, CapabilityName name)
+{
+    SLANG_ASSERT(_getInfo(name).name != nullptr);
+    sb << ((_getInfo(name).name[0] == '_') ? &_getInfo(name).name[1] : &_getInfo(name).name[0]);
```
csyonghe, line 914: should probably use `UnownedStringSlice.tail` function for truncating.

csyonghe (dismissed):

source/slang/slang-capability.h
csyonghe: I am ok with internal and external naming here.

source/slang/slang-capabilities.capdef
```diff
@@ -710,30 +738,27 @@ alias GLSL_410_SPIRV_1_0 = _GLSL_410 | spirv_1_0;
...
+alias METAL_3_1 = metallib_3_1;
 
-alias SPIRV_1_0 = spirv_1_0;
-alias SPIRV_1_1 = spirv_1_1;
-alias SPIRV_1_2 = spirv_1_2;
-alias SPIRV_1_3 = spirv_1_3;
-alias SPIRV_1_4 = spirv_1_4;
-alias SPIRV_1_5 = spirv_1_5;
-alias SPIRV_1_6 = spirv_1_6;
+// Profiles of convience
```
csyonghe, line 761: typo

csyonghe (dismissed):

PR #4707, opened 2024-07-22
Adds a warning for using `[PreferRecompute]` on methods that may contain side effects

### LINKED ISSUE #3303: Warn on marking function with known side-effects with `[PreferRecompute]`
Right now, its possible to accidentally mark a method that relies on global/shared memory side-effects with `[PreferRecompute]`. The additional calls for recomputation will produce inconsistent output, and cause seemingly random gradients. 

We should throw a warning if the compiler data-flow analysis detects a method with side-effects being marked recompute, and provide a way to suppress it if it is intentional (the user may be collecting statistics on how many times the method is recomputed, etc..)

saipraveenb25:
Also adds a way to suppress this warning by using `[PreferRecompute(SideEffects.Allow)]`

Not _technically_ a breaking change since this will only generate a warning.

Resolves #3303


source/slang/core.meta.slang
```diff
@@ -2748,8 +2748,16 @@ attribute_syntax [payload] : PayloadAttribute;
 __attributeTarget(DeclBase)
 attribute_syntax [deprecated(message: String)] : DeprecatedAttribute;
 
+enum SideEffects
```
csyonghe, line 2751: => `SideEffectBehavior`

csyonghe (approved):

csyonghe (approved):

csyonghe:
@saipraveenb25 there are CI failures here, will you take a look?

Issue #4700, opened 2024-07-22
Overlapping explicit binding of uniform buffer produces incorrect values


sivansh11:
explictly setting the same binding and space for multiple uniform buffers doesnt work (gives nonsensical values)
```
[[vk_binding(0,0)]] uniform readonly uint64_t addresses[1000];
[[vk_binding(0,0)]] uniform readonly uint64_t addresses1[1000];
```

gives 
```
addresses 0
addresses1 8000
```
I doubt this is intentional, do let me know if I am wrong.

on commit 5da06d43bb0997455211ca56597c4302b09909ab

ArielG-NV:
This is likely a bug.

csyonghe:
Aliasing binding is allowed, but the code isn’t using the right syntax to define buffers. A buffer should be declared as StructuredBuffer, but the vk_binding in the provided code is declared directly on an ordinary data field that will be placed in the default constant buffer, the vk_binding will have no meaning in this case.

you probably want:

[vk::binding(0,0)] ConstantBuffer<uint[1000]> addresses;
[vk::binding(0,0)] ConstantBuffer<uint[1000]> addresses1;

which defines two constant buffer resources that maps to a same binding. Note that global ordinary iniform data cannot have a binding.

ArielG-NV:
> which defines two constant buffer resources that maps to a same binding. Note that global ordinary iniform data cannot have a binding.

@csyonghe should this produce a warning?

ArielG-NV:
or should we allow mapping `[[vk_binding(0,0)]]` to equivalent `layout` syntax?

csyonghe:
I would say a warning for now on binding attributes declared on uniform data.

ArielG-NV:
I was reading [our user guide](https://shader-slang.com/slang/user-guide/conventional-features.html#global-variables-and-shader-parameters) and it seems we support the behavior described in this issue

> Binding information for Vulkan (and OpenGL) may be specified using [[vk::binding(...)]] attributes

sivansh11:
@csyonghe ahh got you, didnt know ordinary uniforms couldnt have an explicit binding,

I tried out using ConstantBuffer, but I am getting a segfault 
```
[[vk::binding(0, 0)]] uniform ConstantBuffer<uint64_t[1000]> addresses;
[[vk::binding(0, 0)]] uniform ConstantBuffer<uint64_t[1000]> addresses1;

[shader("compute")]
[numthreads(1, 1, 1)]
[...]

PR #4708, opened 2024-07-23
Warn if providing explicit bindings to a object using uniform locations

ArielG-NV:
fixes: #4700

Changes:
* If a uniform object (which uses uniform locations) has explicit bindings we will warn to use `ConstantBuffer<T>` instead. We check for a warning specifically when we know an object uses uniform layouts because objects may not use a uniform-layout register even if tagged with `uniform`. A good example of this is `uniform ConstantBuffer<T>`.


jkwak-work (approved):
Looks good to me

PR #4710, opened 2024-07-23
Overhaul IR lowering of pointer types.

### LINKED ISSUE #4674: slang not annotating array as restricted or aliased
the following shader 
```
struct input_t {
    uint32_t i : 8;
    uint32_t j : 24;
};

struct output_t {
    uint32_t i;
    uint32_t j;
};

uniform void *parameters[1000];

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID, uint groupIndex: SV_GroupIndex) {
    // buffer 0 is input
    // buffer 1 is output
    input_t *input = (input_t *)(parameters[0]);
    output_t *output = (output_t *)(parameters[1]);

    output->i = input->i;
    output->j = input->j;
}
```

doesnt pass spir-v validation, getting the following validation error
```
Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
OpVariable 19: expected AliasedPointer or RestrictPointer for PhysicalStorageBuffer pointer.
  %19 = OpVariable %_ptr_Function__arr__ptr_PhysicalStorageBuffer_void_int_1000 Function
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.3.283.0/linux/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```
currently on commit e19e0474639d579e6f0dc33d6b0c9b2a6d723fae, vulkan sdk version: 1.3.283.0

csyonghe:
Fixes #4674.

This change reimplements how user pointers are handled in our IR passes.
Previously, a user defined pointer type lowers into `IRPtrType(T)` without an explicit address space. This makes it difficult to tell if a pointer type is from a user defined pointer (via T* in user code), or an "address type" for representing internal addresses (e.g. the type of a variable).

We attempted to handle this in an ad-hoc way in spirv legalization pass, by identifying pointer types used in struct fields or global shader parameters as `PhysicalStorageBuffer` pointer and assign address space to these pointers.

However, this logic is problematic and it can miss out corner cases like array of pointers.

This PR implements a more systematic approach, by properly defining `Ptr` type as a `Ptr<T, addrSpace>` type in the front-end, with `addrSpace` defaults to `AddressSpace::UserPointer`. A user defined pointer type will have its address space specified as `UserPointer`, and a variable will have a pointer type with the `Generic` address space. By providing the address space info during lowering, we will know which pointer type is a user-defined pointer without any guessing logic.


source/slang/slang-check-expr.cpp
```diff
@@ -4425,7 +4425,7 @@ namespace Slang
         expr->base = CheckProperType(expr->base);
         if (as<ErrorType>(expr->base.type))
             expr->type = expr->base.type;
-        auto ptrType = m_astBuilder->getPtrType(expr->base.type);
+        auto ptrType = m_astBuilder->getPtrType(expr->base.type, AddressSpace::UserPointer);
```
kaizhangNV, line 4428: Why do we use "UserPointer" here?

source/slang/slang-emit-spirv.cpp
```diff
@@ -1213,6 +1223,19 @@ struct SPIRVEmitContext
         return m_NonSemanticDebugPrintfExtInst;
     }
 
+    SpvStorageClass addressSpaceToStorageClass(AddressSpace addrSpace)
+    {
+        switch (addrSpace)
+        {
+        case AddressSpace::Generic:
+            return SpvStorageClassMax;
```
kaizhangNV, line 1231: I have a question here, in the PR description, I see
"and a variable will have a pointer type with the Generic address space",

but SpvStorageClassMax is an invalid storage class, right?

source/slang/slang-emit-spirv.cpp
```diff
@@ -1411,14 +1434,24 @@ struct SPIRVEmitContext
                     && as<IRStructType>(valueType)
                     && storageClass == SpvStorageClassPhysicalStorageBuffer);
                 SpvId valueTypeId;
-                if (useForwardDeclaration)
+                if (as<IRVoidType>(valueType))
                 {
-                    valueTypeId = getIRInstSpvID(valueType);
+                    // Emit void* as uint*.
```
kaizhangNV, line 1439: why is that?
Does not slang support `void*`?

source/slang/slang-emit-spirv.cpp
```diff
@@ -4333,10 +4366,10 @@ struct SPIRVEmitContext
 
     void maybeEmitPointerDecoration(SpvInst* varInst, IRInst* inst)
     {
-        auto ptrType = as<IRPtrType>(inst->getDataType());
+        auto ptrType = as<IRPtrType>(unwrapArray(inst->getDataType()));
         if (!ptrType)
             return;
-        if (ptrType->getAddressSpace() == SpvStorageClassPhysicalStorageBuffer)
+        if (addressSpaceToStorageClass(ptrType->getAddressSpace()) == SpvStorageClassPhysicalStorageBuffer)
```
kaizhangNV, line 4372: can this change to 
`ptrType->getAddressSpace() == UserPointer` ?

source/slang/slang-emit-spirv.cpp
kaizhangNV: never mind, this is spirv emit.

source/slang/slang-legalize-types.cpp
```diff
@@ -209,8 +209,8 @@ bool isPointerToResourceType(IRType* type)
 {
     while (auto ptrType = as<IRPtrTypeBase>(type))
     {
-        if (ptrType->getAddressSpace() == SpvStorageClassStorageBuffer ||
-            ptrType->getAddressSpace() == SpvStorageClassPhysicalStorageBufferEXT)
+        if (ptrType->getAddressSpace() == AddressSpace(SpvStorageClassStorageBuffer) ||
```
kaizhangNV, line 212: I see the value for SpvStorageClassStorageBuffer is 12, so there is no corresponding address space value.
So can we remove this comparison as it will always be `false`?

tests/spirv/pointer-array-2.slang
```diff
@@ -0,0 +1,27 @@
...
+struct output_t {
+    uint32_t i;
+    uint32_t j;
+};
+
+uniform void *parameters[1000];
+
+//TEST:SIMPLE(filecheck=CHECK): -target spirv
+
+//CHECK: OpEntryPoint
```
kaizhangNV, line 15: what else do we want to check in this test?

kaizhangNV:
The change looks good to me overall, I just left few comments for the trivial stuff.

source/slang/slang-ast-type.cpp
```diff
@@ -352,6 +352,60 @@ Type* NativeRefType::getValueType()
...
+    }
+    switch (addrSpace)
+    {
+    case AddressSpace::Generic:
+        break;
+    case AddressSpace::UserPointer:
+        out << toSlice(", user_ptr");
+        break;
+    case AddressSpace::GroupShared:
+        out << toSlice(", groupshared");
```
kaizhangNV, line 376: Can we make the string consistent?
`group_shared`, `thread_local`

or `userptr`

source/slang/slang-type-system-shared.h
```diff
@@ -58,6 +58,20 @@ FOREACH_BASE_TYPE(DEFINE_BASE_TYPE)
     const int kStdlibTextureIsShadowParameterIndex = 6;
     const int kStdlibTextureIsCombinedParameterIndex = 7;
     const int kStdlibTextureFormatParameterIndex = 8;
+
+    enum class AddressSpace : uint64_t
+    {
+        Generic = 0x7fffffff,
+        ThreadLocal = 1,
```
kaizhangNV, line 65: Actually I suggest we can define the enum by using the storage class directly,
e.g.
`ThreadLocal = (uint64_t)(SpvStorageClassInput),`

so we won't be impacted if there is any change of the spirv.h, though it's very unlikely to change.

source/slang/slang-emit-spirv.cpp
kaizhangNV: If the address space is `Uniform`, which is 4 in your definition.

so it will return `SpvStorageClassWorkgroup`.

Shouldn't return `SpvStorageClassUniform`?

source/slang/slang-type-system-shared.h
ArielG-NV: `ThreadLocal = (uint64_t)(SpvStorageClassInput),` would be invalid semantics, these are not actually equivalents.

SPIRV uses `SpvStorageClassFunction`(7) while other targets use `ThreadLocal`(1).

source/slang/slang-type-system-shared.h
```diff
@@ -58,6 +58,20 @@ FOREACH_BASE_TYPE(DEFINE_BASE_TYPE)
     const int kStdlibTextureIsShadowParameterIndex = 6;
     const int kStdlibTextureIsCombinedParameterIndex = 7;
     const int kStdlibTextureFormatParameterIndex = 8;
+
+    enum class AddressSpace : uint64_t
```
ArielG-NV, line 62: My suggestion is similar to Kai's but slightly different in goal.

Currently if I want to add something like a "ThreadLocal" address space pointer I would either:
1. for Metal assign to the Ptr the address space of `ThreadLocal`.
2. for SPIRV assign to the Ptr the address space of `SpvStorageClassFunction`

My suggestion is to make these the semantics the same or for a comment to be added "address space XYZ with Metal maps to address space ABC for SPIRV". This would make it much easier to maintain address-space using code compatibility with SPIRV/Metal.

source/slang/slang-legalize-types.cpp
ArielG-NV: The comparison will not always be false, Pointers are assigned the address space of either `AddressSpace` or `SpvStorageClass*` as a value.

It is important to note that `AddressSpace` does not have equal values to equal-counter-parts in `SpvStorageClass*`

ArielG-NV (commented):
The additions looks good.

I do have a comment on if we can merge (or at-least unify) the usage of address spaces.

Currently Metal uses address-spaces disjoint from SPIR-V. This makes sense at a high level since Metal does not need an address space for things like `Input` parameters.

The problem with this approach is that some legalization passes which legalize Metal/SPIRV are then incompatible even when the logic is compatible (ThreadLocal vs. SpvStorageClassFunction, note the use of `addressSpaceToStorageClass`). These bugs are also quite hard to track-down since SPIR-V code will fail during validation.

I am wondering if we could try to unify how we manage these address spaces to make things simpler and more reusable? This can be a separate PR or issue, I just wanted to bring up that this is a source of easy confusion and bugs.

source/slang/slang-check-expr.cpp
csyonghe: A user defined pointer type in user code, such as `T*`, should be represented as a user-defined pointer in the IR to differentiate themselves from compiler-internal pointer types used for variables (`IROp_Var` etc.).

This will allow us to easily identify these user defined pointers and translate them to pointers in PhysicalStorageBufferClass address space when we get to spirv legalization.

source/slang/slang-emit-spirv.cpp
csyonghe: `Generic` address space means there is no specific address space info associated with the pointer type. It is the responsibility of the spirv legalization and address propagation step to make sure all pointers in generic address space are turned into pointers with a valid SPIRV Storage class as its address space.

source/slang/slang-emit-spirv.cpp
csyonghe: At this stage, a ptr type can either have AddressSpace::UserPointer or SpvStorageClassPhysicalStorageBuffer, we want both cases to map to PhysicalStorageBuffer, so this is how we can ensure that.

tests/spirv/pointer-array-2.slang
csyonghe: Not much beyond we generate valid spirv.

source/slang/slang-type-system-shared.h
csyonghe: There is no one to one mapping from spv storage class to what is currently defined in `AddressSpace`. The current AddressSpace enum is derived from what metal requires and did not consider the need of SPIRV. For example, ThreadLocal can be SpvFunction or SpvPrivate depending on whether it is a local or global variable.

This will need a future pass for a full conversion so we never directly use the enums for both spirv and metal. For now we just use the SpvStorageClass values directly as AddressSpace.

csyonghe:
Unifying the address space is definitely what we should do, but I don't want to do it just yet in this PR.

source/slang/slang-ast-type.cpp
csyonghe: I changed the logic here so that ptr(T, userpointer) is emitted as "Ptr<T>" and ptr(T, generic) is emitted as "Addr<T>".

source/slang/slang-legalize-types.cpp
kaizhangNV: Sorry, I don't understand your explanation here.

Do you mean 
`ptrType->getAddressSpace() ` could returns a value out of definition in
```
enum class AddressSpace
{...}
```
?

kaizhangNV:
> Unifying the address space is definitely what we should do, but I don't want to do it just yet in this PR.

It makes sense here, this change is already too big.
We should always keep the change as small as possible. (Hope so).

But at least can you please fill a new issue to track this?

source/slang/slang-legalize-types.cpp
csyonghe: That's correct, for now we are still using spv storage class enum values directly as the address space operand when legalizing spirv. We should stop doing that and introduce proper enum values to AddressSpace but that is out of the scope of this PR.

csyonghe:
Created #4742  to  track address space cleanup work.

jkwak-work (commented):
Looks good to me, although the PR is already merged.

PR #4713, opened 2024-07-23
Allow only specific spv storage classes for binding decoration

kaizhangNV:
In
https://registry.khronos.org/vulkan/specs/1.3/html/chap37.html#VUID-StandaloneSpirv-DescriptorSet-06491

it states that
If a variable is decorated by DescriptorSet or Binding, the Storage class must be UniformConstant, Uniform and StorageBuffer.

So apply this rule to our emit-spirv logic.

### DIFF
--- a/source/slang/slang-emit-spirv.cpp
+++ b/source/slang/slang-emit-spirv.cpp
@@ -1791,6 +1791,26 @@ struct SPIRVEmitContext
         }
     }
 
+    static SpvStorageClass getSpvStorageClass(IRPtrTypeBase* ptrType)
+    {
+        SpvStorageClass storageClass = SpvStorageClassFunction;
+        if (ptrType && ptrType->hasAddressSpace())
+        {
+            storageClass = (SpvStorageClass)ptrType->getAddressSpace();
+        }
+        return storageClass;
+    }
+
+    // https://registry.khronos.org/vulkan/specs/1.3/html/chap37.html#VUID-StandaloneSpirv-DescriptorSet-06491
+    // Only UniformConstant, Uniform or StorageBuffer storage class are allowed to be decorated with descriptor
+    // set or binding.
+    static inline bool isBindingAllowed(SpvStorageClass storageClass)
+    {
+        return  (storageClass == SpvStorageClassUniformConstant) ||
+                (storageClass == SpvStorageClassStorageBuffer) ||
+                (storageClass == SpvStorageClassUniform);
+    }
+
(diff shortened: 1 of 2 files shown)


kaizhangNV:
I will add a unit test for this soon.

source/slang/slang-emit-spirv.cpp
```diff
@@ -1791,6 +1791,26 @@ struct SPIRVEmitContext
...
+        }
+        return storageClass;
+    }
+
+    // https://registry.khronos.org/vulkan/specs/1.3/html/chap37.html#VUID-StandaloneSpirv-DescriptorSet-06491
+    // Only UniformConstant, Uniform or StorageBuffer storage class are allowed to be decorated with descriptor
+    // set or binding.
+    static inline bool isBindingAllowed(SpvStorageClass storageClass)
+    {
+        return  (storageClass == SpvStorageClassUniformConstant) ||
```
csyonghe, line 1809: Use `switch` to list all cases.

source/slang/slang-emit-spirv.cpp
```diff
@@ -2196,12 +2220,15 @@ struct SPIRVEmitContext
             case LayoutResourceKind::UnorderedAccess:
             case LayoutResourceKind::SamplerState:
             case LayoutResourceKind::DescriptorTableSlot:
-                emitOpDecorateBinding(
-                    getSection(SpvLogicalSectionID::Annotations),
-                    nullptr,
-                    varInst,
-                    SpvLiteralInteger::from32(int32_t(index)));
-                if (!isDescirptorSetDecorated)
+                if (isBindingDecorationAllowed)
```
csyonghe, line 2223: just do:
```
if (!isBindingDecorationAllowed)
     break;
```

source/slang/slang-emit-spirv.cpp
kaizhangNV: You want me to list all storage classes or just these 3? There are lots of cases.

csyonghe: Just turn this into a switch statement and list the three cases. No need to list everything.

kaizhangNV: OK, I see.
But why we prefer the switch..case to this logic expression? Is just a performance concern?

source/slang/slang-emit-spirv.cpp
kaizhangNV: act

csyonghe (approved):

PR #4714, opened 2024-07-23
Fix auto-diff synthesized method naming conventions

saipraveenb25:
Avoid modifying the name of types that already have a name.. this was causing us to rename self-differential types into `s_diff_<typename>` and messing up the emitted code.


csyonghe (dismissed):

csyonghe:
There are test failures that need to be fixed.

Issue #4720, opened 2024-07-23
Handle uninitialization warnings for target switches and intrinsic asm


venkataram-nv:
Linked to #4698 

Currently the contents of the case blocks in a target switch are not processed by this warning system. Same applies for intrinsic asm.

PR #4727, opened 2024-07-23
Warnings target switch intrinsic asm

venkataram-nv:
Handles #4720 

Also adds a fix to skip generating IR for empty target switches. Previously this would result in generating an unreachable block, which would have necessitated special handling for target switches.

### DIFF
--- a/source/slang/slang-ir-reachability.cpp
+++ b/source/slang/slang-ir-reachability.cpp
@@ -69,20 +69,6 @@ namespace Slang
             }
         }
 
-        // Special cases
-        
-        // Target switches; treat as reachable from any of its cases
-        if (auto tswitch = as<IRTargetSwitch>(to))
-        {
-            IRBlock* upper = getBlock(from);
-            for (Slang::UInt i = 0; i < tswitch->getCaseCount(); i++)
-            {
-                IRBlock* caseBlock = tswitch->getCaseBlock(i);
-                if (caseBlock == upper || isBlockReachable(caseBlock, upper))
-                    return true;
-            }
-        }
-
         return isBlockReachable(getBlock(from), getBlock(to));
     }
 
--- a/source/slang/slang-ir-reachability.cpp
+++ b/source/slang/slang-ir-reachability.cpp
@@ -74,11 +74,21 @@ namespace Slang
         // Target switches; treat as reachable from its cases
         if (auto tswitch = as<IRTargetSwitch>(inst2))
         {
+            HashSet<IRInst*> blocks;
             for (Slang::UInt i = 0; i < tswitch->getCaseCount(); i++)
+                blocks.add(tswitch->getCaseBlock(i));
+
+            IRInst* tmp = inst1->getParent();
+            while (tmp && tmp != tswitch->getParent())
             {
-                IRBlock* block = tswitch->getCaseBlock(i);
-                if (isWithinBlock(block, inst1))
+                if (blocks.contains(tmp))
                     return true;
+
+                // Branching instructions
+                if (tmp->firstUse)
+                    tmp = tmp->firstUse->getUser()->getParent();
+                else
+                    tmp = tmp->getParent();
             }
         }
 
@@ -94,16 +104,5 @@ namespace Slang
         if (!fromId || !toId)
             return true;
         return sourceBlocks[*toId].contains(*fromId);
-    }
-
-    bool ReachabilityContext::isWithinBlock(IRBlock* block, IRInst* tinst)
-    {
-        for (auto inst = block->getFirstInst(); inst; inst = inst->next)
-        {
-            if (inst == tinst)
-                return true;
-        }
-
-        return false;
     }
 }
--- a/source/slang/slang-ir-reachability.cpp
+++ b/source/slang/slang-ir-reachability.cpp
@@ -1,4 +1,5 @@
 #include "slang-ir-reachability.h"
+#include "slang-ir-insts.h"
 
 namespace Slang
 {
@@ -54,6 +55,35 @@ namespace Slang
         }
 
     }
+    
+    bool ReachabilityContext::isInstReachable(IRInst* inst1, IRInst* inst2)
+    {
+        // If inst1 and inst2 are in the same block,
+        // we test if inst2 appears after inst1.
+        if (inst1->getParent() == inst2->getParent())
+        {
+            for (auto inst = inst1->getNextInst(); inst; inst = inst->getNextInst())
+            {
+                if (inst == inst2)
+                    return true;
+            }
+        }
+
+        // Special cases
+
+        // Target switches; treat as reachable from its cases
+        if (auto tswitch = as<IRTargetSwitch>(inst2))
+        {
+            for (int i = 0; i < tswitch->getCaseCount(); i++)
+            {
+                IRBlock* block = tswitch->getCaseBlock(i);
+                if (isWithinBlock(block, inst1))
+                    return true;
+            }
+        }
+
+        return isBlockReachable(as<IRBlock>(inst1->getParent()), as<IRBlock>(inst2->getParent()));
+    }
 
     bool ReachabilityContext::isBlockReachable(IRBlock* from, IRBlock* to)
     {
@@ -64,5 +94,16 @@ namespace Slang
         if (!fromId || !toId)
             return true;
         return sourceBlocks[*toId].contains(*fromId);
+    }
+
+    bool ReachabilityContext::isWithinBlock(IRBlock* block, IRInst* tinst)
+    {
+        for (auto inst = block->getFirstInst(); inst; inst = inst->next)
+        {
+            if (inst == tinst)
+                return true;
+        }
+
+        return false;
     }
 }
--- a/source/slang/slang-ir-reachability.cpp
+++ b/source/slang/slang-ir-reachability.cpp
@@ -1,5 +1,6 @@
 #include "slang-ir-reachability.h"
 #include "slang-ir-insts.h"
+#include "slang-ir-util.h"
 
 namespace Slang
 {
@@ -59,7 +60,7 @@ namespace Slang
     {
         // If inst1 and inst2 are in the same block,
         // we test if inst2 appears after inst1.
-        if (from->getParent() == to->getParent())
+        if (getBlock(from) == getBlock(to))
         {
             for (auto inst = from->getNextInst(); inst; inst = inst->getNextInst())
             {
@@ -73,7 +74,7 @@ namespace Slang
         // Target switches; treat as reachable from any of its cases
         if (auto tswitch = as<IRTargetSwitch>(to))
         {
-            IRBlock* upper = as<IRBlock>(from->getParent());
+            IRBlock* upper = getBlock(from);
             for (Slang::UInt i = 0; i < tswitch->getCaseCount(); i++)
             {
                 IRBlock* caseBlock = tswitch->getCaseBlock(i);
@@ -82,7 +83,7 @@ namespace Slang
             }
         }
 
-        return isBlockReachable(as<IRBlock>(from->getParent()), as<IRBlock>(to->getParent()));
+        return isBlockReachable(getBlock(from), getBlock(to));
     }
 
     bool ReachabilityContext::isBlockReachable(IRBlock* from, IRBlock* to)
--- a/source/slang/slang-ir-reachability.cpp
+++ b/source/slang/slang-ir-reachability.cpp
@@ -55,15 +55,15 @@ namespace Slang
         }
     }
     
-    bool ReachabilityContext::isInstReachable(IRInst* inst1, IRInst* inst2)
+    bool ReachabilityContext::isInstReachable(IRInst* from, IRInst* to)
     {
         // If inst1 and inst2 are in the same block,
         // we test if inst2 appears after inst1.
-        if (inst1->getParent() == inst2->getParent())
+        if (from->getParent() == to->getParent())
         {
-            for (auto inst = inst1->getNextInst(); inst; inst = inst->getNextInst())
+            for (auto inst = from->getNextInst(); inst; inst = inst->getNextInst())
             {
-                if (inst == inst2)
+                if (inst == to)
                     return true;
             }
         }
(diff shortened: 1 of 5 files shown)


source/slang/slang-ir-reachability.cpp
```diff
@@ -54,6 +55,45 @@ namespace Slang
...
+                blocks.add(tswitch->getCaseBlock(i));
+
+            IRInst* tmp = inst1->getParent();
+            while (tmp && tmp != tswitch->getParent())
+            {
+                if (blocks.contains(tmp))
+                    return true;
+
+                // Branching instructions
+                if (tmp->firstUse)
```
csyonghe, line 88: I don't understand what this is tracking.

source/slang/slang-ir-reachability.cpp
```diff
@@ -54,6 +55,45 @@ namespace Slang
...
+        // Special cases
+
+        // Target switches; treat as reachable from its cases
+        if (auto tswitch = as<IRTargetSwitch>(inst2))
+        {
+            HashSet<IRInst*> blocks;
+            for (Slang::UInt i = 0; i < tswitch->getCaseCount(); i++)
+                blocks.add(tswitch->getCaseBlock(i));
+
+            IRInst* tmp = inst1->getParent();
```
csyonghe, line 81: You may want to call getBlock instead of getParent here. GetBlock returns the enclosing block of the instruction. For instructions in spirv asm blocks, their parent is the spirv inst instead of the enclosing block.

source/slang/slang-ir-reachability.cpp
csyonghe: In that case we should be extending isBlockReachable to properly handle target switch insts along with other control flow.

source/slang/slang-ir-reachability.cpp
```diff
@@ -54,6 +55,45 @@ namespace Slang
         }
 
     }
+    
+    bool ReachabilityContext::isInstReachable(IRInst* inst1, IRInst* inst2)
+    {
+        // If inst1 and inst2 are in the same block,
+        // we test if inst2 appears after inst1.
```
jkwak-work, line 62: If a certain order is expected for the input parameters, the naming may be better with `from` and `to`.
```
bool ReachabilityContext::isInstReachable(IRInst* from, IRInst* to)
```
Or `before` and `after`; or `top` and `down`... I don't know.

source/slang/slang-ir-reachability.cpp
```diff
@@ -54,6 +55,45 @@ namespace Slang
...
+        if (inst1->getParent() == inst2->getParent())
```
jkwak-work, line 63: Shouldn't this call `getBlock`?
That's what the comment above says.
```
if (inst1->getBlock() == inst2->getBlock())
```

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -37,30 +37,30 @@ namespace Slang
             || (inst->m_op == kIROp_Var);
     }
 
-    static bool isUndefinedParam(IRParam* param)
+    static bool isPotentiallyUnintended(IRParam* param)
```
jkwak-work, line 40: I think this name much better.
We often use a naming like `maybeXXX`, and I like `Potentially` too.
But shouldn't it be `isPotentiallyUninitialized`?
I am not sure what Unintended means in this context.

tests/diagnostics/uninitialized-out-parameters.slang
```diff
@@ -47,5 +47,48 @@ void control_flow_defined(bool b, out float y)
...
+    case spirv:
+        {
+            if ((x & 2) == 0)
+                v = half(1);
+            else
+                v = half(1);
+            return;
+        }
+    }
+}
```
jkwak-work, line 91: I am wondering about a case like following,
```
void target_switching_blocks(uint x, out half v)
{
    __target_switch
    {
    case hlsl:
        return; // <=== WOULD IT PRINT THE WARNING WHEN TARGETTING HLSL?
    case glsl:
    case spirv:
        {
            v = half(1);
            return;
        }
    }
}
```

tests/diagnostics/uninitialized-out-parameters.slang
```diff
@@ -47,5 +47,48 @@ void control_flow_defined(bool b, out float y)
     return;
 }
 
+// Specialized handling for intrinsic asm
+void instrincs_defined(float x, out float exp)
+{
+    __intrinsic_asm "frexp";
+}
```
jkwak-work, line 54: FYI, the following case should print a warning but I assume that it will be not easy to implement.
```
void instrincs_defined(float x, out float exp)
{
    __intrinsic_asm "frexp(x)";
}
```
We will need to ignore this case for now.

tests/diagnostics/uninitialized-out-parameters.slang
```diff
@@ -47,5 +47,48 @@ void control_flow_defined(bool b, out float y)
...
+
+// Specialized handling for target switches
+void target_switching_undefined(float x, out float exp)
+{
+    //CHK-DAG: warning 41018: returning without initializing out parameter 'exp'
+    __target_switch {}
```
jkwak-work, line 60: Using `-DAG` looks dangerous, because it may match to a warning from totally different part of this test.
You can include the line number into the pattern.

source/slang/slang-ir-reachability.cpp
venkataram-nv: Where is the `getBlock()` method? It doesn't look like `IRInst` has that.

source/slang/slang-ir-reachability.cpp
venkataram-nv: I have simplified the logic for this part to rely on `isBlockReachable` more. We can't exactly extended `isBlockReachable` to handle switches and control flow because then it would relax the reachability to the point where something within the control flow's block is accessible after the control flow.

tests/diagnostics/uninitialized-out-parameters.slang
venkataram-nv: I does print the warning, although not exclusively for hlsl targets. The warning here would be returning without initializing `v`

Should we only warn for the particular target? The diagnostics are being evaluated at the IR level so I'm not sure.

source/slang/slang-ir-use-uninitialized-values.cpp
venkataram-nv: The refactor is kind of alluding the inclusion of checking `inout` parameters. In that case, if the user never writes to the parameter, they might not have meant to keep the parameter `inout` -- so making the parameter `inout` is unintended.

source/slang/slang-ir-reachability.cpp
```diff
@@ -52,17 +53,51 @@ namespace Slang
...
+            IRBlock* upper = as<IRBlock>(from->getParent());
+            for (Slang::UInt i = 0; i < tswitch->getCaseCount(); i++)
+            {
+                IRBlock* caseBlock = tswitch->getCaseBlock(i);
+                if (caseBlock == upper || isBlockReachable(caseBlock, upper))
+                    return true;
+            }
+        }
+
+        return isBlockReachable(as<IRBlock>(from->getParent()), as<IRBlock>(to->getParent()));
```
csyonghe, line 85: `getBlock(from)` and `getBlock(to)`.

source/slang/slang-ir-reachability.cpp
```diff
@@ -52,17 +53,51 @@ namespace Slang
...
+            {
+                if (inst == to)
+                    return true;
+            }
+        }
 
+        // Special cases
+        
+        // Target switches; treat as reachable from any of its cases
+        if (auto tswitch = as<IRTargetSwitch>(to))
```
csyonghe, line 74: Can we move this to `isBlockReachable`?

venkataram-nv: I don't think that works for this case since `tswitch` is a specific instruction. We need to check against the case blocks of `tswitch` rather than `getBlock(tswitch)`

tests/diagnostics/uninitialized-out-parameters.slang
jkwak-work: Aha. I guess, we cannot print the warning if the checking is happening at ir level.

tests/diagnostics/uninitialized-out-parameters.slang
```diff
@@ -3,32 +3,32 @@
 // Using before assigning
 float regular_undefined_use(out float3 v)
 {
-    //CHK-DAG: warning 41015: use of uninitialized out parameter 'v'
+    //CHK-DAG: (7): warning 41015: use of uninitialized out parameter 'v'
```
jkwak-work, line 6: I am AFK now but there is a better way to check the line numbers. Something like line+1 or something.
Please check the other tests for an example.

source/slang/slang-ir-reachability.cpp
venkataram-nv: So I have moved this special case away from `Reachability` and to the `slang-ir-use-uninitialized-values.cpp` file. There is no problem with target switches with respect to reachability, but it adds a case that needs handling for the warnings. For example we want to warn cases like:
```
void foo(float x, out float exp)
{
      __target_switch {}
}
```
But not cases like this:
```
void foo(float x, out float exp)
{
     __target_switch
     {
     case glsl: __intrinsic_asm("frexp");
     }
}
```
To handle the first case, we treat target switches as loads (same thing is done for a return) so they are checked at the end. If it is a terminal instruction (as with both cases above), then in order for the target switch to be OK there needs to be some instruction in (any of) its case blocks which stores into the out parameter. This is why I wanted to establish that the target switch should be reachable from its nested instructions.

In the ordinary use case, thought, I don't think this is appropriate, which is why I have moved that code back into `slang-ir-use-uninitialized-values.cpp`.

csyonghe: I think the problem is not about target switch, it is just that we need to turn off the warning checking for functions parameters if the function body has __intrinsic_asm.

venkataram-nv: But there should still be some check for control flows path even there is a __intrinsic_asm. For example if users early exit before the __intrinsic_asm instruction, there should be some warning. Target switches are involved in this because they can appear as block terminal instructions, and this was previously not considered.

venkataram-nv: With some changes, the TLDR for this PR is:
* Add consideration for `target_switch` as block terminator instructions
* Manually sift through `__intrinsic_asm` instructions found in all blocks
And the rest of the main logic (`cancelLoadStores`) remains the same.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -277,30 +285,47 @@ namespace Slang
...
 
-    static void cancelLoads(ReachabilityContext &reachability, const List<IRInst*>& stores, List<IRInst*>& loads)
+    static bool canStoreReachLoad(ReachabilityContext& reachability, IRInst* store, IRInst* load)
+    {
+        if (reachability.isInstReachable(store, load))
+            return true;
+
+        // Special cases
+        
+        // Target switches; treat as reachable from any of its cases
```
ArielG-NV, line 295: > Target switches are by default not reachable by each of their cases

Would it be possible to add an option/Flag to `ReachabilityContext` such that we don't need to have code that handles an "exception" to the rule of reachability explicitly for `__target_switch`?

csyonghe: What if `load` is an inst inside a target switch case? It doesn't seem like this logic will capture that.

venkataram-nv: > Would it be possible to add an option/Flag to ReachabilityContext
Yes, we could, but I'm not sure its absolutely necessary for the core purpose of `ReachabilityContext`. The exception with target switch for reachability is only for the warnings so far.

> What if load is an inst inside a target switch case? It doesn't seem like this logic will capture that.
If there was an instruction explicitly loading from the parameter then it would be found in the `IRUse` chain for it, so it doesn't need to be handled in a special way.

csyonghe: From what I read the code is doing here: if `load` is a target switch and we want to check if a `store` can reach this target switch, then we go through all the target switch cases and check if that store is reachable from the case block, if it is reachable from any case block, then we say yes?

I don't quite understand  the logic behind this. We are checking if a store can reach a target switch, but then why do we do so by checking if the case blocks can reach the store? It seems reversed...

venkataram-nv: It is indeed reversed, and that is because in this case the store instructions would appear after the target switch (and its case blocks). If we did `isBlockReachable(upper, caseBlock)` instead, it would always be false in this case because a case block is not reachable from an inner instruction.

csyonghe: Then why `canStoreReachLoad` is called with `load = target_switch`? If `load` is before `store`, then the function should always return false as the name suggests.

venkataram-nv: We want to make an exception for `__target_switch`s as we want to include a warning if they are the last instruction of block (and if they don't initialize the out parameter). Just like for `IRReturn`s we treat `__target_switch` as a load instruction so that it goes through `cancelLoadStores`. For it to work out in `cancelLoadStores` we do special handling in `canStoreReachLoad` where for target switches only we allow the store to appear after the load.

I could refactor the method name to something like `canStoreCancelLoad` if it makes more sense.

source/slang/slang-ir-reachability.h
```diff
@@ -16,23 +16,8 @@ struct ReachabilityContext
     ReachabilityContext() = default;
     ReachabilityContext(IRGlobalValueWithCode* code);
 
+    bool isInstReachable(IRInst* from, IRInst* to);
```
csyonghe, line 19: We can revert this change and move `isInstReachable` to its original place?

venkataram-nv: We can but after switching to `getBlock` calls we would also need to `#include "slang-ir-util.h"` in `slang-ir-reachabilty.h` is that fine?

csyonghe: Ah, I see, moving it here is fine.

csyonghe (approved):

PR #4728, opened 2024-07-23
Support 1-dimensional matrix for HLSL

### LINKED ISSUE #4395: Unsupported 1-dimensional matrix types
HLSL has 1-dimensional matrix types that DXC accepts. Metal does not have them.

I think for on ramping it might be just to have them in case someone uses them. 

I don't know if they have specific load/store ordering dependencies. But the DXC has SPIR-V opcodes:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenSPIRV/binary-op.arith-assign.matrix.hlsl

```
void main() {

    float1x1 a, b;
    b += a;

    {

        float1x2 c, d;
        d -= c;

        float1x3 e, f;
        f *= e;

        float1x4 g, h;
        h *= h;
    }
    
    {
        float2x1 c, d;
        d -= c;

        float3x1 e, f;
        f *= e;

        float4x1 g, h;
        h *= h;
    }    
}
```

jkwak-work:
Closes #4395

This commit allows Slang to use 1-dimensional matrix when targetting HLSL. The 1-dimensional matrix is supported by DXC natively.

GLSL/Vulkan doesn't support the 1-dimensional matrix natively. It is not trivial for Slang to convert all of matrix functions to vector or scalar at the emitting step. We can implement this later if there are needs. This commit disallows the use of 1-dimensional matrix for targetting GLSL/Vulkan by the capability system; in other words, the new 1-dimentional functions have "[require(hlsl)]".


jkwak-work:
I spent some time implementing for GLSL but I realized that all of matrix functions need to be mapped to vector/scalar versions when one of dimensions is 1-dimensional.
That seemed overkill for a such a rare case support, and I retreated to support only for HLSL.

csyonghe (approved):

PR #4729, opened 2024-07-23
Fix around extensions and `IDifferentiable` requirement synthesis. 

### LINKED ISSUE #4699: Type constraint error in parameter list
This almost works. I'm seeing the following:
```
struct ArrayLike<T>
{
}
__generic<T : IDifferentiable>
extension ArrayLike<T> : IDifferentiable
{
}

struct ExpectsDiff<T : IDifferentiable>
{
}

void test<T : IDifferentiable>(ExpectsDiff<ArrayLike<T>> t3 /* Not OK */)
{
    ExpectsDiff<T> t; // OK
    ExpectsDiff<ArrayLike<T>> t2; // Also OK
}
```
ArrayLike<T> conforms to IDifferentiable if T conforms to the IDifferentiable interface, but only in the body of the method; if I use it in the parameter list, I get type argument 'ArrayLike<T>' does not conform to the required interface 'IDifferentiable'.

This means I can't use write backwards derivatives, as DifferentialPair in the parameter list does not work. I've been trying a few permutations but no success, do you have any ideas?

csyonghe:
This PR includes two fixes. Closes #4699.

## Check extensions before function parameters

We need to make sure all extensions are discovered and registered before checking the function parameters, otherwise we will fail to recognize that a parameter type is valid due to missing knowledge about the type conformances brought in through extensions. This is done by slightly modifying `checkModuleDecl` to discover all extension decl and call ensureDecl on them before checking other decls.

## Fix DeclRef formation for synthesized differentiable requirements that are inside an extension.

When synthesizing the requirement for `IDifferentiable` inside an extension decl, we need to make sure the decl ref to the synthesized function/associated type contains a correct reference to the enclosing extension decl. This is done by calling `applyExtensionToDecl`.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -5376,7 +5376,7 @@ namespace Slang
             // we need to form a new proper decl ref to the parent extension decl
             // with the correct specialization arguments.
             //
-            if (auto outerGeneric = GetOuterGeneric(context->parentDecl))
+            if (GetOuterGeneric(context->parentDecl))
             {
 
                 auto extDeclRef = applyExtensionToType(parentExtDecl, context->conformingType);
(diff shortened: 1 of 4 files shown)


source/slang/slang-check-decl.cpp
```diff
@@ -5649,13 +5720,13 @@ namespace Slang
         {
             if(!isAssociatedTypeDecl(requiredMemberDecl.getDecl()))
                 continue;
-            auto requiredMemberDeclRef = m_astBuilder->getLookupDeclRef(subTypeConformsToSuperInterfaceWitness, requiredMemberDecl.getDecl());
+            auto requiredMemberDeclRef = m_astBuilder->getLookupDeclRef(subTypeConformsToSuperInterfaceWitness, requiredMemberDecl.getDecl());;
```
saipraveenb25, line 5723: minor typo: extra semicolon

saipraveenb25 (dismissed):

source/slang/slang-check-decl.cpp
csyonghe: Done.

PR #4731, opened 2024-07-24
Cleanup global memory allocations on shutdown.

### LINKED ISSUE #4208: Fix memory leak in slang.dll

csyonghe:
Closes #4208.

We have many global constants that holds heap allocated memory. This PR adds a `slang::shutdown()` API to clean them up.
This isn't strictly necessary to call from user code, but it gets rid of crt memory leak detection errors.

### DIFF
--- a/include/slang.h
+++ b/include/slang.h
@@ -5351,6 +5351,14 @@ Returns nullptr if there isn't an embedded stdlib.
 */
 SLANG_API ISlangBlob* slang_getEmbeddedStdLib();
 
+
+/* Cleanup all global allocations used by Slang, to prevent memory leak detectors from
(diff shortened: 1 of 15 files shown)


jkwak-work (approved):
Looks good to me.
@venkataram-nv for visibility.

PR #4739, opened 2024-07-24
Allow implicitly casting enum types to bool.

### LINKED ISSUE #4738: Enums should be allowed to cast to bool implicitly.
So users can just write:
```
if (x & BitFlags.One)
        outputBuffer[0] = 1;
```

csyonghe:
Fixes #4738.


jkwak-work (approved):
Looks good to me

PR #4741, opened 2024-07-25
Move SPIRV global variables into a context variable

### LINKED ISSUE #4722: Slang fails to output a `static` variable in form that is optimized by downstream compiler
Problem 1: 

Description:
Slang generates redundant Loads which have no side effect if optimized out. These should be optimized out for possible downstream performance implications.

Repro:
1. compile the following shader `slangc -target spirv -ignore-capabilities -fvk-use-entrypoint-name`
```
struct Data
{
    int inner1;
};
static Data staticMagicBufferData;
ByteAddressBuffer buffer;
[shader("anyhit")]
void main1()
{
    staticMagicBufferData = buffer.Load<Data>(0);
    outputBuffer[0] = 1;
}
```
Expected:
`staticMagicBufferData = buffer.Load<Data>(someMagic(InstanceIndex()).inner1);` should be optimized out

What happens:
`staticMagicBufferData = buffer.Load<Data>(someMagic(InstanceIndex()).inner1);` is not optimized out. This line should be optimized out since there is no side-effects if we remove this `Load`.

Note:
Problem 1 is not important currently since downstream compilers normally optimize the use-case. [This branch has a solution though](https://github.com/ArielG-NV/slang/tree/optimize-out-unused-static-global).

--------------------------
Problem 2:

Description:
spirv-opt fails to SROA (translate struct to scalar) when Slang emits SPIR-V globals. This is a possible performance issue since the generated code may be assigning struct members which are unused in a program life-time (may waste registers).

Repro:
1. compile the following with `-target spirv -fvk-use-entrypoint-name`
```
struct Data 
{
uint data0;
uint data1;
uint data2;
uint data3;
uint data4;
uint data5;
uint data6;
uint data7;
uint data8;
};

static Data globalVar;
ByteAddressBuffer bab;

struct Payload_t
{
    uint dataOut;
};

[shader("anyhit")]
void main1(inout Payload_t payload)
{
    globalVar = bab.Load<Data>(0);
    payload.dataOut = globalVar.data0;
}

[shader("anyhit")]
void main2(inout Payload_t payload)
{
    globalVar = bab.Load<Data>(0);
    payload.dataOut = globalVar.data0;
}
```

Expected:
The emitted SPIR-V does not contain a `OpCompositeConstruct` (of type `Data`)

What happens:
The emitted SPIR-V contains `OpCompositeConstruct` (of type `Data`). This code-gen as a result therefore contains unused member variables (inefficient code-gen).

Note:
spirv-opt runs SROA automatically when we compile to spirv with Slang. SROA by default only runs for 100 member variables (spirv-opt `scalar-replacement` option). Any more than 100 member variables and the optimization will not run.

ArielG-NV:
Fixes #4722

Move SPIRV global variables into a context such that spirv-opt can SROA the variables if multiple entry-point's are used.

Note: SROA will only run for 100 elements if spirv-opt does not manually set the spirv-opt `scalar-replacement` option.

Note: some tests need adjustments due to assuming variables are in global scope.


source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -12,8 +12,108 @@ namespace Slang
...
+                    case kIROp_VulkanHitObjectAttributesDecoration:
+                    case kIROp_VulkanHitAttributesDecoration:
+                        return false;
+                    default:
+                        continue;
+                    };
+                }
+            }
+
+            return true;
```
csyonghe, line 93: We are missing builtins and varying in/out variables.

tests/bugs/spirv-opt-SROA-of-globals.slang
```diff
@@ -0,0 +1,130 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv -fvk-use-entrypoint-name
+
+// Note: SROA will only run for 100 elements if spirv-opt does not manually set the spirv-opt `scalar-replacement` option
+//CHECK-NOT: OpCompositeConstruct
+struct Data 
+{
+uint data0;
```
csyonghe, line 7: will `uint data[100]` work?

source/slang/slang-ir-explicit-global-context.cpp
ArielG-NV: Varyings/Built-ins are params from what I recall. All params are ignored when hoisting with SPIR-V currently (`hoistableTypes = HoistableTypes::GlobalVariable;`)

tests/bugs/spirv-opt-SROA-of-globals.slang
ArielG-NV: `uint data[100]` is treated as 1 element in the struct. If `uint data[100]` is alone, SROA will run and hoist members outside the struct. If `uint data[100]` is with 100 other elements, SROA will not run and a composite will be constructed.

csyonghe (approved):

csyonghe:
We need to have a test to make sure we don't break the ray tracing variables and other builtin variables, especially when using glsl.meta.slang.

ArielG-NV:
> We need to have a test to make sure we don't break the ray tracing variables and other builtin variables, especially when using glsl.meta.slang.

👍 I will add a test

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -12,8 +12,108 @@ namespace Slang
 // thread-group, and wrap them up in an explicit "context"
 // type that gets passed between functions.
 
+enum class HoistableTypes : UInt
+{
+    GlobalVariable = 0b1,
```
kaizhangNV, line 17: can we use hex instead of binary, this seems not our coding convention?

ArielG-NV: We can use hex, although I think this may be confusing since this code is meant to function as a Enum to a bit-field (every bit is a unique element).

I could also use a bitfield to represent this data instead of indirectly through an enum?

Note: [We seem to not have coding conventions about this](docs\design\coding-conventions.md)

tests/bugs/spirv-opt-SROA-of-globals.slang
```diff
@@ -0,0 +1,32 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv -fvk-use-entrypoint-name
+// Note: SROA will only run for 100 elements if spirv-opt does not manually set the spirv-opt `scalar-replacement` option
+//CHECK-NOT: OpCompositeConstruct %Data
```
jkwak-work, line 3: I think this test needs a better description of what it is testing.
I am talking about the summary version of description not the details.

tests/spirv/address-space-specialize.slang
```diff
@@ -7,7 +7,10 @@
 static int gArray0[2];
 groupshared int gArray1[2];
 
-// CHECK: %array = OpFunctionParameter %_ptr_Private__arr_int_int_2
+// Note: static globals are inside a context variable
+// CHECK: OpTypeArray %int %int_2
+// CHECK: OpVariable %_ptr_Function__arr_int_int_2 Function
```
jkwak-work, line 12: What is "a context variable"?
Why does it change the storage type from private to function here?

source/slang/slang-ir-explicit-global-context.cpp
jkwak-work: I found the binary expression not very readable.
It is one of things that C/C++ supports but nobody uses.

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -12,8 +12,108 @@ namespace Slang
 // thread-group, and wrap them up in an explicit "context"
 // type that gets passed between functions.
 
+enum class HoistableTypes : UInt
```
jkwak-work, line 15: This enum seems duplicate to the existing enum, `GlobalObjectKind`.
And the use of the word, "type", here is confusing because it is not type like `int`, `float` and so on.

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -12,8 +12,108 @@ namespace Slang
...
+    public:
+        ExplicitContextPolicy(CodeGenTarget target)
+        {
+            switch (target)
+            {
+            case CodeGenTarget::SPIRV:
+            case CodeGenTarget::SPIRVAssembly:
+                hoistableTypes = HoistableTypes::GlobalVariable;
+                requiresFuncTypeCorrectionPass = true;
+                addressSpaceOfLocals = (AddressSpace)SpvStorageClassFunction;
```
jkwak-work, line 43: You may want to leave a TODO comment here so that we can clean up this line after the issue #4742 is resolved.

source/slang/slang-ir-explicit-global-context.cpp
ArielG-NV: > I found the binary expression not very readable.
> It is one of things that C/C++ supports but nobody uses.

To clarify (to ensure I understand correctly) is the suggestion to use: binary variable initializers, hex initializers, or bit fields?

source/slang/slang-ir-explicit-global-context.cpp
```diff
@@ -227,6 +336,27 @@ struct IntroduceExplicitGlobalContextPass
...
+                auto paramCount = func->getParamCount();
+                List<IRType*> paramTypes;
+                paramTypes.reserve(paramCount);
+                for (UInt i = 0; i < paramCount; i++)
+                    paramTypes.add(func->getParamType(i));
+                paramTypes.add(pairOfFuncs.second->getDataType());
+                auto newFuncType = builder.getFuncType(paramTypes, func->getResultType());
+
+                func->setFullType(newFuncType);
+            }
```
jkwak-work, line 358: I am not sure why this is needed.
But I am worried that you are modifying the key for the map variable, `m_mapFuncToContextPtr`.
It may invalidate the map variable.

source/slang/slang-ir-explicit-global-context.cpp
jkwak-work: I think it should be hex

tests/spirv/address-space-specialize.slang
jkwak-work: After reading the all implementation, I learned that your change is expected to change from private to function.
But I am still not sure what "context variable" is.

source/slang/slang-ir-explicit-global-context.cpp
csyonghe: Should be able to call `fixUpFuncType` for this?
The logic around map is fine here because this logic is modifying the value pointed to by `func`, where the map's key is a pointer to func.

tests/spirv/address-space-specialize.slang
ArielG-NV: >What is "a context variable"?

Sometimes Slang is provided global variables or needs to track some state of a variable "globally". The problem is that some targets cannot use global variables for various reasons.

To solve this problem:
1. Slang allocates all "global" variables in the entry-point of a program. "global" variables are now local variables which are only visible to the current invocation.
2. Slang has a **"context variable"** which is passed to (most) functions in Slang. This Composite variable stores a pointer to all "global" variables, this allows all function-calls to access "global" variables. This context variable is only visible to the current invocation


> Why does it change the storage type from private to function here?

Since we use a "context variable" in this scenario we store "global" variables as a local variable only visible to the current invocation. This means we require the storage class of [`Function`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Storage_Class) for the `OpVariable` definition here.

source/slang/slang-ir-explicit-global-context.cpp
ArielG-NV: I agree, I should just be using `GlobalObjectKind`.

tests/spirv/explicit-context-validation-builtin-glsl.slang
```diff
@@ -0,0 +1,19 @@
+//TEST:SIMPLE(filecheck=CHECK): -allow-glsl -target spirv -emit-spirv-directly -stage compute -entry computeMain
+
+// Check to ensure builtin is not moved into a kernelContext (part of entry-point). Ensure builtin is referenced directly.
```
csyonghe, line 3: I am worried more about output variables like gl_Position.

source/slang/slang-ir-explicit-global-context.cpp
ArielG-NV: > I am not sure why this is needed.

I will update the comment I added to be more descriptive.

> But I am worried that you are modifying the key for the map variable, m_mapFuncToContextPtr.
> It may invalidate the map variable.

The key is the pointer, not the data inside the pointer?

csyonghe (dismissed):

csyonghe:
What about global groupshared variables?

tests/spirv/address-space-specialize.slang
jkwak-work: I guess this is same as the context I saw in Metal legalization.
I am wondering if it will cause any problems regarding the memory pointer.
Our user guide says the following,
```
Slang supports pointers to global memory, but not shared or local memory. For example, it is invalid to define a pointer to a local variable.
```
https://shader-slang.com/slang/user-guide/convenience-features.html#pointers-limited

I am not clear on why we don't support a memory pointer to a local variable. I assume that SPIR-V doesn't support it.
Your change will move the pointers from global memory to local/function memory.
And I am not sure if it will cause a problem.

source/slang/slang-ir-explicit-global-context.cpp
jkwak-work: Just to add more information about this.
I think the reason why the binary representation is not popular is that it quickly becomes unreadable when there are a few zeros.
A common way to define bit flags is like (0x01 << 0), (0x01 << 1), and so on.

tests/spirv/address-space-specialize.slang
ArielG-NV: SPIR-V has local pointers.
Looking at SPIR-V spec I don't see anything that could break our use of global's.

@csyonghe, may I ask for clarification on this topic to make-sure we are not missing anything?

csyonghe:
Since this pass has the potential of introducing regressions, let's roll it out progressively. Let's add a `-enable-experimental-passes` flag (default off) and put the pass underneath the flag.

include/slang.h
```diff
@@ -4404,6 +4405,11 @@ namespace slang
         virtual SLANG_NO_THROW void SLANG_MCALL setAllowGLSLInput(
             bool                    value) = 0;
 
+            /** Enable or disable an experimental compiler passes
+             */
```
csyonghe, line 4409: Functions should be added towards the end to avoid breaking binary compatibility. Btw, I don't think this is needed given that the user can just encode the option as an entry when creating the session.

csyonghe (commented):
Looks good, but we should remove the slang header change.

csyonghe (approved):

PR #4744, opened 2024-07-26
Allow passing sized array to unsized array parameter.

### LINKED ISSUE #4701: Support sometimes aliasing unbound arrays (`T[]`) as `T[N]`
Problem:
Currently we will not compile the following code (unless hacked around).
```
void myFunc(int unboundArray[])
{

}

void main()
{
    int notUnboundArray[16];
    myFunc(notUnboundArray);
}
```

Expected behavior:
GLSL allows `void myFunc(int unboundArray[])` to work like `__generic<let N : int> void myFunc(int unboundArray[N])`.
Slang should do the same.

csyonghe:
Closes #4701.

This changes extends the function call specialization pass to be able to specialize the type of a parameter.
If we see a call with sized array arguments to an unsized array parameter, we will specialize the callee to take a sized array parameter instead. This allows us to implement `kIROp_GetArrayLength` opcode that returns the size of an array.

If we cannot optimize out GetArrayLength, we will diagnose an error. This will the case if the user calls GetArrayLength on an unsized array that cannot be resolved to a sized array during compilation.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1040,8 +1040,8 @@ __generic<T, let N:int>
 __magic_type(ArrayExpressionType)
 struct Array : IArray<T>
 {
-    [ForceInline]
-    int getCount() { return N; }
+    __intrinsic_op($(kIROp_GetArrayLength))
+    int getCount();
 
     __subscript(int index) -> T
     {
@@ -1049,7 +1049,6 @@ struct Array : IArray<T>
         get;
     }
 }
-
     /// An `N` component vector with elements of type `T`.
 __generic<T = float, let N : int = 4>
 __magic_type(VectorExpressionType)
--- a/source/slang/slang-ir-specialize-arrays.cpp
+++ b/source/slang/slang-ir-specialize-arrays.cpp
@@ -13,8 +13,11 @@ struct ArrayParameterSpecializationCondition : FunctionCallSpecializeCondition
     // This pass is intended to specialize functions
     // with struct parameters that has array fields
     // to avoid performance problems for GLSL targets.
-
     // Returns true if `type` is an `IRStructType` with array-typed fields.
+    // It will also specialize functions with unsized array parameters into
+    // sized arrays, if the function is called with an argument that has a
+    // sized array type.
+    //
     bool isStructTypeWithArray(IRType* type)
     {
         if (auto structType = as<IRStructType>(type))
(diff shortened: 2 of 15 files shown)


source/slang/slang-ir-specialize-arrays.cpp
```diff
@@ -38,15 +41,55 @@ struct ArrayParameterSpecializationCondition : FunctionCallSpecializeCondition
...
     }
+
+    bool doesParamTypeWantSpecialization(IRParam* param, IRInst* arg)
+    {
+        auto paramType = param->getDataType();
+        auto argType = arg->getDataType();
+        if (auto outTypeBase = as<IROutTypeBase>(paramType))
+        {
+            paramType = outTypeBase->getValueType();
+            SLANG_ASSERT(as<IRPtrTypeBase>(argType));
```
jkwak-work, line 56: I am little worried that this assertion may fail if the user shader is written incorrectly and the argument is not a pointer type.
Shouldn't it be an error or just return false like following?
```
if (!as<IRPtrTypeBase>(argType)) return false;
```

tests/language-feature/unsized-array.slang
```diff
@@ -0,0 +1,36 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-compute -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -compute -output-using-type
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+int test(int arr[])
```
jkwak-work, line 7: This is a question.
I found it interesting that Slang would specialize a function like this and make the array sized.
It is as if the function is written like following,
```
__generic<let N:int>
int test(int arr[N])
```
How does HLSL and GLSL handle the unsized array as a function parameter natively?

jkwak-work (approved):
Looks good to me.

I have a question out of my curiosity.

jkwak-work:
We may meed to update the user guide for this.

source/slang/core.meta.slang
```diff
@@ -1040,16 +1040,15 @@ __generic<T, let N:int>
 __magic_type(ArrayExpressionType)
 struct Array : IArray<T>
 {
-    [ForceInline]
-    int getCount() { return N; }
+    __intrinsic_op($(kIROp_GetArrayLength))
```
ArielG-NV, line 1043: It looks like this member-function should be labeled `[__readNone]`

ArielG-NV (commented):
Code looks good.
Only 1 suggestion/question.

source/slang/slang-ir-specialize-arrays.cpp
csyonghe: If that happens, the IR is ill formed and that would be a bug in the front-end not properly capturing the error.

The purpose of the semantic checking is to make sure we never generate this kind of ill-formed code, so an ASSERT here is good enough.

tests/language-feature/unsized-array.slang
csyonghe: I don't know GLSL, but dxc just inlines it out (even with -O0), and doesn't support this when generating spirv.

source/slang/core.meta.slang
csyonghe: `[__readNone]` is only meaningful for ordinary functions with a body. Its purpose is to make DCE understand that a `IRCall` to the function is side-effect free. Since this function has an `__intrinsic_op` that maps it to an instruction during IR lowering, there will be no `IRCalls` to it, and there is no need for the attribute. We already added a case in `mightHaveSideEffects` for the new opcode, and that should be sufficient for DCE purposes.

csyonghe:
> We may meed to update the user guide for this.

Done.

jkwak-work (approved):

PR #4745, opened 2024-07-26
Disable warnings for input global variables

venkataram-nv:
Fixes an undesired warning issued when using GLSL layout variables. The higher level problem is that the warning system was not skipping checks for input global variables.

### DIFF
--- a/tests/diagnostics/uninitialized-globals.slang
+++ b/tests/diagnostics/uninitialized-globals.slang
@@ -45,7 +45,7 @@ float use_later_writte()
     return writtenLater;
 }
 
-// Should not warn here (glsl layout decoration check)
+// Varying inputs never warn
 layout(location = 0) in vec4 data;
 
 vec4 glsl_layout_in_ok()
--- a/tests/diagnostics/uninitialized-globals.slang
+++ b/tests/diagnostics/uninitialized-globals.slang
@@ -53,7 +53,7 @@ vec4 glsl_layout_in_ok()
     return data;
 }
 
-// Invalid use of layout variables
+// Layout outputs should still be written to at some point
 layout(location = 1) out vec4 x;
 
 vec4 glsl_layout_out_undefined()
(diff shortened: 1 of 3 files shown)


tests/diagnostics/uninitialized-globals.slang
```diff
@@ -45,4 +45,34 @@ float use_later_writte()
     return writtenLater;
 }
 
+// Should not warn here (glsl layout decoration check)
```
ArielG-NV, line 48: I don't think this comment is quite specific enough with "(glsl layout decoration check)".
```suggestion
// Varying inputs never warn
```

tests/diagnostics/uninitialized-globals.slang
```diff
@@ -45,4 +45,34 @@ float use_later_writte()
...
+layout(location = 0) in vec4 data;
+
+vec4 glsl_layout_in_ok()
+{
+    return data;
+}
+
+// Invalid use of layout variables
```
ArielG-NV, line 56: I don't think this comment describes why we have this test case.

I suggest something like the following:
```suggestion
// Varying outputs should be initialized by the end of a shader.
```

tests/diagnostics/uninitialized-globals.slang
```diff
@@ -45,4 +45,34 @@ float use_later_writte()
...
+layout(location = 1) out vec4 x;
+
+vec4 glsl_layout_out_undefined()
+{
+    //CHK-DAG: ([[# @LINE + 1]]): warning 41017: use of uninitialized global variable 'x'
+    return x;
+}
+
+// OK use of layout variables
```
ArielG-NV: I don't think this comment is needed if we explain why we warn (programmer should see that the variable is initialized).

jkwak-work (approved):
Looks good to me

csyonghe (approved):

jkwak-work (approved):

PR #4753, opened 2024-07-30
Fix SPIRV emit for small-integer texture types.

### LINKED ISSUE #4748: SPIRV validation error with `Texture<uint8_t>` and other small int types
```hlsl
RWTexture2D<uint8_t> g_Image;

[numthreads(8, 8, 1)]
void ComputeMain(uint2 tid: SV_DispatchThreadID) {
    g_Image[tid] = uint8_t((tid.x + tid.y) % 2);
}
```

Produces:

```
$ slangc -target spirv test.slang | spirv-val --target-env vulkan1.3
error: line 21: [VUID-StandaloneSpirv-OpTypeImage-04656] Expected Sampled Type to be a 32-bit int, 64-bit int or 32-bit float scalar type for Vulkan environment
  %19 = OpTypeImage %uchar 2D 2 0 0 2 R8ui
```

---

On a related note, I think that integer texture types should not infer the format in OpTypeImage since the float textures are emitted as having unknown format by default, I suspect this would cause UB when the formats don't match (not sure if the validation layers can catch this, it would probably be more problematic with bindless).

csyonghe:
Closes #4748.

### DIFF
--- a/source/slang/slang-ir-util.cpp
+++ b/source/slang/slang-ir-util.cpp
@@ -1557,10 +1557,24 @@ void hoistInstOutOfASMBlocks(IRBlock* block)
 IRType* getSPIRVSampledElementType(IRInst* sampledType)
 {
     auto sampledElementType = getVectorElementType((IRType*)sampledType);
-    if (sampledElementType->getOp() == kIROp_HalfType)
+    
+    IRBuilder builder(sampledType);
+    switch (sampledElementType->getOp())
     {
-        IRBuilder builder(sampledType);
+    case kIROp_HalfType:
         sampledElementType = builder.getBasicType(BaseType::Float);
+        break;
+    case kIROp_UInt16Type:
+    case kIROp_UInt8Type:
+    case kIROp_CharType:
+        sampledElementType = builder.getBasicType(BaseType::UInt);
+        break;
+    case kIROp_Int8Type:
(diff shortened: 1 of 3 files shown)


source/slang/slang-ir-util.cpp
```diff
@@ -1557,10 +1557,24 @@ void hoistInstOutOfASMBlocks(IRBlock* block)
...
-    if (sampledElementType->getOp() == kIROp_HalfType)
+    
+    IRBuilder builder(sampledType);
+    switch (sampledElementType->getOp())
     {
-        IRBuilder builder(sampledType);
+    case kIROp_HalfType:
         sampledElementType = builder.getBasicType(BaseType::Float);
+        break;
+    case kIROp_UInt16Type:
```
ArielG-NV, line 1567: for half/uint16/int16, should we be emitting a `RelaxedPrecision` decoration (the value is still 32bit so we would just be adding the decoration)?

csyonghe: That decoration will need to be added to the sampling instructions, and needs to be propagated through computations. I am not sure how helpful it will be but seems like non trivial amount of effort so I would skip it unless someone tells us that is an issue.

ArielG-NV (approved):
Code looks good to me.

Issue #4580, opened 2024-07-09
Finalize precompiled library prototype support


cheneym2:
Add support for multiple translation units in one slang-module, validate, and submit the -embed-dxil feature change.

PR #4755, opened 2024-07-30
Initial support for precompiled DXIL in slang-modules

cheneym2:
A new option "-embed-dxil" in slangc will cause slang to precompile
DXIL for translation units and embed the binary blob in the TU IR.
When written to a slang-module, the module will contain the IR
with the extra blob of precompiled DXIL.

When linking, if every loaded module contains precompiled DXIL,
slang will directly link together the precompiled DXIL blobs.

In a future change, support will be extended to allow mixtures
of precompiled and raw slang-modules, as well as partially
precompiled modules.

Fixes Issue #4580


source/slang/slang-options.cpp
```diff
@@ -1616,6 +1619,12 @@ SlangResult OptionsParser::_parseProfile(const CommandLineArg& arg)
         }
     }
 
+    // Hack. Library profile targets imply "whole program" compilation
+    if (profileName == "lib_6_6")
```
cheneym2, line 1623: I had an issue where the library profile specified explicitly by the user was overridden implicitly when an entrypoint was discovered. This is my workaround.  Please advise if there's a preferred way to prevent slang from overriding lib* profiles in the presence of entry points.

tests/library/precompiled-dxil.slang
```diff
@@ -7,9 +7,9 @@
 // The test passes if there is no errror thrown.
```
cheneym2, line 7: Looking at the slang-test output in CI, there seems to be a hazard in multi-stage dependent tests like this:

assed test: 'tests/library/export-test.slang' 
passed test: 'tests/library/precompiled-dxil.slang' 
passed test: 'tests/library/library-test.slang' 
passed test: 'tests/library/library-test.slang.1' 
passed test: 'tests/library/export-test.slang.1' 
passed test: 'tests/library/precompiled-dxil.slang.1' 
passed test: 'tests/metal/atomic-texture-buffer.slang' 
passed test: 'tests/library/precompiled-dxil.slang.2' 

Looks like the precompiled-dxil tests are interleaved with other library tests, and since they even use the same export-library shader source, it seems like a genuine risk to look out for.  It so happens that export-test.slang doesn't create an export-library.slang-module to conflict with the export-library.slang-module my test creates, but let me know if I should preemptively fix this, maybe by making my artifact file names more unique.

source/slang/slang-serialize-container.cpp
```diff
@@ -348,6 +348,12 @@ namespace Slang {
             }
         }
 
+	// TODO:
```
cheneym2, line 351: It doesn't look like this is causing any CI test failures, but I've put the "breaking change" label on for now because of the hint that I got that there might be users that are somehow relying on the program level IR in slang-modules. Not sure what to do about this.

source/slang/slang-options.cpp
ArielG-NV: My guess is that since `lib_6_6` is not an actual capability what you were experiencing was `lib_6_6` becoming `sm_6_6`? Apologies if this answer is not exact enough, I don't exactly understand what you mean by "overridden implicitly" since any overridden capability should(?) warn or assert.

As a suggestion to maybe avoid the "are you lib_x_y" code is that we could add to our `.capdef` file a new capability: `def hlsl_lib : hlsl`. 

This way if a compile needs to compile an "hlsl lib", all that needs to be checked is:
1. does your compile capabilities for your target contain the `hlsl_lib` atom.
2. what is your current `sm_x_y` version.

cheneym2: Without the whole program option, slang overrides the explicitly chosen profile and compiles it as cs_6_6 instead since it discovers the computeMain entrypoint.

Using your suggestion, what would the command line change to in the test?

`TEST(windows):COMPILE: tests/library/precompiled-dxil.slang -o tests/library/precompiled-dxil.slang-module -embed-dxil -profile lib_6_6 -incomplete-library`

jkwak-work (approved):
Looks good to me

source/slang/slang-options.cpp
cheneym2: My gut feeling is that it would still be a hack even if I added all the lib versions. Of course, if this is considered a fine approach, I need to add the rest of the versions.  I'm curious to understand Ariel's suggestion.

source/slang/slang-serialize-container.cpp
cheneym2: I don't know how the program IR can be used to even broach the subject.  What would OV be doing to depend on this feature?

I think it might be a viable option to prevent the generation of the program level IR in the first place when embedding precompiled binaries into modules.  Basically, don't call generateOutput() when -embed-dxil/spirv is specified, but I'm unsure.

source/slang/slang.cpp
```diff
@@ -5758,6 +5776,16 @@ void EndToEndCompileRequest::setTargetMatrixLayoutMode(int targetIndex, SlangMat
     getTargetOptionSet(targetIndex).setMatrixLayoutMode(MatrixLayoutMode(mode));
 }
 
+void EndToEndCompileRequest::setTargetGenerateWholeProgram(int targetIndex, bool value)
+{
+    getTargetOptionSet(targetIndex).set(CompilerOptionName::GenerateWholeProgram, value);
```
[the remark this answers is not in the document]
cheneym2: Yes, I think the reason is that there is logic to infer this flag elsewhere in Slang, but it was not necessary to force it on as I do in the "if lib_6_6" hack.

source/slang/slang-options.cpp
jkwak-work: I am wondering if this is actually true for all the time.
> Library profile targets imply "whole program" compilation

Currently, you can use "lib_6_6" as a profile with an option to compile the whole program explicitly.
If the assumption is true that "lib_x_y" should always compile with "whole program" option, we should make this line less hacky.

ArielG-NV: > Using your suggestion, what would the command line change to in the test?

Nothing would have to change

The idea is:
if we find a `-embed-dxil` parameter we will automatically add `hlsl_lib` to the capabilities of our compile. If the capabilities are then invalid, we know that the target is not compatible with dxil libraries. This will also remove the need to check for `lib_6_6` explicitly, only the `hlsl_lib` capability atom will need to be checked for.

`hlsl_lib` will be defined as `def hlsl_lib : _sm_6_6`. This will ensure that if any capability does not conform to a minimum of `hlsl + sm_6_6` we will error.

ArielG-NV: Note: due to how capabilities work we can treat `hlsl_lib+sm_6_7` as valid whilst `hlsl_lib+sm_6_5" is not with a little bit of checking. This means that we use the capability system to manage all of the "are we at a minimum sm_6_6".

tests/library/precompiled-dxil.slang
csyonghe: There are multiple threads running in parallel, causing the interleaving of output. However each file is processed in a single thread so it should be fine.

source/slang/slang-compiler-tu.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "../core/slang-basic.h"
+#include "slang-compiler.h"
+#include "slang-ir-insts.h"
+
+namespace Slang
+{
+    SLANG_NO_THROW SlangResult SLANG_MCALL Module::precompileForTargets(
+        DiagnosticSink* sink,
+        EndToEndCompileRequest* endToEndReq,
```
csyonghe, line 12: why do we need EndToEndCompileRequest here? It shall only be needed by dxc and nothing else.

source/slang/slang-compiler-tu.cpp
```diff
@@ -0,0 +1,79 @@
...
+        TargetRequest* targetReq)
+    {
+        auto module = getIRModule();
+        Slang::Session* session = endToEndReq->getSession();
+        Slang::ASTBuilder* astBuilder = session->getGlobalASTBuilder();
+        Slang::Linkage* builtinLinkage = session->getBuiltinLinkage();
+        Slang::Linkage linkage(session, astBuilder, builtinLinkage);
```
csyonghe, line 19: I am not very sure if this is correct. A linkage is supposed to be self-contained with no references to modules/IComponentTypes defined in another linkage. But the logic here is creating a new linkage, and use that linkage to create CompositeComponentType whose elements are in a different linkage. This may lead to weird referencing errors inside the compiler.

source/slang/slang-ir.cpp
```diff
@@ -2387,6 +2390,29 @@ namespace Slang
         return static_cast<IRStringLit*>(_findOrEmitConstant(keyInst));
     }
 
+    IRBlobLit* IRBuilder::getBlobValue(ISlangBlob* blob)
+    {
+        IRConstant keyInst;
+        memset(&keyInst, 0, sizeof(keyInst));
+
+        UnownedStringSlice inSlice((const char*)(blob->getBufferPointer()), blob->getBufferSize());
```
csyonghe, line 2398: We should allocate the memory from IRModule's memory arena and copy the contents over.

source/slang/slang-options.cpp
csyonghe: I would suggest us using `dxil_lib` as the profile name.

source/slang/slang-compiler-tu.cpp
cheneym2: Would it be sufficent to use endToEndReq()->getLinkage() instead of creating a new one?

tests/library/precompiled-dxil.slang
cheneym2: I'm unsure. The test files, like "tests/library/library-test.slang" or "tests/library/precompiled-dxil.slang" each contain separate TESTs, and there's an assumption in each file that the output from one TEST is available to the next TEST in that file. But if two different test files are using the same output file names, like "export-library.slang-module", isn't there a hazard that the module file on disk is modified by the other test?  IIUC, the only way to resolve this is to make sure there aren't any output filenames shared between two different test files.  I believe that's true now, but it seems risky to have different tests running in parallel and dumping intermediate files to disk which might have filename collisions by accident.

source/slang/slang-options.cpp
cheneym2: > if we find a -embed-dxil parameter we will automatically add/check if our capabilities are a super set of dxil_lib. If our capabilities are not a super set then we know that the compile is not compatible with dxil libraries. This will also remove the need to check for lib_x_y explicitly since only the dxil_lib capability will needed to be checked for to determine conformance.
> 
> dxil_lib will be defined as alias dxil_lib : _sm_6_6. This will ensure that if any capability 'XYZ' does not conform to a minimum of hlsl + sm_6_6 ("conform to a minimum" meaning 'XYZ' is a super-set) we will diagnose and error accordingly for the user.

The goal of checking lib_6_6 here wasn't to ensure we had at least lib_6_6, but rather just to set the GenerateWholeProgram option. AFAICT the actual minimum needed is lib_6_3, so I created a capability alias for _sm_6_3 and added it to the raw target request I created.

I'm missing something though because the capability I added does not cause any new validation error when my test specifies a too-low profile number.  Like if the test specifies "lib_6_1" I just get an error from dxc.

source/slang/slang-options.cpp
```diff
@@ -1655,7 +1655,15 @@ SlangResult OptionsParser::addEmbeddedLibrary(const CodeGenTarget format, Compil
         m_rawTargets.add(rawTarget);
     }
 
-    m_rawTargets.getLast().optionSet.add(option, true);
+    getCurrentTarget()->optionSet.add(option, true);
+    getCurrentTarget()->optionSet.add(CompilerOptionName::GenerateWholeProgram, true);
```
cheneym2, line 1659: Setting the GenerateWholeProgram flag option here negates the need for the "if lib_6_6" hack that we've been talking about.

source/slang/slang-compiler-tu.cpp
cheneym2: When I do that, I get this error:

```
$ ./build/Release/bin/slang-test.exe tests/library/p
Supported backends: fxc dxc glslang spirv-dis visualstudio genericcpp llvm spirv-opt
passed test: 'tests/library/precompiled-dxil.slang'
error: result code = -1
standard error = {
dxc 1.7: : error : error: invalid profile cs_5_1
}
standard output = {
}

FAILED test: 'tests/library/precompiled-dxil.slang.1'
passed test: 'tests/library/precompiled-dxil.slang.2'

===
66% of tests passed (2/3)
===

failing tests:
---
tests/library/precompiled-dxil.slang.1
---
```

source/slang/slang-ir.cpp
cheneym2: Done

source/slang/slang-options.cpp
```diff
@@ -1925,6 +1957,7 @@ SlangResult OptionsParser::_parse(
                 linkage->m_optionSet.set(optionKind, compressionType);
                 break;
             }
+            case OptionKind::EmbedDXIL: SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::DXIL, CompilerOptionName::EmbedDXIL, "dxil_lib")); break;
```
ArielG-NV, line 1960: We should pass in a capability directly here.
```suggestion
            case OptionKind::EmbedDXIL: SLANG_RETURN_ON_FAIL(addEmbeddedLibrary(CodeGenTarget::DXIL, CompilerOptionName::EmbedDXIL, CapabilityName::dxil_lib)); break;
```

source/slang/slang-options.cpp
ArielG-NV: > I'm missing something though because the capability I added does not cause any new validation error when my test specifies a too-low profile number. Like if the test specifies "lib_6_1" I just get an error from dxc.

I think I gave partially misleading advice, I will leave a new code review comment describing what to change.

To give a general answer though of why there are no errors, the end goal is to check if `cookedCapabilities.implies(CapabilitySet(dxil_lib))` is false for when `-embed-dxil` is true.

source/slang/slang-options.cpp
```diff
@@ -1635,6 +1644,29 @@ SlangResult OptionsParser::_parseProfile(const CommandLineArg& arg)
...
+    }
+
+    getCurrentTarget()->optionSet.add(option, true);
+    getCurrentTarget()->optionSet.add(CompilerOptionName::GenerateWholeProgram, true);
+    CapabilityName atom = findCapabilityName(UnownedTerminatedStringSlice(capability));
+    if (atom == CapabilityName::Invalid)
+    {
+        return SLANG_FAIL;
+    }
+    addCapabilityAtom(getCurrentTarget(), atom);
```
ArielG-NV, line 1665: From what I understand the goal here is to error if capabilities are not at a minimum compatible with `dxil_lib`. 

Assuming this is the goal, what we can do is error when `-embed-dxil` is set and our entry-point is not compatible with the capabilities (we can also check if `-ignoreCapabilities` is turned on, if not on we can just warn and upgrade capabilities instead using `maybeDiagnoseWarningOrError`)

Example below
```
// validateEntryPoint
auto targetCaps = target->getTargetCaps();
auto stageCapabilitySet = CapabilitySet(entryPoint->getProfile().getCapabilityName());
targetCaps.join(stageCapabilitySet);
...
if(target->compilingToDxilLib() && !targetCaps.atLeastOneSetImpliedInOther(CapabilitySet(CapabilityName::dxil_lib)))
{
    sink->diagnose(entryPointFuncDecl, Diagnostics::entryPointIncompatibleWithDxilLib, CapabilitySet(CapabilityName::dxil_lib), targetCaps); // error ...: dxil libraries require $0, entry point compiled with $1.
    /// optionally the code here may be changed such that `maybeDiagnoseWarningOrError` is used. We will upgrade capabilities and warn if  `-ignoreCapabilities` is toggled on.
}
```

ArielG-NV (commented):
Comment about the capabilities usage

source/slang/slang-options.cpp
cheneym2: Thanks, I'll try that tomorrow.  Having slang check that the profile is at least lib_6_3 is what I thought you were suggesting, so it's what I was expecting to happen.

I think a better user experience would be if the minimum profile version was automatically applied for the -embed-dxil target request.  Can the "dxil_lib" capability be applied such that it would actually cause the target request to be 6_3 minimum, or higher if the user explicitly requested a newer version?

ArielG-NV: > Can the "dxil_lib" capability be applied such that it would actually cause the target request to be 6_3 minimum, or higher if the user explicitly requested a newer version?

We can do is something like this:
```
// validateEntryPoint
auto targetCaps = target->getTargetCaps();
auto stageCapabilitySet = CapabilitySet(entryPoint->getProfile().getCapabilityName());
targetCaps.join(stageCapabilitySet);
...
auto dxilLibCapabilitySet = CapabilitySet(CapabilityName::dxil_lib);
if(target->compilingToDxilLib() && !targetCaps.atLeastOneSetImpliedInOther(dxilLibCapabilitySet))
{
     // If `RestrictiveCapabilityCheck` is true we will error, else we will warn.
     // error ...: dxil libraries require $0, entry point compiled with $1.
     // warn ...: dxil libraries require $0, entry point compiled with $1, implicitly upgrading capabilities.
    maybeDiagnoseWarningOrError(
                        sink,
                        target->getOptionSet(),
                        DiagnosticCategory::Capability,
                        entryPointFuncDecl->loc,
                        Diagnostics::entryPointIncompatibleWithDxilLib,
                        Diagnostics::entryPointIncompatibleWithDxilLibRestrictive,
                        dxilLibCapabilitySet,
                        targetCaps);
       
       // add dxil_lib to the cooked targetCaps
      auto targetCaps = target->getTargetCaps();
      targetCaps.join(dxilLibCapabilitySet);
      if(targetCaps.isInvalid())
      {
          sink->diagnose(..., CapabilityName::hlsl, target->getTargetCaps().getCompileTarget()); // error ...: failed to upgrade capabilities, '$0' is incompatible with '$1'
      }
      else
      {
          target->setTargetCaps(targetCaps);
      }
}
```

Note: setting `target->setTargetCaps(targetCaps);` may be the wrong way todo this...

cheneym2: Thank you for the code, Ariel! One thing though, validateEntryPoint() would only be applied when targets include an entrypoint. Typically lib targets do not, so I think the validation needs to happen somewhere else.  Perhaps inside precompileForTargets.

ArielG-NV: I see what you are saying, you are correct.
We can do something like this then:
```suggestion
    CapabilityName atom = findCapabilityName(UnownedTerminatedStringSlice(capability));
    if (atom == CapabilityName::Invalid)
    {
        return SLANG_FAIL;
    }
    addCapabilityAtom(getCurrentTarget(), atom);
```
```
// source/slang/slang-compiler-tu.cpp:20
```suggestion
    CapabilityName atom = findCapabilityName(UnownedTerminatedStringSlice(capability));
    if (atom == CapabilityName::Invalid)
    {
        return SLANG_FAIL;
    }
    addCapabilityAtom(getCurrentTarget(), atom);
```
CapabilityName precompileRequirement = CapabilityName::Invalid;
switch (targetReq->getTarget())
{
case CodeGenTarget::DXIL:
    linkage.addTarget(Slang::CodeGenTarget::DXIL);
    precompileRequirement = CapabilityName::dxil_lib;
    break;
default:
    assert(!"Unhandled target");
    break;
}
SLANG_ASSERT(precompileRequirement != CapabilityName::Invalid);
        
auto targetCaps = targetReq->getTargetCaps();
auto dxilLibCapabilitySet = CapabilitySet(CapabilityName::dxil_lib);
if(target->compilingToDxilLib() && !targetCaps.atLeastOneSetImpliedInOther(dxilLibCapabilitySet))
{
     // If `RestrictiveCapabilityCheck` is true we will error, else we will warn.
     // error ...: dxil libraries require $0, entry point compiled with $1.
     // warn ...: dxil libraries require $0, entry point compiled with $1, implicitly upgrading capabilities.
    maybeDiagnoseWarningOrError(
                        sink,
                        target->getOptionSet(),
                        DiagnosticCategory::Capability,
                        entryPointFuncDecl->loc,
                        Diagnostics::entryPointIncompatibleWithDxilLib,
                        Diagnostics::entryPointIncompatibleWithDxilLibRestrictive,
                        dxilLibCapabilitySet,
                        targetCaps);
       
       // add dxil_lib to the cooked targetCaps
      targetCaps.join(dxilLibCapabilitySet);
      if(targetCaps.isInvalid())
      {
          sink->diagnose(..., CapabilityName::hlsl, target->getTargetCaps().getCompileTarget()); // error ...: failed to upgrade capabilities, '$0' is incompatible with '$1'
      }
      else
      {
          targetReq->setTargetCaps(targetCaps);
      }
}
```

File: include/slang.h [hunk 11]
```diff
@@ -872,6 +872,7 @@ extern "C"
             DisableSourceMap,       // bool
             UnscopedEnum,           // bool
             PreserveParameters,       // bool: preserve all resource parameters in the output code.
+            EmbedDXIL,              // bool
```
csyonghe, line 875: This will break binary compatibility, new enums should be added at the bottom of the type.

source/slang/slang-compiler-tu.cpp
csyonghe: You can get the linkage from the module itself, and the astbuilder from the linkage.

source/slang/slang-compiler-tu.cpp
csyonghe: This error is due to the profile being incorrect. You need to make sure to create a temporary target request with the right profile, and generate code with that target request.

tests/library/precompiled-dxil.slang
csyonghe: It is the responsibility of each test to make sure they use a distinct name for its intermediate outputs.  This means that if another test is generating export-library.slang-module, then this test shouldn't be generating the samely named file.

tests/library/precompiled-dxil.slang
```diff
@@ -0,0 +1,24 @@
...
+// The result is two slang-modules that are then linked together in a third slangc invocation.
+// Internally, slang does not use the IR in the modules to link, but rather their embedded DXIL.
+// The test passes if there is no errror thrown.
+// TODO: Check if final linkage used only the precompiled dxil.
+
+//TEST(windows):COMPILE: tests/library/export-library.slang -o tests/library/export-library.slang-module -embed-dxil -profile lib_6_6 -incomplete-library
+//TEST(windows):COMPILE: tests/library/precompiled-dxil.slang -o tests/library/precompiled-dxil.slang-module -embed-dxil -profile lib_6_6 -incomplete-library
+//TEST(windows):COMPILE: tests/library/export-library.slang-module tests/library/precompiled-dxil.slang-module -target dxil -entry computeMain -profile cs_6_6 -o tests/library/linked.dxil
+
+extern int foo(int a);
```
csyonghe, line 14: I would like to point out that we shouldn't be relying on `extern` and `export` symbols for defining the module boundaries.

When we compile a module to dxil, all public symbols in the module should be automatically treated as `export`.

cheneym2: Okay, think I had that working in a previous iteration, but at the time there was an issue with the fact that I did not have the entryPoints properly listed in the CodeGenContext and my ComponentType was at the time just a Module which wasn't working until I created a CompositeType with Module+EntryPoints as children.  I should be able to add that back in.

tests/library/precompiled-dxil.slang
cheneym2: Currently it's fine and there are no clashes. I was just pointing out this issue in case there was a solution to it that I didn't know about, or if there was a naming requirement that I wasn't following, e.g. prefixing output files with the name of the test.

source/slang/slang-serialize-container.cpp
cheneym2: @csyonghe Any feedback on this change to cut out the "target component" IR from slang-modules?

source/slang/slang-options.cpp
cheneym2: Not needed in latest change

source/slang/slang-compiler-tu.cpp
cheneym2: I will address this in a follow-on change that makes this feature work with the Slang API, if that's okay

source/slang/slang-serialize-container.cpp
csyonghe: I actually don't know. I guess we really need to run this against ov tests to find out whether or not this is breaking OV.

cheneym2: It surprises me to find that my change does indeed cause OV regressions, but not because of this.  Taking my change and turning this code back on doesn't help my test results, and taking TOT and turning this code off doesn't hurt test results.  Something else in the change is causing OV regressions, and I'm looking into it now.

include/slang.h
```diff
@@ -4152,6 +4153,13 @@ namespace slang
             int                     targetIndex,
             SlangTargetFlags        flags) = 0;
 
+        virtual SLANG_NO_THROW void SLANG_MCALL setTargetGenerateWholeProgram(
+            int                     targetIndex,
+            bool                    value) = 0;
+
+        virtual SLANG_NO_THROW void SLANG_MCALL setTargetEmbedDXIL(
+            int                     targetIndex,
+            bool                    value) = 0;
```
cheneym2, line 4162: This addition to ICompileRequest is causing my OV regression.

source/slang/slang-serialize-container.cpp
cheneym2: Resolved, this doesn't break known usecases with OV.

source/slang/slang-compiler-tu.cpp
cheneym2: Okay, will fix in next change.

include/slang.h
cheneym2: Fixed

tests/library/precompiled-dxil.slang
cheneym2: Will fix in next change

csyonghe (approved):

PR #4757, opened 2024-07-30
Set `nullptr` to coerce into the default value of `T*` type

### LINKED ISSUE #4754: SPIRV validation error with assign of `nullptr` to pointer variable
Minimal repro:
```hlsl
[vk::push_constant] int* dest;

[numthreads(1)]
void ComputeMain(int tid : SV_DispatchThreadID) {
    int* ptr = nullptr;
    if (tid % 2 == 0) ptr = dest; // prevent const fold
    if (ptr) *ptr = 123;
}
```

Produces:

```
$ slangc -target spirv logs/shader-test/main.slang | spirv-val --target-env vulkan1.3
error: line 56: OpStore Pointer <id> '8[%ptr]'s type does not match Object <id> '28[%28]'s type.
  OpStore %ptr %28
```

The problem is that `nullptr` is being lowered into a void pointer as `OpConvertUToPtr %_ptr_PhysicalStorageBuffer_void %ulong_0` instead of the target pointer type. 

A workaround is to write `(T*)0` instead of nullptr.

ArielG-NV:
Fixes #4754

Changes:
* Currently the way we resolve `nullptr` produces invalid code-gen due to type-mismatching. Now we assign `nullptr` by using `DefaultConstructExpr` to assign a `nullptr`. This correctly assigns `nullptr` to the target `T*` type since the default of `T*` is `nullptr`.

### DIFF
--- a/source/slang/slang-check-conversion.cpp
+++ b/source/slang/slang-check-conversion.cpp
@@ -903,7 +903,11 @@ namespace Slang
                 *outCost = kConversionCost_NullPtrToPtr;
             }
             if (outToExpr)
-                *outToExpr = fromExpr;
+            {
+                auto* defaultExpr = getASTBuilder()->create<DefaultConstructExpr>();
+                defaultExpr->type = QualType(toType);
+                *outToExpr = defaultExpr;
+            }
             return true;
         }
         // none_t can be cast into any Optional<T> type.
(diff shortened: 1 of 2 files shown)


csyonghe (approved):

csyonghe:
This is quick! Thank you for the prompt fix.

Issue #3804, opened 2024-03-20
generic interface cannot be used as a generic argument


jkwak-work:
When an interface uses "__generic", the interface cannot be used as a part of the __generic argument.

It is easier to explain the problem with an example.
The following code gives us a compile error,
```
__generic<T:__BuiltinFloatingPointType> // <== This is the problem
interface IOperation
{
  static bool Equals(T lhs, T rhs);
};
 
__generic<T:__BuiltinFloatingPointType, OpType:IOperation<T>> // <= Not sure if this is a right way
bool Equals(T lhs, T rhs)
{
  return OpType::Equals(lhs, rhs);
}
```
The error message is following,
```
a.slang(8): note 99999: an internal error threw an exception while working on code near this location
bool Equals(T lhs, T rhs)
    ^~~~~~
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unexpected: Generic type/value shouldn't be handled here!
```

The following alternative approach didn't work either,
```
__generic<T:__BuiltinFloatingPointType>
interface IOperation
{
   bool Equals(T lhs, T rhs);
};

__generic<T:__BuiltinFloatingPointType>
struct JKWAK_test
{
    __generic<OpType:IOperation<T>>
    bool Equals(OpType op, T lhs, T rhs)
    {
       return op.Equals(lhs, rhs);
    }
}
```
The error message is identical to the first example.

csyonghe:
This is due to some fundamental limitation we have in our IR. Supporting this is likely something we cannot afford in the short term. Let's try:
```
__generic<T:__BuiltinFloatingPointType>
struct JKWAK_test
{
    bool Equals(IOperation<T>op, T lhs, T rhs)
    {
       return op.Equals(lhs, rhs);
    }
}
```
If this doesn't work, we may be able to fix it.

jkwak-work:
Oh. I got it wrong.
The suggested workaround seems to be working.
Let me try it on UE.

jkwak-work:
This issue was closed as WNF due to the low bang for the buck.
But we should track it as an unfinished feature like https://github.com/shader-slang/slang/issues/4236

PR #4761, opened 2024-07-30
Fix IR lowering for generic interface types.

### LINKED ISSUE #4749: TypeConformance with generics does not work
So when having an interface which is templated, it is possible now to correctly set the type conformance with strings of the interface and implementation struct + the implementing generic type "TestInterface\<float>" "TestInterfaceImpl\<float>". Sadly this still results in an 

```error 50100: No type conformances are found for interface 'TestInterface'. Code generation for current target requires at least one implementation type present in the linkage.```
error

Shortened code:
```cpp
public interface TestInterface<Real> {
    Real sample();
}

struct TestInterfaceImpl<Real> : TestInterface<Real> {
    Real sample() {
        return x;
    }
    Real x;
}

StructuredBuffer<TestInterface<float>> data;
```

csyonghe:
Closes #4749.

The issue being fixed here is that when we lower a generic interface to IR, the function types aren't lowered from proper DeclRef with default substitute info, so they end up being direct `IRGeneric` insts, instead of a `specialize(IRGeneric, ...)` inst. This leads to incorrect interface types after specialization pass. The fix is simply lower the function types using a declref.

### DIFF
--- a/tools/slang-test/slang-test-main.cpp
+++ b/tools/slang-test/slang-test-main.cpp
@@ -4617,8 +4617,8 @@ SlangResult innerMain(int argc, char** argv)
 int main(int argc, char** argv)
 {
     const SlangResult res = innerMain(argc, argv);
-
     slang::shutdown();
+    Slang::RttiInfo::deallocateAll();
 
 #ifdef _MSC_VER
     _CrtDumpMemoryLeaks();
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -8306,54 +8306,50 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         }
 
         UInt entryIndex = 0;
-        auto addEntry = [&](IRStructKey* requirementKey, Decl* requirementDecl)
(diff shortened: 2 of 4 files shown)


tools/slang-test/slang-test-main.cpp
```diff
@@ -4617,8 +4617,8 @@ SlangResult innerMain(int argc, char** argv)
 int main(int argc, char** argv)
 {
     const SlangResult res = innerMain(argc, argv);
-
     slang::shutdown();
+    Slang::RttiInfo::deallocateAll();
```
jkwak-work, line 4621: We may need a document for how to properly shut everything down and deallocate cleanly.
Many user applications will want to shutdown properly to check the memory leaks of their own applications.

jkwak-work (approved):
Looks good to me.

jkwak-work:
I wonder if this can fix #3804 as well

tools/slang-test/slang-test-main.cpp
csyonghe: This is only needed in slang-test-main because slang-test uses the RTTI functionality directly through static linking, so there is a different instance of slang::RTTIInfo that needs to be cleaned up separately. Normal users of slang.dll shouldn't need to do this.

csyonghe:
> I wonder if this can fix #3804 as well

Not quite, that one requires a more fundamental change to our IR rules.

PR #4765, opened 2024-07-31
Allow implicit 'uniform' entrypoint parameters.

### LINKED ISSUE #4554: Error out when varying parameters are used for compute stage
Compute shaders cannot have varying parameters. When the compute shader try to use a varying paramter mistakenly, Slang should print an error.

Any shader that doesn't varying data should have their parameters qualified with `uniform` modifier.

A repro code,
```
 [shader("compute")]
 [numthreads(16, 16, 1)]
 void setcolor(uint3 tid: SV_DispatchThreadID, RWTexture2D<float4> render_texture, int2 pos, int2 size, float4 color)
 {
     if (any(tid.xy >= size))
         return;
     render_texture[tid.xy+pos] = color;
 }
```

csyonghe:
Closes #4554.

Instead of throwing out an error on missing 'uniform' modifier for stages that doesn't allow non-uniform parameters, we will just add the 'uniform' modifier for the user and issue a warning at the same time.


jkwak-work (approved):
Looks good to me.

PR #4766, opened 2024-07-31
Allow generic type deduction from ParameterBlock arguments.

### LINKED ISSUE #2257: Generic deduction error with ParameterBlock
It seems like Slang does not deduce generic parameters when using a ParameterBlock, maybe because an implicit conversion from ParameterBlock to its stored type is happening? 

Minimal example
```cpp
struct TestStruct<Format, let count : int>
{
    RWTexture2D<Format> textures[count];
};

void testFunction<Format, let count : int>(TestStruct<Format, count> data)
{
}

ParameterBlock<TestStruct<float4, 12>> testBlock;

[numthreads(4, 1, 1)]
void main(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    // Does not work: could not specialize generic for arguments of type (ParameterBlock)
    testFunction(testBlock);
    // Works
    testFunction<float4, 12>(testBlock);
    // Works
    TestStruct<float4, 12> temp = testBlock;
    testFunction(temp);
}

```

csyonghe:
Closes #2257.


source/slang/slang-check-constraint.cpp
```diff
@@ -949,6 +949,11 @@ namespace Slang
                     QualType(sndVectorType->getElementType(), snd.isLeftValue));
             }
         }
+
+        if (auto fstUniformParamGroupType = as<UniformParameterGroupType>(fst))
+            return TryUnifyTypes(constraints, QualType(fstUniformParamGroupType->getElementType(), fst.isLeftValue), snd);
+        if (auto sndUniformParamGroupType = as<UniformParameterGroupType>(snd))
+            return TryUnifyTypes(constraints, fst, QualType(sndUniformParamGroupType->getElementType(), snd.isLeftValue));
```
ArielG-NV, line 956: Any reason we don't try to unify directly on `BuiltinGenericType`?

ArielG-NV (dismissed):
Code looks good.

I have 1 question (mostly curiosity).

source/slang/slang-check-constraint.cpp
csyonghe: An alternative is to use `PointerLikeType` here. `BuiltinGenericType` is too general as it is also  the base of `StucturedBufferTypeBase`, which shouldn't be automatically unfolded here.

PR #4767, opened 2024-07-31
Perform type legalization on StructuredBuffer element.

### LINKED ISSUE #4455: Crash with TypeConformance
Using the following code and setting typeconformance for `TestInterface` to `TestImplementation` results in a crash in `composedProgram->getEntryPointCode` `access violation` It seems like this problem comes from interface implementation structs that do not have a member variable.

```csharp
interface TestInterface {
    float sample();
}
struct TestImplementation : TestInterface {
    float sample() {
        return 1.0f;
    }
};
StructuredBuffer<TestInterface> inBuffer;
[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    outputBuffer[0] = inBuffer[0].sample();
}
```

when returning a member var instead, for some reason, it compiles successfully
```csharp
interface TestInterface {
    float sample();
}
struct TestImplementation : TestInterface {
    float x;
    float sample() {
        return x;
    }
};
StructuredBuffer<TestInterface> inBuffer;
[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    outputBuffer[0] = inBuffer[0].sample();
}
```

Slang 2024.1.22

Additionally in the second working case, when the TestImplementation is moved to another file which gets then compiled to a slang-module for later insertion into the main code, this also results in a crash. When including the impl using an import from a TestImplementation.slang file, it works.

csyonghe:
Closes #4455.

The issue here is that we need to make sure to legalize `T` referenced in a `StructuredBuffer<T>`, so that we can generate valid code if some field in T becomes an empty type.


tests/language-feature/interfaces/empty-type-conformance.slang
```diff
@@ -0,0 +1,27 @@
...
+interface TestInterface {
+    float sample();
+}
+struct TestImplementation : TestInterface {
+    float sample() {
+        return 1.0f;
+    }
+};
+
+//TEST_INPUT: set inBuffer = new StructuredBuffer<TestInterface>[new TestImplementation{}];
```
jkwak-work, line 15: Interesting. I didn't know we can do things like this.

jkwak-work (approved):
Looks good to me

Issue #4763, opened 2024-07-31
specializeResourceUsage emit invalid IR code


kaizhangNV:
specializeResourceUsage could generate invalid IR code in some corner cases.

Test code is as follow:

```
RWStructuredBuffer<uint> globalBuffer[] : register(t0, space0);
RWStructuredBuffer<uint> outputBuffer;

int func(RWStructuredBuffer<uint> buffer)
{
    buffer[0] += 10;
    return buffer[0] * 2;
}

[shader("compute")]
[numthreads(1, 1, 1)]
void main(in uint2 id : SV_DispatchThreadID)
{
    uint bufferIdx = id.x;
    uint nonUniformIdx = NonUniformResourceIndex(bufferIdx);

    RWStructuredBuffer<uint> buffer = globalBuffer[nonUniformIdx];

    uint a = func(buffer);

    int bufferIdx2 = id.y;
    RWStructuredBuffer<uint> buffer2 = globalBuffer[NonUniformResourceIndex(bufferIdx2)];

    uint b = func(buffer2);

    outputBuffer[0] = a;
    outputBuffer[1] = b;
}
```
Compile the code with
`-stage anyhit -entry main -target spirv-asm -O0 -dump-ir -o test.spv.asm`.

This test code can expose 2 issues  when we legalize resource type for a function parameter:

1. The function `func` will finally be legalized to `uint func(int)`, however we didn't make the type cast from `uint` to `int` for variable `nonUniformIdx`. This will finally emit invalid spirv code because of the type mismatch. This can be easily observed by running `spirv-val` on the spirv code.
2. In the IR code,  we evaluate the `NonUniformResourceIndex()` right before the call site of `func()`, this will bring an issue that we will not be able to decorate the correct inst to `NonUniform` inside the function of `func()`. So we should evaluate the `NonUniformResourceIndex()` inside  `func()` function, such that the emit logic will know which inst to decorate to `NonUniform`.

csyonghe:
Can you further simplify down this repro to the minimum?

kaizhangNV:
Sure, simplify it.

PR #4769, opened 2024-08-01
Issue/legalize resource

kaizhangNV:
Fix the issue https://github.com/shader-slang/slang/issues/4763.

This PR fixes two issue:
1. Fix the issue that after `specializeFunctionCalls`, `NonUniformResourceIndex` is ignored in the generated specialized
function.

The reason is that if the function has a non-uniform resource parameter, we will legalize it by replacing the resource parameter with a index, and indexing of the resource will be moved inside the specialized function.

e.g.

```
void func(ResourceType resource) { ... }
func(resource[NonUniformResourceIndex(0)])
```
will be specialized into

```
void func(int index) { resource[index]; }
func(0);
```

In this case, inside the function, we will loose the information about whether the resource is a non-uniform. So we add the handling for this corner case by adding insert a `NonUniformResourceIndex` into the specialized function:

```
void func(int index) {
int nonUniformIdx = NonUniformResourceIndex(index);
resource[nonUniformIdx];
}
```

2. Fix the issue that arguments mismatch after `specializeCall()`. 
For example, if the function parameter contains a resource type

```
void func(ResourceType res) { ... }

int index = ...
func(resources[index]);
```

This will be specialized into

```
void func(int index) { resources[index] }

int index = ...
func(index);
```

However, if we have more than 1 call sites, and the other call site doesn't use `int` as the index, e.g.

```
uint index = ...
func(resources[index]);
```
this call site will be specialized into

```
uint index = ...
func(index);
```

this will be invalid, because the argument doesn't match the parameter.

Therefore, after the `specializeCall()`, we will do a type check and cast for the argument list to make the types consistent.

### DIFF
--- a/source/slang/slang-ir-specialize-function-call.cpp
+++ b/source/slang/slang-ir-specialize-function-call.cpp
@@ -564,7 +564,8 @@ struct FunctionParameterSpecializationContext
             // We should also add 2 more things such that our specialization
             // can handle the corner cases that if the oldBase is a nonuniform
             // resource and also the data type of oldIndex will be handled correctly.
-            // By doing so, just add the oldIndex to the key of call info.
+            // By doing so, we form an IRAttributedType to include both information
+            // and add it to the key of call info.
 
             List<IRAttr*> irAttrs;
             if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
--- a/source/slang/slang-ir-specialize-function-call.cpp
+++ b/source/slang/slang-ir-specialize-function-call.cpp
@@ -367,6 +367,14 @@ struct FunctionParameterSpecializationContext
             newFunc = generateSpecializedFunc(oldFunc, funcInfo, callInfo);
             specializedFuncs.add(callInfo.key, newFunc);
         }
+        else
+        {
+            // We don't need to call this function if the specialized function
+            // is not pre-existing, because the we will specialize the function
+            // based on the call site info exactly, so there is no way that the
+            // arguments don't match the parameters of the function.
+            typeCastForNewArguments(newFunc, oldCall, callInfo);
+        }
 
         // Once we've other found or generated a specialized function
         // we need to generate a call to it, and then use the new
@@ -382,6 +390,39 @@ struct FunctionParameterSpecializationContext
         newCall->insertBefore(oldCall);
         oldCall->replaceUsesWith(newCall);
         oldCall->removeAndDeallocate();
+
+    }
+
+    // After specialization of the function, any resource type parameters will be gone, and they could be replaced as
+    // the index of the resources. For example, if the original function is:
+    //
+    // void foo(Texture2D<float> tex) { ... }
+    //
+    // it could be specialized to:
+    //
+    // void foo(int idx) { ... }
+    //
+    // In such case, we need to check the new arguments of the new call site match the type of the parameters of the
+    // new function, because we didn't do type checking when we generate the new call site. This step is necessary,
+    // because in the old call site, it's allowed to use different data type as the index to retrieve the resources,
+    // however, the new function only allows one data type as the input argument.
+    void typeCastForNewArguments(
+        IRFunc* newFunc,
+        IRCall* oldCall,
+        CallSpecializationInfo& callInfo)
+    {
+        UInt argIndex = 0;
+        for( auto newParam : newFunc->getParams() )
+        {
+            auto newArg = callInfo.newArgs[argIndex];
+            if (newParam->getFullType() != newArg->getDataType())
+            {
+                auto castInst = getBuilder()->emitCast(newParam->getFullType(), newArg);
+                castInst->insertBefore(oldCall);
+                callInfo.newArgs[argIndex] = castInst;
+            }
+            argIndex++;
+        }
     }
 
     // Before diving into the details on how we gather information
--- a/source/slang/slang-ir-specialize-function-call.cpp
+++ b/source/slang/slang-ir-specialize-function-call.cpp
@@ -367,14 +367,6 @@ struct FunctionParameterSpecializationContext
             newFunc = generateSpecializedFunc(oldFunc, funcInfo, callInfo);
             specializedFuncs.add(callInfo.key, newFunc);
         }
-        else
-        {
-            // We don't need to call this function if the specialized function
-            // is not pre-existing, because the we will specialize the function
-            // based on the call site info exactly, so there is no way that the
-            // arguments mismatch the parameters of the function.
-            typeCastForNewArguments(newFunc, oldCall, callInfo);
-        }
 
         // Once we've other found or generated a specialized function
         // we need to generate a call to it, and then use the new
@@ -391,57 +383,6 @@ struct FunctionParameterSpecializationContext
         oldCall->replaceUsesWith(newCall);
         oldCall->removeAndDeallocate();
 
-    }
-
-    // After specialization of the function, any resource type parameters will be gone, and they could be replaced as
-    // the index of the resources. For example, if the original function is:
-    //
-    // void foo(Texture2D<float> tex) { ... }
-    //
-    // it could be specialized to:
-    //
-    // void foo(int idx) { ... }
-    //
-    // In such case, we need to check the new arguments of the new call site match the type of the parameters of the
-    // new function, because we didn't do type checking when we generate the new call site. This step is necessary,
-    // because in the old call site, it's allowed to use different data type as the index to retrieve the resources,
-    // however, the new function only allows one data type as the input argument.
-    void typeCastForNewArguments(
-        IRFunc* newFunc,
-        IRCall* oldCall,
-        CallSpecializationInfo& callInfo)
-    {
-        UInt argIndex = 0;
-        for( auto newParam : newFunc->getParams() )
-        {
-            auto newArg = callInfo.newArgs[argIndex];
-            bool isSpecialized = true;
-
-            // If the new argument can be found in the old call site, then it's not specialized.
-            // So we don't need to do any cast.
-            for (UInt i = 0; i < oldCall->getArgCount(); i++)
-            {
-                if (oldCall->getArg(i) == newArg)
-                {
-                    isSpecialized = false;
-                    break;
-                }
-            }
-
-            if (!isSpecialized)
-            {
-                argIndex++;
-                continue;
-            }
-            auto paramType = newParam->getFullType();
-            if (!isTypeEqual(paramType, newArg->getDataType()))
-            {
-                auto castInst = getBuilder()->emitCast(paramType, newArg);
-                castInst->insertBefore(oldCall);
-                callInfo.newArgs[argIndex] = castInst;
-            }
-            argIndex++;
-        }
     }
 
     // Before diving into the details on how we gather information
@@ -612,6 +553,10 @@ struct FunctionParameterSpecializationContext
             // We start by recursively setting up whatever
             // `oldBase` needs:
             //
+            // TODO: We should add check 2 more things in the key
+            // such that our specialization can handle the corner cases
+            // that if the oldBase is a nonuniform resource and also
+            // the data type of oldIndex will be handled correctly.
             getCallInfoForArg(ioInfo, oldBase);
 
             // Then we process `oldIndex` just like we
@@ -620,6 +565,11 @@ struct FunctionParameterSpecializationContext
             // the arguments at the new call site, and
             // don't add anything to the specialization key.
             //
+            // if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
+            // {
+            //
+            // }
+            ioInfo.key.vals.add(oldIndex);
             ioInfo.newArgs.add(oldIndex);
         }
         else if (oldArg->getOp() == kIROp_Load)
--- a/source/slang/slang-ir-specialize-function-call.cpp
+++ b/source/slang/slang-ir-specialize-function-call.cpp
@@ -565,7 +565,16 @@ struct FunctionParameterSpecializationContext
             // can handle the corner cases that if the oldBase is a nonuniform
             // resource and also the data type of oldIndex will be handled correctly.
             // By doing so, just add the oldIndex to the key of call info.
-            ioInfo.key.vals.add(oldIndex);
+
+            List<IRAttr*> irAttrs;
+            if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
+            {
+                IRAttr* attr = getBuilder()->getAttr(kIROp_NonUniformAttr);
+                irAttrs.add(attr);
+            }
+            auto irType = getBuilder()->getAttributedType(oldIndex->getDataType(), irAttrs);
+            ioInfo.key.vals.add(irType);
+
             ioInfo.newArgs.add(oldIndex);
         }
         else if (oldArg->getOp() == kIROp_Load)
--- a/source/slang/slang-ir-specialize-function-call.cpp
+++ b/source/slang/slang-ir-specialize-function-call.cpp
@@ -568,7 +568,7 @@ struct FunctionParameterSpecializationContext
             // and add it to the key of call info.
 
             List<IRAttr*> irAttrs;
-            if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
+            if (findNonuniformIndexInst(oldIndex))
             {
                 IRAttr* attr = getBuilder()->getAttr(kIROp_NonUniformAttr);
                 irAttrs.add(attr);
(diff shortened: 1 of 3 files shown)


jkwak-work:
I think you need a better title for this PR and needs some descriptions so that the reviewer can start reading the code with some ideas of what to expect.

kaizhangNV:
> I think you need a better title for this PR and needs some descriptions so that the reviewer can start reading the code with some ideas of what to expect.

Sure, but next time can you please not start reviewing it until I send review request? This PR is not ready yet. I just want to see all the CI passed.

jkwak-work:
> > I think you need a better title for this PR and needs some descriptions so that the reviewer can start reading the code with some ideas of what to expect.
> 
> Sure, but next time can you please not start reviewing it until I send review request? This PR is not ready yet. I just want to see all the CI passed.

Sure, but next time can you please mark your PR as a draft until it is ready for a review?

tests/compute/nonuniformres-as-function-parameter.slang
```diff
@@ -0,0 +1,95 @@
+//TEST:SIMPLE(filecheck=CHECK-SPV):-target spirv -entry main -stage compute
+//TEST:SIMPLE(filecheck=CHECK-GLSL-SPV):-target spirv -entry main -stage compute -emit-spirv-via-glsl
+//TEST:SIMPLE(filecheck=CHECK-GLSL):-target glsl -entry main -stage compute
+//TEST:SIMPLE(filecheck=CHECK-HLSL):-target hlsl -entry main -stage compute
```
jkwak-work, line 4: Please avoid using minus character(-) as a part of the user-defined keyword.
It is legal to do it but it can cause a confusion, because the words after minus characters are supposed to be a suffix for predefined commands like -NOT or -NEXT.
I suggest to rename them to CHECK_SPV with an underscore character.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -363,9 +364,17 @@ struct FunctionParameterSpecializationContext
...
+            newFunc = generateSpecializedFunc(oldFunc, funcInfo, callInfo);
             specializedFuncs.add(callInfo.key, newFunc);
         }
+        else
+        {
+            // We don't need to call this function if the specialized function
+            // is not pre-existing, because the we will specialize the function
+            // based on the call site info exactly, so there is no way that the
+            // arguments mismatch the parameters of the function.
+            typeCastForNewArguments(newFunc, oldCall, callInfo);
```
csyonghe, line 376: Why is this branch needed? If we had a specialization for int and now calling with uint, shouldn't the tryFindSpecialization fail to find the specialized function in the dictionary?

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -381,6 +390,58 @@ struct FunctionParameterSpecializationContext
...
+    //
+    // it could be specialized to:
+    //
+    // void foo(int idx) { ... }
+    //
+    // In such case, we need to check the new arguments of the new call site match the type of the parameters of the
+    // new function, because we didn't do type checking when we generate the new call site. This step is necessary,
+    // because in the old call site, it's allowed to use different data type as the index to retrieve the resources,
+    // however, the new function only allows one data type as the input argument.
+    void typeCastForNewArguments(
```
csyonghe, line 409: I wonder if we can factor this into the core specialization logic and make the index parameter a part of the specialization key, instead of doing an additional post process step.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -933,6 +995,18 @@ struct FunctionParameterSpecializationContext
...
+        // We need to handle a corner case where the new argument of
+        // the callee of this specialized function could be a use of
+        // NonUniformResourceIndex(), in such case, any indexing operation
+        // on the global buffer by using this new argument should be
+        // decorated with NonUniformDecoration. However, inside the new
+        // specialized function, we don't have that information anymore.
+        // Therefore, we will need to scan the new argument list to find out
+        // this case, and insert the NonUniformResourceIndex() instruction
+        // on the corresponding parameter of the new specialized function.
+        maybeInsertNonUniformResourceIndex(newFunc, funcInfo, callInfo);
```
csyonghe, line 1007: If we call the function once with no uniform and once without, what do we generate? It seems that non-uniform-ness also need to be a part of the specialization key.

tests/compute/nonuniformres-as-function-parameter.slang
jkwak-work, line 1: This test without your fix in the code generates the following GLSL,
```
MyStruct_0 func_0(int _S1)
{
    MyStruct_0 a_1;
    uint _S2 = globalBuffer_0[_S1]._data[uint(0)];
    a_1.a_0 = globalBuffer_0[_S1]._data[uint(0)];
    uint _S3 = _S2 + 1U;
    a_1.b_0 = _S3;
    a_1.c_0 = a_1.a_0 + _S3 + 1U;
    return a_1;
}

void main()
{
    uvec2 _S4 = gl_GlobalInvocationID.xy;
    MyStruct_0 _S5 = func_0(nonuniformEXT(_S4.x));
    MyStruct_0 _S6 = func_0(nonuniformEXT(int(_S4.y)));
```
Note that it emits `nonuniformEXT` and it seems correct.

The PR description says that
> NonUniformResourceIndex is ignored

But it doesn't seem to match to my observation in a way that "nonuniformEXT" is not ignored.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: The specialization is always correctly generated for the first call site, that is the reason we need a branch here.

For your example here, I assume you mean
```
int a = ...;
func(a);

uint b = ...
func(b);
```

then yes, tryFindSpecialization will fail in the second call site of `func`, and it will generate two different functions, so this function `typeCastForNewArguments` won't be called. And this is the easiest case.

But this function here is trying to resolve a more tricky case:
```
void func(Texture<float> tex);
```

considering this case, it won't generate two different functions for the two call sites as follow
```
int int_index = ...
func(  g_tex[int_index])


uint uint_index
func(  g_tex[uint_index])
```

the reason is that we use the function name and parameter type as the key in the dictionary, so in the 2nd call site, it will find the function in the dictionary.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Yea, I was thinking about that path too, but it takes me more time than implementing this path, and this path seems more straightforward (at least to me). So I just proposed this version of solution.

But since you raised the concern here, I think I can spend more time to consider that refactored path.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Right, this is a real concern.
So it looks like we really need to change the key to make it including more modifiers.

source/slang/slang-ir-specialize-function-call.cpp
csyonghe: Right, so we need to find a way to factor that `index` parameter's type into the specialization key in `callInfo`.

csyonghe: If we can do this, it will also be straightforward to also encode the uniform-ness of the `index` parameter in the same `callInfo`.

tests/compute/nonuniformres-as-function-parameter.slang
kaizhangNV: Correct.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Today I revisited this problem, and I think we're fine in this case.
Because I noticed that the `index` parameter has already encode the uniformity info, so if there is another call site with  no uniform, the existing logic will specialize a new function for this call.

I actually add a new test case to cover this situation.

And this also replies your question here: https://github.com/shader-slang/slang/pull/4769#discussion_r1702221574

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: In the new change, I add the data type of the `index` parameter into the key so we will specialize a new function for this case.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -620,6 +565,11 @@ struct FunctionParameterSpecializationContext
             // the arguments at the new call site, and
             // don't add anything to the specialization key.
             //
+            // if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
```
csyonghe, line 568: We should not be checking in commented code.

tests/compute/nonuniformres-as-function-parameter.slang
```diff
@@ -31,65 +37,74 @@ MyStruct func(RWStructuredBuffer<uint> buffer)
...
     RWStructuredBuffer<uint> buffer = globalBuffer[nonUniformIdx];
 
-    // CHECK-SPV: %[[VAR3:[a-zA-Z0-9_]+]] = OpBitcast %int %bufferIdx
-    // CHECK-SPV: %[[VAR1]] = OpAccessChain %_ptr_StorageBuffer_RWStructuredBuffer{{.*}} %{{.*}} %[[VAR3]]
+    // CHECK_SPV: %[[VAR1]] = OpAccessChain %_ptr_StorageBuffer_RWStructuredBuffer{{.*}} %{{.*}} %bufferIdx
 
-    // CHECK-GLSL-SPV: %[[VAR1:[a-zA-Z0-9_]+]] = OpCopyObject %int %{{.*}}
-    // CHECK-GLSL-SPV: %[[VAR2:[a-zA-Z0-9_]+]] = OpAccessChain %_ptr_Uniform_uint %globalBuffer_0 %[[VAR1]] %int_0 %int_0
-    // CHECK-GLSL-SPV: %[[VAR3:[a-zA-Z0-9_]+]] = OpLoad %uint %[[VAR2]]
+    // CHECK_GLSL_SPV: %[[VAR1:[a-zA-Z0-9_]+]] = OpCopyObject %int %{{.*}}
```
csyonghe, line 67: should this be just `[[VAR1]]` since `VAR1` is already captured in the above `OpDecorate` lines?

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -551,6 +553,10 @@ struct FunctionParameterSpecializationContext
             // We start by recursively setting up whatever
             // `oldBase` needs:
             //
+            // TODO: We should add check 2 more things in the key
```
csyonghe, line 556: Is this still todo?

source/slang/slang-ir-specialize-function-call.cpp
csyonghe: I am not getting why you say "the existing logic will specialiize a new function for this call`. It seems that according to `getCallInfoForArg`, the `index` argument is not added to the specialization key, so if you have two callsites on the same resource array, one with nonuniform index and the other with uniform index, we will reuse the same specialization instead of producing different specializations.

Your test case is producing two specializations because the two calls feeds in different buffer arrays.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: No, I will just remove that.

tests/compute/nonuniformres-as-function-parameter.slang
kaizhangNV: Right, I will just remove [[VAR1]] here. it's a typo.

kaizhangNV: Oh, after looking at it carefully, it's not a typo, it's on purpose.
Because we need to check the inst assigning to VAR is decorated as `NonUniform`

kaizhangNV: you can see VAR1 ~ VAR12 are all decorated as `NonUniform`.

csyonghe:
I posted this in on one of the threads above, but I'll repost here in case you miss it:

I am not getting why you say "the existing logic will specialiize a new function for this call. It seems that according to getCallInfoForArg, the index` argument is not added to the specialization key, so if you have two callsites on the same resource array, one with nonuniform index and the other with uniform index, we will reuse the same specialization instead of producing different specializations.

Your test case is producing two specializations because the two calls feeds in different buffer arrays.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: In `getCallInfoForArg`, we add `oldIndex` into the call info key:
`ioInfo.key.vals.add(oldIndex);`

If one callsite use nonuniform index, then this `oldIndex` will be `kIROp_NonUniformResourceIndex`, and if there is another callsite using uniform index, it will be other type of inst instead of `kIROp_NonUniformResourceIndex`.

So we will have two different keys for those two callsites.

I can change my test to use the same buffer arrays, and it will still generate two different functions for those two callsites.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: act

tests/compute/nonuniformres-as-function-parameter.slang
kaizhangNV: ACK

kaizhangNV:
> I posted this in on one of the threads above, but I'll repost here in case you miss it:
> 
> I am not getting why you say "the existing logic will specialiize a new function for this call. It seems that according to getCallInfoForArg, the index` argument is not added to the specialization key, so if you have two callsites on the same resource array, one with nonuniform index and the other with uniform index, we will reuse the same specialization instead of producing different specializations.
> 
> Your test case is producing two specializations because the two calls feeds in different buffer arrays.

I replied the question in above comment: https://github.com/shader-slang/slang/pull/4769#discussion_r1714651292.

But we can continue the discussion on this thread as well.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Please look at my updated test, I already make two callsites on the same buffer, one with nonuniform index, and one without.

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -559,6 +561,11 @@ struct FunctionParameterSpecializationContext
             // the arguments at the new call site, and
             // don't add anything to the specialization key.
             //
+            // We should also add 2 more things such that our specialization
+            // can handle the corner cases that if the oldBase is a nonuniform
+            // resource and also the data type of oldIndex will be handled correctly.
+            // By doing so, just add the oldIndex to the key of call info.
+            ioInfo.key.vals.add(oldIndex);
```
csyonghe, line 568: I think you just want to add uniformity and the type of the index to the key, instead of the index inst itself. This will be too conservative, causing us to generate too many specializations.

source/slang/slang-ir-specialize-function-call.cpp
csyonghe: I saw that change now, but as mentioned in the thread above, I think adding the index inst directly into the key is too much. We should only add the type of the index and the uniformity to the key.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Well, that will work as well.
I have one question, for nonuniform index, can we just add the `oldIndex` directly? I don't know how to represent the uniformity, I guess the `kIROp_NonUniformResourceIndex` inst already a good way to represent it.

For others, I can just use the type inst to represent.

csyonghe: We can introduce an OpCode calls `kIROp_NonUniformAttr`, and then form a `IRAttributedType(int, NonUniformAttr)` to represent the type of the index inst, and use this as the key.

csyonghe: See `IRBuilder::getAttributedType`

kaizhangNV: I make the change that uses `IRAttributedType` as the key now.

source/slang/slang-ir-specialize-function-call.cpp
kaizhangNV: Let's close this thread as duplication below.

csyonghe (dismissed):

source/slang/slang-ir-specialize-function-call.cpp
```diff
@@ -559,6 +561,21 @@ struct FunctionParameterSpecializationContext
...
             // don't add anything to the specialization key.
             //
+            // We should also add 2 more things such that our specialization
+            // can handle the corner cases that if the oldBase is a nonuniform
+            // resource and also the data type of oldIndex will be handled correctly.
+            // By doing so, we form an IRAttributedType to include both information
+            // and add it to the key of call info.
+
+            List<IRAttr*> irAttrs;
+            if (oldIndex->getOp() == kIROp_NonUniformResourceIndex)
```
csyonghe, line 571: we probably also need to scan through any int casts to discover the nonuniform inst.
e.g. IntCast(IntCast(NonUniform(x))) is still a nonuniform.

kaizhangNV: Is `cast` the only operator that doesn't change the uniformity?
How about others?
e.g. NonUniform(x)+1, NonUniform(x) * 2?

I'd like to handle them all.

csyonghe: I think just handling cast right now would be good enough.
The general case should be handle with a proper data flow analysis, and I am fine if we scope that out of this PR.

kaizhangNV: OK, push a new commit to cover this.

csyonghe (approved):

PR #4772, opened 2024-08-02
Support unicode identifier names.

### LINKED ISSUE #4553: Unicode support
Hi there,

I'm looking at converting [a project of mine](https://github.com/ashpil/moonshine) to use slang rather than HLSL.

I use unicode extensively in my math code, as I believe it is clearer to write e.g., `α` instead of `alpha`, or `Λ` instead of `lambda`.

Looks like Slang fails on this: 

```hlsl
shaders/hrtsystem/material.hlsl(32): error 10000: illegal character (0xB1)
        float α2 = pow(α, 2);
                         ^
```
(and many more)

Is this an intentional design choice? Or is there a plan to support unicode in the future?

csyonghe:
Closes #4553.

The main fix here is to handle unicode utf8 bytes during lexing, and treat all unicode characters as a valid identifier character.

### DIFF
--- a/source/compiler-core/slang-lexer.cpp
+++ b/source/compiler-core/slang-lexer.cpp
@@ -5,9 +5,9 @@
 // input bytes and turning it into semantically useful tokens.
 //
 
+#include "core/slang-char-encode.h"
 #include "slang-name.h"
 #include "slang-source-loc.h"
-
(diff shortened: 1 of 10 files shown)


jkwak-work (approved):
I love this.
This can open a door to a very interesting use of Slang coding style for non-english languages.

Issue #4698, opened 2024-07-19
DXIL validation errors with uninitialized fields


venkataram-nv:
Compiling the following shader results in problems at the DXIL validation stage.

```hlsl
struct RadianceHitInfo { int x; };
struct Attributes      { float2 bary; };
struct ShadingState    { float4 tex; };

RWStructuredBuffer<float4> result;

void mdl_init(inout ShadingState state) {}

[shader("closesthit")]
void closesthit(inout RadianceHitInfo payload, Attributes attrib)
{
    ShadingState mdl_state;
    mdl_init(mdl_state);
    result[0] = mdl_state.tex.xyzw;
}
```

When compiling with `slangc [file] -target dxil -stage closesthit -entry closesthit` the following error is displayed:

```
dxc 1.7: hlsl.hlsl(52): error :  Assignment of undefined values to UAV.
dxc 1.7: note : note: at 'call void @dx.op.rawBufferStore.f32(i32 140, %dx.types.Handle %4, i32 0, i32 0, float undef, float undef, float undef, float undef, i8 15, i32 4)' in block '#0' of function '?closesthit@@YAXURadianceHitInfo_0@@UAttributes_0@@@Z'.
dxc 1.7: note : Validation failed.
```

These errors can be resolved by initializing the `tex` field of `mdl_state`.

The proposed two-step solution is as follows:
1. Emit warnings for never assigning to `out` / `inout` / `__ref` parameters. For `inout` we can request the user to fallback to `in` if necessary.
2. Transition uninitialization warnings to errors when emitting DXIL.

jkwak-work:
Are `payload` and `attrib` unrelated to the issue?

csyonghe:
This gets into another tricky detail about dxc. If the uninitialized variable is never used, then dxc will allow that code to pass compilation. This means that if we emit errors, we can potentially break existing code that worked fine, so this may need to stay as a warning for now. We can implement a `-validation-level=strict/relaxed` compiler option that makes this an error under `strict`, and for now we make `relaxed` as default but switch to `strict` some time in the future with enough advance notice.

venkataram-nv:
> Are payload and attrib unrelated to the issue?

Yes, its only there so I can have a raytracing shader entrypoint which I can compile to DXIL.

PR #4777, opened 2024-08-05
Warn when inout parameter is never written

venkataram-nv:
Addresses #4698 as one approach to diagnose the potential problem.

A new intrinsic function `unmodified(inout T)` has been added to explicitly indicate that an `inout` variable will not be modified in the function.

This is only one way to address the specific validation error in #4698. In general it seems that DXC does some more extensive checks on actual struct fields (as opposed to observing arbitrary struct writes), so that will be the next step.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2029,6 +2029,7 @@ T reinterpret(U value);
 // an out paramter.
 __generic<T>
 [__readNone]
+[semanticmutating]
 [ForceInline]
 void unused(inout T){}
 
@@ -2643,6 +2644,9 @@ attribute_syntax [__vulkanHitAttributes] : VulkanHitAttributesAttribute;
 
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [mutating] : MutatingAttribute;
+
+__attributeTarget(FunctionDeclBase)
+attribute_syntax [semanticmutating] : SemanticMutatingAttribute;
 
 __attributeTarget(SetterDecl)
 attribute_syntax [nonmutating] : NonmutatingAttribute;
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2029,7 +2029,6 @@ T reinterpret(U value);
 // an out paramter.
 __generic<T>
 [__readNone]
-[semanticmutating]
 [ForceInline]
 void unused(inout T){}
 
@@ -2650,9 +2649,6 @@ attribute_syntax [__vulkanHitAttributes] : VulkanHitAttributesAttribute;
 
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [mutating] : MutatingAttribute;
-
-__attributeTarget(FunctionDeclBase)
-attribute_syntax [semanticmutating] : SemanticMutatingAttribute;
 
 __attributeTarget(SetterDecl)
 attribute_syntax [nonmutating] : NonmutatingAttribute;
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2025,16 +2025,14 @@ T reinterpret(U value);
 
 // Use an otherwise unused value
 //
-// This can be used to silence the warning about returning before initializing
-// an out paramter.
+// This can be used to silence the warning about returning before initializing an out paramter.
 __generic<T>
 [__readNone]
 [ForceInline]
 __intrinsic_op($(kIROp_Unmodified))
 void unused(inout T){}
 
-// This can be used to silence the warning about not writing before initializing
-// an out paramter.
+// This can be used to silence the warning about not writing to an inout parameter.
 __generic<T>
 [__readNone]
 [ForceInline]
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2039,7 +2039,7 @@ __generic<T>
 [__readNone]
 [ForceInline]
 __intrinsic_op($(kIROp_Unmodified))
-void unmodified(inout T){}
+void unmodified(out T){}
 
 // Specialized function
 
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2030,7 +2030,16 @@ T reinterpret(U value);
 __generic<T>
 [__readNone]
 [ForceInline]
+__intrinsic_op($(kIROp_Unmodified))
 void unused(inout T){}
+
+// This can be used to silence the warning about not writing before initializing
+// an out paramter.
+__generic<T>
+[__readNone]
+[ForceInline]
+__intrinsic_op($(kIROp_Unmodified))
+void unmodified(inout T){}
 
 // Specialized function
 
--- a/source/slang/diff.meta.slang
+++ b/source/slang/diff.meta.slang
@@ -273,7 +273,7 @@ struct TensorView
     __subscript(uint index) -> T
     {
         [ForceInline] [__NoSideEffect] get { return load(index); }
-        [ForceInline] set { store(index, newValue); }
+        [ForceInline] [semanticmutating] set { store(index, newValue); }
 
         [__NoSideEffect]
         ref
@@ -288,7 +288,7 @@ struct TensorView
     __subscript(uint i1, uint i2) -> T
     {
         [ForceInline] [__NoSideEffect] get { return load(i1, i2); }
-        [ForceInline] set { store(i1, i2, newValue); }
+        [ForceInline] [semanticmutating] set { store(i1, i2, newValue); }
         [__NoSideEffect]
         ref
         {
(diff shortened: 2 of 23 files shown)


tests/diagnostics/inout-never-written.slang
```diff
@@ -0,0 +1,47 @@
...
+//CHK-DAG: ([[# @LINE + 1]]): warning 41022: inout parameter 'state' is never written to
+void state_never_assigned(inout State state, inout float v)
+{
+    v = state.v.x;
+}
+
+void state_assigned(inout State state)
+{
+    int3 vi = __slang_noop_cast <int3, float3> (state.v);
+    int3 ni = __slang_noop_cast <int3, float3> (state.n);
```
jkwak-work, line 22: I think you should use `__intCast`

tests/diagnostics/inout-never-written.slang
```diff
@@ -0,0 +1,47 @@
...
+};
+
+__generic <T>
+struct B
+{
+    int state;
+
+    //CHK-DAG: ([[# @LINE + 1]]): warning 41023: method marked `[mutable]` but never modifies `this`
+    [mutating] int next() { return state; }
+};
```
jkwak-work, line 44: A case I am worried about is when the inheritance is involved.
Consider the following code snippet where one of derived structs mutates and the other one doesn't.
```
Interface Base
{
    [mutating] void doSomething();
};

struct NotMutating : Base
{
    [mutating] void doSomething() {} // warning will be printed.
};

struct Mutating : Base
{
    int a = 0;
    [mutating] void doSomething() { a = 1; }
}
```
Note that `[mutating]` is a part of the function signature in Slang.
Would your new attribute solve this problem if I do the following?
```
struct NotMutating : Base
{
    [semanticmutating] void doSomething() {} // is [mutating] still a part of this function signature?
};
```

source/slang/core.meta.slang
```diff
@@ -2650,6 +2651,9 @@ attribute_syntax [__vulkanHitAttributes] : VulkanHitAttributesAttribute;
 __attributeTarget(FunctionDeclBase)
 attribute_syntax [mutating] : MutatingAttribute;
 
+__attributeTarget(FunctionDeclBase)
+attribute_syntax [semanticmutating] : SemanticMutatingAttribute;
```
jkwak-work, line 2655: `semanticmutating` without spacing, underscore and camel-casing is little hard to read.

If this is going to be used only internally, it should start with double underscore like `[__semanticmutating]`.
If it will be used by the users, we should put this in the document and explain how it works with an example.

jkwak-work (commented):
I am not sure if we want to print warnings for `inout`.
I think the intention here is to keep the behavior of DXC in Slang.
But adding a new attribute seems too intrusive for snoozing a warning.

tests/diagnostics/inout-never-written.slang
venkataram-nv: The idea is that it will indeed solve this case (I will confirm locally). The difference between `[semanticmutating]` and `[mutating]` is that the latter will force the `this` parameter to be `InOut` whereas this is not the case with `[semanticmutating]`.

source/slang/core.meta.slang
venkataram-nv: Good point, I think for now making it an internal feature only is enough, since it was necessary to only fix warnings in the stdlib.

csyonghe:
I don't understand what problem we are solving here with semanticnonmutating. Adding attribute like this can have a lot of cascading consequences to the type and I would rather us take a different approach to solve the problem if possible.

csyonghe:
Isn't the solution to the issue 4698 be that we warn if the user is passing an uninitialized value to an inout parameter?

csyonghe:
If we really want to be accurate, we can figure out which fields of an inout parameter will a function write into, and consider only those fields as initialized after the call. So the initialization state if a value is isn't just a single scalar, but a tuple with one element for each field.

venkataram-nv:
@csyonghe 

> Isn't the solution to the issue 4698 be that we warn if the user is passing an uninitialized value to an inout parameter?

The way I am thinking about this is that the user indicated the intention to modify the parameter with `inout` but is never actually writing to it inside the function. For example if a struct is partially uninitialized, and the user wants to update a (uninitialized) field of it using the initialized part but forgets to do so, these warnings could be useful.

> If we really want to be accurate, we can figure out which fields of an inout parameter will a function write into, and consider only those fields as initialized after the call.

That is right, but then we should also consider nested structs. I am planning to tackle this as the next step since it seems that DXC is doing this.

> I don't understand what problem we are solving here with semanticnonmutating. Adding attribute like this can have a lot of cascading consequences to the type and I would rather us take a different approach to solve the problem if possible.

With the current `inout` warnings, some of the stdlib functions will generate warnings (i.e. `unused(inout T)` and the `TensorView` subscript `set` methods). Since they are currently only internal functions, I was thinking of refactoring to `__semanticmutating` like Jay suggested.

csyonghe:
We can have warnings if `inout` is not modified. We can provide a way to silence that warning by providing
```
__intrinsic_op(kIROp_Unmodified)
unmodified<T>(inout T v);
```
in the stdlib. The ir validation pass can then treat `Unmodified(v)` as an operation that initializes `v` inside the function to silence the warning.

csyonghe:
Note that whenever we add a warning, we must also provide a way for the user to properly silence the warning. And I think `unmodified(x);` is a more proper way to do so.

tests/diagnostics/inout-never-written.slang
```diff
@@ -0,0 +1,51 @@
...
+{
+    int state;
+
+    //CHK-DAG: ([[# @LINE + 1]]): warning 41023: method marked `[mutable]` but never modifies `this`
+    [mutating] int next() { return state; }
+    
+    // TODO: unmodified(...)
+    [mutating] int progress()
+    {
+        unmodified(state);
```
jkwak-work, line 36: Hmmm. This makes sense to me but I was expecting more like, `unmodified(this)`.
What happens if I use `this`? Would it snooze the warning, too?

jkwak-work (approved):
Looks good to me.

jkwak-work:
When you squash and merge, don't forget to revise the final commit message because your final code change now looks different.

tests/diagnostics/inout-never-written.slang
venkataram-nv: It would also work. For this example `state` was the only field of the structure, so I was thinking it would be sufficient just to have `unmodified(state)`.

tests/pipeline/ray-tracing/trace-ray-inline.slang
```diff
@@ -55,7 +55,7 @@ void myTriangleClosestHit(inout MyRayPayload payload)
 // In this case, the return value is used to specify
 // whether the hit should be accepted (true) or ignored (false).
 //
-bool myTriangleAnyHit(inout MyRayPayload payload)
+bool myTriangleAnyHit(MyRayPayload payload)
```
jkwak-work, line 58: I think we should keep `inout` modifier here.
And snooze it with `unused`.

source/slang/diff.meta.slang
```diff
@@ -273,7 +273,12 @@ struct TensorView
     __subscript(uint index) -> T
     {
         [ForceInline] [__NoSideEffect] get { return load(index); }
-        [ForceInline] set { store(index, newValue); }
+        
+        [ForceInline] set
+        {
+            unmodified(this);
+            store(index, newValue);
+        }
```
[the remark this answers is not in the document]
venkataram-nv: Yes.

tests/pipeline/ray-tracing/trace-ray-inline.slang
```diff
@@ -73,7 +74,7 @@ void myProceduralClosestHit(inout MyRayPayload payload, MyProceduralHitAttrs att
 {
 	payload.value = attrs.value;
 }
-bool myProceduralAnyHit(inout MyRayPayload payload)
+bool myProceduralAnyHit(MyRayPayload payload)
```
jkwak-work, line 77: I think we should keep `inout` and use `unmodified`

tests/pipeline/ray-tracing/trace-ray-inline.slang
```diff
@@ -85,7 +86,7 @@ bool myProceduralAnyHit(inout MyRayPayload payload)
 // For now we will only deal with the single-intersection
 // case.
 //
-bool myProceduralIntersection(inout float tHit, inout MyProceduralHitAttrs hitAttrs)
+bool myProceduralIntersection(float tHit, MyProceduralHitAttrs hitAttrs)
```
jkwak-work, line 89: I think we should keep `inout` and use `unmodified`

jkwak-work:
Looks good for the intention of PR.
But I am not sure if we need `unmodified` when `unused` supersedes the functionality already.

csyonghe:
If you have an `inout` parameter where your function do read from the parameter but didn't write to the parameter, it isn't very correct to say the parameter is `unused`. That's why we should have both `unmodified` and `unused`. The current implementation can be handwavy by treating `unmodified` as `unused`, but that doesn't mean they are conceptually equivalent things.

jkwak-work (approved):
Looks good to me.

Issue #4784, opened 2024-08-06
macOS binaries should not rely on latest SDK


skallweitNV:
Currently the slang release binaries require the latest 10.15 macOS version.
We can build slang for a different SDK using `CMAKE_OSX_DEPLOYMENT_TARGET`.
This unfortunately doesn't work at the moment because slang uses `std::filesystem` in a few places, which is only supported on the latest macOS version.
It would be best to break that dependency and implement custom routines for filesystem handling.

Issue #4764, opened 2024-07-31
Create a Slang::Path::createDirectories function to create directory recursively


kaizhangNV:
Now, we only have `Slang::Path::createDirectory` function to create a directory, but this function cannot create directories recursively.
So we need to create one function to implement that, such that we can avoid using `std::filesystem`.

PR #4785, opened 2024-08-06
Reduce dependency on std library

skallweitNV:
This is a first effort on reducing dependency on the std library, which is needed for #4784.

Usage of `std::string` is replaced with `Slang::String`.
Usage of `std::filesystem` is replaced with `Slang::String`, `Slang::File` and `Slang::Path` utilities.

`<filesystem>` is now only used in `slang-io.cpp` for `getRelativePath` and `createDirectoryRecursive` implementations. These should be replaced in future PRs.


csyonghe (approved):

csyonghe:
@kaizhangNV I didn't capture this in the review but we should never use std::string or anything that has an equivalent in `core`.

kaizhangNV:
@csyonghe I noticed that this change still uses `std::filesystem::create_directories`.
I think we already talked about this and created a task #4764 for this.

PR #4804, opened 2024-08-07
GitHub action benchmark

venkataram-nv:


jkwak-work (commented):
Looks good to me.

source/core/slang-performance-profiler.cpp
```diff
@@ -31,11 +31,14 @@ namespace Slang
         }
         virtual void getResult(StringBuilder& out) override
         {
-            for (auto func : data)
+            char buffer[512];
+            for (const auto& func : data)
             {
-                out << func.key << ": \t";
+                memset(buffer, 0, sizeof(buffer));
+                snprintf(buffer, sizeof(buffer), "[*] %30s", func.key);
```
jkwak-work, line 38: `memset` is unnecessary, but you may want to make sure that it doesn't cause a buffer overflow.
And 512 seems too big for outputting only 36 characters.
```
char buffer[40];
snprintf(buffer, sizeof(buffer), "[*] %30s", func.key);
buffer[39] = 0;
```

csyonghe, line 38: ms?

csyonghe (approved):

source/core/slang-performance-profiler.cpp
venkataram-nv: The `30s` is for the formatting string

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,47 @@
...
+          cmake --workflow --preset release
+      - name: Setup
+        run: |
+          cd tests/mdl
+          pip install prettytable argparse
+      - name: Run benchmark
+        run: |
+          cd tests/mdl
+          Copy-Item -Path C:\slang-benchmarks -Destination . -Recurse
+          python compile.py --samples 16 --target dxil --ci
```
jkwak-work, line 47: As far as I understand, `tests` directory is just for `slang-test` and nothing else.
You may want to move `compile.py` to a new directory like... `external/mdl` or `tools/mdl`.

jkwak-work (commented):
Looks good to me.
But @expipiplus1 will need to review this PR.

source/core/slang-performance-profiler.cpp
venkataram-nv: We don't have a restriction on the length of `func.key` so I am using a large buffer just in case (so its not just 36 chars).

jkwak-work: Aha. Sorry. My mistake. I thought it was printing only the first 30 characters.

Issue #4812, opened 2024-08-09
"macos-aarch64" package from the release works for Intel chip but not for M3 chip.


jkwak-work:
**Problem description**
When I download the binary release package for macos-aarch64, slang-test passes on all tests even when they should fail.
This was due to the fact that slang-llvm was not actually working; and tests are reported passing incorrectly.

It turned out that the release binary works fine for MacBook Pro 2019 that uses an Intel chip.
But it doesn't work when the chip is M3.

A workaround for this is to locally rebuild slang-llvm when the hardware uses M3 chip.

The issue appears to be limited only to slang-llvm binary; not slangc or other binaries in the release package.

**Goal**
Because the package name has "aarch64" suffix, the binary package should run on M3 chip.

jkwak-work:
@expipiplus1 , do you have any guesses of why this is happening?

expipiplus1:
This should be fixed by https://github.com/shader-slang/slang/pull/4847 If you could give this a test please :)

PR #4813, opened 2024-08-09
Update build steps for MacOS

jkwak-work:
This commit adds a quick build step for MacOS on CONTRIBUTION.md.

It is unclear what are the conditions that slang-llvm needs to be rebuilt locally. An issue is filed for it, #4812


csyonghe (approved):
This is very helpful, thank you for updating the instructions!

PR #4815, opened 2024-08-09
User guide page for SPIR-V target specific information

### LINKED ISSUE #4814: Add user-guide page for SPIR-V target specific functionalities
We need a user-guide document for target specific Slang functionalities.
We should start from SPIR-V target and incrementally append more details for other targets.

Following contents are expected to be described in the document.
- Target specific Compiler options for SPIR-V
- Target specific Attributes for SPIR-V
- Target specific Slang intrinsics for SPIR-V
- Memory pointer for SPIR-V
- Matrix majorness (row/column) for SPIR-V
- Controlling layout per StructuredBuffer.

jkwak-work:
closes #4814


docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,101 @@
...
+
+[spv_target_env_1_3]
+
+[vk_image_format(format : String)]
+
+
+Target specific Slang intrinsics
+--------------------------------
+
+TODO: Cannot think of anything to put here. `spirv_asm` is already well documented in a1-04-interop.md.
```
ArielG-NV, line 73: I think we should link to `a1-04-interop.md`

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,101 @@
+---
+layout: user-guide
+permalink: /user-guide/target-specifics
+---
+
+Target specific functionalities
+============================
+
```
ArielG-NV, line 8: We likely should add a section on "compiling to specific spirv versions".

We should explain how capabilities can affect the compiled spirv version for `-target spirv` and `emit-spirv-via-glsl`:
`-capability spirv_1_5` targets 1.5, `-capability spirv_1_6` targets 1.6, etc

jkwak-work: Thanks for reminding me of that.
I think it makes sense to talk about the compatibility profiles that are specific to SPIR-V and others.
But I am little hesitant that it may blot the document with not much of actual information.

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+This chapter provides information for target specific functionalities and behaviors.
+
+Target specific Compiler options
```
csyonghe, line 11: Let's rename this to "SPIRV specific ***"

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+
+TODO: Need to digest the comments writting in "compute/buffer-layout.slang" and write up a user-friendly document here.
+
+
+Matrix majorness (row/column)
+-----------------------------
+
+### Matrix majorness for SPIR-V
+
+TODO: Not sure how this is handled.
```
csyonghe, line 136: I think we just want to say that we are following the same behavior in DXC to translate a N-row-by-M-column matrix in Slang maps to a N-column-by-M-row matrix in GLSL/SPIRV. If you can talk more about the reasoning behind this decision, that would be great.

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+Buffer layout and alignment
+---------------------------
+
+TODO: Need to digest the comments writting in "compute/buffer-layout.slang" and write up a user-friendly document here.
+
+
+Matrix majorness (row/column)
+-----------------------------
+
+### Matrix majorness for SPIR-V
```
csyonghe, line 134: This title can be "Matrix type translation"

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+    TODO
+
+- [vk::spirv_instruction(op : int, set : String = "")]
+    It is similar to how `__intrinsic_op` works, but it uses the `op-code` of SPIR-V instruction set. You can also specify which instruction set you want to use as a second argument. E.G., `[[vk::spirv_instruction(1, "NonSemantic.DebugBreak")]]`
+
+- [spv_target_env_1_3]
+    TODO: I am not sure what it does. There isn't an example on slang-test
+
+
+Target specific Slang intrinsics
```
csyonghe, line 106: We want to mention that ConstantBuffer translates to uniform buffer, RWByteAddressBuffer, RWStructuredBuffer, ByteAddressBuffer, StructuredBuffer translates to a shader storage buffer, where RW** buffers will have a `readwrite` layout on the buffer and the non RW ones have readonly layout.

Also want to talk about how you can control the buffer layout globally with `-fvk-use-scalar-layout`, or individually on `StructuredBuffer` with `StructuredBuffer<T, Std140Layout>` etc.

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+
+Matrix majorness (row/column)
+-----------------------------
+
+### Matrix majorness for SPIR-V
+
+TODO: Not sure how this is handled.
+
+
+Summary
```
csyonghe, line 139: Also want to talk  about how we translate tessellation shaders to SPIRV. Like DXC, we will merge both the patch function and the domain shader into a single tessellation control shader.

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
+---
+layout: user-guide
+permalink: /user-guide/target-specifics
+---
+
+Target specific functionalities
```
csyonghe, line 6: should have a section to talk about multi-entrypoint support.

docs/user-guide/11-target-specifics.md
csyonghe: And then we need to talk about how `ParameterBlock<T>` and `ConstantBuffer<T>` get laid out to descriptor sets.

csyonghe:
We should also have sections to talk about type legalization and mention all the cases that we enable with our legalization passes.

For example, we allow using resource types in structs, we allow functions that return resource types as return type or `out` parameter, as long as things are statically resolvable. We allow functions that return arrays, we allow putting scalar/vector/matrix/array types directly as element type of a constant buffer or structured buffers. Make sure to go over our legalization passes to build a list of allowed things that people are unsure whether it is allowed.

docs/user-guide/11-target-specifics.md
```diff
@@ -0,0 +1,144 @@
...
+    It can be used on a buffer block that represents a buffer within a shader record as defined in the Ray Tracing API.
+
+- [vk::aliased_pointer]
+    TODO
+
+- [vk_restrict_pointer]
+    TODO
+
+- [vk::spirv_instruction(op : int, set : String = "")]
+    It is similar to how `__intrinsic_op` works, but it uses the `op-code` of SPIR-V instruction set. You can also specify which instruction set you want to use as a second argument. E.G., `[[vk::spirv_instruction(1, "NonSemantic.DebugBreak")]]`
```
csyonghe, line 100: let's not mention `__intrinsic_op`, it is for slang devs only.

csyonghe:
Talk about how we make use of specialization instead of pervasive inlining to preserve the code structure in our `-O0` output to aid debugging.

Talk about nonsemantic debug info and mention renderdoc support.

csyonghe:
We should also mention major HLSL feature sets that are supported for SPIRV: ray tracing, inline ray tracing, mesh shader, tessellation shader, geometry shader, wave intrinsics, barriers, atomics etc. We should allow more atomics intrinsics in spirv than for HLSL. I believe we have native float16 and float32 atomics for SPIRV.

csyonghe:
Talk about `discard` and how they are translated to `OpKill` in spirv 1.5 or earlier, and to `OpDemoteToHelperInvocation` in spirv 1.6, and how you can toggle it to generate `OpDemoteToHelperInvocation` by providing `SPV_EXT_demote_to_helper_invocation` capability via `-capability` or `-profile`.

csyonghe:
Mention that producing SPIRV 1.0, 1.1 and 1.2 is still in experimental state and the compiler may still produce spirv that uses things introduced in later spirv versions when asked to generate these older spirv versions.

csyonghe:
Need to talk about `Sampler2D` etc. for combined texture samplers.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+| SPIR-V |   Yes           |     Yes         |    Yes (SPV1.5+ext)   | Yes (SPV1.5+ext) | Yes (SPV1.5+ext)
++--------+-----------------+-----------------+-----------------------+------------------+------------------+
+
+
+ConstantBuffer, (RW)StructuredBuffer (RW)ByteAddressBuffer
+----------------------------------------------------------
+
+Each member in a `ConstantBuffer` will be emitted as `uniform` parameter.
+StructuredBuffer and ByteAddressBuffer are translated to a shader storage buffer with `readonly` layout.
+RWStructuredBuffer and RWByteAddressBuffer are translated to a shader storage buffer with `read-write` layout.
```
csyonghe, line 90: Should talk about `StructuredBuffer<T, Std140Layout>`, also supports `Std430Layout` and `ScalarLayout`.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+```
+It will generate the following GLSL or SPIR-V code.
+> layout(r32f) uniform image2D typicalTexture_0;
+> %18 = OpTypeImage %float 2D 2 0 0 2 R32f
+
+### [vk::shader_record]
+Same as `shaderRecordEXT` layout qualifier in [GL_EXT_ray_tracing extension](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_ray_tracing.txt).
+It can be used on a buffer block that represents a buffer within a shader record as defined in the Ray Tracing API.
+
+### [vk::aliased_pointer]
```
csyonghe, line 195: this isn't supported right now.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+When two memory pointers point to a same memory location, they are considered `aliased`. Compiler needs to be informed of such cases in order to generate the code that produces a consistant outcome. Slang can figure out whether a pointer needs to be treated as aliased or not, but you can explicitly mark a variable as aliased.
+
+### [vk::restrict_pointer]
```
csyonghe, line 198: this isn't supported right now.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+Opposite to [vk::aliased_pointer] in a way that it tells the compiler to treat the variable as not-aliased.
+
+### [vk::spirv_instruction(op : int, set : String = "")]
```
csyonghe, line 201: this isn't supported right now.

csyonghe: slang users should use spirv_asm instead.

jkwak-work: I will remove it.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will remove it.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will include this line

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+RWStructuredBuffer and RWByteAddressBuffer are translated to a shader storage buffer with `read-write` layout.
+
+
+ParameterBlock for SPIR-V target
+--------------------------------
+
+`ParameterBlock` is a Slang generic type for binding uniform parameters.
+It is similar to `ConstantBuffer` in HLSL, but `ParameterBlock` can include not only constant parameters but also descriptors such as Texture2D or StructuredBuffer.
+
+Because the Vulkan API doesn't natively support `ParameterBlock` that has a mixture of descriptors and constants, Slang emits each member as an individual parameter.
```
csyonghe, line 99: This sentence can be a little bit confusing/misleading, let's just state what slang does if both ordinary data fields and resource typed fields exists in a parameter block: all ordinary data fields will be grouped together into a uniform buffer and appear as a binding 0 of the resulting descriptor set.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+  ...
+}
+```
+
+When targeting SPIR-V, the patch function is merged as a part of the Hull shader, because GLSL nor SPIR-V differentiates them like how HLSL does.
+
+As an example, a Hull shader will be emitted as following,
+```
+void main() {
+    // Set tessellation levels
```
csyonghe, line 293: we will first call the `constants` function, then writes to the outputs on thread 0, then a barrier, then run the `main` function.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+### [vk::spirv_instruction(op : int, set : String = "")]
+When applied to a function, the function will use the `op` value corresponding to a SPIR-V instruction. You can also specify which instruction set you want to use as a second argument. E.G., `[[vk::spirv_instruction(1, "NonSemantic.DebugBreak")]]`
+
+
+Multiple entry points support
+-----------------------------
+
+Slang supports mutiple entry points when targeting SPIR-V.
+It is same to HLSL and SPIR-V that allows a shader source to have multiple entry points.
+Note that GLSL requires the entry point to be named, "main", and a shader source can have only one entry point.
```
csyonghe, line 210: Mention that because of this, the current default behavior is to rename all entrypoints to "main" when emitting SPIRV. This default setting will prevent a SPIRV to have more than one entrypoint defined. To generate valid spirv with more than one entrypoints, use the `-fvk-use-entrypoint-name` compiler option to disable this renaming behavior and preserve the entry point names.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+
+### -g
+Include debug information in the generated code, where possible.
+When targeting SPIR-V, this option emits [SPIR-V NonSemantic Shader DebugInfo Instructions](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc).
+
+### -O<optimization-level>
+Set the optimization level.
+When targeting SPIR-V, this option applies to the downstream compiler.
+"-O0" option can help to identify issues when the problem is from the optimization step of the downstream compiler.
+Compared to the macro inlining, the specialization helps debugging with "-O0" option, because the specialization preserves the code structure much more than the macro inlining.
```
csyonghe, line 124: I think we want to say "Under "-O0" option, Slang will not perform extensive inlining for all functions calls, instead it will preserve the call graph as much as possible to help with understanding the SPIRV structure and diagnosing any downstream toolchain issues."

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+--------------------------------
+
+The following compiler options are specific to SPIR-V.
+
+### -emit-spirv-directly
+Generate SPIR-V output directly (default)
+It cannot be used with -emit-spirv-via-glsl
+
+### -emit-spirv-via-glsl
+Generate SPIR-V output by compiling generated GLSL with glslang
```
csyonghe, line 113: by compiling to glsl source first, then use glslang compiler to produce SPIRV from glsl.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+- UniformConstant: is for uniform constant variables as read-only.
+- Uniform: is for uniform buffer as read-only.
+- StorageBuffer: is for UAV/structuredBuffer/ShaderStorageBuffer as read-write.
+- PhysicalStorageBuffer: same as StorageBuffer but it follows non-logical addressing model.
+- PushConstant: is for variables as read-only stored in the buffer that has `push_constant` layout.
+
+Note that only ones starting with `Physical-` follows the non-logical addressing model.
+
+
+Difference between `std140` layout and `std430` layout
```
csyonghe, line 234: Don't need  to talk about this, spirv/vulkan users already know this very well.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+
+When constant parameters are send to the shader, there are different rules applied to the alignment and paddings.
+`std140` requires alignment of 16 bytes on `struct` types and it adds padding to round up to 16 bytes for each member.
+`std430` requires a natural alignment, which means that the required alignment is same as its size.
+
+
+Matrix type translation
+-----------------------
+
+When targeting SPIR-V, Slang assumes Row-Major matrix, which is same behavior as HLSL/DXC.
```
csyonghe, line 245: I think  this is more a terminology difference rather than an actual divergence.

Matrix type translation is orthogonal to row-major/column-major layout.

In this section we should focus on how matrix types are translated, and not touch into row-major and column-major.

Just have a table showing how hlsl matrices are translated to spirv types, similar to that table in the dxc doc. And say that we follow the same behavior in dxc.

Then we can go for the reasoning behind this design, in that what hlsl calls a row is what glsl will call a column. Essentially, a hlsl row or glsl column is:
- what `matrix[i]` returns.
- what matrices can be constructed from `float3x4(v1, v2, v3)`, or `mat3x4(v1,v2,v3)`.
If we map it the other way, simple operations like `matrix[i]` or construction will become very messy code that may lead to slower performance.

The only consequence of mapping float3x4 to `matrix(vector4, 3)` (or mat3x4, a 3-"column"-by-4-"row" matrix in spirv terminology) is that matrix-vector and matrix-matrix multiplication operations needs to have their operand ordering swapped, which is what slang will do:
```
mul(m, v) ==> v*m
```

csyonghe: Once we talk about this, then there is the row-major and column-major layout. Note that the decision to use row-major or column-major layout has nothing to do with how the matrix type itself is defined in hlsl or spirv. It really just a data layout modifier just like `std140` or `std430`.

We can talk about this in a separate section saying that by default, slang uses row-major layout for reasons documented in [a1-01-matrix-layout], as oppposed to defaulting to column major layout in dxc.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+
+Slang supports mutiple entry points when targeting SPIR-V.
+It is same to HLSL and SPIR-V that allows a shader source to have multiple entry points.
+Note that GLSL requires the entry point to be named, "main", and a shader source can have only one entry point.
+
+
+Memory pointer is experimental
+------------------------------
+
+SPIR-V currently supports five types of `Addressing model`-s, and it can be grouped as Logical addressing model and non-local addressing model.
```
csyonghe, line 216: We don't need to talk about these details here.

Just say that Slang support using pointers when targetting SPIRV. Then show an example of slang code with pointer usage.

Then say pointers will translate to a pointer of PhysicalStorageBuffer storage class in SPIRV. When a slang module uses pointers, the resulting SPIRV will be using the `SpvAddressingModelPhysicalStorageBuffer64` addressing mode. Modules without use of pointers will use `SpvAddressingModelLogical` addressing mode. (no need to go in details explaining what an address mode is).

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+GLSL 4.6 with [GLSL_EXT_shader_atomic_float](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_shader_atomic_float.txt) can use atomic operations for 32-bit float type.
+GLSL 4.6 with [GLSL_EXT_shader_atomic_float2](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GLSL_EXT_shader_atomic_float2.txt) can use atomic operations for 16-bit float type.
+
+SPIR-V 1.5 with [SPV_EXT_shader_atomic_float_add](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float_add.asciidoc) and [SPV_EXT_shader_atomic_float_min_max](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float_min_max.asciidoc) can use atomic operations for 32-bit float type and 64-bit float type.
+SPIR-V 1.5 with [SPV_EXT_shader_atomic_float16_add](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float16_add.asciidoc) can use atomic operations for 16-bit float type
+
++--------+-----------------+-----------------+-----------------------+------------------+------------------+
```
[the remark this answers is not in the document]
jkwak-work: I will double check but I don't think "atomic" is supported for 8-bit.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: Thanks for the information.
Since I am familiar with the reflection API, I wasn't sure.
The comment here is scraped from comments I picked up from the source code.
I will update it as you described.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I omitted the barrier and other parts for the sake of simplicity.
But I guess I shouldn't have.
I will rewrite the code including what you described.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: Aha. I didn't know that the option exists for this reason.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will change as suggested.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: Right. The meaning of my writing is not clear enough.
I will change it as suggested.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will remove it.
But I didn't know about it until I wrote this.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will change it as suggested.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I will change it as suggested.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,316 @@
...
+ - lowp : RelaxedPrecision, on storage variable and operation
+ - mediump : RelaxedPrecision, on storage variable and operation
+ - highp : 32-bit, same as int or float
+
+Slang ignores the keywords above and all of them are treated as `highp`.
+
+
+Supported atomic types for each target
+--------------------------------------
+HLSL 6.2 introduced [16-bit scalar types](https://github.com/microsoft/DirectXShaderCompiler/wiki/16-Bit-Scalar-Types) such as float16 and int16_t, but they didn't come with any atomic operations.
```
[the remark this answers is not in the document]
jkwak-work: Right. I will fix it.

docs/user-guide/11-target-specifics.md
jkwak-work: @ArielG-NV , can you give me some ideas of what could be helpful to include in this document regarding the capability specifically for spirv target?

I feel like the ones you described seems too generic and obvious.
>`-capability spirv_1-5` compiles spirv 1.5.

docs/user-guide/a1-06-spirv-target-specifics.md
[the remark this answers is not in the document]
jkwak-work: I double checked. There is no atomic operators for 8-bit integer type in SPIR-V.

docs/user-guide/a1-06-spirv-target-specifics.md
jkwak-work: I think the correct order is the following,
```
void main() {
    ...
    main(patch, gl_InvocationID);
    barrier();
    if (gl_InvocationID == 0)
    {
        constants(path);
    }
}
```
This matches to how Slangc currently emits them when tested with "tests/spirv/tessellation.slang"
It is implemented in a function, `invokePathConstantFuncInHullShader` [Link](https://github.com/shader-slang/slang/blob/f0ba756c2f982aac8095ff0928d048fc97548315/source/slang/slang-ir-glsl-legalize.cpp#L1049).

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -224,23 +224,48 @@ When a slang module uses a pointer, the resulting SPIRV will be using the SpvAdd
...
-TODO: The only consequence of mapping float3x4 to matrix(vector4, 3) (or mat3x4, a 3-"column"-by-4-"row" matrix in spirv terminology) is that matrix-vector and matrix-matrix multiplication operations needs to have their operand ordering swapped, which is what slang will do:
+An alternative way to emit SPIR-V code is to emit four vectors and each vector has three elements. Slang doesn't do this but this is a more direct translation because SPIR-V spec defines OpTypeMatrix to take "Column Count" not row.
 ```
-mul(m, v) ==> v*m
+%v3float = OpTypeVector %float 3 ; <= float3 type
+%mat4v3float = OpTypeMatrix %v3float 4 ; <= four of float3
 ```
+However, this results in a more complicated access pattern to the elements in a matrix, because `v[i]` will no longer correspond to a vector natively when emitted to SPIR-V.
+
+Another way to put, Slang treats column as row and row as column when targeting GLSL or SPIR-V, although SPIR-V and GLSL use Column-major matrix. This is same to [how DXC handles a matrix when emitting SPIR-V](https://github.com/Microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#appendix-a-matrix-representation).
```
csyonghe, line 264: let's try to avoid the terminlogy "column-major" in this context to avoid confusion over column major data layout. We can delete ", although SPIRV and GLSL use Column-major matrix".

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -224,23 +224,48 @@ When a slang module uses a pointer, the resulting SPIRV will be using the SpvAdd
...
 
-TODO: Once we talk about this, then there is the row-major and column-major layout. Note that the decision to use row-major or column-major layout has nothing to do with how the matrix type itself is defined in hlsl or spirv. It really just a data layout modifier just like std140 or std430.
+Due to the swap of row and column, the matrix multiplication needs to be performed little differently. Slang translates a matrix multiplication, `mul(mat1, mat2)`, to `transpose(mul(transpose(mat2), transpose(mat1)))` when targeting SPIR-V.
```
csyonghe, line 266: Due to the swap of row and column in terminology,

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -224,23 +224,48 @@ When a slang module uses a pointer, the resulting SPIRV will be using the SpvAdd
...
+float3x4 v;
+for (int i = 0; i < 3; ++i)
+{
+  for (int j = 0; j < 4; ++j)
+  {
+    v[i][j] = i * 4 + j;
+  }
+}
+```
+This is similar to how a multi-dimensional array is handled in C and HLSL. When a matrix type is `float3x4`, the first indexing, `i`, corresponds to the first value specified in the matrix type `3`. And the second indexing, `j`, corresponds to the second value specified in the matrix type `4`.
```
csyonghe, line 238: the first indexing -> the first dimension index

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -224,23 +224,48 @@ When a slang module uses a pointer, the resulting SPIRV will be using the SpvAdd
 Matrix type translation
 -----------------------
 
-TODO: Quickly summarize what DXC does as described in [HLSL to SPIR-V Feature Mapping Manual](https://github.com/Microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#appendix-a-matrix-representation) and say Slang follows the same behavior.
+As an example, a Slang shader can iterate each element of a `float3x4` matrix as following,
```
csyonghe, line 227: Start with the conclusion:

A m-row-by-n-column matrix in Slang, represented as `floatmxn` or `matrix<T, m, n>`, is represented as `OpTypeMatrix (OpTypeVector(T, n), m)` in SPIRV. Note that in SPIR-V terminology, this type is referred to a m-column-by-n-row matrix.

The swap of row and column terminology may seem to be confusing at first, but this is the most sensical translation that avoids extra operations that can potentially have negative performance impact. For example, consider the following Slang code:

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,333 @@
...
+
+SPIR-V specific functionalities
+===============================
+
+This chapter provides information for SPIR-V specific functionalities and behaviors.
+
+Experimental support for the older versions of SPIR-V
+-----------------------------------------------------
+
+Support for SPIR-V 1.0, 1.1 and 1.2 is still experimental. When targeting the older SPIR-V profiles, Slang may produce SPIR-V that uses the instructions and keywords that were introduced in the later versions of SPIR-V.
```
csyonghe, line 14: We should say that Slang's SPIRV backend is stable when emitting SPIRV 1.3 and later, however, support for SPIR-V 1.0, 1.1 and 1.2 is still experimental.

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,333 @@
...
+
+As an example, the following command-line arguments can control the behavior of `discard` when targeting SPIR-V.
+```
+slangc.exe test.slang -target spirv -profile spirv_1_5 # emits OpKill 
+slangc.exe test.slang -target spirv -profile spirv_1_6 # emits OpDemoteToHelperInvocation 
+slangc.exe test.slang -target spirv -capability SPV_EXT_demote_to_helper_invocation -profile spirv_1_5 # emits OpDemoteToHelperInvocation 
+```
+
+
+Support HLSL features when targeting SPIR-V
```
csyonghe, line 42: Supported HLSL features...

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,333 @@
...
+SPIR-V 1.5 with [SPV_EXT_shader_atomic_float16_add](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/EXT/SPV_EXT_shader_atomic_float16_add.asciidoc) can use atomic operations for 16-bit float type
+
+|        |  32-bit integer | 64-bit integer  |      32-bit float     |  64-bit float    |   16-bit float   |
+|--------|-----------------|-----------------|-----------------------|------------------|------------------|
+| HLSL   |   Yes (SM5.0)   |   Yes (SM6.6)   | Only bit-wise (SM6.6) |       No         |      No          |
+| GLSL   |   Yes (GL4.3)   | Yes (GL4.4+ext) |    Yes (GL4.6+ext)    | Yes (GL4.6+ext)  | Yes (GL4.6+ext)  |
+| SPIR-V |   Yes           |     Yes         |    Yes (SPV1.5+ext)   | Yes (SPV1.5+ext) | Yes (SPV1.5+ext) |
+
+
+ConstantBuffer, (RW)StructuredBuffer (RW)ByteAddressBuffer
```
csyonghe, line 88: (RW/RasterizerOrdered)

csyonghe (approved):
Looks good. Only a few minor comments.

csyonghe (approved):

docs/user-guide/a1-06-spirv-target-specifics.md
```diff
@@ -0,0 +1,336 @@
+---
+layout: user-guide
+permalink: /user-guide/target-specifics
```
csyonghe, line 3: /user-guide/spirv-target-specific

Instead of putting this to appendix, let's create a chapter for "Target-specific Features", and make this nested under that.

csyonghe: In the future we will also be adding a page for metal and webgpu, all under the same chapter.

csyonghe:
We need a better landing page paragraph.

And I think we also need to document how every `SV_` semantic are translated to spirv builtins.
In addition, we added spirv specific system semantics like `SV_PointSize`.

jkwak-work:
> We need a better landing page paragraph.
> 
> And I think we also need to document how every `SV_` semantic are translated to spirv builtins. In addition, we added spirv specific system semantics like `SV_PointSize`.

I will add it.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -26,6 +26,13 @@ Sampler2D explicitBindingSampler : register(t4): register(s3);
 ```
 
 
+`SV_*` semantics
```
csyonghe, line 29: "System Semantics"

csyonghe:
Remember to update toc.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+| SV_GroupID | BuiltIn WorkgroupId |
+| SV_GroupIndex | BuiltIn LocalInvocationIndex |
+| SV_GroupThreadID | BuiltIn LocalInvocationId |
+| SV_InnerCoverage | BuiltIn FullyCoveredEXT |
+| SV_InsideTessFactor | BuiltIn TessLevelInner |
+| SV_InstanceID | BuiltIn ??TODO?? |
+| SV_IntersectionAttributes | ??TODO?? |
+| SV_IsFrontFace | BuiltIn FrontFacing |
+| SV_OutputControlPointID | BuiltIn InvocationId |
+| SV_PointSize | ??TODO?? |
```
csyonghe, line 58: SpvBuiltInPointSize

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+| SV_Position | BuiltIn Position/FragCoord |
+| SV_PrimitiveID | BuiltIn PrimitiveId |
+| SV_RenderTargetArrayIndex | BuiltIn Layer |
+| SV_SampleIndex | BuiltIn SampleId |
+| SV_ShadingRate | BuiltIn PrimitiveShadingRateKHR |
+| SV_StartVertexLocation | Not supported |
+| SV_StartInstanceLocation | Not suported |
+| SV_StencilRef | BuiltIn FragStencilRefEXT |
+| SV_Target<N> | ??TODO?? Location |
```
csyonghe, line 67: Output Location

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+| SV_DispatchThreadID | BuiltIn GlobalInvocationId |
+| SV_DomainLocation | BuiltIn TessCoord |
+| SV_GSInstanceID | BuiltIn InvocationId |
+| SV_GroupID | BuiltIn WorkgroupId |
+| SV_GroupIndex | BuiltIn LocalInvocationIndex |
+| SV_GroupThreadID | BuiltIn LocalInvocationId |
+| SV_InnerCoverage | BuiltIn FullyCoveredEXT |
+| SV_InsideTessFactor | BuiltIn TessLevelInner |
+| SV_InstanceID | BuiltIn ??TODO?? |
+| SV_IntersectionAttributes | ??TODO?? |
```
csyonghe, line 55: Not supported

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+System-Value semantics
+----------------
+
+The system-value semantics are translated to the following SPIR-V code.
+
+TODO: Found a similar page on [DXC document](https://github.com/microsoft/DirectXShaderCompiler/blob/b766b432678cf5f7a93567d253bb5f7fd8a0b2c7/docs/SPIR-V.rst#implicit-location-number-assignment)
+
+| SV semantic name | SPIR-V code |
+|--|--|
+| SV_Barycentrics | BuiltIn BaryCoord..KHR ??TODO?? |
```
csyonghe, line 38: SpvBuiltInBaryCoordKHR

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+------------
+
+Legalization is a process where Slang applies slightly different approach to translate the input Slang shader to the target.
+This process allows Slang shaders to be written in a syntax that SPIR-V may not be able to achieve natively.
+
+Slang allows to use opaque resource types as members of a struct.
+
+Slang allows functions that return any resource types as return type or `out` parameter as long as things are statically resolvable.
+
+Slang allows functions that return arrays.
```
csyonghe, line 332: These functions will be converted to return the array via an `out` parameter in SPIRV.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+Note that the matrix translation explained above is orthogoal to the memory layout of a matrix. The memory layout is related to how CPU places matrix values in the memory and how GPU reads them. It is like how `std140` or `std430` works. DXC by default uses `column_major` memory layout and Slang uses row-major memory layout. For more information about the matrix memory layout, please see [a1-01-matrix-layout](a1-01-matrix-layout.md).
+
+
+Legalization
+------------
+
+Legalization is a process where Slang applies slightly different approach to translate the input Slang shader to the target.
+This process allows Slang shaders to be written in a syntax that SPIR-V may not be able to achieve natively.
+
+Slang allows to use opaque resource types as members of a struct.
```
csyonghe, line 328: These members will be hoisted out of struct types and become global variables.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+
+Slang allows functions that return any resource types as return type or `out` parameter as long as things are statically resolvable.
+
+Slang allows functions that return arrays.
+
+Slang allows putting scalar/vector/matrix/array types directly as element type of a constant buffer or structured buffers.
```
csyonghe, line 334: Such element types will be wrapped in a struct type when emitting to SPIRV.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -0,0 +1,381 @@
...
+ - [GLSL wiki](https://github.com/KhronosGroup/glslang/wiki/HLSL-FAQ#auto-mapped-binding-numbers)
+
+### -fvk-bind-globals <N> <descriptor-set>
+Places the $Globals cbuffer at descriptor set <descriptor-set> and binding <N>.
+It lets you specify the descriptor for the source at a certain register.
+
+For more information, see the following pages:
+ - [DXC description](https://github.com/Microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#hlsl-global-variables-and-vulkan-binding)
+
+### -fvk-use-scalar-layout, -force-glsl-scalar-layout
```
csyonghe, line 199: We now have -fvk-use-dx-layout

csyonghe (dismissed):

PR #4833, opened 2024-08-13
Variadic Generics Part 1: parsing and type checking.

### LINKED ISSUE #4826: Implement variadic generics: syntax parsing and type system.

csyonghe:
Closes #4826.

This PR is the first part of the implementation for variadic generics, focusing on the type system.

Implemented in this change:
- Parsing `expand` and `each` expressions, both for types and for values.
- Type system entities: `ConcreteTypePack`, `EachType`, `ExpandType`, `TypePackSubtypeWitness`, `EachSubtypeWitness` and `ExpandSubtypeWitness`.
- Subtyping rules for `EachType` and `ExpandType`, e.g. `each T` is a subtype of `IBase` if `T:IBase`.
- Type unification rules for  `ConcreteTypePack`, `EachType`, `DeclRefType(GenericTypePackParamDecl)` and `ExpandType`.
- Generic constraint solver for type pack parameters.
- Overload resolution with generalized argument-parameter matching logic to take account for type packs.


source/slang/slang-check-decl.cpp
```diff
@@ -7076,6 +7076,13 @@ namespace Slang
             {
                 args.add(DeclRefType::create(astBuilder, astBuilder->getDirectDeclRef(genericTypeParamDecl)));
             }
+            else if (auto genericTypePackParamDecl = as<GenericTypePackParamDecl>(mm))
+            {
+                auto packType = DeclRefType::create(astBuilder, astBuilder->getDirectDeclRef(genericTypePackParamDecl));
+                //auto eachType = astBuilder->getEachType(packType);
+                //auto expandType = astBuilder->getExpandType(eachType, makeArrayViewSingle(packType));
```
ArielG-NV, line 7083: I don't quite understand what the comments are supposed to refer to?

source/slang/slang-check-decl.cpp
```diff
@@ -7570,6 +7577,27 @@ namespace Slang
...
+        else if (isTypePack(paramDecl->type.type))
+        {
+            // A parameter pack is always `const` if it is not `inout`.
+            bool hasConstModifier = false;
+            for (auto modifier : paramDecl->modifiers)
+            {
+                if (as<OutModifier>(modifier) || as<InOutModifier>(modifier) || as<RefModifier>(modifier) || as<ConstRefModifier>(modifier))
+                {
+                    getSink()->diagnose(modifier, Diagnostics::parameterPackMustBeConst);
+                }
```
ArielG-NV, line 7589: From what I understand, the code below here is saying is that "A parameter pack is always `const`, otherwise we should error":
`if (as<OutModifier>(modifier) || as<InOutModifier>(modifier) || as<RefModifier>(modifier) || as<ConstRefModifier>(modifier))`

This confuses me since the comment above says, "A parameter pack is always `const` if it is not `inout`". Am I misunderstanding the intention of the code here?

ArielG-NV (commented):
The reviewed code look good and fits nicely into the existing type-system, I have 2 minor comments/questions so-far.

Note: I am not done reviewing

source/slang/slang-check-decl.cpp
csyonghe: Sorry, this is left over code that should be deleted.

source/slang/slang-check-decl.cpp
csyonghe: I didn't implement `inout` yet, so let me change the comment to match the implementation.

source/slang/slang-check-overload.cpp
```diff
@@ -2238,7 +2529,7 @@ namespace Slang
         context.originalExpr = genericAppExpr;
         context.funcLoc = baseExpr->loc;
         context.argCount = args.getCount();
-        context.args = args.getBuffer();
+        context.args = &args;
```
ArielG-NV, line 2532: Why are we no longer using `.getBuffer()`?

source/slang/slang-check-overload.cpp
```diff
@@ -42,20 +57,26 @@ namespace Slang
...
+                if (counts.allowed >= 0)
+                    counts.allowed++;
                 if (!valParam->initExpr)
                 {
                     counts.required++;
                 }
             }
+            else if (as<GenericTypePackParamDecl>(m))
+            {
+                counts.allowed = -1;
```
ArielG-NV, line 78: I noticed we more than once use a `-1` as a special parameter count state.
I wonder if we should make this an unscoped enum (or something more distinct) so it would be easier to refactor/read this code and make sense of what the -1 implies in the future.

ArielG-NV (commented):
Note: reviewed the PR

Rest of the code looks good as-well. I have 3 minor comments/suggestions.

jkwak-work (approved):
Looks good to me.

source/slang/slang-ast-type.h
```diff
@@ -678,21 +691,59 @@ class FuncType : public Type
...
 {
     SLANG_AST_CLASS(TupleType)
 
-    // Construct a unary tupletion
-    TupleType(ArrayView<Type*> memberTypes)
+    Index getMemberCount() const;
+    Type* getMember(Index i) const;
+    Type* getTypePack() const;
+
+};
```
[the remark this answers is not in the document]
csyonghe: Yes, tuple types were in the codebase but never hooked up with any actual type in the stdlib so the user were never able to create them. Instead of deleting it, I am slightly modifying the existing implementation so that it is now allowed in user code as `Tuple<T0,T1,...Tn>`.

Issue #4824, opened 2024-08-12
Issues with `-zero-initialize` when compiling to GLSL


AlexDunn:
I've been testing the new -zero-initialize feature in Slang (2024.9.1) and I have a couple findings that it believe are erroneous.

Here are the slang.c flags used in my experiments: 
`-entry main -target glsl -matrix-layout-column-major -line-directive-mode none -Wno-30081 -zero-initialize`

**Group shared memory is also being zero initialized, and unsafely.**
Consider the following compute shader in slang:
```
groupshared uint samplePixel[16 * 8];
[shader("compute")]
[numthreads(16, 8, 1)]
void main(uint2 thread_id : SV_DispatchThreadID, uint2 localIndex : SV_GroupThreadID)
{
  samplePixel[localIndex.y * 16 + localIndex.x] = 0xffffffff;
  GroupMemoryBarrierWithGroupSync();
  ...
}
```

This is being translated to the following GLSL:
```
shared uint  samplePixel_0[128];
layout(local_size_x = 16, local_size_y = 8, local_size_z = 1) in;
void main()
{
    uint  _S43[128];
    _S12(_S43);
    samplePixel_0 = _S43;
    uvec2 _S44 = gl_GlobalInvocationID.xy;
    uvec2 _S45 = gl_LocalInvocationID.xy;
    samplePixel_0[_S45.y * 16U + _S45.x] = 4294967295U;
    controlBarrier(gl_ScopeWorkgroup, gl_ScopeWorkgroup, gl_StorageSemanticsShared, gl_SemanticsAcquireRelease);
    ...
}
```

It's generally undesirable to initialize shared memory like this, the function _S12 is writing 0 to every element in the shared memory array.  This happens on each thread and so it's a data race.  I would prefer my shared memory to stay uninitialized.  Or worst case, would prefer the inclusion of a memory barrier to make those writes somewhat safe...

**When compiling to GLSL, the zero initialize feature seems to incorrectly initialize RayQuery**
Consider the following line in a slang shader:
```
RayQuery<0> foo;
```

With -zero-initialize this is being translated to:
```
rayQueryEXT _S313;
rayQueryEXT rayQuery_0 = _S313;
...
rayQueryInitializeEXT((rayQuery_0), (TLAS_0), (64U), (maxSortOrder_0), (rayDesc_0.Origin_0), (rayDesc_0.TMin_0), (rayDesc_0.Direction_0), (rayDesc_0.TMax_0));
```

[...]

PR #4838, opened 2024-08-13
Do not zero-initialize groupshared and rayquery variables

ArielG-NV:
Fixes: #4824

`-zero-initialize` option and interface will explicitly not:
1. not set any groupshared values to 'default value'.
2. not set any RayQuery object to a 'default state' (currently invalid code generation).

Additionally:
1. groupshared values error if provided an initializer.

This is a breaking change because:
1. groupshared values error if they are assigned an initializer now


csyonghe:
Let's make it an error if a groupshared variable is being initialized because we can't generate efficient code for them right now.

source/slang/slang-check-decl.cpp
```diff
@@ -1922,10 +1922,59 @@ namespace Slang
...
+                    structDecl = baseTypeStruct;
+                }
+                else if (auto genericDecl = as<GenericDecl>(baseGenericRefType))
+                {
+                    if(auto innerTypeStruct = as<StructDecl>(genericDecl->inner))
+                        structDecl = innerTypeStruct;
+                }
+            }
+        }
+        if (structDecl)
```
csyonghe, line 1961: I think it is safe to say all NonCooyable type are not default initializable.

csyonghe (approved):

PR #4845, opened 2024-08-13
Make precompileForTargets work with Slang API

### LINKED ISSUE #4790: Write a unit test that uses EmbedDXIL option through Slang API
-embed-dxil feature is only supported in slangc and the comparable EmbedDXIL option is not supported in Slang API.

Write a unit test that uses EmbedDXIL and slang api and make necessary fixes in precompileForTargets().

Use slang\tools\gfx-unit-test\precompiled-module.cpp as basis for the test.

cheneym2:
precompileForTargets, renamed to precompileForTarget, does not need an EndToEndCompileRequest and some objects created from it are not necessary either.

Take only a target enum and a diagnostic blob as input and handle everything else internally, such as creating the TargetReq with chosen profile.

Fixes #4790


source/slang-record-replay/record/slang-module.cpp
```diff
@@ -213,6 +213,16 @@ namespace SlangRecord
         return res;
     }
 
+    SLANG_NO_THROW SlangResult ModuleRecorder::precompileForTarget(
+        SlangCompileTarget target,
+        ISlangBlob** outDiagnostics)
+    {
+        // TODO: We should record this call
```
csyonghe, line 220: There needs to be an issue associated with the TODO.

cheneym2: Added

csyonghe (approved):

PR #4848, opened 2024-08-14
Design proposal for Tuple types.

### LINKED ISSUE #4830: Design full tuple type feature

csyonghe:
Closes #4830.

### DIFF
--- a/docs/proposals/008-tuples.md
+++ b/docs/proposals/008-tuples.md
@@ -1,4 +1,5 @@
 SP #008 - Tuples
+==============
 
 Now that we have variadic generics in the language following [SP #007], we should now be able to support `Tuple` type as a core language feature.
 `Tuple` types are useful in many places to reduce boilerplate in user code, such as in function return types to eliminate the need of defining
--- a/docs/proposals/008-tuples.md
+++ b/docs/proposals/008-tuples.md
@@ -1,0 +1,136 @@
+SP #008 - Tuples
+
+Now that we have variadic generics in the language following [SP #007], we should now be able to support `Tuple` type as a core language feature.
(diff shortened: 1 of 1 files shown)


docs/proposals/008-tuples.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+Proposed Approach
+-----------------
+
+With variadic generics support, we can now easily define a Tuple type in stdlib as:
+```
+__generic<each T>
+__magic_type(TupleType)
+struct Tuple
```
jkwak-work, line 33: How would the members be declared?
Should it be something like a following?
```
__generic<T0, each T>
struct Tuple : Tuple<T>
{
     T0 e0;
};
```

docs/proposals/008-tuples.md
```diff
@@ -0,0 +1,137 @@
...
+__intrinsic_op($(kIROp_MakeTuple))
+Tuple<expand each T> makeTuple(expand each T values);
+```
+
+With generic argument inferencing, this will enable user to write:
+```
+makeTuple(1, 2.0f) // returns Tuple<int, float>(1, 2.0f)
+```
+
+### Accessing Tuple Elements
```
jkwak-work, line 55: How can we get the size of the Tuple?
Maybe something like t.getCount()?

jkwak-work (approved):
Looks good to me.

docs/proposals/008-tuples.md
csyonghe: We can't define the members inside the stdlib, just like we can't define `x`, `y` etc in a vector type, it has to be implemented as a special syntax in the parser.

docs/proposals/008-tuples.md
csyonghe: We should be able to do `countof(tupleValue)` just like `countof(T)`.

PR #4849, opened 2024-08-15
Variadic Generics Part 2: IR lowering and specialization.

### LINKED ISSUE #4827: Implement variadic generics: IR lowering and specialization.

### LINKED ISSUE #4829: Implement variadic generics: add basic tuple type support in stdlib.

csyonghe:
Closes #4827.
Closes #4829.

This PR implements the IR lowering and specialization part of the design proposal.


source/slang/core.meta.slang
```diff
@@ -870,6 +870,14 @@ bool operator!=(__none_t noneVal, Optional<T> val)
     return val.hasValue;
 }
 
+__generic<each T>
+__magic_type(TupleType)
+struct Tuple
+{
+    __intrinsic_op($(0))
+    __init(expand each T);
+}
```
jkwak-work, line 879: This is a question.
This implementation is a kind of opaque type whose implementation is not visible to the end users.
Isn't the variadic argument feature good enough to make this not opaque?
I asked the same question on the other review but I am still not sure if there is a reason why the implementation is less on frontend side and more on the c++ side.
```
struct Tuple<T0,T> : Tuple<T>
{
 T0 v;
 ...
}
```
My guess is that currently the implementation for the swizzle and constructor needs to be implemented on c++ side due to a certain limitations Slang still have.
Does this mean the users cannot declare their own version of custom Tuple-like struct because we are still missing certain language feature to support cases like Tuple?

source/slang/slang-ir-insts.h
```diff
@@ -3814,6 +3851,13 @@ struct IRBuilder
 
     IRInst* emitMakeTuple(IRType* type, List<IRInst*> const& args)
     {
+        if (args.getCount() == 1)
+        {
+            if (args[0]->getOp() == kIROp_Expand)
+            {
+                return args[0];
+            }
+        }
```
jkwak-work, line 3860: Should we also handle the zero number of argument case?

source/slang/slang-lower-to-ir.cpp
```diff
@@ -1653,6 +1656,86 @@ struct ValLoweringVisitor : ValVisitor<ValLoweringVisitor, LoweredValInfo, Lower
         return LoweredValInfo::simple(resultVal);
     }
 
+    LoweredValInfo visitConcreteTypePack(ConcreteTypePack* typePack)
+    {
+        ShortList<IRType*> types;
```
jkwak-work, line 1661: I am not sure what the max value can be from getTypeCount.
Maybe we should reserve space fore `types` before start adding elements to it.

jkwak-work (approved):
Looks good to me

source/slang/core.meta.slang
csyonghe: No matter what we do, there needs to have a built in mechanism to represent storage for a type pack, and tuple types is that built in mechanism so we have no way to define it in user space. If the user wish to define a variable that holds values for a list of types, they will use tuple types for that. Tuple type itself is the language mechanism to define storage for type packs so it has to be implemented as builtin functionality. 

And you are right that the user cannot define another thing like tuple type where fields are named differently.

source/slang/slang-ir-insts.h
csyonghe: Zero args here should automatically translate to a tuple type with 0 elements, and will be translated to void during lowering.

source/slang/slang-lower-to-ir.cpp
csyonghe: We can, but this kind of things doesn't really matter that much. Shortlist is already by default reserved to be 16 in length and very few tuples in practice would exceed that anyways. If they do the next allocation would be 32.

tests/language-feature/generics/variadic-0.slang
```diff
@@ -0,0 +1,47 @@
...
+}
+
+[numthreads(1,1,1)]
+void computeMain()
+{
+    // CHECK: 1.0
+    // CHECK: 2.0
+    // CHECK: 3.0
+    // CHECK: 4.0
+    write();
```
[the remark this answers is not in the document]
kaizhangNV: I think it's to verify that the `value` is null in the write function. So should we verify something to indicate that this is handled correctly? e.g. verify `offset` doesn't change?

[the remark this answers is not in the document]
csyonghe: This is to make sure specializing a type pack with zero types works. I guess O can make this empty write after the non empty write to make sure it is not doing anything weird.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2208,6 +2232,239 @@ struct SpecializationContext
...
+            clonePatternVal(*subEnv, builder, childInst->getOperand(i), index);
+        }
+        auto newInst = cloneInst(subEnv, builder, childInst);
+        newInst = builder->replaceOperand(&newInst->typeUse, type);
+        subEnv->mapOldValToNew[childInst] = newInst;
+        IRBuilder subBuilder(*builder);
+        subBuilder.setInsertInto(newInst);
+        for (auto child : childInst->getChildren())
+        {
+            specializeExpandChildInst(*subEnv, &subBuilder, child, index);
```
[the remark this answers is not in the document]
csyonghe: It is not that easy given that we need to handle the an inst differently depending on how the operands are lowered, so the callback is actually useful here. Converting to worklist would be possible but the code can be very messy.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2208,6 +2232,239 @@ struct SpecializationContext
...
+    {
+        if (expandInst->getCaptureCount() == 0)
+            return false;
+
+        bool anyAbstractPack = false;
+        for (UInt i = 0; i < expandInst->getCaptureCount(); i++)
+        {
+            if (!as<IRTupleType>(expandInst->getCapture(i)))
+            {
+                anyAbstractPack = true;
```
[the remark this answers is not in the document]
csyonghe: Good point.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2208,6 +2232,239 @@ struct SpecializationContext
         return false;
     }
 
+    IRInst* specializeExpandChildInst(IRCloneEnv& cloneEnv, IRBuilder* builder, IRInst* childInst, UInt index)
+    {
+        IRCloneEnv freshEnv;
+        IRCloneEnv* subEnv = &cloneEnv;
+        switch (childInst->getOp())
```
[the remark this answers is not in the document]
csyonghe: Usually code like this is to make future extensions easy. If we ever need the same logic for a different op code later, starting with a switch has the benefit that future changes can just be a one liner adding another case. An if statement on opcode is harder to extend later and could result in a lot of nested || as time goes.

PR #4854, opened 2024-08-15
Member-wise constructor generation with visibility control

### LINKED ISSUE #3406: Struct initializer-list with visibility control
Now that we have visibility control, we need to cleanup the initialization part as well:
```
struct S
{
     private int m;
     public int n;
}

void test()
{
     S s = {1,2}; // Initializing private member here seems wrong.
}
```

We should be synthesizing constructors for structs and have initialization list syntax calling into the constructors instead.
This way, our constructor synthesis will fail for `S` because it can't default init `m`. And then that will surface to the user as an error in `test` because it is trying to call a constructor that does not exist.

### LINKED ISSUE #4461: Initializers example in the Slang User Guide does not compile.
I've been hunting down some unexpected behavior, and in the process found that the example from https://shader-slang.com/slang/user-guide/convenience-features.html#initializers won't even compile for me:

The code is:
```
RWStructuredBuffer<uint> output;
struct MyType
{
    int myVal;
    __init(int a, int b)
    {
        myVal = a + b;
    }
}

[numthreads(16, 1, 1)]
void testInitializers(uint3 threadID: SV_DispatchThreadID)
{
    if (threadID.x != 0)
        return;

   MyType instance = {1, 2};
   output[0] = instance.myVal;
}
```

the error is:
```
D:\Projects\Falcor\Source\Tools\FalcorTest\Tests\Slang\InitializerTestsSimple.cs.slang(100): error 30500: too many initializers (expected 1, got 2)
        MyType instance = {1, 2};
```

The behavior seems to be that, contrary to the text in the guide `You may also use C++ style initializer list to invoke a constructor` it does not invoke the constructor, but instead does the C-style member initialization. So instead of calling `__init(int a, int b)` it is trying to initialize the first two members of the struct to values `1` and `2` and since the struct has only one member, it errors out.

This also explains why `Test test;` has a different behavior from `Test test = {}` and `Test test = Test(1)` has a different behavior from `Test test = {1}`. I don't know if the intent is to call `__init()` and it is the codegen that needs a fix, or if the intent is to do this C style struct initialization, in which case the guide/docs needs a fix.

In the latter case also please specify what happens when the initializer list is shorter than the number of member variables. I know that if the variable is in the form `int foo = 3;` then it gets initialized to `3` with the `Test test ={};`, but I don't know if it gets default initialized to any specific value if it is just `int foo;`.

ArielG-NV:
Fixes #3406
Fixes #4461

PR with documentation: #4988

Changes (split up into major files):
`source/slang/diff.meta.slang`
* Addition of `TorchTensorType`. This was added since Slang requires all `TorchTensor` objects to have a CUDAHost attribute. When auto-generating constructors Slang needs to be aware of this type to avoid generating invalid code for CUDAHost variable using functions (automatically add `CUDAHost` if `TorchTensor` is found).

`source/slang/hlsl.meta.slang`
* (19542) A custom `__init()` was added to `TextureFootprint` to more correctly do what the struct requires (this also avoided at the time a weird compile failure that only the stdlib could cause).
* (19965) Slang cannot propagate `literal generics` (`Shape.dimensions`) properly in our use case. We work around this by generating a simple copy of our `literal generic` and letting Slang optimize this value out to directly use `Shape.dimensions`.

`source/slang/slang-ast-decl.h`
* This PR adds constructors to target `private`, `internal`, and `public` groups of members. If a Ctor is used for `public` and `private` and `internal` scopes we need to store this information so that when we look for `InheritanceDecl` ctor's, we know which "member-wise ctor" should be used given a visibility. `ConstructorTags` allows us to store this information.

`source/slang/slang-check-conversion.cpp`
* Initializer lists call into constructors (and no longer uses related hacks unless `UseLegacyLogic` is `true`).
* initialization-list logic for struct objects has been reordered and reformatted due to previous logical incompatibilities.
* Flattened initializers for structs (c-style initializers) are allowed if and only if a struct is a C-Style-Struct (see user-guide for more details, `docs\user-guide\02-conventional-features.md`)

`source/slang/slang-check-decl.cpp`
* we auto-generate member-wise constructors for varying visibilities of members (public, public-internal, public-internal-private).
* Implements visibility and member-wise constructor rules described in #3406.
* Reorders and reformats how we auto-generate constructors. We needed to move generation of constructor "declarations" into `SemanticsDeclConformancesVisitor`.
* If a `TorchTensor` typed variable is found we auto-add a CUDA-host attribute (host only function)
* If we check an invoke and find a `TreatAsDifferentiableExpr`, we do not automatically error with "invalid `no_diff` use-case", this is incorrect behavior. We only error if `TreatAsDifferentiableExpr` is of flavor `no_diff`.

Other changes:
* Addition of `$ZeroInit` to allow us 2 important functionalities: 1. Allows us to use `{}` inside a `__init()` 2. it allows us to zero initialize with `{}` for when a user does not provide a `__init()` (which is expected behavior).
* Due to #4874 we cannot assign functions to global's. As per discussion we will remove this failing functionality inside `tests\compute\type-legalize-global-with-init.slang`.
* We deprecated legacy init-list syntax. This syntax was not removed, only deprecated


ArielG-NV:
Note: to repro the `[  FAILED  ] DiffMaterialTests.cpp:DiffPBRTDiffuse (D3D12)` Falcor failure the following can be inputted into `slangc`

```
-target hlsl -entry testDiffPBRTDiffuse -stage compute ${PATH_TO_FALCOR}\Falcor\Source\Tools\FalcorTest\Tests\DiffRendering\Material\DiffMaterialTests.cs.slang -I ${PATH_TO_FALCOR}\Falcor\Source\Tools\FalcorTest\Tests\ -I C:\code\Falcor\Source\Falcor\Utils\ -I ${PATH_TO_FALCOR}\Falcor\Source\Falcor\ -DSCENE_SDF_GRID_IMPLEMENTATION_NDSDF=1 -DSCENE_SDF_GRID_IMPLEMENTATION_SVS=2 -DSCENE_SDF_GRID_IMPLEMENTATION_SBS=3 -DSCENE_SDF_GRID_IMPLEMENTATION_SVO=4 -DSCENE_SDF_NO_INTERSECTION_METHOD=0 -DSCENE_SDF_NO_VOXEL_SOLVER=1 -DSCENE_SDF_VOXEL_SPHERE_TRACING=2 -DSCENE_SDF_NO_GRADIENT_EVALUATION_METHOD=0 -DSCENE_SDF_GRADIENT_NUMERIC_DISCONTINUOUS=1 -DSCENE_SDF_GRADIENT_NUMERIC_CONTINUOUS=2 -DSCENE_SDF_GRID_COUNT=1 -DSCENE_SDF_GRID_MAX_LOD_COUNT=1 -DSCENE_SDF_GRID_IMPLEMENTATION=0 -DSCENE_SDF_VOXEL_INTERSECTION_METHOD=0 -DSCENE_SDF_GRADIENT_EVALUATION_METHOD=0 -DSCENE_SDF_SOLVER_MAX_ITERATION_COUNT=0 -DSCENE_SDF_OPTIMIZE_VISIBILITY_RAYS=0 -DSCENE_GEOMETRY_TYPES=0 -DMATERIAL_SYSTEM_TEXTURE_DESC_COUNT=1 -DMATERIAL_SYSTEM_SAMPLER_DESC_COUNT=1 -DMATERIAL_SYSTEM_BUFFER_DESC_COUNT=1 -DMATERIAL_SYSTEM_USE_LIGHT_PROFILE=0 -DMATERIAL_SYSTEM_TEXTURE_3D_DESC_COUNT=1 -DSCENE_GRID_COUNT=1 -DFALCOR_MATERIAL_INSTANCE_SIZE=1 -DFALCOR_NVAPI_AVAILABLE=0
```

Note: `import Rendering.Materials.PBRT.PBRTDiffuseMaterial;` must be added to the top of `Source\Tools\FalcorTest\Tests\DiffRendering\Material\DiffMaterialTests.cs.slang` to provide conformances to interfaces.

ArielG-NV:
Note: Falcor is using a c-style-init-list with a non c-style-struct (`TriangleHit`). This is causing a PR failiure.

ArielG-NV:
Note: any commits after this comment will include "Legacy Slang init-list support" to avoid breaking legacy Slang code

source/slang/slang-check-conversion.cpp
```diff
@@ -687,14 +746,32 @@ namespace Slang
            !canCoerce(toType, fromInitializerListExpr->type, nullptr))
             return _failedCoercion(toType, outToExpr, fromInitializerListExpr);
 
-        if(!_readAggregateValueFromInitializerList(toType, outToExpr, fromInitializerListExpr, argIndex))
+        if(!_readAggregateValueFromInitializerList<false>(toType, outToExpr, fromInitializerListExpr, argIndex))
```
csyonghe, line 749: should we aalso try legacy logic here too?

ArielG-NV: we run modern logic here.

below we run legacy logic, we run legacy logic if modern logic fails

csyonghe: but isn't this saying if modern logic failed, return false?

csyonghe: it seems what the code is doing is that if modern logic succeeds but there are still remaining args left, we used to throw an error, now we try legacy.

ArielG-NV: I don't know if we need to handle the other case (return false) for when toType is a `StructType`?

csyonghe: I am confused. when `_readAggregateValueFromInitializerList` returns false, it means the modern approach failed.
Then shoudln't we try legacy logic now, instead of returning `false`?

csyonghe: `_readAggregateValueFromInitializerList<true>` is only invoked when `_readAggregateValueFromInitializerList<false>` returns `true` and `argIndex != argCount`? Is this correct?

source/slang/hlsl.meta.slang
```diff
@@ -19957,8 +19962,13 @@ ${
...
         case hlsl:
             uint isSingleLod = 0;
+            // Note: we cannot pass `Shape.dimensions`
+            // directly as a parameter into a Call, otherwise
+            // AST lookup fails (bug?), creating an intermediary works
+            // so we do this workaround.
+            int shapeDim = Shape.dimensions;
             Footprint footprint =  {__queryFootprint$(CoarseOrFine)NVAPI(
-                Shape.dimensions,
+                shapeDim,
```
csyonghe, line 19971: It seems like this may be due to the lookup scope of the expr isn't correct hooked up when you create the ctor call expr?

source/slang/slang-check-conversion.cpp
```diff
@@ -92,13 +101,24 @@ namespace Slang
...
+        // (which we don't have in our init-list argument list, this will error).
+        // 
+        // This logic should be completly cleaned up at some point by only using constructors for `{...}` and attempting to use 
+        // constructors before fallback coerce logic is tried.
+        if (toType->equals(fromExpr->type))
+            return true;
+
+        // `vector<T,N>` and `vector<T,U>` may not be equal types, but, we have no choice in this senario but to assume coerce logic
+        // can resolve such situations, otherwise we need to fail
+        if (isSameSimpleExprType(toType, fromExpr->type))
```
csyonghe, line 115: how can this not implied by line 110?

source/slang/slang-check-conversion.cpp
```diff
@@ -423,53 +464,211 @@ namespace Slang
...
+                        auto ctorParamCount = ctor->getParameters().getCount();
+                        if (ctorParamCount == 0)
+                            continue;
+
+                        ioArgIndexCandidate = ioArgIndexMirror;
+                        ioArgIndex = ctorParamCount;
+
+                        // if allowCStyleInitList, process any ctor which comes next using the most members possible. ioArgIndex may not be 0.
+                        // if !allowCStyleInitList, process any ctor that exactly matched our argument count. ioArgIndex must start at 0.
+                        if (!allowCStyleInitList && ctorParamCount != Index(argCount))
```
csyonghe, line 574: What if __init takes a variadic parameter pack?

source/slang/slang-constructor-utility.cpp
```diff
@@ -0,0 +1,264 @@
...
+                    return;
+                *defaultCtorOut = ctor;
+            };
+        if (ctorLookupResult.items.getCount() == 0)
+        {
+            lookupResultHandle(ctorLookupResult.item);
+            return ctorList;
+        }
+
+        for (auto m : ctorLookupResult.items)
```
csyonghe, line 66: you can just do ` for (auto m : ctorLookupResult)`

source/slang/slang-constructor-utility.cpp
```diff
@@ -0,0 +1,264 @@
...
+        for (auto funcDecl : structDecl->getMembersOfType<FuncDecl>())
+        {
+            if (!funcDecl->findModifier<ZeroInitModifier>())
+                continue;
+            return funcDecl;
+        }
+        return nullptr;
+    }
+
+    Expr* _constructZeroInitListFuncMakeDefaultCtor(SemanticsVisitor* visitor, StructDecl* structDecl, Type* structDeclType, ConstructorDecl* defaultCtor)
```
csyonghe, line 154: this is a confusing name. I think the function should be named `constructDefaultCtorInvocationExpr`

source/slang/slang-constructor-utility.cpp
```diff
@@ -0,0 +1,264 @@
...
+                    }
+                    callingScope = callingScope->parent;
+                }
+            }
+            if(canCreateCtor)
+                return _constructZeroInitListFuncMakeDefaultCtor(visitor, structDecl, structDeclType, defaultCtor);
+        }
+
+        // 2.
+        if (auto zeroInitListFunc = findZeroInitListFunc(structDecl))
```
csyonghe, line 199: al `zeroInitListFunc` should be named `defaultInitFunc`.

source/slang/slang-constructor-utility.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return true;
+    }
+    List<ConstructorDecl*> _getCtorList(ASTBuilder* m_astBuilder, SemanticsVisitor* visitor, StructDecl* structDecl, ConstructorDecl** defaultCtorOut)
+    {
+        List<ConstructorDecl*> ctorList;
+
+        auto ctorLookupResult = lookUpMember(
+            m_astBuilder,
+            visitor,
+            visitor->getName("$init"),
```
csyonghe, line 39: We should cache this name on SharedASTBuilder, like we do for the name `This`.

source/slang/slang-check-conversion.cpp
```diff
@@ -423,53 +464,211 @@ namespace Slang
...
+                    }
+
+                    // 3. We have a `{arg1, arg2...}`, try to coerce
+                    //
+                    // For this situation we have more steps:
+                    // a. Find constructor candidate for our particular init-list
+                    // b. Collect and coerce init-list arguments into constructor parameters
+                    // c. Create InvokeExpr for our valid ConstructorDecl 
+                    List<ConstructorDecl*> ctorList = _getCtorList(this->getASTBuilder(), this, toStructDecl, nullptr);
+                    bool allowCStyleInitList = checkIfCStyleStruct(this, toStructDecl);
```
csyonghe, line 553: since `ctorList` is already known here, we should have a variant of `checkIfCStyleStreuct` using the ctorList instead of reconstructing them.

source/slang/slang-check-impl.h
```diff
@@ -2980,4 +2993,36 @@ namespace Slang
...
+    enum class ConstructZeroInitListOptions : UInt
+    {
+        None = 0 << 0,
+        PreferZeroInitFunc = 1 << 0,
+        CheckToAvoidRecursion = 1 << 2,
+    };
+    Expr* constructZeroInitListFunc(SemanticsVisitor* visitor, StructDecl* structDecl, Type* structDeclType, ConstructZeroInitListOptions options);
+    FuncDecl* findZeroInitListFunc(StructDecl* structDecl);
+
+    bool checkIfCStyleStruct(SemanticsVisitor* visitor, StructDecl* decl);
```
csyonghe, line 3022: should be named `isCStyleStructDecl`.

source/slang/slang-check-overload.cpp
```diff
@@ -2241,6 +2246,33 @@ namespace Slang
...
     }
 
+
+    /// We allow a special case for when `funcExpr` is expected to be a Default initializer
+    /// but none exist. Note, we cannot just create a default initializer for every variable
+    /// since then we are introducing initialization to every variable through an indirect
+    /// init returning data. Instead we will call `$ZeroInit` through this logic below.
+    Expr* _tryToSpecialCaseOverloadDefaultConstructWithoutInit(SemanticsVisitor* visitor, SemanticsVisitor::OverloadResolveContext& context, Expr* expr, OverloadCandidate* bestCandidate)
+    {
+        if (context.argCount == 0)
```
csyonghe, line 2256: should we make sure the canddiate we are calling is a ctor decl?

what are we trying to do here? If we have a ctor candidate that takes more than 0 parameters but we are trying to call it with 0 args, we form an initializerList coercion instead? Why do we need to do this?

If we have `struct S { __init (x) }` and called with `S s = {};`, it should be an error?

source/slang/slang-constructor-utility.cpp
```diff
@@ -0,0 +1,264 @@
...
+            return ctorList;
+
+        auto lookupResultHandle = [&](LookupResultItem& item)
+            {
+                auto ctor = as<ConstructorDecl>(item.declRef.getDecl());
+                if (!ctor)
+                    return;
+                ctorList.add(ctor);
+                if (ctor->members.getCount() != 0
+                    && !allParamHaveInitExpr(ctor)
```
csyonghe, line 55: this semes to be inconsistent with `getDefaultCtor`, because that function won't categorize `__init(int x =0)` as default ctor.

source/slang/slang-check-conversion.cpp
ArielG-NV: I understand your point, will make a new commit with the addition

ArielG-NV:
I think @kaizhangNV will have to take over and finish review of this PR (from what I understand), I apologize for the inconvenience this may cause.

source/slang/hlsl.meta.slang
ArielG-NV: This is very likely

source/slang/slang-check-conversion.cpp
ArielG-NV: if we have 2 aggregate types we return true

source/slang/slang-check-decl.cpp
```diff
@@ -7833,6 +8045,73 @@ namespace Slang
         return as<SeqStmt>(stmt->body);
     }
 
+    static bool _doesCtorExpectInitializerListUsage(ConstructorDecl* ctor)
```
csyonghe, line 8048: This excludes user defined ctors, accorading to the name of this function, it means user defined ctors does not expect initializer list usage? That seems wrong. What exactly are we testing for here?

source/slang/slang-check-decl.cpp
```diff
@@ -7833,6 +8045,73 @@ namespace Slang
...
+    {
+        return ctor->containsOption(ConstructorTags::Synthesized)
+            && ctor->getParameters().getCount() != 0
+            && !allParamHaveInitExpr(ctor);
+    }
+
+    template<typename T>
+    bool _containsTargetType(ASTBuilder* m_astBuilder, NodeBase* target, HashSet<NodeBase*>& visitedNodes)
```
csyonghe, line 8056: this is very heavy weight recursion, why do we need this?

source/slang/slang-check-decl.cpp
```diff
@@ -7981,19 +8326,325 @@ namespace Slang
...
+        DeclAndCtorInfo* baseStructInfo = nullptr;
+        for(auto& i : inheritanceInfoList)
+        {
+            ensureDecl(i.m_inheritanceDecl, DeclCheckState::VarInitExprAreChecked);
+            if(as<StructDecl>(i.m_inheritanceBaseDecl))
+                baseStructInfo = &i;
+        }
+
+        // pre-calculate any requirements of a CudaHostAttribute
+        HashSet<VarDeclBase*> requiresCudaHostModifier;
```
csyonghe, line 8346: It doesn't seem to worth this much complexity for such a corner case. We need to get rid of all the TorchTensorType` logic here, and just insert the cuda host decoration in a separate legalization pass when compiling to pytorch.

source/slang/slang-check-overload.cpp
ArielG-NV: This was a workaround we kept from the previous implementation (but changed to be less of a hack). The main idea is that some structs may be treated as if they have an __init despite not ever declaring an __init

The goal is that we will zeroInit these objects

source/slang/slang-check-decl.cpp
```diff
@@ -7981,19 +8326,325 @@ namespace Slang
...
+                stmt->expression = assign;
+                stmt->loc = assign->loc;
+
+                seqStmtChild->stmts.add(stmt);
+            }
+            seqStmt->stmts.insert(ctorInfo.m_insertOffset++, seqStmtChild);
         }
 
-        if (structDeclInfo.defaultCtor)
+        // Compiler generated ctor may be destroyed
```
csyonghe, line 8519: Why do we need to remove a ctor? Can we just not generate it in the first place?

source/slang/slang-check-decl.cpp
```diff
@@ -7981,19 +8326,325 @@ namespace Slang
...
             {
-                structDecl->members.remove(structDeclInfo.defaultCtor);
+                structDecl->members.remove(structDeclInfo.m_defaultCtor);
                 structDecl->invalidateMemberDictionary();
                 structDecl->buildMemberDictionary();
             }
         }
+
+        // Fill in our $ZeroInit
+        if (auto zeroInitListFunc = findZeroInitListFunc(structDecl))
```
csyonghe, line 8540: I think we shouldn't need this at all because our IR already knows how to zerzo init a thing.
We can simply do:
```
this = DefaultConstruct(This);
```
at the begining of the ctor, where `DefaultConstruct(T)` lowers to `kIROp_DefaultConstruct(T)`

csyonghe:
@ArielG-NV there is no expectation that you will address any of these.
However if you can put comments on what you think are / are not possible regarding my thoughts, that will be very helpful for @kaizhangNV.

ArielG-NV:
I will leave a 👍 on tasks that should be addressed

source/slang/slang-check-conversion.cpp
ArielG-NV: This requires a test.

I believe coerce logic will function correctly with type packs (meaning type packs will work as intended), but I never thought of making a test.

source/slang/slang-check-decl.cpp
ArielG-NV: This function should be renamed to "_isCtorASynthisizedMemberwiseCtor".

The goal is to detect which functions are member-wise ctor's and need their body filled.

csyonghe: We already have a `ToBeSynthesized` modifier that we can use for this purpose.

source/slang/slang-check-decl.cpp
ArielG-NV: We need this or a cached version of some sort (I agree this is quite an expensive function) to detect if we need our ctor to be tagged with "CudaHost" due to CudaHost only type being found.

csyonghe:
@ArielG-NV do you have a non-NV github account that we can reach you?

source/slang/slang-check-decl.cpp
ArielG-NV: https://github.com/shader-slang/slang/pull/4854/files/c0e59e84f30557389d8dd895198c068c219fb6f7#r1747817868

A legalization pass for 1 target is a better idea than this expensive recursive routine.

ArielG-NV:
My non-NV GitHub would be @16-Bit-Dog

source/slang/slang-check-decl.cpp
ArielG-NV: I think this ctor-deletion routine can just be removed completely, you are correct.

We should no longer be creating __init functions which need to be deleted in our bodyVisitor. 

This must hold true since otherwise we would have a mismatch of conformances from the conformance AST-Pass and actual conformances (since deleting a ctor breaks conformances).

source/slang/slang-check-decl.cpp
ArielG-NV: We could add a flag to DefaultConstructExpr to remove the need for $ZeroInit.

To elaborate, DefaultConstructExpr does not handle initExpr on a struct member (from what I remember).

We need a flag that we can set to tell Slang when lowering a DefaultConstructExpr that it should take into account initExpr of varDecl's.

source/slang/slang-constructor-utility.cpp
kaizhangNV: Addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-dad759cd0132d623bee9253550a0bce736d087bdc3157dd1e11e414cb5f6bc48R70

source/slang/slang-constructor-utility.cpp
kaizhangNV: addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-dad759cd0132d623bee9253550a0bce736d087bdc3157dd1e11e414cb5f6bc48R158

source/slang/slang-constructor-utility.cpp
kaizhangNV: addressed in: https://github.com/shader-slang/slang/pull/5107/commits/d966b3fc88a24a4f45780b651bb918a2c216889c

source/slang/slang-constructor-utility.cpp
kaizhangNV: addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-dad759cd0132d623bee9253550a0bce736d087bdc3157dd1e11e414cb5f6bc48R43

source/slang/slang-check-conversion.cpp
kaizhangNV: addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-dad759cd0132d623bee9253550a0bce736d087bdc3157dd1e11e414cb5f6bc48R226

source/slang/slang-check-impl.h
kaizhangNV: addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-650283cec951a78df98dcededc742d1a4c434f3f2179ac2ab6dcbd29de9b5d4bR553

source/slang/slang-constructor-utility.cpp
kaizhangNV: addressed here:
https://github.com/shader-slang/slang/pull/5107/commits/54014949789e700a589f5eea098443f3e2bc337c#diff-dad759cd0132d623bee9253550a0bce736d087bdc3157dd1e11e414cb5f6bc48R27

source/slang/slang-check-decl.cpp
kaizhangNV: Addressed in this change:
https://github.com/shader-slang/slang/pull/5107/commits/5771007f0f14dad1d69c1d499758d10fa24bc087

I found out that this function is not only check whether a function is synthesized, but also this is a member-wise ctor, so only checking `ToBeSynthesized` could not be enough.

source/slang/slang-check-decl.cpp
kaizhangNV: After reading the code carefully, I found that we cannot avoid removing the ctor based on current change.

This change create a empty default ctor during `SemanticsDeclConformancesVisitor::visitAggTypeDecl`,
and it will fill the body of this default ctor in `SemanticsDeclBodyVisitor::visitAggTypeDecl`.

But there're some situations where it will not fill the body of the ctor, so if that is the case, we will remove the empty default ctor. (Though I still don't quite understand why we have to remove it, removing it will cause lots of errors).

We're actually able to detect those situations during creating the ctor, however the logic to check those situations are part of `SemanticsDeclBodyVisitor::visitAggTypeDecl`, and there is not a very good way to re-use the results of this logic.

So I'd say let's keep it, as cost of removing it could be smaller than "detecting and not create it".

source/slang/slang-check-conversion.cpp
kaizhangNV: already fixed.

source/slang/slang-check-decl.cpp
kaizhangNV: This change already did the 
`this = DefaultConstruct(This);`

but it looks like `DefaultConstruct` will make everything to 0, and doesn't handle the initExpr and the base class.

source/slang/slang-check-conversion.cpp
kaizhangNV: It doesn't work for `variadic parameter pack`.
Test code something like this

```
struct S<T:__BuiltinArithmeticType>
{
    T v;
    __init(T x)
    {
        v = x;
    }
}
struct MyType<T:__BuiltinFloatingPointType>
{
    T val;
    T vals[10];

    __init(S<T> input)       // -> this works fine
    {
        val = input.v;
    }

    __init<each T>(expand S<each T> value)    // -> this doesn't work
    {
    }

    T getVal()
    {
        return val;
    }
}

```
diagnose:
```
error 38029: type argument 'each MyType<T>.T' does not conform to the required interface '__BuiltinArithmeticType'
    __init<each T>(expand S<each T> value)
```

kaizhangNV: @csyonghe to take a look whether I made any syntax error for above test code.
If not, we should create a new task to accomplish this.

csyonghe: This error is because `T` isn't constraint to `__BuiltinArithmeticType` as required by the interface.
You will need:
```
    __init<each T>(expand S<each T> value)
         where T : __BuiltinArithmeticType
    {
    }
```

csyonghe: Also please do another merge with tot, I have recently overhauled the synthesis of generic requirements that should in theory take care of this.

source/slang/hlsl.meta.slang
kaizhangNV: removed the WAR, and it doesn't crash anything.

kaizhangNV: https://github.com/shader-slang/slang/pull/5107/commits/69cb426a73d67ddfdabcca87da4b49f544fcb507

source/slang/slang-check-overload.cpp
kaizhangNV: rename this function to `_tryOverloadWithDefaultInit`, and add some comment to explain this.
https://github.com/shader-slang/slang/pull/5107/commits/2b6f7a26688b60a74fde979fbfff28a24a493e5c

source/slang/slang-check-decl.cpp
kaizhangNV: Close this conversion for now, please re-open it if we want to re-visit it.

source/slang/slang-check-conversion.cpp
kaizhangNV:
```
    __init<each T>(expand S<each T> value)
         where T : __BuiltinArithmeticType
```
this is working.

But 
```

    [mutating]
    T writeData(T value)
    {
        this.val = value;
        return value;
    }

    __init<each T>(expand S<each T> value)
         where T : __BuiltinArithmeticType
    {
        accept(expand writeData((each value).v));
    }
```
but this is wrong again:
```
error 30019: expected an expression of type 'T', got 'each T'
        accept(expand writeData((each value).v));

```

PR #4856, opened 2024-08-15
Tuple swizzling and element access.

csyonghe:
This PR implements tuple swizzling and element access. Partially completes proposal SP#008.


source/slang/slang-ir-inst-defs.h
```diff
@@ -566,6 +571,7 @@ INST(SwizzledStore, swizzledStore, 2, 0)
 /* IRTerminatorInst */
 
     INST(Return, return_val, 1, 0)
+    INST(Yield, yield, 1, 0)
```
jkwak-work, line 574: I just remembered that c# uses the keyword, "yield" for something else. It is similar to how async keyword works for java-script.
From my experience from game engines, the word, "yield", is more often used for thread related control.
I am not sure if it will be good to go with the keyword "yield" for the variadic template feature.

source/slang/slang-ir-specialize.cpp
```diff
@@ -597,7 +606,7 @@ struct SpecializationContext
     {
         // Note: While we currently have named the instruction
         // `lookup_witness_method`, the `method` part is a misnomer
-        // and the same instruction can look up *any* interface
+        // and the same instruction can look up *any* interfacemay
```
jkwak-work, line 609: Typo.
> interface may
With a space character in the middle.

tests/language-feature/tuple/tuple-basic.slang
```diff
@@ -0,0 +1,25 @@
...
+
+void store(inout Tuple<int, float, uint> t)
+{
+    t._1_2 = Tuple<float, uint>(3.0, 4);
+}
+
+[numthreads(4, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+    var t = Tuple<int, float, uint>(1, 2.0, 3u);
```
jkwak-work, line 19: Unrelated to this code but this made me to wonder if we need to worry about alignment when the types are, as an example, like following,
```
Tuple<int8_t, float> // float values may not be aligned???
```

jkwak-work (approved):
Looks good to me.

I know it is little early but Tuple will need to be explained in the user guide later.

tests/language-feature/tuple/tuple-basic.slang
csyonghe: A tuple type is equivalent to a struct type with the same typed fields. So it is equivalent to `struct S { int8_t a; float b};` and will follow struct layout rules, which means under natural layout `b` will be at offset 4.

source/slang/slang-ir-specialize.cpp
csyonghe: Fixed.

source/slang/slang-ir-inst-defs.h
csyonghe: As discussed offline, I believe the yield here has very similar meaning to c#'s yield, so keeping this name.

kaizhangNV (approved):
LGTM

kaizhangNV:
Better to add doc to explain at IR level.

This change involves two different changes. One is described in the PR description, another one is the `countof`.
You'd better to change the description.

Also the PR is too big.

source/slang/slang-check-expr.cpp
```diff
@@ -3815,6 +3861,108 @@ namespace Slang
         return CreateErrorExpr(memberRefExpr);
     }
 
+    Expr* SemanticsVisitor::checkTupleSwizzleExpr(MemberExpr* memberExpr, TupleType* baseTupleType)
```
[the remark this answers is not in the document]
csyonghe: The reason here is similar to vector/matrix swizzles in that if we see `v.xyz` during parsing, since we haven't checked the type for `v` yet (this is done during semantic checking), we don't know if this is a swizzle or just a normal lookup. We only know it is a swizzle after checking `v`, and at that time we will have to parse the member string again to get the swizzle structure.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2416,7 +2574,7 @@ struct SpecializationContext
         bool anyAbstractPack = false;
         for (UInt i = 0; i < expandInst->getCaptureCount(); i++)
         {
-            if (!as<IRTupleType>(expandInst->getCaptureType(i)))
+            if (!as<IRTypePack>(expandInst->getCaptureType(i)))
             {
                 anyAbstractPack = true;
```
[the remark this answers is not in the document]
csyonghe: Done.

csyonghe:
@kaizhangNV I updated the PR per your comments, and updated the description of this PR to include a section about `countof` implementation. I apologize for the big PR but I still need to balance the effort on splitting PRs and the effort on other priorities.

kaizhangNV (approved):

Issue #4823, opened 2024-08-12
Expression type incompatibility with module compilation


venkataram-nv:
Gives the following sources:
```
// hit.slang
import environment;

StructuredBuffer <Environment_sample_data> environment_sample_buffer;

float3 sample_lights(inout uint seed)
{
    return environment_sample(environment_sample_buffer, seed);
}

struct RadianceHitInfo
{
    float3 contribution;
};

struct Attributes
{
    float2 bary;
};

[shader("closesthit")]
void closesthit(inout RadianceHitInfo payload, Attributes attrib) {}
```
and
```
// environment.slang
module environment;

uint lcg(inout uint prev)
{
    const uint LCG_A = 1664525u;
    const uint LCG_C = 1013904223u;
    prev = (LCG_A * prev + LCG_C);
    return prev & 0x00FFFFFF;
}

public float rnd(inout uint prev)
{
    return ((float) lcg(prev) / (float) 0x01000000);
}

public struct Environment_sample_data
{
    uint alias;
    float q;
};

public float3 environment_sample(StructuredBuffer <Environment_sample_data> sample_buffer, inout int seed)
{
    float3 xi;
    xi.x = rnd(seed);
    xi.y = rnd(seed);
    xi.z = rnd(seed);
    return xi.z;
}
```

If I run the following series of commands:
```
slangc.exe hit.slang -stage closesthit -entry closesthit -target dxil -o monolithic.dxil      // Monolithic
slangc.exe environment.slang -o environment.slang-module                                      // Precompiled module
slangc.exe hit.slang -stage closesthit -entry closesthit -target dxil -o precompiled.dxil     // Module linking
```
then the same error pops up the second time I compile:
```
hit.slang(8): error 30019: expected an expression of type 'StructuredBuffer<Environment_sample_data, DefaultDataLayout>', got 'StructuredBuffer<Environment_sample_data, DefaultDataLayout>'
    return environment_sample(environment_sample_buffer, seed);
                              ^~~~~~~~~~~~~~~~~~~~~~~~~
```

The type names reported are exactly the same.

bmillsNV:
@venkataram-nv can you take a look at this and see if you can debug it?

venkataram-nv:
Looking into the issue further, here is what I found:

The error is generated from `_failedCoercion`, from the `_coerce` method. In `_coerce`, the state dump (from VS) is as follows:

[...]

PR #4859, opened 2024-08-15
Include inout cast operation as an aliasing instruction

venkataram-nv:
Noticed this when warnings are issued from #4823, so there are additional tests added to `inout-never-used.slang`.

Previously, the warnings were:
```
environment.slang(22): warning 41022: inout parameter 'seed' is never written to
public float3 environment_sample(StructuredBuffer <Environment_sample_data> sample_buffer, inout int seed)
              ^~~~~~~~~~~~~~~~~~
hit.slang(5): warning 41022: inout parameter 'seed' is never written to
float3 sample_lights(inout uint seed)
       ^~~~~~~~~~~~~
```

With this PR they should not be emitted.


jkwak-work (approved):
Looks good to me.

Issue #4792, opened 2024-08-06
Add support for generics in precompiled DXIL feature


cheneym2:
Add support for Slang modules that use generics as per https://docs.google.com/document/d/1AR9dCZ5Q_0bTMly2pCTxSX1zpL17o7vN7Ud7_lmvLmM/edit design doc.

PR #4860, opened 2024-08-15
Support mixture of precompiled and non-precompiled modules

cheneym2:
This changes the implementation of precompile DXIL modules to accept combinations of modules with precompiled DXIL, ones without, and ones with a mixture of precompiled DXIL and Slang IR.

During precompilation, module IR is analyzed to find public functions which appear to be capable of being compiled as HLSL, and those functions are given a HLSLExport decoration, ensuring they are emitted as HLSL and preserved in the precompiled DXIL blob. The IR for those functions is then tagged with a new decoration AvailableInDXIL, which marks that their implementation is present in the embedded DXIL blob. The DXIL blob is attached to the IR as before, inside a EmbeddedDXIL BlobLit instruction.

The logic that determines whether or not functions should be precompiled to DXIL is a placeholder at this point, returning true always. A subsequent change will add selection criteria.

During module linking, the full module IR is available, as well as the optional EmbeddedDXIL blob. The IR for functions implemented by the blob are tagged with AvailableInDXIL in the module IR.

After linking the IR for all modules to program level IR, the IR for the functions marked AvailableInDXIL are deleted from the linked IR, prior to emitting HLSL and compiling and linking the result.

This change also changes the point of time when the module IR is checked for EmbeddedDXIL blobs. Instead of happening at load time as before, it happens immediately before final linking, meaning that the blob does not need to be independently stored with the module separate from the IR as was done previously.

Work on #4792


source/slang/slang-compiler-tu.cpp
```diff
@@ -59,27 +61,76 @@ namespace Slang
...
 
-        if (res != SLANG_OK)
+        for (auto inst : module->getGlobalInsts())
         {
-            return res;
+            if (inst->getOp() == kIROp_Func)
+            {
+                bool hasResourceType = false;
+
+                // DXIL does not permit HLSLStructureBufferType in exported functions
```
csyonghe, line 73: This won't work if there is a struct type containing the resource. The right way is to mark all functions as "attempt to export" now, and make them real "export" after all the legalization passes if the function body meets requirements.

source/slang/slang-compiler-tu.cpp
```diff
@@ -59,27 +61,76 @@ namespace Slang
...
+            ComPtr<IArtifact> outArtifact;
+            SlangResult res = codeGenContext.emitTranslationUnit(outArtifact);
+            sink.getBlobIfNeeded(outDiagnostics);
+            if (res != SLANG_OK)
+            {
+                return res;
+            }
+
+            // Mark all exported functions as available in dxil
+            for (auto inst : module->getGlobalInsts())
```
csyonghe, line 110: Again, because the final decision will be made during the compilation process, we need a way to associate the linked and legalized IR back to the original module IR and mark the original IR based on the functions in the linked IR.

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -145,6 +145,28 @@ bool removeRedundancyInFunc(IRGlobalValueWithCode* func)
...
+        {
+            continue;
+        }
+        if (globalInst->findDecoration<IRAvailableInDXILDecoration>())
+        {
+            globalInst->findDecoration<IRNameHintDecoration>()->dump();
+            toRemove.add(globalInst);
+        }
+    }
+    for (auto inst : toRemove)
```
csyonghe, line 164: I don't think you can remove the entire inst, because the glue HLSL still need to contain an extern declaration for the function. Also if you just remove the function inst, the IR will be in an invalid state because calls of those function will now be calls into wild pointers. You should just remove the body of the functions so it will be treated as a declaration in our emit logic.

source/slang/slang-compiler-tu.cpp
cheneym2: Do you have some idea on how to do that mapping from linkedIR to original module IR?  I'm picturing the IR undergoing lots of cloning and modification during the link and legalization passes.  Would a decoration that held the pointer to the original IRInst survive until the IR was fully linked and optimized?

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: Could you recommend a test case that would show this problem?

source/slang/slang-compiler-tu.cpp
cheneym2: I'm considering a new decoration like "TransientExport" that I can treat like "HLSLExport" in "_isHLSLExported" but would be easy to remove after.

cheneym2: When marking functions as "real export", please confirm if that means adding kIROp_HLSLExportDecoration.  Also, is the right time to do that during linkAndOptimizeIR() after linkIR()?  That's where I'm looking right now, but my intuition is that I need to inspect IR before lowerGenerics() is called, however that is before much legalization.

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: Maybe my test needs to call a normal export function from inside a generic function in the library module. Seems like that would show the problem since the module IR would have a reference to a function that gets deleted in this patch.

source/slang/slang-compiler-tu.cpp
cheneym2: Since I'm already defining a decoration for "attempt to export" named "TransientExportDecoration", I'm now trying to add a pointer operand to it with the payload being the module IR's Inst to remember the mapping from linked IR back to module IR.

cheneym2: I tried using a PtrLit operand but it looks like cloneValue() deliberately refuses to copy non-null pointer values.  I'll try making up my own integer->IRInst* mapping

cheneym2: I've added an integer operand to the new "TransientExportDecoration" that would allow the mapping of linked IR functions back to the pre-linked and legalized module IR, however, I'm unsure how to correctly do the post-link/legalization filtering and correlation to module IR.

In linkandOptimizeIR(), at the end of the function is where I'm looking to inspect the post-linked IR. I can iterate through the global functions and remove the ones that define or use generics.  Given that, I think I have two approaches.  Either:

1) In linkandOptmizeIR(), after I filter the IR for generics, dig through the module IR and tag things as availableInDXIL that correpond to functions that still exist in the linked IR
or
2) Do the IR filitering at the end of linkAndOptimizeIR() but don't do anything with module IR yet.  Rather, let the function exit and then in precompileForTargets(), dig back into the output artifact and somehow get the linked IR to check for the "TransientExportDecoration" markers.  I don't know the commands to do that if it's currently possible. The output artifact blob contains just DXIL I think.  That is, in precompileForTarget(), I call codeGenContext.emitPrecompiledDXIL(outArtifact), followed by "outartifact->loadBlob()".  Not sure how I would also be able to get a hold of the linked IR.

csyonghe: There is already a `IRExportDeocration` that encodes the mangled name of symbols as long as they are considered "exported" in the slang IR sense. Note that this is not the same as `IRHLSLExportDeocration`, which are for symbols marked with `export` in slang code.

All public symbols should automatically have `IRExportDecoration` already.

cheneym2: I think you're saying that I don't need to add an integer for the cross referencing and I could use mangled names from IRExportDecoration instead. What about the second part though, which is essentially whether it makes more sense to apply the "availalbleInDXIL" decoration to module IR at the end of linkAndOptimizeIR() or if I can inspect the linked IR results from precompileForTargets() after emitTranslationUnit() is finished producing an output DXIL artifact.

csyonghe: I believe either option can work. We can do it in `precomileForTargets()` now for cleanliness, unless we hit any issues with that.

cheneym2: Okay, I have something working using the mangled names for mapping, however, I failed to figure out how to correctly extract the linked IR from the output artifact from codeGenContext.emit*() call.

My upcoming patch revision does work, but it uses what I assume is a hack.  I've added a new pointer to IRModule inside TargetProgram.  I'll be looking for advice on how to do that properly.

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: Current revision guts the function instead of deleting it.

source/slang/slang-compiler-tu.cpp
cheneym2: Current revision does the correlation using mangled name in precompileForTarget, however the retrieval of the linked IR is a hack, I believe.

source/slang/slang-emit.cpp
```diff
@@ -1585,6 +1645,9 @@ SlangResult CodeGenContext::emitEntryPointsSourceFromIR(ComPtr<IArtifact>& outAr
             linkedIR));
         
         auto irModule = linkedIR.module;
+
+        // TODO: How to store this in the artifact?
+        this->getProgram()->linkedIRModule = irModule;
```
cheneym2, line 1650: This was something I didn't figure out, please advise.

source/slang/slang-compiler.h
```diff
@@ -556,6 +556,8 @@ namespace Slang
             ///
         Scope* _getOrCreateScopeForLegacyLookup(ASTBuilder* astBuilder);
 
+        // TODO: What is proper way to get linked IR after emitPrecompiledDXIL?
+        RefPtr<IRModule> linkedIRModule;
```
cheneym2, line 560: Pretty sure I should be able to get this from the output artifact, but didn't understand how.

source/slang/slang-emit.cpp
csyonghe: I don't think we should return the linked IR. Instead, we should collect a list of mangled names that we have exported in the dxil, and attach that string list as the metadata of the output IR in linkAndOptimizeIR.

csyonghe: Please checkout `collectMetadata` and store the list of mangled name into `ArtifactPostEmitMetadata`.

cheneym2: Done!

source/slang/slang-compiler-tu.cpp
cheneym2: Fixed. Getting at linked IR is no longer necessary as the list of exported functions is saved to metadata in the artifact.

source/slang/slang-ir-inst-defs.h
```diff
@@ -782,6 +782,7 @@ INST_RANGE(BindingQuery, GetRegisterIndex, GetRegisterSpace)
     INST(PreciseDecoration,                 precise,                0, 0)
     INST(PublicDecoration,                  public,                 0, 0)
     INST(HLSLExportDecoration,              hlslExport,             0, 0)
+    INST(TransientExportDecoration,         transientExport,        0, 0)
```
csyonghe, line 785: Maybe DownstreamModuleExport ?

source/slang/slang-ir-redundancy-removal.h
```diff
@@ -10,4 +10,6 @@ namespace Slang
     bool removeRedundancyInFunc(IRGlobalValueWithCode* func);
 
     bool eliminateRedundantLoadStore(IRGlobalValueWithCode* func);
+
+    void removeAvailableInDXIL(IRModule* module);
```
csyonghe, line 14: removeAvailableInDownstreamModuleDecorations()

source/slang/slang-emit.cpp
```diff
@@ -1460,6 +1513,13 @@ Result linkAndOptimizeIR(
     if (!targetProgram->getOptionSet().shouldPerformMinimumOptimizations())
         checkUnsupportedInst(codeGenContext->getTargetReq(), irModule, sink);
 
+    if (targetProgram->getOptionSet().getBoolOption(CompilerOptionName::EmbedDXIL))
+    {
+        // We need to remove all the functions that can't be represented
+        // in a DXIL library, eg. with resources.
+        SLANG_RETURN_ON_FAIL(removeNonEmbeddableDXIL(irModule));
```
csyonghe, line 1520: I don't think you can just remove these functions, because I can have a module that is:
```
internal void nonEmbedable(Texture2D t) { ... }

public void embeddable() { nonEmbeddable(...); }
```

All we can do is to say we don't mark `nonEmbeddable` as `export` when emitting HLSL.

source/slang/slang-ir-util.cpp
```diff
@@ -244,6 +244,14 @@ bool isSimpleDataType(IRType* type)
     }
 }
 
+bool isSimpleHLSLDataType(IRInst* inst)
+{
+    // TODO: Add criteria
+    // https://github.com/shader-slang/slang/issues/4792
+    (void)inst; // avoid unused param error
```
csyonghe, line 251: SLANG_UNUSED(inst);

tests/library/precompiled-dxil-generics.slang
```diff
@@ -0,0 +1,28 @@
+// precompiled-dxil-generics.slang
+
+// A test that uses slang-modules with embedded precompiled DXIL and a library containing generics.
+// The test compiles a library slang (export-library-generics.slang) with -embed-dxil then links the
+// library to entrypoint slang (this file).
+// The test passes if there is no errror thrown.
+// TODO: Check if final linkage used only the precompiled dxil.
+
+//TEST(windows):COMPILE: tests/library/export-library-generics.slang -o tests/library/export-library-generics.slang-module -embed-dxil -profile lib_6_6 -incomplete-library
```
csyonghe, line 9: I don't think `(windows)` here is doing anything? Why do we need it?

csyonghe: It is better to use `TEST:SIMPLE(filecheck=CHECK): ...` and use filecheck to verify the compiled results, then just check for the return code of a compile.

cheneym2: -embed-dxil only works on windows

source/slang/slang-ir-metadata.cpp
```diff
@@ -42,8 +42,19 @@ static void _insertBinding(List<ShaderBindingRange>& ranges, LayoutResourceKind
 void collectMetadata(const IRModule* irModule, ArtifactPostEmitMetadata& outMetadata)
 {
     // Scan the instructions looking for global resource declarations
+    // and exported functions.
     for (const auto& inst : irModule->getGlobalInsts())
     {
+        auto func = as<IRFunc>(inst);
+        if (func)
```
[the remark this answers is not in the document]
cheneym2: Done

csyonghe (approved):

csyonghe:
Looks good! Merging.

Issue #4742, opened 2024-07-25
Stop using SpvStorageClass enums as `AddressSpace` values during spirv legalization.


csyonghe:
We should expand `AddressSpace` enum to something that can cover our needs for spirv code generation with cases that maps 1:1 to SpvStorageClass, and stick to values in `AddressSpace` during spirv legalization.

Right now this is messy as we are using SpvStorageClass values directly as AddressSpace operand in pointer types.

PR #4861, opened 2024-08-16
Remove using SpvStorageClass values casted into AddressSpace values

expipiplus1:
Also removes support for specific storage classes in __target_intrinsic snippets

Closes https://github.com/shader-slang/slang/issues/4742


csyonghe (approved):

csyonghe:
Updating the label to non-breaking because I don't think a user would be seeing any breaking behaviors.

Currently we don't promise compatibility of generated slang IR modules between slang versions, so change in IR doesn't count as a breaking change.

PR #4864, opened 2024-08-16
Make sure to resolve overloaded expr for call args.

### LINKED ISSUE #4863: Erroneous Compilation Error with Slang v2024.9.2
After updating from 2024.9.1 to 2024.9.2, we noticed the following (new) compilation error.  We've prepared a simple slang shader repro below.  

Slang Repro: [test.slangRepro.txt](https://github.com/user-attachments/files/16639880/test.slangRepro.txt)

The erroneous errors are as follows:
```
error 39999: no overload for '*' applicable to arguments of type (overload group, overload group)
>
>  Output[ipos] = foo.bar4 * foo.bar1;
```
```
error 30047: argument passed to parameter '0' must be l-value.
>
>  foo.bar1 += Input[ipos].x;
```

Slangc command line:
`-entry main -target glsl -zero-initialize`

Shader:
```
layout(binding = 0)
RWTexture2D<float4> Output; 

layout(binding = 1)
Texture2D<float4> Input; 

interface IFoo
{
  property float4 bar4 {get; set;}
  property float bar1 {get; set;}
};

struct Foo : IFoo
{
  uint4 data;
  property float4 bar4
  {
    get { return float4(asfloat(data.xyz), 1); }
    set { data.xyz = asuint(newValue.xyz); }
  }
  property float bar1
  {
    get { return asfloat(data.w); }
    set { data.w = asuint(newValue); }
  }
};

void doThing(inout Foo foo, uint2 ipos)
{
  foo.bar4 = Input[ipos];
  foo.bar1 = Input[ipos].w;
  foo.bar1 += Input[ipos].x;
}

[shader("compute")]
[numthreads(16, 16, 1)]
void main(uint2 ipos : SV_DispatchThreadID)
{
  Foo foo;
  doThing(foo, ipos);
  Output[ipos] = foo.bar4 * foo.bar1;
}
```

csyonghe:
Closes #4863.

The fix here is to make sure we resolve overloaded exprs for call arguments.

### DIFF
--- a/source/slang/slang-check-expr.cpp
+++ b/source/slang/slang-check-expr.cpp
@@ -2652,7 +2652,7 @@ namespace Slang
         // Next check the argument expressions
         for (auto & arg : expr->arguments)
         {
-            arg = CheckTerm(arg);
+            arg = CheckExpr(arg);
         }
 
         // if the expression is '&&' or '||', we will convert it
--- a/tests/bugs/gh-4863.slang
+++ b/tests/bugs/gh-4863.slang
@@ -1,0 +1,36 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUFFER):-compute -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUFFER):-vk -compute -output-using-type
+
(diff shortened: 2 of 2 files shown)


jkwak-work (approved):
Looks good to me

Issue #10, opened 2017-06-15
Direct generation of binary formats (DXBC, SPIR-V)


tfoleyNV:
There is code in place for allowing the API user to directly ask for binary formats like DX bytecode as output from a compilation, rather than strings, but the actual function for querying output assumes it can return a null-terminated string (no length is passed) and the internal implementation uses the `String` type which isn't really suitable for a byte buffer.

tfoleyNV:
This was implemented in #67 (with fixes in #72).

PR #4865, opened 2024-08-16
Proposal: A simpler and more flexible `IDifferentiable` system

### LINKED ISSUE #4034: Rethink `IDifferentiable` system for buffer/ptr types

saipraveenb25:
Designed to work for both value types as well as ref/ptr/buffer types.

Resolves #4034


docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+	typealias Differential = DPtr<T.Differential>;
+	
+	Buffer<T> buffer;
+	uint64 offset;
+	
+	[BackwardDerivative(get_bwd)]
+	[BackwardDerivative(get_fwd)]
+	T get() { return this.buffer[offset]; }
+	
+	[deriv_method] DifferentialPair<T> get_fwd()
```
csyonghe, line 157: use_fwd?

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+	{ 
+		return diffPair(this.p.buffer[offset], this.d().buffer[offset]);
+	}
+	
+	[deriv_method] void get_bwd(Differential d)
```
csyonghe, line 162: use_bwd?

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+float f_with_use_expansion(float a)
+{
+	DiffPair<float> a_extra = a.use();
+	return add(a, a_extra);
+}
+
+// After fwd-mode:
+DiffPair<float> f_fwd(DiffPair<float> dpa)
+{
+	DiffPair<float> dpa_extra = dpa.use_fwd();
```
csyonghe, line 223: what does `DiffPair<float>.use_fwd` do?

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
+## Problem
```
csyonghe, line 1: Should start with a title
SP #010: New differentiable type system

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Workarounds
+At the moment the primary workaround is to use a **non-differentiable buffer type** with differentiable methods, and always initialize the object with two pointers for both the primal and differential buffers. This is how our `DiffTensorView<T>` object works. 
+Unfortunately, this is a rather hacky workaround with several drawbacks: 
+1.  `DiffTensorView<T>` does not conform to `IDifferentiable`, but is used for derivatives. This makes our type system less useful as checks for `is_subtype` from applications using reflection need workarounds to account for corner cases like these.
+2.  `DiffTensorView<T>` always has two buffer pointers even when used in non-differentiable methods. This is extra data in the struct, and potentially extra tensor allocations (we explicitly handle this case in `slangtorch` by leaving the diff part uninitialized if a primal method is invoked)
+3.  Higher-order derivatives don't work well with this workaround. Differentiating a method twice needs a set of 4 pointers, but we need to account for this ahead of time by using new types like `DiffDiffTensorView` that worsens the problem of carrying around extra data where its not required.
```
csyonghe, line 75: How does higher order diff work with the new system?

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+Some extras in stdlib allow us to constrain the diffpair type for things like `IArithmetic`
+```csharp
+// --- STDLIB EXTRAS ---
+
+interface ISelfDifferentiableValueType : IDifferentiableValueType
+{
+	// Force arithmetic types to be a differential pair of the same two types.
+	// Make it simple to define derivatives of arithmetic operations.
+	//
+	associatedtype Differential : This;
```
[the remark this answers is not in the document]
csyonghe: It is about time we add a `where` clause and implement type equality constraints.

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+	DifferentialPair<DPtr<float>> dpb, 
+	DifferentialPair<DPtr<float>> output)
+{
+	// forward pass.
+	var a_p = dpa.p.get();
+	var b_p = dpb.p.get();
+	// ----
+	
+	// backward pass.
+	float.Differential d_val = DPtr<float>::set_bwd(output); // set_bwd works on the entire pair.
```
csyonghe, line 276: where are the argument to `set`?

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+
+Additionally, we can use `IDifferentiableValueType` as the current `IDifferentiable` for backwards compatibility (it should just work in 95% of cases, since no one really defines dadd/dzero/dmul explicitly anyway) 
+
+Here's the new set of base interfaces:
+```csharp
+interface __IDifferentiableBase { } // Helper type for our implementation.
+interface IDifferentiableValueType : __IDifferentiableBase
+{
+	associatedtype Differential : IDifferentiableValueType & IDefaultInitializable;
+	[Differentiable] This use(); // auto-synthesized
```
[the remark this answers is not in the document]
saipraveenb25: I believe that using `[Differentiable]` in the method signature should automatically require the two derivatives to be provided?

docs/proposals/010-new-diff-type-system.md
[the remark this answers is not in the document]
saipraveenb25: Yes, this part is confusing.. what we want is the equality constraints. 
(We could implement `ISelfDifferentiableValueType` as a special case for now, in case the `where` clause implementation is too complex)

docs/proposals/010-new-diff-type-system.md
saipraveenb25: This is different from `use`. the `IDifferentiablePtrType` explicitly doesn't require a `use` (because we don't need any type-specific handling of bwd_diff(use)`. `use` is invoked internally by the compiler for any extra uses of an value-typed object (if used multiple times)

`get()` is a normal getter/setter method implemented on a buffer or pointer type (that a user can use) and is not required in the differentiable interface. 
It's effectively just derefencing or indexing in this scenario. (Maybe I should have used load() instead?). The core idea is that on these 'pointer-style' types all differentiation happens on the methods used to access the data, so we'll rely on custom derivatives, and simply make sure the pair is passed as-is through all the generated code so the buffer references are available to the custom derivative code.

docs/proposals/010-new-diff-type-system.md
saipraveenb25: In this scenario, we can have a `DTensorView<T> : IDifferentiablePtrType` which can have a single buffer in primal code. 
It becomes `DifferentialPair<DTensorView<T>>` under the first derivative pass (2 buffers), and then becomes `DifferentialPair<DifferentialPair<<DTensorView<T>>>` under the second derivative pass (4 buffers).

docs/proposals/010-new-diff-type-system.md
saipraveenb25: The main reason for introducing `IDifferentiablePtrType` is to make this explicit. It explicitly has no `use` and no way to initialize it, which means that this type will not get 'transposed'. They will behave similar to how `DiffTensorView<T>` works, with the main difference being that auto-diff itself will create the pair of buffers instead of us having to side-step the differentiable type system.

docs/proposals/010-new-diff-type-system.md
saipraveenb25: It is effectively a noop 'passthrough' in this case. It only exists because when converting to reverse-mode, we will replace `use_fwd` into `use_bwd` which is the method that has non-trivial logic. Both `use` and `use_fwd` are always intended to be passthroughs. 
(We could do away with `use` entirely and simply require the implementation of the backward part, which is just `dadd` but not static)

docs/proposals/010-new-diff-type-system.md
```diff
@@ -0,0 +1,283 @@
...
+{ }
+
+extension float
+{
+	// trivial auto-synthesis (maybe we even prevent the user from overriding this)
+	float use() { return this; } 
+	
+	// trivial auto-synthesis (maybe we even prevent the user from overriding this).
+	[ForwardDerivativeOf(use)]
+	[deriv_method] void use_fwd() { return this; }
```
csyonghe, line 134: why can't this be just a static method where the first parameter is `DiffPair<T>`? It seems that `[deriv_method]` isn't necessary.

saipraveenb25: @csyonghe: It's syntactic sugar for a static method that operates on pairs. It's mostly to keep code clarity. If a primal function is a member method that operates on `this`, then the derivative is also a member method that operates on a this-pair. 
But we can certainly avoid adding a new decorator, if it doesn't really help with clarity.

csyonghe: I would rather us not adding new decorations if it can be achieved with existing mechanisms to avoid increasing maintenance costs down the road.

PR #4870, opened 2024-08-20
Support dependent generic constraints.

### LINKED ISSUE #3804: generic interface cannot be used as a generic argument

### LINKED ISSUE #4882: Support generic type constraints that are dependent on another generic type parameter.
We need to allow our compiler to type check and emit code for things like:
```
interface IFoo<T> {...}
int f<T, B:IFoo<T>>()
{}
```

Where the constraint for `B` is dependent on `T`. The main work is to make sure our IR rules around `IRParam` is relaxed so that the type and decorations of an `IRparam` can reference things defined later in the same basic block.

csyonghe:
Closes #4882.
Closes #3804.

This change fixes the type system and handling around `IRParams` so that we can define a generic where one of the type parameter conforms to an generic interface parameterized on another type parameter.

The main fix is to make sure when an IRParam has type or decorations that references another type that came later in the generic block, `clone` doesn't mess things up.

Another change is to make sure we can infer the type argument `T` when it is referenced only through the base interface type, for example, given:
```
struct Foo {}
interface IBar {...}
struct Bar<T> : IBar<T> { ... }

int f<T, B:IBar<T>>(B b)
{
]

void call()
{
    Bar<Foo> obj;
    f(obj); // should be able to infer `f.T` from `Bar<Foo>`.
}
```
Currently, our generic type inference proces is split into two parts: type unification and constraint solving. Type unification step is the step where we gather all the type constraints by structurally matching the argument type and parameter type. Then in the second step, we figure out what each generic type parameter should be to satisfy all the type constraints.

For example, in this `f` call, we have parameter type `B` and argument type `Bar<Foo>`, so type unification step will generate a constraint stating `B == Bar<Foo>`. Since there is no parameter whose type is dependent on `T`, there will be no constraints generated for `T`. When we get to the solving stage, we will conclude that `B` should be `Bar<Foo>` because of the constraint we collected in the first stage, but then we have no info on what `T` should be.

Note that we can deduce `T` if we continue to unify the base interface type of `Bar<Foo>` with the type constraint of type `B`, which means we should unify `IBar<Foo>` and `IBar<T>`, and this will leads to constraint `T==Foo` being added to the system that will tell us what `T` should be. However we don't/cannot check a type argument (in this case `B=Bar<Foo>`) against its stated based type constraint (`IBar<T>`) until the type argument itself is solved. This means that our constraint solving process needs to be iterative: during the intial unification stage, we will discover constraint `B=Bar<Foo>`, and when solving for the constraint, we will discover new constraints that `IBar<Foo>:IBar<T>` and we will add that new constraint and continue the solving process. This means that we should change our generic argument inference from a cleanly separated two-stage process into an iterative process, where we keep going until we have processed all discovered constraints.


source/slang/slang-ir-clone.cpp
```diff
@@ -181,10 +193,14 @@ static void _cloneInstDecorationsAndChildren(
...
-            pair.newInst = newChild;
-            pairs.add(pair);
+            pairs.add({ oldChild, newChild });
+        }
+
+        // Add params to a separate list so that we can clone their types later
+        // after the entire block is cloned.
+        if (oldChild->getOp() == kIROp_Param)
+        {
+            paramPairs.add({oldChild, newChild});
```
kaizhangNV, line 203: Is there any reason that you can't put this at line 183?

source/slang/slang-check-overload.cpp
```diff
@@ -883,7 +883,8 @@ namespace Slang
 
         for( auto constraintDecl : genericDeclRef.getDecl()->getMembersOfType<GenericTypeConstraintDecl>() )
         {
-            DeclRef<GenericTypeConstraintDecl> constraintDeclRef = m_astBuilder->getGenericAppDeclRef(genericDeclRef, substArgs, constraintDecl).as<GenericTypeConstraintDecl>();
+            DeclRef<GenericTypeConstraintDecl> constraintDeclRef = m_astBuilder->getGenericAppDeclRef(
+                genericDeclRef, newArgs.getArrayView(), constraintDecl).as<GenericTypeConstraintDecl>();
```
kaizhangNV, line 887: Is this change to fix a bug?
When I compare to the old code, `substArgs` passed in won't change, while `newArgs` keep changing during iteration.

tests/language-feature/generics/dependent-generic-3.slang
```diff
@@ -0,0 +1,32 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -vk -shaderobj -output-using-type
+
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<int> outputBuffer;
+
+interface IOperation<T : __BuiltinFloatingPointType> // <== This is the problem
```
kaizhangNV, line 7: what is the problem?

kaizhangNV (approved):
LGTM

tests/language-feature/generics/dependent-generic-3.slang
csyonghe: This was copied from a previous GitHub issue, I need to remove the commment. There is no problem.

source/slang/slang-check-overload.cpp
csyonghe: Not technically a bug, but this is to allow subtype witness finding to make use of the already solved args as we make forward progress. This only matters when we have dependent generic parameters and wasn't an issue before.

source/slang/slang-ir-clone.cpp
csyonghe: I can do that.

source/slang/slang-check-overload.cpp
kaizhangNV: Nice to know.

PR #4871, opened 2024-08-20
Implement Path::createDirectoryRecursive

kaizhangNV:
Implement Path::createDirectoryRecursive with existing Path::createDirectory that uses system call instead of c++ standard lib.

### DIFF
--- a/source/core/slang-io.cpp
+++ b/source/core/slang-io.cpp
@@ -564,9 +564,57 @@ namespace Slang
 
     bool Path::createDirectoryRecursive(const String& path)
     {
-        std::error_code ec;
-        std::filesystem::create_directories(path.getBuffer(), ec);
-        return !ec;
+        String finalPath = Path::simplify(path);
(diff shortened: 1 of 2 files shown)


source/core/slang-io.cpp
```diff
@@ -564,9 +564,57 @@ namespace Slang
...
+        {
+            return false;
+        }
+
+        List<String> pathList;
+
+        // Check whether the parent directories exist, and add to the pathList if they are
+        // not, we will create all the directories from back of the list.
+        String parentDir = finalPath;
+        while(1)
```
csyonghe, line 578: `while(1)` may trigger style warnings in some compilers, in Slang we prefer `for(;;)`.

csyonghe (dismissed):

source/core/slang-io.cpp
kaizhangNV: Ah, I see. Will change.
I think I have another place also using that. I will change that as well.

PR #4885, opened 2024-08-20
Track uninitialized values of `Ptr<Specialize<T>>` inside type `T` without hang

### LINKED ISSUE #4878: Uninitialized variable check will infinitely loop if pointer to a specialized IR object is used
Problem:
Uninitialized variable checks will infinitely loop if user code uses the following pattern of:
```
Struct T
{
    Ptr<IRSpecialized<T>> next;
}
```

ArielG-NV:
Fixes #4878

Track uninitialized values of `Ptr<Specialize<T>>` inside type `T` without hang

Note: related to `tests/reflection/ptr/ptr-generic.slang`

### DIFF
--- a/source/slang/slang-ir-use-uninitialized-values.cpp
+++ b/source/slang/slang-ir-use-uninitialized-values.cpp
@@ -180,9 +180,8 @@ namespace Slang
             // Avoid the recursive step if its a
             // recursive structure like a linked list
             IRType* ptype = ptr->getValueType();
-            if (auto spec = as<IRSpecialize>(ptype))
-                if(auto resolvedType = as<IRType>(resolveSpecialization(spec)))
-                    ptype = resolvedType;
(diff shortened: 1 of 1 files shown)


source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -180,6 +180,9 @@ namespace Slang
             // Avoid the recursive step if its a
             // recursive structure like a linked list
             IRType* ptype = ptr->getValueType();
+            if (auto spec = as<IRSpecialize>(ptype))
+                if(auto resolvedType = as<IRType>(resolveSpecialization(spec)))
```
csyonghe, line 184: Shouldn't need `resolveSpecialization` function, just call `getResolvedInstForDecorations(ptype)`.

Issue #4874, opened 2024-08-20
Resource variables used in a global initializer function do not hoist correctly


ArielG-NV:
Problem:
Once compiled, code may leave assignments to resource variables in global scope with local variable assignments in global scope

Repro:
* Run the following test
```
// type-legalize-global-with-init.slang
//
// Confirm that type legalization can handle a global constant
// with a resource type or a type that recursively contains
// resources.
//
//TEST(compute):COMPARE_COMPUTE: -shaderobj
//
//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<uint> outputBuffer;

//TEST_INPUT:ubuffer(data=[1 2 3 4 5 6 7 8], stride=4):name=inputBuffer
RWStructuredBuffer<uint> inputBuffer;

static const RWStructuredBuffer<uint> gBuffer = inputBuffer;

struct Stuff
{
    __init(RWStructuredBuffer<uint> a1, RWStructuredBuffer<uint> b2)
    {
        a = a1;
        b = b2;
    }

    RWStructuredBuffer<uint> a;
    RWStructuredBuffer<uint> b;
}

static const Stuff gStuff = Stuff( inputBuffer, inputBuffer );

uint test(uint x)
{
    return gBuffer[x]
        + gStuff.a[x + 1] * 16
        + gStuff.b[x + 2] * 256;
}

[numthreads(4, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    let tid = dispatchThreadID.x;
    let inVal = tid;
    let outVal = test(inVal);
    outputBuffer[tid] = outVal;
}
```

...

PR #4894, opened 2024-08-21
Hoist resource variables using a global initializer function correctly

ArielG-NV:
Fixes: #4874
Goal:
Hoist resource variables using a global initializer function in such a way that any global with initializer function has any possible resource variables hoisted into their use location directly (otherwise many targets won't work).
Changes:
1. If legalize a global kIROp and it creates a global, make a GlobalVar

[only done for resource variables for simplicity of legalization pass]

2. Add a legalization pass that moves any global GlobalVar stores (initializers) into the GlobalVar init block
3. If we have a GlobalVar which is computed to be an alias (direct assignment to another variable), hoist the alias's value into IRGlobalVar use-sites.


source/slang/slang-ir-legalize-types.cpp
```diff
@@ -343,8 +343,20 @@ struct LegalCallBuilder
                 // result type of the function, so we know that
                 // the legalization funciton/call will use a `void`
                 // result type.
-                //
-                _emitCall(m_context->builder->getVoidType());
+
+                // If our call is in global scope (initializer) we should
+                // hoist the call into the start of every entry-point 
+                if (!m_context->builder->getInsertLoc().getInst()->getParent()
```
csyonghe, line 349: I understand what this is trying to achieve here, but I think this is likely not the right way.

Usually if you have a global variable and you want to initialize it with some code (e.g. a call), you create a `IRGlobalVar`, which is a `IRGlobalValueWithCode` that can have a body just like a `IRFunc`. Inside the body you create a basic block with instructions + a `return` instruct that returns the value of the init expr.

Here instead of inserting a call into global scope, we should be creating that IRGlobalVar, and rely on follow up passes to insert the logic in the body into entry points.

tests/compute/type-legalize-global-with-init-function.slang
```diff
@@ -0,0 +1,49 @@
...
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUF): -shaderobj
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUF): -vk -shaderobj
+//
+//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<uint> outputBuffer;
+
+//TEST_INPUT:ubuffer(data=[1 2 3 4 5 6 7 8], stride=4):name=inputBuffer
+RWStructuredBuffer<uint> inputBuffer;
+
+static const RWStructuredBuffer<uint> gBuffer = inputBuffer;
```
csyonghe, line 16: In general I think we will have issues when there are global resource variables that is being written to dynamically, so I am not sure if this is something we want to allow...

static const currently is overloaded to also mean compile time constants, maybe we should disallow static consts of resource type as well. Is this test coming from some existing user code?

ArielG-NV: > In general I think we will have issues when there are global resource variables that is being written to dynamically, so I am not sure if this is something we want to allow...


We will have an issue with dynamic resource variable accesses.
Slang is lacking a compiler pass to manage non-compile-time known resource accesses such as:
```
RWStructuredBuffer<uint> inputBuffer1;
RWStructuredBuffer<uint> inputBuffer2;
RWStructuredBuffer<uint> inputBuffer3;

[numthreads(4,4,4)]
{
    RWStructuredBuffer<uint> resource;
    if(dispatchThreadID.x == 1)
    {
        resource = inputBuffer1;
    }
    else if(dispatchThreadID.x == 2)
    {
        resource = inputBuffer2;
    }
    else if(dispatchThreadID.x == 3)
    {
        resource = inputBuffer3;
    }

    // This is illegal since resource is not known at compile time.
    outputBuffer[0] = resource[1];
}
```

From what I understand to add this would just piggy-back on our dynamic dispatch system logic:
1. Diagnose all non-constexpr resources, assign all resources of equal type a different ID (like with dynamic dispatch).
2. Any assignments to a resource assigns an ID (not the actual resource)
3. All accesses to a non-compile-time resource specializes all texture intrinsic uses to pick the correct texture using our ID (if/else binary tree)

ArielG-NV: > static const currently is overloaded to also mean compile time constants, maybe we should disallow static consts of resource type as well. Is this test coming from some existing user code?

No test currently uses `static const Stuff gStuff = Stuff( inputBuffer, inputBuffer );`. 
The problem is that this test use `static const Stuff gStuff = { inputBuffer, inputBuffer }`, which once #4854 is merged will be an issue.

csyonghe:
Reading at the code here, it seems clear to me that we need to legalize out the resource typed global variable out of existence before we reach `legalizeType`, instead of letting legalizeType doing weird things such as inserting a `store` in global scope and then try to clean up that mess in a separate pass.

In this case, the global `static const` should have been lowered into a `IRGlobalConstant` whose value is an `IRCall`. We should then recognize these `IRGlobalConstants` and `IRCall`s and inline them before hitting legalization.

ArielG-NV:
> In this case, the global static const should have been lowered into a IRGlobalConstant whose value is an IRCall. We should then recognize these IRGlobalConstants and IRCalls and inline them before hitting legalization.

I will work on this solution 👍

kaizhangNV:
Fix by https://github.com/shader-slang/slang/pull/6098

PR #4895, opened 2024-08-21
Dont require llvm for building generators

expipiplus1:


csyonghe:
@expipiplus1 can you post here once you are able to confirm a test release can be created with this PR?

expipiplus1:
> @expipiplus1 can you post here once you are able to confirm a test release can be created with this PR?

looks good, macos doesn't try to download llvm any more
https://github.com/shader-slang/slang/actions/runs/10484700990/job/29039485323

csyonghe (approved):

csyonghe:
Looks good, thank you!

Will we still have that racing condition where if some build process creates a release first then we can fail?

expipiplus1:
nope, because it shouldn't try to download anything at all. But in the unlikely event that it happens on a user machine (them building from a release before all the assets are populated) they'll now get a much more sensible error message.

Issue #4570, opened 2024-07-09
Use slang examples as initial test cases for the capture/replay system


kaizhangNV:
We need to add some test cases for the capture/replay system when it's done.
Currently, the slang examples are the good test cases because they all implemented by using new slang API.
And it's relative easy to integrate to our CI.

bmillsNV:
2 samples we'll focus on are: gpu printing, cpu hello world

kaizhangNV:
#4898 implement the basic infrastructure that making porting those examples easier. And added two examples: hello world, triangle.

In the follow up patches, I will add more tests and integrate the test flow into slang-unit-test.

kaizhangNV:
cpu-com-example also needs migration to new slang API.

kaizhangNV:
Found a slang bug when running `autodiff-texture`, the fix is by PR: https://github.com/shader-slang/slang/pull/4925.

kaizhangNV:
### Latest status:
Only 3 examples are not done yet.
 - gpu-printing: blocked by #4918
 - shader-toy: blocked by #4918
 - cpu-com-example: Didn't find corresponding API for `ICompileRequest::getTargetHostCallable`.

PR #4898, opened 2024-08-21
Feature/capture unit test

kaizhangNV:
Convert existing examples into tests for capture/replay layer.

- Add a simply TestBase class to the example base class, which handle the command line input parameter "--test-mode", so we can turn the examples into test mode.
- In the test mode, we first disable the swapchain and enable offline rendering, because we will run those tests on headless environment. Then we enable printing the hashcode of all the entry point functions to stdout.
- Since getting hash code is a slang API, it will recored by our recorder. Therefore, it can be replayed by the replayer, and we will also print the hash code the the stdout.
- The test script will parse the hash code from the stdout, and compare the hash code generated by tests with one generated by replayer.

Fix some bugs in record/replay:

- remove release() call for those com ptr.
- don't store the hash code because it's just a blob data not a handle.

### DIFF
--- a/examples/example-base/example-base.cpp
+++ b/examples/example-base/example-base.cpp
@@ -116,7 +116,7 @@ void WindowedAppBase::mainLoop()
     gTransientHeaps[frameBufferIndex]->finish();
 }
 
-void WindowedAppBase::OfflineRender()
+void WindowedAppBase::offlineRender()
 {
     gTransientHeaps[0]->synchronizeAndReset();
     renderFrame(0);
--- a/examples/example-base/example-base.cpp
+++ b/examples/example-base/example-base.cpp
@@ -17,19 +17,6 @@ Slang::Result WindowedAppBase::initializeBase(
     int height,
     DeviceType deviceType)
 {
-    // Create a window for our application to render into.
-    //
-    platform::WindowDesc windowDesc;
-    windowDesc.title = title;
-    windowDesc.width = width;
-    windowDesc.height = height;
-    windowWidth = width;
-    windowHeight = height;
-    windowDesc.style = platform::WindowStyle::Default;
-    gWindow = platform::Application::createWindow(windowDesc);
-    gWindow->events.mainLoop = [this]() { mainLoop(); };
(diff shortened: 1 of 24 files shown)


kaizhangNV:
The changes in `source/slang-record-replay/` directory is just minor bug fix, so feel free to skip it if you are not interested.

examples/example-base/example-base.cpp
```diff
@@ -102,23 +110,27 @@ Slang::Result WindowedAppBase::initializeBase(
...
-    if (frameBufferIndex == -1)
-        return;
 
     gTransientHeaps[frameBufferIndex]->synchronizeAndReset();
     renderFrame(frameBufferIndex);
     gTransientHeaps[frameBufferIndex]->finish();
 }
 
-void WindowedAppBase::createSwapchainFramebuffers()
+void WindowedAppBase::OfflineRender()
```
csyonghe, line 119: offlineRender

examples/example-base/test-base.cpp
```diff
@@ -0,0 +1,58 @@
...
+        if (wcscmp(szArglist[i], L"--test-mode") == 0)
+        {
+            m_isTestMode = true;
+        }
+    }
+
+    return 0;
+}
+#else
+int TestBase::parseOption(int argc, char** argv)
```
csyonghe, line 24: I guess it will be fine if we just use these version.

test-record-replay.sh
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
+
```
csyonghe, line 2: We cannot run this test on windows without bash. It is best to expose this as a slang-unit-test that uses the Process class to create processes.

examples/example-base/example-base.cpp
kaizhangNV: ack

examples/example-base/test-base.cpp
kaizhangNV: act

kaizhangNV:
Finish part of issue #4570.

Issue #4899, opened 2024-08-21
`tests\autodiff\generic-jvp.slang` test is incorrect


ArielG-NV:
Related to: #4877
`tests\autodiff\generic-jvp.slang` test constructs 'myvector' using `{...}` (`kIROp_MakeStruct`) allowing to differentiate when the type (and trivial ctor) is non-differentiable and should not be differentiable.

Issue #4877, opened 2024-08-20
`kIROp_MakeStruct` differentiates when it should not be differentiating


ArielG-NV:
Problem: 
Following code if ran with the following test should be outputting '0' but it is outputting '1' due to the code differentiating `    Data d = { y };`.
```
//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type
//TEST(compute):COMPARE_COMPUTE_EX:-cpu -compute -output-using-type -shaderobj

//TEST_INPUT:ubuffer(data=[0 0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<float> outputBuffer;

// This test isn't actually testing the output, but rather that the compiler doesn't crash upon 
// encountering a specific loop pattern. ('Data' is non-differentiable here, so the expected output is 0)s
// 

typedef DifferentialPair<float> dpfloat;
typedef float.Differential dfloat;

struct P
{
    bool terminated;
    bool isTerminated() { return terminated; }
    bool isHit() { return !terminated; }
};

struct Data
{
    //__init(float tin)
    //{
    //    this.t = tin;
    //}
    float t;
};

void updateData(Data data)
{
    data.t = data.t * data.t;
}

[BackwardDifferentiable]
float test_simple_while(float y, uint n)
{
    Data d = { y };
    //Data d = Data( y );
    P p;
    p.terminated = false;
    int i = n;

    if (p.isTerminated())
        return d.t;

    [MaxIters(4)]
    while (!p.isTerminated())
    {
        updateData(d);
        p.terminated = (i-- == 0);
        if (p.isTerminated())
            break;

        if (!p.isHit())
            break;
    }
    return d.t;
}

[numthreads(1, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    {
        dpfloat dpa = dpfloat(1.0, 0.0);

        __bwd_diff(test_simple_while)(dpa, 2, 1.0f);
        outputBuffer[0] = dpa.d;
    }
}
```

PR #4900, opened 2024-08-21
Correct the `generic-jvp.slang` test

ArielG-NV:
Fixes: #4899

Fixes invalid test results since `{...}` was differentiating the 'trivial constructor' equivalent of `myvector` when it should not (see #4877). This change modifies the test so it is correct so other PRs may be merged if indirectly/directly fixing the old use-case for this test.


tests/autodiff/generic-jvp.slang
```diff
@@ -115,7 +121,7 @@ extension myfloat3 : IDifferentiable
     [ForwardDifferentiable]
     static Differential dmul<T : __BuiltinRealType>(T a, Differential b)
     {
-        return { __realCast<Real, T>(a) * b.val };
+        return myfloat3(__realCast<Real, T>(a) * b.val);
```
csyonghe, line 124: Why can't we use initializer list anymore?

ArielG-NV: `{ __realCast<Real, T>(a) * b.val };` is producing incorrect results. `{__realCast<Real, T>(a) * b.val}` is not supposed to be differentiating but it is from what I understand.

This is related to https://github.com/shader-slang/slang/issues/4877

csyonghe: But with your change, we should map initializer list to ctor call?

ArielG-NV: > But with your change, we should map initializer list to ctor call?

Yes, this is what will happen once #4854 is merged.
Since the correct behavior is to not differentiate the ctor(?), `generic-jvp.slang` will start failing since it is doing something invalid.

csyonghe: In that case we should change the expected result or mark init as differentiable, and keep the initializer list here, right?

ArielG-NV: We can just change the result, that is completely fine
I was just worried about how very confusing that would be to include in the https://github.com/shader-slang/slang/pull/4854 PR rather than as something separate so the change is clearer.

csyonghe (approved):

PR #4901, opened 2024-08-21
Explicitly detach derivative when forming a non-differentiable struct out of differentiable args

saipraveenb25:
This fixes an issue where initializer lists get optimized out and lose information about non-differentiability.

There are 2 places where this could have been fixed:
1. When coercing initializer-list exprs, we can check for non-differentiable aggregate types and use a detach derivative on all the args.
2. Add an extra case in the peephole optimization step that adds detach-derivative when simplifying a make-struct of a non-differentiable type.

Even though solution 2 is more elegant, this PR goes with solution 1 simply to avoid having to use a differentiable-type-conformance-context that is used in the auto-diff IR passes to check for differentiability.

Fixes: #4877


ArielG-NV (approved):
Changes look good to me.

PR #4905, opened 2024-08-22
Make variadic generics work with interfaces and forward autodiff.

csyonghe:
This is a set of changes that enables interface methods with variadic parameters, and allow forward mode autodiff on variadic generic functions.

Partially addresses #4892 and #4891.

Backward autodiff still does not work because they require a variadic parameter to be `inout`, which we haven't implemented yet.

Main work done here:
- Type system rules on differentiating a type pack: if `P` is a type pack or abstract type pack, then `P.Differential` is `expand (each T).Differential`, and a function paramter pack of type `P` will translate to a parameter pack of `expand DifferentialPair<each P>` in its forward derivative.
- Make the specialization pass expand out a function parameter of `IRTypePack` type to actual individual parameters of the function, so that autodiff pass won't see any `IRTypePack` in parameter types.
- Synthesize `IDifferentiable` witness for a tuple/type pack in autodiff pass.
- Make sure autodiff passes can correctly handle tuple/type pack related insts, such as `IRMakeValuePack`.
- Fix name mangling so two functions with different settings of `[mutating]` or `[NoDiffThis]` attributes won't have the same mangled name.
- Added definition of `IFunc`, `IMutatingFunc`, `IDifferentiableFunc` and `IDifferentiableMutatingFunc` interfaces to `core` to enable some tests on autodiff'ing variadic generic functions.


source/slang/slang-check-decl.cpp
```diff
@@ -3975,6 +3967,33 @@ namespace Slang
                     addModifier(synParamDecl, clonedModifier);
                 }
             }
+
```
csyonghe, line 3970: Yong's note: in this function, we are synthesizing a required method in a `struct` type for an interface requirement. For example, if we have
```
interface IFoo<each T>
{
    void bar<each T>(expand each T args);
}
struct Impl : IFoo<int, float>
{
    void bar(int a, float b);
}
```

It is obvious that `Impl::bar` should be able to satisfy the requirement of `IFoo::bar`. However, the requirement `IFoo<int, float>::bar()` will have type `TypePack(int, float)->void`, but `Impl::bar` has type `(int, float)->void`. There is a mismatch in the type system between the implemntation method and the requirement.

To fill this gap, our compiler will invoke this function to try to synthesize a wrapper of `Impl::bar` to see if we can still call into `Impl::bar` from a synthesized wrapper `Impl::bar_wraper(TypePack(int, float) synthesizedArg)`, which have the exact same type as `IFoo::bar()`.

Here we are adding the synthesizing logic so that if our synthesized wrapper has any `TypePack` typed parameters, we will create individual local variables to store the elements of the type pack, and call the actual method with these individual local values, so that we will synthesize:

```
void Impl::bar_wrapper(TypePack<int,float> param)
{
    let a = param._0;
    let b = param._1;
    bar(a, b);
}
```

source/slang/slang-check-expr.cpp
```diff
@@ -1174,6 +1174,23 @@ namespace Slang
             }
         }
 
+        if (auto typePack = as<ConcreteTypePack>(type))
```
csyonghe, line 1177: Yong's note: here we add the front-end logic to derive the differential type of a type pack.
Obviously, a type pack's differential type should be another type pack where each element type is the differential type of the corresponding element in the original type pack. If the original element type is not differentiable, then the corresponding type in the differential type pack is `void`. If none of the types in the original type pack are differentiable, then we treat the entire type pack as non-differnetiable by simply returning `null` here.

source/slang/slang-check-expr.cpp
```diff
@@ -1368,6 +1385,13 @@ namespace Slang
                 });
             return;
         }
+
+        if (auto typePack = as<ConcreteTypePack>(type))
```
csyonghe, line 1389: Yong's note: the logic here is to make sure we register types inside a type pack to the differentiable type dictionary of a differentiable function.

source/slang/slang-check-expr.cpp
```diff
@@ -3598,6 +3652,10 @@ namespace Slang
             if (!isTypePack(baseType) && !as<TupleType>(baseType))
                 goto error;
         }
+        
+        if (auto tupleType = as<TupleType>(baseType))
```
csyonghe, line 3656: Here we are trying to figure out the captured type pack from the inner type of a `each X` type. If `X` is a Tuple type, because we know a tuple type is represented as `TupleType(TypePack(...))`, we should instead use the inner type pack as the captured type pack argument for the enclosing `expand` type.

source/slang/slang-ir-autodiff-transcriber-base.cpp
```diff
@@ -174,179 +174,9 @@ IRInst* AutoDiffTranscriberBase::maybeCloneForPrimalInst(IRBuilder* builder, IRI
 
 IRInst* _lookupWitness(IRBuilder* builder, IRInst* witness, IRInst* requirementKey);
 
-// Get or construct `:IDifferentiable` conformance for a DifferentiablePair.
-IRWitnessTable* AutoDiffTranscriberBase::getDifferentialPairWitness(IRBuilder* builder, IRInst* inOriginalDiffPairType, IRInst* inPrimalDiffPairType)
```
csyonghe, line 178: Yong's note: this code is duplicated with the samely named functions in `DifferentiableTypeConformanceContext`, so removing and bottleneck everything into the same function.

source/slang/slang-ir-autodiff-transcriber-base.cpp
```diff
@@ -486,15 +316,20 @@ IRType* AutoDiffTranscriberBase::_differentiateTypeImpl(IRBuilder* builder, IRTy
     }
 
     case kIROp_TupleType:
+    case kIROp_TypePack:
```
csyonghe, line 319: Yong's note: here we implement the IR-level differenetiation rules for tuple types and type packs.

source/slang/slang-ir-autodiff.cpp
```diff
@@ -963,6 +1009,104 @@ IRInst* DifferentiableTypeConformanceContext::getArrayWitness(IRBuilder* builder
     return table;
 }
 
+IRInst* DifferentiableTypeConformanceContext::getTupleWitness(IRBuilder* builder, IRInst* inTupleType)
```
csyonghe, line 1012: Yong's note: here we implement the IDifferentiable witness synthesis logic for Tuple types. Basically we need to create a witness table for a tuple type that defines how a tuple type satisfies `IDifferentiable`, such as its `Differential` type and its `dzero`/`dadd` methods.

source/slang/slang-ir-specialize.cpp
```diff
@@ -589,9 +590,6 @@ struct SpecializationContext
         case kIROp_Expand:
             return maybeSpecializeExpand(as<IRExpand>(inst));
 
-        case kIROp_ExpandTypeOrVal:
```
csyonghe, line 592: Yong's note: since we are now running the `lowerExpandType` pass after IR lowering, we will no longer need to deal with `ExpandTypeOrVal` inst here. The current specialization logic to deal with expand types is not robust so we delete it.

source/slang/slang-ir-specialize.cpp
```diff
@@ -605,6 +603,15 @@ struct SpecializationContext
 
         case kIROp_CountOf:
             return maybeSpecializeCountOf(inst);
+
+        case kIROp_Func:
+
+            if (tryExpandParameterPack(as<IRFunc>(inst)))
```
csyonghe, line 609: Yong's note: in this new specialization logic, if we see a function that takes a TypePack parameter, we are just going to expand that type pack out of existence. For example, if we have `int f(TypePack<int, float> v)`, we are going to turn it into `int f(int v0, float v1)`. This will make the final emitted code much cleaner, and is the easiest way to make interface method with variadic parameters work so we don't have to worry about packing ordianry arguments into packs.

source/slang/slang-ir-specialize.cpp
```diff
@@ -1010,6 +1017,9 @@ struct SpecializationContext
                     workList.removeLast();
                     workListSet.remove(inst);
 
+                    if (!inst->getParent() && inst->getOp() != kIROp_Module)
```
csyonghe, line 1020: Yong's note: we can run into instructions that we previously add to work list but now have been removed. If the inst has been removed (we check this by seeing if the inst still has a parent), there is no need to process it anymore.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2440,44 +2474,57 @@ struct SpecializationContext
...
+        
+        bool isMultiBlock = as<IRYield>(expandInst->getFirstBlock()->getTerminator()) == nullptr;
 
         for (UInt i = 0; i < elementCount; i++)
         {
             IRCloneEnv cloneEnv;
-            IRBlock* firstBlock = nullptr;
             IRBuilder subBuilder = builder;
-            for (auto childBlock : expandInst->getBlocks())
+            IRBlock* mergeBlock = nullptr;
```
csyonghe, line 2490: Yong's note: here we extend the existing `specailizeExpand` logic so that it can work with `IRExpand` insts that about types in additional to just values. We need this because we just added a pass that turns all `expand` types into `IRExpand` insts as well.

source/slang/slang-ir-specialize.cpp
```diff
@@ -2440,44 +2474,57 @@ struct SpecializationContext
...
+            if (isMultiBlock)
             {
-                auto newBlock = subBuilder.emitBlock();
-                if (!firstBlock)
-                    firstBlock = newBlock;
-                cloneEnv.mapOldValToNew[childBlock] = newBlock;
+                IRBlock* firstBlock = nullptr;    
+                for (auto childBlock : expandInst->getBlocks())
+                {
+                    auto newBlock = subBuilder.emitBlock();
```
csyonghe, line 2496: Yong's note: we want to make sure that if the body of `expand` just contains one block, we can just unroll the expand directly into the enclosing block instead of creating one new block for each unrollled iteration. This simplifies the resulting code, and is necessary for types because types may be used in a place where blocks aren't allowed to exist.

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -2429,25 +2455,38 @@ struct DiffTransposePass
...
         }
+        else if (as<IRTupleType>(baseType) || as<IRTypePack>(baseType))
+        {
+            isTupleType = true;
+            for (UInt i = 0; i < baseType->getOperandCount(); i++)
+            {
+                elementTypes.add((IRType*)baseType->getOperand(i));
+                primalElementTypes.add((IRType*)(aggPrimalType->getOperand(i)));
+            }
+            elementCount = baseType->getOperandCount();
```
jkwak-work, line 2488: Nitpick but elementCount can be set earlier and used in the for-loop.
```
            elementCount = baseType->getOperandCount();
            for (UInt i = 0; i < elementCount ; i++)
            {
                elementTypes.add((IRType*)baseType->getOperand(i));
                primalElementTypes.add((IRType*)(aggPrimalType->getOperand(i)));
            }
```

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -2429,25 +2455,38 @@ struct DiffTransposePass
...
             IRInst* elementCountInst = vectorType->getElementCount();
-            elementType = vectorType->getElementType();
-            primalElementType = as<IRVectorType>(aggPrimalType)->getElementType();
-            SLANG_ASSERT(as<IRIntLit>(elementCountInst));
             elementCount = as<IRIntLit>(elementCountInst)->getValue();
+            for (IRIntegerValue i = 0; i < elementCount; i++)
+            {
+                elementTypes.add(vectorType->getElementType());
+                primalElementTypes.add(as<IRVectorType>(aggPrimalType)->getElementType());
+            }
```
jkwak-work, line 2470: We may want to reserve the space for `elementTypes` with `elementCount`.
It will be a good idea to prepare for an extreme user shader that may use many elements in a tuple.

source/slang/slang-ir-autodiff-transpose.h
```diff
@@ -2493,11 +2536,17 @@ struct DiffTransposePass
...
                             builder->getIntValue(
                                 builder->getIntType(),
                                 sourceIndex)));
+                else if (isTupleType)
+                    accGrad((UIndex)targetIndex,
+                        builder->emitGetTupleElement(
+                            elementTypes[(UIndex)targetIndex],
+                            gradient.revGradInst,
+                            (UInt)sourceIndex));
                 // Case 3: Swizzled input is a scalar.
```
jkwak-work, line 2550: You may want to update this comment by making it Case 4.

source/slang/slang-ir-lower-expand-type.cpp
```diff
@@ -0,0 +1,167 @@
...
+        return newVal;
+    }
+
+    IRInst* clonePatternVal(IRCloneEnv& cloneEnv, IRBuilder* builder, IRInst* val, IRInst* eachIndex)
+    {
+        if (auto clonedVal = cloneEnv.mapOldValToNew.tryGetValue(val))
+            return *clonedVal;
+        cloneEnv.mapOldValToNew[val] = val;
+        auto result = clonePatternValImpl(cloneEnv, builder, val, eachIndex);
+        cloneEnv.mapOldValToNew[val] = result;
```
jkwak-work, line 69: It seems like the first assignment can be removed?
```
cloneEnv.mapOldValToNew[val] = val;
```
I am also wondering what would happen if `val` is `nullptr`.

tests/language-feature/ifunc/ifunc.slang
```diff
@@ -0,0 +1,40 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-dx12 -use-dxil -shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-cpu -shaderobj -output-using-type
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -shaderobj -output-using-type
+
+struct Functor : IFunc<int, int, bool>
+{
+    int __call(int p, bool t)
```
jkwak-work, line 8: I am wondering if we can have `out` or `inout` modifiers on the parameters of `__call`

jkwak-work (approved):
Looks good to me.

tests/language-feature/ifunc/ifunc.slang
csyonghe: That is another restriction of variadic generics, we can't current have them.
It is easy to support `inout` on the whole parameter pack so that all parameters are `inout`, but more difficult to allow each individual parameters to have different directions, because we don't encode `inout` as a type of parameter, but rather a modifier on the parameter itself. Since `inout` is not a part of the type, it cannot be expressed in the `IFunc<...>` interface.

source/slang/slang-ir-lower-expand-type.cpp
csyonghe: This assignment is to prevent stack overflow when trying to recurse into itself, if we ever have a cyclic reference from `val` to itself.

source/slang/core.meta.slang
```diff
@@ -953,6 +953,30 @@ extension Tuple<T> : IComparable
...
+interface IMutatingFunc<TR, each TP>
+{
+    [mutating]
+    TR __call(expand each TP p);
+}
+
+interface IFunc<TR, each TP> : IMutatingFunc<TR, expand each TP>
+{
+    TR __call(expand each TP p);
+}
```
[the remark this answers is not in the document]
csyonghe: Still need to decide on the actual syntax, `operator()` is another option to stay consistent with c++.

Issue #4787, opened 2024-08-06
Reflection API extensions: Support Namespaces


saipraveenb25:
`namespace` containers show up as 'unsupported'. We should add a new decl-kind to resolve this.

Issue #4800, opened 2024-08-06
Reflection API extension: Add endpoint to create specialized references to methods and types using a list of vals


saipraveenb25:
We can currently create a specialized ref through string-based lookups like "MyType<float, 3>", but it is important to be able to construct specializations by providing the base decl and a set of args.

PR #4909, opened 2024-08-23
Add ability to specialize generic references to functions, types and more

saipraveenb25:
This PR adds the following end-points:

- `specializeGeneric()` method that can be called on a generic reflection to substitute arguments for generic type and value parameters. It returns another generic reflection, but this time with the appropriate substitution.
- `applySpecializations()` method to then copy these specializations onto an existing type or function reflection.
- `isSubType()` to check if a type is a subtype of another type (useful to check if a type is differentiable by checking `IDifferentiable`)

This PR also:
 - Adds `DeclReflection::Kind::Namespace` so that namespace containers are correctly reflected when walking the decl-tree. the name can be obtained through `getName()` but there's no need to cast to a namespace (since there's nothing else we can do with a namespace decl)
 - Fixes an issue with name-based lookups that fail if a type or function is referenced without specializations. Its helpful to be able to form a reference to a function with default substitutions, so that we can we can specialize it later (either directly, or via argument types).

Fixes: #4787, #4800


include/slang.h
```diff
@@ -2686,12 +2722,25 @@ extern "C"
...
+    
+    SLANG_API SlangReflectionGeneric* spReflection_specializeGeneric(
+        SlangReflection*                 inProgramLayout,
+        SlangReflectionGeneric*          generic,
+        SlangInt                         argCount,
+        SlangReflectionVariable* const*  params,
+        SlangReflectionGenericArg const* args,
+        ISlangBlob**                     outDiagnostics);
+    
+    SLANG_API bool spReflection_IsSubType(
```
csyonghe, line 2740: isSubType

include/slang.h
```diff
@@ -3498,6 +3579,104 @@ namespace slang
...
+        {
+            return spReflectionGeneric_GetConcreteIntVal((SlangReflectionGeneric*)this, (SlangReflectionVariable*)valueParam);
+        }
+
+        GenericReflection* applySpecializations(GenericReflection* generic)
+        {
+            return (GenericReflection*)spReflectionGeneric_applySpecializations((SlangReflectionGeneric*)this, (SlangReflectionGeneric*)generic);
+        }
+
+        /*GenericReflection* specializeGenericTypeParameter(
```
csyonghe, line 3667: delete?

source/slang/slang-check-decl.cpp
```diff
@@ -7119,6 +7119,99 @@ namespace Slang
 
         return args;
     }
+    
+    SubstitutionSet makeSubstitutionFromIncompleteSet(
```
csyonghe, line 7123: You can call `SemanticsVisitor::inferGenericArguments`, which implement the same logic. Let's try not duplicate the logic here.

saipraveenb25: I think this specific function is different from `inferGenericArguments`. We want to use it for `specializeGeneric` which operates on struct types as well as callables, where we are provided specific argument substitutions for the generic parameters directly (not the arg types of the callables).

I plan on providing a separate function `FunctionReflection::specializeWithArgTypes(...)` that will invoke `inferGenericArguments` directly.

Alternatively, I can change up `inferGenericArguments` to also support struct types. With generic interfaces, we probably need to enforce constraints across the arguments of the same generic.

csyonghe: `inferGenericArguments` is not only for function calls but for types as well.

Basically if you have `S<T:IFoo, int N =  4>` and you pass in `S<FooImpl>`, `inferGenericArguments` will give you `S<FooImpl, 4, FooImpl : IFoo>`.

csyonghe: I guess for here, you wan't to reuse or factor out `checkGenericAppWithCheckedArgs`.

saipraveenb25: Rerote this method with `checkGenericAppWithCheckedArgs` + added support for bool type args.

I think this is a good suggestion, since the new version should work seamlessly with variadics.

csyonghe: We can delete this now?

source/slang/slang.cpp
```diff
@@ -2373,9 +2410,28 @@ DeclRef<Decl> ComponentType::findDeclFromStringInType(
...
+            astBuilder, &visitor, DeclRef(genericDeclRef.getDecl()->inner));
+        result = substituteDeclRef(SubstitutionSet(genericDeclRef), astBuilder, result);
+    }
+
     return result;
 }
 
+bool ComponentType::isSubType(Type* subType, Type* superType)
+{
+    SharedSemanticsContext sharedSemanticsContext(
```
csyonghe, line 2425: At somepoint it may make sense to hold a semantic visitor on the linkage, because it will become extensive to rebuild the type inheritance list for all referenced types everytime `isSubType` is called. These are cached on the semantic visitor and we should probably reuse it to speed up these reflection queries.

csyonghe (commented):
Looks good, but we need to remove `makeSubstitutionFromIncompleteSet`.

PR #4910, opened 2024-08-23
Feature/record unit test

kaizhangNV:
Integrate the record/replay test into slang-unit-test.

We use the slang-unit-test infrastructure to run the examples to test the record/replay layer.

The unit test will run the modified example as a new process, and read the stdout from the process, then it will launch the `slang-replay` as another process, and read the stdout from that process. Then parse the hash code from the stdout and compare to check whether the test passes.

The unit test can be run as
` ./build/Debug/bin/slang-test slang-unit-test-tool/RecordReplay -category unit-test`

Fix the slang-test bug
    
Since we reorganize the build directory, now the libraries are
located at different directory with executables in non-Windows
platform, we have to change the code on how to find the dll directory.

### DIFF
--- a/tools/slang-unit-test/unit-test-record-replay.cpp
+++ b/tools/slang-unit-test/unit-test-record-replay.cpp
@@ -1,0 +1,308 @@
+// unit-test-record-replay.cpp
+
+#include "../../source/core/slang-string-util.h"
+#include "../../source/core/slang-process-util.h"
+
+#include "../../source/core/slang-io.h"
+#include "../../source/core/slang-http.h"
+#include "../../source/core/slang-random-generator.h"
+
+#include "tools/unit-test/slang-unit-test.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#include <shellapi.h>
+#endif
+
+using namespace Slang;
+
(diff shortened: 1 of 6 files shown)


kaizhangNV:
There are some failures that cannot reproduce locally, so I need to debug it by uploading debug commits.
Change this to draft now.

tools/slang-unit-test/unit-test-record-replay.cpp
```diff
@@ -0,0 +1,437 @@
...
+    {
+        return SLANG_OK;
+    }
+
+    StringBuilder msgBuilder;
+    // Path::remove() doesn't support remove a non-empty directory, so we need to implement
+    // a simple function to remove the directory recursively.
+#ifdef _WIN32
+    // https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shfileoperationa
+    SHFILEOPSTRUCTA file_op = {
```
csyonghe, line 340: Might be good to put this logic inside `Path::remove`.

csyonghe (approved):

tools/slang-unit-test/unit-test-record-replay.cpp
kaizhangNV: Sure, let me create a separate small issue for this. (#4916 )

PR #4911, opened 2024-08-23
Ignoring construct field warnings on delegatory methods

### LINKED ISSUE #4872: Compilation Warnings Related to Uninitialized Struct Fields
**Issue Summary:**
The provided code produces the following compilation warnings:

```
warning 41020: exiting constructor without initializing field 'barycentrics'
warning 41020: exiting constructor without initializing field 'primitiveIndex'
```

These warnings occur in the struct `HitInfo`, which has two fields: `float3 barycentrics` and `uint primitiveIndex`. The warnings indicate that the constructors (`__init` functions) may not be properly initializing these fields.

Code Snippet:
```cpp
struct HitInfo
{
    float3 barycentrics;
    uint primitiveIndex;
    
    [[mutating]] void init(float2 hitBarycentrics, uint hitPrimitiveIndex)
    {
        barycentrics = { 1.0 - hitBarycentrics.x - hitBarycentrics.y, hitBarycentrics.x, hitBarycentrics.y };
        primitiveIndex = hitPrimitiveIndex;
    }

    __init(float2 hitBarycentrics, uint hitPrimitiveIndex)
    {
        init(hitBarycentrics, hitPrimitiveIndex);
    }

    __init(BuiltInTriangleIntersectionAttributes attr)
    {
        init(attr.barycentrics, PrimitiveIndex());
    }
}
```

**Root Cause:**
The warnings are triggered because the compiler may not recognize that the `init` function, which is a `[[mutating]]` method, properly initializes the struct members `barycentrics` and `primitiveIndex`. Since these members are not explicitly initialized within the `__init` functions themselves, the compiler assumes they might remain uninitialized, hence the warnings.

**Suggestion for Improvement:**
Would it be possible to enhance compiler to check whether `[[mutating]]` functions called within constructors are initializing all the struct members?

venkataram-nv:
Fixes #4872

### DIFF
--- a/source/slang/slang-ir-use-uninitialized-values.cpp
+++ b/source/slang/slang-ir-use-uninitialized-values.cpp
@@ -441,17 +441,12 @@ namespace Slang
 
     static bool isWrittenTo(IRInst* inst)
     {
-        List<IRInst*> stores;
-        List<IRInst*> loads;
-
         for (auto alias : getAliasableInstructions(inst))
         {
             for (auto use = alias->firstUse; use; use = use->nextUse)
             {
-                collectInstructionByUsage(stores, loads, use->getUser(), alias);
-
-                // ...we will ignore the rest...
-                if (stores.getCount())
+                InstructionUsageType usage = getInstructionUsageType(use->getUser(), alias);
+                if (usage == Store || usage == StoreParent)
                     return true;
             }
         }
@@ -461,16 +456,11 @@ namespace Slang
 
     static bool isDirectlyWrittenTo(IRInst* inst)
     {
-        List<IRInst*> stores;
-        List<IRInst*> loads;
-
         for (auto use = inst->firstUse; use; use = use->nextUse)
         {
             IRInst* user = use->getUser();
-            collectInstructionByUsage(stores, loads, use->getUser(), inst);
-
-            // ...we will ignore the rest...
-            if (stores.getCount())
+            InstructionUsageType usage = getInstructionUsageType(use->getUser(), inst);
+            if (usage == Store || usage == StoreParent)
                 return true;
         }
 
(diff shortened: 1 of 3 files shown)


tests/diagnostics/uninitialized-fields.slang
```diff
@@ -104,5 +104,28 @@ struct Impl
     }
 }
 
+// Calling a method from a constructor to initialize fields
```
csyonghe, line 107: Put this into a separate test. `//CHECK-NOT` will succeed if the line isn't found from the last succesful match. If the message exists early before other `//CHECK` comments, this test will still suceed. Having this in its own test is the only certain way to make sure the warning doesn't exist.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -434,13 +434,56 @@ namespace Slang
...
+
+    static bool isDirectlyWrittenTo(IRInst* inst)
+    {
+        List<IRInst*> stores;
+        List<IRInst*> loads;
+
+        for (auto use = inst->firstUse; use; use = use->nextUse)
+        {
+            IRInst* user = use->getUser();
+            collectLoadStore(stores, loads, user, inst);
```
csyonghe, line 466: This seems very inefficient. All we care about here is whether there are any stores, we don't want to collect all load and stores just to do the check.

venkataram-nv: I've reworked the collection system so that now there are methods to check the usage type of an instruction wrt an operand. With this we can skip the collection and do a pure check.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -214,8 +214,16 @@ namespace Slang
...
+
+    enum InstructionUsageType
+    {
+        None,           // Instruction neither stores nor loads from the soruce (e.g. meta operations)
+        Store,          // Instruction acts as a write to the source
+        StoreParent,    // Instruction's parent acts as a write to the source
+        Load            // Instruciton acts as a load from the source
+    };
+
+    static InstructionUsageType getCallUsageType(IRCall* call, IRInst* inst)
```
csyonghe, line 226: The logic inside this function for handling backward differentiate isn't correct. The inner function's type is usually not the same as the function type of the differentiated function. I suggest we don't check for any autodiff cases here and just assume that if we can't obtain the func type, then whenever a pointer typed argument is passed in a call, it is inout.

csyonghe: If you are not sure if certain logic is correct, make sure to call it out in the comment or the PR description so we don't get incorrect logic slipped through the review process.

venkataram-nv: The diagnostics currently ignore functions marked as differentiable, so this shouldn't be an issue:
```cpp
// for example in checkUninitializedValues
if (isDifferentiableFunc(func))
            return;
```
The reason they are ignored is because at the moment it is not possible to distinguish functions marked as differentiable to be something that the user wrote or that slang synthesized automatically.

csyonghe: If Slang is synthesizing code that leads to warnings, we should fix those synthesis logic. Nevertheless, the logic here is incorrect and we should remove them and replace with an SLANG_UNEXPECTED failure.

csyonghe (approved):

Issue #4126, opened 2024-05-07
Support `-fvk-use-dx-layout`.


csyonghe:
We already have `-force-glsl-scalar-layout` that causes all buffers to follow the scalar layout in the resulting spirv. We also need `-fvk-use-dx-layout` to make them follow the dx layout.

csyonghe:
From Hai:

Here's all the resources that I could scrounge up for this.

For DXC, the selection for the layout rules begins here based on what options were passed in:
[https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L605-L609](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2FDirectXShaderCompiler%2Fblob%2Fmain%2Ftools%2Fclang%2Flib%2FSPIRV%2FSpirvEmitter.cpp%23L605-L609&data=05%7C02%7Cyhe%40nvidia.com%7C1abcdaf37e554073f1da08dc6ee38239%7C43083d15727340c1b7db39efd9ccc17a%7C0%7C0%7C638507170692078715%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=W5vTLZW%2BcJJyPn0vJ3dwRGNJqJWu0%2FRZePtyMJKfbX8%3D&reserved=0)

Looking through this reminded me it was actually the FXC rules that we had to follow, so that's why the enums are prefixed with Fxc.
cbuffer/ConstantBuffer/tbuffers all use the FxcCTBuffer rules.
StructuredBuffer/RWStructuredBuffer + all their variants use the FxcSBuffer rules.
Amplification shader payloads also use FxcSBuffer rules.

The rules (FxcCTBuffer and FxcSBuffer) are spelled out and implemented in this file:
[...]

PR #4912, opened 2024-08-23
Implement `-fvk-use-dx-layout`

ArielG-NV:
Fixes: #4126

Changes:
* Added fvk-use-dx-layout
* Modified `HLSLConstantBufferLayoutRulesImpl` for correctness (ex: Array is always 16 byte aligned)
* Added kFXCShaderResourceLayoutRulesFamilyImpl and kFXCConstantBufferLayoutRulesFamilyImpl to handle fvk-use-dx-layout
* Added `ConstantBufferLayoutRules` to manage constant buffer rules
* Added `alignCompositeElementOfNonAggregate`/`alignCompositeElementOfAggregate` to handle forced alignment of composites for ConstantBuffers
* `StructuredBuffer` rules are mostly equal to `scalar` layout, not much was needed to be changed to support this behavior.


tests/spirv/cbuffer-dx-layout-1.slang
```diff
@@ -0,0 +1,74 @@
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-vk -compute -xslang -fvk-use-dx-layout
+//TEST:SIMPLE(filecheck=SPIRV): -target spirv -entry computeMain -stage compute -fvk-use-dx-layout
+//TEST_INPUT:cbuffer(data=[1 2.0 3.0 4.0   5 6 7 0   8 9 10 11   12 0 0 0  13 0 0 0   14 0 0 0   15 0 0 0   16 17 18 0   19 20 21 22   23 24 0 0   25 0 0 0   26 0 0 0]):name=Test
+
+cbuffer Test
```
csyonghe, line 5: Let's verify the layout with DXC's output to be sure it is correct.

source/slang/slang-ir.h
```diff
@@ -564,6 +564,7 @@ enum class IRTypeLayoutRuleName
     Scalar = Natural,
     Std430,
     Std140,
+    ConstantBuffer,
```
csyonghe, line 567: Is this LegacyD3DConstantBuffer ?

csyonghe (approved):

source/slang/slang-ir.h
ArielG-NV: Yes, `ConstantBuffer` meant the `cbuffer` block `ConstantBuffer`.

tests/spirv/cbuffer-dx-layout-1.slang
ArielG-NV: The output is the same, if it is preferred, I can add a `spirv-via-hlsl`/CROSS_SPIRV_HLSL test option so we can add a test using DXC directly.

csyonghe:
Don't forget to update the documentation on command line /compiler options.

tests/spirv/cbuffer-dx-layout-1.slang
csyonghe: That isn't necessary. As long as it is verified to be correct it is all good.

source/slang/slang-ir.h
csyonghe: Let's name it D3DConstantBuffer to be precise.

tests/spirv/cbuffer-dx-layout-1.slang
ArielG-NV: All layouts work as DXC does for SPIRV.
Attaching one of the 'framework' HLSL programs I used to test with for ease of future testing:
```
cbuffer Test
{
//SPIRV: Offset 0
    uint v0;
//SPIRV: Offset 4
    float3 v1;

//SPIRV: Offset 16
    uint3 v2;

//SPIRV: Offset 32
    uint2 v3;
//SPIRV: Offset 40
    uint2 v4;

//SPIRV: Offset 48
    uint v5[4];

// array always starts on a new register.
//SPIRV: Offset 112
    uint3 v6[2];
//SPIRV: Offset 140
// non-array can pack with a partially filled register
    uint v7;

//SPIRV: Offset 144
    uint2 v8;

// SPIRV: Offset 160
// array always starts on a new register.
    uint v9[2];
}

struct PSInput
{
	float4 color : COLOR;
};

float4 PSMain(PSInput input) : SV_TARGET
{
    float4 res;
    res.x+= v0;
    res.x+= v9[0];
	return res;
}
```

source/slang/slang-type-layout.cpp
```diff
@@ -295,6 +295,92 @@ struct Std430LayoutRulesImpl : GLSLBaseLayoutRulesImpl
...
 };
 
+/// GLSL fvk-use-dx-layout for `ShaderResource`
+struct FXCShaderResourceLayoutRulesImpl : DefaultLayoutRulesImpl
+{
+    // Currently this FXC layout is equal to how we compute 'DefaultLayoutRulesImpl'
+};
+
+/// GLSL fvk-use-dx-layout for `ConstantBuffer`
+struct FXCConstantBufferLayoutRulesImpl : DefaultLayoutRulesImpl
```
csyonghe, line 305: Why can't this be just HLSLConstantBufferLayoutRules?

source/slang/slang-ir-layout.cpp
```diff
@@ -193,7 +193,7 @@ case kIROp_##TYPE##Type:                                        \
                 offset = rules->adjustOffsetForNextAggregateMember(offset, fieldTypeLayout.alignment);
             }
         }
-        *outSizeAndAlignment = rules->alignCompositeElement(structLayout);
+        *outSizeAndAlignment = rules->alignCompositeElementOfAggregate(structLayout);
```
csyonghe, line 196: I don't quite understand the need for alignCompositeElementOfAggregate here. Why can't we just rely on adjustOffsetForNextAggregateMember?

source/slang/slang-ir-layout.cpp
```diff
@@ -250,7 +250,7 @@ case kIROp_##TYPE##Type:                                        \
             resultLayout.size = align(resultLayout.size, fieldTypeLayout.alignment);
             resultLayout.alignment = std::max(resultLayout.alignment, fieldTypeLayout.alignment);
         }
-        *outSizeAndAlignment = rules->alignCompositeElement(resultLayout);
+        *outSizeAndAlignment = rules->alignCompositeElementOfNonAggregate(resultLayout);
```
csyonghe, line 253: Tuple type is the same as struct type so this needs to be consistent with that case.

source/slang/slang-ir-layout.cpp
```diff
@@ -234,7 +234,7 @@ case kIROp_##TYPE##Type:                                        \
         auto anyValType = cast<IRAnyValueType>(type);
         outSizeAndAlignment->size = getIntVal(anyValType->getSize());
         outSizeAndAlignment->alignment = 4;
-        *outSizeAndAlignment = rules->alignCompositeElement(*outSizeAndAlignment);
+        *outSizeAndAlignment = rules->alignCompositeElementOfNonAggregate(*outSizeAndAlignment);
```
csyonghe, line 237: This eventually is lowered into a struct so it need to be consistent with struct type case.

source/slang/slang-ir-layout.cpp
```diff
@@ -272,7 +272,7 @@ case kIROp_##TYPE##Type:                                        \
         IRSizeAndAlignment resultLayout;
         resultLayout.size = size;
         resultLayout.alignment = 4;
-        *outSizeAndAlignment = rules->alignCompositeElement(resultLayout);
+        *outSizeAndAlignment = rules->alignCompositeElementOfNonAggregate(resultLayout);
```
csyonghe, line 275: Eventually a struct so need to be consistent with struct type case.

source/slang/slang-type-layout.cpp
ArielG-NV: We can use `HLSLConstantBufferLayoutRules`.

source/slang/slang-ir-layout.cpp
ArielG-NV: `ConstantBuffer`s requires an alignment of 16 (cannot use partial registers) on all structs, arrays & matrixes (treated like an array). This functionality is equal to how `alignCompositeElement(resultLayout);` functioned and was intended to be used. The issue that we had was that `alignCompositeElement` considers `vector` as the same as `struct` for alignment rules (ConstantBuffer treats these types differently).

csyonghe: Does that mean the current d3d constant buffer layout implementation is incorrect?

csyonghe: Isn't the difference of vector types handled by getVectorSizeAndAlignment?

ArielG-NV: > Isn't the difference of vector types handled by getVectorSizeAndAlignment?

How the old code worked was: `getVectorSizeAndAlignment` gets the vector size (and alignment), `alignCompositeElement` did alignment logic (since std140/std430 computes 'vector size' and 'alignment' differently)

ArielG-NV: > Does that mean the current d3d constant buffer layout implementation is incorrect?

I believe our current implementation does not work as intended (array, matrix, and struct).

Example:
Given the following struct (you may copy paste this struct into https://maraneshi.github.io/HLSL-ConstantBufferLayoutVisualizer/):
```
cbuffer Test
{
    uint v0;
    float3 v1;
    uint3 v2;
    uint2 v3;
    uint2 v4;
    uint v5[4];
    uint3 v6[2];
    uint v7;
    uint2 v8;
    uint v9[2];
};
```

This struct currently (without this PR with `-fvk-use-dx-layout`) will compile into:
```
OpMemberName %SLANG_ParameterGroup_Test_std140 0 "v0"
OpMemberName %SLANG_ParameterGroup_Test_std140 1 "v1"
OpMemberName %SLANG_ParameterGroup_Test_std140 2 "v2"
OpMemberName %SLANG_ParameterGroup_Test_std140 3 "v3"
OpMemberName %SLANG_ParameterGroup_Test_std140 4 "v4"
OpMemberName %SLANG_ParameterGroup_Test_std140 5 "v5"
OpMemberName %SLANG_ParameterGroup_Test_std140 6 "v6"
OpMemberName %SLANG_ParameterGroup_Test_std140 7 "v7"
OpMemberName %SLANG_ParameterGroup_Test_std140 8 "v8"
OpMemberName %SLANG_ParameterGroup_Test_std140 9 "v9"

OpMemberDecorate %SLANG_ParameterGroup_Test_std140 0 Offset 0
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 1 Offset 16
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 2 Offset 32
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 3 Offset 48
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 4 Offset 56
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 5 Offset 64
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 6 Offset 128
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 7 Offset 160
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 8 Offset 168
OpMemberDecorate %SLANG_ParameterGroup_Test_std140 9 Offset 176
```
Which from what I can tell is wrong as per https://maraneshi.github.io/HLSL-ConstantBufferLayoutVisualizer/ and Shader-Playground.

With this PR and `fvk-use-dx-layout` the SPIR-V offsets are now correctly:
```
OpMemberName %SLANG_ParameterGroup_Test_std140 0 "v0"
OpMemberName %SLANG_ParameterGroup_Test_std140 1 "v1"
OpMemberName %SLANG_ParameterGroup_Test_std140 2 "v2"
OpMemberName %SLANG_ParameterGroup_Test_std140 3 "v3"
OpMemberName %SLANG_ParameterGroup_Test_std140 4 "v4"
OpMemberName %SLANG_ParameterGroup_Test_std140 5 "v5"
OpMemberName %SLANG_ParameterGroup_Test_std140 6 "v6"
OpMemberName %SLANG_ParameterGroup_Test_std140 7 "v7"
OpMemberName %SLANG_ParameterGroup_Test_std140 8 "v8"
OpMemberName %SLANG_ParameterGroup_Test_std140 9 "v9"

OpMemberDecorate %SLANG_ParameterGroup_Test_default 0 Offset 0
OpMemberDecorate %SLANG_ParameterGroup_Test_default 1 Offset 4
OpMemberDecorate %SLANG_ParameterGroup_Test_default 2 Offset 16
OpMemberDecorate %SLANG_ParameterGroup_Test_default 3 Offset 32
OpMemberDecorate %SLANG_ParameterGroup_Test_default 4 Offset 40
OpMemberDecorate %SLANG_ParameterGroup_Test_default 5 Offset 48
OpMemberDecorate %SLANG_ParameterGroup_Test_default 6 Offset 112
OpMemberDecorate %SLANG_ParameterGroup_Test_default 7 Offset 140
OpMemberDecorate %SLANG_ParameterGroup_Test_default 8 Offset 144
OpMemberDecorate %SLANG_ParameterGroup_Test_default 9 Offset 160
```

csyonghe: Well, you are running spirv compiles which will never trigger the existing HLSL constant buffer layout logic. You will need to use reflection test and compile it to d3d target to see if the layout computed by slang is correct. In that case the existing HLSL constant buffer layout code will be executed. If you just compile to spirv you are just testing the std140 logic, as indicated in the spirv you pasted.

ArielG-NV: HLSL has correct reflection data.

ArielG-NV: Reflection is based on AST, AST has correct data.

What happened was that the SizeAndAlignment setup (IR-pass) was producing the wrong values for HLSL. Since we don't care about size/alignment for HLSL emitting this was not an issue

source/slang/slang-ir-layout.cpp
```diff
@@ -443,6 +443,38 @@ struct NaturalLayoutRules : IRTypeLayoutRules
...
 };
 
+struct ConstantBufferLayoutRules : IRTypeLayoutRules
+{
+    ConstantBufferLayoutRules()
+    {
+        ruleName = IRTypeLayoutRuleName::D3DConstantBuffer;
+    }
+
+    virtual IRSizeAndAlignment alignCompositeElement(IRSizeAndAlignment elementSize)
```
csyonghe, line 453: Is `alignCompositeElement` ever used? If not we should just change its behavior in-place wihtout adding another variant.

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -108,7 +108,7 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
         auto elementType = inst->getElementType();
         IRSizeAndAlignment elementSize;
         getSizeAndAlignment(m_sharedContext->m_targetProgram->getOptionSet(), layoutRules, elementType, &elementSize);
-        elementSize = layoutRules->alignCompositeElement(elementSize);
+        elementSize = layoutRules->alignCompositeElementOfNonAggregate(elementSize);
```
csyonghe, line 111: why is this nonAggregate?

csyonghe: If this is `nonAggregate`, then `_calcArraySizeAndAlignment` should also call `alignCompositeElementOfNonAggregate`?

source/slang/slang-ir-layout.h
```diff
@@ -56,14 +56,25 @@ struct IRSizeAndAlignment
 
 struct IRTypeLayoutRules
 {
+private:
+    virtual IRSizeAndAlignment alignCompositeElement(IRSizeAndAlignment elementSize) = 0;
+
 public:
     IRTypeLayoutRuleName ruleName;
-    virtual IRSizeAndAlignment alignCompositeElement(IRSizeAndAlignment elementSize) = 0;
+    virtual IRSizeAndAlignment alignCompositeElementOfNonAggregate(IRSizeAndAlignment elementSize)
```
csyonghe, line 64: Need comments here explaining when this function is called, and what it attempts to achieve.

csyonghe (commented):
I feel like it is unecessarily complicated to have both `alignCompositeElementOfNonAggregate` and `alignCompositeElement`, when the only difference we care about is whether or not we are trying to layout a vector typed field in a struct. It appears to me that this should be a specific rule to struct types instead of a property of element types.

It makes more sense to me to keep `alignCompositeElement` as is, and add `adjustOffsetForNextVectorTypedAggregateMember` and call that in `_calcSizeAndAlignment`:

```
if (as<IRMatrixType>(field->getFieldType()) || as<IRArrayTypeBase>(field->getFieldType()) || as<IRStructType>(field->getFieldType()))
            {
                offset = rules->adjustOffsetForNextAggregateMember(offset, fieldTypeLayout.alignment);
            }
            else if (as<IRVectorType>(field->getFieldType())
            {
                offset = rules->adjustOffsetForNextVectorTypedAggregateMember(offset, fieldTypeLayout.alignment);
             }
```

source/slang/slang-ir-layout.cpp
ArielG-NV: The variant can be removed 👍

source/slang/slang-ir-spirv-legalize.cpp
ArielG-NV: Should be `alignCompositeElementOfAggregate` 👍

csyonghe (approved):

csyonghe:
This looks very clean now, I am really glad to see that the existing mechanism just work.

Issue #4913, opened 2024-08-26
GLSL/HLSL Code Emitter: `kIROp_ControlBarrier` is unimplemented


ArielG-NV:
Problem:
* If IR produces a `kIROp_ControlBarrier` Slang will assert when emitting GLSL/HLSL due to not implementing `kIROp_ControlBarrier`.

Repro:
* Run the following slang-test:
```
//TEST:SIMPLE(filecheck=GLSL):-target glsl -entry hullMain -stage hull -allow-glsl
//TEST:SIMPLE(filecheck=HLSL):-target hlsl -entry hullMain -stage hull -allow-glsl

//GLSL: location = 0
//GLSL-NOT: location = 0
//GLSL: location = 1
//GLSL-NOT: location = 1
//GLSL: location = 2

//SPIRV: location 0
//SPIRV-NOT: location 0
//SPIRV: location 1
//SPIRV-NOT: location 1
//SPIRV: location = 2

//SPIRV: location = 0
//HLSL: hullMain

struct HsOut
{
    float2 pos;
    float2 hm;
};

struct HscOut
{
    float EdgeTessFactor[4] : SV_TessFactor;
    float InsideTessFactor[2] : SV_InsideTessFactor;
    uint instanceId;
};

[domain("quad")]
[partitioning("integer")]
[outputtopology("triangle_ccw")]
[outputcontrolpoints(4)]
[patchconstantfunc("constants")]
HsOut hullMain()
{
    HsOut o;
    o.pos = 1;
    o.hm = 2;
    return o;
}

HscOut constants()
{
    HscOut o;
    o.instanceId = 123;
    o.EdgeTessFactor[0] = 1;
    o.EdgeTessFactor[1] = 2;
    o.EdgeTessFactor[2] = 3;
    o.EdgeTessFactor[3] = 4;
    o.InsideTessFactor[0] = 0.5;
    o.InsideTessFactor[1] = 0.5;
    return o;
}
```

Issue #4540, opened 2024-07-03
GLSL/SPIRV: Conflicting slots in tessellation control shader output


n3b:
When custom data added to constant function output that is expected to be accessible in evaluation (domain) shader, the wrong location is generated.

Decompiled GLSL generated by SPIRV-Cross:
```
layout(location = 0) out vec2 entryPointParam_main_gridPosition[4];
layout(location = 1) out vec2 entryPointParam_main_heightmap[4];
layout(location = 0) out uint entryPointParam_constants_instanceId;
```

ArielG-NV:
It would be helpful if a shader that reproduces the problem described can be shared

n3b:
@ArielG-NV sorry about that, here is an example. 
The `HscOut::instanceId` is being bound to the same slot as the `HsOut::pos`.

```hlsl
struct HsOut
{
    float2 pos;
    float2 hm;
};

struct HscOut
{
    float EdgeTessFactor[4] : SV_TessFactor;
    float InsideTessFactor[2] : SV_InsideTessFactor;
    uint instanceId;
};

[domain("quad")]
[partitioning("integer")]
[outputtopology("triangle_ccw")]
[outputcontrolpoints(4)]
[patchconstantfunc("constants")]
HsOut main()
{
    HsOut o;
    o.pos = 1;
    o.hm = 2;
    return o;
}

HscOut constants()
{
    HscOut o;
    o.instanceId = 123;
    o.EdgeTessFactor[0] = 1;
    o.EdgeTessFactor[1] = 2;
    o.EdgeTessFactor[2] = 3;
    o.EdgeTessFactor[3] = 4;
    o.InsideTessFactor[0] = 0.5;
    o.InsideTessFactor[1] = 0.5;
    return o;
}
```

ArielG-NV:
Thank you for passing along an example shader.

Issue #4914, opened 2024-08-26
SPIR-V: Some hull shaders fail during emitting


ArielG-NV:
Problem: 
* Slang will fail when compiling certain hull shaders due to a left-over `kIROp_Param` for when emitting SPIR-V. This `kIROp_Param` is not necessarily wrong, spirv-legalization-passes should be handling the `kIROp_Param`.

Repro:
* Run the following slang-test
```
//TEST:SIMPLE(filecheck=SPIRV):-target spirv -entry hullMain -stage hull -allow-glsl

//GLSL: location = 0
//GLSL-NOT: location = 0
//GLSL: location = 1
//GLSL-NOT: location = 1
//GLSL: location = 2

//SPIRV: location 0
//SPIRV-NOT: location 0
//SPIRV: location 1
//SPIRV-NOT: location 1
//SPIRV: location = 2

//SPIRV: location = 0
//HLSL: hullMain

struct HsOut
{
    float2 pos;
    float2 hm;
};

struct HscOut
{
    float EdgeTessFactor[4] : SV_TessFactor;
    float InsideTessFactor[2] : SV_InsideTessFactor;
    uint instanceId;
};

[domain("quad")]
[partitioning("integer")]
[outputtopology("triangle_ccw")]
[outputcontrolpoints(4)]
[patchconstantfunc("constants")]
HsOut hullMain()
{
    HsOut o;
    o.pos = 1;
    o.hm = 2;
    return o;
}

HscOut constants()
{
    HscOut o;
    o.instanceId = 123;
    o.EdgeTessFactor[0] = 1;
    o.EdgeTessFactor[1] = 2;
    o.EdgeTessFactor[2] = 3;
    o.EdgeTessFactor[3] = 4;
    o.InsideTessFactor[0] = 0.5;
    o.InsideTessFactor[1] = 0.5;
    return o;
}
```

Info:
* The failing `kIROp_Param` is likely created during the following code
```
    auto constantOutputType = constantFunc->getResultType();
    IRTypeLayout* constantOutputLayout = createPatchConstantFuncResultTypeLayout(builder, constantOutputType);
    IRVarLayout::Builder resultVarLayoutBuilder(&builder, constantOutputLayout);
    if (auto semanticDecor = constantFunc->findDecoration<IRSemanticDecoration>())
        resultVarLayoutBuilder.setSystemValueSemantic(semanticDecor->getSemanticName(), 0);
```

projetosgppag:
Download 
[...]

PR #4915, opened 2024-08-26
Fix Varying Variable Location Assignments With Hull Shaders

ArielG-NV:
Fixes: #4913
Fixes: #4540

Changes:
1. Added `kIROp_ControlBarrier` to HLSL/GLSL emitting.
2. Added a method to track 'used' and 'unused' varying's for when legalizing GLSL. This allows us to assign correct offsets to automatically added varying's variables.
    * Added a `getLSBZero` logic to UIntSet for efficient LSBZero finding logic
    
Notes: 
* Due to #4914 test shaders do not compile to SPIR-V


source/slang/slang-emit-hlsl.cpp
```diff
@@ -500,6 +500,11 @@ bool HLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
 {
     switch (inst->getOp())
     {
+        case kIROp_ControlBarrier:
+        {
+            m_writer->emit("GroupMemoryBarrier();\n");
```
csyonghe, line 505: GroupMemoryBatrierWithGroupSync.

Otherwise this is just a memory barrier not a control barrier.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -259,11 +259,20 @@ List<IRInst*> ScalarizedVal::leafAddresses()
...
-    DiagnosticSink*         sink;
-    Stage                   stage;
-    IRFunc*                 entryPointFunc;
+    Session*                   session;
+    GLSLExtensionTracker*      glslExtensionTracker;
+    DiagnosticSink*            sink;
+    Stage                      stage;
+    IRFunc*                    entryPointFunc;
+    
+    //Dictionary<ResourceType, Dictionary<Space, BindingIndex>>
```
csyonghe, line 268: Add more description to the comment as it currently looks too much like a left over debug code.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -265,7 +265,8 @@ struct GLSLLegalizationContext
     Stage                      stage;
     IRFunc*                    entryPointFunc;
     
-    //Dictionary<ResourceType, Dictionary<Space, BindingIndex>>
+    /// This dictionary stores for 'VaryingIn/VaryingOut' ('Dictionary<LayoutResourceKind, ...>') all used
```
csyonghe, line 268: Why would a `varying in/out` has a `space` binding? They should only have a single location number, so it appears to me that `Dictionary<LayoutResourceKind, UIntSet>` is sufficient.

ArielG-NV: > Why would a varying in/out has a space binding? 

They should not have a space other than 0, although, I did not see anywhere that the GLSL legalization code ensures any of this

csyonghe: We can safely assume this, because there is no way in spirv/glsl to specify a "space" for a varying inout.

ArielG-NV: I will add this change👍

csyonghe (approved):

Issue #4924, opened 2024-08-27
Auto-diff texture example fails when using direct-to-SPIRV compilation


saipraveenb25:
Raising on behalf of @kaizhangNV

Reproduced with:
`./build/Debug/bin/slangc examples/autodiff-texture/train.slang -target spirv`

PR #4925, opened 2024-08-27
Avoid inlining functions with inline ASM blocks.

saipraveenb25:
Also removes `[__unsafeForceInlineEarly]` from built-in methods that have inline spir-v

Fixes: #4924


csyonghe (commented):
Also need to change RWByteAddressBuffer, RWStructuredBuffer, and all texture GetDimensions.

saipraveenb25:
@csyonghe: I think the change on line 5288 & 3959 should handle RWStructuredBuffer & RWByteAddressBuffer?

source/slang/hlsl.meta.slang
```diff
@@ -5285,7 +5285,7 @@ struct $(item.name)
     uint DecrementCounter();
 
     [__readNone]
-    [__unsafeForceInlineEarly]
+    [ForceInline]
```
saipraveenb25, line 5288: @csyonghe: I feel like I'm missing something.. 
`[ForceInline]` will work fine with auto-diff because we check for SPIRVAsm blocks.
It's only `[__unsafeForceInlineEarly]` that is a problem.

csyonghe: Ah, sorry, you did change the line, I didn't realize this is for RWStructuredBuffer.

csyonghe (approved):

PR #4926, opened 2024-08-27
Add functor syntax support.

csyonghe:
Partially address #4892.

This PR adds the syntax support for calling a functor directly with `operator ()`.


jkwak-work (approved):
Looks good to me.

Issue #4917, opened 2024-08-26
Capability System: Allow alias's to properly be used with `[shader("...")]` attribute logic


ArielG-NV:
Problem:
Currently if we add `[shader("raygeneration")]` in our code there is no-way for Slang to implicitly require raytracing extensions since Slang only allows 'stage' capabilities inside `[shader("...")]`.

The solution to this is to allow aliases with only 1 stage defined (for each target).

This allows us to add to our `capdef` file an alias to 'raygeneration' and 'raygen' such that `alias raygeneration = _raygeneration + raytracing_extensions_and_versions;`

Issue #4421, opened 2024-06-18
Capabilities System: Rework all uses of Profile to just be an alias/wrapper of a CapabilitySet


ArielG-NV:
Currently we separate `profile` and `capabilities`, where `profile`s are just wrapped `capabilities`. This should be reworked so we no longer use `profile` and just use `capabilities`.

PR #4928, opened 2024-08-27
Allow capabilities to be used with `[shader("...")]`

ArielG-NV:
Fixes: #4917

Changes:
1. Allow using capabilities instead of `Stage`s with `EntryPointAttribute`.
2. When resolving capabilities for an entrypoint+profile (per entrypoint) in `resolveStageOfProfileWithEntryPoint` add our `EntryPointAttribute` capabilities.
3. Added tests and some capabilities related clean-up

Note: due to 'Profile' still being used internally we cannot pass `EntryPointAttribute` capabilities to IR (and require spirv versions using `EntryPointAttribute`), related to  #4421


source/slang/slang-capabilities.capdef
```diff
@@ -414,19 +388,65 @@ alias raytracing_motionblur =  raytracing + motionblur | cuda;
...
+alias _raygeneration = _raygen;
+alias raygen = _raygen + raytracing;
+alias raygeneration = _raygeneration + raytracing;
+alias intersection = _intersection + raytracing;
+alias anyhit = _anyhit + raytracing;
+alias closesthit = _closesthit + raytracing;
+alias callable = _callable + raytracing;
+alias miss = _miss + raytracing;
+
+alias mesh = _mesh + meshshading_internal;
```
csyonghe, line 417: Should it be _meshshading_internal?

csyonghe (dismissed):

source/slang/slang-capabilities.capdef
ArielG-NV: The capability should be renamed👍
I think we can just do `_meshshading` since this capability pairs with `meshshading`.

csyonghe (approved):

PR #4949, opened 2024-08-28
Add `ReportIntersectionOptix` for optix.

### LINKED ISSUE #4948: `ReportHit` not supported in intersection shader for CUDA target
Hi, I am writing a custom intersection shader which I plan to use with a custom primitive in OptiX. However, when I call the `ReportHit` function to report an intersection, the compiler throws an error saying that the instruction is not supported for the target.

The following is a minimal reproducing example of the error:

## Shader
```
[shader("intersection")]
void intersection() {
    ReportHit(0.0, 0, 0);
}
```

## Command
```
slangc.exe .\repro.slang -o test.cu
```

## Output
```
void intersection() {
     ^~~~~~~~~~~~
.\repro.slang(2): error 36107: entrypoint 'intersection' does not support compilation target 'cuda' with stage 'intersection'
void intersection() {
     ^~~~~~~~~~~~
.\repro.slang(3): note: see using of 'ReportHit'
    ReportHit(0.0, 0, 0U);
    ^~~~~~~~~
hlsl.meta.slang(15857): note: see definition of 'ReportHit'
hlsl.meta.slang(15856): note: see declaration of 'require'
```

Modifying the `hlsl.meta.slang` as follows makes it work for integer attribute types but doesn't work for custom structs.
```diff
__generic<A>
[ForceInline]
-[require(glsl_hlsl_spirv, raytracing_intersection)]
+[require(cuda_glsl_hlsl_spirv, raytracing_intersection)]
bool ReportHit(float tHit, uint hitKind, A attributes)
{
    __target_switch
    {
    case hlsl:
        __intrinsic_asm "ReportHit($0, $1, $2)";
+   case cuda: __intrinsic_asm "optixReportIntersection";
    case glsl:
    case spirv:
        [__vulkanHitAttributes]
        static A a;
        a = attributes;
        return __reportIntersection(tHit, hitKind);
    }
}
```

This modification suffices for my current use case since I don't need to pass attributes to my hit shaders. However, it would be nice to have support for this.

csyonghe:
Fixes #4948.


jkwak-work (approved):
Looks good to me

PR #4952, opened 2024-08-28
Metal: Texture write fix

Dynamitos:
In Metal, there is no overload for different component counts of textures, only 4-component writes are allowed. For textures that do not have four components the unused components get discarded, as far as I can gather from the specs.


source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1797,6 +1797,21 @@ namespace Slang
                     }
                 }
             }
+            for(auto inst : block->getChildren())
+            {
+                if(auto write = as<IRImageStore>(inst))
```
csyonghe, line 1802: Can we move this if statement to the loop above, instead of doing a fresh loop?

Dynamitos:
for now i implemented a split between ImageStore and ImageStoreArray, because i cant split the array component from the coordinate in hlsl.meta.slang, but also i dont know when to split it off in the legalize because i dont know if the target texture is an array texture or not. this way if the swizzle does work in the future the entire ImageStoreArray can be deleted and the array coordinate passed the aux coord in ImageStore

csyonghe:
You should be able to check the type of the image operand to get its shape and arrayness.

auto type = imageStoreInst->getOperand(0)->getDataType();
auto imageType = as<IRTextureType>(type);
imageType->xxx //get info about the texture type.

Dynamitos:
Now a much better solution, not needing the ImageStoreArray anymore, but IMO it would be a bit better to swizzle the location in the hlsl.meta.slang instead if doing it in the legalize step

source/slang/hlsl.meta.slang
```diff
@@ -3018,56 +3022,8 @@ extension __TextureImpl<T,Shape,isArray,0,sampleCount,$(access),isShadow, 0,form
...
-                        }
-                        else
-                        {
-                            // void write(Tv color, uint2 coord, uint face, uint lod = 0) const
-                            __intrinsic_asm "$0.write($2, vec<uint,2>(($1).xy), uint(($1).z), uint(($1).w))";
-                        }
-                        break;
-                    }
+                    // last arg will be replaced with the split off array index
+                    __metalImageStore(this, location, newValue, 0);
```
csyonghe, line 3026: I think you should be able to do this completely in stdlib without implementing any logic for kIROp_ImageStore in the compiler backend.

```
[require(metal, texture_sm_4_1)]
static void __metalImageStore(This val, vector<uint, Shape.dimensions> location, T value, uint arrayIndex)
{
     __intrinsic_asm "...";
}

//...
case metal:
    __metalImageStore(this, __vectorReshape<Shape.dimensions>(location), newValue, location[Shape.dimensions+isArray-1]);
     return;
```

source/slang/slang-ir-metal-legalize.cpp
```diff
@@ -1740,6 +1740,46 @@ namespace Slang
         }
     };
 
+    // metal textures only support writing 4-component values, even if the texture is only 1, 2, or 3-component
```
csyonghe, line 1743: You should be able to use `vectorReshape<>` on the value to write into in stdlib as well? That should remove the need for this leglaization step, right?

csyonghe:
@Dynamitos Do you plan to clean up the image store inst with `vectorShape` to change the incoming value into float4?
If not we can also just merge the PR as is.

Dynamitos:
I tried locally but couldnt get it to work because the value type is a generic T and not a vector type, so `__vectorReshape` doesnt compile with the error 

    could not specialize generic for arguments of type (T)
                    __metalImageStore(this, __vectorReshape<Shape.dimensions>(location), __vectorReshape<4>(newValue), location[Shape.dimensions+isArray-1]);

because T is not necessarily a vector, in practice it could be a scalar, but in theory for this function it could be anything, as there are no constraints on the texture type

csyonghe:
OK, I think the solution here is fine as is.

csyonghe (dismissed):

Dynamitos:
Why does the texture write test fail now?

csyonghe:
It seems that we are generating invalid metal code after this fix. Can you try compiling the generated metal source of the test with the metalc compiler and see what errror it is producing?

Dynamitos:
it is not showing any errors when building the resulting metal code in xcode, or when building manually over the commandline

csyonghe:
Is it a metal 3.0 syntax? The current test infra builds for metal 2.4.

Dynamitos:
i cant seem to run the test on windows or mac, the runner always ignores the test no matter what api and category flags i pass

csyonghe:
You will need to install macos metal compiler tools for windows, and add this to PATH:
```
C:\Program Files\Metal Developer Tools\metal\macos\bin
```

Dynamitos:
Finally managed to reproduce on mac, apparently the test executable was out of date. the test failed because there were no metallib test checks in the file even though metallib was a test target. I dont know the layout or syntax for metallib as it seems to be binary, maybe some checks could be added by someone who knows how to do that

csyonghe:
Metallib will automatically be disassembled into text and being checked with filecheck. If you set a breakpoint in _filecheckTest function and inspect the output parameter, you can see what it is like.

csyonghe:
Everything looks good and merged. Thank you for bringing this to full completion!

PR #4953, opened 2024-08-28
Structured binding proposal

### LINKED ISSUE #4832: Design structured binding.

csyonghe:
Closes #4832.

### DIFF
--- a/docs/proposals/011-structured-binding.md
+++ b/docs/proposals/011-structured-binding.md
@@ -1,0 +1,47 @@
+SP #011: Structured Binding
+=================
+
+Tuple types can reduce boilterplate code of defining auxiliary structs, but they can introduce readability issues because the elements are not named.
+To mitigate this issue, we should support structured binding as a convenient way to access tuple elements with meaningful names.
+
+# Status
+
+Status: Proposal in review.
+
+Implementation: N/A
+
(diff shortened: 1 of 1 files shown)


docs/proposals/011-structured-binding.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Alternatives Considered
+
+We could have allowed mutable bindings in the syntax of:
+```
+var [a,b,c] = ...
+```
+That defines mutable variables a,b,c whose values are copied from the structure.
+However, mutable bindings can lead to confusions when modifying `a` doesn't change the value
+int the source composite object from the binding. To avoid this confusion, we simply disallow
```
jkwak-work, line 42: "int" seems to be a typo of "in"

jkwak-work (approved):
Looks good to me 
I wonder how other languages support mutable case

csyonghe:
This is trivial for languages with reference types or pointers. In c++,
auto& [a,b] = expr; translates to 

auto& x = expr;
auto& a = x._0;
auto& b = x._1;

But since it is not possible to support general reference types in Slang, we cannot trivially support this. It is best to leave this for later then introduce more ways to have reference typed things leaking to the user.

PR #4958, opened 2024-08-28
Make sure `NullDifferential` and its witness are removed after autodiff.

### LINKED ISSUE #4957: `NullDifferential_t` appears in output code.

csyonghe:
Closes #4957.

This is to make sure we always remove the `export`/`keepalive` decorations on the builtin `NullDifferential` type regardless of whether autodiff pass is run. To make it easy to identify this type from the linked IR, we add a `[__AutoDiffBuiltin]` attribute so we can easily trigger the logic to remove `export` decoration when we see this decoration.


jkwak-work (approved):
Looks good to me.
Maybe we can have a test for this change?

PR #4963, opened 2024-08-29
Support specialization constants.

### LINKED ISSUE #3581: Shader constants not working?
I want to use a shader with a shader constant like this

```glsl
layout(constant_id = 0) const int MULTIPLIER = 0;
```

but the code generated by slang always wants to use a uniform descriptor at Set 0 Binding 1. Is this not implemented or is this a bug? I am using vulkan.

edit: Not talking about the slang feature #3565 but the functionality used in vulkan, slang just needs to generate spirv that allows this

csyonghe:
This PR adds support for specialization constants in both spirv and glsl backends.

Speecialization constants can now be declared with:
```
[vk::specialization_constant]
const int specConst = 1; // implicit layout

[vk::constant_id(7)] // explicit layout
const int specConst = 1;

layout(constant_id = 9]
const int specConst = 1; // glsl flavor
```

Closes #3581.

### DIFF
--- a/tests/spirv/specialization-constant.slang
+++ b/tests/spirv/specialization-constant.slang
@@ -28,11 +28,10 @@
 [vk::specialization_constant]
 const int constValue0 = 1;
 
-[vk::specialization_constant]
-[vk::location(7)]
+[vk::constant_id(7)]
 const float constValue1 = 3.0f;
 
-[vk::specialization_constant]
+[SpecializationConstant]
 const bool constValue2 = true;
 
 layout(constant_id = 9) const int constValue3 = 111;
--- a/tests/spirv/specialization-constant.slang
+++ b/tests/spirv/specialization-constant.slang
@@ -1,0 +1,49 @@
+//TEST:SIMPLE(filecheck=GLSL): -target glsl
+//TEST:SIMPLE(filecheck=CHECK): -target spirv
+
+// CHECK-DAG: OpDecorate %[[C0:[0-9A-Za-z_]+]] SpecId 0
+// CHECK-DAG: %[[C0]] = OpSpecConstant %int 1
+
+// CHECK-DAG: OpDecorate %[[C1:[0-9A-Za-z_]+]] SpecId 7
+// CHECK-DAG: %[[C1]] = OpSpecConstant %float 3
+
+// CHECK-DAG: OpDecorate %[[C2:[0-9A-Za-z_]+]] SpecId 1
+// CHECK-DAG: %[[C2]] = OpSpecConstantTrue %bool
+
+// CHECK-DAG: OpDecorate %[[C3:[0-9A-Za-z_]+]] SpecId 9
+// CHECK-DAG: %[[C3]] = OpSpecConstant %int 111
+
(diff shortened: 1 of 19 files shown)


cheneym2 (approved):
LGTM

jkwak-work (commented):
Looks good to me.

Sorry for my late reply.

smarton-sc:
@csyonghe I just glanced at the code and I hope I didn't completely miss the ball, but it doesn't seem like this is the solution folks need. We don't just need specialization constants for the vulkan backend. It needs to work for all backends. Specialization constants need to be first class citizens in slang, and be translated to all backends appropriately (specialization constants for vulkan, function constants for metal, #ifdef for glsl, etc.). The same way that SPIRV-Cross translates OpSpecConstant to all backends. 

Having specilization at slang link time is cute, but it doesn't solve the problem of exploding shader permutations and compile times on target platforms. We need to emit platform shaders that are specializable natively for those reasons. Ex: compiled metallib with runtime specializable function constants.

Slang is not meant to be run at runtime for the most part, it should be an offline compiler, so giving it "efficient" specialization at the slang IR level is almost irrelevant. The platform shader that Slang outputs is the one that needs to be specializable efficiently at runtime. At runtime we only want to be running the lowest level platform compiler that's available (ex: create MTLFunction with newFunctionWithName from MTLLibrary while specifying MTLFunctionConstantValues). No other overhead, and no source code of any kind should be seen at runtime (on platforms that don't require source, ie: all but GL and WebGPU - the lost APIs). 

To end on a high note, I LOVE that you added implicit constant id, as that's indispensable and it was going to be my second request :) 

I would love to use Slang in production, as all other decisions made by the language seem great! But, without efficient runtime specialization, it's not usable.

csyonghe:
Thank you for your feedback! I agree that we should add support for it on other targets, this shouldn’t be hard to do. Can you file another issue for this request?

smarton-sc:
Will do. Thanks!

tests/spirv/specialization-constant.slang
```diff
@@ -0,0 +1,48 @@
...
+const bool constValue2 = true;
+
+layout(constant_id = 9) const int constValue3 = 111;
+
+RWStructuredBuffer<float> outputBuffer;
+
+[numthreads(1, 1, 1)]
+void computeMain()
+{
+    if (constValue2)
```
smarton-sc, line 44: @csyonghe can we please add a more complex expression here, like "if (constValue2 && constValue0 >= 1 && uni0 >= 1)" to make sure spec const expressions are working, even when mixed with non-spec-const variables? Also, maybe something like:

float arr[constValue0+1];
for (int i = 0; i< constValue0; ++i)
{
   arr[i] = (i < constValue0 ? i : i*i);
}
return arr[int(constValue1)/2];

csyonghe: Yes, we can definitely enhance the test with more complex expressions. Is this something you would like to submit a patch for?

smarton-sc: I could, but probably not anytime soon :(

PR #4965, opened 2024-08-29
Updated MDL shader sources for benchmarks

venkataram-nv:
The new benchmarking script will checkout the shader-slang MDL-SDK fork and then run only one sample instead of sixteen, to reduce the time taken by the workflow. The plan is for only the commits to master to run the full sixteen samples when pushing to the results page.


tools/benchmark/common.slang
```diff
@@ -0,0 +1,336 @@
+/******************************************************************************
+ * Copyright (c) 2019-2024, NVIDIA CORPORATION. All rights reserved.
```
jkwak-work, line 2: We cannot have a copyrighted code in the master whose license is MIT.
This file has to go to `external` directory as an independent repository.

jkwak-work:
I think you need to modify `.gitmodules` to include a new MDL repo after it is properly moved to shader-slang organization.

jkwak-work (approved):
Looks good to me

PR #4966, opened 2024-08-29
Add API method to specialize function reference with argument types

saipraveenb25:
- Adds `FunctionReflection::specializeWithArgTypes()`
- Fixes an issue with `findFunctionByName` returning a reference to a generic instead of a function-decl.

### DIFF
--- a/include/slang.h
+++ b/include/slang.h
@@ -2588,6 +2588,7 @@ extern "C"
     SLANG_API SlangReflectionType* spReflectionFunction_GetResultType(SlangReflectionFunction* func);
     SLANG_API SlangReflectionGeneric* spReflectionFunction_GetGenericContainer(SlangReflectionFunction* func);
     SLANG_API SlangReflectionFunction* spReflectionFunction_applySpecializations(SlangReflectionFunction* func, SlangReflectionGeneric* generic);
+    SLANG_API SlangReflectionFunction* spReflectionFunction_specializeWithArgTypes(SlangReflectionFunction* func, SlangInt argTypeCount, SlangReflectionType* const* argTypes);
 
     // Abstract Decl Reflection
 
@@ -3585,6 +3586,11 @@ namespace slang
         FunctionReflection* applySpecializations(GenericReflection* generic)
         {
             return (FunctionReflection*)spReflectionFunction_applySpecializations((SlangReflectionFunction*)this, (SlangReflectionGeneric*)generic);
+        }
+
+        FunctionReflection* specializeWithArgTypes(unsigned int argCount, TypeReflection* const* types)
+        {
+            return (FunctionReflection*)spReflectionFunction_specializeWithArgTypes((SlangReflectionFunction*)this, argCount, (SlangReflectionType* const*)types);
         }
     };
 
--- a/source/slang/slang-check-shader.cpp
+++ b/source/slang/slang-check-shader.cpp
@@ -1510,7 +1510,7 @@ namespace Slang
             outGenericArgs.add(argExpr);
         }
     }
-
+    
     Type* Linkage::specializeType(
         Type*           unspecializedType,
         Int             argCount,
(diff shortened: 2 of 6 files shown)


source/slang/slang-check-shader.cpp
```diff
@@ -1510,7 +1510,7 @@ namespace Slang
             outGenericArgs.add(argExpr);
         }
     }
-
+
```
csyonghe, line 1513: trailing space.

source/slang/slang.cpp
```diff
@@ -1348,6 +1347,59 @@ SLANG_NO_THROW slang::TypeReflection* SLANG_MCALL Linkage::specializeType(
     return asExternal(specializedType);
 }
 
+DeclRef<GenericDecl> getGenericParentDeclRef(
```
csyonghe, line 1350: can just use `getInnermostGenericParent`?

source/slang/slang.cpp
```diff
@@ -1348,6 +1347,59 @@ SLANG_NO_THROW slang::TypeReflection* SLANG_MCALL Linkage::specializeType(
...
+    auto genericDeclRef = createDefaultSubstitutionsIfNeeded(astBuilder, visitor, DeclRef(genericDecl)).as<GenericDecl>();
+    return substituteDeclRef(SubstitutionSet(declRef), astBuilder, genericDeclRef).as<GenericDecl>();
+}
+
+DeclRef<Decl> Linkage::specializeWithArgTypes(
+    DeclRef<Decl>   funcDeclRef,
+    List<Type*>         argTypes,
+    DiagnosticSink*     sink)
+{
+    SharedSemanticsContext sharedSemanticsContext(this, nullptr, sink);
```
csyonghe, line 1373: We should really start to hold a semantic visitor on the linkage.

csyonghe (commented):
The fact that we represent generic decl as a parent of the actual func/type decl is an implementation detail and subject to future changes. We really don't want to expose this detail via a publicly visible API that would lock us in this representation.

I would like us to figure out a way to hide this detail from the user, so that the user can always just see the actual decl itself without having to deal with GenericDecl explicitly.

include/slang.h
```diff
@@ -2589,6 +2589,7 @@ extern "C"
     SLANG_API SlangReflectionType* spReflectionFunction_GetResultType(SlangReflectionFunction* func);
     SLANG_API SlangReflectionGeneric* spReflectionFunction_GetGenericContainer(SlangReflectionFunction* func);
     SLANG_API SlangReflectionFunction* spReflectionFunction_applySpecializations(SlangReflectionFunction* func, SlangReflectionGeneric* generic);
+    SLANG_API SlangReflectionFunction* spReflectionFunction_specializeWithArgTypes(SlangReflectionFunction* func, SlangInt argTypeCount, SlangReflectionType* const* argTypes);
```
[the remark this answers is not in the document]
csyonghe: I see. As long as we make sure the API doesn't assume generic decl is the parent of the decl itself, and just treat generic as a different aspect of the same decl, then we should be fine.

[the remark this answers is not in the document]
csyonghe: we should be careful on the ast - children enumeration logic to not list the generic decl as the children, but instead the inner decl.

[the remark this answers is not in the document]
saipraveenb25: Yes, this is currently the case. If the decl-walking API hits a generic decl, the user must cast it to `SlangReflectionGeneric` and call `getInnerDecl()` to get the inner node.

source/slang/slang-check-shader.cpp
saipraveenb25: Fixed!

source/slang/slang.cpp
saipraveenb25: The linkage now holds a ref-ptr to a shared semantics context that is used for all the reflection functions.

source/slang/slang.cpp
saipraveenb25: This is a bit different unfortunately. 
This function returns the decl-ref of the generic-decl while `getInnermostGenericParent` returns the decl-ref of the inner decl of the closest generic-decl (this is so that substitutions are available)

PR #4968, opened 2024-08-30
Support extension on generic type.

### LINKED ISSUE #3960: "Ambiguous reference to This" in extension method returning This
I'm trying to do something like this:

```slang
interface Monoid {
    static const This identity;

    This combine(This b);
}

extension Monoid {
    This wave_reduce() {
        // TODO: actual logic to do wave reduction
        return This.identity;
    }
}
```

The idea is to implement it for the addition monoid using WaveActiveSum, but let it fall to the generic implementation for monoids that don't have such a specialization. When I compile this, I get:

```
monoid.slang(25): error 39999: ambiguous reference to 'This'
```

I'm not sure if this is a real limitation in the language's type system, or just a case not well handled. In Rust, the extension wouldn't be in a separate block, it would just be a default implementation of the method in the trait.

csyonghe:
Closes #3960.

This PR enables the following code:
```
extension<T:IFoo> T : IBar
{
     // Make every type that conforms to IFoo also conform to IBar.
}
```

Our existing type system has almost everything to support this, in that we already allow things like:
```
__generic<T:IFoo>
extension MyType<T> : IBar {}
```

Supporting extending `T` directly isn't that much different from the `MyType<T>` case. The one challenge we need to overcome is we shoudn't be considering the extension for every existing type in the linkage. Instead, we disallow general unconstraint extension such as `extension<T> T {...}` and require all `T` to be constraint with some interface `IFoo`. Then we can register this extension under the `IFoo` decl. 

When considering whether `extension<T:IFoo> T:IBar{}` can be applied to some type `X`, we go through all known base facets of `X` and see if there are any facet corrsponding to `IFoo`. If so, we will try to join `X` with `T`, and if it is successful, then we add the extension to the facet list of `X`. However, do so effectively makes `X` conform to `IBar`, so we need to repeat the process and add any `extesnion<T:IBar> T : IBaz` to the facet list...

The next thing to consider is to make sure not to recurse indefeinitely when there is a "circle" of extensions such as:
```
extension<T:IFoo> T:IBar {}
extension<T:IBar> T:IFoo {}
```
Theoretically we can make this work, since it just means that every type that conforms to `IFoo` also conforms to `IBar` and vice versa. But this is currently not implemented in this PR and we will issue a diagnostic if such circular extension is detected.


jkwak-work:
The syntax at first look was very confusing and I thought it was a typo.
It feels strange to have `<` and `>` right after `extension`, because `<` and `>` have been always some kind of suffix to `template` keyword, function or struct name.

tests/language-feature/extensions/this-in-extension.slang
```diff
@@ -0,0 +1,38 @@
...
+
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK): -shaderobj
+
+interface IFoo
+{
+    static const This identity;
+}
+
+__generic<T:IFoo>
+extension T
```
jkwak-work, line 13: This is little different from the repo example from the issue.
And even with this PR, we still cannot do the following, can we?
```
extension IFoo
```
It looks like `__generic<T:IFoo>` is a workaround for the limitation but I am not sure what the limitation exactly is.
In other words, why we cannot do `extension IFoo` here, instead?

jkwak-work (approved):
Looks good to me.

tests/language-feature/extensions/this-in-extension.slang
csyonghe: The whole purpose of the change is to get rid of `extension IFoo` from the language, because it does not make a lot of sense syntactically: you can't really add a requirement to an interface, because that will make all existing types conforming to the interface no longer conforming.

What `extension IFoo` really means is `extension<T:IFoo> T`, and this PR implements proper support for this, which will replace all existing uses of `extension IFoo`. The currently implementation of the compiler trying to support `extension IFoo` in a very hacky way that is not correct in many cases. It is much cleaner to just remove all these logic from the compiler and ban it forever.

csyonghe:
As for syntax, the only place we can write `<>` without the `__generic` keyword is for it to appear after `extension`, because extension themselves don't really have a name, and you can think of it as:

```
extension omittedNameForExtensionDecl<T:IFoo> for T : IBar
{
}
```

Issue #4883, opened 2024-08-20
Initial precompiled SPIR-V support (-embed-spirv)


cheneym2:
Add support for export keywords in compiled SPIR-V, mirroring HLSLExportDecoration.

Add -embed-spirv support.

Libraries are not linked yet, but SPIR-V is outputted with linkage decorations.

PR #4974, opened 2024-08-30
Initial -embed-spirv support

cheneym2:
Add support for SPIR-V precompilation using the framework established for DXIL.

Work on #4883


source/slang/slang-compiler.h
```diff
@@ -2733,11 +2733,14 @@ namespace Slang
 
         SlangResult emitPrecompiledDXIL(ComPtr<IArtifact>& outArtifact);
 
+        SlangResult emitPrecompiledSPIRV(ComPtr<IArtifact>& outArtifact);
```
csyonghe, line 2736: This is fine for now, but we may want to consider unifying the dxil/spirv methods with a single `emitPrecompiledDownstreamIR(CodeGenTarget target, ComPtr<IArtifact>& outArtifact);`.

source/slang/slang-compiler.h
```diff
@@ -2733,11 +2733,14 @@ namespace Slang
...
+
         void maybeDumpIntermediate(IArtifact* artifact);
 
-        // Used to cause instructions available in precompiled DXIL to be
+        // Used to cause instructions available in precompiled blobs to be
         // removed between IR linking and target source generation.
         bool removeAvailableInDXIL = false;
+        bool removeAvailableInSPIRV = false;
```
csyonghe, line 2743: can we unify them as `removeAvailableInDownstreamIR`?

source/slang/slang-compiler.h
```diff
@@ -2815,6 +2818,7 @@ namespace Slang
         virtual SLANG_NO_THROW void SLANG_MCALL setTargetForceDXLayout(int targetIndex, bool value) SLANG_OVERRIDE;
         virtual SLANG_NO_THROW void SLANG_MCALL setTargetGenerateWholeProgram(int targetIndex, bool value) SLANG_OVERRIDE;
         virtual SLANG_NO_THROW void SLANG_MCALL setEmbedDXIL(bool value) SLANG_OVERRIDE;
+        virtual SLANG_NO_THROW void SLANG_MCALL setEmbedSPIRV(bool value) SLANG_OVERRIDE;
```
csyonghe, line 2821: Can we just use one single option called `embedDownstreamIR`?

source/slang/slang-compiler.h
cheneym2: I don't think so because one module IR may have both embedded DXIL and embedded SPIRV, with a different assortment of coverage in either precompiled blob.

source/slang/slang-compiler.h
cheneym2: Yes

source/slang/slang-compiler.h
csyonghe: But when you are compiling for dxil, you obviously want to remove things that are availble in dxil if the flag is set, and same for spirv. There is no need to think about dxil when generating spirv or vice versa.

source/slang/slang-compiler.h
cheneym2: Done

source/slang/slang-compiler.h
cheneym2: Okay, done.

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -159,26 +159,29 @@ bool removeRedundancyInFunc(IRGlobalValueWithCode* func)
...
-            continue;
-        }
-        if (globalInst->findDecoration<IRAvailableInDXILDecoration>())
-        {
-            // Gut the function definition, turning it into a declaration
-            for (auto inst : funcInst->getChildren())
+            if (auto dec = globalInst->findDecoration<IRAvailableInDownstreamIRDecoration>())
             {
-                if (inst->getOp() == kIROp_Block)
+                if ((dec->getTarget()->getValue() == SLANG_DXIL && target == CodeGenTarget::HLSL) ||
```
csyonghe, line 174: Can this be simplified to just `dec->getTarget()->getValue() == target`?

source/slang/slang-ir-insts.h
```diff
@@ -430,9 +430,14 @@ IR_SIMPLE_DECORATION(NonCopyableTypeDecoration)
 IR_SIMPLE_DECORATION(HLSLMeshPayloadDecoration)
 IR_SIMPLE_DECORATION(GlobalInputDecoration)
 IR_SIMPLE_DECORATION(GlobalOutputDecoration)
-IR_SIMPLE_DECORATION(AvailableInDXILDecoration)
 IR_SIMPLE_DECORATION(DownstreamModuleExportDecoration)
 
+struct IRAvailableInDownstreamIRDecoration : IRDecoration
+{
+    IR_LEAF_ISA(AvailableInDownstreamIRDecoration)
+    IRIntLit* getTarget() { return cast<IRIntLit>(getOperand(0)); }
```
csyonghe, line 438: I recommend having a `CodeGenTarget getTarget()` accessor.
If there are places where we actually need the `IRIntLit` inst, we can have a separate `IRIntLit* getTargetInst()` accessor.

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: The current code gen target is HLSL at this point, but the precompiled blob is DXIL.

source/slang/slang-compiler.h
cheneym2: Unified

source/slang/slang-ir-insts.h
cheneym2: Done

csyonghe (approved):

csyonghe (approved):

PR #4977, opened 2024-08-30
Fix extension override behavior, and disallow extension on interface types.

csyonghe:
With `extension<T:IFoo> T` supported, this PR disallows the ill-defined `extension IFoo` syntax.

This is a technically breaking change, but we must clean up this part of the language before more widespread use of this problematic feature.

Also added a test to make sure the overriding behavior of extensions are correct.

## Breaking Change:

`extension IFoo {}` where `IFoo` is an interface type will now result in a compile time error. This kind of code should always be written as `extension<T:IFoo> T` in the future.

### DIFF
--- a/tests/language-feature/extensions/extension-override.slang
+++ b/tests/language-feature/extensions/extension-override.slang
@@ -1,0 +1,65 @@
+// Test that the override behavior around extensions and generic extensions works as expected.
+
+// When there are multiple ways for a type to conform to an interface, then the expected behavior
+// is that:
+// 1. If the type directly implements an interface, use that conformance.
+// 2. Otherwise, if there is a direct extension on the type that makes it conform to the interface, use that
+//    extension.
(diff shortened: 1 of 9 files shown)


tests/language-feature/extensions/extension-override.slang
```diff
@@ -0,0 +1,65 @@
+// Test that the override behavior around extensions and generic extensions works as expected.
```
ArielG-NV, line 1: I think we need a separate diagnostic to test `Diagnostics::invalidExtensionOnInterface` to ensure this functionality does not break in the future.

ArielG-NV (approved):
Code looks good to me, 1 suggestion.

I don't have too much practice in less conventional applications of the `extension` keyword, it may be a good idea to have someone else also review the code.

tests/language-feature/extensions/extension-override.slang
csyonghe: Done.

PR #4980, opened 2024-09-02
Preserve name in DeclRefExpr for correct highlighitng of `This`.

### LINKED ISSUE #4981: Syntax highlighting/hover info around `This` type is incorrect.
```
struct G
{}

extension G
{
    This my();  // bug: only `T` got highlighted instead of the entire `This` token.
}
```

csyonghe:
Closes #4981.

Fix a bug in language server that incorrectly highlights `This` type declref using the token length of the actual type name instead of just highlighting 4 characters.

The fix is to make sure the `name` field of constructed `DeclRefExpr` for the lookup result uses the `name` from the original `VarExpr` instead of the name from the referenced decl.


kaizhangNV (approved):
LGTM

PR #4982, opened 2024-09-02
Open existential on arguments after overload resolution.

### LINKED ISSUE #4467: Properly handle the mismatching "mutating" signature for implementing interfaces
**Problem description**
When the interface declares a function with `[mutating]`, the implementation also has to have `[mutating]`.
If the implementation didn't have the keyword, Slang crashes with an internal error.

**Goal**
We should print a proper error for the mistake.
Or it should work even when the implementation didn't have the matching `[mutating]` keyword.

**Repro steps**
The following code can reproduce the issue.
```
//TEST:SIMPLE:
//TEST:SIMPLE:-DWITH_MUTATING
interface IHitInfo
{
};

interface IClosestHitQuery
{
    [mutating] void traceRay(out IHitInfo hit);
};

struct GeneratePathClosestHitQuery : IClosestHitQuery
{
#if defined(WITH_MUTATING)
    [mutating]
#endif
    void traceRay(out IHitInfo hit)
    {
        hit = {};
    }
};
```
Note that the modifier `out` is required to reproduce the issue.

csyonghe:
Closes #4467.

The problem revealed here is that in `ResolveInvoke` we are performing `openExistential` on arguments before resolving overloaded callees. This is needed in order to correctly resolve generic arguments for generic callees. However we can't always us the opened existential value as argument if the parameter is mutable.

So the fix here is to switch back to original argument expr after an overload candidate is picked and we can figure out the corresponding parameter is a mutable one.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -2266,6 +2266,7 @@ namespace Slang
         for (auto& arg : expr->arguments)
         {
             arg = maybeOpenRef(arg);
+            arg = maybeOpenExistential(arg);
         }
 
         context.originalExpr = expr;
@@ -2394,6 +2395,10 @@ namespace Slang
             // the user the most help we can.
             if (shouldAddToCache)
                 typeCheckingCache->resolvedOperatorOverloadCache[key] = *context.bestCandidate;
+
+            // Now that we have resolved the overload candidate, we need to undo an `openExistential`
+            // operation that was applied to `out` arguments.
+            //
             auto funcType = context.bestCandidate->funcType;
             ShortList<ParameterDirection> paramDirections;
             if (funcType)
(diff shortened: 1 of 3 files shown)


source/slang/slang-check-overload.cpp
```diff
@@ -2415,6 +2395,45 @@ namespace Slang
...
+            if (funcType)
+            {
+                for (Index i = 0; i < funcType->getParamCount(); i++)
+                {
+                    paramDirections.add(funcType->getParamDirection(i));
+                }
+            }
+            else if (auto callableDeclRef = context.bestCandidate->item.declRef.as<CallableDecl>())
+            {
+                for (auto param : callableDeclRef.getDecl()->getMembersOfType<ParamDecl>())
```
ArielG-NV, line 2413:
```suggestion
                for (auto param : callableDeclRef.getDecl()->getParameters())
```

ArielG-NV (dismissed):
Code looks good to me, 1 suggestion (clean-up).

ArielG-NV (approved):

Issue #4916, opened 2024-08-26
Create util function to remove non-empty directory


kaizhangNV:
In https://github.com/shader-slang/slang/pull/4910, we create `cleanupRecordFiles()` in our test to help to remove a non-empty folder as existing `Path::remove` doesn't support that.

It's nice to have this utility in slang, so will integrate this to slang.

PR #4984, opened 2024-09-03
Add Path::removeNonEmpty() to remove non-empty dir

kaizhangNV:
We've implemented a function in slang-record-replay unit test to remove the non-empty directory, now move this function into slang `Path` namespace to make this function as an utility.

Close issue #4916

### DIFF
--- a/source/core/slang-io.cpp
+++ b/source/core/slang-io.cpp
@@ -17,6 +17,7 @@
 #ifdef _WIN32
 #   include <direct.h>
 #   include <windows.h>
+#   include <shellapi.h>
 #endif
 
 #if defined(__linux__) || defined(__CYGWIN__) || SLANG_APPLE_FAMILY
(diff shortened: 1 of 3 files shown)


source/core/slang-io.cpp
```diff
@@ -777,6 +779,61 @@ namespace Slang
...
+#ifdef _WIN32
+        // https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shfileoperationa
+        // Note: the fromPath requires a double-null-terminated string.
+        String newPath = path;
+        newPath.append('\0');
+        SHFILEOPSTRUCTA file_op = {
+            NULL,
+            FO_DELETE,
+            newPath.begin(),
+            "",
```
ArielG-NV, line 801: > This parameter must be set to NULL if it is not used.

We should probably set this to `NULL`  instead of `""` to avoid any temperamental issues with WinAPI and promotion/demotion to `PCZZSTR`.

ArielG-NV (commented):
1 optional comment/suggestion.
Changes look good to me.

I will let Yong approve the PR.

source/core/slang-io.cpp
kaizhangNV: ack.

csyonghe (approved):

PR #4986, opened 2024-09-03
Support `where` clause and type equality constraint.

### LINKED ISSUE #4890: Support `where` clause for defining generic constraints.

csyonghe:
Closes #4890.

This PR adds syntax for `where` clause and extended the generic syntax so that it can work more nicely with codebases that wish to share with C++.

Generics can now be defined as:
```
__generic<typename T, int N>
struct Vector {}
```

Which is the same as `struct Vector<T, let N:int>` or `struct Vector<T, int N>`.

Generic constraints can now be defined as trailing `where` clauses:

```
struct S<T> where T:IInteger {}
```

In addition, this PR brings support for type equality constraints:
```
interface IFoo { associatedtype A; }

struct S<T>
   where T : IFoo
   where T.A == int   // T.A must be int
{
     int a = T.A(4); // allowed, because T.A is known to be equal to int.
}
```


source/slang/slang-ir-inst-defs.h
```diff
@@ -291,6 +291,10 @@ INST(IndexedFieldKey, indexedFieldKey, 2, HOISTABLE)
 // A placeholder witness that ThisType implements the enclosing interface.
 // Used only in interface definitions.
 INST(ThisTypeWitness, thisTypeWitness, 1, 0)
+
+// A placeholder witness for the fact that two types are equal.
+INST(TypeEqualityWitness, TypeEqualityWitness, 2, HOISTABLE)
+
```
tangent-vector, line 297: When we get around to being more formal with our type system, we may find that we need more nuance here and a type equality witness isn't just a no-op "placeholder."

It really comes down to whether we are talking about *identical* types, or just *equivalent* types. If it is possible in our system to have two types `T` and `U` where each is a subtype of the other, but they are *not* identical, then those types are equivalent, and a witness to their equivalence would basically amount to a pair of witnesses: one for each of the subtype relations involved. Those subtype witnesses would in turn but usable for converting a value of type `T` into one of type `U`, and vice versa, which might not otherwise be possible for the IR codegen to synthesize.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -4989,6 +4998,14 @@ struct ExprLoweringVisitorBase : public ExprVisitor<Derived, LoweredValInfo>
         auto superType = lowerType(context, expr->type);
         auto value = lowerRValueExpr(context, expr->valueArg);
 
+        // First, we check if the witness is a type equality witness.
+        // If so, we can simply emit a bit cast to the target type that should eventually
+        // fold out to a no-op.
+        if (isTypeEqualityWitness(expr->witnessArg))
+        {
+            return LoweredValInfo::simple(getBuilder()->emitBitCast(superType, getSimpleVal(context, value)));
+        }
```
tangent-vector, line 5007: As per the previous comment: this logic assumes that any two equal types must have bit-identical representations. That might be valid, but it isn't something we have carefully thought about.

tangent-vector (approved):
Overall this looks really good.
As discussed elsewhere, one big category of work that keeps this from being truly feature-complete is that we don't yet have a step to canonicalize the constraints on a generic declaration, which will both be important to producing consistent mangled names, and *also* required to detect whether or not two generic function signatures are equivalent (and should be considered redefinitions/redeclarations), or not (in which case they should be treated as distinct overloads).

source/slang/slang-ast-decl.h
```diff
@@ -583,6 +583,11 @@ class GenericTypeConstraintDecl : public TypeConstraintDecl
     TypeExp sub;
     TypeExp sup;
 
+    // If this decl is defined in a where clause, store the source location of the where token.
+    SourceLoc whereTokenLoc = SourceLoc();
+
+    bool isEqualityConstraint = false;
```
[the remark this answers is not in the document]
csyonghe: I thought about having separate subclasses for GenericTypeConstraintDecl but that will lead a much bigger change.

Having a single GenericTypeConstraintDecl with a getType() that returns the relation type seems easier to do.

source/slang/slang-check-impl.h
```diff
@@ -765,6 +765,8 @@ namespace Slang
         InheritanceInfo _getInheritanceInfo(DeclRef<Decl> declRef, DeclRefType* correspondingType, InheritanceCircularityInfo* circularityInfo);
         InheritanceInfo _calcInheritanceInfo(Type* type, InheritanceCircularityInfo* circularityInfo);
         InheritanceInfo _calcInheritanceInfo(DeclRef<Decl> declRef, DeclRefType* correspondingType, InheritanceCircularityInfo* circularityInfo);
+        void getDependentGenericParentImpl(DeclRef<GenericDecl>& genericParent, DeclRef<Decl> declRef);
+        DeclRef<GenericDecl> getDependentGenericParent(DeclRef<Decl> declRef);
```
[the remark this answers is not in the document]
csyonghe: It is looking for the generic parent of `T` when you have `Foo<T>`, because `Foo<T>` may have a constraint defined in `parentDeclOf(T)`.

source/slang/slang-ir-inst-defs.h
csyonghe: Because inheritance isn't really a thing right now in the language, such case won't arise. When we have general support of inheritance, then equivalence vs equality becomes relevant and we will need more formality about it.

Issue #4125, opened 2024-05-07
Document all capability atoms and profiles.


csyonghe:
We need a reference chapter in the user guide to document all supported capability atoms and their meanings.

Start a chapter at the end of user guide with the hierarchy "Reference -> Capability Profiles"

ArielG-NV:
Discussion from another issue about this topic:

> @ArielG-NV , do we have a way to dump the name strings of profiles?
> As far as I can see, there isn't a way to do it currently.

There currently is no way todo this.

> Since there are a lot of profile names, I don't think we can list all of them on the document.
> We should refer to a command-line option that can print out all the available profile name strings.

I believe we could do both options.

The idea I have in mind for documenting every capability/profile:
1. inside `tools\slang-capability-generator\capability-generator-main.cpp` we could add logic to require a `///` comment above the line with the definition of a capability. 
2. store this string inside `slang\build\source\slang\capability\slang-generated-capability-defs-impl.h`.
3. Add build step to Slang (or CI) to autogenerate a user-doc with all capabilities using `slang\build\source\slang\capability\slang-generated-capability-defs-impl.h`
4. output the auto-generate code from '.3' as a a command-line output with slangc if `slangc -help capabilities` is added (or something similar)

Related issue: #4125

csyonghe:
Parsing and storing the comments is a good idea, but for now let's just make sure the comments are in place first.

PR #5004, opened 2024-09-03
Document All Capability Atoms and Profiles

ArielG-NV:
Fixes: #4125

Unimplemented Considerations:
1. This PR does not add support to query all capability-atom's from a command-line option. It is understood that this might be desired, due to this, the logic to generate `docs\user-guide\a3-02-reference-capability-atoms.md` was made to be "command-line friendly" so minimal changes are needed to pipe our documentation into a command-line option if this change is to be added.

Changes:
1. Added a way to document atoms inside `.capdef`. Method to document is described under `source\slang\slang-capabilities.capdef`. The goal is to error if a public atom does not have any form of documentation to ensure we always have up-to-date documentation to guide user on what an atom is/does. * The following `.capdef` file syntax was added * /// [HEADER_GROUP] * /// regular comment
2. When capability generator runs it auto-generates `docs\user-guide\a3-02-reference-capability-atoms.md`
3. Added to the user-guide 3 sections: `Reference`, `Reference -> Capability Profiles`, `Reference -> Capability atoms` section

### DIFF
--- a/docs/user-guide/a2-target-specific-features.md
+++ b/docs/user-guide/a2-target-specific-features.md
@@ -1,4 +1,3 @@
-
 ---
 layout: user-guide
 ---
--- a/docs/user-guide/a3-01-reference-capability-profiles.md
+++ b/docs/user-guide/a3-01-reference-capability-profiles.md
@@ -1,0 +1,49 @@
+---
+layout: user-guide
+---
+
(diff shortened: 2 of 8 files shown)


ArielG-NV:
Note: I need to remake this PR, this one will be closed as is.

Issue #4921, opened 2024-08-27
Semantics when switch init expression and case expression types don't match


aleino-nv:
What is Slang's semantics when the type of the case expression doesn't match the type in the switch init expression?
Example:

    float test(uint x)
    {
        switch (x)
        {
            case -1:
                return 1.0f;

            default:
                break;
        }

        return 0.0f;
    }

The [GLSL specification says](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf):

    The type of init-expression in a switch statement must be a scalar integer. The type of the constantexpression value in a case label also 
    must be a scalar integer. When any pair of these values is
    tested for “equal value” and the types do not match, an implicit conversion will be done to convert
    the int to a uint (see “Implicit Conversions”) before the compare is done.

In GLSL this means `-1` will be treated like `case bitcast<uint>(int(-1))` which means `wrap<uint>(int(-1))` in two's complement representation.

Based on GLSL and HLSL output, it seems Slang's de facto semantics is `case bitcast<uint>(int(-1))`:

- I see that the GLSL output is just `case -1` and as mentioned the GLSL spec reads this as `case bitcast<uint>(int(-1))`
- I don't think the HLSL specification mentions this issue, but for the HLSL backend we output `case int(-1)`, and I believe the DXC compiler ends up turning this into `case bitcast<uint>(int(-1))`.

Can someone confirm?

aleino-nv:
This came up as part of https://github.com/shader-slang/slang/issues/4807 because WGSL doesn't allow these kinds of type mismatches, and doesn't try to do any implicit conversions.

aleino-nv:
This issue [was actually a TODO in the code](https://github.com/shader-slang/slang/blob/87d3d4f73a2b3af3ec00f221350864602d12321d/source/slang/slang-check-stmt.cpp#L311):


    void SemanticsStmtVisitor::visitCaseStmt(CaseStmt* stmt)
    {
        auto expr = CheckExpr(stmt->expr);

        // coerce to type being switch on, and ensure that value is a compile-time constant
        // The Vals in the AST are pointer-unique, making them easy to check for duplicates
        // by addeing them to a HashSet.
        auto exprVal = tryConstantFoldExpr(expr, ConstantFoldingKind::CompileTime, nullptr);
        auto switchStmt = FindOuterStmt<SwitchStmt>();

        if (!switchStmt)
        {
[...]

Issue #4807, opened 2024-08-08
Add initial support for WGSL output


aleino-nv:
This is meant to start addressing issue #4477.

Add support for generating some valid WGSL code for a few very simple shaders, e.g. the ones [here](https://aleino-nv.github.io/issues/4483/results2.html).

Broadly go through the Slang spec (looking at the HLSL spec where Slang detail is missing), and make sure that all of the basic language constructs generate ok WGSL.

aleino-nv:
@csyonghe Any opinions on what we should do with int8_t and uint8_t? WGSL doesn't support 8 bit integer types.

csyonghe:
We should extend our capability system to report errors when trying to uses these types on wgsl. For now just diagnose an error during emit when you see these opcodes.

aleino-nv:
@csyonghe There is also f16, which requires an [extension](https://www.w3.org/TR/WGSL/#extension-f16) in WGSL.
Extensions are enabled in WGSL source using the `enable` directive.

- How should we deal with this in Slang?
  - Should we have a f16 capability that the user must opt in to?
  - Do we silently generate the `enable` directive if `f16` is generated anywhere?

csyonghe:
Yes, we will silently emit that enable directive for now. Check for how this is handled in glsl emit, and we should be able to follow the same technique.

aleino-nv:
I haven't pushed anything to the main branch yet -- I'm still rebasing stuff as the initial WGSL support matures in [my local branch](https://github.com/aleino-nv/slang/tree/aleino/wgsl). I'd say this task is done, at least locally.

I don't want to close the task before I commit, so I will leave it open for now.

aleino-nv:
Another issue is that WGSL [doesn't support matrices with entries of integer type](https://www.w3.org/TR/WGSL/#matrix-types).

aleino-nv:
This has been split into issues #4930, #4931, #4932, #4933, #4934, #4935, #4936, #4937, #4938 and #4939 for more fine-grained tracking.

PR #5006, opened 2024-09-04
Initial WGSL support

aleino-nv:
This adds initial support for WGSL output.
With this, the majority of the shaders in `tests/compute` will output WGSL that compiles.


source/slang/slang-emit-c-like.cpp
```diff
@@ -1657,11 +1672,16 @@ bool CLikeSourceEmitter::shouldFoldInstIntoUseSites(IRInst* inst)
     return true;
 }
 
+bool CLikeSourceEmitter::isPointerSyntaxRequiredImpl(IRInst* /* inst */)
```
csyonghe, line 1675: Why is this needed?

source/slang/slang-emit-c-like.cpp
```diff
@@ -1842,7 +1862,8 @@ void CLikeSourceEmitter::emitInstResultDecl(IRInst* inst)
 
     emitRateQualifiers(inst);
 
-    if(as<IRModuleInst>(inst->getParent()))
+    bool const isConstant(as<IRModuleInst>(inst->getParent()));
```
csyonghe, line 1865: bool isConstant = ...

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+// This may lead to confusion (which is why this note is referenced in several
+// places), but the benefit of doing this is that the generated WGSL code is
+// simpler to generate and should be faster to compile.
+// A "terminology preserving" implementation would have to generate lots of
+// 'transpose' calls, or else perform more complicated transformations that
+// end up duplicating expressions many times.
+
+namespace Slang {
+
+void WGSLSourceEmitter::emitSwitchCaseSelectorsImpl(
```
csyonghe, line 28: We shouldn't be handling this logic here. The frontend should insert the necessary coercion logic to make sure case values have the same type as their condition.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    IRGlobalParam* varDecl, IRUniformParameterGroupType* type
+)
+{
+    auto varLayout = getVarLayout(varDecl);
+    SLANG_RELEASE_ASSERT(varLayout);
+
+    for (auto attr : varLayout->getOffsetAttrs())
+    {
+
+        LayoutResourceKind const kind {attr->getResourceKind()};
```
csyonghe, line 85: style:
LayoutResourceKind kind = attr->getResourceKind();

We don't use `const` if not necessary.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    {
+        m_writer->emit("@builtin(");
+        m_writer->emit(sysSemanticDecor->getSemantic());
+        m_writer->emit(")");
+    }
+
+    CLikeSourceEmitter::emitSimpleFuncParamImpl(param);
+}
+
+void WGSLSourceEmitter::emitTransposedMatrixType(
```
csyonghe, line 209: It feels confusing to have a dedicated `emitTransposedMatrixType`, we just need `emitMatrixType`.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    IRIntegerValue const colCountSlang {getIntVal(matrixType->getColumnCount())};
+    IRIntegerValue const rowCountSlang {getIntVal(matrixType->getRowCount())};
+    IRIntegerValue const colCountWGSL {rowCountSlang};
+    IRIntegerValue const rowCountWGSL {colCountSlang};
+    emitMatrixType(elementType, rowCountWGSL, colCountWGSL);
+}
+
+void WGSLSourceEmitter::emitMatrixType(
+    IRType *const elementType, IRIntegerValue const& rowCountWGSL,
+    IRIntegerValue const& colCountWGSL
```
csyonghe, line 222: For most part of the codebase, we are sticking to `const T*` or `const T&`, so let's stick with this convention if `const` is needed.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    return (n != 0U) && ((n - 1U) & n) == 0U;
+}
+
+void WGSLSourceEmitter::emitStructFieldAttributes(
+    IRStructType * structType, IRStructField * field
+    )
+{
+    // Tint emits errors unless we explicitly spell out the layout in some cases, so emit
+    // offset and align attribtues for all fields.
+    IRSizeAndAlignmentDecoration *const sizeAndAlignmentDecoration {
```
csyonghe, line 251: no initializer list or function call style initialization if we can avoid it because it is not as readable as `T v = value;`

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+
+    m_writer->emit("@align(");
+    m_writer->emit(fieldAlignment);
+    m_writer->emit(")");
+}
+
+bool WGSLSourceEmitter::isPointerSyntaxRequiredImpl(IRInst* inst)
+{
+    // Structured buffers are mapped to 'array' types, which don't need dereferencing
+    if (inst->getOp() == kIROp_RWStructuredBufferGetElementPtr)
```
csyonghe, line 287: I wonder why is this needed.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,1013 @@
...
+
+}
+
+void WGSLSourceEmitter::emitVarKeywordImpl(IRType * type, bool const isConstant)
+{
+    if (isConstant)
+        m_writer->emit("const");
+    else
+        m_writer->emit("var");
+    if (type->getOp() == kIROp_HLSLRWStructuredBufferType)
```
csyonghe, line 521: It feels wrong to test for structured buffer types during emit logic of a `var`. Looks like this needs to be lowered into a variable of pointer type during legalization.

source/slang/slang-ir-wgsl-legalize.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    void LegalizeWGSLEntryPointContext::legalizeEntryPointForWGSL(EntryPointInfo entryPoint)
+    {
+        legalizeSystemValueParameters(entryPoint);
+    }
+
+    void legalizeIRForWGSL(IRModule* module, DiagnosticSink* sink)
+    {
+
```
csyonghe, line 277: extraneous empty lines.

source/slang/slang-emit-c-like.cpp
```diff
@@ -4168,11 +4217,13 @@ void CLikeSourceEmitter::emitGlobalParam(IRGlobalParam* varDecl)
     SLANG_ASSERT(layout);
 
     emitVarModifiers(layout, varDecl, varType);
+    emitVarAttributes(varType, layout);
```
csyonghe, line 4220: I think we can just use `emitVarModifiers` for this purpose.

csyonghe (commented):
Overall, I think the principle we need to stick to is that we should do as little code translation as possible in `emit`, and move as much logic as possible to legalization passes in the IR. After legalization, the IR should be almost 1:1 to target code.

source/slang/slang-emit-c-like.cpp
aleino-nv: At first I just flipped `doesTargetSupportPtrTypes` to `true` for WGSL, but that caused a lot of strange output that didn't make sense (or compile), like e.g:

    *(&array)[i] = 5i; // instead of 'array[i] = 5i'
    
Therefore, I added `isPointerSyntaxRequiredImpl` (defaulting to `doesTargetSupportPtrTypes`) and replaced some calls to `doesTargetSupportPtrTypes` by `isPointerSyntaxRequiredImpl` which also takes the `IRInst` as context in order to be able to opt-in to the "pointer syntax path" in some cases.

csyonghe: My expectation is that `doesTargetSupportPtrTypes` needs to be false for WGSL backend. What are the cases where we want to emit pointer syntax in wgsl?

aleino-nv: Here is the function, with the various cases shere I opt out of the "pointer syntax path":


    bool WGSLSourceEmitter::isPointerSyntaxRequiredImpl(IRInst* inst)
    {
        // Case 1: Structured buffers are mapped to 'array' types, which don't need dereferencing
        if (inst->getOp() == kIROp_RWStructuredBufferGetElementPtr)
            return false;
    
        // Case 2: Don't emit "->" to access fields in resource structs
        if (inst->getOp() == kIROp_FieldAddress)
            return false;
    
        // Case 3: Don't emit "*" to access fields in resource structs
        if (inst->getOp() == kIROp_GlobalParam)
            return false;
    
        // Case 4: Emit 'globalVar' instead of  "*&globalVar"
        if (inst->getOp() == kIROp_GlobalVar)
            return false;
    
        return true;
    }


Here is an example of what goes wrong if I don't opt out in each of the cases:

*Case 1: (tests/compute/array-param.slang)*

    @binding(0) @group(0) var<storage, read_write> globalParams_outputBuffer_0 : array<i32>;

    // ...

    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var dispatchThreadID_1 : vec3<i32> = vec3<i32>(dispatchThreadID_0);
        var b_0 : array<vec3<f32>, i32(4)>;
        writeArray_0(&b_0);
    // error: wgpu: type: 2: message: Error while parsing WGSL: :20:7 error: cannot dereference expression of type 'i32'
        *(&globalParams_outputBuffer_0)[dispatchThreadID_1.x] = i32(b_0[i32(0)].x);
    //    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    // When opting out of "pointer syntax path", I get valid WGSL:
    //  (&globalParams_outputBuffer_0)[dispatchThreadID_1.x] = i32(b_0[i32(0)].x);
    
        
        return;
    }

*Case 2: (tests/compute/cast-zero-to-struct.slang)*

    @binding(0) @group(0) var<storage, read_write> globalParams_gOutputBuffer_0 : array<i32>;
    
    struct S_0
    {
         a_0 : vec2<i32>,
         b_0 : vec2<i32>,
         c_0 : vec2<i32>,
    };
    
    fn test_0( val_0 : i32) -> i32
    {
        var _S1 : vec2<i32> = vec2<i32>(i32(0), i32(0));
        var s_0 : S_0;
    // error: wgpu: type: 2: message: Error while parsing WGSL: :14:11 error: expected '=' for assignment
        (&s_0)->a_0 = _S1;
    //        ^^
    // When opting out of "pointer syntax path", I get valid WGSL:
    //  (&s_0).a_0 = _S1;

        // ...
    }

*Case 3: (tests/compute/dynamic-dispatch-11.slang)*


    struct Tuple_std140_0
    {
        @align(16) value0_0 : vec2<u32>,
        @align(8) value1_0 : vec2<u32>,
        @align(16) value2_0 : _std140_0,
    };
    
    @binding(16) @group(0) var<uniform> globalParams_gCb_0 : Tuple_std140_0;
    
    fn _S7( _S8 : i32) -> i32
    {
    // error: wgpu: type: 2: message: Error while parsing WGSL: :61:60 error: cannot dereference expression of type 'Tuple_std140_0'
        return MyImpl_run_0(unpackAnyValue8_0(unpackStorage_1(*globalParams_gCb_0).value2_0), _S8);
    //                                                         ^^^^^^^^^^^^^^^^^^
    // When opting out of "pointer syntax path", I get valid WGSL:
    //  return MyImpl_run_0(unpackAnyValue8_0(unpackStorage_1(globalParams_gCb_0).value2_0), _S8);    
    }

*Case 4: (tests/compute/global-init.slang)*

    @binding(0) @group(0) var<storage, read_write> globalParams_outputBuffer_0 : array<i32>;
    
    const gVar_0 : i32 = i32(16);
    
    fn test_0( inVal_0 : i32) -> i32
    {
    // error: wgpu: type: 2: message: Error while parsing WGSL: :7:24 error: cannot take the address of 'const gVar_0'
        return inVal_0 + *&gVar_0;
    //                     ^^^^^^
    // When opting out of "pointer syntax path", I get valid WGSL:
    //  return inVal_0 + gVar_0;
    }
    
    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var tid_0 : u32 = dispatchThreadID_0.x;
        (&globalParams_outputBuffer_0)[tid_0] = test_0((&globalParams_outputBuffer_0)[tid_0]);
        return;
    }

source/slang/slang-emit-wgsl.cpp
aleino-nv: I added some examples [here](https://github.com/shader-slang/slang/pull/5006#discussion_r1744880039).

source/slang/slang-emit-c-like.cpp
aleino-nv: I agree this workaround feels wrong, but I'm not sure how to do it right.
It seems I do need the "pointer syntax path" in many cases, but I was pretty confused about the output in the above cases, which I saw when I just set `doesTargetSupportPtrTypes` to return `true` unconditionally.

source/slang/slang-emit-wgsl.cpp
aleino-nv: Done. I tried to find and fix all other instances as well.

source/slang/slang-emit-wgsl.cpp
aleino-nv: Done. (It's a habit from working on safety-critical code -- I think the brace style init forces you to be more explicit about integers in some cases.)

source/slang/slang-emit-c-like.cpp
csyonghe: If I have `int arr[3]` in c++, `*&arr[0]` is valid code. I means take address of `arr`, then dereference that address to get a reference to the array, and then index into the array. It seems that the problem is in wgsl, you can't just do `arr[i]`, and have to do `&arr[i]`, which is weird from C/C++ perspective.

csyonghe: I guess my original question was the other way around: why can't the existing behavior of doesTargetSupportPtrTypes=false work out of the box? Why do we need to customize the behavior?

csyonghe: Why do you need to do `(&globalParams_outputBuffer_0)[tid_0]`? `globalParams_outputBuffer_0` is already an array, and you shouldn't need to take its address before access its elements.

source/slang/slang-emit-c-like.cpp
aleino-nv: Should I make `emitVarModifiers` virtual and then override the whole thing? It's currently not virtual and contains a bunch of calls to virtual functions. I'm not sure which ones of those would fit in this case. 
I guess `emitVarDecorationsImpl` could be close, but it just takes the `decl` so I'd have to replicate some code to get the layout and type as well (or just add those as arguments).

source/slang/slang-emit-c-like.cpp
csyonghe: I think we shouldn't emit things like `(&arr)[i]`, because `(&arr)`'s type is `ptr<array<int>>`, and it is not clear to me what it means to use `[]` affer an `ptr<array<int>>`, if this is c++, then the result of that is `ref<array<int>>`, which shouldn't allow assignment from an int to it.

source/slang/slang-emit-c-like.cpp
csyonghe: will `emitLayoutQualifiersImpl` work?

source/slang/slang-emit-c-like.cpp
csyonghe: If we delete `isPointerSyntaxRequiredImpl` and just use the default `doesTargetSupportPointer = false` behavior, what will break?

source/slang/slang-emit-wgsl.cpp
csyonghe: You just need to modify `SemanticsStmtVisitor::visitCaseStmt` into:
```
    void SemanticsStmtVisitor::visitCaseStmt(CaseStmt* stmt)
    {
        auto expr = CheckExpr(stmt->expr);
        auto switchStmt = FindOuterStmt<SwitchStmt>();

        if (!switchStmt)
        {
            getSink()->diagnose(stmt, Diagnostics::caseOutsideSwitch);
        }
        else
        {
            // coerce to type being switch on, and ensure that value is a compile-time constant
            // The Vals in the AST are pointer-unique, making them easy to check for duplicates
            // by addeing them to a HashSet.
            expr = coerce(CoercionSite::General, switchStmt->condition->type, expr);
        }

        auto exprVal = tryConstantFoldExpr(expr, ConstantFoldingKind::CompileTime, nullptr);
        stmt->expr = expr;
        stmt->exprVal = exprVal;
        stmt->parentStmt = switchStmt;
    }
```

source/slang/slang-type-layout.cpp
```diff
@@ -1817,6 +1817,7 @@ LayoutRulesFamilyImpl* getDefaultLayoutRulesFamilyForTarget(TargetRequest* targe
     case CodeGenTarget::ShaderSharedLibrary:
     case CodeGenTarget::CPPSource:
     case CodeGenTarget::CSource:
+    case CodeGenTarget::WGSL:
```
csyonghe, line 1820: This may not be correct, I think you want wgl to have similar layout rule with GLSL/SPIRV case for now.

csyonghe: If there are actual difference in type alignment and layout rules compared to spirv, we will need to branch out from that and have dedicated type layout classes for WGSL, but I think moving this line to after 1810 is a good start.

source/slang/slang-emit-wgsl.cpp
aleino-nv: (Done.)

This is #4921, which I hoped to defer because I hadn't thought through whether that would break something.

The de-facto behavior of Slang is currently that of GLSL, which will coerce signed to unsigned in case of mismatch, regardless of which is the condition and which is the case expression.

However I don't immediately see how just trying to coerce the case to be of the same type as the condition could cause semantics differing from the current behavior, so I removed this back-end coercion and added a patch to do it in the front-end.

source/slang/slang-emit-c-like.cpp
aleino-nv: "so arr[i] is not an l-value and I can't pass it to a function that expects inout parameter?"

I guess you mean just 'arr'? If so then yes I think that's correct. If you're passing it to a function, the function needs to take a `ptr<arr>` and you need to pass `&arr` to the function.

aleino-nv: > If we delete `isPointerSyntaxRequiredImpl` and just use the default `doesTargetSupportPointer = false` behavior, what will break?

Categories of failures when I never take the "pointer syntax" paths:

1. Passing function parameter without &
2. Assigning to ptr without *
3. Reading T from ptr-to-T without *

Here are examples:

*Category 1 (tests/compute/array-param)*

    fn writeArray_0( a_0 : ptr<function, array<vec3<f32>, i32(4)>>)
    {
        // ...
    }

    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var dispatchThreadID_1 : vec3<i32> = vec3<i32>(dispatchThreadID_0);
        var b_0 : array<vec3<f32>, i32(4)>;
    // Error while parsing WGSL: :19:18 error: type mismatch for argument 1 in call to 'writeArray_0',
    //   expected 'ptr<function, array<vec3<f32>, 4>, read_write>',
    //   got 'array<vec3<f32>, 4>'        
        writeArray_0(b_0);
    //               ^^^
    // When taking the "pointer syntax" path, valid WGSL is emitted:
    //  writeArray_0(&b_0);
        (&globalParams_outputBuffer_0)[dispatchThreadID_1.x] = i32(b_0[i32(0)].x);
        return;
    }

*Category 2 (tests/compute/inout)*

    fn testOut_0( x_0 : i32,  y_0 : ptr<function, i32>)
    {
    // Error while parsing WGSL: :5:5 error: cannot assign to parameter 'y_0'
        y_0 = x_0;
    //  ^^^
    // When taking the "pointer syntax" path, valid WGSL is emitted:
    //  *y_0 = x_0;
        return;
    }


*Category 3 (mutating-and-inout)*

    fn A_doThings_0( this_0 : ptr<function, A_0>,  y_0 : ptr<function, i32>)
    {
        var tmp_0 : i32 = this_0.x_0;
    // Error while parsing WGSL: :11:16 error: cannot assign 'ptr<function, i32, read_write>' to 'i32'
        this_0.x_0 = y_0;
    //             ^
    // When taking the "pointer syntax" path, valid WGSL is emitted:
    //  this_0.x_0 = *y_0;
    
        // ...
        
        return;
    }

source/slang/slang-emit-wgsl.cpp
aleino-nv: > It feels wrong to test for structured buffer types during emit logic of a `var`. 

I agree. This was a temporary hack that I had meant to clean up...

> Looks like this needs to be lowered into a variable of pointer type during legalization.

The equivalent of `RWStructuredBuffer<int> gOutputBuffer;` in WGSL is `var<storage, read_write> gOutputBuffer : array<i32>;` so it can't be a pointer type in this one case when we declare it.

Everywhere else, though, I have the hack described in `WGSLSourceEmitter::emitOperandImpl` to change all places where `gOutputBuffer` is used as an operand to use instead `&gOutputBuffer`.

aleino-nv: So I can lower structured buffer types to pointer types, but then I somehow need to deal with the case where structured buffer globals are declared, because those storage buffer declarations are not supposed to be pointers in WGSL.

That does sound preferable to me, but I'd like to know what you think before fixing it this way.

source/slang/slang-emit-c-like.cpp
aleino-nv: > will `emitLayoutQualifiersImpl` work?

Not currently because that only gets passed the layout, and I expect that I will also need the type:
(Yes I know this code is very wrong, but eventually I need to output "set" and "binding" here, and I need the type for that. For now I just need some code to emit distinct "set" and "binding" so that the output at least compiles.)


    void WGSLSourceEmitter::emitVarAttributesImpl(IRType * type, IRVarLayout* layout)
    {
        for (auto attr : layout->getOffsetAttrs())
        {
            // TODO:
            // This is not correct. For the moment this is just here as a hack to make
            // @binding and @group unique, so that we can pass WGSL compile tests.
            // This will have to be revisited when we actually want to supply resources to
            // shaders.
            if (
                (type->getOp() == kIROp_HLSLRWStructuredBufferType) ||
                (type->getOp() == kIROp_HLSLStructuredBufferType)
                )
            {
                m_writer->emit("@binding(");
                m_writer->emit(attr->getOffset());
                m_writer->emit(") ");
                m_writer->emit("@group(");
                m_writer->emit(attr->getSpace());
                m_writer->emit(") ");
                return;
            }
        }
    }

source/slang/slang-emit-c-like.cpp
aleino-nv: > I think we shouldn't emit things like `(&arr)[i]`...

In cases where arr is a structured buffer type, I currently have this hack (see the comment for my "explanation"):


    void WGSLSourceEmitter::emitOperandImpl(IRInst* inst, const EmitOpInfo& outerPrec)
    {
        // In WGSL, the structured buffer types are converted to ptr<AS, array<E>, AM>
        // everywhere, except for the global parameter declaration.
        // Thus, when these globals are used in expressions, we need an ampersand.
    
        if (inst->getOp() == kIROp_GlobalParam)
        {
            switch (inst->getDataType()->getOp())
            {
            case kIROp_HLSLStructuredBufferType:
            case kIROp_HLSLRWStructuredBufferType:
    
                m_writer->emit("(&");
                CLikeSourceEmitter::emitOperandImpl(inst, outerPrec);
                m_writer->emit(")");
                return;
            }
        }
    
        CLikeSourceEmitter::emitOperandImpl(inst, outerPrec);
    }

Should I solve this some other way, or insert some logic to not do this in the `structuredBuffer[i] = ...` case?

csyonghe: I don't quite understand the comment here, what is `ptr<AS, array<E>, AM>` and why do we need `&` to access the elements in `array<E>`?

It feel like if HLSLStructuredBuffer needs to be mapped to a type with different level of indirection, then this shall be done in the legalization pass, with all use sites updated with the correct conversion/dereferencing insts so emit logic doesn't need to deal with this.

source/slang/slang-emit-c-like.cpp
csyonghe: Instead of checking types, each offset attr provides `getResourceKind()`, and you can use that to decide whether or not to emit these decorations.

source/slang/slang-emit-c-like.cpp
aleino-nv: > I don't quite understand the comment here, what is `ptr<AS, array<E>, AM>` and why do we need `&` to access the elements in `array<E>`?

The  `ptr<AS, T, AM>` (`T = array<E>` in this case) notation comes from https://www.w3.org/TR/WGSL/#ref-ptr-types.
AS = address space (e.g. storage, function, uniform, private, etc...), AM = access mode (R, RW).

We don't need `&` to access elements of `array`, but it doesn't hurt AFAIK.
(Array indexing notation like `buffer[i]` seems to work both when `buffer` is of type `ptr<storage, array<E>, read_write>` and just `array<E>`.)

For all other cases where a structured buffer is used as an operand (such as when it's passed to a function or assigned to a `ptr<AS, array<E>, AM>` typed thing, we need the ampersand because the global object corresponding to a structured buffer cannot be declared as a pointer type.

> It feel like if HLSLStructuredBuffer needs to be mapped to a type with different level of indirection, then this shall be done in the legalization pass, with all use sites updated with the correct conversion/dereferencing insts so emit logic doesn't need to deal with this.

Yeah, I think that there are two cases:

- When translating the global declaration of a structured buffer, `StructuredBuffer<E> buffer` should translate to `var<storage, read_write> buffer : array<E>`.
- Any other place, when referring to `buffer` the references should be translated to expressions of type `ptr<storage, array<E>, read_write>`.

This is essentially what I have done, but yes it feels fragile to do it at the emit stage.

source/slang/slang-emit-wgsl.cpp
aleino-nv: > In this case, we should make sure an IRGlobalParam of StructuredBuffer is in the form of `IRGlobalParam p : Ptr<Array<E>, readwrite_storage>`, and insert `OpLoad` at use sites of the global param during legalization. That way we won't need any specialized logic for structured buffers as they don't exist when we get to emit.

So the `OpLoad` at the usage sites would take a `ptr<storage, array<E>, read_write>` and give us `array<E>`?

I think most often in he WGSL code, the correct type for `buffer` that references a `StorageBuffer<E>` is actually `ptr<uniform, array<E>, read_only>`. For examle if you pass `buffer` to function, WGSL will force you to pass a ptr.
...so I don't think we'd want `OpLoad` to "dereference" and give us `array<E>` in any case that I know of.

csyonghe: After all the legalization step, you shouldn't be seeing buffers being passed around through function calls. They will be legalized into global variables, and all functions will directly access that global variable, instead of load it and pass it as argument to another function.

Back to this problem, I think the solution is:
1. We still lower structured buffer types to `PTR<Array<T>>` during legalization.
2. Structured buffer elements will be accessed with the `kIROp_StructuredBufferLoad` instruction, we should keep these instructions as is during legalization.
3.  Make sure `kIROp_StructuredBufferLoad`, `IRRWStructuredBufferGetElementPtr` is handled correctly, so that we emit `buffer[0]` instead of `&buffer[0]`.

aleino-nv: > After all the legalization step, you shouldn't be seeing buffers being passed around through function calls. They will be legalized into global variables, and all functions will directly access that global variable, instead of load it and pass it as argument to another function.

To clarify, are you saying I'm missing an _existing_ legalization step like this, or to I need to implement a legalization step like this?

If it already exists, could you give me a pointer? I'd like to figure out why it's seemingly not being called.

csyonghe: oh, wait, this is to make `storage, read_write` appear after `var`.
then yeah this is a reasonable place for the logic after all.
sorry for the distraction, i think we are good with what we have.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,968 @@
...
+// This may lead to confusion (which is why this note is referenced in several
+// places), but the benefit of doing this is that the generated WGSL code is
+// simpler to generate and should be faster to compile.
+// A "terminology preserving" implementation would have to generate lots of
+// 'transpose' calls, or else perform more complicated transformations that
+// end up duplicating expressions many times.
+
+namespace Slang {
+
+void WGSLSourceEmitter::emitSwitchCaseSelectorsImpl(
```
csyonghe, line 28: you should be able to delete this now.

source/slang/slang-check-stmt.cpp
```diff
@@ -308,8 +308,11 @@ namespace Slang
         }
         else
         {
-            // TODO: need to do some basic matching to ensure the type
-            // for the `case` is consistent with the type for the `switch`...
+            // Try to coerce the case expression to the same type as the condition
+            Expr* condition = switchStmt->condition;
+            expr = coerce(CoercionSite::General, condition->type, expr);
```
csyonghe, line 313: you want to run `tryConstantFoldExpr` (line 302) after `coerce`.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -0,0 +1,968 @@
...
+    uint32_t fieldAlignment = 1U;
+    while (((fieldAlignment & (structAlignment | fieldOffset)) == 0U))
+        fieldAlignment = fieldAlignment << 1U;
+
+    m_writer->emit("@align(");
+    m_writer->emit(fieldAlignment);
+    m_writer->emit(")");
+}
+
+bool WGSLSourceEmitter::isPointerSyntaxRequiredImpl(IRInst* inst)
```
csyonghe, line 239: I don't think you need this anymore by making `doesTypeSupportPointers` to `false`.

source/slang/slang-emit-c-like.cpp
csyonghe: After the legalization, all references to structured buffers should be direct references to the global parameter, and there should not be any reference passing in the IR, so we shouldn't need to worry about case 2.

csyonghe:
Looking good, except that we probably want to just undo `emitSwitchCaseSelectorsImpl`.

source/slang/slang-emit-wgsl.cpp
aleino-nv: There is still the issue that cases sharing a block have different syntax in WGSL, [as mentioned in the description of the commit that adds this hook to the C-like emitter](https://github.com/shader-slang/slang/pull/5006/commits/432a57ae3bd7af2d521beedc1f4aef8a814d58b6):



    The C-like emitter will emit code like this:
    
        switch(a.x)
        {
    
        case 0:
        case 1:
        {
           ...
        } break;
    
        ...
    
        }
    
    This is not allowed in WGSL. Instead, selectors for cases that share a body must [1] be
    separated by commas, like this:
    
        switch(a.x)
        {
    
        case 0, 1:
        {
           ...
        } break;
    
        ...
    
        }
    
    To prepare for addressing issue #4807, this patch makes the emission of switch case
    selectors overloadable.
    
    [1] https://www.w3.org/TR/WGSL/#syntax-case_selectors

Would you handle it some other way?

csyonghe: I see, this makes sense then. Can you include this explanation in the comment?

csyonghe (dismissed):

csyonghe:
Not sure why CI is not running in this PR. Maybe we need to close this and reopen a new PR to see if that fixes the problem.

aleino-nv:
https://github.com/shader-slang/slang/pull/5025#issuecomment-2335047811

> I think we should not allow the bad cast in falcor code, but at the same time it is probably not a good idea to introduce breaking change as part of this PR.
>
> Given it would have been a breaking change to have the coercion logic in the frontend, let's not do that for now and use your original emit logic hack to workaround this issue.

@csyonghe (on original PR) Done: I dropped the front-end fix and added a patch with the WGSL-emitter workaround for now.
Once Falcor is fixed, the inverse can be done to clean up the workaround and get the front-end fix.

csyonghe (approved):

PR #5011, opened 2024-09-04
Support entrypoints defined in a namespace.

### LINKED ISSUE #2144: Namespace Functions not supported as EntryPoint
```
struct PSInput
{
	float4 color : COLOR;
};

namespace Pixel
{
	float4 Main(PSInput input) : SV_TARGET
	{
		return input.color;
	}    
}
```
In the following example, specifying `Pixel::Main` as the entry point fails to compile in Slang, but will work in HLSL compilers FXC and DXC.

csyonghe:
Closes #2144.


tests/language-feature/namespaces/entrypoint-in-namespace.slang
```diff
@@ -0,0 +1,20 @@
...
+namespace Pixel
+{
+    struct MyType
+    {
+        static float4 Main(PSInput input) : SV_TARGET
+        {
+            return input.color;
+        }
+    }    
+}
```
ArielG-NV, line 20: From what I understand this PR also allows us to define entry-points inside a `struct` (but not inside a namespace), should we add that to this test?

ArielG-NV (approved):
Code looks good to me, 1 question.

tests/language-feature/namespaces/entrypoint-in-namespace.slang
csyonghe: This test does define the entrypoint inside a struct and a namespace.

csyonghe: if things can work for function defined in a struct and a namespace, then it should also work if it is in a struct but not in a namespace, so I don't think we need a separate test for that.

ArielG-NV:
```
// My File

struct MyType
{
    static float4 Main(PSInput input) : SV_TARGET
    {
        return input.color;
    }    
}
```

PR #5014, opened 2024-09-05
Remove unused package in primary benchmark script

venkataram-nv:
Also fixes the machine used for running so that the benchmark results will have reduced variance.


jkwak-work (approved):
Looks good to me. If there is an issue for this, don't forget to mention it on the description.

Issue #4817, opened 2024-08-12
`reinterpret` with structs containing `uint64_t` or `uint8_t` incorrectly throws “T contains fields that cannot be packed into AnyValue” error


natevm:
I’m finding that with a recent build of Slang, I can no longer use `reinterpret`. Even for very simple cases, I get this error: `cannot be packed into AnyValue`

I seem to be able to repro with something like:

struct A {
    uint64_t data;
}

struct B {
    uint64_t data;
}

A a;
B b = reinterpret<B>(a);

natevm:
I’ve also tried having both of these types extend a common interface but I get the same error.

This is with the slang->SPV backend. Perhaps it has something to do with me using the scalar struct layout flag? 🤔

natevm:
Looks like it has something to do with structs containing "uint64_t". If I use "uint", then I don't get this compilation error.

natevm:
Looks like I also get this issue with reinterprets involving uint8_t.

natevm:
Looking into this a bit, it appears that the logic in the marshalling code supports 16-bit and 32-bit types, but not 8-bit or 64-bit. 

https://github.com/shader-slang/slang/blob/master/source/slang/slang-ir-any-value-marshalling.cpp#L328-L338

https://github.com/shader-slang/slang/blob/master/source/slang/slang-ir-any-value-marshalling.cpp#L514-L520

https://github.com/shader-slang/slang/blob/b2ca2d5a4efeae807d3c3f48f60235e47413b559/source/slang/slang-ir-any-value-marshalling.cpp#L699-L702

natevm:
If it's any help, here's the current reproducer I'm using:

```
RWStructuredBuffer<uint64_t> result1;
RWStructuredBuffer<uint8_t> result2;
[shader("compute")]
[numthreads(1, 1, 1)]
void main()
{
    uint8_t[8] bytes = { 1, 2, 3, 4, 5, 6, 7, 8 };
    uint64_t i64 = reinterpret<uint64_t>(bytes);
    result1[0] = i64;

    uint8_t[8] backToBytes = reinterpret<uint8_t[8]>(i64);
    result2[0] = backToBytes[0];
    result2[1] = backToBytes[1];
    result2[2] = backToBytes[2];
    result2[3] = backToBytes[3];
    result2[4] = backToBytes[4];
    result2[5] = backToBytes[5];
    result2[6] = backToBytes[6];
    result2[7] = backToBytes[7];
}
```

[...]

PR #5020, opened 2024-09-06
Promote `bitfieldExtract` and `bitfieldInsert` to become Slang intrinsics

natevm:
For context, see the related [issue here](https://github.com/shader-slang/slang/issues/4817)

In short, bitfield insertion and extraction are useful tools for reinterpreting structures where types have differing sizes, as they allow for simpler and more hardware-efficient instructions for byte manipulation.

Before this change, `bitfieldExtract` and `bitfieldInsert` were only supported outside of the compiler, when using `-allow-glsl`. 

Here, I'm introducing `KIROp_BitfieldExtract` and `KIOROp_BitfieldInsert` instructions, which by default generate the original fallback bitfield manipulation logic but in CLikeSourceEmitter.

Then, when emitting SPIR-V or GLSL, these compilation paths override this default logic to instead tap into more specialized intrinsics.


source/slang/core.meta.slang
```diff
@@ -2193,6 +2193,19 @@ __generic<T, U>
 __intrinsic_op($(kIROp_Reinterpret))
 T reinterpret(U value);
 
+// Bitfield extract / insert
+__generic<T>
+[__readNone]
+[__unsafeForceInlineEarly]
+__intrinsic_op($(kIROp_BitfieldInsert))
+    T bitfieldInsert(T base, T insert, int offset, int bits);
```
csyonghe, line 2201: leading space.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3667,6 +3677,167 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
...
+            m_writer->emit("(1u <<");
+            emitOperand(bts, getInfo(EmitOp::General));
+            m_writer->emit(") - 1");
+        }
+        m_writer->emit(")");
+    }
+    m_writer->emit(")");
+}
+
+void CLikeSourceEmitter::emitBitfieldInsertImpl(IRInst* inst) {
```
csyonghe, line 3764: brace in its own line.

csyonghe (commented):
This needs to come with a test,

natevm:
> This needs to come with a test.

Agreed. Looks like the current test suite is passing which is good. I’ll see if I can write something up which tests on a couple different target backends.

natevm:
@csyonghe What's the correct approach to constructing vectors for the CUDA backend? 

bitfieldExtract is a component-wise intrinsic, so I need to support a function like this as a right hand side which can be assigned to the left:

```
uint4 bfe(uint val, int off, int bits) {
    return ((val >> off) & ((1u << bits) - 1));
}
```

However, every time I try to construct an int4 from `off`, I get errors similar to the following:
```
nvrtc 11.2: (25): error : no suitable constructor exists to convert from "int" to "int4"
nvrtc 11.2: (25): error : no suitable constructor exists to convert from "unsigned int" to "int4"
nvrtc 11.2: (25): error : no suitable constructor exists to convert from "int" to "int4"
nvrtc 11.2: (25): error : no suitable constructor exists to convert from "int" to "int4"
```

I've tried constructing an int4 as `int4(off, off, off, off)` but this does not work either. 

I could try emitting a `make_int4` as a special case for CUDA, but this doesn't generalize to uint8_t, uint16_t, or uint64_t.

natevm:
For the time being, looks like a special case lookup table works for CUDA.

natevm:
On adding these tests, I noticed that the current logic in glsl.meta.slang for bitfield extraction in the signed integer case was incorrect. For signed integers, the extracted bits are assumed to be sign-extended. 

The current logic now correctly accounts for this, and I have some explicit tests to back that up now too.

tests/language-feature/bitfield/bitfield-extract-i32.slang
```diff
@@ -0,0 +1,56 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-cpu -output-using-type
```
csyonghe, line 1: We need to cover all the other targets, with this line duplicated where `-cpu` is replaced with `-vk`, `-d3d`, `-metal` and `-cuda`.

natevm: I'll add these in and see if the existing tests pass. We might also want to consider adding overrides for metal and cuda if these languages have direct bitfield extraction intrinsics. 

For CUDA, at least for scalars, we could be using the following PTX:
```
unsigned int bitfieldExtract(unsigned int val, int pos, int len) { 
    unsigned int r; asm("bfe.u32 %0, %1, %2, %3;" : "=r"(r) : "r"(val), "r"(pos), "r"(len)); 
    return r; 
}
unsigned int bitfieldInsert(unsigned int src, unsigned int dst, int pos, int len) { 
    unsigned int r; 
    asm("bfi.b32 %0, %1, %2, %3, %4;" : "=r"(r) : "r"(src), "r"(dst), "r"(pos), "r"(len)); 
    return r; 
}
```
We'd need to add a `.s32` for the sign extension case, but that's not bad. I can look into a solution for Metal as well, but hopefully in the meantime the c-like fallback will work there too.

natevm:
Looks like there are some newly failing tests on my end after adding these flags for the other targets. I'll see if I can figure out what's going on.

natevm:
I'm a bit confused with the Vulkan spec... I think I might have found a typo?...

According to the spec:
> The Base operand of any OpBitCount, OpBitReverse, **OpBitFieldInsert**, **OpBitFieldSExtract**, or **OpBitFieldUExtract** instruction must be a 32-bit integer scalar or a vector of 32-bit integers.

However, at least with NVIDIA, I'm finding that the tests here reveal that 64-bit integers are supported, and do give correct results. 

So I'm wondering then if I'm missing a SPIR-V extension or something...

natevm:
It appears the two outstanding issues are: 
1. When `SLANG_RUN_SPIRV_VALIDATION=1`, SPIR-V validation does not correctly consider how `SpvCapabilityInt64` extends the capabilities of `OpBitfieldSExtract`/`OpBitfieldUExtract`/`OpBitfieldInsert` intrinsics. This is causing the one windows CI build to fail. Setting `SLANG_RUN_SPIRV_VALIDATION=0` and ignoring the validation errors, I get correct results. 

2. For Metal, it appears that when targetting release mode, Slang appears to be translating my `vec<int64_t, 4>` Metal type into `int64_t4`, which is not an existing type in the Metal shading language. 

For 1, as far as I know, this is an issue on Khronos' end. I'm not sure what policy is here with handling ambiguities in the SPIR-V spec, but my thoughts are that I should test to see how 64-bit / 16-bit / 8-bit integers interact with bitfield intrinsics on an Intel ARC card and then on an AMD card. I do have an A770 and an AMD 6750XT, and could setup a test bench. 

For 2, this seems like a bug with the current Slang Metal backend, and isn't directly tied to my changes as far as I can tell. 

@csyonghe For 1, what are your thoughts? Do we have a process for reporting issues in SPIR-V validation? And then for 2, do you know what's causing the release build on MacOS to fail, but the debug build on MacOS to pass?

csyonghe:
Only release build runs the full suite of tests, debug build doesn't run most tests.

natevm:
> You can add -skip-spirv-validation in the //TEST line to disable validation for that specific test. You can search for it in the codebase to see how it is used elsewhere for similar situations.

I did try to use this flag like so:
`//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -skip-spirv-validation`

But this throws an error that `-skip-spirv-validation` is not a valid command line argument. Perhaps it's because I'm using this `TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK)` pattern?

The other test using this flag follows a different pattern:
`//TEST:SIMPLE(filecheck=SPIRV):-target spirv -entry computeMain -stage compute -emit-spirv-directly -skip-spirv-validation`

csyonghe:
For compute tests, you need -slang -skip-spirv-validation.

natevm:
> For 2, this seems like a bug with the current Slang Metal backend, and isn't directly tied to my changes as far as I can tell.
> 
> @csyonghe For 2, do you know what's causing the release build on MacOS to fail with `vec4(int64_t, N)`, but the debug build on MacOS to pass?

@csyonghe looks like this is the last major issue at the moment. I don't have a Mac, so I can't reproduce this locally to debug it.

csyonghe:
You should be able to fix this by changing the way we emit vector types when the element type is 64bit int in slang-emit-metal.cpp, look for kIROp_VectorType in that file to find the place to emit the type.

You can test this locally on a windows machine by adding a 
//TEST:SIMPLE:-target metal
On the offending test file to trigger a compile only test with the source, that will run the metal compiler on windows to give you any metal compile errors without actually running the shader.

natevm:
@csyonghe looks like all relevant tests are passing. 

There’s one windows CI machine which is reporting that bash is missing, but I see the same error on ToT, so I’m assuming that’s okay to see.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3704,6 +3714,233 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
...
+    bool isSigned; 
+    int bitWidth;
+    if (!tryGetIntInfo(elementType, isSigned, bitWidth)) 
+    {
+        SLANG_DIAGNOSE_UNEXPECTED(getSink(), SourceLoc(), "non-integer element type given to bitfieldExtract");
+        return;
+    }
+
+    String one;
+    switch(bitWidth) {
```
natevm, line 3813:
```suggestion
    switch(bitWidth)
    {
```

source/slang/slang-emit-glsl.cpp
```diff
@@ -2449,6 +2449,28 @@ void GLSLSourceEmitter::emitFuncDecorationImpl(IRDecoration* decoration)
     }
 }
 
+void GLSLSourceEmitter::emitBitfieldExtractImpl(IRInst* inst) {
```
natevm, line 2452:
```suggestion
void GLSLSourceEmitter::emitBitfieldExtractImpl(IRInst* inst)
{
```

source/slang/slang-emit-glsl.cpp
```diff
@@ -2449,6 +2449,28 @@ void GLSLSourceEmitter::emitFuncDecorationImpl(IRDecoration* decoration)
...
+    m_writer->emit("bitfieldExtract(");
+    emitOperand(inst->getOperand(0), getInfo(EmitOp::General));
+    m_writer->emit(",");
+    emitOperand(inst->getOperand(1), getInfo(EmitOp::General));
+    m_writer->emit(",");
+    emitOperand(inst->getOperand(2), getInfo(EmitOp::General));
+    m_writer->emit(")");
+}
+
+void GLSLSourceEmitter::emitBitfieldInsertImpl(IRInst* inst) {
```
natevm, line 2462:
```suggestion
void GLSLSourceEmitter::emitBitfieldInsertImpl(IRInst* inst)
{
```

source/slang/slang-emit-spirv.cpp
```diff
@@ -5594,6 +5600,43 @@ struct SPIRVEmitContext
         return emitInst(parent, inst, SpvOpConvertUToPtr, inst->getFullType(), kResultID, inst->getOperand(0));
     }
 
+    SpvInst* emitBitfieldExtract(SpvInstParent* parent, IRInst* inst) {
```
natevm, line 5603:
```suggestion
    SpvInst* emitBitfieldExtract(SpvInstParent* parent, IRInst* inst)
    {
```

source/slang/slang-emit-spirv.cpp
```diff
@@ -5594,6 +5600,43 @@ struct SPIRVEmitContext
...
+
+        // NM: technically, using bitfield intrinsics for anything non-32-bit goes against 
+        // VK specification: VUID-StandaloneSpirv-Base-04781. However, it works on at least 
+        // NVIDIA HW. 
+        SpvOp opcode = i.isSigned ? SpvOpBitFieldSExtract : SpvOpBitFieldUExtract;
+        return emitInst(parent, inst, opcode, inst->getFullType(), kResultID, 
+                    inst->getOperand(0), inst->getOperand(1), inst->getOperand(2));
+    }
+
+    SpvInst* emitBitfieldInsert(SpvInstParent* parent, IRInst* inst) {
```
natevm, line 5620:
```suggestion
    SpvInst* emitBitfieldInsert(SpvInstParent* parent, IRInst* inst)
    {
```

natevm (commented):
Quick pass over braces

source/slang/slang-emit-c-like.cpp
```diff
@@ -3704,6 +3714,234 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
...
+                case kIROp_UInt16Type: m_writer->emit("ushort"); break;
+                case kIROp_UIntType:   m_writer->emit("uint"); break;
+                case kIROp_UInt64Type: m_writer->emit("ulonglong"); break;
+                default: SLANG_ABORT_COMPILATION("Unhandled type emitting CUDA vector");
+            }
+
+            m_writer->emitRawText(std::to_string(N).c_str());
+        }
+        // Special handling required for Metal target
+        else if (isMetalTarget(getTargetReq()))
```
csyonghe, line 3756: We shouldn't need this, and instead we should make sure the metal backend can emit all vector types correctly.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3704,6 +3714,234 @@ void CLikeSourceEmitter::emitFuncDecorationsImpl(IRFunc* func)
...
+    Slang::IROp type = elementType->getOp();
+    if (!(type >= kIROp_Int8Type && type <= kIROp_UInt64Type)) return false;
+    isSigned = (type >= kIROp_Int8Type && type <= kIROp_Int64Type);
+
+    Slang::IROp stype = (isSigned) ? type : Slang::IROp(type - 4);
+    bitWidth = 8 << (stype - kIROp_Int8Type);
+    return true;
+}
+
+void CLikeSourceEmitter::emitVecNOrScalar(IRVectorType* vectorType, std::function<void()> emitComponentLogic)
```
csyonghe, line 3728: naming suggestion: `maybeEmitMakeVectorFromScalar`. Instead of doing call-back style with std::function (which is a virtual function call), just use `bool maybeEmitMakeVectorFromScalar(IRType* type);` and `void maybeCloseMakeVectorFromScalar(bool v)`.

source/slang/slang-emit-metal.cpp
```diff
@@ -538,7 +538,8 @@ bool MetalSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inO
 
 void MetalSourceEmitter::emitVectorTypeNameImpl(IRType* elementType, IRIntegerValue elementCount)
 {
-    emitSimpleTypeImpl(elementType);
+    // NM: Passing count here, as Metal 64-bit vector type names do not match their scalar equivalents.
+    emitSimpleTypeKnowingCount(elementType, elementCount);
```
csyonghe, line 542: I think you can special case the 64bit types in the switch statement below, and emit `vec<T, N>` instead of `TN` for those types. We shouldn't need the `emitSimpleTypeKnowingCount()` change.

natevm:
Apologies on the delay here. Been a bit tied up with a move between states. I should have a bit more time to work on this in the upcoming week.

natevm:
@kaizhangNV Going to try digging into this today. Sounds like there's not much left needed to get this done.

kaizhangNV:
It looks like the main concern for metal is addressed already.
The only thing remaining is https://github.com/shader-slang/slang/pull/5020#discussion_r1761623703

kaizhangNV:
@csyonghe This PR is in a good shape. In the latest change, I see Nate already addressed the special handling for metal backend. Do you have any other concerns except the naming suggestion? I feel it's not worth for me to take over this as I probably will have to create a new PR.

csyonghe:
@kaizhangNV The remaining work here is to make sure it works on WGPU. @natevm do you have time for this or do you want @kaizhangNV to take over?

csyonghe:
@kaizhangNV has already landed this change in another PR, so closing this one.

PR #5023, opened 2024-09-06
Look through attributes and rates when determining by reference initialization

### LINKED ISSUE #5022: groupshared types emit false uninitialized warnings
for example

```
groupshared int myInt;

void set(__ref groupshared int i){ i = 1; }

void main(){
  set(myInt);
  use(myInt);
}
```

Both set and used lines generate uninitialized global warnings.

This is because the uninitialized checker's `checkCallUsage` doesn't correctly look through rate qualified types.

expipiplus1:
Closes #5022


jkwak-work (approved):
Looks good to me

source/slang/slang-ir.cpp
```diff
@@ -7869,6 +7869,21 @@ namespace Slang
         return type;
     }
 
+    IRType* IRInst::getUnratedAndUnattributedType()
```
csyonghe, line 7872: there is already a `unwrapAttributedType` defined in `slang-ir-util.h` that does exactly this.

expipiplus1: ah, I missed that searching for IRType->IRType functions, thanks

source/slang/slang-ir-util.cpp
```diff
@@ -175,7 +175,9 @@ IRInst* maybeSpecializeWithGeneric(IRBuilder& builder, IRInst* genericToSpecaili
 // Returns true if is not possible to produce side-effect from a value of `dataType`.
 bool isValueType(IRInst* dataType)
 {
-    dataType = getResolvedInstForDecorations(unwrapAttributedType(dataType));
+    if(auto t = as<IRType>(dataType))
```
csyonghe, line 178: This if is unnecessary.

expipiplus1: I changed the type of unwrapAttributedType to be on IRType, so it's required now.

csyonghe: That means we are checking the opcode twice, once for the dynamic cast and once for the actual unwrapping, which seems unnecessary.

csyonghe: We should feel comfortable to accept IRInst where a type is expected, and rely on static cast to convert between IRInst and IRType because there is not a clean way to separate a type and an inst anyways. IRType has no more meaning other than clarifying what an IRInst is for in code, but for unwrapAttributeType, it is verify clear what the param is, so it accepting IRInst is not wrong and I don't see a need for the change.

expipiplus1: ok, I've reverted that change

csyonghe (approved):

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -265,7 +265,7 @@ namespace Slang
 
         // Consider it as a store if its passed
         // as an out/inout/ref parameter
-        auto type = as<IRType>(unwrapAttributedType(ftype->getParamType(index)));
+        auto type = cast<IRType>(unwrapAttributedType(ftype->getParamType(index)));
```
csyonghe, line 268: `cast` will throw if the cast failed. should use `as` here.

csyonghe (changes requested):

source/slang/slang-ir-use-uninitialized-values.cpp
expipiplus1: can just remove the whole thing actually :)

csyonghe (approved):

PR #5025, opened 2024-09-06
Initial WGSL support

aleino-nv:
(Duplicate of https://github.com/shader-slang/slang/pull/5006 to see if that fixes CI issues)

This adds initial support for WGSL output.
With this, the majority of the shaders in tests/compute will output WGSL that compiles.


source/compiler-core/slang-artifact.h
```diff
@@ -143,6 +143,7 @@ enum class ArtifactPayload : uint8_t
     CUDA,           ///< CUDA source
     Metal,          ///< Metal source
     Slang,          ///< Slang source
+    WGSL,           ///< WGSL source
```
csyonghe, line 146: This needs to be placed at the end to avoid breaking binary compatibilty I think.

csyonghe:
It seems that the switch coercion change causes breakage in Falcor, since they are doing
```
switch(uint)
{
case EnumCase:
    ...
}
```

which is bad and I would rather have this breaking change.

However we will need to change Falcor code to make the test to pass.

csyonghe:
I think we should not allow the bad cast in falcor code, but at the same time it is probably not a good idea to introduce breaking change as part of this PR. 

Given it would have been a breaking change to have the coercion logic in the frontend, let's not do that for now and use your original emit logic hack to workaround this issue.

aleino-nv:
The [original PR](https://github.com/shader-slang/slang/pull/5006) (with all review comments) is now passing the checks. Closing this one.

PR #5040, opened 2024-09-09
split slang-test from build in ci yaml

jkwak-work:
This PR creates three jobs in ci.yml.
1. load matrix_*.json.
2. spawn instances for build jobs.
3. spawn instances for slang-test jobs.

TODO: the first job, "load_matrix", is no longer necessary.
Initially it was to share the matrix values for both build and test.
But now the matrix values diverged and each job got its own matrix values.
The matrix values in matrix_*.json can be inlined in ci.yml now.

This PR adds build instance for AARCH64-debug.
Before this PR, we haven't been building nor testing for debug build of AARCH64.
slang-test is still disabled but we can take the advantage of free runners for the build checks.


.github/workflows/ci.yml
```diff
@@ -115,8 +81,71 @@ jobs:
...
+        with:
+          name: test-${{matrix.os}}-${{matrix.platform}}-${{matrix.compiler}}-${{matrix.config}}
+          path: build/${{ env.CONFIG_CAMEL }}
+
+      - name: "[Debug] list files"
+        run: ls -R build
+      
       - name: Test Slang
         run: |
+          chmod a+x "$bin_dir/"*
```
expipiplus1, line 148: I'm very surprised that this is necessary, can we at least have a comment?

.github/workflows/ci.yml
```diff
@@ -15,85 +15,51 @@ on:
...
-          - os: windows
-            config: release
-            compiler: cl
-            platform: x86_64
-            test-category: full
-            full-gpu-tests: true
-            runs-on: [Windows, self-hosted]
-            has-gpu: true
       fail-fast: false
+      matrix: ${{ fromJson(needs.load_matrix.outputs.matrix_build) }}
```
expipiplus1, line 42: I think it's probably neater to have this inline, because then we can at least have comments. Is it possible to do the GITHUB_OUTPUT thing without having that file separate?

.github/workflows/ci.yml
```diff
@@ -115,8 +81,71 @@ jobs:
...
+          name: test-${{matrix.os}}-${{matrix.platform}}-${{matrix.compiler}}-${{matrix.config}}
+          if-no-files-found: error
+          retention-days: 1
+          path: |
+            build/${{ env.CONFIG_CAMEL }}
+            !**/*.pdb
+            !**/*.ilk
+            !**/*.lib
+            !**/*.exp
+            !**/*.a
```
expipiplus1, line 100: The files to upload should already be captured by the pack step from cmake,

.github/workflows/ci.yml
```diff
@@ -115,8 +81,71 @@ jobs:
...
+
+      - uses: ./.github/actions/common-setup
+        with:
+          os: ${{matrix.os}}
+          compiler: ${{matrix.compiler}}
+          platform: ${{matrix.platform}}
+          config: ${{matrix.config}}
+
+      - name: Set capitalized config
+        run: echo "CONFIG_CAMEL=$(echo ${{ matrix.config }} | sed 's/debug/Debug/;s/release/Release/')" >> $GITHUB_ENV
```
expipiplus1, line 135: Might be easier to set this from common-setup/action.yml, can just export the already made `Config` variable (although I don't know if windows env is case sensitive?)

.github/workflows/ci.yml
jkwak-work: Without this line, the executables don't have "x" permission on Linux based platforms which are all but Windows.
I think this is due to the fact that the artifact uploader doesn't preserve the file permissions.

.github/workflows/ci.yml
jkwak-work: As I explained on the description of the PR, I am planning to delete the new files, "matrix_build.json" and "matrix_test.json".
I will inline them in the CI.yml.
Because json file doesn't allow us to leave any comments, I think it is better to keep them in YAML file as it has been.

.github/workflows/ci.yml
jkwak-work: I was thinking to put it in the `matrix` like the following,
```
include:
  { config: debug, config_camel: Debug }
  { config: Release, config_camel: Release }
```
Then I wouldn't need to spawn a bash only to handle the string data.

jkwak-work:
Closing the PR until it is ready; or until I need to test CI again.

PR #5041, opened 2024-09-09
Fix inccorect dropping of declref during Unification of DeclaredSubtypeWitness.

### LINKED ISSUE #5033: Inferring generic value parameters fails for generic constraints
It looks like generic value parameters can't be inferred if they are part of a type constraint that is generic. This means extensions don't work when these kinds of type constraints are used.

Say I have an interface with value parameter:
```
interface IValueGeneric<let D : int> {}
struct ValGenericImpl<let D : int> : IValueGeneric<D> {}
```
And something that takes an implementation of `IValueGeneric`:
```
struct NestedValueGeneric<let D : int, S : IValueGeneric<D>>
{
}
```
I now can't pass `NestedValueGeneric` around:
```
void acceptor<let D : int, S : IValueGeneric<D>>(NestedValueGeneric<D, S> x)
{
}
void test(NestedValueGeneric<2, ValGenericImpl<2>> x)
{
    // could not specialize generic for arguments of type (NestedValueGeneric<2, ValGenericImpl<2>>)
    acceptor(x);
}
```
In this case, I can work around this by setting `D` explicitly, via `acceptor<2>(x)`.

However, where I can't work around this is with extensions: Extensions involving a value generic constraint won't work
```
__generic<let D : int, S : IValueGeneric<D>>
extension NestedValueGeneric<D, S>
{
    void foo();
}
void test2(NestedValueGeneric<2, ValGenericImpl<2>> x)
{
    // 'foo' is not a member of 'NestedValueGeneric<2, ValGenericImpl<2>>'
    x.foo();
}
```
Interestingly, this does work if `D` is a type parameter instead of a value parameter.

csyonghe:
Closes #5033.

### DIFF
--- a/source/slang/slang-check-constraint.cpp
+++ b/source/slang/slang-check-constraint.cpp
@@ -779,8 +779,8 @@ namespace Slang
                 SLANG_ASSERT(constraintDecl2);
                 return TryUnifyTypes(constraints,
                     unifyCtx,
-                    constraintDecl1.getDecl()->getSup().type,
-                    constraintDecl2.getDecl()->getSup().type);
+                    getSup(m_astBuilder, constraintDecl1),
+                    getSup(m_astBuilder, constraintDecl2));
             }
         }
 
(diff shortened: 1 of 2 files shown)


saipraveenb25 (approved):
Looks good to me, but it looks like #5044 has duplicate commits?

Issue #5035, opened 2024-09-06
Support synthesizing generic adaptor methods for interface requirements.


tunabrain:
There seems to be a corner case issue with generic return values in interfaces.

Let's pretend I had an interface like this:
```
interface IStack<let D : int>
{
    IStack<D - 1> pop();
}
struct StackImpl<let D : int> : IStack<D>
{
    StackImpl<D - 1> pop() { /* ... */ }
}
```
This works. However, this does not:
```
interface IStack<let D : int>
{
    __generic<let N : int>
    IStack<D - N> popN();
}
struct StackImpl<let D : int> : IStack<D>
{
    // member 'popN' does not match interface requirement.
    __generic<let N : int>
    StackImpl<D - N> popN() { /* *** */ }
}
```
Is it intended to begin with that a `pop` return type of `StackImpl<D-1>` satisfies the interface requirement of a `IStack<D-1>` return type?
If so, is it possible to support the `IStack<D-N>` case as well?

csyonghe:
Seems like a bug that needs to be fixed.

csyonghe:
The problem here is that the implementation method 
```
__generic<let N : int>
    StackImpl<D - N> popN() { /* *** */ }
```
does not match exactly the requirement:
```
__generic<let N : int>
    IStack<D - N> popN();
```

Because `StackImpl<D-N>` is not the same type as `IStack<D-N>`.
Normally if this is not a generic method, slang can synthesize a wrapper to satisfy the requirement. However since `popN` is a generic method, we currently don't have the logic to synthesize the wrapper and you will need to manually provide one, or simply modify the implementation method to return `IStack<D-N>`.

Doing that revealed another issue that the compiler cannot properly specailize functiosn that returns an interface type to functions that return a concrete type even if the return type is known at compile time.
PR #5044 addresses this issue and can produce correctly specialized code.

PR #5044, opened 2024-09-09
Specialize existential return types when possible.

csyonghe:
Paritally addresses #5035.

This PR enhances our specialization logic so that functions that returns an existential value can now be specailized into a function that returns a concrete type. This avoid us generating unnecessary dynamic dispatch logic when the return type of the function is statically known.


saipraveenb25 (approved):
Looks good to me!

Issue #5034, opened 2024-09-06
Extending __subscript with set fails


tunabrain:
It looks like it's not possible to extend `__subscript` to add e.g. a `set` portion conditionally.

Say I had the following interfaces for abstracting things that allow reading/writing at an index:
```
interface Readable<T>
{
    T read(int i);
}
interface Writable<T> : Readable<T>
{
    void write(int i, T t);
}
```
For convenience I want to support `[]` access that internally calls read/write. I can't put code in the interface, but I can use an extension like this:
```
__generic<S, T : Readable<S>>
extension T
{
    __subscript(int i)->S
    {
        get { return read(i); }
    }
}
__generic<S, T : Writable<S>>
extension T
{
    __subscript(int i)->S
    {
        set { return write(i, newValue); }
    }
}
```
This works for reading:
```
T readAtZero<T>(Readable<T> x)
{
    return x[0];
}
```
But not writing:
```
void writeAtZero<T>(Writable<T> x)
{
    // ambiguous call to 'operator[]' with arguments of type (int)
    x[0] = 1.0f;
}
```
I think this is because `__subscript` as a whole is treated as a function, and not `get` / `set` individually. The extension then adds a competing `__subscript` rather than supplying a complementary `set` method.

Extensions involving `__subscript` are a useful pattern, is it possible to support this case?

csyonghe:
Yes, it should be possible to support this.

PR #5046, opened 2024-09-10
Disambiguate subscript decls by preferring the one with more accessors.

csyonghe:
Closing #5034.

Directly allowing defining `set` and `get` in different subscript decls and allow deferred overload resolution is tricky to support. However it is trivial to add a disambiguation rule to prefer the one with more complete accessors. This should allow example like #5034 to work, but in a slightly different way:

```
//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-slang -compute -shaderobj -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-vk -compute -shaderobj -output-using-type

// Test that we can disambiguiate subscript decls by prefering the candidate that contains a super set of
// accessors than the other candidates.
interface IBuf<T>
{
    T read(int x);
}
interface IRWBuf<T> : IBuf<T>
{
    [mutating]
    void write(int x, T v);
}

extension<T, U : IBuf<T>> U
{
    __subscript(int x) -> T { get { return read(x); } }
}

extension<T, U : IRWBuf<T>> U
{
  // This would have introduced ambiguity, however by providing both getter and setter
  // here, we can make use the disambiguation logic to prefer this one.
    __subscript(int x)->T { get { return read(x); } set { write(x, newValue); } }
}

struct MyArray<T> : IRWBuf<T>
{
    T data[4];
    T read(int x) { return data[x]; }
    [mutating]
    void write(int x, T v) { data[x] = v; }
}


[numthreads(1,1,1)]
void computeMain()
{
    MyArray<int> arr = {};
    arr[0] = 1;
    arr[1] = 2;
    // CHECK: 1
    // CHECK: 2
    outputBuffer[0] = arr[0];
    outputBuffer[1] = arr[1];
}

```

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -1199,6 +1199,8 @@ namespace Slang
         return parent;
     }
 
+    // Returns -1 if left is preferred, 1 if right is preferred, and 0 if they are equal.
+    //
     int SemanticsVisitor::CompareLookupResultItems(
         LookupResultItem const& left,
         LookupResultItem const& right)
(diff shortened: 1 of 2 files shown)


saipraveenb25 (commented):
I agree that this works for the read-only/RW buffer scenario, but it feels kind of fragile.

Is there a specific reason why we can't prefer the subscript method in the more derived interface?

csyonghe:
The interfaces may not have any subclass relationship, but you may still expect this work:
```
extension MyType
{
      __subscript {get;set;}
}
extension MyType
{
      __subscript{get;}
}
```

PR #5048, opened 2024-09-10
Precompiled SPIR-V import support

### LINKED ISSUE #4992: SPIR-V library support for import decorations on function declarations
To have linkable spir-v libraries, it will be necessary to allow "import"s.  Any IR lowered to SPIR-V which contains function declarations without definitions should be tagged with import linkage attributes.

cheneym2:
Adds appropriate linkage and function declaration syntax for SPIR-V functions that are declared, to be imported from another SPIR-V module.

Unlike DXIL, stripping the Slang IR for a function down to a declaration requires retaining a block of parameters, as the function declaration must be emitted to SPIR-V with the same parameters as a definition. Because that thwarts the logic in Slang to tell the difference between a declaration and definition, and explicit decoration is introduced to explicitly mark functions which need to be treated as declarations during emit phase.

Fixes #4992


source/slang/slang-ir.cpp
```diff
@@ -8766,6 +8766,7 @@ namespace Slang
         switch (val->getOp())
         {
         case kIROp_Func:
+            if (val->findDecoration<IRDownstreamModuleImportDecoration>() != nullptr) return false;
```
csyonghe, line 8769: return in its own line.

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -177,11 +178,29 @@ void removeAvailableInDownstreamModuleDecorations(CodeGenTarget target, IRModule
                     // Gut the function definition, turning it into a declaration
                     for (auto inst : funcInst->getChildren())
                     {
+                        // For HLSL, we can delete the whole block
```
csyonghe, line 181: for (auto block : funcInst->getBlocks())

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -177,11 +178,29 @@ void removeAvailableInDownstreamModuleDecorations(CodeGenTarget target, IRModule
...
                         if (inst->getOp() == kIROp_Block)
                         {
-                            toRemove.add(inst);
+                            if (target == CodeGenTarget::HLSL)
+                            {
+                                toRemove.add(inst);
+                            }
+                            else if (target == CodeGenTarget::SPIRV)
+                            {
+                                for (auto blockInst : inst->getChildren())
```
csyonghe, line 192: You cannot modify the children while iterating it with getChildren.
need to call `inst->getModifiableChildren()`.

See https://github.com/shader-slang/slang/blob/master/docs/design/ir.md#ir-global-and-hoistable-value-deduplication

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -177,11 +178,29 @@ void removeAvailableInDownstreamModuleDecorations(CodeGenTarget target, IRModule
                     // Gut the function definition, turning it into a declaration
                     for (auto inst : funcInst->getChildren())
                     {
+                        // For HLSL, we can delete the whole block
+                        // but for SPIR-V we have to keep the block with only the
+                        // parameters inside, no other instructions.
                         if (inst->getOp() == kIROp_Block)
                         {
-                            toRemove.add(inst);
+                            if (target == CodeGenTarget::HLSL)
```
csyonghe, line 186: Why can't HLSL be handled by the same SPIRV path? I would think that HLSL emitter also needs to know the parameters to be able to emit the correct declaration.

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: Isn't it acceptable that I'm just adding the block instructions to a list of things, and then I remove them all at the end?

source/slang/slang-ir-redundancy-removal.cpp
cheneym2: That hasn't been an issue so far, so it's confusing to me.  HLSL emission manages to preserve the function parameters even when deleting all function blocks.  I don't understand how it's able to.  The code that recognized empty functions and emitted HLSL function declarations existed already and I have yet to see how it worked.

Current code still produces desirable declarations for HLSL:

```
int normalFunc_0(int _0);
int normalFuncUsesGeneric_0(int _0);
```

source/slang/slang-ir-redundancy-removal.cpp
csyonghe: Ah you are right that works too.

source/slang/slang-ir-redundancy-removal.cpp
csyonghe: Can we unify the two cases a bit by having spirv emit logic use the func type to emit params too?

We want to have some consistency in our IR in terms of what it means for all backends.

cheneym2: It should be doable.  I see emitFuncDecl (c-like) as the example of what HLSL does.
I'm confused why function declarations in SPIR-V require OpFunctionParameters.  It seems redundant to the OpTypeFunction which is already a field of the OpFunction itself.

csyonghe: That's probably to allow you to put decorations on the parameters.

source/slang/slang-emit-spirv.cpp
```diff
@@ -2631,14 +2632,75 @@ struct SPIRVEmitContext
...
         //
-        m_sink->diagnose(irFunc, Diagnostics::internalCompilerError);
-        SLANG_UNEXPECTED("function declaration in SPIR-V emit");
-        UNREACHABLE_RETURN(nullptr);
+        auto funcType = irFunc->getDataType();
+        auto paramCount = funcType->getParamCount();
+        for (UInt pp = 0; pp < paramCount; ++pp)
+        {
+            auto paramType = funcType->getParamType(pp);
+            emitOpFunctionParameter(spvFunc, nullptr, paramType);
```
csyonghe, line 2684: Need to make sure to emit necessary decorations for pointer-typed parameters as well. Need to call `maybeEmitPointerDecoration` here.

cheneym2: Looking at the implementation, it's for PhysicalStorageBuffer pointers and I'm unclear how to write a test case for that.  What function parameter would I define in Slang that corresponds to that in spir-v?
I'm currently trying to find a test case that hits this test case for function definitions to model after.

cheneym2: Nm, I believe I have a test case now.

cheneym2: Exporting this shows an "Alias" decoration applied from the function definition side.
```
public int normalFunc(int a, const float *b)
{
    return a - floor(*b);
}
```

Trying to replicate that on the import side function declaration.

source/slang/slang-emit-spirv.cpp
```diff
@@ -2628,14 +2630,76 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
...
+        // blocks in a function declaration, so we will emit the parameters
+        // as derived from the function's type.
+        //
+        auto funcType = irFunc->getDataType();
+        auto paramCount = funcType->getParamCount();
+        for (UInt pp = 0; pp < paramCount; ++pp)
+        {
+            auto paramType = funcType->getParamType(pp);
+            SpvInst* spvParam = emitOpFunctionParameter(spvFunc, nullptr, paramType);
+            maybeEmitPointerDecoration(spvParam, paramType, false, kIROp_Param);
```
cheneym2, line 2683: Added pointer decoration

csyonghe (dismissed):

csyonghe (approved):

Issue #4476, opened 2024-06-25
Unexpected ambiguous call.


tomas-davidovic:
In code like this:
```
uniform uint gData;

uint getData()
{
    return gData;
}

struct DataObtainer
{
    uint data;
    uint getData()
    {
        return data;
    }

    uint getValue()
    {
        return getData(); /// This is ambiguous call
    }
}

[numthreads(16, 1, 1)]
void testGetData(uint3 threadID: SV_DispatchThreadID)
{
    DataObtainer obtainer;
    output[0] = getData();
    output[1] = obtainer.getData();
    output[2] = obtainer.getValue();
}
```
The call to `getData()` inside `getValue()` is reported as an ambiguous call. This is incosistent with C/C++, where the compiler will automatically prefer the member function `getData()`, without needing to explicitly specify `this.getData()` or `::getData()`.

ArielG-NV:
I don't know if this is intentional behavior, but the solution should be simple. Overload resolution would just need to prefer functions with equal parents.

csyonghe:
Yes this is just an oversight and we should fix it.

PR #5065, opened 2024-09-16
Lower the priority of looking up the rank of scope

kaizhangNV:
In the previous change of #5060, we propose a way to resolve the ambiguous call when considering the scope of a function.

But this rule should be considered as a low priority than "specialized candidate", aka. we should consider more "specialized candiate" first.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -1454,7 +1454,7 @@ namespace Slang
         for (auto node = prefixDecl; node != nullptr; node = node.getParent())
         {
             Decl* key = node.getDecl();
-            uint32_t value = refPath.getCount();
+            uint32_t value = (uint32_t)refPath.getCount();
             refPath.add(key, value);
         }
 
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -1448,58 +1448,52 @@ namespace Slang
 
         DeclRef<Decl> prefixDecl = referenceSiteScope->containerDecl;
 
-        // find the common prefix decl of reference site and left
-        int leftDistance = 0;
-        int rightDistance = 0;
-        auto distanceToCommonPrefix = [](DeclRef<Decl> const& candidate, DeclRef<Decl> const& reference) -> int
+        // Hold the path from reference site to the root
+        // key: Decl node, value: distance from reference site
+        Dictionary<Decl*, uint32_t> refPath;
+        for (auto node = prefixDecl; node != nullptr; node = node.getParent())
         {
-            bool found = false;
-            int distanceFromRefToPrefix = 0;
-            DeclRef<Decl> prefixDecl = reference;
-            for (;;)
-            {
-                if (prefixDecl == nullptr && !found)
-                {
-                    return -1;
-                }
-                DeclRef<Decl> decl = candidate;
-                int distanceFromPrefixToCandidate = 0;
-                for(;;)
-                {
-                    if (decl == nullptr)
-                        break;
-
-                    // found the common prefix
-                    if (decl == prefixDecl)
-                    {
-                        found = true;
-                        break;
-                    }
-
-                    decl = decl.getParent();
-                    distanceFromPrefixToCandidate++;
-                }
-                if (found == true)
-                    return distanceFromRefToPrefix + distanceFromPrefixToCandidate;
-
-                prefixDecl = prefixDecl.getParent();
-                distanceFromRefToPrefix++;
-            }
-        };
-
-        leftDistance = distanceToCommonPrefix(left, prefixDecl);
-        rightDistance = distanceToCommonPrefix(right, prefixDecl);
-
-        if (leftDistance == rightDistance)
-        {
-            return 0;
+            Decl* key = node.getDecl();
+            uint32_t value = refPath.getCount();
+            refPath.add(key, value);
         }
 
-        if (leftDistance == -1)
-            return 1;
+        // find the common prefix decl of reference site and left
+        uint32_t leftDistance = 0;
+        uint32_t rightDistance = 0;
+        auto distanceToCommonPrefix = [](DeclRef<Decl> const& candidate, Dictionary<Decl*, uint32_t> refPath) -> uint32_t
+        {
+            uint32_t distanceToReferenceSite = 0;
+            uint32_t distanceToCandidate = 0;
 
-        if (rightDistance == -1)
-            return -1;
+            // Sanity check
+            if (candidate.getDecl() == nullptr)
+                return UINT32_MAX;
+
+            // search from candidate to root, once we found the first node in the reference path, that is the first
+            // common prefix, and we can stop searching.
+            for (auto node = candidate; node != nullptr; node = node.getParent())
+            {
+                Decl* key = node.getDecl();
+                if (refPath.tryGetValue(key, distanceToReferenceSite))
+                {
+                    break;
+                }
+                distanceToCandidate++;
+            }
+
+            // If we don't find the common prefix, there must be something wrong, return the max value.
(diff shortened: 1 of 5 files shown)


source/slang/slang-check-overload.cpp
```diff
@@ -1561,6 +1547,22 @@ namespace Slang
             auto overloadRankDiff = getOverloadRank(right->item.declRef) - getOverloadRank(left->item.declRef);
             if (overloadRankDiff)
                 return overloadRankDiff;
+
+            // We need to consider the distance of the declarations to the global scope to resolve this case:
```
csyonghe, line 1551: This need to be before `overloadRank` comparison above. Overload rank is a hack that we should always use as last resort disambiguation method.

kaizhangNV: ACK.

kaizhangNV:
In the latest change, I update the algorithm to compare the candidates by calculating distance
from reference site to declaration site via nearest common prefix
in the scope tree.

This will involve finding the common parent node of two child nodes
and how sum the distance from the common parent to the two child nodes.

kaizhangNV:
@csyonghe, this is not the same as we discussed offline. So can you please look at that to see if would agree with the rule like this?

source/slang/slang-check-overload.cpp
```diff
@@ -1479,6 +1440,70 @@ namespace Slang
...
+    {
+        if (!referenceSiteScope)
+            return 0;
+
+        DeclRef<Decl> prefixDecl = referenceSiteScope->containerDecl;
+
+        // find the common prefix decl of reference site and left
+        int leftDistance = 0;
+        int rightDistance = 0;
+        auto distanceToCommonPrefix = [](DeclRef<Decl> const& candidate, DeclRef<Decl> const& reference) -> int
```
csyonghe, line 1454: Instead of doing a O(n^2) loop, let's just put all parents of `candidate` and `reference` into a `ShortList<Decl*>`, and then reverse the lists to find the common prefix in `O(n)`.

kaizhangNV: Yea, this is just a proof of concept. Just want to make sure you agree with this idea.

kaizhangNV:
Again, we hit the issue that we should not put this logic here.
When looking up for the generic functions, it doesn't gaurantee to provide the valid candidates to function `CompareOverloadCandidates`, so it breaks the assumption again.

To be more specific, I can provide the repro shader here:

```
namespace A
{
    struct HPackedArray<let SIZE : uint>
    {
        uint data[SIZE];
    };

    static HPackedArray<Z1> mad<let Z0 : uint, let Z1 : uint>(
       HPackedArray<Z0> ip)
    {
        HPackedArray<Z1> res;
        return res;
    }
};

namespace A
{
    struct HCoopVector<let SIZE : uint>
    {
        HPackedArray<SIZE> vec_simt;
    }

    static HCoopVector<Z1> mad<let Z0 : uint, let Z1 : uint>(
        HCoopVector<Z0> ip)
    {
        HCoopVector<Z1> res;
        res.vec_simt = mad<Z0, Z1>(ip.vec_simt);  // it will fails here, it should choose above "mad"
         return res;
    }
};

```

In previous implementation, `CompareOverloadCandidates` won't find the best match, so it will leave the job to the following logic, while this PR gives the wrong answer because we assume both candidates are valid.

source/slang/slang-check-overload.cpp
csyonghe: I think it should work.

source/slang/slang-check-overload.cpp
```diff
@@ -1590,10 +1584,17 @@ namespace Slang
             //          float g(float y) { return f(y); }   // will call S::f() instead of ::f()
             //      }
             //  we will count the distance from the reference site to the declaration in the scope tree.
-
-            auto scopeRank = getScopeRank(left->item.declRef, right->item.declRef, this->m_outerScope);
-            if (scopeRank)
-                return scopeRank;
+            //  NOTE: DON'T do this for the generic function, because for the generic function compare, we only check the
+            //  template parameter/arguments, not the function parameters/arguments. This could break the assumption
```
csyonghe, line 1588: template->generic.

source/slang/slang-check-overload.cpp
```diff
@@ -1590,10 +1584,17 @@ namespace Slang
...
+            //  for this logic where we assume that all the candidates are valid.
+            //  This logic will be evaluated later when "Flavor" becomes "Func" or "Expr", because by then the type checks for
+            //  the candidate will be done.
+            if (left->flavor != OverloadCandidate::Flavor::Generic && left->flavor != OverloadCandidate::Flavor::UnspecializedGeneric)
```
csyonghe, line 1592: what if right is generic?

source/slang/slang-check-overload.cpp
```diff
@@ -1448,58 +1448,52 @@ namespace Slang
...
+            Decl* key = node.getDecl();
+            uint32_t value = refPath.getCount();
+            refPath.add(key, value);
+        }
+
         // find the common prefix decl of reference site and left
-        int leftDistance = 0;
-        int rightDistance = 0;
-        auto distanceToCommonPrefix = [](DeclRef<Decl> const& candidate, DeclRef<Decl> const& reference) -> int
+        uint32_t leftDistance = 0;
```
csyonghe, line 1462: always prefer `int` over unsigned int.

source/slang/slang-check-overload.cpp
kaizhangNV: Hmm, that is my oversight. I think the left and right should always the same flavor. Isn't that the case?

kaizhangNV: updated.

source/slang/slang-check-overload.cpp
kaizhangNV: ack.

source/slang/slang-check-overload.cpp
```diff
@@ -1557,10 +1582,32 @@ namespace Slang
...
+            //          float f(float x);
+            //          float g(float y) { return f(y); }   // will call S::f() instead of ::f()
+            //      }
+            //  we will count the distance from the reference site to the declaration in the scope tree.
+            //  NOTE: DON'T do this for the generic function, because for the generic function compare, we only check the
+            //  generic parameter/arguments, not the function parameters/arguments. This could break the assumption
+            //  for this logic where we assume that all the candidates are valid.
+            //  This logic will be evaluated later when "Flavor" becomes "Func" or "Expr", because by then the type checks for
+            //  the candidate will be done.
+            if ((left->flavor != OverloadCandidate::Flavor::Generic && left->flavor != OverloadCandidate::Flavor::UnspecializedGeneric) ||
```
csyonghe, line 1598: I feel like this needs to be && instead of ||?
Can you change it to use early return instead of nested if?
I feel like the overloadRankDiff logic down below should be guarded by the flavor check too.

csyonghe (approved):

source/slang/slang-check-overload.cpp
kaizhangNV: Why does it needs to by &&?
Either left or right is generic, we can't make the compare here anyway.

> I feel like the overloadRankDiff logic down below should be guarded by the flavor check too

I'm not very familiar with this, just did a search, it's a attribute `[OverloadRank(#)]`, if this can be applied to generic too, then yes, we should guard this. Because only the valid candidates should be in the rank compare, correctness is always the highest priority.

I will make a change to include this and add some commend.

kaizhangNV: push a new change.

csyonghe (approved):

kaizhangNV:
followup PR to issue #4476.

PR #5068, opened 2024-09-16
release-note.sh fixes

cheneym2:
Add a gh.exe search path that works with cygwin installations.

Do not bail out from the breaking changes loop when a change is found without PR#. Instead, continue to the next one. This fixes an issue where [BREAKING] changes appear in the detailed change list by not in the summary of breaking changes.


jkwak-work (approved):
Thanks for fixing the bug.

PR #5069, opened 2024-09-16
Fixing bug when writing primitives in metal mesh shaders

Dynamitos:
I have no idea how i missed this, fixing mesh shading primitive writing


csyonghe (approved):

csyonghe:
Can we get a regresion test added for mesh shader primitive writes?

Dynamitos:
Just so i understand correctly, you mean a test that checks for this change?

csyonghe:
Yes, we need to have a test that exposes the issue being fixed here.

csyonghe (approved):

PR #5070, opened 2024-09-16
Use GITHUB_TOKEN if set

jkwak-work:
We have been having an error from Github saying that API rate limit exceeded for IP. It hits the limit more often if your public IP is shared by many other collegues.

> {"message":"API rate limit exceeded for 216.228.112.22. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}

The suggested solution from GitHub is to use gh.exe tool with "auth login" arguments. It will store a token and allow us to use API more than the limit set for IP without token.

However, our cmake build doesn't use gh.exe and API is called via REST. This commit adds an extra header to the HTTP request with the infomration of the github token if the value is.

Usage: cmake.exe --preset vs2019 -DSLANG_GITHUB_TOKEN=your_token_here


expipiplus1 (commented):
What happens when one doesn't specify a token?

We might want to issue a `message(WARNING` if we detect a failure that adding a token would help with.

jkwak-work:
The token is optional.
When not specified, the default value is an empty string. I tested this case locally.

I can add the suggested message.

csyonghe (approved):

PR #5078, opened 2024-09-17
Implement math intrinsics for WGSL

jkwak-work:
This commit implements math related intrinsics and a few others for WGSL.

The implementation is based on the following doc,
 https://www.w3.org/TR/WGSL

### DIFF
--- a/tests/wgsl/math.slang
+++ b/tests/wgsl/math.slang
@@ -1,10 +1,4 @@
 //TEST:SIMPLE(filecheck=WGSL): -stage compute -entry computeMain -target wgsl
-//TEST:SIMPLE(filecheck=GLSL): -stage compute -entry computeMain -target glsl
-//TEST:SIMPLE(filecheck=GLSL_SPIRV): -stage compute -entry computeMain -target spirv -emit-spirv-via-glsl
-//TEST:SIMPLE(filecheck=SPIR): -stage compute -entry computeMain -target spirv -emit-spirv-directly
-//TEST:SIMPLE(filecheck=HLSL): -stage compute -entry computeMain -target hlsl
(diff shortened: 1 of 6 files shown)


csyonghe (approved):
Looks good!

tests/wgsl/math.slang
```diff
@@ -0,0 +1,285 @@
+//TEST:SIMPLE(filecheck=WGSL): -stage compute -entry computeMain -target wgsl
```
csyonghe, line 1: Since we only have `WGSL:` comments in the test, I guess we don't need the other test lines here.

jkwak-work:
Many CI actions are failing but they don't seem to be related to my change, because they seem to be failing to checkout from github repository.
The important one, "Windows Release", is passing, which is a good sign.

I will re-trigger the failed tests later and see if it works out.

tests/wgsl/math.slang
jkwak-work: I removed other targets.

csyonghe:
@jkwak-work any idea on the failed CI runs?

jkwak-work:
> @jkwak-work any idea on the failed CI runs?

I am still investigating.

jkwak-work:
It turned out that the CI failures are due to my change.
It was hitting an assertion failure when I reproduced it on WSL debug build at
https://github.com/shader-slang/slang/blob/25d155937402120292cff9e6f5a31c1d45a2ecbf/tools/slang-test/slang-test-main.cpp#L957

The target, SLANG_WGSL, was not handled in this function.

jkwak-work:
All CI are now passing.
Merging now.

PR #5089, opened 2024-09-17
Update spirv-tools version

cheneym2:


jkwak-work:
I think you need to include external/spirv-tools-generated in this PR, isn't it?

cheneym2:
> I think you need to include external/spirv-tools-generated in this PR, isn't it?

Maybe, I'm curious if the PR passes without that, but in any case, I will check to see what differences exist in the generated files.

jkwak-work (approved):
Looks good to me.

jkwak-work:
> I was hasty and added the generated files to the PR before finding out if the checks would have passed. Anyways, PR is updated.

I was curious about it too.
But to share my experience, I was getting a compile error when I didn't update the generated files. That's when I learned that external/spirv-tools-generated existed.
So you may get lucky without updating it for this time, but I am pretty sure it is not a proper way to upgrade the spirv-header.

csyonghe:
Yeah, I do think updating the generated files is needed.

csyonghe (approved):

PR #5096, opened 2024-09-17
Fix WGSL frexp and modf that returns a struct

jkwak-work:
Two WGSL functions have little different behavior compared to other shader languages: frexp and modf.  They return a struct to return two values.

For more details, see:
https://www.w3.org/TR/WGSL/#frexp-builtin
https://www.w3.org/TR/WGSL/#modf-builtin


jkwak-work:
There might be a better way to implement this.
If you have any suggestions, please let me know.

csyonghe (approved):

PR #5097, opened 2024-09-18
Add `IRWArray` interface, and make StructuredBuffer conform to them.

### LINKED ISSUE #5098: Missing `IRWArray` in stdlib.
We have `IArray` that provides an interface for read-only arrays, but we also need `IRWArray` so we can write generic functions that mutates an array like type.

### LINKED ISSUE #5099: Make `StructuredBuffer` conform to `IArray`.
So that they can be plugged into generic functions that operates on array liked types.

csyonghe:
Closes #5098.
Closes #5099.

This adds `IRWArray` interface to generic functions to read/write arrays.

Makes StructuredBuffer conform to IArray, and RWStructuredBuffer conform to IRWArray.

Most part of this PR is about adding the ability to synthesize a __subscript implementation from an interface requirement.
For example,
```
interface IArray<T> { __subscript(int index) -> T {get;}}

struct Array<T> : IArray<T>
{
     // case 1:
      // compiler should be able to automatically synthesize IArray::__subscript without
     // specifying it explicitly here, because arrays already has a builtin `[]` syntax.
}

struct MyArray<T>
{
     // case 2:
     // the signature of subscript doesn't exactly match the interface
    // but the compiler should be able to synthesize a matching one
    // from the provided subscript operator.
     __subscript(uint index)->T {ref{...}}
}
```

The PR fills in the missing pieces in `trySynthesizeMemberRequirement` to handle the subscript requirement case, just like how properties are handled. In fact a lot of the logic in synthesizing the accessors of a property decl is refactored into a shared sub-routine and reused for the subscript case.

This PR also removes the warning on unmodified `inout` parameter since it frequently gets in the way if there is an interface that requires `inout parameter` and the implementation doesn't actually modify it. There is adhoc logic to silence the warning if the method is implementing an interface method, but that doesn't really work for generic functions. Overall I think this warning is  kind of annoying and not easy to get it right, so removing.


tests/language-feature/generics/irwarray.slang
```diff
@@ -0,0 +1,34 @@
...
+    float arr[3] = { 1.0, 2.0, 3.0 };
+    float4 v = float4(1.0, 2.0, 3.0, 4.0);
+    float2x2 m = float2x2(1.0, 2.0, 3.0, 4.0);
+
+    // CHECK: 1.0
+    writeToBuffer(outputBuffer, 0, 1.0f);
+
+    // CHECK: 4.0
+    writeToArray(arr, 0, 4.0f);
+    outputBuffer[1] = readFromArray(arr, 0);
```
jkwak-work, line 23: I am little confused by the inconsistency with `writeToBuffer`.
If I do `writeToBuffer(outputBuffer, ...)`, it will modify `outputBuffer`.
But if I do `writeToBuffer(arr, ...)`, it wouldn't modify `arr`.

When I look at the function signature alone, it shouldn't modify the first argument because it doesn't have `out` nor `inout` modifier.
```
void writeToBuffer<U, T : IRWArray<U>>(T array, int index, U value)
```
When used with RWStructuredBuffer, `T array` works as if it is `inout T array`, which seems inconsistent.

jkwak-work (approved):
Code looks good to me.

tests/language-feature/generics/irwarray.slang
csyonghe: Yeah, this is a little unfortunate, but at the moment there isn't a better solution.

RWStructuredBuffer<T> itself should be viewed as `T*`, that is a handle to an array. Therefore, The structuredbuffer itself is a non-mutable object (represented as a uniform parameter), and therefore you can't pass it into a function that expects an `inout` parameter, because you can't assign `buffer` to something else, it is readonly.

In contract, for arrays you do want to mutate its content, so `inout` is needed there.

csyonghe: The essence in this is that `RWStructuredBuffer` is `Array<T>*` and not an array, but there is no correponding interface to represent a "reference to an array" right now, and even if we do, we can't construct a reference object to a real array in order to satisfy that interface anyways.

PR #5107, opened 2024-09-18
Member-wise constructor generation with visibility control

kaizhangNV:
Rebased version of PR #4854.


kaizhangNV:
Just try to run the CI now to see everything is fine after rebase.

source/slang/slang-check-conversion.cpp
```diff
@@ -388,88 +429,246 @@ namespace Slang
...
+                        auto ctorParamCount = ctor->getParameters().getCount();
+                        if (ctorParamCount == 0)
+                            continue;
+
+                        ioArgIndexCandidate = ioArgIndexMirror;
+                        ioArgIndex = ctorParamCount;
+
+                        // if allowCStyleInitList, process any ctor which comes next using the most members possible. ioArgIndex may not be 0.
+                        // if !allowCStyleInitList, process any ctor that exactly matched our argument count. ioArgIndex must start at 0.
+                        if (!allowCStyleInitList && ctorParamCount != Index(argCount))
```
csyonghe, line 574: problem here is that in the case of a variadic ctor, the arg count will not match param count, because the param count is just 1 for the pack, that can match to many different args.

source/slang/slang-check-conversion.cpp
```diff
@@ -388,88 +429,246 @@ namespace Slang
...
 
-                    // No point in trying further if any argument fails
-                    if(!argResult)
-                        return false;
+                    auto toStructDecl = toStructDeclRef.getDecl();
+                    // 1. Ensure our `StructDecl` has declaratations created for all auto-generated constructors
+                    ensureDecl(toStructDecl, DeclCheckState::DefaultConstructorReadyForUse);
 
-                    if( coercedArg )
+                    // 2. We have a `{}` (0 arguments), try to coerce
```
csyonghe, line 535: I think we should delete all these logic here from 535 till 671, and reuse the existing `ResolveInvoke` logic to resolve the constructor call.

Once we reach here, we are about to turn initializer list `{...}` into a ctor call that is `StructType(...)`.
This can be done by:
1. create a AST node that is `InvokeExpr(calleeExpr, coercedArgs)`, where:
    `calleeExpr` is a `VarExpr(declRef = toStructDeclRef, type = TypeType(toType), checked = true)`.
2. call `ResolveInvoke` on the invokeExpr, and that is the `*outToExpr` we need to return.

This should handle all the generic case or the variadic case just fine using the same logic in ResolveInvoke.

source/slang/slang-check-conversion.cpp
```diff
@@ -128,7 +148,7 @@ namespace Slang
         // for aggregate initialization.
         //
         auto firstInitExpr = fromInitializerListExpr->args[ioInitArgIndex];
-        if(shouldUseInitializerDirectly(toType, firstInitExpr))
+        if (shouldUseInitializerDirectly(toType, firstInitExpr) && canCoerce(toType, firstInitExpr->type, firstInitExpr))
```
csyonghe, line 151: delete shouldUseInitializerDirectly, instead, call `canCoerce` here.

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +221,24 @@ namespace Slang
         return baseStructDeclRef;
     }
 
+    GenericAppDeclRef* _getGenericAppDeclRefType(Type* argType)
+    {
+        auto argDeclRefType = as<DeclRefType>(argType);
```
csyonghe, line 226:
```
auto declRef = isDeclRefTypeOf<Decl>(argType);
if (!declRef)
    return nullptr;
return as<GenericAppDeclRef>(declRef.getDeclRefBase());
```

source/slang/slang-check-conversion.cpp
```diff
@@ -35,81 +35,101 @@ namespace Slang
...
-        if(isEffectivelyScalarForInitializerLists(fromExpr->type))
+        if (isEffectivelyScalarForInitializerLists(fromExpr->type))
             return false;
 
-        // Once the above cases are handled, the main thing
-        // we want to check for is whether a direct initialization
-        // is possible (a type conversion exists).
-        //
-        return canCoerce(toType, fromExpr->type, fromExpr);
+        // If 2 types are "broadly equal" we know the types can be coerced directly
```
csyonghe, line 104: I want to understand why this change here, along with the change in line 151 is needed.

I think it should be fine if we revert all changes to this function and in 151.

If things break, I want to understand why.

source/slang/slang-check-decl.cpp
```diff
@@ -8475,19 +8829,327 @@ namespace Slang
...
-            if (seqStmt && seqStmt->stmts.getCount() == 0)
+            ensureDecl(i.m_inheritanceDecl, DeclCheckState::VarInitExprAreChecked);
+            if(as<StructDecl>(i.m_inheritanceBaseDecl))
+                baseStructInfo = &i;
+        }
+
+        // pre-calculate any requirements of a CudaHostAttribute
+        HashSet<VarDeclBase*> requiresCudaHostModifier;
+        for (auto member : membersOfStructDeclInstance)
+            if (containsTargetType<TorchTensorType>(m_astBuilder, member.getDecl()->type.type))
```
csyonghe, line 8851: This is going to be bad for performance because it revisits the entire type hierarchy for all the member types.

source/slang/slang-check-decl.cpp
```diff
@@ -8475,19 +8829,327 @@ namespace Slang
...
             if (seqStmtChild->stmts.getCount() == 0)
                 continue;
-            seqStmt->stmts.insert(insertOffset, seqStmtChild);
+
+            auto seqStmt = _ensureCtorBodyIsSeqStmt(m_astBuilder, ctor);
+            seqStmt->stmts.insert(ctorInfo.m_insertOffset++, seqStmtChild);
         }
 
-        if (structDeclInfo.defaultCtor)
+        // Note: we assume only 1 inheritance decl currently.
```
csyonghe, line 8837: There is a lot of code added to this checkAggType method. We really need to split them into separate functions for readability.

source/slang/slang-check-decl.cpp
csyonghe: An alternative is to add the required CudaHost attribute in an IR pass to make things cleaner.

source/slang/slang-check-decl.cpp
```diff
@@ -8475,19 +8829,327 @@ namespace Slang
...
+                auto stmt = m_astBuilder->create<ExpressionStmt>();
+                stmt->expression = assign;
+                stmt->loc = assign->loc;
+
+                seqStmtChild->stmts.add(stmt);
+            }
+            seqStmt->stmts.insert(ctorInfo.m_insertOffset++, seqStmtChild);
+        }
+
+        // Compiler generated ctor may be destroyed
```
csyonghe, line 9024: What will go wrong if we just leave the empty ctor there without deleting it?

source/slang/slang-check-decl.cpp
```diff
@@ -8475,19 +8829,327 @@ namespace Slang
...
+            // Part 3
+            // Assign initExpr to all members of structDecl
+            for (auto memberRef : membersOfStructDeclInstance)
+            {
+
+                auto member = memberRef.getDecl();
+                addCudaHostModifierIfRequired(defaultInitListFunc, member, foundCudaHostModifier);
+
+                auto initExpr = member->initExpr;
+                if (!initExpr)
```
csyonghe, line 9122: if !initExpr, and zero-init flag is set, we want to create a DefaultConstruct expr to create a 0-initialized value of member->type, and assign it to member.

source/slang/slang-check-decl.cpp
```diff
@@ -8475,19 +8829,327 @@ namespace Slang
...
+            2. Insert `super::$DefaultInit()` into `this->$DefaultInit`
+            3. Assign initExpr to all members
+            */
+
+            bool foundCudaHostModifier = false;
+
+            auto defaultConstructExpr = createDefaultConstructExprForType(m_astBuilder, structDeclType, seqStmt->loc);
+
+            // Part 1
+            // Assign DefaultConstructExpr to `this`
```
csyonghe, line 9067: I think we only need to execute part 1 if "-zero-init" option is set.

source/slang/slang-check-overload.cpp
```diff
@@ -2385,6 +2390,35 @@ namespace Slang
...
     }
 
+
+    /// We allow a special case for when `funcExpr` is expected to be a Default initializer
+    /// but none exist. Note, we cannot just create a default initializer for every variable
+    /// since then we are introducing initialization to every variable through an indirect
+    /// init returning data. Instead we will call `$DefaultInit` through this logic below.
+    ///     struct S { int x; }
+    ///     S s = S(); // This will call S.$DefaultInit().
+    Expr* _tryOverloadWithDefaultInit(SemanticsVisitor* visitor, SemanticsVisitor::OverloadResolveContext& context, Expr* expr, OverloadCandidate* bestCandidate)
```
csyonghe, line 2400: Can you explain this again? I still don't understand what it is trying to do. For the example shown in the comment here, haven't we already synthesized the default ctors for `S`, so that the overload resolution will resolve into a call to that defualt ctor already? Why do we need to run the coercion logic again here?

csyonghe (commented):
Looks good overall, but I think there are few important cleanups we need to do before checking this in.

source/slang/slang-check-decl.cpp
kaizhangNV: I just figure out why Ariel wants to delete the empty default ctor.
Considering this case
```
struct S
{
    int a;
}

S s;
```
In this case, based on currently logic, we will synthesize a empty default ctor: `S.$init`, and for `S s;`, we will call this default ctor `S.$init`. So this will cause an issue that slang will report a warning that `a` is not initialized.
While, if there is no default ctor (removing the default ctor), the default ctor won't be called to construct `s`, no warning will be reported.

I didn't participant to your previous discussion about what the default behavior it is. I know that for C++, the default ctor will be called, and C++ won't report warning for that.

If we still want to insist this behavior, I think the warning reported here is quite helppful because it kind of reminds developers to initialize it or use `s = {}` to zero init it.

BTW, there is actually noting wrong that if we don't delete this ctor.

kaizhangNV: @csyonghe to comment this reply first.

16-Bit-Dog: It is worth noting that there was a second reason that makes this change slightly problematic.

By auto generating `init()` functions, we will be changing how code-gen is structured since certain previously uninitialized `struct` values will now have an assignment to a function that effectively does nothing.

This means that:
1. Certain tests will need to be adjusted.
2. Code-gen will include `init()` functions that do nothing but are still used in code-gen.

source/slang/slang-check-overload.cpp
16-Bit-Dog: This is code that mostly carries over from how TOT did things (put into its own function). I will break down why we have this function into 2 parts.

*What `_tryOverloadWithDefaultInit` does:*

If I type the following code, `_tryOverloadWithDefaultInit` will be used to resolve `MyStruct()`:
```
struct MyStruct
{
    float a;
};

[numthreads(1,1,1)]
void computeMain(int3 tid : SV_DispatchThreadID)
{
    MyStruct val = MyStruct(); // effectively `MyStruct val = {}`
}
```

*Why do we have this function:*

[using the code above] We have 2 requirements for any call to `MyStruct()` or related (Ex: `RayDesc()`):
1. We need to treat `MyStruct()` as a 'default-initialize'. This is a requirement since lots of code relies on this behavior.
2. We cannot treat `MyStruct()` as a 'default-initialize' using a compiler generated `init()` since this will force all uninitialized `struct` types into calling their `init()` ('default-initialize'). This is undesired behavior.

An example where rule 2 is important:
```
struct MyStruct
{
    float a;
};

[numthreads(1,1,1)]
void computeMain(int3 tid : SV_DispatchThreadID)
{
    MyStruct val1 = MyStruct(); // effectively `MyStruct val = {};`. This variable is default initialized.
    MyStruct val2; // Effectively `MyStruct val2;`. This variable is uninitialized.
}
```

source/slang/slang-check-conversion.cpp
kaizhangNV: Create a sub-task to address this issue: https://github.com/shader-slang/slang/issues/5134.
Note, I won't submit a new PR for this. It just that I found out the original task took much longer than we estimated (estimation is just 1 day! And I already spent 3 days). And it could take longer, so split the task smaller so that I will have a clear goal what I'm going to resolve.

source/slang/slang-check-decl.cpp
kaizhangNV: Create sub-task #5135 to address comment.

source/slang/slang-check-overload.cpp
kaizhangNV: Create a sub-task to remove this hack: https://github.com/shader-slang/slang/issues/5138

source/slang/slang-check-decl.cpp
kaizhangNV: create sub-task to remove this hack: Create a sub-task to remove this hack: https://github.com/shader-slang/slang/issues/5138

PR #5112, opened 2024-09-19
Add diagnostic to verify the left hand side of a generic constraint.

### LINKED ISSUE #5113: Add validation of left side of `where` clause.
After adding `where` clause, it is now possible for user to write `void foo<T>() where int : Bar{}` which doesn't make sense.

We need to add a check to guarantee that the left hand side of the where clause is dependent on at least one generic type parameter.

csyonghe:
(This PR is based on top of PR #5111.)

Add validation on left hand side of type generic. For example, we need to diagnose an error for:
```
void foo<T>() where int : X {} // int cannot be left hand side of a generic constraint.
```

Closes #5113.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -2522,7 +2522,7 @@ namespace Slang
                 // interface IFoo {
                 //    associatedtype T
                 //        where T : IFoo  // OK, constraint is on the associatedtype T itself.
-                //        where T.T == X; // OK, constraint is on the associated type of T.
+                //        where T.T == X  // OK, constraint is on the associated type of T.
                 //        where int == X; // Error, int is not a valid left hand side of a constraint.
                 //  }
                 // ```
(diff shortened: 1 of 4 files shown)


source/slang/slang-check-decl.cpp
```diff
@@ -2487,6 +2489,69 @@ namespace Slang
...
+            {
+                // If the constraint is on an associated type, then it should either be the associated type itself,
+                // or a associated type of the associated type.
+                // For example,
+                // ```
+                // interface IFoo {
+                //    associatedtype T
+                //        where T : IFoo  // OK, constraint is on the associatedtype T itself.
+                //        where T.T == X; // OK, constraint is on the associated type of T.
+                //        where int == X; // Error, int is not a valid left hand side of a constraint.
```
jkwak-work, line 2526: I think you meant to say 
```
where int : X // Error ...
```
I mean that `==` should be replaced with `:` and `;` should be dropped.

jkwak-work (dismissed):
Looks good to me.

source/slang/slang-check-decl.cpp
csyonghe: `==` is also a valid constraint operator, for type equality constraints.

PR #5120, opened 2024-09-19
Re-enable slang-test for WGSL

jkwak-work:
My previous commit disabled the WGSL test by a mistake. This commit fixes the mistake and run the slang-test for WGSL tests.

frexp and modf were still not working for the vector types.
And this commit fixes it.


jkwak-work:
I can confirm that CI now runs WGSL test,
```
passed test: 'tests/wgsl/math.slang' 
```

csyonghe (approved):

Issue #4999, opened 2024-09-03
Allow custom derivative definitions for non-static methods of a differentiable struct type.


saipraveenb25:
We currently have a gap in our custom derivatives system. We can't provide custom derivatives for member methods of differentiable struct types. 
Example:
```csharp
struct MyDiffType : IDifferentiable
{
     // No way to provide a user-written derivative, because the 'this' object becomes a pair type.
     float foo()
     {  /* ... */ }
}
```

We can fix this by allowing custom derivatives to be defined as static methods that take the appropriate differentiable pair as the input.
```csharp
struct MyDiffType : IDifferentiable
{
     // Proposed approach to do this
     [BackwardDerivative(foo_bwd)]
     float foo()
     {  /* ... */ }
     
     static void foo_bwd(inout DifferentialPair<MyDiffType> dp_this)
     {  /* ... */ }
}
```

PR #5124, opened 2024-09-20
Support custom derivatives of member functions of differentiable types

saipraveenb25:
- Also support definition of custom derivatives via extensions.

Fixes: #4999


source/slang/slang-check-decl.cpp
```diff
@@ -7096,13 +7096,13 @@ namespace Slang
             // Two such constraints are equivalent if their `sub`
             // and `sup` types are pairwise equivalent.
             //
-            auto leftSub = leftConstraint->sub;
-            auto rightSub = getSub(m_astBuilder, rightConstraint);
+            auto leftSub = leftConstraint->sub.type;
+            auto rightSub = substInnerRightToLeft.substitute(m_astBuilder, rightConstraint.getDecl()->sub.type);
```
csyonghe, line 7100: I thought `rightConstraint` already contains everything needed to specialize `sub` and `sup`, why isn't that working?

csyonghe (approved):

tests/autodiff/diff-member-func-custom-derivative.slang
```diff
@@ -0,0 +1,36 @@
...
+//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type
+
+//TEST_INPUT:ubuffer(data=[0], stride=4):out,name=outputBuffer
+RWStructuredBuffer<float> outputBuffer;
+
+struct A : IDifferentiable
+{
+    float x;
+
+    [ForwardDerivative(diff_f)]
```
csyonghe, line 11: should test backward derivative as well, because the parameter direction there is different.

source/slang/slang-check-decl.cpp
saipraveenb25: The issue is that when referring directly to `GenericTypeConstraintDecl` or `GenericTypeParam` we don't construct the substitutions tree at all. There's special-case logic that always uses the direct-decl-ref. As a result, there are never any substitutions on the constraint decl-ref.

PR #5125, opened 2024-09-20
Initial `Atomic<T>` type implementation.

### LINKED ISSUE #5084: WGSL: Intrinsics: atomics
3 functions

csyonghe:
Closes #5084.

Todo: test new Atomic<T> type for all existing targets as well as wgsl target.

### DIFF
--- a/docs/proposals/003-atomic-t.md
+++ b/docs/proposals/003-atomic-t.md
@@ -9,7 +9,7 @@ Author: Yong He
 
 Status: Implementation in review.
 
-Implementation: N/A
+Implementation: [PR 5125](https://github.com/shader-slang/slang/pull/5125)
 
 Reviewed by: Theresa Foley, Jay Kwak
 
--- a/docs/proposals/003-atomic-t.md
+++ b/docs/proposals/003-atomic-t.md
@@ -34,27 +34,26 @@ We define an `Atomic<T>` type that functions as a wrapper of `T` and provides at
 ```csharp
 enum MemoryOrder
 {
-    Relaxed = $(kIRMemoryOrder_Relaxed),
-    SeqCst = $(kIRMemoryOrder_SeqCst),
+    Relaxed = 0,
+    Acquire = 1,
+    Release = 2,
+    AcquireRelease = 3,
+    SeqCst = 4,
 }
 
 [sealed] interface IAtomicable {}
-[sealed] interface IArithmeticAtomicable : IAtomicable {}
-[sealed] interface IBitAtomicable : IArithmeticAtomicable {}
+[sealed] interface IArithmeticAtomicable : IAtomicable, IArithmetic {}
+[sealed] interface IBitAtomicable : IArithmeticAtomicable, IInteger {}
 
 [require(cuda_glsl_hlsl_metal_spirv_wgsl)]
 struct Atomic<T : IAtomicable>
 {
-    __intrinsic_op($(kIROp_AtomicLoad))
     T load(MemoryOrder order = MemoryOrder.Relaxed);
 
-    __intrinsic_op($(kIROp_AtomicStore))
     [__ref] void store(T newValue, MemoryOrder order = MemoryOrder.Relaxed);
 
-    __intrinsic_op($(kIROp_AtomicExchange))
     [__ref] T exchange(T newValue, MemoryOrder order = MemoryOrder.Relaxed); // returns old value
 
-    __intrinsic_op($(kIROp_AtomicCompareExchange))
     [__ref] T compareExchange(
         T compareValue,
         T newValue,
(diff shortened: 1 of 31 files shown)


docs/proposals/003-atomic-t.md
```diff
@@ -31,42 +31,70 @@ Proposed Approach
 -----------------
 
 We define an `Atomic<T>` type that functions as a wrapper of `T` and provides atomic operations:
-```
+```csharp
+enum MemoryOrder
+{
+    Relaxed = $(kIRMemoryOrder_Relaxed),
+    SeqCst = $(kIRMemoryOrder_SeqCst),
```
jeffbolznv, line 38: FWIW Vulkan doesn't actually support sequentially consistent atomics. We do support AcquireRelease which is close, but doesn't support some obscure use cases. I think it would be fine to only have Relaxed for now.

csyonghe: I see. `SeqCst` was added because metal supports it.

csyonghe: I guess we can have AcquireRelease here and map it to SeqCst on metal.

source/slang/core.meta.slang
```diff
@@ -2721,6 +2721,132 @@ __Addr<T> __getLegalizedSPIRVGlobalParamAddr(T val);
...
+    __intrinsic_op($(kIROp_AtomicAnd))
+    [__ref] T and(T value, MemoryOrder order = MemoryOrder.Relaxed);      // returns original value
+    __intrinsic_op($(kIROp_AtomicOr))
+    [__ref] T or(T value, MemoryOrder order = MemoryOrder.Relaxed);         // returns original value
+    __intrinsic_op($(kIROp_AtomicXor))
+    [__ref] T xor(T value, MemoryOrder order = MemoryOrder.Relaxed); // returns original value
+    __intrinsic_op($(kIROp_AtomicInc))
+    [__ref] T increment(MemoryOrder order = MemoryOrder.Relaxed);
+    __intrinsic_op($(kIROp_AtomicDec))
+    [__ref] T decrement(MemoryOrder order = MemoryOrder.Relaxed);
```
jkwak-work, line 2792: Would they return the original value too?

source/slang/core.meta.slang
```diff
@@ -2721,6 +2721,132 @@ __Addr<T> __getLegalizedSPIRVGlobalParamAddr(T val);
...
+[sealed] interface IArithmeticAtomicable : IAtomicable, IArithmetic {}
+[sealed] interface IBitAtomicable : IArithmeticAtomicable, IInteger {}
+
+extension int : IBitAtomicable {}
+extension uint : IBitAtomicable {}
+extension int64_t : IBitAtomicable {}
+extension uint64_t : IBitAtomicable {}
+extension double : IArithmeticAtomicable {}
+extension float : IArithmeticAtomicable {}
+extension half : IArithmeticAtomicable {}
```
jkwak-work, line 2745: SM 6.6 added bit-wise atomic operation for float.
https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html#interlockedcomparestorefloatbitwise
https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html#interlockedcompareexchangefloatbitwise

source/slang/slang-emit-metal.cpp
```diff
@@ -664,28 +836,32 @@ void MetalSourceEmitter::emitSimpleTypeImpl(IRType* type)
...
         case kIROp_UInt16Type:
             m_writer->emit("ushort");
             return;
         case kIROp_IntPtrType:
-            m_writer->emit("int64_t");
+            m_writer->emit("long");
             return;
         case kIROp_UIntPtrType:
-            m_writer->emit("uint64_t");
+            m_writer->emit("ulong");
```
jkwak-work, line 864: Thanks for fixing this.

jkwak-work (approved):
Looks good to me.

source/slang/core.meta.slang
csyonghe: Yes, all atomic methods returns original values.

source/slang/core.meta.slang
csyonghe: The interface definition is fine, because IArithmeticAtomicable is IAtomicable, and compareExchange is a IAtomicable method.

However this reminds me that the currently code emit logic for floating point atomics on HLSL is wrong. We should be using the InterlockedAddF32/F16 etc instead of InterlockedAdd.

csyonghe:
I am going to merge this first, and then attempt to fix the floating point intrinsics on hlsl later.

PR #5128, opened 2024-09-20
Add interfaces for retrieving separate linkable downstream binaries

cheneym2:


source/slang-record-replay/record/slang-module.cpp
```diff
@@ -213,17 +213,6 @@ namespace SlangRecord
         return res;
     }
 
-    SLANG_NO_THROW SlangResult ModuleRecorder::precompileForTarget(
```
csyonghe, line 216: We must provide implementation for record-replay.

csyonghe (commented):
For any compilation interface added to the slang.h header, we should implement the record/replay logic for the interface.

csyonghe:
The new interface needs to be marked as experimental, probably name it `IModulePrecompileService_` and clearly state it as experimental and subject to removal or change in the comment.

include/slang.h
```diff
@@ -5574,13 +5574,34 @@ namespace slang
             SlangInt32 index) = 0;
 
         virtual SLANG_NO_THROW DeclReflection* SLANG_MCALL getModuleReflection() = 0;
+    };
+    
+    #define SLANG_UUID_IModule IModule::getTypeGuid()
+
+    /* Experimental interface for doing target precompilation of slang modules */
```
cheneym2, line 5581: Already marked experimental

cheneym2:
> The new interface needs to be marked as experimental, probably name it `IModulePrecompileService_` and clearly state it as experimental and subject to removal or change in the comment.

I did already comment it as Experimental in slang.h.  I can additionally rename it.

PR #5129, opened 2024-09-20
Fix handling of pointer logic in wgsl backend.

csyonghe:
This PR overhauls the logic of handling pointer dereference in WGSL backend.

Basically we treat WGSL as a target that doesn't support pointers, so we never emit any pointer logic for variable reference or structured buffer reference, i.e. no `*`, `&`, `->` etc. through out the emit logic like what we are doing for hlsl and glsl, except:

1. When emitting an `inout` parameter, emit it as `ptr<function, T> param`.
2. When referencing a `inout` parameter as operand, emit it as `(*param)` so the rest of logic won't need to deal with pointers.
3. When calling a function with an argument to an `inout` parameter, emit the argument as `&arg`.

Aside from call sites, we treat wgsl as a language that doesn't have pointers everywhere else. This should work exactly as WGSL expected, since pointers in WGSL is almost solely used to express mutable parameters. So we are basically treating `&` and `*` as the syntax around mutable parameters instead of something really meaningful.

The other legalization step introduced in this PR is to work around the WGSL restriction on not allowing to pass a pointer to a sub part of a composite value as argument to an inout parameter. To make such code work, we will need to introduce a temp var for the argument and pass the temp var into the function instead, and then write back the new value of the temp var to the original pointer arg after the call.

### DIFF
--- a/source/slang/slang-emit-c-like.cpp
+++ b/source/slang/slang-emit-c-like.cpp
@@ -1998,6 +1998,11 @@ void CLikeSourceEmitter::emitIntrinsicCallExprImpl(
     }
 }
 
+void CLikeSourceEmitter::emitCallArg(IRInst* inst)
+{
+    emitOperand(inst, getInfo(EmitOp::General));
+}
+
 void CLikeSourceEmitter::_emitCallArgList(IRCall* inst, int startingOperandIndex)
 {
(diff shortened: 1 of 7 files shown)


jkwak-work (approved):
Looks good to me.

PR #5130, opened 2024-09-21
Implemented Combined-texture for WGSL

jkwak-work:
The implementation in hlsl.meta.slang for combined texture was already committed and it appears to be working fine.
The trouble was more on the layout rule where the synthesized textures and samplers were not getting the binding offset values correctly.

When the combined-texture is used, it is replaced with a texture and a sampler.
For WGSL, both cases get layout info as `LayoutResourceKind::DescriptorTableSlot`.
This is little different from how it is handled in HLSL where Texture got `LayoutResourceKind::ShaderResource` and Sampler got `LayoutResourceKind::SamplerState`

This commit also adds a new layout-rule for WGSL.
It is mostly identical to the layout-rule for GLSL.
The difference is that Combined-texture is handled little differently.
This is because for GLSL, there is an actual combined sampler from GLSL language level and Slang don't need to synthesize one. For WGSL, there isn't combined sampler and Slang has to synthesize one.

### DIFF
--- a/source/slang/slang-type-layout.cpp
+++ b/source/slang/slang-type-layout.cpp
@@ -1867,13 +1867,12 @@ LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getAnyValueRules()
     return &kGLSLAnyValueLayoutRulesImpl_;
 }
 
-LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getConstantBufferRules(CompilerOptionSet& compilerOptions)
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getConstantBufferRules(CompilerOptionSet&)
 {
-    (void) compilerOptions;
     return &kWGSLConstantBufferLayoutRulesImpl_;
 }
 
-LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getParameterBlockRules(CompilerOptionSet& compilerOptions)
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getParameterBlockRules(CompilerOptionSet&)
 {
     return &kStd140LayoutRulesImpl_;
 }
@@ -1888,7 +1887,7 @@ LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getShaderRecordConstantBufferRules()
     return &kGLSLShaderRecordLayoutRulesImpl_;
 }
 
-LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getTextureBufferRules(CompilerOptionSet& compilerOptions)
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getTextureBufferRules(CompilerOptionSet&)
 {
     return &kStd430LayoutRulesImpl_;
 }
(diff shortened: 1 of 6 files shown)


source/slang/slang-type-layout.cpp
```diff
@@ -1816,6 +1856,103 @@ LayoutRulesImpl* MetalLayoutRulesFamilyImpl::getHitAttributesParameterRules()
...
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getShaderRecordConstantBufferRules()
+{
+    return &kGLSLShaderRecordLayoutRulesImpl_;
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getTextureBufferRules(CompilerOptionSet& compilerOptions)
+{
+    if (compilerOptions.shouldUseScalarLayout())
```
csyonghe, line 1898: I don't think it is allowed in wgsl to declare a layout that isn't consistent with std140/std430.

source/slang/slang-type-layout.cpp
```diff
@@ -1816,6 +1856,103 @@ LayoutRulesImpl* MetalLayoutRulesFamilyImpl::getHitAttributesParameterRules()
...
+    return &kGLSLVaryingOutputLayoutRulesImpl_;
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getSpecializationConstantRules()
+{
+    return &kGLSLSpecializationConstantLayoutRulesImpl_;
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getShaderStorageBufferRules(CompilerOptionSet& compilerOptions)
+{
```
csyonghe, line 1922: This can only be std430.

source/slang/slang-type-layout.cpp
```diff
@@ -1816,6 +1856,103 @@ LayoutRulesImpl* MetalLayoutRulesFamilyImpl::getHitAttributesParameterRules()
...
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getHitAttributesParameterRules()
+{
+    return &kGLSLHitAttributesParameterLayoutRulesImpl_;
+}
+
+LayoutRulesImpl* WGSLLayoutRulesFamilyImpl::getStructuredBufferRules(CompilerOptionSet& compilerOptions)
+{
+    if (compilerOptions.shouldUseScalarLayout())
```
csyonghe, line 1948: Can only be std430.

tests/wgsl/texture-sampler-less.slang
```diff
@@ -0,0 +1,387 @@
...
+
+        // W-GSL: textureSampleGrad({{\(*}}t2DArray
+        && all(Tvn(T(0)) == t2DArray.SampleGrad(float3(u2, u, 0.0f), float2(ddx, ddx), float2(ddy, ddy), int2(0, 0)))
+#endif // #if TEST_WHEN_CONSTEXPR_WORKS_FOR_OFFSET
+        ;
+
+    return result;
+}
+
+void fragMain()
```
csyonghe, line 365: Maybe add a check for the binding location?

jkwak-work: That's a good idea. I will do that.

csyonghe (approved):

Issue #5117, opened 2024-09-19
Wasm build: Integrate to CI


aleino-nv:
Integrate the Wasm build into CI and make our releases include webassembly builds.

aleino-nv:
https://github.com/aleino-nv/slang/blob/aleino/wasm/docs/building.md#webassembly-build

aleino-nv:
This will need the CI machine to have emsdk: https://emscripten.org/docs/getting_started/downloads.html, as mentioned in the doc link I just posted above

PR #5131, opened 2024-09-23
Enable emscripten builds to compile slang.dll to WebAssembly.

aleino-nv:
Add support for Wasm build target

### DIFF
--- a/source/slang/slang-ast-dump.cpp
+++ b/source/slang/slang-ast-dump.cpp
@@ -304,9 +304,9 @@ struct ASTDumpContext
 
     void dump(uint32_t v)
     {
-        m_writer->emit(UInt(v));
+        m_writer->emit((uint64_t)v);
     }
-    void dump(UInt v)
+    void dump(uint64_t v)
     {
         m_writer->emit(v);
     }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,6 +253,13 @@ add_custom_target(
     all-generators
     COMMENT "meta target which depends on all generators"
 )
+
+if(CMAKE_HOST_WIN32)
+  set(CMAKE_HOST_EXECUTABLE_SUFFIX ".exe")
+else()
+  set(CMAKE_HOST_EXECUTABLE_SUFFIX "")
+endif()
+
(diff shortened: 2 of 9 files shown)


source/slang/slang-ast-dump.cpp
```diff
@@ -304,9 +304,9 @@ struct ASTDumpContext
 
     void dump(uint32_t v)
     {
-        m_writer->emit(UInt(v));
+        m_writer->emit((uint64_t)v);
     }
-    void dump(UInt v)
```
aleino-nv, line 309: (With WASM target, `UInt` is the same type as `uint32_t` and so the function was redefined.)

aleino-nv: There may be some Emscripten options (`-sWASM_BIGINT` and/or `-sMEMORY64=1`) that I believe would let us have `long` be 64 bits...

csyonghe:
PR titles needs to be a descriptive sentence and the description needs to be summrizing the changes in the PR.

source/slang/slang-ast-dump.cpp
csyonghe: We should probably turn that flag on.

csyonghe (approved):

csyonghe:
We need a CI setup to prevent regression that breaks emscripten builds.

source/slang/slang-ast-dump.cpp
aleino-nv: I experimented with these flags but for some reason they don't seem to help with these particular issues. (In fact I think these are linker flags.)

64 bit integers are emulated in JS and so there's be a perf hit involved.
https://emscripten.org/docs/porting/guidelines/portability_guidelines.html#code-that-compiles-but-might-run-slowly

`WASM_BIGINT` just skips a legalization steps where i64s become pairs of i32s in favor of emitting i64 operations and relying on special VM/browser support for i64s. I guess whatever flag makes long be 64 bits is orthogonal.

In any case I'd like to defer the decision since it seems to be working as it is right now.

aleino-nv:
> We need a CI setup to prevent regression that breaks emscripten builds.

Yeah, that's this task: https://github.com/shader-slang/slang/issues/5117

PR #5141, opened 2024-09-23
Feature/wgsl intrinsic texture gather

jkwak-work:
This PR implements the texture gather functions for WGSL.

    The pattern was very similar to how Metal was implemented.
    Before copy and paste from the Metal implementation, I had to
    clean up the Metal implementation to make it more readable
    and maintainable.

    Gather functions are available only for 2D and 3D textures.
    Their `array` and `depth` variants may or may not be supported depending on the target.
    `static_assert` ensures that Gather functions are available only for 2D and 3D textures.

    Removed incorrect use of "$p" argument for targeting GLSL.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -2190,11 +2190,11 @@ Array<T,4> __makeArray<T>(T v0, T v1, T v2, T v3);
 
 
 // Beginning of Texture Gather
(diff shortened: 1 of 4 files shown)


jkwak-work:
It seems like my clean up caused a problem for `GatherCmp` when the texture is non-depth texture.
GatherCmp in HLSL requires the texture to be a depth texture.
But it seems like GLSL doesn't require it.
I will fix it.

csyonghe (approved):

csyonghe (approved):

PR #5142, opened 2024-09-23
Design proposal for initialization.

### LINKED ISSUE #5149: Design how initialization works in slang.

csyonghe:
Closes #5149.


kaizhangNV:
The proposal is not in *..md extension, github cannot render it into formatted doc.
Can you update that?

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+Implemtation: N/A
+
+Author: Yong He
+
+Reviewer: 
+
+Background
+----------
+
+Slang has introduced several different syntax around initialization to provide syntactic compatibility with HLSL/C++. As the language evolve, there aree many corners where
```
kaizhangNV, line 20: typo: aree -> area

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+the semantics around initialization are not well-defined, and causing confusion or leading to surprising behaviors.
+
+This proposal attempts to provide a design on where we want to language to be in turns of how initialization is handled in all different places.
```
kaizhangNV, line 23: want to -> want

kaizhangNV: "in turns of" ?

is "in terms of"?

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+## Variable Initialization
+
+If the type of a local variable is default-initializable, then its default initializer will be invoked at its declaration site implicitly to intialize its value:
+```c++
+int x; // x will be default initialized to 0 because `int` is default-initializable.
+// The above is equivalent to:
+int x = int();
+
+struct S { int x; int y; }
+S s; // s will be default initialized to {0, 0} because `S` is default-initializable.
```
kaizhangNV, line 78: Can we add few sentences to make it clear that

`S s;`
equal to 
`S s = S();`

And I have a question:
Is `S s = S()` equal to `S s = {}` ?

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+void foo<T>()
+{
+    T t; // error, `t` is uninitialized.
+}
+```
+
+## Automatic Synthesis of Default-Initializer
+
+If a `struct` type is determined to be default-initializable but a default constructor isn't explicitly provided by the user, the Slang compiler should
+synthesize such a constructor for the type. The synthesis logic should be recursively invoke defualt initialization on all members.
```
kaizhangNV, line 103: typo: defualt -> default.

Suggestion:
"defualt initialization" => "default initializer"

docs/proposals/004-initialization
kaizhangNV: Never mind, I see you you covered question this later.

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+Slang will try to convert the code into:
+```
+S obj = S(1,2);
+```
+
+As a special case, an empty initializer list will translate into a default-initialization:
+```csharp
+S obj = {};
+// equivalent to:
+S obj = S();
```
kaizhangNV, line 127: By knowing this, I think the PR makes a mistake.

Because it treats these two differently.

That's why there are both `$init(This)` and `$defaultInit(This)` functions synthesized.

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+A type X is default initializable if:
+- It explicitly declares that `X` implements `IDefaultInitializable`.
+- It explicitly provides a default constructor `X::__init()` that takes no arguments, in which case we treat the type as implementing `IDefaultInitializable` even if
+  this conformance isn't explicitly declared.
+- It is a struct type where all its members are default-initializable. A member is considered default-initializable if the type of the member is default-initializable,
+  or if the member has an initialization expression that defines its default value.
+- It is a sized-array type where the element type is default-initializable.
+- It is a tuple type where all element types are default-initializable.
+- It is an `Optional<T>` type for any `T`.
+
```
kaizhangNV, line 64: How about the type defined in shader resource type, e.g.:

cbuffer Uniforms
{
	default-initializable-type a;
}

is `a` still considered default-initializable? I think it is.

docs/proposals/004-initialization
```diff
@@ -0,0 +1,177 @@
...
+```
+
+As a special case, an empty initializer list will translate into a default-initialization:
+```csharp
+S obj = {};
+// equivalent to:
+S obj = S();
+```
+
+If the above code passes type check, then it will be used as the way to initialize `obj`.
```
expipiplus1, line 130: Wouldn't it be neater from a typing point of view to just synthesize a constructor which perform this, then we wouldn't need to special case any initializer list application and it would always resolve to a constructor call

expipiplus1 (commented):
This doesn't mention `out` parameters, which at the moment are at risk of being uninitialized at their declaration site. I would be all for requiring such parameters be default initializable (and make it easier to return multiple values from a function to facilitate situations where this isn't possible)

kaizhangNV:
This doesn't mention `static` variable in a class. Is that default initializable?

csyonghe:
`static` variables are a tricky case because there is currently no efficient way to initialize them on d3d and vulkan. So the answer for now has to be that they won't be default initialized.

csyonghe:
I think for `out` parameters, we also don't want to default initialize them, because an `out` parameter must be coming from a location that is either default initlaized because it is a var, or can't be efficiently default inited because it is a static global var.

docs/proposals/004-initialization
csyonghe: it is considered default-initializable, although it will have no meaning since the location is in a read-only position.

tangent-vector:
To follow up on the discussion of `out` parameters: the existence of `out` in Slang should be another factor pushing us to have a model where the semantic rule is *not* that it is an error to fail to initialize a variable at its declaration point, and instead should have a rule that it is an error to *use* a variable at a point where it might not be (fully) initialized.

We should ideally treat use of a variable (or field of a `struct` etc.) as an argument for an `out` parameter as equivalent to an assignment for the purposes of deciding whether a variable has been initialized, and where.

tangent-vector:
If we do make the rules be that the initialization point of a variable need not be its declaration, that would also allow us to have variables declared with `let` be initialized via assignment or being used as an `out` argument:

    let x : Int; // okay: just not initialized yet
    if(something) x = 1; else x = 2;

This is only a slight extension of the rules needed for the `var` case:

* For any read of a variable, the variable must be fully initialized along every control-flow path that reaches that read

* For any write of a `let` variable, the variable must be fully *uninitialized* along every control-flow path that reaches that write

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,191 @@
...
+Slang will try to convert the code into:
+```
+S obj = S(1,2);
+```
+
+As a special case, an empty initializer list will translate into a default-initialization:
+```csharp
+S obj = {};
+// equivalent to:
+S obj = S();
```
kaizhangNV, line 126: I just want to be very clear that

`S obj = {}; `

won't zero-initialize any members of S. It's not a C-style zero initializer list anymore.

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,317 @@
...
+However, the Slang language has been allowing implicit initialization of variables whose types are default initializable types.
+For example,
+```csharp
+struct MyType1 {
+  int x;
+  __init() { x = 0; }
+}
+void foo() {
+  MyType t1; // `t1` is initialized with a call to `__init`.
+}
```
kaizhangNV, line 65: can you also make an example that if there is no `__init()` defined for `MyType1`, what will happen?

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,317 @@
...
+A type is considered "default-initializable" if it provides a constructor that can take 0 arguments, so that it can be constructed with `T()`.
+
+### Variable Initialization
+
+Generally, a variable is considered uninitialized at its declaration site without an explicit value expression.
+For example,
+```csharp
+void foo()
+{
+  MyType t; // t is considered uninitialized.
```
kaizhangNV, line 51: Can you clarify that there is no `__init()` defined for `MyType`?

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,317 @@
...
+  Empty s0 = {}; // Works, `s` is considered initialized.
+  Empty s1; // `s1` is considered initialized bcause `Empty` is `IDefaultInitializable`.
+}
+
+struct CLike {int x; int y; }
+void test1()
+{
+  CLike c0; // `c0` is uninitialized.
+  CLike c1 = {}; // initialized with legacy initializaer list logic, `c1` is now `{0,0}`.
+  CLike c2 = {1}; // initialized with legacy initializaer list logic, `c1` is now `{1,0}`.
```
kaizhangNV, line 200: Looking at this example, so we still zero-initialize members?

kaizhangNV: And we still want to support legacy initializaer list?

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,317 @@
...
+}
+
+public struct Visibility2
+{
+  internal int x = 1;
+  public int y = 0;
+}
+void test7()
+{
+  Visibility2 t = {0, 0}; // error, no matching ctor
```
kaizhangNV, line 270: From user-guide:

`internal` symbols are visible throughout the same module, regardless if it is referenced from the same type or same file. But they are not visible to other modules. 

so why is that?
Visibility2 t = {0, 0}; // error, no matching ctor 

are Visibility2 and test7 in the different module?

docs/proposals/004-initialization.md
csyonghe: No, it should not, if `S` defines a default ctor, `S obj = {}` should be calling that default ctor.

docs/proposals/004-initialization.md
csyonghe: The general principle is never to initialize `t`. The fact that we still need to initialize `t` if `MyType` has default `__init` is just for backward compatbility, and we may in the future no longer do so. If the user is defining this as modern syntax such as `var t : MyType`, we will never initialize whether or not `MyType` has `__init`.

csyonghe: So: if `MyType` has a default `__init()`, then:
1. `MyType s;` will default initialize now, but we need to think about a migration path so we don't default initilaize it in the future.
2. `var s : MyType` will not initialize.
3. In a modern module, i.e. a module defined with `module moduleName;`, `MyType s;` will not default initialize.

docs/proposals/004-initialization.md
csyonghe: However, if `S` is a C-style struct, then `S s = {}` will still fall back to c-style initailization logic according to this proposal because the ctor match will fail. In that case, we will still zero initialize.

docs/proposals/004-initialization.md
csyonghe: Yes, that is a still fallback if ctor match failed.

docs/proposals/004-initialization.md
csyonghe: This will fail because `Visiblity2` type contains members at different visiblity levels that disqualifies it from being a "C-style" struct type, and therefore we will NOT try the C-Style fallback logic to resolve the initializer list. If ctor match failed and the type is not c-style struct, it is an error.

docs/proposals/004-initialization.md
kaizhangNV: Yes, I understand the rules here.

But what I'm trying to suggest is to change this example as
```
struct MyType1 {
  int x;
}
void foo() {
  MyType t1; // `t1` is initialized with a call to `__init`.
}
```

So we know why it's different from the example in line 59-65. Because there is no default `__init()` defined in `MyType1`.

docs/proposals/004-initialization.md
kaizhangNV: Can we also make a clear definition in this proposal that what is "C-style" struct?
I can see there is function in our code base, but it's more clear that people can find the definition in the doc as well.

kaizhangNV: For the implementation purpose, we hope it's  very clear that which path is a special case that we could potentially remove in the future.

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,318 @@
...
+
+Note that initializer list of a single argument still translates directly into a constructor call and not a type cast. For example:
+```csharp
+void test()
+{
+  MyType t = {1};
+  // translates to:
+  // MyType t = MyType.__init(1);
+  // which is not
+  // MyType t = MyType(t)
```
kaizhangNV, line 156: I'm not quite understand this examples, what's the different between
`MyType t = MyType.__init(1);`
and
`MyType t = MyType(t)`

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,318 @@
...
+
+If the above code passes type check, then it will be used as the way to initialize `obj`.
+
+If the above code does not pass type check, Slang continues to check if `S` meets the standard of a "legacy C-style struct` type.
+A type is a "legacy C-Style struct" if all of the following conditions are met:
+- It is a user-defined struct type or a basic scalar, vector or matrix type, e.g. `int`, `float4x4`.
+- It does not contain any explicit constructors defined by the user.
+- All its members have higher or equal visibility than the type.
+- All its members are legacy C-Style structs or arrays of legacy C-style structs.
+In such case, we perform a legacy "read data" style consumption of the initializer list, so that the following behavior is valid:
```
kaizhangNV, line 170: Can "legacy C-style struct" have member init expression?
e.g.
```
struct S
{
      int a = 5;
}
```

csyonghe: Yes, it can.

docs/proposals/004-initialization.md
csyonghe: MyType.__init(1) calls the ctor directly.

MyType(1) means (MyType)1 which will first try a set of builtin coercion rules to convert 1 to MyType. If the builtin rules doesn't apply, then fallback to calling MyType.__init(1).

docs/proposals/004-initialization.md
csyonghe: Why? MyType1 shouldn't default initialize because it doesn't even have a user defined ctor.

docs/proposals/004-initialization.md
kaizhangNV: Oh, I see that we want to differentiate from
`Single argument constructor call` sub-section above?

csyonghe: Yes. 

MyType s = {1}
Always calls the ctor.

MyType(1) goes through type cast.

docs/proposals/004-initialization.md
```diff
@@ -0,0 +1,318 @@
...
+
+```csharp
+struct Inner { int x; int y; };
+struct Outer { Inner i; Inner j; }
+
+Outer o = {1, 2, 3}; // Initializes `o` into `{ Inner{1,2}, Inner{3,0} }`.
```
kaizhangNV, line 176: Is this already supported now?

csyonghe: Yes, this is already implemented in the legacy logic.

docs/proposals/004-initialization.md
kaizhangNV: I see.

docs/proposals/004-initialization.md
kaizhangNV: I can verify that this is already supported by current slang.

tangent-vector (approved):

docs/proposals/004-initialization.md
kaizhangNV: If this is the case, you may want to update this example
```
struct DefaultMember {
  int x = 0;
  int y = 1;
}
void test3()
{
  DefaultMember m; // `m` is uninitialized.
  DefaultMember m1 = {}; // calls `__init()`, initialized to `{0,1}`.
  DefaultMember m2 = {1}; // calls `__init(1)`, initialized to `{1,1}`.
  DefaultMember m3 = {1,2}; // calls `__init(1,2)`, initialized to `{1,2}`.
}
```
because it looks like it will go to the legacy initializer list logic

PR #5150, opened 2024-09-24
Fix the incorrect use of GLSL textureGather

jkwak-work:
`$p` was incorrectly removed on the previous commit. This commit brings it back and fixes the problem.

This is to fix the regression on CI.


csyonghe:
Can we have a test here to prevent future regressions?

csyonghe (approved):

jkwak-work:
> Can we have a test here to prevent future regressions?

Sorry. I missed this comment.
I will address it tomorrow in a separate PR.

PR #5155, opened 2024-09-25
Move texture format inference to frontend and add reflection api for it.

### LINKED ISSUE #5148: Add reflection API to query the texture format on a texture binding.

csyonghe:
Closes #5148.

This PR moves texture format inference from glsl/spirv backends to the frontend, which will add a `[format]` attribute if one doesn't exist.

This allows us to provide the reflection API to query the format of a texture.


jkwak-work (approved):
Looks good to me.

PR #5158, opened 2024-09-25
Implement texture_storage_Xd in WGSL

### LINKED ISSUE #5082: WGSL: Intrinsics: texture
15 functions

jkwak-work:
Closes #5082

This commit implements `texture_storage_Xd` in WGSL, which is similar to RWTextureXD in HLSL.

It is intresting that `texture_storage_Xd` doesn't take the shader type as its input argument at all.
Instead, it takes "texel format" enum value as its first template parameter, which can be found here:
https://www.w3.org/TR/WGSL/#storage-texel-formats

As an example, `texture_storage_2d<rg32uint, read_write>` expects vec4<u32> as a value type for `Load` and `Store`, where Z-component will be ignored and treated as zero and W-component will be treated always as 1.

WGSL requires `texture_storage_Xd` values to always be global variables and they cannot be passed into a function as a function argument.

### DIFF
--- a/source/slang/slang-emit-wgsl.cpp
+++ b/source/slang/slang-emit-wgsl.cpp
@@ -300,6 +300,9 @@ void WGSLSourceEmitter::emit(const AddressSpace addressSpace)
 
 static const char* getWgslImageFormat(IRTextureTypeBase* type)
 {
+    // You can find the supported WGSL texel format from the URL:
+    // https://www.w3.org/TR/WGSL/#storage-texel-formats
+    //
     ImageFormat imageFormat = type->hasFormat() ? (ImageFormat)type->getFormat() : ImageFormat::unknown;
     switch (imageFormat)
     {
@@ -319,7 +322,12 @@ static const char* getWgslImageFormat(IRTextureTypeBase* type)
     case ImageFormat::rgba32ui:    return "rgba32uint";
     case ImageFormat::rgba32i:     return "rgba32sint";
     case ImageFormat::rgba32f:     return "rgba32float";
-    default:                       return "rgba8unorm";
+    case ImageFormat::unknown:
+        // Unlike SPIR-V, WGSL doesn't have a texel format for "unknown".
+        return "rgba32float";
+    default:
+        // We may need to print a warning for types WGSL doesn't support
+        return "rgba32float";
     }
 }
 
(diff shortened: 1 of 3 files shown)


source/slang/slang-emit-wgsl.cpp
```diff
@@ -298,6 +298,31 @@ void WGSLSourceEmitter::emit(const AddressSpace addressSpace)
     }
 }
 
+static const char* getWgslImageFormat(IRTextureTypeBase* type)
+{
+    ImageFormat imageFormat = type->hasFormat() ? (ImageFormat)type->getFormat() : ImageFormat::unknown;
+    switch (imageFormat)
+    {
+    case ImageFormat::rgba8:       return "rgba8unorm";
```
csyonghe, line 306: No r8unorm/rg8unorm?

source/slang/slang-emit-wgsl.cpp
```diff
@@ -298,6 +298,31 @@ void WGSLSourceEmitter::emit(const AddressSpace addressSpace)
...
+    case ImageFormat::r32ui:       return "r32uint";
+    case ImageFormat::r32i:        return "r32sint";
+    case ImageFormat::r32f:        return "r32float";
+    case ImageFormat::rg32ui:      return "rg32uint";
+    case ImageFormat::rg32i:       return "rg32sint";
+    case ImageFormat::rg32f:       return "rg32float";
+    case ImageFormat::rgba32ui:    return "rgba32uint";
+    case ImageFormat::rgba32i:     return "rgba32sint";
+    case ImageFormat::rgba32f:     return "rgba32float";
+    default:                       return "rgba8unorm";
```
csyonghe, line 322: For read write textures, default should be rgba32f to match Vulkan behavior.

jkwak-work:
I rebased to include Yong's change.
But I still need to address the comment.
I need to go over each format more carefully.

source/slang/slang-emit-wgsl.cpp
csyonghe: r16unorm/rg16unorm/rgba16unorm and snorm formats as well.

csyonghe (dismissed):

source/slang/slang-emit-wgsl.cpp
jkwak-work: Right. One channel format can only be 32bit.
Two channel format can only be 64bit in total.

source/slang/slang-emit-wgsl.cpp
jkwak-work: I changed the default to "rgba32float" as suggested.

PR #5164, opened 2024-09-26
Add CI step to build WebAssembly version of Slang

### LINKED ISSUE #5117: Wasm build: Integrate to CI

jkwak-work:
Closes #5117

This adds a build step of WebAssembly version of Slang for all platforms and all configurations.
The new build step appears to take about double times of time compared to the regular Slang build step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,12 +25,15 @@ jobs:
         os: [linux, macos, windows]
         config: [debug, release]
         compiler: [gcc, clang, cl]
-        platform: [x86_64, aarch64]
+        platform: [x86_64, aarch64, wasm]
         exclude:
           # Default to x64, but aarch64 on osx
           - { os: linux, platform: aarch64 }
           - { os: windows, platform: aarch64 }
           - { os: macos, platform: x86_64 }
+          - { os: linux, config: debug, platform: wasm }
+          - { os: windows, platform: wasm }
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -94,6 +94,20 @@ jobs:
           platform: ${{matrix.platform}}
           config: ${{matrix.config}}
           build-llvm: true
+      - name: Build Slang for Wasm
```
csyonghe, line 97: We don't need to run this build for every platform matrix.
Instead, wasm should be one of the entries in `matrix.platform`, and this secion should be what "Build Slang" step is doing for that `wasm` platform.

jkwak-work: I see.
But `wasm` is an actual platform. Do we want to use Windows for that?
I can make wasm build for Windows Release and Debug if that's what you meant.

csyonghe: Let's use linux to build it.

csyonghe: And let's only do a release build.

jkwak-work: Ok. Linux Release it is, then.

aleino-nv (approved):

csyonghe (dismissed):

PR #5171, opened 2024-09-27
export CMAKE_SYSTEM_PROCESSOR for cross compilation builds

skallweitNV:
This should prevent `slang-rhi` from linking dawn on aarch64 target platform when using cross-compilation.


expipiplus1 (approved):

expipiplus1:
Thanks!

PR #5177, opened 2024-09-27
Fix l-value computation for subscript call.

### LINKED ISSUE #4971: Invalid Slang Code-Generation
Problem:
With the provided shader attached below, when compiled, the following is observed: 
1. Slang will compile invalid HLSL.
`static RWStructuredBuffer<int > test_val_0;` is created and assigned to nothing.

2. Slang will compile invalid SPIR-V.
```
error: line 24: In Logical addressing, variables may not allocate a pointer type
  %test_val = OpVariable %_ptr_Private__ptr_StorageBuffer_RWStructuredBuffer Private
```

3. Slang will ASSERT when compiling GLSL.
ASSERT while attempting to emit `IRHLSLStructuredBufferTypeBase`, which is invalid for GLSL targets.

Expected:
if this the repro-shader is an incorrect use-case: Slang warns/errors without creating invalid code-gen.
if this the repro-shader is a correct use-case: Slang should compile the code correctly.

To Repro:
1. Compile the shader below 3 times, 1 for each target: `-target glsl -entry computeMain -stage compute myTestShader.slang` and `-target hlsl -entry computeMain -stage compute myTestShader.slang` and `-target spirv -entry computeMain -stage compute myTestShader.slang` 

```
// myTestShader.slang
struct Test {
    RWStructuredBuffer<int> val;
    __subscript(int x, int y)->int
    {
        get { return val[x * 3 + y]; }
        set { val[x * 3 + y] = newValue; }
    }
}
Test test;

[shader("compute")]
[numthreads(32, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID) {
    test[1,1] = 1;
}
```

csyonghe:
Closes #4971.

This PR fixes a bug in determining l-value ness of an expression.

If we have:
```
struct Test
{
    __subscript(int x) { get; set; }
}

uniform Test t;
void main()
{
   t[0] = 2; // should error, because `t` is not l-value.
}
```

If the `setter` is marked as `[nonmutating]`, then the subscript call should be treated as l-value even if base object is immutable:
```
struct Test
{
    __subscript(int x) { get; [nonmutating] set; }
}

uniform Test t;
void main()
{
   t[0] = 2; // should work, because `t.__subscript.setter` is `[nomutating]` and therefore does not require `this` to be mutable.
}
```


tests/bugs/gh-4971.slang
```diff
@@ -0,0 +1,25 @@
...
+
+struct Test {
+    RWStructuredBuffer<int> val;
+    __subscript(int x, int y)->int
+    {
+        get { return val[x * 3 + y]; }
+        [nonmutating] set { val[x * 3 + y] = newValue; }
+    }
+}
+Test test;
```
jkwak-work, line 12: I think two tests can be combined,
```
//TEST:SIMPLE(filecheck=CHECK): -target spirv -DERROR_CASE
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -output-using-type
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-d3d11 -output-using-type

struct Test {
    RWStructuredBuffer<int> val;
    __subscript(int x, int y)->int
    {
        get { return val[x * 3 + y]; }
#if !defined(ERROR_CASE)
        [nonmutating]
#endif
        set { val[x * 3 + y] = newValue; }
    }
}
Test test;
```

jkwak-work (approved):
Looks good to me.

tests/bugs/gh-4971.slang
csyonghe: I think both ways are fine, personally it is just cleaner to me to keep things separate.

PR #5181, opened 2024-09-27
Fix bug in `translateGlslGlobalVar()`.

### LINKED ISSUE #5027: When targeting SpirV, Slang will misscompile global variables bound to a semantic when they are used in the same expression as WorkgroupSize().
Hi !

Slang misscompiles this code: 

```
RWStructuredBuffer<uint> out_buffer;

in uint3 gid : SV_GroupID;

[shader("compute")]
[numthreads(8, 8)]
void comp_main() {
    uint expr = WorkgroupSize().x + gid.x;
    out_buffer[0] = expr;
}
```

When compiled with `slangc -target spirv test.slang -o test.spv` then dumped into glsl using `spirv-cross test.spv`, I get this:

```
#version 450
layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;

layout(binding = 0, std430) buffer RWStructuredBuffer
{
    uint _m0[];
} out_buffer;

void main()
{
    uvec3 gid;
    out_buffer._m0[0] = uint(8 + int(gid.x));
}
```

Here `gid` isn't initialized, which leads to problems (first detected as a device lost due to an incorrect index computation leading to an out of bound access).

Weirdly this doesn't happen if I use `gid` directly without mixing it with `WorkgroupSize()` or if I take `SV_GroupID` as main argument. 
It isn't specific to `SV_GroupID`, other semantics are also affected. It appears to be specific to  `WorkgroupSize()` though. 

This is a regression, it didn't happen before v2024.1.24. A quick bisection seems to point to commit 0b5190226b3061597f0fd88ca6b1f304f2c6af38

csyonghe:
Closes #5027.


jkwak-work (approved):
Looks good to me.

PR #5182, opened 2024-09-28
Fix target switch lowering bug.

csyonghe:
Fix the lowering bug that can lead to IR validation assert on debug builds.


jkwak-work (approved):
Looks good to me

PR #5196, opened 2024-09-30
Disable the building of slang-llvm when targeting wasm platform

### LINKED ISSUE #5195: Don't build slang-llvm in CI when targeting wasm platform
PR #5164 added a new platform target, "wasm", for building WebAseembly build of Slang.
It is expected to take about 7 minutes but when the sccache fail to work, the build time goes above 30 minutes.
Most of the build time is spent for building slang-llvm locally, and slang-llvm is not needed for WebAssembly at the moment.
We should avoid building slang-llvm when targeting wasm platform in the ci workflow.

jkwak-work:
Closes #5195

Disable the building of slang-llvm when targeting wasm platform

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
           compiler: ${{matrix.compiler}}
           platform: ${{matrix.platform}}
           config: ${{matrix.config}}
-          build-llvm: true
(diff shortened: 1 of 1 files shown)


jkwak-work:
I looked at the workflow logs.
I can conform that the was build in ci workflow is skipping the build of slang-llvm while the other platforms are still working as before.

csyonghe (approved):

Issue #5191, opened 2024-09-30
countof funciton doesn't work in stdlib


expipiplus1:
insert into hlsl.meta.slang

```
void foo<each U>(expand each U args)
{
    static_assert(countof(U) == 1, "");
}
```

compile slang

observe that countof has resolved to bottom type

```
hlsl.meta.slang(21120): error 39999: ambiguous call to '==' with arguments of type (error, int)
    static_assert(countof(U) == 1, "");
                             ^~
core.meta.slang(2899): note 39999: candidate: func ==(uintptr_t, uintptr_t) -> bool
__intrinsic_op(288) bool operator==(uintptr_t left, uintptr_t right);
                                 ^~
core.meta.slang(2891): note 39999: candidate: func ==(uint64_t, uint64_t) -> bool
__intrinsic_op(288) bool operator==(uint64_t left, uint64_t right);
                                 ^~
core.meta.slang(2883): note 39999: candidate: func ==(uint, uint) -> bool
__intrinsic_op(288) bool operator==(uint left, uint right);
                                 ^~
core.meta.slang(2875): note 39999: candidate: func ==(uint16_t, uint16_t) -> bool
__intrinsic_op(288) bool operator==(uint16_t left, uint16_t right);
                                 ^~
core.meta.slang(2867): note 39999: candidate: func ==(uint8_t, uint8_t) -> bool
__intrinsic_op(288) bool operator==(uint8_t left, uint8_t right);
                                 ^~
core.meta.slang(2859): note 39999: candidate: func ==(double, double) -> bool
__intrinsic_op(288) bool operator==(double left, double right);
                                 ^~
core.meta.slang(2851): note 39999: candidate: func ==(float, float) -> bool
__intrinsic_op(288) bool operator==(float left, float right);
                                 ^~
core.meta.slang(2843): note 39999: candidate: func ==(half, half) -> bool
__intrinsic_op(288) bool operator==(half left, half right);
                                 ^~
core.meta.slang(2835): note 39999: candidate: func ==(intptr_t, intptr_t) -> bool
__intrinsic_op(288) bool operator==(intptr_t left, intptr_t right);
                                 ^~
core.meta.slang(2827): note 39999: candidate: func ==(int64_t, int64_t) -> bool
__intrinsic_op(288) bool operator==(int64_t left, int64_t right);
                                 ^~
hlsl.meta.slang(21120): note 39999: 4 more overload candidates
    static_assert(countof(U) == 1, "");
                             ^~
[...]

PR #5205, opened 2024-10-01
SizeOfLikeExpr has type int

expipiplus1:
As to why this ever worked in non-stdlib code, I still don't know

Fixes https://github.com/shader-slang/slang/issues/5191


expipiplus1:
why should this make `tests/hlsl-intrinsic/size-of/size-of-generic.slang` fail only for vulkan, and only on the ci machines (works locally), this change should have nothing to do with any backend...

csyonghe (approved):

jkwak-work:
> ~why should this make `tests/hlsl-intrinsic/size-of/size-of-generic.slang` fail only for vulkan, and only on the ci machines (works locally), this change should have nothing to do with any backend...~
> 
> oh they're passing now, slangwin4-2 strikes again?

This is failing on my local machine with the latest now.

PR #5206, opened 2024-10-01
Remove sccache

expipiplus1:


jkwak-work:
If we disable the cache, CI workflow will take a lot longer.
Because it will use the fallback mechanism of building slang-llvm locally.

If we are going to disable the cache, shouldn't we have a better way to avoid the local build of slang-llvm?
Probably we should try to download the prebuilt binary of slang-llvm as a replacement for the caching step?

expipiplus1:
the llvm build is cached separately, it has never been cached with sccache, so that shouldn't be a concern

bmillsNV:
Closing this request based on discussion in dev meeting. We don't need this one right now.

csyonghe (approved):

Issue #5192, opened 2024-09-30
[SP004]: Implement the translation from initialize list to constructor with visibility control


kaizhangNV:
This task is tracking the implementation of one topic in the design proposal of https://github.com/shader-slang/slang/blob/master/docs/proposals/004-initialization.md.

If a struct is initialized with initialize list, we will need to synthesize a constructor, and translate the initialize list to the call to such synthesized constructor.

PR #5215, opened 2024-10-02
SP004: implement initialize list translation to ctor

kaizhangNV:
First PR for #5192.

This is the first change to implement part of SP004, what this change does:

1. We synthesize a member-wise constructor for each struct follow the rules described in SP004.

2. Add logic to translate the initialize list to constructor invoke

TODO: what needs to be done next

1. We didn't fill the body of the synthesized constructor
  - This is relative easy task.

2. The fall back to legacy initialize list logic is not done
  - I found out that it's straight forward to put this logic after ResolveInvoke() call, because after ResolveInvoke() we will need to do a final 'coerce()' check to determine whether the candidate is applicable to use. If that fails, we can do the C-Style struct check and fall back to legacy logic.

3. Should we enable this for stdlib as well?
  - Our stdlib has some syntax that external users cannot use, so whether this change could break anything is unknown for now. It needs some extra effort to investigate and debugging the potential compile issue.

### DIFF
--- a/source/slang/slang-check-conversion.cpp
+++ b/source/slang/slang-check-conversion.cpp
@@ -201,21 +201,6 @@ namespace Slang
         return baseStructDeclRef;
     }
 
-    bool SemanticsVisitor::_hasExplicitConstructor(StructDecl* structDecl)
-    {
-        bool result = false;
-        for (auto ctor : getMembersOfType<ConstructorDecl>(getASTBuilder(), structDecl, MemberFilterStyle::All))
-        {
-            // constructor that is not synthesized must be user defined.
-            if (ctor.getDecl()->findModifier<SynthesizedModifier>() == nullptr)
-            {
-                result = true;
-                break;
-            }
-        }
-        return result;
-    }
-
     // TODO: We might need to find a good way to get the synthesized constructor instead of traversing all of constructors.
     ConstructorDecl* SemanticsVisitor::_getSynthesizedConstructor(StructDecl* structDecl)
     {
@@ -230,6 +215,114 @@ namespace Slang
         return nullptr;
     }
 
+    static StructDecl* _getStructDecl(Type* type)
+    {
+        if( as<VectorExpressionType>(type) ||
+            as<MatrixExpressionType>(type) ||
+            as<ArithmeticExpressionType>(type) ||
+            as<BuiltinType>(type))
+        {
+            return nullptr;
+        }
+
+        if(auto declRefType = as<DeclRefType>(type))
+        {
+            auto structDecl = as<StructDecl>(declRefType->getDeclRef());
+            return structDecl.getDecl();
+        }
+
+        return nullptr;
+    }
+
+    bool SemanticsVisitor::_cStyleStructBasicCheck(Decl* decl)
+    {
+        // 1. It has to be a user-defined struct type, or a basic scalar, vector or matrix type
+        if (isFromStdLib(decl))
+            return false;
+
+        if (auto varDecl = as<VarDecl>(decl))
+        {
+            auto type = varDecl->getType();
+            if( as<VectorExpressionType>(type) ||
+                as<MatrixExpressionType>(type) ||
+                as<BasicExpressionType>(type))
+                return true;
+            else
+                return false;
+        }
+
+        auto structDecl = as<StructDecl>(decl);
+        if (!structDecl)
+            return false;
+
+        // 2. It cannot have inheritance
+        if (structDecl->getMembersOfType<InheritanceDecl>().getCount() > 0)
+            return false;
+
+        // 3. It cannot have explicit constructor
+        if (_hasExplicitConstructor(structDecl))
+            return false;
+
+        // 4. All of its members have to have the same visibility as the struct itself.
+        DeclVisibility structVisibility = getDeclVisibility(structDecl);
+        for (auto varDeclRef : getMembersOfType<VarDeclBase>(getASTBuilder(), structDecl, MemberFilterStyle::Instance))
+        {
+            auto varDecl = varDeclRef.getDecl();
+            if (getDeclVisibility(varDecl) != structVisibility)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // TODO: We need to cache the result of this check, though it's not a heavy call.
+    bool SemanticsVisitor::isCStyleStruct(StructDecl* structDecl)
+    {
+        // rules 1-4 are checked in _cStyleStructBasicCheck for all the non-array members
+        if(!_cStyleStructBasicCheck(structDecl))
+            return false;
+
+        // 5. All its members are legacy C-Style structs or arrays of legacy C-style structs
+        for (auto varDeclRef : getMembersOfType<VarDeclBase>(getASTBuilder(), structDecl, MemberFilterStyle::Instance))
+        {
+            auto varDecl = varDeclRef.getDecl();
+
+            // if the member is an array, check if the element is legacy C-style rule.
+            if (auto arrayType = as<ArrayExpressionType>(varDecl->getType()))
+            {
+                auto* elementType = arrayType->getElementType();
+                ArrayExpressionType* nextType = nullptr;
+                while(nextType = as<ArrayExpressionType>(elementType))
+                {
+                    elementType = nextType->getElementType();
+                }
+                if(auto structDecl = _getStructDecl(elementType))
+                {
+                    if (!_cStyleStructBasicCheck(structDecl))
+                        return false;
+                }
+                else
+                {
+                    // if the element is not a struct, it has to be a scalar, vector or matrix.
+                    if( !as<VectorExpressionType>(elementType) &&
+                        !as<MatrixExpressionType>(elementType) &&
+                        !as<BasicExpressionType>(elementType))
+                    {
+                        return false;
+                    }
+                }
+            }
+            else
+            {
+                // all the other members still go through the basic check.
+                if (!_cStyleStructBasicCheck(varDecl))
+                    return false;
+            }
+        }
+        return true;
+    }
+
     Expr* SemanticsVisitor::_prepareCtorInvokeExpr(Type* toType, const SourceLoc& loc, const List<Expr*>& coercedArgs)
     {
         auto* varExpr = getASTBuilder()->create<VarExpr>();
@@ -240,9 +333,7 @@ namespace Slang
         constructorExpr->functionExpr = varExpr;
         constructorExpr->arguments.addRange(coercedArgs);
         constructorExpr->loc = loc;
-        auto resolvedConstructorExpr = CheckExpr(constructorExpr);
-
-        return resolvedConstructorExpr;
+        return constructorExpr;
     }
 
     // translation from initializer list to constructor invocation if the struct has constructor.
@@ -262,6 +353,7 @@ namespace Slang
                 if (_hasExplicitConstructor(toStructDeclRef.getDecl()))
                 {
                     auto ctorInvokeExpr = _prepareCtorInvokeExpr(toType, fromInitializerListExpr->loc, fromInitializerListExpr->args);
+                    ctorInvokeExpr = CheckTerm(ctorInvokeExpr);
                     if (outExpr && ctorInvokeExpr)
                     {
                         *outExpr = ctorInvokeExpr;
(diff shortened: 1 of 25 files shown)


source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
 
+    bool SemanticsVisitor::_hasExplicitConstructor(StructDecl* structDecl)
+    {
+        bool result = false;
+        for (auto ctor : getMembersOfType<ConstructorDecl>(getASTBuilder(), structDecl, MemberFilterStyle::All))
+        {
+            // constructor that is not synthesized must be user defined.
+            if (ctor.getDecl()->findModifier<SynthesizedModifier>() == nullptr)
+            {
+                result = true;
```
csyonghe, line 212: just `return true`.

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+                break;
+            }
+        }
+        return result;
```
csyonghe, line 216: `return false;`

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+    }
+
+    // TODO: We might need to find a good way to get the synthesized constructor instead of traversing all of constructors.
```
csyonghe, line 219: You can have a `SLANG_UNREFLECTED` on the structdecl to cache the result of this function.

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+            if (ctor.getDecl()->findModifier<SynthesizedModifier>() != nullptr)
+            {
+                return ctor.getDecl();
+            }
+        }
+
+        return nullptr;
+    }
+
+    Expr* SemanticsVisitor::_prepareCtorInvokeExpr(Type* toType, const SourceLoc& loc, const List<Expr*>& coercedArgs)
```
csyonghe, line 233: createCtorInvokeExpr

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+        return resolvedConstructorExpr;
+    }
+
+    // translation from initializer list to constructor invocation if the struct has constructor.
+    bool SemanticsVisitor::_invokeExprForExplicitCtor(Type* toType, InitializerListExpr* fromInitializerListExpr, Expr** outExpr)
+    {
+        if(auto toDeclRefType = as<DeclRefType>(toType))
+        {
+            auto toTypeDeclRef = toDeclRefType->getDeclRef();
+            if(auto toStructDeclRef = toTypeDeclRef.as<StructDecl>())
```
csyonghe, line 254: We need to unify the style, but most of the code we have puts a space between the keyword `if`, `for` etc. and the `(`.

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+        constructorExpr->functionExpr = varExpr;
+        constructorExpr->arguments.addRange(coercedArgs);
+        constructorExpr->loc = loc;
+        auto resolvedConstructorExpr = CheckExpr(constructorExpr);
+
+        return resolvedConstructorExpr;
+    }
+
+    // translation from initializer list to constructor invocation if the struct has constructor.
+    bool SemanticsVisitor::_invokeExprForExplicitCtor(Type* toType, InitializerListExpr* fromInitializerListExpr, Expr** outExpr)
```
csyonghe, line 249: createExplicitCtorInvokeExpr

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+            return true;
+
+        auto synthesizedConstructor = _getSynthesizedConstructor(structDecl);
+        SLANG_ASSERT(synthesizedConstructor);
+
+        List<Expr*> coercedArgs;
+        auto ctorInvokeExpr = _prepareCtorInvokeExpr(toType, fromInitializerListExpr->loc, fromInitializerListExpr->args);
+        if (ctorInvokeExpr)
+        {
+            // The reason we need to check the coercion again is that the ResolveInvoke() could still find us
```
csyonghe, line 301: We need to fix `ResolveInvoke` to not resolve into base class's ctor call.

source/slang/slang-check-conversion.cpp
```diff
@@ -201,11 +201,137 @@ namespace Slang
...
+            if(auto toStructDeclRef = toTypeDeclRef.as<StructDecl>())
+            {
+                structDecl = toStructDeclRef.getDecl();
+            }
+        }
+
+        if (!structDecl || isFromStdLib(structDecl))
+            return true;
+
+        auto synthesizedConstructor = _getSynthesizedConstructor(structDecl);
```
csyonghe, line 294: This is not used?
Perhaps we need a version of `_prepareCtorInvokeExpr` where the ctor decl is known so we don't redundantly resolve overloads in `ResolveInvoke`.

source/slang/slang-check-conversion.cpp
```diff
@@ -547,8 +673,27 @@ namespace Slang
...
+            return true;
+        }
+
+        // try to invoke the synthesized constructor if it exists
+        if (!_invokeExprForSynthesizedCtor(toType, fromInitializerListExpr, outToExpr))
+        {
+            // TODO: check if it's C-style initializer list
+            // if it's not, return error.
+            // if it is, return, fall-back to legacy logic
+            getSink()->diagnoseRaw(Severity::Error, "Synthesized Constructor is not found\n");
```
csyonghe, line 688: should turn into a proper diagnostic, instead of `diagnoseRaw`.

source/slang/slang-check-conversion.cpp
```diff
@@ -547,8 +673,27 @@ namespace Slang
...
+            return false;
+        }
+
+        // We will fall back to the legacy logic of initialize list.
         if(!_readAggregateValueFromInitializerList(toType, outToExpr, fromInitializerListExpr, argIndex))
+        {
```
csyonghe, line 694: `_readAggregateValueFromInitializerList` needs to be rewritten to turn into something that forms the argument list for the synthesized ctor call.

`_coerceInitializerList` should always set `outToExpr` to an invokeExpr of some ctor.

csyonghe: But this can be cleanly separated out into its own PR.

source/slang/slang-check-decl.cpp
```diff
@@ -218,6 +220,15 @@ namespace Slang
...
 
         void visitExtensionDecl(ExtensionDecl* decl);
+
+        // Synthesize the constructor declaration for a struct during header visit, as we
+        // need to have such declaration first such that the overloading resolution can lookup
+        // such constructor and complete the initialize list to constructor translation.
+        //
+        // We will defer the actual implementation of the constructor to the body visit, because
+        // we will have full information about each field in the struct during that stage.
+        void _synthesizeEmptyConstructor(StructDecl* structDecl);
```
csyonghe, line 230: EmptyConstructor is confusing name, should be named `synthesizeCtorSignature`.

source/slang/slang-check-decl.cpp
```diff
@@ -1960,13 +1971,32 @@ namespace Slang
...
+            for (auto i : ctor->getParameters())
+                if (!i->initExpr)
+                    return false;
+            return true;
+        };
+
+        // 1. default ctor must have definition
+        // 2. either 2.1 or 2.2 is safisfied
+        // 2.1. default ctor must have no parameters
+        // 2.2. default ctor can have parameters, but all parameters have init expr (Because we won't differentiate this case from 2.)
```
csyonghe, line 1987: I think we will consider 2.2 to be not a default ctor.

source/slang/slang-check-decl.cpp
```diff
@@ -6916,6 +6945,111 @@ namespace Slang
         }
     }
 
+    void SemanticsDeclBasesVisitor::_searchMembersWithHigherVisibility(StructDecl* structDecl, const DeclVisibility& ctorVisibility, List<VarDeclBase*>& resultMembers)
```
csyonghe, line 6948: DeclVisibility is a enum, so don't pass it by reference.

source/slang/slang-check-decl.cpp
```diff
@@ -6916,6 +6945,111 @@ namespace Slang
...
+            }
+        };
+
+        for (auto inheritanceMember : structDecl->getMembersOfType<InheritanceDecl>())
+        {
+            auto declRefType = as<DeclRefType>(inheritanceMember->base.type);
+            if (!declRefType)
+                continue;
+
+            auto structOfInheritance = as<StructDecl>(declRefType->getDeclRef().getDecl());
```
csyonghe, line 6968: can be named `baseTypeDecl`

csyonghe, line 6968: line 6964 through 6968 can be simplified as:
```
if (baseTypeDecl = isDeclRefTypeOf<StructDecl>(inheritanceMember->base.type)
```

source/slang/slang-check-decl.cpp
```diff
@@ -6916,6 +6945,111 @@ namespace Slang
...
+    {
+        // 1st condition is easy, we can just use the init expression as the default value.
+        if (varDecl->initExpr)
+        {
+            return varDecl->initExpr;
+        }
+
+        // For the 2nd condition, we need to check if the type is default initializable, if so,
+        // we can use the default constructor.
+        return constructDefaultInitExprForVar(visitor, varDecl);
```
csyonghe, line 6997: Feels more this should be named `constructDefaultInitExprForType`

source/slang/slang-check-decl.cpp
```diff
@@ -6916,6 +6945,111 @@ namespace Slang
...
+
+            if (!stopProcessingDefaultValues)
+                ctorParam->initExpr = _getParamDefaultValue(this, member);
+
+            if (!ctorParam->initExpr)
+                stopProcessingDefaultValues = true;
+
+            ctorParam->parentDecl = ctor;
+            StringBuilder prefix;
+            prefix << "param_" << paramIndex++ << "_";
```
csyonghe, line 7038: Don't add prefix here, because in the future we may want to support:
```
T v = {member0: expr, member1: expr}
```
that translates to:
```
T v = T(member0: expr, member1: expr)
```
If we added a prefix, this won't work anymore.

source/slang/slang-check-decl.cpp
```diff
@@ -6916,6 +6945,111 @@ namespace Slang
...
+        {
+            auto declRefType = as<DeclRefType>(inheritanceMember->base.type);
+            if (!declRefType)
+                continue;
+
+            auto structOfInheritance = as<StructDecl>(declRefType->getDeclRef().getDecl());
+            if (!structOfInheritance)
+                continue;
+
+            findMembers(structOfInheritance);
```
csyonghe, line 6972: For base types, the ctor parameters should be the base type's ctor's parameters, instead of base type's members.

For example, if I have
```
struct Base { int x, int y; __init(int p) {...} }

struct Derived : Base
{
        int z;
}
```

The synthesized ctor for `Derived` should be `__init(int p, int z)` instead of `__init(int x, int y, int z)`.

source/slang/slang-check-conversion.cpp
kaizhangNV: Definitely, it's changed in the later commit.

source/slang/slang-check-conversion.cpp
kaizhangNV: Yea. 
You may notice in my later commits of this PR, I don't change `_readAggregateValueFromInitializerList` this function at all, as I want to keep this function as it for the legacy logic only.

source/slang/slang-check-decl.cpp
kaizhangNV: But base type could have more than one ctors, how are we going to handle that?

source/slang/slang-check-decl.cpp
kaizhangNV: Changed the name in this PR.

source/slang/slang-check-conversion.cpp
kaizhangNV: done

source/slang/slang-check-conversion.cpp
kaizhangNV: update this for every location I have touched.

source/slang/slang-check-decl.cpp
kaizhangNV: Inheritance is a new issue we haven't clarify in the design doc.

We were aware of that there could be issues we were not able to capture before the implementation. So can we clarify this in the design doc and make this piece more clear, instead of making the notes in github conversation?

kaizhangNV:
I rebase this PR to my another PR #5231 which just refactor the logic of synthesizing the body of the default constructor.

This will make my next implementation easier.

source/slang/slang-check-decl.cpp
csyonghe: If base type has user defined explicit ctor, it means we cannot synthesize the ctor the derived type. This is consistent with ordinary fields: if we have a field member T whose type has explicit ctors and the member does not have a default expr, then we cannot synthesize the ctor for the type.

Feel free to update the design doc in this PR.

kaizhangNV: > if we have a field member T whose type has explicit ctors and the member does not have a default expr, then we cannot synthesize the ctor for the type.

I'm not quite sure if this could bring us some problem?

So for your previous example:
```
struct Base { int x, int y; __init(int p) {...} }         // this __init is an explicit ctor

struct Derive: Base
{
    int z;
}
```

in this case, Derive will not have a ctor (neither explicit nor synthesized), so it means that user cannot use initialize list
nor ctor experssion (`Derive d = Derive()`) to initialize it. Is this correct?

csyonghe: Correct, this is desired behavior.

kaizhangNV: Got it, thanks for explanation, I will update the design doc.

Issue #5229, opened 2024-10-04
WGSL: Global variable declared without specifying address space


aleino-nv:
This issue was split from https://github.com/shader-slang/slang/issues/5173.

Affected tests under tests/compute (currently excluded):

1. `global-init`

slang-test output for `global-init`:

    ...
    WGPU error: Error while parsing WGSL: :3:1 error: module-scope 'var' declarations that are not of texture or sampler types must provide an address space
    var gVar_0 : i32 = i32(16);
    ^^^^^^^^^^^^^^^^^^^^^^^^^^

     - While validating [ShaderModuleDescriptor]
     - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
    ...

Generated WGSL for `global-init`:

    @binding(0) @group(0) var<storage, read_write> outputBuffer_0 : array<i32>;

    var gVar_0 : i32 = i32(16);
    
    fn test_0( inVal_0 : i32) -> i32
    {
        return inVal_0 + gVar_0;
    }
    
    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var tid_0 : u32 = dispatchThreadID_0.x;
        outputBuffer_0[tid_0] = test_0(outputBuffer_0[tid_0]);
        return;
    }

aleino-nv:
I have a simple fix for this issue, but I can't submit it until pull-request https://github.com/shader-slang/slang/pull/5174 is merged.

PR #5236, opened 2024-10-08
WGSL emitter: Specify private address space for global non-handle variable declarations

aleino-nv:
Closes issue #5229.

### DIFF
--- a/source/slang/slang-emit-wgsl.cpp
+++ b/source/slang/slang-emit-wgsl.cpp
@@ -602,6 +602,21 @@ void WGSLSourceEmitter::emitVarKeywordImpl(IRType * type, IRInst* varDecl)
         m_writer->emit("storage, read");
         m_writer->emit(">");
     }
+    else if(varDecl->getOp() == kIROp_GlobalVar)
+    {
+        // Global ("module-scope") non-handle variables need to specify storage space
+
+        // https://www.w3.org/TR/WGSL/#var-decls
+        // "
+        // Variables in the private, storage, uniform, workgroup, and handle address
+        // spaces must only be declared in module scope, while variables in the function
+        // address space must only be declared in function scope. The address space must
+        // be specified for all address spaces except handle and function. The handle
+        // address space must not be specified. Specifying the function address space is
+        // optional.
+        // "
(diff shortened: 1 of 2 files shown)


csyonghe (approved):

source/slang/slang-emit-wgsl.cpp
```diff
@@ -602,6 +602,21 @@ void WGSLSourceEmitter::emitVarKeywordImpl(IRType * type, IRInst* varDecl)
...
+        // https://www.w3.org/TR/WGSL/#var-decls
+        // "
+        // Variables in the private, storage, uniform, workgroup, and handle address
+        // spaces must only be declared in module scope, while variables in the function
+        // address space must only be declared in function scope. The address space must
+        // be specified for all address spaces except handle and function. The handle
+        // address space must not be specified. Specifying the function address space is
+        // optional.
+        // "
+        m_writer->emit("<private>");
```
csyonghe, line 618: what about global groupshared var?

csyonghe:
I think the right way to do this is to run address space propagation pass to figure out the address space for all pointers, and during emit just grab the address space operand on  the pointer type.

source/slang/slang-emit-wgsl.cpp
aleino-nv: I believe this case should be handled by this line, already:

https://github.com/shader-slang/slang/pull/5236/commits/26f6c47adefa746359b1f73050e47503af64901a#diff-9dc6d2e5a48c5258bcab72361355a660ee872bd88326ae87c28d6a26d5db9fe6L590

I think groupshared maps to WGSL's `workgroup` address space:
https://www.w3.org/TR/WGSL/#address-spaces-workgroup

aleino-nv:
> I think the right way to do this is to run address space propagation pass to figure out the address space for all pointers, and during emit just grab the address space operand on the pointer type.

I filed https://github.com/shader-slang/slang/issues/5238 for this.
I suppose we can consider this a 'nice to have' so I didn't schedule this one yet and it's not part of any WGSL milestone. Let me know if you disagree.

PR #5239, opened 2024-10-08
Enable WebGPU tests in CI

aleino-nv:
This copies the .dll files necessary to enable wgpu and wgsl-spriv-asm tests in CI. (slang-test looks for these .dll's in order to determine what is supported)


csyonghe (approved):

csyonghe:
all failed wgpu test needs to be added to ignore list.

aleino-nv:
> all failed wgpu test needs to be added to ignore list.

It's not working as I would expect: I see

> FAILED test: 'tests/autodiff/bug-1.slang.1 syn (wgpu)'  

https://github.com/shader-slang/slang/actions/runs/11249895548/job/31277668049?pr=5239#step:5:152

in spite of 

https://github.com/shader-slang/slang/pull/5239/commits/99792f32c03035a8f4e7ab933aa5860f760f0912#diff-67561c6c931ea940fa4e948364074339ec3e5f2c0aef7d7417d0175338536414R9

csyonghe (approved):

aleino-nv:
Test slang step said that this fails:

> tests/serialization/obfuscated-serialized-module-test.slang.3 syn (dx12)

tests/serialization/obfuscated-serialized-module-test.slang.3.expected.txt
```diff
@@ -0,0 +1,4 @@
+0
```
aleino-nv, line 1: I have no idea why this file is suddenly required to get CI to pass. It shouldn't have anything to do with my changes I think?!

aleino-nv:
@csyonghe I added a workaround to disable WGPU tests in Debug mode. The issue is the presence of d3d12SDKLayers.dll in slang-binaries. When it's picked up instead of the system one, there can be mismatches with the SDK version.

I'm planning to fix this later. For now I would just like to at least enable WGPU tests in Release mode. That seems to finally work after adding obfuscated-serialized-module-test.slang.3.expected.txt, for some reason...

csyonghe (approved):

tests/serialization/obfuscated-serialized-module-test.slang.3.expected.txt
jkwak-work: It seems like this is expected.
A better approach would be to delete all files `tests/serialization/obfuscated-serialized-module-test.slang.*.expected.txt` and add the following lines in `tests/serialization/obfuscated-serialized-module-test.slang`
```
//TEST:COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-slang -compute -Xslang... -r tests/serialization/obfuscated-serialized-module.slang-module -obfuscate -X. -shaderobj

//BUF:0
//BUF-NEXT:1
//BUF-NEXT:2
//BUF-NEXT:3
```

PR #5242, opened 2024-10-08
Use user defined type name in glsl buffer declarations.

### LINKED ISSUE #5056: Add deterministic naming for Constant / Uniform buffers when targeting GLSL source output
Currently, Slang names uniform buffers on a 'first come, first served basis' when outputting GLSL source. Consider the following example Slang shader code:
```
struct InputAssembler
{
	float4 position : POSITION;
}

struct Vertex
{
	float4 position : SV_Position;
}

struct Pixel
{
    float4 color : SV_Target0;
}

struct TestData1
{
	float4x4 transform;
}

struct TestData2
{
	float4 color;
}

[vk::binding(0)]
ConstantBuffer<TestData1> test1 : register(b0);

[vk::binding(1)]
ConstantBuffer<TestData2> test2 : register(b1);

[shader("vertex")]
void Stage_Vertex(in InputAssembler input, out Vertex output)
{
	output.position = mul(input.position, test1.transform);
}

[shader("pixel")]
void Stage_Pixel(in Vertex input, out Pixel output)
{
    output.color = test2.color;
}
```
`TestData1` is only used during the vertex stage, and `TestData2` is only used during the pixel stage. Because of this the GLSL output will only emit the uniform buffer that is actually used in each stage (which is desired), however as a result, both TestData1 and TestData2 will have a generated symbol name of '_S1' .

Vertex Shader Ouput:
```
#line 27
layout(binding = 0)
layout(std140) uniform _S1
{
    mat4x4 transform_0;
}test1_0;
```

Pixel Shader Output:
```
#line 30
layout(binding = 1)
layout(std140) uniform _S1
{
    vec4 color_0;
}test2_0;
```

Why is this an issue? Some GLSL compilers (particularly one relevant to console games) compile a whole program from vertex+pixel shader code and will fail with an error since there are two different definitions of the uniform with "_S1".

Suggested solution:
instead of using the `_S(x)` naming scheme, generate a symbol based on the original constant buffer name in the slang source code. That way, even if only the relevant uniform buffers are emitted for individual shader stage GLSL code, no symbol collision can occur.

I am happy to help out and provide a fix, but would just need to confirm how your team would like to approach any change to naming symbols.

Thanks!

csyonghe:
Closes #5056.

### DIFF
--- a/source/slang/slang-emit-glsl.cpp
+++ b/source/slang/slang-emit-glsl.cpp
@@ -454,9 +454,8 @@ void GLSLSourceEmitter::_emitGLSLParameterGroup(IRGlobalParam* varDecl, IRUnifor
     }
 
     // Generate a name for the block.
-    m_writer->emit("block_");
-    m_writer->emit(getName(type->getElementType()));
-
+    m_writer->emit(_generateUniqueName(
+        (StringBuilder() << "block_" << getUnmangledName(type->getElementType()).getUnownedSlice()).getUnownedSlice()));
 
     auto elementType = type->getElementType();
     auto structType = as<IRStructType>(elementType);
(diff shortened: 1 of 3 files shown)


kaizhangNV (approved):
LGTM

PR #5243, opened 2024-10-09
Fix spirv lowering logic around pointer to unsized array.

### LINKED ISSUE #5061: Indexing unbounded struct array from pointer results in SPIRV validation error
Another silly one :)

Simplified repro:
```hlsl
struct MeshVertex {
    float3 Pos;
    float2 TexCoord;
};
struct MeshData { 
    float4x4 ModelMat;
    MeshVertex Vertices[];
};
struct DispatchParams {
    MeshData* Mesh;
    float3* Dest;
};

[vk::push_constant] DispatchParams pc;

[numthreads(64)]
void ComputeMain(uint tid: SV_DispatchThreadID) {
    pc.Dest[tid] = pc.Mesh->Vertices[tid].Pos;
}
```

spirv-val output:

```
error: line 85: OpAccessChain result type (OpTypeStruct) does not match the type that results from indexing into the base <id> (OpTypeStruct).
  %36 = OpAccessChain %_ptr_PhysicalStorageBuffer_MeshVertex_natural %33 %10
```

csyonghe:
Closes #5061.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -8430,6 +8430,11 @@ namespace Slang
     {
         auto typeExpr = paramDecl->type;
 
+        if (!as<ArrayExpressionType>(paramDecl->type) && doesTypeHaveTag(paramDecl->type, TypeTag::Unsized))
+        {
+            getSink()->diagnose(paramDecl, Diagnostics::paramCannotBeUnsized, paramDecl);
+        }
+
         // The "initializer" expression for a parameter represents
         // a default argument value to use if an explicit one is
         // not supplied.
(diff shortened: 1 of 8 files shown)


expipiplus1 (approved):
Do we have any designs on supporting this for webgpu?

wijiler:
Im having trouble building this branch
image
both ninja and vs2022 fail for me, just following the build instructions [here](https://github.com/csyonghe/slang/blob/gh-5061/docs/building.md)

expipiplus1:
@wijiler 
As a quick fix, pass `-DSLANG_SLANG_LLVM_FLAVOR=DISABLE`, this will prevent slang trying to download the prebuilt llvm which it's failing at.

It's failing because it's trying to fetch it from a three year old release https://github.com/shader-slang/slang/releases/tag/v0.19.8, it must detect 0.19.8 as the latest tag? Can you fetch tags maybe?

It should however still succeed in this case and fall back to fetching the latest release which does have that file: https://github.com/shader-slang/slang/releases/tag/v0.19.8 I'll open an issue

csyonghe:
> Do we have any designs on supporting this for webgpu?

This feature requires general pointer support, so it is not possible to implement it on any targets other than spirv right now. Let's not worry about it.

PR #5255, opened 2024-10-10
slang-test: retry failed test at the end.

csyonghe:


csyonghe:
We just retry it one more time, after all other tests are run, in serial.

kaizhangNV (dismissed):
LGTM

Issue #4758, opened 2024-07-30
Slang C API Deprecation - Step 1


csyonghe:
We need to start deprecate all `sp*` functions for shader compilation (not for reflection).
The first step is to create a `slang-deprecated.h` in `include/` directory, make `slang.h` `#include` `slang-deprecated.h`, and move all compilation related `sp*` functions + the `ICompileRequest` interface to the `slang-deprecated.h` file so new users don't get to see them in `slang.h`. We can't move `IGlobalSession::createCompileRequest` because it is inside an interface, so we may just mark the function as `[[deprecated]]` now.

This issue tracks this first step.

The next step is then to mark all functions in slang-deprecated.h as `[[deprecated]]` so users will start to receive compiler warnings. We may do this some time in 2025.

The final step is to remove everything in slang-deprecated.h, but this depends on how our users are migrated out of the C API. This will happen after 2026.

expipiplus1:
I've confirmed that OV are no longer using any of the to-be-deprecated functions, with the exception of `spGetBuildTagString` which I think is reasonable to keep.

PR #5301, opened 2024-10-15
 Move C interface from slang.h to slang-deprecated.h

expipiplus1:
spGetBuildTagString remains, because it's useful to have before the
    global session exists.

    This C API is used quite pervasively in the C++ helpers (for example
    slang::UserAttribute. It's not trivial to move these to
    slang-deprecated.h as they're entangled with some enums which are
    themselves used elsewhere in the compiler.

    The fact that these helpers use the C API can be viewed as an
    implementation detail for now, and this usage moved to slang-deprecated
    in due course, probably promptly as the next step in the C api deprecation

    Closes https://github.com/shader-slang/slang/issues/4758


csyonghe (approved):

csyonghe:
For now let's keep all reflection API as is.

Issue #3905, opened 2024-04-07
Improve README 4: Improve and cleanup the build instruction


kaizhangNV:
Cleanup the build instructions.

User should successfully build slang by following the commands, so we should list all the steps. Also need sub-sections to explain some "option" field, e.g. --enable-embed-stdlib=true, --no-progress, --build-glslang, etc..

Issue #5247, opened 2024-10-09
Improve behaviour when building with no tags


expipiplus1:
We should make sure we fall back to grabbing the latest slang-llvm

Initially from here:

              Im having trouble building this branch
![image](https://github.com/user-attachments/assets/eb843f80-50b4-4916-9413-2eaaf6291c87)
both ninja and vs2022 fail for me, just following the build instructions [here](https://github.com/csyonghe/slang/blob/gh-5061/docs/building.md)

_Originally posted in https://github.com/shader-slang/slang/issues/5243#issuecomment-2402503402_

PR #5302, opened 2024-10-15
review and revise building instructions

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/3905

Would be good to tackle this as a follow on https://github.com/shader-slang/slang/issues/5247


jkwak-work (approved):
Looks good to me.

PR #5315, opened 2024-10-15
Fix D3D12Core.dll loading problem for debug build

### LINKED ISSUE #5276: slang-test assert on usePassthru
This assert triggers

        case RenderApiType::WebGPU:
            target = SLANG_WGSL;
            SLANG_ASSERT(!usePassthru);
            break;

Affected test:

- tests/bugs/texture2d-gather.hlsl (**NOTE**: .hlsl extension)

I think the reason `usePassthru` is true is due to this special case logic above, for this particular test, due to file extension of `.hlsl`:

                if (languageRenderType != RenderApiType::Unknown)
                {
                    foundLanguageRenderType = languageRenderType;

                    // Use the pass thru compiler if these are the sources
                    usePassthru |= (argName == "hlsl" || argName == "glsl");

                    continue;
                }

### LINKED ISSUE #5305: Debug build of Slang fails to load d3d12SDKLayers.dll properly
**Problem description**
Debug build of Slang is failing to load d3d12SDKLayers.dll
The error message is,
```
D3D12_ERROR_INVALID_REDIST – 887e0003
```
This results in a situation where Windows/Debug configuration fails to support DX12 tests in CI workflow.
Note that Windows/Release works fine and the issue is limited to the Debug configuration.

**Quick triage**
It seems that Slang is loading DX12Core.dll from the c:\windows\system32 instead of one in the build directory.

**More detail**
Currently we have DX12Core.dll and d3d12SDKLayers.dll under a directory `external/slang-binaries/bin/windows-xxx`.
They are copied to the build directory where slang.dll gets generated.
When d3d12SDKLayers.dll is loaded, a matching DX12Core.dll should loaded from the same directory.
But this doesn't seem to be happening and DX12Core.dll appears to be loaded from c:\windows\system32.

This issue prevents us from running WGSL tests on CI actions for Windows/Debug configuration.
Once the issue is resolved, we should revert a [workaround commit](https://github.com/shader-slang/slang/pull/5239/commits/009a575234c34f1dc86ce4ec75dca2773618806d).

**Goal**
We need to get DX12 supported for Debug build.

jkwak-work:
D3D12Core.dll had been copied to a wrong directory and slang has been using D3D12Core.dll from the system directory, C:\windows\system32.

D3D12Core.dll has to be copied from `external/slang-binaries/bin/windows-x64` to `build/Release/bin/D3D12` not to `build/Release/bin`.
The same is true for the debug build and it had to be copied to `build/Debug/bin/D3D12` not `build/Debug/bin`.

It hasn't been a problem for Release build, because the debug-layer is not enabled for Release build and it didn't cause the version mismatching problem with D3D12SDKLayers.dll. The Release build was loaded from either `build/Release/bin` or from `C:\windows\system32`, and it didn't matter which one was used.

The Debug build, however, got into a problem where D3D12Core.dll was loaded from the system directory whereas D3D12SDKLayers.dll was loaded from `build/Debug/bin` and it failed to load D3D12.dll entirely. This caused D3D12 to be "Not supported" for "Windows/Debug" configuration. Note that our CI explicitly excludes DX12 tests for the "Windows/Debug" configuration with a command-line argument "-api all-dx12", and DX12 tests were going to be ignored anyway.

The actual problem was observed when WGPU is implemented. WGPU started printing explicit errors for the load failure of D3D12.dll.

See more detailed explanation:
  https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/#d3d12sdkpath-should-not-be-the-same-directory-as-the-application-exe

Closes #5305
Closes #5276

### DIFF
--- a/tools/slang-test/slang-test-main.cpp
+++ b/tools/slang-test/slang-test-main.cpp
@@ -1132,7 +1132,8 @@ static SlangResult _extractRenderTestRequirements(const CommandLine& cmdLine, Te
             break;
         case RenderApiType::WebGPU:
             target = SLANG_WGSL;
-            SLANG_ASSERT(!usePassthru);
+            nativeLanguage = SLANG_SOURCE_LANGUAGE_WGSL;
+            passThru = SLANG_PASS_THROUGH_TINT;
             break;
     }
 
@@ -4661,6 +4662,7 @@ SlangResult innerMain(int argc, char** argv)
             for (auto& test : context.failedFileTests)
             {
                 FileTestInfoImpl* fileTestInfo = static_cast<FileTestInfoImpl*>(test.Ptr());
+                TestReporter::SuiteScope suiteScope(&reporter, "tests");
                 TestReporter::TestScope scope(&reporter, fileTestInfo->testName);
                 reporter.addResult(TestResult::Fail);
             }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -487,13 +487,21 @@ endif()
 
 if(SLANG_ENABLE_PREBUILT_BINARIES)
     if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+        # DX Agility SDK requires the D3D12*.DLL files to be placed under a sub-directory, "D3D12".
+        # https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/#d3d12sdkpath-should-not-be-the-same-directory-as-the-application-exe
         file(GLOB prebuilt_binaries "${CMAKE_SOURCE_DIR}/external/slang-binaries/bin/windows-x64/*")
+        file(GLOB prebuilt_d3d12_binaries "${CMAKE_SOURCE_DIR}/external/slang-binaries/bin/windows-x64/[dD]3[dD]12*")
+        list(REMOVE_ITEM prebuilt_binaries ${prebuilt_d3d12_binaries})
         add_custom_target(
             copy-prebuilt-binaries ALL
             COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/$<CONFIG>/${runtime_subdir}
             COMMAND ${CMAKE_COMMAND} -E copy_if_different 
                 ${prebuilt_binaries}
                 ${CMAKE_BINARY_DIR}/$<CONFIG>/${runtime_subdir}
+            COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/$<CONFIG>/${runtime_subdir}/D3D12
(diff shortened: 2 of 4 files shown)


jkwak-work:
With this PR, the "[Windows/Debug](https://github.com/shader-slang/slang/actions/runs/11355301383/job/31584387450?pr=5315)" test in CI shows the following,
```
Compiling stdlib on debug build, this can take a while.
Compiling stdlib took 21.00 seconds.
Supported backends: fxc dxc glslang spirv-dis clang visualstudio gcc genericcpp llvm spirv-opt tint
Check dx12,d3d12: Supported
Check dx11,d3d11: Supported
Check cuda: Not Supported
```
Note that D3D12 is supported for Debug build.

Without this PR, the [previous](https://github.com/shader-slang/slang/actions/runs/11354984008/job/31583392799) output shows that D3D12 was not supported for Debug build
```
Compiling stdlib on debug build, this can take a while.
Compiling stdlib took 21.00 seconds.
Supported backends: fxc dxc glslang spirv-dis clang visualstudio gcc genericcpp llvm spirv-opt tint
Check dx12,d3d12: Not Supported
Check dx11,d3d11: Supported
Check cuda: Not Supported
Warning: D3D12CreateDevice failed
 - While initializing adapter (backend=BackendType::D3D12)
    at InitializeImpl (D:\a\webgpu-dawn-binaries\webgpu-dawn-binaries\build\_deps\dawn-src\src\dawn\native\d3d12\PhysicalDeviceD3D12.cpp:[93](https://github.com/shader-slang/slang/actions/runs/11354984008/job/31583392799#step:5:94))
```

tools/slang-test/slang-test-main.cpp
```diff
@@ -1132,7 +1132,7 @@ static SlangResult _extractRenderTestRequirements(const CommandLine& cmdLine, Te
             break;
         case RenderApiType::WebGPU:
             target = SLANG_WGSL;
-            SLANG_ASSERT(!usePassthru);
+            //SLANG_ASSERT(!usePassthru);
```
aleino-nv, line 1136: Please don't just disable this, if possible. I have a bug for investigating this:
https://github.com/shader-slang/slang/issues/5276

Last I checked this only triggers in a single test, so that test should then be disabled.

aleino-nv (dismissed):
LGTM but can the test be disabled instead of commenting out the assert?
The assert seems to be just because of some file extension specific logic, and should otherwise hold.

tools/slang-test/slang-test-main.cpp
csyonghe: I think we can just set passThru to `SLANG_PASS_THROUGH_NONE` here.

jkwak-work: I think I now have a proper fix for this.
But it will be on another PR.
Marking this as resolved.

PR #5316, opened 2024-10-15
Fix type checking on generic extensions.

### LINKED ISSUE #5140: Compiler crash on defining extension with additional constraints
The following code leads to an error:

```
public interface IPDE<T, let DIM : uint>
    where T : IFloat
{
    property float absorptionCoeff { get; set; }              // must be positive or equal to zero
    property bool robinConditionsArePureNeumann { get; set; } // set to false if Robin coefficients are non-zero anywhere

    // returns source term
    T source(vector<float, DIM> x);

    // returns Dirichlet boundary conditions
    T dirichlet(vector<float, DIM> x, bool returnBoundaryNormalAlignedValue = false);

    // returns Robin boundary conditions and coefficients
    T robin(vector<float, DIM> x, bool returnBoundaryNormalAlignedValue = false);          // dual purposes as values for Neuamnn BCs
    float robinCoeff(vector<float, DIM> x, bool returnBoundaryNormalAlignedValue = false); // must be positive or equal to zero
};

public extension<let DIM : uint, T : IFloat, S : IPDE<T, DIM>> S
{
    bool hasNonZeroRobinCoeff(vector<float, DIM> x)
    {
        return robinCoeff(x, true) > 0.0 ||
               robinCoeff(x, false) > 0.0;
    }
}
```

Error message:
```fatal error 39999: cyclic reference 'GenericTypeConstraintDecl'. public extension<let DIM : uint, T : IFloat, S : IPDE<T, DIM>> S```

Raising on behalf of @rohan-sawhney

csyonghe:
Add fcpw library to test suite.

Closes #5140.

### DIFF
--- a/source/slang/slang-check-inheritance.cpp
+++ b/source/slang/slang-check-inheritance.cpp
@@ -437,8 +437,6 @@ namespace Slang
             // representation would need to take into account canonicalization of
             // constraints.
 
-            ensureDecl(&visitor, genericDeclRef.getDecl(), DeclCheckState::CanSpecializeGeneric);
-
             if (auto extensionDecl = as<ExtensionDecl>(genericDeclRef.getDecl()->inner))
             {
                 if (isDeclRefTypeOf<GenericTypeParamDecl>(extensionDecl->targetType.type) == declRef)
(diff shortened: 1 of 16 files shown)


jkwak-work (approved):
Looks good to me.

But wouldn't it be better to have fcpw under the `external/fcpw` directory?

csyonghe:
I am not sure what is best. We are not really depending on it, but rather just consume it as a test, so putting it in test seems fit.
We also have approval from the original author to use it however we want.

PR #5320, opened 2024-10-16
Fix entrypoint naming in glsl backend.

csyonghe:
When there is
```
[shader("compute")]
void inner() { ... }

[shader("compute")]
void outer() { inner(); }
```

And the user trys to compile `outer` to glsl, we are currently generating two functions named `main` in the output glsl because both functions are marked as entrypoint, and there is the logic to emit all entrypoints as `main`. This is wrong if a user-declared entrypoint isn't being selected as the entrypoint in the current compilation session.


jkwak-work (approved):
Looks good to me

PR #5330, opened 2024-10-17
Cleanup definition of `printf`.

csyonghe:
Now that we have variadic generics, we can clean up the `printf` declaration to use variadic instead of overloads of different number of parameters. This means that we need to define `printf` as a slang IR opcode, so we can emit proper spirv, since we don't have a way to expand a variadic pack in a spirv asm block.


jkwak-work (approved):
Looks great to me.

PR #5344, opened 2024-10-18
Cross compilation using cmake toolchain file

skallweitNV:
Using `CMAKE_TOOLCHAIN_FILE` to setup cross compilation to aarch64 on Ubuntu. This allows to properly set the `CMAKE_SYSTEM_PROCESSOR` value to the target processor architecture, fixing issues when building `slang-rhi` which depends on that cmake system value.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,14 +68,12 @@ jobs:
       - name: Change dev tools to host arch (linux and macos)
         run: |
           if [[ "${{matrix.os}}" == linux* && "${{matrix.platform}}" == "aarch64" && "$(uname -m)" != "aarch64" ]]; then
-            export CC=aarch64-linux-gnu-gcc
-            export CXX=aarch64-linux-gnu-g++
+            export CMAKE_TOOLCHAIN_FILE=cmake/ubuntu-cross-aarch64.cmake
           fi
           CMAKE_OSX_ARCHITECTURES="${{matrix.platform}}"
           CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES//aarch64/arm64}
 
-          echo "CC=$CC" >> "$GITHUB_ENV"
-          echo "CXX=$CXX" >> "$GITHUB_ENV"
+          echo "CMAKE_TOOLCHAIN_FILE=$CMAKE_TOOLCHAIN_FILE" >> "$GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


cmake/ubuntu-cross-aarch64.cmake
```diff
@@ -0,0 +1,13 @@
+# Target operating system / architecture.
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR aarch64)
+
+# C/C++ compiler.
+set(CMAKE_C_COMPILER "/usr/bin/aarch64-linux-gnu-gcc")
+set(CMAKE_CXX_COMPILER "/usr/bin/aarch64-linux-gnu-g++")
```
expipiplus1, line 7: Does this have to be an absolute path, could we make it more portable by just having `aarch64-linux-gnu-gcc`?

.github/workflows/release.yml
```diff
@@ -68,14 +68,12 @@ jobs:
       - name: Change dev tools to host arch (linux and macos)
         run: |
           if [[ "${{matrix.os}}" == linux* && "${{matrix.platform}}" == "aarch64" && "$(uname -m)" != "aarch64" ]]; then
```
expipiplus1, line 70: It might be cleaner now to use `matrix.runs-on = ubuntu*` rather than checking for `os` if the toolchain file is ubuntu specific.

expipiplus1 (approved):
It's fine, but wouldn't it have been easier to just add CMAKE_SYSTEM_PROCESSOR to the env like we do with CMAKE_OSX_ARCHITECTURES?

(of course, this way does have the benefit of making it more obvious to any end user?)

skallweitNV:
> It's fine, but wouldn't it have been easier to just add CMAKE_SYSTEM_PROCESSOR to the env like we do with CMAKE_OSX_ARCHITECTURES?
> 
> (of course, this way does have the benefit of making it more obvious to any end user?)

That's what I initially tried. But cmake overwrites that value, you cannot set it either from an env variable or within the cmake file. Please try yourself, maybe I've missed something, but from what I gathered, the only way to set the system variables is through a toolchain file.

skallweitNV:
I just found out that `CMAKE_SYSTEM_PROCESSOR` doesn't get set properly on windows at all. So this is pointless. Instead I'm now determining the target architecture by querying the compiler, which should be more robust. See here: https://github.com/shader-slang/slang-rhi/blob/main/cmake/DetermineTargetArchitecture.cmake

Will open a new PR to update slang-rhi which should fix slang's release pipeline.

PR #5345, opened 2024-10-18
Update slang-rhi 

skallweitNV:
This adds a more robust way for determining the target architecture which should fix slang's release pipeline.


skallweitNV:
I also added aarch64 builds to slang-rhi CI build pipeline now, so things should be more stable moving forward.

PR #5356, opened 2024-10-20
Mark SPIRV artifact as non-executable.

### LINKED ISSUE #5353: File Output Sets Executable Bit
This is a pretty minor issue - it seems like on at least mac (probably all *nix) slang outputs the output file with the executable bit set (755 instead of 644). This caused a bunch of churn in git as windows' git commits the files as 644, but slang was outputting 755.

csyonghe:
Closes #5353.

This avoids slangc marking the output file as executable in linux.


jkwak-work (approved):
Looks good to me

PR #5369, opened 2024-10-22
Document derivative core module functions

### LINKED ISSUE #5307: document Global Functions - Atomic Derivative, bit operation functions

cheneym2:
Work on #5307

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -8554,6 +8554,7 @@ matrix<T, N, M> frexp(matrix<T, N, M> x, out matrix<int, N, M, L> exp)
 }
 
 /// Texture filter width.
+/// Calculates the sum abs(ddx(`p`)) + abs(ddy(`p`)).
 /// @param p The value to sum x and y partial derivative magnitudes for.
 /// @return The sum of abs(ddx(`p`)) and abs(ddy(`p`)).
 /// @remarks For SPIR-V, this function maps to `OpFwidth`.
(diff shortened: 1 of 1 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -7073,6 +7073,10 @@ ${{{{
 const char* diffDimensions[2] = {"x", "y"};
 for (auto xOrY : diffDimensions) {
 }}}}
+/// Take the partial derivative of `x` with respect to $(xOrY) in screen space.
+/// @param x The value to take partial derivative for.
+/// @return The partial derivative of `x`
+/// @remarks for SPIR-V, this function maps to `OpDPd$(xOrY)`.
```
csyonghe, line 7079: For SPIR-V

csyonghe: 'x' -> '$(xOrY)'

csyonghe (commented):
The x mentioned in the documentation needs to be replaced with the macro.

source/slang/hlsl.meta.slang
cheneym2: I'm going to set up the gem workflow for previewing the rendered markup today so I can see what you mean, but I thought the parameter was named `x` in all cases, so line 7076, 7077, and 7078 where I wrote `x` I think it's correct, but I need to see the rendered result anyway.

cheneym2: The rendered version looks like what I expected.  I think it's confusing to have the parameter named 'x' when dealing with x or y partial derivatives though.  The description of ddy is:
Take the partial derivative of `x` with respect to y in screen space.

Any objection to renaming the parameter p instead of x?

cheneym2: I've updated the parameter name to p, as to not be confused with x,y

csyonghe (approved):

Issue #5267, opened 2024-10-11
Slang release builds on windows has no debug info.


csyonghe:

expipiplus1:
You can build with the config `RelWithDebInfo` to get this (`cmake --build build --config RelWithDebInfo`, I think in VS there's like a dropdown selection for this?). Would you like for this to be made default for the `release` preset?

jkwak-work:
I see what Ellie says,
![image](https://github.com/user-attachments/assets/786f232d-07cc-46a3-84d1-347a3cbfd28d)
What would be a good way to verify that "RelWithDebugInfo" actually contains the debug info?

PR #5370, opened 2024-10-22
Put debug symbols in releases

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/5267


.github/workflows/ci.yml
```diff
@@ -110,12 +110,12 @@ jobs:
...
-              [ -f "build.em/Release/lib/libslang.a" ]
-              [ -f "build.em/Release/lib/libcompiler-core.a" ]
-              [ -f "build.em/Release/lib/libcore.a" ]
+              cmake --build --preset emscripten --config "$cmake_config" --target slang
+              [ -f "build.em/$cmake_config/lib/libslang.a" ]
+              [ -f "build.em/$cmake_config/lib/libcompiler-core.a" ]
+              [ -f "build.em/$cmake_config/lib/libcore.a" ]
           else
-            if [[ "${{ matrix.os }}" =~ "windows" && "${{ matrix.config }}" != "release" ]]; then
+            if [[ "${{ matrix.os }}" =~ "windows" && "${{ matrix.config }}" != "release" && "${{ matrix.config }}" != "releaseWithDebugInfo" ]]; then
```
jkwak-work, line 118: I think we can keep both cases "release" and "releaseWithDrbugInfo", here.
```
&& "${{ matrix.config }}" != "release" && "${{ matrix.config }}" != "releaseWithDebugInfo"
```

jkwak-work (dismissed):
Looks good to me.

csyonghe:
Have you verified this works on windows? From the build log it is still outputting to the Release folder.

.github/workflows/ci.yml
expipiplus1: sorry, what do you mean?

jkwak-work: Oops. Sorry I read the diff incorrectly.
Please ignore my comment above.

expipiplus1:
> Have you verified this works on windows? From the build log it is still outputting to the Release folder.

Where are you observing this, in CI it's compiling into RelWithDebInfo 
image


I changed things to make RelWithDebInfo the default config btw

csyonghe:
There are a  lot of warnings on windows builds:

```
cl : Command line warning D9025 : overriding '/MD' with '/MT'
```

expipiplus1:
yeah, I'm looking into that now

expipiplus1:
ok, looks like it's fixed

clean.sh
```diff
@@ -0,0 +1 @@
+../slang-ellie/clean.sh
```
csyonghe, line 1: what's this?

csyonghe:
Question: when this lands, what is our prebuilt binary package directory structure? Are things under `bin/RelWithDebInfo/` instead of `bin/Release`. I really don't like `RelWithDebInfo`.

expipiplus1:
image
This has been the layout for a while now, no `Release` or `RelWithDebInfo` either way

In the build tree it'll look like `build/RelWithDebInfo/bin/*` yes. It's very possible with CMake to change that to `build/bin` for example (or `build/Foo/bin`) if that's something you want?

csyonghe (approved):
On Linux, are the debug symbols embedded into the so? Can we separate the debug symbols into their own files like on windows?

csyonghe:
The package layout is good, no need for changes then.

expipiplus1:
It's definitely possible to do separate debug info on linux (but no on MacOS I think?), but it's not very common in my experience. 

In fact, I think at the moment we're embedding the debug info on Windows as well (sccache doesn't support separate debug info iirc)

csyonghe:
IIRC the binaries with debug info embedded is very big, something about 50MB? Not sure if it actually is a good idea to include it for all of our users, but it is probably still the better choice? We could provide a separate withDebInfo package for Linux and windows x64 only.

PR #5373, opened 2024-10-22
Document bit operations

### LINKED ISSUE #5307: document Global Functions - Atomic Derivative, bit operation functions

cheneym2:
Fixes #5307


csyonghe (approved):
If the function supports more than spirv, we should mention what they translate into on other targets (e.g. metal) as well.

Issue #4198, opened 2024-05-20
AD Enhancement: Enhance DCE to avoid situations where mutating primal methods are called multiple times during the reverse-mode pass, potentially causing slow-down


saipraveenb25:
Primal methods that mutate global/shared variables will remain in the backward pass because dead-code elimination will not remove them. This shouldn't affect the correctness of the backward pass because the necessary primal values are explicitly recorded onto the context, but it can severely slow-down the generated derivative code, especially if the calls are within a loop.

One way to solve this issue is to change the rules of DCE when processing the recompute blocks of the derivative methods and the primal context generation methods, since we don't care about global/shared mutation code.

Here is a test that reproduces this behavior:

```C
//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type -g0
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type -g0

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<float> outputBuffer;

groupshared float shared[10];
groupshared float d_shared[10];

float load(int i)
{
    return shared[i];
}

[BackwardDerivativeOf(load)]
void rev_d_load(int i, float value)
{
    d_shared[i] = value;
}

void store(int i, float value)
{
    shared[i] = value;
}

[BackwardDerivativeOf(store)]
void rev_d_store(int i, inout DifferentialPair<float> dpv)
{
    dpv = DifferentialPair<float>(dpv.getPrimal(), d_shared[i]);
}

[Differentiable]
[PreferRecompute]
float f(float a)
{
    return a * a;
}

[Differentiable]
void g()
{
    for (int i = 0; i < 10; i++)
    {
        store(i, f(load(i)));
    }
}

[numthreads(1, 1, 1)]
void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    float a[10] = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
    float da[10] = { 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f };

    for (int i = 0; i < 10; i++)
        shared[i] = a[i];
    for (int i = 0; i < 10; i++)
        d_shared[i] = da[i];

    __bwd_diff(g)();

    outputBuffer[0] = d_shared[0];
[...]

PR #5394, opened 2024-10-23
Assorted auto-diff enhancements for increased performance & more streamlined auto-diff results

saipraveenb25:
This PR makes a laundry list of changes:

- Adds `IROptimizableTypeDecoration` for types whose fields can be removed if there are no loads or extracts from it. The DCE pass now performs this field removal, though it is a **limited** analysis. All intermediate context types are optimizable by default.
- Adds `IRIgnoreSideEffectsDecoration` for synthesized functions that may have side-effects that are known to be duplicates (and therefore can be safely ignored for the purposes of DCE. All calls to primal-context generator functions are marked this way since the side-effects are effectively duplicates of the original function. The DCE pass has special handling for such functions. 
     Note: In the future, it may be better to make this a decoration of the call-site rather than the function.
     
- Explicitly removes the primal blocks from the `bwd_prop` function during `extractPrimalFunc`. This step is because, at this point, all instructions have either been recomputed in the recompute blocks or have been cloned into the primal_ctx function from where they are stored into the context struct. There should be nothing left in the primal blocks.

- Changes the recompute policy to recompute all functions without side-effects by default. Also classifies all `OpVar` as recompute as long as it can be recomputed.
- Fixes issues with 'use-chains' that are used to clone multiple insts at a time (where the intermediate insts are 'hoistable' and therefore cannot be cloned properly one at a time)
- Change header paths `<ankerl/unordered_dense.h>` to their relative path version (need to make sure slang-embed works correctly and serializes all the headers into one big prelude file)
- Disable several checkpoint-info-reporting tests: these were being reported incorrectly due to a bug in the function that computes struct sizes. Will re-enable in the future after verifying the results after the bugs are fixed for checkpoint-reporting.
- Fixed bug with the checkpoint-info-reporting (unwrap attributed types)
- Fix issue with function parameter source-locs not getting transferred into the IR properly.

Resolves: https://github.com/shader-slang/slang/issues/5387 and https://github.com/shader-slang/slang/issues/4198


csyonghe:
There is a CI failure on the WGPU test. We should either disable the test or figure out why it is failing.

source/slang/slang-ir-dce.cpp
```diff
@@ -294,6 +294,235 @@ bool isFirstBlock(IRInst* inst)
...
+}
+
+bool isStructEmpty(IRType* type)
+{
+    auto structType = as<IRStructType>(type);
+    if (!structType)
+        return false;
+    
+    UCount fieldCount = 0;
+    for (auto field : structType->getFields())
```
csyonghe, line 468: what if one of the field is also an empty struct?
should we just run type legalization first to cleanup empty struct types?

csyonghe (approved):

source/slang/slang-ir-dce.cpp
saipraveenb25: For now, I've changed it to check the fields recursively, but I'll do a follow-up PR to re-arrange the passes a bit. The optimizable type pass probably doesn't need to run for every call to `simplifyIR()`.

PR #5402, opened 2024-10-24
Fix incorrect debug assert in `getLanguagePrelude`.

### LINKED ISSUE #5401: `getEntryPointHash` assert hit when `targetDesc.format = SLANG_GLSL;`
With a target format like `SLANG_GLSL`, I get an InternalError while calling `getEntryPointHash`. 

The reason is this line in `source/slang/slang.cpp` `getLanguagePrelude()`
```cpp
SLANG_ASSERT(int(sourceLanguage) > int(SourceLanguage::Unknown) && int(sourceLanguage) < int(SourceLanguage::CountOf));
```

passThroughMode is none,  therefore SourceLanguage is Unknown. This only happens when I compile slang myself and SLANG_HAS_EXCEPTIONS is activated.

Not exactly sure what the intention here is.

csyonghe:
Closes #5401.


jkwak-work (approved):
Looks good to me

PR #5403, opened 2024-10-24
wasm: Add compile target option when creating slang session

kaizhangNV:
Also add a new interface to return spirv code which is binary,

because 'std::string ComponentType::getEntryPointCode' is not suitable for returning the binary data.

We use a more standard way that wrap the binary data by using emscripten::val as the return type.

### DIFF
--- a/source/slang-wasm/slang-wasm.cpp
+++ b/source/slang-wasm/slang-wasm.cpp
@@ -53,6 +53,7 @@ CompileTargets::CompileTargets()
         MAKE_PAIR(HLSL),
         MAKE_PAIR(WGSL),
         MAKE_PAIR(SPIRV),
+        MAKE_PAIR(METAL),
     };
 }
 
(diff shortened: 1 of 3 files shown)


source/slang-wasm/slang-wasm.cpp
```diff
@@ -38,15 +44,40 @@ GlobalSession* createGlobalSession()
...
-Session* GlobalSession::createSession()
+CompileTargets::CompileTargets()
+{
+#define MAKE_PAIR(x) { #x, SLANG_##x }
+
+    m_compileTargetMap = {
+        MAKE_PAIR(GLSL),
+        MAKE_PAIR(HLSL),
+        MAKE_PAIR(WGSL),
+        MAKE_PAIR(SPIRV),
```
csyonghe, line 55: Metal?

csyonghe (dismissed):

source/slang-wasm/slang-wasm.cpp
kaizhangNV: OK, will add

PR #5404, opened 2024-10-24
Use DebugDeclare instead of DebugValue.

### LINKED ISSUE #5398: Use OpDebugDeclare instead of OpDebugValue
We are using OpDebugValue as if it is storing a value into a debug variable, but this is not what the instruction means.

Given how our IR works, we need to use OpDebugDeclare instead.

csyonghe:
Closes #5398.

Fix missing DebugLine before loop conditions.
Fix duplicate DebugEntryPoint declarations.


jkwak-work (approved):
Looks good to me

PR #5408, opened 2024-10-25
Switch back to fetching upstream llvm project 

expipiplus1:
Now that sccache isn't evicting llvm from the cache we should be able to fetch it from upstream again (and archive our fork)

I did however keep the behaviour of using github credentials to fetch llvm


jkwak-work (approved):
Looks good to me

PR #5410, opened 2024-10-25
Add documentation for buffer types

kaizhangNV:
The preview page can be found here: https://kaizhangnv.github.io/stdlib-reference/types/buffer_types.html


source/slang/hlsl.meta.slang
```diff
@@ -4216,6 +4364,10 @@ struct $(item.name)
         }
     }
 
+    /// Load two 32-bit unsigned integers from the buffer at the specified location with alignment
+    /// of `uint2`.
```
csyonghe, line 4368: Speak out what the alignment is, I. This case it is 8.

source/slang/hlsl.meta.slang
```diff
@@ -5042,8 +5283,22 @@ struct $(item.name)
         }
     }
 
+    /// Increment the object's hidden counter.
```
csyonghe, line 5286: Add that this function is currently not implemented for non HLSL targets.

source/slang/hlsl.meta.slang
```diff
@@ -5023,8 +5259,13 @@ When generating code for other targets, this parameter is ignored and has no eff
 **/
 struct $(item.name)
 {
+    /// Decrements the object's hidden counter.
```
csyonghe, line 5262: Add that not implemented for HLSL targets.

source/slang/hlsl.meta.slang
```diff
@@ -5042,8 +5283,22 @@ struct $(item.name)
...
+    /// @param location The index of buffer.
+    /// @param[out] status The status of the operation.
+    /// @return The element at the specified index.
+    ///
+    /// @remarks
+    /// You can't access the output parameter `status` directly; instead,
+    /// pass the status to the `CheckAccessFullyMapped` intrinsic function.
+    /// `CheckAccessFullyMapped` returns TRUE if all values from the corresponding Sample,
+    /// Gather, or Load operation accessed mapped tiles in a tiled resource.
+    /// If any values were taken from an unmapped tile, `CheckAccessFullyMapped` returns FALSE.
```
csyonghe, line 5301: For all functions that returns status, the status we return is 0 on non HLSL targets.(they are not implemented).

source/slang/hlsl.meta.slang
```diff
@@ -4969,21 +5200,26 @@ ${{{{
...
     [ForceInline]
-    void Store<T>(int offset, T value, uint alignment)
+    void Store<T>(uint address, T value, uint alignment)
     {
-        __byteAddressBufferStore(this, offset, alignment, value);
+        __byteAddressBufferStore(this, address, alignment, value);
     }
 
+    /// Set four values to the buffer at the specified location, the address will be aligned
+    /// to the size of `T`.
```
csyonghe, line 5215: To the alignment of T.

source/slang/hlsl.meta.slang
```diff
@@ -106,12 +111,23 @@ struct AppendStructuredBuffer
...
 
-/// @category buffer_types
+//@public:
+/**
+Represents an opaque handle to a read-only buffer allocated in global memory that is indexed in bytes.
+ByteAddressBuffer can be used when working with raw buffers. Raw buffer can be viewed as a bag of bits to
+which you want raw access, that is, a buffer that you can conveniently access through chunks of one to
+four 32-bit typeless address values.
+ @remarks
+This type is supported natively when targeting HLSL.
```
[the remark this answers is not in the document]
kaizhangNV: ack

source/slang/hlsl.meta.slang
```diff
@@ -4066,9 +4170,15 @@ __intrinsic_type($(kIROp_HLSLConsumeStructuredBufferType))
 [require(cpp_cuda_glsl_hlsl_spirv, consumestructuredbuffer)]
 struct ConsumeStructuredBuffer
 {
+    /// Removes an element  from the end of the buffer.
```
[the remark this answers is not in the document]
kaizhangNV: ack.

source/slang/hlsl.meta.slang
kaizhangNV: This seems still 8.
I just checked:
source/slang/slang-ir-layout.cpp:236

source/slang/hlsl.meta.slang
kaizhangNV: ack.

source/slang/hlsl.meta.slang
kaizhangNV: ack

source/slang/hlsl.meta.slang
kaizhangNV: ack

source/slang/hlsl.meta.slang
kaizhangNV: ack

source/slang/hlsl.meta.slang
```diff
@@ -181,28 +227,46 @@ struct ByteAddressBuffer
...
         __target_switch
         {
         case hlsl: __intrinsic_asm ".Load2";
-        default:
-            return __byteAddressBufferLoad<uint2>(this, location, __naturalStrideOf<uint2>());
         }
     }
 
+    /// Load two 32-bit unsigned integers from the buffer at the specified location with alignment
+    /// of `uint2`, which is 4.
```
csyonghe, line 240: I believe this is 8?

source/slang/hlsl.meta.slang
```diff
@@ -231,28 +295,45 @@ struct ByteAddressBuffer
...
     {
         __target_switch
         {
         case hlsl: __intrinsic_asm ".Load3";
+        default:
+            return __byteAddressBufferLoad<uint3>(this, location, __naturalStrideOf<uint3>());
         }
     }
 
+    /// Load four 32-bit unsigned integers from the buffer at the specified location.
```
csyonghe, line 324: this should be 16.

csyonghe (commented):
I think the alignments for LoadAligned is not correct: 2-vector should be 8 bytes (offset must be multiple of 8 too), and 4-vector should be 16.

kaizhangNV:
> I think the alignments for LoadAligned is not correct: 2-vector should be 8 bytes (offset must be multiple of 8 too), and 4-vector should be 16.

I think offset is different from alignment.
If you see some definition here:
void __byteAddressBufferStore<T>(RWByteAddressBuffer buffer, int offset, int alignment, T value);

Alignment put the restriction on the address.

Also, I checked the alignment by looking at our code at

source/slang/slang-ir-layout.cpp:236

```
    case kIROp_VectorType:
    {
        auto vecType = cast<IRVectorType>(type);
        IRSizeAndAlignment elementTypeLayout;
        getSizeAndAlignment(optionSet, rules, vecType->getElementType(), &elementTypeLayout);
        *outSizeAndAlignment = rules->getVectorSizeAndAlignment(elementTypeLayout, getIntegerValueFromInst(vecType->getElementCount()));
        return SLANG_OK;
    }
```

outSizeAndAlignment do shows that uint{2,3,4} have alignment of 4.

source/slang/hlsl.meta.slang
kaizhangNV: uint4 is four 16bits integer?

source/slang/hlsl.meta.slang
```diff
@@ -153,6 +184,21 @@ struct ByteAddressBuffer
         }
     }
 
+    /// Load two 32-bit unsigned integers from the buffer at the specified location
+    /// with additional alignment.
+    ///@param location The input address in bytes.
+    ///@param alignment Specifies the alignment of the location, which must be a multiple of 4.
```
csyonghe, line 190: 8

source/slang/hlsl.meta.slang
csyonghe: it is four 32 bit integers, and LoadAlign will align to the vector boundary, which is 16 bytes.

csyonghe:
OK, I think the current behavior is LoadAligned<T> will treat the buffer as an array of T, and the offset must be at the boundary of T's stride. Therefore, the result will be undefined if you do `LoadAligned<float4>(4)`.

You can use slang-playground to compile this code to spirv:
```
RWByteAddressBuffer bb;

[numthreads(1,1,1)]
void computeMain()
{
    var v = bb.LoadAligned<float4>(16);
    bb.Store(0, v);
}
```
You can see in the spirv we have:
```
%computeMain = OpFunction %void None %3
          %4 = OpLabel
         %11 = OpAccessChain %_ptr_StorageBuffer_v4float %bb %int_0 %int_1
         %16 = OpLoad %v4float %11
         %17 = OpAccessChain %_ptr_StorageBuffer_v4float %bb %int_0 %int_0
               OpStore %17 %16
               OpReturn
               OpFunctionEnd
```
Which is treating `bb` as `float4[]`.

csyonghe:
Looks like we are generating wrong code for composite types...
Try 
```
RWByteAddressBuffer bb;

struct VV
{
    float4 v1;
    float2 v2;
};
[numthreads(1,1,1)]
void computeMain()
{
    var v = bb.LoadAligned<VV>(16);
    bb.Store(0, v);
}
```

The load for the `v2` field is at a completely invalid address.

csyonghe:
Let's put in the document that this currently only works for scalar, vector and matrix types.

csyonghe:
and the semantics is that it assumes the address to load/store is aligned at the stride boundary of the vector type.

kaizhangNV:
OK, I will update the document accordingly.

kaizhangNV:
I tried the code like:

```
ByteAddressBuffer byteBuffer;
RWStructuredBuffer<uint2> output;

[shader("compute")]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
    uint2 out;
    out = byteBuffer.Load2Aligned(4);
    output[0] = out;
}
```
 and it generates:

```
%10 = OpAccessChain %_ptr_StorageBuffer_uint %byteBuffer %int_0 %int_1
%15 = OpLoad %uint %10
%17 = OpAccessChain %_ptr_StorageBuffer_uint %byteBuffer %int_0 %int_2
```

and you can see that, we treat the whole buffer as uint type, so the alignment will always use 4 as the alignment.
I tried uint2, uint3, and uint4, same behavior. They're all aligned to 4.


But if I give an input location that is aligned to size of the vector, it will generate treat the buffer as the expecting types, e.g.

```
byteBuffer.Load2Aligned(8);  => OpAccessChain %_ptr_StorageBuffer_v2uint %byteBuffer %int_0 %int_1

byteBuffer.Load3Aligned(12);  => OpAccessChain %_ptr_StorageBuffer_v3uint %byteBuffer %int_0 %int_1 

byteBuffer.Load4Aligned(16); => OpAccessChain %_ptr_StorageBuffer_v3uint %byteBuffer %int_0 %int_1 
```

so it looks to me that, as long as the minimum alignment is met (which is 4), it can still generate the correct result.

kaizhangNV:
So as long as we define that

```
out = byteBuffer.Load2Aligned(4); 

=>

%10 = OpAccessChain %_ptr_StorageBuffer_uint %byteBuffer %int_0 %int_1
%17 = OpAccessChain %_ptr_StorageBuffer_uint %byteBuffer %int_0 %int_2
```

is an unexpected behavior, otherwise, alignment of 4 should be valid statement.

kaizhangNV:
@csyonghe add a new commit to update all the aligned version of load and store, please review.
For other load and store methods, doc remains unchanged.

csyonghe (approved):

Issue #5335, opened 2024-10-17
Swap out all uses of stdlib for core or "standard library" to core module and standard modules throughout docs & code


bmillsNV:
Terminology
Slang Standard Modules → the modules you get when you download slang
Slang Core Module — the module that starts every slang file — thing that can’t opt out of that has all our core types/functions/etc that every module starts with

We don’t want to use stdlib or “library” anywhere. Slang does modules. So we should use modules.

jkwak-work:
I have three questions regarding the terminology change.

1. Does it include the name of repo like [shader-slang/stdlib-reference](https://github.com/shader-slang/stdlib-reference)?
If yes, I will rename the repo.

2. Does it include the name of projects?
   There are a few project who name includes `stdlib`.
   Should they be changed too?
   - slang-embedded-stdlib
   - slang-embedded-stdlib-source
   - slang-no-embedded-stdlib
   - slang-no-embedded-stdlib-source
   - generate_stdlib_headers

3. Does it include the name of files?
   There are two files whose name include `stdlib`.
   - slang-stdlib.cpp
   - slang-stdlib-textures.cpp

jkwak-work:
The answer during the meeting yesterday was,
1. No, we don't want to rename the repo yet. And when we rename it later, it will be "standard-reference".
2. Yes, we should rename the project names.
3. Yes, we should rename the file names.
4. Yes, we should rename the enum and constant variable names.

jkwak-work:
I like to share some update, although it may make sense only to myself.

I locally renamed all filenames, enum names and project names.
It compiles and slang-test runs all green.

I still need to do the followings,
- rename `slang-stdlib.bin` to `slang-core-module.bin`.
- rename the command-line arguments that use the keyword `stdlib` such as `-load-stdlib` and `-save-stdlib`
- rename function names that uses the keyword, `stdlib`.
- modify comments to use either "core module" or "standard module" depending on the context.
- modify documents to use either "core module" or "standard module" depending on the context.
- rename `stdlib` in some of python, bash and pwsh scripts.

jkwak-work:
I figured that `compileStdLib()` is just for a core-module, and it should be `compileCoreModule()`.
But `loadStdLib()` and `saveStdLib()` are for multiple modules so it will have to be renamed to be `loadStandardModules()` and `saveStandardModules()`.

[...]

PR #5411, opened 2024-10-25
Replace stdlib with core-module on files and projects

jkwak-work:
This commit renames the files and projects to prefer "core-module" over "stdlib".

The directory name `source/slang-stdlib` needs to be renamed too, and there will be another commit for it soon.


jkwak-work:
This PR is for https://github.com/shader-slang/slang/issues/5335, but it wouldn't close the issue yet.

I prefer to make multiple smaller PRs.
And I still need to work more on it.

csyonghe (approved):

PR #5421, opened 2024-10-28
Improvements to formatting bot

expipiplus1:
It now opens a PR against the target branch, rather than committing directly. this means we can give the bot far fewer permissions.

Also, it formats using the script in the destination branch, which by virtue of the actions being hosted in this repository is something we've signed off on, so we should be less concerned with the bot executing arbitrary code from a PR. The disadvantage of this is that one can't change external/formatting.sh and also call `/format` to apply these changes in the same pr


jkwak-work (approved):
Looks good to me

Issue #5372, opened 2024-10-22
[CTS] constant-folding not working for the switch-case labels


jkwak-work:
**Problem description**
One of recent changes (PR #5341) caused a regression.
The expression in a switch-case label is no longer folded at the compile-time.
It caused CTS failures and four tests are disabled by [PR #8](https://github.com/shader-slang/VK-GL-CTS/pull/8)

**Repro**
The following shader should compile without error,
```
#version 310 es
layout(location = 0) in highp vec4 a_position;
layout(location = 1) in highp vec4 a_coords;

layout(location = 0) out mediump vec4 v_color;
layout (std140, set=0, binding=0) uniform buffer0 { highp int ui_two; };

void main (void)
{
    gl_Position = a_position;
    highp vec4 coords = a_coords;
    mediump vec3 res = vec3(0.0);

    const int t = 2;
    switch (ui_two)
    {
        case int(0.0): res = coords.xyz; break;
        case 2-1:      res = coords.wzy; break;
        case 3&(1<<1): res = coords.yzw; break;
        case t+1:      res = coords.zyx; break;
    }

    v_color = vec4(res, 1.0);
}
```
Currently it prints the following error message.
```
$ slangc.exe test.slang.vert
test.slang.vert(17): error 39999: expression does not evaluate to a compile-time constant
        case int(0.0):  res = coords.xyz;   break;
```

**Goal**
Fix the regression and re-enable the disabled CTS tests.

PR #5436, opened 2024-10-29
Constant-fold for the type-casting in switch-case labels

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/5372

When the expresion used as a switch-case labels is a type-casting from floating-point type to an integer type, the floating-point literal also need to be constant-folded.

### DIFF
--- a/source/slang/slang-check-expr.cpp
+++ b/source/slang/slang-check-expr.cpp
@@ -2069,12 +2069,13 @@ IntVal* SemanticsVisitor::tryConstantFoldExpr(
         if (!isValidCompileTimeConstantType(substType))
             return nullptr;
 
-        IntVal* val = nullptr; 
+        IntVal* val = nullptr;
         if (auto floatLitExpr = typeCastOperand.as<FloatingPointLiteralExpr>())
         {
-            // When explicitly casting from float type to integer type, let's fold it as an integer value.
+            // When explicitly casting from float type to integer type, let's fold it as an integer
+	    // value.
             const IntegerLiteralValue value = IntegerLiteralValue(floatLitExpr.getExpr()->value);
-            val = m_astBuilder->getIntVal(m_astBuilder->getFloatType(), value);
+            val = m_astBuilder->getIntVal(substType, value);
         }
         else
         {
--- a/source/slang/slang-check-expr.cpp
+++ b/source/slang/slang-check-expr.cpp
@@ -2069,17 +2069,16 @@ IntVal* SemanticsVisitor::tryConstantFoldExpr(
         if (!isValidCompileTimeConstantType(substType))
             return nullptr;
 
-        IntVal* val = nullptr;
-        if (auto floatLitExpr = typeCastOperand.as<FloatingPointLiteralExpr>())
+        IntVal* val = tryConstantFoldExpr(typeCastOperand, kind, circularityInfo);
+        if (!val)
         {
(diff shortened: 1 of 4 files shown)


source/slang/slang-check-expr.cpp
```diff
@@ -2068,7 +2068,19 @@ IntVal* SemanticsVisitor::tryConstantFoldExpr(
...
+        IntVal* val = nullptr; 
+        if (auto floatLitExpr = typeCastOperand.as<FloatingPointLiteralExpr>())
+        {
+            // When explicitly casting from float type to integer type, let's fold it as an integer value.
+            const IntegerLiteralValue value = IntegerLiteralValue(floatLitExpr.getExpr()->value);
+            val = m_astBuilder->getIntVal(m_astBuilder->getFloatType(), value);
+        }
+        else
+        {
+            val = tryConstantFoldExpr(typeCastOperand, kind, circularityInfo);
```
csyonghe, line 2081: Will this lead to infinite recursion?

jkwak-work: I don't see how it can become infinite recursion.

The function, tryConstantFoldExpr, is recursively called. But it peels one step for each call.

And that is not the code change I am making.
My change only adds a line to handle the case where the value is a floating-point literal.

csyonghe: Ah I see, yeah this should work.

source/slang/slang-check-expr.cpp
```diff
@@ -2068,7 +2068,19 @@ IntVal* SemanticsVisitor::tryConstantFoldExpr(
...
         if (!isValidCompileTimeConstantType(substType))
             return nullptr;
-        auto val = tryConstantFoldExpr(typeCastOperand, kind, circularityInfo);
+
+        IntVal* val = nullptr; 
+        if (auto floatLitExpr = typeCastOperand.as<FloatingPointLiteralExpr>())
+        {
+            // When explicitly casting from float type to integer type, let's fold it as an integer value.
+            const IntegerLiteralValue value = IntegerLiteralValue(floatLitExpr.getExpr()->value);
+            val = m_astBuilder->getIntVal(m_astBuilder->getFloatType(), value);
```
csyonghe, line 2077: the type here should be substType, not float type.

jkwak-work: After replacing it with `substType`, some of tests started failing.

The previous change of mine was copied from how Boolean type was handled in `SemanticsVisitor::tryConstantFoldExpr` ([link](https://github.com/shader-slang/slang/blob/368c2d6eb4c5049dccc1291d3350b4539cdd3c85/source/slang/slang-check-expr.cpp#L1989))
```
    if (auto boolLitExpr = expr.as<BoolLiteralExpr>())
    {
        // If it's a boolean, we allow promotion to int.
        const IntegerLiteralValue value = IntegerLiteralValue(boolLitExpr.getExpr()->value);
        return m_astBuilder->getIntVal(m_astBuilder->getBoolType(), value);
    }
```

So I think it should be `m_astBuilder->getFloatType()` not substType.

jkwak-work: I think you are right. It should be `substType`.

But then, I am not sure what the existing test started failing.

csyonghe (approved):
You can post "/format" in the thread to let the bot format the change for you.

csyonghe (approved):

PR #5455, opened 2024-10-30
Add design doc for write-only textures.

csyonghe:


jkwak-work (dismissed):
Looks good to me.
But it wasn't clear to me how "write-only" can solve the problem "read-write" has until I check the table that Kai referred to.
https://www.w3.org/TR/webgpu/#texture-format-caps
It appears that "write-only" and "read-only" supports more texel types compared to "read-write".

tangent-vector (approved):
Looks good. Only question I have is why there's this prefix of `kCoreModule_*` on the resource access enum values.

csyonghe:
The entire `_Texture` type and argument values to pass to the generic parameters are all internal now, and the enum in this case doesn't really have a presence in the actual slang module yet. They are all referenced through backquotes as in `$(kCoreModule_ResourceAccessReadOnly)`, so this should be considered as implementation detail.

PR #5461, opened 2024-10-30
Find slangc in more places in doc generation script

expipiplus1:


csyonghe:
Because I wrote it and as a terrible linux user I can only write powershell...

csyonghe (approved):

PR #5478, opened 2024-11-01
Allow enum cases to be used in attribute arguments.

### LINKED ISSUE #5477: Enums should be compile time constants
I'm attempting to use a shared header between host and shader code to specify binding locations, rather than relying on reflection. Ideally I'd have an enum defined like
```
enum BufferIndices {
        CameraStuff,
        LightingStuff,
        ModelStuff,
};
```

which I could reference in the shader with 
```
[[vk::binding(BufferBindings::CameraStuff)]]
StructuredBuffer<MyCameraInfo> cameraStuff;
```

Currently, this fails with error 39999: expression does not evaluate to a compile-time constant. However, if I explicitly set the values for the enum variants, it compiles successfully, so it seems like this should be allowed to work. Obviously its not a serious blocker since I can just set the values manually, but it would be nice to not need to do that.

csyonghe:
Closes #5477.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -6602,6 +6602,7 @@ bool SemanticsVisitor::checkInterfaceConformance(
     {
         if (!isAssociatedTypeDecl(requiredMemberDecl.getDecl()))
             continue;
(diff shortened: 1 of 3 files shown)


jkwak-work (approved):
Looks good to me

PR #5480, opened 2024-11-01
Revert uint<->int implicit cast cost to prefer promotion to unsigned.

### LINKED ISSUE #5473: Comparison between int and uint is being cast to int, instead of uint
I have a piece of code ~ to this 

`if (abs((int)frameNumber - (int)timeStamp) > age)`
where `age` is uint

And Slang compiles this to
`if ((int)abs((int)frameNumber - (int)timeStamp) > (int)age)`

This differs from C like languages. C standard says: `Otherwise, if the operand that has unsigned integer type has rank greater or
equal to the rank of the type of the other operand, then the operand with
signed integer type is converted to the type of the operand with unsigned
integer type.`

csyonghe:
This is a breaking change, but the change is to make Slang's overload resolution aligned with C specification.

Basically, when there is a binary operator that takes (int, int) or (uint, uint), and the argument is of type (int, uint), the expected behavior is to convert the `int` to `uint` instead of the other way around.

Closes #5473.

### DIFF
--- a/source/slang/slang-ast-support-types.h
+++ b/source/slang/slang-ast-support-types.h
@@ -127,10 +127,10 @@ enum : ConversionCost
 
     // Same-size size unsigned->signed conversions are potentially lossy, but they are commonly
     // allowed silently.
-    kConversionCost_SameSizeUnsignedToSignedConversion = 250,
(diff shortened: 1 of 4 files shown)


jkwak-work (approved):
Looks good to me.

Issue #5228, opened 2024-10-04
WGSL: Passing runtime-sized arrays (RWStructuredBuffer) by value


aleino-nv:
This issue was split from https://github.com/shader-slang/slang/issues/5173.

Affected tests under tests/compute (currently excluded):

- `func-resource-param`

Output from slang-test, for `func-resource-param`:

    ...
    WGPU error: Error while parsing WGSL: :5:25 error: runtime-sized arrays can only be used in the <storage> address space
    fn helper_0( buffer_0 : array<i32>,  index_0 : i32) -> i32
                            ^^^^^^^^^^
    
    :5:14 note: while instantiating parameter buffer_0
    fn helper_0( buffer_0 : array<i32>,  index_0 : i32) -> i32
                 ^^^^^^^^
    
    
     - While validating [ShaderModuleDescriptor]
     - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
    ...

Generated WGSL for `func-resource-param`:

    @binding(1) @group(0) var<storage, read_write> inputBuffer_0 : array<i32>;
    
    @binding(0) @group(0) var<storage, read_write> outputBuffer_0 : array<i32>;
    
    fn helper_0( buffer_0 : array<i32>,  index_0 : i32) -> i32
    {
        return buffer_0[index_0];
    }
    
    fn test_0( val_0 : i32) -> i32
    {
        var _S1 : i32 = helper_0(inputBuffer_0, val_0);
        return _S1 + val_0;
    }
    
    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var _S2 : u32 = dispatchThreadID_0.x;
        var outVal_0 : i32 = test_0(i32(_S2));
        outputBuffer_0[_S2] = outVal_0;
        return;
    }

aleino-nv:
@csyonghe This is one reason I had that weird logic for `RWStructuredBuffer` (and similar) in the WGSL emitter. This used to work but I guess the pointer overhaul/cleanup broke it.

csyonghe:
I think we should translate `StructuredBuffer` types consistently to `ptr<storage, array<T>>` in all places, including both global and function parameter declarations.

aleino-nv:
> I think we should translate `StructuredBuffer` types consistently to `ptr<storage, array<T>>` in all places, including both global and function parameter declarations.

[...]

PR #5483, opened 2024-11-04
Make structured buffers illegal parameter types in WGSL

aleino-nv:
Structured buffer types translate to array types in the WGSL emitter. WGSL doesn't allow passing runtime-sized arrays to functions.

This closes issue #5228.


source/slang/slang-ir-specialize-resources.cpp
```diff
@@ -67,7 +67,10 @@ struct ResourceParameterSpecializationCondition : FunctionCallSpecializeConditio
             else
                 return isIllegalGLSLParameterType(type);
         }
-
+        else if (targetRequest->getTarget() == CodeGenTarget::WGSL)
```
csyonghe, line 70: We need a `isWGPUTarget()` that takes account for all variants like WGSL_SPIRV_ASM, WGSL_SPIRV etc.

source/slang/slang-ir-specialize-resources.cpp
```diff
@@ -1349,4 +1352,16 @@ bool isIllegalSPIRVParameterType(IRType* type, bool isArray)
     }
     return false;
 }
+
+bool isIllegalWGSLParameterType(IRType* type)
+{
+    // The WGSL emitter translates structured buffer types to runtime-sized array types.
+    // Functions cannot take parameters of runtime-sized array type in WGSL.
```
csyonghe, line 1359: It should be more than this. Basically every illegal glsl/spirv parameter type is also illegal wgsl parameter type.

We can simply have this function call into isIllegalGLSLParameterType for now.

aleino-nv: Ok, done. I'd personally rather just copy-paste the function in case these need to diverge at any point.

I don't think it's obvious that these two sets of IR types should be equal for GLSL and WGSL, in all present and future cases.

expipiplus1 (dismissed):

PR #5488, opened 2024-11-04
Update user-guide top bar nav links.

csyonghe:


jkwak-work (approved):
Looks good.

I didn't notice they were swapped.

PR #5489, opened 2024-11-04
Fix issue of infinity float literal

### LINKED ISSUE #5280: WGSL: infinity float literal not working
Affected tests:

- tests/bugs/inf-float-literal.slang

It seems a float suffix of 'f' is appended to 'inf':

    WGPU error: Error while parsing WGSL: :13:19 error: unresolved value 'inff'
                a_0 = inff;

kaizhangNV:
Closes #5280.


csyonghe (approved):

kaizhangNV:
I noticed that this change doesn't work, as WGSL doesn't allow any constant infinity during compile time.
1.0/0.0 will report compile error.

csyonghe:
NaN is not supported by the wgsl spec, but we should be able to work around the compiler error by emitting something like:
```
let tmp = 0.0;
let nan = 1.0/tmp;
```

kaizhangNV:
That's correct

But I don't know how to emit the code for this. As the code is emit for a float literal.

So we might need to wrap this as a inline function. Don't know if there is better option?

csyonghe:
An easy way to do this is to copy `ensurePrelude` from `slang-emit-metal.cpp`, and call
```
ensurePrelude(R"(fn getNan()->f32 {var tmp = 0.0f; return 1.0f/nan;}\n");
```
before emitting `getNan()` for the nan literal.

kaizhangNV:
Ah, that's great.
I thought about prelude, but I'm not aware of that we have prelude for wgsl.

kaizhangNV:
This new change should work:
image

source/slang/slang-emit-wgsl.cpp
```diff
@@ -26,6 +26,39 @@
 namespace Slang
 {
 
+// In WGSL, expression of "1.0/0.0" is not allowed, it will report compile error,
```
aleino-nv, line 29: I don't know how solid the foundations for infinity/etc are in our other back-ends, but at least for WGSL there seems to be little/no foundation.

https://www.w3.org/TR/WGSL/#differences-from-ieee754

I suppose we can add this support just for the test here, but what are we really testing if we have basically no guarantees from WGSL?

aleino-nv: "Implementations may assume that overflow, infinities, and NaNs are not present at runtime."

csyonghe: I am OK with us generating out of spec code that works on majority of the platforms that most of our users care about. We can promise our user nothing, but still provide a decent solution that works most of the time, instead of enforcing restrictions that just adds friction.

aleino-nv: I suspected that's the policy for Slang, but good to have it confirmed.

While on the subject, what's the plan in case Slang does start specifying some of this stuff?
Are we fine with breaking apps that may be relying on behavior that Slang never promised?

csyonghe (approved):

Issue #5282, opened 2024-10-14
WGSL: Missing return statement


aleino-nv:
Affected tests:
- tests/hlsl-intrinsic/scalar-float.slang
- tests/hlsl-intrinsic/vector-float.slang

WGPU eror:

    WGPU error: Error while parsing WGSL: :29:1 error: missing return at end of function

kaizhangNV:
The actual problem is we emit
```
log10()
```

to an empty function in wgsl

```
fn log10_0( x_0 : f32) -> f32
{
}
```


this can be easily reproduced here: 

https://shader-slang.com/slang-playground/?target=SPIRV&code=eJxNkEEKwjAQRfc5xUcQEis2AXVTPIInUBfFkBKonVJjUUrv7iRW2kAmYebNz5_4R0tdQFuXn6qjV2MLIXryFm3nm3AufSOVGIQAL07A2zdO8Sa1KlI2BVdTGeBiiYENpNlphxxyz6dickkFxjTnF_15DvcgO2EB2Ql3ekr37xxSnDX6y41FBpgtjI77FzTGYiaTTLRaU2W07C_s7aaQ8eMHN7kfZw9xPHqFaTzu3sAc_wbSd8jV2l6b1TZiXBi_HYpNqw

PR #5491, opened 2024-11-05
Fixing the wrong implementation of some math intrinsic

kaizhangNV:
Close the issue #5282

The root cause of the issue is that log10 is not supported in wgsl. So add the implementation.

Also ldexp in wgsl doesn't support float type exponent, so fix the implementation of the intrinsic.


tests/hlsl-intrinsic/scalar-float.slang
```diff
@@ -4,7 +4,7 @@
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj
 //DISABLED_TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-cuda -compute -shaderobj
-//DISABLE_TEST(compute):COMPARE_COMPUTE_EX:-wgpu
+//TEST(compute):COMPARE_COMPUTE_EX:-wgpu -compute -shaderobj
```
aleino-nv, line 7: I think you should also remove a line from `tests/expected-failure-github.txt`

tests/hlsl-intrinsic/vector-float.slang
```diff
@@ -3,7 +3,7 @@
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -dx12 -use-dxil -output-using-type -shaderobj
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -output-using-type -shaderobj
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-cuda -compute -output-using-type -shaderobj
-//DISABLE_TEST(compute):COMPARE_COMPUTE_EX:-wgpu
+//TEST(compute):COMPARE_COMPUTE_EX:-wgpu -compute -output-using-type -shaderobj
```
aleino-nv, line 6: I think you should also remove a line from `tests/expected-failure-github.txt`

csyonghe (approved):

Issue #5275, opened 2024-10-14
Slang-RHI: Intel driver crashes in vkCreateDevice during DeviceImpl::initVulkanInstanceAndDevice


aleino-nv:
Note again that this is for the **intel** driver, `igvk64.dll`.
Edit: However, there are issues even when running on discrete GPU: https://github.com/shader-slang/slang/issues/5275#issuecomment-2410225134

I ran `slang-test -bindir %BINDIR% -category full -api wgpu tests/bugs/atomic-coerce.slang`.
I don't understand why slang-test is creating a Vulkan device.

Affected tests:

- tests/bugs/atomic-coerce.slang
- tests/bugs/frexp-double.slang
- tests/bugs/frexp.slang
- tests/bugs/generic-type-duplication.slang

csyonghe:
slang-test will attempt to create a device at start with all possible APIs to determine which APIs are available on the system.

aleino-nv:
> slang-test will attempt to create a device at start with all possible APIs to determine which APIs are available on the system.

Ok I see.

    in slang-test.exe: Microsoft C++ exception: VK::Exception at memory location 0x000000AEAD598A58


Should we wrap this in a catch and just conclude that VK is not available in this case? Currently it seems to tear down the whole test.

aleino-nv:
For the moment I'll set slang-test to use the discrete GPU on my machine.

aleino-nv:
> For the moment I'll set slang-test to use the discrete GPU on my machine.

That didn't help much. Still crashing but now instead on WGPU device/instance destructors. Need to investigate this case as well.

cheneym2:
There is already try, catch guards in render-test's innerMain() function around _innerMain(), which should catch crashes, such as from vkCreateDevice.  I find that it doesn't work though.  The exception is still thrown and uncaught.  I am able to force a crash by providing bad input to vkCreateDevice, such as by nulling out pQueueCreateInfos:

```
        deviceCreateInfo.pQueueCreateInfos = nullptr;        //&queueCreateInfo;
        deviceCreateInfo.enabledExtensionCount = uint32_t(deviceExtensions.size());
        deviceCreateInfo.ppEnabledExtensionNames = deviceExtensions.data();
        if (m_api.vkCreateDevice(m_api.m_physicalDevice, &deviceCreateInfo, nullptr, &m_device) != VK_SUCCESS)
```

[...]

PR #5495, opened 2024-11-05
Catch all exceptions

cheneym2:
In MSVC, the /EHsc flag is used by default, it causes only
C++ (synchronous) exceptions to be caught by try/catch
blocks. The /EHa flag can instead be used to catch both
synchronous C++ exceptions as well as structured
asynchronous exceptions such as those seen in segfaults
or other typical bugs.

Using /EHa allows render-test to not crash completely if
there is a buggy graphics driver in the system.

Issue 5275


cmake/CompilerFlags.cmake
```diff
@@ -152,6 +152,9 @@ function(set_default_compile_options target)
 
     add_supported_cxx_flags(${target} PRIVATE ${warning_flags})
 
+    # Support try/catching both asynchronous and synchronous exceptions
+    add_supported_cxx_flags(${target} PRIVATE /EHa)
```
jkwak-work, line 156: Does it work for gcc?

cheneym2: I don't think it does, though I don't know what the gcc equivalent is.  I was basically trying this out in CI to see what happened.  I think I need to do more than just filter it for MSVC though.  glslang is compiled with "/EHs" (IIRC) and there's an error thrown for conflicting "/EHa" and "/EHs" flags there.

So I think I need to limit this flag to the render-test project on MSVC.

csyonghe:
Is there a reason to use structural exception handling? For slang-test, we already run render-test in a separated process to isolate the crashes, so it shouldn't matter if render-test crashes during CI testing, and slang-test can still resume.

csyonghe:
Allowing render-test to exit gracefully with a fail return status is definitely nice to have so we can definitely try to enable this if nothing else breaks. But I think we will be OK not having it if this is too difficult to set up.

cheneym2:
I think I just need to restrict the option to MSVC now, since it's making the linux builds barf.

I find that slang-test crashes completely when it encounters an issue like what Anders reported in https://github.com/shader-slang/slang/issues/5275, even when using "-use-test-server" in slang-test.

Is there a distinction between "render-test" and "render-test-tool.dll" maybe?  Here's the stack trace of the artificial segfault I caused to have something to debug.

```
>	render-test-tool.dll!rhi::vk::DeviceImpl::initVulkanInstanceAndDevice(const rhi::NativeHandle * handles, bool enableValidationLayer) Line 986	C++
 	render-test-tool.dll!rhi::vk::DeviceImpl::initialize(const rhi::DeviceDesc & desc) Line 1027	C++
 	render-test-tool.dll!rhi::createVKDevice(const rhi::DeviceDesc * desc, rhi::IDevice * * outRenderer) Line 512	C++
 	render-test-tool.dll!rhi::_createDevice(const rhi::DeviceDesc * desc, rhi::IDevice * * outDevice) Line 373	C++
 	render-test-tool.dll!rhi::RHI::createDevice(const rhi::DeviceDesc & desc, rhi::IDevice * * outDevice) Line 408	C++
 	render-test-tool.dll!_innerMain(Slang::StdWriters * stdWriters, slang::IGlobalSession * session, int argcIn, const char * const * argvIn) Line 1407	C++
 	render-test-tool.dll!innerMain(Slang::StdWriters * stdWriters, slang::IGlobalSession * sharedSession, int inArgc, const char * const * inArgv) Line 1486	C++
 	slang-test.exe!spawnAndWaitSharedLibrary(TestContext * context, const Slang::String & testPath, const Slang::CommandLine & cmdLine, Slang::ExecuteResult & outRes) Line 837	C++
 	slang-test.exe!_getAvailableRenderApiFlags(TestContext * context) Line 1331	C++
 	slang-test.exe!_runTestsOnFile(TestContext * context, Slang::String filePath) Line 4160	C++
 	slang-test.exe!runTestsInDirectory::__l2::<lambda>(Slang::String file) Line 4402	C++
 	slang-test.exe!runTestsInDirectory(TestContext * context, Slang::String directoryPath) Line 4432	C++
 	slang-test.exe!innerMain(int argc, char * * argv) Line 4828	C++
 	slang-test.exe!main(int argc, char * * argv) Line 4899	C++

```

cmake/CompilerFlags.cmake
jkwak-work: My question was more of if gcc prints a warning that may get treated as an error; something like unrecognized argument.
I gave a quick try on my WSL, which I believe uses either gcc or clang, and I didn't see any warning or error messages.
I thought there would be but I am not sure why I didn't see any.

cheneym2: Googling, I couldn't tell what SEH support there is in any version of gcc.  Sounds like maybe it's at least tolerated on some version of gcc.  On linux though, it does make g++ barf, so I'm still working on getting this option applied for only MSVC.  I don't know enough about the SEH support in gcc on windows to try to keep it enabled for non-MSVC.

csyonghe: I believe SEH is a windows only thing.

csyonghe:
@cheneym2 The stack trace you are showing seems to be that the test is being run directly from the slang-test process and not through the test-server process. When `-use-test-server` is enabled, slang-test will start `test-driver.exe`, which loads and calls `render-test-tool.dll` instead.

cmake/CompilerFlags.cmake
cheneym2: Yeah, SEH is windows only.  Not sure if that makes it MSVC-only though, though I think it is.  IIUC, we only support using MSVC on the slang windows build, so it's not an important distinction.  The only CI protected build that barfs with "/EHa" flag are the linux builds.

cheneym2:
> @cheneym2 The stack trace you are showing seems to be that the test is being run directly from the slang-test process and not through the test-server process. When `-use-test-server` is enabled, slang-test will start `test-driver.exe`, which loads and calls `render-test-tool.dll` instead.

My stack trace is from a non-server run, true.  However, even with "-use-test-server" slang-test behaves the same way with this crash.

csyonghe:
Ah, I see the problem is in `_getAvailableRenderApiFlags`, this is the step where we are trying to determine whether an API/backend is available by attempting to create a device of that API type. If device creation leads to fatal crashes, we currently have no way to recover from it.

Issue #4985, opened 2024-09-03
WGSL: WGSL doesn't support arrays in uniform memory, with element strides less than 16 bytes


aleino-nv:
The tests/compute/buffer-layout.slang sample contains a struct with an element of type `float2 c[2];`, and the struct is used in a constant buffer.

The corresponding WGSL member declaration is `array<vec2<f32>, i32(2)>`, so the array element stride is `8`.
WGSL requires that array elements have alignment `16`, but it doesn't silently change the layout of the array.
Instead it prints an error message suggesting to use vec4 in the array type, instead.

    Error while parsing WGSL: :14:22 error: 'uniform' storage requires that array elements are aligned to 16 
    bytes, but array element of type 'vec2<f32>' has a stride of 8 bytes. Consider using a vec4 instead.
        @align(16) c_0 : array<vec2<f32>, i32(2)>,
                         ^^^^^^^^^^^^^^^^^^^^^^^^

We should work around this.
If we instead emit `array<S, i32(2)>` where `S` is a struct containing the `vec2<f32>` (or whatever other type with stride less than 16), then it should work because the size of `S` will get aligned up to `16`.

aleino-nv:
@bmillsNV Can we find someone to work on this for Sprint 13?

aleino-nv:
> @bmillsNV Can we find someone to work on this for Sprint 13?

As discussed during yesterday's sync meeting, this is not needed for WGSL milestone 1.

PR #5497, opened 2024-11-05
[WGSL] Enable arbitrary arrays in uniform buffers.

csyonghe:
Closes [#4985](https://github.com/shader-slang/slang/issues/4985)

WGSL does not allow declaring arrays of scalars or vectors whose alignment is not a multiple of 16 in uniform buffers.
This means that if the user defines a `float[3]` or `float3x2` in the uniform buffer, we will be generating invalid code.

The fix here is to extend `lower-buffefr-element-type` pass to convert float/float2 arrays to float4 arrays when mapping a type to a buffer storage type, and make sure the insert the packing/unpacking logic for such translated arrays.


source/slang/slang-ir-lower-buffer-element-type.h
```diff
@@ -8,6 +8,14 @@ class TargetProgram;
 struct IRTypeLayoutRules;
 struct IRType;
 
+struct BufferElementTypeLoweringOptions
+{
+    bool lowerBufferPointer = false;
+
+    // For WGSL, we can only create arrays that has a stride of 16 bytes for constant buffers.
+    bool use16ByteArrayElementForConstantBuffer = false;
```
kaizhangNV, line 16: Should we rename this to useAlignment16ForConstantBuffer?

source/slang/slang-ir-lower-buffer-element-type.cpp
```diff
@@ -319,6 +381,15 @@ struct LoweredElementTypeContext
         }
     }
 
+    IRIntegerValue get16ByteAlignedVectorSize(IRType* elementType, IRIntegerValue minCount)
```
kaizhangNV, line 384: Does this function actually return the number of elements after alignment instead of the size?

kaizhangNV (dismissed):

jkwak-work (dismissed):
Looks good to me.

I think it is strange to treat all scalar array elements as 16 byte, but I guess that is what WGSL does.

source/slang/slang-ir-lower-buffer-element-type.cpp
csyonghe: yeah, by `VectorSize`, I really just mean the number elements. Will rename it to VectorElementCount.

csyonghe:
The reason for this is that WGSL requires arrays in a uniform buffer to have 16-byte stride.
However, there isn't a way in WGSL to specify array stride explicitly, and they have to be derived from the element type.
And float always have 4 byte alignment, and float2 always have 8 byte alignment.
These two rules combined means that it is not possible to form array of `float2`s in wgsl and their official advise is to just use float4 in arrays. We obviously don't want to pass that restriction to slang users, so we need to run this legalization to do the conversion work.

PR #5499, opened 2024-11-05
[WGSL] [WASM] Add reflection endpoints + Fix bit manipulation operations

saipraveenb25:
- Add necessary layout reflection endpoints to Slang API's WASM bindings.
- Emit parenthesis around bit-shift, bit-xor and bit-or operations to avoid WGSL compiler complaining about precendence issues.


source/slang-wasm/slang-wasm.cpp
```diff
@@ -386,6 +386,47 @@ HashedString* ComponentType::loadStrings()
     return hashedStrings;
 }
 
+ProgramLayout* ComponentType::getLayout(unsigned int targetIndex)
```
csyonghe, line 389: This is going to making really hard in javascript to ensure all these reflection objects are explicitly `delete()`'d when they are no longer in use. 

We should instead just return a raw pointer to the layout without allocating a fresh object on the heap, and use `allow_raw_pointers` in emscripten bind, to free the javascript side from the responsibility of manually calling delete().

csyonghe (dismissed):

source/slang-wasm/slang-wasm.cpp
saipraveenb25: Okay, let me see if I can change up the wasm bindings

csyonghe:
@saipraveenb25 this PR is still not ready to merge, can you fix the formatting?

csyonghe:
@saipraveenb25 do you have plans to change the pointer ownership model in the wasm binding here?

csyonghe (dismissed):
need another formatting pass. You can setup vscode to perform a format-on-save, so you don't need to worry about formatting anymore.

PR #5500, opened 2024-11-05
Fix WGSL parameter block binding.

### LINKED ISSUE #5222: slang-RHI/WebGPU: Binding missing from bind group layout
This issue was split from https://github.com/shader-slang/slang/issues/5172.

Affected tests under `tests/compute` (currently disabled):
- Category 1: `Binding doesn't exist in...`
   - column-major
   - constant-buffer-memory-packing
   - func-cbuffer-param
   - parameter-block
   - non-square-column-major
- Category 2: `Number of entries X did not match the expected number of entries Y...`
   - kernel-context-threading
   - matrix-layout (.hlsl)
   - non-square-row-major
   - row-major

Tasks:
- Investigate and fix the issue(s)
- Try to enable the tests listed above

csyonghe:
Closes #5222.


source/slang/slang-emit-wgsl.cpp
```diff
@@ -87,9 +87,23 @@ void WGSLSourceEmitter::emitParameterGroupImpl(
     auto varLayout = getVarLayout(varDecl);
     SLANG_RELEASE_ASSERT(varLayout);
 
-    for (auto attr : varLayout->getOffsetAttrs())
+    EmitVarChain blockChain(varLayout);
```
expipiplus1, line 90: we're finally web3

expipiplus1 (dismissed):

source/slang/slang-emit-wgsl.cpp
aleino-nv: > we're finally web3

Speaking of block chains...

csyonghe: I am not accountable for this. The code and naming is copied from the glsl implementation. I am curious who did this.

PR #5507, opened 2024-11-06
Reflection compiler option

### LINKED ISSUE #5492: Option to generate binding and layout information for slangc
As far as I understand the documentation, layout and  binding information (unless manually specified) is chosen by the language, and the only way know what was chosen is to use the reflection API. With something like a separate json output file that includes the descriptor set and binding for globals and the type layout, slangc could be used standalone for compilation.

tareksander:
Fixes #5492

Some questions:

- Is `OSFilesystem` the correct API to write a file?
- Does `saveFile` expect a NULL-terminated string? I kind of just expected it to, since there is no length argument.

All the reflection tests still seemed to pass.


source/core/slang-pretty-writer.h
```diff
@@ -0,0 +1,122 @@
+#ifndef SLANG_CORE_PRETTY_WRITER_H
```
csyonghe, line 1: Can you put this file under `compiler-core` instead of `core`?

csyonghe (commented):
Thank you for putting these together! It all looks good to me. Only minor comments.

Would also be good if we can also get this added to our COM-based compilation API, in the form of `IComponentType::getReflectionJson(IBlob** outBlob)`.

csyonghe:
For formatting, you can post a message in this thread with the content "/format" and you will get a pull request to your branch for the formatting changes.

source/slang/slang-reflection-json.cpp
```diff
@@ -0,0 +1,1195 @@
...
+        writer << "]";
+    }
+
+    // If code generation has been performed, print out the parameter usage by this entry point.
+    if ((request->getCompileFlags() & SLANG_COMPILE_FLAG_NO_CODEGEN) == 0)
+    {
+        writer << ",\n\"bindings\": [\n";
+        writer.indent();
+
+        auto parameterCount = programReflection->getParameterCount();
```
csyonghe, line 1079: This line has a warning that breaks the build on CI:
```
warning C4456: declaration of 'parameterCount' hides previous local declaration
```

expipiplus1:
It's not necessary to do, but it would be nice if this handled passing `-` to output the json to stdout.

csyonghe:
The code still needs to be formated.

csyonghe:
There are linking errors.

expipiplus1:
/format

(@tareksander, I just fixed a bug where non-write users couldn't trigger the formatting bot, I've done it now but it should work for you in the future)

source/slang/slang.cpp
```diff
@@ -6861,10 +6861,13 @@ SlangResult EndToEndCompileRequest::compile()
     auto reflectionPath = getOptionSet().getStringOption(CompilerOptionName::EmitReflectionJSON);
     if (reflectionPath.getLength() != 0)
     {
-        reflectionPath.appendChar('\0');
         auto bufferWriter = PrettyWriter();
         emitReflectionJSON(this, this->getReflection(), bufferWriter);
-        if (SLANG_FAILED(OSFileSystem::getMutableSingleton()->saveFile(reflectionPath.getBuffer(), bufferWriter.getBuilder().getBuffer(), bufferWriter.getBuilder().getLength())))
+        if (reflectionPath == "-")
```
expipiplus1, line 6866: thank you!

tareksander:
> Would also be good if we can also get this added to our COM-based compilation API, in the form of `IComponentType::getReflectionJson(IBlob** outBlob)`.

I don't have enough knowledge of that API to do that.

> There are linking errors.

Weird. `slang-reflection-test` links with the `slang` library, but the linker can't find `Slang::emitReflectionJSON`.

csyonghe:
This is because `slang` is compiled as an DLL, and functions in a dll is generally not visible to other projects, unless you explicitly `export` it.  You need to expose it in slang-deprecated.h similar to other reflection functions.

```
    SLANG_API SlangRESULT spReflection_ToJson(SlangReflection* reflection, IBlob** outBlob);
```

Then you will be able to call this from reflection-test.

tareksander:
> This is because `slang` is compiled as an DLL, and functions in a dll is generally not visible to other projects, unless you explicitly `export` it. You need to expose it in slang-deprecated.h similar to other reflection functions.

Weird, since I'm on Linux, and symbols should be visible by default. That fixed it anyways, thanks.

Now I get a segfault in `spReflection_GetTypeParameterCount` during the test, no idea why.

include/slang.h
```diff
@@ -884,6 +884,7 @@ typedef uint32_t SlangSizeT;
         DumpWarningDiagnostics,
         InputFilesRemain,
         EmitIr,                        // bool
+        EmitReflectionJSON,            // bool
```
csyonghe, line 887: This will break binary compatibility. New enum values should be added at the end.

csyonghe:
The failures are due to C++ object lifetime issues.

You need to do:
```
*outObject = StringBlob::moveCreate(...).detach();
```
Otherwise the object will be deleted upon return.

Also changed the callsite to use
```
ComPtr<ISlangBlob> b;
spReflection_ToJson(..., b.writeRef());
```
instead of using raw `ISlangBlob*` pointer for `b` to ensure consistent life-time management.

csyonghe:
Everything looks good after fixing the memory management issue.
Thank you very much for your contribution!

Issue #5481, opened 2024-11-02
slang not building with cmake fetch content


sivansh11:
slang is not building with cmake fetch content

on adding this to my root cmake
```
FetchContent_Declare(
  slang
  GIT_REPOSITORY https://github.com/shader-slang/slang.git
)
FetchContent_MakeAvailable(slang)
```

getting this error while compiling
```
path_to_horizon/build/_deps/slang-src/external/slang-rhi/include/slang-rhi.h:7:10: fatal error: slang-com-ptr.h: No such file or directory
    7 | #include <slang-com-ptr.h>
```

cmake version 3.31.20241102-g08878dc
I dont think the commit hash is relevent since I dont specify it in fetchContent_Declare, non the less I can try doing it if you need me to.

bmillsNV:
@expipiplus1 can you help on this one?

expipiplus1:
Building this CMake projects works now

```cmake
cmake_minimum_required(VERSION 3.22)
project(foo)
include(FetchContent)
FetchContent_Declare(
  slang
  GIT_REPOSITORY https://github.com/shader-slang/slang
  GIT_TAG dc331db11c837bc0c4b27dff1811f4c624da33b6
)
FetchContent_MakeAvailable(slang)
```

sivansh11:
I dont think this problem is fully fixed, I still cant include slang.h 

```
path_to_project/horizon/examples/example_0/main.cpp:3:10: fatal error: slang.h: No such file or directory
    3 | #include <slang.h>
      |          ^~~~~~~~~
```
I made sure to add
target_link_library(my_project
  PUBLIC slang
)
and
target_include_directories(my_project
  PUBLIC slang
)

expipiplus1:
yeah, this is a different issue, caused when we moved things into the `include` dir instead of root (https://github.com/shader-slang/slang/pull/4636)

I've fixed this in https://github.com/shader-slang/slang/pull/5539

PR #5534, opened 2024-11-11
Use correct slang source dir for slang-rhi

expipiplus1:
Fixes https://github.com/shader-slang/slang/issues/5481

We should consider changing the CMAKE_BINARY_DIR usages also in our build descs


csyonghe (approved):

csyonghe:
Why are tests passing with the wrong dir?

csyonghe (approved):

expipiplus1:
Because it's only incorrect when slang isn't the top level cmake project, could test better regarding this https://github.com/shader-slang/slang/issues/5537

PR #5539, opened 2024-11-12
Correct include dir for libslang

expipiplus1:
technically a breaking change, but anyone using the old behaviour ought to change includes from `<include/slang.h>` to just `<slang.h>`


jkwak-work (dismissed):
Looks good to me

csyonghe:
Can you provide more context on why this PR is needed?

expipiplus1:
ah sure, that was in the commit message but didn't make it into the PR desc:

This stops adding the repo root to the include path for anything linking
with slang. This enabled a bunch of convenient includes, but might lead
to confusing behavior for anyone including slang. Not to mention
differences including it from an install vs source. 

Somebody already hit this in fact: https://github.com/shader-slang/slang/issues/5481#issuecomment-2469762252

expipiplus1:
(ci failure is simple, I can get to that and it reflects something wrong in our build system not falcor's usage of it)

csyonghe (approved):

PR #5544, opened 2024-11-12
Push buffer load to end of access chain.

csyonghe:
Closes #5538.

PR #5552, opened 2024-11-13
Simplify defer buffer load.

csyonghe:
This PR simplifies #5544 by not doing adhoc deduplication and instead rely on the existing redundancy removal pass to get rid of redundant FieldExtract/GetElement insts before deferring buffer loads.


csyonghe:
I am going to merge this to see if it addresses a regression.

PR #5556, opened 2024-11-13
Remove 'raw' default construct insts for SPIRV emit

saipraveenb25:
Our auto-diff step currently emits a few 'raw' `OpDefaultConstruct` insts for intermediate data types, mostly to silence HLSL errors about uninitialized variables. This is a redundant inst, because our auto-diff pass will always make sure all generated fields are individually set.
 
Usually, these intermediate types are made up of zero-initializable types so they get lowered to `OpMakeStruct`, so targets like SPIRV where `OpDefaultConstruct` does not exist don't run into issues.

For some special types, it is not easy (or safe) to emit such zero initializers, so we will instead run a pass to remove `OpDefaultConstruct` for backends where it is not available (thankfully, these backends do not complain about uninitialized variables, so the code correctness is unaffected)

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -89,6 +89,7 @@
 #include "slang-ir-ssa.h"
 #include "slang-ir-string-hash.h"
 #include "slang-ir-strip-cached-dict.h"
+#include "slang-ir-strip-default-construct.h"
 #include "slang-ir-strip-witness-tables.h"
 #include "slang-ir-strip.h"
 #include "slang-ir-synthesize-active-mask.h"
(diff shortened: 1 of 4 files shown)


source/slang/slang-emit.cpp
```diff
@@ -1419,6 +1420,19 @@ Result linkAndOptimizeIR(
     // we will need to disable this pass.
     stripWitnessTables(irModule);
 
+    switch (target)
+    {
+    // On targets that don't support default initialization, remove 'raw' default construct
+    // insts because our code-gen will not have any way to emit them.
+    //
+    case CodeGenTarget::SPIRV:
```
csyonghe, line 1428: What if the raw default constructor has uses? If we just remove them the rest of the IR will be ill-formed.

csyonghe: Oh, I see the code is detecting `store(var, defaultConstruct())` and remove the `store` along with the defaultConstruct.

PR #5563, opened 2024-11-14
Add proposal for #feature directive

tangent-vector:
This proposal is a companion to SP#012.

### DIFF
--- a/docs/proposals/013-experimental-feature-directive.md
+++ b/docs/proposals/013-experimental-feature-directive.md
@@ -1,0 +1,195 @@
+SP #013: Introduce a `#feature` Directive
+=========================================
+
+Status: Design Review
+
+Implementation: -
+
+Author: Theresa Foley <tfoley@nvidia.com>
+
+Reviewer: -
+
+Introduction
+------------
+
+We propose to introduce a `#feature` directive, which can be used to explicitly opt in to not-yet-stabilized features:
+
+    #language slang 2025
+    #feature quantumTextures
+
+In the above hypothetical example, a developer has chosen to code against the Slang 2025 language version (as introduced in SP#012), which does not support a hypothetical feature `quantumTextures`, because it is not yet stabilized.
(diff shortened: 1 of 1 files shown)


csyonghe:
I am worried that supporting individual feature flags can lead to fragmentation of possible compiler modes that we need to test and maintain support for.

Can we simply have something simpler with versions, by having `#language slang 2025.1 stable` and `#language slang 2025.1 experimental`. We are already suffering from the divergent behavior for some existing compiler flags such as -allow-glsl, and I want to be cautious to not make this situation worse.

tangent-vector:
I agree that the fragmentation could be a problem, and that it makes sense to consider a simpler alternative like an `experimental` language version/mode.

The main argument *against* making it an all-or-nothing binary switch is something like:

* There is a spectrum of just how unstable unstable features are. Some might be close to stabilization, while others might be completely new ideas that are only half-implemented.

* We expect many/most Slang users (including those doing production work) to write code that uses unstable features - especially those that are close to stabilization. Almost nobody doing serious work on top of Slang would be using the non-`experimental` mode.

* If `experimental` mode is all-or-nothing, these developers will not be able to easily get confidence that their code isn't unintentionally making use of other unstable features beyond what they intended to use. It is hard for them to determine how much churn could be inflicted on them by changes across Slang compiler releases.

I suspect that on the technical side it will not be much harder for us to *implement* fine-grained enabling/disabling of features as compared to the all-or-nothing case, so ideally we could defer the final decision on policy until after we have more of the mechanism in place.

csyonghe:
I really don't want to see every piece of slang code comes with a huge list of boilerplate `#feature` nonsense listing every Slang feature added in the past two years. I don't want to encourage people writing that kind of code or even make it possible to do so. I think the GLSL `#extension require` is particularly bad, can we please not do that.  

I think your concern can be addressed with probably one more feature level, like `beta`, `experimental` etc.
I would rather us not overdo fragmentation in the beginning unless it is causing a real problem.

csyonghe:
Closing since this is done in  the spec repo.

PR #5569, opened 2024-11-15
Embed core module in wasm build.

csyonghe:
This change allows us to embed prebuilt core module in WebAssembly.

The main issue being fixed here is that some of our AST nodes contain fields of type `Index` or `UInt`, whose sizes are architecture dependent. This means that the serialized module produced from x64 build system won't correct load by the web assembly module, which is 32-bit. The fix is to get rid of these arch-dependent int types in AST nodes.

To prevent future regression, the slang-cpp-extract application is updated to report an error if it sees an AST node field is of such types.

### DIFF
--- a/tools/slang-cpp-extractor/parser.cpp
+++ b/tools/slang-cpp-extractor/parser.cpp
@@ -1849,7 +1849,8 @@ SlangResult Parser::_maybeParseContained(Node** outNode)
                     "UIndex",
                     "UCount",
                     "PtrInt",
-                    "intptr_t"};
+                    "intptr_t",
+                    "uintptr_t"};
                 for (const auto& illegalType : illegalTypes)
                 {
                     int index = typeName.indexOf(UnownedStringSlice(illegalType));
--- a/tools/slang-cpp-extractor/parser.cpp
+++ b/tools/slang-cpp-extractor/parser.cpp
@@ -1524,6 +1524,11 @@ bool Parser::_isCtor()
     return isCtor;
 }
 
+bool isAlphaNumeric(char c)
+{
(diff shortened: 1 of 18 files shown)


tools/slang-cpp-extractor/parser.cpp
```diff
@@ -1833,7 +1838,36 @@ SlangResult Parser::_maybeParseContained(Node** outNode)
...
-
+            if (fieldNode->m_reflectionType == ReflectionType::Reflected)
+            {
+                static const char* illegalTypes[] = {
+                    "size_t",
+                    "Int",
+                    "UInt",
+                    "Index",
+                    "Count",
+                    "UIndex",
```
aleino-nv, line 1849: How will this code get updated if the names these types change?

tools/slang-cpp-extractor/parser.cpp
```diff
@@ -1833,7 +1838,36 @@ SlangResult Parser::_maybeParseContained(Node** outNode)
...
+                    "UCount",
+                    "PtrInt",
+                    "intptr_t"};
```
aleino-nv, line 1852: There is also `uintptr_t`

aleino-nv (approved):
Looks good to me

tools/slang-cpp-extractor/parser.cpp
csyonghe: There is not a good way to guard against that unfortunately. But we shouldn't be changing the names of these types.

tools/slang-cpp-extractor/parser.cpp
csyonghe: Done.

Issue #4943, opened 2024-08-28
WGSL: Aim for getting all GFX tests to pass


aleino-nv:

jkwak-work:
The plan is to try to get `slang\tests\render\cross-compile-entry-point.slang` working for WGSL and see what comes out next.

jkwak-work:
After spending some time understanding how `tests/render/cross-compile-entry-point.slang` works, here is how I understood this issue.

There are four slang-test "test-types" related to "rendering"; the full list of "test-types" can be found [here](https://github.com/shader-slang/slang/blob/master/tools/slang-test/README.md#test-types).:
 - COMPARE_HLSL_RENDER
 - COMPARE_HLSL_CROSS_COMPILE_RENDER
 - COMPARE_HLSL_GLSL_RENDER
 - COMPARE_RENDER_COMPUTE

Because this task is limited to "all GFX tests", "COMPARE_RENDER_COMPUTE" isn't relevant.

There are only 7 tests that use one of the three test-types:
```
tests/bugs/texture2d-gather.hlsl
tests/render/cross-compile-entry-point.slang // currently disabled
tests/render/nointerpolation.hlsl // currently disabled
tests/render/render0.hlsl // currently disabled
tests/render/cross-compile0.hlsl // currently disabled
tests/render/imported-parameters.hlsl // currently disabled
tests/render/unused-discard.hlsl // currently disabled
```
The goal of this task will be to make those tests working for WGSL.

For the disabled tests, we will need to re-evaluate if they should be re-enabled.
If enabled, they will also need to work for WGSL.
As an example, "tests/render/cross-compile-entry-point.slang" can be simply re-enabled by changing the test-type from `COMPARE_HLSL_CROSS_COMPILE_RENDER` to `COMPARE_HLSL_RENDER`.

When the task is completed, we should have a following line on those rendering tests,
```
//TEST:COMPARE_HLSL_RENDER:-wgsl
```
It will generate two PNG image files: one with HLSL/DXC and another with WGSL/DAWN.
slang-test will compare the images and check if they are identical.

csyonghe:
`COMPARE_RENDER_COMPUTE` is actually relevant, it is a vertex-fragment pipeline that writes results into a buffer instead of a framebuffer, and reads back the buffer values for comparison, so that test is a vertex-fragment shader test.

jkwak-work:
I see.
Thanks for the information.

[...]

PR #5572, opened 2024-11-15
render-test: Limit 1d textures to 1 MIP level

aleino-nv:
This is due to a limitation in WGSL.
See https://www.w3.org/TR/webgpu/#abstract-opdef-maximum-miplevel-count.

This helps to address issue #4943.


aleino-nv:
This is not the right fix. I need to deal with this WebGPU limitation in some other way...

PR #5589, opened 2024-11-19
Implement semantics for WGSL

aleino-nv:
This helps to address issue #4943.


jkwak-work:
Shouldn't we add some test cases or enable existing tests for this change?

csyonghe:
Yes, we should. I don't consider this PR to close any specific issue due to lack of tests, but I would rather have this merged in instead of having no support at all.

aleino-nv:
This change itself doesn't close the issue as mentioned in the change description.
As mentioned in the linked issue title, the issue is indeed about enabling GFX tests.
Before closing the issue, I will enable all tests that should be passing.

Issue #5591, opened 2024-11-19
[Reflection API] `getFullName()` does not return a fully-qualified name in many cases.


saipraveenb25:
`getFullName()` uses the parent information encoded in the `DeclRef`, but our current system only stores parent info if there's relevant specialization info.
e.g. 
```csharp
struct A<T>
{
    struct B
    { /* */ };
};

// A needs specialization info
getFullName(A<float>.B) -> "A<float>.B"

struct A
{
    struct B
    {  /* ... */  };
};

// A needs no specialization info
getFullName(A.B) -> "B" // 'A' is left out
```

We need to change up how `DeclRef::toText()` and `DeclRef::getParentDeclRef()` work, so that we return a properly usable name.

Issue raised on behalf of @ccummingsNV

PR #5592, opened 2024-11-19
Change how DeclRef::toText works

saipraveenb25:
We now ignore the decl-ref heirarchy since that only includes nodes with specialization info & simply walk up the tree of decls, while emitting any specializations present in the decl-ref.

Fixes: https://github.com/shader-slang/slang/issues/5591


csyonghe (commented):
Instead of changing DeclRef::toText, which mess up all the tests, you can use `ASTPrinter` to get the full qualified name of a declref.

Call `ASTPrinter::addOverridableDeclPath` and then grab the content of the astprinter's string builder.

saipraveenb25:
@csyonghe I fixed up the failing tests. (Only a couple of them were failing)
Further, one of those tests required this fix: `tests/diagnostics/mismatching-types.slang` was expecting full names for the nested types, but had the same problem that only a subset of the nodes were printed out.

csyonghe (approved):

PR #5609, opened 2024-11-20
WGSL: Support isnan, isinf, etc..

aleino-nv:


csyonghe (approved):

source/slang/hlsl.meta.slang
```diff
@@ -9566,12 +9568,14 @@ bool isnan(T x)
         __intrinsic_asm "$P_isnan($0)";
     case spirv:
         return spirv_asm { result:$$bool = OpIsNan $x};
+    case wgsl:
+         __intrinsic_asm "$0 != $0";
```
jkwak-work, line 9572: I think this can be `default`

aleino-nv: Right, yeah I was considering that... I opted to be conservative since shader languages sometimes have unexpected guarantees about floating point.

aleino-nv: Hopefully we don't have a ton of backends to add in the future! :)

PR #5617, opened 2024-11-20
Enable WGPU texture sampling test

aleino-nv:
- Update Slang-RHI to get WGPU fixes
- Fix and enable WGPU texture sampling test

https://github.com/shader-slang/slang/issues/4943

### DIFF
--- a/tests/compute/texture-sampling-no-1d-arrays.slang
+++ b/tests/compute/texture-sampling-no-1d-arrays.slang
@@ -3,7 +3,9 @@
 //TEST(compute):COMPARE_RENDER_COMPUTE: -shaderobj -output-using-type -mtl
 
 
-//TEST_INPUT: Texture1D(size=4, content = one):name=t1D
+// WebGPU only supports 1 MIP level for 1d textures
+// https://www.w3.org/TR/webgpu/#abstract-opdef-maximum-miplevel-count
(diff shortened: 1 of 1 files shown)


csyonghe (approved):

tests/compute/texture-sampling-no-1d-arrays.slang
```diff
@@ -3,7 +3,9 @@
 //TEST(compute):COMPARE_RENDER_COMPUTE: -shaderobj -output-using-type -mtl
 
 
-//TEST_INPUT: Texture1D(size=4, content = one):name=t1D
+// WebGPU only supports 1 MIP level for 1d textures
+// https://www.w3.org/TR/webgpu/#abstract-opdef-maximum-miplevel-count
```
jkwak-work, line 7: I think this limitation should be mentioned on our document although it is a limitation on WGSL itself not Slang.
https://github.com/shader-slang/slang/blob/master/docs/user-guide/a2-03-wgsl-target-specific.md

aleino-nv: So far the things that have gone into that document have been strictly about the generated WGSL.
This would be a limitation you can only detect on the WebGPU API side, so I'm not sure it fits well with the other information.

PR #5621, opened 2024-11-21
[Docs] Minor fixes to auto-diff documentation

saipraveenb25:
Resolves: https://github.com/shader-slang/slang/issues/5551
Resolves: https://github.com/shader-slang/slang/issues/5311


csyonghe (approved):

docs/user-guide/07-autodiff.md
```diff
@@ -167,6 +167,9 @@ interface IDifferentiablePtrType
 }
 ```
 
+> #### Note ####
+> Support for `IDifferentiablePtrType` is still experimental. 
+
 Types should not conform to both `IDifferentiablePtrType` and `IDifferentiable`. Such cases will result in a compiler error.
```
csyonghe, line 173: I wonder if IDifferentiablePtrType should be the base type of IDifferentiable? This way we won’t have to define what it means if a type is both IDifferentiable and IDifferentiablePtr.

PR #5634, opened 2024-11-22
ci: Update `actions/checkout` to always use v4

waywardmonkeys:
CI was using a mix of version 3 and 4. Version 3 results in a warning within the GitHub Actions UI as it was using a version of NodeJS that is deprecated within the GitHub Actions infrastructure.


csyonghe (approved):
Looks good to me. Thank you for your contribution!

PR #5636, opened 2024-11-22
[Docs] Extensions -  better wording for c# equivalent

burak-efe:
slang extensions looks same as c# extension methods, not partial classes. Partial classes allow extending types to introduce more variables unlike slang extensions. And you cannot extend that types not partial to begin with. Where extension methods allows that. Also c# users that use extension methods much more  than partial types, probably.


csyonghe:
Agreed. Thanks for the correction!

csyonghe (approved):

PR #5638, opened 2024-11-22
[Docs] fix spirv dead links on toc.html

burak-efe:
I believe this will fix the dead links that occurred because of sub section renames.


csyonghe:
Thank you for spotting and fixing this issue!

the toc.html is generated by running build_toc.ps1 in docs/. Is this produced by the script?

burak-efe:
> Thank you for spotting and fixing this issue!
> 
> the toc.html is generated by running build_toc.ps1 in docs/. Is this produced by the script?

oh, no. I just copy fasted section names. I guess you can consider this just as a issue rather than pr, if its not gonna solve the issue

csyonghe (approved):
I am going to merge this in assuming the script generated toc.html will be identical. If this breaks again, we should consider it an issue.

Issue #5625, opened 2024-11-21
WGSL: nointerpolate doesn't work


aleino-nv:
Failing tests:
- `tests/render/nointerpolation.hlsl (wgpu)` is failing to generate the expected output (solid red triangle)

aleino-nv:
Filing separate issue for tracking, but this will be done as part of https://github.com/shader-slang/slang/issues/4943

aleino-nv:
It should probably map to ['interpolate' attribute](https://www.w3.org/TR/WGSL/#interpolate-attr) with https://www.w3.org/TR/WGSL/#interpolation-type-flat

aleino-nv:
I'm working on a fix in the branch [aleino/nointerpolate](https://github.com/aleino-nv/slang/tree/aleino/nointerpolate), but the resulting `@interpolate` attribute gets added to the CoarseVertex struct, whereas the actual output struct for the vertex shader (which also gets a `@location` attribute, as is required when using `@interpolate`) is the result of struct flattening.

I noticed that [the struct flattening function](https://github.com/shader-slang/slang/blob/master/source/slang/slang-ir-wgsl-legalize.cpp#L712) does not move the interpolation mode decoration from the 'color' field over to the new flattened struct.


    ...
    
    struct VertexStageOutput_0
    {
        @location(0) color_0 : vec3<f32>,            // @interpolate *should* be here
        @builtin(position) sv_position_0 : vec4<f32>,
    };
    
    struct VertexStageInput_0
    {
        @location(0) position_0 : vec3<f32>,
        @location(1) color_1 : vec3<f32>,
    };
    
    struct AssembledVertex_0
    {
         position_1 : vec3<f32>,
         color_2 : vec3<f32>,
    };
    
    struct VertexStageInput_1
    {
         assembledVertex_0 : AssembledVertex_0,
    };
    
    struct CoarseVertex_0
    {
        @interpolate(flat)  color_3 : vec3<f32>,    // @interpolate *should not* be here
    };
    
    struct VertexStageOutput_1
    {
         coarseVertex_0 : CoarseVertex_0,
         sv_position_1 : vec4<f32>,
    };
    
    @vertex
    fn vertexMain( input_0 : VertexStageInput_0) -> VertexStageOutput_0
    {
        ...
        var _S3 : VertexStageOutput_0;
        ...
        return _S3;
    }

PR #5641, opened 2024-11-22
Support interpolation modifiers for WGSL

aleino-nv:
- Move interpolation decorations to flattened structs during legalization
- Add some code to emit interpolate attributes
- Enable 'nointerpolate' test

https://github.com/shader-slang/slang/issues/5625


source/slang/slang-ir-wgsl-legalize.cpp
```diff
@@ -674,6 +675,13 @@ struct LegalizeWGSLEntryPointContext
             auto newKey = builder.createStructKey();
             copyNameHintAndDebugDecorations(newKey, oldKey);
 
+            if (auto interpolationModeDecoration =
```
csyonghe, line 678: should we transfer all decorations from oldKey to newKey?

csyonghe (commented):
Looks good but I think it is more robust and correct to just transfer all decorations over.

csyonghe (approved):

source/slang/slang-emit-wgsl.cpp
```diff
@@ -1506,4 +1506,59 @@ void WGSLSourceEmitter::emitIntrinsicCallExprImpl(
...
+    {
+        switch (mode)
+        {
+        case IRInterpolationMode::Sample:
+            m_writer->emit(", sample");
+            break;
+        case IRInterpolationMode::Centroid:
+            m_writer->emit(", centroid");
+            break;
+        default:
```
jkwak-work, line 1536: I think this `switch` will always hit the `default` case here, because in order for `interpolationTypeEmitted` to be true, `mode` has to be be one of three cases that is non of the cases listed in this `switch`.
That seems like a coding mistake.

source/slang/slang-ir-wgsl-legalize.cpp
aleino-nv: Seems to work for the `nointerpolation` test, at least.
Should we do the same thing for the field, and not just the key?

expipiplus1 (approved):

Issue #5633, opened 2024-11-22
WGSL target doesn't increase location id for attributes


lmariscal:
Simple rasterization pipeline shader doesn't increase the location id for the different inputs.

```slang
struct VertexOutput {
    float4 position : SV_Position;
    float4 color : F_COLOR;
}

struct VertexInput {
    float2 position : V_POSITION;
    float4 color : V_COLOR;
}

[shader("vertex")]
VertexOutput vert(VertexInput input) {
    var output: VertexOutput;
    output.position = float4(input.position, 0.0, 1.0);
    output.color = input.color;
    return output;
}

struct FragmentInput {
    float4 color : F_COLOR;
}

[shader("fragment")]
float4 frag(FragmentInput input) : SV_Target
{
    return input.color;
}
```

Output:
```wgsl
struct VertexOutput_0
{
    @builtin(position) position_0 : vec4<f32>,
    @location(0) color_0 : vec4<f32>,
};

struct VertexInput_0
{
    @location(0) position_1 : vec2<f32>,
    @location(0) color_1 : vec4<f32>,
};

@vertex
fn vert( input_0 : VertexInput_0) -> VertexOutput_0
{
    var output_0 : VertexOutput_0;
    output_0.position_0 = vec4<f32>(input_0.position_1, 0.0f, 1.0f);
    output_0.color_0 = input_0.color_1;
    return output_0;
}

struct pixelOutput_0
{
    @location(0) output_0 : vec4<f32>,
};

struct FragmentInput_0
{
    @location(0) color_0 : vec4<f32>,
};

@fragment
fn frag( input_0 : FragmentInput_0) -> pixelOutput_0
{
    var _S1 : pixelOutput_0 = pixelOutput_0( input_0.color_0 );
    return _S1;
}
```

I would expect the location ids inside `VertexInput_0` to increase, so `position_1` has `@location(0)` while `color_1` has `@location(1)`.

Have already tried playing with the semantics defined, renaming them, but the output remains the same.

As per the documentation inside ["WGSL specific functionalities"](https://shader-slang.com/slang/user-guide/wgsl-target-specific.html):
> Parameters without semantics are given automatic location indices. (See the @location attribute.)

I would believe they shouldn't even need a user defined semantics.  
But if I remove the defined semantics I get the following result:

```slang
struct VertexOutput {
    float4 position : SV_Position;
    float4 color : F_COLOR;
}

struct VertexInput {
    float2 position;
    float4 color;
}

[...]

PR #5642, opened 2024-11-22
wgsl: Fix field semantics even if input struct doesn't need flattening

aleino-nv:
This fixes an issue where location indices of stage inputs were repeated

https://github.com/shader-slang/slang/issues/5633


csyonghe (approved):

jkwak-work (changes requested):
I feel uncomfortable with this change.

The code is copied exactly from slang-ir-metal-legalize.cpp
And with this PR, WGPU side works but Metal side is failing.

I suggest to investigate why Metal is failing and fix it properly before merging this change.

csyonghe:
I feel like metal needs the same treatment.

source/slang/slang-ir-wgsl-legalize.cpp
```diff
@@ -145,11 +145,10 @@ struct LegalizeWGSLEntryPointContext
                     structType,
                     mapOldFieldToNewField,
                     semanticInfoToRemove);
+                // Validate/rearange all semantics which overlap in our flat struct.
+                fixFieldSemanticsOfFlatStruct(flattenedStruct);
```
jkwak-work, line 149: If this function is going to be called even when it is not flattened, the name of the function may need to change to match

jkwak-work:
> I feel like metal needs the same treatment.

Yes, that must be it.

jkwak-work (approved):

aleino-nv:
> > I feel like metal needs the same treatment.
> 
> Yes, that must be it.

aleino-nv:
> > I feel like metal needs the same treatment.
> 
> Here is a PR #5666

Ok the PR fails, I filed https://github.com/shader-slang/slang/issues/5667.
The render0 test is also broken for Metal, so it's probably best to fix that one first, and then enable this test.

Issue #5631, opened 2024-11-21
Speed up core library loading, bring compilation time of small files down to <10ms mark


kanashimia:
## Description:

Core module takes a long time to load, which in turn slows down all operations with CLI, improve that.
Compilation of `hello.slang` with `slangc` should be around 1ms mark.

This will require a number of changes in the codebase.

---

#### Old description:

```
> time slangc --help
slangc --help  1.72s user 0.11s system 98% cpu 1.855 total
```
Simple call takes almost two seconds! Even chromium starts faster on my machine.
For comparison glslang:
```
> time glslang --help
glslang --help  0.00s user 0.00s system 95% cpu 0.006 total
```

ltrace tells that it is stuck during `slang_createGlobalSession`

strace shows that it tries to newfstatat `libslang.so` then tries to openat `slang-core-module.bin` and then it is a long long chain of `brk` calls and it is stuck during brk calls it seems.

Looks like it tries to allocate a loooooot of memory during startup? using brk????
To open `libslang.so`??
Shouldn't happen.

slang rev: dbc28b4fe9b0a6e8215640c04a9f245c150150a8
os: NixOS 24.11
custom build of slang, if that is related.

bmillsNV:
Thanks for raising this issue. Is it possible to try the prebuilt releases [here](https://github.com/shader-slang/slang/releases/) to see if you can repro the issue? On my machine it takes .2 seconds for slangc --help. One difference with the prebuilt binaries is that the core module is embedded into the libslang.so itself so there wont be any filesystem calls.

dmitsuki:
I'm using the prebuilt binaries. For me slangc takes 1.33s, whereas glslang similarly takes 0.00s.

kanashimia:
> Thanks for raising this issue. Is it possible to try the prebuilt releases [here](https://github.com/shader-slang/slang/releases/) to see if you can repro the issue? On my machine it takes .2 seconds for slangc --help. One difference with the prebuilt binaries is that the core module is embedded into the libslang.so itself so there wont be any filesystem calls.

Prebuilt indeed faster. 
`./slangc --help  0.12s user 0.06s system 99% cpu 0.173 total`
But system calls more or less look the same.

So, I found the issue, it seems slangc tries to write `lib/slang-core-module.bin` during first launch, and then it takes longer, but on my system that is impossible because artefacts directory is read only, so it always takes long time.
Why is that the case? 

I could reproduce the issue like this:
```
› time ./bin/slangc
[...]

PR #5644, opened 2024-11-22
Enable core module embedding by default in CMake.

kanashimia:
It seems that release workflows were unexpectedly broken when STDLIB was renamed to CORE_MODULE.
Changed the option default because I think it just makes more sense. Removed explicit calls to enable it. Fixed docs.

Should `SLANG_EMBED_CORE_MODULE_SOURCE` be disabled too? Does it do anything useful?

As reported in https://github.com/shader-slang/slang/issues/5631

NOTE: I haven't really tested it at this point, but should work.

### DIFF
--- a/.github/workflows/release-linux-glibc-2-17.yml
+++ b/.github/workflows/release-linux-glibc-2-17.yml
@@ -26,7 +26,7 @@ jobs:
 
             cd /home/app
             git config --global --add safe.directory /home/app
-            cmake --preset default --fresh -DSLANG_SLANG_LLVM_FLAVOR=DISABLE -DSLANG_EMBED_STDLIB=1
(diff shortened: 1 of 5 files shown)


csyonghe (approved):
Looks good, thanks for your contribution!

csyonghe:
The only reason to enable `SLANG_EMBED_CORE_MODULE_SOURCE` is so that the language server can provide useful documentation in hovering hints. We could disable this for most builds, and enable it only when building for language server, but that likely needs to come as its own PR. We can change the language server code to fetch the core module source from a separately deployed file that ships with `slangd` in the vs extensions, and get rid of this embeding in slang.dll to reduce binary size.

jkwak-work:
I appreciate fixing the mistake I made.
I guess I didn't carefully check directories whose name starts with dot, "."

PR #5645, opened 2024-11-22
Don't treat StrcturedBuffer<IFoo> as a specializable param.

### LINKED ISSUE #5629: Always add vtable to struct?
Is there a way to always add a vtable to structs that implement an interface?

When such a type is directly used (without an interface or second type implementing this interface) in some shader this vtable space is often removed. This makes it hard to reuse buffers in cases with and without vtable. Is there an option to always add the vtable for interface implementations?

```c#
interface IShape {}

struct Sphere : IShape {
    float radius;
};
struct Cube : IShape {
    float size;
};
```

```c#
StructuredBuffer<IShape> shapes;
// if typeconformance includes Sphere and Cube -> radius/size starts at offset 16 bytes (vtable)
// if typeconformance includes just Sphere -> radius starts at offset 0 bytes (no vtable)

StructuredBuffer<Sphere> spheres;
// radius starts at offset 0 bytes (no vtable)
```
This makes it really hard to manage the backend buffer of shapes/spheres, would be nice if interface types would always include the vtable.

csyonghe:
Fixes #5629.

When we have a `StructuredBuffer<IFoo>` parameter, we shouldn't treat it as a statically specializable parameter because we never want to directly replace `IFoo` with any concrete type. Instead we want to always rely on the dynamic dispatch handling logic to insert proper header to maintain data layout consistency.

### DIFF
--- a/source/slang/slang-check-shader.cpp
+++ b/source/slang/slang-check-shader.cpp
@@ -79,15 +79,6 @@ static void _collectExistentialSpecializationParamsRec(
             loc);
         return;
     }
-    else if (auto structuredBufferType = as<HLSLStructuredBufferTypeBase>(type))
-    {
-        _collectExistentialSpecializationParamsRec(
-            astBuilder,
-            ioSpecializationParams,
-            structuredBufferType->getElementType(),
-            loc);
-        return;
-    }
 
     if (auto declRefType = as<DeclRefType>(type))
     {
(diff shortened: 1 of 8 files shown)


jkwak-work (approved):
Looks good to me.

I think I have seen it some place before but what does `_ST` stands for?

csyonghe:
"_S" is the start of any mangled name.
"T" means the thing being mangled is a type.

jkwak-work:
> "_S" is the start of any mangled name. "T" means the thing being mangled is a type.

Aha. Thanks for the explanation.

PR #5650, opened 2024-11-23
Fix unused struct in  sample shader

yknishidate:
Hi, I'm new here.

I was looking at the triangle sample and noticed that it declared a struct `Fragment` for the fragment shader output but didn't actually use it, so I decided to fix that.

Of course, I could have just deleted `Fragment`, but I chose to use it to demonstrate that we can use a custom struct for output.


cheneym2 (approved):
Looks good to me, thanks

PR #5666, opened 2024-11-25
metal: Fix field semantics even if input struct doesn't need flattening

aleino-nv:
As per the comment in https://github.com/shader-slang/slang/pull/5642, the Metal back-end may need this kind of fix.


jkwak-work (approved):
Looks good to me.

But it looks like Metal CI is failing.

aleino-nv:
> Looks good to me.
> 
> But it looks like Metal CI is failing.

Yeah, I filed an issue about it. https://github.com/shader-slang/slang/issues/5667

PR #5674, opened 2024-11-25
Add SlangConfig.cmake with slang build targets

### LINKED ISSUE #5649: Provide SlangConfig.cmake in releases
This is more of a nice-to-have than a must-have, but it would be really helpful that releases provide such a file (generated by CMake's install process) to ease the import of precompiled binaries in CMake setups!

ov-l:
Closes #5649


ov-l:
I need to reformat the CMakeLists file.

obhi-d:
I am not able to make the wasm build work, so probably I need to skip adding slang as target. I tried to add through this macro:
```cmake

macro(export_slang_targets)
  set(SlangExportTargetList)
  foreach(target IN ITEMS ${ARGN})
    message(STATUS "----------------> ADDING TARGET ${target}")
    if(TARGET ${target})
      list(APPEND SlangExportTargetList ${target})
    endif()
  endforeach()
  
  if (SlangExportTargetList)
  
    install(TARGETS ${SlangExportTargetList} EXPORT SlangExportTarget)
    
    install(
        EXPORT SlangExportTarget
        FILE ${PROJECT_NAME}Targets.cmake
        NAMESPACE ${PROJECT_NAME}::
        DESTINATION cmake
    )
    
    install(
        FILES
            "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
            "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
        DESTINATION cmake
    )
  
  endif()
endmacro()

# 
export_slang_targets(
  core
  prelude
  unordered_dense
  compiler-core
  slang-capability-defs
  slang-capability-lookup
  slang-reflect-headers
  slang-lookup-tables
  SPIRV-Headers
  slang-common-objects
  slang
)
```

But other issues crop up if I do so, notably : Target "slang-reflect-headers" INTERFACE_INCLUDE_DIRECTORIES property
contains path: <build-path>,  which is prefixed in the build directory.
  
Exporting all targets specified in SlangTargets is also not an option because it seems SlangTargets include every target declared within the library and they should not be linked to (for ex. executable targets).

For my use-case, only single `slang` target works for both Linux and Windows. Probably this needs to be handled correctly instead of how I am doing.

eliemichel:
It might be ok to just not try to find a prebuilt version of Slang when cross-compiling with emscripten (i.e., enclose the `find_package` within a `if (NOT EMSCRIPTEN)` and fallback to "manually" locating `slangc` executable in that case (because I suppose that's all what is needed when the target is WebAssembly -- unless one wants to link slang library into a WebAssembly module but there is no precompiled release for WebAssembly anyways, so one would build from scratch in that case).

obhi-d:
> It might be ok to just not try to find a prebuilt version of Slang when cross-compiling with emscripten (i.e., enclose the `find_package` within a `if (NOT EMSCRIPTEN)` and fallback to "manually" locating `slangc` executable in that case (because I suppose that's all what is needed when the target is WebAssembly -- unless one wants to link slang library into a WebAssembly module but there is no precompiled release for WebAssembly anyways, so one would build from scratch in that case).

Yes, I missed your point earlier. I will try that. Thanks for clarifying.

expipiplus1 (changes requested):
Would it be possible to add some documentation to the new code explaining what's going on please. And could you add some documentation to docs/building.md explaining how the config cmake files can be used from an installed slang.

Thanks also! this looks like a good addition!

obhi-d:
Yes, I will add some comments and add docs.

csyonghe (approved):

csyonghe:
For webassembly builds, we will produce the slang-wasm.wasm which is a library that contains the wasm binding to slangc library callable from JavaScript. The binding is defined in slang-wasm.cpp.

csyonghe:
And we should update the CI setup to include prebuilt binaries for web assembly in our releases.

csyonghe:
@expipiplus1 Is this change good to merge?

expipiplus1 (approved):
Thanks!

Issue #5565, opened 2024-11-14
Slang "C" header is not valid C99/C11/C19/C23.


moonheart08:
Self explanatory, slang does not provide any C compatible API at this time, making creation of bindings unnecessarily difficult.

# Reproduction
`gcc -std=c11 slang.h`

csyonghe:
Since most of our core APIs are defined as COM interfaces, we are no longer maintaining C compliance for the header right now.

To target C, we will need to parse the existing COM interfaces and code-gen the C bindings for them. We welcome contributions to make this possible.

laurooyen:
I’ll be tackling this. My work will be based on the `slang-cpp-extractor` which is a C++ parser used for generating the AST reflection. I already have a proof of concept working for my personal Rust bindings. I’ll ensure to include C as well. They both have unique challenges and solving those together should ensure extensibility for other languages.

laurooyen:
@csyonghe A lot of code in `slang-cpp-extractor` can be shared with the binding generator. I feel like the latter should be its own executable though so how should I go about extracting the cpp parsing code? Move it to a new static library under the tools folder maybe called `slang-cpp-parser`?  Does the binding generator itself belong under tools too? I’d like to make a PR to set this separation up before submitting the actual work on the binding generator itself.

csyonghe:
Yes, let’s move it to its own static library.

csyonghe:
The binding generator can be placed under the same directory as other cpp extractor.

csyonghe:
I'd say we just hard code the logic for now and perform the include only for the enum header. What  do you think?

bmillsNV:
We're unable to prioritize this right now, placing on backlog. We would greatly welcome contribution if someone wants to take this.

PR #5675, opened 2024-11-25
Move c++ parsing code from slang-cpp-extractor to static library

laurooyen:
As part of #5565, this is the initial work on extracting c++ parsing code from `slang-cpp-extractor` to a separate static library. This will allow me to reuse the parser for my work on the language bindings generator. The current `slang-cpp-extractor` and newly created `slang-cpp-parser` can still use some refactoring but I feel like that is better left for future PRs once I know how much additional code can be shared with the binding generator.

Please ensure especially my changes to the CMake file look good, I'm not too familiar with it.


csyonghe:
@expipiplus1 can you please take a look?

tools/slang-cpp-parser/unit-test.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
aleino-nv, line 1: Ideally this should have been created with `git mv`, so that history is retained.

tools/slang-cpp-extractor/unit-test.h
```diff
@@ -1,17 +0,0 @@
-#ifndef CPP_EXTRACT_UNIT_TEST_H
```
aleino-nv, line 1: Ideally this should have been moved with `git mv`, so that history is retained.

tools/slang-cpp-parser/unit-test.h
laurooyen: I used `git mv` for all files. This one is small and changed a lot so git can't track that. Performing multiple commits should solve it but would break again once this PR gets squash merged.

tools/slang-cpp-extractor/cpp-extractor-main.cpp
```diff
@@ -14,14 +14,14 @@
...
 #include "../../source/core/slang-string.h"
 #include "../../source/core/slang-writer.h"
-#include "diagnostics.h"
-#include "file-util.h"
+#include "../tools/slang-cpp-parser/diagnostics.h"
+#include "../tools/slang-cpp-parser/file-util.h"
+#include "../tools/slang-cpp-parser/node.h"
+#include "../tools/slang-cpp-parser/options.h"
+#include "../tools/slang-cpp-parser/parser.h"
+#include "../tools/slang-cpp-parser/unit-test.h"
```
jkwak-work, line 22: Can we avoid using "../" here?
We already have a few but we should try to remove them not adding them.

jkwak-work (commented):
Is there a related issue for this PR?
I am not sure what problem this PR is trying to solve.
I am also not sure if it is safe to rename from `CppExtract` to `CppParse` without causing troubles on the users who might be already using them.

laurooyen:
This was discussed in #5565. My bad for not including it, I updated the PR description. It allows us to reuse the C++ parser code for generating C (and other) language bindings by parsing `slang.h`. I'd like to work on that in future PRs.

I didn't see any harm in renaming parts of `CppExtract` to `CppParse` because it felt like an internal API but I could be wrong.

tools/slang-cpp-extractor/cpp-extractor-main.cpp
laurooyen: I think we can by passing `INCLUDE_DIRECTORIES_PUBLIC .` to the `slang-cpp-parser` target in the CMake file.
Should I remove the relative includes in all other modified files too? I just verified it's already possible to include directly from `core/`, `comiler-core/` and `slang/` in those. What annoys me with that though is the clang formatter intertwining includes from the target itself with includes from other targets by just sorting alphabetically.

tools/slang-cpp-parser/unit-test.h
expipiplus1: yeah, fwiw `git mv` doesn't actually do any special tracking, it's just short for normal `mv`, untrack old file, track new file. the detects file moves purely from text similarity.

tools/slang-cpp-extractor/cpp-extractor-main.cpp
expipiplus1: yeah, I think we ought to do exactly that and just `#include "slang-cpp-parser/node.h"` etc. If you want to change the other ones feel free.

I wouldn't worry about the include ordering, but if it really bothers you the I guess we could have a .clang-format file in this subdirectory which inherits from the root and specifies `IncludeCategories`?

expipiplus1 (dismissed):
looks great, thanks!

tools/slang-cpp-extractor/cpp-extractor-main.cpp
jkwak-work: I am not sure how it can be done in cmake syntax but it sounds fine to me.

I am strongly against using the "../" for the include path. I have seen it including unexpected files with same file name before.

laurooyen: I removed all relative includes.

csyonghe (approved):

PR #5678, opened 2024-11-25
Enable Link-Time-Code-Generation for Release builds

jkwak-work:
Closes https://github.com/shader-slang/slang/discussions/5632

This commit enables Link Time Code Generation, or LTCG for short, for the release builds.

Because "incremental" build is enabled, this ends up using "/LTCG:incremental" rather than the full "/LTCG". It is still expected to more performant than without any "/LTCG" at all.

In order to use the full LTCG, we will need to give up the incremental build option for the Release build. We will need to decide if we wants to do it later.

### DIFF
--- a/cmake/SlangTarget.cmake
+++ b/cmake/SlangTarget.cmake
@@ -156,6 +156,13 @@ function(slang_add_target dir type)
         return()
     endif()
 
+    # Enable link-time optimization for release builds
+    # See: https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html
+    set_target_properties(
+        ${target}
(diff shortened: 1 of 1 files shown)


zamazan4ik:
> Closes https://github.com/shader-slang/slang/discussions/5632

One note: this PR doesn't close the linked discussion completely - it just closes a part of it regarding enabling LTO. The question about applying PGO and PLO to Slang remains open.

csyonghe (approved):

jkwak-work:
> > Closes #5632
> 
> One note: this PR doesn't close the linked discussion completely - it just closes a part of it regarding enabling LTO. The question about applying PGO and PLO to Slang remains open.

I think that will be tricky to apply unless we can get the profile data from a lot of use cases; or find more representative use cases.

With Profile guided optimization, the performance can shift and bias to only a few use cases if not careful.

PR #5680, opened 2024-11-26
Fix issue with slang-embed & include ordering

saipraveenb25:
Related: https://github.com/shader-slang/slang-torch/issues/23

Our slang-torch repository keeps breaking due to relative paths being occasionally re-written (e.g. 
`#include "slang.h"` vs. `#include "../../include/slang.h"`), exacerbated by the fact that we don't have slang-torch on CI yet.

This PR adds an include directory to the slang-embed tool to avoid this issue in the future.

### DIFF
--- a/tools/slang-embed/slang-embed.cpp
+++ b/tools/slang-embed/slang-embed.cpp
@@ -63,6 +63,7 @@ struct App
 {
     char const* appName = "slang-embed";
     char const* inputPath = nullptr;
+    char const* includeDir = nullptr;
     char const* outputPath = nullptr;
     Slang::HashSet<Slang::String> includedFiles;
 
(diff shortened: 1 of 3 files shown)


tools/slang-embed/slang-embed.cpp
```diff
@@ -91,7 +98,7 @@ struct App
 
         if (!inputPath || (argc != 0))
         {
-            fprintf(stderr, "usage: %s inputPath [outputPath]\n", appName);
+            fprintf(stderr, "usage: %s inputPath includeDir [outputPath]\n", appName);
```
cheneym2, line 101: If includeDir is required, maybe add "|| !includeDir" to the condition above, otherwise if includeDir is optional, it should be shown in brackets here.

cheneym2 (approved):

tools/slang-embed/slang-embed.cpp
saipraveenb25: Uh-oh, I think auto-merge merged it in before I could react to the comment..

Let me open another PR for the fix

Issue #5283, opened 2024-10-14
WGSL: Signed int initialized using unsigned int


aleino-nv:
Affected tests:

- tests/hlsl-intrinsic/byte-address-buffer/byte-address-struct.slang

WGPU errors:

    WGPU error: Error while parsing WGSL: :17:5 error: cannot initialize 'var' of type 'i32' with value of type 'u32'
        var _S1 : i32 = inputBuffer_0[(offsetX_0)/4];

Generated WGSL:

    @binding(2) @group(0) var<storage, read_write> outputBuffer_0 : array<i32>;
    
    @binding(0) @group(0) var<storage, read_write> inputBuffer_0 : array<u32>;
    
    @binding(1) @group(0) var<storage, read_write> tmpBuffer_0 : array<u32>;
    
    fn test_0( val_0 : i32) -> i32
    {
        var offsetX_0 : i32 = val_0 * i32(8);
        var offsetY_0 : i32 = offsetX_0 & i32(255);
        var _S1 : i32 = inputBuffer_0[(offsetX_0)/4];
        var _S2 : i32 = offsetX_0 + i32(4);
        var _S3 : f32 = inputBuffer_0[(_S2)/4];
        tmpBuffer_0[(offsetY_0)/4] = _S1;
        tmpBuffer_0[(offsetY_0 + i32(4))/4] = _S3;
        var _S4 : i32 = tmpBuffer_0[(offsetX_0)/4];
        var _S5 : f32 = tmpBuffer_0[(_S2)/4];
        return _S4 * i32(256) * i32(16) + i32(_S5);
    }
    
    @compute
    @workgroup_size(4, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        var tid_0 : i32 = vec3<i32>(dispatchThreadID_0).x;
        var _S6 : i32 = test_0(tid_0);
        outputBuffer_0[tid_0] = _S6;
        return;
    }

aleino-nv:
> slang/source/slang/slang-emit.cpp

Ok this patch doesn't seem to help, at least:


    NVIDIA.COM+aleino@NV-HT9ZR14 MINGW64 ~/workspaces/slang (aleino/test-clean)
    $ git diff
    diff --git a/source/slang/slang-emit.cpp b/source/slang/slang-emit.cpp
    index 010b4bc9..ff4efb45 100644
    --- a/source/slang/slang-emit.cpp
[...]

PR #5681, opened 2024-11-26
wgpu: Enable Metal-like legalization for byte addressible buffers

aleino-nv:
Enable the same options as for Metal, plus the "bitcast from UInt" option.

Closes issue #5283.


jkwak-work (approved):
Looks good to me

Issue #5605, opened 2024-11-20
WGPU: f16 extension not allowed in current environment


aleino-nv:
Affected tests:

- `tests/language-feature/generics/variadic-0.slang`
- `tests/language-feature/generics/tuple.slang`

Example output:

    WGPU error: Error while parsing WGSL: :1:8 error: extension 'f16' is not allowed in the current environment
    enable f16;
           ^^^
    
    
     - While validating [ShaderModuleDescriptor]
     - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).

aleino-nv:
I found that Dawn is not returning the required `WGPUFeatureName_ShaderF16` feature from `wgpuAdapterEnumerateFeatures`.
On my machine, I get these features:

1. `WGPUFeatureName_DepthClipControl`
2. `WGPUFeatureName_Depth32FloatStencil8`
3. `WGPUFeatureName_TimestampQuery`
4. `WGPUFeatureName_TextureCompressionBC`
5. `WGPUFeatureName_IndirectFirstInstance`
6. `WGPUFeatureName_RG11B10UfloatRenderable`
7. `WGPUFeatureName_BGRA8UnormStorage`
8. `WGPUFeatureName_Float32Filterable`
9. `WGPUFeatureName_DawnInternalUsages`
10. `WGPUFeatureName_DawnMultiPlanarFormats`
11. `WGPUFeatureName_DawnNative`
12. `WGPUFeatureName_ImplicitDeviceSynchronization`
13. `WGPUFeatureName_SurfaceCapabilities`
14. `WGPUFeatureName_DualSourceBlending`
15. `WGPUFeatureName_Unorm16TextureFormats`
16. `WGPUFeatureName_Snorm16TextureFormats`
17. `WGPUFeatureName_MultiPlanarRenderTargets`
18. `WGPUFeatureName_AdapterPropertiesMemoryHeaps`
19. `WGPUFeatureName_AdapterPropertiesD3D`
20. `WGPUFeatureName_FormatCapabilities`
21. `WGPUFeatureName_Norm16TextureFormats`
22. `WGPUFeatureName_SharedTextureMemoryDXGISharedHandle`
23. `WGPUFeatureName_SharedFenceDXGISharedHandle`

I guess this situation is something we should handle in slang-test, but first I want to understand why `WGPUFeatureName_ShaderF16` is apparently not supported on my machine...

aleino-nv:
[...]

PR #5694, opened 2024-11-27
wgsl: Conditionally enable some tests based on 'half' support

aleino-nv:
Require the 'half' rendering feature for WebGPU, so that we can detect when the f16 feature is not available, for the following tests:

- tests/language-feature/generics/tuple.slang
- tests/language-feature/generics/variadic-0.slang

This helps to address issue #5605.


saipraveenb25 (approved):
Looks good!

Issue #5691, opened 2024-11-26
undefined identifier `nonuniformEXT` compiler error


theblackunknown:
Hi team, 

While porting our GLSL shaders to be compiled with `slangc`, I am running into a compiler error where `nonuniformEXT` is unexpected.
It is a feature offered by [GL_EXT_nonuniform_qualifier](https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_nonuniform_qualifier.txt) that we are using in our shaders, I should expect Slang to resolve it out of the box, or if I need to tell the Slang compiler that I need this extension for things to resolve correctly ?

See below for a minimal repro case.
If you can point me to how to port it as a Slang test I would gladly contribute it if that helps.

Given the following `nonuniformExt.slang` file
```slang
#version 450 core

layout(location = 0) out mediump vec4 o_color;
layout(location = 0) flat in int i;
layout(set = 0, binding = 0) uniform sampler2D tex[2];

void main()
{
    o_color = texture(tex[nonuniformEXT(i)], vec2(0.25, 0.75));
}
```

compiling the above file using the following command line
```slangc.exe -target spirv -stage fragment -entry main nonuniformExt.slang```

Will produce the following errors
```
tests\glsl\nonuniformExt.slang(9): error 30015: undefined identifier 'nonuniformEXT'.
    o_color = texture(tex[nonuniformEXT(i)], vec2(0.25, 0.75));
                          ^~~~~~~~~~~~~
tests\glsl\nonuniformExt.slang(9): error 30014: index expression must evaluate to int.
    o_color = texture(tex[nonuniformEXT(i)], vec2(0.25, 0.75));
```

**Comment by jkwak-work** [comment]
This appears to be a missing feature support for GLSL.
As a workaround, you can have a following line,
```
#define nonuniformEXT NonUniformResourceIndex
```

theblackunknown:
Thanks @jkwak-work, I have applied your suggestion and it seems to work !
Would you be able to point me at few places if I want to contribute to add this feature ?

jkwak-work:
I have a quick PR for this issue #5695

theblackunknown:
I see that #5695 has been merged, should we close this issue now ?

jkwak-work:
Closing as completed as #5695

PR #5695, opened 2024-11-27
Support nonuniformEXT for GLSL

jkwak-work:
Adding nonuniformEXT intrinsic for GLSL


jkwak-work:
Closes https://github.com/shader-slang/slang/issues/5691

expipiplus1 (approved):
lgtm, thanks

source/slang/glsl.meta.slang
```diff
@@ -9601,3 +9601,14 @@ public bool allInvocationsEqual(bool value)
...
     return WaveActiveAllEqual(value);
 }
+
+
+/// extensions
+/// https://github.com/KhronosGroup/GLSL/blob/main/extensions/ext/GL_EXT_nonuniform_qualifier.txt
+
+__generic<T:__BuiltinArithmeticType>
+__intrinsic_op($(kIROp_NonUniformResourceIndex))
+[require(cpp_cuda_glsl_hlsl_spirv, nonuniformqualifier)]
```
aleino-nv, line 9611: "cpp_cuda_glsl_hlsl_spirv" seems surprising. Isn't this a GLSL-only thing?

expipiplus1: it's implemented by the slang IR op `kIROp_NonUniformResourceIndex`, which is supported on those platforms. The requirement here is for the backend, not the source language. It should be exactly the same implementation as HLSL's `NonUniformResourceIndex`

jkwak-work: I am piggy backing on how HLSL implemented.

Alternative implementation could be something like,
```
#define nonuniformEXT NonUniformResourceIndex
```
But this would be ugly.

aleino-nv: I see, thanks for the explanation!

PR #5699, opened 2024-11-28
Test toc generation bot

expipiplus1:


expipiplus1:
> Wow. What are you cooking up here? I am gonna call you bot-maker from now one. ;-)

I noticed a few times that the table of contents got out of date, so we should check it in CI, but it was also a pain to regenerate, so why not have the bot do it

expipiplus1:
ok, all working :partying_face:

jkwak-work:
Is `/regenerate-toc` working now?
I think I need to update the CONTRIBUTING.md for these magic bot commands.

PR #5700, opened 2024-11-28
Make ToC regeneration script path handling more robust

expipiplus1:
Should allow it to be used from the toc fix bot


jkwak-work:
Looks good to me.

PR #5716, opened 2024-12-01
Add wgsl missing float image format inference

### LINKED ISSUE #5704: Compiling RWTexture2D<float> to wgsl leads to rgba32float instead of r32float
Currently:
`RWTexture2D<uint4>` => `texture_storage_2d<rgba32uint, read_write>`
`RWTexture2D<uint>` => `texture_storage_2d<r32uint, read_write>`
`RWTexture2D<float4>` => `texture_storage_2d<rgba32float, read_write>`
`RWTexture2D<float>` => `texture_storage_2d<rgba32float, read_write>`

`RWTexture2D<float>` should become `texture_storage_2d<r32float, read_write>` to match the behavior of other languages.

This bug will lead to storage images requiring a format specifier in the playground once https://github.com/shader-slang/slang-playground/pull/67 is merged.

Devon7925:
fixes #5704


csyonghe (approved):

csyonghe:
Looks good to me. Thank you for your contribution!

csyonghe:
There is a test whose expected result needs to be updated.

Devon7925:
Not sure how to do that.

csyonghe:
Actually the test failed because the generated spirv failed validation. We need to investigate which validation rule is being violated.

Devon7925:
On my computer that test doesn't fail so I'm not sure how to proceed.

Devon7925:
The test fails due to this line
`%sampledImageType = OpTypeSampledImage $$Texture2D<float>;`
It still infers unknown for the texture format and that doesn't match the newly inferred type of the texture. The test also fails independently of this pr if `half` is used instead of `float`. 

I've been unable to figure out where to put the inference code though I'm still looking.

csyonghe:
Ah, actually we do want to have `Texture2D<float>` or `Texture2D<float4>` to map to `unknown` format when generating SPIRV, so we shouldn't be adding the image format inference logic here.

The reasoning is that when the user defines something like `Texture2D texture;` without a format attribute, it automatically expands to `Texture2D<float4> texture;`, and in this case we shouldn't produce SPIRV code that declares the format to be `RGBA32F`. Most often the texture will NOT be in that format and instead be something like `rgba8unorm`. Specifying `unknown` here is the only correct behavior.

Devon7925:
I could infer for `float` but not `float4` if the issue is just the `Texture2D` to `Texture2D<float4>` conversion which would fix the issue in wgsl. Would that work?

Devon7925:
It would require modifying the test to use float4 instead of float, but since the test doesn't work with half, it is already testing a special case.

csyonghe:
Inferring for `float` maybe OK, since it is a less common case. But still the argument applies for `R8UNORM` formats.

It is safer to not infer for `float`, because all unorm/snorm formats can map to `float`.

The right fix to this problem IMO, is to allow user to write `Texture2D<r8unorm>` to be more explicit on the format,  but t his requires some refactoring on how texture types are represented internally.

Devon7925:
People can use attributes to specify format as well. `R10G10B10A2_UINT` creates ambiguity for `UInt4`. If we want to prevent ever inferring the wrong format maybe removing format inference all together would be better. I would argue getting `rgba32` for `float` in wgsl due to not inferring is just as bad as getting other undesired formats due to inferring.

csyonghe:
For WGSL where there isn't a "unknown" format, I agree that any inference is better.
But we need to be careful to not break SPIRV applications by adding more inferencing.

csyonghe:
Perhaps we should do this during wgsl emit time, so we can do a last minute infer for wgsl when the default front-end infer failed.

csyonghe (dismissed):

csyonghe:
Looks good, need a formatting pass.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -343,6 +345,38 @@ static const char* getWgslImageFormat(IRTextureTypeBase* type)
     //
     ImageFormat imageFormat =
         type->hasFormat() ? (ImageFormat)type->getFormat() : ImageFormat::unknown;
+
+    if (imageFormat == ImageFormat::unknown)
```
jkwak-work, line 349: The `switch` statement right after this `if`, there is a case for `ImageFormat::unknown`.
This code change could be placed inside of the switch.

Devon7925: I originally had it inside the switch statement. Having it outside makes it a bit easier to mentally parse imo. It's the difference between
```
handle edge cases
do main operation
```
and
```
do main operation except in the case of edge cases where you handle them
```

PR #5717, opened 2024-12-02
Conform non-suffixed integer literals

### LINKED ISSUE #5458: Incorrect declaration of 64 bit static arrays
I have found a very perplexing problem, which occurs when declaring a 64 bit global array as a bitmask.
    
    struct ComputeParams
    {
        RWStructuredBuffer<uint64_t> bitFieldBuffer;
    };
    ParameterBlock<ComputeParams> pParams;

    static const uint64_t OCBT_bit_mask[18] = { 0xffffffff, // Root 17
                                0xffffffff, // Level 16
                                0xffffffff, // level 15
                                0xffffffff, // level 14
                                0xffffffff, // level 13
                                0xffffffff, // level 12
                                0xffffffff, // level 11

                                0xffff, // level 10
                                0xffff, // level 9
                                0xffff, // level 8
                                0xff, // level 8

                                0xffffffffffffffff, // level 7
                                0xffffffff, // Level 6
                                0xffff, // level 5
                                0xff, // level 4
                                0xf, // level 3
                                0x3, // level 2
                                0x1, // level 1
    };

    [numthreads(1, 1, 1)]
    [shader("compute")]
    void TestHeap(uint dispatchID: SV_DispatchThreadID)
    {
        pParams.bitFieldBuffer[dispatchID] = OCBT_bit_mask[dispatchID];
    }

This produces the following GLSL code:

        #version 450
    #extension GL_EXT_shader_explicit_arithmetic_types_int64 : require
    layout(row_major) uniform;
    layout(row_major) buffer;

    #line 34 0
    layout(std430, binding = 0) buffer StructuredBuffer_uint64_t_0 {
        uint64_t _data[];
    } pParams_bitFieldBuffer_0;

    #line 8
    const uint64_t  OCBT_bit_mask_0[18] = { 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 65535UL, 65535UL, 65535UL, 255UL, 18446744073709551615UL, 18446744073709551615UL, 65535UL, 255UL, 15UL, 3UL, 1UL };

    #line 32
    void main()
    {

    #line 32
        uint _S1 = gl_GlobalInvocationID.x;

        pParams_bitFieldBuffer_0._data[uint(_S1)] = OCBT_bit_mask_0[_S1];
        return;
    }

In the GLSL output, the array has replaced the 32 bit masks with 64 bit masks. The problem persits when compiling with `emit-spirv-directly` and decompiling with `spriv-cross`

fairywreath:
Fixes #5458. 

Make non-suffixed integer literals conformant to the [C++ standard](https://en.cppreference.com/w/cpp/language/integer_literal). The type of a decimal non-suffixed integer literal is the first integer type from the list [`int`, `int64_t`] which can represent the specified literal value. If the value cannot fit, the literal is represented as  an `uint64_t` and a warning is given. The type of hexadecimal non-suffixed integer literal is the first type from the list [`int`, `uint`, `int64_t`, `uint64_t`] that can represent the specified literal value.  The current non-suffixed behavior forces all integer literals to be represented as signed 32 bit integers.

 The change consists of:
- Implementation in the parser with a new "integer is too large" warning, similar to clang and gcc
- Tests modification to properly test the new behaviour
- Documentation changes

Unfortunately there is an edge case when parsing `-9223372036854775808`(smallest int64 value). Because the lexer handles the negative(-) part of the literal separately. `9223372036854775808` is parsed by itself and it exceeds `9223372036854775807`(largest int64 value), and a warning is emitted although it should not as `-9223372036854775808` is within the range of int64 . When negation is parsed the value will still be properly assigned to `-9223372036854775808`. This compiler warning is also seen in clang and gcc. A proper fix for this would require lexer changes.


csyonghe (approved):
Looks good to me, thank you for the fixes!

We don’t need to address in this PR, but the change does reveal that we are missing a check during coercion of integer literals when the literal is out of range.

for example, we should still warn when the user writes
int x=0xffffffffffff;

But this check can’t be done in the lexer and needs to happen during coerce().

csyonghe:
Mark the PR as a breaking change because some corner case behaviors around literal types are changed.

csyonghe:
The `tests/slang-extension/atomic-int64-byte-address-buffer` test needs to be updated as well.

fairywreath:
Regarding coercion,

` x=0xffffffffffff` does give a warning:
```
1.slang(34): warning 30081: implicit conversion from 'int64_t' to 'int' is not recommended
    int x = 0xffffffffffff;
            ^~~~~~~~~~~~~~
```
But there is a weird issue when coercing UINT64_MAX which I have written down [here](#5728) which I fill fix in the future.
The current(without this PR's changes) behavior also does not give any warning, while it should give the "integer literal truncated" warning.

csyonghe (approved):
Looks good to me. Thank you for fixing the tests!

PR #5731, opened 2024-12-03
WGPU: Add new test and explanation for disabled test

### LINKED ISSUE #5612: Slang-RHI/WGPU: Invalid bind group layout entry used
Affected tests:

- `tests/bugs/buffer-swizzle-store.slang`

Example output:

    WGPU error: BindGroupLayoutEntry had none of buffer, sampler, texture, storageTexture, or externalTexture set
     - While validating entries[0]
     - While validating [BindGroupLayoutDescriptor]
     - While calling [Device].CreateBindGroupLayout([BindGroupLayoutDescriptor]).

aleino-nv:
- WGPU cannot support read-write access for storage textures of the format in the disabled test, so an explanation is added.
- A similar test is added, which uses structured buffers instead and should work also on WGSL.

See comments on https://github.com/shader-slang/slang/issues/5612 for more details.

This closes #5612.


tests/bugs/buffer-swizzle-store.slang
```diff
@@ -1,7 +1,7 @@
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type
 //TEST(compute):COMPARE_COMPUTE:-slang -shaderobj -mtl -output-using-type
-// Slang-RHI/WGPU: Invalid bind group layout entry used #5612
+// Not supported in WGSL: read-write storage texture with "rg16f" format
```
jkwak-work, line 4: If the format is only problem for WGSL, can we change the format to one of formats WGSL supports?
Or maybe we can copy this test with a different format.

jkwak-work (approved):
Looks good to me

tests/bugs/buffer-swizzle-store.slang
aleino-nv: Yeah I can give it a shot.
There are [some Slang-RHI issues related to storage textures](https://github.com/shader-slang/slang/issues/5612#issuecomment-2514257576) I need to fix first though -- it currently doesn't create the bind group layout entry properly. I think I should set the `access` field to `read-write`, but I'm not quite sure how I should then get the required format information. I think I'll have to check in with Simon.

csyonghe (approved):

csyonghe:
Needs another re-format.

jkwak-work (approved):

Issue #5504, opened 2024-11-06
Write design proposal for extended length vector support


expipiplus1:

jkwak-work:
Does this contradict to #5471 ?

csyonghe:
I don't think so. Until we actually implemented long vectors, we need the documentation to state the limitation.

PR #5734, opened 2024-12-03
Extended vector proposal

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/5504


csyonghe (commented):
What about standard library functions around vectors?
sin(v) continues to work for `vector<float, 256>`?

expipiplus1:
> What about standard library functions around vectors? sin(v) continues to work for `vector<float, 256>`?

Yup, I've amended the proposal to reflect this. Any operation which at the moment is generic over 1,2,3,4 vectors will be supported (barring platform limitations)

csyonghe (approved):

PR #5736, opened 2024-12-03
Add intrinsics for aligned load/store.

### LINKED ISSUE #5737: Add aligned load/store intrinsic.
We should add `T loadAligned<alignment, T>(T*)` and `void storeAligned<alignment, T>(T*, T val)` intrinsics to the core module that maps to `OpLoad` and `OpStore` with a `Aligned` memory operand when generate SPIRV code from pointers.

This should allow the downstream compiler to emit more efficient load/store code.

Note that we are intentionally not adding alignment as a property of the pointer type, to avoid having to deal with the problem of additional type coercions and alignment inferencing rules for access chain operations.

csyonghe:
Closes #5737.


cheneym2 (approved):
The implementation looks good, is this below the threshold of complexity where we need to go through the feature proposal process?

cheneym2 (approved):

csyonghe:
@cheneym2 we should follow the feature change process. But since this is a trivial change, I am checking in the proposal itself along with the implementation.

cheneym2 (approved):

PR #5738, opened 2024-12-03
Core lib Metal math function fixes

### LINKED ISSUE #5580: fmax and fmin not implemented to metal spec for non-metal targets
It appears to be an oversight in the implementation of fmax() and fmin() for non-metal targets.

Same issue with fmin is present as described below with fmax().

In the implementation of fmax, metal implements using an intrinsic, so presumably meets the Apple definition of fmax() which is:

```
Returns y if x < y, otherwise returns x. If one
argument is a NaN, fmax() returns the other
argument. If both arguments are NaNs, fmax()
returns a NaN. If x and y are denormals and the
GPU doesn’t support denormals, either value
may be returned.
```

It looks like there was an attempt to match that in the fmax implementation for other targets, but for some reason it only checks one of the special conditions, notably, it doesn't return x if isnan(y), and it doesn't return NaN if x and y are both NaN.

```
[__readNone]
[require(cpp_cuda_glsl_hlsl_metal_spirv, sm_4_0_version)]
T fmax(T x, T y)
{
    __target_switch
    {
    case metal: __intrinsic_asm "fmax";
    default:
        if (isnan(x)) return y;
        return max(x, y);
    }
}
```

### LINKED ISSUE #5581: fmax3 implemented incorrectly for non-metal targets
By inspection, this appears to have a bug where X is NaN and Y and Z are not specials.

fmin3() has a similar issue as described below for fmax3()

fmax3(NaN, 1, 2) should return 2 but would return 1.

```
[__readNone]
[require(cpp_cuda_glsl_hlsl_metal_spirv, sm_4_0_version)]
T fmax3(T x, T y, T z)
{
    __target_switch
    {
    case metal: __intrinsic_asm "fmax3";
    default:
    {
        bool isnanX = isnan(x);
        bool isnanY = isnan(y);
        bool isnanZ = isnan(z);

        if (isnanX)
        {
            return isnanY ? z : y;
        }
        else if (isnanY)
        {
            if (isnanZ)
                return x;
            return max(x, z);
        }
        else if (isnanZ)
        {
            return max(x, y);
        }

        return max(y, max(x, z));
    }
    }
}
```

### LINKED ISSUE #5587: powr() implemented inefficiently for non-metal targets
It seems like powr() is implemented inefficiently on non-metal targets.  If I understand correctly, powr() is only defined for x>=0 but for some reason we coerce x to be abs(x) before delegating to regular pow() for other targets, instead of letting it be undefined.

```
/// @category math
[__readNone]
[require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, sm_4_0_version)]
T powr(T x, T y)
{
    __target_switch
    {
    case metal: __intrinsic_asm "powr";
    default:
        return pow(abs(x), y);
    }
}
```

cheneym2:
In order to emit fast target implementations of some Metal-based functions (fmin(), fmax(), fmin3(), fmax3()) on all targets, remove some specification regarding the handling of NaNs, and also remove the enforcement of the specification.

These functions are now documented to be basically undefined now in the presence of NaN input, to make the common "is a number" case fast.

Also, clarify that powr() is undefined when given a non-positive base input value, allowing us to remove an additional abs() operation that was unnecessarily coercing results to be predictable on non-Metal targets.

Closes #5580
Closes #5581
Closes #5587

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -10275,12 +10275,11 @@ vector<T,N> max3(vector<T,N> x, vector<T,N> y, vector<T,N> z)
     }
 }
 
-/// Floating-point maximum considering NaN.
+/// Floating-point maximum.
 /// @param x The first value to compare.
 /// @param y The second value to compare.
-/// @return The larger of the two values, element-wise if vector typed, considering NaN.
-/// @remarks For metal, if either value is NaN, the other value is returned. If both values are NaN, NaN is returned.
-/// For other targets, if `x` is NaN, `y` is returned, otherwise the larger of `x` and `y` is returned.
+/// @return The larger of the two values, element-wise if vector typed.
+/// @remarks Result is `y` if `x` < `y`, either `x` or `y` if both `x` and `y` are zeros, otherwise `x`. Which operand is the result is undefined if one of the operands is a NaN.
 /// @category math
 __generic<T : __BuiltinFloatingPointType>
 [__readNone]
@@ -10291,7 +10290,6 @@ T fmax(T x, T y)
     {
     case metal: __intrinsic_asm "fmax";
     default:
-        if (isnan(x)) return y;
         return max(x, y);
     }
 }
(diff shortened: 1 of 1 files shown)


jkwak-work (approved):
Looks good to me

PR #5759, opened 2024-12-04
Relocate the step of 'Add bash to PATH'

kaizhangNV:


csyonghe (approved):

jkwak-work (approved):
Looks good to me. 

@expipiplus1 added a step to check if the change is limited to documents. And it uses bash.

Issue #5607, opened 2024-11-20
WGSL: Functions cannot be called at module scope


aleino-nv:
Failing tests:

- `tests/language-feature/constants/static-const-in-generic-interface.slang`

Example output:

    WGPU error: Error while parsing WGSL: :8:24 error: functions cannot be called at module-scope
    const result_0 : i32 = EnsureCompileTimeEval_getValue_0();
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    
    
     - While validating [ShaderModuleDescriptor]
     - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).

aleino-nv:
@csyonghe I haven't looked into this yet, but from the name it sounds like Slang is emitting code where `EnsureCompileTimeEval_getValue_0()` is supposed to be evaluated by WGSL at compile time.

Is this something Slang does? WGSL has some support for something similar to "constexpr", but I think they have to be explicitly marked as such.

csyonghe:
I think this is supposed to be inlined and evaluated by slang so it never actually appears in the generated wgsl code.

aleino-nv:
For some reason I cannot reproduce by the problematic code by just doing a stand-alone compilation:

    slangc -o test.wgsl -target wgsl -stage compute -entry computeMain static-const-in-generic-interface.slang

This produces the following code:

    @binding(0) @group(0) var<storage, read_write> outputBuffer_0 : array<i32>;
    
    fn EnsureCompileTimeEval_getValue_0() -> i32
    {
        return i32(4);
    }
    
    @compute
    @workgroup_size(1, 1, 1)
    fn computeMain(@builtin(global_invocation_id) dispatchThreadID_0 : vec3<u32>)
    {
        outputBuffer_0[i32(0)] = EnsureCompileTimeEval_getValue_0();
        return;
    }

I'll debug a bit to see why slang-test somehow comes up with different code, and what the complete generated code looks like.

aleino-nv:
This is the code that actually gets used to create the WebGPU shader module.
It has the problem that the error message is complaining about

    @binding(0) @group(0) var<storage, read_write> outputBuffer_0 : array<i32>;
    
    fn EnsureCompileTimeEval_getValue_0() -> i32
    {
        return i32(4);
    }
    
    const result_0 : i32 = EnsureCompileTimeEval_getValue_0();
    @compute
    @workgroup_size(1, 1, 1)
[...]

PR #5768, opened 2024-12-05
WGSL: Fix issue where global calls are generated

aleino-nv:
- Refactor the SPIR-V inlining legalization code so that it's reusable by other targets
- Apply said legalization code for WGSL
- Extend the legalization code such that globals left unused after inlining are removed

The last step (also the last commit) gets rid of the global call and fixes the issue.

https://github.com/shader-slang/slang/issues/5607


source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+#include "slang-ir-clone.h"
+#include "slang-ir-util.h"
+
+namespace Slang
+{
+
+void GlobalInstInliningContextGeneric::inlineGlobalValuesAndRemoveIfUnused(IRModule* module)
+{
+    Dictionary<IRInst*, List<IRUse*>> globalInstUsesToInline;
```
csyonghe, line 11: Why do we need to add everything into a dictionary first?

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,231 @@
...
+                    globalInstUsesToInline.getOrAddValue(globalInst, List<IRUse*>()).add(use);
+                }
+            }
+        }
+    }
+
+    for (auto globalInstUses : globalInstUsesToInline)
+    {
+        auto inst = globalInstUses.first;
+        auto uses = globalInstUses.second;
```
csyonghe, line 30: This is creating an unnecessary copy of the list.

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: I just changed it to a dictionary so that I can iterate over the keys when deleting later.

Alternatives:
- Use a separate list to record global insts to consider for deletion. These will be the same as the keys in the dictionary.
- Get the global insts from the uses we record, and somehow ensure we're not double-deleting

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: Thanks, changing to `auto&`. (I haven't been using `auto` much.)

aleino-nv:
Some tests were failing:

- `tests/spirv/global-compute.slang` fails for SPIR-V, seemingly because `OpEntryPoint` is removed
  - I have a hacky workaround and a TODO comment for this, at the moment.
  - I will try refining the logic so that only global insts _which actually had any uses to begin with_ are considered for removal
- Some WGSL tests failed because the inlining code produced `IRGlobalValueRef` which were not handled during emit.
  -  I added some code for this

source/slang/slang-ir-legalize-global-values.cpp
csyonghe: Yeah, just use a List<IRInst*> to store the insts for deletion.

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,233 @@
...
+            {
+                if (getParentFunc(use->getUser()) != nullptr)
+                {
+                    globalInstUsesToInline.getOrAddValue(globalInst, List<IRUse*>()).add(use);
+                }
+            }
+        }
+    }
+
+    for (auto globalInstUses : globalInstUsesToInline)
```
csyonghe, line 27: this will also trigger a deep copy to get `globalInstUses`.

aleino-nv: The code should now be back to what it was. I think this is now just a pointer copy.

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: Ok I switched to a list. The code also slightly changed meaning (the keys under consideration for deletion may now be a superset of the keys in the dictionary), but AFAIK it should still be ok.

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,233 @@
...
+        auto val = maybeInlineGlobalValue(builder, use->getUser(), use->get(), cloneEnv);
+        if (val != use->get())
+            builder.replaceOperand(use, val);
+    }
+
+    // Since certain globals that appear in the IR are considered illegal for all targets,
+    // e.g. calls to functions, we delete globals which no longer have uses after inlining.
+    for (auto& globalInst : globalInstsToConsiderDeleting)
+    {
+        // TODO: Explain why tests/spirv/global-compute.slang fails if we don't exclude
```
csyonghe, line 44: Todo in comment should be addressed.

aleino-nv: The issue was that it wasn't actually safe to remove the global inst, because I didn't check whether the inst _or any of its descendants_ had any usages.

For the mentioned global-compute test, there were some `SPIRVAsmOperandInst` descendants of the global inst that had usages left after inlining. (They didn't get inlined and so the inlined `SPIRVAsm` itself was using them.)

Is the fact that the `SPIRVAsmOperandInst`s didn't get inlined an issue we should fix?
If so I would like to file a new task for that, and do it in a later commit. (Right now it's not causing any problem.)

aleino-nv:
I got a test timeout...

      internal/renderpasses/test_InternalPathTracerVolumeTransmittance_vulkan : FAILED (600.0 s)
        Process killed due to timeout
        View test at: http://SlangWin5:8080//unknown/internal/renderpasses/test_InternalPathTracerVolumeTransmittance_vulkan

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,261 @@
...
+        if (val != use->get())
+            builder.replaceOperand(use, val);
+    }
+
+    // Since certain globals that appear in the IR are considered illegal for all targets,
+    // e.g. calls to functions, we delete globals which no longer have uses (or children with uses)
+    // after inlining.
+    for (auto& globalInst : globalInstsToConsiderDeleting)
+    {
+        if (!hasDescendantSatisfyingPredicate(
```
csyonghe, line 72: This is a band aid instead of an actual fix.

As we discussed in the meeting, the true problem is when we clone the global SpirvASM inst to a function, we didn't update the operand references to point to the cloned version of the local spirv operand insts, as a result, the inlined asm block is still referencing the child insts in the original global asm block.

The fix here just prevents removal of the global asm block, but the right fix is to make sure we dont have such inter block references back to the original global asm block.

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+#include "slang-ir-clone.h"
+#include "slang-ir-util.h"
+
+namespace Slang
+{
+
+// Returns true iff 'rootInst', or any of it's descendants, satisfies 'predicate'.
+template<typename P>
+bool hasDescendantSatisfyingPredicate(IRInst* rootInst, P predicate)
```
csyonghe, line 11: I don't think this is necessary. See my comment below.

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: Ok, I must have misunderstood something. 

Yesterday I stepped through the inlining code in the debugger and noticed that the `SPIRVAsmOperandInst`s didn't get inlined (unlike all of the other operands) and so there was no clone to be found.

I figured this is something that can happen in general, and so we must check for uses recursively before deleting the global inst.

So you are saying that the existing inlining logic is broken?

I will take another look.

aleino-nv: @csyonghe Should the fix be that, when we decide to inline SPIRV ASM insts then we force-inline things it contains, like the `SPIRVAsmOperandInst`s etc.. ?

csyonghe: When we looked through the code together, it appears to me that the SPIRVAsm parent inst is being cloned, and its children are also being cloned. It is just the operands of the clone seems to be still referencing the original ones instead of the cloned ones.

e.g.

```
%b = SPIRV_Asm {
    %x = SPIRVAsmOperandInst ...
    %y = SPIRVAsmOperandInst ...
     %z = SPIRVAsmInst %x %y
}
```

got cloned into `main` as:

```
func %main  {
%block = block {
    %b' = SPIRV_Asm {
       %x' = SPIRVAsmOperandInst ...
       %y' = SPIRVAsmOperandInst ...
        %z' = SPIRVAsmInst %x %y // !!! should be %x' here, instead of %x
    }
}
}
```

aleino-nv: The reason that the SPIRV Operands are not inlined is that they are legal global insts, according to this:

https://github.com/shader-slang/slang/blob/772c48af91951bda9ff7064fc7a089550c759f76/source/slang/slang-ir-spirv-legalize.cpp#L1614

All of the other operands of the SPIRVAsmInst being inlined are not considered legal global insts, and so they are inlined.

aleino-nv: Oh, I think I see now!

The `SPIRVAsmOperand`s are special in that they are parented to the `SPIRVAsm` which we would like to delete.

The `inlineInst` function does indeed inline its children, but then later when we get to the last child which is a `SPIRVAsmInst` taking `SPIRVAsmOperand`s as operands, we decide not to inline them.

I see that there is a dictionary called `m_mapGlobalInstToShouldInline` that is first consulted.
Perhaps whenever we inline some inst we should set it as `true` in `m_mapGlobalInstToShouldInline`...

aleino-nv: > Oh, I think I see now!
> 
> The `SPIRVAsmOperand`s are special in that they are parented to the `SPIRVAsm` which we would like to delete.
> 
> The `inlineInst` function does indeed inline its children, but then later when we get to the last child which is a `SPIRVAsmInst` taking `SPIRVAsmOperand`s as operands, we decide not to inline them.
> 
> I see that there is a dictionary called `m_mapGlobalInstToShouldInline` that is first consulted. Perhaps whenever we inline some inst we should set it as `true` in `m_mapGlobalInstToShouldInline`...

That works, but only because the SPIRVAsmInst taking these SPIRVAsmOperandInst's as operands happen to appear after the SPIRVAsmOperandInst's children of the SPIRVAsm... 

If I instead mark `m_mapGlobalInstToShouldInline` to true for any inlined insts first thing in `inlineInst`, and then I can change `shouldInlineInstImpl` to return `true` in case any ancestors of the inst in question should be inlined according to `m_mapGlobalInstToShouldInline`.

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: Deleted.

Yeah it shouldn't be needed anymore if my fix to the inlining logic works as I expect.
The fix is detecting if any ancestor of anything used by the inlined value is being inlined, and forces inlining of that thing if so.

That should mean that the inlined values don't use stuff from anywhere under the thing being inlined..

source/slang/slang-ir-legalize-global-values.cpp
aleino-nv: I did something similar to the above, and it seems to work.

source/slang/slang-ir-legalize-global-values.cpp
```diff
@@ -124,6 +139,15 @@ bool GlobalInstInliningContextGeneric::isInlinableGlobalInst(IRInst* inst)
 
 bool GlobalInstInliningContextGeneric::shouldInlineInstImpl(IRInst* inst)
 {
+    // If 'inst' has an ancestor that is currently being inlined, then we
+    // better inline it since we'll be removing the ancestor.
+    bool ancestorShouldBeInlined = false;
+    for (IRInst* ancestor = inst->parent; ancestor != nullptr; ancestor = ancestor->parent)
+        if (m_mapGlobalInstToShouldInline.tryGetValue(inst, ancestorShouldBeInlined) && ancestorShouldBeInlined)
+            break;
```
csyonghe, line 148: Just return true here.

aleino-nv: Done.

csyonghe (approved):
Looks good to me.

PR #5777, opened 2024-12-05
Fix recursive function checks for Metal

### LINKED ISSUE #5684: Compiler hangs on recursive functions when targeting Metal
Attempting to compile the following code targeting Metal results in the compiler hanging:
```
[shader("compute")]
func m() {
    r();
}

func r() {
    r();
}
```

I stumbled on this while messing around in the playground: https://shader-slang.com/slang-playground/?target=METAL&code=eJyLLs5ITEkt0lBKzs8tKC1JVdKM5UorzUtWyNXQVKjmUgCCIg1Na65aLohwEZowAHEsEU0

fairywreath:
Fixes #5684 .

Move recursive function checks to be done earlier during `linkAndOptimizeIR`. The current implementation hangs on Metal because recursive checks are handled towards the end of this function inside `checkUnsupportedInst`. This is after Metal's IR legalization pass, where `specializeAddressSpace` is called and cause an infinite worklist run.


fairywreath:
`source/slang/slang-ir-check-recursive-type.cpp` is renamed to `source/slang/slang-ir-check-recursion.cpp` and the function recursive checks are moved from `source/slang/slang-ir-check-unsupported-inst.cpp` to this renamed file. Github does not consider this a rename because the rename detection is too high and too much new code is added to the file.

csyonghe (approved):
Change looks good to me. Thank you for the fix!

PR #5783, opened 2024-12-06
Emit debug info for Release builds

expipiplus1:


expipiplus1:
Annoyingly the debug files for MacOS are very small (empty) with only some cryptic messages in the logs... Could someone with a Mac @csyonghe, @kaizhangNV (I think?) Maybe take a look at this, as debugging via actions is just not practical

PR #5792, opened 2024-12-06
Report error on nested functions.

### LINKED ISSUE #5624: Local functions
Without checking the manual too closely I tried writing a local function like this:

```
float4 fragmentMain(VertexOutput vertex) : SV_Target
{
  static float GetValue(int val) { return val; } // Simplified example
  float a = GetValue(vertex.Input);

  // etc
}
```

So from what I understand this is not something that's actually supported, there's currently no support for local functions/lambda functions/closures/etc. But it still compiles into correct code - is this parsed as a member function? It has a few problems:

* It crashes the language server in VS Code, i.e. code completion stops working.
* When I tried to use a value from an outer function (thinking closures are supported) it compiled, but generated invalid GLSL.

Unless I'm completely off the mark here, declaring a function like GetValue inside a function should probably just produce an error.

csyonghe:
Closes #5624.

### DIFF
--- a/source/slang/slang-parser.cpp
+++ b/source/slang/slang-parser.cpp
@@ -5796,6 +5796,9 @@ DeclStmt* Parser::parseVarDeclrStatement(Modifiers modifiers)
     else if (as<TypeDefDecl>(decl))
     {
     }
+    else if (as<UsingDecl>(decl))
+    {
+    }
     else
     {
         sink->diagnose(decl->loc, Diagnostics::declNotAllowed, decl->astNodeType);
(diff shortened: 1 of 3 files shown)


source/slang/slang-parser.cpp
```diff
@@ -5783,6 +5783,27 @@ DeclStmt* Parser::parseVarDeclrStatement(Modifiers modifiers)
...
+    }
+    else if (as<AggTypeDecl>(decl))
+    {
+    }
+    else if (as<TypeDefDecl>(decl))
+    {
+    }
+    else if (as<UsingDecl>(decl))
+    {
+    }
```
expipiplus1, line 5801: Quite a weird way of avoiding `||`?

PR #5804, opened 2024-12-08
Parse slang.h for future language binding generation

laurooyen:
This implements parsing of *slang.h* as part of #5565. It adds the specific requirements to parse that file to *slang-cpp-parser* which was introduced in #5675. Additionally this adds a new target *slang-binding-generator* that can be used in the future to implement the language bindings in. For now it just provides a testing environment.

Currently a draft PR so people know this is being worked on and to collect feedback on how the lack of a preprocessor in the parser should be handled, see `TODO:` comments in the code.

Before this gets merged I'd still like to do some refactoring to reduce code duplication between the newly created *slang-binding-generator* and *slang-cpp-extractor* which it was derived from.

Even though the newly added target doesn't generate code yet, we should probably already run it as part of the build to ensure this work doesn't break once merged. Suggestions on how to do that in CMake are welcome.

To test the parser, you may run a command similar to this: `slang-binding-generator -d C:\dev\slang\include slang.h -dump`.

### DIFF
--- a/include/slang.h
+++ b/include/slang.h
@@ -1729,8 +1729,6 @@ public:                                                              \
     typedef slang::IGlobalSession SlangSession;
 
 
-    typedef struct SlangProgramLayout SlangProgramLayout;
-
     /*!
     @brief A request for one or more compilation actions to be performed.
     */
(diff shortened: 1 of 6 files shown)


QuantumDeveloper:
@laurooyen Is there any news on this? I am really interested of getting C bindings for slang. Do you know when this PR could be done and merged?

Issue #4681, opened 2024-07-18
Need further development of built-in interfaces


swoods-nv:

bmillsNV:
Needs proper design doc and design discussions. We need to focus on this if we're going to have folks convert their HLSL template shaders to use generics/interfaces.

bmillsNV:
This is for the arithmetic interfaces.

PR #5810, opened 2024-12-09
Add Abstract Algebra Interface Hierarchy proposal

expipiplus1:
closes https://github.com/shader-slang/slang/issues/4681

### DIFF
--- a/docs/proposals/015-abstract-algebra-interface-hierarchy.md
+++ b/docs/proposals/015-abstract-algebra-interface-hierarchy.md
@@ -1,0 +1,347 @@
+# SP #015: Abstract Algebra Interface Hierarchy
+
+This proposal introduces a hierarchy of algebraic structure interfaces in
+Slang, providing a foundation for principled generic mathematical
+operations.
+
+## Status
+
+Status: Design Review
+
+Implementation: N/A
+
+Author: Ellie Hermaszewska
+
+Reviewer: TBD
+
+## Background
+
+Currently, Slang lacks a formal way to express algebraic properties of types.
+While basic arithmetic operations are supported, there's no standard vocabulary
+to express a type's conformance to algebraic laws. This is particularly
+relevant for graphics programming where we frequently work with a variety of
+mathematical objects, each fulfilling different properties described by
+abstract algebra, for example:
+
+- Vector spaces
+- Quaternions
+- Dual numbers
+- Matrices
+- Complex numbers
+
+Additionally, many types support only a subset of arithmetic operations. For
+example:
+
+- Strings support concatenation (forming a monoid) but not subtraction
+- Colors support addition and scalar multiplication but not division
+- Matrices support multiplication but aren't always invertible
+- Non-empty containers support concatenation (forming a semigroup) but lack an
+  identity element
+
+A single catch-all "Arithmetic" interface would be insufficient as it would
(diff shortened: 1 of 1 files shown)


docs/proposals/015-abstract-algebra-interface-hierarchy.md
```diff
@@ -0,0 +1,347 @@
...
+- Scala: Abstract algebra typeclasses in libraries like Spire
+
+## Proposed Approach
+
+Implement the following interfaces, replacing in part `__BuiltinArithmeticType`
+
+```slang
+// Semigroup represents types with an associative addition operation
+// Examples: Non-empty arrays (concatenation), Colors (blending)
+interface Semigroup
```
csyonghe, line 68: All interfaces should be named with `I` prefix in Slang's core module naming convention.

csyonghe (commented):
While adopting concepts from abstract algebra gives us flexibility and comprehensiveness, it is not necessarily the most accessible way to approach game and system developers who may not have good knowledge of abstract algebra. I recommend we rename the same interfaces to something more descriptive: `IAddable`, `IMultiplicable`, `ISubtractable`. We don't need to have `CommutativeGroup` for now.

Another big problem with the approach right now is there is no distinction on scalar types and composite types. How do I write a generic function constrained on builtin scalar types only? This is necessary to expose intrinsics or defining derived functions based on the intrinsics. We sometimes need to define generic functions on all integer scalar types, and sometimes need to define generic functions on all floating point scalar, vector or matrix types. How can we express that?

docs/proposals/015-abstract-algebra-interface-hierarchy.md
```diff
@@ -0,0 +1,347 @@
...
+
+// CommutativeGroup ensures addition order doesn't matter
+// Examples: Vectors, Colors (in linear space)
+interface CommutativeGroup : Group
+{
+    // Enforces a + b == b + a
+};
+
+// Semiring adds multiplication with identity to Monoid
+// Examples: Matrices, Quaternions
```
aleino-nv, line 95: Matrices and quaternions are full rings, aren't they? Of course that means they're also semirings but it's more interesting to list examples here that are not full rings. I guess the non-negative integers would be an example. (?)

docs/proposals/015-abstract-algebra-interface-hierarchy.md
```diff
@@ -0,0 +1,347 @@
...
+interface Monoid : Semigroup
+{
+    static T zero();  // Additive identity
+};
+
+// Group adds inverse elements to Monoid
+// Examples: Vectors under addition, Matrices under addition
+interface Group : Monoid
+{
+    T operator-();  // Additive inverse
```
aleino-nv, line 84: If the group is "additive" meaning the group operation is written + and inverse is written -, that always also implies that the group is commutative. This is sometimes called an "additive group".

Group in the most general sense usually uses some kind of operator that looks more like * (or when written, just concatenation: a*b = ab).

Should we call this AdditiveGroup and subclass it from a more general group interface that just has some interface (maybe *) for a more generic group operation? The more general Group would then also need an interface to get the inverse of an element.

expipiplus1:
Closing in favor of https://github.com/shader-slang/spec/pull/7

Issue #5573, opened 2024-11-15
Update SPIRV submodules to include "Update SPV_AMDX_shader_enqueue"


jkwak-work:
One of our customers requested to update the SPIR-V submodules to include [Update SPV_AMDX_shader_enqueue](https://github.com/KhronosGroup/SPIRV-Headers/commit/07ddb1c0f1ffa929262d4568481a692bb0fb1535)

PR #5815, opened 2024-12-10
Update SPIRV submodules

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/5573

One of tests started failing with the latest SPIR-V submodules:
`tests/gpu-feature/texture/query/footprint/nv-shader-texture-footprint.slang`

`[ForceInline]` is added to GLSL texture functions in order to inline `%true` and `%false`.
Without it, the value was indirectly passed down via a function parameter, which broke the existing test.

Also the test is modified to use `-DAG`, because the order unpredictably changed for `Grad` variants due to additional `[ForceInline]` marks for Gradient functions.


csyonghe:
The test still has validation error on linux and macos?

jkwak-work:
I will work on the CI failures on Linux and MacOS

jkwak-work:
It turned out that the Windows build is also not working properly.
The cause of the failure is that the test starts having a Vulkan Validation Error, which was introduced three month ago in SPIRV-Tools repo:
https://github.com/KhronosGroup/SPIRV-Tools/commit/7ba72f14f2f59a6f65465e5fb3e849dc81ae462d
The description of the commit is here,
```
spirv-val: Disallow stores according to VUID 06924 (https://github.com/KhronosGroup/SPIRV-Tools/pull/5368)
Ensure that the validator rejects stores to objects of types
`OpTypeImage`, `OpTypeSampler`, `OpTypeSampledImage`,
`OpTypeAccelerationStructureKHR`, and arrays of these types, according
to `VUID-StandaloneSpirv-OpTypeImage-06924`.

Guard the check behind the before_hlsl_legalization option, as
sometimes we may have temporaries or local variables that are expected
to get optimized away.
```

I will file an issue for this.

jkwak-work:
I applied Ellie's fix and all CI tests are now passing.

csyonghe (approved):

PR #5819, opened 2024-12-10
A small mistake.

HappyLi3:
The 'samplerInfo.magFilter' should match the 'desc.magFilter'.


csyonghe:
@skallweitNV This needs to be fixed in slang-rhi as well.

PR #5823, opened 2024-12-10
Fix attribute reflection.

### LINKED ISSUE #5811: Access to COM Attribute GUID via Reflection API
For a `TypeReflection` object, there are only methods like `getUserAttributeByIndex` to access user defined attributes. However, for purposes such as generating COM binding from slang source, it is necessary to extract the GUID from the COM interface declaration.

csyonghe:
Closes #5811.


expipiplus1 (changes requested):
Shouldn't this be a breaking change?

jkwak-work (approved):
Looks good to me.

PR #5839, opened 2024-12-11
Add an example for using the reflection API

tangent-vector:
The example program is meant to accompany a document that goes into more detail about the mental model behind the reflection API and the way this program drives it. Ideally this program can land before the document goes live, and then the document can be published with a link to the example. After that, the example could be updated to include links into the live document.

Along with adding the example program, this change also adds some convenience functions to the reflection API to avoid cases where the program would otherwise need to cast between `slang::ParameterCategory` and `SlangParameterCategory`.


csyonghe (approved):

csyonghe (approved):

csyonghe:
The example code doesn't build with clang.

tangent-vector:
I'll get on top of the fixes today.

csyonghe (approved):

PR #5846, opened 2024-12-12
use Release over RelWithDebInfo for ci

expipiplus1:
This will necessetate changing the "required tests" setting on the GitHub repo


csyonghe:
Release is -O3 and RelWithDebInfo is -O2. This may technically lead to tests passing in CI but failing with prebuilt release. For example, -O3 will optimize using assume but -O2 will not. We need to make sure the prebuilt release is using the exact same config as CI.

expipiplus1:
that's what this PR is doing, releases are already switched to Release, this is correcting an oversight not making CI do this already

csyonghe (approved):

csyonghe:
I see. Thanks for the clarification!

PR #5847, opened 2024-12-12
Bit extract

kaizhangNV:
The only commit in this PR is this:

[correct wgsl code gen](https://github.com/shader-slang/slang/pull/5847/commits/094b0b8a779e90c26bb017af8bafa48df53d6e46)


kaizhangNV:
The only change of this PR is this commit:
[correct wgsl code gen](https://github.com/shader-slang/slang/pull/5847/commits/094b0b8a779e90c26bb017af8bafa48df53d6e46)

csyonghe (approved):

source/slang/slang-emit-c-like.cpp
```diff
@@ -3906,13 +3906,20 @@ void CLikeSourceEmitter::emitVecNOrScalar(
         // In other languages, we can output the Slang vector type directly
         else
         {
-            emitType(vectorType);
+            if (getTarget() == CodeGenTarget::WGSL)
```
csyonghe, line 3909: why is this needed? Shouldn't `emitType` be a virtual call into `WGSLSourceEmitter::emitTypeImpl` or something that handles the emit of vectorType properly?

source/slang/slang-emit-c-like.cpp
```diff
@@ -3906,13 +3906,20 @@ void CLikeSourceEmitter::emitVecNOrScalar(
...
+            }
+            else
+                emitType(vectorType);
         }
 
         m_writer->emit("(");
         for (int i = 0; i < N; ++i)
         {
-            emitType(elementType);
+            (getTarget() == CodeGenTarget::WGSL) ? m_writer->emit("u32") : emitType(elementType);
```
csyonghe, line 3922: same here, why is `emitType` not correct?

kaizhangNV: This is restriction of WGSL.

Since this function is used to emit the number of the shifted bits, and WGSL requires the type has to be 'u32', so we will always have to convert the type to `u32`.  

emitType(elementType); could also generate `i32`, which will throw an wgsl compile error.

kaizhangNV: To be honest, I don't like the current implementation as well. I think we should refactor the code to handle wgsl separately, and make it less hacky.

csyonghe: In that case we need to define the parameter type as uint in the core module, and emit the type of the index operand here.

kaizhangNV: As long as we can change the intrinsic signature, I think we should be good here.
Let me try that to see if there is other issues.

kaizhangNV:
In the latest change, I update the function signature to use `uint` as the type for the `offset` and `bit` parameters.
Update the test correspondingly. Now we can get rid of most of the special handling of wgsl backend.

csyonghe (approved):

PR #5848, opened 2024-12-12
Fix prebuilt dependencies filenames on non-windows platforms

expipiplus1:
CMAKE_SHARED_LIBRARY_PREFIX is still `lib` when compiling on linux, but the deps don't have that in them


jkwak-work (approved):
Looks good to me

Issue #5820, opened 2024-12-10
Investigate slang binary size increase


expipiplus1:
Yong noted that there has been an unexpected increase in slang binary size recently. We should discover the cause and rectify it if possible/desirable

PR #5850, opened 2024-12-12
Reduce export size with gcc

expipiplus1:
Hide some more symbols

affects https://github.com/shader-slang/slang/issues/5820 slightly


jkwak-work (approved):
Looks good to me

Issue #5843, opened 2024-12-12
Allow cross compiling via mingw?


Parad0x84:
As far as I understand, currently I can't cross compile from linux to windows using mingw, since slang code uses macros like MSVC_VER which is not defined by mingw.
Are there any plans to support cross compiling?

As far as I understand solution is as easy as converting this `#ifdef MSVC_VER` to this `#if defined(MSVC_VER) || defined(__MINGW32__ ) || defined(__MINGW64__ )`

expipiplus1:
I made a change yesterday which should fix that.

If I disable most optional features (DXIL (header confuses msvc and windows), tests (depends on gfx), NVAPI (MSVC only header), GFX (too old d3d12.h bundled with mingw)) then I can build libslang.dll and slangc.exe with mingw

![image](https://github.com/user-attachments/assets/2c07b22d-f3b5-49d5-8e5f-971fc03f1784)
 
I haven't tested this, so ymmv.

I would be happy to review any PRs making this better supported if you decide to experiment!

Parad0x84:
I pretty much disabled everything except `SLANG_ENABLE_SLANG_GLSLANG`. I'm using spirv-tools, but I tried disabling it too, still same result.

```cmake
set(SLANG_LIB_TYPE STATIC)
#set(SLANG_EMBED_CORE_MODULE OFF)
set(SLANG_ENABLE_CUDA OFF)
set(SLANG_ENABLE_OPTIX OFF)
set(SLANG_ENABLE_NVAPI OFF)
set(SLANG_ENABLE_XLIB OFF)
set(SLANG_ENABLE_AFTERMATH OFF)
set(SLANG_ENABLE_DX_ON_VK OFF)
set(SLANG_ENABLE_SLANG_RHI OFF)
set(SLANG_ENABLE_DXIL OFF)
set(SLANG_ENABLE_FULL_IR_VALIDATION OFF)
set(SLANG_ENABLE_IR_BREAK_ALLOC OFF)
set(SLANG_ENABLE_ASAN OFF)
set(SLANG_ENABLE_PREBUILT_BINARIES OFF)
set(SLANG_ENABLE_GFX OFF)
set(SLANG_ENABLE_SLANGD OFF)
set(SLANG_ENABLE_SLANGC OFF)
set(SLANG_ENABLE_SLANGRT OFF)
set(SLANG_ENABLE_TESTS OFF)
set(SLANG_ENABLE_EXAMPLES OFF)
[...]

PR #5851, opened 2024-12-12
Correctly distinguish between windows and MSVC

expipiplus1:
Partially sorts https://github.com/shader-slang/slang/issues/5843

### DIFF
--- a/source/compiler-core/windows/slang-win-visual-studio-util.cpp
+++ b/source/compiler-core/windows/slang-win-visual-studio-util.cpp
@@ -69,7 +69,7 @@ static SlangResult _readRegistryKey(const char* path, const char* keyName, Strin
 
 // Make easier to set up the array
 
-static DownstreamCompilerMatchVersion _makeVersion(int main)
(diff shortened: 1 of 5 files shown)


source/compiler-core/windows/slang-win-visual-studio-util.cpp
```diff
@@ -69,15 +69,15 @@ static SlangResult _readRegistryKey(const char* path, const char* keyName, Strin
 
 // Make easier to set up the array
 
-static DownstreamCompilerMatchVersion _makeVersion(int main)
+[[maybe_unused]] static DownstreamCompilerMatchVersion _makeVersion(int main)
```
jkwak-work, line 72: I think we have a macro define for maybe_unised.
Something like SLANG_MAYBE_UNUSED or something

jkwak-work (approved):
Looks good to me

source/compiler-core/windows/slang-win-visual-studio-util.cpp
expipiplus1: now that we're on c++17 we should avoid that macro

PR #5853, opened 2024-12-12
Add SLANG_ENABLE_DXIL

expipiplus1:
At the moment there's no way to easily disable this functionality 

(Happy that the linux support was so good until now that there was no call for this :D)


jkwak-work (approved):
Looks good to me

PR #5864, opened 2024-12-13
Fix: Missing cross compile symbols

Parad0x84:
It seems like mingw-w64 doesn't define some symbols like `fread_s` and `_wstat32` when doing cross compile https://github.com/shader-slang/slang/issues/5843#issuecomment-2540549691

To clarify, I still encounter other errors because some tools required during the build process are being built for the target
 ```
system, making them unable to run on the host system 
[ 41%] Built target slang-embed
[ 42%] Generating slang-torch-prelude.h.cpp
/bin/sh: 1: slang-embed: not found
```

Full commands I used to build (from a fresh install of ubuntu 24.04)
```
sudo apt install mingw-w64
sudo snap install cmake --classic
git clone --recurse-submodules https://github.com/Parad0x84/slang.git cd slang
cmake -B build/ -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=./cmake/cross-windows-x86_64.cmake cmake --build build/ --config Release
```

### CODE REVIEW

File: cmake/cross-windows-x86_64.cmake [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+set(CMAKE_SYSTEM_VERSION 10.0)
+set(CMAKE_SYSTEM_PROCESSOR x86_64)
+
+set(CMAKE_C_COMPILER x86_64-w64-mingw32-gcc)
+set(CMAKE_CXX_COMPILER x86_64-w64-mingw32-g++)
+set(CMAKE_RC_COMPILER x86_64-w64-mingw32-windres)
+set(CMAKE_FIND_ROOT_PATH /usr/x86_64-w64-mingw32)
+
+# Slang specific
+set(SLANG_LIB_TYPE STATIC)
```
csyonghe, line 12: Why are we setting static library as default?

Parad0x84: The toolchain file is only intended for an easy way to reproduce same build, not intended for merging. You can ignore that file. This is Work in Progress anyways. I'm not expecting to merge as is. Open to ideas

Parad0x84:
Fixed formatting issue.
By the way, I didn't know I should label PR about breaking or not. This change shouldn't be breaking unless you are using mingw on windows and targeting windows. Shouldn't be a common use case (I'm open to ideas on that)

csyonghe (approved):
Looks good to me!

expipiplus1 (changes requested):
Would you be able to either remove the toolchain file, or mention it in a comment and in building.md that it's intended as an example config and isn't officially supported.

FWIW I didn't need to make the other changes when I built with mingw, but I can look at that when I next go down that path, so if this helps you then happy to merge

Parad0x84:
> Would you be able to either remove the toolchain file, or mention it in a comment and in building.md that it's intended as an example config and isn't officially supported.
> 
> FWIW I didn't need to make the other changes when I built with mingw, but I can look at that when I next go down that path, so if this helps you then happy to merge

Please read this commit message, if you didn't yet: https://github.com/Parad0x84/slang/commit/f25f9ea35afbdf5a58eeedc3c659a172397eb5a4
Are you sure you aren't getting same missing symbols issue when cross compiling from linux to windows?

This PR is not useful as-is. Currently slang-embed can't run at build time, since when cross compiling, it gets built for windows (target) and not for linux (host). If we decide on how to move forward solving this issue, I'm happy to implement it (if I can).
Please take a look at this: https://github.com/shader-slang/slang/issues/5843#issuecomment-2543891027

Parad0x84:
Since nobody replied, I'm assuming we don't wanna move forward with solving this issue

Issue #5829, opened 2024-12-10
Auto-diff crashes when attempting to dynamic-dispatch on differentiable associated types


saipraveenb25:
Example code that causes this crash:

```csharp
// Test calling differentiable function through dynamic dispatch.

//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type

//TEST_INPUT:ubuffer(data=[0 0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<float> outputBuffer;

//TEST_INPUT: set g_materials = new StructuredBuffer<MaterialDataBlob>[new MaterialDataBlob{new MaterialHeader{[0, 0, 0, 0]}, new MaterialPayload{[1.0, 1.2, 0.3, 0.5]}}];
RWStructuredBuffer<MaterialDataBlob> g_materials;

public struct ShadingInput
{
    public float scale;
}

struct MaterialHeader
{
    uint4 header;
};
struct MaterialPayload
{
    float4 data;
};
struct MaterialDataBlob
{
    MaterialHeader header;   // 16B
    MaterialPayload payload; // 16B
};

interface IMaterial : IDifferentiable
{
    associatedtype MaterialInstance : IMaterialInstance;

    [Differentiable]
    MaterialInstance setupMaterialInstance( ShadingInput input );
}

interface IMaterialInstance : IDifferentiable
{
    [Differentiable]
    float eval( float x );
}

[open]
struct MaterialBase
{
    MaterialHeader header;
}

[BackwardDerivative(getMaterial_bwd)]
IMaterial getMaterial(int id)
{
    return createDynamicObject<IMaterial, MaterialDataBlob>(id, g_materials[id]);
}

void getMaterial_bwd(int id, IDifferentiable d)
{
    // Something random
    outputBuffer[id] = 2.f;
}

struct Material1: MaterialBase, IMaterial
{
    typedef MaterialInstance1 MaterialInstance;

    float a;
    float b;
    float c;

    [Differentiable]
    MaterialInstance1 setupMaterialInstance( ShadingInput input )
    {
        MaterialInstance1 instance;
        instance.a = a * input.scale;
        instance.b = b * input.scale;
        instance.c = c * input.scale;
        return instance;
    }

}
struct MaterialInstance1: IMaterialInstance
{
    float a;
    float b;
    float c;

[...]

PR #5866, opened 2024-12-13
[Auto-diff] Overhaul auto-diff type tracking + Overhaul dynamic dispatch for differentiable functions

saipraveenb25:
This is a large overhaul PR containing several fixes and overhauls:

## **Overhauled differentiable type lowering**: 
- `IRDifferentiableTypeDictionaryDecoration` is deprecated in favor of `IRDifferentiableTypeAnnotation` instructions that can appear anywhere in the code (module/generic/function scopes) depending on the scope of the type. This makes it much simpler to lower differentiable run-time types. 
- `IRDifferentiableTypeAnnotation(type, witness)` instructions are generated by registering a 'type lowering hook' callback during IR lowering that checks the type against the list of differentiable types & generates an annotation for that type. 
- `IRDifferentiableTypeAnnotation` is hoistable, so it is emitted at the same parent as the type itself.
- `DifferentiableTypeConformanceContext`, used by the auto-diff passes, scans the function being differentiated, as well as all parent scopes, to build the full list of differentiable types in the context of the function.

## **Overhaul differentiation of dynamic types**
- Previously, a limited version of dynamic-dispatch lowering was performed before auto-diff to make things simpler for auto-diff, but this meant that advanced uses (associated types) were not supported properly. 
- The overhauled system moves all dynamic-dispatch lowering to after the auto-diff passes. 
- Under the new system, differential pair of an interface type`IInterface` is a pair of two interface types:`IInterface` & `IDifferentiable`, while the differential pairs of existential types & associated types are replaced with lookups of a _synthesized_ differential pair type.
- E.g. for a system like this: 
```csharp
interface IFoo : IDifferentiable
{
     associatedtype Bar : IDifferentiable;
     [Differentiable] Bar doThing(This);
};

struct A : IFoo 
{ 
     float x; 
     /* ... */
};
```
the derivative of `doThing()` is represented using associated types that are synthesized to represent the pairs. 
In the above example, `IFoo` is automatically extended to add the new type requirements (in addition to the derivative method requirements, which is already an existing feature)

```csharp
interface IFoo : IDifferentiable
{
     associatedtype Bar : IDifferentiable;
     
     // Synthesized type requirements.
    associatedtype DiffPair_This;
    associatedtype DiffPair_Bar;

    // Synthesized makePair requirements.
    DiffPair_This This_makePair(This, This.DIfferential);
    DiffPair_Bar Bar_makePair(Bar, Bar.DIfferential);

    // Synthesized getPrimal requirements.
    This This_getPrimal(DiffPair_This);
    Bar Bar_getPrimal(DiffPair_Bar);
    
    // Synthesized getDiff requirements.
    This.Differential This_getDiff(DiffPair_This);
    Bar.Differential Bar_getDiff(DiffPair_Bar);
     
     float doThing(This, Bar);
     
     // Synthesized diff requrirements (logic for this is already present
     // before this PR.
     void doThing_bwd(inout DiffPair_This, inout DiffPair_Bar, float);
     DiffPair<float> doThing_fwd(DiffPair_This, DiffPair_Bar);
};
```

- After the interface is modified, all available concrete witness tables are extended with synthesized methods to meet these new requirements.

- Note that this type & method synthesis happens on the IR side during auto-diff pair lowering for now. 
During auto-diff, these types are still represented as `DifferentialPair<LookupWitness(table, assoc_type_key)>` though it is not a proper type.
- Ideally, the next step is to perform this on the AST side when creating the method requirements for the differentiable methods.

## Additional fixes
There are several other minor fixes to bugs that appeared once the auto-diff step was moved to after the dynamic-dispatch lowering step.
- `UseGraph`: During the primal availability step, primal instructions that are used in differential contexts are stored into an intermediate variables & loaded from later. Many instructions (such as `ExtractExistentialType`) cannot be stores, and so is considered a 'passthrough' inst that should be cloned in rather than stored. The previous system used a set of such 'use paths' called `UseChain`s to facilitate this, but it fails in cases where there is a common node in these paths. This PR introduces `UseGraph` to represent & clone a full graph of instruction between a primal inst and its differential uses.
- **Support for lowering differential pairs of type-packs**: We were previously lowering these as a struct of two type-packs, but this is not a valid use & will break specialization. We now lower these as type-packs of lowered pair types (with similar logic for value packs)

Fixes: #https://github.com/shader-slang/slang/issues/5829


source/slang/slang-emit.cpp
```diff
@@ -857,9 +868,20 @@ Result linkAndOptimizeIR(
...
+    eliminateDeadCode(irModule, deadCodeEliminationOptions);
+
+    // After auto-diff, we can perform more aggressive specialization with dynamic-dispatch
+    // lowering.
+    //
+    if (!codeGenContext->isSpecializationDisabled())
+    {
+        SpecializationOptions specOptions;
+        specOptions.lowerWitnessLookups = true;
+        specializeModule(targetProgram, irModule, codeGenContext->getSink(), specOptions);
```
csyonghe, line 883: Performing lowerWitnessLookups can open up new opportunities for specializations, so we need to run that specialization-optimization loop again.

This is really calling for cleaning up the specialization pass to be just a peephole optimization pass.

source/slang/slang-ir-autodiff-fwd.cpp
```diff
@@ -827,17 +860,72 @@ InstPair ForwardDiffTranscriber::transcribeCall(IRBuilder* builder, IRCall* orig
                 }
             }
         }
+
+        {
+            // --WORKAROUND--
```
csyonghe, line 865: Put all these in a function, and do early returns instead of nested if.

source/slang/slang-emit.cpp
saipraveenb25: I believe `specializeModule()` already runs the witness lowering in a loop with the rest of the specialization pass?

csyonghe (dismissed):
I am going to merge this PR in for now, but we need to address the comments in a follow up work.

csyonghe (approved):

Issue #5520, opened 2024-11-07
Integrate all examples in CI


csyonghe:
We get bug reports that our examples are crashing.
We should add `-test` commandline argument to all examples, and when that flag is step, we render 1 frame and exit with 0.

Then we can setup ci script to run all examples one by one with the -test flag, and fail the CI if any runs returns non 0 code.

aleino-nv:
@csyonghe Can you set priority on this issue?

aleino-nv:
Lots of samples are crashing when run in CI. The only OS where some samples complete successfully is MacOS.

On Linux and Windows, I believe there is probably something about the environment that prevents successful initialization of graphical applications, e.g. no window manager running, or `DISPLAY` environment variable not set.
I don't exactly know what kind of "session" or "shell" is being created to run CI.

As a first step, I'll see if I can reproduce some kind of crash/failure by simply logging on to a CI machine and running the examples there.

aleino-nv:
Ah, there is already a "headless test mode" for the example base framework.

I'll get that running in CI as a first step, and then see what it would take to make the app run with a full WM and whether that's worth the extra coverage on Linux and Windows. 
(At least in Linux I think it's usually pretty easy with a headless WM...)

aleino-nv:
For some reason the CI script is currently failing to run anything on Linux and Windows, even with headless mode (`--test-mode`).

I logged onto a test machine via SSH, it runs fine there:


    slangadmin@slangadmin-desktop:~/test/slang/build/Debug/bin$ ./triangle --test-mode
    [GraphicsLayer]: INFO: CreateDevice: Debug layer is enabled.
    [Slang]: WARNING: (0): warning 50011: Slang's SPIR-V backend only supports SPIR-V version 1.3 and later. Use `-emit-spirv-via-glsl` option to produce SPIR-V 1.0 through 1.2.

    [Slang]: WARNING: (0): warning 50011: Slang's SPIR-V backend only supports SPIR-V version 1.3 and later. Use `-emit-spirv-via-glsl` option to produce SPIR-V 1.0 through 1.2.

    callIdx: 0, entrypoint: 0, target: 0, hash: 1A5FD3EF8CB6418720C6F80C10FF516605714C76
    callIdx: 1, entrypoint: 1, target: 0, hash: 4479172016FA47274F6BE8E8483F124D4079D6C1
    slangadmin@slangadmin-desktop:~/test/slang/build/Debug/bin$ echo $?
    0

[...]

PR #5879, opened 2024-12-16
Add slang example tests to CI

aleino-nv:
- Reuse the existing `--test-mode` option to run samples in "headless mode" in CI.
- Add some missing error handling and make minor fixes to samples.

https://github.com/shader-slang/slang/issues/5520


.github/workflows/ci.yml
```diff
@@ -176,6 +176,107 @@ jobs:
               -expected-failure-list tests/expected-failure-github.txt \
               -expected-failure-list tests/expected-failure-record-replay-tests.txt
           fi
+      - name: Run Slang examples
```
kaizhangNV, line 179: can we put this into a bash script file to simplify the yml file?

csyonghe:
@expipiplus1 can you also review this? I am not an expert on bash scripts. I want to make sure the script is clean.

aleino-nv:
> @expipiplus1 can you also review this? I am not an expert on bash scripts. I want to make sure the script is clean. I feel like there could be a more structural way to define which tests to run on which platform instead of using duplicated for loops and adhoc ifelse commands.

I initially used associative arrays to record the skips, but it seems the bash on our MacOS build machines is not new enough to support those.

.github/workflows/ci.yml
aleino-nv: Should be possible, I will try it out. I think I'll have to evaluate stuff like `"${{ matrix.os }}"` within the yaml file, but IMO it's cleaner to hoist that stuff out to the top level anyway, and just pass it as arguments to the script.

csyonghe:
Perhaps we can define the test list in a separate file? I am open to other ideas as well.

@kaizhangNV @expipiplus1 any thoughts?

.github/workflows/ci.yml
csyonghe: Are there any examples that actually runs on macOS? If not I am OK with skipping macOS for now and just run the tests on Windows where we can use the most number of APIs.

aleino-nv: > Are there any examples that actually runs on macOS? If not I am OK with skipping macOS for now and just run the tests on Windows where we can use the most number of APIs.

I don't have a Mac to test on, but several of the examples seem to "pass" in CI.

In fact, even when I didn't run them headless with the `--test-mode` flag, several examples passed CI on MacOS.
(Initially I had implemented a `--single-frame-test-mode` flag that actually created a window and called `exit(0)` right after the first frame was presented. However that complicates CI because it would require a window system environment, so I just reused the existing headless rendering test mode instead.)

aleino-nv: Do you mean we should disable MacOS and/or Linux tests that are currently passing?

aleino-nv:
> Perhaps we can define the test list in a separate file? I am open to other ideas as well.
> 
> @kaizhangNV @expipiplus1 any thoughts?

I have a patch in progress for this. It's currently here https://github.com/shader-slang/slang/pull/5903 because I'm still testing it in CI. (Don't want to push a million times to this branch now that review has started.)

expipiplus1 (approved):

aleino-nv:
I was going to make sure all review comments were addressed by all reviewers.

I didn't realize that auto-merge would go through unless all reviewers approved, sorry!

Issue #4495, opened 2024-06-27
Add Support for Sparse/Tiled Texture Operations (HLSL, GLSL, SPIRV)


ArielG-NV:
Problem:
Currently Slang does not support sparse/tiled texture operations.

Solution:
Support for [Gather operations with a status](https://github.com/shader-slang/slang/pull/4490) are being implemented, but we still require to implement more functions to complete the support for `hlsl.meta.slang`
1. `Sample()` with `status`
2. `Load()` with `status` [[Spec](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/t2d-load-float-int-uint-)]
3. `CheckAccessFullyMapped()` to fetch the meaning of `status`

GLSL implements the operations with: [GL_ARB_sparse_texture2](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_sparse_texture2.txt)

SPIRV implements the operations with: OpCapability [SparseResidency](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html)

bmillsNV:
Should be supported w/several enhancements over the last year.

PR #5901, opened 2024-12-18
Add HLSL texture sample with status intrinsics

fairywreath:
This is part of #4495.

Add HLSL intrinsics for texture sample functions that accept a status out parameter. These are used for sparse textures and described in the SM 5.0 [specs](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/samplelevel-s-float-float-int-uint-). This change only supports HLSL targets since SPIRV/GLSL sparse images are still not supported. Also added more cases for the HLSL texture intrinsics test to use all variants of Sample*.


csyonghe (dismissed):

csyonghe:
The test failure needs to be addressed before we can land this PR.

fairywreath:
Turns out some of the additional test cases I added for other Sample* variants require derivatives and the test runs on a compute shader(it is a compute compare ex test case). We can either mangle the gfx test runner and eg. enable sm6.6 and vk extensions or just make a new test with a fragment shader. I reverted these additional cases for now and they can be added along with tests for spirv sparse textures once they are implemented.

csyonghe (approved):
Looks good.

PR #5902, opened 2024-12-18
command-line-slangc-reference.md: fixed <a> HTML tag

stanoddly:
Fixed the broken tag introduced by #5507.

Also, the file seemed to have UTF-16LE encoding unlike other files I checked. Since I was lazy and did the changes through GitHub UI, it automatically converted the changes to UTF-8. Let me know if this is a problem, but UTF-16LE sounds bit unintentional than a "feature" anyway.


csyonghe:
This file is generated by running `slangc -h`, so we shouldn't be changing this file directly. Modifications should be made to `source\slang\slang-options.cpp` or to the source code that produces this md file.

PR #5907, opened 2024-12-18
Support explicit `[vk::location(n)]` binding on metal/wgsl.

### LINKED ISSUE #5883: MSL Backend does not respect the location used in [[vk::location(x)]]
Filed on behalf of a client developer:

The following slang code:
```
slang
struct VSInput
{
    [[vk::location(0)]] float3 inPositions;
#ifdef HAS_NORMAL
    [[vk::location(1)]] float3 inNormals;
#endif
#ifdef HAS_TANGENT
    [[vk::location(2)]] float4 inTangents;
#endif
#ifdef HAS_TEXCOORD0
    [[vk::location(3)]] float2 inTexCoords0;
#endif
};
```
Would translate with HAS_NORMAL and HAS_TEXCOORD0 defined to:
```
msl
struct VSInput_0
{
    float3 inPositions_0;
    float3 inNormals_0;
    float2 inTexCoords0_0;
};
```
We wanted to use fixed attribute locations and expected to have for the `inTexCoords0` attribute the [[attribute(3)]] where the msl code would need to bind this attribute at [[attribute(2)]], leading to a crash in the metal validation layers or an invalid output.
We need to rely on the reflection to have the location where we should be confident on the attribute to use
You already have this interop feature for `[[vk::constant_id]]` maybe the translation from `[[vk::location(x)]]` to `[[attribute(x)]]` could be added as well?

csyonghe:
Closes #5883.

### DIFF
--- a/docs/user-guide/a2-02-metal-target-specific.md
+++ b/docs/user-guide/a2-02-metal-target-specific.md
@@ -276,6 +276,8 @@ Since metal does not differentiate a constant buffer, a shader resource (read-on
 
 `spaceN` specifiers inside `register` semantics are ignored.
 
+The `[vk::location(N)]` attributes on stage input/output parameters are respected.
+
 ## Specialization Constants
 
 Specialization constants declared with the `[SpecializationConstant]` or `[vk::constant_id]` attribute will be translated into a `function_constant` when generating Metal source.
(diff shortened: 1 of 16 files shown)


jkwak-work (approved):
Looks good to me.

PR #5909, opened 2024-12-19
Don't lower float matrix types when emitting SPIRV.

csyonghe:


csyonghe:
Actually, we can't do this for many nuanced reasons.

SPIRV only allows the ColMajor or RowMajor or the MatrixStride decorations to be placed on a struct member. This means that we will need to wrap matrix types in struct types in many situations no matter what. For example, `float4x4*` will always need to be translated into a `struct_wrapper_float4x4*` in order to decorate the inner matrix. Matrices of integer or bool types will also need to translated no matter what.

If we create `struct struct_wrapper_float4x4 { float4x4 inner; };`, we will also need to be careful not to fall into infinite loops trying to recursively lower the type of `inner`. Our lower to storage type pass isn't architected to easily allow generating conditional wrappers like this.

PR #5918, opened 2024-12-19
Add base vertex and base instance system values

fairywreath:
Adds base vertex and base instance system value semantics that were introduced in [SM 6.8](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_ShaderModel6_8.html#extended-command-information). Parallel implementations already exist on GLSL, SPIRV and Metal but not WGSL.


csyonghe (dismissed):
looks great!

csyonghe (approved):

PR #5919, opened 2024-12-19
Lower varying parameters as pointers instead of SSA values.

csyonghe:
This PR changes the way we lower entry point varying parameters so that they are emitted as a `ConstRef<T>` typed pointer value instead of a SSA value.

This is required to properly implement things like `GetAttributeAtVertex` or `EvaluateAttributeAtCentroid`, where the argument is expected to be a reference to a vertex attribute declaration.

Previously this is implemented as a hack assuming the `x` in `GetAttributeAtVertex(x)` is a `load(attribute_var)`, which isn't safe or correct.

By lowering all varying parameters as pointers, we can properly implement `GetAttributeAtVertex(x)` assuming `x` is a pointer in the `Input` address space.

This also allowed us to support using types that needs storage lowering for vertex attributes, such as column_major matrix types or bool types, by reusing the existing `lowerBufferElementTypeToStorageType` pass, which requires any such type to be used via a buffer or a pointer. This lowering is required to support `Optional<T>` in vertex attributes.


source/slang/slang-ir-fix-entrypoint-callsite.cpp
```diff
@@ -0,0 +1,101 @@
...
+            case kIROp_UserExternDecoration:
+            case kIROp_HLSLExportDecoration:
+            case kIROp_EntryPointDecoration:
+            case kIROp_LayoutDecoration:
+            case kIROp_NumThreadsDecoration:
+            case kIROp_ImportDecoration:
+            case kIROp_ExternCDecoration:
+            case kIROp_ExternCppDecoration:
+                decorsToRemove.add(decor);
+                break;
```
kaizhangNV, line 44: need a default?

kaizhangNV (approved):
lgtm overall.

Issue #5656, opened 2024-11-24
[request] HTML5 example.


0wwafa:
It would be great to have a simple example on how to load and apply a SLANG shader (for example a CRT emulation) to a DIV/CANVAS in HTML5.

aleino-nv:
Starting on a bear-bones example now. 
I'll do a simple hello-world type of application first using the WebGPU shading language (WGSL) and then just swapping it out for a Slang shader by calling slangc as an offline step.

aleino-nv:
Here is my first draft of an example: https://github.com/aleino-nv/slang/tree/aleino/wgpu-sample/examples/wgpu-html5

The corresponding pull request is here: https://github.com/shader-slang/slang/pull/5923

aleino-nv:
It was also requested that I add a simple example showing how to use the WASM build of the Slang API to compile a Slang shader to WGPU in the browser. I'm working on that next.

aleino-nv:
Now there is a PR (https://github.com/shader-slang/slang/pull/5996) showing how to use the Slang API from JavaScript (i.e. using the WASM build of Slang) to compile WGSL code in the browser.

aleino-nv:
@0wwafa I hope that the added examples are helpful. Please let me know if you have any questions.

0wwafa:
@aleino-nv the example is ok, thanks, but I was thinking more like: apply a crt shader to a canvas element.

aleino-nv:
> @aleino-nv the example is ok, thanks, but I was thinking more like: apply a crt shader to a canvas element.

I think the Slang-related part of the example I added would be about the same for such an example.

I haven't tried it, but I think you would use https://developer.mozilla.org/en-US/docs/Web/API/GPUCanvasContext/getCurrentTexture to get a texture which you read from/write to.

PR #5923, opened 2024-12-20
Add a basic WebGPU example

aleino-nv:
- Add a simple example of using WebGPU and Slang together

https://github.com/shader-slang/slang/issues/5656

### DIFF
--- a/examples/wgpu-html5/example.js
+++ b/examples/wgpu-html5/example.js
@@ -81,17 +81,9 @@ let Example = {
             device.queue.submit([commandBuffer]);
         }
 
-        const fileNames = ["shader.vertex.wgsl", "shader.fragment.wgsl"];
-        Promise.all(fileNames.map(n => fetch(n))).then(responses => {
-            for(var i = 0; i < fileNames.length; i++) {
-                if(!responses[i].ok) {
-                    console.error("Failed to fetch '" + fileNames[i] + "'");
-                    return;
-                }
-            }
-            Promise.all(responses.map(r => r.text())).then(codes => {
-                render({vertex: codes[0], fragment: codes[1]});
-            });
+        render({
+            vertex: await fetch("shader.vertex.wgsl").then(r => r.text()),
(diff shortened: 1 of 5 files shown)


examples/wgpu-html5/example.js
```diff
@@ -0,0 +1,115 @@
+"use strict";
+
+let Example = {
```
csyonghe, line 3: Use the simpler fetch API to read files.

examples/wgpu-html5/example.js
```diff
@@ -0,0 +1,115 @@
...
+            pass.setPipeline(pipeline);
+            const vertexBufferSlot = 0;
+            pass.setVertexBuffer(vertexBufferSlot, vertexBuffer);
+            pass.draw(vertices.length / 2);
+            pass.end();
+            const commandBuffer = encoder.finish();
+            device.queue.submit([commandBuffer]);
+        }
+
+        Promise.all([
```
csyonghe, line 105: Use async/await instead of explicit promise.

examples/wgpu-html5/example.js
aleino-nv: Done. Thanks, I wasn't aware of the `fetch` API.

examples/wgpu-html5/example.js
aleino-nv: You mean like this?

    var vertexShaderResponse = await fetch("shader.vertex.wgsl");
    var vertexShaderCode = await vertexShaderResponse.text();
    var fragmentShaderResponse = await fetch("shader.fragment.wgsl");
    var fragmentShaderCode = await fragmentShaderResponse.text();

That would load the shaders serially one at a time, which is why I opted for `Promise.all` which I believe allows you to overlap the fetches.
Not a big deal for this app though, so I'll change it if you want.

aleino-nv: Done. The last commit switches to serial `await`.

csyonghe (approved):

PR #5932, opened 2024-12-21
Fix parsing GLSL SSBO arrays / bindless descriptors

### LINKED ISSUE #5788: glsl bindless descriptors do not compile
When using glsl compatibility mode, bindless descriptor set usage does not compile. For example

```
struct RenderObject
{
    mat4 Model; 
    mat4 Position; 
    mat4 Rotation; 
    mat4 Scale; 
    uint MaterialIndex;
};


layout(set = 0, binding = 0, std430) buffer RenderObjects 
{
    RenderObject Objects[]; 
} renderObjects[]; 

struct Material
{
    uint Diffuse;
}; 

layout(set = 0, binding = 0, std430) buffer Materials
{
    Material Mats[];
} materials[];

```
will return the error 

`TexturedMesh.vert(55): error 39999: declaration does not declare anything
} renderObjects[];
  ^~~~~~~~~~~~~
TexturedMesh.vert(65): error 39999: declaration does not declare anything
} materials[];
`

juliusikkala:
Should fix #5788. Unfortunately, it doesn't _quite_ work yet, as there's some issue with indexing into these. The entries aren't interpreted as l-values for some reason. As I'm completely unfamiliar with the codebase, I'm very lost on how to debug that.

Example source:
```
#version 460

layout(local_size_x = 1) in;
layout(binding = 0) buffer buf
{
    float data;
} buffers[];

void main()
{
    buffers[0].data = buffers[1].data;
}
```

slangc output with the changes from this PR:
```
test.comp(11): error 30011: left of '=' is not an l-value.
    buffers[0].data = buffers[1].data;
```
If the contents of `main` are commented out or the l-value diagnostic is disabled, slangc silently finishes and generates appropriate-looking GLSL and HLSL output from this source.

Any hints are appreciated!


juliusikkala:
I fixed that l-value issue by making isMutableGLSLBufferBlockVarExpr work correctly for SSBO arrays as well. As such, I think the PR is now ready.

source/slang/slang-parser.cpp
```diff
@@ -3360,6 +3360,18 @@ static Decl* ParseBufferBlockDecl(
         reflectionNameModifier->nameAndLoc = bufferVarDecl->nameAndLoc;
         parser->ReadToken(TokenType::Semicolon);
     }
+    else if (parser->options.allowGLSLInput &&
+             parser->LookAheadToken(TokenType::Identifier) &&
+             parser->LookAheadToken(TokenType::LBracket, 1) &&
+             parser->LookAheadToken(TokenType::RBracket, 2) &&
```
csyonghe, line 3366: I think the lookahead should stop at "[". We shouldn't need to ensure it is followed by "[];", as sized arrays should also be allowed.

csyonghe: Since `parseBracketTypeSuffix` already looks for '[', you can just call it unconditionally.

csyonghe: The logic here is to allow structs and other block declarations to end without ";".

If we want the disambiguation logic to be correct, we need to skip until we find the matching `]` and see if it is followed by ';'.

But since we are in GLSL compatbiliity mode (`parser->options.allowGLSLInput`), it is fine to just look for '[' and be loose about not allowing missing semicolons here.

juliusikkala: Okay, thanks for the review. Would it be better to just skip the `parser->ReadToken(TokenType::Semicolon)`? I left it in there for now.

csyonghe (commented):
Looks good to me. Thank you for contributing the fix!

csyonghe (dismissed):

csyonghe (approved):

PR #5498, opened 2024-11-05
Legalize the Entry-point for WGSL

jkwak-work:
Closes #5486

The return type of the entry-point needs to be legalized when targeting WGSL.
This commit flattens the nested-structs of the return type and the input parameters of the entry-point.

Most of code is copied from the legalization code for Metal. The following functions are exactly same to the implementation for Metal or almost same.
- flattenInputParameters() : 136 lines
- reportUnsupportedSystemAttribute() : 7 lines
- ensureResultStructHasUserSemantic() : 46 lines
- struct MapStructToFlatStruct : 176 lines
- flattenNestedStructs() : 95 lines
- maybeFlattenNestedStructs() : 42 lines
- _replaceAllReturnInst() : 19 lines
- _returnNonOverlappingAttributeIndex() : 16 lines
- _replaceAttributeOfLayout() : 23 lines
- tryConvertValue() : 41 lines
- legalizeSystemValueParameters() : 11 lines

They need to be refactored to reduce the duplication later.

Issue #4657, opened 2024-07-16
Translate all metal entry point legalization (`LegalizeMetalEntryPointContext`) to use `EntryPointVaryingParamLegalizeContext`


ArielG-NV:
Problem:
Currently Metal has its own unique entry-point legalization logic. This should be refactored such that `LegalizeMetalEntryPointContext` runs using the `EntryPointVaryingParamLegalizeContext` entry-point legalization logic so we: 1. reduce code-reuse 2. CUDA/CPP benefit from all Metal improvements. This will require significant changes to `EntryPointVaryingParamLegalizeContext`.

fairywreath:
I would like to add a note that this refactoring will remove a lot of code duplication in legalization between Metal and WGSL.

PR #5937, opened 2024-12-24
Fix WGSL varying input attributes

### LINKED ISSUE #5933: wgsl @location attribute not emitted for entry-point varyings
I'm trying to slang-ify a very simple slang "hello triangle" example flor WebGPU, targeting WGSL. Here's the minimal slang shader:
```
struct VertexOutput {
    float4 clipPos : SV_POSITION;
    float3 color;
}

[shader("vertex")]
func vertexMain(float3 pos: POSITION, float3 color: COLOR0)->VertexOutput {
    VertexOutput out;
    out.clipPos = float4(pos, 1.0);
    out.color = color;
    return out;
}
```

I compile it on the 2024.17 release using this command:
```
.\bin\slangc.exe shaders.slang -target wgsl -entry vertexMain -o ..\src\vertex.wgsl
```

And then I get this invalid WGSL as the output:
```
struct VertexOutput_0
{
    @builtin(position) clipPos_0 : vec4<f32>,
    @location(0) color_0 : vec3<f32>,
};

@vertex
fn vertexMain( pos_0 : vec3<f32>,  color_1 : vec3<f32>) -> VertexOutput_0
{
    var out_0 : VertexOutput_0;
    out_0.clipPos_0 = vec4<f32>(pos_0, 1.0f);
    out_0.color_0 = color_1;
    return out_0;
}
```

This fails to compile in WGSL with the error:
```
      Entry point vertexMain at Vertex is invalid
        Argument 0 varying error
          Entry point arguments and return values must all have bindings
```

If I manually edit the WGSL to add  `@location` attributes, it's fine:
```
fn vertexMain( @location(0) pos_0 : vec3<f32>,  @location(1) color_1 : vec3<f32>) -> VertexOutput_0
```

---

Note that if I instead use an input structure instead of entry parameters, the wgsl `@location` attributes are correctly emitted. The problem is specific to entry parameters. Here's some slang using an input struct:
```
struct VertexInput {
    float3 pos : POSITION;
    float3 color : COLOR0;
}

struct VertexOutput {
    float4 clipPos : SV_POSITION;
    float3 color;
}

[shader("vertex")]
func vertexMain(VertexInput v)->VertexOutput {
    VertexOutput out;
    out.clipPos = float4(v.pos, 1.0);
    out.color = v.color;
    return out;
}
```

And here's the valid WGSL it produces. Note the `@location` attributes correctly applied to the members of `VertexInput_0`:
```
struct VertexOutput_0
{
    @builtin(position) clipPos_0 : vec4<f32>,
    @location(0) color_0 : vec3<f32>,
};

struct VertexInput_0
{
    @location(0) pos_0 : vec3<f32>,
    @location(1) color_1 : vec3<f32>,
};

@vertex
fn vertexMain( v_0 : VertexInput_0) -> VertexOutput_0
{
    var out_0 : VertexOutput_0;
    out_0.clipPos_0 = vec4<f32>(v_0.pos_0, 1.0f);
    out_0.color_0 = v_0.color_1;
    return out_0;
}
```

fairywreath:
Closes #5933.

This is a quick fix that copies code from Metal legalization (function packStageInParameters) to WGSL legalization, where individual varying inputs are placed into a global struct. Currently, WGSL does not properly add location qualifiers to the individual varying inputs.

This is not a permanent solution and will need to be refactored later. Issue #4657 tracks the refactoring effort of applying metal's legalization to the common entry point legalization context This change is also related to PR #5498, where code duplication already exists, further highlighting the need for refactoring and share the metal specific legalization logic to other backends.

[Draft because explicit vk locations are broken]


source/slang/slang-ir-wgsl-legalize.cpp
```diff
@@ -1348,6 +1356,195 @@ struct LegalizeWGSLEntryPointContext
         }
     }
 
+    void hoistEntryPointParameterFromStruct(EntryPointInfo entryPoint)
```
csyonghe, line 1359: Are these exactly the same as metal? if so we can move it to a shared file and call the function from both backends.

csyonghe (approved):

source/slang/slang-ir-wgsl-legalize.cpp
fairywreath: I initially copied them but I am getting crashes and test regressions, I will need more time to fix this. Another option is to just add new code to add semantic attributes to the parameters, but reusing code from Metal and refactoring(including refactoring of the existing) wgsl legalization code that is duplicated from Metal is still worth it IMO.

fairywreath:
I will create a fresh separate PR for this, changes from this PR are not valid anymore after the refactor.

PR #5952, opened 2024-12-29
Fix invalid uninitialized-use warning on hitAttributeEXT-decorated globals

### LINKED ISSUE #5947: Hit attributes in closest-hit shader triggering erroneous warning in GLSL compatibility mode
In GLSL compatibility mode, a global variable decorated with hitAttributeEXT is mistakenly reported as uninitialized. It isn't, as it is an input from the intersection shader.
 
```glsl
// test.rchit
#version 460
#extension GL_EXT_ray_tracing : require

layout(location = 0) rayPayloadInEXT vec2 payload;
hitAttributeEXT vec2 hit_attribs;

void main()
{
    payload = hit_attribs;
}
```

Output:
```
test.rchit(10): warning 41017: use of uninitialized global variable 'hit_attribs'
    payload = hit_attribs;
```

juliusikkala:
Fixes #5947. There's one issue with this though; `hitAttributeEXT` globals really are uninitialized in intersection shaders, so the warning would've been valid there. Still, they're rarer than any-hit and closest hit shaders, and this erroneous warning seems to trigger in basically any non-trivial ahit/chit GLSL shader with perfectly legitimate use.


csyonghe (approved):
LGTM, thank you!

PR #5953, opened 2024-12-29
Prevent constant folding for specialization constants

### LINKED ISSUE #5949: Erroneous warning about 0 iteration loops with specialization constants in GLSL compatibility mode
When the loop length is dependent on a specialization constant that is initialized to zero, slangc reports that the "loop runs for 0 iterations and will be removed. It shouldn't, since the value of the specialization constant isn't actually known at compile time. Luckily, even with -O2, it doesn't actually appear to get removed, judging from `-target glsl` output still emitting the loop anyway. So it's probably just the diagnostics.

Source for replicating:
```glsl
// test.comp
#version 460
layout(constant_id = 0) const uint A = 0;
layout(local_size_x = 1) in;

layout(binding = 0) buffer output_buffer
{
    float output[];
};

void main()
{
    for(uint i = 0; i < A; ++i)
    {
        output[i] = i;
    }
}
```
(The output buffer is only there to prevent the compiler from optimizing the loop away, it is not particularly necessary to replicate this bug.)

Output:
```
test.comp(13): warning 30505: the loop runs for 0 iterations and will be removed.
    for(uint i = 0; i < A; ++i)
    ^~~
```

juliusikkala:
Fixes #5949. This issue occurred in constant folding in `SemanticsStmtVisitor::tryInferLoopMaxIterations`; since specialization constants were being folded with their default values, there may have been other issues caused by the same problem as well.


csyonghe (approved):
LGTM, thank you for the fix!

PR #5955, opened 2024-12-29
Fix wrapper type member synthesis.

### LINKED ISSUE #5942: Support synthesized properties in specialization's `=` sugar syntax
According to the User's guide, the `=` sugar syntax in the type definition creates a wrapper on the underlying inner type (https://shader-slang.com/slang/user-guide/link-time-specialization.html). However, properties defined in the interface and implemented by a field are not properly wrapped.

I know that automatically synthesized `getter` and `setter` methods are not supported (based on the User's guide https://shader-slang.com/slang/user-guide/convenience-features.html#properties), but a code like this works pretty well:
```
public interface ITest {
    public property int val;
};

public struct TestImpl : ITest {
    public int val;
}
```

### Current result:

The problem occurs if I want to use this structure as part of shader specialization:
```
public interface ITest {
    public property int val;
};

public struct TestImpl : ITest {
    public int val;
}

public struct Test : ITest = TestImpl;

StructuredBuffer<Test> data;

[shader("compute")]
[numthreads(1,1,1)]
void computeMain()
{
    int val = data[0].val;
}
```

This source code is compiled with an error:
```
(0): error 30027: '$syn_property_val' is not a member of 'TestImpl'.
```

Tested on binaries from the latest release (v2024.17) and the build from the current master (7cecc51).

### Expected result:

The shader will be successfully compiled and the code above will load the `val` field into the local variable.

### LINKED ISSUE #5960: Support arguments with non-`in` direction in link-time wrapper types.
Functions that contain arguments with directions other than `in` (i.e. `out`, `inout`, `__ref`) lead to compilation errors about non-compliance l-value.

### Current result:
```
public interface ITest {
    public int testDir(inout int a);
};

public struct TestImpl : ITest {
    public int testDir(inout int a) {
        int oldA = a;
        a = 5;
        return a;
 }
}

public struct Test : ITest = TestImpl;

[shader("compute")]
[numthreads(1,1,1)]
void computeMain()
{
    Test data;
    int a = 516;
    int b = data.testDir(a);
}
```

Compilation of this source code produces an error:
```
(0): error 30047: argument passed to parameter '0' must be l-value.
```

Tested on binaries from the latest release (v2024.17), a build from the current master (89dd2b1), and just in case against PR #5955.

### Expected result:

The shader will be successfully compiled, `a` will have the value `5` and `b` will have the value `516` at the end of the function.

csyonghe:
Closes #5942.

Closes #5960.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -4394,14 +4394,14 @@ void SemanticsVisitor::addRequiredParamsToSynthesizedDecl(
     //
     for (auto paramDeclRef : getParameters(m_astBuilder, requirement))
     {
-        auto paramType = getType(m_astBuilder, paramDeclRef);
+        auto paramType = QualType(getType(m_astBuilder, paramDeclRef));
 
         // For each parameter of the requirement, we create a matching
         // parameter (same name and type) for the synthesized method.
         //
         auto synParamDecl = m_astBuilder->create<ParamDecl>();
         synParamDecl->nameAndLoc = paramDeclRef.getDecl()->nameAndLoc;
-        synParamDecl->type.type = paramType;
+        synParamDecl->type.type = paramType.type;
 
         // We need to add the parameter as a child declaration of
         // the method we are building.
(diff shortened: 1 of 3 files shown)


cezneluk:
Thank you for resolving all my issues really quickly.

If I understand it correctly, all changes from this PR are already in `master` due to the merge of PR #5964.

csyonghe:
You are right, we can close this PR now.

PR #5963, opened 2024-12-30
Implement specialization constant support in numthreads / local_size

juliusikkala:
Adds support for using SPIR-V specialization constants in `numthreads`:

```hlsl
[vk::constant_id(0)] const int WG_SIZE = 8;

[shader("compute")]
[numthreads(1, WG_SIZE, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
}
```

Also adds support in GLSL mode:
```glsl
#version 460

layout(constant_id = 0) const uint size = 8;
// The standard local_size_x_id is supported, but also local_size_x = SPECCONSTNAME.
layout(local_size_x_id = 0, local_size_y = size) in;

void main()
{
}
```

Similar to the `local_size_x_id` feature from GLSL, this allows users to generate shader permutations with different work group shapes / thread counts after the shader has already been compiled to SPIR-V.

Currently, this works with the SPIR-V (using `LocalSizeId` execution mode) and GLSL (using `local_size_{x,y,z}_id`) targets. I'm not sure if there even is anything that could be done for the other targets, other than to just present diagnostics about this feature not being supported there.


juliusikkala:
I've now tested this in a real Vulkan program in a couple of different contexts, and this feature seems to now work correctly.

csyonghe (commented):
Can you add a test for this feature?

source/slang/slang-ast-modifier.h
```diff
@@ -1038,9 +1043,12 @@ class NumThreadsAttribute : public Attribute
...
     // ordinary `args` list, rather than side data.
-    IntVal* x;
-    IntVal* y;
-    IntVal* z;
+    IntVal* extents[3];
+
+    // References to specialization constants, for defining the number of
+    // threads with them. If set, the corresponding axis is set to nullptr
+    // above.
+    DeclRef<VarDeclBase> specConstExtents[3];
```
csyonghe, line 1051: It is unfortunate that we need to use this side-band mechanism to handle the specialization constant case, but I also cannot think of any simpler way without complicating the `IntVal` hierarchy and the constant folding logic.

csyonghe (approved):
Looks good to me. I wish there is a cleaner way to add this feature, but what is implemented in this PR is probably the easiest path forward. If I understand the code correctly, this implementation won't allow something like `[numthreads(specConst+1, 1, 2)]`, right?

juliusikkala:
It currently doesn't allow doing math on the specialization constant, because I'm fairly sure that that can't be expressed in SPIR-V. I'll fix that compiler warning later today, I've been developing on Linux only so I guess MSVC errors are to be expected.

source/slang/slang-ast-modifier.h
juliusikkala: Yeah, I don't like it either. Especially the `axisIsSpecConstId` is nasty in the GLSL version, but since GLSL can legally reference undeclared specialization constant IDs, there's probably not much else that can be done.

juliusikkala:
Oh no... Sorry for the spam, I made a mistake.

csyonghe (approved):

fknfilewalker:
It seems that this implementation does not cover setting the wave size via constants
```csharp
[vk::constant_id(0)] const int CONST_VAR = 1;
[WaveSize(CONST_VAR, 1, 1)]
```

juliusikkala:
Is there a platform where you can set wave size / subgroup size with a specialization constant? With a quick scan through the SPIR-V spec, I couldn't find an instruction that would allow for that.

AFAIK numthreads() and NumThreads() should be equivalent (at least they map to the same AST node in `core.meta.slang`). If there's a difference, that's probably a bug. I don't have the time to look at this for a few days, but it'd help if you can specify in which way it doesn't work. Does it compile, does the SPIR-V pass validation, or does it silently just not do what it should during runtime? What does the SPIR-V disassembly look like for a simple program that fails?

fknfilewalker:
With `it does not work` I meant that setting the size via specialization constants does not compile for `NumTheads` as well as `WaveSize`. Explicitly specifying `NumTheads` or `WaveSize` does compiles, though `WaveSize` seems to be doing nothing. I am not sure what WaveSize is for...

juliusikkala:
Huh, odd, I can't replicate that. `NumThreads` is working just fine for me, identical to `numthreads`. Can you post a minimal code example of a full slang shader that doesn't compile with `NumThreads`? The one in the first message of this thread does work for me, with `slangc -target spirv test.slang -o test.spv`, and produces identical output for both the lowercase and uppercase `NumThreads` versions.

fknfilewalker:
Sorry I was wrong, `NumThreads` works. I mixed things up. It is just `WaveSize` that does not work, and as we discussed, not sure if it is even useful. Sorry for the troubles.

juliusikkala:
No problem!

Btw, about the WaveSize thing: "Wave" is Subgroup in Khronos terminology, you can read more about that feature [here](https://www.khronos.org/blog/vulkan-subgroup-tutorial). Some hardware allows you to select how wide the "Wave"/subgroup is (IIRC at least AMD RDNA3 has 32/64 options and Intel Arc GPUs has even more alternatives).

fknfilewalker:
Interesting that AMD and Intel can configure this size. I think for SPIRV this is set via the execution mode #35.
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Execution_Mode

Issue #5900, opened 2024-12-17
Segfault when accessing interface extension 


tunabrain:
The following snippet causes a compiler segfault:

```
interface IFoo
{
    float get();
}

extension<FooType : IFoo> FooType {
    float load()
    {
        return get();
    }
}

struct Foo : IFoo
{
    StructuredBuffer<float> buffer;
    int dummy;

    float get() { return buffer[0]; }
}

float bugTest(IFoo t)
{
    return t.load();
}

uniform Foo input;
uniform RWStructuredBuffer<float> output;

[shader("compute")]
[numthreads(32, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    output[0] = bugTest(input);
}
```
Variations of it (e.g. commenting out `dummy`) generates invalid HLSL instead.

bmillsNV:
@kaizhangNV can you take a look at this in the current sprint?

kaizhangNV:
Yes

kaizhangNV:
You need to change

```
float bugTest(IFoo t)
```
to
```
float bugTest<T: IFoo>(T t)
```

for now.

I'm not sure whether `float bugTest(IFoo t)` is a correct syntax here.

kaizhangNV:
From what I understand, 

```
extension<FooType : IFoo> FooType {
```

is not actually extending `IFoo` itself,  it extends every type that conforms to `IFoo`, so it uses generic extension here.
I don't know the fundamental reason why slang doesn't support interface extension.

So when you try to use `load` method, you have to use generic as well.

kaizhangNV:
@csyonghe 

I think what we should really fix is to report error at:

```
float bugTest(IFoo t)
{
    return t.load();    // -> report error, because t doesn't have load member function.
}
```

csyonghe:
@kaizhangNV Actually this call is allowed in Slang.

When defining a variable or a parameter whose type is an interface, it means an "existential type". The function will then be specialized by the actual type of the argument, if the type is statically known, or fallback to dynamic dispatch if the type isn't statically known.

kaizhangNV:
> @kaizhangNV Actually this call is allowed in Slang.
> 
> When defining a variable or a parameter whose type is an interface, it means an "existential type". The function will then be specialized by the actual type of the argument, if the type is statically known, or fallback to dynamic dispatch if the type isn't statically known.

[...]

PR #5970, opened 2024-12-31
Fix issue on specialize 'this_type(interface)'

kaizhangNV:
close issue #5900

### DIFF
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -1328,7 +1328,51 @@ struct SpecializationContext
         // If we never found a parameter or return type worth specializing, we should bail out.
         //
         if (!returnTypeNeedSpecialization && !argumentNeedSpecialization)
+        {
+            auto firstParam = calleeFunc->getFirstParam();
+            auto firstArg = inst->getArg(0);
+            if (!firstParam || !firstArg)
+                return false;
+
+            // If the first parameter is `this_type(interface_type)`, we will try to see if we
+            // can specialize this type to some concrete type.
+            //
+            // This is a corner case that we can't handle during maybeSpecializeGeneric() call. For the
+            // specialize call like this `specialize(base, this_type(interface_type), witness_table)`,
+            // this will specialize the function in the witness_table by using the 'this_type(interface_type)'
+            // as the first argument. However, we cannot handle this case in maybeSpecializeGeneric() because
+            // we the concrete type of the 'this_type(interface_type)' is not known at that time. So we have to
+            // handle this case here.
+            if (auto thisType = as<IRThisType>(firstParam->getDataType()))
+            {
+                if(as<IRInterfaceType>(thisType->getConstraintType()))
+                {
+                    auto argType = firstArg->getDataType();
+                    if (isCompileTimeConstantType(argType))
+                    {
+                        IRBuilder builderStorage(module);
+                        auto builder = &builderStorage;
+                        auto newParam = builder->createParam(argType);
+
+                        // Replace the first parameter with the new parameter.
+                        auto firstBlock = calleeFunc->getFirstBlock();
+                        firstBlock->insertParamAtHead(newParam);
+                        firstParam->transferDecorationsTo(newParam);
+                        firstParam->replaceUsesWith(newParam);
+                        firstParam->removeAndDeallocate();
+
+                        // Replace the first parameter of the function type with the
+                        // concrete type.
+                        auto functionType = calleeFunc->getDataType();
+                        auto firstParamType = functionType->getParamType(0);
+                        firstParamType->replaceUsesWith(argType);
+                        firstParamType->removeAndDeallocate();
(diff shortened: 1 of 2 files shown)


source/slang/slang-ir-specialize.cpp
```diff
@@ -1328,7 +1328,52 @@ struct SpecializationContext
...
+            // can specialize this type to some concrete type.
+            //
+            // This is a corner case that we can't handle during maybeSpecializeGeneric() call. For
+            // the specialize call like this `specialize(base, this_type(interface_type),
+            // witness_table)`, this will specialize the function in the witness_table by using the
+            // 'this_type(interface_type)' as the first argument. However, we cannot handle this
+            // case in maybeSpecializeGeneric() because we the concrete type of the
+            // 'this_type(interface_type)' is not known at that time. So we have to handle this case
+            // here.
+            if (auto thisType = as<IRThisType>(firstParam->getDataType()))
```
csyonghe, line 1347: Should we just check for ThisType in the loop above at line 1316?

source/slang/slang-ir-specialize.cpp
```diff
@@ -1328,7 +1328,52 @@ struct SpecializationContext
...
+                        // Replace the first parameter with the new parameter.
+                        auto firstBlock = calleeFunc->getFirstBlock();
+                        firstBlock->insertParamAtHead(newParam);
+                        firstParam->transferDecorationsTo(newParam);
+                        firstParam->replaceUsesWith(newParam);
+                        firstParam->removeAndDeallocate();
+
+                        // Replace the first parameter of the function type with the
+                        // concrete type.
+                        auto functionType = calleeFunc->getDataType();
```
csyonghe, line 1367: firstParamType->replaceUsesWith() is dangerous/incorrect, because there may still be other uses of the type (e.g. outside the function we are processing), and they shouldn't be replaced.

If the intention here is just to create a new IRFuncType and set it to calleeFunc, you can just call fixupFuncType before returning.

source/slang/slang-ir-specialize.cpp
```diff
@@ -1328,7 +1328,52 @@ struct SpecializationContext
         // If we never found a parameter or return type worth specializing, we should bail out.
         //
         if (!returnTypeNeedSpecialization && !argumentNeedSpecialization)
+        {
+            auto firstParam = calleeFunc->getFirstParam();
```
csyonghe, line 1332: I am not understanding the underlying issue we are trying to fix here.

How can a param type of the callee end up being an ThisType? IIRC `ThisType` is only used when declaring an interface requirement, and should never appear in any actual methods. I am very confused why the callee here can have its first param type being an IRThisType. They should only just be IRInterfaceType and be handled with the existing code.

csyonghe (commented):
I want to get better understanding why IRThisType can appear as the parameter type of an actual function, this seems wrong to me.

csyonghe:
The PR should also include a regression test for the issue we are fixing.

source/slang/slang-ir-specialize.cpp
kaizhangNV: `call specialize(base, this_type(interface_type), witness_table) (xxx)`

for above IR, `maybeSpecializeGeneric()` will evaluate `specialize`, and this will generate us a specialized function with `this_type(interface_type)` as the input parameter.

Let's say the function is named `func`, then this IR will be

`call func(xxx)`, and the function will be

```
void func(this_type(interface_type))
```

kaizhangNV: Based on what you said, you mean 

```
call specialize(base, this_type(interface_type), witness_table) (xxx)
```

`this_type(interface_type)` should never appear in `specialize` call?

kaizhangNV:
> The PR should also include a regression test for the issue we are fixing.

Yea, I will add.

source/slang/slang-ir-specialize.cpp
kaizhangNV: I don't know about that.
But if this is the case, then the bug is in more previous stage.

csyonghe:
I did some investigation here, and it appears that the IR for `bugTest` is wrong after lower-to-ir.

It is something like:
```
call(specialize(extension::load, ThisType), ...)
```

The `specialize(extension::load, ThisType)` is an invalid inst. We should never use `ThisType` to specialize anything.

This usually occurs when the front-end produces wrong DeclRef to `extension::load` after type checking. It currently is something like `load<ThisType>`, which makes no sense. It really should be invoking `load<ExtractExistentialType(bugTest::t)>` instead.

csyonghe:
I drafted an attempt to fix this issue in the front-end here: https://github.com/shader-slang/slang/pull/5971

It may still not be 100% correct, but this is the direction we should be exploring.

kaizhangNV:
It makes sense to me that we should fix the front-end if the `ThisType` should not appear in specialize().
Close this PR.

PR #5973, opened 2024-12-31
Add backtraces to examples

aleino-nv:
- Add backtrace-printing support for Windows, for both `main` and `WinMain` examples.
- For `WinMain` example programs, redirect stdout to `stdout-[example].txt` and print that in CI if it's found.

This helps to address #5520.


examples/stacktrace-windows/common.cpp
```diff
@@ -1,13 +1,14 @@
-#include <vector>
+#include "common.h"
+
+#include <dbghelp.h>
```
aleino-nv, line 3: `windows.h` must be included before dbghelp.h. Need to figure out a way to make formatting not reorder the headers like this.

examples/stacktrace-windows/common.cpp
```diff
@@ -0,0 +1,202 @@
...
+    {
+        SLANG_EXAMPLE_LOG_ERROR();
+        return false;
+    }
+
+    fileName = std::string(buffer.data(), buffer.data() + buffer.size());
+    return true;
+}
+
+// NOTE: This function is not thread-safe, due to usage of StackWalk64 and static buffers.
```
expipiplus1, line 73: We should use a mutex here to make this function reentrant

expipiplus1 (changes requested):

examples/stacktrace-windows/common.cpp
```diff
@@ -0,0 +1,202 @@
+#include "common.h"
+
+#include <inttypes.h>
+#include <string>
+#include <vector>
+#include <windows.h>
+// dbghelp.h needs to be included after windows.h
```
aleino-nv, line 7: This can also be done by just putting the dbghelp.h include in it's own block surrounded by empty lines.

examples/stacktrace-windows/common.cpp
aleino-nv: It's currently only called from the main thread.

I suppose it's possible some other thread will call these dbghelp API's, is that the concern?

If so I would prefer it if the function was "externally synchronized" i.e. the mutex is around the call site.

tools/platform/window.h
```diff
@@ -240,30 +240,14 @@ class Application
...
-        void* /*instance*/,                          \
-        void* /* prevInstance */,                    \
-        void* /* commandLine */,                     \
-        int /*showCommand*/                          \
-    )                                                \
-    {                                                \
-        platform::Application::init();               \
-        auto result = APPLICATION_ENTRY(0, nullptr); \
-        platform::Application::dispose();            \
-        GFX_DUMP_LEAK                                \
```
aleino-nv, line 254: `GFX_DUMP_LEAK` should also be expanded in the new main wrapper.

csyonghe (approved):

PR #5976, opened 2024-12-31
Correct IR generation for no-diff pointer type

### LINKED ISSUE #5805: Pointers as parameters to differentiable functions cause compiler to crash
[Here's a little reproducer on Slang playground](https://shader-slang.com/slang-playground/?target=SPIRV&code=eJylVNtu00AQffdXjCoh7OLagSIeGuCFqm-IioLUClC1jWeTVfdi9pLEQv13Zte5OW1FgShx7NmZM3Nmzriu4ctMOMAlU61EcDOzcEAX8AaCQ7iQTE-fO2DBG8W8mEAjOEeL2gt6NBo4Mh8sZnUdYyZGtcEj-BlCg1bMyWmOrsri-ZmxoIxFEJobq_p4-j4KXoJDwjGToMjW-zN_ErFm3rfupK7djFGeIxfrrCh7ne5qqt0eTYNosI7oEbeaeSWzTKjWWA-tZN3UmqCbcartFLnQCAycSJ3gQU_6fLoBxewtCA_M7VZ4I7HKvp0ODD8yLg3z4IL6xC9-BmbR5b1pWUJ_05WgzXUE6g3Hh-DR-SL7lQF9LFI_NSzhkH4voKP_bpzdZdnciAZaK7T_yITO1_5rDB2kPPcW3kG-MhV5IOc3r699PiqKcfKW6CmDC9KT46DKV9WohON4WSGtIlJGnh98TlEn8Ix_1wflCoR8khN18CvJhca6YLa5J5I9ZUwtawSdguHp2QRPZ7CobOUrYm207KpNuRHskqrliyZ1Ld8tu8ij6ZwJ2xN4WY2KEja2xGeUbA-zas5gQXJYbnilbFXzdGKtNS2bsn1q3BpFQo-8WmaZQo_Wkf8adcV5zmTAPa5X_0P25RPIJlF3Q8pX9zjfsMntlvRfsN4Z6ZYvk_J-O2gborOjZ_BCUSPW7h8sRnC2t3AULGzKz5wjvBRO9BSTfSsTNTq3OEExnw_L6_u8s7Ey9u1tWpf30LRRZcMOR-mM_xB1tRt1PIwiJucPd2oo_ZzGxotYeDTu96mv--YRSbSXZaxiM_ReA49MfrWGDb3dhZ7uzXg74q022n4ZHgDr_gGsV9ndb0YAIPQ)

It seems that whenever I pass a slang pointer to a differentiable function, the compiler crashes. 

This occurs regardless of if I mark that pointer with no_diff or not. 

It also occurs regardless independently from whether that pointer is actually touched by the code or not. In the example above, I don't actually read from the pointer, but I still see the issue.

kaizhangNV:
Close #5805.

There is an issue on checking whether a pointer type parameter is no_diff, we should first check whether this parameter is an Attribute type first, then check the data type.

In the back-propagate pass, for the pointer type parameter, we should load this parameter to a temp variable, then pass it to the primal function call. Otherwise, the temp variable will no be initialized, which will cause the following calculation wrong.


source/slang/slang-ir-autodiff.cpp
```diff
@@ -126,6 +126,12 @@ static IRInst* _getDiffTypeWitnessFromPairType(
 
 bool isNoDiffType(IRType* paramType)
 {
+    if (auto attrType = as<IRAttributedType>(paramType))
```
csyonghe, line 129: It seems to me that we need to fuse the two loops below into one, instead of adding another check here.

source/slang/slang-ir-autodiff.cpp
```diff
@@ -133,12 +133,13 @@ bool isNoDiffType(IRType* paramType)
     }
 
     while (auto ptrType = as<IRPtrTypeBase>(paramType))
-        paramType = ptrType->getValueType();
-    while (auto attrType = as<IRAttributedType>(paramType))
     {
-        if (attrType->findAttr<IRNoDiffAttr>())
+        paramType = ptrType->getValueType();
```
csyonghe, line 137: I think it should be a general loop that can peel off either an attributed type or a ptr type during each iteration.

csyonghe (approved):

source/slang/slang-ir-autodiff.cpp
```diff
@@ -126,20 +126,22 @@ static IRInst* _getDiffTypeWitnessFromPairType(
 
 bool isNoDiffType(IRType* paramType)
 {
-    if (auto attrType = as<IRAttributedType>(paramType))
+    for(;;)
```
csyonghe, line 129: I would just write while(parameter) here to get rid of any compiler warnings on unreachable code.

kaizhangNV:
There is some error in the deubgprintf

Issue #5979, opened 2025-01-01
Support arbitrary test directories


bspeice:
I'm currently trying to set up a new project, and I'd like to use `slang-test` to help with making sure my shaders function properly - it's easier than writing a test harness myself. While I can do so by making my own `tests/` directory, it would be much easier to solve this adding support for tests being located in an arbitrary directory.

PR #5980, opened 2025-01-01
Allow explicit test root

bspeice:
Intended to address #5979.

From what I can tell, this should be a non-breaking change. The existing tests still run correctly, and I've verified on my computer that I can use an arbitrary test directory with the `-test-dir` option.

While attempting to implement this, I did also notice an off-by-one issue with the binary name on Linux. While `_calcExectuablePath` properly null-terminates the program path, it overwrites the final non-null character of the name. This ends up breaking functions like `Path::getCanonical` later on, because `slang-tes` does not exist. If that change should be in a separate PR, please let me know.

### DIFF
--- a/source/core/slang-io.cpp
+++ b/source/core/slang-io.cpp
@@ -992,12 +992,12 @@ static SlangResult _calcExectuablePath(char* outPath, size_t* ioSize)
     {
         return SLANG_FAIL;
     }
-    if (size_t(resSize) >= bufferSize)
+    if (size_t(resSize + 1) >= bufferSize)
     {
         return SLANG_E_BUFFER_TOO_SMALL;
     }
     // Zero terminate
-    outPath[resSize - 1] = 0;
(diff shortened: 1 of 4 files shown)


csyonghe (approved):

csyonghe:
This will make slang-test easier to use during development. Thank you for contributing the patch!

csyonghe:
The change seems to break existing tests.

bspeice:
...and specifically on Windows for some reason. Will take a look at what's going on, thanks for the heads up.

bspeice:
Ah, tests are breaking because of an unintentional change in the path separator. I was using `tests` as the directory name, which led Windows to use filenames like `test\...`. Using a directory name of `tests/` fixes things up for me locally, waiting on actions to complete to make sure it's fixed for good.

csyonghe (approved):

PR #5984, opened 2025-01-02
Fix some robustness issues in the examples

aleino-nv:
Do some robustness fixes for a few of the examples.
This fixes a hang in the raytracing samples, and fixes the test failure in the hello-world Windows/Debug config.

This helps to address #5520.


examples/hello-world/main.cpp
```diff
@@ -22,6 +22,7 @@ struct HelloWorldExample : public TestBase
 {
     // The Vulkan functions pointers result from loading the vulkan library.
     VulkanAPI vkAPI;
+    bool vkAPIValid = false;
```
csyonghe, line 25: Why is this needed?

aleino-nv: I check this value in `HelloWorldExample::~HelloWorldExample` which previously called some uninitialized VK API functions in the case where VK API initialization failed.

csyonghe: Uninitialized vk api should be nullptr, you should be able to check for the function pointer before calling it.

aleino-nv: Sure, but then we would need to check every function pointer.
This variable means that all of the pointers should be initialized so we can just check this one value.

csyonghe: If it is the destructor that is causing the problem, checking for null vk device handle before calling destroy should be more than enough. If device is null, just return immediately.

aleino-nv: > I still don’t understand why this is needed. If the basic Vulkan functions are not found, shouldn’t we exit already?
The destructor of the app still runs, regardless of whether `initializeVulkanDevice` succeeded.

On the other hand, I believe `initializeVulkanDevice` could successfully set some of the VK API pointers (but not others) if it doesn't make it all the way to the end and returns 0, so I don't think it's enough to just check one of the pointers.

aleino-nv: Here is what the current code looks like

    int main(int argc, char* argv[])
    {
        initDebugCallback();
        HelloWorldExample example;
        example.parseOption(argc, argv);
        return example.run(); // ...destructor still runs and uses VK API
    }
    
    int HelloWorldExample::run()
    {
        RETURN_ON_FAIL(initVulkanInstanceAndDevice()); // <--- This can fail if VK API initialization fails
        RETURN_ON_FAIL(createComputePipelineFromShader());
        RETURN_ON_FAIL(createInOutBuffers());
        RETURN_ON_FAIL(dispatchCompute());
        RETURN_ON_FAIL(printComputeResults());
        return 0;
    }

aleino-nv: > The VkAPI class already provides some methods that returns whether or not it is successfully loaded, no?

Not as far as I can tell. Do you want me to move the boolean into the VkAPI struct and set it to `true` just before returning 0?

aleino-nv: > If it is the destructor that is causing the problem, checking for null vk device handle before calling destroy should be more than enough. If device is null, just return immediately.

Ok I did something like this, please have a look.

examples/hello-world/main.cpp
```diff
@@ -511,15 +515,18 @@ int HelloWorldExample::printComputeResults()
 
 HelloWorldExample::~HelloWorldExample()
 {
-    vkAPI.vkDestroyPipeline(vkAPI.device, pipeline, nullptr);
-    for (int i = 0; i < 3; i++)
+    if (vkAPI.device != VK_NULL_HANDLE)
```
csyonghe, line 518: always prefer early return over nesting `if`:
```
if (!vkAPI.device)
   return;
```

aleino-nv: Done.

csyonghe (approved):
Looks good!

Issue #5992, opened 2025-01-03
Properly set `D3D12SDKPath` as `.\\D3D12\\`


jkwak-work:
When we load D3D12.DLL, it loads D3D12Core.DLL either from C:\windows\system32 or a directory the application specifies.
By default, the application specified directory is `.\\D3D12\\` relative to where the executable file exists.

The directory path, however, can be set explicitly via `D3D12SDKPath` variable in the application.
It has been working fine without explicitly setting it but we found edge cases where it doesn't work.
More specifically we saw that slang-test fails to find D3D12Core.dll when ran with `-use-test-server` option.

It will be safe to follow what MS officially recommends:
https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/#2.-set-agility-sdk-parameters

PR #5993, opened 2025-01-03
Properly set D3D12SDKPath

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/5992

This commit have an effect on slang-test.exe and test-server.exe.

When we run slang-test.exe, it uses D3D12Core.DLL located at `build/{Release,Debug}/bin/D3D12`, which is `D3D12` from where the executable file is located at.

This has been working fine but we found a case where it doesn't work properly.  This commit sets D3D12SDKPath value explicitly to avoid potential problems.


tools/slang-test/slang-test-main.cpp
```diff
@@ -47,6 +47,12 @@
 
 using namespace Slang;
 
+#if defined(_WIN32)
+extern "C" { __declspec(dllexport) extern const uint32_t D3D12SDKVersion = 711; }
```
aleino-nv, line 51: It would be nice to have a comment pointing to the documentation for these, or some kind of explanation of what effect these have.

aleino-nv (approved):
LGTM but would be good to document the exports a bit.

tools/slang-test/slang-test-main.cpp
jkwak-work: Good idea

jkwak-work:
I thought this PR will have an effect on both slang-test and test-server, but it turned out that it worked only for slang-test.
I have another PR for test-server.
https://github.com/shader-slang/slang/pull/6095

PR #6000, opened 2025-01-03
  Check whether array element is fully specialized

### LINKED ISSUE #5776: Invalid HLSL generated for bwd_diff function
I'm running into an issue where slang will generate invalid HLSL for the following snippet:

```
struct Foo
{
    float foo(int idx) { return 5.0f; }
}
struct WrappedBuffer<T, let D : int>
{
    StructuredBuffer<T> buffer;
    int[D] shape;

    T get(int idx) { return buffer[idx]; }
}
struct GradInBuffer<T : IDifferentiable, let D : int>
{
    RWStructuredBuffer<T> primal;
    WrappedBuffer<T.Differential, D> grad_in;
}

[Differentiable, BackwardDerivative(set_bwd)]
void set(GradInBuffer<float[1], 1> t, int idx, float[1] value)
{
    t.primal[idx] = detach(value);
}
void set_bwd(GradInBuffer<float[1], 1> t, int idx, inout DifferentialPair<float[1]> grad)
{
    // Generates invalid HLSL, "candidate function not viable: no known conversion from 'WrappedBuffer_0' to 'WrappedBuffer_1' for 1st argument"
    grad = diffPair(grad.p, t.grad_in.get(idx));

    // This works:
    //grad = diffPair(grad.p, t.grad_in.buffer[idx]);
}

struct CallData
{
    Foo weights;
    RWStructuredBuffer<float> dOut;
    GradInBuffer<float[1], 1> result;
}
ParameterBlock<CallData> call_data;

[shader("compute")]
[numthreads(32, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    int idx = dispatchThreadID.x;

    float[1] primal = { call_data.weights.foo(idx) };
    DifferentialPair<float[1]> dResult = diffPair(primal);
    bwd_diff(set)(call_data.result, idx, dResult);

    call_data.dOut[idx] = dResult.d[0];
}

```
It seems to generate two identical specializations for WrappedBuffer (`WrappedBuffer_0` and `WrappedBuffer_1`), and calls the get method of `WrappedBuffer_1` with an instance of `WrappedBuffer_0`. This only happens for the backwards pass and is very finicky; e.g. removing `int[D] shape;` or changing T from `float[1]` to `float` makes it go away.

kaizhangNV:
close #5776
    
When we start specialize a "specialize" IR, we should
make sure all the elements are fully specialized, but
we miss checking the elements of an array. This change
will check the it.


source/slang/slang-ir-specialize.cpp
```diff
@@ -123,6 +123,12 @@ struct SpecializationContext
                 }
                 return false;
             }
+        case kIROp_ArrayType:
+            {
+                auto array = as<IRArrayType>(inst);
```
csyonghe, line 128: I wonder why this case is necessary. 
If we have an type that is `T.Differential[N]`, then it shall be encoded as `IRArrayType(IRLookupWitness(w, "Differential"), N)`, and since `IRLookupWitness` has operand `w`, it won't appear in global scope, and therefore the IRArrayType itself won't appear directly in the global scope, and therefore it won't be reported as `true` by the default case logic.

Basically, if the element type is not fully specialized, then the array type itself shouldn't appear in the global scope. What IR are you seeing?

kaizhangNV: Yes, this case this necessary, the unit test can reproduce the case.

csyonghe: I ask this question because in theory this shouldn’t happen, otherwise we will have the same issue for Ptr type and all other kinds of wrapper types.

kaizhangNV: If you dump the IR from the unit test, you will find out:
```
Ptr(specialize(%84, specialize(%150, Float, 1 : Int, %168)))	= get_field_addr(%callx5Fdata, %result)
```
this could result as:

```
	struct %GradInBuffer	: Type
	{
		field(%gradx5Fin, specialize(%82, lookupWitness( witness_table_t(%IDifferentiable)(Array(Float, 1)),   %37)))
	}
```

and finally this lookup table will give us

```
Array(lookupWitness(%168, %37), %N6)
```

here is the case I need to handle.

csyonghe: I see. In this case we do need to also handle all other opcodes by recursively checking their operands. Because we could have PtrType(lookupWitness()), or ArrayType(int, lookupwitness()), etc.

kaizhangNV: I verified that `vector<float, N>` seems not having this problem, but not sure about pointer or matrix, etc.

csyonghe: We need to at least include ptrType, AttributedType, all the texture and buffer types, etc.

Basically everything needs to bottleneck through areAllOperandsFullySpecialized. But doing so can be slow, due to redundant checks.

We probably need to refactor the code so we sweep the global scope once and insert all fully specialized insts into a hash set, and run the deep check if an inst isn't in the hash set.

csyonghe: vector<T.Differential, N> should have the same problem, no?

csyonghe: the case should be handled by calling:
```
return areAllOperandsFullySpecialized(inst);
```

this is because for array<T, N>, we should not only verify `T`, but also `N`.

kaizhangNV: After comparing the witness table between vector and array, I found out only array could have this kind of issue:
```
        witness_table %324      : witness_table_t(%IDifferentiable)(Vec(%T39, %N36))
        {
                witness_table_entry(%46,Vec(%T39, %N36))

                witness_table_entry(%45,%321)

                witness_table_entry(%49,specialize(%203, %T39, %N36, %320))

                witness_table_entry(%51,specialize(%191, %T39, %N36, %320))

                witness_table_entry(%53,specialize(%173, %T39, %N36, %320))

        }
```

where %T39 and %N36 are concret. 

```
generic %150	: witness_table_t(%IDifferentiable)
{
block %151(
		[TypeConstraintDecoration(%IDifferentiable)]
		[nameHint("T")]
		param %T7	: type_t,
		[nameHint("N")]
		param %N6	: Int,
		param %152	: witness_table_t(%IDifferentiable)):
	let  %153	: Type	= lookupWitness(%152, %37)
	let  %154	: witness_table_t(%IDifferentiable)	= lookupWitness(%152, %36)
	witness_table %155	: witness_table_t(%IDifferentiable)(Array(%T7, %N6))
	{
		witness_table_entry(%37,Array(%153, %N6))

		witness_table_entry(%36,specialize(%150, %153, %N6, %154))

		witness_table_entry(%40,specialize(%142, %T7, %N6, %152))

		witness_table_entry(%42,specialize(%121, %T7, %N6, %152))

		witness_table_entry(%44,specialize(%95, %T7, %N6, %152))

	}
}
```
Where %T7 is concret, but %153 is a lookupWitness.

It's not easy for me to reproduce the issue on other types.
I tried `Ptr<>` type, however, it's not conformed to IDifferentiable type.

kaizhangNV: Same for the texture and buffer types. (Not conformed to IDifferentiable).

kaizhangNV: Actually, I'm wondering more that why it's 

`witness_table_entry(%37,Array(%153, %N6))` instead of `witness_table_entry(%37,Array(%T7, %N6))`

csyonghe: This issue is not specific to `IDifferentiable` though. It feels like I can craft an interface `IFoo` and have vector, ptr and array conform to that interface, where during specialization there can be a `IRPtrType<LookupWitness()>` being created in global scope.

csyonghe: Theoretically, there can be any type opcode `IRTypeX(arg0, arg1, ...)`, and we shouldn't consider this type as specializable if any of `arg_i` is not specializable.

kaizhangNV: I want to give it a try to craft an interface

```
__generic<T:IFoo, let N:int>
extension Array<T, N> : IFoo
{
    float foo()
    {
        return 0.0;
    }
}
```

is this the right syntax?

csyonghe: In your syntax you specified `T:IFoo`, since `float !: IFoo`, `Array<float,1> !: IFoo`.

kaizhangNV: Oh, I see the problem, I forgot removing that line.

OK, I will see if I can craft a shader to reproduce this issue.

csyonghe: For a repro, you probably need to define an associatedtype in the interface for lookupwitness to occur inside an array type

for example, something like:
```
interface IFoo
{
      associatedtype FA : IFoo;
     int get();
}
struct X : IFoo { typealias FA = X;  int get() { return -1; } }
extension<T, int n> Array<T, n> : IFoo
{
      typealias FA = X;
}

int test<T:IFoo>() {
     T.FA v[2];
     return v[0].get();
}

...
```

kaizhangNV: I wrote another shader a little bit different from yours, but now I can get stable reproduce.

kaizhangNV:
Update the patch to cover all wrapper types, also added more test cases.

source/slang/slang-ir-specialize.cpp
```diff
@@ -123,12 +123,20 @@ struct SpecializationContext
...
+        // For all the wrapper type, we need to make sure the operands are fully specialized.
         case kIROp_ArrayType:
-            {
-                auto array = as<IRArrayType>(inst);
-                auto elementType = array->getElementType();
-                return isInstFullySpecialized(elementType);
-            }
+        case kIROp_TextureType:
+        case kIROp_VectorType:
+        case kIROp_MatrixType:
```
csyonghe, line 130: Also: TupleType, OptionalType, TypePack.

we should define a isBuiltinGenericType() utility in slang-it-util that returns true for these opcodes so that we don’t have to duplicate these switch cases all over the place.

kaizhangNV: sure.

An extra question, in the test, I actually want to cover the case like:

StructuedBuffer< StructuredBuffer<float> > input;

but I'm struggled with how to feed data to this type.

From the generated code, I do see we handle this situation correctly, but I didn't cover this in the test as I don't know how to feed data.

csyonghe: This is an invalid type. Slang should check and report error on this.

kaizhangNV: miss the inner generic parameter, is this the invalid type?
`StructuedBuffer< StructuredBuffer<float> >`

kaizhangNV: > Also: TupleType, OptionalType, TypePack.
> 
> we should define a isBuiltinGenericType() utility in slang-it-util that returns true for these opcodes so that we don’t have to duplicate these switch cases all over the place.

Done of adding this.

csyonghe (approved):

PR #6001, opened 2025-01-03
Use disassemble API from SPIRV-Tools

jkwak-work:
This commit uses C API version of SPIRV disassemble function rather than calling spirv-dis.exe.

This allows us to use a correct version of SPIRV disassble function that Slangc.exe is using.

The implementation is mostly copied from `external/spirv-tools/tools/dis/dis.cpp`, which is a source file for building spirv-dis.exe.


jkwak-work:
Converting to draft while working on the failing test

csyonghe (dismissed):

PR #6002, opened 2025-01-04
Add `Bindless<T>` proposal.

csyonghe:


docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+    // to `SamplerState`.
+    t.Sample(sampler, float2(0,0));
+}
+```
+
+A `Bindless<T>` type is a concrete type whose size is always 8 bytes and is internally a 64-bit integer.
+This means that you can use a `Bindless<T>` type in any context where an ordinary data type, e.g. `int` type
+is allowed, such as in buffer elements.
+
+On targets where resource handles are already concrete and sized types, `Bindless<T>` simply translates to just `T`.
```
tangent-vector, line 86: This statement might need some clarification on what happens if the opaque resource types have a size/alignment other than 8 bytes. Ideally, I would expect that `Bindless<T>` is always stored with 8-byte size/alignment, even on targets where `T` might have smaller size/alignment.

That still leaves the open question of what happens if `T` has a native size or alignment that is greater than 8 bytes...

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+
+// All texture and buffer handles are defined in descriptor set 100.
+[vk::binding(0, 100)]
+__DynamicResource<__DynamicResourceKind.General> resourceHandles[];
+
+// All sampler handles are defined in descriptor set 101.
+[vk::binding(0, 101)]
+__DynamicResource<__DynamicResourceKind.Sampler> samplerHandles[];
+
+export getResourceFromBindlessHandle<T>(uint64_t handle) where T : IOpaqueHandle
```
tangent-vector, line 131: I'm behind on reading up on the link-time specialization feature, but is there any note that needs to be made on what happens if multiple user modules try to implement this method?

I'm not entirely comfortable with this kind of thing as a mechanism, but I also have to concede that I can't think of anything better that we could implement in a reasonable timeframe.

docs/proposals/015-bindless-t.md
csyonghe: Done.

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+    if (T.getResourceKind() == ResourceKind.Sampler)
+        return (T)samplerHandles[handle];
+    else
+        return (T)resourceHandles[handle];
+}
+```
+
+### Invalid Handle
+
+We reserve `uint64_t.maxValue` as a special handle value of `Bindless<T>` types to mean an invalid/null resource.
```
tangent-vector, line 142: It's unfortunate that we can't easily make the encoding of an invalid handle be a zero-valued integer, to keep things consistent with the encoding of pointers.

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+This will allow us to optimize `Optional<Bindless<Texture2D>>` to use the reserved value to mean no-value.
```
tangent-vector, line 143: I initially didn't catch this sentence, which seems to be very important.

Notably, we seem to be saying that in all other contexts, the Slang compiler will assume that a non-optional `Bindless<T>` has a valid handle value, and that it is programmer error (I'm going to hold back from saying "undefined behavior") if they attempt to use a `Bindless<T>` that has an invalid handle value.

(This isn't actually surprising, since there will no doubt be many integers that don't represent valid handles, and using any of those would be a programmer error too. The point is more that the compiler is allowed to assume a `Bindless<T>` is valid, and programmers are expected to use an `Optional<>` if they want to allow for invalid handles.)

tangent-vector: I assume that `Optional<Bindless<T>>` will Just Work with an `if let`?

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+The user should also be able to use `Bindless<T>.invalid` to refer to such an invalid value:
+
+```slang
+struct Bindless<T> where T:IOpaqueHandle
+{
+    static const Bindless<T> invalid = Bindless<T>(uint64_t.maxValue);
+}
+```
+
+## Conclusion
```
tangent-vector, line 154: What I see missing here is any discussion of existing or alternative solutions, as well as possible directions for improvement/extension.

(To be clear, I don't think any of these *need* to be added for this PR to land; I'm more just saying they would be good to have added to this initial proposal over time.)

tangent-vector: One very concrete example of a possible future direction that I know we've discussed before is making `Bindless<T>` applicable to all types, effectively doing a structural recursion over types. E.g., a `Bindless<float4>` would basically just be a `float4`, while a `Bindless<Foo>` given `struct Foo { float4 a; Texture2D b; }` would be more-or-less equivalent to `struct Bindless_Foo { Bindless<float4> a; Bindless<Texture2D> b; }`.

Such an approach would in theory allow developers to author modules/effects/subsystems that can be used in both bindless and "bind-full" contexts, with the code *using* the module deciding between `SomeType` and `Bindless<SomeType>` depending on their application's own design considerations.

At this point I'm not clear on whether that feature would be useful enough that we'd ever get to it, but it is worth having around as a historical note at the least.

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+Where `IOpaqueHandle` is an interface that is implemented by all texture, buffer and sampler state types:
+
+```slang
+enum ResourceKind
+{
+    Unknown, Texture, ConstantBuffer, StorageBuffer, Sampler
+}
+interface IOpaqueHandle
+{
+    static ResourceKind getResourceKind();
```
tangent-vector, line 43: Should this effectively be an associated constant?

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,160 @@
...
+
+Direct3D Shader Model 6.6 introduces the "Dynamic Resources" capability, which further simplifies the way to write bindless shader code by removing the need to even declare the global array.
+
+We believe that graphics developers will greatly benefit from a system defined programming model for the bindless parameter passing idom that is versatile and cross-platform, which will provide a consistent interface so that different shader libraries using the bindless pattern can interop with each other without barriers.
+
+## Proposed Approach
+
+We introduce a `Bindless<T>` type that is defined as:
+```
+struct Bindless<T> : IComparable<Bindless<T>>
```
tangent-vector, line 28: Unrelated: I wonder what it would take for us to give the type parameter of `IComparable<T>` a default argument of `This`. That would allow a developer to simply write `struct MyType : IComparable` instead of `struct MyType : IComparable<MyType>` or even an explicit `struct MyType : IComparable<This>`.

Chewing on it a bit, I suspect it would actually be harder to implement than I realize...

tangent-vector (dismissed):
This seems solid, and there is a very real need for a feature like this. The very different idioms for implementing bindless across target platforms can lead to a lot of platform-specific code spread throughout a codebase, and this feature seems like it will help more code be written in a fashion that doesn't have to care about the platform-specific details.

docs/proposals/015-bindless-t.md
tangent-vector: I'm just now paging in a major counter-point against the more expansive `Bindless<T>` that I've described in this comment.

For targets where bindless isn't an option (which really only includes our older or more "legacy" targets), a `Bindless<T>` will just map to a `T` anyway. A developer who wants to write maximally portable code can in principle just use `Bindless<T>` liberally in their type declarations, and still have code that works more-or-less anywhere.

docs/proposals/015-bindless-t.md
csyonghe: Yes, it should.

docs/proposals/015-bindless-t.md
csyonghe: Added discussion on this.

docs/proposals/015-bindless-t.md
csyonghe: yes that's better, fixed.

docs/proposals/015-bindless-t.md
natevm: Is there a reason why 0 cannot be used instead of uintMax?

docs/proposals/015-bindless-t.md
```diff
@@ -0,0 +1,199 @@
...
+
+While relaxing `T` this way adds an extra layer of convenience, it introduces complicated
+semantic rules to the type system, and there is increased chance of exposing tricky corner
+cases that are hard to get right.
+
+An argument for allowing `T` to be general composite types is that it enables sharing the same
+code for both bindless systems and bindful systems. But this argument can also be countered by
+allowing the compiler to treat `Bindless<T>` as `T` in a special mode if this feature is found to be useful.
+
+For now we think that restricting `T` to be an `IOpaqueHandle` type will result in a much simpler implementation, and is likely sufficient for current needs. Given that the trend of modern GPU architecture is moving towards bindless idioms and the whole idea of opaque handles may disappear in the future, we should be cautious at inventing too many heavy weight mechanisms around opaque handles. Nevertheless, this proposal still allows us to relax this requirement in the future if it becomes clear that such feature is valuable to our users.
```
natevm, line 189: I’m not sure where to put this, but figured I’d ask, is there a way to specify that a bindless structured buffer is “globally coherent”? 

This appears to be a limitation with slang’s buffer pointers, so I’m currently carrying buffers around in a single mutable descriptor array, where I then create an unofficial integer addressing like described here. Is the idea to have the user add this via what’s retrieved from getResourceFromBindlessHandle?

natevm, line 189: One concern, a student / newcomer to slang and GPU programming might by mistake try declaring an opaque handle directly, eg a Texture2D, in a structure expecting a size, eg as a uniform parameter to a closest hit program. It’s a bit unnatural to wrap the opaque handle with this Bindless syntax (though I’d be happy to have this over what I’m doing currently)

At a higher level, Binding<T> is quite a bit more verbose than *IOpaqueType, and won’t be as familiar to those coming from something like CUDA/C. A CUDA user would probably be confused as to why they can’t continue to represent their textures via pointers. I just worry it’s going to be too easy a mistake for users to make, to forget to wrap their handles with Binding<intendedType>. 

I suppose I’m just a bit confused why we couldn’t extend pointers to accomplish this same behavior, but I’m a much more concise way, eg via a special pointer referencing an opaque type. If the user casts this special pointe, it can only be  to another opaque handle type, which is ultimately still a “opaque handle” pointer, so it remains a 64-bit integer. If the user adds an integer offset to this special pointer type, then one could handle subsequent dereferencing behavior this through some operator overloading equivalent to the “getResourceFromBindlessHandle” function proposed above.

Otherwise, perhaps some compiler flag could be specified to somehow interpret opaque handles in structs / uniforms as if they were already marked as being bindless?

docs/proposals/015-bindless-t.md
natevm: On alternatives, the benefits I see with this proposal have a lot to do with portability.

I can effectively implement my own version of this Bindless wrapper myself, but then others trying to use my slang code will have their own idiomatic mechanisms, so they’d have to modify the code I give them to account for those differences. Having some standard solution to this could help.

I also think we could be doing more than just Bindless<T> here. As an end user, it’s hard to override certain operators / syntax patterns involving pointers, so that might also be worth considering.

docs/proposals/015-bindless-t.md
maxime-modulopi: The bindless handle is generally an index into a descriptor heap/descriptor binding, so 0 is a valid value

natevm: If we could get 0 to act as invalid, then using pointer syntax would make more sense, given that nullptr is 0.

In my code base, I’ll normally allocate a “default” texture/sampler, eg a single magenta pixel, then assign this to slot 0 as if it were an invalid handle. Perhaps something like this could be standardized in the proposal?

docs/proposals/015-bindless-t.md
csyonghe: This will need to be a separate change that is not related to this proposal. We should probably extend StructuredBuffer and ByteAddressBuffer to take one additional generic parameter to specify the memory semantic, just like how we extended the types to include layout.

csyonghe: Supporting general pointer type can be difficult. If we have a struct type whose field is a texture, what does it mean to have a pointer to the struct type?

That will be similar to the more general form of Bindless type we discussed here and it will be make the system significantly more complicated.

natevm: > If we have a struct type whose field is a texture, what does it mean to have a pointer to the struct type?

Note, what I’m proposing is a special “opaque handle pointer” type, which cannot be cast to a non-opaque type. A struct is not opaque, so this isn’t an issue. 

Like with C++, a struct type containing a field which is a texture would mean a struct type containing a uint64_t. Behavior is well defined, so long as that structure’s uint64_t maps to a valid descriptor in the users global mutable descriptor array.

natevm: For non-opaque types, so long as structs are allocated in a device buffer, users today can already use slang’s pointer syntax to refer to these structures or fields inside them in a “bindless” way.

So I don’t personally see any additional benefits to be had from additionally supporting non-opaque types.

natevm: Another way to think of this is having two address spaces. The pointers slang supports today refer to a traditional device address space. Opaque handle pointers would refer to the mutable descriptor allocation. The two pointer types would then not be allowed to refer to each others memory spaces.

csyonghe: Isn’t that type exactly what Bindless<T> is?

csyonghe: Are you suggesting we rename Bindless to HandlePtr or something like that, and keep everything in the proposal?

csyonghe: The problem with using just a pointer type, is then we can’t map Texture2D* to just Texture2D on platforms where resource handles are already sized handles (e.g. CUDA), if we do that it can lead to more confusions.

natevm: > Isn’t that type exactly what Bindless is?

Yes, pointers are exactly Bindless. It’s exactly this reason why i think it’s a bad idea to invent a new idiom in favor of more intuitive pointer syntax. 

I’m arguing for this:
```
Sampler2D* sampler; <- opaque pointer type
t.Sample(*sampler, float2(0,0)); <- standard practice dereferencing syntax
```

Rather than this:
```
Bindless<Sampler2D> sampler; <- non-standard syntax
t.Sample(sampler, float2(0,0)); <- sketchy implicit type conversion
```

I kinda don’t see the point in this feature otherwise, since I could just implement what’s being proposed in this document myself otherwise…

natevm: > The problem with using just a pointer type, is then we can’t map Texture2D* to just Texture2D on platforms where resource handles are already sized handles (e.g. CUDA), if we do that it can lead to more confusions.

Wouldn’t it be less confusing and more portable to instead have the same behavior for Texture2D* for all targets?

I find it artificially limiting that we’re currently reserving Texture2D* for CUDA. I also suspect Bindless<Texture2D> would lead to more confusion than what I’m proposing. Eg, does this map to Bindless<Texture2D*>? If so, why not just Texture2D*?

natevm: Ultimately I’m trying to advocate for a more consistent solution to what C/CUDA/OptiX programmers are comfortable with. 

There is value in supporting some bindless paradigm like this.

Personally, my gut is telling me that what’s being proposed right now, Bindless<T>, will give off the impression that this is more of a WAR to a deeper rooted issue, namely that Textures cannot be addressed using pointers, without much rhyme or reason. 

I worry that a workaround like this is going to lead to tech debt, especially for a feature like bindless textures which will see a lot of use. 

If eventually if SPIRV exposes a mechanism to address textures like it does for Buffers and Acceleration Structures, it would be much easier to port already existing pointer syntax to those instructions. Having users rewrite their Bindless<T> to *T would be much harder, and we’d more likely end up unnecessarily needing to support two separate ways to address opaque handles.

csyonghe: I don’t like the idea of using pointer syntax for this because it is a much more limited version of pointers and a lot of normal pointer operations won’t work.

for example, there is no way to support
Texture2D * p = &gParams.texture;

there is ambiguity in what Texture2D* mean on platforms where Texture2D is already a non-opaque handle.

generally in Slang we want to discourage people from using pointer pervasively l, instead of ecouraging more use of pointers.

I also disagree that this is what a user can do without compiler support. The implicit dereferencing and casting is not what a user can write on their own, and the language needs to standardize how Bindless code is written so different libraries can interop with each other instead of rolling their own representation of Bindless textures.

csyonghe: On CUDA, Texture2D and Bindless<Texture2D> should both map to sampler2D. There is no such thing as Texture2D* in CUDA. 

With pointer syntax, if would be very confusing if Texture2D* maps to sampler2D and not sampler2D*.

csyonghe: A key semantic of `Bindless<T>` that is not embeded within `Ptr<T>` is that `Bindless<T>` can map to just `T` on platforms where `T` is already a physical value, while `Ptr<T>` should always map to a pointer to `T` instead of just `T` in target code. Using dedicated `Bindless<T>` instead of just `T*` gives the compiler the freedom to implement the implied semantic without fearing of confusing with what users would expect from a real pointer.

Issue #5989, opened 2025-01-02
Invalid SPIR-V for shared memory atomics


Axel-Reactor:
```
#version 460

layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;

shared uint32_t SHARED_MEMORY_test_atomics_SharedMemory_shared_u32;

[shader("compute")] void main()
{
    atomicAdd((SHARED_MEMORY_test_atomics_SharedMemory_shared_u32), 1u);
}
```
```
slangc test.slang -target spirv -o test.spv
spirv-val test.spv
```
->
```
error: line 17: AtomicIAdd: storage class forbidden by universal validation rules.
  %6 = OpAtomicIAdd %uint %SHARED_MEMORY_test_atomics_SharedMemory_shared_u32 %uint_1 %uint_64 %uint_1
```

csyonghe:
We do not support `shared` keyword at the moment, consider use `groupshared` instead.

csyonghe:
The current `shared` keyword is occupied by the D3D11 effect system syntax, an it should be safe for us to repurpose it to mean groupshared when in GLSL mode.

Axel-Reactor:
I still think it's a bug to not error and instead produce invalid SPIR-V?

juliusikkala:
FWIW, I just got the exact same SPIR-V validation error, but in the GLSL mode using the `shared` keyword. I can take a look at mapping `shared` to `groupshared` in GLSL mode over the weekend unless it's already fixed by then :)

aleino-nv:
@csyonghe @juliusikkala What is left to do here?
With the[PR](https://github.com/shader-slang/slang/pull/6003), the generated SPIR-V is now passing validation.

juliusikkala:
The PR only fixed the 'shared' keyword in GLSL to parse into `HLSLGroupSharedModifier`, but not the root cause of generating invalid SPIR-V if the AST contains `HLSLEffectSharedModifier`. This issue can probably be closed if it's not possible to cause this situation via the HLSL syntax anymore either, but I haven't yet tested if that's possible.

aleino-nv:
Before the PR, the `shared` modifier was basically just ignored (i.e. if you remove it you get the same code) as you would expect since in HLSL it's a "hint to the compiler" https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax. 

[...]

PR #6003, opened 2025-01-04
Make 'shared' keyword map to 'groupshared' in GLSL mode

juliusikkala:
The "shared" keyword in GLSL mode should be equivalent to "groupshared" in HLSL & Slang. It got interpreted as a D3D11 effect system thing instead, but that would never appear in GLSL code. Because of this, "shared" was ignored, breaking effectively any GLSL shader using shared memory. This PR fixes that by interpreting "shared" as HLSLGroupSharedModifier when GLSL input is allowed.

Related to #5989 but does not close that issue.


csyonghe (approved):
LGTM, thank you!

PR #6004, opened 2025-01-05
Add SampleCmpLevel intrinsics

fairywreath:
Adds `SampleCmpLevel` intrinsics introduced in [SM 6.7](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_7_Advanced_Texture_Ops.html#samplecmplevel) which performs depth comparison sampling operations with an explicit LOD level. Parallel implementations exist in SPIRV, GLSL and Metal. The SM 6.7 requirements are bumped down to "texture_shadowlod"/approximately SM 4.1 internally to reduce other SM 6++ restrictions.

The meta functions are taken from the already implemented `SampleCmpLevelZero` functions where a parameter `level` is used for the LODs instead of hardcoded level zero.


csyonghe (approved):
Looks good!

csyonghe (dismissed):

csyonghe (approved):

PR #6006, opened 2025-01-05
Fix GLSL mix(T,T,bool) SPIR-V output

juliusikkala:
`mix(a,b,t)` should be `a` when t=0 and `b` t=1. The cases where `t` is bool have special overloads in `glsl.meta.slang`, which use the OpSelect instruction. The [parameters for OpSelect](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpSelect) were flipped, so `mix(a,b,true)` returned `a` instead of `b`. This PR fixes that.


csyonghe (approved):
LGTM, thanks for the fix!

PR #6017, opened 2025-01-06
OpImageQueryLod requires OpCapability ImageQuery

andreasschultes:
OpImageQueryLod requires OpCapability ImageQuery

Unit test are currently missing for this case.


fairywreath:
Hi - I was working on something that coincidentally also includes these fixes #6018 and also includes tests.

PR #6018, opened 2025-01-06
Add CalculateLevelOfDetail* overloads for comparison samplers

fairywreath:
`CalculateLevelOfDetail*` overloads that accept `SamplerComparisonState` were introduced as part of [SM 6.8's expanded comparison sampling](https://microsoft.github.io/hlsl-specs/proposals/0014-expanded-comparison-sampling.html#detailed-design). This feature is available in GLSL, SPIRV and Metal. The internal implementation is an exact copy of the existing `CalculateLevelOfDetail*` for `SamplerState` - GLSL's textureQueryLod accepts sampler2DShadow as an overload and there is no type difference between cmp and non-cmp samplers for metal and SPIRV. Additionally adds the required OpImageQuery capability required for SPIRV texture lod querying.


csyonghe (dismissed):
Looks good, thank you!

csyonghe:
There are still a couple test regressions that need to be fixed.

csyonghe (approved):

PR #6019, opened 2025-01-06
Pass compiler specific args to GCC and Visual Studio too

juliusikkala:
I was playing around with `-target exe` and noticed that `-Xgenericcpp -w` wasn't silencing the GCC warnings as it's supposed to. Well, the reason was that `compilerSpecificArguments` weren't being added to the command line in the first place. So this fixes that for both GCC and Visual Studio.


csyonghe (approved):
Looks good. Thank you!

PR #6021, opened 2025-01-07
slang-test: Ensure runtime is consistent

aleino-nv:
I hit issue #5305 again after re-generating the build directory, building and then running slang-test.
In other words, D3D12 and Slang-RHI/WebGPU/D3D12 failed to create a device due to the runtime being incosistent. Building copy-prebuilt-binaries fixed it, and so it's best to make this step a dependency of slang-test.


jkwak-work:
I am not sure why this change is needed.
On my machine with ToT, I can do the following,
```
$ cmake.exe --preset vs2019
$ cmake.exe --build --preset release
$ find build -iname d3d12core.dll
build/Release/bin/D3D12/D3D12Core.dll
```

It looks like this commit assumes that `build/Release/bin/D3D12/D3D12Core.dll` doesn't exist with the command, `cmake.exe --preset vs2019 && cmake.exe --build --preset release`.
But it works for my machine and I am wondering why it didn't work for your system.

aleino-nv:
> I am not sure why this change is needed. On my machine with ToT, I can do the following,
> 
> ```
> $ cmake.exe --preset vs2019
> $ cmake.exe --build --preset release
> $ find build -iname d3d12core.dll
> build/Release/bin/D3D12/D3D12Core.dll
> ```
> 
> It looks like this commit assumes that `build/Release/bin/D3D12/D3D12Core.dll` doesn't exist with the command, `cmake.exe --preset vs2019 && cmake.exe --build --preset release`. But it works for my machine and I am wondering why it didn't work for your system.
> 
> With this commit, we are effectively ignoring `SLANG_ENABLE_PREBUILT_BINARIES` setting. I am not sure if we want that.

Try clearing the build directory and then building+running only `slang-test`. I think `d3d12` and `wgpu` will not be supported.
That is what failed for me.

csyonghe (approved):

PR #6022, opened 2025-01-07
Add overload for select to handle Optional<T>

### LINKED ISSUE #6020: No overload of `select` or `operator?:` for Optional<T>
Currently the overloads `select(bool condition, T ifTrue, __none_t ifFalse)` `select(bool condition, __none_t ifTrue, T ifFalse)` don't exist if the stored value is Optional<T>.
Technically it's possible to wrap it this way, `select(true, Optional<T>(some), Optional<T>(none));` but it's not elegant.

mTvare6:
Fixes #6020


source/slang/core.meta.slang
```diff
@@ -727,6 +727,18 @@ __generic<T, let N : int> __intrinsic_op(select) vector<T,N> operator?:(vector<b
 __generic<T> __intrinsic_op(select) T select(bool condition, T ifTrue, T ifFalse);
 __generic<T, let N : int> __intrinsic_op(select) vector<T,N> select(vector<bool,N> condition, vector<T,N> ifTrue, vector<T,N> ifFalse);
 
+[ForceInline]
+__generic<T> __intrinsic_op(select) Optional<T> select(bool condition, __none_t ifTrue, T ifFalse)
```
csyonghe, line 731: functions with `__intrinsic_op` do not need to have a body, we just need put `;` after the signature here.

mTvare6: @csyonghe but select doesn't have a intrinsic overload for none type, does it? I should I might have to enter in the intrinsic part because select is an intrinsic.

mTvare6: Alright, I've removed it to reflect that.

csyonghe (approved):

Issue #5541, opened 2024-11-12
Shader attributes ignored?


Ollhax:
I have a slang file containing two entry points for the vertex and fragment shaders:

```
[shader("vertex")]
VertexOutput vertexMain(VertexInput vertexInput)
{
	...
}

[shader("fragment")]
float4 fragmentMain(CoarseVertex vertex: CoarseVertex) : SV_Target
{
	...
}
```

When I run `slangc shader.slang -target glsl -stage vertex` (or via `-o out.vert`) I expect it to be able to find the `[shader("vertex")]` and figure out that `vertexMain` is the entry point it needs, but instead I get an error:

`shader.slang(3): error 38000: no function found matching entry point name 'main'`

I can rename vertexMain to "main" and it'll work, but it doesn't work if I rename both vertexMain and fragmentMain to "main".

csyonghe:
The commandline tool doesn't respect the predefined entrypoint attributes, but the compilation API can correctly identify them.

If you are using the slangc tool, you need to explicitly specify which entrypoint to compile:
```
slangc shader.slang -target glsl -entry vertexMain -stage vertex
```

Ollhax:
Alright, thanks for the clarification. I can work around it, but it would be nice to have support for entrypoint attributes in the commandline tool. It's easy to trip on, if nothing else.

bmillsNV:
Closing given the documentation update in #6033

cheneym2:
https://github.com/shader-slang/slang/pull/6033 is still in review, re-opening

PR #6033, opened 2025-01-08
[docs] Admonish slangc entry points / shader attributes

Ben-Fields:
Admonish the related non-functional compilation command in the reference manual until #5541 is addressed.


docs/user-guide/08-compiling.md
```diff
@@ -179,6 +179,13 @@ we can compile the `computeMain()` entry point to SPIR-V using the following com
 slangc hello-world.slang -target spirv -o hello-world.spv
 ```
 
+> #### Note ####
+> The `slangc` CLI [does not currently support](https://github.com/shader-slang/slang/issues/5541) `[shader(...)]` attributes. Complile the above example with the following command instead:
+> 
+> ```bat
+> slangc hello-world.slang -target spirv -entry computeMain -o hello-world.spv
```
csyonghe, line 186: This is not completely true though. `slangc hello-world.slang -target spirv` should work just fine. You need to specify the entry point explicitly only when targeting HLSL and GLSL, but not SPIRV.

natevm: +1, I use this feature regularly. Might be good to clarify what does or doesn't work, but slangc does support at least some mixed entry point work. 

I thought at one point we discussed supporting mixed entry points via partially-compiled slang modules too. I haven't really investigated this though.

Ben-Fields: Updated.

docs/user-guide/08-compiling.md
```diff
@@ -179,6 +179,13 @@ we can compile the `computeMain()` entry point to SPIR-V using the following com
 slangc hello-world.slang -target spirv -o hello-world.spv
 ```
 
+> #### Note ####
+> The `slangc` CLI [does not currently support](https://github.com/shader-slang/slang/issues/5541) `[shader(...)]` attributes for targets other than SPIRV. For other targets, the `-entry` argument is mandatory. Example:
```
csyonghe, line 183: I don't think it is accurate to say `slangc` doesn't support `[shader()]` attribute because shaders with `[shader()]` attribute will be correctly compiled by slangc and the attribute will be reflected in the resulting code.

It is just that for targets without multi-entrypoint support, `-stage` or `-profile` argument is still needed even if the shader contains `[shader()]` attribute when using `slangc`.

Ben-Fields: Updated. Please suggest alternate wording if still preferred.

natevm: I believe that OptiX also supports multiple entry points via the [shader()] attribute. Might be better to more specifically list the targets that are currently known to not support multiple entry points.

csyonghe: How about 
> The `slangc` CLI currently cannot automatically deduce `-entrypoint` and `-stage`/`-profile` options from the `[shader(...)]` attributes when generating code for targets other than SPIRV, Metal, CUDA, or Optix. For targets such as HLSL, please continue to specify `-entry` and `-profile` options even when compiling a file with `[shader(...)]` attributes on the entrypoints.

cheneym2:
Is this patch still needed?

docs/user-guide/08-compiling.md
Ben-Fields: Updated with specified targets. Added hyperlinks.

Ben-Fields:
Updated per feedback.

expipiplus1 (approved):

PR #6035, opened 2025-01-08
Fix implicit string conversion breaking NativeStrings

### LINKED ISSUE #6034: String -> NativeString conversion incorrectly omitted
`NativeString`s are incorrectly assumed to be equivalent to `String`s, without going through the proper constructors. This causes compilation errors in e.g. the following program:

```hlsl
void string_conversion(String title)
{
    NativeString c_title = NativeString(title);
    printf(c_title);
}

export __extern_cpp int main(int argc, int8_t** argv)
{
    string_conversion("asdf");
    return 0;
}
```

Compiled with `slangc -target exe test.slang -o test`:

```
gcc 13.3: testi.slang(3): error :  cannot convert ‘Slang::String’ to ‘const char*’
    3 | void string_conversion(String title)
      |            ^~~~~~~
      |            |
      |            Slang::String
gcc 13.3: /usr/include/x86_64-linux-gnu/bits/stdio2.h(84): note :    initializing argument 1 of ‘int printf(const char*, ...)’
   84 | printf (const char *__restrict __fmt, ...)
      |         ~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

The generated code looks like this:

```
#line 1 "test.slang"
static void string_conversion_0(String title_0)
{

#line 1
    printf(title_0);



    return;
}


#line 7
int32_t main(int32_t argc_0, int8_t * * argv_0)
{
    string_conversion_0(Slang::toTerminatedSlice("asdf"));
    return int(0);
}
```

I have a PR lined up for this, but I'm not 100% sure if it's the correct solution, so I made this issue to keep track of this in case the PR isn't the right way to fix this.

juliusikkala:
Fixes #6034 by replacing the existing conversion code (that assumes the expressions are equivalent) by going through core.meta.slang and using the already-existing intrinsic ops for the conversion instead.


csyonghe (approved):

juliusikkala:
Locally, the tests were also failing in debug mode, seems I was careless. I reinstated the coercion but only for string literals, that should be safe both for CPU and non-CPU targets. I also added a test to make sure that passing strings around doesn't break again.

csyonghe (approved):

PR #6040, opened 2025-01-09
Disambiguate 'target' anchors in slangc reference

AndrewBrownK:
There is a misleading/confusing link in the slangc reference when read on GitHub ([here](https://github.com/shader-slang/slang/blob/master/docs/command-line-slangc-reference.md)).

There are a few different links/anchors identified as "target" (try case insensitive search `id="target` while editing the file). One of these sections talks about the `-target` option itself (`id="target-1"`). Another section talks about other options related to target code generation (`id="Target"` with capital T). The last lists the specific values that you can feed to the `-target` option (`id="target"` with lowercase t). 

This might work as intended in some markdown viewers/editors (I don't know). However, GitHub converts the element IDs to lowercase and so two of these links are ambiguous. Thus when reading this documentation and seeking valid options for `-target`, clicking the link to find those options leads to the wrong section.

In this PR I have changed the title and id of `Target` to `Target Options`, so now each target related hyperlink works as intended. This can be tested on my branch [here](https://github.com/AndrewBrownK/slang/blob/disambiguate-target-anchor/docs/command-line-slangc-reference.md#target). I cannot attest/confirm that this is bug free in other markdown viewers/editors.

I'm not sure why GitHub is showing this diff is a binary file/change.


csyonghe:
This markdown file is generated by running `slangc` so we need to fix the docgen logic in the compiler code instead of fixing directly the md file.

Should change `source\slang\slang-options.cpp` L114.

jhelferty-nv:
@AndrewBrownK Have you had a chance to read Yong's review comments?

mkeshavaNV:
Setting to draft since there has been no recent activity on this here. @AndrewBrownK - please set it to ready when this is ready for another round of review. Thanks.

PR #6042, opened 2025-01-09
Fix Metal type layout reflection for nested parameter blocks.

csyonghe:
When computing layout for Metal target for the following type using MetalArgumentBuffer layout:
```
struct Params
{
     ParameterBlock<Data> data;
     Texture2D t;
}
```
`Params::t` should receive `id=1`, but due to a bug around parameter group type layout, if struct field ha ParameterBlock type and if the ParameterBlock consumes any actual non space layout units, they are not accounted for during construction of a struct layout. This PR fixes this issue so we can correct report t at id = 1.


source/slang/slang-type-layout.cpp
```diff
@@ -2351,7 +2362,9 @@ static SimpleLayoutInfo _getParameterGroupLayoutInfo(
         //
         // TODO: wouldn't it be any different to just allocate this
```
cheneym2, line 2363: Is this TODO still relevant?

csyonghe: I have not tested this yet. I believe there are reasons why things are done this way but I didn't have time to trace to the bottom of this.

PR #6047, opened 2025-01-09
Fix potential test failures due to SPIRV validation failure

jkwak-work:
There are a few tests that are currently passing but could fail when conditions change little bit.

When slang-test runs with `test-server`, the spirv validation fails and the test gets reported as failed even though they are actually passing.

To avoid the potential problem, this commit adds `-skip-spirv-validation` option to avoid unnecessary error.


csyonghe:
We shouldn't be generating invalid spirv though so it seems like we need to properly fix these issues.

jkwak-work:
> We shouldn't be generating invalid spirv though so it seems like we need to properly fix these issues.

I will add empty entry points to avoid the validation error.

csyonghe (approved):

PR #6053, opened 2025-01-10
`[[shader_record]] / [[push_constant]]` Param Annotations Proposal

natevm:
When reviewing documentation on the current parameter block specification, @tangent-vector and I discussed some of the current inconsistencies regarding the use of `uniform` to specify the rate of change of parameter bindings in ray tracing pipelines. 

The desired usage of `uniform` is to map to `push_constants` on all entry point types when targeting SPIR-V---however, this requires a breaking change of the existing behavior regarding uniform mapping to shader record annotation. 

At the moment, I'm proposing `[[shader_record]]` and `[[push_constant]]`, as we already have these annotation today. They just don't seem to be respected when used in the parameters of an entry point. 

I'm also proposing that we allow users to drop `uniform` when these attributes are specified, since `uniform` becomes implied by these more rate-specific annotations. 

Beyond uniform, I think we should consider similar annotations to disambiguate payload registers and hit attributes. As I show at the end of the proposal, it's pretty easy for code to unintentionally map to the wrong system values.


csyonghe (approved):
These are positive changes to the language, and is a good step in removing ambiguity.
This is also necessary to allow using push constants in ray tracing entrypoints.

csyonghe:
Can you update the proposal number to 017?

natevm:
Hm, didn’t mean to dismiss any reviews 😅

Should be updated now.

csyonghe (approved):

PR #6054, opened 2025-01-10
Fix `markNonContextParamsAsSideEffectFree`.

### LINKED ISSUE #6039: Values accessed via properties not saved for backwards pass
We're using properties for access control so that values can be supplied to downstream code without them accidentally getting stomped on.

However, it looks like values accessed through properties are not properly saved in the primal intermediates when the backwards pass is generated and are replaced by 0. In the following code, `write_bwd` only ever sees 0 for `idx`, rather than the expected value of 5, and accesses the wrong memory location.

```
public struct ReadOnlyIndex
{
    private int _idx;
    public property int idx { get { return _idx; } }
}
struct GradientBuffer
{
    RWStructuredBuffer<float> primal;
    StructuredBuffer<float> grad;

    [Differentiable]
    void write(int idx, float v) { primal[idx] = detach(v); }

    [BackwardDerivativeOf(write)]
    void write_bwd(int idx, inout DifferentialPair<float> d) { d = diffPair(d.p, grad[idx]); }

    [Differentiable]
    void store(ReadOnlyIndex idx, float v) { write(idx.idx, v); }
}
[Differentiable]
void test(GradientBuffer buf, ReadOnlyIndex b, float x)
{
    buf.store(b, x);
}
public float repro(RWStructuredBuffer<float> primal, StructuredBuffer<float> grad)
{
    DifferentialPair<float> result = diffPair(1.0f);
    GradientBuffer buf = { primal, grad };
    bwd_diff(test)(buf, { 5 }, result);
    return result.d;
}
```

### LINKED ISSUE #6041: Trivial primal function causes backward function to be optimized away
In the following snippet, a buffer is wrapped to supply gradients for a backwards pass. This is done with a custom backward derivative `write_bwd` and an empty primal `write` to avoid side effects.

However, it looks both `write` and `write_bwd` are optimized away, and the code compiles to a noop. Replacing `write` with a non-trivial body causes `write_bwd` to be run as expected.

```
struct GradientBuffer
{
    StructuredBuffer<float> grads;

    [Differentiable]
    void write(int idx, float value) { /* Discard write */ }

    [BackwardDerivativeOf(write)]
    void write_bwd(int idx, inout DifferentialPair<float> d)
    {
        d = diffPair(d.p, grads[idx]);
    }
}

[Differentiable]
void test(GradientBuffer dst, int idx, float v)
{
    dst.write(idx, v);
}

uniform StructuredBuffer<float> grad_in;
uniform RWStructuredBuffer<float> grad_out;

[shader("compute")]
void repro()
{
    GradientBuffer grads = { grad_in };
    DifferentialPair<float> result = diffPair(1.0f);
    bwd_diff(test)(grads, 0, result);
    grad_out[0] = result.d; // Should return grad_in[0], but returns 0.0f instead
}
```

### LINKED ISSUE #6048: [MaxIters(N)] causes variables to not be saved for backwards pass
This is possibly related to #6039, with similar behavior but a different trigger. Using `[MaxIters]` in the following example causes the parameter `context` of `store` to not be saved during primal execution. `write_bwd` in the backwards pass only sees a value of `0` for `idx[0]` and reads from the wrong memory location. Replacing `[MaxIters]` with `[ForceUnroll]` fixes the issue.

```
struct GradientBuffer<let D : int>
{
    RWStructuredBuffer<float> primal;
    StructuredBuffer<float> grad;
    int strides[D];

    int toIndex(int idx[D]) {
        int result = 0;
        for (int i = 0; i < D; ++i)
            result += strides[i] * idx[i];
        return result;
    }

    [Differentiable]
    void write(int[D] idx, float v) { primal[toIndex(idx)] = detach(v); }

    [BackwardDerivativeOf(write)]
    void write_bwd(int[D] idx, inout DifferentialPair<float> d) { d = diffPair(d.p, grad[toIndex(idx)]); }

    [Differentiable]
    void store<let N : int>(int context[D - 1], in float value[N])
    {
        int idx[D];
        //[ForceUnroll] /* Using ForceUnroll instead of MaxIters makes it work */
        [MaxIters(2)]
        for (int i = 0; i < D - 1; ++i)
            idx[i] = context[i];
        [ForceUnroll]
        for (int i = 0; i < N; i++) {
            idx[D - 1] = i;
            write(idx, value[i]);
        }
    }
}

[Differentiable]
void test(GradientBuffer<2> buf, int[1] base, float[3] value)
{
    buf.store(base, value);
}

[shader("compute")]
float3 repro(RWStructuredBuffer<float> primal, StructuredBuffer<float> grad)
{
    float input[3];
    input[0] = input[1] = input[2] = 1.0f;
    var result = diffPair(input);
    GradientBuffer<2> buf = { primal, grad, {3, 1} };
    bwd_diff(test)(buf, { 1 }, result);
    return float3(result.d[0], result.d[1], result.d[2]);
}

```

csyonghe:
`markNonContextParamsAsSideEffectFree` is supposed to mark all parameters except the intermediate context parameter as `IgnoreSideEffect`. However, the function is not implemented correctly. It checks if a parameter is an intermediate context by checking if its type is `IRIntermediateContextType`. However at the time this function is called, the intermediate context type is already lowered into a `struct`, so the current implementation will always mark all parameters as `IngoreSideEffect`. which result in the compiler to aggressively remove/ignore many primal function calls during autodiff pass when it shouldn't.

This PR fixes this bug by explicitly marking the intermediate context parameter with a `PrimalContext` decoration, and instead of checking parameter types in `markNonContextParamsAsSideEffectFree`, we simply look for this decoration to test if a parameter is an intermediate context.

Closes #6039.
Closes #6041.
Closes #6048.


jkwak-work (approved):
Looks good to me

Issue #6024, opened 2025-01-07
Reflection returns invalid size for OptixTraversableHandle


skallweitNV:
Reflection returns a uniform size/stride of 16 for `OptixTraversableHandle`. The actual size should be 8 bytes.
Maybe 16 was correct in older versions of OptiX?

PR #6055, opened 2025-01-10
Fix CUDA reflection for acceleration structure handle size.

csyonghe:
Closes [#6024](https://github.com/shader-slang/slang/issues/6024).


jkwak-work (dismissed):
Looks good to me

PR #6056, opened 2025-01-10
Add exhaustive associativity test for WGSL

### LINKED ISSUE #6005: wgsl output doesn't use parenthesis for mixed operations, generating invalid code
See https://github.com/gfx-rs/wgpu/issues/6857

I got this error in Chromium

```
Compilation log for [Invalid ShaderModule (unlabeled)]:
6 error(s) generated while compiling the shader:
:67:54 error: mixing '+' and '^' requires parenthesis
        var _S4 : u32 = _S3 + (((((_S2 << (u32(4)))) + u32(2738958700) ^ (_S2 + sum_1)) ^ (((_S2 >> (u32(5)))) + u32(3355524772))));
```

slang code: https://github.com/NVIDIAGameWorks/Falcor/blob/9dc819c162b2070335c65060436041690b7937f8/Source/Falcor/Utils/Math/HashUtils.slang#L62-L74

aleino-nv:
- Add lots of tests to check how binary operators associate in various cases.
- Add a fix to the C-like emitter to make them all pass.

This closes #6005.


aleino-nv:
I realize the test case is a bit ridiculous, but I managed to catch lots of missing parentheses with it.

tests/wgsl/associativity.slang
```diff
@@ -0,0 +1,952 @@
...
+    // bb = ['&&', '||', '==', '!=']
+    // rl = ['<', '>', '<=', '>=']
+    // eq = ['==', '!=']
+    // lg = ['&&', '||']
+    //
+    // iidx = 0
+    // oidx = 0
+    //
+    // for (l, r) in pairs(nn, nn):
+    //     print("na = input[{}];".format(iidx + 0))
```
cheneym2, line 37: I think the test would be slightly easier to comprehend if the input values were all inlined instead of indirectly retrieved from the RWStructuredBuffer data.  Or even 

```
    na = 3;
    nb = 7;
    nc = 8;
    output[0] = na ^ nb ^ nc;
```

or even

`    output[0] = 3 ^ 7 ^ 8;`

cheneym2 (approved):

tests/wgsl/associativity.slang
aleino-nv: If they were constants that would probably just get calculated at compile time and it would defeat the purpose of the test. `na ^ nb ^ nc` would look nicer but I don't think it's possible to trick the compiler into generating that.

tests/wgsl/associativity.slang
```diff
@@ -0,0 +1,952 @@
...
+    nb = input[460];
+    nc = input[461];
+    output[206] = uint(na << nb != nc);
+    output[207] = uint(na != nb << nc);
+    na = input[462];
+    nb = input[463];
+    nc = input[464];
+    output[208] = uint(na >> nb < nc);
+    // error 39999: expected a generic when using '<...>' (found: 'uint')
+    // output[209] = uint(na < nb >> nc);
```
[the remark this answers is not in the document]
aleino-nv: Kudos for catching that. :)
Yes, I suppose I should file an issue for it. I just wanted to convince myself it's not the intended grammar and we're supposed to work around it.

tests/wgsl/associativity.slang
cheneym2: That makes sense, good call.

PR #6058, opened 2025-01-10
Feature/initialize list side branch

kaizhangNV:


docs/proposals/004-initialization.md
```diff
@@ -129,6 +129,39 @@ MyType x = MyType(y); // equivalent to `x = y`.
 
 The compiler will attempt to resolve all type casts using type coercion rules, if that failed, will fall back to resolve it as a constructor call.
 
+### Inheritance Initialization
+For derived struct, slang will synthesized the constructor by bring the parameters from the base struct's constructor if the base struct also has a synthesized constructor. For example:
```
cheneym2, line 133: Suggest: "For derived structs"
"constructor by bringing"

csyonghe:
We still need to fix the CUDA export error before we can land this PR.

tests/autodiff/differential-type-constructor.slang
```diff
@@ -14,12 +14,12 @@ RWStructuredBuffer<float> outputBuffer;
 void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
 {
     {
-        var dp = diffPair(MyStruct(), MyStruct.Differential());
+        var dp = diffPair(MyStruct(0.0, 0), MyStruct.Differential(0.0));
 
         outputBuffer[0] = dp.d.a;
 
-        var dp2 = diffPair<MyStruct>(MyStruct(), MyStruct.Differential(1.f));
+        var dp2 = diffPair<MyStruct>(MyStruct(0.0, 0), MyStruct.Differential(1.f));;
```
csyonghe, line 21: double semicolon

csyonghe (commented):
We need to add tests for all the scenarios discussed in the proposal, and ensure they all have correct code/error generated.

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+    }
+
+    return nullptr;
+}
+
+bool SemanticsVisitor::_cStyleStructBasicCheck(Decl* decl)
+{
+    // 1. It has to be a user-defined struct type, or a basic scalar, vector or matrix type
+    StructDecl* structDecl = nullptr;
+    if (isFromCoreModule(decl))
```
csyonghe, line 240: Is this saying all structs from core module are not c-style? This isn't true if this function is called with a structDecl representing the Vector/Matrix type declaration.

In general, we should avoid `isFromCoreModule` checks in our frontend, unless absolutely necessary.

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+    {
+        getShared()->cacheCStyleStruct(structDecl, false);
+        return false;
+    }
+
+    // 5. All its members are legacy C-Style structs or arrays of legacy C-style structs
+    for (auto varDecl : structDecl->getMembersOfType<VarDeclBase>())
+    {
+        // if the member is an array, check if the element is legacy C-style rule.
+        if (auto arrayType = as<ArrayExpressionType>(varDecl->getType()))
```
csyonghe, line 303: Unsized arrays shouldn't be C-style type no matter what the element type is.

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+    {
+        if (getDeclVisibility(varDecl) != structVisibility)
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool SemanticsVisitor::isCStyleStruct(StructDecl* structDecl)
```
csyonghe, line 284: It's weird to have the array case separated out from the structs.

I think the code should be restructured to recurse on type instead of decls. Basically you just need:
```
bool isCStyleType(Type* type);
```
An array can be a C-style type if it is sized and the element type is cstyle. A struct can be C-style if all the 5 conditions are met, etc.

This way we can easily extend the logic to cover more types than just structs as c-style if we want, and we can easily return false if the type is scalar,vector,matrix or enum types.

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+    return constructorExpr;
+}
+
+// translation from initializer list to constructor invocation if the struct has constructor.
+bool SemanticsVisitor::_invokeExprForExplicitCtor(
+    Type* toType,
+    InitializerListExpr* fromInitializerListExpr,
+    Expr** outExpr)
+{
+    if (auto toStructDeclRef = isDeclRefTypeOf<StructDecl>(toType))
```
csyonghe, line 371: if we have `isCStyleType`, this just become `if (isCStyleType(toType))`

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+            getSink()->diagnoseRaw(
+                Severity::Error,
+                static_cast<char const*>(blob->getBufferPointer()));
+        }
+        return false;
+    };
+
+    if (ctorInvokeExpr)
+    {
+        // The reason we need to check the coercion again is that the ResolveInvoke() could still
```
csyonghe, line 454: I think the right thing to do here is that when we lookup for constructor candidates from a type, we should exclude all base decls in the lookup, because it makes no sense to call a constructor from base to construct a derived type.

In `SemanticsVisitor::AddTypeOverloadCandidates`, we should setup the lookup context to include `IgnoreInheritance` option.

This way, we will never resolve the invoke expr to call the base ctor, and we can remove this additional checking logic here.

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+            {
+                *outExpr = ctorInvokeExpr;
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+bool SemanticsVisitor::_invokeExprForSynthesizedCtor(
```
csyonghe, line 410: should be named `createInvokeExprForSynthesizedCtor`

source/slang/slang-check-decl.cpp
```diff
@@ -8013,6 +8139,40 @@ SemanticsContext SemanticsDeclBodyVisitor::registerDifferentiableTypesForFunc(
 
 void SemanticsDeclBodyVisitor::visitFunctionDeclBase(FunctionDeclBase* decl)
 {
+    if (auto constructorDecl = as<ConstructorDecl>(decl))
+    {
+        // When checking the synthesized constructor, it's possible to hit error, but we don't want
```
csyonghe, line 8144: Checking and then removing a ctor seems dangerous, because the following sequence may be theoretically possible:
1. The ctor signature is synthesized.
2. Some other code that calls the ctor is being checked and resolved into a call to this ctor.
3. The ctor is being synthesized and found errors, thus removed.
4. Now that  other code will contain a call to a non-existent function.

How can we make sure this does not happen?

What are the cases when ctor synthesis can fail?

docs/proposals/004-initialization.md
kaizhangNV: ack

tests/autodiff/differential-type-constructor.slang
kaizhangNV: ack.

source/slang/slang-ast-decl.h
```diff
@@ -165,6 +165,12 @@ class AggTypeDecl : public AggTypeDeclBase
 class StructDecl : public AggTypeDecl
 {
     SLANG_AST_CLASS(StructDecl);
+
+    SLANG_UNREFLECTED
+    // We will use these auxiliary to help in synthesizing the member initialize constructor.
+    Slang::HashSet<VarDeclBase*> m_membersVisibleInCtor;
+    Dictionary<int, ConstructorDecl*> m_synthesizedCtorMap;
+    bool m_hasExplicitCtorInExtension = false;
```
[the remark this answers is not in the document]
kaizhangNV: This is not to prevent synthesize the ctor, but to favor the calling of the explicit ctor defined in extension during 
`_coerceInitializerList`.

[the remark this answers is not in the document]
kaizhangNV: Well, this seems like a corner case we haven't discussed before.
I thought extension will be part of the struct, so it can also determine whether we should define the ctor.

The thing is that if the syntheized ctor is exactly the same as the explicit ctor in extension, won't that be a compile error?

source/slang/slang-check-conversion.cpp
```diff
@@ -203,6 +203,292 @@ DeclRef<StructDecl> findBaseStructDeclRef(
...
+
+static StructDecl* _getStructDecl(Type* type)
+{
+    if (as<VectorExpressionType>(type) || as<MatrixExpressionType>(type) ||
+        as<ArithmeticExpressionType>(type) || as<BuiltinType>(type))
+    {
+        return nullptr;
+    }
+
+    if (auto declRefType = as<DeclRefType>(type))
```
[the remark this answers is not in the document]
kaizhangNV: ack

source/slang/slang-check-conversion.cpp
kaizhangNV: ack

source/slang/slang-check-conversion.cpp
kaizhangNV: ack.

source/slang/slang-check-conversion.cpp
kaizhangNV: I actually tried this option very early, but when I tested with the example like this;

```
struct Foo
{
    float3 a;
}

Foo test()
{
    float16_t3 a;
    Foo foo = {a};
    return foo;
}
```

it complains about
```
test1.slang(49): error 30019: expected an expression of type 'float', got 'vector<half,3>'
    Foo foo = {a};
               ^
```

I haven't digged into the root cause, because current solution is working.
But I will take another look into this issue.

kaizhangNV: Fixed in the latest commit.

source/slang/slang-ir-check-differentiability.cpp
```diff
@@ -249,13 +249,25 @@ struct CheckDifferentiabilityPassContext : public InstPassBase
         if (outerFuncInst->findDecoration<IRTorchEntryPointDecoration>())
             return;
 
+        bool isSynthesizeConstructor = false;
```
csyonghe, line 252: Instead of working around here, I think the right solution may just be that we don't synthesize any constructors for types that have `__intrinsic_type` modifier on it, since these types maps t o special IR types and synthesized constructors for these types are not meaningful.

This way we wouldn't be synthesizing any constructor for the `TorchTensor` type, and won't run into this problem.

source/slang/slang-check-overload.cpp
```diff
@@ -2175,7 +2175,8 @@ void SemanticsVisitor::AddTypeOverloadCandidates(Type* type, OverloadResolveCont
         type,
         context.sourceScope,
         LookupMask::Default,
-        LookupOptions::NoDeref);
+        context.checkForSynthesizedCtor ? LookupOptions::IgnoreInheritance
```
csyonghe, line 2178: why does this need to be conditional on `checkForSynthesizedCtor`, instead of just always ignore inheritance for all constructor lookups?

source/slang/slang-ir-check-differentiability.cpp
kaizhangNV: This is not just handling that case, if there is a struct having a TorchTensor type members, we will still have to synthesize constructor for that struct.

source/slang/slang-check-overload.cpp
kaizhangNV: Though I'm not 100% sure, it could be the reason on the comment, this lookup is also used by the coerce logic.

e.g.:
 `float3 <= vector<half, 3>`

with only `LookupOptions::IgnoreInheritance`, the coerce could fail.

kaizhangNV: I think maybe this is a bug but I thought it's valid,

e.g.
```
struct Foo
{
    float3 a;
    __init(float3 a) {this.a = a;}
}

float16_t3 x;
Foo f = Foo(x);
```
Is this a valid syntax? Can `float16_t3` implicitly converts to `float3`?

kaizhangNV: The reason I'm asking is because currently Falcor is using such syntax in their code base, but we don't treat this like an error. So that's why I'm thinking this is a valid syntax.

csyonghe: I wonder why it will fail. These conversions should be defined in extensions and not in base types. IgnoreInheritance shouldn't be skipping ctors from extensions, right?

csyonghe: In your example, `Foo(x)` is valid, but implicit conversions between float16_t3 and Foo should be invalid.
e.g. if you have a function that `helper(Foo x)` and you pass a float16_t3 to helper, it should be an error.

csyonghe: But I still don't see why these two issues are related. In this scenario, if we try to coerce float16_t3 to Foo, we will be looking for a Foo::__init, and we shouldn't be looking for __init in any base classes of Foo, right?

kaizhangNV: Assuming this
`Foo f = {x};`

it will look up the explicit ctor, and it will find it, then we will coerce each argument to parameter type,
so coerce will trigger this for type `float3`. If the option is still `LookupOptions::IgnoreInheritance`, the coerce will fail.

But unfortunately, I don't know the reason yet.

csyonghe: I think it is worth finding out, because float3 does not have a base type defines float3.__init(float16t_3), and such constructors should be discoverable even with IgnoreInheritance.

A possible reason is that the ctor is defined in a generic extension on BuiltinArithmeticType, and somehow IgnoreInheritance is causing us to ignore extensions defined on some base interface type.

kaizhangNV: Yes, I think the root cause is that it prevent us from find those ctors defined in extension

they'are defined in core module
```
            sb << "    __implicit_conversion(" << cost << ")\n";
            sb << "    __intrinsic_op(" << int(op) << ")\n";
            sb << "    __init(vector<" << kBaseTypes[ff].name << ",N> value);\n";
```

kaizhangNV: And I did some cross validation, revert my change, and look at the the best candidate.

csyonghe: Are you saying IgnoreInheritance is ignoring all extensions? If so, we need a different option that is ignoring just the base types.

source/slang/slang-ir-check-differentiability.cpp
csyonghe: I see, this really starting to get into serious heterogeneous programming models if we need to be correct. I can see us using this workaround for now until we get better support for heterogeneous code.

source/slang/slang-ast-decl.h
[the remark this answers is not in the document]
csyonghe: Yes, this what I was saying that during overload resolution, if we see a user defined ctor that is applicable, it should always be preferred over implicitly generated ones.

The one reason that ctor synthesis shouldn't depend on extensions is if a module defines public X and an internal extension on X that adds a ctor, should we synthesize ctor for X or not? When this decision is affected by extensions, we will say we don't. But then from the user of the module's perspective this is weird because they don't get to see the extension and yet X for some reason doesn't have the expected ctor.

If we say the decision is not dependent on extensions, then we will always synthesize the ctor. But inside the module that sees both the extension and X itself, the user defined ctor will be preferred over the synthesized one if it is applicable, so we will have no issues. From the outside of the module, they still can use the synthesized ctor without surprises.

source/slang/slang-check-overload.cpp
csyonghe: This is a bug then. We need to make IgnoreInheritance work as expected. The `__init` in extensions should be discoverable in all cases.

I looked at `_lookupMembersInSuperTypeFacets` and found this:

```
else if (
    int(request.options) & int(LookupOptions::IgnoreInheritance) &&
    (facet.getImpl()->directness != Facet::Directness::Self &&
     (!extensionFacet || !extensionFacet->targetType.type->equals(selfType))))
{
    continue;
}
```

This is not entirely correct. If we have `void test<T:IFoo>(T param) { param = T(); }`, we should at least allow lookup in the `IFoo` facet. Also need to make sure:
```
interface IFoo {}
extension<T:IFoo> T {__init(x); }

struct X : IFoo{};
void t() { X x = X(...);  } // should also be discoverable.
```

kaizhangNV: Issue is fixed in #6146

source/slang/slang-check-conversion.cpp
kaizhangNV: act in latest commit.

source/slang/slang-check-decl.cpp
```diff
@@ -12332,8 +12295,8 @@ void SemanticsDeclAttributesVisitor::_synthesizeCtorSignature(StructDecl* struct
 
     // Only the members whose visibility level is higher or equal than the
     // constructor's visibility level will appear in the constructor's parameter list.
-    List<VarDeclBase*> resultMembers;
-    if (!_searchMembersWithHigherVisibility(structDecl, ctorVisibility, resultMembers))
+    List<KeyValuePair<VarDeclBase*, String>> resultMembers;
```
csyonghe, line 12298: I don't think it is necessary for `_searchInitializableMembers` to return the name of the enclosing struct, because you can easily obtain the struct decl from `VarDeclBase->getParent`.

tests/language-feature/interfaces/zero-init-interface.slang
```diff
@@ -26,7 +27,9 @@ struct MyType
 [numthreads(1, 1, 1)]
 void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
 {
-    MyType t = {};
+    // Note 'MyType' is not a C-Style struct because it has an interface typed member.
+    // slang will synthesize a constructor for it. __init(IFoo foo)
+    MyType t = {{}};
```
csyonghe, line 32: We don't need to clean this up in  this PR, but I think using `{}` to initialize a `IFoo` typed variable/member is ill-formed. It doesn't have a meaning. A `IFoo` variable must be initiialized with something that has a concrete type.

source/slang/slang-check-overload.cpp
```diff
@@ -2565,7 +2563,7 @@ Expr* SemanticsVisitor::ResolveInvoke(InvokeExpr* expr)
     // type coercion.
     bool typeOverloadChecked = false;
 
-    if (expr->arguments.getCount() == 1)
+    if (expr->arguments.getCount() == 1 && m_checkForSynthesizedCtor == false)
```
csyonghe, line 2566: style: `!m_checkForSynthesizedCtor` for simplicity.
But why is `m_checkForSynthesizedCtor` needed at all?

source/slang/hlsl.meta.slang
```diff
@@ -20793,11 +20793,12 @@ ${
...
         case hlsl:
             uint isSingleLod = 0;
-            Footprint footprint =  {__queryFootprint$(CoarseOrFine)NVAPI(
+            __queryFootprint$(CoarseOrFine)NVAPI(
                 Shape.dimensions,
                 __getRegisterSpace(this), __getRegisterIndex(this),
                 __getRegisterSpace(sampler), __getRegisterIndex(sampler),
-                __vectorReshape<3>(coords), granularity, /* out */isSingleLod), false};
+                __vectorReshape<3>(coords), granularity, /* out */isSingleLod);
+            Footprint footprint = {false};
```
csyonghe, line 20801: It seems that this change is not equivalent to the original code. The return value of `__queryFootprint` is dropped.

source/slang/slang-check-decl.cpp
kaizhangNV: There could a case that this varDeclBase is from base struct's ctor parameter list.
so varDeclBase->parentDecl will be the "__init", I have to use 
varDeclBase->parentDecl->parentDecl.

source/slang/slang-check-decl.cpp
kaizhangNV: Removing this in the latest commit.

source/slang/slang-check-decl.cpp
csyonghe: You can call `getParentAggTypeDecl(decl)` for this.

source/slang/hlsl.meta.slang
kaizhangNV: This is working previously in the legic initializer list because it initialize the base as well, looking at the definition:
struct TextureFootprint<let ND:int> : __TextureFootprintData<ND>
{
    bool _isSingleLevel;

    property isSingleLevel : bool
    {
        [__NoSideEffect]
        get
        {
            return _isSingleLevel;
        }
    }
}

struct __TextureFootprintData<let ND:int>
{
    // effectively empty struct
}

so 
`Footprint footprint = {    return of __queryFootprint,       false}; `
was working.

But it doesn't work now because it's not a c-style type anymore because we don't allow inheritance from a type other than interface.

source/slang/slang-check-decl.cpp
kaizhangNV: ack

tests/language-feature/interfaces/zero-init-interface.slang
kaizhangNV: reverted it back already.

source/slang/slang-ast-decl.h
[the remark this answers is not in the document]
kaizhangNV: done.

source/slang/slang-check-overload.cpp
kaizhangNV: I remember that this could bring us to a infinite loop, because this is a special case handling for type cast, but we  get here because of ctor invoke from `_coerceInitializerList`. So if we go into the coerce logic here, it will finally go to _coerceInitializerList again and form a infinite loop.

I can't remember exactly which test catch this issue.

source/slang/slang-ast-expr.h
```diff
@@ -135,6 +135,9 @@ class InitializerListExpr : public Expr
 {
     SLANG_AST_CLASS(InitializerListExpr)
     List<Expr*> args;
+
+    SLANG_UNREFLECTED
+    bool m_synthesizedForTypeCastZero = false;
```
csyonghe, line 140: This should be reflected, and let's rename it to be more descriptive for what it does, how about `useCStyleInitialization`?

source/slang/slang-check-decl.cpp
```diff
@@ -28,6 +28,8 @@ static List<ConstructorDecl*> _getCtorList(
     SemanticsVisitor* visitor,
     StructDecl* structDecl,
     ConstructorDecl** defaultCtorOut);
+static Expr* constructDefaultInitExprForType(SemanticsVisitor* visitor, VarDeclBase* varDecl);
+void addVisibilityModifier(ASTBuilder* builder, Decl* decl, DeclVisibility vis);
```
csyonghe, line 32: We should have access to ASTBuilder directly via `this`?

source/slang/slang-check-decl.cpp
```diff
@@ -94,6 +96,18 @@ struct SemanticsDeclAttributesVisitor : public SemanticsDeclVisitorBase,
...
     void visitVarDecl(VarDecl* varDecl) { checkVarDeclCommon(varDecl); }
+
+    // Synthesize the constructor declaration for a struct during header visit, as we
+    // need to have such declaration first such that the overloading resolution can lookup
+    // such constructor and complete the initialize list to constructor translation.
+    //
+    // We will defer the actual implementation of the constructor to the body visit, because
+    // we will have full information about each field in the struct during that stage.
+    void _synthesizeCtorSignature(StructDecl* structDecl);
+    bool _searchInitializableMembers(
```
csyonghe, line 107: naming: `collectInitializableMembers`

source/slang/slang-check-decl.cpp
```diff
@@ -331,6 +346,7 @@ struct SemanticsDeclBodyVisitor : public SemanticsDeclVisitorBase,
             StructDecl* parent,
             const bool getOnlyDefault)
         {
+            this->parent = parent;
```
csyonghe, line 349: avoid parameter names that collide with member names. can rename the parameter to `inParent`.

source/slang/slang-check-decl.cpp
```diff
@@ -2241,7 +2310,7 @@ void SemanticsDeclBodyVisitor::checkVarDeclCommon(VarDeclBase* varDecl)
     if (getOptionSet().hasOption(CompilerOptionName::ZeroInitialize) && !varDecl->initExpr &&
         as<VarDecl>(varDecl))
     {
-        varDecl->initExpr = constructDefaultInitExprForVar(this, varDecl);
+        varDecl->initExpr = constructDefaultInitExprForType(this, varDecl);
```
csyonghe, line 2313: Can be done in a later PR, but we should remove zero init logic from the front-end.

tools/gfx/gfx.slang
```diff
@@ -1718,7 +1725,7 @@ public struct SlangDesc
 public struct ShaderCacheDesc
 {
     // The root directory for the shader cache. If not set, shader cache is disabled.
-    public NativeString shaderCachePath;
+    public NativeString shaderCachePath = {0};
```
csyonghe, line 1728: I wonder why this is needed. NativeString should be treated as a default-initializable type.

tools/gfx/gfx.slang
```diff
@@ -1698,17 +1705,17 @@ public interface IDebugCallback
 
 public struct SlangDesc
 {
-    public NativeRef<slang::IGlobalSession> slangGlobalSession; // (optional) A slang global session object. If null will create automatically.
+    public NativeRef<slang::IGlobalSession> slangGlobalSession = {slang::IGlobalSession()}; // (optional) A slang global session object. If null will create automatically.
```
csyonghe, line 1708: why do we need this change?

tools/gfx/gfx.slang
```diff
@@ -105,9 +105,9 @@ public enum class ShaderModuleSourceType
 
 public struct ShaderProgramDesc2
 {
-    public ShaderModuleSourceType sourceType;
-    public void *sourceData;
-    public Size sourceDataSize;
+    public ShaderModuleSourceType sourceType = ShaderModuleSourceType::SlangSource;
```
csyonghe, line 108: I find it weird to force initialize everything in these descs, why do we need this?

tests/autodiff/self-differential-type-synthesis.slang
```diff
@@ -17,7 +17,7 @@ struct Ray : IDifferentiable {
 [numthreads(1, 1, 1)]
 void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
 {
-    Ray ray = Ray();
+    Ray ray = Ray(0, float3(0, 0, 0), float3(0, 0, 0));
```
csyonghe, line 20: According to the spec, since float and float3 are all default-initializable types, the synthesized ctor should contain default values for these parameters, so why can't we call `Ray()` here?

source/slang/slang-check-decl.cpp
```diff
@@ -9064,25 +9260,59 @@ void SemanticsDeclBodyVisitor::synthesizeCtorBodyForMember(
...
-    memberExpr->baseExpression = thisExpr;
-    memberExpr->declRef = member->getDefaultDeclRef();
-    memberExpr->scope = ctor->ownedScope;
-    memberExpr->loc = member->loc;
-    memberExpr->name = member->getName();
-    memberExpr->type = DeclRefType::create(getASTBuilder(), member->getDefaultDeclRef());
+    Expr* initExpr = nullptr;
+    auto structDecl = as<StructDecl>(member->parentDecl);
+    bool useParamList = isMemberInitCtor;
+    useParamList = isMemberInitCtor && structDecl->m_membersVisibleInCtor.contains(varDeclBase);
```
csyonghe, line 9280: Whether or not a member appears in the param list of a ctor depends on the visibility of the member and the ctor, so this is different for each ctor. I feel it is incorrect to use a single hash set to determine `useParamList`.

We probably shouldn't be storing `m_membersVisibleInCtor` in the struct decl, and just compute that on-demand for each ctor.

source/slang/slang-ast-decl.h
```diff
@@ -165,6 +165,11 @@ class AggTypeDecl : public AggTypeDeclBase
 class StructDecl : public AggTypeDecl
 {
     SLANG_AST_CLASS(StructDecl);
+
+    SLANG_UNREFLECTED
+    // We will use these auxiliary to help in synthesizing the member initialize constructor.
+    Slang::HashSet<VarDeclBase*> m_membersVisibleInCtor;
+    Dictionary<int, ConstructorDecl*> m_synthesizedCtorMap;
```
csyonghe, line 172: what is `m_synthesizedCtorMap` used for? why do we need it?

source/slang/slang-check-decl.cpp
```diff
@@ -8018,6 +8124,29 @@ void SemanticsDeclBodyVisitor::visitFunctionDeclBase(FunctionDeclBase* decl)
...
+            auto structDecl = as<StructDecl>(constructorDecl->parentDecl);
+            ConstructorDecl* defaultCtor = nullptr;
+            if (structDecl->m_synthesizedCtorMap.tryGetValue(
+                    (int)ConstructorDecl::ConstructorFlavor::SynthesizedDefault,
+                    defaultCtor))
+            {
+                structDecl->members.remove(defaultCtor);
+                structDecl->invalidateMemberDictionary();
+                structDecl->buildMemberDictionary();
+                structDecl->m_synthesizedCtorMap.remove(
```
csyonghe, line 8145: why removing things?

source/slang/slang-check-decl.cpp
```diff
@@ -9026,31 +9155,98 @@ static SeqStmt* _ensureCtorBodyIsSeqStmt(ASTBuilder* m_astBuilder, ConstructorDe
...
+    Scope* scope,
+    Decl* member)
+{
+    MemberExpr* memberExpr = m_astBuilder->create<MemberExpr>();
+    memberExpr->baseExpression = thisExpr;
+    memberExpr->declRef = member->getDefaultDeclRef();
+    memberExpr->scope = scope;
+    memberExpr->loc = member->loc;
+    memberExpr->name = member->getName();
+    memberExpr->type = DeclRefType::create(getASTBuilder(), member->getDefaultDeclRef());
```
csyonghe, line 9169: member->getDefaultDeclRef() is a declref to the member and not to a type, it can't be used in DeclRefType::create() IIRC.

The declref you pass to DeclRefType::create() can only refer to a struct, enum, typedef or interface.

source/slang/slang-check-decl.cpp
```diff
@@ -9026,31 +9155,98 @@ static SeqStmt* _ensureCtorBodyIsSeqStmt(ASTBuilder* m_astBuilder, ConstructorDe
...
-        ctorToInvoke->name = declInfo.defaultCtor->getName();
-        ctorToInvoke->loc = declInfo.defaultCtor->loc;
+        ctorToInvoke->declRef = baseCtor->getDefaultDeclRef();
+        ctorToInvoke->name = baseCtor->getName();
+        ctorToInvoke->loc = baseCtor->loc;
         ctorToInvoke->type = m_astBuilder->getFuncType(ArrayView<Type*>(), ctor->returnType.type);
 
         auto invoke = m_astBuilder->create<InvokeExpr>();
         invoke->functionExpr = ctorToInvoke;
+        invoke->arguments.insertRange(0, argumentList);
```
csyonghe, line 9244: `addRange`.

source/slang/slang-check-expr.cpp
```diff
@@ -2909,7 +2909,7 @@ Expr* SemanticsExprVisitor::convertToLogicOperatorExpr(InvokeExpr* expr)
 
     if (auto varExpr = as<VarExpr>(expr->functionExpr))
     {
-        if ((varExpr->name->text == "&&") || (varExpr->name->text == "||"))
+        if (varExpr->name && ((varExpr->name->text == "&&") || (varExpr->name->text == "||")))
```
csyonghe, line 2912: `getText(varExpr->name)`.

Never call `name->text` directly because `name` can be `nullptr`.

source/slang/slang-check-overload.cpp
csyonghe: I feel like adding another parameter to `ensureDecl()` isn't the right solution to this problem. I would like to understand this problem better and see if we can find a more elegant way around it.

csyonghe: Generally, we want to think twice when adding any new modes, flags and special case branches, and make sure anything we add is truly essential. Because special cases is where we will introduce bugs.

source/slang/slang-check-decl.cpp
kaizhangNV: because we don't want to keep two synthesized constructors.

source/slang/slang-check-overload.cpp
kaizhangNV: As per discussed offline.
We fix the infinite loop issue by using alternative solution, where we will create a  ExplicitCtorInvokeExpr type inherit from InvokeExpr type, and check the AST node in this call to indicate this is not a type cast operation.

Also during digging into this issue, we found a new bug regarding to auto-diff.
#6176 is created for tracking.

source/slang/slang-ast-expr.h
kaizhangNV: ack.

tools/gfx/gfx.slang
kaizhangNV: NativeRef also has explicit constructor defined.

source/slang/slang-check-expr.cpp
kaizhangNV: ack

source/slang/slang-check-decl.cpp
kaizhangNV: ack

source/slang/slang-check-decl.cpp
kaizhangNV: This should be some cleanup work after we decide to remove default constructor.

kaizhangNV: I will do that in next commit.

source/slang/slang-ast-decl.h
kaizhangNV: I use this as a cache to quick find constructor.
But since we won't have two ctors in the first place, I might be able to remove this cache and just use getMembersOfType<ConstructorDecl> to find it.

tools/gfx/gfx.slang
csyonghe: We should provide a default ctor for the type.

tools/gfx/gfx.slang
csyonghe: should provide a default ctor here.

source/slang/slang-ast-decl.h
csyonghe: we could be having one ctor for each visibility level, right?

source/slang/slang-check-decl.cpp
kaizhangNV: I'm not sure about this
>> Whether or not a member appears in the param list of a ctor depends on the visibility of the member and the ctor

I agree this, but this list is only used for the synthesized ctor, so we pretty much know the visibility of the ctor is the same as the struct. So the parameter's visibility is determined for synthesized ctor. I might need to change the name
`m_membersVisibleInCtor` to `m_membersVisibleInSynthesizedCtor`.

csyonghe: Ah, you are right. I forgot that we will be synthesizing just one ctor that has the same visibility as the type.

source/slang/slang-ast-decl.h
kaizhangNV: what do you mean by "for each visibility level"?

tests/autodiff/self-differential-type-synthesis.slang
kaizhangNV: According to our new rule,

float and float3 won't be default-initializable types anymore, because they all have explicit ctor defined.
If we want to make them default-initializable type, we have to define default ctor for them.

source/slang/slang-check-decl.cpp
kaizhangNV: remove this already, we're now only creating one ctor at most.
waiting for the CI now

source/slang/slang-ast-decl.h
csyonghe: never mind, at some point the proposal is saying for public structs, we will synthesize one public ctor, one private ctor and one internal ctor that allows you to initialize private/internal members from where those members are visible.

But this is removed from latest proposal so this is no longer the case we need to worry about here.

tests/autodiff/self-differential-type-synthesis.slang
csyonghe: I think this may require the same treatment as C-style types, in that the explicit ctor rule applies to user defined types, but for builtin vectors, scalars, matrices we still want to treat them as default initializable. 

I am not sure about the consequences yet, so maybe we do this as a follow up.

kaizhangNV: Yes, I recommend we should keep this as it, every time when we change the proposal, it could potential break things.

tools/gfx/gfx.slang
kaizhangNV: ack.

tools/gfx/gfx.slang
kaizhangNV: Providing default ctor for this type is not trivial.
Can we do that in a follow-up?

csyonghe: Yes, that is OK. I guess we might need a [__cstyleType] modifier for our internal use.

tools/gfx/gfx.slang
kaizhangNV: Because in the gfx-smoke.slang, we use 

```
    gfx.ShaderProgramDesc2 programDesc = {};
```

to initialize those descs, but those descs are not C-Style type, for example NativeString is not C-Style type in this struct.

tools/gfx/gfx.slang
kaizhangNV: Yes, I think it's the simplest way handle the our core module types.
I created issue #6189 for the follow-up.

tests/autodiff/self-differential-type-synthesis.slang
kaizhangNV: This can also be tracked in #6189.

tools/gfx/gfx.slang
kaizhangNV: See #6189.

tools/gfx/gfx.slang
kaizhangNV: I provide a default constructor for this type as it's simple to add.

source/slang/slang-check-decl.cpp
kaizhangNV: astBuild is a protected member of SemanticsVisitor.

`addVisibilityModifier` and its callee functions are not members function of SemanticsVisitor, I might need to move all of those functions to SemanticsVisitor.

source/slang/slang-check-decl.cpp
kaizhangNV: ack

source/slang/slang-check-decl.cpp
kaizhangNV: ack.

source/slang/slang-check-decl.cpp
kaizhangNV: Yes, this is tracked in #6109.
I currently just did minimal to disable zero init feature.

source/slang/slang-check-decl.cpp
kaizhangNV: ack

source/slang/slang-check-decl.cpp
kaizhangNV: ack.

source/slang/slang-check-overload.cpp
```diff
@@ -1313,6 +1313,21 @@ int SemanticsVisitor::CompareLookupResultItems(
     bool rightIsExtension = as<ExtensionDecl>(rightDeclRefParent.getDecl()) != nullptr;
     if (leftIsExtension != rightIsExtension)
     {
+        // Add a special case for constructors, where we prefer the one that is not synthesized,
+        if (auto leftCtor = as<ConstructorDecl>(left.declRef.getDecl()))
+        {
+            auto rightCtor = as<ConstructorDecl>(right.declRef.getDecl());
+            bool leftIsSynthesized =
+                leftCtor->containsFlavor(ConstructorDecl::ConstructorFlavor::SynthesizedDefault);
```
csyonghe, line 1321: I think we want the same behavior for `MemberInit` flavor.

source/slang/slang-check-decl.cpp
```diff
@@ -7638,6 +7705,42 @@ void SemanticsVisitor::validateEnumTagType(Type* type, SourceLoc const& loc)
...
+
+    if (_hasExplicitCtor(structDecl))
+        return true;
+
+    if (!checkBaseType)
+        return false;
+
+    for (auto inheritanceMember : structDecl->getMembersOfType<InheritanceDecl>())
+    {
+        if (auto baseTypeDecl = isDeclRefTypeOf<StructDecl>(inheritanceMember->base.type))
```
csyonghe, line 7751: `isDeclRefTypeOf<AggTypeDecl>() && !as<InterfaceDecl>(..)`

We want to include both the `Struct` and `Class` decl cases.

source/slang/slang-check-overload.cpp
kaizhangNV: The flavor is bitmask, so MemberInit also have this flag on.

source/slang/slang-check-decl.cpp
kaizhangNV: ack

source/slang/slang-check-conversion.cpp
```diff
@@ -264,6 +264,10 @@ bool SemanticsVisitor::isCStyleType(Type* type)
         {
             Type* varType = varDecl->getType();
 
+            // Skip the type if it is a reference to the struct itself.
```
csyonghe, line 267: This is likely not enough, because I can have struct X that has field Y that has a field of type X.

kaizhangNV: Yea, you are right.

kaizhangNV: Wait, can slang support such case?
```
struct A
{
    B b;
}

struct B
{
    A a;
}

```

kaizhangNV: slang seems not even supporting cyclic reference to itself.

csyonghe: We don’t, but in the case of such code, your logic will overflow the stack, right?

kaizhangNV: Yes, I have the fix, but I'm just trying to have a test case, but found out that I'm not able to create one.

tests/initializer-list/defaultMember.slang
```diff
@@ -0,0 +1,35 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUFFER):-shaderobj -vk
```
csyonghe, line 1: All file names should be lower case and separated with -.

tests/initializer-list/explicitCtor-diagnostic.slang
```diff
@@ -0,0 +1,19 @@
+//DIAGNOSTIC_TEST:SIMPLE(filecheck=CHECK):
```
csyonghe, line 1: filename: `explicit-ctor-diagnostic`

tests/initializer-list/explicitCtor.slang
```diff
@@ -0,0 +1,41 @@
+//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=BUFFER):-shaderobj -vk
```
csyonghe, line 1: filename

csyonghe (approved):
Let's merge this so we can start working on follow ups and other issues that are related to initialization lists.

csyonghe (dismissed):

Issue #3591, opened 2024-02-15
vector<int8_t, N> and vector<uint8_t, N> not supported in dxc backend


skallweitNV:
Using `int8_t_X` and `uint8_t_X` vector types lead to compilation errors in the dxc backend:

```
dxc 1.7: note :     vector<int8_t,2> _S1 = vector<int8_t,2>(int8_t(-4), int8_t(-3));
dxc 1.7: note :            ^
dxc 1.7: c:/projects/kali/src/libkali/kali/device/tests/test_print.slang(93): error :  use of undeclared identifier 'int8_t'
dxc 1.7: note :     vector<int8_t,3> _S2 = vector<int8_t,3>(int8_t(-2), int8_t(-1), int8_t(0));
```

csyonghe:
D3D doesn't have a uint8_t type and there isn't much we can do there besides an error.

PR #6059, opened 2025-01-10
Implement AnyValue marshalling for 8-bit integers

### LINKED ISSUE #4817: `reinterpret` with structs containing `uint64_t` or `uint8_t` incorrectly throws “T contains fields that cannot be packed into AnyValue” error

juliusikkala:
I think this closes #4817?

Based on the discussion in that issue, I implemented this using the now-existing BitfieldInsert & BitfieldExtract instead of the shift and mask approach (which is currently used for 16-bit types) and added a test.

I only realized that there was already an issue about this after having already implemented the shift + mask version, and I have to say that this BitfieldInsert & Extract approach is much cleaner to implement as well.


csyonghe (dismissed):
I didn't realize that this is missing. Thank you for completing the implementation!

csyonghe (dismissed):

juliusikkala:
I took out that DXIL test. I figured out that it's not possible to compile code containing uint8_t there, this is in the same boat as #3591. Which is why that test fails too. This feature has been [backlogged for a future shader model](https://github.com/microsoft/hlsl-specs/issues/337) on their side, so there's not much we can do for now.

juliusikkala:
Heh, formatting issue from master. Since it's now fixed there, I'll just click that Update branch button and let it sort itself out.

csyonghe (approved):

PR #6063, opened 2025-01-11
Reuse code for Metal and WGSL entry point legalization

### LINKED ISSUE #5933: wgsl @location attribute not emitted for entry-point varyings

fairywreath:
Related to #4657, related to #5933.

Reuse common code for Metal and WGSL entry point legalization. 

Hoisting + packing is also currently not done for WGSL entry point parameters and they are required for #5933 as only struct entry points are given attributes. I will have a follow up fix for that as that change will break lots of test and I do not want to make this PR any bigger.

#4657 ultimately aims to also unify the legalization logic with C/CPP/CUDA backends - this PR is a good first step to unify the shading language backends.


fairywreath:
I recommend comparing the (old) individual implementations of the legalization in their separate files with the new unified implementation to make the review process easier.

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -1544,4 +1547,2333 @@ void depointerizeInputParams(IRFunc* entryPointFunc)
...
+
+        ShortList<IRType*> permittedTypes;
+        bool isUnsupported = false;
+
+        // Only used by Metal.
+        bool isSpecial = false;
+
+        SystemValueInfo()
+        {
+            // most commonly need 2
```
csyonghe, line 1595: This shouldn't be needed, as shortlist by default will statically allocate 16 elements on stack.

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -1,11 +1,14 @@
 // slang-ir-legalize-varying-params.cpp
 #include "slang-ir-legalize-varying-params.h"
 
+#include "core/slang-common.h"
 #include "slang-ir-clone.h"
 #include "slang-ir-insts.h"
 #include "slang-ir-util.h"
 #include "slang-parameter-binding.h"
 
+#include <set>
```
csyonghe, line 10: this shouldn't be needed?

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -1,11 +1,14 @@
 // slang-ir-legalize-varying-params.cpp
 #include "slang-ir-legalize-varying-params.h"
 
+#include "core/slang-common.h"
```
csyonghe, line 4: this shouldn't be needed.

csyonghe (approved):
It's great to see the duplicated code being removed. Thank you for putting together the cleanup!

I wonder if we can make this even cleaner by having a base class, and move all the metal/wgsl specific logic into derived classes?

fairywreath:
> I wonder if we can make this even cleaner by having a base class, and move all the metal/wgsl specific logic into derived classes?

Yes this is doable. Should I modify this PR or have these changes as a follow-up?

fairywreath:
> Yes this is doable. Should I modify this PR or have these changes as a follow-up?

I will update this PR - it should be easy to view the changes commit by commit so readability/reviewing is not an issue

source/slang/slang-ir-legalize-varying-params.cpp
fairywreath: `std::set` is actually used and was added here #4536. Looking at that PR's commit history, A custom TreeMap/TreeSet was considered to be added but `std::set` ended up being used.

csyonghe (approved):
Looks good to me. Thanks for the cleanup!

Our CPU/CUDA backend is in great need of a refactor to reuse the same varying legalization logic we have for other targets. We are missing a lot of legalization there. This PR is a good step towards the greater unification.

csyonghe:
Looks like there are still conflicts.

fairywreath:
Sorry for the plethora of dirty commits but the last two commits are the only changes, the rest are reverted(I had incorrect merge commits). I had to move additional logic from #5963.

csyonghe (approved):

PR #6066, opened 2025-01-12
Fix nullptr in generic specialization

### LINKED ISSUE #6064: Using nullptr inside generic function breaks generics specialization
```hlsl
void breakIt<T>()
{
    T* ptr;
    ptr = nullptr;
    printf("nullptr = %p\n", ptr);
}

export __extern_cpp int main(int argc, int8_t** argv)
{
    breakIt<int>();
    return 0;
}
```
`slangc -target cpp test.slang -o test.cpp`

In a release build, this puts some random uninitialized number into `ptr`, causing rather funny prints like `nullptr = 0x7d`. It generates C++ with an incorrect literal (125 in my case) instead of 0 or nullptr, but does not omit the assignment. Luckily, in a debug build, it's caught by an assertion:

```
(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: assert failure: !as<IRConstant>(oldInst)
```

This assertion comes from slang-ir-clone.cpp:63 (cloneInstAndOperands).

juliusikkala:
Fixes #6064. May fix other stuff as well, there appears to have been uninitialized values through `cloneInst` in `slang-ir-link.cpp` (IRConstants were cloned incorrectly).

There's one thing here that is unclear to me, however. `cloneInstAndOperands` had a comment that mentioned that it shouldn't come across IRConstants, but in this case it did. The approach I took in this PR seems to work just fine, but it's not necessarily the cleanest way to do this. I wonder if a better fix would've been to figure out why IRPtrLit survives that far?


source/slang/slang-ir-clone.cpp
```diff
@@ -51,16 +51,16 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
-    // TODO: It would be easy enough to extend this logic
-    // to handle constants gracefully, if it ever comes up.
-    //
-    SLANG_ASSERT(!as<IRConstant>(oldInst));
+    // IRConstants contain data other than just the operands, so cloning them
+    // is done separately.
+    IRConstant* oldConstant = as<IRConstant>(oldInst);
+    if (oldConstant)
+    {
+        IRConstant* newConstant = builder->getClonedConstantValue(*oldConstant);
```
csyonghe, line 59: This will always create a new inst without going through the deduplication logic. If we already have an inst that represents the cloned inst, this will break the invariant that all constants in global scope are deduplicated. 

Given that the case will only ever arise for pointer literals because their types may be generic, we should handle just the PtrLit case and leave other opcodes as is, and then call emitPtrLit instead of doing all these cloning.

juliusikkala: Okay, sounds good. This would also keep the PR much smaller & cleaner since `getClonedConstantValue` would no longer need to separated from `_findOrEmitConstant`.

source/slang/slang-ir-insts.h
```diff
@@ -4433,6 +4433,8 @@ struct IRBuilder
 
     IRGlobalConstant* emitGlobalConstant(IRType* type, IRInst* val);
 
+    IRConstant* emitPtrLit(IRType* type, void* ptr);
```
csyonghe, line 4436: Instead of adding this, we should change `IRBuilder::_getPtrValue` to take a type. We should bottleneck all constant inst creation through `findOrEmitConstant` as in `_getPtrValue`.

csyonghe:
> I added emitPtrLit to IRBuilder, since there wasn't one before. I'm not at all familiar with the invariants here, so not sure if it's at all correct. It's not deduplicating at least, but if I try to use `getNullPtrValue()` / `_findOrEmitConstant()`, I get:
> 
> ```
> test.slang(5): internal error 40007: IR validation failed: def must be ancestor of use
>     printf("%p\n", ptr);
> ```

You would get this error when type isn’t defined in global scope. I think we shouldn’t be specializing the nullptr constant if the type is not fully specialized.

juliusikkala:
I took a bit more time to better understand how the specialization infrastructure works. As you mentioned, the issue was that the type of the constant wasn't fully specialized. And that's because I was using the unspecialized type (`oldType`) for the cloned pointer as well, when I should've used `newType` instead :sweat_smile:

csyonghe (approved):

PR #6067, opened 2025-01-13
Fix #5354, change diagnostic definition error message and expect output in test

### LINKED ISSUE #5354: Minor Improvement: Make no-return warning message more user friendly
rather than
```
shaders/sprite.slang(56): warning 41010: control flow may reach end of non-'void' function
float3 transform(Vertex vert, uint32_t id)
       ^~~~~~~~~
```
I feel the warning would be much more helpful if the verbage was closer to clang, where it says
```
non-void function does not return in all cases
```

wijiler:
This PR changes the warning message of 41010
from 
`control flow may reach end of non-'void' function`
to
`non-void function does not return in all cases`

This verbiage is much clearer on what the actual issue is, and addresses issue #5354


wijiler:
In my testing, works on windows release, and debug passing all tests with vs2022

csyonghe (approved):

wijiler:
@csyonghe could you add pr:non-breaking, im a first time contributor and cant do this myself

PR #6068, opened 2025-01-13
Implement Packed Dot Product intrinsics

### LINKED ISSUE #5574: Fix SM 6.4 math operation implmentation and documentation
Three math intrinsics added for HLSL SM 6.4 are currently implemented as pass-through declarations. They need to be properly implemented with target switch, as done in other operations, and documented, requiring HLSL SM 6.4 e.g. "[require(hlsl, sm_6_4_version)]"

```
//
// Shader Model 6.4
//

/// Treats `left` and `right` as 4-component vectors of `UInt8` and computes `dot(left, right) + acc`
/// @category math
uint dot4add_u8packed(uint left, uint right, uint acc);

/// Treats `left` and `right` as 4-component vectors of `Int8` and computes `dot(left, right) + acc`
/// @category math
int dot4add_i8packed(uint left, uint right, int acc);

/// Computes `dot(left, right) + acc`.
/// May not produce infinities or NaNs for intermediate results that overflow the range of `half`
/// @category math
float dot2add(float2 left, float2 right, float acc);

```

fairywreath:
Closes #5574.

Adds [SM 6.4 Math intrinsics](https://github.com/microsoft/DirectXShaderCompiler/wiki/Shader-Model-6.4) for all targets.


expipiplus1 (approved):

fairywreath:
While fixing the test I discovered a funky issue with HLSL floating point literals described here #6078. `dot2add` accepts float as its last parameter and slang can generated something like `dot2add(x, y, 1.5)` but dxc fails to compile this because its (current default) behavior is to tread `1.5` as a double.

csyonghe (approved):

PR #6070, opened 2025-01-13
C-like emitter: Add parenthesis when combining relational and bitwise…

### LINKED ISSUE #6005: wgsl output doesn't use parenthesis for mixed operations, generating invalid code

aleino-nv:
… operators

This closes #6005.


expipiplus1 (changes requested):
Please can you add a test for these cases.

Happy to merge this since it's an improvement, but this code really feels as though it should be done generically by associating a precedence to each one of these enums and comparing that ranking...

aleino-nv:
> Please can you add a test for these cases.
> 
> Happy to merge this since it's an improvement, but this code really feels as though it should be done generically by associating a precedence to each one of these enums and comparing that ranking...

There are already some tests, but as mentioned [here](https://github.com/shader-slang/slang/issues/6005#issuecomment-2586357113) these cases are kind-of tricky to hit because WGSL requires the result of the `<` to be converted to a `u32`, which adds parentheses. 

The ranking in this case would have to be backend -specific e.g. WGSL -specific, so the ranking cannot be directly associated with just the enum -- target also matters.

aleino-nv:
I agree that the function is getting a bit long, but no easy cleanups come to mind.

source/slang/slang-emit-c-like.cpp
```diff
@@ -827,6 +827,48 @@ bool CLikeSourceEmitter::maybeEmitParens(EmitOpInfo& outerPrec, const EmitOpInfo
     {
         needParens = true;
     }
+    // a | b < c => a | (b < c)
```
csyonghe, line 830: These conditions here are so messy, can we clean them up with a `switch` statement?

We can also just set `needParens = true` when `outerPrec.leftPrecedence` != `prec.leftPrecedence` and one of prec or outerPrec is `BitOr`, `BitAnd`, `BitXor`, `Or`, `And` `Xor`, `Shl` or `Shr`,

aleino-nv: Done. I rewrote the conditions for adding the not-strictly-required parentheses as nested switch statements.

source/slang/slang-emit-c-like.cpp
```diff
@@ -681,151 +681,118 @@ bool CLikeSourceEmitter::maybeEmitParens(EmitOpInfo& outerPrec, const EmitOpInfo
...
-    // a == b < c => a == (b < c)
-    else if (
-        prec.leftPrecedence == EPrecedence::kEPrecedence_Relational_Left &&
-        outerPrec.leftPrecedence == EPrecedence::kEPrecedence_Equality_Right)
+
+    switch (outerPrec.leftPrecedence)
     {
-        needParens = true;
+    case EPrecedence::kEPrecedence_BitAnd_Right:
+        switch (prec.leftPrecedence)
```
csyonghe, line 688: instead of listing every combination, we should be able to fold all these cases by checking just `prec.leftPrecedence < outerPrec.leftPrecedence`.

aleino-nv: That should already be covered by [this line at the top of the function](https://github.com/shader-slang/slang/pull/6070/files/e826e0a9b268bf1800f7332eb53143cc4ab84c7a#diff-6394749074a29fcbed918a58f53b6c1b46170dd00edbb5e4af0753282ca8d6b2R677):

    bool CLikeSourceEmitter::maybeEmitParens(EmitOpInfo& outerPrec, const EmitOpInfo& prec)
    {
        bool needParens = (prec.leftPrecedence <= outerPrec.leftPrecedence) ||
                          (prec.rightPrecedence <= outerPrec.rightPrecedence);
       ...

csyonghe: Then it is unclear to me what this switch is doing, as it appears to be manually enumerating all meaningful cases of `prec.leftPrecedence < outerPrec.leftPrecedence`, and we should be able to delete this entire first switch statement and be fine.

source/slang/slang-emit-c-like.cpp
```diff
@@ -681,151 +681,118 @@ bool CLikeSourceEmitter::maybeEmitParens(EmitOpInfo& outerPrec, const EmitOpInfo
...
+            break;
+        }
+        break;
     }
-    // a < b == c => (a < b) == c
-    else if (
-        prec.rightPrecedence == EPrecedence::kEPrecedence_Relational_Right &&
-        outerPrec.rightPrecedence == EPrecedence::kEPrecedence_Equality_Left)
+
+    switch (outerPrec.rightPrecedence)
```
csyonghe, line 737: This is just `return outerPrec.rightPrecedence < prec.rightPrecedence` when `outerPrec.rightPrecedence` is one of `kEPrecedence_BitAnd_Left`, `kEPrecedence_BitOr_Left`, `kEPrecedence_BitXor_Left`. `kEPrecedence_Equality_Left` and `kEPrecedence_Shift_Left`.

I think the whole function simplifies down to

```
bool needParens = (prec.leftPrecedence <= outerPrec.leftPrecedence) ||
                      (prec.rightPrecedence <= outerPrec.rightPrecedence);
if (needParens) return true;

if (isBitLogicalOrRelational(prec))
{
     if (outerPrec.leftPrecedence > kEPrecedence_General_Left)
         return true;
}
if (isBitLogicalOrRelational(outerPrec))
     return true;
return false;
```

where `isBitLogicalOrRelational` returns `true` for any relational, bitand, bit or, bit xor, shift, or, and `and` precedences.

source/slang/slang-emit-c-like.cpp
aleino-nv: This switch is enumerating cases where inner precedence is _higher_ than the outer, not the other way around.

    // While Slang correctly removes some of parentheses, many compilers print warnings
    // for common mistakes when parentheses are not used with certain combinations
    // of the operations. We emit parentheses to avoid the warnings.
    //


    switch(outerPrec.leftPrecedence)
    {
    case EPrecedence::kEPrecedence_BitAnd_Right:
        switch(prec.leftPrecedence)
        {
        case EPrecedence::kEPrecedence_Relational_Left: // a & b < c => a & (b < c)
        case EPrecedence::kEPrecedence_Additive_Left: // a & b + c => a & (b + c)
        case EPrecedence::kEPrecedence_Multiplicative_Left: // a & b * c => a & (b * c)
            needParens = true;
            break;
        }
        break;
    ...

Outer precedence is `BitAnd`.
Inner precedences are `Relational`, `Additive` and `Multiplicative`.

Here is the precedence table:


    #define SLANG_PRECEDENCE(x) \
        x(None,         NON_ASSOC) \
        x(Comma,        LEFT) \
        \
        x(General,      NON_ASSOC) \
        \
        x(Assign,       RIGHT) \
        \
        x(Conditional,  RIGHT) \
        \
        x(Or,           LEFT) \
        x(And,          LEFT) \
        x(BitOr,        LEFT) \
        x(BitXor,       LEFT) \
        x(BitAnd,       LEFT) \
        \
        x(Equality,     LEFT) \
        x(Relational,   LEFT) \
        x(Shift,        LEFT) \
        x(Additive,     LEFT) \
        x(Multiplicative, LEFT) \
        x(Prefix,       RIGHT) \
        x(Postfix,      LEFT) \
        x(Atomic,       NON_ASSOC)

source/slang/slang-emit-c-like.cpp
aleino-nv: > where `isBitLogicalOrRelational` returns `true` for any relational, bitand, bit or, bit xor, shift, or, and `and` precedences.

So `isBitLogicalOrRelational` would take a `EmitOpInfo`. That contains both a left and right precedence, so I suppose the function should check if either the left or the right is one of the listed classes of operations?

I did something like that and found that I had to throw in `Equality` as well as the listed operations.

If I do that then the `tests/wgsl/associativity.slang` passes, although I'm pretty sure we're now adding parentheses in lots of cases where the WGSL compiler wouldn't require them.

csyonghe (approved):

csyonghe:
There is a test that needs to be updated.

csyonghe (approved):

Issue #5972, opened 2024-12-31
Crash in constant propagation


Axel-Reactor:
This is nonsensical test code from internal code gen testing, but I reliably crash with this:

```
#version 460

struct TYPE_QBmQRbmlpcTGbHxh { uint32_t variant; };
bool operator==(TYPE_QBmQRbmlpcTGbHxh lhs, TYPE_QBmQRbmlpcTGbHxh rhs) { return lhs.variant == rhs.variant; }
const TYPE_QBmQRbmlpcTGbHxh TYPE_kxUyNIzptPXewEJs = TYPE_QBmQRbmlpcTGbHxh(0);
const TYPE_QBmQRbmlpcTGbHxh TYPE_JLGyglBSdGSCICTU = TYPE_QBmQRbmlpcTGbHxh(1);
const TYPE_QBmQRbmlpcTGbHxh TYPE_ASIGaEGzVWSeCIec = TYPE_QBmQRbmlpcTGbHxh(2);

struct TYPE_eyQZDoMvVLZElrDX {
    vector<float32_t, 4> m_ujigZxHKbJHSjjnn;
    vector<float32_t, 4> m_ICCAonGNzdVjdlFA;
};

struct TYPE_YFSaXGWERCxuTrrK {
    TYPE_QBmQRbmlpcTGbHxh m_mdQTHFyBjntnjvIN;
};

layout (scalar, set = 1, binding = 0) uniform UBO_Global {
   float64_t HREOIZIQbKUVImYu;
   int64_t TCQXdQpMJLfmpkBk;
   uint64_t BxMLeBcLoJtfhYSt;
   TYPE_QBmQRbmlpcTGbHxh qBiKxIBkevFdwQpf;
   float32_t NNcjGUneJoKLqabz;
   int32_t sGnohaMoeSNpKPid;
   uint32_t NBitKKiZyGtZUjVI;
   uint32_t wERQxwcenXQxjkLv;
   uint8_t sNZWJyWSVFVhBiIM;
} UNIFORMS_Global;
layout (set = 1, binding = 1) SamplerState BINDING_wizPzNhlwuWdxKiV;
layout (location = 0) in vector<float32_t, 2> BINDING_jishczyZNXwjphtt;
layout (location = 0) out vector<float32_t, 4> BINDING_AWsuafnviCutmiDT;
layout (location = 1) out vector<float32_t, 4> BINDING_pffaEcYwacxRzGLr;

[shader("fragment")] void main()
{
[...]

PR #6077, opened 2025-01-13
Fix simplify if-else

cheneym2:
The if-else optimization observes that at if at least one true/false block is merely an unconditional jump to the after block, and both are "trivial", that the whole if-else can be replaced with a jump to the after block. But it's important to copy the phi arguments from the aforementioned unconditional jump, rather than what is present in the 'true' block, since the 'true' block might actually just be the after block itself.

Below, the ifElse() would be replaced with an unconditional jump to block %39, but with the `phi` arguments copied from the branch to %29, which is an unrelated block.

ifElse(%38, %39, %40, %39)

block %40:
    unconditionalBranch(%39)

block %39:
    unconditionalBranch(%29, 0 : Float)

block %29(
        [nameHint("ret")]
        param %ret  : Float):

Fixes issue #5972


source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -490,10 +490,17 @@ static bool trySimplifyIfElse(IRBuilder& builder, IRIfElse* ifElseInst)
     bool isFalseBranchTrivial = false;
     if (isTrivialIfElse(ifElseInst, isTrueBranchTrivial, isFalseBranchTrivial))
     {
-        // If both branches of `if-else` are trivial jumps into after block,
+        // If either branch of `if-else` is a trivial jump into after block,
         // we can get rid of the entire conditional branch and replace it
         // with a jump into the after block.
-        if (auto termInst = as<IRUnconditionalBranch>(ifElseInst->getTrueBlock()->getTerminator()))
+        IRUnconditionalBranch* termInst;
+        if (termInst = as<IRUnconditionalBranch>(ifElseInst->getTrueBlock()->getTerminator()))
```
csyonghe, line 497: style: fold the initialization to the declaration site of termInst, and just do `if (termInst & termInst->getTargetBlock() != ifElseInst->getAfterBlock())`

csyonghe (approved):

PR #6084, opened 2025-01-14
Fix issue with specialization using arithmetic expressions

saipraveenb25:
This PR runs SCCP in tandem with specialization so that arithmetic expressions are simplified before they can be used to specialize generic objects.

Fixes an issue where the same effective specialization can be replicated several times, causing a type mismatch.


csyonghe (approved):

csyonghe:
Can we add a test case?

Issue #5938, opened 2024-12-26
register usage incorrectly produces a warning and suggests [[vk::binding(...)]]


stanoddly:
If I try to compile this shader:
```
Sampler2D CombinedSampler : register(t0): register(s0);

[shader("fragment")]
float4 main(float2 TexCoord : TEXCOORD0) : SV_Target0
{
    return CombinedSampler.Sample(TexCoord);
}
```

via command line:
```
/opt/slang/bin/slangc shader.slang -target "spirv" -o "shader.spv"
```

it seems to generate `shader.spv` as expected, but also a warning is produced:
```
 warning 39013: shader parameter 'CombinedSampler' has a 'register' specified for D3D, but no '[[vk::binding(...)]]` specified for Vulkan
```

However, such usage is explicitly documented in "SPIR-V specific functionalities" ([link](shader-slang.com/slang/user-guide/spirv-target-specific.html#combined-texture-sampler)):

>  ## Combined texture sampler
> 
> Slang supports Combined texture sampler such as `Sampler2D`. Slang emits SPIR-V code with `OpTypeSampledImage` instruction.
> 
> You can specify two different register numbers for each: one for the texture register and another for the sampler register.
> 
> ```
> Sampler2D explicitBindingSampler : register(t4): register(s3);
> ```
> 

So, there should be no warning for a `register` usage, shouldn't be? I know the warning could be disabled via `-warnings-disable`, however I would expect there is nothing to warn about or alternatively register usage shouldn't be documented.

I would personally say that any usage of register shouldn't produce any warning. I think the behavior should be aligned with how `dxc` - `dxc` produces no warnings when compiling to SPIR-V ([source](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#implicit-binding-number-assignment) + verified manually).

csyonghe:
We should remove this warning now that we now support the same -fvk-shift options as in DXC.

cheneym2:
@csyonghe What is the syntax in https://github.com/orgs/shader-slang/projects/10/views/shader-slang.com/slang/user-guide/spirv-target-specific.html#combined-texture-sampler supposed to do for a SPIRV target?

csyonghe:
A Sampler2D maps to a single combined texture sampler descriptor in Vulkan, so it will have only one binding index. This is basically what sampler2D in glsl maps to.

cheneym2:
@csyonghe My understanding is that we should:

[...]

PR #6085, opened 2025-01-14
Remove warning about unspecified vkbinding

cheneym2:
It is permitted that users omit explicit [vk::binding] settings on register definitions of HLSL-style, as they are opting into the default implicit mapping, as with DXC.

That does mean register IDs can clash between resources which have different register type but the same register number. For example:

Such a
problem can be resolved either with explicit bindings in source, or using the "-fvk-binding" slangc options which match DXC behavior.

Closes issue #5938

### DIFF
--- a/source/slang/slang-parameter-binding.cpp
+++ b/source/slang/slang-parameter-binding.cpp
@@ -1060,22 +1060,6 @@ static void addExplicitParameterBindings_HLSL(
     }
 }
 
-static void _maybeDiagnoseMissingVulkanLayoutModifier(
-    ParameterBindingContext* context,
-    DeclRef<VarDeclBase> const& varDecl)
-{
-    // If the user didn't specify a `binding` (and optional `set`) for Vulkan,
-    // but they *did* specify a `register` for D3D, then that is probably an
-    // oversight on their part.
-    if (auto registerModifier = varDecl.getDecl()->findModifier<HLSLRegisterSemantic>())
-    {
-        getSink(context)->diagnose(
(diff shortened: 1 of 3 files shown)


source/slang/slang-parameter-binding.cpp
```diff
@@ -1254,21 +1231,12 @@ static void addExplicitParameterBindings_GLSL(
...
-        }
-    }
 
-    if (warnedMissingVulkanLayoutModifier)
+    if (hlslToVulkanLayoutOptions == nullptr || !hlslToVulkanLayoutOptions->canInferBindings() ||
+        !hlslToVulkanLayoutOptions->canInfer(vulkanKind, hlslInfo.space))
     {
-        // If we warn due to invalid bindings and user did not set how to interpret 'hlsl style
-        // bindings', we should map `register` 1:1 with equivlent vulkan bindings.
+        // If the user did not set how to interpret 'hlsl style bindings', we should map
```
csyonghe, line 1238: Instead of removing the warning completely, let's remove the warning only if `hlslToVulkanLayoutOptions` is not set or `hlslToVulkanLayoutOptions->canInfer(vulkanKind, hlslInfo.space)` is `false`, and amend the warning message to ask the user to either use `[vk::binding]` or specify `-fvk-binding-` options.

cheneym2: Will do

csyonghe:
@cheneym2 any updates on this PR?

PR #6093, opened 2025-01-15
Add validation for destination of atomic operations

aleino-nv:
- Rewrite GLSL atomic operations in terms of the intrinsic IR atomic operations
- Add validation for destination of atomic operations
- Add some tests that the validation is working as expected

This closes issue #5989.


aleino-nv:
@csyonghe This validation currently breaks the following tests:


    // tests/spirv/address-space-specialize.slang
    groupshared int gArray1[2];
    void atomicOp(inout int array[2])
    {
        InterlockedAdd(array[0], 1); 
    }
    [numthreads(1,1,1)]
    void main()
    {
        atomicOp(gArray1);
    }

1. How can I detect that `array` is groupshared in the above case?

2. The test [`tests/spirv/ptr-vector-member.slang` ](https://github.com/shader-slang/slang/blob/master/tests/spirv/ptr-vector-member.slang) seems to be doing atomics on a push constant. Is that supposed to be allowed?

csyonghe:
you need to call `getRootAddr(inst)` to get the root of any addressing instruction (e.g. GetElementAddr, FieldAddress), then check the type of the root inst.

If the root addr is `kIROp_RWStructuredBufferGetElementPtr`, then the addr is in a hlsl structured buffer so atomic operations should be allowed.

If the root addr is a pointer where the address space is one of `UserPointer`, `Global` or `StorageBuffer`, then atomic operations should be allowed.

In [tests/spirv/ptr-vector-member.slang](https://github.com/shader-slang/slang/blob/master/tests/spirv/ptr-vector-member.slang), the operand of the atomic operation is a pointer (whose address space is UserPointer), the pointer itself is a push constant, but that is irrelevant to the issue: the address where we are performing the atomic operation on is in PhyiscalStorageBuffer address space since we are doing atomics on a user defined pointer.

aleino-nv:
> you need to call `getRootAddr(inst)` to get the root of any addressing instruction (e.g. GetElementAddr, FieldAddress), then check the type of the root inst.
> 

Yes, I'm following addressing instructions to get to the root.
(I will switch to use `getRootAddr` where I can once I get all the tests passing. At the moment I believe I need to chase through a few more indirections than `getRootAddr` currently does, namely `Load`, `GetElementPtr` and `GetOffsetPtr`.)

> If the root addr is a pointer where the address space is one of `UserPointer`, `Global` or `StorageBuffer`, then atomic operations should be allowed.

Thanks, I hadn't added the ptr case yet and that's why `ptr-vector-member` was failing.

> In [tests/spirv/ptr-vector-member.slang](https://github.com/shader-slang/slang/blob/master/tests/spirv/ptr-vector-member.slang), the operand of the atomic operation is a pointer (whose address space is UserPointer), the pointer itself is a push constant, but that is irrelevant to the issue: the address where we are performing the atomic operation on is in PhyiscalStorageBuffer address space since we are doing atomics on a user defined pointer.

Ok, the syntax mislead me. I think would have expected the syntax `InterlockedAdd(v, 1)` for what you're describing, but instead the test code reads:

    uint * v = &push2.value[0].x;
    InterlockedAdd(*v, 1);

aleino-nv:
I'm still having an issue with this case:

    // tests/spirv/address-space-specialize.slang
    groupshared int gArray1[2];
    void atomicOp(inout int array[2])
    {
        InterlockedAdd(array[0], 1); 
    }
    [numthreads(1,1,1)]
    void main()
    {
        atomicOp(gArray1);
    }

When I chase `array[0]` to get to a root addr, I just end up at `IRParam`.
This is not really surprising, but I wonder what the solution should be.

I see a few options: 

1. `atomicOp` must be inlined before running this validation.
2. The parameter of `atomicOp` must be marked up somehow so that it only accepts parameters that lead back to valid atomic operation destinations.
3. The validation logic should be elaborated such that in this case we find all calls to this function and then run the validation on the function call argument corresponding to the parameter.

csyonghe:
I do t think we should be chasing OpLoad, because InterlockedAdd(*ptr, …) does not translate to AtomicAdd(Load(ptr)) and is just AtomicAdd(ptr).

*ptr returns a reference, and at IR level a reference is still a pointer.

using atomic operation on an inout array is technically unsound and we may want to disallow this in the future, but for now, make sure your check is after address space specialization pass to get rid of the issue. After address space specialization, the IRParam’s type will be a IROutTypeBase with a specialized address space, and you can apply the same rules to check that address space.

aleino-nv:
> I do t think we should be chasing OpLoad, because InterlockedAdd(*ptr, …) does not translate to AtomicAdd(Load(ptr)) and is just AtomicAdd(ptr).
> 

Done -- removed.
(This was not specifically for `AtomicAdd`. I think I don't need to chase through `OpLoad` anymore after I added detection for the ptr cases.)

> using atomic operation on an inout array is technically unsound and we may want to disallow this in the future, but for now, make sure your check is after address space specialization pass to get rid of the issue. After address space specialization, the IRParam’s type will be a IROutTypeBase with a specialized address space, and you can apply the same rules to check that address space.

Thanks, that made the test pass.

source/slang/slang-emit.cpp
```diff
@@ -1322,6 +1322,11 @@ Result linkAndOptimizeIR(
             byteAddressBufferOptions);
     }
 
+    // For SPIR-V, this function is called elsewhere, so that it can happen after address space
```
csyonghe, line 1325: This could mean we will still be generating errors when emitting hlsl or other non-spirv target code.

I think we should amend `validateAtomicOperations` such that for any non-spirv target code, we silently treat inout/out `IRParam` as valid target for atomic operations for now.

source/slang/slang-ir-validate.cpp
```diff
@@ -410,4 +410,96 @@ void validateIRModuleIfEnabled(CodeGenContext* codeGenContext, IRModule* module)
...
+        return isValidAtomicDest(fieldAddress->getBase());
+
+    return false;
+}
+
+void validateAtomicOperations(DiagnosticSink* sink, IRInst* inst)
+{
+    // There may be unused functions containing violations after address space specialization.
+    if (auto func = as<IRFunc>(inst))
+        if (!(func->hasUses() || func->findDecoration<IREntryPointDecoration>()))
```
csyonghe, line 470: I think we can remove this check because all functions that exist in the IR at this point are existing for a reason, due to DCE has already run.

tests/glsl-intrinsic/atomic/invalidDestAdd.slang
```diff
@@ -0,0 +1,13 @@
+//DIAGNOSTIC_TEST:SIMPLE(filecheck=CHECK):  -allow-glsl -stage compute -entry computeMain -target glsl -DTARGET_GLSL
```
csyonghe, line 1: You can merge all these tests into one single file, and use
```
// CHECK: ([[# @LINE+1]]): error 41043
```
before all the atomic calls to ensure there are error messages for each one of them.

source/slang/slang-ir-validate.cpp
aleino-nv: I added this because of the IRParam issue: it seemed like the original function (with IRParam) was left in the IR with no uses after address space specialization had run.

However, if I change the code to treat inout/out IRParam's as valid then I guess I won't hit this.

source/slang/slang-emit.cpp
aleino-nv: I checked that it works for Metal, at least. (Address space specialization is run in Metal IR legalization, which happens before this point in the code.)

> I think we should amend validateAtomicOperations such that for any non-spirv target code, we silently treat inout/out IRParam as valid target for atomic operations for now.

Sure, then I guess I don't need two separate call sites for this function.

source/slang/slang-ir-validate.cpp
aleino-nv: Confirmed: I still hit this issue in the call from slang-ir-spirv-legalize.cpp (since that's SPIR-V target I don't allow skipping the validation of the IRParam, as for other targets), if I delete this check.

Some options I can see:
- Do the check in address space specialization and clean up any non-entrypoint functions left without uses there.
- Keep the check here.
- Move the call to `validateAtomicOperations` to some other place where the function has been cleaned up.

tests/glsl-intrinsic/atomic/invalidDestAdd.slang
aleino-nv: Done.

source/slang/slang-emit.cpp
aleino-nv: Done: I added a parameter that lets the caller specify whether to skip full IRParam checks in this way.
The caller can set this parameter to false for SPIR-V and true for other targets.

csyonghe (dismissed):

aleino-nv:
I changed the entrypoint check into an assert, as discussed yesterday.

csyonghe (approved):

PR #6099, opened 2025-01-15
Fix resource specialization issue where store insts from inlined calls are not considered properly.

saipraveenb25:


csyonghe (approved):

csyonghe:
Can we get a test case?

saipraveenb25:
Yes, working on a test case now.

saipraveenb25:
@csyonghe: I tried for a bit to reproduce this with a smaller test case, but it's really tricky to reproduce this minimally. 
It needs the global insts to be emitted in just the right order to cause the last-minute inlining step to become necessary (which seems to only occur with a certain combination of public/module/etc..)

The larger test-case is part of slangpy, and since it will be on CI, I think it may not be worth too much spending time on this?

csyonghe:
No worries then. We will use the slangpy test.

Issue #5894, opened 2024-12-17
shared memory + atomic addition?


Jaisiero:
Hi there! I'm using shared memory and trying to increment it with an atomic operation in a compute shader but it's not working.
I have tried to simplify the code as much as possible.  Could you guys take a look?
```cpp
// Edit: I forgot this guys
static const uint RADIX_SORT_WORKGROUP_SIZE = 256; // assert WORKGROUP_SIZE >= RADIX_SORT_BINS
static const uint RADIX_SORT_BINS = 256;

// groupshared variables
groupshared uint _histogram[RADIX_SORT_BINS];

// Entry point with HLSL semantics
[numthreads(RADIX_SORT_WORKGROUP_SIZE, 1, 1)] void entry_radix_sort_histogram(uint3 pixel_i : SV_DispatchThreadID,
                                                                                                     uint3 GroupThreadID : SV_GroupThreadID,
                                                                                                                                  uint3 GroupID : SV_GroupID)
{
  uint lID = GroupThreadID.x;
  uint wID = GroupID.x;

  Ptr<uint> global_histograms =
      Ptr<uint>(RBRSH.task_head.global_histograms);

  Ptr<SimConfig> sim_config =
      Ptr<SimConfig>(RBRSH.task_head.sim_config);

  uint shift = sim_config->radix_shift;

  uint num_of_elements = sim_config->rigid_body_count;

  // Initialize histogram
  if (lID < RADIX_SORT_BINS)
  {
    _histogram[lID] = 0U;
  }
  GroupMemoryBarrierWithGroupSync();

  // Add to histogram
  if (lID < RADIX_SORT_BINS)
  {
    uint old_value = 0;
    InterlockedAdd(_histogram[lID], 1, old_value);
    // ++_histogram[lID]; <-- this works to increase count but I obviously need it in an atomic manner
  }
  GroupMemoryBarrierWithGroupSync();

  if (lID < RADIX_SORT_BINS)
  {
    uint count = _histogram[lID];
    global_histograms[RADIX_SORT_BINS * wID + lID] = count;
    if(count > 0)
      printf("Histogram global bin %d, bin %d, count %d\n", RADIX_SORT_BINS * wID + lID, lID, count);
  }
}
````

Expected value:  It should return at least one (more than zero for sure, printf is performed when regular increment is used).
Value return: When I use atomics, value returned is always zero (not printf performed at all).

Thank you!

Jaime.

csyonghe:
[...]

PR #6107, opened 2025-01-16
Add groupshared atomic array test.

csyonghe:
Related to #5894.

Add a test derived from the user reported issue and verify that we are translating groupshared atomic array to SPIRV correctly.


jkwak-work (approved):
Looks good to me

PR #6110, opened 2025-01-16
Fix cyclic lookups with UnscopedEnums

### LINKED ISSUE #6096: Multiple unscoped enums with explicit underlying type cause cyclic reference error
Related to #3959.

If there are more than one enums with `[UnscopedEnum]`, and they explicitly specify the underlying type, a cyclic reference is incorrectly detected.

Example code:
```cpp
[UnscopedEnum]
enum EnumA : int32_t { 
    ENUM_A_VALUE
};

[UnscopedEnum]
enum EnumB : uint32_t { 
    ENUM_B_VALUE
};
```

Compiling the above produces the following output:
```
test.slang(2): fatal error 39999: cyclic reference 'EnumA'.
enum EnumA : int32_t {
```

While the root cause is highly similar, the fix for #3959 doesn't help here, because only one of the enum declarations can be ignored via `LookupRequest::declToExclude` at a time. So the sequence of events is:

1. EnumA is looking up int32_t (declToExclude = EnumA) and ends up `ensureDecl()`ing EnumB
2. EnumB looks up uint32_t (declToExclude = EnumB) and ends up `ensureDecl()`ing EnumA
3. Cyclic reference, as EnumA is still being checked.

Seems a bit tricky to fix. Maybe declToExclude could be a set of decls and collect both EnumA and EnumB along the way?

juliusikkala:
Fixes #6096 by ignoring transparent decls when looking up inheritances of a transparent parent decl, as that situation leads to cyclic lookups.


csyonghe (approved):
Looks good, thank you for fixing this issue!

csyonghe (approved):

Issue #6117, opened 2025-01-17
Derivatives do not propagate through generic __subscript in interfaces


tunabrain:
Follow-up on #5921: The bug reappears if the subscript is generic. Modified slang test below that reproduces it on current ToT:

```
//TEST:COMPARE_COMPUTE(filecheck-buffer=CHK): -output-using-type
interface ITest
{
    __generic<I : IInteger>
    __subscript(I i) -> float
    {
        [BackwardDifferentiable] get;
    }
}
struct Test : ITest
{
    __generic<I : IInteger>
    __subscript(I i) -> float
    {
        [BackwardDifferentiable] get { return 5.0f * i.toInt(); }
    }
}
[Differentiable]
float test(ITest arg)
{
    return arg[1];
}
//TEST_INPUT:set output = out ubuffer(data=[0 0 0 0], stride=4)
RWStructuredBuffer<float> output;
[numthreads(1,1,1)]
void computeMain()
{
    Test t = {};
    output[0] = test(t);
    // CHK: 5.0
}
```

PR #6123, opened 2025-01-17
Fix interface requirement lowering for generic accessors

saipraveenb25:
Fixes: https://github.com/shader-slang/slang/issues/6117

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -8633,6 +8633,9 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         UInt operandCount = 0;
         for (auto requirementDecl : decl->members)
         {
+            if (as<GenericDecl>(requirementDecl))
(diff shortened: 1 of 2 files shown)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -8782,6 +8785,13 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
             auto requirementKey = getInterfaceRequirementKey(requirementDecl);
             if (!requirementKey)
             {
+                if (auto genericDecl = as<GenericDecl>(requirementDecl))
```
csyonghe, line 8788: should this logic be folded into `getInterfaceRequirementKey` instead?

saipraveenb25: This might be a bit tricky since we'll need to fold the `PropertyDecl`/`SubscriptDecl` cases below it into `getInterfaceRequirementKey` as well, and by that point, each `requirementDecl` may provide more than one key (getter + setter)

csyonghe (approved):

PR #6124, opened 2025-01-17
Allow __subscript<T> syntax.

### LINKED ISSUE #6125: Support generic syntax on `__subscript`.
Currently, `__subscript` is missing parsing logic for generic signatures, and the user has to fallback to use `__generic<T>` construct before `__subscript` to define generic subscripts. We should apply the same generic parsing logic to `__subscript`, so that the following syntax is allowed:

```
__subscript<T>(T index)->float
    where T:IInteger
{
     get{} set{}
}
```

csyonghe:
Closes #6125.


saipraveenb25 (dismissed):
Looks good! This will simplify a lot of the SlangPy code

PR #6126, opened 2025-01-17
Make -depfile work for binary slang-modules

juliusikkala:
Previously, `-depfile <filename>` would create an empty file when `.slang-module` is requested and there is no `-target`. This PR makes `-depfile` also output dependencies for `slang-module`s. This can be used in build systems (e.g. CMake, that's my use case) to make sure that all files importing a module get rebuilt when that module is edited.


csyonghe (approved):
Looks good to me!

csyonghe (approved):

Issue #5407, opened 2024-10-25
Interface with 2 implementations from separate compiled slang-modules -> crash


fknfilewalker:
So I have this simple interface and the following simple shader:

```csharp
// emitter.slang
public interface IEmitter<Real : __BuiltinFloatingPointType> {
    [Differentiable] vector<Real, 3> illuminate(vector<Real, 3>, vector<Real, 3>);
    [Differentiable] vector<Real, 3> sample();
} 
```

```csharp
import emitter;
StructuredBuffer<IEmitter<float>> emitters;
RWStructuredBuffer<float> outputBuffer;

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    vector<float, 3> P = { 0.0, 0.0, 0.0 };
    vector<float, 3> N = { 0.0, 1.0, 0.0 };
    outputBuffer[0] = emitters[0].illuminate(P, N).x;
}
```

I compile two different implementations (PointLight and SpotLight) of this interface to two slang-modules, which then get added when compiling the main shader above. If I only use one implementation it compiles but when using two, it results in a crash.

Directly compiling to spirv gives me "Unhandled global inst in spirv-emit: witness_table %1	: %2(%PointLight);"
Compiling to glsl gives me an internal error 99999

So the problem comes from the last line where the compiler has an `kIROp_swizzle` with `kIROp_WitnessTable` as operand0, which then in the func `defaultEmitInstExpr` can not be handled and results in the error.

https://github.com/fknfilewalker/dyslang contains the code 
`tests/load_plugins_one` works but `tests/load_plugins_two` crashes (only tried it on windows)
`DYSLANG_SLANG_FROM_SUBMODULE TRUE` uses the latest github code.

Devon7925:
I got this error as well. But I can trigger it without a generic interface and without using multiple modules. This is as close to a minimal example as I could get.
```slang
RWStructuredBuffer<float> buffer;

void doSomethingToBuffer() {
    buffer[0] += 0.0;
}

interface Processor {
    void process();
}

struct Layer<let SIZE : uint> : Processor {
    void process() {
        doSomethingToBuffer();
    }
}

struct Wrapper {
    Processor inner;

    void run() {
        inner.process();
    }
}

void eval() {
    Wrapper(Layer<1>()).run();
}

[shader("compute")]
[numthreads(64, 1, 1)]
void computeMain()
{
    eval();
}
```
[...]

PR #6129, opened 2025-01-18
Fix generic witness table SequentialIDDecorations

juliusikkala:
Fixes (hopefully) #5407, although I couldn't get OPs code compiling with the current master branch version of Slang. At least this PR fixes the two similar cases from the comments. This is likely not the correct way to do it, but it's probably easiest to discuss this in a PR since I want to fix this in any case.

The problem appears to be that witness tables for generic structs don't get `SequentialIDDecoration`s. This is because `ensureWitnessTableSequentialIDs` checks for LinkageDecoration, which those generic witness tables don't have (but should they?). When `addLinkageDecoration` is called for witness tables in `visitInheritanceDecl` (`slang-lower-to-ir.cpp`), it starts by finding the outermost generic and attaches the decoration to that, which in this case is not the witness table itself.

I tried adding a new flag to `addLinkageDecoration` to skip that outermost-generic-search, skipping it only in `visitInheritanceDecl`, but that broke some tests, so it's probably not that simple.

So the overkill approach in this PR is to give `SequentialIDDecoration` to every single witness table for which `ensureWitnessTableSequentialIDs` is called. It doesn't seem to break any tests and fixes the code examples from that issue, but I guess there must be some negative consequences since clearly some effort was seen to only add the IDs in a very specific corner case.


csyonghe:
I can no longer recall why we need to skip assigning IDs for witness tables without a linkage. I don't see the downside of assigning every witness table a sequential ID. If all tests are passing, then this is fine and we will deal with any consequences if we find one in the future.


At this point all witness tables should have a linkage decoration due to the way we lower them to IR. And really the specialized generic witness tables are the only ones that do not have a linkage decoration but they shouldn't be excluded. So removing this logic seems the correct approach.

csyonghe (approved):

PR #6131, opened 2025-01-19
Remove unnecessary parameters from Metal entry point signature

### LINKED ISSUE #6112: Bug in metal backend which push all buffers as input to every shader function inside file
Code like this

```cxx
[shader("compute")]
void peaseFFTHalf(
    RWStructuredBuffer<ComplexHalf> peaseFFTHalfData,
    ConstantBuffer<uint> bufferSize,
    ConstantBuffer<uint> stage,
    uint3 dispatchThreadID: SV_DispatchThreadID
) {
}



[shader("compute")]
void peaseFFTHalfMagintude(
    RWStructuredBuffer<ComplexHalf> data,
    ConstantBuffer<uint> bufferSize,
    ConstantBuffer<uint> stage,
    ConstantBuffer<bool> isFinalStage,
    uint3 dispatchThreadID: SV_DispatchThreadID
) {
}

struct AudioParams {
    uint frameSize;
    uint hopLength;
    uint numFrames;
    uint melFilters;
    float sampleRate;
};

[shader("compute")]
void peaseFFTHalf(
) {
    RWStructuredBuffer<float> framingInput;
    RWStructuredBuffer<half> framingOutput;
    ConstantBuffer<AudioParams> framingParams;
}
```

Produce results like this

```cxx

#line 25
[[kernel]] void peaseFFTHalf(uint3 dispatchThreadID_0 [[thread_position_in_grid]], ComplexHalf_0 device* entryPointParams_peaseFFTHalfData_1 [[buffer(0)]], uint constant* entryPointParams_bufferSize_2 [[buffer(1)]], uint constant* entryPointParams_stage_2 [[buffer(2)]], ComplexHalf_0 device* entryPointParams_data_1 [[buffer(3)]], uint constant* entryPointParams_bufferSize_3 [[buffer(4)]], uint constant* entryPointParams_stage_3 [[buffer(5)]], bool constant* entryPointParams_isFinalStage_1 [[buffer(6)]], float device* entryPointParams_framingInput_1 [[buffer(7)]], half device* entryPointParams_framingOutput_1 [[buffer(8)]], AudioParams_0 constant* entryPointParams_framingParams_1 [[buffer(9)]])
{

[[kernel]] void peaseFFTHalfMagintude(uint3 dispatchThreadID_1 [[thread_position_in_grid]], ComplexHalf_0 device* entryPointParams_peaseFFTHalfData_2 [[buffer(0)]], uint constant* entryPointParams_bufferSize_4 [[buffer(1)]], uint constant* entryPointParams_stage_4 [[buffer(2)]], ComplexHalf_0 device* entryPointParams_data_2 [[buffer(3)]], uint constant* entryPointParams_bufferSize_5 [[buffer(4)]], uint constant* entryPointParams_stage_5 [[buffer(5)]], bool constant* entryPointParams_isFinalStage_2 [[buffer(6)]], float device* entryPointParams_framingInput_2 [[buffer(7)]], half device* entryPointParams_framingOutput_2 [[buffer(8)]], AudioParams_0 constant* entryPointParams_framingParams_2 [[buffer(9)]])
{

#line 135
[[kernel]] void framingAndWindowing(uint3 dispatchThreadID_2 [[thread_position_in_grid]], ComplexHalf_0 device* entryPointParams_peaseFFTHalfData_3 [[buffer(0)]], uint constant* entryPointParams_bufferSize_6 [[buffer(1)]], uint constant* entryPointParams_stage_6 [[buffer(2)]], ComplexHalf_0 device* entryPointParams_data_3 [[buffer(3)]], uint constant* entryPointParams_bufferSize_7 [[buffer(4)]], uint constant* entryPointParams_stage_7 [[buffer(5)]], bool constant* entryPointParams_isFinalStage_3 [[buffer(6)]], float device* entryPointParams_framingInput_3 [[buffer(7)]], half device* entryPointParams_framingOutput_3 [[buffer(8)]], AudioParams_0 constant* entryPointParams_framingParams_3 [[buffer(9)]])
{
```

So every shader is taking every buffer as input which seems like big performance hit and bug.

fairywreath:
Related to #6112.

Ensures Metal entry point parameters only contain global parameters that originates from itself(not other entrypoints) and from globally scoped global parameters.


csyonghe:
I think we need to keep the offset calculation to be for whole file for now. This is because if you have both a vertex and a fragment shader defined in the same module, you actually don’t want the parameters defined in different entry points to overlap. Parameters defined in global scope should get the same offset for each generated EntryPoint though.

source/slang/slang-ir-insts.h
```diff
@@ -2842,6 +2842,14 @@ struct IRGlobalVar : IRGlobalValueWithCode
 struct IRGlobalParam : IRInst
 {
     IR_LEAF_ISA(GlobalParam)
+
+    /// Get the entry point if this global param originates from an entry point parameter,
```
csyonghe, line 2846: Instead of extending global param with a new operand, I think it is better to add a new decoration on the global param, so that global params that is not coming from EntryPoint does not need to carry a null operand.

source/slang/slang-emit-c-like.cpp
```diff
@@ -5011,18 +5026,21 @@ void CLikeSourceEmitter::ensureInstOperandsRec(ComputeEmitActionsContext* ctx, I
         requiredLevel = EmitAction::ForwardDeclaration;
     }
 
-    for (UInt ii = 0; ii < operandCount; ++ii)
+    if (!areOperandsViewOnly(inst))
```
fairywreath, line 5029: I am not a big fan of this code but I do not see an easier/cleaner way to do this.
Without this `ensureInstOperands*` will complain about cyclic  code gen as `IREntryPointParamDecoration` contains a reference entry point func `IRFunc*`.
Another option is to not have  `IREntryPointParamDecoration` directly contain an operand to its originating entry point func, but I am not aware of a clean way to do this.

source/slang/slang-ir-insts.h
```diff
@@ -5226,6 +5234,11 @@ struct IRBuilder
     {
         addDecoration(inst, kIROp_CheckpointIntermediateDecoration, func);
     }
+
+    void addEntryPointParamDecoration(IRInst* inst, IRFunc* entryPointFunc)
+    {
+        addDecoration(inst, kIROp_EntryPointParamDecoration, entryPointFunc);
```
fairywreath, line 5240: `kIROp_EntryPointParamDecoration`, albeit being useless in the current code(before this PR), already exists with similar intentions of what this PR is doing and I am (re)using it.

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -110,10 +110,11 @@ static void registerLegalizedValue(
     context->mapValToLegalVal[irValue] = legalVal;
 }
 
-struct IRGlobalNameInfo
```
fairywreath, line 113: This struct exists but was useless/dead code.

source/slang/slang-emit-c-like.cpp
csyonghe: We can add a step during legalization to remove all entrypointParam decorations from global param insts. These decorations are no longer useful after legalization is done.

machineko:
Just to add my 2cents do we need Metal stuff to live in "global" scope if inputs to functions are always local to them? Is it not possible to start kernels from buffer(0) ... buffer(x-1)?

If i understand correctly Metal max size for index of input buffer is 31 so using more inputs in single file would break everything anyway.

fairywreath:
@machineko  There are a few reasons on why we are moving params to the global scope in Metal during IR legalization(@csyonghe correct me if I am wrong):
- As @csyonghe said if you have both a vertex and a fragment shader defined in the same module, you actually don’t want the parameters defined in different entry points to overlap. Hence we declare the entry point params to live in the global scope to cleanly process this, as we need to consider all parameters, including those that do not originate from entry point parameters themselves and actually live in the global scope.
- The metal backend legalization code is shared with other targets, where "global parameters" actually live in the global scope(which is not the case for Metal). We want to minimize code duplication hence we move Metal parameters out to the global scope so that we can reuse existing code. This is strictly for code quality.

Your request is a valid use case and we can definitely add an option/feature to allow overlapping offsets and have kernels start from  buffer(0) ... buffer(x-1) as you desire. It will be a separate PR from this, this PR mainly addresses the unnecessary parameters that is being added to all entry points.

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -3648,7 +3656,7 @@ static LegalVal legalizeGlobalVar(IRTypeLegalizationContext* context, IRGlobalVa
                 LegalVarChain(),
                 nameHint,
                 irGlobalVar,
-                &globalNameInfo,
+                nullptr,
```
csyonghe, line 3659: why do we need to pass in nullptr?

csyonghe (approved):

source/slang/slang-ir-legalize-types.cpp
fairywreath: The original `globalNameInfo` is actually dead code(it is not used at all in any of the child functions), at least before this PR. I am recycling that info pointer and it is only required to pass down information from a (non-legalized) global param to new vars that are created in the legalization process of the global param.

The highlighted code is inside `legalizeGlobalVar` which performs action on a global var and not a global param hence we do not actually need to pass down any info. `nullptr` is passed here because the parameter is still required as syntax for `declareVars` function signature. I am using the already existing signature and I'd rather pass a nullptr to explicitly state that argument will not and should not be used at all.

PR #6137, opened 2025-01-20
Fix maximal gcc downstream compiler optimization level

spking11:
There is no `-O4` in [gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html), and [anything above -O3 is the same as -O3](https://stackoverflow.com/a/30308151) actually.
More importantly, metal compiler will generate a warning `"metal 32023.404: : warning : -O4 is equivalent to -O3 [-Wdeprecated]"` when it is used with `-O4`.
So it‘s better to use `-O3` for `OptimizationLevel::Maximal`.


csyonghe (approved):
Looks good, thank you for the fix!

PR #6141, opened 2025-01-21
Use SPIRV integer vector dot product instructions

### LINKED ISSUE #6135: Use OpUDot/OpSDot for SPIRV
Integer dot products are supported by raw SPIRV instructions through the DotProduct capability, which is core in SPIRV 1.6 and requires and the `SPV_KHR_integer_dot_product` extension for SPIRV versions < 1.6. This should be used instead of the manual for loop that is currently implemented.

The only downside here is that Vulkan apps that use this need to enable the dot product extension, and some devices may not support it.

fairywreath:
Closes #6135.

Use SPIRV `OpUDot/OPSDot` instructions from the `DotProduct` capability to implement `dot`.


csyonghe (dismissed):
Looks good!

aleino-nv (approved):
Looks good to me!

(Aside: I think `OpUDot` and `OpSDot` would both work the same in the case where the operands are vectors (not packed) and the return type `T` is the same as both operand scalar types, but it doesn't hurt to distinguish AFAIK).

PR #6143, opened 2025-01-21
Catch all exceptions from loadModule* API calls.

### LINKED ISSUE #5950: Undefined exceptions thrown by public API
As far as I can tell `Slang::Exception` is not defined in the public headers, but sometimes due to ICE they get thrown out of the public API anyway.
`catch(...)` works, but there is no way to get any diagnostic text.

aleino-nv:
This closes #5950.


cheneym2 (approved):
LGTM

PR #6146, opened 2025-01-21
Fix bug: IgnoreInheritance in lookup

kaizhangNV:
When specifying IgnoreInheritance in lookup, it will ignore all members in the self extension for generic, the reason is that it doesn't specialize the target type of the extension decl when comparing with self type, so it will result that every type is unequal to the target type.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -2175,7 +2175,7 @@ void SemanticsVisitor::AddTypeOverloadCandidates(Type* type, OverloadResolveCont
         type,
         context.sourceScope,
         LookupMask::Default,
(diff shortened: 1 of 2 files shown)


source/slang/slang-check-overload.cpp
```diff
@@ -2175,7 +2175,7 @@ void SemanticsVisitor::AddTypeOverloadCandidates(Type* type, OverloadResolveCont
         type,
         context.sourceScope,
         LookupMask::Default,
-        LookupOptions::NoDeref);
```
csyonghe, line 2178: I think you want both NoDeref and IgnoreInheritance here.

csyonghe (approved):

PR #6153, opened 2025-01-22
Some usability improvements to formatting script

aleino-nv:
- Don't require binaries that wouldn't be used in certain cases
- Add option to only format files since a given revision (speeds up local use a lot)
  - Example: `extras/formatting.sh --cpp --no-version-check --since master`

### DIFF
--- a/extras/formatting.sh
+++ b/extras/formatting.sh
@@ -108,12 +108,12 @@ require_bin() {
 }
 
 require_bin "git" "1.8"
-require_bin "gersemi" "0.17"
-require_bin "xargs" "3"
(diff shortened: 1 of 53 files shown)


extras/formatting.sh
```diff
@@ -4,6 +4,7 @@ set -e
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 source_dir="$(dirname "$script_dir")"
+since_rev=
```
csyonghe, line 7: =0

aleino-nv: Done.

csyonghe (approved):

PR #6160, opened 2025-01-22
Add intptr_t abs/min/max operations for CPU & CUDA targets

juliusikkala:
This PR allows using abs/min/max with uintptr_t & intptr_t (= size_t as well).

This only covers the CPU and CUDA targets because they use the same code path, where this is easy to implement. For the rest of the targets, there seems to be many more obstacles, so I didn't do those in this PR (`Slang::getIntTypeInfo()` can't handle these types, so there's quite a few contexts where they don't work, it's not just these builtins)

Previously, e.g.
```hlsl
{
    size_t a = 1;
    size_t b = 2;

    printf("%lu\n", max(a, b));
    return 0;
}
```
would result in
```
(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: unexpected: unexpected type in intrinsic definition
```

### DIFF
--- a/prelude/slang-cpp-scalar-intrinsics.h
+++ b/prelude/slang-cpp-scalar-intrinsics.h
@@ -765,6 +765,39 @@ SLANG_FORCE_INLINE int64_t I64_max(int64_t a, int64_t b)
     return a > b ? a : b;
 }
 
+// ----------------------------- UPTR -----------------------------------------
+
+SLANG_FORCE_INLINE uintptr_t UPTR_abs(uintptr_t f)
+{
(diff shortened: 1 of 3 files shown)


csyonghe (dismissed):
Looks good to me!

juliusikkala:
So apparently NVRTC doesn't have intptr_t & uintptr_t. I defined them with ptrdiff_t & size_t for now. If that's not OK, I guess these abs/min/max operations can't be supported for the CUDA target.

csyonghe (approved):

PR #6161, opened 2025-01-23
Add an example for reflection of parameter blocks

tangent-vector:
The example loads a shader program, compiles it for Vulkan, and then uses reflection information to set up descriptor set layouts and a pipeline layout. It then validates that the pipeline layout that is created is compatible with the compiled code, by using them together to make a pipeline state object with the validation layer enabled.

The basic flow of the application follows the presentation in the companion article, and references its sections.

This is example could be expanded in a few ways:

* A D3D12 path could be added, to show the comparable operations for creating a root signature from reflection data

* The existing shader could be modified to touch/use more of its parameter data, to help ensure that any errors would be caught by the validation layer

* The set of shader files/modules that get loaded automatically could be expanded, to test more cases

* The code could be expanded to handle more than just compute shaders, by allowing for multiple-entry-point files to be loaded for rasterization or ray-tracing pipelines


csyonghe:
Need to fix build and format errors.

csyonghe (approved):

PR #6162, opened 2025-01-23
Fix incorrect resolve of specialization instance

jkwak-work:
While checking the uninitialized variables, we were not resolving the specialized instance correctly. This commit repeats the resolve while the result is a specialization instance. A new test is added for this:
  tests/diagnostics/uninitialized-generic.slang

After the problem is fixed, it revealed another problem in existing tests:
  tests/compute/nested-generics2.slang
  tests/diagnostics/uninitialized-local-variables.slang

When a struct has a member variable whose type is a generic type, we cannot iterate over its member variables yet, because the type is unknown until the generic function/struct is specialized. We will have to give up checking for these cases.

### DIFF
--- a/source/slang/slang-ir-use-uninitialized-values.cpp
+++ b/source/slang/slang-ir-use-uninitialized-values.cpp
@@ -142,8 +142,16 @@ static bool isDifferentiableFunc(IRInst* func)
 static IRInst* resolveSpecialization(IRSpecialize* spec)
 {
     IRInst* base = spec->getBase();
-    IRGeneric* generic = as<IRGeneric>(base);
-    return findInnerMostGenericReturnVal(generic);
+    while (auto baseSpecialize = as<IRSpecialize>(base))
+    {
+        base = resolveSpecialization(baseSpecialize);
+    }
+
+    auto baseGeneric = as<IRGeneric>(base);
+    if (!baseGeneric)
+        return base;
+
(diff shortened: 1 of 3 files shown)


source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -142,8 +142,16 @@ static bool isDifferentiableFunc(IRInst* func)
 static IRInst* resolveSpecialization(IRSpecialize* spec)
 {
     IRInst* base = spec->getBase();
-    IRGeneric* generic = as<IRGeneric>(base);
-    return findInnerMostGenericReturnVal(generic);
+    while (auto baseSpecialize = as<IRSpecialize>(base))
```
csyonghe, line 145: We shouldn't need this `resolveSpecialization` at all, just call `getResolvedInstForDecoration()`.

source/slang/slang-ir-use-uninitialized-values.cpp
```diff
@@ -174,6 +182,10 @@ static bool canIgnoreType(IRType* type, IRType* upper)
     if (as<IRInterfaceType>(type))
         return true;
 
+    // We don't know what TypeType will be yet.
+    if (as<IRParam>(type) && as<IRTypeType>(type->getDataType()))
```
csyonghe, line 186: shouldn't need the `as<IRTypeType>` part.

csyonghe (approved):

PR #6168, opened 2025-01-24
Fix depth texture sampling on Metal.

### LINKED ISSUE #6166: Texture2D should translate to depth2d in metal when used in SampleCmp
sample_compare function is only defined on depthND type and not textureND type in metal.

csyonghe:
We translate `Texture2D` to `texture2d` in Metal.

However, when the user calls `Texture2D::SampleCmp()`, we cannot translate it to `texture2d::sample_compare`, because `sample_compare` is only defined on `depth2d` type.

This PR fixes this issue by doing a bitcast to convert `texture2d` to `depth2d` before calling `sample_compare`, which is UB but seems to be compiling just fine by the metal compiler.

Closes #6166.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -855,6 +855,10 @@ float __glsl_texture_level_offset_1d_shadow<TTexture, TCoord, TOffset>(TTexture 
     }
 }
 
+__intrinsic_op($(kIROp_MetalCastToDepthTexture))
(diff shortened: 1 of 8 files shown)


source/slang/hlsl.meta.slang
```diff
@@ -1676,6 +1680,96 @@ extension _Texture<T,Shape,isArray,isMS,sampleCount,access,isShadow,0,format>
         }
     }
 }
+__generic<T:ITexelElement, Shape: __ITextureShape, let isArray:int, let isMS:int, let sampleCount:int, let format:int>
```
jkwak-work, line 1683: Maybe you can add [require(metal)] here.

jkwak-work (approved):
Looks good to me

PR #6169, opened 2025-01-24
Improve error message on generic value parameters decls

### LINKED ISSUE #6139: Improve compiler error message on malformed generic signature
I accidentally wrote the following incorrect generic code:
```
interface ITest
{
    static const uint kValue0;
    static const uint kValue1;
};

struct TestImpl<let TValue0 : uint, let TValue1> : ITest // Note the missing : uint after TValue1
{
    static const uint kValue0 = TValue0;
    static const uint kValue1 = TValue1;
};
```

This generated the following error:
`(0): error 30620: a variable declaration without an initial-value expression must be given an explicit type`
without giving the specific line on which the problem occured. I do not think this is a good error message for this case.
The struct has not been actually used anywhere, the error occured just when it was defined like this in the code.

fairywreath:
Closes #6139.


csyonghe (approved):
Looks good, thank you for fixing this!

PR #6177, opened 2025-01-24
Fix an incorrect iteration with a workList

jkwak-work:
We cannot modify workList while iterating it, because its type `List` is actually an array container.

### DIFF
--- a/source/slang/slang-ir-variable-scope-correction.cpp
+++ b/source/slang/slang-ir-variable-scope-correction.cpp
@@ -121,7 +121,7 @@ void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
 
     auto instAfterParam = funcInst->getFirstBlock()->getFirstOrdinaryInst();
 
-    for (int i = 0; i < workList.getCount(); i++)
+    for (Index i = 0; i < workList.getCount(); i++)
     {
         auto inst = workList[i];
         if (auto loopHeaderList = loopHeaderMap.tryGetValue(getBlock(inst)))
--- a/source/slang/slang-ir-variable-scope-correction.cpp
+++ b/source/slang/slang-ir-variable-scope-correction.cpp
@@ -121,12 +121,28 @@ void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
 
     auto instAfterParam = funcInst->getFirstBlock()->getFirstOrdinaryInst();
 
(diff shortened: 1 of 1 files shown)


csyonghe:
Actually it is fine to append items to a `List` while iterating it, and the newly appended item will be processed.

This is one major divergence from std containers, which doesn't allow this behavior. In practice we find allowing this makes a lot of code simpler (and potentially faster).

Is the code leading to any crashes or bugs?

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -121,12 +121,28 @@ void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
 
     auto instAfterParam = funcInst->getFirstBlock()->getFirstOrdinaryInst();
 
-    for (auto inst = workList.begin(); inst != workList.end(); inst++)
+    while (true)
```
kaizhangNV, line 124: I think we should consistently using `for(;;)`

kaizhangNV:
> Actually it is fine to append items to a `List` while iterating it, and the newly appended item will be processed.
> 
> This is one major divergence from std containers, which doesn't allow this behavior. In practice we find allowing this makes a lot of code simpler (and potentially faster).
> 
> Is the code leading to any crashes or bugs? The for loop should be iterating with an integer index instead of using the iterator, if the iterator is just a pointer (and that pointer will change if the list reallocates).

kaizhangNV:
>>Is the code leading to any crashes or bugs? The for loop should be iterating with an integer index instead of using the iterator, if the iterator is just a pointer (and that pointer will change if the list reallocates).

Yea, I have a same ask here

jkwak-work:
It was causing a crash.
I will try with index as suggested.

source/slang/slang-ir-variable-scope-correction.cpp
```diff
@@ -121,11 +121,12 @@ void VariableScopeCorrectionContext::_processFunction(IRFunc* funcInst)
 
     auto instAfterParam = funcInst->getFirstBlock()->getFirstOrdinaryInst();
 
-    for (auto inst = workList.begin(); inst != workList.end(); inst++)
+    for (int i = 0; i < workList.getCount(); i++)
```
csyonghe, line 124: Should be `for (Index i = 0`.

csyonghe (approved):

PR #6178, opened 2025-01-25
Fix CompileOptions defined in targetDesc not loaded into TargetRequest.

### LINKED ISSUE #6179: CompileOptions defined in targetDesc are not loaded into TargetRequest.

csyonghe:
Also add a test to cover type-conformance compilation API.

Closes #6179


EtienneParmentier:
Just in case, this PR does not closes #6178 . The original source code is not embeded in the output spirv.

jkwak-work (approved):
Looks good to me.

source/slang/slang.cpp
```diff
@@ -1337,6 +1337,7 @@ void Linkage::addTarget(slang::TargetDesc const& desc)
     optionSet.setProfile(Profile(desc.profile));
     optionSet.set(CompilerOptionName::LineDirectiveMode, LineDirectiveMode(desc.lineDirectiveMode));
     optionSet.set(CompilerOptionName::GLSLForceScalarLayout, desc.forceGLSLScalarBufferLayout);
+    optionSet.load(desc.compilerOptionEntryCount, desc.compilerOptionEntries);
```
aleino-nv, line 1340: I see, yes this might also do what I need if load will overwrite the already existing debug setting (which is copied from the linkage options in my case).

aleino-nv (approved):

Issue #6175, opened 2025-01-24
Support GLSL/WGSL float packing intrinsics


Runi-c:
I'm converting some WGSL to Slang and had a slight hitch on `unpack2x16float` and `unpack2x16snorm`.

GLSL and SPIR-V has these via GLSL.std.450 under the names `UnpackHalf2x16`, `UnpackSnorm2x16`, etc.

But it seems like there's no way to emit these instructions on these build targets with Slang, this would be nice to have.

Alternatives I'm aware of:

- Implement manually leveraging `f16tof32`
- Use GLSL syntax directly using `-allow-glsl` (but this won't compile to my target WGSL)

csyonghe:
Yes, we should have it defined. On HLSL I guess we can just implement it with bit operations + type cast.

csyonghe:
Yes, we should create one, but given this is a simple intrinsic addition, we can keep the proposal short and simple.

fairywreath:
Ok thanks, I can work on this

PR #6191, opened 2025-01-27
Add floating-point pack/unpack proposal

fairywreath:
Related to #6175. Adds a proposal to define floating-point pack/unpack intrinsics.


docs/proposals/018-floating-point-packed-data-intrinsics.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Normalized float conversion rules
+Normalized float conversion rules are standard across GLSL/SPIRV, Metal and WGSL. Slang follows these standards. Details of the conversion rules for each target can be found in:
+- Section 8.4, `Floating-Point Pack and Unpack Functions`, of the GLSL language specification, which is also used by the SPIR-V extended instrucsion for GLSL.
+- Section 7.7.1, `Conversion Rules for Normalized Integer Pixel Data Types`, of the Metal Shading language specification.
+- Sections [16.9  Data Packing Built-in Functions](https://www.w3.org/TR/WGSL/#pack-builtin-functions) and [16.10 Data Unpacking Built-in Functions](https://www.w3.org/TR/WGSL/#unpack-builtin-functions) of the WebGPU Shading language specification.
+
+### Built-in packed Datatypes
+Unlike HLSL's implementation with introduces new packed datatypes, `uint8_t4_packed` and `int8_t4_packed`, unsigned 32-bit integers are used directly 
+and no new pakced datatypes are introduced.
```
fairywreath, line 80: I don't have a strong preference here and would really appreciate any suggestions or input. If new datatypes are to be introduced we would add something like `half2_unorm_packed`, `half2_snorm_packed`, `half2_packed`, then another full set for the 8-bit variants.

csyonghe (approved):
Looks good to me. Are these functions named the same way after the corresponding HLSL intrinsic? Or should we name them using Slang's naming convention?

fairywreath:
> Looks good to me. Are these functions named the same way after the corresponding HLSL intrinsic? Or should we name them using Slang's naming convention?

The proposal currently names them based on HLSL's intrinsics - I believe using Slang's naming convention is better. I will change them and will also add new intrinsics for the integer pack/unpack functions that follows Slang's naming convention in addition to the hlsl intrinsics.

csyonghe (approved):

PR #6204, opened 2025-01-28
Fix exact-match witness synthesis for static functions

### LINKED ISSUE #6195: Slang asserts when interface and implementation mismatch on static keyword
When interface declares a method as `static` but implementation implements it as not-static, Slang does not catch that in syntax analysis and instead fails this assert `SLANG_RELEASE_ASSERT(callParamCount == callArgCount);` in `propagateConstExprBackward` in `slang-ir-constexpr.cpp`.

A better error message would be desired.

Full repro case:
```
interface Base
{
    static int getValue();
}

struct Impl : Base
{
    int getValue() { return 5; } // add static to fix
}

int callGet<Foo : Base>()
{
    return Foo::getValue();
}

RWStructuredBuffer<int> result;

[numthreads(1, 1, 1)]
void testHashedStrings()
{
    result[0] = callGet<Impl>();
}
```

fairywreath:
Closes #6195. 

Removes non-static lookup items when trying to synthesize an exact-match witness for a static interface method.
The current behavior allows them and asserts/crashes the compiler during lowering because the non-static method has an extra this argument that the static interface cannot handle. This change  removes them from being considered during synthesis to properly generate an error message instead.

Also adds a test case to ensure static methods can implement non-static method interfaces.


csyonghe (approved):
Looks good!

PR #6205, opened 2025-01-28
Fix UIntSet calcSubtract to handle mismatched buffer sizes

### LINKED ISSUE #6201: Missing capability names in warning output
Hi, I recently updated from an older version of slang, and got the following warning from the newer version of the compiler:

```
entry point 'pixel_main' uses additional capabilities that are not part of the specified profile 'spirv_1_3'. The profile setting is automatically updated to include these capabilities: ''
```

Unfortunately the warning does not specify the missing capabilities, making the source of this warning difficult to track down, but by trial and error I found that adding `texture_gather` fixes it.  I thought I should raise an issue due to this name being missing from the warning output.

This can be reproduced on godbolt with slang 2025.2.2 using the following shader:

```slang
[shader("pixel")]
float4 pixel_main(float2 pos : SV_Position, Sampler2D<float> foo) : SV_Target
{
    return foo.Gather(uint2(pos));
}
```

Compiling with args `-profile spirv_1_3` produces the warning above, changing args to `-profile spirv_1_3+texture_gather` silences it.

fairywreath:
Closes #6201.

Fixes an issue where missing capability names are not printed out. A `UIntSet` created from `calcSubtract` was deemed empty because `calcSubtract` does not properly copy over elements.


csyonghe (approved):
Yeah, this set subtract is obviously wrong. Thank you for the fix!

PR #6207, opened 2025-01-28
Fix combined sampler documentation and warning

cheneym2:
This change updates documentation around support for combined texture/samplers
for SPIR-V. The existing text implied that it was correct to use two register assignments
for explicit binding in the SPIR-V docs, but there is actually no way for Slang to derive
one vkbinding from two registers.  The warning was misleading.

Related, there is a set of -fvk-xxx-shift options in slangc that _do_ offer a way to specify
mappings from registers to vk bindings, however, they only work for 1:1 situations, not
the 2:1 mapping that comes up with combined texture samplers.

The warning which comes up when a user specifies "register" but not "vkbinding" nor
-vk-xxx-shift, is now updated to add "-vk-xxx-shift" as a suggestion for fixing the problem,
but only for the case of non-combined texture samplers.

Closes Issue #5938


source/slang/slang-diagnostic-defs.h
```diff
@@ -2019,6 +2019,13 @@ DIAGNOSTIC(
     notValidVaryingParameter,
     "parameter '$0' is not a valid varying parameter.")
 
+DIAGNOSTIC(
+    39029,
+    Warning,
+    registerModifierButNoVkBindingNorShift,
+    "shader parameter '$0' has a 'register' specified for D3D, but no '[[vk::binding(...)]]` "
+    "specified for Vulkan, nor is `-fvk-xxx-shift` used.")
```
csyonghe, line 2027: let's be more concrete: `-fvk-$1-shift`

tests/diagnostics/hlsl-to-vulkan-sampler-diagnostic.hlsl
```diff
@@ -0,0 +1,30 @@
+//DIAGNOSTIC_TEST:SIMPLE:-target glsl -profile ps_4_0 -entry main -fvk-t-shift 5 all -fvk-t-shift 7 2  -fvk-s-shift -3 0 -fvk-u-shift 1 2 -no-codegen
+//DIAGNOSTIC_TEST:SIMPLE:-target glsl -profile ps_4_0 -entry main -no-codegen
+
+// This tests that combined texture sampler objects which have D3D style register assignments, but no vk::binding, show an appropriate
+// warning, even in the presence of -fvk-xxx-shift options. Warning should not recommend using -fvk-xxx-shift.
```
csyonghe, line 5: the comment here may need to be updated.

cheneym2: Done, though I'm not sure if my update meets your expectations.  What I said in this version of the comment is what I believe to be correct.  Hoping it's just a matter of unclear wording.

source/slang/slang-diagnostic-defs.h
cheneym2: Done

csyonghe (approved):

tests/diagnostics/hlsl-to-vulkan-sampler-diagnostic.hlsl
csyonghe: Ah, never mind, the comment is correct. I misunderstood it.

PR #6212, opened 2025-01-29
Fix loophole in hoisting where an `OpVar`'s uses might not be properly registered for replacement

saipraveenb25:


source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -500,6 +514,15 @@ RefPtr<HoistedPrimalsInfo> AutodiffCheckpointPolicyBase::processFunc(
                             callVarWorkList.add(callUser);
                     }
                 }
+
+                // This is a bit of a hack.. ideally we need to add the var to the worklist for
+                // further processing rather than replicating those operations here.
+                //
```
csyonghe, line 520: Why can’t we just do as the comment said?

csyonghe (approved):

source/slang/slang-ir-autodiff-primal-hoist.cpp
saipraveenb25: I believe the body of the loop needs a small rewrite. I can do this in a follow-up PR.

saipraveenb25: Opened an issue to track this cleanup: https://github.com/shader-slang/slang/issues/6214

PR #6213, opened 2025-01-29
Add support for WGSL subgroup operations

### LINKED ISSUE #6208: Support shader subgroup/wave intrinsics for WGSL

fairywreath:
Related to #6208.

Adds support for WGSL subgroup operations. Most map directly to HLSL and GLSL intrinsics, with some exceptions being:
- WGSL's `quadBroadcast` only accepts const ids while HLSL's and GLSL's intrinsics accept non-const. The current HLSL and GLSL functions in Slang cannot target WGSL.

There are some things that still need to be done in a follow-up PR:
- `WaveGetLaneIndex()` and `WaveGetLaneCount()` are not implemented. They are functions in HLSL but are built-in system values in WGSL. A non-trivial implementation of adding built-in values to the generated entry point parameter list is required.
- Some GLSL `subgroup*` arithmetic/bitwise intrinsics also accept bool while WGSL does not allow bool. The current implementation currently allows them.
- We now have multiple IR extension AST types for different targets(and maybe even more in the future), unify them to a single type.
- f16 subgroup operations require the `subgroups_f16` extension and they are not handled cleanly in this PR(although the produced WGSL code is valid), they should be improved to check for the for the float type in Slang meta file.


csyonghe (approved):
Looks good to me. Thank you!

csyonghe (commented):
Would be good if we have some test cases.

source/slang/slang-ir-insts.h
```diff
@@ -422,6 +422,15 @@ struct IRRequireGLSLExtensionDecoration : IRDecoration
     UnownedStringSlice getExtensionName() { return getExtensionNameOperand()->getStringSlice(); }
 };
 
+struct IRRequireWGSLExtensionDecoration : IRDecoration
+{
```
csyonghe, line 426: I feel like we may want something like `RequireTargetExtension(glsl, "...")` instead of having one op code per target. But we don't need to change this in this PR.

csyonghe (dismissed):

fairywreath:
I still have some todos left listed in the PR description, but I'd rather have them completed in follow-up PRs.

csyonghe (dismissed):

csyonghe:
needs a format fix.

csyonghe (approved):

PR #6223, opened 2025-01-30
Support cooperative vector

jkwak-work:
This commit has Slang implementations for supporting cooperative vector.
Vulkan and DX12 tests are disabled for now.


source/slang/slang-ir.cpp
```diff
@@ -4171,6 +4197,18 @@ IRInst* IRBuilder::emitGetNativeString(IRInst* str)
     return emitIntrinsicInst(getNativeStringType(), kIROp_getNativeStr, 1, &str);
 }
 
+IRInst* IRBuilder::emitGetElement(IRType* type, IRInst* arrayLikeType, IRIntegerValue element)
```
csyonghe, line 4200: This shouldn't be needed, we already have `emitElementExtract` and `emitElementAddress`.

source/slang/slang-emit.cpp
```diff
@@ -1016,6 +1017,16 @@ Result linkAndOptimizeIR(
 #endif
     validateIRModuleIfEnabled(codeGenContext, irModule);
 
+    switch (target)
```
csyonghe, line 1020: We need to do a prescan for `IRCoopVecType` in `calcRequiredLoweringPassSet`, and only call the pass if we detected any related insts. This is for performance.

source/slang/core.meta.slang
```diff
@@ -1664,6 +1664,18 @@ struct matrix : IRWArray<vector<T,C>>
     int getCount() { return R; }
 }
 
+__generic<T:__BuiltinArithmeticType, let N:int>
+__magic_type(CoopVectorExpressionType)
+__intrinsic_type($(kIROp_CoopVectorType))
+struct CoopVector : IArray<T>
```
csyonghe, line 1670: Should delete this too.

This PR contains: `struct CoopVector`, `struct CoopVec`, and `struct coopvecNV`. We just need `CoopVec`.

source/slang/slang-ir.cpp
jkwak-work: It doesn't seem like I can simply replace the keyword `emitGetElement` with `emitElementExtract`.

I will address this as a separate PR later.

source/slang/slang-emit.cpp
jkwak-work: I will address this as a separate PR later.

source/slang/slang-ast-type.cpp
```diff
@@ -355,6 +355,29 @@ Type* AtomicType::getElementType()
...
+}
+
+IntVal* CoopVectorExpressionType::getElementCount()
+{
+    return as<IntVal>(_getGenericTypeArg(this, 1));
+}
+
+void CoopVectorExpressionType::_toTextOverride(StringBuilder& out)
+{
+    out << toSlice("CoopVector<") << getElementType() << toSlice(",") << getElementCount()
```
csyonghe, line 372: CoopVec

csyonghe (approved):
Looks good, minor comments.

csyonghe:
Why is this marked as breaking? What is breaking?

jkwak-work:
> Why is this marked as breaking? What is breaking?

If the users were using CoopVector or coopvecNV, it will break.
Do you think it should be non-breaking?

source/slang/slang-ast-type.cpp
jkwak-work: I think this code is for targeting hlsl.

source/slang/hlsl.meta.slang
```diff
@@ -12414,10 +12430,50 @@ __generic<T:__BuiltinArithmeticType, U:__BuiltinArithmeticType>
 __intrinsic_op($(kIROp_IntCast))
 T __int_cast(U val);
 
+__generic<T:__BuiltinArithmeticType, U:__BuiltinArithmeticType>
+__intrinsic_op($(kIROp_FloatCast))
+T __float_cast(U val);
```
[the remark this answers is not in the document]
jkwak-work: I will remove it

csyonghe:
No existing Slang code in the public will contain those, so this is not a breaking change.

source/slang/hlsl.meta.slang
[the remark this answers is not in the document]
jkwak-work: Hmmm. It seems like I cannot simply replace __float_cast with __realCast.
There are two overloadings for `__realCast()`:
 - __BuiltinRealType -> __BuiltinRealType and
 - __BuiltinRealType -> __BuiltinIntegerType

But what we need with `__float_cast()` is:
 - __BuiltinArithmeticType -> __BuiltinArithmeticType

I will rename __float_cast to __real_cast to follow the pattern.

source/slang/slang-ast-type.cpp
csyonghe: No, we shouldn’t need this override. toText here should return the string that represent the type in Slang, not any specific target language. If the user call toText and then use the result to generate Slang code, that Slang code should compile by Slang.

csyonghe: I think we should delete the override here and use what is provided in the base class (DeclRefType).

PR #6224, opened 2025-01-30
Fix ConstantIntVal::toText when the val is a enum.

### LINKED ISSUE #6225: Type::toText should print enum names instead of raw int values.
Given:
```
namespace ns {
enum Test : uint32_t
{
    A = 1,
    B = 2,
}

struct Foo<let T : Test>
{
}
}

ns.Foo<ns.Test.A> first;
```

When printing the type of `first`, we should output `"ns.Foo<ns.Test.A>"` instead of `"ns.Foo<1>"`. This can be an issue if the user is using this reflection info to generate new Slang code, since `ns.Foo<1>` will fail type check.

csyonghe:
Closes #6225.

### DIFF
--- a/source/slang/slang-ast-decl.h
+++ b/source/slang/slang-ast-decl.h
@@ -210,6 +210,8 @@ class EnumCaseDecl : public Decl
 
     // Tag value
     Expr* tagExpr = nullptr;
+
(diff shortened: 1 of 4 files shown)


saipraveenb25 (approved):
Looks good to me!

PR #6228, opened 2025-01-30
Support cooperative vector including Vulkan-Header

jkwak-work:
Adding a Slang support for cooperative vector with vulkan-header update.


csyonghe (approved):

jkwak-work:
Coopvec tests were failing for "slang-test via glsl" step.
I disabled those tests, and they can be reenabled once we have glsl side support on glslang.

PR #6231, opened 2025-01-30
Declare `VariablePointers` capability when needed.

### LINKED ISSUE #6217: Support VariablePointers capability
We will need to emit `VariablePointers` capability whenever we are emitting OpLoad or OpStore whose value type is a pointer to physical storage buffer storage class.

csyonghe:
Closes #6217.


jkwak-work (approved):
Looks good to me.

PR #6233, opened 2025-01-31
Distribute slang-glsl-module.bin in release packages.

### LINKED ISSUE #6232: Extend CMake build process to run slangc to produce slang-glsl-module.bin and include it in CPACK.
During the first run of slangc.exe, we will generate slang-glsl-module.bin if it doens't exist.

We should include this file in our releases.

This can be done by extending our cmake file to run slangc after building it, and then make sure the generated slang-glsl-module.bin is included in cpack.

csyonghe:
Closes #6232.


aleino-nv (approved):

source/slang-record-replay/util/emum-to-string.h
cheneym2, line 1: Enum

PR #6242, opened 2025-02-01
Support HLSL target for CoopVec::MatMul(Add) with (RW)StructuredBuffer

jkwak-work:
This commit allows HLSL targeting CoopVec to use MatMul(Add) with (RW)StructuredBuffer.

However, this commit itself is not enough to provide the functionality.
I am going to keep this as draft until DXC support becomes available.


jkwak-work:
This is related to https://github.com/shader-slang/slang/issues/6252

PR #6243, opened 2025-02-01
Fix build issue when SLANG_EMBED_CORE_MODULE is OFF.

csyonghe:


source/slang-glsl-module/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 #
 # Generate an embeddable glsl module
 #
+if (${SLANG_EMBED_CORE_MODULE})
 
 set(glsl_module_generated_header_dir ${CMAKE_CURRENT_BINARY_DIR})
 set(glsl_module_generated_header
```
jkwak-work, line 7: You may want to add an indentation inside of the new if-endif

jkwak-work (approved):
Looks good to me.
And I tested it on my local machine. It worked.

Issue #6144, opened 2025-01-21
Build option NVAPI_ROOT_DIR doesn't work properly


jkwak-work:
It appears that the CMake build option, NVAPI_ROOT_DIR, is not working.

**Repro step**
Download NVAPI SDK from [here](https://developer.nvidia.com/tools-downloads#?search=nvapi)
```
cmake.exe --preset vs2019 -DNVAPI_ROOT_DIR=../../nvapi/R560-developer/
cmake.exe --build build --config Release

D:\sbf\git\slang\nvapi_root_dir_not_working_properly\tools\gfx\d3d11\../nvapi/nvapi-include.h(21,10): fatal error C1083: Cannot open include file: 'nvShaderExtnEnums.h': No such file or directory (compiling source file D:\sbf\git\slang\nvapi_root_dir_not_working_properly\tools\gfx\d3d11\d3d11-helper-functions.cpp) [D:\sbf\git\slang\nvapi_root_dir_not_working_properly\build\tools\gfx.vcxproj]
```

jkwak-work:
I see that there is cmake file that is related to NVAPI.
https://github.com/shader-slang/slang/blob/master/cmake/FindNVAPI.cmake

But it has been unused since we moved from Premake to CMake.

The missing part looks to be that NVAPI_ROOT_DIR should be used as an include directory, but it is not used at all.

@expipiplus1  for visibility.

jkwak-work:
It appears that we are using a feature in CMake to find NVAPI and OpiX.
But they are not properly wired.

https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html?utm_source=chatgpt.com
https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules

expipiplus1:
These files are found by the auto_option function, the problem is that the include directory wasn't being passed to gfx

PR #6263, opened 2025-02-04
Correct NVAPI include

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/6144


jkwak-work (approved):
Looks good to me.

I tested it on my local machine and it worked.

Issue #5755, opened 2024-12-04
DCE eliminates empty functions even if they have user-defined derivatives


saipraveenb25:
Test-case to reproduce this behavior
```csharp
//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type
//TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -shaderobj -output-using-type

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<float> outputBuffer;

void foo_bwd(float a, inout DifferentialPair<float> dpx)
{
    outputBuffer[2] = 2.f;
}

[Differentiable, BackwardDerivative(foo_bwd)]
void foo(no_diff float a, float x)
{ }

[Differentiable]
float outerFunc(no_diff float a, float x)
{
    foo(a, x);
    return 1.f;
}

[numthreads(1, 1, 1)]
[shader("compute")]
void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
{
    float a = 10.0;
    DifferentialPair<float> dpx = DifferentialPair<float>(4.f, 1.f);
    bwd_diff(outerFunc)(a, dpx, 1.0);
    
    // Expect 2.0 in outputBuffer[2]
}
```

kaizhangNV:
I think DCE might not be the only problem, I found out that we might generate wrong code for the output write instruction in the bwd functions. Please just take a notice about that.

bmillsNV:
@saipraveenb25 sounds like Tony needs some help from you on this. Can you help?

saipraveenb25:
This test actually doesn't reproduce this behavior.. 

Turns out our tests are run with `-g` on, which inserts `DebugVar` insts for debugging and when these are around, a lot of functions are not DCE'd.

To reproduce the result, you should run the test with the top line replaced with: 
```//TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -shaderobj -output-using-type -g0```

PR #6272, opened 2025-02-04
Fix DCE for calls to functions that have associations

saipraveenb25:
DCE will now check for side-effect-ness on other functions that are associated with the callee being checked.

Fixes: #5755

### DIFF
--- a/source/slang/slang-ir-util.cpp
+++ b/source/slang/slang-ir-util.cpp
@@ -1218,8 +1218,14 @@ bool isSideEffectFreeFunctionalCall(IRCall* call, SideEffectAnalysisOptions opti
 // Enumerate any associated functions of 'func'
 // that might be used by a pass (e.g. auto-diff)
 //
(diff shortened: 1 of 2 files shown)


source/slang/slang-ir-util.cpp
```diff
@@ -1215,19 +1215,79 @@ bool isSideEffectFreeFunctionalCall(IRCall* call, SideEffectAnalysisOptions opti
...
     }
-    return true;
+
+    // If the callee has no side effect, check if any of its associated functions have side effect.
+    // If so, we want to keep the callee around.
+    //
+    // Typically, once the relevant pass has completed, the association is removed,
+    // and at that point we can remove the function.
+    //
+    if (!sideEffect)
```
csyonghe, line 1278: Given that this function is called very frequently, let's avoid allocating memory here. `getAssociatedFunction` can perhaps turned into `enumerateAssociatedFunction(callback)`.

saipraveenb25: Fixed now

PR #6275, opened 2025-02-04
Fix autodiff issue for vector<T, N>

### LINKED ISSUE #6154: [slangpy] Torch integration test failure
The polynomial test from `test_torchintegration.py` from [slangpy](https://github.com/shader-slang/slangpy/tree/dev/ccummings/nativeslangpy2) tests currently fails due to incorrect gradients from slang (gradients are unexpectedly 0)

Image

kaizhangNV:
Close #6154

We didn't implement correctly for vector<T, N> regarding the differentiablity. As we check differentiable before specialization, however according to the definition of vector, it has to be specialized to IFloat to know it's conformed to IDifferential type. Therefore for parameter type vector<T, N> will become no_diff.

Therefore, we change the implementation a to make it explicit conform to IDifferential type.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -2054,8 +2054,8 @@ extension matrix<T,N,M,L> : IFloat
     __init(float v) { this = matrix<T,N,M>(T(v)); }
 }
 
-__generic<T:__BuiltinFloatingPointType, let N : int, let M : int, let L : int>
-extension matrix<T,N,M,L> : IFloat
+__generic<T:IDifferentiable, let N : int, let M : int, let L : int>
+extension matrix<T,N,M,L> : IDifferentiable
 {
     // IDifferentiable.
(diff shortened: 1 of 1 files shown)


kaizhangNV:
It passes `test_torchintegration.py` on my local.

source/slang/core.meta.slang
```diff
@@ -2042,7 +2052,11 @@ extension matrix<T,N,M,L> : IFloat
     [__unsafeForceInlineEarly]
     __implicit_conversion($(kConversionCost_ScalarToMatrix))
     __init(float v) { this = matrix<T,N,M>(T(v)); }
+}
 
+__generic<T:__BuiltinFloatingPointType, let N : int, let M : int, let L : int>
```
csyonghe, line 2057: This should be :IDifferentiable, right?

csyonghe (approved):

PR #6289, opened 2025-02-05
maxtessfactor attribute should take a floating point value

jkwak-work:
It looks like `[maxtessfactor(XX)]` is useful only when targeting HLSL.
There isn't corresponding keyword for GLSL nor SPIR-V.


source/slang/slang-check-modifier.cpp
```diff
@@ -312,6 +312,22 @@ AttributeDecl* SemanticsVisitor::lookUpAttributeDecl(Name* attributeName, Scope*
...
 
+bool SemanticsVisitor::hasFloatArgs(Attribute* attr, int numArgs)
+{
+    if (int(attr->args.getCount()) != numArgs)
+    {
+        return false;
+    }
+    for (int i = 0; i < numArgs; ++i)
+    {
+        if (!as<FloatingPointLiteralExpr>(attr->args[i]))
```
csyonghe, line 323: we should allow integer values as well.

jkwak-work: I tried to use `tryFoldIntegerConstantExpression()` as suggested.
But I couldn't figure how to properly use it.

`slang-check-modofier.cpp` is just for printing errors.
And `_getFloatFromAttribute()` is where it needs to call `tryFoldIntegerConstantExpression()` but they belong to different classes and I couldn't figure where is a good place to fold it.

csyonghe: I think the current PR is fine, it will just not work when the user writes `[maxtessfactor(1+1)]`, which I don't think will come up in practice.

If you scroll down the function you can see lines like:
```
auto cint = checkConstantIntVal(attr->args[0]);
if (cint)
    forceUnrollAttr->maxIterations = (int32_t)cint->getValue();
```
If `cint` is nullptr, then you know it is not an int.

csyonghe (approved):

tests/render/tess.hlsl
```diff
@@ -50,6 +50,7 @@ HS_CONSTANT_OUTPUT HSConst()
 [partitioning("integer")]
 [outputtopology("line")]
 [outputcontrolpoints(4)]
+[maxtessfactor(16.0)]
```
csyonghe, line 53: we should also add a separate spirv compile test and verify the right spirv execution model gets emitted from this.

jkwak-work: `[maxtessfactor]` is ignored when targeting GLSL and SPIRV.
Because there isn't any spirv instructions for `maxtessfactor`.

But I guess we can emit the execution model when `maxtessfactor` is used.

jkwak-work: It looks like the execution mode for hull shader is emitted by `[partitioning("XXX")]`.
https://github.com/shader-slang/slang/blob/f6cbb81e1c0080518185294ee94705f5e93aa849/source/slang/slang-emit-spirv.cpp#L4415

I guess `maxtessfactor` by itself cannot tell use which mode it should be.

Issue #6279, opened 2025-02-04
Support SM 6.8 in Slang Gfx


pclarberg:
sm68 is needed for CoopVector support on D3D12.

jkwak-work:
In order to use CoopVec feature with dxcompiler.dll, Gfx had to enable "d312 experimental feature".
Initially I tried to enable it all the time.
But I observed that things became unstable when it is enabled.
So I had to make it turned off by default and it will be enabled with a command line argument "-dx12-experimental".

PR #6290, opened 2025-02-05
Enable D3D12 experimental feature to use coopvec

jkwak-work:
Close https://github.com/shader-slang/slang/issues/6279

This commit enables "D3D12-experimenta-feature" in render-test.
This is required to use CoopVec feature with dxcompiler.dll.

But it is enabled only when "-dx12-experimental" is used, because it appears that DX12 becomes unstable when the experimental feature is enabled, which causes bunch of tests randomly failing.


jkwak-work:
A lot of dx12 tests are failing; if not all.
I need to investigate.

tools/render-test/render-test-main.cpp
```diff
@@ -13,6 +13,10 @@
 #include "slang-support.h"
 #include "window.h"
 
+#if defined(_WIN32)
```
csyonghe, line 16: Why is this include needed?

jkwak-work: It is needed for the type `UUID`.

jkwak-work:
In order to use CoopVec feature with dxcompiler.dll, Gfx had to enable "d312 experimental feature".
Initially I tried to enable it all the time.
But I observed that things became unstable when it is enabled.
So I had to make it turned off by default and it will be enabled with a command line argument "-dx12-experimental".

csyonghe (approved):

PR #6291, opened 2025-02-05
Import into the scope of the import decl, rather than module scope

### LINKED ISSUE #6221: Multiple source files with module imports only work on the last source file
When doing a single step compile with multiple source files using slangc, any imports for any source file other than the last one in the command line don't seem to be visible.  The following, for example:
A.slang:
```
module A;

public int test() { return 1; }
```
source1.slang:
```
import A;

[shader("closesthit")]
void CH1() { int a = test(); }
```
source2.slang:
```
[shader("closesthit")]
void CH2() {}
```
After compiling A.slang to A.slang-module, trying to compile source1 and source2 together (`slangc source1.slang source2.slang -o test.spv`) results in "undefined identifier: 'test'".  If the source1 and source2 arguments are reversed in the command line, it compiles.  If I compile source1 and source2 to their own slang-modules and then link all 3 together, it also compiles fine.
I did not see anything in the docs suggesting multiple source files aren't allowed when using modules.

aleino-nv:
When a module consists of multiple source files, the module scope gets over-written for each source file that's parsed into the module.
On the other hand, SemanticsVisitor::importModuleIntoScope will import into the scope of the module corresponding to the import declaration. The result is that if you do something like the following, where source1.slang contains an import statement, then the imported module will get imported into the module scope corresponding to source2.slang.

    slangc source1.slang source2.slang # 1 module from 2 source files

This patch fixes this problem by importing into the scope of the import statement, rather than the scope of the corresponding module, since that may have been overwritten.

This closes #6221.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -10178,7 +10178,12 @@ void SemanticsDeclHeaderVisitor::visitImportDecl(ImportDecl* decl)
     // the module's scope.
 
     auto name = decl->moduleNameAndLoc.name;
-    auto scope = getModuleDecl(decl)->ownedScope;
+
+    // We should import the module into the same scope as the decl.
+    // Note that if the module consists of multiple source files, the scope of the module may not
+    // correspond to the scope of this import declaration, since the module scope gets overwritten
+    // each time a new source file is parsed into it.
+    auto scope = decl->scope;
 
     // Try to load a module matching the name
     auto importedModule = findOrImportModule(


aleino-nv (commented):
I will try to repro with the new compilation API and add some test coverage for this.

aleino-nv:
From tests/language-feature/modules/import-in-include/c.slang it seems like imports from another file in the translation unit making up the module
should be visible to other files in the translation unit.

Some other references:

From https://shader-slang.org/slang/user-guide/compiling.html#concepts:
> Source units (such as files) are grouped into translation units, and each translation unit will produce a single module when compiled.

From https://shader-slang.org/slang/user-guide/compiling.html#command-line-compilation-with-slangc:
> If there are any .slang files, then all of them will be grouped into a single translation unit

From https://shader-slang.org/slang/user-guide/modules.html:
> A module can contain more than one file. The additional files are pulled into the module with the __include syntax


...with all this in mind, the fix is instead to not be pushing new scopes when we parse files from the same translation unit.

aleino-nv:
See https://github.com/shader-slang/slang/pull/6292

PR #6292, opened 2025-02-05
Don't overwrite existing container decl scopes when parsing new sourc…

### LINKED ISSUE #6221: Multiple source files with module imports only work on the last source file

aleino-nv:
…e files.

When a module consists of multiple source files, the module scope gets over-written for each source file that's parsed into the module.
The result is that if you do something like the following, where source1.slang contains an import statement, then the imported module will get imported into the module scope corresponding to source2.slang, but won't be found from the scope of source1.slang.

slangc source1.slang source2.slang # 1 module from 2 source files

This patch fixes this problem by not over-writing existing container decl scope when parsing new source files into the container.

This closes #6221.


csyonghe:
We should include a test for #6221, you can create two files:
`tests/language-feature/modules/multi-file-tu/a.slang` - for module A.
`tests/language-feature/modules/multi-file-tu/source1.slang`
`tests/language-feature/modules/multi-file-tu/source2.slang`

then you can add a `//TEST_IGNORE_FILE:` on a.slang and source2.slang, and add this line to source1.slang:
```
//TEST:SIMPLE(filecheck=CHECK): 
```

PR #6301, opened 2025-02-06
Allow tuples to work with initializer list.

### LINKED ISSUE #6159: Coercing Initializer List to `Tuple` leads to invalid SPIRV.
When compiling my code to SPIRV I get this warning. The codebase is quite large, so it is hard to find the code that causes this. Nevertheless the code seems to work just fine so maybe it is a false positive validation layer warning? In case you need more info I could attach the spv.

```
VUID-VkShaderModuleCreateInfo-pCode-08737(ERROR / SPEC): msgNum: -1520283006 - Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
OpVariable 8458: expected AliasedPointer or RestrictPointer for PhysicalStorageBuffer pointer.
  %8458 = OpVariable %_ptr_Function_1573 Function
The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.4.304.0/windows/1.4-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```

csyonghe:
Closes #6159.


source/slang/slang-check-conversion.cpp
```diff
@@ -238,6 +238,21 @@ bool SemanticsVisitor::isCStyleType(Type* type, HashSet<Type*>& isVisit)
         return cacheResult(true);
 
 
+    // A tuple type is C-style if all of its members are C-style.
```
kaizhangNV, line 241: Looks like we need to update proposal a little bit?

kaizhangNV (approved):

source/slang/slang-check-conversion.cpp
csyonghe: Updated.

Issue #6250, opened 2025-02-03
Enable DX12 coopvec tests on CI


jkwak-work:
DXC repo that supports coopvec is released publicly.
https://github.com/NVIDIA-RTX/DirectXShaderCompiler/tree/CooperativeVector

We should start using dxcompiler.dll from the repo and enable the related tests.

jkwak-work:
When I used the dxcompiler.dll, I observed crash on one of runners, `SLANGWIN10X64-1`.

When I debugged with visual studio, the call stack looked almost same to one of known issues on DXC repo.
https://github.com/microsoft/DirectXShaderCompiler/issues/6916

The suggested solution is to upgrade MSVC runtime binaries of v143.
And when I upgraded it, it started working.

But the same solution didn't work for another runner, `SlangWin10-2`.
Since the repro is 100% consistent, I will be able to collect more debugging information if I debug it with Visual Studio.

jkwak-work:
To reproduce the issue, I used the following steps,

1. Login to the build machine, "SlangWin10X64-1", with RDP.
2. Clone "https://github.com/shader-slang/slang.git"
3. Replace external/slang-binaries/bin/windows-x64/dxcompiler.dll with one that support coopvec
4. Delete external/slang-binaries/bin/windows-x64/dxil.dll
5. Build as usual and run the test.
6. Observe a few tests failing.

```
cmake.exe --preset vs2022
cmake.exe --build build --preset release
build/Release/bin/slang-test.exe -use-test-server -server-count 8
.............
.............

===
99% of tests passed (4012/4021), 830 tests ignored
===

failing tests:
---
gfx-unit-test-tool/RayTracingTestAD3D12.internal
gfx-unit-test-tool/mutableRootShaderObjectD3D12.internal
gfx-unit-test-tool/shaderCacheSourceStringD3D12.internal
gfx-unit-test-tool/shaderCacheEvictionD3D12.internal
gfx-unit-test-tool/RayTracingTestBD3D12.internal
gfx-unit-test-tool/shaderCacheSpecializationD3D12.internal
gfx-unit-test-tool/computeSmokeD3D12.internal
gfx-unit-test-tool/mutableShaderObjectD3D12.internal
gfx-unit-test-tool/precompiledTargetModule2Vulkan.internal
---
```

jkwak-work:
I realized that there is a build configuration called "MinSizeRel" when compiling dxcompiler.dll
[...]

PR #6302, opened 2025-02-06
Use coopvec supporting dxcompiler dll

jkwak-work:
This commit uses dxcompiler.dll and dxil.dll that is built from:
https://github.com/NVIDIA-RTX/DirectXShaderCompiler/tree/CooperativeVector

Related to https://github.com/shader-slang/slang/issues/6250


jkwak-work:
Falcor is also failing. I need to investigate

PR #6281, opened 2025-02-04
Use two-stage parsing to disambiguate generic app and comparison.

csyonghe:
Closes #6130.
Closes #6285.
Closes #3382.

PR #6303, opened 2025-02-06
Uncomment a previously failing test case

aleino-nv:
This case now parses after the following PR was merged. https://github.com/shader-slang/slang/pull/6281

### DIFF
--- a/tests/wgsl/associativity.slang
+++ b/tests/wgsl/associativity.slang
@@ -745,8 +745,7 @@ void computeMain()
     nb = input[463];
     nc = input[464];
     output[208] = uint(na >> nb < nc);
-    // error 39999: expected a generic when using '<...>' (found: 'uint')
-    // output[209] = uint(na < nb >> nc);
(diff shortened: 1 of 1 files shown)


tests/wgsl/associativity.slang
```diff
@@ -745,8 +745,7 @@ void computeMain()
     nb = input[463];
     nc = input[464];
     output[208] = uint(na >> nb < nc);
-    // error 39999: expected a generic when using '<...>' (found: 'uint')
-    // output[209] = uint(na < nb >> nc);
+    output[209] = uint(na < nb >> nc);
```
csyonghe, line 748: Do you also need to modify the expected result?

aleino-nv: No, it was already `0`. (`na, nb, nc = 3,8,7` and it should be evaluated like like `3 < (8 >> 7) = 3 < 0 = 0`)

csyonghe (approved):

Issue #6307, opened 2025-02-06
Logical operator short-circuit errors after upgrading to dxc 1.8 and above


jkwak-work:
**Problem**
The default behavior of the logical operator short-circuiting is changed from DXC 1.8.
When we upgrade to DXC 1.9 for coopvec support, we get compilation errors for shaders that worked before.

**Repro steps**
The following slang-test can reproduce the issue with dxc 1.8 or above.
```
//TEST(compute):COMPARE_COMPUTE: -shaderobj -output-using-type

int test()
{
    float3 p = {};
    float3 minPoint = {};
    float3 maxPoint = {};
    if (all(p >= minPoint && p <= maxPoint))
        return 0;
    return 1;
}

//TEST_INPUT:ubuffer(data=[0 0 0 0], stride=4):out,name=outputBuffer
RWStructuredBuffer<float> outputBuffer;

[Shader("compute")]
[NumThreads(4, 1, 1)]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
        int tid = dispatchThreadID.x;
        outputBuffer[tid] = test();
}
```
The error message looks something like this, as an example.
```
WGPU error: Error while parsing WGSL: :6:12 error: no matching overload for 'operator && (vec3<bool>, vec3<bool>)'
1 candidate operator:
 • 'operator && (bool  ✗ , bool  ✗ ) -> bool'
    if(all((p_0 >= p_0) && (p_0 <= p_0)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 - While validating [ShaderModuleDescriptor]
 - While calling [Device].CreateShaderModule([ShaderModuleDescriptor]).
```

You can also reproduce the issue with Falcor unit test,
```
FalcorTest.exe -f AABB -d d3d12
```
Observe the following error message, which can be seen only when a debugger is attached:
```
(Info) Created GPU device 'NVIDIA GeForce RTX 3090' using 'Direct3D 12' API (SM6.6).
(Error) GFX Error: dxc 1.9: D:/sbf/git/slang/falcor/Falcor/Source/Falcor/Utils/Math/AABB.slang(108): error :  operands for short-circuiting logical binary operator must be scalar, for non-scalar types use 'and'
dxc 1.9: note :     return (AABB_valid_0(this_6)) && (all((p_2 >= (this_6.minPoint_0)) && (p_2 <= (this_6.maxPoint_0))));
[...]

PR #6310, opened 2025-02-07
Use and() and or() functions for logical-AND and OR

jkwak-work:
With this commit, Slang will emit function calls to `and()` and `or()` for the logical-AND and logical-OR when the operands are non-scalar and the target profile is SM6.0 and above. This is required change from SM6.0.

For WGSL, there is no operator overloadings of `&&` and `||` when the operands are non-scalar. Unlike HLSL, WGSL also don't have `and()` nor `or()`. Alternatively, we can use `select()`.

Closes https://github.com/shader-slang/slang/issues/6307


jkwak-work:
It looks like my change is failing on a few Falcor image tests.
I will take a look.
internal/renderpasses/test_MegakernelPathTracerGBuffer_d3d12

source/slang/slang-emit-wgsl.cpp
```diff
@@ -1312,6 +1312,37 @@ bool WGSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+            // WGSL doesn't have operator overloadings for `&&` and `||` when the operands are
+            // non-scalar. Unlike HLSL, WGSL doesn't have `and()` and `or()`.
+            if (as<IRBasicType>(inst->getDataType()))
+                return false;
+
+            if (inst->getOp() == kIROp_And)
+            {
+                m_writer->emit("select(");
+                emitType(inst->getDataType());
+                m_writer->emit("(false), ");
```
csyonghe, line 1327: (false) should come after operand0?

jkwak-work:
The failure on Falcor is due to the outdated DX validator.
The latest validator doesn't have any problem with it.

With my change, the following code causes the validation to fail:
```
float3 outColor_1 = outColor_0 + select(vector<bool,3>(and((int3)gData_0.params_0.clampSamples_0, int3((path_2.L_0) > (gData_0.params_0.clampThreshold_0)))), (float3)gData_0.params_0.clampThreshold_0, path_2.L_0);
```

Before my change, it was:
```
float3 outColor_1 = outColor_0 + (vector<bool,3>(((int3)gData_0.params_0.clampSamples_0) && int3((path_2.L_0) > (gData_0.params_0.clampThreshold_0))) ? (float3)gData_0.params_0.clampThreshold_0 : path_2.L_0);
```

Note that the following passes the validation, so the problem is on the type given to `and()` is not bool type.
```
float3 outColor_1 = outColor_0 + select(vector<bool,3>(and((bool3)(int3)gData_0.params_0.clampSamples_0, (bool3)int3((path_2.L_0) > (gData_0.params_0.clampThreshold_0)))), (float3)gData_0.params_0.clampThreshold_0, path_2.L_0);
```

csyonghe (dismissed):

source/slang/slang-emit-wgsl.cpp
```diff
@@ -1312,6 +1312,37 @@ bool WGSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+    case kIROp_And:
+    case kIROp_Or:
+        {
+            // WGSL doesn't have operator overloadings for `&&` and `||` when the operands are
+            // non-scalar. Unlike HLSL, WGSL doesn't have `and()` and `or()`.
+            if (as<IRBasicType>(inst->getDataType()))
+                return false;
+
+            if (inst->getOp() == kIROp_And)
+            {
```
csyonghe, line 1324: The signature of `select` in wgsl is `select(falseVal, trueVal, condition)`, so you need to change the order of operands accordingly.

source/slang/slang-emit-wgsl.cpp
```diff
@@ -1317,28 +1317,31 @@ bool WGSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+            // The function signature for `select` in WGSL is different from others:
+            // @const @must_use fn select(f: T, t: T, cond: bool) -> T
             if (inst->getOp() == kIROp_And)
             {
-                m_writer->emit("select(");
-                emitType(inst->getDataType());
-                m_writer->emit("(false), ");
-                emitOperand(inst->getOperand(0), getInfo(EmitOp::General));
-                m_writer->emit(", ");
+                m_writer->emit(" select(vec");
```
csyonghe, line 1328: Why is there a space before “select”?

jkwak-work: There isn't specific reason.
I probably removed and rewrote the line in a little different way.
I will remove it

jkwak-work:
I am fixing the Metal case.

jkwak-work:
Yeah¡ CI all green

PR #6311, opened 2025-02-07
Support stage_switch.

### LINKED ISSUE #3354: Support GLSL builtin variables that maps to different semantics depending on stage.
Some of the GLSL builtin variables maps to different HLSL semantics depending on the current stage. For those builtins, we can define them as global properties where the getter/setter routes to internal declarations based on stage.

For example, 
```
uint __builtinControlPointID : sv_outputcontrolpointid;
uint __builtinGSInstanceID : sv_gsinstanceid;

property uint gl_InvocationID
{
     get
     {
         __target_switch
        {
        case geometry:
               return __builtinGSInstanceID;
        case hull:
              return __builtinGSInstanceID;
        }
     }
}
```

This requires us to expand our `target_switch` mechanism to support stage-based specialization. Currently, there is a dedicated specialization pass to remove all `__target_switch` in the IR by selecting the path suitable for the current compilation target. In the case of  a compilation of mixed entry points, the stage of each invocation might be different, depending on where the call is initiated from.

We need to expand this pass to:
1. Identify all function bodies that has a stage dependent `__target_switch`, and mark such functions as `StageSpecializable`.
2. Propagate `StageSpecializable`  tag on the call graph, such that a function is stage specializable if it transitively calls a stage specializable function.
3. Traverse the call graph from entry points, and specialize `StageSpecializable` callees with the stage of the entry point.
4. Fully remove `__target_switch` by combining the stage the function is specialized to and other capability flags of the current compilation request.

### LINKED ISSUE #6007: gl_PrimitiveID in GLSL fragment shader causes SPIR-V validation error
`gl_PrimitiveID` should be available in fragment shaders as well. Currently, if you try to use it, a SPIR-V validation issue occurs:

```
error: line 8: Operand 3 of Decorate requires one of these capabilities: Geometry Tessellation RayTracingKHR MeshShadingNV MeshShadingEXT RayTracingNV 
  OpDecorate %gl_PrimitiveID BuiltIn PrimitiveId
```

This is because `OpCapability Geometry` should be required if `gl_PrimitiveID` is used in a fragment shader. 

**Analysis:**

This case would work correctly if `glsl.meta.slang` had just `public in int gl_PrimitiveID : SV_PrimitiveID;`, because there's already logic in `slang-emit-spirv.cpp` to handle `SV_PrimitiveID` correctly. Unfortunately, `gl_PrimitiveID` is also used in the ray tracing shaders, and to handle that use case, `glsl.meta.slang` actually has this instead:

```hlsl
public property int gl_PrimitiveID 
{
    [require(cuda_glsl_hlsl_spirv, raytracing_anyhit_closesthit_intersection)]
    get 
    {
        setupExtForRayTracingBuiltIn();
        return PrimitiveIndex();
    }
}
```

This `PrimitiveIndex()` generates `PrimitiveId` using `spirv_asm`. When done this way, the SPIR-V emitter does not add the required capability in a fragment shader, but behaves correctly in the ray tracing shaders as they don't need additional capabilities for this builtin.

It would probably not be too hard to just check through the `SPIRVAsmOperandBuiltinVar`s and require the Geometry capability if `PrimitiveId` is encountered, but that feels "hacky" to me and doesn't fix the GLSL emitter unnecessarily requiring `GL_EXT_ray_tracing` because of `setupExtForRayTracingBuiltIn()`. I'm not yet super familiar with the intricacies here; is there a way to have this `gl_PrimitiveID` property behave differently depending on what shader stage is accessing it?

### LINKED ISSUE #6270: Allow __target_switch to use the shader stages
We should implement `__target_switch` to use the shader stages such as `compute`, `vertex`, `fragment` and so on.

For example, the code can be usefull when compute or vertex shader cannot call `discard`.
```
__target_switch
{
    case fragment: discard(); break;
    default: break;
}
```

csyonghe:
Closes #6007.
Closes #6270.
Fixes #3354.

### DIFF
--- a/docs/proposals/020-stage-switch.md
+++ b/docs/proposals/020-stage-switch.md
@@ -1,0 +1,93 @@
+# SP#020: `stage_switch`
+
+## Status
+
+Author: Yong He
+
+Status: In Experiment
+
+Implementation: [PR]()
+
+Reviewed by:
+
+## Background
+
+We need to provide a mechanism for authoring stage-specific code that works with the capability system. For example, the user may want to define a function `ddx_or_zero(v)` that returns `ddx(v)` when called from a fragment shader, and return `0` when called from other shader stages. Without a mechanism for writing stage-specific code, there is no way to define a valid function that can be used from both a fragment shader and a compute shader in a single compilation.
+
+The user can workaround this problem with the preprocessor:
+
+```
+float ddx_or_zero(float v)
+{
+#ifdef FRAGMENT_SHADER
+    return ddx(v);
+#else
+    return 0.0;
+#endif
+}
+
+[shader("compute")]
+[numthread(1,1,1)]
+void computeMain() { ddx_or_zero(...); }
+
+[shader("fragment")]
+float4 fragMain() { ddx_or_zero(...); }
+```
+
+However, this require the application to compile the source file twice with different pre-defined macros. It is impossible to use a single compilation to generate one SPIRV module that contains both the entrypoints.
+
+## Proposed Approach
+
+We propose to add a new construct, `__stage_switch` that works like `__target_switch` but switches on stages. With `__stage_switch` the above code can be written as:
+
+```
+float ddx_or_zero(float v)
+{
+    __stage_switch
+    {
+    case fragment:
+        return ddx(v);
+    default:
+        return 0.0;
+    }
+}
+
+[shader("compute")]
+[numthread(1,1,1)]
+void computeMain()
+{
+    ddx_or_zero(...); // returns 0.0
+}
+
+[shader("fragment")]
+float4 fragMain()
+{
+    ddx_or_zero(...); // returns ddx(...)
+}
+```
+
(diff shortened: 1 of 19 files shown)


jkwak-work (approved):
Looks good to me.

It is an exciting new feature.

docs/proposals/020-stage-switch.md
```diff
@@ -0,0 +1,93 @@
...
+[shader("fragment")]
+float4 fragMain()
+{
+    ddx_or_zero(...); // returns ddx(...)
+}
+```
+
+With `__stage_switch`, the two entrypoints can be compiled into a single SPIRV in one go, without requiring setting up any preprocessor macros.
+
+Unlike `switch`, there is no fallthrough between cases in a `__stage_switch`. All cases will implicitly end with a `break` if it is not written by the user. However, one special type of fallthrough is supported, that is when multiple `cases` are defined next to each other with nothing else in between, for example:
```
pdjonov, line 71: Nit: I suppose this is aligned with `__target_switch`, but "things that calls themselves `switch` don't always behave like `switch`" feels like more time spent remembering and teaching the exception than a mandatory `break` would cost in keystrokes.

PR #6324, opened 2025-02-09
Add WGSL to list of formats supporting multiple entry points

### LINKED ISSUE #6322: WGSL: Support multiple entry points in output file
Follow up from https://github.com/shader-slang/slang/issues/3974#issuecomment-2645927641

It would be great if like spirv, slang could support outputting multiple entry points when targeting wgsl. Looking at the [example](https://github.com/shader-slang/slang/tree/master/examples/wgpu-html5) If I do: 
```sh
slangc -target wgsl -o shader.wgsl .\shader.slang
```
I'm getting: 
```
error 70: the output path 'shader.wgsl' is not associated with any entry point; a '-o' option for
a compiled kernel must follow the '-entry' option for its corresponding entry point
```

And I'm not seeing any additional args I can pass when looking at the [guide](https://shader-slang.org/slang/user-guide/wgsl-target-specific.html). I also tried setting mutliple `-entry` with `-stage` arguments, and I wasn't able to get that working either.   

Thanks for the great project! Very much looking forward to using slang for WebGPU.

davidar:
The WGSL code generator already supports this, it just wasn't possible to use it with `slangc`, e.g.:

```sh
cd examples/wgpu-html5
slangc -target wgsl -entry vertexMain -stage vertex -entry fragmentMain -stage fragment -o shader.wgsl shader.slang
```

Fixes #6322

### DIFF
--- a/source/slang/slang-options.cpp
+++ b/source/slang/slang-options.cpp
@@ -3509,6 +3509,7 @@ SlangResult OptionsParser::_parse(int argc, char const* const* argv)
                     case CodeGenTarget::MetalLib:
                     case CodeGenTarget::MetalLibAssembly:
                     case CodeGenTarget::Metal:
+                    case CodeGenTarget::WGSL:
                         rawOutput.isWholeProgram = true;
                         break;
                     case CodeGenTarget::SPIRV:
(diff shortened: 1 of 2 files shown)


csyonghe:
Can we add a test to verify multi entrypoint compile is working for wgsl?

davidar:
I think that test works, but honestly I'm not particularly familiar with slang-test

csyonghe (approved):

csyonghe:
@davidar yes, the test looks great.

PR #6325, opened 2025-02-09
Make -output-includes look like cl /showIncludes

mikejsavage:
This change makes it easier to integrate slang into C++ build systems (e.g. ninja) that know how to rebuild files when their dependencies change.

The code is obviously terrible and punches through abstractions because I don't know the codebase well enough to do a good job of this, but this works as a proof of concept.

Demo:

```
> cat compute_shader.slang
#include "blah.slang"
[rest of the shader trimmed]
> cat build.ninja
rule slangc
    command = slangc $in -I. -output-includes -target metal -stage compute -o $out
    deps = msvc
build compute_shader.metal: slangc compute_shader.slang
> ninja
[1/1] slangc compute_shader.slang -I. -output-includes -target metal -stage compute -o compute_shader.metal
> ninja
ninja: no work to do.
> touch blah.slang
> ninja -f shader.ninja
[1/1] slangc compute_shader.slang -I. -output-includes -target metal -stage compute -o compute_shader.metal
> ninja -f shader.ninja
[1/1] slangc compute_shader.slang -I. -output-includes -target metal -stage compute -o compute_shader.metal
```

You can see ninja rebuilds compute_shader.slang when I touch blah.slang, which is what this patch enables. (It keeps rebuilding it because slangc doesn't update the output file's modtime, but that should be fixed separately)

### DIFF
--- a/source/compiler-core/slang-diagnostic-sink.cpp
+++ b/source/compiler-core/slang-diagnostic-sink.cpp
@@ -150,7 +150,7 @@ static void formatDiagnostic(
     DiagnosticSink::Flags flags,
     StringBuilder& outBuilder)
 {
-    if (flags & DiagnosticSink::Flag::HumaneLoc)
+    if (diagnostic.severity != Severity::OutputIncludes && (flags & DiagnosticSink::Flag::HumaneLoc))
     {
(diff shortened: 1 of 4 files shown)


mikejsavage:
Ah I see -depfile does this already, sorry!

PR #6338, opened 2025-02-11
Fix crash when handling uninitialized resource type

### LINKED ISSUE #6328: Default member initialization crashes compiler
Default initialized RW/Textures/Resources seem to be crashing compiler despite a condition is set to false, this is started happening after I bumped slang version from `v2024.17.4` to `v2025.5`

```cpp
struct Foo {
    bool sample_bar = false;
    Texture2D<float4> bar = {};
};

float4 sample_foo(in Foo foo) {
    if (foo.sample_bar) {
        // Compiler crashes here despite `sample_bar` is set to false
        return bar.Sample(...);
    }

    return float4(0.0);
}

float4 main() {
    Foo foo;
    return sample_foo(foo);
}
```

Repro: https://gist.github.com/exdal/2d40533d7a8df813a6996771b1f9f95d
Command line:
```
./slangc -capability spirv_1_4 -target spirv test.slang -o ./test.bin
test.slang(59): warning 41016: use of uninitialized variable 'foo'
    const let result = process(foo);
                              ^
[1]    28517 segmentation fault (core dumped)  ./slangc -capability spirv_1_4 -target spirv test.slang -o ./test.bin
```

Version: v2025.5 (`57b09a8986668626c37055e431fa0ac6449d7214`)
Target profile: `spirv_1_4`

Link to original Discord post: https://discord.com/channels/1303735196696445038/1337851943942619298/1337851943942619298

kaizhangNV:
close #6328.

When declare a var with struct type, if the struct has a resource type field and it doesn't provide explicit constructor, because slang won't implicit construct such variable at declare site if user doesn't explicitly call the initializer list, we should report the resource type field uninitialized as an error to prevent future possible crash when legalize any use of such variable.

### DIFF
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,11 +2551,7 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
-DIAGNOSTIC(
-    56003,
-    Error,
-    useOfUninitializedResouceType,
-    "use of uninitialized resource type '$0'.")
+DIAGNOSTIC(56003, Error, useOfUninitializedResourceType, "use of uninitialized resource type '$0'.")
 
 // Metal
 DIAGNOSTIC(
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,7 +2551,7 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
-DIAGNOSTIC(56003, Error, useOfUninitializedResourceType, "use of uninitialized resource type '$0'.")
+DIAGNOSTIC(56003, Error, useOfUninitializedResourceType, "use of uninitialized opaque type '$0'.")
 
 // Metal
 DIAGNOSTIC(
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,6 +2551,12 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
+DIAGNOSTIC(
+    56003,
+    Error,
+    useOfUninitializedResouceType,
+    "use of uninitialized resource type '$0'.")
+
 // Metal
 DIAGNOSTIC(
     56100,
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,7 +2551,7 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
-DIAGNOSTIC(56003, Error, useOfUninitializedResourceType, "use of uninitialized opaque type '$0'.")
+DIAGNOSTIC(56003, Error, useOfUninitializedOpaqueType, "use of uninitialized opaque type '$0'.")
 
 // Metal
 DIAGNOSTIC(
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,7 +2551,7 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
-DIAGNOSTIC(56003, Error, useOfUninitializedOpaqueHandle, "use of uninitialized opaque handle '$0'.")
+DIAGNOSTIC(56003, Fatal, useOfUninitializedOpaqueHandle, "use of uninitialized opaque handle '$0'.")
 
 // Metal
 DIAGNOSTIC(
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2551,7 +2551,7 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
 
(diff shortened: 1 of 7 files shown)


cheneym2:
There is a typo in the message enum "Resouce":

useOfUninitializedResouceType

source/slang/slang-diagnostic-defs.h
```diff
@@ -2551,6 +2551,8 @@ DIAGNOSTIC(
     attemptToQuerySizeOfUnsizedArray,
     "cannot obtain the size of an unsized array.")
```
csyonghe, line 2553: Maybe reword this to “use of uninitialized opaque handle” to be more accurate.

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -2046,6 +2051,25 @@ static LegalVal coerceToLegalType(IRTypeLegalizationContext* context, LegalType
     }
 }
 
+static LegalVal legalizeUndefined(IRTypeLegalizationContext* context, IRInst* inst)
+{
+    if (auto structType = as<IRStructType>(inst->getFullType()))
+    {
+        for (auto field : structType->getFields())
+        {
+            if (isResourceType(field->getFieldType()))
```
csyonghe, line 2060: I think you want to define a isOpaqueType() functions that returns true for handles and structs that contains opaque type fields. This logic won’t work if there are two levels of nesting.

source/slang/slang-diagnostic-defs.h
```diff
@@ -2551,6 +2551,8 @@ DIAGNOSTIC(
...
 
+DIAGNOSTIC(56003, Error, useOfUninitializedResourceType, "use of uninitialized opaque type '$0'.")
```
csyonghe, line 2554: Need to update the diagnostic code name as well.

source/slang/slang-legalize-types.cpp
```diff
@@ -198,6 +198,23 @@ bool isResourceType(IRType* type)
     return false;
 }
 
+IRType* isOpaqueType(IRType* type)
+{
+    if (isResourceType(type))
+        return type;
+
+    if (auto structType = as<IRStructType>(type))
```
csyonghe, line 206: Whenever you are recursing on a struct type, also consider other composite types that may apply, including:

- array type base (sized and unsized)
- tuple type

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -2046,6 +2051,19 @@ static LegalVal coerceToLegalType(IRTypeLegalizationContext* context, LegalType
     }
 }
 
+static LegalVal legalizeUndefined(IRTypeLegalizationContext* context, IRInst* inst)
+{
+    if (auto opaqueType = isOpaqueType(inst->getFullType()))
+    {
+        context->m_sink->diagnose(
+            inst->getFullType(),
```
csyonghe, line 2059: Instead of using the type inst’s source loc for the diagnostic, traverse the uses to find the first user of the undef inst, and report using that user’s source loc. if a user doesn’t have a valid source loc, consider the next user. If none of the users has source loc, fall back to use the source loc of the type inst.

you can abstract this into a method called SourceLoc findBestSourceLocFromUses(IRInst*)

kaizhangNV: I'm actually thinking another way around, the undefined inst doesn't have sourceLoc field, we could set it during IR lowering right?

csyonghe: That would be tricky because this is introduced during SSA pass as a stub value, so it doesn’t have a well defined source loc.

source/slang/slang-legalize-types.cpp
kaizhangNV: should include pointer type?

kaizhangNV: Also, how to get element type of a `IRTupleType`,  I didn't find there is any member functions to get that.

csyonghe: Just call `getOperand` to get element types.

csyonghe: it is fine to leave pointer behind, since it should never appear in this context.
if we add pointer check, then we need to be careful not to run into infinite recursion because you can have a pointer that point to the struct type itself.

source/slang/slang-diagnostic-defs.h
csyonghe: a "type" cannot be "initialized", so here it should be "opaque handle", not "opaque type".

source/slang/slang-legalize-types.cpp
kaizhangNV: OK, I can leave pointer there.

I'm wondering if we already legalize a Tuple type to struct type in previous legalization pass? So we actually don't really need to handle Tuple type here?

csyonghe: Yes, we do already legalize a tuple to struct when we reach legalizeResourceTypes.

However if this function is expected to be a general utility that can be called everywhere, it is better to make sure tuple types are handled here.

csyonghe (dismissed):

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -2046,6 +2051,24 @@ static LegalVal coerceToLegalType(IRTypeLegalizationContext* context, LegalType
...
+    if (isOpaqueType(inst->getFullType(), opaqueTypes))
+    {
+        auto opaqueType = opaqueTypes[0];
+        auto containerType = opaqueTypes.getCount() > 1 ? opaqueTypes[1] : opaqueType;
+        SourceLoc loc = findBestSourceLocFromUses(inst);
+
+        if (!loc.isValid())
+            loc = getDiagnosticPos(containerType);
+
+        context->m_sink->diagnose(loc, Diagnostics::useOfUninitializedOpaqueHandle, opaqueType);
```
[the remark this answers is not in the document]
kaizhangNV: Oh, I see.

Issue #4760, opened 2024-07-30
Move slang-test and render-test to use the new compilation API


csyonghe:
All C functions + the ICompileRequest interface is the old API.
We should move to use COM interfaces directly, and not use the ICompileRequest interface.

jkwak-work:
I am very interested in working on this.
It will be a good chance for me to learn more about our reflection APIs, which I am not very familiar at this point.

csyonghe:
I think this is more about the compilation API, the reflection logic can most likely remain unchanged. But this is a good opportunity to get familiar with the compilation API.

aleino-nv:
> I am very interested in working on this. It will be a good chance for me to learn more about our reflection APIs, which I am not very familiar at this point.

@jkwak-work Would it be ok if I took this task, or are you planning to get to it soon?

aleino-nv:
@csyonghe What's meant by "new compilation API"? COM-based rather than C API?

aleino-nv:
@csyonghe Does the non-deprecated API provide anything like `ICompileRequest::compile`?

There is one usage of `ICompileRequest` in `slang-test`, and that's this line in `slangc-tool.cpp`:

https://github.com/shader-slang/slang/blob/0fee8c16e25521d13b50427ac6bf87d7f3419b9c/tools/slang-test/slangc-tool.cpp#L58

csyonghe:
That use is fine, because `slangc` is currently still on legacy API, and we just have to keep it that way.

The main body of work is to get render-test to use the new API, which amounts to changing `ShaderCompilerUtil::_compileProgramImpl` function to not use any `sp*` APIs or the `ICompileRequest` API.

aleino-nv:
Ok, I worked on render-test today. I've almost got a sketch of a patch ready.

aleino-nv:
I had to debug quite a bit today, because some reflection data seems to have broken with my patch.
I created a simpler clean-up patch which is now in review, as a stepping stone.

I still need at least a day on this task to actually get rid of the compile request stuff from render-test.

aleino-nv:
I created https://github.com/shader-slang/slang/pull/6440 today.

However it runs into a seemingly pre-existing issue where there's a crash in loadModuleFromSourceString if you call it again from a second session. I will file a bug for this tomorrow and try to fix that issue first.

aleino-nv:
[...]

PR #6343, opened 2025-02-12
Migrate slang-test away from deprecated Slang API

aleino-nv:
This helps to address #4760.


csyonghe:
Please note that this is only a small part of https://github.com/shader-slang/slang/issues/4760.
We still need to update render-test, specifically the `ShaderCompilerUtil` class to use the COM API.

PR #6350, opened 2025-02-13
explicitly add slang-without-embedded-core-module as a dependency to …

expipiplus1:
…the slang-bootstrap call


csyonghe:
The fix is already in tot.

PR #6351, opened 2025-02-13
Add support for GLSL interface blocks

cheneym2:
This adds support for GLSL style syntax for shader inputs, e.g.:

in VertexData
{
    vec2 texcoord_0;
    vec2 texcoord_1;
} vd;


csyonghe:
In slang-ir-translate-global-var.cpp:97, instead of creating a new field type layout for the global var, there should already be a layout created on `input` (stored in a `IRLayoutDecoration`), we should just use that as `fieldTypeLayout` instead of building it.

cheneym2:
> In slang-ir-translate-global-var.cpp:97, instead of creating a new field type layout for the global var, there should already be a layout created on `input` (stored in a `IRLayoutDecoration`), we should just use that as `fieldTypeLayout` instead of building it.

Okay, thanks, that seems to be working, though I don't know what to do about the call to "addResourceUsage" on the fieldTypeLayout builder which would go unused in the case that I reuse a layout from IRLayoutDecoration.

csyonghe:
`varLayoutBuilder` takes a TypeLayout to initialize, so you just delete the existing `fieldTypeLayout` completely, and pass the type layout you get from `input`'s layout decoration to initailize `varLayoutBuilder`.

cheneym2:
> `varLayoutBuilder` takes a TypeLayout to initialize, so you just delete the existing `fieldTypeLayout` completely, and pass the type layout you get from `input`'s layout decoration to initailize `varLayoutBuilder`.

That's what I think I'm doing, and it mostly works.  I'll just upload a commit to explain.

source/slang/slang-ir-translate-glsl-global-var.cpp
```diff
@@ -94,9 +111,12 @@ struct GlobalVarTranslationContext
                 }
                 else
                 {
-                    fieldTypeLayout.addResourceUsage(
-                        LayoutResourceKind::VaryingInput,
-                        LayoutSize(1));
+                    if (!hasExistingLayout)
+                    {
+                        fieldTypeLayoutBuilder.addResourceUsage(
```
cheneym2, line 116: The doubt I was having was about this call to addResourceUsage.  I think it doesn't make sense if I'm pulling fieldTypeLayout from IRLayoutDecoration.

cheneym2:
glsl-intrinsic/fragment-processing/fragment-processing.slang is failing due to the missing shadowed globals replacement optimization.

cheneym2:
With the change to use tryReplaceUsesofStageInput for the case of nested input, compilation of the test case outputs this:

```
#version 450
layout(row_major) uniform;
layout(row_major) buffer;

#line 31 0
vec4 _S1;


#line 31
layout(location = 0)
out vec4 entryPointParam_fragmentMain_out1_0;


#line 1974 1
layout(location = 0)
in vec2 vd_texcoord_0_0;


#line 1974
layout(location = 1)
in vec2 vd_texcoord_1_0;


#line 1974
layout(location = 2)
in vec4 vd_inner_texcoord_2_0;


#line 19 0
struct innerData_0
{
    vec4 texcoord_2_0;
};


#line 33
void main()
{

#line 33
    innerData_0 _S2 = { vd_inner_texcoord_2_0 };

#line 33
    _S3.texcoord_0_0 = vd_texcoord_0_0;

#line 33
    _S3.texcoord_1_0 = vd_texcoord_1_0;

#line 33
    _S3.inner_0 = _S2;

    vec4 _S4 = vec4(vd_texcoord_0_0, vd_texcoord_1_0.x, vd_inner_texcoord_2_0.y);

#line 35
    _S1 = _S4;

#line 35
    entryPointParam_fragmentMain_out1_0 = _S4;

#line 35
    return;
}
```

csyonghe:
I guess the left over global variable _S3 can be removed if we add an optimization pass that removes non-output global variables that has only writes but no reads.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -3290,45 +3306,54 @@ void legalizeEntryPointParameterForGLSL(
...
+                }
+                SLANG_ASSERT(realGlobalVar);
+
+                // Remove all stores into the global var introduced during
+                // the initial glsl global var translation pass since we are
+                // going to replace the global var with a pointer to the real
+                // input, and it makes no sense to store values into such real
+                // input locations.
+                traverseUses(
+                    globalVar,
```
csyonghe, line 3340: we can also follow uses of any FieldAddr on the globalVar and remove the stores to those field addrs to get rid of the store to _S3 in the test case.

csyonghe (approved):

csyonghe (approved):

PR #6363, opened 2025-02-14
Deduplicate witness table while specializing generic functions

jkwak-work:
This commit fixes an issue that Slang sometimes emits duplicated DiffPair_XX structs when targeting HLSL.

This fix is required for upgrading DXC version from 1.7 to 1.9. With DXC 1.7, it had been fine even with the duplicated structs as long as their member variables are identical. But with DXC 1.9, it causes a compile error.

Closes https://github.com/shader-slang/slang/issues/6364

### DIFF
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -3123,27 +3123,26 @@ IRInst* specializeGenericImpl(
             //
             if (auto witnessTable = as<IRWitnessTable>(ii))
             {
+                // CLone type and concreteType to use as cache-key
                 auto witnessTableType = witnessTable->getFullType();
+                IRType* clonedWitnessTableType =
+                    as<IRType>(cloneInst(&env, builder, witnessTableType));
 
-                // We need to apply specialization from `env` to `concreteType`.
                 auto concreteType = witnessTable->getConcreteType();
                 IRType* clonedConcreteType = as<IRType>(cloneInst(&env, builder, concreteType));
 
                 SpecializationContext::WitnessTableKey cacheKey(
-                    witnessTableType,
+                    clonedWitnessTableType,
                     clonedConcreteType);
 
                 if (!context->mapClonedWitnessTable.tryGetValue(cacheKey, clonedInst))
                 {
-                    // Not found from cache
-                    // Need to create a new one
-                    IRInst* newOperands[] = {clonedConcreteType};
-
+                    // Not found from cache and need to create a new one
                     clonedInst = builder->emitIntrinsicInst(
-                        witnessTableType,
+                        clonedWitnessTableType,
                         kIROp_WitnessTable,
                         1,
-                        newOperands);
+                        (IRInst* const*)&clonedConcreteType);
 
                     clonedInst->sourceLoc = ii->sourceLoc;
 
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -3123,7 +3123,7 @@ IRInst* specializeGenericImpl(
             //
             if (auto witnessTable = as<IRWitnessTable>(ii))
             {
-                // CLone type and concreteType to use as cache-key
+                // Clone type and concreteType to use as cache-key
                 auto witnessTableType = witnessTable->getFullType();
                 IRType* clonedWitnessTableType =
                     as<IRType>(cloneInst(&env, builder, witnessTableType));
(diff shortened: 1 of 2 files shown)


source/slang/slang-ir-autodiff.h
```diff
@@ -510,7 +510,34 @@ struct DifferentialPairTypeBuilder
     };
 
     // Cache from pair types to lowered type.
-    Dictionary<IRInst*, IRInst*> pairTypeCache;
+    //
```
csyonghe, line 513: This won't fully resolve the issue, if there is a `IRDiffPair<IRDiffPair<x, w0>, w1>` and  `IRDiffPair<IRDiffPair<x, w3>, w1>`, we will still generate duplicates because the inner diff pair types are different.

The diff pair type can be used as array element type, and other types as well, causing them  to be different. So the only guaranteed way is to make sure we deduplicate the witness tables.

jkwak-work: I agree that when diff pair type is an element of an array, this implementation wouldn't be able to remove the witness table. I can improve that if it is a really reproducible case.

But the comment sounds more like the source of the problem is on other places where the witness tables are created.
One of the commits on this PR attempted to remove the duplication where the witness table gets created.
https://github.com/shader-slang/slang/pull/6363/commits/f06f0dd5fef745011a37d33398f5e48ebc1cf49f

But it turned out that they weren't really duplicates, because they have differences on their decorations-and-children.

My understanding so far is that `pairTypeCache` here is where we should remove the duplications caused by the witness table. If it doesn't look to be the case, please give me some help to figure out where is the best location to apply the deduplication.

csyonghe: Differences in decoration and children are fine, and witness tables is supposed to be equivalent when their type and concrete types match. We just need to pick one witness table and use it everywhere.

jkwak-work:
The latest change on this PR removes the duplicated witness table while specializing the generics.
This is the most preferred way to handle the issue.

source/slang/slang-ir-specialize.cpp
```diff
@@ -3098,6 +3122,23 @@ IRInst* specializeGenericImpl(
             //
             IRInst* clonedInst = cloneInst(&env, builder, ii);
 
+            // Avoid duplicated witness tables.
```
csyonghe, line 3125: Can we move this test before `cloneInst`? That we can avoid cloning and then removing.

We just detect if `ii` is a witness table, if so, clone its type and operand0, then use that as key to lookup, if nothing found, proceed to call `cloneInst`, otherwise just use the found value as `clonedInst`.

jkwak-work: Yes I can do that

jkwak-work:
I made changes as suggested.

jkwak-work:
From a quick inspection with LLM, it appears that the decorations need to be handled properly.
In fact, they are not decorations but children for indicating how to handle dzero, dadd, dmul and such.

jkwak-work:
It appears the the witness-table-entry as the children are same when the type and concrete type are same.
I left a comment on the source code for an extra information.

source/slang/slang-ir-specialize.cpp
```diff
@@ -54,6 +54,29 @@ struct SpecializationContext
...
+        {
+            return witnessTableType == other.witnessTableType && concreteType == other.concreteType;
+        }
+
+        HashCode getHashCode() const
+        {
+            return combineHash(HashCode(witnessTableType), HashCode(concreteType));
+        }
+    };
+    Dictionary<WitnessTableKey, IRInst*> mapClonedWitnessTable;
```
csyonghe, line 78: This dictionary needs to be populated upon initialization of `SpecializationContext`, otherwise, if a specialization will result in a witness table that already exists in the global scope, we will not be able to dedup.

The current code can only dedup witness tables created during a single specialization pass. Note that the specialization pass is invoked in a loop, so it can run more than once, each time specializing a couple more insts.

csyonghe: I still feel like the cleanest way to solve this is to make IRWitnessTable a `HOISTABLE` inst so it can be deduplicated by the existing global value deduplication mechanism in our IR.

jkwak-work: I will make changes based on the first comment.

For the second comment, I can give another try but when I tried, it was crashing.
I think WitnessTableEntry will have to become HOISTABLE as well in order to make WitnessTable HOISTABLE.
Because WitenessTable gets WitnessTableEntry as children.
I will try to make both of them HOISTABLE and see what happens.

csyonghe: We definitely don’t want witness table entry to be hoistable, since they doesn’t matter for deduplication purpose. Deduplication uses only type and operand as key, and that is exactly what we need here. 

I am not saying turning the witness table inst itself as hoistable is the only required change, we need to deal with the consequences. For example, hoistable insts requires knowing its type and operands upon creation, you can’t create an hoistable inst and then modify its type or operands afterwards.

jkwak-work: I see. I will not touch the witness table entry.

jkwak-work:
Closing this PR because a better version of PR(https://github.com/shader-slang/slang/pull/6417) is merged

PR #6367, opened 2025-02-15
expose value of constant integers in module reflection

ennis:
One of the pain points of working with shaders is needing to keep the definitions of interface types in sync between shader and device code. With the reflection API of slang, it is possible to automate this process by generating a header file for the host language from the reflection data, that contains all interface types used in the shader.
However, one thing missing is the ability to get the value of compile-time constants defined in the shaders. This is useful if you have some constant that an algorithm or a data structure depends on, like a static block or tile size, and you don't want to have to keep them in sync manually between host and shader.

To solve this (at least for integers) this PR adds `VariableReflection::getDefaultValueInt` to get the default value of a variable if it is a compile-time constant integer. The reflection-api example has also been updated to print all the static const variables of a module and their values.

Internally, this uses the `VarDeclBase::val` field that should contain the constant-folded initializer value. However, currently this field is null if, for instance, the initializer is a variable reference (as is the case for THREADGROUP_SIZE_Y in the `compute-simple.slang` of the reflection-api example). It seems to be initialized in `SemanticsDeclHeaderVisitor::checkVarDeclCommon`, which then goes through `SemanticsVisitor::tryConstantFoldExpr`, which does try to follow variable references. However it seems that at this point the reference cannot be resolved. I'd like to make this work before trying to merge this.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -1445,7 +1445,10 @@ IntVal* SemanticsVisitor::_validateCircularVarDefinition(VarDeclBase* varDecl)
     //
     if (!isScalarIntegerType(varDecl->type))
         return nullptr;
-    return tryConstantFoldDeclRef(DeclRef<VarDeclBase>(varDecl), ConstantFoldingKind::LinkTime, nullptr);
+    return tryConstantFoldDeclRef(
+        DeclRef<VarDeclBase>(varDecl),
+        ConstantFoldingKind::LinkTime,
+        nullptr);
 }
 
 void SemanticsDeclModifiersVisitor::visitStructDecl(StructDecl* structDecl)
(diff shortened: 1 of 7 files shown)


ennis:
Following a suggestion on discord I've added code to update VarDecl's constant value at a later stage during DeclBodyVisitor. Since constant folding is already done inside _validateCircularVarDefinition, this commit reuses the constant that is evaluated there.

source/slang/slang-check-decl.cpp
```diff
@@ -2348,7 +2348,14 @@ void SemanticsDeclBodyVisitor::checkVarDeclCommon(VarDeclBase* varDecl)
         // a constant with a circular definition.
         //
         varDecl->setCheckState(DeclCheckState::DefinitionChecked);
-        _validateCircularVarDefinition(varDecl);
+        auto intVal = _validateCircularVarDefinition(varDecl);
```
csyonghe, line 2351: I think you can now move the call into the `if`, because if `varDecl->val` is not null, then there couldn't be any circular references.

csyonghe (approved):
Looks good to me!

source/slang/slang-check-decl.cpp
ennis: I've made the suggested changes

csyonghe (approved):

Issue #6210, opened 2025-01-29
Support shader subgroup/wave intrinsics for Metal


fairywreath:
Metal has support for subgroup intrinsics through "SIMD groups", Slang should support generating them.

expipiplus1:
@fairywreath can this be closed after those two PRs?

fairywreath:
@expipiplus1 I made a new issue #6610 for the last remaining TODO and am closing this one.

PR #6371, opened 2025-02-16
Add WaveGetLane* support for Metal and WGSL

### LINKED ISSUE #6208: Support shader subgroup/wave intrinsics for WGSL

fairywreath:
Closes #6208, related to #6210.

Adds support for `WaveGetLane*` intrinsics for Metal and WGSL through the use of GLSL flavored in/out global vars. 

These intrinsics require built-in values to be passed through the entry point parameters, but the intrinsics themselves are "global" functions. The implementation reuses the already existing code that handles glsl `gl_*` global builtin variables and adds them to the entry point parameter list. `translateGLSLGlobalVar` is renamed because it also handles non-glsl code.


csyonghe (dismissed):
Looks good!

source/slang/slang-emit.cpp
```diff
@@ -100,7 +100,7 @@
 #include "slang-ir-strip-default-construct.h"
 #include "slang-ir-strip-legalization-insts.h"
 #include "slang-ir-synthesize-active-mask.h"
-#include "slang-ir-translate-glsl-global-var.h"
+#include "slang-ir-translate-in-out-global-var.h"
```
csyonghe, line 103: Can name it translate-global-varying-var.

source/slang/slang-emit-c-like.cpp
```diff
@@ -3055,6 +3055,11 @@ void CLikeSourceEmitter::defaultEmitInstExpr(IRInst* inst, const EmitOpInfo& inO
             emitOperand(as<IRGlobalValueRef>(inst)->getOperand(0), getInfo(EmitOp::General));
             break;
         }
+    case kIROp_RequireWGSLExtension:
```
csyonghe, line 3058: Can we just use a single RequireTargetExtension for all targets?

csyonghe (approved):

fairywreath:
Not sure why the Metal glsl variant test fails while the HLSL variant passes - the produced code is similar and they both compile fine using the Metal compiler. I do not have an Apple device to test, hopefully my most recent change fixes it.

fairywreath:
The glsl variant test still failed on Metal and I am not sure why. I loosened up the check requirements for the Metal case and added another test for the glsl syntax that is a copy of the hlsl syntax test(that passes just fine in Metal), so the new test should definitely pass.

csyonghe:
Hmm, if it is a hardware problem, then the HLSL syntax should also fail?

Can you compare the metal shader output from both HLSL and GLSL syntax and see if there are any differences?

fairywreath:
Yea somehow the new GLSL syntax test(that is a 1-1 copy of the passing HLSL syntax test) is failing. Here is a [diff](https://editor.mergely.com/7tnXKLJb) of the outputs of the HLSL and GLSL syntax. They are very similar.

csyonghe:
I noticed that the HLSL case, the compiler is able to fold the access to wave index directly as a use of `_builtinWaveLaneIndex_0` parameter, where in the GLSL case, the parameter is first written into the kernel context and then loaded from the context.

Which should be fine, but I wonder if that's the reason that caused the wrong value.

csyonghe:
We are currently generating:
```
#line 16
    (&kernelContext_0)->_S3 = _builtinWaveLaneCount_0;

#line 16
    (&kernelContext_0)->_S2 = _builtinWaveLaneIndex_0;

#line 16
    (&kernelContext_0)->_S1 = gl_GlobalInvocationID_0;
```
and this is in reverse order of the fields, might worth clean this up and make the assignment in the consistent order of the fields.

fairywreath:
Thanks for the help. I have made changes so that the GLSL and HLSL outputs are exactly the same ([diff](https://editor.mergely.com/uv7xaA7Q)).  Can you run the test again?

fairywreath:
It is surprising that the GLSL syntax variant still fails. Here is a [diff of the hlsl and glsl tests](https://editor.mergely.com/Sht38SwD) and [diff of the Metal output](https://editor.mergely.com/CobPshx1). The produced Metal code is exactly the same.

csyonghe:
One test is using COMPARE_COMPUTE, and the other is using COMPARE_COMPUTE_EX, can we make both of them the same?

csyonghe:
This is needing more debug. Let's disable the glsl test and get this PR merged first. We can leave it as future work.

tests/glsl-intrinsic/shader-subgroup/shader-subgroup-builtin-variables-2.slang
```diff
@@ -1,7 +1,10 @@
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-vk -compute -entry computeMain -allow-glsl -emit-spirv-directly
 //TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-wgpu -compute -entry computeMain -allow-glsl
-//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-metal -compute -entry computeMain -allow-glsl
+
+
+// TODO: There are some issues with the Metal backend when using glsl-style syntax, test is disabled for now.
+//DISABLE_TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=BUF):-metal -compute -entry computeMain -allow-glsl
```
csyonghe, line 7: Instead of disable test here, let’s run the test, but add the test name to expected-failure.txt so we can always use that file to track tests that we still need to work on.

csyonghe (approved):

csyonghe (approved):

fairywreath:
Whoops I put the expected to fail tests in the wrong file list, I have corrected them.

csyonghe (approved):

Issue #4251, opened 2024-05-31
Throw an error when using `[[vk::binding]]` on global shader parameters


pedrotrschneider:
### Discussed in https://github.com/shader-slang/slang/discussions/4232

Basically, when you write something like this:

```HLSL
[[vk::binding(0, 1)]]
uniform float my_uniform;
```

The slang compiler ignores the request for the specific descriptor set and binding group and packs all uniforms into one global uniform at descriptor set 0 binding 0. As we discussed in the Discussion Thread, this is intended behavior, but the compiler should warn the user (or, I'd argue, throw a compilation error) that their request was ignored, and maybe include steps on what they need to do to get their desired outcome.

For me, this worked to force slang to abide by my `[[vk::binding]]` directives, but I don't know if it's the optimal solution:

```HLSL
[[vk::binding(0, 1)]]
ConstantBuffer<float> my_uniform
``` 

I would also suggest updating the example in the [user's guide](https://shader-slang.com/slang/user-guide/conventional-features.html#global-variables-and-shader-parameters), as it does not do what the guide suggests it does.

![image](https://github.com/shader-slang/slang/assets/62211260/29ab1e48-09ca-462b-9ed7-a906b73674a2)

PR #6373, opened 2025-02-17
Add warning for ignored binding attributes on uniforms

mkeshavaNV:
Fix for #4251 

When binding attributes (like [[vk::binding]]) are specified on uniforms that get packed into the default constant buffer, these binding attributes are effectively ignored since the uniform will always be placed at descriptor set 0, binding 0. This can be confusing for users who expect their explicit bindings to take effect.

This change adds a new warning (71) that informs users when their binding attributes on uniforms will be ignored, and suggests declaring the uniform inside a constant buffer to preserve the explicit binding.

The warning helps users understand:
1. Why their binding attribute isn't having the expected effect
2. That the uniform is being packed into the default constant buffer
3. How to fix it by using a constant buffer declaration

Added test case in tests/bugs/binding-attribute-ignored.slang to verify the warning behavior.

### DIFF
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -2055,6 +2055,14 @@ DIAGNOSTIC(
     "shader parameter '$0' has a 'register' specified for D3D, but no '[[vk::binding(...)]]` "
     "specified for Vulkan, nor is `-fvk-$1-shift` used.")
 
+DIAGNOSTIC(
+    39071,
+    Warning,
+    bindingAttributeIgnoredOnUniform,
+    "binding attribute on uniform '$0' will be ignored since it will be packed into the default "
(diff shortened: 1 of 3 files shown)


source/slang/slang-diagnostic-defs.h
```diff
@@ -248,6 +248,14 @@ DIAGNOSTIC(
     "the output path '$0' is not associated with any entry point; a '-o' option for a compiled "
     "kernel must follow the '-entry' option for its corresponding entry point")
 
+DIAGNOSTIC(
```
csyonghe, line 251: This diagnostic should be number 3xxxx around other binding related messages.

mkeshavaNV: @csyonghe - Fixed.

PR #6377, opened 2025-02-18
Implement sparse texture Sample* intrinsics for SPIRV

fairywreath:
Related to #4495.

Adds SPIRV support for sparse texture `Sample*` intrinsics or `Sample*` with status intrinsics. They are implemented through the `SparseResidency` capability.

`Load` and `Gather` intrinsic variants are still not implemented for SPIRV and will be addressed in separate PRs.


csyonghe (dismissed):
Looks good!

csyonghe:
`tests/hlsl-intrinsic/texture/partial-resident-texture.slang` test failed.

fairywreath:
There seems to be a problem with slang-test when running DX12 compute functional tests with sampler2D - I tested this on a Windows machine with an AMD gpu. The test just fails with error code 1 - is this expected?

I have separate the original test into two, one for regular textures and the other for combined texture+sampler resources. Can you run them to confirm that this is also happening in the CI?

csyonghe (approved):

Issue #2890, opened 2023-05-19
assertion failure with zero-length array


expipiplus1:
Fails in `legalizeArrayReturnType`

```
[numthreads(4, 1, 1)]
void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
{
    bar();
}
func bar() -> int[0];
```

natduca:
Seems like q3 is where we're putting language completeness items. Medium b/c compiler crash so inherently more important than quality of life

PR #6379, opened 2025-02-18
Bugfix/gh 2890

mkeshavaNV:


mkeshavaNV:
I see that issue #2890 no longer reproduces on ToT.  Added a test for this. 
Cc: @expipiplus1 for visibility.

expipiplus1 (approved):

expipiplus1:
Thanks!

tests/diagnostics/array-zero-size.slang
```diff
@@ -0,0 +1,13 @@
...
+
+//TEST:SIMPLE:
+
+[numthreads(4, 1, 1)]
+void computeMain(uint3 dispatchThreadID : SV_DispatchThreadID)
+{
+    bar();
+}
+
+func bar() -> int[0];  // expected-error 30025 "array size must be larger than zero."
```
jkwak-work, line 13: I am not sure how this test is expected to work.
When I compiled, I don't see any error message,
```
$ slangc.exe tests/diagnostics/array-zero-size.slang

#pragma pack_matrix(column_major)
#ifdef SLANG_HLSL_ENABLE_NVAPI
#include "nvHLSLExtns.h"
#endif

#ifndef __DXC_VERSION_MAJOR
#pragma warning(disable : 3557)
#endif

void bar_0(out int  _0[int(0)]);

[numthreads(4, 1, 1)]
void computeMain(uint3 dispatchThreadID_0 : SV_DispatchThreadID)
{
    int  _S1[int(0)];
    bar_0(_S1);
    return;
}
```

jkwak-work (commented):
@expipiplus1 , can you give me some explanations of why this is approved?

tests/diagnostics/array-zero-size.slang
mkeshavaNV: @jkwak-work thanks for noticing this. Yeah I agree that the test doesn't really check this error correctly. It just passes always. I will reopen the issue.

mkeshavaNV: @expipiplus1 @jkwak-work - I have created a separate PR for this: https://github.com/shader-slang/slang/pull/6399/

@jkwak-work 
1. I was running a release build, hence never saw the assert, but now I realize that I also missed seeing the compiler error.
2. As for the test, the diagnostic tests expect an "expected" file, which also was missing. I've now added it.
Can you please review PR 6399

PR #6381, opened 2025-02-18
HLSL: Add 'f' suffix to float literals in code generation

### LINKED ISSUE #6078: Issues with HLSL floating point literals
HLSL and dxc traditionally defaults(?) non-suffixed floating literals to 64-bit double, but with https://github.com/llvm/llvm-project/issues/85714 the behavior is updated to default to 32-bit floats. The latest DXC releases however still require the `-HV 202x` flag to enable this, or else it will still default to double.

This slang shader, for example
```
float4 fragmentMain(float2 uv : TEXCOORD) : SV_Target {
    float val = 1.5;
    return float4(1.0, ddx(val), 0.0, 1.0); 
}
```
generates this shader:
```
float4 fragmentMain(float2 uv_0 : TEXCOORD) : SV_TARGET
{
    return float4(1.0, ddx(1.5), 0.0, 1.0);
}
```
which fails to compile and triggers a validation on `dxc` without the `-HV 202x` flag because `1.5` is treated as a double `ddx` only accepts float. This behavior is consistent with all other builtins that accept float but not double:
```
error: validation errors
ss2.hlsl:15:24: error: DXIL intrinsic overload must be valid.
note: at '%1 = call double @dx.op.unary.f64(i32 83, double 1.500000e+00)' in block '#0' of function 'fragmentMain'.
Validation failed.
```
Changing `ddx(1.5)` to `ddx(1.5f)` allows the code to compile through `dxc`.

This issue can be resolved by adding the f suffix to float literals when emitting HLSL code. We can also just ignore this and  enforce the use of the -HV 202x flag, or wait for a newer DXC release that changes the default behavior.

mkeshavaNV:
Fixes #6078

Previously, Slang would emit HLSL float literals without a suffix (e.g.,"1.5"), which caused DXC to interpret them as 64-bit doubles by default unless the -HV 202x flag was used. This could cause validation errors when these literals were used with intrinsics that only accept 32-bit floats (like ddx, ddy).

This change modifies the HLSL emitter to explicitly add 'f' suffix to 32-bit float literals, ensuring they are correctly typed regardless of DXC's version or flags. For example:
- "1.5" becomes "1.5f"
- "(0.0 / 0.0)" becomes "(0.0f / 0.0f)" for NaN
- "float4(1.0, 2.0, 3.0, 4.0)" becomes "float4(1.0f, 2.0f, 3.0f, 4.0f)"

Added a test case to verify the behavior with various float literal scenarios including basic literals, intrinsic calls, and vector construction.


mkeshavaNV:
Here is the output of the hlsl shader:

`#pragma pack_matrix(column_major)
#ifdef SLANG_HLSL_ENABLE_NVAPI
#include "nvHLSLExtns.h"
#endif

#ifndef __DXC_VERSION_MAJOR
#pragma warning(disable : 3557)
#endif


#line 1 "slang-repro.slang"
float4 fragmentMain(float2 uv_0 : TEXCOORD) : SV_TARGET
{
    return float4(1.0f, ddx(1.5f), 0.0f, 1.0f);
}
`

Verified that this works OK with dxc.

Cc: @csyonghe for reference.

source/slang/slang-emit-hlsl.cpp
```diff
@@ -1252,23 +1252,30 @@ void HLSLSourceEmitter::emitSimpleValueImpl(IRInst* inst)
...
             default:
-                break;
+                {
+                    m_writer->emit(constantInst->value.floatVal);
+                    // Add 'f' suffix for 32-bit float literals to ensure DXC treats them as float
+                    if (constantInst->getDataType()->getOp() == kIROp_FloatType)
+                    {
+                        m_writer->emit("f");
+                    }
+                    return;
```
jkwak-work, line 1276: Before your change, it was "break" instead of "return".
Your PR is making a change to the behavior that is not described on the PR description.
Can you explain why this change is needed?

jkwak-work (commented):
Were you able to reproduce the error form DXC?
I am about to upgrade DXC from 1.7 to 1.9 and I like to know if the issue is specific to a certain version of DXC or not.

source/slang/slang-emit-hlsl.cpp
csyonghe: You need `return` here to avoid the fallback call to `Super::emitSimpleValueImpl` since we are overriding the behavior for ordinary float literals here.

csyonghe:
as long as all tests are passing, we are fine.

csyonghe (approved):

jkwak-work:
Please use the following test, instead of one in this PR,
```
//TEST:SIMPLE(filecheck=HLSL):-target hlsl -profile ps_6_6 -entry fragmentMain
//TEST:SIMPLE(filecheck=DXIL):-target dxil -profile ps_6_6 -entry fragmentMain

float4 fragmentMain(float2 uv : TEXCOORD) : SV_Target
{
    //HLSL:, ddx({{.*}}1.5f)
    //DXIL: = call float @dx.op.unary.f32(i32 {{.*}} ; DerivCoarseX(value)
    float val = 1.5;
    return float4(1.0, ddx(val), 0.0, 1.0);
}
```

mkeshavaNV:
> To answer my question, I was able to reproduce it with DXC 1.7.

@jkwak-work - Yes I was able to reproduce the issue with dxc. I used this version of dxc:
```
C:\WINDOWS\system32>dxc --version
dxcompiler.dll: 1.8 - 1.8.0.4775 (d39324e06)
```

> Please use the following test, instead of one in this PR,

Ok, you want to add dxil coverage too, right?

PR #6382, opened 2025-02-18
-dump-intermediates for precompiled targets

cheneym2:
When using slangc with precompiled target modules, dump the raw downstream target IR to disk if "-dump-intermediates" is used.

Will create a file named "./precompiled-<target>-<module>.bin" eg. "precompiled-spirv-mymodule.bin"

### DIFF
--- a/source/slang/slang.cpp
+++ b/source/slang/slang.cpp
@@ -18,6 +18,7 @@
 #include "../core/slang-file-system.h"
 #include "../core/slang-memory-file-system.h"
 #include "../core/slang-writer.h"
+#include "../core/slang-type-text-util.h"
 #include "core/slang-shared-library.h"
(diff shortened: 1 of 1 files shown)


source/slang/slang.cpp
```diff
@@ -3726,6 +3727,37 @@ SlangResult EndToEndCompileRequest::executeActionsInner()
...
+                       }
+                       path << "-";
+                       path << translationUnit->getModule()->getName();
+                       path << ".bin";
+
+                       SLANG_RETURN_ON_FAIL(File::writeAllBytes(path, blob->getBufferPointer(), blob->getBufferSize()));
+                   }
+                   else
+                   {
+                       printf("Failed to get precompiled target code\n");
```
cheneym2, line 3758: Should use diagnostics

PR #6392, opened 2025-02-18
Support for dynamic array of textures access for Texture footprint.

jkwak-work:
This is PR implemented by jsmall on Oct 2023. The change is old and it is unclear if we still need this code or not.

We need to decide if we want to merge or drop.

Note that the source code comment stated that the implementation is "hack".
> // TODO(JS): This a hack to make this work for arrays of resource type.


jkwak-work:
@csyonghe , as discussed on our meeting last week, can you review the change and decide if we want to have the change or drop it?

csyonghe (commented):
What happens when you run the test without the code change?

jkwak-work:
> What happens when you run the test without the code change?

I am getting a compile error like the following when targeting HLSL/DXIL,
```
$ slangc.exe -target hlsl -entry fragmentMain -stage fragment tests/tmp.slang

tests/tmp.slang(59): error 81111: could not determine register/space for a resource or sampler used with NVAPI
void fragmentMain(
     ^~~~~~~~~~~~
```

But it seems like targeting SPIRV is fine.

csyonghe (commented):
it is fine to merge this in, even if it is a partial solution.

csyonghe (approved):

Issue #6358, opened 2025-02-14
Slang takes ~30x as long as shaderc to compile a simple compute shader


NBickford-NV:
Hi Slang team! I've been running into some issues affecting hot-reload workflows, where re-compiling small shaders is common.

The ToT version of Slang (as of https://github.com/shader-slang/slang/commit/944c19b408f20e6c5495678028f62e4e7040aa3b) takes 48-49 ms on my Windows computer to compile the following 841 bytes of source to SPIR-V. This imports no modules, does not include Slang global session creation time or I/O time, uses optimization level 0, and is averaged over 128 runs. I've included a [benchmark](https://github.com/NBickford-NV/slang-compile-timer) you can use to reproduce this issue; more information about it below.

```hlsl
// shader.slang
struct PushConstantCompute
{
  uint64_t bufferAddress;
  uint     numVertices;
};

struct Vertex
{
  float3 position;
};

[[vk::push_constant]]
ConstantBuffer<PushConstantCompute> pushConst;

[shader("compute")]
[numthreads(256, 1, 1)]
void main(uint3 threadIdx : SV_DispatchThreadID)
{
  uint index = threadIdx.x;

  if(index >= pushConst.numVertices)
    return;

  Vertex* vertices = (Vertex*)pushConst.bufferAddress;

  float angle = (index + 1) * 2.3f;

  float3 vertex = vertices[index].position;

  float cosAngle = cos(angle);
  float sinAngle = sin(angle);
  float3x3 rotationMatrix = float3x3(
    cosAngle, -sinAngle, 0.0,
    sinAngle,  cosAngle, 0.0,
         0.0,       0.0, 1.0
  );

  float3 rotatedVertex = mul(rotationMatrix, vertex);

  vertices[index].position = rotatedVertex;
}
```

The options and targets used in my benchmark are:

```cpp
    m_options = {{slang::CompilerOptionName::EmitSpirvDirectly, {slang::CompilerOptionValueKind::Int, 1}},        //
                 {slang::CompilerOptionName::VulkanUseEntryPointName, {slang::CompilerOptionValueKind::Int, 1}},  //
                 {slang::CompilerOptionName::Optimization, {slang::CompilerOptionValueKind::Int, 0}}};
    m_targets = {slang::TargetDesc{.format = SLANG_SPIRV, .profile = m_globalSession->findProfile("spirv_1_6")}};
```

[...]

PR #6396, opened 2025-02-19
Improve performance when compiling small shaders.

csyonghe:
Addresses #6358.

This PR address three performance issues when compiling small shaders:
1. Avoid copying witness table entries that are not getting used during linking.
2. Avoid copying auto-diff related decorations and derivative functions during linking, if the user modules doesn't use autodiff.
3. Cache operator overload resolution results on global session, so each new Session doesn't need to repetitively run through overload resolution from scratch.

These three optimizations combined achieves a 75% reduction in the benchmark provided in #6358.

Copilot summary:

This pull request includes several changes aimed at improving the handling of GLSL modules, enhancing the differentiation capabilities, and optimizing the IR linking process. The most important changes are grouped by their themes below:

### GLSL Module Handling:
* Added handling for GLSL modules in the `SemanticsVisitor::importModuleIntoScope` method to set the `glslModuleDecl` when the module name is "glsl".
* Updated `OperatorOverloadCacheKey` to include an `isGLSLMode` flag and adjusted its hash code calculation accordingly.
* Modified `ResolveInvoke` to set the `isGLSLMode` flag based on the presence of the GLSL module and cache the resolved operator overloads conditionally. [[1]](diffhunk://#diff-2eecdb041b541df14ab93a0cfe35b4e0fbbcb63ef7fdcb004401829bfacdeb77R2512) [[2]](diffhunk://#diff-2eecdb041b541df14ab93a0cfe35b4e0fbbcb63ef7fdcb004401829bfacdeb77R2726-R2733)

### Differentiation Enhancements:
* Introduced the `IDifferentiable` interface with the `KnownBuiltin("IDifferentiable")` attribute.
* Changed the `NullDifferential` struct to remove the `export` keyword.
* Updated the method for finding the `IDifferentiable` interface to use `IRKnownBuiltinDecoration` instead of `IRNameHintDecoration`.

### IR Linking Optimization:
* Implemented the `ImmutableHashedString` class for efficient string handling in the IR linking process.
* Updated the `IRSharedSpecContext::SymbolDictionary` to use `ImmutableHashedString` and added a new dictionary for imported symbols. [[1]](diffhunk://#diff-e9d9d1281f79ee0fed14ff718e315e6a3b2b6aa3584bcb7cf3fadfc09f2e9265L55-R76) [[2]](diffhunk://#diff-e9d9d1281f79ee0fed14ff718e315e6a3b2b6aa3584bcb7cf3fadfc09f2e9265L72-R105)
* Enhanced the cloning process for witness tables to handle decorations and entries more efficiently, including a deep clone condition based on specific decorations. [[1]](diffhunk://#diff-e9d9d1281f79ee0fed14ff718e315e6a3b2b6aa3584bcb7cf3fadfc09f2e9265R685-R715) [[2]](diffhunk://#diff-e9d9d1281f79ee0fed14ff718e315e6a3b2b6aa3584bcb7cf3fadfc09f2e9265R726-R775)

### Type Checking Cache:
* Refactored `TypeCheckingCache` to inherit from `RefObject` and updated its usage in `Linkage` and `Session` classes. [[1]](diffhunk://#diff-de6635fcced55ee0d7f18e86d265f8a58e13fca3b40ac0123ce0ef14306d40eaL302-R304) [[2]](diffhunk://#diff-d90bb752ab912aca45fce6f0c319d91d28245bd4726188b447263613d2e58f8eL2212-R2212) [[3]](diffhunk://#diff-d90bb752ab912aca45fce6f0c319d91d28245bd4726188b447263613d2e58f8eR3592-R3594)

### Debugging and Specialized Targets:
* Added a GLSL-specific debug break function with the `__specialized_for_target(glsl)` attribute and corresponding SPIR-V instruction.

These changes collectively improve the handling of GLSL modules, enhance differentiation capabilities, and optimize the IR linking process, making the codebase more efficient and maintainable.


source/core/slang-string.h
```diff
@@ -790,6 +790,57 @@ class SLANG_RT_API String
     UnownedStringSlice getUnownedSlice() const { return StringRepresentation::asSlice(m_buffer); }
 };
 
+class ImmutableHashedString
+{
+public:
+    String slice;
+    HashCode64 hashCode;
```
jkwak-work, line 797: Probably we should make the members private.
Not sure if we want to make them `const` as well since it is Immutable.

source/slang/hlsl.meta.slang
```diff
@@ -20309,7 +20309,7 @@ void ReorderThread( HitObject HitOrMiss )
 ///
 /// There doesn't appear to be an equivalent for debugBreak for HLSL
 
-
+[require(glsl)]
 __specialized_for_target(glsl)
 [[vk::spirv_instruction(1, "NonSemantic.DebugBreak")]]
 void __glslDebugBreak();
```
jkwak-work, line 20315: Maybe we should move this to glsl.meta.slang?

source/slang/slang-ir-link.cpp
```diff
@@ -52,15 +52,28 @@ struct IRSharedSpecContext
...
+
     IRBuilder builderStorage;
 
     // The "global" specialization environment.
     IRSpecEnv globalEnv;
 };
 
+void insertGlobalValueSymbol(IRSharedSpecContext* sharedContext, IRInst* gv);
+
+struct WitnessTableCloenInfo : RefObject
```
jkwak-work, line 70: Copilot found  typo here.
WitnessTableCloenInfo  -> WitnessTableCloneInfo

jkwak-work (approved):
Looks good to me.

source/slang/slang-ir-link.cpp
csyonghe: good catch, how did you setup copilot to do the review?

PR #6404, opened 2025-02-19
Fix slang raytracing test

ksavoie-nv:
TestRenderApp is missing handling for raytracing tests in a few places, causing the test `tests/pipeline/ray-tracing/raygen.slang (cuda)` to fail. This PR adds the missing handling, fixing the test.


cheneym2:
Does CI pass if you enable optix tests by default?  I think it's this in slang-test:
https://github.com/shader-slang/slang/blob/master/tools/slang-test/slang-test-main.cpp#L4902

csyonghe (approved):

hellfiretoad:
Looks like raygen does not pass.  Is that expected ?

ksavoie-nv:
> Looks like raygen does not pass. Is that expected ?

No, it passed for me locally. I'll investigate.

cheneym2:
I confirmed also that this PR works and it seems to be an issue with the CI machine's optix installation.  Seems like that could be handled subsequently and just get the original fix merged without that bit.

cheneym2:
This also shows that missing a proper OptiX installation causes unexpected failures in slang-test.  I think the symptom should have been that optix tests are unsupported and skipped rather than fail.  There was really no way to tell that optix wasn't being found as it stands.

cheneym2 (approved):

cheneym2:
@ksavoie-nv please re-run the failed windows release test.  Looks like a spurious failure.

Issue #6349, opened 2025-02-12
Handle error case gracefully when SV_TessFactor is incorrectly used.


jkwak-work:
**Problem description**
The repro shader causes SPIRV validation error; and it actually doesn't run at runtime.
Slang should print an error so that the users can fix it.

**Repro steps**
The following shader is from [tests/hlsl/simple-hull-shader-1.slang](https://github.com/shader-slang/slang/blob/master/tests/hlsl/simple-hull-shader-1.slang).

Change the array size as described below,
```
struct HscOut
{
    float EdgeTessFactor[3] : SV_TessFactor; // <=== CHANGED FROM 4
    float InsideTessFactor[1] : SV_InsideTessFactor; // <== CHANGED FROM 2
    uint instanceId;
};
```

Observer the following spirv validation errors:
```
error: line 141: OpStore Pointer <id> '87[%gl_TessLevelOuter]'s type does not match Object <id> '84[%84]'s type.
  OpStore %gl_TessLevelOuter %84

error: line 141: OpStore Pointer <id> '91[%gl_TessLevelInner]'s type does not match Object <id> '88[%88]'s type.
  OpStore %gl_TessLevelInner %88
```

**More details**
According to chatGPT, this has something to do with a fact that `gl_TessLevelOuter` in GLSL is declared as `float[4]`.
When the shader tries to copy from float[3] to float[4], it causes the trouble.

Slang will check if the array side for SV_TessFactor is always 4 and error out if not.

For `SV_InsideTessFactor`, it has the same reason that `gl_TessLevelInner` in GLSL is declared as `float[2]`.

csyonghe:
I believe this can be handled by extending our system value legalization logic to do the type conversion for the user.

jkwak-work:
I agree that we can legalize and fix the mistake for the users; but with a warning.

jkwak-work:
I was wrong to suggest the warning message.
The SPIRV spec doesn't state it explicitly but it seems to be assumed that the unused components will be ignored.
When the tessellation is in "isolines" mode, it will use the first two components of TessLevelOuter.
When it is in "tri" mode, it will use the first three components.
When it is in "quad" mode, it will use all of four components.

I don't think it is a mistake on the user shader side to use 3 components when tessellation mode is in "tri".

PR #6409, opened 2025-02-20
Legalize array size of SV_TessFactor and SV_InsideTessFactor

jkwak-work:
When targeting SPIR-V or GLSL, the type for SV_TessFactor must be float[4]; note that it is not a vector but an array. Similarly the type of SV_InsideTessFactor has to be an array of float[2].

When the user shader declare them as arrays smaller than the required size, Slang will legalize them to the required sizes.

Note that it is not the user mistake to declare floar[3] when the hull mode is in "tri", as an example. The unused components are expected to be ignored.

Note also that HLSL allows the type to be float[2|3|4] whereas GLSL and SPIR-V requires it to be always float[4].

Closes https://github.com/shader-slang/slang/issues/6349

### DIFF
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -605,6 +605,11 @@ DIAGNOSTIC(30019, Error, typeMismatch, "expected an expression of type '$0', got
 DIAGNOSTIC(30021, Error, noApplicationFunction, "$0: no overload takes arguments ($1)")
 DIAGNOSTIC(30022, Error, invalidTypeCast, "invalid type cast between \"$0\" and \"$1\".")
 DIAGNOSTIC(30023, Error, typeHasNoPublicMemberOfName, "\"$0\" does not have public member \"$1\".")
+DIAGNOSTIC(
+    30024,
+    Error,
+    cannotConvertArrayOfSmallerToLargerSize,
+    "Cannot convert array of size $0 to array of size $1 as this would truncate data")
 DIAGNOSTIC(30025, Error, invalidArraySize, "array size must be larger than zero.")
 DIAGNOSTIC(
     30026,
--- a/source/slang/slang-ir-glsl-legalize.cpp
+++ b/source/slang/slang-ir-glsl-legalize.cpp
@@ -1543,6 +1543,17 @@ ScalarizedVal createSimpleGLSLGlobalVarying(
                     if (declaredArraySize == requiredArraySize)
                         break;
 
+                    auto toSize = getIntVal(requiredArraySize);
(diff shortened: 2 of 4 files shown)


source/slang/slang-diagnostic-defs.h
```diff
@@ -2426,6 +2426,18 @@ DIAGNOSTIC(
     Error,
     invalidTessellationDomain,
     "'Domain' should be either 'triangles' or 'quads'.")
+DIAGNOSTIC(
```
csyonghe, line 2429: These are not used any more, can remove?

jkwak-work: Good catch. I will remove it

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -1971,6 +1997,38 @@ ScalarizedVal adaptType(IRBuilder* builder, IRInst* val, IRType* toType, IRType*
...
+                    auto element = builder->emitElementExtract(
+                        fromArray->getElementType(),
+                        val,
+                        builder->getIntValue(builder->getIntType(), i));
+                    elements.add(element);
+                }
+
+                // Fill remaining elements with default value up to target size
+                auto elementType = toArray->getElementType();
+                auto defaultValue = builder->emitDefaultConstruct(elementType);
```
csyonghe, line 2023: we should create `defaultValue` only when toSize>fromSize.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -1971,6 +1997,38 @@ ScalarizedVal adaptType(IRBuilder* builder, IRInst* val, IRType* toType, IRType*
...
+            if (fromArray->getElementCount() != toArray->getElementCount())
+            {
+                List<IRInst*> elements;
+
+                // Get array sizes once
+                auto fromSize = getIntVal(fromArray->getElementCount());
+                auto toSize = getIntVal(toArray->getElementCount());
+
+                // Extract elements one at a time up to the source array size
+                for (Index i = 0; i < fromSize; i++)
```
csyonghe, line 2012: what if toSize is smaller than fromSize?

source/slang/slang-ir-glsl-legalize.cpp
jkwak-work: I see. I will change it

source/slang/slang-ir-glsl-legalize.cpp
jkwak-work: Hmmm. I should print an error fir that

csyonghe (approved):

PR #6410, opened 2025-02-20
Do not perform deep checkout of submodules

### LINKED ISSUE #5465: Don't perform a deep checkout for CI and releases
we probably don't need the tag for CI and releases should be made from the tagged revision itself

we definitely don't need deep checkout for submodules.

mkeshavaNV:
Fixes #5465


csyonghe:
Is this PR ready for review?

mkeshavaNV:
> Is this PR ready for review?

Yeah I think so. I wanted one confirmation from @expipiplus1 , but we can get that in the review. I'm not sure how else we can confirm that indeed we are doing a -depth=1 clone of submodules. Setting as "Ready for review".

.github/workflows/release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: "recursive"
-          fetch-depth: "0"
+          fetch-depth: "1" # Fetch only the latest commit (shallow clone)
```
csyonghe, line 47: I think this won’t work for our release CI, which requires a tag to be available so it can use the tag name as the version string during build. If you just do a shallow checkout, we won’t be able to use git describe to retrieve the current release tag.

csyonghe: To test, you can create a release from your own fork, download the binaries from the release build and run slangc —version, and check if it is reporting the version number correctly from the release tag.

mkeshavaNV: > which requires a tag to be available so it can use the tag name as the version string during build.
I see, ok I will revert the change for the release CI.

expipiplus1 (approved):

PR #6413, opened 2025-02-20
Metal fix

kaizhangNV:
Fix the issue in creating texture view in metal.
  In newTextureView, levelRange should represent the mipmap level range,
  while sliceRange should represent the texture layer range for texture
  array. But the implementation inverses those two wrongly.

Fix the issue when emitting read only texture_buffer<>
  texture_buffer doesn't support access::sample, we should emit access::read when it's read only buffer.

### DIFF
--- a/tools/gfx/metal/metal-device.cpp
+++ b/tools/gfx/metal/metal-device.cpp
@@ -653,8 +653,8 @@ Result DeviceImpl::createTextureView(
     MTL::PixelFormat pixelFormat = desc.format == Format::Unknown
                                        ? textureImpl->m_pixelFormat
                                        : MetalUtil::translatePixelFormat(desc.format);
-    NS::Range levelRange(sr.baseArrayLayer, sr.layerCount);
-    NS::Range sliceRange(sr.mipLevel, sr.mipLevelCount);
+    NS::Range sliceRange(sr.baseArrayLayer, sr.layerCount);
+    NS::Range levelRange(sr.mipLevel, sr.mipLevelCount);
 
     viewImpl->m_textureView = NS::TransferPtr(textureImpl->m_texture->newTextureView(
         pixelFormat,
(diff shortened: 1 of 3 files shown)


csyonghe (dismissed):

tools/gfx/metal/metal-device.cpp
```diff
@@ -404,7 +404,7 @@ Result DeviceImpl::createTextureResource(
         break;
     }
 
-    bool isArray = desc.arraySize > 0;
+    bool isArray = desc.arraySize > 1;
```
jkwak-work, line 407: This was discussed when it was initially submitted.
Please see the discussion:
https://github.com/shader-slang/slang/pull/4331#discussion_r1633978813

jkwak-work: @skallweitNV  for an opinion.

kaizhangNV: I see this, maybe I should fix this problem on sgl side. Will revert this.

csyonghe: I wonder why this is needed? What issue is this fixing?

Are there no 1-element texture arrays in metal?

kaizhangNV: I fix this issue on SGL side. So no fix needed here.

kaizhangNV: The issue I'm trying to fix is that SGL presume that `desc.arraySize == 1` is a non-array texture, however, slang-gfx assume this is a texture array with array size of 1. Those assumption conflicts.

kaizhangNV:
In the new commit, I fix an obvious bug in metal.

kaizhangNV:
> In the new commit, I fix an obvious bug in metal. [8db65b6](https://github.com/shader-slang/slang/commit/8db65b6335745c82784743ce835e5fe719d12324)

FYI @skallweitNV the same issue is also in slang-rhi.

csyonghe (approved):

PR #6416, opened 2025-02-20
Add inner texture type to reflection json

Devon7925:
This will allow integer storage textures to be used in playground by letting us infer more formats.


csyonghe (dismissed):

Devon7925:
I don't understand what's going on with the `tests/reflection/unbounded-arrays.hlsl.1` test. The names look completely different from the test in the file, and the test succeeds on my computer.

csyonghe (approved):

csyonghe (approved):

csyonghe:
what do you mean by "test failed on my computer"? which test is failing?

Devon7925:
I think my testing version was wrong, seems to work fine now. Although it looks like something is generally wrong with testing right now since my pr shouldn't be breaking that CI test.

csyonghe:
Yeah, that CI failure is irrelevant is already being fixed in top of tree. The PR looks good!

Issue #6364, opened 2025-02-14
Slang generates duplicated structs with identical members


jkwak-work:
**Problem description**
Slang generates duplicated struct-s whose names are different but their member variables are identical.
And Slang use them interchanging way, which is fine with DXC 1.7 but causes an error with DXC 1.9.

**Repro step**
The following shader can reproduce the issue.
```
// repro.slang
RWTexture2D<float> gOutputColor;

struct ShadingFrame : IDifferentiable
{
    float3 T;
}

[Differentiable]
float computeRay()
{
    float3 dir = 1.f;
    return dot(dir, dir);
}

[Differentiable]
float paramRay()
{
    DifferentialPair<float> dpDir = fwd_diff(computeRay)();
    return dpDir.p;
}

[Shader("compute")]
[NumThreads(1, 1, 1)]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
    DifferentialPair<float> dpColor = fwd_diff(paramRay)();
    gOutputColor[0] = dpColor.p;
}
```

Compile the shader
```
slangc.exe -stage compute -entry computeMain -target hlsl -o repro.hlsl repro.slang
```

Observer two structs whose names are different but the members are identical.
```
// repro.hlsl
struct DiffPair_1
{
    DiffPair_vectorx3Cfloatx2C3x3E_0 primal_0;
    DiffPair_vectorx3Cfloatx2C3x3E_0 differential_0;
};

struct DiffPair_2
{
    DiffPair_vectorx3Cfloatx2C3x3E_0 primal_0;
    DiffPair_vectorx3Cfloatx2C3x3E_0 differential_0;
};

DiffPair_0 s_fwd_d_dot_0(DiffPair_2 dpdpx_0, DiffPair_2 dpdpy_0)

// ... omit ...

    DiffPair_1 _S6 = { _S4, _S5 };
    DiffPair_0 _S7 = s_fwd_d_dot_0(_S6, _S6); // ERROR: DXC expects DiffPair_2 type not DiffPair_1
```

jkwak-work:
This might be a recent regression.
I am investigating.

jkwak-work:
This issue is caused by a recent change.
 - https://github.com/shader-slang/slang/pull/6275

@kaizhangNV , can we revert the change and find a better fix?

jkwak-work:
Currently WIP PR is here:
- https://github.com/shader-slang/slang/pull/6417

[...]

PR #6417, opened 2025-02-21
Make IRWitnessTable HOISTABLE

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6364

# Make `IRWitnessTable` Hoistable

## Intention of the PR
This commit makes `IRWitnessTable` Hoistable so that we can avoid duplicated `IRWitnessTable`.

## Problems
This commit tries to address the following issues arise after turning `IRWitnessTable` into Hoistable:
 1. A Hoistable instance is immutable.
 2. When tries to create a duplicated child, you will get a previously created instance of `IRWitnessTable`, instead of a new one.
 3. We don't actually want to hoist `IRWitnessTable`.
 4. There can be only one instance of Hoistable and it cannot appear as childs multiple times.
 5. Different import/export mangled names were used for the same Witness-table when its type is "enum" interface.

### Solution for "1. A Hoistable instance is immutable."
`IRWitnessTable::setConcreteType()` is removed, because when an `IRInst` is Hoistable, it is treated as immutable. Any `IRInst::setXXX()` methods don't work anymore.

There were two places calling `setConcreteType()` and their logic had to change little bit.

`DeclLoweringVisitor::visitInheritanceDecl()` in `source/slang/slang-lower-to-ir.cpp` was calling `setConcreteType()`. It had a little strange logic around `lowerType()`. The `IRWitnessTable` was added with `context->setGlobalValue()` first and its `concreteType` was changed later. This commit works around in a way that it sets the parent of `IRWitnessTable` temporarily and reset it with the correct `IRWitnessTable`. Without this logic, it went into an infinite recursion.

`AutoDiffPass::fillDifferentialTypeImplementation()` in `source/slang/slang-ir-autodiff.cpp` was calling `setConcreteType()`. It was changing the concreteType of `innerResult.diffWitness`. This commit creates a new `IRWitnessTable` and copies its `IRWitnessTableEntry`.

### Solution for "2. When tries to create a duplicated child, you will get a previously created instance of IRWitnessTable, instead of a new one"

After a call to `IRBuilder::createWitnessTable()`, this commit checks if the returned `IRWitnessTable` is a brand new or not. If it is not a new one, we have to avoid adding the decorations and children.

This commit decides when to add decorations and children based on whether `IRWitnessTable` has any of decorations or children already. It doesn't seem like a proper way to check. But when I tried, it was difficult to find a bottleneck point where the decorations and children are added to `IRWitnessTable` first time. Note that we are not trying to find when `IRWitnessTable` is created for the first time; we need to find if the decorations and children were added once.

It might be fine to have duplicated `IRWitnessTableEntry` in most of the cases, but I noticed that it fails an assertion check when `shouldDeepCloneWitnessTable()` returns false in `cloneWitnessTableImpl()`.

### Solution for "3. We don't actually want to hoist IRWitnessTable."

The reason why this commit makes `IRWitnessTable` is to prevent the duplicated instances of `IRInst`. But we don't really want to "Hoist" them.

When an `IRWitnessTable` gets Hoisted out, it causes unexpected problems and the specialization process fails due to the missing `IRWitnessTable` in the input.

This commit prevent from hoisting `IRWitnessTable` in `_replaceInstUsesWith()`. The way this is implemented feel little hack but we discussed on Slack and decided to go with this. One of the proper approaches could be to add a new flag in `IROpFlags` and have a new one like `kIROpFlag_Deduplicate`, which is different from just `kIROpFlag_Hoistable`.

### Solution for "4. There can be only one instance of Hoistable and it cannot appear as childs multiple times."

When `IRWitnessTable` is Hoistable, there can be only a unique set of instances. And we cannot have an instance as a duplicated childs. It is because `IRInst` has only one set of `IRInst* next` and `IRInst* prev`.

Before this commit, an instance of `IRGeneral` could have duplicated instances of `IRWitnessTable`. As an example, `IInteger` interface inherits two other interfaces, `IArithmetic` and `ILogical`. And they both inherits from `IComparable`.
```
interface IInteger : IArithmetic, ILogical {}
interface IArithmetic : IComparable {}
interface ILogical : IComparable
```
When we specialize it in `specializeGenericImpl()`, an `IRBlock` gets the following list of children:
- IRWitnessTable for IComparable,
- IRWitnessTable for IArithmetic,
- IRWitnessTable for IComparable,
- IRWitnessTable for ILogical,

For the cloning during the specialize, "IRWitnessTable for `IComparable`" must be cloned before the cloning of "IRWitnessTable for `IArithmetic`". Because "IRWitnessTable for `IArithmetic`" refers "IRWitnessTable for `IComparable`" as its `IRWitnessTableEntry`. The order they appear in the `IRBlock` as children decides which instances will be cloned first. And "IRWitnessTable for `IComparable`" must appear before "IRWitnessTable for `IArithmetic`".

Note that "IRWitnessTable for `IComparable`" appears twice, The first one was added for "IRWitnessTable for `IArithmetic`". And the second one is added for "IRWitnessTable for `ILogical`".

With this commit "IRWitnessTable for `IComparable`" can appear as a child only once in `IRBlock`. So it causes an error if it gets the following list:
- IRWitnessTable for IArithmetic,
- IRWitnessTable for IComparable,
- IRWitnessTable for ILogical,

In order to resolve the problem, "IRWitnessTable for `IComparable`" must appear before both "IRWitnessTable for `IArithmetic`" and "IRWitnessTable for `ILogical`" as following:
- IRWitnessTable for IComparable,
- IRWitnessTable for IArithmetic,
- IRWitnessTable for ILogical,

To address the problem, the instances of `IRWitnessTable` is always added to the end of the children list. If it is already added to the list, we don't move. This works out because the AST tree is built based on the dependencies.

### Solution for "5. Different import/export mangled names were used for the same Witness-table when its type is "enum" interface."

This issue was found while testing with Falcor tests where it uses Conformance-type feature of Slang.
We are using different import and export mangled names for a same Witness-table when the witness-table is for "Enum" interface.

The way we simplify the implementation of "Enum" causes a problem when it comes to generate export/import for the witness-table. And the exact repro step is still unclear.

There were two suggested solutions for the problem and this PR adopted the first option for now. Maybe we want to improve it with the second option later.

option 1, when we produce mangled names for those witness-table, we can use a mangled name with the underlying "int" type instead of the name of the enum type. In this way, all witness-tables for enum types whose underlying type is same will get the same mangled name. It will allow us to deduplicate the witness-table during the linking.

option 2, we can preserve type info for enum type when generating IR. We can still erase all other uses of the type info of enum types for now. But when we generate the witness-table, instead of filling the conforming type operand to IntType, we fill it as EnumType(IntType) where EnumType is a new global IROp code to represent all enum types (like InterfaceType/StructType). This way the operands for the two witness-tables will be different.

"option 1" is more quick and dirty and "option 2" is more proper way to address it.
I should go with "option 1" and improve it with "option 2" approach later.


jkwak-work:
This PR is still work in progress, but any comments and feedback will be welcomed.

Currently 17 tests are failing with this PR:
- tests/autodiff/existential-1.slang
- tests/autodiff/existential-2.slang
- tests/bugs/type-legalize-bug-1.slang
- tests/compute/assoctype-func-param.slang
- tests/compute/byte-address-buffer.slang
- tests/compute/column-major.slang
- tests/language-feature/generics/dependent-generic-2.slang
- tests/language-feature/generics/dependent-generic-3.slang
- tests/language-feature/generics/dependent-generic.slang
- tests/language-feature/generics/generic-assoc-type-constraint.slang
- tests/language-feature/generics/generic-interface-2.slang
- tests/language-feature/generics/generic-return-type-requirement.slang
- tests/language-feature/generics/generic-witness-derived.slang
- tests/language-feature/generics/variadic-0.slang
- tests/language-feature/interfaces/generic-requirement-synth-2.slang
- tests/language-feature/interfaces/is-as-dynamic.slang
- tests/metal/texture.slang

jkwak-work:
I pushed more code change.
It is still failing on two tests and I am working on it:
- tests/compute/assoctype-func-param.slang
- tests/metal/texture.slang

jkwak-work:
For the remaining problem, here is a minimal repro shader,
```
RWStructuredBuffer<bool> outputBuffer;


__generic<T : IArithmetic>
bool test()
{
    return T(0) == T(0);
}

[numthreads(1, 1, 1)]
void computeMain()
{
    outputBuffer[0] = test<int4>();
}
```

When compiled, I get the following error,
```
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: unimplemented: Unhandled global inst in spirv-emit:
let  %1 : %2    = undefined
```

It appears that some of IRinst-s are removed by DCE and their uses are replaced by "undefined" incorrectly.
When dumped IR, I see that some of functions are not specialized "AFTER-SPECIALIZE", probably because witness-tables are not properly applied.
I will keep digging.

jkwak-work:
With my latest changes, the only test failing is
- tests/compute/assoctype-func-param.slang

jkwak-work:
It looks like my PR is failing on a new test,
 - tests/bugs/gh-6504-linker.slang

And Falcor test is also crashing at
 - BSDFIntegratorTests.cpp:BSDFIntegrator (D3D12)

jkwak-work:
Changing to draft until I fix the failing test and falcor tests

jkwak-work:
It turned out that the test failure of the following test was due to an incorrect resolve of the merge conflict.
 - `tests/bugs/gh-6504-linker.slang`

jkwak-work:
Changed the status from "draft" to "ready for review".

source/slang/slang-ir.cpp
```diff
@@ -1687,12 +1687,34 @@ void addHoistableInst(IRBuilder* builder, IRInst* inst)
     //
     IRInst* insertBeforeInst = parent->getFirstChild();
 
-    // Hoistable instructions are always "ordinary"
-    // instructions, so they need to come after
-    // any parameters of the parent.
-    //
-    while (insertBeforeInst && insertBeforeInst->getOp() == kIROp_Param)
-        insertBeforeInst = insertBeforeInst->getNextInst();
+    if (inst->getOp() == kIROp_WitnessTable)
```
csyonghe, line 1690: I think instead of doing this adhoc logic for witness tables, you want to return immediately without moving the instruction if you are creating a hoistable inst and find that the hoistable inst already exists. We should also look at the call side of addHoistableInst and just don’t call it if we are dealing with an inst that already exists. We can make it so that a freshly created houstable inst has parent == nullptr, so whenever we see a inst has parent != nullptr just exit immediately.

source/slang/slang-ir.cpp
```diff
@@ -4621,6 +4643,18 @@ void addGlobalValue(IRBuilder* builder, IRInst* value)
         parent = builder->getModule()->getModuleInst();
     }
 
+    // If the value is already in the parent, keep it as-is.
+    // Because WitnessTable is Hoistable, the parent can have
+    // only one instance of this WitnessTable. The order among
+    // siblings should remain because the later siblings may
+    // have dependency to the earlier siblings.
+    //
+    if (parent == value->parent && value->getOp() == kIROp_WitnessTable)
```
csyonghe, line 4652: Same here, I believe we can change the logic to exit early if inst already has a parent.

source/slang/slang-mangle.cpp
```diff
@@ -824,6 +824,37 @@ String getMangledNameForConformanceWitness(ASTBuilder* astBuilder, Type* sub, Ty
...
+// described on the issue 6364 is more proper and ideal solution for the issue.
+//
+String getMangledNameForConformanceWitness(ASTBuilder* astBuilder, Type* sub, Type* sup, IROp subOp)
+{
+    SLANG_AST_BUILDER_RAII(astBuilder);
+
+    ManglingContext context(astBuilder);
+    emitRaw(&context, "_SW");
+
+    if (ASTNodeType(sup->getClassInfo().m_classId) == ASTNodeType::EnumTypeType)
```
csyonghe, line 845: `isDeclRefTypeOf<EnumDecl>()`

source/slang/slang-ir-autodiff.cpp
```diff
@@ -1862,17 +1862,21 @@ IRInst* DifferentiableTypeConformanceContext::buildDifferentiablePairWitness(
...
-            table,
-            sharedContext->differentialAssocTypeStructKey,
-            diffDiffPairType);
-        builder->createWitnessTableEntry(
-            table,
-            sharedContext->differentialAssocTypeWitnessStructKey,
-            table);
-        builder->createWitnessTableEntry(table, sharedContext->addMethodStructKey, addMethod);
-        builder->createWitnessTableEntry(table, sharedContext->zeroMethodStructKey, zeroMethod);
+        // Add WitnessTableEntry only once
```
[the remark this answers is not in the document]
jkwak-work: I will do that.

source/slang/slang-ir.cpp
jkwak-work: I will test if that approach works.

jkwak-work: After testing a few different implementation for this idea, I found that the condition `inst->parent != nullptr` is not enough. I didn't check for the details of why but it looks like the condition must be `parent != inst->parent`.

And the current implementation is the earliest point where we know what `parent` is.

source/slang/slang-ir.cpp
jkwak-work: As as the other case. I was trying to find an earlier point where we can exit.
But the condition needs to be `parent == value->parent` and `value->parent != nullptr` is not enough.
The current implementation is the earliest location where we know the value of `parent`.

jkwak-work:
I pushed a new change with minor clean up as Yong suggested.
But I couldn't adopt the idea of moving the exit logic earlier.

jkwak-work:
It seems like I used `isDeclRefTypeOf()` incorrectly.
I am gonna re-test the idea of exiting earilier.

source/slang/slang-mangle.cpp
jkwak-work: It turned out that I can just use `as<EnumTypeType>(inst)`

jkwak-work:
I pushed a new change after resolving all the Falcor failure issues that was recently regressed by another commit.

The code now skips `addHoistableInst()` for hoistable instances whose parent value is already set.

jkwak-work:
All CI tests are (finally) passing.

source/slang/slang-ir.cpp
csyonghe: This change should not be needed anymore, right?

csyonghe (commented):
Looks good overall. I think the addHoistableInst change can be reverted now that the function is not even called when the inst being created is an existing inst.

source/slang/slang-ir.cpp
jkwak-work: This is still needed, because of the order where the inst is inserted.
Without my change, the witness-table is inserted earlier than it should be, which causes a dependency problem during the specialize.
I tried to replace the existing code with my new code, and it was also causing troubles, which I didn't investigate deeper.

jkwak-work:
> Looks good overall. I think the addHoistableInst change can be reverted now that the function is not even called when the inst being created is an existing inst.

That code is still needed when a Witness-Table is inserted for the first time.

csyonghe (commented):
I think what we are implementing here is a new category of Inst: deduplicable but not hoistable.

Instead of changing addHoistableInst, we should just not call addHoistableInst for witness tables, and call addInst() instead to insert the witness table at the current insert location of the IRBuilder.

jkwak-work:
> I think what we are implementing here is a new category of Inst: deduplicable but not hoistable.
> 
> Instead of changing addHoistableInst, we should just not call addHoistableInst for witness tables, and call addInst() instead to insert the witness table at the current insert location of the IRBuilder.

I see. I will try it tomorrow.

source/slang/slang-ir.cpp
```diff
@@ -2613,7 +2636,16 @@ IRInst* IRBuilder::_findOrEmitHoistableInst(
...
     }
 
-    addHoistableInst(this, inst);
+    // When an hoistable inst is already a child, skip adding it.
+    if (inst->parent == nullptr)
+    {
+        // In order to de-duplicate them, Witness-table is marked as Hoistable.
+        // But it is not exactly a hoistable type and it can be added simpler.
+        if (inst->getOp() == kIROp_WitnessTable)
+            addDeduplicatedInst(this, inst);
```
csyonghe, line 2645: wouldn't `addInst(inst)` work?

jkwak-work: I will try

jkwak-work: I need to take it back.
This doesn't seem to be working.
It is causing a slang-test failure on one of existing tests, `tests/autodiff/dynamic-dispatch-material.slang`.

I am gonna revert the last commit of this PR.

jkwak-work: I am gonna need to retry this, because I made a mistake when I was testing my change for this.

I noticed that there is a member method, `IRBuilder::addInst(IRInst*)`.
So when I added a new function with the same name, it was probably calling the existing one not the new one.

jkwak-work: It worked as young suggested. I just had to rename the function from addInst to another.

jkwak-work:
This PR is ready for another review.

source/slang/slang-ir.cpp
```diff
@@ -2613,7 +2642,16 @@ IRInst* IRBuilder::_findOrEmitHoistableInst(
...
     }
 
-    addHoistableInst(this, inst);
+    // When an hoistable inst is already a child, skip adding it.
+    if (inst->parent == nullptr)
+    {
+        // In order to de-duplicate them, Witness-table is marked as Hoistable.
+        // But it is not exactly a hoistable type and it can be added simpler.
+        if (inst->getOp() == kIROp_WitnessTable)
+            addDeduplicatedInst(inst);
```
csyonghe, line 2651: Ah, sorry I wasn’t communicating my meaning clearly earlier. My question was is it sufficient to just call the existing addInst function here, instead of calling the new addDeduplicatedInst?

jkwak-work: When I used the exisitng `IRBuilder::addInst()` , one of tests were failing. So no, I couldn't use the function.
```
tests/autodiff/dynamic-dispatch-material.slang
```

jkwak-work: I think I can investigate why it fails if `IRBuilder::addInst()` is preferred.
Considering that it is only failing test, I think there might be a problem for the specific use case.

jkwak-work: I couldn't make it to work with `IRBuilder::addInst()` but it worked with `addGlobalValue()`.

csyonghe (approved):

Issue #6391, opened 2025-02-18
[WorkGraphs] experimental shader code causes Slang to crash


jkwak-work:
The following code uses WG feature and it causes Slang to crash.
```
struct sPayloadType {
    uint value;
};

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 dispatchThreadId : SV_GroupThreadID)
{
    spirv_asm {
        OpCapability ShaderEnqueueAMDX;
        OpExtension "SPV_AMDX_shader_enqueue";
        OpExecutionModeId $main ShaderIndexAMDX $(0);
        OpExecutionMode $main StaticNumWorkgroupsAMDX $(1) $(1) $(1);
    };

    DispatchNodeInputRecord<sPayloadType> inputData;
    NodePayloadPtr<sPayloadType> payload = inputData.Get();

    payload.value = 0;
}
```
When compiled, Slang prints the following error message,
```
slangc.exe -target spirv-asm -entry main -stage compute test.slang

(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: assert failure: The key already exists in Dictionary.
```

jkwak-work:
This issue is caused by my mistake on my previous [PR](https://github.com/shader-slang/slang/pull/6052).
The following line is causing the problem.
```
    auto spvValueType = ensureInst(valueType);
    auto spvNodePayloadType = emitOpTypeNodePayloadArray(inst, spvValueType);
    valueTypeId = getID(spvNodePayloadType);
```
The call to `ensureInst()` should have performed `emitOpTypeNodePayloadArray()`.

A quick workaround is to make the following change,
```
    auto spvNodePayloadType = emitOpTypeNodePayloadArray(inst, spvValueType); // <==== FROM
    auto spvNodePayloadType = emitOpTypeNodePayloadArray(nullptr, spvValueType); // <==== TO
```

I will work on a proper fix.

jkwak-work:
I think the workaround proposed on my previous comment is actually a proper way to fix the problem.

jkwak-work:
I had a PR for this issue.
- https://github.com/shader-slang/slang/pull/6470

It handles the problem case properly without the workaround mentioned on the previous comments.

Closing the issue.

PR #6423, opened 2025-02-22
Fix SPIRV error for WorkGraphs

jkwak-work:
This commit resolves an error message that is caused while emitting SPIRV code.

The cause was on the fact that there is no corresponding IRInst for one of SPIRV instructions when emitting,
```
%in_payload_t = OpTypeNodePayloadArrayAMDX $$T;
%in_payload_ptr_t = OpTypePointer NodePayloadAMDX %in_payload_t;
```

There is an IRInst for `OpTypePointer` but there isn't IRInst corresponding to `OpTypeNodePayloadArrayAMDX`. When Slang emits `OpTypePointer` with its second operand `OpTypeNodePayloadArrayAMDX`, it is going to be emitted without a matching IRInst.

Closes https://github.com/shader-slang/slang/issues/6391

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -24407,9 +24407,7 @@ struct DispatchNodeInputRecord
         case spirv:
             return spirv_asm
             {
-                %in_payload_t = OpTypeNodePayloadArrayAMDX $$T;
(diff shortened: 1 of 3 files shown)


jkwak-work:
Closing it because I realized that this is not a good way to solve this.

PR #6424, opened 2025-02-22
Allow `.member` syntax on vector and scalars.

csyonghe:


source/slang/slang-language-server-completion.cpp
```diff
@@ -724,6 +726,18 @@ CompletionResult CompletionContext::createSwizzleCandidates(
             result.add(item);
         }
     }
+    else if (auto scalarType = as<BasicExpressionType>(type))
+    {
+        String typeStr;
+        if (scalarType)
```
cheneym2, line 732: Redundant conditional

cheneym2:
I don't see a test that verifies .member on scalars, as the commit title suggests.  Sorry if I'm just not seeing what I'm seeing.  Can you extend a scalar with a method and use it like in vector-member.slang?

csyonghe:
Sure, will add a test for scalar.

source/slang/slang-language-server-completion.cpp
csyonghe: Fixed.

cheneym2 (approved):

PR #6436, opened 2025-02-24
set depth 2 for CI submodules

mkeshavaNV:


expipiplus1 (approved):

mkeshavaNV:
> What problem does this solve? Did we have a problem with this?

yes, depth=1 was just behaving as in no changes were added. So CIs would just pass. Depth=2 was what we needed.

PR #6444, opened 2025-02-24
Fix TypeCheckingCache concurrency and candidate lifetime.

csyonghe:
If `candidate.candidate.item.declRef.declRefBase` is a genericapp declref created in previous session, it will be an invalid pointer, and `as<DirectDeclRef>(candidate.candidate.item.declRef.declRefBase)` will crash. We should instead use the decl to check if it is a generic decl ref.


kaizhangNV (approved):
LGTM

Issue #6394, opened 2025-02-18
Slang.D backward mode doesn't run lines of code after a for loop


yashbelhe:
The code below shows a simple example of the bug. 

```
[Differentiable]
[AutoPyBindCUDA]
[CUDAKernel]
void run(
    DiffTensorView<float> x, DiffTensorView<float> ret)
{
    int idx = (cudaBlockIdx() * cudaBlockDim() + cudaThreadIdx()).x;
    if (idx >= 1) return;

    if (idx == 0) {
        printf("entered the kernel \n");
    }

    for (int i = 0; i < 1; i++) {
        if (idx > 0) {
            printf("this doesn't print since idx>0 is never true \n");
            return;
        }

        if (idx == 0) {
            printf("inside the loop \n");
        }
    }

    if (idx == 0) {
        printf("reached outside \n");
    }
    ret[idx] = x[idx,0];
}
```

In forward mode, when the code is run with x=[0.1,0.2], the following is printed
```
entered the kernel 
inside the loop 
reached outside 
entered the kernel 
inside the loop 
entered the kernel 
inside the loop
```
and x.grad = [0.0, 0.0], which is incorrect.

Ideally, the code should produce
```
entered the kernel 
inside the loop 
reached outside 
entered the kernel 
inside the loop 
reached outside 
entered the kernel 
inside the loop 
reached outside
```
and x.grad=[1,0].

A slightly modified version of the code, which comments out the for loop doesn't produce this bug

```
[Differentiable]
[AutoPyBindCUDA]
[CUDAKernel]
void run(
    DiffTensorView<float> x, DiffTensorView<float> ret)
{
    int idx = (cudaBlockIdx() * cudaBlockDim() + cudaThreadIdx()).x;
    if (idx >= 1) return;

    if (idx == 0) {
        printf("entered the kernel \n");
    }

    // for (int i = 0; i < 1; i++) {
        if (idx > 0) {
            printf("this doesn't print since idx>0 is never true \n");
            return;
        }

        if (idx == 0) {
            printf("inside the loop \n");
        }
    // }

    if (idx == 0) {
        printf("reached outside \n");
    }
    ret[idx] = x[idx,0];
}
```
The snippet above produces the right output with the correct print statements and gradients. This means the for loop is correlated with the bug.

Additionally the bug seems to correlated with the return statement within the for loop. The first if statement within the for loop never evaluates to true, yet, its presence causes the bug.

If you remove that return statement, we get the following code, 
```
[...]

Issue #5999, opened 2025-01-03
Error -11 when using printf in differentiable function


half-potato:
I get an error -11 when using printf in a differentiable function. I am using slangtorch on a function that is being bound to a differentiable pytorch function.

slangtorch version: 1.3.4

Also, side note, I need to add a ```#define __STDC_FORMAT_MACROS``` to the top of ```tools/render-test/shader-input-layout.cpp``` and ```source/slang/slang-emit-source-writer.cpp``` to avoid the following error for the latest build:
```
slang/tools/render-test/shader-input-layout.cpp:1151:33: error: expected ')' before 'PRId64'
 1151 |                 writer.print("%" PRId64 "\n", v);
      |                             ~   ^~~~~~~
      |                                 )
```

half-potato:
Still quite a difficult bug to work with. No printline debugging for me :(

saipraveenb25:
Sorry it took so long to get around to this issue. I accidentally assumed a prior PR fixed it. [#6600](https://github.com/shader-slang/slang/pull/6600) should close this issue

PR #6446, opened 2025-02-24
Fix a bug with hoisting 'IRVar' insts that are used outside the loop

saipraveenb25:
- We introduce a 'OpCheckpointObject' inst and use that to split loop state insts into two pieces (one for within-loop uses and one for outside-loop uses.
- This allows the two kinds of uses to be handled separately by the hoisting mechanism
- `OpCheckpointObject` is then lowered to a no-op after hoisting is complete.

Also added a test to check for this.

Fixes: https://github.com/shader-slang/slang/issues/6394 and https://github.com/shader-slang/slang/issues/5999

### DIFF
--- a/source/slang/slang-ir-autodiff-primal-hoist.cpp
+++ b/source/slang/slang-ir-autodiff-primal-hoist.cpp
@@ -281,6 +281,138 @@ static Dictionary<IRBlock*, IRBlock*> createPrimalRecomputeBlocks(
     return recomputeBlockMap;
 }
 
+// Checks if list A is a subset of list B by comparing their primal count parameters.
+//
+// Parameters:
+//   indicesA - First list of IndexTrackingInfo to compare
+//   indicesB - Second list of IndexTrackingInfo to compare
+//
+// Returns:
+//   true if all indices in indicesA are present in indicesB, false otherwise
+//
+bool areIndicesSubsetOf(List<IndexTrackingInfo>& indicesA, List<IndexTrackingInfo>& indicesB)
+{
+    if (indicesA.getCount() > indicesB.getCount())
+        return false;
+
+    for (Index ii = 0; ii < indicesA.getCount(); ii++)
+    {
+        if (indicesA[ii].primalCountParam != indicesB[ii].primalCountParam)
+            return false;
+    }
+
+    return true;
+}
+
+bool canInstBeStored(IRInst* inst)
+{
+    // Cannot store insts whose value is a type or a witness table, or a function.
+    // These insts get lowered to target-specific logic, and cannot be
+    // stored into variables or context structs as normal values.
+    //
+    if (as<IRTypeType>(inst->getDataType()) || as<IRWitnessTableType>(inst->getDataType()) ||
+        as<IRTypeKind>(inst->getDataType()) || as<IRFuncType>(inst->getDataType()) ||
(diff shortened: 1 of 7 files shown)


source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -281,6 +281,138 @@ static Dictionary<IRBlock*, IRBlock*> createPrimalRecomputeBlocks(
...
+                else
+                    afterLoopUses.add(use);
+            }
+
+            // If inst has uses both inside and after the loop, create a copy for after-loop uses
+            if (loopUses.getCount() > 0 && afterLoopUses.getCount() > 0)
+            {
+                setInsertAfterOrdinaryInst(&builder, inst);
+                auto copy = builder.emitCheckpointObject(inst);
+                copy->sourceLoc = inst->sourceLoc; // Copy source location so that checkpoint
```
csyonghe, line 402: long multiline comment should be placed before the line.

source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -281,6 +281,138 @@ static Dictionary<IRBlock*, IRBlock*> createPrimalRecomputeBlocks(
...
+            if (as<IRTerminatorInst>(inst) || !canInstBeStored(inst))
+                continue;
+            // Shouldn't see any vars.
+            SLANG_ASSERT(!as<IRVar>(inst));
+
+            HashSet<IRUse*> loopUses;
+            HashSet<IRUse*> afterLoopUses;
+
+            // Get the indices for the condition block
+            auto condBlockIndices = indexedBlockInfo.getValue(condBlock);
```
csyonghe, line 378: I want us to be extra careful when using `List` types inside loops like this. The code here would be alloc/free-ing memory like crazy just to represent the temporary list objects. Should return and use references if `indexedBlockInfo` is immutable in this loop.

source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -281,6 +281,138 @@ static Dictionary<IRBlock*, IRBlock*> createPrimalRecomputeBlocks(
...
+        //
+        for (auto inst = condBlock->getFirstInst(); inst; inst = inst->getNextInst())
+        {
+            // Skip terminators and insts that can't be stored
+            if (as<IRTerminatorInst>(inst) || !canInstBeStored(inst))
+                continue;
+            // Shouldn't see any vars.
+            SLANG_ASSERT(!as<IRVar>(inst));
+
+            HashSet<IRUse*> loopUses;
```
csyonghe, line 374: any reason to use `HashSet`? it seems `List` would suffice.

also move these outside of the loop, and do a `loopUses.clear()` here to avoid frequent `alloc`.

source/slang/slang-ir-autodiff-primal-hoist.cpp
saipraveenb25: Fixed!

source/slang/slang-ir-autodiff-primal-hoist.cpp
saipraveenb25: Fixed!

csyonghe (approved):

Issue #6437, opened 2025-02-24
performing layout generation after linking causes non local structs to have no layout information


expipiplus1:
Consider the following

main.slang:
```slang
public interface IFoo
{
    public float4 getFoo();
};
public extern struct S : IFoo;

[shader("vertex")]
float4 vs_main(S params) : SV_Position
{
    return params.getFoo();
}

```

foo.slang:
```slang
import main;

export public struct S : IFoo
{
    public float4 getFoo() { return this.foo; }

    float4 foo;
}
```

When these are compiled and linked, we generate the layout information for S before linking with the definition of S in foo.slang

This is the diff between a good version, in which the definition for S is available, and the bad version above. Observe how the details of the struct layout are missign 

```diff
11161,11164c11111
< let  %10        : Void  = varLayout(%11, %12, %6)
< let  %11        : Void  = structTypeLayout(%13, %14)
< let  %14        : Void  = fieldLayout(%foo, %15)
< let  %15        : Void  = varLayout(%16, %12, %6)
---
> let  %10        : Void  = varLayout(%11, %6)
11166,11168c11113
< let  %12        : Void  = offset(5 : Int, 0 : Int)
< let  %16        : Void  = typeLayout(%13)
< let  %13        : Void  = size(5 : Int, 1 : Int)
---
> let  %11        : Void  = structTypeLayout
11190c11135
< let  %17        : rtti_type     = rtti_object(%S)
---
> let  %12        : rtti_type     = rtti_object(%S)
11199c11144
< block %18(
---
> block %13(
11202,11203c11147,11148
<         let  %19        : Vec(Float, 4 : Int)   = get_field(%this, %foo)
<         return_val(%19)
---
>         let  %14        : Vec(Float, 4 : Int)   = get_field(%this, %foo)
>         return_val(%14)
11213c11158
< block %20(
---
> block %15(
11217,11219c11162,11164
<         let  %21        : %S    = load(%params)
<         let  %22        : Vec(Float, 4 : Int)   = call %Sx5FgetFoo(%21)
<         return_val(%22)
---
>         let  %16        : %S    = load(%params)
>         let  %17        : Vec(Float, 4 : Int)   = call %Sx5FgetFoo(%16)
>         return_val(%17)
11222,11243c11167
< #version 450
< layout(row_major) uniform;
< layout(row_major) buffer;
< struct S_0
< {
<     vec4 foo_0;
< };
<
< vec4 S_getFoo_0(S_0 this_0)
< {
[...]

Issue #5994, opened 2025-01-03
System values in a link time structure are not legal when compiling to targets other than hlsl/dxil


lodinukal:
Here is a set of files which can be used to reproduce this https://github.com/lodinukal/slang-link-type-bug-repro.

Code generated for hlsl looks correct, with the system value being passed properly and used:
```hlsl
#line 12832
    float _S1 = float(params_0.inner_0._instance_id_offset_0 + BaseAttributes_0.base_instance_0);
```
However the code doesn't link when targeting spriv/metal, with it erroring when attempting to legalize the entrypoint parameters (in slang-ir-glsl-legalize.cpp and the corresponding metal file). When debugging, I found that usually the structTypeLayout has at least 1 operand usually, but with linking the module, it had 0. Not sure if that helps.

expipiplus1:
Blocked on https://github.com/shader-slang/slang/issues/6437

bmillsNV:
Still not passing CI. Moving back to In Progress.

PR #6450, opened 2025-02-25
Resolve 'extern' types during type layout generation if possible

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/5994 Closes https://github.com/shader-slang/slang/issues/6437


csyonghe:
I wonder what will happen if you have an ordinary struct type with a field whose type is extern, and then you query the type layout of the outer struct type.

Another thing is if I lookup and get type layout first from a single IModile, then lookup and the type layout from a linked program on the same struct with extern field, will I get the up to date layout? iIRC we cache type layout on the type itself inside a dictionary owned by the module or something, and if so we may return the wrong result.

csyonghe (approved):

PR #6451, opened 2025-02-25
Improve entry point lookup function documentation

aleino-nv:
- ~~Add a unit test for the case where the code does not specify entry points, but they are instead specified via the API~~
- ~~Make the entry point lookup logic check the information in the options~~
- Add user guide documentation that explains which lookup function to use when there is no [shader(...)] attribute on an entry point
- Add a documentation comment in slang.h mentioning the restriction on findEntryPointByName

https://github.com/shader-slang/slang/issues/6452

See also issue #4760.

### DIFF
--- a/docs/user-guide/02-conventional-features.md
+++ b/docs/user-guide/02-conventional-features.md
@@ -655,6 +655,7 @@ In this example, the `vertexMain` shader indicates that it is meant for the `ver
 Rasterization, compute, and ray-tracing pipelines each define their own stages, and new versions of graphics APIs may introduce new stages.
 
 For compatibility with legacy codebases, Slang supports code that leaves off `[shader(...)]` attributes; in these cases application developers must specify the names and stages for their entry points via explicit command-line or API options.
+Such entry points will not be found via `IModule::findEntryPointByName()`. Instead `IModule::findAndCheckEntryPoint()` must be used, and a stage must be specified.
 It is recommended that new codebases always use `[shader(...)]` attributes both to simplify their workflow, and to make code more explicit and "self-documenting."
 
 > #### Note ####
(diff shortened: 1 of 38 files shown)


cheneym2 (commented):
Looks good to me, but someone else should approve

csyonghe:
There is no bug to be fixed here. If the source code did not specify [shader()] attribute, the host side should call `findAndCheckEntryPoint`.

aleino-nv:
> There is no bug to be fixed here. If the source code did not specify [shader()] attribute, the host side should call `findAndCheckEntryPoint`.

@csyonghe IIUC the user guide states that [shader(...)] should not be needed: https://github.com/shader-slang/slang/issues/4760#issuecomment-2681052524

I ran into this because several tests don't specify [shader(...)], and those fail to look up entry points when I migrate away from the deprecated build request API.

csyonghe:
It is documented here: https://shader-slang.org/docs/compilation-api/#query-entry-points

aleino-nv:
At least I see this doc comment for `findAndCheckEntryPoint`:

```
    /// Find and validate an entry point by name, even if the function is
    /// not marked with the `[shader("...")]` attribute.
    virtual SLANG_NO_THROW SlangResult SLANG_MCALL findAndCheckEntryPoint(...)
```

The `findEntryPointByName` function should probably also have a doc comment that mentions this, and the user guide should probably direct the user to `findAndCheckEntryPoint`.

I will change the patch accordingly.

aleino-nv:
@csyonghe I added a mention in the user guide and slang.h as well. I think it's worth mentioning there since it would have saved me some time...

csyonghe (dismissed):

csyonghe:
looks good! needs a formatting fix.

csyonghe (approved):

PR #6455, opened 2025-02-25
Fix precompiledTargetModule tests

cheneym2:
Add SPIRV-Tool linker support to gfx unit tests
and use the linker in precompileModule tests
that use precompiled modules to reconstitute
SPIRV shaders that were modularly compiled.

Fix a Slang reference count bug in the
precompile service.


tools/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ if(SLANG_ENABLE_GFX)
             $<$<BOOL:${SLANG_ENABLE_XLIB}>:X11::X11>
             $<$<BOOL:${SLANG_ENABLE_CUDA}>:CUDA::cuda_driver>
             $<$<BOOL:${SLANG_ENABLE_NVAPI}>:${NVAPI_LIBRARIES}>
+            SPIRV-Tools-link
```
csyonghe, line 144: we should wrap spirv link in slang-glslang, and load the function dynnamically from the dll so we have clearly separately dependency bottlenecked through slang-glslang.

cheneym2: Done

tools/gfx/d3d12/d3d12-shader-program.cpp
```diff
@@ -10,14 +10,14 @@ using namespace Slang;
 
 Result ShaderProgramImpl::createShaderModule(
     slang::EntryPointReflection* entryPointInfo,
-    ComPtr<ISlangBlob> kernelCode)
+    List<ComPtr<ISlangBlob>> kernelCodes)
```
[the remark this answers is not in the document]
cheneym2: Done

tools/gfx/vulkan/vk-shader-program.cpp
```diff
@@ -69,17 +70,62 @@ VkPipelineShaderStageCreateInfo ShaderProgramImpl::compileEntryPoint(
     return shaderStageCreateInfo;
 }
 
+static ComPtr<ISlangBlob> LinkUsingSPIRVTools(List<ComPtr<ISlangBlob>> kernelCodes)
```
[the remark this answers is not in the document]
cheneym2: Done

csyonghe (approved):

tools/gfx/renderer-shared.cpp
```diff
@@ -1103,32 +1103,106 @@ void ShaderProgramBase::init(const IShaderProgram::Desc& inDesc)
...
+                diagnostics.writeRef());
+            if (diagnostics)
+            {
+                DebugMessageType msgType = DebugMessageType::Warning;
+                if (compileResult != SLANG_OK)
+                    msgType = DebugMessageType::Error;
+                getDebugCallback()->handleMessage(
+                    msgType,
+                    DebugMessageSource::Slang,
+                    (char*)diagnostics->getBufferPointer());
```
kaizhangNV, line 1131: If there is error here, we should either return error,  or not add `downstreamIR` into `kernerlCodes`.

cheneym2: > If there is error here, we should either return error, or not add `downstreamIR` into `kernerlCodes`.
https://github.com/shader-slang/slang/pull/6570/files

PR #6459, opened 2025-02-26
Add Slang-specific intrinsics for integer pack/unpack

fairywreath:
Related to #6175.

Adds Slang-specific integer pack/unpack intrinsics based on the [proposal](https://github.com/shader-slang/spec/blob/main/proposals/018-packed-data-intrinsics.md).

This change is mostly moving the code from the already implemented HLSL pack/unpack intrinsics with code improvements. Floating point intrinsics will be added in a follow-up PR - I am making a separate PR for the integer intrinsics to reduce the size of this follow-up PR.


csyonghe:
According to the proposal, we should just delete int8_t4_packed as a special builtin basic type, and just define it as a typealias to uint.

I now think it wasn’t not a great idea to introduce the type as another basic type, because it will introduce too many additional overloaded functions/operators for no benefit. Let’s remove it from our existing code.

fairywreath:
I've removed the builtin types - the change now looks big but the builtin type removal can be found in this commit 
[remove builtin packed types, use typealias instead](https://github.com/shader-slang/slang/pull/6459/commits/f9330144a2c766932f811525ccffe0e59249062d) [f933014](https://github.com/shader-slang/slang/pull/6459/commits/f9330144a2c766932f811525ccffe0e59249062d). The rest of the changes in this PR are just hlsl.meta.slang changes and test update.

source/slang/hlsl.meta.slang
```diff
@@ -24017,383 +24017,426 @@ T workgroupUniformLoad<T>(__ref T v)
...
+/// clamping each value to the range [-128, 127] to ensure it fits within 8 bits.
+[__readNone]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, shader5_sm_5_0)]
+int8_t4_packed pack_clamp_s8(int16_t4 unpackedValue)
+{
+    return packInt4x8Clamp(unpackedValue);
+}
+
+// Work-graphs
```
csyonghe, line 24143: why is work-graph showing up here?

csyonghe (dismissed):

csyonghe (approved):

PR #6461, opened 2025-02-26
Fix a bug where Session::parseCommandLineArguments returns invalid data

aleino-nv:
This helps to address issue #4760.


source/slang/slang.cpp
```diff
@@ -1108,17 +1108,17 @@ SLANG_NO_THROW SlangResult SLANG_MCALL Session::parseCommandLineArguments(
     if (outDesc->structureSize < sizeof(slang::SessionDesc))
         return SLANG_E_BUFFER_TOO_SMALL;
     RefPtr<ParsedCommandLineData> outData = new ParsedCommandLineData();
-    SerializedOptionsData optionData;
     RefPtr<EndToEndCompileRequest> tempReq = new EndToEndCompileRequest(this);
     tempReq->processCommandLineArguments(argv, argc);
+    outData->options.add({});
+    SerializedOptionsData& optionData = outData->options.getLast();
```
aleino-nv, line 1114: The invalid data was being returned on line 1127:

```
outDesc->compilerOptionEntries = optionData.entries.getBuffer();
```

The issue was that optionData would previously go out of scope, so that the entries buffer would get released.

csyonghe: I don't think the fix here is correct though. When you call `outData->options.add` in line 1119, `outData->options` can resize, making every existing reference to its elements (`optionData`) invalid.

csyonghe (changes requested):
I think we need to call `outData->options->setCount()` to allocate the space before everything, to avoid calling `add` that may invalidate the buffer content in `outData->options`.

source/slang/slang.cpp
aleino-nv: Right! I realized after logging off yesterday... Need to adjust the patch.

csyonghe (approved):

PR #6468, opened 2025-02-26
Map `SV_InstanceID` to `gl_InstanceIndex-gl_BaseInstance`

### LINKED ISSUE #6457: `SV_InstanceID` should translate consistently to SPIRV.
should be emitted as `gl_InstanceIndex - gl_BaseInstance`.

csyonghe:
Closes #6457.

This PR adds a new mechanism to represent builtin GLSL/SPIRV variables in Slang IR, by decorating a global variable with the `IRTargetBuiltinVarDecoration`. This new mechanism is used to represent SV_Instance, gl_InstanceIndex and gl_BaseInstance.

In a follow up PR, we will refactor the existing builtin variable implementation in GLSL/SPIRV legalization to use IRTargetBuiltinVarDecoration instead of relying on the SystemSemanticAttr. The refactor is not done in thie PR to keep the change small and focused.


source/slang/slang-artifact-output-util.cpp
```diff
@@ -6,6 +6,8 @@
 #include "../core/slang-string-util.h"
 #include "../core/slang-type-text-util.h"
 
+#include <chrono>
+
```
jkwak-work, line 10: Is this needed?

jkwak-work (commented):
Looks good to me.

source/slang/slang-artifact-output-util.cpp
csyonghe: yes, this is needed to silence another <chrono> related error in latest vs.

PR #6470, opened 2025-02-26
[WIP] Allow duplication of OpTypeNodePayloadArrayAMDX

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6391

This commit is mainly to allow OpTypeNodePayloadArrayAMDX to be duplicated in SPIRV when using WorkGraphs.

AMD spec requires that there should be dedicated type declaration with OpTypeNodePayloadArrayAMDX for each "node". And each of them should be decorated with PayloadNodeNameAMDX.
https://docs.vulkan.org/features/latest/features/proposals/VK_AMDX_shader_enqueue.html

This behavior is very different from how "types" are declared. Normally types are declared only once and they are shared across the whole program. But for the work-graphs, each "RecordType" belongs to each node and they need to be declared for each node.

Consider the following HLSL example where a producer triggers two consumer nodes. Each node will be differentiated by "NodeID" attribute.

```
[Shader("node")]
[NodeLaunch("broadcasting")]
[NumThreads(4,1,1)]
void myFancyNode(...,
    [NodeID("myNode1")] NodeOutput<MY_RECORD> myRecords1,
    [NodeID("myNode2")] NodeOutput<MY_RECORD> myRecords2,
    ...)
{
    ThreadNodeOutputRecords<MY_RECORD> myRecord1 = myRecords1.GetThreadNodeOutputRecords(1);
    myRecord1.myData = 1;
    myRecord1.OutputCompelete();

    ThreadNodeOutputRecords<MY_RECORD> myRecord2 = myRecords2.GetThreadNodeOutputRecords(1);
    myRecord2.myData = 1;
    myRecord2.OutputCompelete();
}
```

Even though both are using the same type, "MY_RECORD", each of them needs to be declared separately. The generated SPIR-V code for the HLSL above is something like the following,
```
%str_node1 = OpConstantStringAMDX "myNode1";
%str_node2 = OpConstantStringAMDX "myNode1";

%type_node1 = OpTypeNodePayloadArrayAMDX %type_myRecord;
%type_node2 = OpTypeNodePayloadArrayAMDX %type_myRecord;

OpDecorateId %type_node1 PayloadNodeNameAMDX %str_node1;
OpDecorateId %type_node2 PayloadNodeNameAMDX %str_node2;
```
Note that `%type_node1` and `%type_node2` are equivalent, but they must be declared separately because each of them will be decorated with different nodeID string.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -4211,6 +4211,25 @@ attribute_syntax [QuadDerivatives] : QuadDerivativesAttribute;
 __attributeTarget(FuncDecl)
 attribute_syntax [RequireFullQuads] : RequireFullQuadsAttribute;
 
-__generic<T>
-typealias NodePayloadPtr = Ptr<T, $( (uint64_t)AddressSpace::NodePayloadAMDX)>;
+// Work-graphs
+
+/// @internal
+/// __SPIRVNodePayloadArray is to emit OpTypeNodePayloadArrayAMDX and
+/// SpvDecorationPayloadNodeNameAMDX for it.
+///
+__generic<RecordType, let nodeID : int>
+__intrinsic_type($(kIROp_SPIRVNodePayloadArrayType))
+struct __SPIRVNodePayloadArray;
+
+/// @internal
+/// NodePayloadArrayPtr is a pointer type for "NodePayloadAMDX" storage-class and it points to an array of work-graph nodes
+///
+__generic<RecordType, let nodeID : int>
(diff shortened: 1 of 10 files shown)


jkwak-work:
This PR is not ready for review nor merge.

The main problem is that it is not able to take "nodeID" in a String type.
Current implementation takes an int-type and convert it to "NodeID_%d" during the emitting, which is a hack.

I need to get the nodeID from a new attribute `[NodeID("name")]` to make it a proper PR.

csyonghe:
I am closing this PR due to inactivity, assuming it is no longer being worked on.

PR #6476, opened 2025-02-27
Don't add multiple matrix layout options when creating a session

aleino-nv:
The default matrix layout mode was applied in addition to any related options, and this caused the wrong matrix layout mode to be used.
For example, tests/compute/column-major was failing when attempting to migrate to the new compilation API.

This helps to address issue #4760


source/slang/slang-compiler.h
```diff
@@ -27,6 +27,8 @@
 #include "slang-syntax.h"
 #include "slang.h"
 
+#include <chrono>
```
csyonghe, line 30: this is not needed anymore.

aleino-nv: Done, I removed it.

csyonghe (approved):

PR #6478, opened 2025-02-27
Bump glslang and deps

expipiplus1:
supersedes https://github.com/shader-slang/slang/pull/6477

This uses the latest public versions, rather than the latest tag


mkeshavaNV:
LGTM, approved.

aleino-nv (dismissed):

aleino-nv (dismissed):

expipiplus1:
merging because the only failure is also present on the target branch

PR #6487, opened 2025-02-27
Allow partial specialization of existential arguments.

### LINKED ISSUE #6482: Strange compilation bug
This is a reduced version of my code but for some reason it does not compile and gives me the following error. Not exactly sure what the problem is but it seems to come from the `while (rayQuery.Proceed()) {` loop.
eg removing `float rand = rng.next_float();` seems to let it compile

```
(20): error 41011: type 'Scene' does not fit in the size required by its conforming interface.
struct Scene : IScene {
       ^~~~~
(20): note 41012: sizeof(Scene) is 0, limit is 0
```

```csharp

public interface IRandom {
    [mutating] uint32_t next_uint();
    [mutating] float next_float();
};

public struct TEA : IRandom {
    public __init(uint32_t v0, uint32_t v1) {}
    [mutating] public uint32_t next_uint() {
        return 0;
    }
    [mutating] public float next_float() {
        return 0;
    }
};

public interface IScene {
    property RaytracingAccelerationStructure as;
};
struct Scene : IScene {
    RaytracingAccelerationStructure as;
};

public interface IIntegrator {
    public float3 sample(IScene scene, int * sensor, RayDesc ray, IRandom rng);
};
namespace integrator {
    public struct NoShading : IIntegrator {
        public float3 sample(IScene scene, int *sensor, RayDesc _ray, IRandom rng) {
            return float3( 0.0f, 0.0f, 0.0f );
        }
    };

    struct Test {

        float4 sample(IScene scene, int *sensor, RayDesc _ray, IRandom rng, int pixel_aabb_uv, uint3 id) {
            float4 grad = { 0.0f, 0.0f, 0.0f, 1.0f };

            RayDesc ray = _ray; uint32_t depth = 0;
            RayQuery<0> rayQuery;
            while (rayQuery.Proceed()) {
                float rand = rng.next_float();

                IIntegrator integrator = integrator::NoShading();
                let color_in = integrator.sample(scene, nullptr, {}, rng);
                let color_out = integrator.sample(scene, nullptr, {}, rng);
                
            }
            return grad;
        }
    };
};

[[vk::binding(0, 0)]] RWTexture2D<float4> output;

[vk::constant_id(0)] const int WGS_X = 1;
[vk::constant_id(1)] const int WGS_Y = 1;
[shader("compute"), numthreads(WGS_X, WGS_Y, 1)]
void main(
    uint3 id : SV_DispatchThreadID
)
{
    IRandom rng = TEA(id.y * id.x, 1);
    Scene scene = { RaytracingAccelerationStructure(0) };

    integrator::Test integrator = integrator::Test();
    float4 grad = integrator.sample(scene, nullptr, {}, rng, {}, id);
    output[id.xy] += float4(grad.xyz, grad.w);
}
```

csyonghe:
Closes #6482.


tests/bugs/gh-6482-interface-method-existential-specialize.slang
```diff
@@ -0,0 +1,93 @@
+//TEST:SIMPLE(filecheck=CHECK): -target spirv
+
+// This is a test that checks that we can apply partial specialization to a function
+// that takes existential parameters.
+
+// CEHCK: OpRayQueryProceedKHR
```
cheneym2, line 6: CHECK

csyonghe: Fixed. Thank you for catching this.

cheneym2 (approved):

source/slang/slang-ir-layout.cpp
```diff
@@ -376,7 +376,9 @@ static Result _calcSizeAndAlignment(
     default:
         break;
     }
-    if (as<IRResourceTypeBase>(type) || as<IRSamplerStateTypeBase>(type))
+
+    if (as<IRResourceTypeBase>(type) || as<IRSamplerStateTypeBase>(type) ||
+        type->getOp() == kIROp_RaytracingAccelerationStructureType)
```
Devon7925, line 381: Why is this check here rather than in the switch statement?

csyonghe: Because there are many opcodes that are treated as subclass of ResourceTypeBase, and we do not want to duplicate them all in the switch. For example, if we add a new resource type, we don’t want to modify all switches in the code base and add a case for it.

Devon7925: I meant specifically the `kIROp_RaytracingAccelerationStructureType` check. Feels weird for it to be grouped into this set of checks instead of handled with the other opcodes.

csyonghe: Ah, I see what you mean. I put it here because it logically groups together with other resource types, but putting it in the same cases with the pointer types should also work. I don't have a strong preference here, so I can move it up.

csyonghe: done.

PR #6500, opened 2025-02-28
Support SPIR-V deferred linking option

### LINKED ISSUE #4994: SPIR-V linkage support modules with precompiled blobs
When slang-modules are linked which contain precompiled spirv, strip the precompiled IR from the linked IR prior to emitting SPIR-V. Link the modules with spirv-link.

cheneym2:
The new option "SkipDownstreamLinking" will defer final downstream IR linking to the user application. This option only has an effect if there are modules that were precompiled to the target IR using precompileForTarget().

Until now, the default behavior for SPIR-V was to use deferred linking, and the default behavior for DXIL was to use immediate/internal linking in Slang.

This change only affects the SPIR-V behavior such that both deferred and non-deferred linking is supported based on the new option.

To support the non-deferred option, Slang will internally call into SPIRV-Tools-link to reconstitute a complete SPIR-V shader program when necessary (due to modules having been precompiled to target IR). Otherwise, if SkipDownstreamLinking is enabled, the shader returned by e.g. getTargetCode() or getEntryPointCode() may have import linkage to the SPIR-V embedded in the constituent modules.

Closes #4994

### DIFF
--- a/tools/gfx-unit-test/precompiled-module-2.cpp
+++ b/tools/gfx-unit-test/precompiled-module-2.cpp
@@ -17,7 +17,7 @@ static Slang::Result precompileProgram(
     gfx::IDevice* device,
     ISlangMutableFileSystem* fileSys,
     const char* shaderModuleName,
-    bool precompileToTarget)
+    PrecompilationMode precompilationMode)
 {
     Slang::ComPtr<slang::ISession> slangSession;
     SLANG_RETURN_ON_FAIL(device->getSlangSession(slangSession.writeRef()));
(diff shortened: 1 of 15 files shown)


tools/gfx-unit-test/precompiled-module-2.cpp
```diff
@@ -121,6 +122,17 @@ void precompiledModule2TestImplCommon(
     }
     sessionDesc.targets = &targetDesc;
     sessionDesc.fileSystem = memoryFileSystem.get();
+
+    std::vector<slang::CompilerOptionEntry> options;
```
csyonghe, line 126: List.

csyonghe (approved):

PR #6506, opened 2025-03-03
Fix codegen bug when targeting PTX with new API

aleino-nv:
- add a repro of the problem
  - This is added as a unit test, because the new API is not yet in use for the tests in tests/ 
  - This unit test can be removed when the new API is used for tests/
- Simple fix that detects entry points more robustly when applying some "ConstRef hack"
  - When using the old API, an explicit EntryPointAttribute was being attached to the entry point, even though none was present in the code

See issue #6507 for debugging notes.

This closes issue #6507.

### DIFF
--- a/tests/expected-failure-github-runner.txt
+++ b/tests/expected-failure-github-runner.txt
@@ -1,0 +1,1 @@
+slang-unit-test-tool/cudaCodeGenBug.internal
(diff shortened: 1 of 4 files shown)


tests/expected-failure-github-runner.txt
cheneym2, line 1: I'm curious. What is the "runner" test list meant for, and why couldn't expected-failure-github.txt be used?

aleino-nv: Our GitHub runners don't have the CUDA toolkits installed, so they can't run all tests.
The unit test I add targets PTX and so needs the CUDA toolkit.

Ideally this should be detected at runtime, but I don't think we can readily do that.
The test should be short-lived, so I just added the runner list as a means to filter.

PR #6510, opened 2025-03-03
Fix on missing warning on some cases of integer literal coercion #5728

### LINKED ISSUE #5728: Missing warning on some cases of integer literal coercion
These properly give warnings:
```
    int x1 = 0xffffffffffff;
    int x2 = -9223372036854775807;
```
```
1.slang(36): warning 30081: implicit conversion from 'uint64_t' to 'int' is not recommended
    int x1 = 0xffffffffffff;
             ^~~~~~~~~~~~~~~
1.slang(38): warning 30081: implicit conversion from 'int64_t' to 'int' is not recommended
    int x2 = -9223372036854775807;
              ^~~~~~~~~~~~~~~~~~~
```
But these do not(note they are assigned to UINT64_MAX)
```
    int x3 = 18446744073709551615;
    int x4 = 0xFFFFFFFFFFFFFFFF;
```
This needs to be investigated and fixed.

zlatinski:
Fix missing warnings on:

int x3 = 18446744073709551615;
int x4 = 0xFFFFFFFFFFFFFFFF;

fixes #5728

### DIFF
--- a/source/slang/slang-parser.cpp
+++ b/source/slang/slang-parser.cpp
@@ -6897,7 +6897,7 @@ static IntegerLiteralValue _fixIntegerLiteral(
         // If the masked value is 0 or equal to the mask, we 'assume' no information is
         // lost
         // This allows for example -1u, to give 0xffffffff
-        // It also means 0xfffffffffffffffffu will give 0xffffffff, without a warning.
+        // It also means 0xffffffffffffffffu will give 0xffffffff, without a warning.
         if ((!(maskedValue == 0 || maskedValue == mask)) && sink && token)
         {
             // Output a warning that number has been altered
@@ -6954,8 +6954,11 @@ static BaseType _determineNonSuffixedIntegerLiteralType(
     {
         baseType = BaseType::UInt64;
 
-        if (isDecimalBase)
+        // Emit warning if the value is too large for signed 64-bit, regardless of base
+        // This fixes the inconsistency between decimal and hex literals
+        if (sink && token)
         {
+
             // There is an edge case here where 9223372036854775808 or INT64_MAX + 1
             // brings us here, but the complete literal is -9223372036854775808 or INT64_MIN and is
             // valid. Unfortunately because the lexer handles the negative(-) part of the literal
--- a/tests/diagnostics/integer-literal-warnings.slang
+++ b/tests/diagnostics/integer-literal-warnings.slang
@@ -1,0 +1,24 @@
+//TEST:SIMPLE(filecheck=CHECK): -target dxbc-assembly
+
+// Test that UINT64_MAX literals produce appropriate warnings
+// This test verifies the fix from commit 0a3c1c8dbd6aa9501ea94a8de69d1a5967034b2c
+// which ensures literals equivalent to UINT64_MAX produce appropriate warnings
+
+void main()
+{
+    // Decimal form of UINT64_MAX
+    // CHECK: warning 39999: integer literal is too large to be represented in a signed integer type, interpreting as unsigned
+    uint64_t a = 18446744073709551615;
+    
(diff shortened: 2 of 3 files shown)


csyonghe (commented):
Can you add a test?

I am not sure I understand what is being fixed here. We don’t want a warning for int64_t v = -9223372036854775808 and I am not seeing how the patch fixes that.

zlatinski:
It is fixing the issues reported:
 int x3 = 18446744073709551615;
 int x4 = 0xFFFFFFFFFFFFFFFF;
  
>  We don’t want a warning for int64_t v = -9223372036854775808 and I am not seeing how the patch fixes that.  
Are you saying that we have an issue with the above assignment? Let me check ...

csyonghe:
I am just seeing the comment being deleted is discussing this case, so it is best to verify things are working as expected with tests.

zlatinski:
I have the test added for the integer literal coercion.

source/slang/slang-parser.cpp
```diff
@@ -6954,16 +6954,10 @@ static BaseType _determineNonSuffixedIntegerLiteralType(
...
+        // This fixes the inconsistency between decimal and hex literals
+        if (sink && token)
         {
-            // There is an edge case here where 9223372036854775808 or INT64_MAX + 1
-            // brings us here, but the complete literal is -9223372036854775808 or INT64_MIN and is
-            // valid. Unfortunately because the lexer handles the negative(-) part of the literal
-            // separately it is impossible to know whether the literal has a negative sign or not.
-            // We emit the warning and initially process it as a uint64 anyways, and the negative
-            // sign will be properly parsed and the value will still be properly stored as a
-            // negative INT64_MIN.
```
jkwak-work, line 6965: Please do not remove this comment, because the concern hasn't been addressed yet.
When I tested, the value, "-9223372036854775808", still prints an incorrect warning.
It is a valid value representable with unsigned 64bits.

The code I tests is following,
```
    int64_t i1 = -9223372036854775807;
    int64_t i2 = -9223372036854775808;
```
i2 printed a warning and i1 didn't.
Both were expected not to print a warning.
```
tests/compute/frem.slang(17): warning 39999: integer literal is too large to be represented in a signed integer type, interpreting as unsigned
    int64_t i2 = -9223372036854775808;
                  ^~~~~~~~~~~~~~~~~~~
```

jkwak-work (approved):
Change looks good to me.

source/slang/slang-parser.cpp
zlatinski: > Please do not remove this comment, because the concern hasn't been addressed yet.

Good point, Jay! Thank you for catching that! 
I have restored the comment.

csyonghe:
The fundamental problem of this issue is that we are trying to determine the type of a literal or issue a warning too early.
We should wait until the semantic checking stage to issue this kind of warnings.

When we parse the int literal, we just store the token into the AST, and parse the int into a uint64_t field. Let's not determine the type of the literal expr yet.

Then when we are checking the expr in SemanticExprVisitor, we can always fold the `OperatorExpr(IntLiteralExpr)` into a single `IntLiteralExpr`, and then do the checking on the `IntLiteralExpr` to see if it is out of bounds.

zlatinski:
< When we parse the int literal, we just store the token into the AST, and parse the int into a uint64_t field. Let's not determine the type of the literal expr yet.

< Then when we are checking the expr in SemanticExprVisitor, we can always fold the OperatorExpr(IntLiteralExpr) into a single IntLiteralExpr, and then do the checking on the IntLiteralExpr to see if it is out of bounds.

Thank you, @csyonghe, for the feedback. However, I'm not sure I'm qualified to make the proposed changes.

bmillsNV:
Closing. no follow up from dev

PR #6511, opened 2025-03-03
Update SPIRV-Tools and fix new validation errors.

### LINKED ISSUE #6508: Invalid SPIR-V Explicit Layout on texture descriptor indexing
(Slangc 2025.6.1 with latest SPIRV-Tools)

```
[[vk::binding(0, 0)]]
Texture2D texture[];
[[vk::binding(1, 0)]]
SamplerState sampler;

struct StorageBuffer {
    uint data;  // Will be one
    float4 color;
};

[[vk::binding(2, 0)]]
RWStructuredBuffer<StorageBuffer> storageBuffer;

[shader("compute")]
void main() {
    uint dataIndex = storageBuffer[0].data;
    float4 sampledColor = texture[dataIndex].Sample(sampler, float2(0));
    storageBuffer[0].color = sampledColor;
}
```

ran with `slangc in.spv -o out.spv` will produce invalid SPIR-V (https://godbolt.org/z/zjsPMPfTx) 

> Invalid explicit layout decorations on type for operand '23[%_ptr_UniformConstant__runtimearr_22]'
> %texture = OpVariable %_ptr_UniformConstant__runtimearr_22 UniformConstant

I think this occurs starting with https://github.com/KhronosGroup/SPIRV-Tools/pull/6012

csyonghe:
Closes #6508.

## Issue

Recent SPIRV-Tools update adds a new verification rule that disallows explicit layouts (e.g. `Offset` decoration) on types used from Function/Private/Workgroup etc. address spaces after SPIRV 1.4.

## Background

SPIRV makes a distinction of physical and logical types. A physical type is a type with explicit memory layout and can be stored in a user-transparent way in memory. A logical type in contrast does not have a defined layout, and cannot be decorated with any explicit layout deocrations, such as `Offset`, `ArrayStride` etc.

The SPIRV spec requires that all buffer elements or global input/outputs to be a physical type, and all variables in the function/private(i.e. ordinary global vars)/workgroup address spaces to be logical.

This means that if the user writes:
```
struct Foo { int x, y; }

StructuredBuffer<Foo> f;  // or ConstantBuffer, etc.

void main()
{
     Foo localVar;
     localVar = f[0];
     f[1] = localVar;
}
```

We need to represent `Foo` as two distinct types when generating SPIRV: the one used inside `StructuredBuffer` must be a physical type with explicit layout (specifying the offset of `Foo::x` and `Foo::y`), and the one used for `localVar` must be a logical type. Therefore we generate something like:

```
struct Foo { int x, y; }
struct Foo_std430 { [Offset(0)] int x;  [Offset(4)] int y; }
// If `Foo` is used in a `ConstantBuffer`, we would generate a `Foo_std140` type too.

StructuredBuffer<Foo_std430> f;  // or ConstantBuffer, etc.

Foo_std430 packStorage(Foo f) { ... }
Foo unpackStorage(Foo_Std430 f) {...}

void main()
{
     Foo localVar;
     localVar = unpackStorage(f[0]);
     f[1] = packStorage(localVar);
}
```

Where `Foo_std430` is the physical type for `Foo` that can be placed in a buffer, and `packStorage` and `unpackStorage` functions to convert between physical and logical type when needed.

The above translation is what the existing `lowerBufferElementTypeToStorageType` is doing.

Currently, when we emit SPIRV, we always emit `Offset` decorations for both `Foo` and `Foo_std430`. However this is against the SPIRV rules, which disallows a type to be declared as a local variable, if the type has explicit layout.

## Solution

To fix the SPIRV validation error complaining `Foo` cannot be used to declare `localVar` because it has explicit layout, we simply need to mark the lowered storage types (`Foo_std430` in the above example) with a `[PhysicalType]` decoration, and make sure we only emit SPIRV layout decorations (ArrayStride, MatrixStride, ColMajor, RowMajor, Offset) for structs with the `[PhyiscalType]` decoration.

The rest of the PR is to deal with a bunch of consequences after changing the way we emit SPIRV explicit layout based on whether or not a type is marked `[PhysicalType]`.

- `GLSLStorageBuffer` type weren't lowered by `lowerBufferElementTypeToStorageType` previously, but now needs to be in order to produce valid SPIRV so that the element type is properly marked as physical.
- This means that `lowerBufferElementTypeToStorageType` needs to generate pack/unpack functions for unsized array types that can appear in glsl storage buffers, which we weren't handling previously.
- Array types without explicit stride are now treated as logical type and won't result in a stride being decorated in the resulting SPIRV. We need to be careful to never declare a local variable whose type is an array with explicit stride operand -- that will lead to invalid SPIRV. This leads to a change in the next bullet.
- packStorage/unpackStorage functions can no longer operate on SSA values, and has to operate on memory pointers instead, because we are no longer to `elementExtract` from a SSA register of a physical array type (can't copy it to a OpVariable and then do AccessChain, because OpVariable must be a logical type). These functions are inlined before emitting final code, because SPIRV doesn't allow us to pass pointers to functions without using VariablePointers capability. Because we no longer need to lower `ElementExtract(array, i)` to first copy to local var then do access chain, the resulting code should be much simpler for pack/unpack of large arrays and should be faster to compile by the downstream compiler.
- When packStorage/unpackStorage operate on pointers, we are effectively deferring the load of a composite to load of separate leaf fields, this means that `loadAlign(ptr)` is going to be split into a lot of smaller loads, so we must propagate the alignment on the load through the addresses, so that each individual load/store gets the correct alignment.
- Reimplement ConstBufferType as a wrapper of native pointers instead of a special builtin type, because we no longer want to maintain that code that special cases for ConstBufferTypes with this change. This means we need to:
- Implemented pointers for GLSL target. The only reason for ConstBufferPointer to exist as a builtin was that it provides some minimal coverage of buffer_reference when targeting GLSL. With pointers supported for GLSL, there is no reason to keep it as a compiler-builtin.

The updated `lowerBufferElementTypeToStorageType` pass now translates the above example code to the following:
```
struct Foo { int x, y; }
[PhysicalType] struct Foo_std430 { [Offset(0)] int x;  [Offset(4)] int y; }

StructuredBuffer<Foo_std430> f;  // or ConstantBuffer, etc.

// Pack/unpack function are now inlined by Slang, because they
// need to work on references instead of values.
[ForceInline] void packStorage(Foo_std430* dest, Foo f) { ... }
[ForceInline] Foo unpackStorage(Foo_Std430* src) {...}

void main()
{
     Foo localVar;

     // unpack function takes in a pointer to the source value, instead of an SSA value.
     localVar = unpackStorage(&f[0]);

     // destination driven packing function that writes directly into a pointer,
     // instead of returning an SSA value.
     packStorage(&f[1], localVar); 
}
```

## Copilot Summary
This pull request includes significant changes to the handling of pointer types and various optimizations in the SPIR-V and GLSL code generation. The most important changes include the refactoring of `ConstBufferPointerType`, updates to GLSL source emission, and improvements to SPIR-V emission and optimization.

### Refactoring and Simplification:

* [`source/slang/hlsl.meta.slang`](diffhunk://#diff-645e6e8cf2915cbda6d0a72af868d90c90a6da0b2a277e2020dae14d2b7fd5b3L21612-R21641): Refactored `ConstBufferPointer` to simplify its structure and improve inline functions.
* [`source/slang/slang-ast-type.h`](diffhunk://#diff-fb7bbdb9f44cc4db32d9f9951580a9adf1ce3088f1815a444c8e9848028bd213L635-L641): Removed the `ConstBufferPointerType` class to streamline the pointer type definitions.

### GLSL Source Emission:

* [`source/slang/slang-emit-glsl.cpp`](diffhunk://#diff-8f5119455d06dc493f194357a9c49f97190b825a64e9eda5880ebd66c083402cL2039-R2062): Updated various methods to handle the new pointer type structure, including `emitBufferPointerTypeDefinition`, `tryEmitInstExprImpl`, and `emitSimpleTypeImpl`. [[1]](diffhunk://#diff-8f5119455d06dc493f194357a9c49f97190b825a64e9eda5880ebd66c083402cL2039-R2062) [[2]](diffhunk://#diff-8f5119455d06dc493f194357a9c49f97190b825a64e9eda5880ebd66c083402cR2114-R2150) [[3]](diffhunk://#diff-8f5119455d06dc493f194357a9c49f97190b825a64e9eda5880ebd66c083402cL3207-L3210)

### SPIR-V Emission and Optimization:
* [`source/slang/slang-emit-spirv.cpp`](diffhunk://#diff-6b376cb3b3756ec6837e1486125ec8c9a478b6f47c3b387787745fc53bc4c868R1440-R1453): Improved handling of forward-declared pointers and added checks to emit array stride and offset decoration only for types marked as physical. [[1]](diffhunk://#diff-6b376cb3b3756ec6837e1486125ec8c9a478b6f47c3b387787745fc53bc4c868R1440-R1453) [[2]](diffhunk://#diff-6b376cb3b3756ec6837e1486125ec8c9a478b6f47c3b387787745fc53bc4c868L1657-L1692)
* [`source/slang/slang-emit.cpp`](diffhunk://#diff-646f42e4400aedca89782098c7b26aba8772b08bd0f588305b857ac565fba0a8R1647-R1651): Added force inlining and other optimization steps after `lowerBufferElementTypeToStorageType` pass. [[1]](diffhunk://#diff-646f42e4400aedca89782098c7b26aba8772b08bd0f588305b857ac565fba0a8R1647-R1651) [[2]](diffhunk://#diff-646f42e4400aedca89782098c7b26aba8772b08bd0f588305b857ac565fba0a8L1672-R1674)

These changes collectively enhance the handling of pointer types and improve the efficiency and correctness of the generated GLSL and SPIR-V code.


source/slang/slang-emit-glsl.cpp
```diff
@@ -2036,21 +2036,30 @@ bool GLSLSourceEmitter::_tryEmitBitBinOp(
...
-    auto constPtrType = as<IRHLSLConstBufferPointerType>(ptrType);
     auto ptrTypeName = getName(ptrType);
-    auto alignment = getIntVal(constPtrType->getBaseAlignment());
+    IRSizeAndAlignment sizeAlignment;
+    getNaturalSizeAndAlignment(
+        m_codeGenContext->getTargetProgram()->getOptionSet(),
+        ptrType->getValueType(),
+        &sizeAlignment);
+    auto alignment = sizeAlignment.alignment;
     m_writer->emit("layout(buffer_reference, std430, buffer_reference_align = ");
```
kaizhangNV, line 2055: I recall in the commit description, it says if the buffer is a constant buffer, we will use "std_140"?

source/slang/slang-emit-glsl.cpp
```diff
@@ -2372,10 +2418,57 @@ bool GLSLSourceEmitter::tryEmitInstExprImpl(IRInst* inst, const EmitOpInfo& inOu
...
+
+                maybeCloseParens(needClose);
+                return true;
+            }
             // Use the default
             break;
         }
+    case kIROp_GetOffsetPtr:
+        {
+            _requireGLSLExtension(UnownedStringSlice("GL_EXT_buffer_reference2"));
```
kaizhangNV, line 2469: wondering why emit nothing here?

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1850,6 +1853,127 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                            auto base = inst->getOperand(0);
+                            auto ptrType = as<IRPtrTypeBase>(base->getDataType());
+                            if (!ptrType)
+                                break;
+                            // Propagate address alignment if possible.
+                            auto alignDecor = base->findDecoration<IRAlignedAddressDecoration>();
+                            if (!alignDecor)
+                                break;
+                            auto valueType = ptrType->getValueType();
+                            auto layout = valueType->findDecoration<IRSizeAndAlignmentDecoration>();
```
kaizhangNV, line 1883: nullptr check on layout?

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1850,6 +1853,127 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+                                IRBuilder builder(inst);
+                                builder.addAlignedAddressDecoration(
+                                    inst,
+                                    alignDecor->getAlignment());
+                            }
+                        }
+                        break;
+                    case kIROp_FieldAddress:
+                        {
+                            auto base = inst->getOperand(0);
```
kaizhangNV, line 1896: Looks like all those code can be reused.

tests/compute/byte-address-buffer-array.slang
```diff
@@ -43,35 +30,9 @@ void computeMain(uint3 threadId : SV_DispatchThreadID)
...
-    // CHECK3-DAG: %[[V14:[a-zA-Z0-9_]+]] = OpAccessChain %[[SBF]] %buffer_0 {{.*}}
-    // CHECK3-DAG: %[[V15:[a-zA-Z0-9_]+]] = OpLoad %float %[[V14]]
-    // CHECK3-DAG: %[[V16:[a-zA-Z0-9_]+]] = OpCompositeConstruct %v4float %[[V9]] %[[V11]] %[[V13]] %[[V15]]
-    // CHECK3-DAG: %[[V17:[a-zA-Z0-9_]+]] = OpAccessChain %[[SBVF]] %buffer_1 {{.*}}
-    // CHECK3-DAG: %[[V18:[a-zA-Z0-9_]+]] = OpLoad %v4float %[[V17]]
-    // CHECK3-DAG: %[[V19:[a-zA-Z0-9_]+]] = OpAccessChain %[[SBVF]] %buffer_1 {{.*}}
-    // CHECK3-DAG: OpStore %[[V19]] %[[V16]]
-    // CHECK3-DAG: %[[V20:[a-zA-Z0-9_]+]] = OpAccessChain %[[SBVF]] %buffer_1 {{.*}}
-    // CHECK3-DAG: OpStore %[[V20]] %[[V18]]
+    // CHECK3: OpAccessChain
```
kaizhangNV, line 33: Is there no alignment decor in the load inst?

tests/hlsl-intrinsic/const-buffer-pointer.slang
```diff
@@ -25,12 +22,9 @@ struct MyStruct
 // CHECK-NEXT: BufferPointer_MyStruct{{.*}} pStruct
 // CHECK-NEXT: }
 
-// CHECK: void main
-// CHECK: MyStruct{{.*}} s{{.*}} = ((gGlobals{{.*}}.pStruct{{.*}})._data);
-
 struct Globals
 {
-    ConstBufferPointer<MyStruct> pStruct;
+    MyStruct* pStruct;
```
kaizhangNV, line 27: why is this change?

source/slang/slang-emit-glsl.cpp
csyonghe: here we are translating a pointer type `e.g. T*` to glsl `buffer_reference`. In GLSL, buffer references are always std430.

source/slang/slang-emit-glsl.cpp
csyonghe: `return false` falls back to the default emit logic implemented in the base `CLikeEmitter` class, which will emit the pointer add logic there. Here we just want to ensure the right extension is being declared.

tests/compute/byte-address-buffer-array.slang
csyonghe: For this specific test, no, we simply make sure we are generating typed loads instead of loading individual elements of a vector.

tests/hlsl-intrinsic/const-buffer-pointer.slang
csyonghe: This shouldn't be needed. I was at one point thinking we just delete `ConstBufferPointer` type, but then I changed my mind to keep it and instead reimplement it as a wrapper.

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: Fixed.

source/slang/slang-ir-spirv-legalize.cpp
csyonghe: Merged the two cases to reuse code.

jkwak-work (approved):
Looks good to me.

We were talking about the logical storage and physical storage in WorkGraphs meeting today coincidently.
I guess this is a thing going around recently.

PR #6514, opened 2025-03-04
Migrate render-test away from deprecated compile request API

aleino-nv:
- Add a simple test to more easily debug specialization parameter issues
- Do some required fixes for existing tests
- Replace code using compile requests with non-deprecated API calls

This closes issue #4760.


tools/render-test/slang-support.cpp
```diff
@@ -48,136 +50,307 @@ static SlangResult _compileProgramImpl(
...
+        sessionTargetDesc.flags |= SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY;
     else
+        sessionTargetDesc.flags = 0;
+
+    // Not expecting argument parsing to have added any targets
+    SLANG_ASSERT(sessionDesc.targetCount == 0);
+    sessionDesc.targetCount = 1;
+    sessionDesc.targets = &sessionTargetDesc;
+
+    if (options.generateSPIRVDirectly)
```
cheneym2, line 202: Do you know why debug info depends on generateSPIRVDirectly?  I see that this behavior is maintained from the previous code, but it seems odd.

cheneym2 (approved):

tools/render-test/slang-support.cpp
aleino-nv: I agree it seems strange, but I didn't find out why. 
I figured it's best to do the migration as strictly as possible, and then possibly it should get cleaned up later when/if it's ever a problem.

PR #6527, opened 2025-03-05
Fix a bug in default ctor synthesizing

kaizhangNV:
This is fix for the implementation bug, when a struct has explict ctor we should not synthesize the default ctor anymore.

### DIFF
--- a/source/slang/slang-check-conversion.cpp
+++ b/source/slang/slang-check-conversion.cpp
@@ -408,7 +408,12 @@ bool SemanticsVisitor::createInvokeExprForSynthesizedCtor(
         // When a struct has no constructor and it's not a C-style type, the initializer list is
         // invalid.
         isCStyle = isCStyleType(toType, isVisit);
-        if (!isCStyle)
+
+        // WAR: We currently still has to allow legacy initializer list for array type until we have
+        // more proper solution for array initialization, so if the right hand side is an array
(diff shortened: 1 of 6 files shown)


csyonghe (approved):

source/slang/slang-check-conversion.cpp
```diff
@@ -396,14 +396,35 @@ bool SemanticsVisitor::createInvokeExprForSynthesizedCtor(
...
+    {
+        // When a struct has no constructor and it's not a C-style type, the initializer list is
+        // invalid.
+        isCStyle = isCStyleType(toType, isVisit);
+
+        // WAR: We currently still has to allow legacy initializer list for array type until we have
+        // more proper solution for array initialization, so if the right hand side is an array
+        // type, we will not report error and fall-back to legacy initializer list logic.
+        bool isArrayType = as<ArrayExpressionType>(toType) != nullptr;
+        if (!isCStyle && !isArrayType)
```
csyonghe, line 416: I think the ordering of steps that needs to happen is:
Try match to a ctor call, if succeeded done.
If array type and empty init list, try construct a single value of the element type with empty init list, if succeed, create a MakeArrayFromSingleValue(singleValue) node for the array value, done.
If that failed, fallback to cstyle logic if applicable.
If that failed, error.

PR #6528, opened 2025-03-05
Update build to allow setting external paths

jarcherNV:
Update the build to allow setting user-specific paths for the external modules. This allows building Slang without also fetching the external modules, assuming they are already present elsewhere locally.


CMakeLists.txt
```diff
@@ -166,6 +166,43 @@ option(
 )
 mark_as_advanced(SLANG_SPIRV_HEADERS_INCLUDE_DIR)
 
+# Options for user defined paths for external modules.
+advanced_option(
+    SLANG_LZ4_PATH
```
jkwak-work, line 171: The naming pattern for CMake options are like SLANG_VERBE_NOUN.
So I like to suggest some thing like
SLANG_OVERRIDE_PATH_LZ4

jarcherNV: Fixed.

jkwak-work (dismissed):
Looks good to me

jkwak-work (approved):
Reapproving after merging with ToT

PR #6536, opened 2025-03-07
Fix SPIRV Generator ID.

csyonghe:
The SPIRV Generator ID is actually in the high 16 bits.


cheneym2:
I checked out the spec and I agree. https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/spir-v.xml#L35

cheneym2 (approved):

jkwak-work (approved):
Looks good to me

PR #6537, opened 2025-03-07
Add option to exclude webgpu-dawn and slang-tint

jarcherNV:
Add option to exclude webgpu-dawn and slang-tint from the build, for environments that need to build Slang without internet access.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,7 +199,19 @@ advanced_option(
 )
 advanced_option(
     SLANG_OVERRIDE_GLSLANG_PATH
-    "Build using user defined path for glslang, this also requires "
+    "Build using user defined path for glslang"
+    OFF
+)
+
+advanced_option(
+    SLANG_EXCLUDE_DAWN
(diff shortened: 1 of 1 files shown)


jkwak-work (dismissed):
It looks good to me.

@aleino-nv may want to take a look?

aleino-nv:
I think the more canonical way of doing this is to have two CMake options (one for dawn and one for tint), and set the defaults to indicate that we want to get these packages.

There are some examples of this in the same CMake file.

jarcherNV:
Second revision splits this into 2 separate options as suggested by aleino.

jkwak-work (approved):
It looks good enough.

But the naming pattern we have is SLANG_USE_SYSTRM_XXX, isn't it?

samestep:
Possibly of note: right now I'm attempting to build Slang with Nix which prevents internet access, and enabling this new `SLANG_EXCLUDE_DAWN` flag did not work for me; instead, I had to disable the `SLANG_RHI_ENABLE_WGPU` flag added by shader-slang/slang-rhi#48.

PR #6543, opened 2025-03-07
Make make_struct call consistent

### LINKED ISSUE #6541: Invalid code emit for Metal
```
struct nestStruct
{
    RWStructuredBuffer<int> buffer;
    int strides;
    int shape;
}

struct MyStruct
{
    nestStruct a;
}

ParameterBlock<MyStruct> param;

[shader("compute")]
[numthreads(1, 1, 1)]
void computeMain(
    uint tid: SV_DispatchThreadID,
)
{
    param.a.buffer[0] = param.a.strides;
    param.a.buffer[1] = param.a.shape;
}
```

For this code, slang will emit invalid Metal shader:
```
struct nestStruct_0
{
    int device* buffer_0;
    int strides_0;
    int shape_0;
};

struct MyStruct_0
{
    nestStruct_0 a_0;
};


#line 17
[[kernel]] void computeMain(uint3 tid_0 [[thread_position_in_grid]], MyStruct_0 constant* param_0 [[buffer(0)]])
{

#line 18
    nestStruct_0 _S1 = param_0->a_0;


    *((&param_0->a_0)->buffer_0+int(0)) = param_0->a_0.shape_0;
    *((&param_0->a_0)->buffer_0+int(1)) = _S1.shape_0;
    return;
}
```

the reference for `strides` is disappeared. 

It's already disappeared in IR level, so it should not be emit logic bug.

kaizhangNV:
Close #6541.

Previously in type legalization pass, we skip the `VoidType` field when call `make_struct`, however in some optimization pass we keep counting the `VoidType` field. We have to make this behavior consistently over all our codebase.

So in this change, we spot the `make_struct` call and leave `VoidType` field as it.

### DIFF
--- a/source/slang/slang-ir-legalize-types.cpp
+++ b/source/slang/slang-ir-legalize-types.cpp
@@ -1962,8 +1962,6 @@ static LegalVal coerceToLegalType(IRTypeLegalizationContext* context, LegalType 
             ShortList<IRInst*> fields;
             for (auto field : structType->getFields())
             {
-                if (as<IRVoidType>(field->getFieldType()))
-                    continue;
                 auto fieldVal = coerceToLegalType(
                     context,
                     LegalType::simple(field->getFieldType()),


source/slang/slang-ir-peephole.cpp
```diff
@@ -402,6 +402,10 @@ struct PeepholeContext : InstPassBase
                     Index i = 0;
                     for (auto sfield : structType->getFields())
                     {
+                        if (as<IRVoidType>(sfield->getFieldType()) &&
```
csyonghe, line 405: It seems the problem is not with peephole optimization, but the code that results the consistency between struct type and make struct. Whenever we remove the void field from the struct type, we also need to update all make struct insts to keep consistent.

csyonghe:
We need to find the pass that causes inconsistency between make struct and struct type, and make sure we never violate the invariant that make struct has the same number of operands as the number of  struct fields.

source/slang/slang-ir-peephole.cpp
kaizhangNV: I fixed this problem because of looking at the code here:
https://github.com/shader-slang/slang/blob/master/source/slang/slang-ir-legalize-types.cpp#L1963

The location where we make this inconsistent is the `legalizeResourceTypes`.
When legalizing the struct with resource field, Meta will use VoidType for the resource field.
slang-legalize-types.cpp:558-559.

Later on, when we legalize the `kIROp_FieldAddress`, we will call `coerceToLegalType` to construct such new struct , and it won't create field for VoidType.
https://github.com/shader-slang/slang/blob/master/source/slang/slang-ir-legalize-types.cpp#L1963

I'm aware of that this fix could be problematic, as it's possible that there is some locations making struct with that "void" field, but I'm just not able to find it.

It will be help if you can find me some example.

csyonghe:
Need to update PR title and description to reflect what is being fixed.

kaizhangNV:
> Need to update PR title and description to reflect what is being fixed.

Done updating description.

PR #6550, opened 2025-03-09
Make image format reflection easier

Devon7925:
This makes image format reflection easier and adds known image formats to reflection json.


source/slang/slang-reflection-json.cpp
```diff
@@ -2,6 +2,7 @@
 #include "slang-reflection-json.h"
 
 #include "../core/slang-blob.h"
+#include "slang-ir-util.h"
```
fairywreath, line 5: Use direct include `slang-ast-support-types.h` instead of `slang-ir-util.h`

fairywreath (approved):

PR #6565, opened 2025-03-11
Add cooperative matrix 1 support

### LINKED ISSUE #6397: VK_KHR_cooperative_matrix (coopmat1) support
SPIRV: [SPV_KHR_cooperative_matrix](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_cooperative_matrix.html).
GLSL:  [GLSL_KHR_cooperative_matrix](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GLSL_KHR_cooperative_matrix.txt).

These help accelerate machine learning workloads. There are no (direct) DirectX, Metal or WGSL equivalents.

fairywreath:
Closes #6397 

Adds cooperative matrix 1(`VK_KHR_cooperative_matrix`, `SPV_KHR_cooperative_matrix`) support for SPIRV targets. The current Slang syntax is a straight copy of the GLSL(`GLSL_KHR_cooperative_matrix`) syntax plus a few `CoopVec` syntax. The implementation and test code is mostly derived from `CoopVec`. Full functionality is verified using a [fork of vk_cooperative_matrix_perf](https://github.com/fairywreath/vk_cooperative_matrix_perf_slang) where the GLSL shaders are replaced by Slang shaders.


source/slang/hlsl.meta.slang
```diff
@@ -21921,6 +21921,528 @@ extension<T, L : IBufferDataLayout> RasterizerOrderedStructuredBuffer<T, L> : IR
...
+    Subgroup = 3,
+    QueueFamily = 5,
+};
+
+enum CoopMatMatrixUse
+{
+    A = 0,
+    B = 1,
+    Accumulator = 2,
+};
```
jkwak-work, line 22413: Don't we need more descriptive names for these enums?

source/slang/hlsl.meta.slang
```diff
@@ -21921,6 +21921,528 @@ extension<T, L : IBufferDataLayout> RasterizerOrderedStructuredBuffer<T, L> : IR
     int getCount() { uint count; uint stride; this.GetDimensions(count, stride); return count; }
 }
 
+//
+// Cooperative Matrix type
+//
+
+__intrinsic_type($(kIROp_CoopMatrixType))
+[require(cooperative_matrix)]
+struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, let N : int, let R : CoopMatMatrixUse> : IArray<T>, IArithmetic
```
jkwak-work, line 21930: I am not sure what `CoopMatMatrixUse` is for, but I am wondering if it will be more consistent to other matrix types with the following order of the generic arguments,
```
CoopMat<T, M, N, S, R>
```

jkwak-work (commented):
What is the current status of this PR.
I reviewed it and it seems near complete, if not complete.
Do you have an ETA for this PR?

fairywreath:
> What is the current status of this PR. I reviewed it and it seems near complete, if not complete. Do you have an ETA for this PR?

Thanks for the review!

The remaining tasks are creating the proposal and adding  `coopMatMulAdd` tests. Syntax changes that can result from the proposal should not change this PR by a lot. I also recently learned that HLSL has a spec for [vulkan coop matrix](https://github.com/microsoft/hlsl-specs/blob/dd3931a15c18843313ac5a349735755a6abc6bd0/proposals/0021-vk-coop-matrix.md) (that seems to not be implemented yet). I am curious whether we should use this syntax define here instead, although I still prefer Slang to have its own set of coop mat intrinsics(that is similar to CoopVec syntax).

I will try to get this done by the end of this week or this weekend.

source/slang/hlsl.meta.slang
jkwak-work: I take this back, because the HLSL syntax uses your argument order.

jkwak-work:
@fairywreath , do you have any update on this PR?
We like to know ETA for a planning purpose.

fairywreath:
@jkwak-work Sorry I was not able to finish this last week. I will push the remaining test cases and spec proposal(which is very simple) by tomorrow.

fairywreath:
Remaining TODO is to add in-code documention for the stdlib reference, I will do this in a separate PR.

source/slang/slang-ir-util.cpp
```diff
@@ -23,6 +23,8 @@ IRType* getVectorElementType(IRType* type)
         return vectorType->getElementType();
     if (auto coopVecType = as<IRCoopVectorType>(type))
         return coopVecType->getElementType();
+    if (auto coopMatType = as<IRCoopMatrixType>(type))
```
csyonghe, line 26: This may not be needed.

jkwak-work: @fairywreath , can you address this comment?

fairywreath: Nice catch, I completely disregarded the `vector` part of this function name. This function is used intensively inside the SPIRV emitter to handle arithmetic and conversion operations. As coop matrix type is not a matrix type the SPIRV emitter will default to this function because it thinks it is working with a vector type. Cooperative matrices more like vectors than matrices as far as SPIRV is concerned as artihmetic OPs can be performed directly on them without looping through the matrix and accessing the vector elements individually. So this code is required(and just works) as this function gets called everywhere and proper SPIRV code is emitted because arithmetic operations that work on vectors also work on cooperative matrices.

I have added comments and created a new function `getVectorOrCoopMatrixElementType` to make this intention more clear.

csyonghe (approved):

source/slang/slang-ir-util.cpp
csyonghe: I see, thanks for the explanation!

PR #6569, opened 2025-03-11
Add referenced modules as libraries when creating a session

aleino-nv:
In the legacy compile request based API, the referenced modules are added to the request's linkage libraries as part of compiler option parsing. In the non-legacy compilation API, the argument parsing creates a temprary compile request and so those libraries only survive as options.

This change will look for such options when creating an ISession object, and again add the referenced modules to the libraries of the new linkage that's contained in the ISession object.

This is done in two steps:

1. Factor out a helper to create a referenced module artifact in the same way as it's done during legacy option parsing.
2. Use the helper function to create artifacts to add to the linkage libararies, when the session is created.

This helps to address issue #4760, because it enables passing in downstream modules via options, as is required for the following tests:
```
tests/library/library-test.slang.2 (dx12)
tests/library/export-test.slang.2 (dx12)
```

### DIFF
--- a/source/slang/slang-options.cpp
+++ b/source/slang/slang-options.cpp
@@ -1748,13 +1748,17 @@ SlangResult OptionsParser::_expectInt(const CommandLineArg& initArg, Int& outInt
     return SLANG_OK;
 }
 
-SlangResult OptionsParser::addReferencedModule(String path, SourceLoc loc, bool includeEntryPoint)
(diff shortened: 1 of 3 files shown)


aleino-nv:
@cheneym2 This was the only way I could figure out how to support referenced downstream DXIL modules with the new compilation API.

cheneym2 (approved):

PR #6576, opened 2025-03-11
Don't crash when precompiling twice

### LINKED ISSUE #6516: Precompiled Target Test: Precompile twice
Using the compilation API, ensure that calling precompileForTarget() twice for the same target does not cause an error.  The call should be a no-op.

cheneym2:
Abort precompileForTarget if it's already done.

Fixes #6516


kaizhangNV (approved):
LGTM

PR #6577, opened 2025-03-11
Implement parameter block to slang-gfx for Metal backend

### LINKED ISSUE #4816: Parameter block tests are failing on MacOS
The following tests are currently failing on MacOS and CI tests are ignoring for an unknown reason.

```
tests/bindings/nested-parameter-block-2.slang.3 (mtl)
tests/bindings/nested-parameter-block-3.slang.3 (mtl)
```

We need to fix the failures and make sure CI doesn't ignore them.

kaizhangNV:
Close #4816

Implement the parameter block support for slang-gfx metal backend.

On Metal, we will use argument buffer to bind parameter block, so we have to use argument buffer tier2 rule to calculate the layout for parameter block.

When using argument buffer, everything resources the argument buffer will be bound as bindless resource via device memory pointer or resource IDs, so we will have to notify the encoder to hazard track those resources.

### DIFF
--- a/tools/gfx/metal/metal-shader-object.cpp
+++ b/tools/gfx/metal/metal-shader-object.cpp
@@ -402,6 +402,7 @@ void ShaderObjectImpl::writeOrdinaryDataIntoArgumentBuffer(
 }
 
 BufferResourceImpl* ShaderObjectImpl::_ensureArgumentBufferUpToDate(
+    BindingContext* context,
     DeviceImpl* device,
     ShaderObjectLayoutImpl* layout)
 {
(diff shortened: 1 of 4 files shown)


tools/gfx/metal/metal-shader-object.cpp
```diff
@@ -401,94 +436,78 @@ BufferResourceImpl* ShaderObjectImpl::_ensureArgumentBufferUpToDate(
...
-        int bindingRangeIndex = 0;
-        SLANG_ASSERT(
-            defaultTypeLayout->getBindingRangeCount() == typeLayout->getBindingRangeCount());
-
-        int bufferBindingIndexOffset = layout->getTotalOrdinaryDataSize() != 0 ? 1 : 0;
-
-        for (unsigned int bindingRangeIndex = 0;
-             bindingRangeIndex < defaultTypeLayout->getBindingRangeCount();
-             bindingRangeIndex++)
+        // For parameter blocks, all the feilds are flattened as ordinary data, so the size of the
```
aleino-nv, line 439: Typo: feilds -> fields

aleino-nv (commented):
LGTM apart from the typo but probably someone else should also review

tools/gfx/metal/metal-shader-object.cpp
kaizhangNV: thanks, updated.

saipraveenb25 (approved):
LGTM, but I don't see any test coverage here. Is this handled elsewhere?

kaizhangNV:
> LGTM, but I don't see any test coverage here. Is this handled elsewhere?

No, it's because salng-test doesn't use slang-gfx anymore. But when we copy the change to slang-rhi, there will be test covered.

PR #6587, opened 2025-03-12
update slang-rhi

skallweitNV:
Update to latest `slang-rhi` version.

- Update render test application due to slang-rhi API changes
- Update texture format strings used in tests, they now use a `CamelCase` style

I had to disable two tests (adding them `to expected-failure-github.txt`):
- `tests/compute/cbuffer-legalize.slang.2 syn (mtl)` (@kaizhangNV maybe this is related to the metal parameter block change?!)
- `tests/hlsl-intrinsic/classify-float.slang.5 syn (wgpu)` (`isnan` bug reported here #6881)

In addition the following tests require the `half` render feature now for WGPU, because slang-rhi switched WGPU backend to Vulkan which doesn't support the `f16` extension. Once we can switch back to D3D12 backend again those tests will automatically be re-enabled as the `half` render feature will be reported again:
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic_Exclusive.slang`
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-arithmetic_None.slang`
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-ballot.slang`
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-quad.slang`
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle-relative.slang`
- `tests/glsl-intrinsic/shader-subgroup/shader-subgroup-shuffle.slang`
- `tests/hlsl-intrinsic/dot-accumulate.slang`
- `tests/hlsl-intrinsic/packed/pack-unpack-float.slang`

Explicitly removed testing on D3D11 for the following tests. Before, they were ignored because D3D11 didn't report the `hardware-device` render feature:
- `tests/glsl-intrinsic/unpack-float.slang`
- `tests/hlsl-intrinsic/packed/pack-unpack-float.slang`
- `tests/hlsl-intrinsic/packed/pack-unpack.slang`
- `tests/hlsl-intrinsic/packed/unpack-float.slang`
- `tests/hlsl-intrinsic/texture/texture-intrinsics.slang` (also fixed expected result, which now matches D3D12)

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -181,8 +181,8 @@ if(SLANG_ENABLE_SLANG_RHI)
 
     # Not disabling these break tests (issue #5474)
     set(SLANG_RHI_FETCH_DXC OFF)
-    set(SLANG_RHI_FETCH_AGILITY_SDK OFF)
-
+    set(SLANG_RHI_ENABLE_AGILITY_SDK OFF)
+    
     set(SLANG_RHI_ENABLE_NVAPI ${SLANG_ENABLE_NVAPI})
 
     if(SLANG_ENABLE_DX_ON_VK)
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -182,7 +182,7 @@ if(SLANG_ENABLE_SLANG_RHI)
     # Not disabling these break tests (issue #5474)
     set(SLANG_RHI_FETCH_DXC OFF)
     set(SLANG_RHI_ENABLE_AGILITY_SDK OFF)
-    
+
     set(SLANG_RHI_ENABLE_NVAPI ${SLANG_ENABLE_NVAPI})
 
     if(SLANG_ENABLE_DX_ON_VK)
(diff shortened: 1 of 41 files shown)


external/CMakeLists.txt
```diff
@@ -181,7 +181,7 @@ if(SLANG_ENABLE_SLANG_RHI)
 
     # Not disabling these break tests (issue #5474)
     set(SLANG_RHI_FETCH_DXC OFF)
-    set(SLANG_RHI_FETCH_AGILITY_SDK OFF)
+    set(SLANG_RHI_ENABLE_AGILITY_SDK OFF)
```
jkwak-work, line 184: I think we should use the agility sdk by default because it is required for cooperative vector to work.

jkwak-work (approved):
Looks good to me.

external/CMakeLists.txt
westlicht: Lets do that in a separate PR. I'm sure it will break stuff.

jkwak-work:
CI was failing on the following test,
```
tests/autodiff/array-param.slang (dx11)
```
I restarted the test but it may not be an intermittent failure.
We will see.

Issue #6580, opened 2025-03-11
sporadic corruption in slang-core-module-generated.h


lunarpapillo:
I'm seeing sporadic corrupted values appearing mostly in `slang-core-module-generated.h` and occasionally in `slang-glsl-module-generated.h`.  When the corruption appears, it often causes a compile failure; when it does not, the corrupted value seems to be integrated silently, which could be just as bad.  We've observed the corruption on **Windows 10** and **Ubuntu** machines, typically when we're building **slang** for inclusion in the **Vulkan SDK**.

Both `slang-core-module-generated.h` and `slang-glsl-module-generated.h` seem to be ASCII lists of hexadecimal bytes, 16 per line, with less than 16 usually appearing on the last line, e.g.:
```
0x25, 0x7f, 0x19, 0xfb, 0x3a, 0x7f, 0x19, 0x01, 0xba, 0x09, 0x11, 0x00, 0xf0, 0x00, 0x3b, 0x7f,
```
The most common corruption we've identified is an `0x` disappearing and being replaced by a newline, causing what should be a single line of 16 values broken into two lines:
```
0x19, 0x24, 0xfa, 0xec, 0x55, 0x1e, 0xb8, 0xb9, 0x54, 0xdf, 0x65, 0x00,
1e, 0x4e, 0x0a, 0x00,
```
This particular failure broke the build because `1e` is not a valid integer value.  We've seen corruption that didn't break the build, e.g.:
```
0x19, 0xfb, 0x50, 0x7f, 0x19, 0x01, 0xba, 0x0a, 0x01, 0xfa, 0xae, 0x58, 0x00, 0x00,
88, 0x15,
```
but this is concerning because the decimal 88 value will be used, when likely the hexadecimal 0x88 was intended.

The most unusual corruption we've observed is the insertion of many thousands of `nul` bytes (ASCII 0), e.g.:
```
0xfb, 0x50, 0x7f, 0x19, 0x01, 0xba, 0x0a, 0x01, 0xfa, 0xae, 0x58, 0x00,
^@^@...0x19, 0x01, 0xb9, 0xdc, 0x11, 0x00,
```
In this case, there were a total of 2208780 `nul` bytes inserted at the location marked with `...`.

On Linux we've observed another corruption, where the `0x` becomes a newline and `x`, e.g.:
```
[...]

Issue #6591, opened 2025-03-12
missing dependency on slang-glsl-module-generated.h


lunarpapillo:
While attempting to build for the **Vulkan SDK**, among the other sporadic cross-platform build failures we've been seeing is:

```
C:\j\msdk1\build\slang\repo\source\slang-glsl-module\slang-embedded-glsl-module.cpp(6,10): fatal error C1083: Cannot open include file: 'slang-glsl-module-generated.h': No such file or directory [C:\j\msdk1\build\slang\repo\dbuild\source\slang-glsl-module\slang-glsl-module.vcxproj]

```
This appears only infrequently in our builds (other issues, like #6580 , are much more frequent), but suggests that there's a missing dependency causing a race condition between creating `slang-glsl-module-generated.h` and its consumers.

expipiplus1:
This is so weird,

the only place this header is used is in `source/slang-glsl-module/slang-embedded-glsl-module.cpp` which is used in `source/slang-glsl-module/CMakeLists.txt` to generate the `slang-glsl-module` target. This target has an explicit dependency on the `generate_core_module_headers` custom target which depends on that header. 

Are you able to reproduce this with a single threaded build or with a different cmake generator?

bmillsNV:
@lunarpapillo can this issue be closed now or do you still see the issue?

PR #6594, opened 2025-03-13
Use an explicit dependency for generated headers instead of source file dependency

expipiplus1:
From the CMake Docs

> Do not list the output in more than one independent target that may build in parallel or the instances of the rule may conflict. Instead, use the add_custom_target() command to drive the command and make the other targets depend on that one. See the Example: Generating Files for Multiple Targets below.

Hopefully closes https://github.com/shader-slang/slang/issues/6580 and https://github.com/shader-slang/slang/issues/6591


gtong-nv (approved):
Verified locally. Running the build 10 times with MSBuild and they all pass and produce headers files with all hex numbers.

Issue #6566, opened 2025-03-11
Segfault with differentiable associatedtype


tunabrain:
The following snippet crashes the compiler:

```
interface IFoo<int N>
{
    associatedtype Params : IDifferentiable;

    [BackwardDifferentiable]
    Params bar();
}
```
Making `Params` or `bar` not differentiable, or removing the generic param, bypasses the crash.

PR #6600, opened 2025-03-13
Fix lowering of associated types in generic interfaces

saipraveenb25:
Fixes: #6566


source/slang/slang-lower-to-ir.cpp
```diff
@@ -1919,6 +1919,28 @@ struct ValLoweringVisitor : ValVisitor<ValLoweringVisitor, LoweredValInfo, Lower
...
+            SubstitutionSet(pairType->getDeclRef())
+                .forEachSubstitutionArg(
+                    [&](Val* arg)
+                    {
+                        auto argVal = lowerVal(context, arg).val;
+                        SLANG_ASSERT(argVal);
+                        operands.add(argVal);
+                    });
+
+            auto undefined = getBuilder()->emitUndefined(operands[1]->getFullType());
```
csyonghe, line 1937: This is another case where things would be much cleaner if DifferentialPair is just an associated type, and we can simply lower this normally as other associated types. The `undefined` operand for witness just feels wrong, and is suggesting our current handling of DifferentialPair isn't correct.

csyonghe (approved):

source/slang/slang-lower-to-ir.cpp
saipraveenb25: Yup, strong agree.. I'm adding this point to the "autodiff 2.0 type system" overhaul.

saipraveenb25:
Also fixes: #5999

csyonghe (approved):

PR #6603, opened 2025-03-14
Eliminate empty struct on metal target

### LINKED ISSUE #6573: Metal: emit empty struct
In metal, we should not emit empty struct or variable who type is an empty struct.
Note that metal has strict rule that such variable cannot be used in buffer pointee.

(We may need to legalization pass to eliminate all the empty struct in IR.)

kaizhangNV:
Close #6573.

We previously disabled the type legalization for ParameterBlock on Metal, but Metal doesn't allow empty struct in the argument buffer which is mapped from ParameterBlock, so we will need legalizeEmptyTypes on Metal target.

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -1278,6 +1278,15 @@ Result linkAndOptimizeIR(
         //
         legalizeResourceTypes(targetProgram, irModule, sink);
 
+        // We also need to legalize empty types for Metal targets.
+        switch (target)
+        {
+        case CodeGenTarget::Metal:
+        case CodeGenTarget::MetalLib:
+        case CodeGenTarget::MetalLibAssembly:
+            legalizeEmptyTypes(targetProgram, irModule, sink);
+            break;
+        }
         //  Debugging output of legalization
 #if 0
         dumpIRIfEnabled(codeGenContext, irModule, "LEGALIZED");
--- a/source/slang/slang-ir-legalize-types.cpp
+++ b/source/slang/slang-ir-legalize-types.cpp
@@ -4125,6 +4125,11 @@ struct IREmptyTypeLegalizationContext : IRTypeLegalizationContext
 
     bool isSimpleType(IRType* type) override
     {
+        if (isMetalTarget(targetProgram->getTargetReq()))
+        {
+            return false;
+        }
+
         // If type is used as public interface, then treat it as simple.
         for (auto decor : type->getDecorations())
         {
--- a/source/slang/slang-ir-legalize-types.cpp
+++ b/source/slang/slang-ir-legalize-types.cpp
@@ -4147,6 +4147,8 @@ struct IREmptyTypeLegalizationContext : IRTypeLegalizationContext
     {
         return LegalType();
     }
+
+    virtual bool isEmptyTypeLegalize() override { return true; }
 };
 
 // The main entry points that are used when transforming IR code
--- a/source/slang/slang-ir-legalize-types.cpp
+++ b/source/slang/slang-ir-legalize-types.cpp
@@ -4153,7 +4153,10 @@ struct IREmptyTypeLegalizationContext : IRTypeLegalizationContext
         return LegalType();
     }
 
-    virtual bool isEmptyTypeLegalize() override { return true; }
+    virtual bool shouldLegalizeParameterBlockElementType() override
+    {
+        return isMetalTarget(targetProgram->getTargetReq());
+    }
 };
 
 // The main entry points that are used when transforming IR code
--- a/source/slang/slang-legalize-types.cpp
+++ b/source/slang/slang-legalize-types.cpp
@@ -1211,7 +1211,8 @@ LegalType legalizeTypeImpl(TypeLegalizationContext* context, IRType* type)
         LegalType legalElementType;
 
         if (isMetalTarget(context->targetProgram->getTargetReq()) &&
-            as<IRParameterBlockType>(uniformBufferType) && !context->isEmptyTypeLegalize())
+            as<IRParameterBlockType>(uniformBufferType) &&
+            !context->shouldLegalizeParameterBlockElementType())
         {
             // On Metal, we do not need to legalize the element type of
             // a parameter block because we can translate it directly into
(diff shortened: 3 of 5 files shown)


source/slang/slang-emit.cpp
```diff
@@ -1189,6 +1189,15 @@ Result linkAndOptimizeIR(
         //
         legalizeResourceTypes(targetProgram, irModule, sink);
 
+        // We also need to legalize empty types for Metal targets.
```
csyonghe, line 1192: Can this be combined with line 1211 so we run this pass if it is metal or cpu/cuda?

kaizhangNV: I think it can, but are we sure that
`options.shouldLegalizeExistentialAndResourceTypes` only false when target is C/C++?

csyonghe: I think the logic there for C++ is wrong too. We need to run legalizeEmptyTypes regardless of `shouldLegalizeExistentialAndResourceTypes`.

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -4125,6 +4125,11 @@ struct IREmptyTypeLegalizationContext : IRTypeLegalizationContext
 
     bool isSimpleType(IRType* type) override
     {
+        if (isMetalTarget(targetProgram->getTargetReq()))
```
csyonghe, line 4128: Why is this needed? Seems very hacky.

source/slang/slang-legalize-types.cpp
```diff
@@ -1211,11 +1211,14 @@ LegalType legalizeTypeImpl(TypeLegalizationContext* context, IRType* type)
         LegalType legalElementType;
 
         if (isMetalTarget(context->targetProgram->getTargetReq()) &&
-            as<IRParameterBlockType>(uniformBufferType))
+            as<IRParameterBlockType>(uniformBufferType) && !context->isEmptyTypeLegalize())
```
csyonghe, line 1214: I feel like this `if` condition is getting messy. Can we just hoist this into a virtual `shouldLegalizeParameterBlockElementType()` method, which returns `true` by default, but overrided by `IRResourceTypeLegalizationContext` to return `false` when target is metal?

source/slang/slang-emit.cpp
csyonghe: The code here assumes that `legalizeResourceType` or `legalizeExistentialType` passes are supposed to also cleanup any empty types along the way. I don't understand why we are failing to clean it up for Metal. Is it due to the `ParameterBlock` special case handling logic?

csyonghe: Yeah, after looking at the test case, it is definitely caused by the skipping logic for metal parameter block element types, and that's why a separate empty type legalization pass is needed only for metal. This change here is OK then, no need to merge with the C++ case.

source/slang/slang-ir-legalize-types.cpp
kaizhangNV: Because all the following decor check are based on C/C++. 
For shader target, we always want to handle the type recursively. Otherwise, it will just return a LegalType::simple(type);

source/slang/slang-legalize-types.cpp
kaizhangNV: Yes, sounds good.

source/slang/slang-ir-legalize-types.cpp
kaizhangNV: What really should be treated as non-simple type in this case is a type whose leaf type is a struct. However, we shouldn't repeat ourselves to do such check here, because there are already existing logic in `legalizeTypeImpl()` doing the same thing. We just want to make sure on Metal target, this will execute the logic in that function without early return.

kaizhangNV: But I feel that the logic is almost 100% overlap with `legalizeEmptyTypes`. So there is no need to create different `TypeLegalizationContext`.

kaizhangNV: And, we need to run `legalizeEmptyTypes` if the target is metal, that is the point of this change actually.

saipraveenb25: Hmm.. 
I guess, since the legalization context does accept a `TargetProgram`, it's fine if target-specific stuff happens there.

saipraveenb25 (approved):
Looks good!

jkwak-work:
This PR seems to be causing CI failures.
https://github.com/shader-slang/slang/actions/runs/14092090746

The new test fails with the following error,
```
metal 32023.155: C:\Users\jkwak\AppData\Local\Temp\unkB06B.tmp.metal(30): error :  cannot initialize a member subobject of type 'int' with an lvalue of type 'device int *const constant'
    MyStruct_0 _S1 = { (*param_0).buffer_1, (*param_0).value_1 };
                       ^~~~~~~~~~~~~~~~~~~
1 error generated.
```
It looks like `MyStruct_0` gets only one member variable whose type is `int`.
So it is an error to assign two values on the construction.

The following shader is metal shader generated,
```
#include <metal_stdlib>
#include <metal_math>
#include <metal_texture>
using namespace metal;
void emptyStruct_set_0(int device* buffer_0, int value_0)
{
    *(buffer_0+int(0)) = value_0;
    return;
}

struct MyStruct_0
{
    int value_1;
};

void MyStruct_set_0(MyStruct_0 this_0, int device* this_buffer_0)
{
    emptyStruct_set_0(this_buffer_0, this_0.value_1);
    return;
}

struct MyStruct_1
{
    int device* buffer_1;
    int value_1;
};

[[kernel]] void computeMain(uint3 tid_0 [[thread_position_in_grid]], MyStruct_1 constant* param_0 [[buffer(0)]])
{
    MyStruct_0 _S1 = { (*param_0).buffer_1, (*param_0).value_1 };
    MyStruct_set_0(_S1, (*param_0).buffer_1);
    return;
}
```

I am wondering how this test passed CI, if it did.

fairywreath:
The incorrect Metal code is caused by having `RWStructuredBuffer<int> buffer;` inside a struct. I have a temporary fix #6709 for the empty struct test that resolves this. Issue seems to be unrelated to the empty struct elimination, a separate fix for having `RWStructuredBuffer` not emit incorrect metal code is required.

PR #6605, opened 2025-03-14
fix trimming of "v" from version string

Parad0x84:
Current releases are removing "v" from "vulkan-sdk-1.4.309.0" when they were supposed to remove it from something like "v1.2.4" https://github.com/shader-slang/slang/releases/tag/vulkan-sdk-1.4.309.0

Fixed it so it only removes "v", if next character is a number


csyonghe (approved):
Looks good to me!

Parad0x84:
I keep forgetting these pr labels. It's obviously a non-breaking change. It just renames output files in releases.
Are we good with merging this?

PR #6615, opened 2025-03-15
Support spirv ops added by multiple extensions

ramenguy99:
Some spirv ops are added by multiple extensions and capabilities. This PR adds support to avoid emitting unnecessary extensions and capabilities if one of the options is already required by some other op.

For example `OpRaytracingAccelerationStructureKHR` is provided by both `SPV_KHR_ray_tracing` and `SPV_KHR_ray_query`
and the respective capabilities `RayTracingKHR` and `RayQueryKHR`, but devices are not required to support both extensions (see  https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_ray_tracing.asciidoc#interactions-with-spv_khr_ray_query).

Before this change if a shader uses `RaytracingAccelerationStructure`  (this used to require `SPV_KHR_ray_tracing` and `RayTracingKHR`) and `RayQuery` (requires  `SPV_KHR_ray_query` and `RayQueryKHR`) ops for both extensions and capabilities were emitted. With this change only `SPV_KHR_ray_query` and the `RayQueryKHR` capability are emitted.

This is my first code contribution to the project, I apologize if there are mistakes. A few things for which I think some input could be useful:
- Naming might not be optimal. I tried to keep names short but some variables and methods are not super self-explanatory, let me know if you have any suggestions.
- I did not see many comments in the surrounding code, and most of the logic should be self explanatory, let me know if you think something could be added.
- I did not add any tests, not sure if there is an easy way to add a test that checks that some extension/capability is not emitted.

~ Dario


expipiplus1 (changes requested):
Is it possible to add a test for this change please

ramenguy99:
> Is it possible to add a test for this change please

I am not super familiar with the testing system, is there a way to test for a capability/extension to not be emitted? I see tests using things like:
```
// CHECK: OpCapability RayQueryKHR
// CHECK: OpExtension "SPV_KHR_ray_query"
```
But this change only reduces the number of capabilities that will be generated, so you would need a negative check.

csyonghe:
You can use CHECK-NOT:

see llvm filecheck documentation: https://llvm.org/docs/CommandGuide/FileCheck.html

csyonghe:
Add a line to existing test should be good enough.

ramenguy99:
Ended up adding the negative check for `RayQuery` to an existing tests but had to add a new one that does `RayQuery` in compute to generate a valid example that should not have `RayTracingKHR`

csyonghe (approved):

PR #6619, opened 2025-03-16
Add defer statement

juliusikkala:
Discussed here: https://github.com/shader-slang/slang/discussions/6549
More info here: https://github.com/shader-slang/spec/blob/main/proposals/legacy/003-error-handling.md

I tried to implement this as non-invasively as possible. If you don't use `defer`, the generated IR even before any passes should be the exact same as before.


juliusikkala:
Note: `SemanticsStmtVisitor::visitDeferStmt` should probably check that the statement does not contain return, continue or break. Those would make defer more or less impossible to implement, and are also banned in other languages with defer (e.g. Zig, [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Defer-Statement)).

juliusikkala:
That last commit probably needs some explanation. I realized that `defer` cannot be implemented reasonably in the IR level without any scoping information. I added `IRDeferHookDecoration` to mark a block as containing the end of a scope. Each `IRDefer` contains a "scope ID" that matches an `IRDeferHookDecoration`. Now, the contents of `defer` are pasted in front of branches that:

1. Are terminators of blocks dominated by an `IRDefer` AND
2. Target a block that is not dominated by `IRDefer` AND/OR occur after the block with the corresponding `IRDeferHookDecoration`

A common pattern with `defer` is the following:
```slang
{
    var resource = createResource(...);
    defer free(resource);
    // Do something with resource
}

{
    var otherResource = createOtherResource(...);
    defer free(otherResource);
    // Do something with resource
}
```

Without this additional `IRDeferHookDecoration` that is inserted for block statements, there is no info in the IR that would reveal that `defer free(resource);` should occur before `var otherResource = ...`.

Another case that was fixed at the same time is

```slang
for(int i = 0; i < 4; ++i)
{
    defer printf("%d ", 2*i+1);
    printf("%d ", 2*i);
}
```

`++i` is also a block that is dominated by the `defer` in the loop, so `defer` gets pasted after that. This made the program previously print `0 3 2 5 4 7 6 9`, which (IMHO) is clearly undesirable and not how other languages work. With the new scope information provided by `IRDeferHookDecoration`, this prints `0 1 2 3 4 5 6 7` as one would expect.

juliusikkala:
Some extra notes that I'd like to get feedback on in review:

* `IRDeferHookDecoration`:
    - Is the added complexity to `IRGenContext` & `StmtLoweringVisitor::visitBlockStmt` OK?
    - Do you have other, cleaner ideas for how to implement similar capabilities? I.e. how to detect the end of a block statement when scopes have been erased from the IR.
    - Currently, there's a check that worries me somewhat: `blockSucceedsDeferScope` in `slang-ir-lower-defer.cpp`. It makes assumptions about the order of blocks. Is it a safe assumption that blocks are in the order of control flow at this point?
* There is a legacy proposal that mentions `defer`, but it is not specifically about `defer` but error handling in general. Should a new proposal be written, that more clearly specifies what `defer` does?
* This is my first time working with IR extensively in the Slang compiler; please read the code with the assumption that I'm doing dumb things! I tried to learn how stuff works here from other files in the project.
* Bikeshedding: I'm not particularly attached to any names I've used in the code. If you have better ideas, let me know.

Additionally, I haven't tested the compile times of this implementation with larger pieces of code. If it seems too bad, it would probably be possible to cache information in `blockSucceedsDeferScope`, as it often ends up reading over the same blocks over and over again.

csyonghe (commented):
Looking at the code, I think it may be cleaner to represent defer the same way as IRLoop and IRIfElse in the IR (as normal blocks instead of children of a IRDefer inst). This should allow you to get rid of the HookDecoration, and clean up the lower-to-ir logic.

We need to include a test to verify that `defer` works with autodiff too.

jkwak-work:
I think we need more tests to amke sure defer works with the control flow keywords like `break` and `continue`.

juliusikkala:
Thanks for the feedback. I added more tests, which revealed a bug in the current implementation. It'll take me a while to take Yong He's feedback into account, but hopefully that'll simplify the implementation and make it more robust.

Regarding adding tests involving `throw`: how much of the try-catch mechanism actually works by now? Is it possible to already use `throw` in some way?

csyonghe:
IIRC throw is implemented since it is just lowering to a return, but catch isn’t fully implemented. We can skip testing for throw in this PR, and cover that when we get to implementing catch.

juliusikkala:
It turns out that the "bug" preventing the autodiff test from passing was actually a "feature": the defer runs after the return value has been computed. This is the same behaviour as in Swift (https://godbolt.org/z/4fWrdhhMo), so I think it's OK to leave it as it is and just change the expected values for the test.

I've now rewritten the patch to use the suggested `IRLoop` style blocks and deleted `deferHookDecoration`. This simplified the code a bit and removed the wishy-washy order-dependent behaviour in determining scope exits. It's controlled by dominance now, so it should be more robust as well. I did have to add a third operand though to handle surrounding scopes (like in `tests/language-feature/defer/scoped.slang`):

`defer <firstDeferBlock> <mergeBlock> <scopeEndBlock>`

The `scopeEndBlock` replaces the functionality previously provided by `deferHookDecoration`. The deferred blocks are placed in front of every terminator that is
1. Dominated by `defer`
2. Jumps to a location dominated by `scopeEndBlock` or not dominated by `defer`

csyonghe (commented):
We also need a test where the `defer` statement contains a loop with `break`/`continue` inside the loop.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -6429,10 +6434,35 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
...
+        IRBlock* prevScopeEndBlock = context->scopeEndBlock;
+
+        auto builder = getBuilder();
+        context->scopeEndBlockActive = false;
+        context->scopeEndBlock = builder->createBlock();
+
+        // To lower a block (scope) statement, just lower its body.
         lowerStmt(context, stmt->body);
+
+        if (context->scopeEndBlockActive)
```
csyonghe, line 6449: you should be able to call `context->scopeEndBlock->hasUses()` instead of maintaining a separate `scopeEndBlockActive` variable.

csyonghe (commented):
The implementation looks good, I commented inline for a few suggestions and improvements.

csyonghe:
Also remember to update `docs/user-guide/conventional-features.md #statements` section to talk about `defer` statement.

juliusikkala:
Thanks, I implemented the suggested improvements and updated the docs. I also found an issue with the previous logic for determining scope-exiting branches.

Basically, there's a way to branch to a location that is dominated by defer, but is outside of the scope, yet not dominated by the end of the scope. That case is `continue` in for-loops; the "increment" block of the for loop can be jumped into from both the continue statement and the end of the scope, so neither dominates it. I have added special logic to handle this case.

Also, there was a test failing with DX11 and DX12, but the reason seemed to not be related to defer but rather overwriting an RWStructuredBuffer value. I inserted a barrier in hopes of fixing this, but it'll be a few days until I can get a hold of my Windows laptop to debug it myself in case it still fails.

tests/language-feature/defer/switch.slang
```diff
@@ -0,0 +1,34 @@
+//TEST(compute):COMPARE_COMPUTE: -shaderobj
+//TEST(compute):COMPARE_COMPUTE: -vk -shaderobj
+//DISABLE_TEST(compute):COMPARE_COMPUTE:-cpu -shaderobj
```
csyonghe, line 3: Any reason to not include the test on cpu?

source/slang/slang-ir-lower-defer.cpp
```diff
@@ -0,0 +1,256 @@
...
+        // itself. The only case like this is the continue statement of a for
+        // loop.
+        if (dom->properlyDominates(scopeEndBlock, block))
+            return false;
+
+        // The continuation block is not always dominated by the end of the
+        // scope, since a continue statement is another way to get there.
+        // Luckily, we can just check if the scope end is directly succeeded by
+        // the target block, which is the case in for loops.
+        for (auto successor : scopeEndBlock->getSuccessors())
```
csyonghe, line 117: I wonder if we can get rid of this work around, if we handle the situation in lower-to-ir.

If you are lowering a block statement directly inside a loop, then don’t create a dedicated scopeEnd block, just use the continueBlock of the loop.

csyonghe:
This makes me wonder if things like
`for(…) defer f();`

will work properly?
What about 
`defer defer f();`?

Basically, if defer is inside some control flow structure but not directly nested inside a block stmt, what will happen?

tests/language-feature/defer/switch.slang
juliusikkala: Good point, I should enable them on the CPU as well. They were left disabled because I had troubles getting CPU tests to run on my machine. I have now figured out why that is (#6734).

csyonghe:
Do we need the same pushScope treatment for all other control flow constructs, like while, do?

What about if? Given

{
if(cond) defer X();
Y();
}

Should X happen before or after Y?

csyonghe:
Other than the semantics around defer in control flow constructs without {}, the PR looks good to me. I want to make sure we make a decision that makes sense is consistent with other languages.

juliusikkala:
Alright, I did a bit of a survey of languages with defer:

**Go**: Defers are collected and always run at the end of the function, regardless of scope.
**Swift**: This situation cannot arise, since {} is always required with control flow statements.
**Perl**: Braceless if syntax doesn't appear to allow `defer` statements.
**Zig**: Same issue as Swift
**Odin**:
```odin
package main

import "core:fmt"

main :: proc() {
    if(true) do defer fmt.println("A")
    fmt.println("B")
}
```
prints A first, then B.

**Nim**:
```nim
proc main() =
    if true:
        defer: echo "A"
    echo "B"
main()
```
prints "A" first, then "B". Although, : begins a block so I think this is actually in the same boat with Swift, where a block is always required.


**C2y defer TS 25755**:
    (this appears to be the latest development on the C front regarding defer, AFAIK)
    Example 4 from:
    https://thephd.dev/_vendor/future_cxx/technical%20specification/C%20-%20defer/C%20-%20defer%20Technical%20Specification.pdf
```c
#include <stdio.h>
#include <stdlib.h>
int main () {
    {
        defer {
            fputs(" meow", stdout);
        }
        if (true)
            defer fputs("cat", stdout);
        fputs(" says", stdout);
    }
    // "cat says meow" is printed to standard output
    exit(0);
}
```

It appears that while there isn't much to go off of here, existing practice seems to "anchor" defer to the surrounding statement, even if {} is not used.

So I think I should add tests for blockless `if`, `while` and `do-while` as well, and add pushScope for those as well.

csyonghe:
Yeah, I think anchoring the lifetime to be the enclosing control flow structure regardless of {} seems better and maintains consistency.

csyonghe:
Thanks for the quick fix, the PR looks good to me.

csyonghe (approved):

juliusikkala:
Sorry, MSVC had a warning over a variable name which failed the Windows build, fixed that now.

csyonghe (approved):

csyonghe:
Everything looks good and merged. Thank you for implementing this feature!

PR #6623, opened 2025-03-17
Force optimization

cheneym2:


jkwak-work:
I am not sure what this PR is for.
It doesn't have a link to an issue.

And the change in this PR doesn't make sense to me that default to map to default not none.

I think we should close this since the author is no longer going to make any changes to this.

PR #6635, opened 2025-03-18
Add Nix flake for direnv

samestep:
This PR makes it easy for developers using [Nix](https://nixos.org/) to build Slang from source without already having any of the prerequisites installed globally, by adding the following:

- A `flake.nix` file including a `devShell` with all the prerequisites from [Nixpkgs](https://github.com/NixOS/nixpkgs).
- A `flake.lock` file pinning the flake's dependencies to specific versions, making it reproducible (anyone using it will get the exact same versions of all the prerequisites).
- A line in `.gitignore` for `.envrc` in case people want to use [direnv](https://direnv.net/), along with the `.direnv` folder that would get automatically created in that case.
- A short section in `docs/building.md` documenting all this.

I've tested this on the following machines:

- My desktop with an x86 CPU running NixOS 24.11.
- My MacBook with an Apple Silicon M1 chip running macOS Sequoia 15.3.1.


flake.nix
```diff
@@ -0,0 +1,32 @@
+{
+  description = "A shading language";
```
saipraveenb25, line 2: Use something like this: "The Slang Shading Language and Compiler"?

samestep: @saipraveenb25 done!

expipiplus1 (changes requested):
This is really good to see, thank you!

I've actually been using a much more substantial flake myself for a couple of years https://github.com/expipiplus1/slang/blob/ellie-tools/flake.nix it's grown quite "organically" over that time and certainly isn't very neat or friendly; I can look into merging that in later

I think we shouldn't check in the .envrc file, and mention in the user guide that the user can do something like `echo 'use flake' >> .envrc && direnv allow`.

Please can you also mention in the readme that this is unofficial and not an officially supported or tested configuration?

samestep:
@expipiplus1 wow, that is indeed substantial! Yeah if you're able to merge in some of the useful stuff from your existing Nix config, that'd be awesome.

> I think we shouldn't check in the .envrc file, and mention in the user guide that the user can do something like `echo 'use flake' >> .envrc && direnv allow`.
> 
> Please can you also mention in the readme that this is unofficial and not an officially supported or tested configuration?

I've now done both of these things. Anything else you'd like me to do in this PR?

PR #6638, opened 2025-03-18
Add -dump-module command to slangc

### LINKED ISSUE #6599: Standalone command or utility to dump slang-module IR
As a Slang developer, it would be very useful to have a command or utility that would dump the slangIR in a slang-module file on disk.  It's not always convenient to dump IR contents during a `slangc` command using the `-dumpIR` argument.

Basically, just a command-line command that we could run which would disassemble and print the contents encoded in a slang-module.

It could maybe be a mode in `slangc`, or a new executable.  My humble opinion is that a new executable would be best.

Just an idea:

`slang-dump mymodule.slang-module -dumpIR`

or maybe an option to the compiler as in dxc.exe:

`slangc -dumpbin mymodule.slang-module`

cheneym2:
The new -dump-module command to slangc will load and disassemble a slang module, similar to what would be seen by the -dump-ir command, except that -dump-ir tells slangc to print IR as it performs some compilation command. That is, -dump-ir requires some larger compilation task.

-dump-module on the otherhand requires no additional goal and will simply load a module and print its IR to stdout independently from other compilation steps.

Its intended purpose is to inspect .slang-module files on disk. It can also be used on .slang files which will be parsed and lowered.  Slang will _not_ automatically load a ".slang-module" to disassemble if the user requested a ".slang" file.  This is intentional to give the user control to inspect modules loaded from source independently from ones loaded from precompiled slang-modules on disk.

The compilation API is extended with a new IModule::disassemble() method which retrieves the string representation of the dumped IR.

Closes #6599


source/slang/slang-options.cpp
```diff
@@ -2947,6 +2948,87 @@ SlangResult OptionsParser::_parse(int argc, char const* const* argv)
...
+                    if (SLANG_FAILED(module->disassemble(disassemblyBlob.writeRef())))
+                    {
+                        m_sink->diagnose(arg.loc, Diagnostics::cannotDisassemble, fileName.value);
+                        return SLANG_FAIL;
+                    }
+                    else
+                    {
+                        m_sink->diagnoseRaw(
+                            Severity::Note,
+                            (const char*)disassemblyBlob->getBufferPointer());
```
mkeshavaNV, line 3017: Is this the part that dumps onto the console, and is readable by slang-test?

mkeshavaNV (approved):

source/slang/slang-options.cpp
cheneym2: Yep, the new test "dump-module.slang" depends on this.  It calls slangc with the -dump-module argument and is able to scrape the output from this diagnoseRaw() call.

source/slang/slang-options.cpp
```diff
@@ -2947,6 +2948,87 @@ SlangResult OptionsParser::_parse(int argc, char const* const* argv)
...
+                auto desc = slang::SessionDesc();
+                ComPtr<slang::ISession> session;
+                m_session->createSession(desc, session.writeRef());
+                ComPtr<slang::IBlob> diagnostics;
+
+                // Coerce Slang to load from the given file, without letting it automatically
+                // choose .slang-module files over .slang files.
+                // First try to load as source string, and fall back to loading as an IR Blob.
+                // Avoid guessing based on filename or inspect the file contents.
+                FILE* file;
```
csyonghe, line 2966: We should never be using FILE directly. Use File::writeAllText or FileStream.

cheneym2: ack

expipiplus1 (approved):

PR #6641, opened 2025-03-18
Output SPV_KHR_compute_shader_derivatives extension string instead of the NV extension

### LINKED ISSUE #6639: Mixing of the KHR and NV version of some SPIR-V capabilites
```slang
[[vk::binding(0, 0)]]
Texture2D texture[];
[[vk::binding(1, 0)]]
SamplerState sampler;

struct StorageBuffer {
    float4 color;
};

[[vk::binding(2, 0)]]
RWStructuredBuffer<StorageBuffer> storageBuffer;

[shader("compute")]
void main() {
    storageBuffer[0].color = texture[0].Sample(sampler, float2(0));
}
```

https://godbolt.org/z/zTqG6nG1G (tested with 2025.6.2 as well

will produce the KHR capability, but the old NV extension (instead of `SPV_KHR_compute_shader_derivatives`)

```
OpCapability ComputeDerivativeGroupQuadsKHR
OpExtension "SPV_NV_compute_shader_derivatives"
```

which is not wrong, but can be confusing to people who don't realize it is all aliased

Curious are you parsing the spirv grammar as I [see here](https://github.com/KhronosGroup/SPIRV-Headers/blob/09913f088a1197aba4aefd300a876b2ebbaa3391/include/spirv/unified1/spirv.core.grammar.json#L16372-L16377) if you are just grabbing the first extension string it is the NV version... would a "fix" be to flip it in the spirv-header so the array of promoted things always puts the "newest" things at the front?

fairywreath:
Closes #6639.

`SPV_NV_compute_shader_derivatives` was ratified to a KHR extension. Slang currently outputs the hardcoded extension string as the NV extension and relies on the spirv header enums for capability, which since then has been changed to output the KHR version. This causes Slang to output a mix of both NV and KHR.

This PR changes the hardcoded NV extension string to the KHR variant.


source/slang/slang-capabilities.capdef
```diff
@@ -527,7 +527,11 @@ def SPV_NV_shader_image_footprint : _spirv_1_0;
 
 /// Represents the SPIR-V extension for compute shader derivatives.
 /// [EXT]
-def SPV_NV_compute_shader_derivatives : _spirv_1_0;
+def SPV_KHR_compute_shader_derivatives : _spirv_1_0;
+
+/// Represents the SPIR-V extension for compute shader derivatives.
+/// [EXT]
+alias SPV_NV_compute_shader_derivatives = SPV_KHR_compute_shader_derivatives;
```
fairywreath, line 534: I added this alias to not break the slangc/api interface and still allow usage of the NV variant as a Slang capability.

expipiplus1 (approved):

Issue #6606, opened 2025-03-14
Dead code path in collectExportedConstantInContainer()


cheneym2:
Noticed by inspection, in the function, collectExportedConstantInContainer(), there isn't a code path where `isConst` is false.  It looks buggy.

https://github.com/shader-slang/slang/blob/master/source/slang/slang.cpp#L2869

```
static void collectExportedConstantInContainer(
    Dictionary<String, IntVal*>& dict,
    ASTBuilder* builder,
    ContainerDecl* containerDecl)
{
    for (auto m : containerDecl->members)
    {
        auto varMember = as<VarDeclBase>(m);
        if (!varMember)
            continue;
        if (!varMember->val)
            continue;
        bool isExported = false;
        bool isConst = true;
        bool isExtern = false;
        for (auto modifier : m->modifiers)
        {
            if (as<HLSLExportModifier>(modifier))
                isExported = true;
            if (as<ExternAttribute>(modifier) || as<ExternModifier>(modifier))
            {
                isExtern = true;
                isExported = true;
            }
            if (as<ConstModifier>(modifier))
                isConst = true;
            if (isExported && isConst)
                break;
        }
        if (isExported && isConst)
        {
            auto mangledName = getMangledName(builder, m);
            if (isExtern && dict.containsKey(mangledName))
                continue;
            dict[mangledName] = varMember->val;
        }
    }

    for (auto member : containerDecl->members)
    {
        if (as<NamespaceDecl>(member) || as<FileDecl>(member))
        {
            collectExportedConstantInContainer(dict, builder, (ContainerDecl*)member);
        }
    }
}
```

PR #6645, opened 2025-03-19
Correct modifier isConst initialization

gtong-nv:
The previous implementation had two issues in the modifier processing loop:
1. isConst was incorrectly initialized to true, making the const check redundant
2. Premature loop break could skip processing important modifiers. e.g. isExtern

Changes:
- Initialize isConst to false by default
- Remove early break condition to process all modifiers

Fixes: #6606


gtong-nv:
I'm not entirely clear on the specs, but I assume the modifier should default to non-const to make the subsequent check and setting logical. Please correct me if I'm wrong.

csyonghe (approved):

PR #6651, opened 2025-03-20
Fix crash when using GLSL global uniforms and varying inputs/ouputs together

### LINKED ISSUE #6637: Fix compilation bug for global uniform variables
(Reported in discord)

This code fails to compile without any error messages, and crashes on an assert.
```
#version 330 core

layout(location=0) in float3 a_position;
layout(location=1) in float4 a_color;

out float4 v_color;

uniform matrix<float,4,4> u_transform;

[Shader("vertex")]
void main() {
    gl_Position = u_transform * vec4(a_position, 1.0);
    v_color = a_color;
}
```
Output is
```
(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: assert failure: globalParam
```

The problematic line is 
```
uniform matrix<float,4,4> u_transform;
```
Slang does not officially support non-vulkan flavored GLSL syntax so this does not work, but Slang should still return error with a proper message.

fairywreath:
Closes #6637. Fixes a crash caused by using GLSL global uniforms and varying inputs/ouputs together inside `CollectGlobalUniformParametersContext`. This pass collects all global uniforms and combines them into a single struct.

This crash is caused by calling `builder->setInsertBefore(globalParam)`(or for debug builds, a failed assertion on`SLANG_ASSERT(globalParam)` ) when `globalParam` is nullptr. `globalParam` is an `IRGlobalParam` type that is casted from a field of `globalScopeVarLayout`, and is nullptr when the field is not an `IRGlobalParam` type. It is valid however for a field to be an `IRGlobalVar` type, which is the case for varying inputs/outputs, hence `globalParam` can be nullptr and we get a crash.

The fix is to use the field instruction instead(which is guaranteed to not be null) rather than `globalParam`. There is already existing logic that skips processing the `globalParam` variable when the field is not a uniform type(which is the case for varying inputs/outputs), and when this skip is done the `globalParam` variable is not used at all.


tests/glsl/global-uniform-with-varyings.slang
```diff
@@ -0,0 +1,18 @@
+//TEST:SIMPLE(filecheck=CHECK_SPIRV): -target spirv -entry main -stage vertex
+//TEST:SIMPLE(filecheck=CHECK_GLSL): -target spirv -entry main -stage vertex
```
csyonghe, line 2: Should be -target glsl?

source/slang/slang-ir-collect-global-uniforms.cpp
```diff
@@ -180,21 +180,17 @@ struct CollectGlobalUniformParametersContext
...
-
-            // Once we have decided to do replacement, we need to
-            // set ourselves up to emit the replacement code.
-            //
-            builder->setInsertBefore(globalParam);
-
             // This global parameter needs to be turned into a field of the global
             // parameter structure type, and that field will need a key.
             //
+            builder->setInsertBefore(fieldLayoutAttr->getFieldKey());
```
csyonghe, line 186: Is this needed?

csyonghe:
Can you describe what is being fixed? I am seeing a coupe setInsertBefore being added, and an assert moved after an early return. What’s causing the crash?

source/slang/slang-ir-collect-global-uniforms.cpp
fairywreath: Yes this is not required

fairywreath:
I have added details of the crash and the fix in the PR description.

source/slang/slang-ir-collect-global-uniforms.cpp
```diff
@@ -216,6 +211,9 @@ struct CollectGlobalUniformParametersContext
             if (!globalParamLayout->getTypeLayout()->findSizeAttr(LayoutResourceKind::Uniform))
                 continue;
```
fairywreath, line 212: This skip will be invoked for varying inputs/ouputs, and `globalParam` is not used at all in this case, hence we can ignore it and only use it after this skip logic.

source/slang/slang-ir-collect-global-uniforms.cpp
fairywreath: Actually after running more tests it is required - the insert position needs to be set at a non-global instruction with a parent so `createStructKey` can work - it does not have to be of a `globalParam` type however.

csyonghe (approved):

Issue #6618, opened 2025-03-16
`-target exe` broken due to header issues


juliusikkala:
Trying to compile anything with `slangc -target exe` (at least on Linux), now yields the following:
```
gcc 13.3: /tmp/unknown-9vMR85.cpp(7097): error :  ankerl/unordered_dense.h: No such file or directory
 7097 | #include <ankerl/unordered_dense.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Appears to be caused by #6528. Since the ankerl include went from `""` to `<>`, `slang-embed` no longer embeds that header in the prelude. Luckily, this doesn't seem hard to fix. Is there any harm in changing the ankerl include to use `""`? Otherwise, slang-embed could also be changed to embed `<>` includes with some heuristic.

EDIT: Seems like quotes alone won't help, since `slang-embed` won't find that header in any case.

juliusikkala:
This has been fixed by #6675 before #6662 got merged.

PR #6662, opened 2025-03-21
Add unordered_dense as an include path for gcc and msvc

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/6618

This PR restores the old behavior, however we should probably modify slang-embed to also embed `#include`d files or install the unordered_dense headers


csyonghe:
This is not needed anymore, closing.

PR #6665, opened 2025-03-21
Add GLSL array length syntax support

### LINKED ISSUE #6652: issues when using slangc to compile glsl to spirv
slangc does not like glsl `debugPrintfEXT` and `array.length()` notation:
```glsl
#version 450

#extension GL_EXT_debug_printf : enable

void main() {
    int a[3];
    int l = a.length();
    debugPrintfEXT("hey\n");
}
```
```
slang.comp.glsl(7): error 30027: 'length' is not a member of 'int[3]'.
    int l = a.length();
              ^~~~~~
slang.comp.glsl(8): error 30015: undefined identifier 'debugPrintfEXT'.
    debugPrintfEXT("hey\n");
```
but this code compiles with glslang.
am i missing some flags perhaps ?

DarrelFW321:
Implement array.length() function for GLSL syntax.
 Closes #6652

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -321,6 +321,17 @@ ${{{{
 }
 }}}}
 
+/// Array length
+__generic<T, let N : int>
+public extension Array<T, N>
+{
+    [ForceInline]
+    public int length()
+    {
+        return this.getCount();
+    }
+}
+
 //
 // Section 8.1. Angle and Trigonometry Functions
 //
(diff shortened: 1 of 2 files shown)


fairywreath:
This is related #6652, can you add "closes #6652" in the PR description.

csyonghe:
Unrelated: this makes me realize that we can use the OpArrayLength opcode to implement StructuredBuffer.getDimension on spirv.

csyonghe (approved):

rcmz:
> Unrelated: this makes me realize that we can use the OpArrayLength opcode to implement StructuredBuffer.getDimension on spirv.

Oh yes you should, cause OpArrayLength also works for runtime arrays but getDimension does not.

Issue #6331, opened 2025-02-10
N5Slang13InternalErrorE: assert failure: typeParamDeclVal && typeParamDeclVal->val


rhoarau:
Hi,

We are using the OptiX Slang backend with some patches ([our fork](https://github.com/United-Visual-Researchers/slang)).
Later, We plan to submit a pull request once we have time to improve the patch quality with some guidance.

Recently, I rebased our fork to incorporate the latest upstream fixes (v2025.5). However, the introduction of ITexelElement ([PR #6115](https://github.com/shader-slang/slang/pull/6115)) introduced breaking changes that required modifications in our codebase.

While updating our code, I encountered a compiler crash. 
The error message is as follows:

```
.../generic-crash.slang(58): note 99999: an internal error threw an exception while working on code near this location
    public void accumulate(uint32_t2 pDestIdx, uint64_t pSampleIdx, vector<T, N> pSampleValue) {
                ^~~~~~~~~~
(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: assert failure: typeParamDeclVal && typeParamDeclVal->val
```

Steps to Reproduce:
**This crash is reproducible on the upstream master branch** with a simple slangc invocation (slangc ./generic-crash.slang). Below is a simplified code snippet (generic-crash.slang) that triggers the issue:
```
////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
[ForceInline]
__generic<T : __BuiltinFloatingPointType>
public T accumulateSample(uint64_t pSampleIdx, T pOldValue, T pNewValue) {
    T lTmp = isfinite(pOldValue) ? pOldValue : T(0);
    return lTmp + (pNewValue - lTmp) / T(pSampleIdx + 1);
}
[ForceInline]
__generic<T : __BuiltinFloatingPointType, let N : int>
public vector<T, N> accumulateSample(uint64_t pSampleIdx, vector<T, N> pOldValue, vector<T, N> pNewValue) {
    vector<T, N> lTmp = select(isfinite(pOldValue), pOldValue, T(0));
    return lTmp + (pNewValue - lTmp) / vector<T, N>(pSampleIdx + 1);
}
////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////

////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
public struct RenderBuffer2D<T : ITexelElement> {
    public Texture2D<T> handle;
    public uint16_t2 offset;
[...]

PR #6671, opened 2025-03-24
Fixed name mangling of generic extensions

SuperRonan:
I have had some problems recently with genereics lately, mostly generic extensions. 
I made simple shaders to expose the problem.

example1.slang
```
interface IValuable
{
	uint getValue();
};

__generic <Int : __BuiltinIntegerType>
extension Int : IValuable
{
	uint getValue()
	{
		return 0;
	}
};

__generic <Float : __BuiltinFloatingPointType>
extension Float : IValuable
{
	uint getValue()
	{
		return 1;
	}
};


RWByteAddressBuffer buffer;


[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 gl_GlobalInvocationID : SV_DispatchThreadID)
{
	uint i = gl_GlobalInvocationID.x;
	float f = float(i) / float(10);	

	buffer.StoreAligned<uint>(i, i.getValue()); // (a)
	buffer.StoreAligned<uint>(i + 1000, f.getValue()); // (b)
}
```

When compiling this shader (I did it in SPIR-V and GLSL), there is no error, but the result is incorrect.
Lines (a) and (b) should get compiled to:
```
buffer.StoreAligned<uint>(i, 0); // (a)
buffer.StoreAligned<uint>(i + 1000, 1); // (b)
```
But instead, they are compiled to:
```
buffer.StoreAligned<uint>(i, 0); // (a)
buffer.StoreAligned<uint>(i + 1000, 0); // (b)
```
Moreover, If we swap line (b) and (a) (store the `float` before the `uint`), we get:
```
buffer.StoreAligned<uint>(i + 1000, 1); // (b)
buffer.StoreAligned<uint>(i, 1); // (a)
```
It looks like it keeps only the first implementation of `getValue()`. 

But if we wrote non generic extensions like:
```
extension uint : IValuable
{
	uint getValue()
	{
		return 0;
	}
};

extension float : IValuable
{
	uint getValue()
	{
		return 1;
	}
};
```
We get a correct result. So the issue appears to come from generic extensions.
I looked at the IR dumps to try to find the issue. 
In the LOWER-TO-IR dump, both `getValue()` implementations are present, but in the next dump (POST IR VALIDATION) only one of the two is present (the first one being called).
Importantly, both functions were decorated with the same export tag: `[export("_S5test3XGP0I5test39IValuable8getValuep0pu")]`, which appears to be the source of the issue. (When using non generic extensions, both funtions get a different export tag.)

example2.slang
```
interface IValuableImpl
{
	uint getValue();
}

interface IValuable
{
	IValuableImpl getImpl();
};

__generic <Int : __BuiltinIntegerType>
extension Int : IValuableImpl
{
	uint getValue()
	{
		return 0;
	}
};

__generic <Float : __BuiltinFloatingPointType>
extension Float : IValuableImpl
{
	uint getValue()
	{
		return 1;
	}
};

__generic <ValuableImpl : IValuableImpl>
extension ValuableImpl : IValuable
{
	ValuableImpl getImpl()
	{
		return this;
	}
};


RWByteAddressBuffer buffer;


[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 gl_GlobalInvocationID : SV_DispatchThreadID)
{
	uint i = gl_GlobalInvocationID.x;
	float f = float(i) / float(10);	

	buffer.StoreAligned<uint>(i, i.getImpl().getValue());
	buffer.StoreAligned<uint>(i + 1000, f.getImpl().getValue());
}
```

The same problem arises with the second example, but this time it is less forward. 
This time, the problemn comes from the witness tables that are not exported properly. 
The witness table for `IValuableImpl` for both `__BuiltinIntegerType` and `__BuiltinFloatingPointType` get exported with: `[export("_SWGP05test413IValuableImpl")]`, and only one of the two remains during later stages of the compilation (the first one being called). 

Solution: 
I looked into the name mangler which generates these export names. 
Generic extensions don't appear to be properly handled by the mangler.
I removed some early returns in `emitQualifiedName(...)` that were preventing the emition of the generic declaration when existing.
Also when lowring an inheritance declaration (in `visitInheritanceDecl(...)`), I explicitely handled the case of a generic extension, so that the witness table gets a mangled name with the generic included.

For the first example, `__BuiltinIntegerType::getValue` is now decorated with `[export("_S5test3XGP0I5test39IValuableg2TCGP04core20__BuiltinIntegerType8getValuep0pu")]`, and `__BuiltinFloatingPointType::getValue` is now decorated with `[export("_S5test3XGP0I5test39IValuableg2TCGP04core26__BuiltinFloatingPointType8getValuep0pu")]`.

For the second example, the witness table of `IValuableImpl` for `__BuiltinIntegerType` is now decorated with `[export("_SW5test4XGP0I5test413IValuableImplg2TCGP04core20__BuiltinIntegerType5test413IValuableImpl")]`, and for `__BuiltinFloatingPointType`, it is now decorated with `[export("_SW5test4XGP0I5test413IValuableImplg2TCGP04core26__BuiltinFloatingPointType5test413IValuableImpl")]`.

(This is first time I look into Slang's implementation, so my solution might not be correct.)

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -8242,6 +8242,16 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         {
             subType = DeclRefType::create(context->astBuilder, makeDeclRef(parentDecl));
         }
+        bool isGenericExtension = false;
+        // Test if we are in a generic extension context
+        if (parentDecl->parentDecl)
+        {
+            auto genDecl = as<GenericDecl>(parentDecl->parentDecl);
+            if (genDecl)
+            {
+                isGenericExtension = true;
+            }
+        }
 
         // What is the super-type that we have declared we inherit from?
         Type* superType = inheritanceDecl->base.type;
@@ -8304,14 +8314,17 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
         {
             // Construct the mangled name for the witness table, which depends
             // on the type that is conforming, and the type that it conforms to.
-            //
-            // TODO: This approach doesn't really make sense for generic `extension`
-            // conformances.
-            auto mangledName = getMangledNameForConformanceWitness(
-                context->astBuilder,
-                subType,
-                superType,
-                irSubType->getOp());
+            String mangledName;
+            if (isGenericExtension)
+            {
+                mangledName =
+                    getMangledNameForConformanceWitness(context->astBuilder, parentDecl, superType);
+            }
+            else
+            {
+                mangledName =
+                    getMangledNameForConformanceWitness(context->astBuilder, subType, superType);
+            }
 
             // TODO(JS):
             // Should the mangled name take part in obfuscation if enabled?
--- a/source/slang/slang-mangle.cpp
+++ b/source/slang/slang-mangle.cpp
@@ -391,6 +391,7 @@ void emitVal(ManglingContext* context, Val* val)
 
 void emitQualifiedName(ManglingContext* context, DeclRef<Decl> declRef, bool includeModuleName)
 {
+    bool ignoreName = false;
     if (!includeModuleName)
     {
         if (as<ModuleDecl>(declRef))
@@ -445,19 +446,6 @@ void emitQualifiedName(ManglingContext* context, DeclRef<Decl> declRef, bool inc
         return;
     }
 
-    // Inheritance declarations don't have meaningful names,
-    // and so we should emit them based on the type
-    // that is doing the inheriting.
-    if (auto inheritanceDeclRef = declRef.as<TypeConstraintDecl>())
-    {
-        emit(context, "I");
-        emitType(context, getSup(context->astBuilder, inheritanceDeclRef));
-        return;
-    }
-
-    // Similarly, an extension doesn't have a name worth
-    // emitting, and we should base things on its target
-    // type instead.
     if (auto extensionDeclRef = declRef.as<ExtensionDecl>())
     {
         // TODO: as a special case, an "unconditional" extension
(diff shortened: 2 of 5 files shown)


csyonghe:
Can you add a test derived from your description?

SuperRonan:
I added the tests, but when running them, it triggered an exception error in bugs/gh-6331.slang, so I disabled it for now. 
I reduced the problem to this test (that I will add later), which crashes even with the old mangling. 
```
//TEST:SIMPLE(filecheck=CHECK): -target spirv
// CHECK: OpEntryPoint

public uniform RWTexture2D<float4> gOutput;

__generic<T : __BuiltinFloatingPointType, let N : int>
void Write(uint2 location, vector<T, N> value)
    where vector<T, N> : ITexelElement 
{
    float4 f4;
    for(int i = 0; i < min(4, N); ++i)  f4[i] = value[i].toFloat();
    gOutput[location] = f4;
}

[shader("compute")]
[numthreads(16, 16, 1)]
void main(uint3 pDispatchThreadIdx: SV_DispatchThreadID) {
    const uint2 lPixelCoordinates = { pDispatchThreadIdx.x, pDispatchThreadIdx.y };
}
```
Mangling the clause `where vector<T, N> : ITexelElement` triggers an assertion error in `getCanonicalGenericConstraints(context->astBuilder, parentGenericDeclRef)`. This function expects the sub type to be a `DeclRefType`, not an `ExpressionType`. 
The assert did not trigger before in bugs/gh-6331.slang because we were not mangling its clause `where vector<T, N> : ITexelElement` since it was in a generic extension. 
I'll fix that new bug.

kaizhangNV:
the windows CI failures should already be fixed on ToT, please rebase your PR.

SuperRonan:
Now that #6787 has been merged, tests/bugs/gh-6331.slang works again.

SuperRonan:
Hopefully the falcor tests pass now...

csyonghe (approved):

Issue #6674, opened 2025-03-24
slang-torch: Cannot import "ankerl/unordered_dense.h"


saipraveenb25:
This happens if the `slang-embed` tool sees an include directive that is not in scope. 
The embed tool was mostly designed to look through files right next to the prelude file that is being 'flattened' by recursively pasting the text of included files.

This process breaks everytime there's a change to the directory structure or change between relative/absolute include paths.

To fix this once and for all, the `slang-embed` tool needs to be able to see the effective set of include directories that the Slang project is using.

juliusikkala:
I think this is also related to #6618 and #6663, the PR may fix them too.

PR #6675, opened 2025-03-24
Improve embed tool to search all include directories as determined by CMake

saipraveenb25:
Fixes: #6674 

This change makes it so that we emit the full set of include directories from CMake as a command line argument to slang-embed. 

Hopefully this puts an end to prelude generation issues, and makes it so that we don't have to do elaborate workarounds to our include paths to keep `slang-embed` happy.

### DIFF
--- a/tools/slang-embed/slang-embed.cpp
+++ b/tools/slang-embed/slang-embed.cpp
@@ -63,47 +63,119 @@ struct App
 {
     char const* appName = "slang-embed";
     char const* inputPath = nullptr;
-    char const* includeDir = nullptr;
     char const* outputPath = nullptr;
+    Slang::List<Slang::String> includeDirs;
     Slang::HashSet<Slang::String> includedFiles;
+    size_t charCount = 0;
+    bool useNewStringLit = true;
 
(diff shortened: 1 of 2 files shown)


tools/slang-embed/slang-embed.cpp
```diff
@@ -63,47 +63,119 @@ struct App
...
+                    // Check if the directory has a semicolon or quotes at the end
+                    if (dirEnd > dirStart && (*(dirEnd - 1) == ';' || *(dirEnd - 1) == '"'))
+                        dirEnd--;
+
+                    // Save the current directory by creating a substring
+                    if (dirEnd > dirStart)
+                    {
+                        // Create a null-terminated copy
+                        size_t dirLen = dirEnd - dirStart;
+                        char* tempDir = (char*)malloc(dirLen + 1);
```
csyonghe, line 130: Use `List<char>` or `UnownedStringSlice` instead of `malloc`.

saipraveenb25: Okay, switched to Slang's String representation

kaizhangNV (approved):

Issue #3429, opened 2024-01-01
(GLSL/SPirV) Code generation - Inlining breaks shader


Paul-Hi:
While cleaning up one of my projects, I stumbled across something that seems like a bug to me.

I factored out functionality to the function `getIntersectionInfo` - which led to a significant drop in performance. So I looked at the generated output and realized that the problem is the missing inlining of the function, and tried to force the compiler to do so via [ForceInline]. The compiler then inlines the function, but ignores the condition for two other function calls with global buffer access in the conditional branches, which leads to crashes (probably an out-of-bounds access, since `getTriangle` and `getSphere` access information with their primitiveId in different buffers).

Shader compilation is done with `-profile glsl_460 -target glsl -o ${OUTPUT_SPIRV_FILE} -entry main -O3`. No difference could be determined for other optimization levels.

```
[ForceInline]
IntersectionInfo getIntersectionInfo(const RayPayload payload)
{
  IntersectionInfo iInfo;  
  if(payload.hitInfo.hitKind == EHitKind::triangleMesh)
  {
    DeviceTriangle triangle = getTriangle(payload.hitInfo.primitiveId, payload.hitInfo.instanceCustomIndex, ...);
    iInfo.material = gMaterials[triangle.materialId];
    iInfo.uv = triangle.interpolatedUV;
    iInfo.lightId = triangle.lightId;
    iInfo.space = CoordinateSpace(triangle.interpolatedNormal, ...);
  }
  else if(payload.hitInfo.hitKind == EHitKind::proceduralSphere)
  {
    DeviceSphere sphere = getSphere(payload.hitPoint, payload.hitInfo.primitiveId, ...));
    iInfo.material = gMaterials[sphere.materialId];
    iInfo.uv = sphere.uv;
    iInfo.lightId = sphere.lightId;
    iInfo.space = CoordinateSpace(sphere.normal, true);
  }
  return iInfo;
}

...

{

...

IntersectionInfo iInfo = getIntersectionInfo(payload); // compiler should inline here

...

}
```

The expected output for the call would look something like that:

```
bool _S170 = payload_1.hitInfo_0.hitKind_0 == 0U;
bool _S171 = payload_1.hitInfo_0.hitKind_0 == 1U;

...

IntersectionInfo_0 iInfo_0;
if(_S170)
{
[...]

PR #6677, opened 2025-03-25
Add check for the variable requirement

### LINKED ISSUE #6304: Usage of Buffer and RWBuffer yields invalid WGSL code with no errors
From [discord issue 'Oddity With Buffers When Targeting Wgsl in 2025.3.2'](https://discord.com/channels/1303735196696445038/1336924417871319142/1336924417871319142).

The following code

    Buffer<float> InputTyped;
    RWBuffer<float> OutputTyped;
    
    [numthreads(64, 1, 1)]
    void Main(uint3 DTid : SV_DispatchThreadID)
    {
        OutputTyped[DTid.x] = InputTyped[DTid.x];
    }

Compiling with `.\slangc .\test.slang -target wgsl -entry Main -stage compute -o test.txt` produces invalid WGSL without any errors:

    @compute
    @workgroup_size(64, 1, 1)
    fn Main(@builtin(global_invocation_id) DTid_0 : vec3<u32>)
    {
    }

gtong-nv:
This change adds the capability check for the variables requirement. With this check, the shader
```
[require(cpp_cuda_glsl_hlsl_metal_spirv)]
Buffer<float> InputTyped;
[require(cpp_cuda_glsl_hlsl_metal_spirv)]
RWBuffer<float> OutputTyped;
```
will issue error if targeting to WSGL
e.g. `.\build\Debug\bin\slangc .\tests\wgsl_no_buffer.slang -o wgsl_no_buffer.txt -target wgsl -entry Main -stage compute`

```
.\tests\wgsl_no_buffer.slang(2): error 36108: 'InputTyped' has dependencies that are not compatible on the required target 'wgsl'. Buffer<float> InputTyped;
              ^~~~~~~~~~
.\tests\wgsl_no_buffer.slang(4): error 36108: 'OutputTyped' has dependencies that are not compatible on the required target 'wgsl'. RWBuffer<float> OutputTyped;
                ^~~~~~~~~~~
```
Fixes #6304


kaizhangNV:
Overall, you should also:
1. add test case
2. check if the cap propagation correctly.

jkwak-work:
I don't have a first hand experience with the capability checking but the change looks reasonable.

I am not sure what you mean by "implicit" on #3. It sounds more like explicit to me if you are going to add [require(...)] attribute.

If you can add test cases to make sure that the expected errors are actually printed, we can prevent regressions in the future.

gtong-nv:
> I don't have a first hand experience with the capability checking but the change looks reasonable.
> 
> I am not sure what you mean by "implicit" on #3. It sounds more like explicit to me if you are going to add [require(...)] attribute.
> 
> If you can add test cases to make sure that the expected errors are actually printed, we can prevent regressions in the future.

Thanks Jay. "implicit" means that even if the user doesn't put `[require(...)] attribute` to the .slang code, if the Buffer/RWBuffer is used and the compile target is WGSL, we should still prints errors.

I think we can acommplish this by annotated the Buffer/RWBuffer with `[require(...)] attribute` in the core module?

gtong-nv:
> Overall, you should also:
> 
> 1. add test case
> 2. check if the cap propagation correctly.

Thanks @kaizhangNV , I added 3 simple tests. Currently, `var-capability-wgsl.slang` and `var-capability-wgsl-2.slang` will pass as they add the explicit `[require(cpp_cuda_glsl_hlsl_metal_spirv)]` and the compiler is able to detect the incompability. 
For var-implicit-capability-wgsl.slang, I'm still looking ways to add the capabliity requirement to the core modules so that the user doesn't need to write it out.

jkwak-work:
> > I don't have a first hand experience with the capability checking but the change looks reasonable.
> > I am not sure what you mean by "implicit" on #3. It sounds more like explicit to me if you are going to add [require(...)] attribute.
> > If you can add test cases to make sure that the expected errors are actually printed, we can prevent regressions in the future.
> 
> Thanks Jay. "implicit" means that even if the user doesn't put `[require(...)] attribute` to the .slang code, if the Buffer/RWBuffer is used and the compile target is WGSL, we should still prints errors.
> 
> I think we can acommplish this by annotated the Buffer/RWBuffer with `[require(...)] attribute` in the core module?

Yes, we should annotate [require(...)] in the core module "explicitly", I think.

tests/language-feature/capability/var-capability-wgsl-2.slang
```diff
@@ -0,0 +1,23 @@
+//TEST:SIMPLE(filecheck=CHECK): -target wgsl -entry Main -stage compute
+//TEST:SIMPLE(filecheck=CHECK_IGNORE_CAPS): -target wgsl -entry Main -stage compute -ignore-capabilities
+
+//CHECK:error 36108: 'InputTyped' has dependencies that are not compatible on the required target 'wgsl'.
+//CHECK_IGNORE_CAPS-NOT: error 36108
+struct inputWrapper {
+    [require(cpp_cuda_glsl_hlsl_metal_spirv)]
```
csyonghe, line 7: I don't think we should be adding `[require]` to parameter/var declarations. Instead they should be added to the `struct RWBuffer` decl, and we should infer the requirement of a vardeclbase from its type.

source/slang/slang-check-decl.cpp
```diff
@@ -12945,6 +12945,27 @@ CapabilitySet getStatementCapabilityUsage(SemanticsVisitor* visitor, Stmt* stmt)
 
 void SemanticsDeclCapabilityVisitor::checkVarDeclCommon(VarDeclBase* varDecl)
 {
+    auto declaredCaps = getDeclaredCapabilitySet(varDecl);
```
csyonghe, line 12948: This check should only be needed for global variables (variables that doesn't have a function parent), since for local variables their requirements are already propagated to the enclosing function and we are already diagnosing errors on the function.

tests/language-feature/capability/var-implicit-capability-wgsl.slang
```diff
@@ -0,0 +1,13 @@
+//TEST:SIMPLE(filecheck=CHECK): -target wgsl -entry Main -stage compute
```
csyonghe, line 1: We should have general tests for scenarios like:

```
//TEST:SIMPLE(filecheck=CHECK):-target spirv
[require(hlsl)]
struct MyType
{}


[numthreads(1,1,1)]
void f1()
{
    MyType t;  // compile to spirv should result error here.
}

[numthreads(1,1,1)]
void f2(uniform ConstantBuffer<MyType> t)  // compile to spirv should result error here.
{
     
}

ConstantBuffer<MyType> t2; // error here too.
```

source/slang/slang-check-decl.cpp
```diff
@@ -12962,6 +12962,32 @@ void SemanticsDeclCapabilityVisitor::checkVarDeclCommon(VarDeclBase* varDecl)
         },
         [this, varDecl](DiagnosticCategory category)
         { _propagateSeeDefinitionOf(this, varDecl, category); });
+
+    if (isGlobalDecl(varDecl))
+    {
+        auto declaredCaps = getDeclaredCapabilitySet(varDecl);
+        auto inferedCaps = varDecl->inferredCapabilityRequirements;
+        declaredCaps.nonDestructiveJoin(inferedCaps);
```
gtong-nv, line 12970: It takes me way too long to find the correct way to get the inferred requirement from the stdlib, and I'm still not sure if this the best place to add.
For now it fixes the issue and correclty report that capabilities not compatible with target for Buffer/RWBuffer used in WGSL. see test case `var-capability-wgsl-implicit.slang`

tests/language-feature/capability/cpp-target-has-cpp-impl-from-default.slang
```diff
@@ -1,6 +1,6 @@
 //TEST:SIMPLE(filecheck=CHECK): -target cpp -entry computeMain -stage compute -allow-glsl
 
-// CHECK: computeMain(
+// CHECK: error 36108: 'outputBuffer' has dependencies that are not compatible on the required target 'cpp'.
```
gtong-nv, line 3: This 2 tests -`cpp-target-has-cpp-impl-from-default` and `cpp-target-has-cpp-impl-from-default2` failed with this check added.  According to the comment below, it should "Fail". 
I updated the tests to make them PASS, assuming this is valid usage in C++ target.

gtong-nv (commented):
Hi @csyonghe, can you please take another look at this long-standing PR? 
Thanks!

tests/language-feature/capability/var-capability-incompatible.slang
```diff
@@ -0,0 +1,22 @@
+//TEST:SIMPLE(filecheck=CHECK1):-target spirv
+//TEST:SIMPLE(filecheck=CHECK2):-target spirv
```
csyonghe, line 2: there should be no need to run compile twice. Just a single `//TEST:SIMPLE(filecheck=CHECK)` line should suffice.

source/slang/slang-check-decl.cpp
csyonghe: I am not sure why we need this logic at all? Are we not reporting any errors without this code?

tests/language-feature/capability/var-capability-incompatible.slang
```diff
@@ -0,0 +1,22 @@
...
+}
+
+[numthreads(1,1,1)]
+void f2(uniform ConstantBuffer<MyType> t)  // compile to spirv should result error here.
+{
+     
+}
+
+//CHECK2: error 36107: entrypoint 'f2' uses features that are not available in 'compute' stage for 'spirv' target.
+ConstantBuffer<MyType> t2; // error here too.
```
csyonghe, line 22: we should only error here when `t2` is referenced by an entrypoint. So I would add some uses of `t2` from `f2`.
If `t2` is not referenced anywhere, it's fine to leave it there without any errors.

tests/language-feature/capability/var-capability-wgsl-2.slang
```diff
@@ -0,0 +1,17 @@
+//TEST:SIMPLE(filecheck=CHECK): -target wgsl -entry Main -stage compute
+//TEST:SIMPLE(filecheck=CHECK_IGNORE_CAPS): -target wgsl -entry Main -stage compute -ignore-capabilities
+
+//CHECK:error 36108: 'input' has dependencies that are not compatible on the required target 'wgsl'.
```
csyonghe, line 4: I don't think this should result in an error, because `Main` doesn't reference `inputWrapper` at all, so it shouldn't be considered part of program when compiling `Main`. You will need to reference `input` from `Main` to trigger an error.

source/slang/slang-check-decl.cpp
csyonghe: Looking at your tests, I don't think this logic is needed, because we shouldn't error on a global variable declaration if it is not referenced by an entrypoint.

source/slang/slang-parser.cpp
```diff
@@ -4965,8 +4965,9 @@ static DeclBase* ParseDeclWithModifiers(
                 };
                 if (AdvanceIf(parser, "buffer"))
                 {
-                    decl = as<Decl>(
-                        parseGLSLShaderStorageBufferDecl(parser, getLayoutArg("Std430DataLayout")));
+                    decl = as<Decl>(parseGLSLShaderStorageBufferDecl(
+                        parser,
+                        getLayoutArg("DefaultDataLayout")));
```
ArielG-NV, line 4970: Is this change supposed to be in this PR, It looks to fix a problem regarding GLSL `buffer` blocks (not capabilities)?

tests/bugs/gh-3429.slang
```diff
@@ -14,7 +14,7 @@ struct PushConstants
     bool bufferHasOnlyOneElement;
 };
 
-[[vk::push_constant]] ConstantBuffer<PushConstants> gPushConstants;
+[[vk::push_constant]] ConstantBuffer<PushConstants, ScalarDataLayout> gPushConstants;
```
ArielG-NV, line 17: Is this change related to the issue being solved (#6304)?

tests/language-feature/capability/var-capability-incompatible.slang
```diff
@@ -0,0 +1,19 @@
...
+[numthreads(1,1,1)]
+void f1()
+{
+    MyType t;  // compile to spirv should result error here.
+}
+
+ConstantBuffer<MyType> t2;
+//CHECK: error 36107: entrypoint 'f2' uses features that are not available in 'compute' stage for 'spirv' target.
+[numthreads(1,1,1)]
+void f2()  // compile to spirv should result error here.
```
ArielG-NV, line 16:
```suggestion
//CHECK: ([[# @LINE+2]]): error 36107
[numthreads(1,1,1)]
void f2()
```
We can check for the location of where an error points to using `(SOME_LINE_NUMBER):` in our check.

`[[# @LINE+2]]` fetches a line location relative to the current line plus 2 lines.

tests/language-feature/capability/var-capability-incompatible.slang
```diff
@@ -0,0 +1,19 @@
+//TEST:SIMPLE(filecheck=CHECK):-target spirv
+[require(hlsl)]
+struct MyType
+{}
+
+//CHECK: error 36107: entrypoint 'f1' uses features that are not available in 'compute' stage for 'spirv' target.
+[numthreads(1,1,1)]
+void f1()
+{
+    MyType t;  // compile to spirv should result error here.
```
ArielG-NV, line 10:
```suggestion
//CHECK: ([[# @LINE+4]]): error 36107
[numthreads(1,1,1)]
void f1()
{
    MyType t;
```
We probably do not want the full error in case we change the diagnostic wording.
This test is ensuring we check capabilities, not if the capability system error outputs the correct string.

Instead of a comment to where the error will be, we can add this to the test using hard-coded line numbers or a relative line location (`[[# @LINE+4]]`)

ArielG-NV (commented):
2 Questions on why certain changes were added

2 Recommendation for testing

tests/bugs/gh-3429.slang
gtong-nv: Similar story here. The default layout of ConstantBuffer is again changed to Std430DataLayout in `slang-check-decl.cpp:1993` . Have to specify it otherwise this test fails.

tests/language-feature/capability/var-capability-incompatible.slang
gtong-nv: Thanks! Updated

tests/bugs/gh-3429.slang
csyonghe: Changing default of ConstantBuffer layout to Scalar is wrong and will break a lot of things. All these defaults should just be DefaultBufferLayout that means platform and buffer type specific.

gtong-nv: In `lang-check-decl.cpp:1993`
```
        // If we see a ConstantBuffer<T, DefaultLayout> parameter marked as "push_constant", we need
        // to set its type to ConstantBuffer<T, Std430>.
        if (auto cbufferType = as<ConstantBufferType>(varDecl->type))
        {
            if (cbufferType->getLayoutType() == m_astBuilder->getDefaultLayoutType())
            {
                varDecl->type.type = getConstantBufferType(
                    cbufferType->getElementType(),
                    m_astBuilder->getStd430LayoutType());
            }
        }
```

With the default `ConstantBuffer<PushConstants>`, the layout would be changed to Std430 because of the block above, But we are targing hlsl in this test.
The PR introduces the capability checking, so for Std430 only spirv and glsl are supported. 

I'm not sure what is the rational of that block to change ConstantBuffer with push_constant to Std430 though.

gtong-nv: Or maybe this test should be targeting hlsl?

csyonghe:
I think the logic to change from DefaultLayout to Std430Layout directly at the AST level is problematic with the capability requirement. We should remove that logic and keep it as DefaultLayout, but change the parameter layout reflection logic for the Khronos targets to select std430 layout when it sees a ConstantBuffer with DefaultLayout and decorated with PushConstant attribute.

tests/bugs/gh-3429.slang
ArielG-NV: It may be that the capabilities are outdated and should be changed.

The following works in Slang-Playground for hlsl, glsl, metal, spirv. It internal errors for WGSL
```
import playground;

struct DataStruct
{
    int32_t v1;
    int16_t v2;
    int32_t v3;
}
ConstantBuffer<DataStruct, Std430DataLayout> cb1;
ConstantBuffer<DataStruct, Std140DataLayout> cb2;

RWStructuredBuffer<int> outBuffer;
void printMain()
{
    outBuffer[0] = cb1.v1+cb2.v1;
    outBuffer[1] = cb1.v2+cb2.v2;
    outBuffer[2] = int(cb1.v3)+int(cb2.v3);
}
```

gtong-nv:
> I think the logic to change from DefaultLayout to Std430Layout directly at the AST level is problematic with the capability requirement. We should remove that logic and keep it as DefaultLayout, but change the parameter layout reflection logic for the Khronos targets to select std430 layout when it sees a ConstantBuffer with DefaultLayout and decorated with PushConstant attribute.

Do you suggest moving the DefaultLayout -> Std430Layout logic to 
`slang\slang-parameter-binding.cpp` getTypeLayoutForGlobalShaderParameter()

Not sure if I'm doing the right thing but it seems there is a lot need to be done to be able to create `ConstBufferType` with `Std430DataLayoutType` from here. I'm not sure how the witness works -- see the TODO in the block below..

```
        // For Khronos targets, if the ConstantBuffer has DefaultLayout, change it to Std430Layout
        if (isKhronosTarget(context->getTargetRequest()))
        {
            if (auto cbufferType = as<ConstantBufferType>(type))
            {
                auto layoutType = cbufferType->getLayoutType();
                if (layoutType->astNodeType == ASTNodeType::DefaultDataLayoutType)
                {
                    // Create a new ConstantBufferType with Std430Layout instead of DefaultLayout
                    auto astBuilder = context->getASTBuilder();
                    auto std430LayoutType = astBuilder->getStd430LayoutType();
                    
                    // TODO: Create the layout witness using the same logic as in slang-check-type.cpp ?
                    // Do we need to replicate the isSubtype logic here? 
                    // auto witness = isSubtype(layoutType, iBufferDataLayoutType, IsSubTypeOptions());

                    auto newCBufferType = astBuilder->getConstantBufferType(
                        cbufferType->getElementType(),
                        std430LayoutType,
                        witness );
                    type = newCBufferType;
                }
            }
        }
```

csyonghe (approved):

PR #6686, opened 2025-03-25
Fix mul operator followed by global scope

### LINKED ISSUE #6684: Failed to parse arithmetic operator before scope operator
Slang current fails to parse code like this

```
2.0f * ::ns::numbers::pi
```

but this will work
```
::ns::numbers::pi * 2.0f
```

one can use this URL to check the compile error.

https://shader-slang.org/slang-playground/?target=SPIRV&code=eJxdjcEKwjAQRO_7FUtBaKWIetzgob37B17WtJFAuwlpFETy78ZUEJzbvMcwdvYuRPQTP2_B3WVQAMLzuHjWI3bwAsz5kb70lX6yRI5Wo5WIVzzhQf0bMzmOqLM77vZm1QkSwMPZAX3IyzNbqZvvVSF1tTEXqdqywS0SdUQ9kW4UpDc9YTEv

This should be a parse bug.

gtong-nv:
This should fix expr like `2.0f * ::a::b::c`

But it will no longer parse something like
```
extension<T> Ptr<T> { static void foo(); }
int*::foo() // won't work, but this is a less common case
```

Fixes #6684

### DIFF
--- a/tests/language-feature/namespaces/simple-namespace.slang
+++ b/tests/language-feature/namespaces/simple-namespace.slang
@@ -3,9 +3,11 @@
 //TEST(compute):COMPARE_COMPUTE: -shaderobj
 
 // Test that simple `namespace` declarations work as expected
+// Test that the global scope operator `::` works as expected
 
 namespace A
 {
+    static int num = 16;
     struct X
     {
         int val;
@@ -47,7 +49,10 @@ namespace A
 int test(int val)
 {
     A.X a = A::makeX(val);
-    B::X b = B.makeX(val*16, val*256);
+    // Use the global scope operator "::A::num" to access the static member of namespace A
+    // Test it with mul operator
+    int num = 16*::A::num;
+    B::X b = B.makeX(val*16, val*num);
 
     return a.getVal() + b.getHead() + b.getTail();
 }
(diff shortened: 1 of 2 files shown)


csyonghe:
need to include a test.

gtong-nv:
> need to include a test.

Not sure if I should make a new test. I updated the simpe-namespace.slang to verify this issue.

jkwak-work:
I am not sure if this PR should be treated as a breaking change or not.

tests/language-feature/namespaces/simple-namespace.slang
```diff
@@ -47,7 +49,10 @@ namespace A
 int test(int val)
 {
     A.X a = A::makeX(val);
-    B::X b = B.makeX(val*16, val*256);
+    // Use theglobal scope operator "::A::num" to access the static member of namespace A
```
jkwak-work, line 52: Typo. "The global".

LLM will be very helpful to catch mistakes like this, but I am not sure what would be a best way to wire it up.

jkwak-work (approved):
Looks good to me

tests/language-feature/namespaces/simple-namespace.slang
gtong-nv: Thanks Jay!
I should have asked LLM to review first!
I used to have it read `git diff > patch.txt` and do proofreading. Should automate the process.

csyonghe:
need to assign a pr:non-breaking label to the PR.

csyonghe (approved):

PR #6688, opened 2025-03-25
Fixed generic interface specialization crashes (#6601):

### LINKED ISSUE #6601: Generic interfaces + dynamic-dispatch is currently crashing
Example test to reproduce

```hlsl
// Test calling differentiable function through dynamic dispatch.

//TEST_INPUT:ubuffer(data=[0 0 0 0 0], stride=4):out,name=outputBuffer

interface IGetter
{
    float get(uint id);
}

struct GetterImpl : IGetter
{
    float[8] data;

    __init(float[8] data)
    { this.data = data; }

    float get(uint id)
    {
        return data[id];
    }
}
interface IFoo<int N>
{
    associatedtype Params : IGetter;

    Params bar();
}

struct FooImpl1: IFoo<8>
{   
    typealias Params = GetterImpl;

    __init()
    { }

    Params bar()
    {
        float x = outputBuffer[0];
        return GetterImpl({x, x+1, x+2, x+3, x+4, x+5, x+6, x+7});
    }
}

struct FooImpl2: IFoo<8>
{
    typealias Params = GetterImpl;

    __init()
    { }

    Params bar()
    {
        float x = 2 * outputBuffer[0];
        return GetterImpl({x, x+5, x+7, x+9, x+11, x+13, x+15, x+17});
    }
}

IFoo<8> getFoo(uint id)
{
    if (id == 0)
        return FooImpl1();
    else
        return FooImpl2();
}

float doThing(uint id)
{
    IFoo<8> foo = getFoo(id);
    return foo.bar().get(0);
}

[shader("compute")]
{
    outputBuffer[0] = doThing(0);
    outputBuffer[1] = doThing(1);
}
```
Error:
```
(0): error 99999: Slang compilation aborted due to an exception of class Slang::InternalError: assert failure: interfaceLinkage && "An interface type does not have a linkage," "but a witness table associated with it has one."
```

Issue appears to be that the `ExportDecoration` on the interface type is being dropped during specialization of the interface type. 
We probably just need to create a new mangled name during specialization.

SuperRonan:
As mentioned in #6601, generic interface specialization were missing an export tag -> they are considered as dead code -> can't link them in later compilation stages. 
The solution was to check when specializing an interface if it had an export tag, and create a new export tag for the specialization.
I wrote a basic IR mangler to generate the specialized export tag, but it is probably not a very durable solution (Is there already an IR mangler? I did not find one). 

Closes #6601.


source/slang/slang-ir-clone.cpp
```diff
@@ -104,6 +104,75 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
+                            {
+                                sb.append("m");
+                                rec_f(matrixType->getRowCount(), rec_f);
+                                sb.append("x");
+                                rec_f(matrixType->getColumnCount(), rec_f);
+                                rec_f(matrixType->getElementType(), rec_f);
+                            }
+                            else
+                            {
+                                SLANG_E_NOT_IMPLEMENTED;
```
csyonghe, line 164: SLANG_E_NOT_IMPLEMENTED is just an error code, you probably mean `SLANG_UNIMPLEMENTED_X("ir type mangle")` here to throw an error.

source/slang/slang-ir-clone.cpp
```diff
@@ -104,6 +104,75 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
+        {
+            StringBuilder sb;
+            sb.append(genExport->getMangledName());
+            for (UInt i = 0; i < oldInst->getArgCount(); ++i)
+            {
+                auto arg = oldInst->getArg(i);
+                // TODO call the mangler (but it works one the AST, not the IR?)
+                // This will do for now
+                sb.append(i);
+                auto emit = [&sb](IRInst* inst, const auto & rec_f) -> void
```
csyonghe, line 123: maybe hoist this out into a proper utility function.

source/slang/slang-ir-clone.cpp
```diff
@@ -104,6 +104,75 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
+                {
+                    if (auto intValue = as<IRIntLit>(inst))
+                    {
+                        sb.append("V");
+                        sb.append(intValue->getValue());
+                    }
+                    else if (auto type = as<IRType>(inst))
```
csyonghe, line 130: `as<IRType>` is dangerous, since other opcodes like `LookupWitness` or `Specialize` can also be used as types,
 but `as<IRType>(IRSpecialize)` is nullptr, because IRSpecialize is not considered a subtype of IRType...

consider using a big switch statement after `if basictype` for all other cases.

source/slang/slang-ir-clone.cpp
SuperRonan: Thanks, I did not know!
Resolved with 1fc1bcb.

source/slang/slang-ir-clone.cpp
SuperRonan: Resolved with 1fc1bcb.

SuperRonan:
I replaced the small manual mangler that I wrote with `getTypeNameHint(StringBuilder& sb, IRInst* type)` which does mangling of IR types (but not with exactly the same syntax as the AST mangling, I don't know if it could be an issue).

source/slang/slang-ir-clone.cpp
```diff
@@ -104,69 +105,38 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
-                // TODO call the mangler (but it works one the AST, not the IR?)
-                // This will do for now
+                specializationProvider = targetAsSpec;
+            }
+            for (UInt i = 0; i < specializationProvider->getArgCount(); ++i)
+            {
+                auto arg = specializationProvider->getArg(i);
                 sb.append(i);
-                auto emit = [&sb](IRInst* inst, const auto & rec_f) -> void
+                if (auto typeExport = arg->findDecoration<IRExportDecoration>())
```
csyonghe, line 128: should be `IRLinkageDecoration` to also take into account for any `IRImportDecoration`s.

source/slang/slang-ir-clone.cpp
```diff
@@ -104,69 +105,38 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
+                {
+                    sb.append(typeExport->getMangledName());
+                }
+                else if (auto nameHint = arg->findDecoration<IRNameHintDecoration>())
```
csyonghe, line 132: this shouldn't be needed, as `getTypeNameHint` will do the same thing.

source/slang/slang-ir-clone.cpp
```diff
@@ -104,69 +105,38 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
...
-                            }
-                            else
-                            {
-                                SLANG_E_NOT_IMPLEMENTED;
-                            }
-                        }
-                    }
-                };
-                emit(arg, emit);
+                    getTypeNameHint(sb, arg);
```
csyonghe, line 138: getTypeNameHint may introduce characters that can't be part of an identifier name, e.g. `.` or `<>`.

We need to wrap this with a post process that escapes thses characters with `x99` where `99` is the hex of the character value.

source/slang/slang-ir-clone.cpp
```diff
@@ -300,6 +265,11 @@ static void _cloneInstDecorationsAndChildren(
         newParam->setFullType(newType);
         newParam->sourceLoc = oldParam->sourceLoc;
     }
+
+    if (auto oldAsSpec = as<IRSpecialize>(oldInst))
+    {
+        specializeExportDecorationIFP(newInst, oldAsSpec, builder);
```
csyonghe, line 271: what does `IFP` stand for? we don't use this postfix in the codebase, `specializeExportDeocrations` is fine.

source/slang/slang-ir-clone.cpp
```diff
@@ -104,6 +105,44 @@ IRInst* cloneInstAndOperands(IRCloneEnv* env, IRBuilder* builder, IRInst* oldIns
     return newInst;
 }
 
+static void specializeExportDecorationIFP(IRInst* target, IRSpecialize* oldInst, IRBuilder* builder)
```
csyonghe, line 108: Need a comment here to explain what the pass is doing.

source/slang/slang-ir-clone.cpp
SuperRonan: `IFP` = IF Possible (and `IFN` = IF Necessary), I could also have called the function `maybeSpecializeExportDeocrations`. 
I find it more explicit. `specializeExportDeocrations` is also fine.

source/slang/slang-ir-clone.cpp
SuperRonan: I noticed that too, but when I tested it with some of these chars (like with vector which included `<` and `>`), it worked, so I though it was ok. 
Maybe it's because I tested it only on functions that were directly compiled to a target, and not compiled to some module and then imported in another?

source/slang/slang-ir-clone.cpp
SuperRonan: Fixed with bd89a6f.

source/slang/slang-ir-clone.cpp
SuperRonan: Fixed with bd89a6f.

source/slang/slang-ir-clone.cpp
SuperRonan: Fixed with bd89a6f.

source/slang/slang-ir-clone.cpp
SuperRonan: Fixed with bd89a6f.

source/slang/slang-ir-clone.cpp
SuperRonan: Fixed with bd89a6f. the function also specializes the import declaration.

csyonghe (approved):

csyonghe:
The change looks good, can we add a test to verify  #6601 is fixed?

csyonghe (approved):
Looks good, thanks for adding the test!

jarcherNV:
It looks like this fix is ready so I have enabled auto-merging and rebased it to ToT. This should be merged once the tests have passed.

PR #6689, opened 2025-03-26
Ignore failure to fetch webgpu_dawn and slang-tint

### LINKED ISSUE #6683: Slang build fails to download webgpu-dawn and slang-tint
Some Slang builds have started failing recently due to being unable to download webgpu-dawn and slang-tint.

Example error:
CUSTOMBUILD : error : downloading 'https://github.com/shader-slang/webgpu-dawn-binaries/releases/download/v131.0.6738.0/webgpu-dawn-131.0.6738.0-windows-x86_64.zip' failed [C:\slang\build\_deps\dawn-subbuild\dawn-populate.vcxproj]
            status_code: 56
            status_string: "Failure when receiving data from the peer"

Previously I had fixed github download errors by using -DSLANG_GITHUB_TOKEN=XXXX but that does not fix this.
The build works for me locally but not from my remote PC in the office, there I get an error when trying to access the url.

gtong-nv:
Ignore failure to fetch webgpu_dawn and slang-tint

Fixes #6683

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,7 +371,11 @@ add_subdirectory(external)
 # trees are too large and onerous for us to depend on.
 # We ignore the failure to fetch the library, since it's not required for the build to succeed.
 if(SLANG_WEBGPU_DAWN_BINARY_URL)
-    copy_fetched_shared_library("webgpu_dawn" "${SLANG_WEBGPU_DAWN_BINARY_URL}" IGNORE_FAILURE)
+    copy_fetched_shared_library(
+        "webgpu_dawn"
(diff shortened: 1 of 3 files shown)


jarcherNV (approved):
This seems reasonable. It would be nice if we could find a fix that also downloads and builds these however.

gtong-nv:
> This seems reasonable. It would be nice if we could find a fix that also downloads and builds these however.
yes, in the long run we might need to consider store the prebuilt in other storage provider - might after the google cloud pipeline is ready.

csyonghe (approved):

Issue #281, opened 2017-11-15
IR: type legalization doesn't apply to `cbuffer`s


tfoleyNV:
We currently scalarize a `struct` type like this:

```hlsl
struct P { Texture2D t; SamplerState s; };

void test(P p);
```

so that we get a compiled function like:

```hlsl
void test(Texture2D p_t, SamplerState p_s);
```

However, if we call that function by directly passing in a field of a `cbuffer`:

```hlsl
cbuffer C { P p; }
void main()
{
    test(p);
}
```

Then the code lowers to still contain a call to `test` with a single argument of type `P` (rather than the scalarized fields of `P`), and the `cbuffer` still declares a field of the original type. This leads to compilation failure in the downstream compiler.

tfoleyNV:
Fixed by #282.

PR #6690, opened 2025-03-26
fix(d3d11): correct parameter in VSSetConstantBuffers1 from uavCount …

### LINKED ISSUE #6531: [Regression] slang-rhi update causes rendering tests to fail
One of recent PRs causes a regression,
- https://github.com/shader-slang/slang/pull/6251

The following slang-tests are failing but it reports as passing,
- tests/render/cross-compile-entry-point.slang
- tests/render/cross-compile0.hlsl
- tests/render/imported-parameters.hlsl
- tests/render/multiple-stage-io-locations.slang
- tests/render/nointerpolation.hlsl
- tests/render/render0.hlsl

Those tests are supposed to render a triangle and store the result in "*.expected.png" file.
Currently slang-test generates those files with an empty black image with white background.
They should render a triangle.

Note that the current Github workflow CI is green, because both "*.expected.png" and "*.actual.png" are blank images incorrectly.

szihs:
fix(d3d11): correct parameter in VSSetConstantBuffers1 from uavCount to cbvCount (fixes #6531)
Root cause - Incorrect parameter passing in slang-rhi

1. slang-rhi #281 - Add the correct cbvCount for setting Constant Buffer
2. Prevent render tests from overwriting reference images
    * Add missing tests/render/multiple-stage-io-locations.slang.3.expected.png


jkwak-work (commented):
I think we should keep the change in slang-rhi but drop the rest changes that are trying to skip the generation of expected images.

jkwak-work (approved):
After offline discussion, I agree to skip the reference generation step for CI workflow.

PR #6693, opened 2025-03-26
Use correct syntax for WGSL array transpiling

jarcherNV:
This patch updates handling of Array and ConstantBuffer types for WGSL transpiling, giving correct syntax for arrays of buffers in WGSL.


jkwak-work (approved):
Looks good to me.

Issue #6648, opened 2025-03-19
Excessive storage during reverse-mode AD over nested loops


saipraveenb25:
Nested loops can cause several arrays to be allocated to store the 'end-state' of the loop counter of the inner loops, even in cases where the loops iteration count can be statically deduced. 

E.g.:
```csharp
for (uint i = 0; i < 15; i++)
{
    for (uint j = 0; j < 10; j++)
    { 
           /* .. simple body with no breaks and no modifications on i or j .. */
    }
}
```

Under reverse-mode AD, this loop will still store an array of size 15 (+2) to store the last state of the "j" for each of the 15 iterations of the outer loop. 

Our inductive value detection pass _does_ work here and deduces that "j" and "i" are inductive, but it says nothing about the value of "j" and "i" at the time of exiting the loop. 
We need to extend our current pass to also detect cases where we can say something definite about the value of loop state parameters at exit. In the above case, we should be able to detect the invariant that "j = 10" and "i = 15" at loop exit.

This should allow simple nested loops to work seamlessly (which should cover 95% of use-cases)

saipraveenb25:
Current approach is to implement a pass very similar to the `collectInductionValue`. That is, try to construct a "proof" by walking the predecessors of the loop condition block. In this case, we'll try to prove that the loop's break condition is true if and only if some relation holds true for a given phi param. 

E.g. in the sample loop above, we should be able to prove that the `breakFlag == true` **if and only if** `(i < 15) == false`. We can then combine this fact with the inductive value proof to state that "i == 15" at loop exit.

This involves implementing a (minimal) subset of relational algebra, so we can reason about things like predicates and negation, but for now it will be very minimal. If something is too hard to prove, we can always fall back to storing the loop exit value, and 99% of the cases can be detected with only a handful of rules.

PR #6696, opened 2025-03-26
Add a loop analysis step to infer the exit values of loop phi parameters.

saipraveenb25:
This PR adds a static analysis system that can infer simple inequalities on the loop phi parameters that must be true when the loop stops iterating (i.e. breaks)

The algorithm used in this system is fairly simplistic: 
- The goal is to compute a relationship between the loop's break flag (which is false when the loop is ready to exit), and the phi parameters. E.g. relations of the form `~breakFlag => (i >= 3)`, where `=>` is the "implies" operator from propositional calculus.
- We generalize this with a pass that can computes all such "implications" (such as `i >= 3`) that are true for a given block, under a given predicate relation (e.g.`breakFlag == true` or `breakFlag == false`)
- This pass operates on the principle that if we can discover a statement for the predecessor block such that `A' => B'`, and we can show that `A=>A'` (where `A` is the predicate for this block), then we can assert that `B'` is also true for the successor block.
- Based on this principle, our algorithm has an 'upward' phase where predicates `A` are propagated to predecessors, and then a 'downward' phase, where implications `B` are propagated downward.
- Since static analysis is an undecidable problem, to avoid exponential complexity, we compute a single predicate per block. If there are multiple successors for a block, we take the disjunction (i.e. if a block has `A1` and `A2` predicates from two successors, it must compute a `B` such that `A1 v A2 -> B` (where `v` is the disjunction operator, aka "OR")

To support this system, we have a `SimpleRelation` struct that can represent a single innequality or equality statement on integer or boolean types. Additionally, `SimpleRelation` can also be `Any` or `Impossible` to represent the fact that we can't say anything about a parameter, or that a certain situation is impossible (i.e. often this denotes a branch that cannot be taken under the given predicate, and can be ignored).

Additionally, we have helper functions that build a simple calculus around `SimpleRelation` (i.e. conjunction, disjunction and implication) and a helper struct called `StatementSet` that holds one relation per inst and represents a conjunction of independent statements.

This system allows us to infer loop bounds **exactly** (**not** just an upper bound, but an exact count of the loop iterations), and removes the need to store the final value of loops. 
This is particularly important for nested loops, since the inner loop no longer needs to store its final value into an array (as long as it is statically determinable!). This should lead to dramatically faster nested loop systems.

Fixes: #6648


source/slang/slang-ir-autodiff-primal-hoist.cpp
```diff
@@ -1046,6 +1063,174 @@ void AutodiffCheckpointPolicyBase::collectInductionValues(IRGlobalValueWithCode*
...
+                            builder.getIntValue(loopCounter->getDataType(), exitIValue);
+                    }
+                };
+
+                if (counterFactor > 0 && statement.comparator == SimpleRelation::GreaterThanEqual)
+                {
+                    // Find the smallest value that satisfies counterFactor * i + counterOffset >=
+                    // relationValue
+                    //
+                    IRIntegerValue exitIValue =
```
csyonghe, line 1212: This kind of computation without considering the original type of the induction variable can be dangerous because it doesn't factor in possible int overflow behavior. We need to take extra care to make sure we are handling things like int8_t arithmetic correctly.

source/slang/slang-ir-autodiff-loop-analysis.h
```diff
@@ -0,0 +1,181 @@
...
+
+#include "slang-ir-autodiff-region.h"
+#include "slang-ir-autodiff.h"
+#include "slang-ir-dominators.h"
+#include "slang-ir-insts.h"
+#include "slang-ir.h"
+
+namespace Slang
+{
+struct SimpleRelation
```
csyonghe, line 12: Can we handle the case where the right hand side of a relation is another variable and not just a constant?

csyonghe:
We also need test cases where there are `continue` and `break` inside the loops.

source/slang/slang-ir-autodiff-loop-analysis.h
saipraveenb25: Yes, it should be possible to extend the current system. 
Basically, we need a way to do 'substitution' on both sides as we move propagate through the blocks and phi parameters have to be translated into their args. Right now, we assume that each relation references only one inst.

I think for now, this covers a lot of cases, but I agree that we should extend this to support those cases too. Preferably in a follow-up PR, since this one has gotten rather large?

saipraveenb25:
Added more tests that use `continue`
For now, I'm not adding tests that use `break` since we won't usually be able to infer anything useful in these cases. I think this is fine, since the type of loops we are looking to analyse are simple "for" loops without additional break logic.

csyonghe (approved):

PR #6701, opened 2025-03-29
Fix compilation of global builtin variables inside generics

### LINKED ISSUE #6501: GLSL builtin variables do not work inside generics
GLSL builtin variables, i.e. global variables with in/out decorations that supposedly contains `IRGlobalInputDecoration` and is legalized through `translateGlobalVaryingVar` do not properly emit system value/builtin entry point param and store operation; the global variables are not initialized.

Consider this shader:
```

T getValue<T: IInteger>(T value)
{
    return T(gl_GlobalInvocationID.x) + value;
}

[shader("compute")]
void computeMain()
{
    outputBuffer[0U] = getValue(100U);
}
```
The WGSL output is
```
@binding(0) @group(0) var<storage, read_write> outputBuffer_0 : array<u32>;

var<private> gl_GlobalInvocationID_0 : vec3<u32>;

fn getValue_0( value_0 : u32) -> u32
{
    return u32(i64(gl_GlobalInvocationID_0.x)) + value_0;
}

@compute
@workgroup_size(1, 1, 1)
fn computeMain()
{
    outputBuffer_0[u32(0)] = getValue_0(u32(100));
    return;
}
```
Note `gl_GlobalInvocationID_0` is not initialized properly and the entry point is missing the builtin parameter. Similar behavior can be seen when compiling for Metal and HLSL. The output shader is correct without generics.

The issue seems to be caused by `IRGlobalInputDecoration` not being added/transferred/copied properly or accidentally when lowering  generics, which causes the global var inst to not be detected during `translateGlobalVaryingVar`.

fairywreath:
Closes #6501.

Global builtin variables inside generic functions currently fail to compile properly - they are not added as entry point parameter list for targets that require it (HLSL, Metal and WGSL).

The logic that handles adding global builtin variables to the entry point parameter list  is done inside `translateGlobalVaryingVar`.  A call graph is constructed to obtain the referencing entry points for each global variable. This constructed call graph however does not include generic functions, hence builtin global variables inside generics  return no referenced entry points inside `translateGlobalVaryingVar` and are not added to entry point parameter list. The fix here is to include generics when constructing the call graph, inside `buildEntryPointReferenceGraph`.

### DIFF
--- a/source/slang/slang-ir-call-graph.cpp
+++ b/source/slang/slang-ir-call-graph.cpp
@@ -88,6 +88,7 @@ void buildEntryPointReferenceGraph(
             case kIROp_GlobalParam:
             case kIROp_GlobalVar:
             case kIROp_SPIRVAsmOperandBuiltinVar:
+            case kIROp_Generic:
                 addToWorkList({entryPoint, operand});
                 break;
             }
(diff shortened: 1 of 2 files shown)


csyonghe (approved):
Looks good, thank you for the fix!

PR #6702, opened 2025-03-30
Add sparse texture Load intrinsic for SPIRV

DarrelFW321:
Implements sparse texture.load intrinsics for SPIRV. Uses the existing SPIRV code for texture loads and sparse texture samples as reference code.


tests/hlsl-intrinsic/texture/partial-resident-texture-combined.slang
```diff
@@ -177,6 +180,25 @@ bool TEST_sampler<T>(
         ;
 }
 
+bool TEST_load<T>(
```
jkwak-work, line 183: Can you rename this to reflect that it is for the partial resident texture testing?
Something like TEST_sparse?

jkwak-work (dismissed):
Looks good to me.

fairywreath (approved):

PR #6712, opened 2025-03-31
Use GITHUB_TOKEN for fetching prebuilt

### LINKED ISSUE #6683: Slang build fails to download webgpu-dawn and slang-tint

gtong-nv:
This PR extends Commit c6b702c to use GITHUB_TOKEN if set for fetching prebuilt binaries.
This allows webgpu-dawn and slang-tint to be downloaded for certain IPs where the github API rate is limited.

Fixes #6683


gtong-nv:
This fixes the github downloading issue.
We still need to update the file download logic in the slang-rhi repo, or have to turn off `SLANG_ENABLE_SLANG_RHI`. But this is out of scope of this PR

sricker-nvidia (commented):
Looks ok to me, but I'll give others a chance to review as I've only just started taking a look at PRs.

CMakeLists.txt
```diff
@@ -375,6 +375,7 @@ if(SLANG_WEBGPU_DAWN_BINARY_URL)
         "webgpu_dawn"
         "${SLANG_WEBGPU_DAWN_BINARY_URL}"
         IGNORE_FAILURE
+        SLANG_GITHUB_TOKEN ${SLANG_GITHUB_TOKEN}
```
jarcherNV, line 378: This looks like the kind of thing I had in mind, but I'm wondering if we should now not ignore the failure if SLANG_GITHUB_TOKEN is defined? I'm thinking that people would probably intend for this to be built if they provide the token so they may want to see the error if one appears.

Perhaps we could throw in yet another if statement to check if the token is defined, then call the function without the IGNORE_FAILURE arg if it is.

gtong-nv: I updated the [FetchedSharedLibrary.cmake](https://github.com/shader-slang/slang/pull/6712/files#diff-74a95d679f648bc9ea4e8345449ad666ce8183d6a55e1285283a7b396c168024) to include this check. Eventually we should drop IGNORE_FAILURE when everything is stable... probably when we find a better storage provider for hosting these prebuilt binaries.

jarcherNV (approved):
LGTM

jkwak-work (approved):
Looks good to me.

sricker-nvidia (dismissed):
LGTM

cmake/FetchedSharedLibrary.cmake
```diff
@@ -56,7 +70,11 @@ endfunction()
 # for the shared library to produce the install rule.
 function(copy_fetched_shared_library library_name url)
     cmake_parse_arguments(ARG "IGNORE_FAILURE" "" "" ${ARGN})
-    if(ARG_IGNORE_FAILURE)
+    cmake_parse_arguments(ARG "" "SLANG_GITHUB_TOKEN" "" ${ARGN})
```
expipiplus1, line 73: It's unnecessary to use multiple calls to cmake_parse_arguments here, these can be condensed :)

expipiplus1 (changes requested):

cmake/FetchedSharedLibrary.cmake
gtong-nv: Thanks! Updated it with a single cmake_parse_arguments call

gtong-nv:
\format

gtong-nv:
Hi @expipiplus1 , can I get another look at this PR? Thanks!

jkwak-work (approved):
Looks good to me

Issue #5660, opened 2024-11-25
importing a file and including it as a cli arguments leads to bad error message


expipiplus1:
`nullType.slang`:
```slang
import nullTypeModule;

struct S { }

extension S : I { }
```

`nullTypeModule.slang`:
```slang
interface I { } 
```

`slangc -stage compute tests/bugs/nullTypeModule.slang tests/bugs/nullType.slang`

```
tests/bugs/nullType.slang(5): error 39999: ambiguous reference to 'I'
extension S : I { }
              ^
tests/bugs/nullTypeModule.slang(1): note 39999: candidate: interface I
interface I { }
          ^
tests/bugs/nullTypeModule.slang(1): note 39999: candidate: interface I
interface I { }
          ^
```

This "ambiguous" reference refers to the same name

bmillsNV:
@aleino-nv can you help to verify if this issue still occurs? We think you might have fixed something similiar.

aleino-nv:
> [@aleino-nv](https://github.com/aleino-nv) can you help to verify if this issue still occurs? We think you might have fixed something similiar.

Yes, this seems similar to https://github.com/shader-slang/slang/issues/6221. I'll take a look.

aleino-nv:
It still happens today on master (133d705dcaa69f10f7a713605d5080405e49875c), even after the #6221 fix:

```
C:\Users\aleino\Downloads>C:\src\slang\build\Debug\bin\slangc.exe -stage compute nullTypeModule.slang nullType.slang   || exit 1 
nullType.slang(5): error 39999: ambiguous reference to 'I'
extension S : I { }
              ^
nullTypeModule.slang(1): note 39999: candidate: interface I
interface I { }
          ^
nullTypeModule.slang(1): note 39999: candidate: interface I
interface I { }
          ^
(0): error 30015: undefined identifier 'main'.
nullTypeModule.slang(1): error 38000: no function found matching entry point name 'main'
interface I { }
^~~~~~~~~
```

aleino-nv:
Assigning to @bmillsNV, since I probably won't have time to investigate this sprint.

kaizhangNV:
@expipiplus1 
I don't quite understand why we should compile in this way?

```
slangc -stage compute nullTypeModule.slang nullType.slang
```

we should only compile `nullType.slang`, because `nullTypeModule.slang` is imported to `nullType.slang` in the source code.

so I don't know if this is actually legal?
```
[...]

PR #6713, opened 2025-04-01
Report an error if a module imports itself

jarcherNV:
Fixes issue #5660

Undefined behavior was reported when files in a given module attempted to import that module. Now when a module imports itself we report error 38200 for a recursive import.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -10499,19 +10499,30 @@ void SemanticsDeclHeaderVisitor::visitImportDecl(ImportDecl* decl)
     // the module's scope.
 
     auto name = decl->moduleNameAndLoc.name;
+    auto loc = decl->moduleNameAndLoc.loc;
     auto scope = getModuleDecl(decl)->ownedScope;
 
     // Try to load a module matching the name
     auto importedModule = findOrImportModule(
         getLinkage(),
         name,
-        decl->moduleNameAndLoc.loc,
+        loc,
         getSink(),
         getShared()->m_environmentModules);
 
     // If we didn't find a matching module, then bail out
     if (!importedModule)
         return;
+
+    // If we're attempting to import a module into itself then give a
+    // warning.
+    if (auto parentDecl = getModuleDecl(decl->parentDecl))
+    {
+        if (parentDecl && getText(parentDecl->getName()) == name->text)
+        {
+            getSink()->diagnose(loc, Diagnostics::moduleSelfImport, name);
+        }
+    }
 
     // Record the module that was imported, so that we can use
     // it later during code generation.
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -1927,6 +1927,8 @@ DIAGNOSTIC(
     vectorWithDisallowedElementTypeEncountered,
     "vector with disallowed element type '$0' encountered")
 
+DIAGNOSTIC(38204, Warning, moduleSelfImport, "module '$0' imported more than once")
+
 // 39xxx - Type layout and parameter binding.
 
 DIAGNOSTIC(
--- a/tests/diagnostics/self-import-extra.slang
+++ b/tests/diagnostics/self-import-extra.slang
@@ -1,0 +1,8 @@
+//DIAGNOSTIC_TEST:SIMPLE(filecheck=CHECK): tests/diagnostics/self-import.slang -stage compute
+
+// This test compiles two Slang shaders via the command line, where the second
+// shader imports this one. This should produce a warning.
+
+interface I { }
+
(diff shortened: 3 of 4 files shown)


source/slang/slang-check-decl.cpp
```diff
@@ -10499,20 +10499,32 @@ void SemanticsDeclHeaderVisitor::visitImportDecl(ImportDecl* decl)
...
 
     // If we didn't find a matching module, then bail out
     if (!importedModule)
         return;
 
+    // If we're attempting to import a module into itself then bail out
+    // and give an error.
+    if (decl->parentDecl)
+    {
+        ModuleDecl* parentDecl = getModuleDecl(decl->parentDecl);
```
kaizhangNV, line 10521: please double check can this simplify to 

if (auto parentDecl = getModuleDecl(decl->parentDecl))
{
    ...
}

source/slang/slang-check-decl.cpp
```diff
@@ -10499,20 +10499,32 @@ void SemanticsDeclHeaderVisitor::visitImportDecl(ImportDecl* decl)
...
+        if (parentDecl && parentDecl->getName()->text == name->text)
```
kaizhangNV, line 10522: nullptr check on `parentDecl->getName()`.
use `getText(parentDecl->getName()) ` instead.

tests/diagnostics/self-import-extra.slang
```diff
@@ -0,0 +1,8 @@
+//DIAGNOSTIC_TEST:SIMPLE(filecheck=CHECK): tests/diagnostics/self-import.slang -stage compute
+
+// This test compiles two Slang shaders via the command line, where the second
+// shader imports this one. This should produce a warning.
+
+interface I { }
+
+// CHECK: warning 38204
```
kaizhangNV, line 8: can you also add warning message here for matching.

source/slang/slang-check-decl.cpp
```diff
@@ -10499,20 +10499,32 @@ void SemanticsDeclHeaderVisitor::visitImportDecl(ImportDecl* decl)
...
+        loc,
         getSink(),
         getShared()->m_environmentModules);
 
     // If we didn't find a matching module, then bail out
     if (!importedModule)
         return;
 
+    // If we're attempting to import a module into itself then bail out
+    // and give an error.
```
kaizhangNV, line 10518: error -> warning

source/slang/slang-diagnostic-defs.h
```diff
@@ -1927,6 +1927,8 @@ DIAGNOSTIC(
     vectorWithDisallowedElementTypeEncountered,
     "vector with disallowed element type '$0' encountered")
 
+DIAGNOSTIC(38204, Warning, moduleSelfImport, "module '$0' imports itself")
```
kaizhangNV, line 1930: should we change the warning message to something like "module xxx are imported more than once"?

I see the original issue is 
`slangc A.slang B.slang`

where `B.slang` import `A.slang`, so it's really that we import `A.slang` more than once.

csyonghe: We should clarify what it means when you specify more than one filenames to slangc. It could mean two things:

1. each file represents a slang module, and we create a composite component type that includes all the modules.
2. Each file is just one file in the same module, and the result is a single module that is parsed from all the specified files.

I think it makes more sense for slangc to assume case 1, because in modern slang, a slang module always specifies all the files via __include. In this case slangc A.slang b.slang where A imports B shouldn’t be an error, and the module B just got deduplicated when we create the composite component type.

source/slang/slang-check-decl.cpp
jarcherNV: I have changed this to use getText as you suggest, which has its own internal nullptr check which should be good enough for this.

tests/diagnostics/self-import-extra.slang
jarcherNV: Done.

source/slang/slang-check-decl.cpp
jarcherNV: Done.

source/slang/slang-diagnostic-defs.h
jarcherNV: So should we instead not issue a warning, and create separate modules for each input file?

csyonghe: yes, I think this is what we want to move to. Note that if we implement this now it will be a breaking change. We should make it very clear in our documentation what this means.

jarcherNV:
This PR has become stale as I've not had time to come back to this, and it sounds like we want a very different solution to the one  prepared here. So that I'm not cluttering up the list of open PRs I am going to close this and create a new one when I have a chance to look at the issue again.

PR #6719, opened 2025-04-01
Use coopvec supporting dxcompiler.dll and dxil.dll

### LINKED ISSUE #4919: Upgrade packaged DXC to 1.8 or greater
The current dxcompiler.dll is version 1.7, but TOT is >= 1.8.

TOT has been observed to contain additional validation that caught the linking of DXIL libraries with duplicate function definitions.

https://github.com/microsoft/DirectXShaderCompiler/releases

Slang should upgrade.

jkwak-work:
Closes #4919
Related to #6250 

This commit upgrades DXC version from 1.7 to 1.9 that supports CoopVector.

The DXC is compiled from a [public repo](https://github.com/NVIDIA-RTX/DirectXShaderCompiler/tree/CooperativeVector), and it uses `MinSizeRel` build config.

The version information from a command "dxc.exe --help":

Version: dxcompiler.dll: 1.9 - 1.8.0.4817 (CooperativeVector, 018d767f9); dxil.dll: 1.9(1.8.0.4817)

### DIFF
--- a/tests/hlsl-intrinsic/ray-tracing/rt-pipeline-intrinsics-ahit.slang
+++ b/tests/hlsl-intrinsic/ray-tracing/rt-pipeline-intrinsics-ahit.slang
@@ -47,15 +47,13 @@
 // DXIL: call i32 @dx.op.instanceID.i32
 // DXIL: call i32 @dx.op.geometryIndex.i32
 // DXIL: call i32 @dx.op.primitiveIndex.i32
+
 // DXIL: call float @dx.op.objectRayOrigin.f32
 // DXIL: call float @dx.op.objectRayDirection.f32
 // DXIL: call float @dx.op.objectToWorld.f32
 // DXIL: call float @dx.op.worldToObject.f32
 // DXIL: call i32 @dx.op.hitKind.i32
 
-// DXIL: call float @dx.op.objectRayOrigin.f32
-// DXIL: call float @dx.op.objectRayDirection.f32
-// DXIL: call float @dx.op.rayTCurrent.f32
 // DXIL: call void @dx.op.acceptHitAndEndSearch
 // DXIL: call void @dx.op.ignoreHit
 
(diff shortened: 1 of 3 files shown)


jkwak-work:
Note that there are many modifications in slang-binary submodule in this PR.
But most of them are unrelated to the upgrade of DXC version.

We have been using an out-dated commit hash of slang-binary submodule.
Probably we forgot to update the hash value in external/slang-binary when we made changes to master branch in slang-binary repo.

tests/hlsl-intrinsic/ray-tracing/rt-pipeline-intrinsics-ahit.slang
jkwak-work, line 1: The changes in this file is because the different versions of DXC generates slightly different DXIL code.

tests/metal/texture.slang
jkwak-work, line 1: DXC 1.9 requires the thread size to be at least 2x2 when the ddx/ddy feature is used in compute shader.

external/slang-binaries
jkwak-work, line 1: The only real change in this PR for slang-binaries is on bin/windows-x64/dxcompiler.dll and dxil.dll
The rest are from previous changes made to the submodule.

sricker-nvidia: Interesting, do you know why the PR diff treats external/slang-binaries as a single file and then lists 40 file diffs as part of that file diff?

Also wondering if it's worth mentioning the previous changes made to the submodule that got picked up in the commit message, but otherwise LGTM.

aidanfnv (commented):
LGTM

external/slang-binaries
jkwak-work: There are a few "submodules" that Slang repo uses. And "external/slang-binaries" is one of them.
Each submodule is treated as a single hash value, which represents an entire repo.
Github system is nice enough to visualize what the hash value difference actually means in terms of the file differences.

Some of the changes picked up from the master branch of slang-binaries are the binary executables of spirv-tools that we no longer use and workflow settings that has nothing to do with slang repo. I ignored the details because they are too trivial.

sricker-nvidia: Make sense, thanks for the details!

Issue #6138, opened 2025-01-20
Improve using generics to select float precision and array sizes.


tomas-davidovic:
I am trying to turn out MLPs from hardcoded values and defines to a more generics based solution with interface like this:
```
interface IMLP
{
    associatedtype PrecisionType : __BuiltinFloatingPointType; // precision of the IO, can be float or half
    static const uint kInputWidth;
    static const uint kOutputWidth;
    void eval(PrecisionType input[kInputWidth], out PrecisionType output[kOutputWidth]); // other parameters not relevant
}
```

This is then used in this manner:
```
void decorate_and_eval<MLP : IMLP, let TLatentSizes : uint>(
    float latents[TLatentSizes],
    out float result[MLP::kOutputWidth],
    out float resultExponentiated[MLP::kOutputWidth]
)
{
    MLP::PrecisionType inputs[MLP::kInputWidth];
    for (uint i = 0; i < TLatentSizes; ++i)
    {
        inputs[i] = latents[i]; // This does not work, it won't cast, even though it can be cast with any conforming type
        inputs[i] = (MLP::PrecisionType)latents[i]; // requires explicit cast
    }

    for (uint i = TLatentSizes; i < MLP::kInputWidth; ++i)
    {
        inputs[i] = 0.f;                     // This does not work either
        inputs[i] = (MLP::PrecisionType)0.f; // must have an explicit cast
    }

    MLP::PrecisionType outputs[MLP::kOutputWidth] = {}; // this does not work, says the size is not statically known.

    // Must do this explicit thing as a workaound to get equivalent result
    [ForceUnroll]
    for (uint i = 0; i < MLP::kOutputWidth; ++i)
        outputs[i] = MLP::PrecisionType(0.f);

    MLP mlp;
    mlp.eval(inputs, outputs);

    for (uint i = 0; i < MLP::kOutputWidth; ++i)
    {
        result[i] = outputs[i];                    // this does not work
        result[i] = (float)outputs[i];             // nor does this
        result[i] = float(outputs[i]);             // nor this
        result[i] = __realCast<float>(outputs[i]); // must do this
    }

    for (uint i = 0; i < MLP::kOutputWidth; ++i)
    {
        resultExponentiated[i] = exp(outputs[i]); // this does not work, despite exp having generic arg of type __BuiltinFloatingPointType
        resultExponentiated[i] = exp(__realCast<float>(outputs[i])); // if PrecisionType is half, this is unnecessarily expensive
[...]

PR #6720, opened 2025-04-01
Support Array Sizes using Generic arguments to be initialized via {}

psruthikNV:
Support Array Sizes using Generic arguments to be initialized via {}

Fixes one subissue of #6138

This change adds support for initializing Arrays with Generic size arguments via {}


source/slang/slang-check-conversion.cpp
```diff
@@ -613,17 +613,27 @@ bool SemanticsVisitor::_readAggregateValueFromInitializerList(
...
-                if (outToExpr)
+                if (auto typeCastElementCount = as<TypeCastIntVal>(toElementCount))
                 {
-                    getSink()->diagnose(
-                        fromInitializerListExpr,
-                        Diagnostics::cannotUseInitializerListForArrayOfUnknownSize,
-                        toElementCount);
+                    if (auto genericElementCount = as<GenericParamIntVal>(typeCastElementCount->getBase()))
+                    {
+                        elementCount = 0;
```
jkwak-work, line 620: I am not familiar with this part of source code.
Can you give us some explanations of how this code resolves the problem?

psruthikNV: Ack. IIUC GenericParamIntVal types get their concrete values post initialization.

So in my change, I just check if the input array size was of a GenericParamIntVal type and if yes, do not consider this as the element count not being specified or statically unknown.

What I'm not sure about is setting elementCount to zero here for GenericParamIntVal Array sizes.

expipiplus1 (changes requested):
Yeah, it's definitely quite suspicious to just be inventing the array size `0` out of thin air, if it's true that these are statically known then we should organize things such that we can get that value when it comes time to handle the initializer list.

source/slang/slang-check-conversion.cpp
psruthikNV: I've cleaned up the code a bit so that we don't initialise elementCount to 0 redundantly.

source/slang/slang-check-conversion.cpp
```diff
@@ -611,7 +611,7 @@ bool SemanticsVisitor::_readAggregateValueFromInitializerList(
             {
                 elementCount = (UInt)constElementCount->getValue();
             }
-            else
+            else if (!as<GenericParamIntVal>(as<TypeCastIntVal>(toElementCount)->getBase()))
```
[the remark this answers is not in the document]
psruthikNV: Done

source/slang/slang-ast-expr.h
```diff
@@ -139,6 +139,12 @@ class StringLiteralExpr : public LiteralExpr
     FIDDLE() String value;
 };
 
+FIDDLE()
+class ArrayWithGenericSizeExpr : public Expr
```
csyonghe, line 143: This naming is confusing. The Expr node represents an option that is to create an array from a single element value. It doesn't matter if the array type has generic size or not. This should just named `MakeArrayFromElementExpr`.

psruthikNV: Done

expipiplus1 (approved):

csyonghe (commented):
This PR doesn’t handle the case where the element type of the array doesn’t have a default value, in which case we should be issuing an error.


For example,

struct X{ int v; __init(int c){v=c;}}
// if struct has user defined ctor but not user defined default ctor, the type won’t have a default value.

X arr[3] ={}; // should error
X arr[] = {}; // should error

source/slang/slang-check-conversion.cpp
```diff
@@ -613,19 +613,13 @@ bool SemanticsVisitor::_readAggregateValueFromInitializerList(
...
-                if (outToExpr)
-                {
-                    getSink()->diagnose(
-                        fromInitializerListExpr,
-                        Diagnostics::cannotUseInitializerListForArrayOfUnknownSize,
-                        toElementCount);
-                }
-                return false;
-            }
+                auto toMakeArrayFromElementExpr = m_astBuilder->create<MakeArrayFromElementExpr>();
```
csyonghe, line 616: This expr should store a valueExpr that represents the default value of Element style. If we cannot for. That valueExpr, then we need to issue an error.

psruthikNV: @csyonghe I'm planning on addressing this in the below fashion:

- If we cannot get a default value for a particular Element style, my understanding is that getDefaultVal() will return a LoweredValInfo with LoweredValInfo::val set to nullptr
- In the visitor implementation of MakeArrayFromElementExpr, we add a check for the above nullptr and if such a check gets triggered, we raise a SLANG_UNEXPECTED error stating that the default element for the element cannot be constructed.

Does this make sense?

csyonghe: First of all, you can’t call anything that returns LoweredValInfo from the semantic visitor, if. Function returns LoweredValInfo, it belongs to IR lowering, not semantic checking. You will need to find something else that checks if a type is default constructable. Check with @kaizhangNV on what you need to call.

second,  we should never do SLANG_UNEXPECTED on anything that the user can write. It should be a proper diagnostic output.

PR #6711, opened 2025-03-31
Revert "Eliminate empty struct on metal target (#6603)"

jkwak-work:
This reverts commit b3deec2001ea34e20e9a6af8ddf5cf3866cafac0.

The commit causes CI failures and I cannot come up with an easy fix for it.
Reverting it for now.

PR #6725, opened 2025-04-01
Metal remove void field

kaizhangNV:
Remove argument in make_struct call corresponding to void field

    
- This is a follow-up of #6543, where we leave the VoidType field as it in
  make_struct call during legalization pass.
    
  So during cleaning_void IR pass, when we remove "VoidType" from struct,
  we will have to also clean up the argument corresponding to the
  "VoidType" field.

- Reapply "Eliminate empty struct on metal target (#6603)" which is reverted in (#6711)
  Because #6603 is not the root casuse. #6543 is.

### DIFF
--- a/source/slang/slang-legalize-types.cpp
+++ b/source/slang/slang-legalize-types.cpp
@@ -1211,11 +1211,15 @@ LegalType legalizeTypeImpl(TypeLegalizationContext* context, IRType* type)
         LegalType legalElementType;
 
         if (isMetalTarget(context->targetProgram->getTargetReq()) &&
-            as<IRParameterBlockType>(uniformBufferType))
+            as<IRParameterBlockType>(uniformBufferType) &&
+            !context->shouldLegalizeParameterBlockElementType())
         {
             // On Metal, we do not need to legalize the element type of
             // a parameter block because we can translate it directly into
             // an argument buffer.
+            //
+            // But we do need empty type legalized for Metal, because Metal doesn't
+            // allow empty struct in argument buffer.
             legalElementType = LegalType::simple(originalElementType);
         }
         else
--- a/source/slang/slang-legalize-types.h
+++ b/source/slang/slang-legalize-types.h
@@ -660,6 +660,10 @@ struct IRTypeLegalizationContext
         LegalType legalElementType,
         IRInst* layoutOperand) = 0;
 
+    /// Customization point to decide whether a parameter block type should be legalized.
+    ///
+    /// This function is called in `legalizeTypeImpl` to decide whether a parameter block
+    /// type should be legalized. Not all legalization passes need to legalize parameter block.
     virtual bool shouldLegalizeParameterBlockElementType() { return false; }
 };
 
--- a/source/slang/slang-legalize-types.h
+++ b/source/slang/slang-legalize-types.h
@@ -659,6 +659,8 @@ struct IRTypeLegalizationContext
         IROp op,
         LegalType legalElementType,
         IRInst* layoutOperand) = 0;
+
+    virtual bool shouldLegalizeParameterBlockElementType() { return false; }
 };
 
 // This typedef exists to support pre-existing code from when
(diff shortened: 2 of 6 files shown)


kaizhangNV:
The only actual change in this PR is in this file: source/slang/slang-ir-cleanup-void.cpp

tests/metal/empty-struct-remove.slang
```diff
@@ -0,0 +1,33 @@
+
+//TEST:SIMPLE(filecheck=LIB):-target metallib -entry computeMain -stage compute -DMETAL
+//TEST:SIMPLE(filecheck=METAL):-target metal -entry computeMain -stage compute -DMETAL
```
jkwak-work, line 3: I don't think `-DMETAL` is needed here, because this file doesn't use the macro.
I guess this lines are copied from another test case where the test uses "METAL" macro in a way like `#ifdef METAL`

source/slang/slang-legalize-types.h
```diff
@@ -659,6 +659,8 @@ struct IRTypeLegalizationContext
         IROp op,
         LegalType legalElementType,
         IRInst* layoutOperand) = 0;
+
+    virtual bool shouldLegalizeParameterBlockElementType() { return false; }
```
jkwak-work, line 663: I think this function deserves a comment that describes what it is for.
The other functions have explanations.

source/slang/slang-legalize-types.cpp
```diff
@@ -1211,11 +1211,15 @@ LegalType legalizeTypeImpl(TypeLegalizationContext* context, IRType* type)
         LegalType legalElementType;
 
         if (isMetalTarget(context->targetProgram->getTargetReq()) &&
-            as<IRParameterBlockType>(uniformBufferType))
+            as<IRParameterBlockType>(uniformBufferType) &&
+            !context->shouldLegalizeParameterBlockElementType())
         {
```
jkwak-work, line 1216: This seems little confusing to me.
It reads as "isMetalTarget && .. && ! isMetalTarget".
It looks like it has two conflicting conditions.

In what situation this condition is expected to be true??

kaizhangNV: why do you read as this ? 
"isMetalTarget && .. && ! isMetalTarget"


it's 

```
isMetalTarget &&
!context->shouldLegalizeParameterBlockElementType()
```

tests/metal/empty-struct-remove.slang
kaizhangNV: updated.

source/slang/slang-legalize-types.h
kaizhangNV: done

source/slang/slang-legalize-types.cpp
jkwak-work: That part is what's little confusing to me.
I see that
```
    virtual bool shouldLegalizeParameterBlockElementType() override
    {
        return isMetalTarget(targetProgram->getTargetReq());
    }
```
So it reads as "isMetalTarget && !isMetalTarget", but probably that's not the case.

kaizhangNV: This is only for one pass, but the other passes, this does not necessarily depends on metal.
The `TypeLegalizationContext` is an abstract class, that all the type legalization passes inherit from it. So only when it's the empty type legalization pass, we will return true for metal, and we do want to legalize parameter block for metal, so it will not hit this if condition (Note this if condition means that we don't want to legalize a parameter block as it simple return a `simple` legal vale). But for other type legalization passes, e.g. resource/existantial, we always return `false` for parameter block.

This is the pattern that how to legalize different aspects of a type for different targets.

kaizhangNV:
I Just rebase my PR to re-run the CI because I merged a new change to enable MacOS CI test https://github.com/shader-slang/slang/pull/6731

PR #6726, opened 2025-04-01
Enable coopvec tests with dxc backend

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6250

This PR enables existing CoopVec tests with DX12 backend.
In order to use the CoopVec feature with DX12 backend, we have to use an option, "-dx12-experimental", because the current implementation of CoopVec feature in dxcompiler.dll requires "experimental feature".

Note that when the "experimental feature" is enabled, slang-test becomes less stable.
For that reason, we should use the option "-dx12-experimental" only when it is needed.

All tests for GLSL are deleted because CoopVec support for GLSL in Slang is deprecated and no longer supported.

Some of CoopVec tests are still disabled for DX12 backend because:
 1. DXC doesn't support 8bit integer types and
 2. Some of CoopVec features are not implemented in DXC backend.


sricker-nvidia (approved):
LGTM

Issue #6727, opened 2025-04-01
Fix wasm build error


jarcherNV:
The following error has been reported at ToT from the wasm build:

CPack: Create package using ZIP
CPack: Install projects
CPack: - Install project: slang [Release]
CPack: -   Install component: generators
CPack: Create package
CPack: - package: /home/runner/work/slang/slang/build/dist-release/slang-generators.zip generated.
CMake Error at build/tools/cmake_install.cmake:53 (file):
  file INSTALL cannot find
  "/home/runner/work/slang/slang/build/generators/Debug/bin/slang-cpp-extractor":
  No such file or directory.
Call Stack (most recent call first):
  build/cmake_install.cmake:77 (include)

We should fix this as it is causing failures in PRs at CI / build (linux, release, gcc, wasm) (pull_request)

PR #6730, opened 2025-04-02
Fix the broken workflow for linux/release/gcc/wasm

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6727


jarcherNV (approved):
LGTM

PR #6735, opened 2025-04-02
Return non-escaped strings from user defined attributes

### LINKED ISSUE #6624: Strings in user defined attributes are returned as escaped
I have this Used Defined Attribute:
```
[__AttributeUsage(_AttributeTargets.Var)]
struct DefaultValueAttribute
{
    string mxValueString;
}
```
and I set it like this:
```
[DefaultValue("-1.000000, 0.000000, 0.000000, 0.000000, 0.000000, 1.000000, 0.000000, 0.000000, 0.000000, 0.000000, -1.000000, 0.000000, 0.000000, 0.000000, 0.000000, 1.000000")]
```
When I use this code to retrieve the value:
```
size_t strSize = 0;
const char* str = attr->getArgumentValueString(0, &strSize);
std::string defaultValue = std::string(str, strSize);
```
I get the string including the `"` around it.

This seems unintended and I'd expect the string to be the actual thing in the quotes, not including the quotes.

aidanfnv:
Fixes #6624

When calling getArgumentValueString() from a user defined attribute, we currently return the string including the surrounding quotation marks, rather than just the string itself.

This changes the behavior of getArgumentValueString() to return the string's value, instead of returning the string's token, as that token also contains the surrounding quotation marks. This also modifies the relevant unit test accordingly, to not check for the surrounding quotations.


jkwak-work (approved):
Looks good to me.

PR #6739, opened 2025-04-03
void field rework

kaizhangNV:
This change is a rework on previous PR #6725.

The background of this issue is that we have inconsistent IR instructions cross our code base for make_struct and get_field regarding the `void` field.

Think about the struct:
```
struct
{
void a;
int b;
}
```
this is generated during the resource type legalization pass which will finally be eliminated. However, during IR passes we still need to access such intermediate struct.

So if we call `make_struct(a, b)`, we will have to call `get_field(0)` for a and `get_field(1)` for b, while if we call `make_struct(b)`, we will need to call `get_field(0)` for b.

Both ways are the correct fix, we just need to see which way take us less effort.

Previously in #6725, we choose the 1st way that when constructing construct, we just keep the `void` field in `make_struct` call. However, this causes lots more trouble because we could dynamically emit new `get_field` instruct during legalization, and we will have to do all the cleanup in the same pass and later pass to remove that `void`.

This change change to the 2nd way. where we keep original method to always skip the `void` field, and we just need to calculate the correct index for the field in very few locations.

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -1716,6 +1716,8 @@ Result linkAndOptimizeIR(
     //
     eliminateDeadCode(irModule, deadCodeEliminationOptions);
 
+    cleanUpVoidType(irModule);
+
     if (isKhronosTarget(targetRequest))
     {
         // As a fallback, if the above specialization steps failed to remove resource type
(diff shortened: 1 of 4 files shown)


kaizhangNV:
For reviewer:

This PR basically just revert the previous change.
The only change in this PR is in this file: source/slang/slang-ir-peephole.cpp.

jkwak-work (approved):
Looks good to me.

Issue #6654, opened 2025-03-20
ForceInline moves instructions before a condition


cezneluk:
As in the example below, if `ForceInline` is used, parts of expressions (%58 - %67) are evaluated before condition branches, which does not follow the principle of least surprise and, in my case, probably even hurts performance (reducing time between load and usage).

Tested against current master (9b1e10e).

**Example source code:**

```slang
public struct Triangle {
	float4 v0;
	float4 v1;
	float4 v2;

	[ForceInline]
	public bool intersect(float3 origin, float3 direction) {
		float t = (v0.w - dot(origin, v0.xyz)) / dot(direction, v0.xyz);
		if(t < 0.f)
			return false;

		float u = v1.w + dot(origin, v1.xyz) + t * dot(direction, v1.xyz);
		if(u < 0.f)
			return false;

		float v = v2.w + dot(origin, v2.xyz) + t * dot(direction, v2.xyz);
		if(v < 0.f || u + v > 1.f)
			return false;

		return true;
	}
};

struct PC {
	Triangle* triangles;
	bool* intersected;

	float3 rayOrigin;
	float3 rayDirection;
};

[[vk::push_constant]]
PC pc;

[shader("compute")]
[numthreads(1,1,1)]
void main() {
	*pc.intersected = pc.triangles[0].intersect(pc.rayOrigin, pc.rayDirection);
}
```

**Disassembled SPIR-V:**
```spirv
[...]
         %40 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %37 %int_0
         %41 = OpLoad %v4float %40 Aligned 4
         %42 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %37 %int_1
         %43 = OpLoad %v4float %42 Aligned 4
         %45 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %37 %int_2
         %46 = OpLoad %v4float %45 Aligned 4
         %48 = OpAccessChain %_ptr_PushConstant_v3float %pc %int_2
         %49 = OpLoad %v3float %48
         %51 = OpAccessChain %_ptr_PushConstant_v3float %pc %int_3
         %52 = OpLoad %v3float %51
         %53 = OpCompositeExtract %float %41 3
         %54 = OpVectorShuffle %v3float %41 %41 0 1 2
         %55 = OpDot %float %49 %54
         %56 = OpFSub %float %53 %55
         %57 = OpDot %float %52 %54
[...]

PR #6740, opened 2025-04-04
Add flag to hoist instructions

jarcherNV:
This fixes issue #6654

Only hoist instructions that are optimized by prepareFuncForForwardDiff. Do this by adding a flag tryToHoist to IRSimplificationOptions and set this to true only if called from prepareFuncForForwardDiff, then only hoist if the flag is set. Additionally, do not hoist loops if they only have a single trivial iteration.


jkwak-work:
You need a test case to prevent a regression.

jarcherNV:
I have added a test for this to confirm that the dot is after the branch, to verify that it is not hoisted.

jarcherNV:
The test seems to be failing with some builds, I'm guessing due to different SPIR-V asm. I'll need to investigate.

source/slang/slang-ir-simplify-cfg.h
```diff
@@ -14,6 +25,11 @@ struct CFGSimplificationOptions
     static CFGSimplificationOptions getFast() { return CFGSimplificationOptions{false, false}; }
 };
 
+bool isTrivialSingleIterationLoop(
```
csyonghe, line 28: isTrivialSingleIterationLoop is only using `CFGSimplificationContext::domTree`, instead of exposing the entire `CFGSimplicationContext` here, just modify isTrivialSingleIterationLoop to take only `domTree`, which is available at our new callsite.

source/slang/slang-ir-ssa-simplification.h
```diff
@@ -20,6 +20,7 @@ struct IRSimplificationOptions
 
     bool minimalOptimization = false;
     bool removeRedundancy = false;
+    bool tryToHoist = false;
```
csyonghe, line 23: tryToHoist ==> `hoistLoopInvariantInsts`.

source/slang/slang-ir-redundancy-removal.cpp
```diff
@@ -20,6 +21,11 @@ struct RedundancyRemovalContext
             auto terminatorInst = parentBlock->getTerminator();
             if (auto loop = as<IRLoop>(terminatorInst))
             {
+                // Don't bother hoisting if a loop has only a single trivial iteration.
+                CFGSimplificationContext ctx;
+                if (isTrivialSingleIterationLoop(ctx, func, loop))
```
csyonghe, line 26: pass `dom` to `isTrivialSingleIterationLoop` to avoid redundant computation of the dominator tree.

source/slang/slang-ir-simplify-cfg.h
jarcherNV: I missed this simplification. Good idea. Fixed.

source/slang/slang-ir-redundancy-removal.cpp
jarcherNV: Done.

PR #6748, opened 2025-04-04
Implemented #pragma warning

SuperRonan:
Based on [MSVC's #pragma warning](https://learn.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170).

All specifiers are implemented. 
- `default`: does not change the severity of the warning message
- `disbale` disables the warning message.
- `error` promotes the warning to an error
- `once` shows the warning only once. The documentation is not very explicit on how it should work: once per TU? once per segment? How does it overrides other specifiers? I choosed to do it per segment since it makes more sense and it is more flexible. So for example 
```
#pragma warning(once : id)
emit warning id; // emitted
emit warning id; // once was consumed
#pragma warning(once : id) // once is reseted
emit warning id; // emitted
```
emits the warning twice, but
```
#pragma warning(once : id) // (a)
emit warning id;
#pragma warning(suppress : id)
emit warning id; // warning is suppressed
emit warning id; // state of (a) has been restored, once was already consumed, the warning is disabled
```
emits only one warning. 
These examples are edge cases that whould probably not appear in actual use cases.
- `suppress` disables the warning for the next line. I also tested it in compiler explorer and it disables only the next line (it ignores if the line is extended with a `\`, which I think should not be ignored to allow to suppress multiple lines). I had to add a function to the `Lexer` that looks forward for the end of the next line, without considering the tokenization. The `suppress` specifier is not properly handled when it can't find the end of the next line (emits a warning), but again its an edge case that is not very realistic (writing a  `#pragma warning(suppress : id)` at the end of file, to impact the next file on the stack). 

Multiple warnings can be modified in the same line like: `#pragma warning(disable : 1 2 3; error : 4 5 6)`.

- `push` stores the state of all warnings on a stack. I did not implement `(push, n)` where `n` is the warning level, because it does not appear there is such a thing in Slang. 
- `pop` restores the state of all warnings from the latest entry in the stack. 

Specifiers and warning IDs can be macro invocations. 

I will add some tests soon.


SuperRonan:
So I forgot an important point: to check how this works with `#includes` directives. 
I was using the `SourceLoc` to get the latest `#pragma warning`, but `SourceLoc` are not sorted as I thought.  
Here is an example:
```
// a.slang
#pragma warning (disable : id) // (a)
```
```
// b.slang
#include "a.slang"
emit warning id; // (b)
```
When compiling b.slang: all the tokens in a.slang have a SourceLoc after b.slang, so line (a) does not afect line (b). To keep the same implementation, we need to map each `SourceLoc` to an 'absolute index' that is ordered after the inclusion (basicaly the index in the preprocessed file after all the `#includes` are resolved).

I constructed this mapping in each `SourceView` when pushing or popping a file in the preprocessor. Each `SourceView` is segmented at each `#include` point with a new 'absolute index' base for the rest of the segment.
Then the `WarningStateTracker` automatically does the mapping from `SourceLoc` to 'absolute index'.

source/compiler-core/slang-source-loc.cpp
```diff
@@ -1018,4 +1018,45 @@ PathInfo SourceManager::getPathInfo(SourceLoc loc, SourceLocType type)
...
+}
+
+SourceLoc::RawValue SourceManager::getAbsoluteLocation(SourceLoc location) const
+{
+    SourceLoc::RawValue res = 0;
+    const SourceView* view = findSourceView(location);
+    if (view)
+    {
+        res = view->getAbsoluteLocation(location);
+    }
```
fairywreath, line 1058: Nit
```
if (const SourceView* view = findSourceView(location))
{
    res = view->getAbsoluteLocation(location);
}
```

tests/preprocessor/pragma-warning/default-1.slang
```diff
@@ -0,0 +1,13 @@
+//TEST:SIMPLE:
```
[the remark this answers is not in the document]
SuperRonan: Thanks, that's much easier indeed!

source/slang/slang-preprocessor.cpp
```diff
@@ -467,6 +477,19 @@ struct InputStreamStack
...
 
+    SourceLoc findNextLineEnd(SourceLoc from, UInt lineCount = 1) const
+    {
+        auto top = m_top;
+        SourceLoc res = from;
+        res = top->findNextLineEndImpl(res, lineCount);
+        if (lineCount > 0)
+        {
+            // TODO pop the stack?
+            res = {};
```
fairywreath, line 488: What stack does this refer to? You can just remove this comment if this behavior is not desired.

source/compiler-core/slang-source-loc.h
```diff
@@ -406,6 +410,46 @@ class SourceView
                               ///< locations. Relative to the line number in the underlying file.
     };
 
+    // Represents a segment of a source
+    // All SourceLoc in segment are linearly mapped relative to absoluteBegin
+    struct AbsoluteSegment
+    {
+        SourceLoc begin = {};
+        SourceLoc::RawValue absoluteBegin = {};
+    };
```
fairywreath, line 419: Can you add a comment on how `begin` is different from `absoluteBegin`. From my understanding `begin` is the the default location provided by `SourceLoc` while `absoluteBegin` is the location ordered by includes.

source/slang/slang-preprocessor.cpp
```diff
@@ -3819,6 +4160,130 @@ SLANG_PRAGMA_DIRECTIVE_CALLBACK(handlePragmaOnceDirective)
...
+                    return finish();
+                }
+            }
+        }
+    }
+    else
+    {
+        GetSink(context)->diagnose(tk, Diagnostics::syntaxError);
+    }
+    Expect(context, TokenType::RParent, Diagnostics::syntaxError);
```
fairywreath, line 4284: Exit early when this case is met `tk.type != TokenType::Identifier`, or call EXCEPT, or wrap the whole `tk.type == TokenType::Identifier` if case in another function. This is so that the  if block `tk.type == TokenType::Identifier` can be removed to reduce nesting. In general try to reduce nesting or use smaller functions.

source/slang/slang-preprocessor.cpp
```diff
@@ -3819,6 +4160,130 @@ SLANG_PRAGMA_DIRECTIVE_CALLBACK(handlePragmaOnceDirective)
...
+        else if (tk.getContent() == "pop")
+        {
+            AdvanceToken(context);
+            context->m_preprocessor->warningStateTracker->pop(tk.loc, GetSink(context));
+        }
+        else
+        {
+            while (true)
+            {
+                SourceLoc specifierLocation = PeekRawToken(context).loc;
```
fairywreath, line 4187: Is this location/token returned from `PeekRawToken` different from `Token tk = PeekToken(context);`?

source/slang/slang-preprocessor.cpp
```diff
@@ -3819,6 +4160,130 @@ SLANG_PRAGMA_DIRECTIVE_CALLBACK(handlePragmaOnceDirective)
...
+                {
+                    GetSink(context)->diagnose(
+                        specifierLocation,
+                        Diagnostics::pragmaWarningUnknownSpecifier,
+                        id.getContent());
+                    return finish();
+                }
+                Expect(context, TokenType::Colon, Diagnostics::syntaxError);
+                while (true)
+                {
```
fairywreath, line 4232: Add a comment before this while block to explain what it is doing. From a quick glance I can understand that it is consuming warning numbers, but a comment would be nice to separate this logic from the previous warning string consume logic(or better yet create a small function for this).

fairywreath (commented):
Thanks for this significant work! I will probably need another review pass to fully understand this.

source/slang/slang-preprocessor.cpp
SuperRonan: It refers to the inputStream stack (would have same logic as in `InputStreamStack::peekToken()`). As I mentioned it is an edge case that would not really happen (writing `#pragma warning (suppres)` at the end of an included file to affect the next line in the includer maybe?). I agree it can be clarified in the comments.

source/slang/slang-preprocessor.cpp
SuperRonan: Good remark! It is different. `PeekToken` returns a token that can be in macro expansion, but `PeekRawToken` returns the macro invocation token. It happens when using `#defined` IDs. I will add a comment to clarify it.

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+struct WarningTimeline
+{
+    struct Entry
+    {
+        PragmaWarningSpecifier specifier = {};
+        SourceLoc::RawValue location = {}; // Absolute location
+        // Used for the once specifier
+        // -1 points to this, but not consumed
+        // -2 points to this, but was consumed
+        // >= 0 points to a previous once entry (sharing the payload)
```
csyonghe, line 1023: Can you explain why we need to reference the previous once entry?

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+    };
+    // Sorted by location
+    List<Entry> entries = {};
+
+    const Entry* findEntry(SourceLoc::RawValue location) const
+    {
+        const Entry* res = nullptr;
+        if (entries.getCount() && location >= entries.getFirst().location)
+        {
+            res = ::std::upper_bound(
```
csyonghe, line 1041: auto entryIndex = std::upper_bound(...);
res = entryIndex - 1;

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+                    sink->diagnose(
+                        SourceLoc::fromRaw(prevEntry->debugLocation),
+                        Diagnostics::pragmaWarningPointSuppress,
+                        id);
+                }
+            }
+        }
+    }
+
+    void popEntry(
```
csyonghe, line 1163: addEntryForPragmaPop

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+    {
+        return sourceManager ? sourceManager->getAbsoluteLocation(loc) : loc.getRaw();
+    }
+
+    virtual Severity consumeWarningSeverity(SourceLoc location, int id, Severity severity) override
+    {
+        SourceLoc::RawValue absoluteLoc = getAbsoluteLocation(location);
+        Severity res = severity;
+        WarningTimeline* timeline = entries.tryGetValue(id);
+        if (timeline)
```
csyonghe, line 1202: prefer early return style over nested if.

if (!timeline)
    return res;

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+            getAbsoluteLocation(location),
+            PragmaWarningSpecifier::Suppress,
+            nullptr,
+            sink,
+            id,
+            location);
+        timeline.addEntry(getAbsoluteLocation(nextLineEnd), prev, lastEntry, sink, id, nextLineEnd);
+    }
+
+    void push(SourceLoc location) { stack.add(location); }
```
csyonghe, line 1246: addPragmaPush

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+
+    void pop(SourceLoc location, DiagnosticSink* sink = nullptr)
```
csyonghe, line 1248: addPragmaPop

source/slang/slang-preprocessor.cpp
```diff
@@ -2409,6 +2731,15 @@ static void SkipToEndOfLine(PreprocessorDirectiveContext* context)
     }
 }
 
+static SourceLoc FindNextEndOfLine(
```
csyonghe, line 2734: `findNextEndOfFile`

source/slang/slang-preprocessor.cpp
```diff
@@ -3157,8 +3488,21 @@ static SlangResult readFile(
...
 }
 
-void Preprocessor::pushInputFile(InputFile* inputFile)
+void Preprocessor::pushInputFile(InputFile* inputFile, SourceLoc loc)
 {
+    if (m_currentInputFile)
+    {
+        SourceView* sourceView = m_currentInputFile->getLexer()->m_sourceView;
+        SourceLoc::RawValue offset = SourceRange(sourceView->getLastSegment().begin, loc).getSize();
+        ;
```
csyonghe, line 3497: superfluous ;

source/slang/slang-preprocessor.cpp
```diff
@@ -3819,6 +4163,152 @@ SLANG_PRAGMA_DIRECTIVE_CALLBACK(handlePragmaOnceDirective)
...
+                while (true)
+                {
+                    // Same logic as for the specifierLocation
+                    SourceLoc idLocation = PeekRawToken(context).loc;
+                    Token warningNumberToken = PeekToken(context);
+                    if (warningNumberToken.type == TokenType::IntegerLiteral)
+                    {
+                        AdvanceToken(context);
+                        int warningNumber = stringToInt(warningNumberToken.getContent());
+                        if (specifier == PragmaWarningSpecifier::Suppress)
```
csyonghe, line 4262: can we bottleneck everything to `addEntry` to cleanup this `if-else`?

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+                else
+                {
+                    e.debugLocation = debugLoc.getRaw();
+                }
+                entries.add(e);
+            }
+        }
+        else
+        {
+            if (sink)
```
csyonghe, line 1148: A potential problem here is that different files may get allocated to different SourceLoc ranges, and there is no guarantee that sources with higher SourceLoc numbers are processed before those that have lower SourceLoc numbers.
For example, I could have loaded `a.slang` and `b.slang` to source manager, so b.slang will get sourcelocs that is bigger than a.slang. Now if I check b.slang first and then a.slang, I will hit the path the tries to insert entries before the current known max location, and hit an error here, although it shouldn't error.

csyonghe (commented):
This PR implements the pragma setting/state tracking almost completely in `DiagnosticSink`. This will work for all warnings produced by the front-end, but won't work for warnings that will be produced during or post linking, where the DiagnosticSink may be populated with the pragma settings when the module is loaded from a precompiled module, instead of being parsed from source. 

One way to solve this problem is to make `#pragma warning` converted to just a `Token`. When we encounter this special Token during parsing, we will build a non-mutable `Timeline` stored in the current `ModuleDecl` that we are parsing. This `Timeline` is then a part of the AST and can be serialized out to a precompiled module.

Whenever we are checking/parsing a module, we make sure the DiagnosticSink sees the `Timeline` and can use it to determine warning behavior. The consumed status is then maintained separately on the DiagnosticSink so we don't mutate the AST state.

The only downside I see is if the lexer itself is emitting some warning, then that warning can't be controlled by the pragma. But there are only a handful warnings for the lexer so it is probably a worthwhile compromise for cleaniless.


By converting `#pragma` into something like a `Token(TokenType::PoundPragma)`, we can void the mess to deal with absolute source loc altogether, and will allow us to extend the pragma system to support things like controlling other compiler options (e.g. fastmath) in the parser. For example, a fastmath pragma can just be parsed into a `PragmaDecl` storing the compiler option.

source/slang/slang-preprocessor.cpp
SuperRonan: This is what the "absolute location mapping" solves. It maps `SourceLoc` emitted by the source manager to an index that takes into account the `#inclusion` order. 
Or is it another issue?

source/slang/slang-preprocessor.cpp
```diff
@@ -950,6 +1002,272 @@ struct InputFile
...
+            sink,
+            id,
+            debugLoc);
+    }
+};
+
+struct WarningStateTracker : SourceWarningStateTrackerBase
+{
+    SourceManager* sourceManager = nullptr;
+    Dictionary<int, WarningTimeline> entries = {};
```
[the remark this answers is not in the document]
SuperRonan: The key is the warning id.

source/slang/slang-preprocessor.cpp
SuperRonan: As I mentioned in the PR message, the behaviour of `#pragma warning (once : id)` is not precisely defined (in MSVC's doc at least). 
The simplest implementation would be to keep a `bool consumed` per translation unit, but I feel it is a bit limitted. 
I choosed to once per segment: 
```
#pragma warning (once : id) // (a) Emit id only once in the next segment
// some code
#pragma warning (push)
#pragma warning (disable : id)
#include "file.slang" // Emits a lot of warning 'id', but we don't care about those
#pragma warning (pop) // Restores (a), but we want (a)'s segment to extand from here, so we should keep some kind of link to (a)
// some code
#pragma warning (once : id) // (b) Resets the once for the next file segment (no link to (a))
```

(In this context a file segment is the span between two `#pragma warning` on the same id). 

I agree it is not the cleanest implementation, but it works.

source/slang/slang-preprocessor.cpp
SuperRonan: All other static functions here start with an upper-case letter (like `SkipToEndOfLine`, `PeekToken`, ...).

csyonghe (approved):
This looks good. The PR is providing good value to our users, and I don't think there is anything that should block this PR from landing. I have some thoughts on how #pragma can be implemented more generally at the parsing stage, but that can be left for future explorations.

Issue #5995, opened 2025-01-03
Errors with Filenames including file extension


kphillisjr:
A little bit of clarification is needed in how file resolution works. I have a public Git repository ( see: https://github.com/kphillisjr/Shader_Slang_tests ) ) that has compile errors when i write the following code examples:

```D:\Shader_Slang_Test\Source>D:\Slang\bin\slangc.exe Shaders/ExampleShader.slang -target cpp -o cpu.cpp -DUSING_FILENAME_EXTENSTION=1
Shaders\ExampleShaderModule.slang(38): error 30502: the included source file is expected to implement module 'ExampleShaderModule', but it is implementing 'ExampleShaderModule.slang' instead.
__include "TestEmbeddedDirectory/EmbeddedSubDirectory/InterfaceFoo2.slang";
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Shaders\ExampleShaderModule.slang(41): error 30502: the included source file is expected to implement module 'ExampleShaderModule', but it is implementing 'ExampleShaderModule.slang' instead.
__include "TestEmbeddedDirectory/EmbeddedSubDirectory/MyFooType2.slang";
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Shaders/ExampleShader.slang(1): error 39999: import of module 'ExampleShaderModule' failed because of a compilation error
import ExampleShaderModule;
       ^~~~~~~~~~~~~~~~~~~
(0): fatal error 39999: compilation ceased

D:\Shader_Slang_Test\Source>

```

Tested Versions of Shader-SlangC:

```
# This is from Shader-Slang github release.
D:\Shader_Slang_Test\Source>D:\Slang\bin\slangc.exe -v
2024.17
```

```
# This is the default included in Vulkan SDK.
D:\Shader_Slang_Test\Source>slangc -v
2024.13
```

dir/subdirectory/a.slang
```
// dir/subdirectory/a.slang
implementing m;
void f_a() {}
```

dir/subdirectory/b.slang
```
// dir/subdirectory/b.slang
implementing "m"; // alternate syntax.
__include a; // pulls in `a` to module `m`.
void f_b() { f_a(); }

```

dir/subdirectory/c.slang
```
// dir/subdirectory/c.slang
[...]

PR #6757, opened 2025-04-07
warn when the user puts a file extension in an implementing directive

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/5995

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -10589,6 +10589,7 @@ void SemanticsDeclHeaderVisitor::visitIncludeDecl(IncludeDecl* decl)
 
     if (fileDecl->members.getCount() == 0)
         return;
+
     auto firstMember = fileDecl->members[0];
     if (auto moduleDeclaration = as<ModuleDeclarationDecl>(firstMember))
     {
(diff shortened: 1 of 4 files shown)


szihs (approved):
Looks good to me.

source/slang/slang-check-decl.cpp
```diff
@@ -10629,16 +10631,44 @@ void SemanticsDeclHeaderVisitor::visitIncludeDecl(IncludeDecl* decl)
...
+                // If the normalized name would match, emit warning but continue
+                if (normalizedName.getUnownedSlice().caseInsensitiveEquals(
+                        expectedModuleNameStr.getUnownedSlice()))
+                {
+                    getSink()->diagnose(
+                        implementing->moduleNameAndLoc.loc,
+                        Diagnostics::moduleImplementationHasFileExtension,
+                        moduleNameStr,
+                        normalizedName);
+                    return;
```
jkwak-work, line 10656: Since this is a warning, should we let it through here instead of returning?
I understand that it makes no different as it is but it looks little confusing to me.

jkwak-work (approved):

source/slang/slang-check-decl.cpp
expipiplus1: the early return is what stops this becoming an error

PR #6761, opened 2025-04-07
Add slang-test check for D3D11 double support

### LINKED ISSUE #6171: slang-test should query for D3D11_FEATURE_DOUBLES to determine support for certain tests
1. https://github.com/shader-slang/slang/issues/5601#issuecomment-2609240712
2. https://github.com/shader-slang/slang/issues/5601#issuecomment-2611705996

The following tests fail (shader creation fails with `E_INVALIDARG`) on my integrated Intel GPU (Iris Xe, driver version 31.0.101.5333)

1. `tests/bugs/dxbc-double-problem.slang`
2. `tests/bugs/frexp-double.slang (dx11)`
3. `tests/compute/transcendental-double.slang`
4. `tests/hlsl-intrinsic/bit-cast-double.slang`
5. `tests/hlsl-intrinsic/classify-double.slang`
6. `tests/hlsl-intrinsic/scalar-double-simple.slang`

On the same GPU, querying for `D3D11_FEATURE_DOUBLES` would reveal this and let us skip the test instead of reporting a failure.

I propose that 

1. We specify statically (rather than detecting e.g. `SAD4` instructions in the generated DXBC), in each of the test files listed above, that the test requires `D3D11_FEATURE_DOUBLES`.
2. If `D3D11_FEATURE_DOUBLES` is not available, we skip the test rather than failing it.

aidanfnv:
Fixes #6171

This adds logic for reporting double support to the d3d11 backend, to allow for checking for that support when running tests on GPUs that do not support D3D11_FEATURE_DOUBLES, and this adds checks for that support to tests that require the feature.


tests/bugs/dxbc-double-problem.slang
```diff
@@ -1,6 +1,7 @@
 //TEST(compute):COMPARE_COMPUTE_EX:-slang -compute -output-using-type -shaderobj
 //TEST(compute, vulkan):COMPARE_COMPUTE_EX:-vk -compute -output-using-type -shaderobj -render-feature double
 //TEST(compute):COMPARE_COMPUTE_EX:-cpu -compute -output-using-type -shaderobj
+//TEST(compute):COMPARE_COMPUTE_EX:-dx11 -compute -output-using-type -shaderobj -render-feature half,double
```
jkwak-work, line 4: I think the issue is reproducible without adding a new test.
I don't think you want to add this one, which is probably duplicate to the existing test lines in this file.

jkwak-work (commented):
Were you able to reproduce the problem on Intel GPU as described on the issue description?

My understanding is that you can reproduce the problem without adding new tests.
It may not be obvious but some of "//TEST" lines may actually execute "dx11".
I am not sure which one does it, but I suspect that "-slang" may execute all rendering targets like vulkan, dx11 and dx12 but I am not sure.

jkwak-work:
The plan is to borrow a laptop with Intel GPU for testing.

tests/bugs/dxbc-double-problem.slang
aidanfnv: It was a duplicate. The test still failed because it was running the dx11 test in the first line. I added the -render-feature double to there instead.

aidanfnv:
I acquired an Intel laptop to test this with, repro'd the original issue, and then tested my changes.
After removing the duplicate test entries that I added and adding -render-feature double to the existing test entries that run dx11, I can confirm that this change fixes the issue for DX11.

jkwak-work (approved):
Looks good to me.

Issue #6525, opened 2025-03-04
Providing mismatched type for input semantic crash the compiler when targeting spirv


gan74:
Declaring a semantic as `in` with the wrong type crashes the compiler with no error message.
The crash comes from  `SLANG_ASSERT(realGlobalVar);` in `legalizeEntryPointParameterForGLSL`.

Here is the repro:
```
// SV_VertexID should be an INT not an UINT. Declaring it properly fixes the problem
in uint gl_VertexIndex : SV_VertexID;

static const float2 uvs[] = { float2(0.0, 0.0), float2(0.0, 2.0), float2(2.0, 0.0) };

[shader("vertex")]
float4 vert_main() : SV_Position{
    const float2 uv = uvs[gl_VertexIndex];
    return float4((uv * 2.0 - 1.0), 0.0, 1.0);
}
```

PR #6762, opened 2025-04-07
Get real value for typeAdapter

gtong-nv:
When the type is mismatch and typeAdapter is used, 
get the real value from typeAdapter so that we don't get nullptr for irValue.

This fixes the assert if uint is used for SV_VertexID, which is an int in the system binding semantic.

Fixes: #6525

### DIFF
--- a/source/slang/slang-ir-glsl-legalize.cpp
+++ b/source/slang/slang-ir-glsl-legalize.cpp
@@ -3541,6 +3541,13 @@ void legalizeEntryPointParameterForGLSL(
                         if (elem.key == key)
                         {
                             realGlobalVar = elem.val.irValue;
+                            if (!realGlobalVar &&
+                                ScalarizedVal::Flavor::typeAdapter == elem.val.flavor)
+                            {
+                                auto typeAdapterVal =
+                                    as<ScalarizedTypeAdapterValImpl>(elem.val.impl);
+                                realGlobalVar = typeAdapterVal->val.irValue;
+                            }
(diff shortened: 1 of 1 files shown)


kaizhangNV:
add a unit test please.

source/slang/slang-ir-glsl-legalize.cpp
```diff
@@ -3541,6 +3541,13 @@ void legalizeEntryPointParameterForGLSL(
                         if (elem.key == key)
                         {
                             realGlobalVar = elem.val.irValue;
+                            if (!realGlobalVar &&
+                                ScalarizedVal::Flavor::typeAdapter == elem.val.flavor)
+                            {
+                                auto typeAdapterVal =
+                                    as<ScalarizedTypeAdapterValImpl>(elem.val.impl);
+                                realGlobalVar = typeAdapterVal->val.irValue;
```
kaizhangNV, line 3549: non-nullptr check on `typeAdapterVal`?

kaizhangNV (changes requested):

source/slang/slang-ir-glsl-legalize.cpp
gtong-nv: Thanks! nullptr check added

gtong-nv:
> add a unit test please.
Not sure what language feature is this, I just added a test for this issue in bugs/

kaizhangNV (approved):

PR #6769, opened 2025-04-09
Fix #6544: Properly format nested type names in extensions

### LINKED ISSUE #6544: `getFullName` on type defined in extension returns incorrect string
In the following slang code, a different implementation of `State` is chosen via extensions:

```
public interface IOptimizer<T : __BuiltinFloatingPointType>
{
    public associatedtype State;
}

public struct FullPrecisionOptimizer<T : __BuiltinFloatingPointType>
{
}

public extension FullPrecisionOptimizer<float> : IOptimizer<float>
{
    // Parameter precision is already a float -> Just use it directly
    public typealias State = float;
}

public extension FullPrecisionOptimizer<half> : IOptimizer<half>
{
    // Parameter precision is in half precision -> Store full precision copy
    public struct State
    {
        public float paramFullPrecision;
        public half paramHalfPrecision;
    }
}
```

Looking up `FullPrecisionOptimizer<half>::State` via reflection succeeds. However, calling `getFullName()` on the resulting type reflection returns `.State`, and not `FullPrecisionOptimizer<half>.State` as expected.

Minimum repro via sgl:
```
import sgl

device = sgl.Device()

m = device.load_module("repro.slang")
state = m.layout.find_type_by_name("FullPrecisionOptimizer<half>::State")
print(state.full_name) # Returns ".State", should be "FullPrecisionOptimizer<half>.State"
```

szihs:
Modify DeclRefBase::toText to properly handle types defined in extensions by qualifying them with their parent type name. This ensures getFullName() returns the full name like 'FullPrecisionOptimizer<half>.State' instead of just '.State'. 

Also add diagnostic tests for the same


kaizhangNV (dismissed):

kaizhangNV:
Is there an issue number associated with this?

szihs:
> Is there an issue number associated with this?
https://github.com/shader-slang/slang/issues/6544

source/slang/slang-ast-decl-ref.cpp
```diff
@@ -318,6 +318,20 @@ void DeclRefBase::toText(StringBuilder& out)
         return;
     }
 
+    // Handle extension types - format as "ParentType.MemberName"
+    auto currentDecl = getDecl();
+    if (currentDecl && currentDecl->parentDecl && as<ExtensionDecl>(currentDecl->parentDecl))
+    {
+        auto extDecl = as<ExtensionDecl>(currentDecl->parentDecl);
+        if (extDecl->targetType)
```
csyonghe, line 326: This is dropping generic arguments for the parent.
If I have
```
extension<T:IFoo, int N> Array<T, N>
{
        void member() {}
}

struct Bar : IFoo {};
```
And then I have a declref to `Array<Bar,3>.member`, the current logic will print `Array<T,N>.member`.

csyonghe: This should be:
```
getTargetType(getCurrentASTBuilder(), getParent())->toText(out);
out << ".";
...
```

csyonghe (changes requested):
Requesting changes since I think it is incorrect to drop declref info in `toText`.

source/slang/slang-ast-decl-ref.cpp
szihs: > This is dropping generic arguments for the parent. If I have
> 
> ```
> extension<T:IFoo, int N> Array<T, N>
> {
>         void member() {}
> }
> 
> struct Bar : IFoo {};
> ```
> 
> And then I have a declref to `Array<Bar,3>.member`, the current logic will print `Array<T,N>.member`.

@csyonghe can you help with the reproducer of this case. I want to understand how would it break . Thanks

szihs: Done - Added test case and used getTargetType

source/slang/slang-ast-decl-ref.cpp
```diff
@@ -333,29 +333,54 @@ void DeclRefBase::toText(StringBuilder& out)
...
     }
 
-    decls.reverse();
+    declRefs.reverse();
 
     bool first = true;
-    for (auto decl : decls)
+    for (auto declRef : declRefs)
     {
+        auto decl = declRef->getDecl();
```
csyonghe, line 368: In this loop you are already iterating through all parent decls, why checking for decl->parentDecl again?

Just check if decl is a extension decl and apply the logic there.

PR #6773, opened 2025-04-09
Fix benchmark/compile.py nested f-string and deprecated constant issues

### LINKED ISSUE #6772: benchmark\compile.py fails due to nested f-string syntax error
When running the benchmark in CI, the following error occurs:

```
Notice:  To update, run: python.exe -m pip install --upgrade pip
  File "C:\actions-runner\_work\slang\slang\tools\benchmark\compile.py", line 188
    print(f'Total: **{timings[f'full/{target_ext}/precompilation']['compileInner']} ms**\n')
                                ^^^^
SyntaxError: f-string: unmatched '['
Error: Process completed with exit code 1.
```
The error is caused by nested f-strings with incompatible quote usage. The nested f-string f'full/{target_ext}/precompilation' inside the outer f-string creates a syntax error.

gtong-nv:
This commit resolves two issues in the benchmark script:
1. Fixed nested f-string syntax error that was causing CI failures (fixes #6772)
2. Fixed key access error handling for timing dictionary

The nested f-string was causing a syntax error in CI, and the proper fix is to create the key variable separately before using it. Also improved error handling for cases where compilation fails and timing data might be missing.

Fixes: #6772

### DIFF
--- a/tools/benchmark/compile.py
+++ b/tools/benchmark/compile.py
@@ -73,9 +73,9 @@ def run(command, key):
         try:
             results = subprocess.check_output(command, stderr=subprocess.STDOUT, shell=True).decode('utf-8')
         except subprocess.CalledProcessError as exc:
+            print(f"[Error] Failed to run command: {command}")
             print(exc.output.decode('utf-8'))
-            return
-            # exit(-1)
+            return  # Return without adding to timings
 
         p = parse(results)
         if len(profile) == 0:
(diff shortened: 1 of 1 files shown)


jkwak-work:
The change looks good.
I re-triggered the benchmark CI until it picks up the older machine to make sure the change works also on the older machine, `SlangWin10-2`.
The first attempt of it picked up the new GCP machine, `slang-win-2022-runner-4`, and it passed the test.

jkwak-work (approved):
The change appears to be working for both GCP and the regular runners.

PR #6776, opened 2025-04-09
Add a workflow for auto labeling issues

sricker-nvidia:
Change introduces a workflow that will automatically run on new issues and add a "Dev Opened" label if opened by select dev members. This workflow can be used to add additional labels in the future.


.github/workflows/add-issue-labels.yml
```diff
@@ -0,0 +1,61 @@
...
+              team_slug: 'dev'
+            });
+
+            // For testing, use the provided username otherwise use the actual issue creator
+            const username = '${{ github.event.inputs.test_username || github.event.issue.user.login }}';
+
+            // Check if user is in the Nvidia dev team
+            const isTeamMember = response.data.some(member =>
+              member.login === username &&
+              member.login !== 'fairywreath'
```
kaizhangNV, line 37: Why do we filter this user out?

kaizhangNV (approved):
LGTM.

For Windows CI, just re-trigger it, as it runs in GCP runner again.

.github/workflows/add-issue-labels.yml
sricker-nvidia: Brandon wanted the dev list minus Darren

PR #6777, opened 2025-04-10
Document CoopVec functions

### LINKED ISSUE #6297: Polish the generated document for CoopVec
Currently we have a proposal document that has most detailed information about CoopVec:
https://github.com/shader-slang/slang/blob/master/docs/proposals/019-cooperative-vector.md

But we also need to polish and update the generated document for CoopVec:
https://shader-slang.org/stdlib-reference/types/coopvec-04/index.html

jkwak-work:
Closes #6297 

Note that some of comments are removed or converted from `///` to `//`, because the doc generation tool prints warnings about them.
It is because the overloading functions can have `///` style comment only once.


jkwak-work:
I reviewed the changes with Jekyll on my local machine.
It looked as expected.

It appears that the stdlib-reference repo has its own bot to update the doc.
I don't think I need to manually submit the generated document files.
https://github.com/shader-slang/stdlib-reference/commits/main/

source/slang/hlsl.meta.slang
```diff
@@ -22410,9 +22450,12 @@ struct CoopVec<T : __BuiltinArithmeticType, let N : int> : IArray<T>, IArithmeti
     }
 
     //
-    // Equality and ordering
+    // IComparable
```
sricker-nvidia, line 22453: Not sure I understand this change, could you explain?

sricker-nvidia: Based on what I see below, maybe this should be something like, "Equality and ordering functions that exist only to conform to IComparable" ?

source/slang/hlsl.meta.slang
```diff
@@ -22371,8 +22405,10 @@ struct CoopVec<T : __BuiltinArithmeticType, let N : int> : IArray<T>, IArithmeti
         return N;
     }
 
+    /// Access an individual element in the Cooperative vector by index.
     __subscript(int index) -> T
     {
+        [ForceInline]
```
sricker-nvidia, line 22411: Could you explain this edit? I saw a TODO somewhere that mentions a dxc bug, but I'm not familiar. Is this just to match some of the other functions?

source/slang/hlsl.meta.slang
```diff
@@ -22488,11 +22548,15 @@ struct CoopVec<T : __BuiltinArithmeticType, let N : int> : IArray<T>, IArithmeti
 
     __intrinsic_op($(kIROp_Sub))
     This __pureSub(This other);
+
     [mutating]
     [require(hlsl)]
     void __mutSub(This other)
```
sricker-nvidia, line 22554: This seems like a helper function, so it probably doesn't need documentation, but is there anything that indicates that it is a help function?

source/slang/hlsl.meta.slang
```diff
@@ -24020,7 +24223,6 @@ CoopVec<T, M> coopVecMatMulPacked(
     }
 }
 
-// specialized coopVecMatMul for non-packed inputs
```
sricker-nvidia, line 24023: Shouldn't this still have some documentation as it differs from the other coopVecMatMul?

source/slang/hlsl.meta.slang
```diff
@@ -24288,7 +24490,7 @@ uint32_t4 unpack_u8u32(uint8_t4_packed packed)
     return unpackUint4x8ToUint32(packed);
 }
 
-/// Pack a vector of 4 unsigned 32 bit integers into a packed value of 4 8-bit integers, dropping unused bits.
+/// Pack a vector of 4 unsigned 32/16 bit integers into a packed value of 4 8-bit integers, dropping unused bits.
```
sricker-nvidia, line 24493: Could you explain the addition of "16"?

source/slang/hlsl.meta.slang
```diff
@@ -24297,7 +24499,7 @@ uint8_t4_packed pack_u8(uint32_t4 unpackedValue)
     return packUint4x8(unpackedValue);
 }
 
-/// Pack a vector of 4 signed 32 bit integers into a packed value of 4 8-bit integers, dropping unused bits.
+/// Pack a vector of 4 signed 32/16 bit integers into a packed value of 4 8-bit integers, dropping unused bits.
```
sricker-nvidia, line 24502: Is this intended to cover the removed line at 24309? If so I don't  think this is technically accurate. This function seems to take in a signed value "Pack a vector of 4 signed 32 bit integers" and the next function that seems to refer to "Pack a vector of 4 signed16 bit integers" actually takes in 4 unsigned 16 bit integers, rather than signed as stated in the comment.

source/slang/hlsl.meta.slang
```diff
@@ -24324,7 +24524,7 @@ int8_t4_packed pack_s8(int16_t4 unpackedValue)
     return packInt4x8(unpackedValue);
 }
 
-/// Pack a vector of 4 unsigned 32 bit integers into a packed value of 4 8-bit integers,
+/// Pack a vector of 4 unsigned 32/16 bit integers into a packed value of 4 8-bit integers,
```
sricker-nvidia, line 24527: Seeing signed vs unsigned issues with a lot of these comments.

source/slang/hlsl.meta.slang
```diff
@@ -24631,8 +24825,8 @@ uint packInt4x8(int16_t4 unpackedValue)
     }
 }
 
-/// Pack a vector of 4 signed 32 bit integers into a packed value of 4 8-bit integers,
-/// clamping each value to the range [-128, 127] to ensure it fits within 8 bits.
+/// Pack a vector of 4 signed 32/16 bit integers into a packed value of 4 8-bit integers,
+/// clamping each value to the range [0, 255] to ensure it fits within 8 bits.
```
sricker-nvidia, line 24829: Some of these functions seem to return a signed int vs an unsigned int holding the 4 8-bit packed values. Is it worth indicating sign vs unsigned in the description? I guess it's implied by the range?

source/slang/hlsl.meta.slang
jkwak-work: I am glad you asked those questions.

For this comment, it has something to do with what CoopVec struct implements.
The declaration of CoopVec is following,
```
struct CoopVec<T : __BuiltinArithmeticType, let N : int> : IArray<T>, IArithmetic
```
And IArithmetic is defined as,
```
interface IArithmetic : IComparable
```

So CoopVec is required to implement functions that IComparable promised to provide, which are
```
/// Represents types that can be compared.
/// Implemented by all builtin integer and floating point scalar or vector types.
interface IComparable
{
    /// Returns true if two values of the conforming type are equal.
    __builtin_requirement($( (int)BuiltinRequirementKind::Equals))
    bool equals(This other);

    /// Returns true if `this` is less than `other`.
    __builtin_requirement($( (int)BuiltinRequirementKind::LessThan))
    bool lessThan(This other);

    /// Returns true if `this` is less than or equal to `other`.
    __builtin_requirement($( (int)BuiltinRequirementKind::LessThanOrEquals))
    bool lessThanOrEquals(This other);
}
```

The intention of the comment change is to... make it more consistent to the other comment down there.
Maybe I can change it to
```
// Implement the member functions required by IComparable
```
But that seems too verbose.

source/slang/hlsl.meta.slang
jkwak-work: I think there is a bug in the current DXC implementation that when a function returns CoopVec instance without any operation applied to it, it doesn't function correctly like,
```
CoopVec<T,N> func(CoopVec<T,N> a) { return a; }
```
In order workaround it temporarily, it is required to use `[ForceInline]`; or at least that's what I think is happening.

But this line has nothing to do with it.
In general, we want to use `[ForceInline]` when the body of the function is very simple; or just a single line.
And this one happened to be one of those cases.

In other words, this line change is more for the cosmetic reason when HLSL is emitted from Slang rather than any functional reason.

source/slang/hlsl.meta.slang
jkwak-work: When function names start with two underscore characters, they are considered "internal" functions.
And doc generating tool treats it in a way that those functions don't show up on the document.

source/slang/hlsl.meta.slang
jkwak-work: This one was my mistake to remove.
I am gonna bring it back.

source/slang/hlsl.meta.slang
jkwak-work: For this function, the generated document can be found from here,
https://shader-slang.org/stdlib-reference/global-decls/pack_u8.html

As you can see on the document, when two or more functions are overloaded, they are squashed into one document page.
And there can be only one comment that describe what the function does.
When we add comments on two functions whose names are same, the doc generation tool prints a warning and the second one is ignored.

I am combining the descriptions for both into one comment here.

source/slang/hlsl.meta.slang
jkwak-work: This is to combine with the line 24318 not 24309.

source/slang/hlsl.meta.slang
sricker-nvidia: Good to know, thanks!

source/slang/hlsl.meta.slang
sricker-nvidia: Thanks for the explanation, I think it's fine then.

source/slang/hlsl.meta.slang
sricker-nvidia: ack thanks

source/slang/hlsl.meta.slang
jkwak-work: I think all of the packing functions return unsigned int.
The sign-ness is applied to the input.
When values are considered packed, there is no way to know if each should be treated as signed or unsigned.
The function name implies what is expected. But the comment explicitly describes the clamping range; LLM did it not me.

source/slang/hlsl.meta.slang
sricker-nvidia: Thanks for the explanation, that makes a lot more sense now.

source/slang/hlsl.meta.slang
sricker-nvidia: I see, based on your other comment about how the documentation treats the overloaded functions, this description is applying to the 2 signed functions. I did not understand how the documentation was generated and just assumed that the comment referred to both this function and the one immediately below. Thanks for explaining.

source/slang/hlsl.meta.slang
sricker-nvidia: Disregard, I think this is irrelevant now that I understand how the docs are generated.

source/slang/hlsl.meta.slang
sricker-nvidia: ah I see, I think that's fine then, thanks

sricker-nvidia (approved):
LGTM

PR #6786, opened 2025-04-10
Fix various intptr_t issues by defining its width in `getIntTypeInfo`

juliusikkala:
The `SLANG_PTR_IS_32` & `SLANG_PTR_IS_64` macros seem to be quite heavily used around the codebase to determine the size of `intptr_t` (e.g. `getTypeBitSize`, any-value marshalling, `isIntValueInRangeOfType`, etc.), so I extended that to `getIntTypeInfo` as well.

This fixes at least the following internal compiler errors:
* `vector<intptr_t, N>` types which were broken by peephole optimization trying to use `getIntTypeInfo`
* Bit shifting `intptr_t` types (SCCP wanted to just get the sign with `getIntTypeInfo`)

AFAIK there's not really any way for this to break existing code, since the previous behaviour was just to crash when these types were encountered.


juliusikkala:
There was a pre-existing test for intptr_t, so I just extended that one for the two cases I ran across.

fairywreath (dismissed):

juliusikkala:
The CUDA error wasn't too hard to fix, but I've no clue what's going in the failing Mac OS test, and can't test locally.

csyonghe:
I wonder if it is because SLANG_PTR_IS_64 doesn't work on macos?
Could try changing:
```
#elif defined(_M_ARM64) || defined(__aarch64__)
    #define SLANG_PROCESSOR_ARM_64 1
```

in slang.h to:
```
#elif defined(_M_ARM64) || defined(__aarch64__)||defined(__ARM_ARCH_ISA_A64 )
    #define SLANG_PROCESSOR_ARM_64 1
```

juliusikkala:
Hmm, looks like that unfortunately wasn't it. I'm kind of considering whether it'd make sense to just disable that test on Macs for now and add a new issue about that to handle it later?

juliusikkala:
Alright, finally managed to find out what the Mac issue was. So basically, the problem there is that `uintptr_t == unsigned long`, while `uint64_t == unsigned long long`. Despite both being 8-byte unsigned integers, they're not the same type, so none of the overloads for vector ops matched. However, we can't just always define the overloads for `intptr_t`; on Windows and Linux, `uintptr_t == uint64_t` (or uint32_t on a 32-bit machine), and the overloads would cause redefinitions. So they only need to be defined on Macs and nowhere else.

fairywreath (approved):

PR #6787, opened 2025-04-10
Added getCanonicalGenericConstraints2 (sorts constraints and allows more generic expressions)

SuperRonan:
While working on #6671, an error was triggered in test bugs/gh-6331.slang. 
The problem is that there is a constaint not directly on a `GenericTypeParam`, but on a `VectorExpressionType` using a `GenericTypeParam`. 
The problem did not arise before because it is in a generic extension, but with #6671, generic extensions are fully mangled, and `getCanonicalGenericConstraints(...)` was called. 
But this function, there is an `assert` to check that all constraints are `GenericTypeParam`. 
So I created a simpler example (the new test where-5.slang) which exposes this issue even without the changes of #6671. 
Note that this new example only crashes in Debug (with `assert` enabled), not in Release. If the `assert` is ignored, the later code fails to retrive the `GenericTypeParam` on a `VectorExpressionType` and does not add the constraint on the vector type to the cannonical constraints, so the mangled name is incomplete (but it still generally works). 

To fix this issue, I added a new function `getCanonicalGenericConstraints2` which handles any `Type` in generic constraints. 
I also noticed in the first `getCanonicalGenericConstraints(...)` a 'TODO sort types'. 
Constrained types were previously sorted by `GenericTypeParam`'s index, but now we have a more generic `Type`. 
So I started to implement a comparator (I did not find any, maybe there was already one?) to sort the types in the generic constraint. 
I am not sure the way I wrote the comparator is the right one, so I only implemented it partially to cover enough cases. 

I am also thinking the generic constraints should be canonized when creating the `GenericDecl`, not just before mangling it. 
But maybe it should be the subject of another PR.


source/slang/slang-check-decl.cpp
```diff
@@ -11353,6 +11353,257 @@ static void _dispatchDeclCheckingVisitor(Decl* decl, DeclCheckState state, Seman
...
+        }
+        res = specRes.value_or(0);
+    }
+    return res;
+}
+
+int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
+{
+    SLANG_UNUSED(astBuilder);
+    // No risk of overflow, replace with <=> in C++20
```
csyonghe, line 11544: I think a general logic for decl comparison can be:
- Find the lowest common parent decl for both lhs and rhs.
- If they there isn't a common parent decl, e.g. they come from two different modules, just return fallback to name comparison the module.
- If there is a common parent decl, check the order of the child nodes. If the child node containing lhs is appearing before rhs, return -1, otherwise return 1.

(To speed up checks, we may need to add a `mapDeclMemberToIndex` dictionary to `ContainerDecl` after `memberDictionary`.)

source/slang/slang-check-decl.cpp
```diff
@@ -11353,6 +11353,257 @@ static void _dispatchDeclCheckingVisitor(Decl* decl, DeclCheckState state, Seman
...
+                    }
+                }
+            }
+        }
+        else if (auto lDRT = dynamicCast<DeclRefType>(&lhs))
+        {
+            auto rDRT = dynamicCast<DeclRefType>(&rhs);
+            auto lDecl = lDRT->getDeclRef().getDecl();
+            auto rDecl = rDRT->getDeclRef().getDecl();
+            specRes = compareDecls(astBuilder, lDecl, rDecl, feedback);
```
csyonghe, line 11530: If compareDecl returns 0 (equal), we need to continue compare the specialization arguments (if any).

- If both decls are direct decl ref, return 0.
- If left and right declref are of different types in the `DeclRefBase` hierarchy, return `lhsDeclRef.getDeclRefBase()->astNodeType` - `rhsDeclRef.getDeclRefBase()->astNodeType`.
- If both declref are `GenericAppDeclRef`, recursively compare the vals in `GenericAppDeclRef::args`.
- If both declref are `LookupDeclRef`, recursively compare the witness.
- If both declref are `MemberDeclRef`, recursively compare its `getParentOperand()`.

csyonghe (commented):
Looks good to me. We don't yet have a Type/Val comparer and this is a good time to add it.

I have some comments on how to improve the type comparer.

source/slang/slang-check-decl.cpp
```diff
@@ -11395,8 +11652,58 @@ OrderedDictionary<GenericTypeParamDeclBase*, List<Type*>> getCanonicalGenericCon
...
+        {
+            if (!genericConstraints.containsKey(typeToAdd))
+            {
+                genericConstraints[typeToAdd] = HashSet<Type*>();
+            }
+            genericConstraints[typeToAdd].add(supExpr.type);
+        }
+    }
+    const auto typeComparator = [&](Type* lhs, Type* rhs)
+    { return compareTypes(astBuilder, *lhs, *rhs) < 0; };
```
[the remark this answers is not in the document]
SuperRonan: The formater won't let me...

source/slang/slang-check-decl.cpp
```diff
@@ -11353,6 +11353,257 @@ static void _dispatchDeclCheckingVisitor(Decl* decl, DeclCheckState state, Seman
...
+}
+
+int compareTypes(ASTBuilder* astBuilder, Type& lhs, Type& rhs, int* feedback)
+{
+    // No risk of overflow, replace with <=> in C++20
+    int res = static_cast<int>(rhs.astNodeType) - static_cast<int>(lhs.astNodeType);
+    if (res == 0)
+    {
+        std::optional<int> specRes = {};
+        if (auto lVec = dynamicCast<VectorExpressionType>(&lhs))
```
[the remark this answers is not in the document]
SuperRonan: Fixed with 6378115.

source/slang/slang-check-decl.cpp
SuperRonan: Not yet implemented in 6378115.

source/slang/slang-check-decl.cpp
SuperRonan: Implemented in 6378115
(except for `SubtypeWitness`, which I forgot).

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
     return res;
 }
 
+int compareDeclRefBases(ASTBuilder* astBuilder, DeclRefBase& lhs, DeclRefBase& rhs, int* feedback)
+{
+    SLANG_UNUSED(astBuilder);
+    // No risk of overflow, replace with <=> in C++20
+    int res = static_cast<int>(rhs.astNodeType) - static_cast<int>(lhs.astNodeType);
+    if (res == 0)
```
csyonghe, line 11575: prefer early return:
```
if (res)
    return res;
```

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
...
+    {
+        std::optional<int> specRes = {};
+
+        if (auto lDirect = dynamicCast<DirectDeclRef>(&lhs))
```
csyonghe, line 11579: as<DirectDeclRef>

source/slang/slang-check-decl.cpp
```diff
@@ -11468,67 +11496,24 @@ int compareTypes(ASTBuilder* astBuilder, Type& lhs, Type& rhs, int* feedback)
...
-        }
-        else if (auto lMat = dynamicCast<MatrixExpressionType>(&lhs))
-        {
-            auto rMat = dynamicCast<MatrixExpressionType>(&rhs);
-            auto lR = lMat->getRowCount();
-            auto rR = rMat->getRowCount();
-            int RComp = compareIntVals(astBuilder, lR, rR, feedback);
-            if (RComp != 0)
+            auto rDRT = dynamicCast<DeclRefType>(&rhs);
+            auto lDecl = lDRT->getDeclRef().getDecl();
```
csyonghe, line 11502: I would do
```
auto lDeclRef = lDRT->getDeclRef();
auto rDeclRef = rDRT->getDeclRef();
return compareDeclRef(lDeclRef, rDeclRef);
```

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
     return res;
 }
 
+int compareDeclRefBases(ASTBuilder* astBuilder, DeclRefBase& lhs, DeclRefBase& rhs, int* feedback)
```
csyonghe, line 11570: make this `compareDeclRef`, and compare the decl first in the function.
If decl is the same, continue with the rest of the logic here.

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
...
+            else
+            {
+                // No risk of overflow, replace with <=> in C++20
+                specRes = static_cast<int>(rGeneric->getArgCount() - rGeneric->getArgCount());
+            }
+        }
+        else if (auto lLookup = dynamicCast<LookupDeclRef>(&lhs))
+        {
+            auto rLookup = dynamicCast<LookupDeclRef>(&rhs);
+            specRes = compareWitnesses(
```
csyonghe, line 11616: this should just be `compareVals`.

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
...
+                astBuilder,
+                lLookup->getWitness(),
+                rLookup->getWitness(),
+                feedback);
+        }
+        else if (auto lMember = dynamicCast<MemberDeclRef>(&lhs))
+        {
+            auto rMember = dynamicCast<MemberDeclRef>(&rhs);
+            specRes = compareDeclRefBases(
```
csyonghe, line 11625: if we have a `compareDeclRef`, this bottlenecks becomes a recursive call to `compareDeclRef`.

source/slang/slang-check-decl.cpp
```diff
@@ -11582,6 +11567,139 @@ int compareDecls(ASTBuilder* astBuilder, Decl& lhs, Decl& rhs, int* feedback)
...
+}
+
+int compareWitnesses(ASTBuilder* astBuilder, Witness& lhs, Witness& rhs, int* feedback)
+{
+    SLANG_UNUSED(astBuilder);
+    // No risk of overflow, replace with <=> in C++20
+    int res = static_cast<int>(rhs.astNodeType) - static_cast<int>(lhs.astNodeType);
+    if (res == 0)
+    {
+        std::optional<int> specRes = {};
```
csyonghe, line 11648: for witnesses, we can simply compare `subType` and `supType`.

source/slang/slang-check-decl.cpp
SuperRonan: I don't understand, `compareDeclRefs` should be implemented in `compareVals`? It can be done, but `compareVals` will be a bit too long if we implemented all specialized comps in it. 
Or is it that we should just do a generic array comparison on all `Val::m_operands`?

source/slang/slang-check-decl.cpp
SuperRonan: Fixed with df4e3e5?

source/slang/slang-check-decl.cpp
SuperRonan: Fixed with df4e3e5.

source/slang/slang-check-decl.cpp
csyonghe: Yeah, I think implementing the generic comparison on the `Val::m_operands` array can serve as a good fallback path for a lot of the cases, and will greatly simply the code.

SuperRonan: So I added the "generic" operands comparison in c19531f, but should we replace most (if not all) the specialized comparison code with it?
For now, some operands are not compared (operand 1 `getGenericDeclRef` in `GenericAppDeclRef`, or the type of `ConstantIntVal`), but is it relevant to compare them too?

SuperRonan: Simplified compareVals in a7e54be. 
I removed the `feedback` pointer because we should handle all cases now. 
I feel I should also remove the `astBuilder`, originally I thought it might be necessary at some point, but it turned out to be never used. Could there be case in which it could be useful?

csyonghe: I think it would be a good idea to use general comparison logic where it applies. The only exception I can think of is when comparing DirectDeclRefs, that you want to dispatch to the current compareDecl logic. For anything else, the default should just work?

csyonghe: Yeah, feel free to remove unused parameters.

csyonghe (dismissed):
The new comparison is much cleaner. PR looks good to me.

source/slang/slang-check-decl.cpp
SuperRonan: A `DirectDeclRef` has only one operand, its `Decl` which will be compared, so there is no need for special case?

csyonghe: You are right, the current implementation is good as is. I was really talking about the case where an operand is a decl*, which is already handled.

SuperRonan:
I have been investigating the failling tests on my build:
These four tests fail also on cb361223b88a21446cea60d316e3c01b7f8c2546:
```
error: failed to run test 'tests/diagnostics/break-outside-loop.slang'
error: failed to run test 'tests/diagnostics/global-uniform.slang'
error: failed to run test 'tests/diagnostics/no-type-conformance.slang'
```
`tests/glsl-intrinsic/array-legnth.slang` sometimes crashes, sometimes not (probably some UB, it does not crash in `Debug`, but crashes in `Release` or `ReleaseWithDebInfo`), even on cb361223b88a21446cea60d316e3c01b7f8c2546.
So I will not be considering these here.
- Concerning this PR: 
`slang-unit-test-tool/declTreeReflection.internal` Seems to fail on this PR.
```
error: getTypeFullName(typeParamTConstraintType1) == "IArithmetic" - E:\libraries\mySlang\tools\slang-unit-test\unit-test-decl-tree-reflection.cpp (212)
error: getTypeFullName(typeParamTConstraintType2) == "IFloat" - E:\libraries\mySlang\tools\slang-unit-test\unit-test-decl-tree-reflection.cpp (215)
FAILED test: 'slang-unit-test-tool/declTreeReflection.internal'
```
Here constraints `IFloat` and `IArithmetic` appear the be swapped, because they are sorted now (I also added a `sorting` of constraints in the original `getCanonicalGenericConstraints`). 
One thing that bugs me it the `IFloat` and `IArithmetic` are sorted by order of declaration, which does not seem to be the most reliable for testing (If we change the order of declaration of two interfaces, the test might fail). For these kind of tests: maybe we should rather check that both `IFloat` and `IArithmetic` are present, regardless of their order.
Since all constraints form a conjunction, their order should not matter? 

(A side note on testing, we could add a `SLANG_CHECK_EQUAL(X, Y)` macro for testing: it would automatically output that X does not match Y, with the values. On the example above, we would have something like:
`error: getTypeFullName(typeParamTConstraintType1): "IFloat" doest not match "IArithmetic": "IArithmetic"`).

csyonghe:
Yeah, we could improve the robustness of the test by checking the existence of a conformance, instead of assuming any particular order.

csyonghe (approved):

PR #6788, opened 2025-04-10
Add a more specific diagnostic message when passing concrete value to interface-typed output parameter

### LINKED ISSUE #6779: `out` and `inout` causes l-value error with interface parameters
I initially stumbled on this with IRWArray, as below:
```slang
void func(inout IRWArray<int> arr)
{
    arr[0] = 1;
}

{
    int outputArray[10] = {};
    func(outputArray);
    return 0;
}
```

When compiling this, I get:

```slang
array.slang(9): error 30047: argument passed to parameter '0' must be l-value.
    func(outputArray);
```

But it seems to also occur with other interfaces as well, like with the code below:
```slang
interface IThing
{
    [mutating]
    void f();
}

struct TestStruct : IThing
{
    int a;
    [mutating]
    void f()
    {
        a = 1;
    }
}

void func(inout IThing t)
{
    t.f();
}

{
    TestStruct ts;
    func(ts);

    return 0;
}
```

Rewriting `func` in the below manner instead is a workaround in both cases:
```slang
void func<T:IThing>(inout T t)
```

which is somewhat interesting, as the [user guide](https://shader-slang.org/slang/user-guide/interfaces-generics.html#interface-typed-values) states that "From the Slang compiler’s view, [...] will be compiled to the same target code" in a similar situation.

juliusikkala:
Fixes #6779 with a more helpful error message.

### DIFF
--- a/source/slang/slang-check-overload.cpp
+++ b/source/slang/slang-check-overload.cpp
@@ -711,7 +711,35 @@ bool SemanticsVisitor::TryCheckOverloadCandidateTypes(
         }
         else
         {
-            arg.argExpr = coerce(CoercionSite::Argument, paramType, arg.argExpr);
+            Expr* coercedExpr = coerce(CoercionSite::Argument, paramType, arg.argExpr);
+
+            // Check if concrete-to-interface coercion caused loss of l-valueness.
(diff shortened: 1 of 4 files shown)


fairywreath:
Can you add a test?

fairywreath (approved):

Issue #6319, opened 2025-02-08
Slang needs to error out when duplicate or overlappping system value names are used


fairywreath:
Slang successfully compiles shader with duplicate built-in decorations, for example:
```
float4 main(uint vertexID : SV_VertexID, uint anotherVertexID : SV_VertexID) : SV_Position {
    return float4(vertexID * 0.1, 0, 0, 1);
}
```
This compiles successfully to invalid shaders such as:
```
struct vertexOutput_0
{
    @builtin(position) output_0 : vec4<f32>,
};

@vertex
fn main(@builtin(vertex_index) vertexID_0 : u32, @builtin(vertex_index) anotherVertexID_0 : u32) -> vertexOutput_0
{
    var _S1 : vertexOutput_0 = vertexOutput_0( vec4<f32>(f32(vertexID_0) * 0.10000000149011612f, 0.0f, 0.0f, 1.0f) );
    return _S1;
}
```
Built-in/system values cannot be duplicated in the entry point parameter list.

fairywreath:
I cannot assign myself this issue but I will be working on it.

jkwak-work:
Pushing to Q4 2025.

@fairywreath , do you have an ETA for this?
I see a PR related to this issue is merged, but I am not sure if it is directly related.

jkwak-work:
@nv-slang-bot , can you make a PR for this issue if we don't already have one?

Issue #4961, opened 2024-08-28
Slang need to print an error for an invalid system-value semantic name


jkwak-work:
**Problem description**
Slang should print an error when an invalid system-value sematic is used.
When a semantic name starts with `SV_`, it must be one of the reserved semantic names.
Slang silently allows any names but DXC prints an error.

**Repro**
The following example uses an invalid System-value-like semantic name, SV_DispatchThreadIndex.
A correct name is `SV_DispatchThreadID` with "ID" not "Index".
```
RWStructuredBuffer<int> outputBuffer;

[numthreads(1, 1, 1)]
void computeMain(uint tid : SV_DispatchThreadIndex)
{
    outputBuffer[tid] = tid;
}
```
Try the following command and observe that Slang doesn't print any errors.
```
$ slangc.sh -stage compute -entry computeMain test.slang
```
Try the following command and observe that DXC prints an error.
```
$ dxc.exe -T cs_6_4 -E computeMain test.slang
test.hlsl:4:31: error: invalid semantic 'SV_DispatchThreadIndex' for cs 6.4
void computeMain(uint tid_0 : SV_DispatchThreadIndex)
                              ^
```

fairywreath:
I am working on a fix for this and I need a few clarifications.

Currently Slang does print an error when an invalid system-value sematic name is used, but only when a target is specified during IR legalization. This does not apply for HLSL however and no error is given. When no target is specified(i.e. the command specified in the repro descirption), no error is given. I just want to confirm that the desired behavior is:
- Give an error that is backend/target agnostic when an invalid SV name is used. This will check against all valid SV names, for all targets.
- "Unimplemented" errors can be given as well in target specific code when a valid SV name is unimplemented or invalid for the specific target(this currently already exists except for HLSL).

fairywreath:
After some investigation this issue is best handled together with #411. I originally intended for a quick fix but a proper solution with semantics in the stdlib is a lot more involving since I do not have adequate knowledge(yet at least) on how write proper meta stuff, and my work on this will be paused for now.

PR #6795, opened 2025-04-12
Add SV_PointCoord to match gl_PointCoord

juliusikkala:
Adds support for [`gl_PointCoord`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_PointCoord.xhtml) in the GLSL compatibility mode and a simple test for it.

It'd probably be nice to have this functionality available even without the GLSL compatibility mode, but that requires some kind of API decisions since HLSL doesn't have any equivalent value, and it would only work with the GLSL and SPIRV targets.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -96,21 +96,7 @@ public typealias dmat4x4 = matrix<double, 4, 4>;
 
 public out float4 gl_Position : SV_Position;
 public out float gl_PointSize : SV_PointSize;
-public property float2 gl_PointCoord {
-    [require(glsl_spirv, fragment)]
-    get
-    {
-        __target_switch
-        {
-        case glsl:
-            __intrinsic_asm "(gl_PointCoord)";
-        case spirv:
-            return spirv_asm {
-                result:$$float2 = OpLoad builtin(PointCoord:float2);
-            };
-        }
-    }
-}
+public in float2 gl_PointCoord : SV_PointCoord;
 
 public in vec4 gl_FragCoord : SV_Position;
 public out float gl_FragDepth : SV_Depth;
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -96,6 +96,22 @@ public typealias dmat4x4 = matrix<double, 4, 4>;
 
 public out float4 gl_Position : SV_Position;
 public out float gl_PointSize : SV_PointSize;
+public property float2 gl_PointCoord {
+    [require(glsl_spirv, fragment)]
+    get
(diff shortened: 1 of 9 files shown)


fairywreath:
I think we should add a `SV_PointCoord` system value. Metal supports this as well through `point_coord`.

source/slang/glsl.meta.slang
```diff
@@ -96,6 +96,22 @@ public typealias dmat4x4 = matrix<double, 4, 4>;
 
 public out float4 gl_Position : SV_Position;
 public out float gl_PointSize : SV_PointSize;
+public property float2 gl_PointCoord {
```
fairywreath, line 99: Add `SV_PointCoord` for SPIRV, GLSL and Metal and use it here directly. The current system value implementation is a bit messy and you need to edit 3 separate files/function - `slang-emit-spirv.cpp`, `slang-ir-glsl-legalize.cpp`, and `slang-ir-legalize-varying-params.cpp`(for Metal).

juliusikkala: Alright, thanks for the hints - I've done this now.

csyonghe (dismissed):
Thanks, this looks good. 

We really need some refactor around system values to clean up all the duplicate/ messy adhoc handling across different backends and passes, and make it more extensible and working with the capability system etc…

csyonghe:
Needs to update the doc here too: docs/user-guide/a2-01-spirv-target-specific.md

fairywreath:
> Thanks, this looks good.
> 
> We really need some refactor around system values to clean up all the duplicate/ messy adhoc handling across different backends and passes, and make it more extensible and working with the capability system etc…

Yes I agree and I am planning to work on this. A good refactor is needed to fix these issues across all targets, for example:
- #6319
- #4961

tests/glsl-intrinsic/point-coord.slang
```diff
@@ -0,0 +1,15 @@
+//TEST:SIMPLE(filecheck=CHECK_SPIRV): -entry main -stage fragment -target spirv
+//TEST:SIMPLE(filecheck=CHECK_GLSL): -entry main -stage fragment -target glsl
+
```
fairywreath, line 3: Also add a test targetting Metal here.

fairywreath:
docs/user-guide/a2-02-metal-target-specific.md also needs to be updated, sorry for not noticing this earlier.

juliusikkala:
No problem! The Metal docs had no mention of `SV_PointSize` either but already supports it, so I documented that as well since it's kinda related.

fairywreath (approved):

PR #6797, opened 2025-04-13
Add `vk::offset` to specify member offsets for push constants

### LINKED ISSUE #6660: Attribute equivalent to layout(offset=n) in slang
If i have two push constants, one is for vertex shader, one is for fragment shader, and the struct is as below:

```
// vertex
struct Params
{
    float4x4 matrix;
}
```

```
// fragment
struct Params
{
    // layout(offset = 64)
    float2 offset;
    int numCells;
};
```
There's no attribute similar to layout(offset=n) in slang to indicate the offset relative to the previous struct size. Of course we can combine two struct into one, but this will make two struct have to share same layout attributes even if their binding, index, location, etc. are different.

Is there any way to solve this problem? Or perhaps adding this offset feature?

fairywreath:
Closes #6660.

Adds a `vk::offset` modifier to specify offsets for struct fields. The main use-case for this is for push constants to specify member variables for different shader stages. The `vk::offset` is derived from DXC, this has already been implemented there.

### DIFF
--- a/source/slang/slang-emit-glsl.cpp
+++ b/source/slang/slang-emit-glsl.cpp
@@ -5,7 +5,6 @@
 #include "slang-emit-source-writer.h"
 #include "slang-ir-call-graph.h"
 #include "slang-ir-entry-point-decorations.h"
-#include "slang-ir-insts.h"
 #include "slang-ir-layout.h"
 #include "slang-ir-util.h"
 #include "slang-legalize-types.h"
--- a/source/slang/slang-emit-glsl.cpp
+++ b/source/slang/slang-emit-glsl.cpp
@@ -245,7 +245,7 @@ void GLSLSourceEmitter::emitStructFieldAttributes(
 
     if (allowOffsetLayout)
     {
-        if (auto offsetDecoration = structKey->findDecoration<IRGLSLStructOffsetDecoration>())
+        if (auto offsetDecoration = structKey->findDecoration<IRVkStructOffsetDecoration>())
         {
             m_writer->emit("layout(offset = ");
             m_writer->emit(offsetDecoration->getOffset()->getValue());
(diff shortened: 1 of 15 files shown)


source/slang/slang-emit-glsl.cpp
```diff
@@ -234,6 +234,26 @@ void GLSLSourceEmitter::emitMemoryQualifiers(IRInst* varDecl)
     _emitMemoryQualifierDecorations(varDecl);
 }
 
+
+void GLSLSourceEmitter::emitStructFieldAttributes(
+    IRStructType* structType,
+    IRStructField* field,
+    bool allowOffsetLayout)
```
jkwak-work, line 241: Wouldn't it make more sense to take `IRGLSLStructOffsetDecoration` as the 3rd parameter?

jkwak-work (commented):
It looks good to me.
But I am not sure if there is any reason why we may not want to support this new syntax.

source/slang/slang-emit-glsl.cpp
fairywreath: The problem is this function is not called directly but is a virtual function of the base c-like emitter class and is invoked only by this base class. Passing in `IRGLSLStructOffsetDecoration` directly would mean having the caller(that is for all C-like targets) perform a `findDecoration` instead of performing the find only on the GLSL emitter implementation. The `bool allowOffsetLayout` parameter is also consistent with this function's parent virtual function `emitStructDeclarationsBlock` that has the same exact bool parameter.

fairywreath:
I renamed cases of `GLSLStructOffset` to `VkStructOffset` as this is mainly for Vulkan(regular struct member offsets are not supported in GLSL) and added an emit-spirv-via-glsl test cases in the latest commit.

source/slang/slang-emit-glsl.cpp
jkwak-work: That makes sense.
Thanks for the explanation.

jkwak-work (approved):
Looks good to me.

PR #6811, opened 2025-04-14
Update the access tokens used for auto labeling CI

sricker-nvidia:
The auto labeling CI has been failing as a result of the SLANGBOT_PAT token not having all of the necessary permissions. Change updates the tokens used for requests to use new tokens that were created with the specific permissions needed for the operations in this CI workflow.


jkwak-work (approved):
Looks good to me

PR #6818, opened 2025-04-15
Fix regression in partial specialization of existential arguments

### LINKED ISSUE #6589: Regression found by SlangPy
There is a regression found by running SlangPy test, it reports error:
tensor.slang(331): error 41014: type 'GradInOutTensor' contains fields that cannot be packed into ordinary bytes for dynamic dispatch.

Binary search points to the PR: #6487.

[repro.txt](https://github.com/user-attachments/files/19214494/repro.txt)

Repro steps:

- Clone slangpy repo, or just download `https://github.com/shader-slang/slangpy/tree/main/slangpy/slang` this folder

- rename repro.txt to repro.slang (because I cannot upload *.slang file in GitHub), copy repro.slang to `slangpy/slang` folder
- run slangc on repro.slang to whatever target

kaizhangNV:
Close #6589.

In PR #6487, we support partial specialization. However there is a corner case we didn't handle correctly.

For the IR like this:

```
%val: specialize(...) = some inst;
%arg1: specialize(...) = makeExistential(%val, ...);
%arg2: %SomeConcreteType: load(...);
call func(%arg1, %arg2);
```

when we specialize the `call func` instruct, we will also specialize the function parameters.

On our existing logic, when we find an argument is a `makeExistential`, we will always extract the existential value, and use its type as the new parameter.

But in this case, %arg1 is not fully specialized yet, so it's type will still be a `specialize`. In this case, we will change the function's first parameter from an existential type to a `specialize`. This will result in that we lose the chance to specialize the first argument in the next iteration, because the first parameter of this function is not an existential type any more.

The reason behind this is that we should always keep specializing the arguments and parameters at the same time.

So this PR just does a check before specializing the parameters that if the argument cannot be fully specialized, we won't specialize the parameter this early. Instead, we will wait for the next iteration until the argument can be specialized.

### DIFF
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -1789,11 +1789,12 @@ struct SpecializationContext
                 auto valType = val->getFullType();
                 if (!isCompileTimeConstantType(valType))
                 {
-                    // If the `argument` is not specialized yet, don't aggressively specialize the parameter.
+                    // If the `argument` is not specialized yet, don't aggressively specialize the
+                    // parameter.
                     //
-                    // If we specialize the parameter type too early, we will lose the opportunity to specialize the
-                    // callee later. The principal is to always let the specialization happen at the same time for both
-                    // on argument and parameter.
+                    // If we specialize the parameter type too early, we will lose the opportunity
+                    // to specialize the callee later. The principal is to always let the
+                    // specialization happen at the same time for both on argument and parameter.
                     continue;
                 }
                 else if (auto extractExistentialType = as<IRExtractExistentialType>(valType))
--- a/source/slang/slang-ir-specialize.cpp
+++ b/source/slang/slang-ir-specialize.cpp
@@ -1787,7 +1787,8 @@ struct SpecializationContext
                 // created.
                 //
                 auto valType = val->getFullType();
-                if (isCompileTimeConstantType(valType))
+                if (isCompileTimeConstantType(valType) &&
+                    isCompileTimeConstantType(oldParam->getFullType()))
                 {
                     if (auto extractExistentialType = as<IRExtractExistentialType>(valType))
                     {
@@ -1847,14 +1848,14 @@ struct SpecializationContext
             }
 
             // If we go here, then the parameter is either not an existential type,
-            // or the argument is not a specialized yet.
+            // or the argument/parameter is not specialized yet.
             //
             // For first case there is nothing interesting to do. The new function
             // will also have a parameter of the exact same type, and we'll use that
             // instead of the original parameter.
             //
             //
-            // For the second case if the argument is not specialized yet, don't
+            // For the second case if the argument/parameter is not specialized yet, don't
             // aggressively specialize the parameter.
             //
             // If we specialize the parameter type too early, we will lose the opportunity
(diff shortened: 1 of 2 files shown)


kaizhangNV:
I'm still working on creating a minimal shader test. Since this issue is only observed on one slangpy test, and crafting that test shader is hard. But it does pass that test on my local. So it shouldn't be a blocker for the review.

source/slang/slang-ir-specialize.cpp
```diff
@@ -1787,7 +1787,25 @@ struct SpecializationContext
...
+                    // Note we should not use `createParam` here, because this call won't assign the
+                    // parent to the new parameter, therefore during the cloning process, some
+                    // existential related IR inst could be hoisted to the global scope, which is
+                    // unexpected. Instead, we should use cloneInst here, such that the new
+                    // parameter will be inserted into the function scope.
+                    auto newParam = (IRParam*)cloneInst(&cloneEnv, builder, oldParam);
+                    newParams.add(newParam);
+                    continue;
+                }
+                else if (auto extractExistentialType = as<IRExtractExistentialType>(valType))
```
csyonghe, line 1808: This branch will never hit with the new if branch? Because is `valType` is a compiletime constant type, then it cannot be a extract existential type.

source/slang/slang-ir-specialize.cpp
```diff
@@ -1787,7 +1787,25 @@ struct SpecializationContext
                 // created.
                 //
                 auto valType = val->getFullType();
-                if (auto extractExistentialType = as<IRExtractExistentialType>(valType))
+                if (!isCompileTimeConstantType(valType))
```
csyonghe, line 1790: I wonder if we should make the fix more general, by just calling `createParam` here, but add another version of `cloneInstDecorationsAndChildren` that takes a list of parameters, and insert that list of parameters in the first block, before cloning any remaining insts.

Because we still have other branches in thie function that are making calls to `createParam` that theoretically can have the same hoisting problem.

source/slang/slang-ir-specialize.cpp
```diff
@@ -1787,32 +1787,38 @@ struct SpecializationContext
                 // created.
                 //
                 auto valType = val->getFullType();
-                if (auto extractExistentialType = as<IRExtractExistentialType>(valType))
+                if (isCompileTimeConstantType(valType) &&
+                    isCompileTimeConstantType(oldParam->getFullType()))
```
csyonghe, line 1791: why do we need to check oldParam->getFullType() as well?

csyonghe (approved):
Looks good.

source/slang/slang-ir-specialize.cpp
kaizhangNV: This is another case similar to the description, but kind of "inverse"

```
%val: concrete_type(...) = some inst;
%arg1: specialize(...) = makeExistential(%val, ...);
%arg2: %SomeConcreteType: load(...);
call func(%arg1, %arg2);

%func: (void, %concrete_type, %specialize)
```

in this case, we won't specialize the call site because the parameter 2 is not an existential type yet. But based on currently logic, we will specialize parameter 2 in `createExistentialSpecializedFunc`, because `%arg1` has a concrete wrapped value (`%val`). But if we do that, the call site will also lose the chance to be specialized because all the parameters of `%func` are not existential type anymore.

The principle in this change is to always make argument and parameter to the concrete type at the same time.

Issue #6370, opened 2025-02-16
Loop fails to unroll and invalid SPIR-V is generated for OpImageFetch


mifreundorfer:
Tested with v2025.5.1, compiling the following code:

```
Texture2D<float4> input;
RWTexture2D<float4> output;

float3 LoadSourceColor(uint2 pixelPos, int2 offset) {
    return input.Load(int3(pixelPos, 0), offset).rgb;
}

[shader("compute")]
void Test(uint3 thread_index : SV_DispatchThreadID) {
    float4 result = 0;

    [unroll]
    for (int i = 0; i < 3 * 3 - 1; i++) {
        //result.rgb += input.Load(int3(thread_index.xy, 0), int2( i % 3, i / 3 )).rgb;
        result.rgb += LoadSourceColor(thread_index.xy, int2( i % 3, i / 3 )).rgb;
    }

    output[thread_index.xy] = result;
}
```

- Slang fails to unroll the loop when a function is called inside the loop. Inlining the load correctly unrolls the loop.
- When the loop fails to unroll, invalid SPIR-V is generated for `OpImageFetch` with `ConstOffset` instead of `Offset`, using a non constant value.

csyonghe:
[unroll] maps to a downstream decoration/hint and does not instructs slang compiler to unroll the loop.

You can use [ForceUnroll] to make unrolling happen in Slang.

mifreundorfer:
[ForceUnroll] does indeed work as intended. Thank you for the hint.

There is still an issue though. If you remove the [unroll] attribute and leave the loop entirely undecorated, invalid SPIR-V is generated when the function is called inside the loop. If the function is not called and the function body is inlined (see the commented line), everything works as expected.

The issue is, `OpImageFetch` is generated with the `ConstOffset` flag set, but the variable referenced as the offset is not constant. The flag should be set to `Offset` instead (it references the loop counter). `ImageGatherExtended` must be specified in the capabilites for this to work.

mifreundorfer:
Note that in glsl `texelFetchOffset` requires that the offset parameter must be a compile time constant. I am not sure if the same applies to hlsl. 

If i emulate this behaviour, propagating it to the offset paramter of `LoadSourceColor` manually, by setting it to `constexpr`, the behaviour is as i would expect. The loop is unrolled and the generated SPIR-V references constant values for the offsets.

fairywreath:
[...]

PR #6820, opened 2025-04-15
check constexpr of the nested calls

gtong-nv:
This change aims to report error when the nested func requires constexpr, and the caller calls wrapper func with non-cost params.
For example, the following should failed to compile because RWTexture2D<T>::Load requires offset to be constexpr.

```
Texture2D<float3> input;
RWTexture2D<float3> output;

float3 LoadSourceColor(int2 offset) {
    int3 pixelPos = int3(0,0,0);
    return input.Load(pixelPos, offset);
}

[shader("compute")]
void Test() {
    float3 result = 0;

    [ForceUnroll]
    for (int i = 0; i < 3; i++) {
        //result += input.Load( int3(0,0,0), int2( i, i ));
        result += LoadSourceColor(int2( i, i ));
    }

    output[0] = result;
}
```
Fixes: #6370

### DIFF
--- a/source/slang/slang-ir-constexpr.cpp
+++ b/source/slang/slang-ir-constexpr.cpp
@@ -178,7 +178,10 @@ IRLoop* isLoopPhi(IRParam* param)
 bool opCanBeConstExprByBackwardPass(IRInst* value)
 {
     if (value->getOp() == kIROp_Param)
-        return isLoopPhi(as<IRParam, IRDynamicCastBehavior::NoUnwrap>(value));
+    {
+        return true;
+        // return isLoopPhi(as<IRParam, IRDynamicCastBehavior::NoUnwrap>(value));
+    }
     if (opCanBeConstExpr(value->getOp()))
         return true;
     if (auto callInst = as<IRCall>(value))
--- a/source/slang/slang-ir-constexpr.cpp
+++ b/source/slang/slang-ir-constexpr.cpp
@@ -177,13 +177,11 @@ IRLoop* isLoopPhi(IRParam* param)
 
 bool opCanBeConstExprByBackwardPass(IRInst* value)
 {
-    if (value->getOp() == kIROp_Param)
+    if (opCanBeConstExpr(value->getOp()))
     {
         return true;
-        // return isLoopPhi(as<IRParam, IRDynamicCastBehavior::NoUnwrap>(value));
     }
-    if (opCanBeConstExpr(value->getOp()))
-        return true;
+
     if (auto callInst = as<IRCall>(value))
     {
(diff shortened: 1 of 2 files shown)


source/slang/slang-ir-constexpr.cpp
```diff
@@ -343,154 +344,175 @@ bool propagateConstExprForward(PropagateConstExprContext* context, IRGlobalValue
...
-                        auto genericInst = as<IRGeneric>(specInst->getBase());
-                        if (!genericInst)
-                            break;
-
-                        auto returnVal = findGenericReturnVal(genericInst);
-                        if (!returnVal)
-                            break;
-
-                        callee = returnVal;
+                        changedInThisInst = checkInstConstExprRecursively(context, ii);
```
juliusikkala, line 448: I think this should be `changedThisIteration |= checkInstConstExprRecursively(context, ii);`, otherwise `changedThisIteration` may reset back to false if a later instruction didn't change.

source/slang/slang-ir-constexpr.cpp
```diff
@@ -343,154 +344,175 @@ bool propagateConstExprForward(PropagateConstExprContext* context, IRGlobalValue
...
-                            }
-                        }
-                    }
-                }
+        bool changedThisIteration = false;
+        for (auto bb = code->getLastBlock(); bb; bb = bb->getPrevBlock())
+        {
+            for (auto ii = bb->getLastInst(); ii; ii = ii->getPrevInst())
+            {
+                changedThisIteration = checkInstConstExprRecursively(context, ii);
```
juliusikkala, line 510: Same here, `changedThisIteration` may be reset back to false if a later iteration is false.

source/slang/slang-ir-constexpr.cpp
juliusikkala: The recursion going through this line causes a stack overflow if a recursive function is given in input, like in `tests/diagnostics/recursion.slang`.

source/slang/slang-ir-constexpr.cpp
```diff
@@ -343,154 +344,175 @@ bool propagateConstExprForward(PropagateConstExprContext* context, IRGlobalValue
...
+            auto arg = inst->getOperand(aa);
+            if (isConstExpr(arg))
+                continue;
+
+            if (!opCanBeConstExprByBackwardPass(arg))
+                continue;
+
+            if (maybeMarkConstExprBackwardPass(context, arg))
+            {
+                changedInThisInst = true;
```
kaizhangNV, line 367: I'm wondering for all of those `changedInThisInst = true;` can we just early return? It doesn't look like we need to go through everything.

PR #6822, opened 2025-04-15
Add CoopVec load/store pointer overloads

fairywreath:
Adds CoopVec load/store function overloads to accept `Ptr` types. SPIRV accepts the pointers as pointers to an array of scalars/vectors. Hence the user's pointer-to-scalar must be translated to a pointer-to-array, more specifically a runtime/unsized array, where a `stride` decoration is required.


tests/cooperative-vector/load-store-pointer.slang
```diff
@@ -0,0 +1,35 @@
...
+// CHECK-NEXT: 2
+// CHECK-NEXT: 3
+// CHECK-NEXT: 4
+
+struct A
+{
+int32_t* ptr;
+}
+
+//TEST_INPUT:ubuffer(data=[1 2 3 4 5 6 7 8], stride=4),name=buf
```
csyonghe, line 21: ptr isn't initialized with anything that is actually valid to read/write to.

TEST_INPUT allows you to assign a `ubuffer` to a pointer value, can you try that instead and turn this into an executable test?

csyonghe: e.g.
```
//TEST_INPUT: set pObj = ubuffer(data=[2 0 0 0], stride=4);
uniform Obj* pObj;
```

fairywreath: Thanks - is it possible to have these as output?

csyonghe: Should be? I haven’t tried that yet…

csyonghe (approved):

csyonghe:
I think we also need overloads of coopMatMulAdd and friends that takes pointers for the matrix instead of a ByteAddressBuffer + offset.

Issue #6766, opened 2025-04-09
[GCP] Skip CoopVec tests when the hardware doesn't support it


jkwak-work:
## Problem

When we run slang-test, cooperative vector tests are failing when the hardware doesn't support.
This causes problem with the new GCP workflow machines, because they work only with the older graphics drivers.

## Proposed solution

We should skip the cooperative vector tests when the feature is not available on the hardware/driver.

We can add a new feature keyword `coopvec` for render-test and add the following command-line arguments for all of the cooperative vector tests.
```
-render-features coopvec
```

juliusikkala:
I'm also having trouble with CoopVec tests failing on my computer (non-Nvidia GPU), all renders from `slang-unit-test-tool/RecordReplay.internal` fail with the following:

```
[Vulkan Loader] ERROR:          ICD associated with VkPhysicalDevice does not support GetPhysicalDeviceCooperativeVectorPropertiesNV
```

The problem here seems to be that `vkGetPhysicalDeviceCooperativeVectorPropertiesNV` is attempted to be used before checking that the relevant extension is present on the physical device. The function pointer check doesn't help, as the vulkan loader happily hands out a function pointer that just issues the above error when called.

https://github.com/shader-slang/slang/blob/0f0818722268464775981520c2d6e975a9cf14fa/tools/gfx/vulkan/vk-api.cpp#L89-L98

jkwak-work:
I am escalating this issue and assigned to myself.
This is a blocker for GCP to work.

PR #6829, opened 2025-04-16
Coopvec tests are expected to fail on GCP servers

jkwak-work:
Temporarily disables cooperative vector tests, because GCP servers cannot use the latest graphics driver and coopvec tests are always failing.


jkwak-work:
Some of gfx-unit-tests are also failing for the same reason that the driver doesn't support the coop vec.
- https://github.com/shader-slang/slang/issues/6766

jkwak-work:
Dropping this PR because we have a proper fix now.
- https://github.com/shader-slang/slang/pull/6849

PR #6830, opened 2025-04-16
Add subscript operator support in cuda

mkeshavaNV:
This CL adds support for the subscript operator for Read Only
textures in cuda. Also adds a test for this.

Now that we support .Load for read only textures in cuda, an existing test also needs to be modified.


source/slang/hlsl.meta.slang
```diff
@@ -3610,14 +3610,42 @@ extension _Texture<T,Shape,isArray,0,sampleCount,0,isShadow,isCombined,format>
...
+                {
+                    switch(Shape.flavor)
+                    {
+                    case $(SLANG_TEXTURE_1D):
+                        __intrinsic_asm "<PTX does not support 1D texture read>";
+                    case $(SLANG_TEXTURE_2D):
+                        __intrinsic_asm "tex2Dfetch_int<$T0>($0, ($1).x, ($1).y)";
+                    case $(SLANG_TEXTURE_3D):
+                        __intrinsic_asm "tex3Dfetch_int<$T0>($0, ($1).x, ($1).y, ($1).z)";
+					case $(SLANG_TEXTURE_CUBE):
```
skallweitNV, line 3644: are there no PTX intrinsics for cube map reads?

mkeshavaNV: nope, cubemap requires float coordinates. Confirmed with the ptx team as well. I even saw error:
`line 84; error   : Floating-point coordinates required`

source/slang/hlsl.meta.slang
```diff
@@ -3610,14 +3610,42 @@ extension _Texture<T,Shape,isArray,0,sampleCount,0,isShadow,isCombined,format>
...
+                    default:
+                        __intrinsic_asm "<invalid intrinsic>";
+                    }
+                }
+                else
+                {
+                    switch(Shape.flavor)
+                    {
+                    case $(SLANG_TEXTURE_1D):
+                        __intrinsic_asm "<PTX does not support 1D texture read>";
```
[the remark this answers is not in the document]
mkeshavaNV: ok seems like that was because of missing default. I've added that back, and the static_assert. Please check.

prelude/slang-cuda-prelude.h
```diff
@@ -3366,3 +3366,66 @@ struct TensorView
...
+
+template<typename T>
+SLANG_FORCE_INLINE SLANG_CUDA_CALL T tex3Dfetch_int(CUtexObject texObj, int x, int y, int z)
+{
+    T result;
+    float dummy;
+    asm("tex.3d.v4.f32.s32 {%0, %1, %2, %3}, [%4, {%5, %6, %7, %8}];"
+        : "=f"(result), "=f"(dummy), "=f"(dummy), "=f"(dummy)
+        : "l"(texObj), "r"(x), "r"(y), "r"(z), "r"(z));
+    // Note: The repeated z is a WAR for a bug in PTX. That is a dummy parameter.
```
szihs, line 3406: can you add more details. Does PTX only support vec4 fetch?

expipiplus1: Can we write a test which will start failing when this bug is fixed downstream?

source/slang/hlsl.meta.slang
```diff
@@ -3610,14 +3610,41 @@ extension _Texture<T,Shape,isArray,0,sampleCount,0,isShadow,isCombined,format>
...
             __intrinsic_asm ".Load";
+		case cuda:
+            if (isArray != 0)
+                {
+     				static_assert(Shape.flavor == $(SLANG_TEXTURE_2D) || Shape.flavor == $(SLANG_TEXTURE_3D),
+                              "Integer coordinates are supported for texture reads only for 2D and 3D textures.");
+
+                    switch(Shape.flavor)
+                    {
+                    case $(SLANG_TEXTURE_2D):
```
expipiplus1, line 3629: Do we need to handle 3d here as well, inline with the static assert (I think that an if/else here would be clearer)

expipiplus1 (changes requested):

prelude/slang-cuda-prelude.h
mkeshavaNV: oh that is not a bug. Fixed the note after I confirmed with the ptx team. It is how the 3D textures are addressed.

From the docs:
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#texture-instructions-tex

```
Operand c is a scalar or singleton tuple for 1d textures; is a two-element vector for 2d textures; 
and is a four-element vector for 3d textures, 
```

source/slang/hlsl.meta.slang
mkeshavaNV: I don't think we support 3D array textures in slang? I could not find `Texture3DArray`. I have updated the static_assert to be clearer, and also made this an `if`. Please let me know if this is OK.

Issue #6784, opened 2025-04-10
[Diagnostic]Windows: slangc generate no code for compute shader


RickSchanze:
Compute shader below generate no code using command line:
```
slangc.exe ./irradiance.slang -target glsl -o irradiance.glsl -entry compute
```
irradiance.slang:
```
struct Params {
    float output_size_x;
    float output_size_y;
    float sample_delta;
    float sky_intensity;
}

ConstantBuffer<Params> params;
Texture2D<float3> input_equirect_map;
SamplerState common_samler;
RWTexture2D<float4> output_irradiance;
static const float PI = 3.1415926535;

float2 DirToUV(float3 dir) {
    float phi = atan2(dir.y, dir.x) + PI;      
    float theta = acos(dir.z);                
    return float2(phi / (2 * PI), theta / PI);
}

float3 SphericalToDir(float theta, float phi) {
    return float3(
        sin(theta) * cos(phi),
        sin(theta) * sin(phi),
        cos(theta)
    );
}

[shader("compute")]
[numthreads(16, 16, 1)]
void compute(uint3 tid: SV_DispatchThreadID) {
    float2 uv = (tid.xy + 0.5) / float2(params.output_size_x, params.output_size_y);
    float phi = uv.x * 2 * PI;
    float theta = (1 - uv.y) * PI; 
    float3 N = float3(sin(theta) * cos(phi), cos(theta), sin(theta) * sin(phi));
    float3 up = (0, 1, 0);
    float3 right = normalize(cross(up, N));
    up = normalize(cross(N, right));

    float3 irradiance = 0;
    float nSample = 0;

    for (float sample_phi = 0.0; sample_phi < 2.0 * PI; sample_phi += params.sample_delta) {
        for (float sample_theta = 0.0; sample_theta < 0.5 * PI; sample_theta + params.sample_delta) { //
            float3 tangent_sample = SphericalToDir(sample_theta, sample_phi);
            float3 world_sample = tangent_sample.x * right + tangent_sample.y * up + tangent_sample.z * N;
            float2 sample_uv = DirToUV(world_sample);
            float3 color = input_equirect_map.Sample(common_samler, sample_uv);
            irradiance += color * cos(theta) * sin(theta);
            nSample++;
        }
    }
    irradiance = PI * irradiance / nSample;
    output_irradiance[tid.xy] = float4(irradiance * params.sky_intensity, 1.0f);
}
```
And output is:
```
#version 450
layout(row_major) uniform;
[...]

PR #6832, opened 2025-04-16
Fix#6784 - Add diagnostic for for-loop update expressions with no effect

szihs:
This commit introduces a new warning diagnostic (30506) for cases where the update expression in a for-loop does not modify any variables, potentially leading to infinite loops. The implementation includes checks in the `SemanticsStmtVisitor` to identify such expressions and provide appropriate warnings. Additionally, tests have been added to verify the new diagnostic behavior.

### DIFF
--- a/tests/diagnostics/for-loop-warning.slang
+++ b/tests/diagnostics/for-loop-warning.slang
@@ -56,5 +56,8 @@ float doSomething(int x)
     {
         i--;
     }
+    for (int i = 1; i < 100; i + 2) //warn
+    {
+    }
     return 0.0;
 }
--- a/tests/diagnostics/for-loop-warning.slang.expected
+++ b/tests/diagnostics/for-loop-warning.slang.expected
@@ -21,6 +21,9 @@ tests/diagnostics/for-loop-warning.slang(45): warning 30505: the loop runs for 0
 tests/diagnostics/for-loop-warning.slang(51): warning 30505: the loop runs for 0 iterations and will be removed.
     for (int i = 1; i > 1; i--) // warn
     ^~~
+tests/diagnostics/for-loop-warning.slang(59): warning 30506: the for loop update expression '+' has no effect; this will result in an infinite loop
+    for (int i = 1; i < 100; i + 2) //warn
+                               ^
 tests/diagnostics/for-loop-warning.slang(23): warning 30504: the for loop is statically determined to terminate within 3 iterations, which is less than what [MaxIters] specifies.
     [MaxIters(6)] // warn
      ^~~~~~~~
--- a/source/slang/slang-check-stmt.cpp
+++ b/source/slang/slang-check-stmt.cpp
@@ -240,6 +240,31 @@ void SemanticsStmtVisitor::visitForStmt(ForStmt* stmt)
     if (stmt->sideEffectExpression)
     {
         stmt->sideEffectExpression = CheckExpr(stmt->sideEffectExpression);
+
+        // Check for common mistake where update expression has no side effect
+        auto sideEffect = stmt->sideEffectExpression;
+        if (auto infixExpr = as<InfixExpr>(sideEffect))
+        {
+            // Check if the operation is a binary operation like "+" and not "+="
+            auto funcExpr = as<DeclRefExpr>(infixExpr->functionExpr);
+            if (funcExpr && funcExpr->declRef.getDecl())
+            {
+                // Check if the operator name doesn't contain "="
+                auto opName = funcExpr->declRef.getName();
+                if (opName && !String(opName->text).contains("="))
+                {
+                    // This is a potential issue - binary operators without "=" in for-loop updates
+                    // usually don't modify variables
+                    StringBuilder builder;
+                    builder << opName->text;
+
+                    getSink()->diagnose(
+                        sideEffect,
+                        Diagnostics::forLoopUpdateExpressionHasNoEffect,
+                        builder.produceString());
+                }
+            }
+        }
     }
     subContext.checkStmt(stmt->statement);
 
--- a/source/slang/slang-diagnostic-defs.h
+++ b/source/slang/slang-diagnostic-defs.h
@@ -1397,6 +1397,11 @@ DIAGNOSTIC(
     loopRunsForZeroIterations,
     "the loop runs for 0 iterations and will be removed.")
 DIAGNOSTIC(
+    30506,
+    Warning,
+    forLoopUpdateExpressionHasNoEffect,
+    "the for loop update expression '$0' has no effect; this will result in an infinite loop")
+DIAGNOSTIC(
     30510,
     Error,
     loopInDiffFuncRequireUnrollOrMaxIters,


csyonghe (dismissed):
The name based checking is a big code smell, and always should be avoided.

The right way to check is to implement a general side effect detection logic in post lowering validation phase next to where we are detecting missing returns. If we see a loop whose condition is a constant and the loop does not have any exit jumps, then there should be a warning.

szihs:
> The name based checking is a big code smell, and always should be avoided.
> 
> The right way to check is to implement a general side effect detection logic in post lowering validation phase next to where we are detecting missing returns. If we see a loop whose condition is a constant and the loop does not have any exit jumps, then there should be a warning.

@csyonghe  by post lowering validation  does that mean 
slang-lower-to-ir.cpp`
checkForMissingReturns(module, compileRequest->getSink(), CodeGenTarget::None, true);`

csyonghe:
Correct, we should add another pass after the check missing return pass to validate this.

szihs:
Okay. Would require to redo the logic.

tests/diagnostics/for-loop-warning.slang
```diff
@@ -56,5 +58,9 @@ float doSomething(int x)
     {
         i--;
     }
+    for (int i = 1; i < 10; i+1) // warn.
```
szihs, line 61: @csyonghe - I observed that `Diagnostics::potentialInfiniteLoop` is only emitted for first occurence in `doSomething`.  In the code we are doing a graph search - so ideally this should be caught, but i am not able to wrap my head on this.

source/slang/slang-ir-potential-infinite-loop.cpp
```diff
@@ -0,0 +1,131 @@
...
+    for (auto childInst : inst->getChildren())
+    {
+        checkForPotentialInfiniteLoopsInInst(childInst, sink, diagnoseWarning);
+    }
+}
+
+void checkForPotentialInfiniteLoops(IRModule* module, DiagnosticSink* sink, bool diagnoseWarning)
+{
+    // Look for loops in the module
+    checkForPotentialInfiniteLoopsInInst(module->getModuleInst(), sink, diagnoseWarning);
```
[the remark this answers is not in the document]
szihs: We start procesing all IR instructions - The IRloop check is done inside this function at L105

szihs:
I've been working on enhancing Slang's IR analysis to detect potential infinite loops, specifically focusing on a case where a loop's condition, 0.0 < (0.5 * %PI), should evaluate to a constant true, thereby indicating an infinite loop. Initially, my checkForPotentialInfiniteLoops pass and its helper getConstantBoolValue were modified to identify such constant conditions. I discovered through GDB that this check was happening before critical IR optimization passes, like simplifyIR (which includes Sparse Conditional Constant Propagation - SCCP), had a chance to run. Consequently, %PI wasn't being resolved to its numerical value, and the loop condition wasn't simplified to a boolean literal.


The current status is that I attempted to use tryConstantFoldInst within my analysis pass as a way to locally evaluate the condition. However, GDB traces revealed that tryConstantFoldInst, when called with the default SCCPContext (where code = nullptr), doesn't fully resolve expressions involving non-literal IR instructions within function bodies (like the IRMul for 0.5 * %PI) if their operands (like %PI itself, if it's a global symbol) aren't direct literals. The getLatticeVal helper within this context treats such cases as Flavor::Any. The primary issue seems to be that my loop check pass executes before the main simplifyIR pass, which is responsible for the more global and thorough constant folding that would simplify the problematic condition. The definition of %PI in the IR is still needed to confirm if it's a true compile-time constant that simplifyIR should be folding.


We do have warnings working as seen in 

 tests/diagnostics/bug-6784-for-loop-warning.slang          | 16 ++++++++++++
 tests/diagnostics/bug-6784-for-loop-warning.slang.expected |  8 ++++++

@csyonghe - Can you suggest on if we can move the detection to a later stage ? or does the current code look good enough ?
- Note we still  face the issue as in https://github.com/shader-slang/slang/pull/6832/#issuecomment-2840722205

tests/diagnostics/for-loop-warning.slang
szihs: I think its because we return warning from outer loop... I think it should be suffiient.

szihs:
Investigating how we can do this check in a later stage since the constant folding doesn't allow us to do validate in the current `generateIRForTranslationUnit` func

tests/diagnostics/bug-6784-for-loop-warning.slang.expected
```diff
@@ -0,0 +1,8 @@
+result code = 0
+standard error = {
```
csyonghe: Why is this checked in?

tests/diagnostics/for-loop-warning.slang.expected
```diff
@@ -30,6 +30,9 @@ tests/diagnostics/for-loop-warning.slang(28): warning 30504: the for loop is sta
 tests/diagnostics/for-loop-warning.slang(32): warning 30504: the for loop is statically determined to terminate within 5 iterations, which is less than what [MaxIters] specifies.
     [MaxIters(6)] // warn
      ^~~~~~~~
+tests/diagnostics/for-loop-warning.slang(10): warning 41008: loop might execute indefinitely
+    for (int j = 0; j < 3; j += 0) // warn.
```
csyonghe, line 34: This line is not necessary.

tests/diagnostics/bug-6784-for-loop-warning.slang.expected
```diff
@@ -0,0 +1,8 @@
...
+tests/diagnostics/bug-6784-for-loop-warning.slang(10): warning 41008: loop might execute indefinitely
+            for (float sample_theta = 0.0; sample_theta < 0.5 * PI; sample_theta + sample_delta) { //
```
csyonghe: This line is not necessary

source/slang/slang-emit.cpp
```diff
@@ -164,7 +165,8 @@ EntryPointLayout* findEntryPointLayout(ProgramLayout* programLayout, EntryPoint*
     return nullptr;
 }
 
-/// Given a layout computed for a scope, get the layout to use when lookup up variables.
+/// Given a layout computed for a scope, get the layout to use when lookup up
```
csyonghe: These changes are unrelated to the issue. If needed, please submit them in a separate PR.

tests/diagnostics/for-loop-warning.slang.expected
szihs: The CL emits additional (correct - 0 update) warning - so this is required as `expected ` output changes

tests/diagnostics/bug-6784-for-loop-warning.slang.expected
szihs: Removed file.

tests/diagnostics/bug-6784-for-loop-warning.slang.expected
szihs: removed file

source/slang/slang-emit.cpp
szihs: fixed.

szihs:
@csyonghe - updated the review comments. 
We should keep the line for warning for [tests/diagnostics/for-loop-warning.slang.expected](https://github.com/shader-slang/slang/pull/6832/files#diff-39894ad7844893454431db5b62797bb85a645933a6090b1ec39031d8fcd50203)

szihs:
Ping @csyonghe for review.

source/slang/slang-ir-potential-infinite-loop.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+                if (alwaysTakenBlock == breakBlock)
+                {
+                    // The constant condition forces the loop to take a path that is the break
+                    // block.
+                    return true;
+                }
+
+                if (neverTakenBlock == breakBlock)
+                {
```
csyonghe, line 100: This doesn’t seem to be correct? while(true) return; will exit.

csyonghe (commented):
I think the algorithm is more complicated than it needs to be. We shouldn’t be implementing any constant folding logic in this pass, and instead should wait until that constant folding is done first before running this pass.


The DFS search also seemed like an over kill. We just need to detect simplest possible mistakes. The algorithm should be:
1. Collect all exiting jumps within the loop.
2. If there are no exiting jump, warn.

exiting jumps can be collected by running loop analysis to get all blocks belonging to a loop, and see if any of the blocks ends with a jump to the break block, or a return/ discard.

szihs:
> I think the algorithm is more complicated than it needs to be. We shouldn’t be implementing any constant folding logic in this pass, and instead should wait until that constant folding is done first before running this pass.
> 
> The DFS search also seemed like an over kill. We just need to detect simplest possible mistakes. The algorithm should be:
> 
> 1. Collect all exiting jumps within the loop.
> 2. If there are no exiting jump, warn.
> 
> exiting jumps can be collected by running loop analysis to get all blocks belonging to a loop, and see if any of the blocks ends with a jump to the break block, or a return/ discard.

instead should wait until that constant folding is done first before running this pass.

The DFS search also seemed like an over kill. We just need to detect simplest possible mistakes. The algorithm should be:

Collect all exiting jumps within the loop.
If there are no exiting jump, warn.
exiting jumps can be collected by running loop analysis to get all blocks belonging to a loop, and see if any of the blocks ends with a jump to the break block, or a return/ discard.
While working on the issue to test driven development.

1. Ensure that the newly added for loop test works. Us `slang-test`.
- [ ] `tests/diagnostics/bug-6784-for-loop-warning.slang` 
- [ ] `tests/diagnostics/for-loop-warning.slang` and 
- [ ] `tests/bugs/gh-6784.slang`
2.  Remove the constant folding as directed.
3. Check if the tests work after removing DFS. (as from review comment)

szihs:
@claude  can you help work on this review comment to update the existing PR.

> I think the algorithm is more complicated than it needs to be. We shouldn’t be implementing any constant folding logic in this pass, and instead should wait until that constant folding is done first before running this pass.
> 
> The DFS search also seemed like an over kill. We just need to detect simplest possible mistakes. The algorithm should be:
> 
> 1. Collect all exiting jumps within the loop.
> 2. If there are no exiting jump, warn.
> 
> exiting jumps can be collected by running loop analysis to get all blocks belonging to a loop, and see if any of the blocks ends with a jump to the break block, or a return/ discard.

instead should wait until that constant folding is done first before running this pass.

The DFS search also seemed like an over kill. We just need to detect simplest possible mistakes. The algorithm should be:

Collect all exiting jumps within the loop.
If there are no exiting jump, warn.
exiting jumps can be collected by running loop analysis to get all blocks belonging to a loop, and see if any of the blocks ends with a jump to the break block, or a return/ discard.
While working on the issue to test driven development.

1. Ensure that the newly added for loop test works. Us `slang-test`.
- [ ] `tests/diagnostics/bug-6784-for-loop-warning.slang` 
- [ ] `tests/diagnostics/for-loop-warning.slang` and 
- [ ] `tests/bugs/gh-6784.slang`
2.  Remove the constant folding as directed.
3. Check if the tests work after removing DFS. (as from review comment)

jhelferty-nv:
@szihs I see no activity in 2 wks. Is this one blocked?

szihs:
It's not blocked, but I have not been able to spend any cycles on this.

szihs:
Closing this PR in favor of https://github.com/shader-slang/slang/pull/8422

PR #6835, opened 2025-04-16
Eliminate back-reference in ChildStmt

tangent-vector:
This change is part of a larger effort to improve the code for AST serialization in the Slang compiler.

Tree structures are understandably easier to serialize than DAGs, and DAGs are easier than fully generaal graphs.

The Slang AST nodes form a tree structure... except when they don't. Among the exceptions to nice tree-structured ASTs are:

1. References to `Decl`s are encoded as pointers to the AST `Decl` nodes themselves. This can result in cycles in the graph, and requires care in serialization.

2. Nodes that inherit from `Val` represent, well, *values* instead of actual pieces of syntax, and as such they are deduplicated so that identical values will (hopefully) be identical pointers. This results in a DAG structure for `Val`s, but at least it's not a general graph (except for cycles that go through a `Decl`).

3. There are some minor cases of DAG-structured sharing that the parser can introduce to deal with cases when a traditional-style declaration includes multiple declarators. E.g., given:

  ```
  static int a, b;
  ```

  The resulting `DeclGroup` will include distinct `Decl`s for `a`
  and `b`, which will share the `static` modifier through a
  `SharedModifiers` node, and the `int` type specifier through a
  `SharedTypeExpr` node.

  This duplication can be ignored, for the purposes of serialization,
  since duplicating those parts of the AST has no major down-sides.

4. There is the case of `ChildStmt`, used for things like `break` and `continue`, which stores a direct `Stmt*` to the enclosing parent statement being targetted. Storing the target is useful so that IR lowering doesn't need to repeat the work that the semantic checking logic did to associate each child statement with its parent.

  The parent link inside of `ChildStmt` creates a cycle in the AST
  `Stmt` hierarchy, since the outer statement contains the inner,
  and the inner statement stores a pointer to the outer.

This change eliminates the last of these sources of complication for AST serialization, by changing the `ChildStmt` type to stored an integer ID for the enclosing statement that it matches to, and having each `BreakableStmt` (used to represent the outer `switch`, or loop, or whatever) generate its own unique ID as part of semantic checking.

Note: if necessary, it is reasonable for the outer statement to have its unique ID generated as part of parsing, rather than semantic checking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           - { os: macos, compiler: gcc }
           - { os: macos, compiler: cl }
         include:
-          - { os: linux, runs-on: ubuntu-20.04 }
+          - { os: linux, runs-on: ubuntu-latest }
           - { os: macos, runs-on: macos-latest }
           - { os: windows, runs-on: windows-latest }
           # Warnings are treated as errors by default.
@@ -58,7 +58,7 @@ jobs:
             platform: aarch64
             test-category: smoke
             full-gpu-tests: false
-            runs-on: ubuntu-22.04-arm
+            runs-on: ubuntu-24.04-arm
             has-gpu: false
             build-llvm: false
           # Self-hosted full gpu build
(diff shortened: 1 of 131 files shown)


csyonghe (dismissed):

tangent-vector:
Sigh. This approach is incompatible with the current AST serialization logic, because serialization and deserialization of a module will automatically use the same IDs that were in place at the time it was serialized (and ignore the fact that the current `ASTBuilder` might have its own tracking). That's frustrating, but I think it can be fixed somewhat reasonably.

csyonghe:
Should this id be tracked on the ModuleDecl instead?

tangent-vector:
AFAICT, the test failures on `(windows, release, cl, x86_64)` aren't really things related to the content of this PR.

csyonghe (approved):

PR #6836, opened 2025-04-16
CI check for gcp runners

aasgaonkar:


jkwak-work:
Changing this to "draft" because it is intended only for a testing.

PR #6837, opened 2025-04-16
Remove support for "module libraries"

tangent-vector:
This change *also* removes support for the `-r` command-line option to `slangc`, and the ability to pass `.slang-module` files as command-line arguments.

It may turn out that some (or all) of the functionality being removed in this change is required by at least some existing users of the slang API or command-line compiler. The intention of this change is to provide a way to evaluate whether these changes would break any user code that is being tested by various automated CI processes.


tangent-vector:
@juliusikkala Thanks for the feedback!

This PR is definitely just an attempt on my part to find the boundaries between what is and isn't being used when it comes to this functinality, so knowing that passing a `.slang-module` on the command line is a useful behavior helps a lot.

Also, FWIW, I strongly agree that the behavior you describe is desirable, for the reasons you went into. It's not that I don't think `slangc` should have that behavior, and its more just that the *way* that behavior is currently implemented in the compiler is tangled up in several overly-complicated features, so that ideally we could do something a lot more narrow that would meet your needs.

I'm going to keep iterating on this work, but for now your comment plus the various test failures (including some user code integration tests I've run internally) tell me that I can't just do the easy thing and eliminate this functionality entirely.

csyonghe:
Closing as it seems we might not be able to remove this just yet.

PR #6841, opened 2025-04-16
Skip cooperative vector tests when the graphics driver doesn't support it

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6766

Currently DX12 doesn't officially support the cooperative vector yet.
And this PR will be effectively disabled the coopvec tests with DX12 always.

This will be addressed later:
1. when DX12 officially supports it next month, or
2. if we can upgrade the graphics driver on GCP to the latest.


tangent-vector (approved):

jkwak-work:
I verified the followings.

With this PR, cooperative-vector tests are:
- passing for vk on non-GCP servers
- ignored for dx12 on non-GCP servers
- ignored for vk and dx12 on GCP servers

PR #6844, opened 2025-04-17
Add Yet Another Source Code Generator

tangent-vector:
This change introduces an offline source code generation tool, provisionally called `fiddle`. More information about the design of the tool can be found in `tools/slang-fiddle/README.md`.

Yes... this is yet another code generator in a project that already has too many. Yes, this could easily be a very obvious instnace of [XKCD 927](https://xkcd.com/927/).

This change is part of a larger effort to change how the AST types are being serialized, and the way code generation for them is implemented.

Right now, the source code for the new tool is being checked in and the relevant build step is enabled, just to make sure everything is working as intended, but please note that this change does *not* introduce any code in the repository that actually makes use of the new generator. All of the AST-related reflection information that feeds the current serialization system is still being generated using `slang-cpp-extractor`.

The design of the new tool is primarily motivated by the new approach to serialization that I'm implementing, and once that new approach lands we should be able to deprecate the `slang-cpp-extractor`.

In addition, the new tool should in principle be able to handle many of the kinds of code generation tasks that are currently being implemented with other tools like `slang-generate` (used for the core and glsl libraries). This tool should also be well suited to the task of generating more of the code related to the IR instructions.

### DIFF
--- a/tools/slang-fiddle/slang-fiddle-template.cpp
+++ b/tools/slang-fiddle/slang-fiddle-template.cpp
@@ -5,552 +5,526 @@
 
 namespace fiddle
 {
-    struct TextTemplateParserBase
+struct TextTemplateParserBase
+{
+protected:
+    TextTemplateParserBase(
+        SourceView* inputSourceView,
+        DiagnosticSink* sink,
+        UnownedStringSlice source)
+        : _inputSourceView(inputSourceView)
+        , _sink(sink)
+        , _cursor(source.begin())
+        , _end(source.end())
     {
-    protected:
-        TextTemplateParserBase(
-            SourceView* inputSourceView,
-            DiagnosticSink* sink,
-            UnownedStringSlice source)
-            : _inputSourceView(inputSourceView)
-            , _sink(sink)
-            , _cursor(source.begin())
-            , _end(source.end())
-        {}
+    }
 
-        SourceView* _inputSourceView = nullptr;
-        DiagnosticSink* _sink = nullptr;
-        char const* _cursor = nullptr;
-        char const* _end = nullptr;
+    SourceView* _inputSourceView = nullptr;
+    DiagnosticSink* _sink = nullptr;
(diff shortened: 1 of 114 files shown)


tangent-vector:
Oh my. The new tool is crashing during the build on several of our target platforms. This will be fun to debug.

tools/slang-fiddle/slang-fiddle-template.cpp
```diff
@@ -0,0 +1,530 @@
...
+                if (wasAtStartOfLine && !depthInSplice)
+                {
+                    addRaw(currentSpanBegin, currentLineBegin);
+                    isInScriptLine = true;
+                    currentSpanBegin = _cursor;
+                }
+                break;
+
+            case '$':
+                if (isInScriptLine)
```
csyonghe, line 179: We need a way to escape, e.g. $$ means just a $.

tools/slang-fiddle/slang-fiddle-template.cpp
```diff
@@ -0,0 +1,530 @@
...
+                    continue;
+                if (depthInSplice)
+                    throw 99;
```
csyonghe, line 182: output a diagnostics instead of throwing?

csyonghe (commented):
We should use this to generate the .meta.slang files as well.

tools/slang-fiddle/slang-fiddle-template.cpp
tangent-vector: Yes, I was lazy at several points while developing this and did a quick `throw` instead of diagnosing properly.
We may still want to fail-fast on errors rather than try to soldier on, just given that the tool itself isn't 100% robust against bad input.

csyonghe: you can do a fatal error style handling, just write to diagnostic and throw, and make sure we handle the exception in main and print out the diagnostics.

Issue #6433, opened 2025-02-24
`countbits()` does not return correct results for non-32 bit integer types.


natevm:
Related to #5858 , we should add 16bit and 64bit tests for the `countbits` intrinsic. At the moment, this produces incorrect results in slang. 

```
uint64_t x = uint64_t(7) << 32;
print("count: ", countbits(x));
print("count: ", countbits(uint(x & 0xFFFFFFFF)) + countbits(uint(x >> 32) & 0xFFFFFFFF));
```

The above prints `count: 0` which is incorrect. 
The expected result is `count: 3`. 

Vulkan should soon support 64-bit integers for `countbits`.
CUDA also directly supports 64-bit integers in their pipc intrinsic.
HLSL will likely need us to add some sort of emulation, like `countbits(uint(x & 0xFFFFFFFF)) + countbits(uint(x >> 32) & 0xFFFFFFFF))`

sricker-nvidia:
Able to repro, investigating.

sricker-nvidia:
Posting an update, I was able to successfully test some edits that resolve the issue for cuda and cpp. That is:
`uint64_t x = uint64_t(7) << 32;
print("count: ", countbits(x));`
will now print 3 as expected. Working on emulation for backends that do not support 64bit countbits natively and starting on 16bit and 64bit tests.

sricker-nvidia:
Created an initial PR here: https://github.com/shader-slang/slang/pull/6845.

I had trouble tracking down info regarding Vulkan supporting 64-bit integers for countbits in the future. https://github.com/shader-slang/slang/issues/5858 refers to a new Q2 2025 spec, but I'm having trouble finding details for this upcoming spec and whether or not this will include 64-bit countbits. Interestingly, the other Spir-v integer intrinsics (min, max, abs) already support 64-bit. It seems like instructions from the SPIR-V Extended Instructions for GLSL (https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html) support 64-bit by default (or at least some of them do).

Trying to run the new countbits64 test from the PR above with Vulkan results in a VUID error:
https://docs.vulkan.org/spec/latest/appendices/spirvenv.html#VUID-StandaloneSpirv-Base-04781
"VUID-StandaloneSpirv-Base-04781
[...]

PR #6845, opened 2025-04-17
Implement 64bit countbits intrinsic for select targets (#6433)

sricker-nvidia:
Change modifies the countbits intrinsic to use generics in order to support 64bit countbits on select platforms where this is supported.

This should align with the implementation of other uint64_t intrinsics such as abs, min, max and clamp.

Added new countbits64 test to verify changes.

Updated documentation for 64bit-type-support.html

### DIFF
--- a/docs/64bit-type-support.md
+++ b/docs/64bit-type-support.md
@@ -147,7 +147,7 @@ D3D12    | FXC/DXBC         |      No        |          No        |   2
 2) uint64_t support requires https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12, so DXBC is not a target.
 
 The intrinsics available on `uint64_t` type are `abs`, `min`, `max`, `clamp` and `countbits`.
-The intrinsics available on `uint64_t` type are `abs`, `min`, `max` and `clamp`.
+The intrinsics available on `int64_t` type are `abs`, `min`, `max`, `clamp` and `countbits`.
 
 GLSL
 ====
--- a/prelude/slang-cpp-scalar-intrinsics.h
+++ b/prelude/slang-cpp-scalar-intrinsics.h
@@ -661,6 +661,23 @@ SLANG_FORCE_INLINE double I32_asdouble(int32_t low, int32_t hi)
     return u.d;
 }
 
+SLANG_FORCE_INLINE uint32_t I32_countbits(int32_t v)
+{
+#if SLANG_GCC_FAMILY && !defined(SLANG_LLVM)
+    return __builtin_popcount(uint32_t(v));
+#elif SLANG_PROCESSOR_X86_64 && SLANG_VC
+    return __popcnt(uint32_t(v));
+#else
+    uint32_t c = 0;
+    while (v)
+    {
+        c++;
+        v &= v - 1;
+    }
+    return c;
+#endif
+}
+
 // ----------------------------- U32 -----------------------------------------
 
 SLANG_FORCE_INLINE uint32_t U32_abs(uint32_t f)
@@ -729,9 +746,6 @@ SLANG_FORCE_INLINE uint64_t U64_max(uint64_t a, uint64_t b)
     return a > b ? a : b;
 }
 
-// TODO(JS): We don't define countbits for 64bit in the core module currently.
-// It's not clear from documentation if it should return 32 or 64 bits, if it exists.
-// 32 bits can always hold the result, and will be implicitly promoted.
 SLANG_FORCE_INLINE uint32_t U64_countbits(uint64_t v)
 {
 #if SLANG_GCC_FAMILY && !defined(SLANG_LLVM)
@@ -763,6 +777,11 @@ SLANG_FORCE_INLINE int64_t I64_min(int64_t a, int64_t b)
 SLANG_FORCE_INLINE int64_t I64_max(int64_t a, int64_t b)
 {
     return a > b ? a : b;
+}
+
+SLANG_FORCE_INLINE uint32_t I64_countbits(int64_t v)
+{
+    return U64_countbits(uint64_t(v));
 }
 
 // ----------------------------- UPTR -----------------------------------------
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -8028,7 +8028,8 @@ vector<T,N> cospi(vector<T,N> x)
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, shader5_sm_5_0)]
-uint countbits(uint value)
+__generic<T : __BuiltinIntegerType>
+uint countbits(T value)
 {
     __target_switch
     {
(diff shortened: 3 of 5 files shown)


sricker-nvidia:
Noting that I attempted to emulate 64-bit countbits for spir-v using something like:
`     case spirv:
-        return spirv_asm {OpBitCount $$uint result $value};
+        return spirv_asm {
+            %u64 = OpTypeInt 64 0;
+            %u32 = OpTypeInt 32 0;
+            %u64_32 = OpConstant %u64 32;
+            %high = OpShiftRightLogical $value %u64_32;
+            %low32 = OpUConvert %u32 $value;
+            %high32 = OpUConvert %u32 %high;
+            %lowCount = OpBitCount %low32;
+            %highCount = OpBitCount %high32;
+            result : $$uint = OpIAdd %lowCount %highCount
+        };`
But got complaints that OpConstant was being provided 5 words while expecting 4, presumably because it was using a 64-bit type. However, I don't understand the failure as this should be legal so long as "OpCapability Int64" is also given. I also see spirv-tools tests using similar spir-v assembly, so I'm not sure why it doesn't work.  The constant itself is needed as OpShiftRightLogical expects scalars of the same width (and value is 64-bit).

That aside, I'm also wondering if emulation like the above would have a negative impact on perf for the 32-bit case as now more instructions will be used to accomplish the same.

csyonghe:
you can use 
```
...
case spirv:
    if(T is int64_t || T  is uint64_t) return spirv_asm { // emulate};
    else return spirv_asm { // direct }
```

sricker-nvidia:
Thanks Yong, I wasn't sure if I could mix conditionals with intrinsics. I can give that a try if I figure out the issue with OpConstant.

tests/hlsl-intrinsic/countbits64.slang
```diff
@@ -0,0 +1,46 @@
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHK):-slang -compute -cpu
+// Spirv lacks support for 64bit countbits, see https://docs.vulkan.org/spec/latest/appendices/spirvenv.html#VUID-StandaloneSpirv-Base-04781
```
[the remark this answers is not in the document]
sricker-nvidia: Would this be something like:

%u32 = OpTypeInt 32 0;
%uint2 = OpTypeVector %u32 2;
%u64asuint2 = OpBitcast %uint2 $value
%bitcounts = OpBitCount  %u64asuint2
result : $$uint = OpIAdd ...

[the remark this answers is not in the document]
csyonghe: You can write the bit cast as slang code before the spirv_asm.

juliusikkala:
I wonder if the mac test failure here is the same kind as in #6786. There, the test involving a `vector<uintptr_t, 2>` fails only on Mac. Maybe there's something about 64-bit vectors that breaks on Mac OS builds?

prelude/slang-cpp-scalar-intrinsics.h
```diff
@@ -765,6 +779,23 @@ SLANG_FORCE_INLINE int64_t I64_max(int64_t a, int64_t b)
     return a > b ? a : b;
 }
 
+SLANG_FORCE_INLINE uint32_t I64_countbits(int64_t v)
```
jkwak-work, line 782: Can this be just like this without copy&paste?
```
SLANG_FORCE_INLINE uint32_t I64_countbits(int64_t v)
{
     return U64_countbits(uint64_t(v));
}
```

prelude/slang-cpp-scalar-intrinsics.h
```diff
@@ -661,6 +661,23 @@ SLANG_FORCE_INLINE double I32_asdouble(int32_t low, int32_t hi)
     return u.d;
 }
 
+SLANG_FORCE_INLINE uint32_t I32_countbits(int32_t v)
```
jkwak-work, line 664: Don't we need uint32_t variant as well?

docs/64bit-type-support.md
```diff
@@ -146,8 +146,10 @@ D3D12    | FXC/DXBC         |      No        |          No        |   2
 
 2) uint64_t support requires https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12, so DXBC is not a target.
 
+3) There are additional SPIR-V requirements applying to Vulkan shaders. uint64_t support for countbits is not available as the base operand of OpBitCount must be a 32-bit integer scalar or a vector of 32-bit integers (https://docs.vulkan.org/spec/latest/appendices/spirvenv.html#VUID-StandaloneSpirv-Base-04781)
```
jkwak-work, line 149: I don't think this statement is necessary for emulating things that SPIRV natively doesn't support.
We have so many places doing emulations like that.

source/slang/hlsl.meta.slang
```diff
@@ -7960,7 +7960,8 @@ vector<T,N> cospi(vector<T,N> x)
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, shader5_sm_5_0)]
-uint countbits(uint value)
+__generic<T : __BuiltinIntegerType>
+uint countbits(T value)
 {
     __target_switch
```
jkwak-work, line 7966: As Yong pointed out you should use `if ( T is XX)` thing that doesn't have a performance impact at the runtime.
You can also use `static_assert(T is XX)` to explicitly prevent the users from using a certain types by making them an error.

sricker-nvidia:
Interesting, thanks for sharing Julius. I'm taking a look now.

I've also just noticed that some similar 64-bit intrinsic testing in scalar-uint64.slang is devoid of a metal subtest, so perhaps there's some history we're not aware of?

The metal spec that I'm looking at (https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) doesn't explicitly state 64-bit support for things like min, abs, countbits, etc, but it feels implied as there is a row in Table 6.2. _Integer functions in the Metal standard library_:

`T popcount(T x) Returns the number of nonzero bits in x.`

in section 6.3 _Integer Functions_ and this section also has a paragraph stating:

`T is one of the scalar or vector integer types. Tu is the corresponding unsigned scalar or vector integer type. T32 is one of the scalar or vector 32-bit int or uint types.`

The fact that the table did not list T32 popcount(T32 x) ... to me implies 64-bit support, but that doesn't seem to be the case in practice. I see another line in the spec that mentions unsigned long uint64_t support being available in Metal 2.2 and later. Maybe a higher version needs to be requested?

prelude/slang-cpp-scalar-intrinsics.h
sricker-nvidia: Sure, I'll update.

docs/64bit-type-support.md
sricker-nvidia: I figured it would be good to have since the current version doesn't have spir-v emulation working. I'll remove this line if I'm able to get it functional based on some of the above suggestions.

source/slang/hlsl.meta.slang
sricker-nvidia: Thanks, that's good to know. My limited understanding of intrinsics made me think that I needed to avoid most of the normal slang semantics when implementing functions like countbits.

prelude/slang-cpp-scalar-intrinsics.h
sricker-nvidia: I made this update for I64_countbits, but not for I32_countbits due to the layout of the header file. In the header file U64_countbits is defined/declared above I64_countbits, but I32_countbits is not defined above/before U32_countbits. The header lumps all of these integer functions together in sections in this format:

--- I32 ---
...
--- U32 ---
...
--- U64 ---
...
--- I64 ---

If it makes sense to move the U32 section above the I32 section then I can fix I32_countbits as well.

jkwak-work (approved):
Looks good to me.

docs/64bit-type-support.md
```diff
@@ -147,7 +147,7 @@ D3D12    | FXC/DXBC         |      No        |          No        |   2
 2) uint64_t support requires https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12, so DXBC is not a target.
 
 The intrinsics available on `uint64_t` type are `abs`, `min`, `max`, `clamp` and `countbits`.
-The intrinsics available on `uint64_t` type are `abs`, `min`, `max` and `clamp`.
+The intrinsics available on `int64_t` type are `abs`, `min`, `max, clamp` and `countbits`.
```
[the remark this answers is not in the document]
sricker-nvidia: ah, good catch, I'll fix

jkwak-work (approved):
Looks good to me

PR #6851, opened 2025-04-17
Add a script to reproduce issues with vk-gl-cts

jkwak-work:
This commit adds a simple script that can help reproducing an issue with vk-gl-cts. The script will download necessary files and print out command-line arguments that execute vk-gl-cts tests.

Alternatively, we could write a document of how to manually run the vk-gl-cts test on the runner machine. But the script is simple enough to work as a documentation as well.


jkwak-work:
I am not happy with the limitations of the script.
Dropping the change.

PR #6852, opened 2025-04-17
Document matrix multiplication operators

jkwak-work:
This is to address one of cases where a user was confused about why Slang doesn't have operator overloading of multiplication for matrices.

It is little confusing that glsl.meta.slang implemented `operator*()` for matrices but Slang/HLSL doesn't. It turned out that HLSL/Slang performs a component-wise multiplication with `operator*()` whereas GLSL performs a matrix multiplication.


sricker-nvidia:
Looks ok to me besides the above suggestion.

sricker-nvidia (approved):
LGTM

fknfilewalker:
Could we get an option to deactivate the elementwise default operators? Not a big fan of them and you basically need them never.

PR #6857, opened 2025-04-18
Fix compiler warning on Windows about -build-id

jkwak-work:
This PR fixes the compiler warning like following, lld-link: warning: ignoring unknown argument '--no-undefined' lld-link: warning: ignoring unknown argument '--build-id', did you mean '-build-id'

chatGPT said that those options are for ELF and not for Windows:

These warnings happen because the Clang toolchain on Windows is invoking lld-link, which is the LLVM linker frontend for link.exe compatibility (i.e., it acts like the MSVC linker). The arguments --no-undefined and --build-id are ELF-specific flags, which are valid when targeting Linux, but not valid in the PE/COFF (Windows) environment.

Here's a breakdown:

⚠️ Warning messages explained:
lld-link: warning: ignoring unknown argument '--no-undefined' → This is a Linux/ELF flag. On Windows, this has no effect and is unknown.

lld-link: warning: ignoring unknown argument '--build-id', did you mean '-build-id' → Again, --build-id is for ELF binaries to include a unique build ID. Not valid for Windows targets.

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -152,8 +152,8 @@ function(set_default_compile_options target)
 
     add_supported_cxx_flags(${target} PRIVATE ${warning_flags})
 
-    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
-        # valid linker options only for GNU/Clang
+    if(NOT WIN32)
+        # these options are for ELF specific and not for Windows
         add_supported_cxx_linker_flags(
             ${target}
             PRIVATE


sricker-nvidia:
Are there equivalent args that we should be passing in for Windows?

jkwak-work:
> Are there equivalent args that we should be passing in for Windows?

It seems like there is no equivalent option for Windows.

The option for ELF binaries does the following,
```
--no-undefined: This linker flag instructs the linker to ensure that all symbols (functions, variables, etc.) used in the code are resolved during the linking process. If any undefined symbols are found, the linker will produce an error and fail.
```

sricker-nvidia (approved):
LGTM

PR #6861, opened 2025-04-18
Make peephole optimization not replace param with itself

### LINKED ISSUE #6860: Compilation hangs on unusual control flow with uninitialized variable
I'm using Slang 2025.6.4.

```
func breaker()->float {
    var x: float;
    for (int i = 0; i < 1; ++i) {
        if (true) {
        } else {
            x = 0.0;
        }
    }
    return x;
}


[shader("fragment")]
    : SV_Target
{
    let res = breaker();
    return float4(0, 0, 0, 0);
}

```

This code results in compilation hanging (tested for SpirV and WGSL targets). If `x` is assigned an initial value in breaker(), it compiles.

This is a locally minimal example. The body of the for loop cannot be extracted, the branch cannot be removed, nor can the body of the else be made empty. The bug only occurs when the assignment to x is in the else branch, if branch works fine.

I found this while chasing a different, invalid WGSL generation bug that creates `var<...>` annotations for local variables inside function bodies. Maybe the two are related.

juliusikkala:
Fixes #6860. The peephole optimization can get stuck in an infinite loop if changes are made on every iteration in `processFunc`. In this case, an IRParam was getting constantly replaced with itself, which marked a change on each iteration when nothing changed in reality.

Although, it's kind of interesting that this can happen in the first place. I'll probably dig around for a while in case there's something deeper happening here.

### DIFF
--- a/source/slang/slang-ir-peephole.cpp
+++ b/source/slang/slang-ir-peephole.cpp
@@ -1034,7 +1034,7 @@ struct PeepholeContext : InstPassBase
                         break;
                     }
                 }
-                if (argValue)
+                if (argValue && argValue != inst)
                 {
                     if (inst->hasUses())
                     {
--- a/source/slang/slang-ir-peephole.cpp
+++ b/source/slang/slang-ir-peephole.cpp
@@ -1022,7 +1022,10 @@ struct PeepholeContext : InstPassBase
                     SLANG_ASSERT(terminator->getArgCount() > paramIndex);
                     auto arg = terminator->getArg(paramIndex);
                     if (arg->getOp() == kIROp_undefined)
-                        continue;
+                    {
+                        argValue = nullptr;
+                        break;
+                    }
                     if (argValue == nullptr)
                         argValue = arg;
                     else if (argValue == arg)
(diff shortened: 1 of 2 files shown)


csyonghe:
Thanks for taking this issue! Can we add a test for the fix?

juliusikkala:
Sure, I'll add one soon too. I figured out what's actually happening here. For some reason, `x` gets promoted into a parameter for the loop, and then gets replaced with itself in the `true` branch. So it seems that there's probably a bigger problem going on here.

csyonghe:
Yeah, given that `x` is not read by anything in the loop, it doesn't need to be promoted as an inductive phi node in the loop. The check you added makes sense though. We may also want to consider being more conservative and don't try to replace things if one of the phi argument is an `undefined`.

juliusikkala:
Alright, I made the 'kIROp_undefined' check more conservative. I also tested that it alone would be enough to fix this bug as well, so now it's double-fixed :-)

juliusikkala:
Looks like the more conservative `undefined` check breaks some tests, so I'll revert for now.

juliusikkala:
Closing in favor of #6870, the situation that this PR fixes shouldn't be possible anymore after that PR.

PR #6863, opened 2025-04-19
Fixed various queryInterface implementations

AlexisPollonni:
Hello all,

I am doing something similar as https://github.com/shader-slang/slang/pull/5978 where i am writing COM wrappers for slang. All went good for the core slang lib and reflections APIs thus far, but i hit a few issues with the gfx wrapper.
First one is in the IDevice implementation, in the case another uuid was queried than IDevice it would return null with the OK code. This would crash when the .Net runtime tries to query the interface.

https://github.com/shader-slang/slang/blob/043278a527ab5744674417a08d924c67a60a486b/tools/gfx/renderer-shared.cpp#L398-L410

I hit another snag in the debug transient heap, weirdly the inner object was returned instead of the debug wrapper version. I supposed this was unintended looking at the code. It would cause an access violation when subsequently using the compute encoder to bind a pipeline state. Let me know if those changes are ok or if i need to adjust anything, its my first time digging and debugging this codebase!
So far i ran the shader-object, raytracing and model viewer examples without error with the changes.
I am going to continue going through the interfaces and see if i hit other issues that i can do something about.

### DIFF
--- a/tools/gfx/debug-layer/debug-transient-heap.cpp
+++ b/tools/gfx/debug-layer/debug-transient-heap.cpp
@@ -43,7 +43,7 @@ Result DebugTransientResourceHeap::finish()
 }
 
 Result DebugTransientResourceHeap::createCommandBuffer(ICommandBuffer** outCommandBuffer)
-{ 
+{
     SLANG_GFX_API_FUNC;
     RefPtr<DebugCommandBuffer> outObject = new DebugCommandBuffer();
     outObject->m_transientHeap = this;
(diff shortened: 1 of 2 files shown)


tools/gfx/debug-layer/debug-transient-heap.cpp
```diff
@@ -41,7 +43,7 @@ Result DebugTransientResourceHeap::finish()
 }
 
 Result DebugTransientResourceHeap::createCommandBuffer(ICommandBuffer** outCommandBuffer)
-{
+{
```
jkwak-work, line 46: This seems unnecessary white space.

jkwak-work (approved):
Looks good to me.

tools/gfx/debug-layer/debug-transient-heap.cpp
AlexisPollonni: woops sorry! Seems visual studio inserted some sneaky space i missed, thanks for noticing should be all good now

PR #6865, opened 2025-04-19
Stop using `chmod` on Windows host to fix build targeting Android

James2022-rgb:
Hi, this PR fixes building for Android on Windows host, by making the usage of the UNIX `chmod` command conditional on `CMAKE_HOST_WIN32`.


jkwak-work:
Can you give us more explanation of why this is needed when it is not needed on x64? I don't think any of committees can reproduce the problem locally so we need more information

James2022-rgb:
Hi, without the fix, the build fails because the `chmod` command doesn't exist on Windows.
(Also Windows NTFS doesn't use Unix-style permission bits)

It results in something like the following, failing the build:
```
chmod not recognized as internal or external command
```

(I'm on Japanese Windows so not sure this is the exact wording on an English Windows)

cmake/SlangTarget.cmake
```diff
@@ -277,7 +277,22 @@ function(slang_add_target dir type)
...
-                    COMMAND chmod 644 $<TARGET_FILE:${target}>.dwarf
+                    WORKING_DIRECTORY ${output_dir}
+                    VERBATIM
+                )
+                # We may be building for Android on a Windows host, where chmod isn't available or needed.
+                if(NOT CMAKE_HOST_WIN32)
+                    add_custom_command(
+                        TARGET ${target}
+                        POST_BUILD
+                        COMMAND chmod 644 $<TARGET_FILE:${target}>.dwarf
```
jkwak-work, line 288: Copilot suggests to use the following,
```
COMMAND ${CMAKE_COMMAND} -E file CHMOD 644 $<TARGET_FILE:${target}>.dwarf
```
That is a platform agnostic way with cmake.
You can remove `if(NOT CMAKE_HOST_WIN32)` when using this command, I assume.

Can you give a try on Windows Android?

James2022-rgb: Undoing the commit and replacing the `chmod 644` line with `COMMAND ${CMAKE_COMMAND} -E file CHMOD 644 $<TARGET_FILE:${target}>.dwarf`,
I get:

```
[110/353] Linking CXX shared module Debug\lib\libslang-glslang.so
  FAILED: Debug/lib/libslang-glslang.so 
  cmd.exe /C "cd . && <Long chain of commands> && <Path to Android SDK>\cmake\3.22.1\bin\cmake.exe -E file CHMOD 644 <Path to Android project>/app/.cxx/Debug/4j3cu1c3/arm64-v8a/Debug/lib/libslang-glslang.so.dwarf && <Long chain of commands>"
  CMake Error: cmake version 3.22.1-g37088a8-dirty
  Usage: <Path to Android SDK>\cmake\3.22.1\bin\cmake.exe -E <command> [arguments...]
  Available commands: 
    capabilities              - Report capabilities built into cmake in JSON format
    cat <files>...            - concat the files and print them to the standard output
    chdir dir cmd [args...]   - run command in a given directory
    compare_files [--ignore-eol] file1 file2
                                - check if file1 is same as file2
    copy <file>... destination  - copy files to destination (either file or directory)
    copy_directory <dir>... destination   - copy content of <dir>... directories to 'destination' directory
    copy_if_different <file>... destination  - copy files if it has changed
    echo [<string>...]        - displays arguments as text
    echo_append [<string>...] - displays arguments as text but no new line
    env [--unset=NAME]... [NAME=VALUE]... COMMAND [ARG]...
                              - run command in a modified environment
    environment               - display the current environment
    make_directory <dir>...   - create parent and <dir> directories
    md5sum <file>...          - create MD5 checksum of files
    sha1sum <file>...         - create SHA1 checksum of files
    sha224sum <file>...       - create SHA224 checksum of files
    sha256sum <file>...       - create SHA256 checksum of files
    sha384sum <file>...       - create SHA384 checksum of files
    sha512sum <file>...       - create SHA512 checksum of files
    remove [-f] <file>...     - remove the file(s), use -f to force it (deprecated: use rm instead)
    remove_directory <dir>... - remove directories and their contents (deprecated: use rm instead)
    rename oldname newname    - rename a file or directory (on one volume)
    rm [-rRf] <file/dir>...    - remove files or directories, use -f to force it, r or R to remove directories and their contents recursively
    sleep <number>...         - sleep for given number of seconds
    tar [cxt][vf][zjJ] file.tar [file/dir1 file/dir2 ...]
                              - create or extract a tar or zip archive
    time command [args...]    - run command and display elapsed time
    touch <file>...           - touch a <file>.
    touch_nocreate <file>...  - touch a <file> but do not create it.
    create_symlink old new    - create a symbolic link new -> old
    create_hardlink old new   - create a hard link new -> old
    true                      - do nothing with an exit code of 0
    false                     - do nothing with an exit code of 1
  Available on Windows only:
    delete_regv key           - delete registry value
    env_vs8_wince sdkname     - displays a batch file which sets the environment for the provided Windows CE SDK installed in VS2005
    env_vs9_wince sdkname     - displays a batch file which sets the environment for the provided Windows CE SDK installed in VS2008
    write_regv key value      - write registry value
```

jkwak-work: In that case, I guess it is good as it was.

jkwak-work (approved):
Looks good to me.

PR #6866, opened 2025-04-19
Avoid dead code causing broken & invalid Phi nodes

### LINKED ISSUE #6862: Unusual control flow with unreachable code generates invalid WGSL
Using Slang 2025.6.4.

```
func dummy(b: StructuredBuffer<float>)->float {
    return 0;
}

func breaker(b: StructuredBuffer<float>)->float {
    var x: float = 0;
    for (int i = 0; i < 1; ++i) {
        x = dummy(b);
        if (true) {
        } else {
            return 0;
            x = 0;
        }
    }
    return x;
}

StructuredBuffer<float> b;


[shader("fragment")]
    : SV_Target
{
    let res = breaker(b);
    return float4(res, 0, 0, 0);
}
```
This code results in the following WGSL:
```
@binding(0) @group(0) var<storage, read> b_0 : array<f32>;

fn dummy_0( b_1 : array<f32>) -> f32
{
    return 0.0f;
}

struct pixelOutput_0
{
    @location(0) output_0 : vec4<f32>,
};

fn breaker_0() -> f32
{
    var<storage, read> _S1 : array<f32> = b_0;
    var x_0 : f32 = 0.0f;
    var i_0 : i32 = i32(0);
    for(;;)
    {
        if(i_0 < i32(1))
        {
        }
        else
        {
            break;
        }
        var _S2 : f32 = dummy_0(_S1);
        var i_1 : i32 = i_0 + i32(1);
        x_0 = _S2;
        i_0 = i_1;
    }
    return x_0;
}

@fragment
fn fragment(@builtin(position) in_0 : vec4<f32>) -> pixelOutput_0
{
    var _S3 : pixelOutput_0 = pixelOutput_0( vec4<f32>(breaker_0(), 0.0f, 0.0f, 0.0f) );
    return _S3;
}
```
Note the `var<storage, read>` annotation on the first body line of `breaker_0` - this is invalid, `var<storage, read>` declarations are module-scope only.
Swapping the order of the two lines in the `else` branch of `breaker` results in Slang no longer warning about the assignment `x=0` being unreachable, and the code compiles to valid WGSL.

I suspect that this may have something to do with issue https://github.com/shader-slang/slang/issues/6860, due to the very similar control flow used to find the bug. @juliusikkala thoughts?

juliusikkala:
Fixes #6862.

```slang
func breaker(b: StructuredBuffer<float>)->float {
    var x: float = 0;
    for (int i = 0; i < 1; ++i) {
        x = dummy(b);
        if (true) {
        } else {
            return 0;
            x = 0;
        }
    }
    return x;
}
```

This one was a real doozy to debug. What's happening here, is that a phi node is inserted for `b` in the loop, because the `after`/reconvergence block of the if-else is preceded by both the empty true-branch and the `x = 0;`. There's no issue in recursively finding a value for `b` in the true-branch, because it has predecessors. However, because the `x = 0;` block is dead code, it has no predecessors and thus yields an undefined value for `b` in `readVarRec`. This means that SSA doesn't "know" that the value of 'b' never actually changes, because figuring out its value hits a false dead-end at the dead code.

So the phi for `b` has to select between the known defined value from the true-branch and an undefined value from the false-branch, and cannot be removed. This causes the creation of an unnecessary variable for `b`, which apparently isn't valid in WGSL.

### DIFF
--- a/source/slang/slang-ir-ssa.cpp
+++ b/source/slang/slang-ir-ssa.cpp
@@ -435,8 +435,8 @@ IRInst* tryRemoveTrivialPhi(ConstructSSAContext* context, PhiInfo* phiInfo)
 
         auto parentBlock = as<IRBlock>(usedVal->parent);
         bool deadCode = usedVal->getOp() == kIROp_undefined && parentBlock &&
-            parentBlock->getPredecessors().getCount() == 0 &&
-            parentBlock->getPrevBlock() != nullptr;
+                        parentBlock->getPredecessors().getCount() == 0 &&
+                        parentBlock->getPrevBlock() != nullptr;
 
         if (usedVal == same || usedVal == phi || deadCode)
         {
--- a/source/slang/slang-ir-ssa.cpp
+++ b/source/slang/slang-ir-ssa.cpp
@@ -428,27 +428,10 @@ IRInst* tryRemoveTrivialPhi(ConstructSSAContext* context, PhiInfo* phiInfo)
     // to the phi itself.
 
     IRInst* same = nullptr;
-    IRInst* undefinedVal = nullptr;
-    bool multipleUndefined = false;
     for (auto u : phiInfo->operands)
     {
         auto usedVal = u.get();
         SLANG_ASSERT(usedVal);
-
-        if (usedVal->getOp() == kIROp_undefined)
-        {
-            // Since undefined values are undefined, we can
-            // ignore them in favor of a defined value in
-            // the phi. We still need to keep it around
-            // though, in case we don't find any defined
-            // values and have to use it instead.
-            if (undefinedVal != usedVal && undefinedVal != nullptr)
-            {
-                multipleUndefined = true;
-            }
-            undefinedVal = usedVal;
-            continue;
-        }
 
         if (usedVal == same || usedVal == phi)
         {
@@ -456,7 +439,6 @@ IRInst* tryRemoveTrivialPhi(ConstructSSAContext* context, PhiInfo* phiInfo)
             // to the same value we've seen already.
             continue;
         }
-
         if (same != nullptr)
         {
             // We've found at least two distinct values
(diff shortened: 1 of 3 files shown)


juliusikkala:
Notes: if the DCE approach is not suitable, another option is to somehow mark the no-predecessor case `readVarRec` differently than just "undefined". A different instruction may have *unforeseen consequences*, so perhaps a decoration could be used to mark the fact that it's undefined because it's dead code or an entry block.

Then, based on that decoration, `tryRemoveTrivialPhi` can skip values stemming from dead code.

juliusikkala:
The current approach here was the only one that didn't cause failures in existing tests and also fixes the issue and the new test. It makes the `tryRemoveTrivialPhi()` skip undefined values that stem from blocks that aren't the entry block and have no predecessors (= dead code). Such blocks are unreachable, so they aren't a "real" option for the Phi and can be ignored. In the end, no additional decorations were necessary with this approach.

csyonghe (commented):
I wonder if the right fix should be to run dead code elim pass to cleanup the cfg before doing ssa. 

Theoretically, the usedVal could be defined in a block that is transitively unreachable(in a block that has predecessor but the predecessor block doesn’t have an predecessor), instead of directly unreachable as being tested here. The only correct way to test this would be to run a systematic cleanup pass first.

juliusikkala:
I also think just running dead code elimination before ssa is a cleaner way to fix this. I tried that in https://github.com/shader-slang/slang/pull/6866/commits/a0b0abdea999fdc9c8b5834dab6b52fb93a0fad1, but this caused one seemingly unrelated test to fail (`tests/autodiff/backward-diff-check.slang`), which is why I went back to this current form. I can figure out why the test breaks tomorrow.

csyonghe:
I think it isn’t technically an error to create the additional phi node if the cfg hasn’t been cleaned up before the ssa pass, so long as our optimization passes can remove the redundant phi later. I thought you have a fix that allows our optimization to clean that up?

juliusikkala:
I'm still very unfamiliar with this area of the compiler, so I'm going to have to do more investigation/studying on this later. It's seems odd to me that the phi didn't get removed in a later pass, despite the fact that the dead code leading to it did get removed.

By the fix, do you mean #6861? I did try it with this issue too, but it doesn't affect the output here. It only fixes the hang from #6860. Other than that, I don't really have anything else right now.

csyonghe:
Yeah, I thought they were fixing the same issue but I am obviously wrong…

It does seem to me that making sure we can correctly cleanup the phi after dead code pass might be the more robust solution here.

juliusikkala:
Closing in favor of #6870, which is a separate PR due to taking a completely different approach and also fixing the other issue as well.

PR #6870, opened 2025-04-21
Allow simplifying self-referential Phi parameters

### LINKED ISSUE #6862: Unusual control flow with unreachable code generates invalid WGSL

juliusikkala:
Fixes #6862 and #6860, and supersedes #6861 and #6866.

In both cases, unnecessary phis were sticking around, as they had two options: the value before the loop, and themselves. Since in this case the only possible value it could have is the value before the loop, the phi can be removed. This fixes both issues, as both are related to unnecessary self-referential phis messing up optimization or WGSL output.

### DIFF
--- a/source/slang/slang-ir-simplify-cfg.cpp
+++ b/source/slang/slang-ir-simplify-cfg.cpp
@@ -728,9 +728,9 @@ static bool simplifyBoolPhiParams(IRBlock* block)
 
 static bool removeTrivialPhiParams(IRBlock* block)
 {
-    // We can remove a phi parmeter if:
-    // 1. all arguments to a parameter is the same (not really a phi).
-    // 2. the arguments to the parameter is always the same as arguments to another existing
+    // We can remove a phi parameter if:
+    // 1. all non-self-referential arguments to a parameter are the same (not really a phi).
+    // 2. the arguments to the parameter are always the same as arguments to another existing
     // parameter (duplicate phi).
 
     bool changed = false;
@@ -765,7 +765,10 @@ static bool removeTrivialPhiParams(IRBlock* block)
         termInsts.add(termInst);
         for (UInt i = 0; i < termInst->getArgCount(); i++)
         {
-            if (args[i].areKnownValueSame)
+            // Self-referential parameters can be skipped, as they cannot
+            // introduce a new value. The phi can only have multiple different
+            // values if non-self-referential arguments differ.
+            if (args[i].areKnownValueSame && termInst->getArg(i) != params[i])
             {
                 if (args[i].knownValue == nullptr)
                     args[i].knownValue = termInst->getArg(i);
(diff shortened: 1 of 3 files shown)


source/slang/slang-ir-simplify-cfg.cpp
```diff
@@ -765,7 +765,10 @@ static bool removeTrivialPhiParams(IRBlock* block)
         termInsts.add(termInst);
         for (UInt i = 0; i < termInst->getArgCount(); i++)
         {
-            if (args[i].areKnownValueSame)
+            // Self-referential parameters can be skipped, as they cannot
+            // introduce a new value. The phi can only have multiple different
+            // values if non-self-referential arguments differ.
+            if (args[i].areKnownValueSame && termInst->getArg(i) != params[i])
```
csyonghe, line 771: If args[i].knownVal is nullptr, do we still want to assign the param to it as in line 774, even when the arg is self referential?

juliusikkala: AFAIK, there's no way to end up with a phi that only has self-referential arguments (that would implicate a block where all predecessors are back jumps), so `knownValue` will still always get some non-nullptr value in the end.

csyonghe: Ah, that's true.

csyonghe (approved):

PR #6879, opened 2025-04-22
Add a new SM profile 6.8 and 6.9

jkwak-work:
This is needed later for coopvec with DX12 backend.


jkwak-work:
Merging this because slang-rhi side [PR 325](https://github.com/shader-slang/slang-rhi/pull/325) is merged and it may start failing without this PR.

Issue #6815, opened 2025-04-15
SlangPy regression with Slang 2025.6.4


ccummingsNV:
When building SGL with Slang 2025.6.4, the SlangPy test "test_pytorch.py: test_tensor_arguments" and "test_pytorch.py: test_tensor_interfaces" fail (in addition to the regressions in 2025.6.2).

Both are as a result of a slang internal error in getEntryPointCode (screenshot attached).

![Image](https://github.com/user-attachments/assets/92d344fc-dc7c-47e7-a60e-ad7c036f225c)

gtong-nv:
Using Slang 2025.6.4, I'm able to reproduce some failures like `FAILED test_pytorch.py::test_tensor_interfaces[DeviceType.d3d12] - SystemError: nanobind::detail::nb_func_error_except(): exception could not be translated!`
```
self = Object(0x1e4d72e6b88)
args = (<slangpy.torchintegration.wrappedtensor.WrappedTensor object at 0x000001E508D31450>, <slangpy.torchintegration.wrappe...sor object at 0x000001E508D2AE40>, <slangpy.torchintegration.wrappedtensor.WrappedTensor object at 0x000001E508D2AB70>)
kwargs = {'_result': <slangpy.torchintegration.wrappedtensor.WrappedTensor object at 0x000001E508D55A90>}
resval = <slangpy.torchintegration.wrappedtensor.WrappedTensor object at 0x000001E508D55A90>

    def call(self, *args: Any, **kwargs: Any) -> Any:
        """
        Call the function with a given set of arguments. This will generate and compile
        a new kernel if need be, then immediately dispatch it and return any results.
        """
        # Handle result type override (e.g. for numpy) by checking
        # for override, and if found, deleting the _result arg and
        # calling the function with the override type.
        resval = kwargs.get('_result', None)
        if isinstance(resval, (type, str)):
            del kwargs['_result']
            return self.return_type(resval).call(*args, **kwargs)
        else:
            try:
>               return self._native_call(self.module.call_data_cache, *args, **kwargs)
               SystemError: nanobind::detail::nb_func_error_except(): exception could not be translated!

..\core\function.py:207: SystemError
```

But when running with VS debugger for slang-rhi, it doesn't raise exception at  `Device::getEntryPointCodeFromShaderCache`

gtong-nv:
By enable all exceptions in VS, I'm able to see the issue described at `Device::getEntryPointCodeFromShaderCache`

[...]

PR #6890, opened 2025-04-23
Auto-diff loop analysis: Handle case where either side of a conditional branch is empty

saipraveenb25:
Addresses the auto-diff crash for #6815


gtong-nv:
Confirmed this fixes the crash.

gtong-nv (approved):

csyonghe (approved):

PR #6897, opened 2025-04-24
Add countbits 16-bit and 8-bit support (#6433)

sricker-nvidia:
Change adds 16-bit and 8-bit support for countbits intrinsic. In cases where a backend's native counbits lacks support, support is emulated.

New tests are added for 16-bit and 8-bit support. Additional testing added for 32-bit and minor updates made to 64-bit countbits.


source/slang/hlsl.meta.slang
```diff
@@ -8018,6 +8018,16 @@ vector<T,N> cospi(vector<T,N> x)
     }
 }
 
+// emulate 64-bit countbits when not natively supported.
+[__readNone]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, shader5_sm_5_0)]
+uint emulatedCountbits64(uint64_t value)
```
csyonghe, line 8025: `internal uint emulatedCountbits64`.

source/slang/hlsl.meta.slang
```diff
@@ -8055,17 +8073,28 @@ uint countbits(T value)
...
-            // emulate 64-bit
-            uint2 value_uint2 = bit_cast<uint2>(value);
-            uint2 counted_bits_uint2 = countbits(value_uint2);
-            return counted_bits_uint2.x + counted_bits_uint2.y;
+            return emulatedCountbits64(reinterpret<uint64_t, T>(value));
+        }
+        else if (T is int16_t || T  is uint16_t)
+        {
+            // emulate 16-bit
+            return countbits(reinterpret<uint32_t, T>(value));
```
csyonghe, line 8081: `__intCast<uint32_t>(value)`

source/slang/hlsl.meta.slang
```diff
@@ -8094,15 +8130,24 @@ vector<uint, N> countbits(vector<T, N> value)
...
         else
         {
             return spirv_asm {OpBitCount $$vector<uint, N> result $value};
         }
     case wgsl:
-        __intrinsic_asm "countOneBits";
+        // wgsl only supports 32-bit integers
+        if (T is int32_t)
+        {
+            return countbits(reinterpret<vector<uint32_t, N>, vector<T, N>>(value));
```
csyonghe, line 8148: `__intCast<vector<uint32_t,N>(value)`

source/slang/hlsl.meta.slang
jkwak-work: I didn't know `internal` is a valid keyword; good to know.
To @sricker-nvidia , the naming convention for the internal functions that the Slang users aren't supposed to use to start with two underscore characters like `__emulatedCountbits64`.

source/slang/hlsl.meta.slang
```diff
@@ -8031,19 +8041,27 @@ vector<T,N> cospi(vector<T,N> x)
 __generic<T : __BuiltinIntegerType>
 uint countbits(T value)
 {
+    // Emulate 8-bit support
+    // 8-bit support is not currently supported anywhere natively
+    if (T is int8_t || T  is uint8_t)
+    {
+        return countbits(reinterpret<uint32_t, T>(value));
```
jkwak-work, line 8048: Shouldn't this be just
```
return countbits(uint32_t(value));
```
I thought reinterpret is same as bit-cast and the input and output types should be in a same size, isn't it?

source/slang/hlsl.meta.slang
```diff
@@ -8075,6 +8104,13 @@ __generic<T : __BuiltinIntegerType, let N : int>
 [require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, shader5_sm_5_0)]
 vector<uint, N> countbits(vector<T, N> value)
 {
+    // Emulate 8-bit support
+    // 8-bit support is not currently supported anywhere natively
+    if (T is int8_t || T  is uint8_t)
+    {
+        VECTOR_MAP_UNARY(uint, N, countbits, value);
```
jkwak-work, line 8111: I am wondering if we can do this little better like how we do 64bits with two of 32bits?
But I guess it will be tricky to implement.

source/slang/hlsl.meta.slang
csyonghe: you are right, reinterpret is a bit cast, and can’t change the size of the value.

source/slang/hlsl.meta.slang
sricker-nvidia: ack

source/slang/hlsl.meta.slang
sricker-nvidia: ack

source/slang/hlsl.meta.slang
sricker-nvidia: What did you have in mind? We can't take 4 uint8_t's and stick them into 1 uint32_t and run the 32bit countbits as we need to return a vector of the individual counts rather than a sum.

source/slang/hlsl.meta.slang
sricker-nvidia: Using the above gives a compile error:
`29>hlsl.meta.slang(8047): error 30019: expected an expression of type 'uint', got 'T'`
`29>        return countbits(uint32_t(value));`

source/slang/hlsl.meta.slang
sricker-nvidia: Using this is giving me a build error:
`27>hlsl.meta.slang(8143): error 39999: could not specialize generic for arguments of type (vector<T,N>)`
`27>            return countbits(__intCast<vector<uint32_t,N>>(value));`

sricker-nvidia:
Looks like 8-bit and 16-bit emulation does not work on macos for some reason according to the CI. Will likely disable those for now and investigate in a follow up.

source/slang/hlsl.meta.slang
jkwak-work: Something like that. But let's not worry about that now.

Issue #5782, opened 2024-12-06
Language server should display default argument values


skallweitNV:
When hovering over functions it would be nice to show default argument values instead of just a plain list of all arguments. This helps the user to see what arguments he **must** provide vs. optional arguments..

PR #6902, opened 2025-04-24
Output default parameter values on language server hover

expipiplus1:
Closes https://github.com/shader-slang/slang/issues/5782


source/slang/slang-ast-print.cpp
```diff
@@ -74,6 +74,1052 @@ void ASTPrinter::addType(Type* type)
...
+        sb << ")";
+    }
+    else if (const auto invokeExpr = as<InvokeExpr>(expr))
+    {
+        if (const auto operatorExpr = as<OperatorExpr>(invokeExpr))
+        {
+            if (const auto infixExpr = as<InfixExpr>(operatorExpr))
+            {
+                // Binary operator
+                if (invokeExpr->arguments.getCount() == 2)
```
csyonghe, line 302: This doesn't seem to handle insertion of () to properly express operator priorities. We could just emit () for all cases for now.

PR #6904, opened 2025-04-24
Modify markdown writing to produce readthedocs pages

aidanfnv:
For https://github.com/shader-slang/stdlib-reference/issues/3

This makes multiple changes to the slang doc markdown writer to make the stdlib reference pages usable with readthedocs.
- Adds tables of contents at the bottom of every page with children. TOCs contain category landing pages where applicable and contain uncategorized child pages
- Changes links on the pages to use relative paths instead of absolute paths
- Adds explicit .html extensions to links
- Changes anchor tags to use HTML syntax instead of markdown syntax

The TOC for the stdlib reference on readthedocs will look the same as the TOC seen on the existing docs website, and I have verified that links and anchors will still work.


source/slang/slang-doc-markdown-writer.cpp
```diff
@@ -100,6 +100,72 @@ String getDocPath(const DocumentationConfig& config, String path)
     return config.rootDir + Path::getPathWithoutExt(path);
 }
 
+String getRelativePath(String fromPath, String toPath, bool includeExt)
```
csyonghe, line 103: Can you just use `Path::getRelativePath` defined in `slang-io.h`?

source/slang/slang-doc-markdown-writer.cpp
```diff
@@ -100,6 +100,72 @@ String getDocPath(const DocumentationConfig& config, String path)
...
+    }
+
+    // Add the .html extension if requested
+    if (includeExt)
+        sb << ".html";
+
+    return sb.produceString();
+}
+
+String getToctreeEntry(const String& name, const String& fromPath, const String& toPath)
```
csyonghe, line 160: `getTocTreeEntry`

source/slang/slang-doc-markdown-writer.cpp
aidanfnv: Ah, I missed `Path::getRelativePath`. I could just use that, and where necessary I could use `Path::getPathWithoutExt` to strip the extension and then append `.html`.

source/slang/slang-doc-markdown-writer.cpp
aidanfnv: Fixed

source/slang/slang-doc-markdown-writer.cpp
aidanfnv: I've removed getRelativePath in favor of using Path::getRelativePath and Path::getPathWithoutExt.

jkwak-work (approved):
Looks good to me.

expipiplus1 (approved):

aidanfnv:
I have built a preview of the readthedocs site with the docs generated with these changes:
https://slang-documentation.readthedocs.io/en/latest/external/stdlib-reference/

jkwak-work (approved):
Looks good to me.

Issue #6892, opened 2025-04-23
Module serialization crash with for loops


juliusikkala:
The problem seems to occur whenever there's a for-loop. It appears that this started happening after #6854 (I tested this to work correctly with the commit right before that PR was merged).

This is the absolute minimal example I could find.
```slang
void f()
{
    for (int i = 0; i < 1; ++i)
        ;
}
```

Compile with `slangc`:
```
slangc test.slang -o test.slang-module
```

to observe a crash:

```
terminate called after throwing an instance of 'Slang::InternalError'
```

The crash occurs during encoding at https://github.com/shader-slang/slang/blob/785669c8771c01d57914ec0075315e8ae61b6a31/source/core/slang-riff.cpp#L855

tangent-vector:
@juliusikkala I'm going to start looking into this. Do you know whether the code also runs into problems in a release build, or is it just related to that assertion?

The `isChunkOk` thing is one of those validation checks that can often be more bothersome than helpful, so I have a strong suspicion that the fix will amount to just silencing it (but will try to confirm first). In my own feature branch where I'm continuing to build on the AST serialization work, I've actually refactored the RIFF handling subsystem so that it doesn't even need the `isChunkOk` check (it just automatically updates the side-band data if it has been invalidated).

tangent-vector:
Thank you for providing such a compact reproducer for this issue.

I am seeing a crash at a different location, which may mean that there is more than one issue here.

The crash I am seeing relates to the fact that the `LiteralExpr` class was marked as abstract for the purposes of the `slang-fiddle` code generation tool (it uses `FIDDLE(abstract)`), but there is code in `slang-check-stmt.cpp` that was creating a `LiteralExpr` as part of `tryInferLoopMaxIterations`. If that function is able to infer a maximum iteration count (which it can for the repro code), it will construct an `InferredMaxItersAttribute` with a `LiteralExpr` (rather than an `IntegerLiteralExpr`) for the iteration count.

The fix for that issue is as simple as changing the `create<LiteralExpr>` to `create<IntegerLiteralExpr>`, so I will now go on to see whether I can repro a crash after that fix.

tangent-vector:
PR #6905 contains a fix for the issue I described just above. With that fix I am unable to repro a crash using the shader code from the original bug report, but I am reluctant to close out the bug until the fix has been separately confirmed, given that I was unable to reproduce a crash at the location that was originally reported.

juliusikkala:
[...]

PR #6905, opened 2025-04-24
Fix attempt to construct an abstract AST node

tangent-vector:
Work on #6892

The function `tryInferLoopMaxIterations` in `slang-check-stmt.cpp` was doing:

    auto litExpr = m_astBuilder->create<LiteralExpr>();

but the declaration of `LiteralExpr` in `slang-ast-expr.h` had been marked as abstract, during the switch to using the `slang-fiddle` tool to generate code:

    FIDDLE(abstract)
    class LiteralExpr : public Expr
    { ... }

In this case, the intention of the AST design is that `LiteralExpr` should be kept abstract, and only the concrete subclasses should ever be instantiated.

Because of some historical design choices, the `ASTNodeType` enumeration includes both the concrete and abstract AST classes, so the code ended up constructing a `LiteralExpr` that had the tag `ASTNodeType::LiteralExpr`. Then attempts to use the `ASTNodeDispatcher` code on such a node caused crashes, because the dispatcher code only included `case` statements for the non-abstract `ASTNodeType`s.

The quick fix here was to change the `tryInferLoopMaxIterations` function to instead do:

    auto litExpr = m_astBuilder->create<IntegerLiteralExpr>();

A test case was added to help catch any future regressions on this specific issue.

A more long-term fix should involve introducing code that statically and/or dynamically prohibits the creation of instances of AST node classes that have been marked abstract.

### DIFF
--- a/source/slang/slang-check-stmt.cpp
+++ b/source/slang/slang-check-stmt.cpp
@@ -880,7 +880,7 @@ void SemanticsStmtVisitor::tryInferLoopMaxIterations(ForStmt* stmt)
     // if the loop body modifies the induction variable.
     //
     auto maxItersAttr = m_astBuilder->create<InferredMaxItersAttribute>();
-    auto litExpr = m_astBuilder->create<LiteralExpr>();
+    auto litExpr = m_astBuilder->create<IntegerLiteralExpr>();
     litExpr->type.type = m_astBuilder->getIntType();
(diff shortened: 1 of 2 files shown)


juliusikkala (approved):
I tested this on my PC, and it has fixed the issue. Thanks!

PR #6910, opened 2025-04-25
reduce server count to 4 for testing

aasgaonkar:


jkwak-work:
I think we should close this PR in a favor of the retry logic,
 - https://github.com/shader-slang/slang/pull/6912

PR #6914, opened 2025-04-26
Initial support for immutable lambda expressions.

csyonghe:
This PR adds initial support for lambda expressions.

The implementation in this PR does not allow lambdas that mutates the captured variables, nor does it allow lambdas to have `out` or `ref` parameters. Two syntaxes are supported:
1. Lambda body that is a simple expression: `(float x, float y) => x + y`.
2. Lambda body that is a block of statements: `(float x, float y) => { return x+y; }`

Example test case enabled by this PR:
```

//TEST:COMPARE_COMPUTE(filecheck-buffer=CHECK):-vk -output-using-type
//TEST:COMPARE_COMPUTE(filecheck-buffer=CHECK):-cpu -output-using-type

//TEST_INPUT: set outputBuffer = out ubuffer(data=[0 0 0 0], stride=4)

struct Matrix
{
    float data[16];

    [mutating]
    void map(IFunc<float, float> f)
    {
        for (int i = 0; i < 16; ++i)
        {
            data[i] = f(data[i]);
        }
    }
}

struct Applier
{
    float c;
    void apply(inout Matrix m)
    {
        m.map((float x) =>
        {
            if (x > 2.5)
                return x * c;
            else
                return x + c;
        });
    }
}

[numthreads(1,1,1)]
void computeMain()
{
    Matrix m = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
    Applier applier = { 1.5 };
    applier.apply(m);
    outputBuffer[0] = m.data[1];
    // CHECK: 3.5
    outputBuffer[1] = m.data[3];
    // CHECK: 6.0
}
```

The implementation is very simple: for a lambda expression, we will convert it to a functor struct during semantic checking. Given
```
int c = 2;
let lambda = (float x, float y)=>x+y;
```
We will convert it to:
```
struct _Lambda_UID : IFunc<float, float>
{
    int c;
    float operator()(float x, float y) { return x + y; }
};
let lambda = _Lambda_UID(2);
```


juliusikkala:
I wonder what happens if one tries to capture a NonCopyable type like RayQuery inside the lambda?

csyonghe:
You are absolutely right, I have addressed that in my propsoal here: https://github.com/shader-slang/spec/pull/13/files

Will update the implementation shortly to check for this as well.

tests/language-feature/lambda/lambda-0.slang
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+}
+
+[numthreads(1,1,1)]
+void computeMain()
+{
+    Matrix m = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
+    int c = 2;
+    m.map((float x) => (float)(x * c));
```
jkwak-work, line 27: I made some changes locally and played with it a bit.
I found that the following code fails to compile,
```
IFunc<float, float> createFunc(int c)
{
    return (float x) => (float)(x * 2);
}

    int c = 2;
    let func = createFunc(c);
    m.map(func);
```
The error message is,
```
(0): error 38101: 'this' expression can only be used in members of an aggregate type
```
I guess the current implementation needs to improve to capture from a function-parameter?

jkwak-work (approved):
Looks good to me.

tests/language-feature/lambda/lambda-simple.slang
```diff
@@ -0,0 +1,16 @@
...
+//TEST:COMPARE_COMPUTE(filecheck-buffer=CHECK):-cpu -output-using-type
+
+//TEST_INPUT: set outputBuffer = out ubuffer(data=[0 0 0 0], stride=4)
+RWStructuredBuffer<uint> outputBuffer;
+
+[numthreads(1,1,1)]
+void computeMain()
+{
+    int c = 1;
+    let lambda = (int x, int y) => x + y + c;
```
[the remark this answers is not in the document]
jkwak-work: It works! great!

tests/language-feature/lambda/lambda-0.slang
csyonghe: let me take a look at this.

jkwak-work (dismissed):
Great! Looks good to me.

PR #6943, opened 2025-04-29
Legalise out parameters for vertex shaders on metal

### LINKED ISSUE #6025: Compiling raster-simple example fails for metal
I am trying to compile raster-simple example for metal. Compilation fails with error 99999. Here is the command I'm using and the output I get:

`slangc -target metal Shaders.slang -o output.metal`

`(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: assert failure: toStyle != TypeCastStyle::Unknown`

Shader: [raster-simple](https://github.com/shader-slang/slang/blob/master/examples/reflection-api/raster-simple.slang)

expipiplus1:
Closes #6025

This packs these parameters into a struct which is returned, layout information and semantic decorations are preserved


source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -1822,6 +1824,7 @@ class LegalizeShaderEntryPointContext
             auto structType = as<IRStructType>(param->getDataType());
             builder.setInsertBefore(func->getFirstBlock()->getFirstOrdinaryInst());
             auto varLayout = findVarLayout(param);
+            SLANG_ASSERT(varLayout);
```
kaizhangNV, line 1827: I'm wondering if this will take effect in release build?

tests/metal/simple-vertex-position.slang
```diff
@@ -0,0 +1,27 @@
...
+//METAL: [[position]]
+//METALLIB: @vertexMain
+
+// Vertex Shader which writes to position
+void vertexMain(out float4 position : SV_Position, in uint vertexID : SV_VertexID)
+{
+    position = float4(0.6, 0.1, 0.6, 0.33);
+}
+
+// struct VertexOut
```
kaizhangNV, line 13: Is this still useful?

source/slang/slang-ir-lower-out-parameters.cpp
```diff
@@ -0,0 +1,686 @@
...
+        return;
+
+    // Create new entry point layout
+    auto entryPointLayout = builder.getEntryPointLayout(paramsVarLayout, resultVarLayout);
+
+    // Add layout decoration to the function
+    builder.addLayoutDecoration(newFunc, entryPointLayout);
+}
+
+IRFunc* lowerOutParameters(IRFunc* func, DiagnosticSink* sink, bool alwaysUseReturnStruct)
```
kaizhangNV, line 335: my overall suggestion is to split the logic in this function to several pieces, it's quite easy to get lost when reading this big function.

If I understand this function correctly, the logic will be

- collect parameters info
- create new entry point with new parameters
- call old entry point inside new entry point
- transfer entry point decor from old entry point to new entry point.


-

source/slang/slang-ir-lower-out-parameters.cpp
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// Main function that orchestrates the transformation
+IRFunc* lowerOutParameters(
+    IRFunc* func,
+    [[maybe_unused]] DiagnosticSink* sink,
+    bool alwaysUseReturnStruct)
+{
+    {
+        SourceManager s;
```
kaizhangNV, line 433: remove those debugging code.

source/slang/slang-ir-lower-out-parameters.cpp
```diff
@@ -0,0 +1,528 @@
...
+        outSemanticIndex = semanticDecor->getSemanticIndex();
+        return true;
+    }
+
+    // Check for semantic in entry point layout
+    IREntryPointLayout* entryLayout = nullptr;
+    IRVarLayout* paramsLayout = nullptr;
+    IRVarLayout* resultLayout = nullptr;
+
+    if (findEntryPointLayoutInfo(func, entryLayout, paramsLayout, resultLayout) && resultLayout)
```
kaizhangNV, line 84: `&& resultLayout` doesn't seem necessary

source/slang/slang-ir-lower-out-parameters.cpp
```diff
@@ -0,0 +1,528 @@
...
+    // // Set up layout information
+    // if (returnStruct)
+    // {
+    //     setupEntryPointLayout(func, newFunc, outKeys, paramToKeyMap, origToNewParamMap, builder);
+    // }
+
+    // Handle cleanup of original function
+    handleOriginalFunction(func, callResult);
+
+    {
```
kaizhangNV, line 519: same, cleanup these debugging code.

source/slang/slang-ir-legalize-varying-params.cpp
```diff
@@ -4005,11 +4009,43 @@ class LegalizeWGSLEntryPointContext : public LegalizeShaderEntryPointContext
     const UnownedStringSlice userSemanticName = toSlice("user_semantic");
 };
 
+void legalizeVertexShaderOutputParamsForMetal(DiagnosticSink* sink, EntryPointInfo& entryPoint)
+{
+    const auto oldFunc = entryPoint.entryPointFunc;
+    const bool alwaysUseReturnStruct = true;
```
kaizhangNV, line 4015: Is this variable always `true`? Then why do we still need it?

source/core/slang-list.h
```diff
@@ -192,6 +192,18 @@ class List
 
     Index getCount() const { return m_count; }
     Index getCapacity() const { return m_capacity; }
+    template<typename Predicate>
+    Index countIf(Predicate predicate) const
```
kaizhangNV, line 196: where is this function used?

source/slang/slang-ir-lower-out-parameters.cpp
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// Structure to hold parameter information
+struct ParamInfo
+{
+    IRParam* origParam;       // Original parameter
+    IRType* valueType;        // Parameter value type (without out/inout wrapper)
+    bool isOut;               // Is an out parameter
+    bool isInOut;             // Is an inout parameter
+    IRParam* newParam;        // New param (nullptr for pure out params)
```
kaizhangNV, line 108: it looks like `newParam` kind of duplicating its purpose with `isOut`?
So if newParam is non-nullptr, it must be out, right?

source/core/slang-list.h
expipiplus1: ah, it's vestigial, but seems like a useful thing to have

source/slang/slang-ir-legalize-varying-params.cpp
expipiplus1: it's true so far, but in the future there could be callers for which they don't want things automatically wrapped in a return struct

source/slang/slang-ir-lower-out-parameters.cpp
expipiplus1: ah, not always, this just hangs on to the pointer once it's created, for example this is always null just after we have collected the param info, I'll update the comment

mkeshavaNV (approved):

PR #6956, opened 2025-04-30
Fix: Preserve inout param modifications with OptiX IgnoreHit()

szihs:


source/slang/glsl.meta.slang.temp.h
```diff
@@ -0,0 +1,9930 @@
+SLANG_RAW("// TODO: These keywords are not recognized but they should be.\n")
```
juliusikkala, line 1: Is this file included by accident?

source/slang/hlsl.meta.slang
```diff
@@ -17266,6 +17266,7 @@ bool ReportHitOptix(float tHit, uint hitKind, expand each T attribs)
 /// Ignores the current intersection and continues traversal.
 /// @remarks Used in any-hit shaders to reject potential intersections
 /// @category raytracing
+[__noReturn]
 [require(cuda_glsl_hlsl_spirv, raytracing_anyhit)]
 void IgnoreHit()
```
juliusikkala, line 17271: Shouldn't `AcceptHitAndEndSearch()` also be noReturn?

szihs:
Based on internal discussions - Here is the update

> Yong: I don’t think there is a perfect solution for this problem. The way `optixIgnoreIntersection` works is incompatible with the `inout` parameter passing convention in Slang. They just can’t work together. I think the `inout HitBuffer buffer` parameter really needs to be a true pointer, so the user code does not rely on the copy in and copy out semantics of arguments before/after the call. Writes to that buffer should take effect immediately.
> 
> Here’s my proposed solution:
> 
> 1) implement an IR pass that undo the variable copy-in and copy-outs for inout parameters, and just replace it with direct pointers.
> 
> Right now if I have:
> ```
> Void f(inout v) {...}
> Void test(inout u) { f(u); }
> ```
> We will lower the above code to IR as:
> ```
> Void test(int * u) {
>    Int temp_v = *u;
>    f(&temp_v);
>    *u = *temp_v;
> }
> ```
> When targeting CUDA (and other targets with native pointer support), we could just get rid of this copying by passing `u` directly to `f`.
> 
> To make this possible, we will need some new decorations to identify the temp vars introduced for parameter passing. For example, we can introduce a [TempCallArgVar] decoration on `temp_v` so we lower the AST to:
> ```
> Void test(int * u) {
>    [TempCallArgVar]
>    Int temp_v = *u;
>    f(&temp_v);
>    *u = *temp_v;
> }
> ```
> 
> Then in an IR pass we can identify all [TempCallArgVar]-decorated IRVars, and find their corresponding source pointer by looking at the first OpStore on the var, and it shall be in the form of 
> %l = opLoad(some_ptr)
> opStore(%tempVar, %l)
> And we can know that the `tempVar` can be replaced by `some_ptr`. We will just remove all load/stores to tempVar and do the replacement. So the code will become:
> ```
>  Void test(int * u) {
>    f(u);
> }
> ```
> 
> This pass should be run as one of the first things in `linkAndOptimizeIR` when targeting CUDA or optix.
> 
> 2) update slang-ir-legalize-varying-params.cpp::processMutableParam to decorate the temp var it generates with [TempCallArgVar]. This should clean up your code to:
> void __anyhit__anyHitShader()
> {
>     HitBuffer_0 * _S11 = (HitBuffer_0 *)getOptiXRayPayloadPtr(); //local copy of <in>payload
>     [TempCallArgVar]
>     HitBuffer_0 _S12 = *_S11; // in stack for keeping local modifications
> 
>     HitBuffer_0 * _S13 = (HitBuffer_0 *)getOptiXRayPayloadPtr(); // <out> payload
> 
>     float _S14 = optixGetRayTmax();
> 
>     uint _S15 = optixGetInstanceId();
> 
>     RayHit_0 hit_1 = { _S14, int(_S15) };
>     int _S16 = HitBuffer_insert_0(&_S12, hit_1); // Maps to buffer.insert(hit) 
> 
>     if(_S14 < HitBuffer_last_0(_S12))
>     {
>         optixIgnoreIntersection();
>     }
> 
>     *_S13 = _S12;
>     return;
> }
> And now apply the temp-var copy undoing pass, you will get:
> void __anyhit__anyHitShader()
> {
>     HitBuffer_0 * _S11 = (HitBuffer_0 *)getOptiXRayPayloadPtr(); //local copy of <in>payload
> 
>     float _S14 = optixGetRayTmax();
> 
>     uint _S15 = optixGetInstanceId();
> 
>     RayHit_0 hit_1 = { _S14, int(_S15) };
>     int _S16 = HitBuffer_insert_0(_S11, hit_1); // Maps to buffer.insert(hit) 
> 
>     if(_S14 < HitBuffer_last_0(*_S11))
>     {
>         optixIgnoreIntersection();
>     }
>     return;
> }

I am currently looking to update PR based on newer logic

szihs:
The output looks mostly correct.

`#line 203
extern "C" __global__ void __anyhit__anyHitShader()
{

#line 203
    HitBuffer_0 * _S4 = (HitBuffer_0 *)getOptiXRayPayloadPtr();

#line 203
    HitBuffer_0 * _S5 = (HitBuffer_0 *)getOptiXRayPayloadPtr();



    float _S6 = optixGetRayTmax();

#line 207
    uint _S7 = optixGetInstanceId();

#line 207
    RayHit_0 hit_1 = RayHit_x24init_0(_S6, int(_S7));


    int _S8 = HitBuffer_insert_0(_S4, hit_1);


    if((hit_1.t_0) < (HitBuffer_last_0(*_S4)))
    {
        optixIgnoreIntersection();

#line 213
    }

#line 213
    *_S5 = *_S4;



    return;
}
`
We do see additional `_S5` load/store though. But since its pointer it should be a no-op.

source/slang/hlsl.meta.slang
szihs: updated code as per https://github.com/shader-slang/slang/pull/6956#issuecomment-2878524808

csyonghe:
I believe you can get rid of the redundant `getOptiXRayPayloadPtr` by adding 
`case kIROp_GetOptiXRayPayloadPtr:` to `mightHaveSideEffects`, and make `kIROp_GetOptiXRayPayloadPtr` to be a `HOISTABLE` inst in `slang-ir-inst-defs.h`.

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,141 @@
...
+                {
+                    if (varInst->findDecoration<IRTempCallArgVarDecoration>())
+                    {
+                        IRStore* initializingStore = nullptr;
+                        IRInst* originalParamPtr = nullptr;
+
+                        // Scan for the store that initializes this varInst
+                        // This store should be in the same block, after varInst.
+                        // The value stored should be an IRLoad from the original parameter pointer.
+                        for (auto scanInst = varInst->getNextInst();
```
csyonghe, line 68: Instead of traversing instructions following `varInst`, you should traverse its uses:

```
for (auto use = varInst->firstUse; use; use = use->nextUse)
{
     auto user = use->getUser();
     if (auto storeInst == as<IRStore>(user))
     {
              if (storeInst->getPtr() == varInst)
                 // we found the store into `varInst`.
     }
}
```

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,141 @@
...
+                             scanInst && scanInst->getParent() == block;
+                             scanInst = scanInst->getNextInst())
+                        {
+                            if (auto storeInst = as<IRStore>(scanInst))
```
csyonghe, line 72: I think you need to break whenever you see a `IRCall`. Because we may have:
```
IRVar a;
call func(&a);
b  = *a;
```
In this case `a` doesn't have a store before the call, and we should break out of the loop once we see a call.

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,141 @@
...
+                                        instsToRemove.add(initializingStore);
+                                        instsToRemove.add(varInst);
+                                    }
+
+                                    break; // Found the initializing store for varInst
+                                }
+                            }
+                            // Optimization: if we see another var declaration or a terminator, stop
+                            // scanning for this varInst's store.
+                            if (as<IRVar>(scanInst) || as<IRTerminatorInst>(scanInst))
```
csyonghe, line 91: terminator inst's nextInst is always nullptr, so no need to break when seeing terminator inst, the loop will already terminate.

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,141 @@
...
+        // Pass 2: Replace uses of temp vars with their original parameter pointers.
+        for (auto& pair : tempVarToOriginalParamPtrMap)
+        {
+            IRInst* tempVar = pair.first;
+            IRInst* originalParamPointer = pair.second;
+            tempVar->replaceUsesWith(originalParamPointer);
+        }
+
+        // Pass 3: Remove the temp vars and their initializing stores.
+        for (Index i = instsToRemove.getCount(); i > 0; --i)
```
csyonghe, line 119: it is fine to just `for (auto inst : instsToRemove)` here.

source/slang/slang-ir-undo-param-copy.cpp
szihs: Done

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,141 @@
...
+                    if (varInst->findDecoration<IRTempCallArgVarDecoration>())
+                    {
+                        IRStore* initializingStore = nullptr;
+                        IRInst* originalParamPtr = nullptr;
+
+                        // Scan for the store that initializes this varInst
+                        // This store should be in the same block, after varInst.
+                        // The value stored should be an IRLoad from the original parameter pointer.
+                        for (auto scanInst = varInst->getNextInst();
+                             scanInst && scanInst->getParent() == block;
```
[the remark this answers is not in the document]
szihs: Updated

source/slang/slang-ir-undo-param-copy.cpp
szihs: That's a good point. So effictively `call` acts as a terminator. Updated code

szihs:
With changes to 
- Marks `kIROp_GetOptiXRayPayloadPtr` as side-effect-free within and 
- Flags `GetOptiXRayPayloadPtr` as `HOISTABLE` in its definition
we can see the output seems correct.
```
#line 203
extern "C" __global__ void __anyhit__anyHitShader()
{


    float _S4 = optixGetRayTmax();

#line 207
    uint _S5 = optixGetInstanceId();

#line 207
    RayHit_0 hit_1 = RayHit_x24init_0(_S4, int(_S5));


    int _S6 = HitBuffer_insert_0((HitBuffer_0 *)getOptiXRayPayloadPtr(), hit_1);


    if((hit_1.t_0) < (HitBuffer_last_0(*(HitBuffer_0 *)getOptiXRayPayloadPtr())))
    {
        optixIgnoreIntersection();

#line 213
    }

#line 213
    *(HitBuffer_0 *)getOptiXRayPayloadPtr() = *(HitBuffer_0 *)getOptiXRayPayloadPtr();



    return;
}

```

source/slang/slang-emit.cpp
```diff
@@ -1594,6 +1595,13 @@ Result linkAndOptimizeIR(
     case CodeGenTarget::Metal:
     case CodeGenTarget::CPPSource:
     case CodeGenTarget::CUDASource:
+        // For CUDA/OptiX like targets, add our pass to replace inout parameter copies with direct
+        // pointers
+        undoParameterCopy(irModule);
+#if 0
```
mkeshavaNV, line 1601: please remove the debug code

source/slang/slang-ir-undo-param-copy.cpp
```diff
@@ -0,0 +1,139 @@
+#include "slang-ir-undo-param-copy.h"
+
+#include "slang-ir-dce.h"
+#include "slang-ir-insts.h"
+#include "slang-ir.h"
+
+#include <stdio.h> // For fprintf
```
mkeshavaNV, line 7: I don't see it used?

mkeshavaNV:
Can we add some test coverage for this?

szihs:
> Can we add some test coverage for this?

Yes, we should. I was thinking about run-test to compute payload modifications prior and post 'IgnoreHit'.

Can we utilise 'COMPUTE_COMPARE' test @mkeshavaNV …?

source/slang/slang-emit.cpp
szihs: There are similar debug constructs present for earlier passes... I can remove it.

csyonghe:
I don't think we have the infrastructure to execute OptiX test in `slang-test`. We will need to support optix in `slang-rhi` before we can run such tests. For now just do a //TEST:SIMPLE test and filecheck the output.

csyonghe:
What's going on with the:
```
 *(HitBuffer_0 *)getOptiXRayPayloadPtr() = *(HitBuffer_0 *)getOptiXRayPayloadPtr();
```
Seems superfluous and left-over from the undoParamCopy pass.
Have we make sure we remove any `stores` into the tempvar before replacing it with original pointer?

szihs:
> What's going on with the:
> 
> ```
>  *(HitBuffer_0 *)getOptiXRayPayloadPtr() = *(HitBuffer_0 *)getOptiXRayPayloadPtr();
> ```
> 
> Seems superfluous and left-over from the undoParamCopy pass. Have we make sure we remove any `stores` into the tempvar before replacing it with original pointer?

The `undoParameterCopy` pass replaces the temporary variable used for an inout parameter with the original parameter. While it removes the initial store into the temporary variable, it doesn't explicitly remove the final store from the temporary variable back to the original parameter. After the replacement, this final store becomes `store(originalParam, load(originalParam))`, which is redundant.

To fix this - the visitor now tracks the original parameter pointers that were involved in the parameter copy undoing.
It does so by specifically identifying IRStore instructions where the destination is an original parameter pointer and the value being stored is an IRLoad from that same original parameter pointer.

These identified redundant stores are now also marked for removal.


-Added test case as well.

csyonghe (approved):

PR #6963, opened 2025-04-30
Fix compiler warning with clang 18.1.8 on windows

jkwak-work:
`include/slang.h` has `SLANG_NO_THROW` on a few member functions.
The replay implementation didn't use them and compiler was printing warnings which is treated as an error on my machine.


jarcherNV (approved):
LGTM

Issue #6952, opened 2025-04-29
[Bindless] Invalid SPIRV


skallweitNV:
I'm trying to get bindless working with Vulkan but hitting a few issues:

For every accessed bindless resource type, slang generates a `__slang_resource_heap` descriptor set with a fixed set index (defined by the compiler option `BindlessSpaceIndex`) and binding 0. This is not working, because for every descriptor type we need a different binding index. While we can alias the texture binding to all different types of textures (dimensions, arrayness etc) we cannot alias the same binding index for buffer/texture/sampler/etc descriptors, as they need different descriptor types.

In addition, Vulkan doesn't allow "gaps" in descriptor sets in a pipeline layout, unless an extension is enabled, so having a fixed descriptor set index is not really working well.

Here are my suggestions for making bindless work with Vulkan:

- Dynamically allocate the bindless descriptor set index next to the last parameter block descriptor set index, so we get a contiguous list of descriptor sets. If we have global parameters + 3 parameter blocks, we should allocate 4 descriptor sets for the globals + parameter blocks and make the bindless descriptor set have index `4` for a total of 5 descriptor sets.
- Allocate one binding for each descriptor type that is supported. For now that could be:
  - `VK_DESCRIPTOR_TYPE_SAMPLER` on binding 0
  - `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER` on binding 1
  - `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE` on binding 3
  - `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE` on binding 4
  - `VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER` on binding 5
  - `VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER` on binding 6
  - `VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER` on binding 7
  - `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` on binding 8
  - `VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR` on binding 9
- The `descriptor type -> binding` mapping needs to be consistent across all compiled shaders, otherwise we cannot share the bindless descriptor set between them.
[...]

PR #6967, opened 2025-04-30
Add `IOpaqueDescriptor::descriptorAccess`.

csyonghe:
This is to allow user defined `getDescriptorFromHandle` function to tell more precisely the type of the resource.

Partially addresses #6952.


jarcherNV (approved):
LGTM

Issue #6957, opened 2025-04-30
Add CI test for slang-playground


jkwak-work:

jkwak-work:
Both slang and playground repos now has CI with the smoke-test step.
Closing as completed.

PR #6969, opened 2025-04-30
Smoke test WASM as a part of CI

jkwak-work:
This adds a smoke-test.js that can be used for testing slang-wasm binary.

The test will run as a part of WASM build of the current CI.

Related to,
-  https://github.com/shader-slang/slang/issues/6957


.github/workflows/ci.yml
```diff
@@ -135,8 +135,12 @@ jobs:
               cmake --install build --config Release --component generators --prefix generators
               emcmake cmake -DSLANG_GENERATORS_PATH=generators/bin --preset emscripten -DSLANG_SLANG_LLVM_FLAVOR=DISABLE
               cmake --build --preset emscripten --config "$cmake_config" --target slang-wasm
-              [ -f "build.em/$cmake_config/bin/slang-wasm.wasm" ]
-              [ -f "build.em/$cmake_config/bin/slang-wasm.js" ]
+              cd "build.em/$cmake_config/bin"
+              [ -f "slang-wasm.wasm" ]
+              [ -f "slang-wasm.js" ]
+              mkdir test
+              cp ../../../tests/wgsl/rand_float.slang ../../../source/slang-wasm/smoke-test.js test/
```
csyonghe, line 142: can we put both the .slang and the .js file under `tests/wasm/smoke`?

jkwak-work: I can do that.
But I made a change just now that rand_float.slang will be downloaded with git-clone.
I will move smoke-test.js to tests/wasm/smoke

csyonghe: Are we pulling playground repo just to get that file?
What else is needed in order to test the wasm? Ideally the CI here does not depend on the playground repo.

csyonghe:
I have a question in one of my comments that might have been missed:

Are we pulling playground repo just to get that file?
What else is needed in order to test the wasm? Ideally the CI here does not depend on the playground repo.

jkwak-work:
> I have a question in one of my comments that might have been missed:
> 
> Are we pulling playground repo just to get that test slang shader? What else is needed in order to test the wasm? Ideally the CI here does not depend on the playground repo.

Yes, I am cloning the repo only to get the testing shader.
I guess the redundancy is not a big problem in this case.
I will bring back to make a copy of the shader.

csyonghe (approved):

PR #6971, opened 2025-05-01
Fix build on GCC 15

juliusikkala:
`isConsole` was triggering `-Woverloaded-virtual` (there's a virtual function with the same name) and `slang-glslang.h` was relying on `stdint.h` being transitively included.


jarcherNV (approved):
LGTM

PR #6973, opened 2025-05-01
Add IREnumType to distinguish enums from ints and each other

### LINKED ISSUE #6964: Witness tables for all enums are shared
```slang
interface IThing
{
    void declareSelf();
}

enum FirstEnum
{
    A,
    B
};

enum SecondEnum
{
    C,
    D
};

extension FirstEnum: IThing
{
    void declareSelf()
    {
        printf("FirstEnum: %d\n", this);
    }
}

extension SecondEnum: IThing
{
    void declareSelf()
    {
        printf("SecondEnum: %d\n", this);
    }
}

void declareType<T: IThing>(T val)
{
    val.declareSelf();
}

{
    declareType(FirstEnum.A);
    declareType(SecondEnum.C);
    return 0;
}
```

Surprisingly, this prints "FirstEnum" twice!
Before lowering to IR, the witness tables have distinct pointers. After lowering, they are the same. Perhaps there is some erroneous caching going on here.

The IR witness table is also implemented for `Int`, but I guess that isn't a problem?
```
witness_table %199      : witness_table_t(%IThing)(Int)
{
        witness_table_entry(%115,%declareSelf)
}
```

This issue is blocking #6916, because this prevents disambiguating between separate enums (which should have different witness tables, even if the IR type is the same), so an incorrect `catch` block can run if there are multiple catches.

juliusikkala:
Fixes #6964. At least one questionable part is my implementation of `processEnumCast()`; I'm not sure if I need to do something to the `value` if it's of an enum type (like swap out the type of the `inst` from IREnumType to the underlying integer type)? Didn't figure out how to do that, and this seems to work as-is.


jarcherNV (approved):
I can't really speak to your comment on the implementation of processEnumCast, but as long as this passes the CI tests I assume it's fine.

This looks like a reasonable implementation of Yong's proposed IREnumType to me.

juliusikkala:
Whoops, the .expected.txt was missing for the test, causing failures.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -9228,18 +9228,14 @@ struct DeclLoweringVisitor : DeclVisitor<DeclLoweringVisitor, LoweredValInfo>
...
-        // An `enum` declaration will currently lower directly to its "tag"
-        // type, so that any references to the `enum` become referenes to
-        // the tag type instead.
-        //
         // TODO: if we ever support `enum` types with payloads, we would
         // need to make the `enum` lower to some kind of custom "tagged union"
         // type.
 
         IRType* loweredTagType = lowerType(subContext, decl->tagType);
+        IRType* enumType = subBuilder->createEnumType(loweredTagType);
```
csyonghe, line 9236: When we do so, we will need to make sure enum types have linkage decorations, so they can be deduplicated during linking.

csyonghe:
I think the enum lowering pass also need to happen much later in the pipeline until at least lower generics pass is done. Because lowering enums can leads to witness tables being incorrectly deduplicated.

tests/bugs/gh-6964.slang
```diff
@@ -0,0 +1,53 @@
+//TEST(compute):COMPARE_COMPUTE: -shaderobj
+//TEST(compute):COMPARE_COMPUTE: -vk -shaderobj
+//TEST(compute):COMPARE_COMPUTE: -cpu -shaderobj
```
fairywreath, line 3: Use a filecheck-buffer here and do `CHECK: ` inside the test file instead of having an additional .expected.txt file.

source/slang/slang-lower-to-ir.cpp
juliusikkala: Okay, I've added a call to `addLinkageDecoration`.

source/slang/slang-ir-lower-enum-type.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+            processInst(inst);
+
+            for (auto child = inst->getLastChild(); child; child = child->getPrevInst())
+            {
+                addToWorkList(child);
+            }
+        }
+
+        // Replace all optional types with lowered struct types.
```
csyonghe, line 127: The comments seemed to be copied from another pass and need to be updated here.

juliusikkala: Whoops, good catch. I had used `slang-ir-lower-optional-type.cpp` as reference.

csyonghe:
The type replacement you are doing right now should work fine for enum types constants and values.

csyonghe:
Delaying enum type lowering could mean that we also need to take care of  enum types in `slang-ir-layout.cpp`, `_calcSizeAndAlignment`.

Also in autodiff, we need to make sure the new opcodes are marked as non differentiable in `ForwardDiffTranscriber::transcribeInstImpl`.

juliusikkala:
Thanks, I completely missed those. I also spotted a couple more switches in `slang-ir-util.cpp` where it made sense to handle enum types specifically.

csyonghe (approved):
Looks good to me! Thanks for addressing the comments quickly!

csyonghe:
As a future work, we could improve our target code generation for HLSL, metal, etc. by emitting the enum type as a enum in the target language for better readability.

PR #6986, opened 2025-05-03
Remove extra User Guide readthedocs TOC entries by turning H1 headings into H2

aidanfnv:
readthedocs treats H1 headings as document titles in the sidebar table of contents.
This change turns all extra H1 headings into H2, H2 to H3, etc. in the "Metal-Specific Functionalities" doc.

### DIFF
--- a/docs/user-guide/a2-02-metal-target-specific.md
+++ b/docs/user-guide/a2-02-metal-target-specific.md
@@ -3,12 +3,12 @@ layout: user-guide
 permalink: /user-guide/metal-target-specific
 ---
 
-# Metal-Specific Functionalities
+## Metal-Specific Functionalities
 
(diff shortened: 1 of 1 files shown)


aidanfnv:
Ah, this won't work, the headings need to be changed to the same style as the SPIRV- and WGSL-Target Specific Functionalities docs

aidanfnv:
Actually, that wouldn't work either. Using the same style as the other docs results in having multiple H1 headings in the other style, which readthedocs ignores, but the build_toc script takes the last H1 heading as the doc title, leading to an incorrect title in toc.html.
On second look, I am not quite sure that the additional H1 headings should even be H1.

csyonghe (approved):

Issue #6414, opened 2025-02-20
ASCII single character support (as int8_t)


KoT3isGood:
Syntax can be same as in C
eg:
`'a', '\t'`

bmillsNV:
Closing as #6989 is merged

PR #6989, opened 2025-05-03
Parse char literals as integers

### LINKED ISSUE #6984: Feature Request: support character literals like HLSL does
Hello!

When using slang to convert HLSL shaders to WGSL, slang complains when using character literals such as '1'.

Character literals are helpful when making debugging functionality, like printf from a shader.

HLSL is just fine with these character literals and converts them into their integral ascii value.

Could we have the same functionality in slang?

Thanks!

juliusikkala:
Closes #6984 and #6414. Conveniently, there was already a token type for character literals in the codebase, so only a conversion to an integer and small parser changes were needed to make this work. The approach taken here is to treat char literals as integer literals corresponding to the Unicode codepoint of the given char.

This PR leaves actual char literal AST / IR types as future work (not volunteering for that one, seems like it's quite a lot more work and is not immediately useful to me or the mentioned issues). They would be necessary to keep char literals as char literals on targets that support them, like HLSL and C++.


csyonghe:
Yeah, we still need to decide what `char` type in Slang should be. I am leaning towards that `char` should be a 32bit unicode code point, and String.operator[] should be returning the decoded unicode char. But this could be mean the operation is more heavyweight than what people would expect. So we may not have a type named exactly as `char` to avoid confusion with C++'s `char`, which isn't really a char.

source/compiler-core/slang-lexer.cpp
```diff
@@ -829,17 +829,47 @@ FloatingPointLiteralValue getFloatingPointLiteralValue(
     return value;
 }
 
-static void _lexStringLiteralBody(Lexer* lexer, char quote)
+IntegerLiteralValue getCharLiteralValue(Token const& token)
 {
+    const UnownedStringSlice content = token.getContent();
```
csyonghe, line 834: what if the token content contains escape sequences?

e.g. `'\t'`?

csyonghe: May want to use `StringEscapeUtil::unquote()` here.

juliusikkala: That's a really good point. I've now added an escaped char in the test and unquoted appropriately. It had to be done in a bit of a roundabout way; `StringEscapeUtil::unquote()` doesn't unescape, and `StringEscapeUtil::unescapeString` assumes that the string starts with `"` and not `'`.

juliusikkala:
I also agree that chars should definitely be 32-bit unicode codepoints. `char` carries a lot of C baggage. One option would be to just not have a 'char' type, and instead just treat char literals as `uint32_t` (which is what this PR does). At least that would be pretty explicit about the size, but wouldn't help at all with preserving literals into target languages. A name like `char32_t` could be nice.

Making `String.operator[]` decode UTF-8 to UTF-32 seems like a bad idea. For random access to a char in the middle of the string, you'd need to resolve the sizes of all preceding chars. So iterating over a string would suddenly become an O(n^2) operation, unless the string stores some lookup structure, but at that point, why not just make the String UTF-32 in the first place? So perhaps it'd make more sense to have some more controlled way of iterating over the string, like

```slang
uint32_t offset = 0;
while(let c = str.advance(offset))
{
}
```

Or just make Strings actually store chars in UTF-32.

csyonghe:
That being said, we also need a language-wise convention for iterators. This quickly gets into lifetime/ownership problems which we don’t have a proper language spec for yet…

source/compiler-core/slang-lexer.cpp
```diff
@@ -829,17 +830,48 @@ FloatingPointLiteralValue getFloatingPointLiteralValue(
     return value;
 }
 
-static void _lexStringLiteralBody(Lexer* lexer, char quote)
+IntegerLiteralValue getCharLiteralValue(Token const& token)
 {
+    String unquotedContent = StringEscapeUtil::unquote('\'', token.getContent());
+    StringBuilder unescaped(4);
+    auto escapeHandler = StringEscapeUtil::getHandler(StringEscapeUtil::Style::Cpp);
+    escapeHandler->appendUnescaped(unquotedContent.getUnownedSlice(), unescaped);
```
csyonghe, line 838: can just call `appendUnquoted` which calls `unquote + unescape` inside.

csyonghe (approved):

Issue #6708, opened 2025-03-31
Support missing float16 type keywords in GLSL


jkwak-work:
Following types are valid types in GLSL with an extension, "GL_EXT_shader_explicit_arithmetic_types_float16". Slang currently doesn't recognize them.
- f16vec2
- f16vec3
- f16vec4
- f16mat2
- f16mat3
- f16mat4
- f16mat2x3
- f16mat2x4
- f16mat3x2
- f16mat3x4
- f16mat4x2
- f16mat4x3

We need to add them in glsl.meta.slang as typealias like how other types are aliased.

kaizhangNV:
can you update the sprint?

PR #6997, opened 2025-05-05
Add interger pack/unpack intrinsic to glsl

### LINKED ISSUE #6949: Missing integer pack/unpack intrinsic

### LINKED ISSUE #7007: bitcast issue when pack integers to 32bit interger
assuming we have an array of vector<int16_t, 2> = [-3498, 1], and want to cast it into a 32bit integer

bitCast([-3498, 1]) 
will be lowered into
```
  let  %62    : UInt  = bitCast(-3498)                 0xFFFFF256
  let  %64    : UInt  = bitCast(1)                         0x00000001
  let  %65    : UInt  = shl(%64, 16 : UInt)         0x00010000
  let  %66    : UInt  = add(%62, %65)               0xFFFFF256 + 00010000 = 0x0000F256 (overflow)
  let  %67    : Int   = bitCast(%66)                     0x0000F256
```

These is definitely wrong because after bitcast, the input and output should be the same regarding to bit.

So we should fix the lowering logic.

kaizhangNV:
Add unpack8, unpack16 and pack32 intrinsics to glsl.
Close #6949.

Also fix the issue in lowering `bitcast`.
close #7007.

### DIFF
--- a/source/slang/glsl.meta.slang
+++ b/source/slang/glsl.meta.slang
@@ -736,178 +736,71 @@ public uvec2 unpackDouble2x32(double v)
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
-public u8vec4 unpack8(uint32_t u)
-{
-    __target_switch
-    {
-    case glsl: __intrinsic_asm "unpack8";
-    default:
-        uint32_t4 unpackRes = unpackUint4x8ToUint32(u);
-        u8vec4 res = u8vec4(unpackRes);
-        return res;
-    }
-}
+__intrinsic_op($(kIROp_BitCast))
+public u8vec4 unpack8(uint32_t u);
 
 /// Unpack a 32-bit signed integer into 4 8-bit signed integers.
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
-public i8vec4 unpack8(int32_t u)
-{
-    __target_switch
-    {
-    case glsl: __intrinsic_asm "unpack8";
-    default:
-        uint32_t4 unpackRes = unpackInt4x8ToInt32(u);
-        i8vec4 res = i8vec4(unpackRes);
-        return res;
-    }
-}
+__intrinsic_op($(kIROp_BitCast))
+public i8vec4 unpack8(int32_t u);
 
 /// Unpack a 16-bit unsigned integer into 2 8-bit unsigned integers.
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
-public u8vec2 unpack8(uint16_t u)
-{
-    __target_switch
-    {
-    case glsl: __intrinsic_asm "unpack8";
-    case spirv:
-        return spirv_asm
-        {
-            result:$$vector<uint8_t, 2> = OpBitcast $u;
-        };
-    default:
-        return u8vec2
-        (
-            uint8_t(u),
-            uint8_t(u >> 8U),
-        );
-    }
-}
+__intrinsic_op($(kIROp_BitCast))
+public u8vec2 unpack8(uint16_t u);
 
 /// Unpack a 16-bit signed integer into 2 8-bit signed integers.
 [__readNone]
 [ForceInline]
 [require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
-public i8vec2 unpack8(int16_t u)
-{
-    __target_switch
-    {
-    case glsl: __intrinsic_asm "unpack8";
-    case spirv:
-        return spirv_asm
-        {
-            result:$$vector<int8_t, 2> = OpBitcast $u;
-        };
-    default:
-        return i8vec2
-        (
-            int8_t(u),
-            int8_t(u >> 8U),
-        );
-    }
-}
+__intrinsic_op($(kIROp_BitCast))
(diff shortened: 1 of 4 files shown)


source/slang/glsl.meta.slang
```diff
@@ -724,6 +732,183 @@ public uvec2 unpackDouble2x32(double v)
...
+[__readNone]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
+public u8vec4 unpack8(uint32_t u)
+{
+    __target_switch
+    {
+    case glsl: __intrinsic_asm "unpack8";
+    default:
+        uint32_t4 unpackRes = unpackUint4x8ToUint32(u);
```
gtong-nv, line 745: what does `unpackUint4x8ToUint32` do here? from the comment
 > Unpack 4 unsigned 8-bit values into a vector of 32 bit integers.

Does it mean it turns a 4 8-bit unsigned integers (in the form of 32-bit uint) to a 4 32-bit vec? And then we cast that to a 4 8-bit vec with ` u8vec4(unpackRes);`  ?
Feels like a bit of waster of space here?

gtong-nv:
Looks good to me in general. Left a question. 
Do we need to add a test?

gtong-nv (dismissed):
LGTM

jkwak-work:
Is there an issue related to this PR?

kaizhangNV:
> Is there an issue related to this PR?

It has actually, let me update the description.

source/slang/glsl.meta.slang
```diff
@@ -17,6 +17,10 @@
 // Section 4.1. 'asic Types'
 //
 
+public typealias f16vec2 = half2;
+public typealias f16vec3 = half3;
+public typealias f16vec4 = half4;
```
jkwak-work, line 22: It doesn't look like this PR needs these types.
Can we remove them?

Note that these types are not available by default in GLSL.
We need to require the following extension,
```
GL_EXT_shader_explicit_arithmetic_types_float16
```
See:
- https://github.com/shader-slang/slang/issues/6708

source/slang/glsl.meta.slang
kaizhangNV: I just don't want to repeating the same code.
Also for spirv target, those will be optimized out. So should be fine.

source/slang/glsl.meta.slang
```diff
@@ -724,6 +732,183 @@ public uvec2 unpackDouble2x32(double v)
...
+        uint32_t res = __packUint2x16(v);
+        return res;
+    }
+}
+
+/// Pack 2 16-bit signed integers into a 32-bit signed integer.
+[__readNone]
+[ForceInline]
+[require(cpp_cuda_glsl_hlsl_spirv, pack_vector)]
+public int32_t pack32(i16vec2 v)
```
jkwak-work, line 871: Don't we need an overloading for `f16vec2` as well?
```
public int32_t pack32(f16vec2 v)
```

kaizhangNV:
This PR is driven by the GGML, so I didn't any any extra that are not used in GGML.

source/slang/glsl.meta.slang
kaizhangNV: No, there is no such intrinsic.

source/slang/glsl.meta.slang
csyonghe: Why can't we just map all of these to `__intrinsic_op($(kIROp_BitCast))`?

source/slang/glsl.meta.slang
kaizhangNV: > This PR is driven by the GGML, so I didn't any any extra that are not used in GGML.

So can't remove them.

csyonghe:
I think all the pack/unpack is just bitcast. It maybe better to mark them as `__intrinsic_op($(kIROp_BitCast))` so we don't need to implement them at all.

source/slang/glsl.meta.slang
kaizhangNV: Is those GLSL extension meaningful in slang?
I think we will ensure to emit correct spirv cap and extension when we see those 16 bit floating numbers.
Slang will just ignore all those GLSL extensions.

kaizhangNV:
> I think all the pack/unpack is just bitcast. It maybe better to mark them as __intrinsic_op($(kIROp_BitCast)) so we don't need to implement them at all.

I can try this.

kaizhangNV:
kIROp_BitCast can work in most cases, but I found there is an issue in this test:

```
    i16vec2 h = {0xF256, 0x1234};
    outputBuffer.h = pack32(h);
    // BUF-NEXT: 1234F256
```


this result us `1233F256`, which shows that there seems a bug in constant folding.

kaizhangNV:
Note it doesn't have problem for `uvec2` type, only `ivec2`.
And when packing 2 16 bits into 32-bit, the first 16 bit always minus 1.

` i16vec2 h = {Y, X}; => (X-1)Y. `

I don't think this is correct behavior.

source/slang/glsl.meta.slang
kaizhangNV: I've made this change in the latest PR.
But there is a bug in BitCast, so I also fix it in this PR as well.

csyonghe (approved):

kaizhangNV:
File an issue https://github.com/shader-slang/slang/issues/7013 to track the refactor.

Issue #6669, opened 2025-03-23
Assert/segfault on simple vertex shader


Please-just-dont:
So I don't know if the syntax I'm using is correct because I'm only new, but the point is that slangc is failing the compilation and returning an error value, but not writing anything to stdout or stderr:

```
[[vk::location(0)]] in int2 inPos;


struct PushConstants {
    float point_size;
    float padding;
    float2 resolution;
};



[shader("vertex")]
void main(uniform PushConstants push_constants, out float4 pos: SV_Position)
{



    float2 wnd_position = float2( 1, 1 );
    float2 vertex_position = inPos.xy;
    float2 screenspace_pos = inPos.xy;
    

    
    screenspace_pos = wnd_position + vertex_position;


    
    float2 NDC_pos = float2(screenspace_pos.x / 3840 * 2 - 1, screenspace_pos.y / 2160 * 2 - 1);

    pos = float4(NDC_pos, 1, 1);


}
```

And here's the weird part, the bug goes away by deleting different parts of the code, so it's not one particular part of the code that makes it happen. Try commenting out different parts and see.

I'm using the slangc compiler that comes with the Vulkan SDK 1.4.304.1

expipiplus1:
This fails in more than one way:

`build/Debug/bin/slangc a.slang -stage vertex -target metal` -> segmentation fault

`build/Debug/bin/slangc a.slang -stage vertex -target spirv` -> assert in glsl legalization

sricker-nvidia:
Started investigating. Able to repro the metal and spirv cases. Issue seems related to mixing global variable inputs and a uniform struct as a param to a shader entry point. I created a smaller repro case:
`in int2 inPos;

struct PushConstants {
    float point_size;
    float padding;
    float2 resolution;
};

[shader("vertex")]
void main(uniform PushConstants push_constants, out float4 pos: SV_Position)
{
    pos = float4(inPos.x, inPos.y, 0, 0);
}`

Using `void main(PushConstants push_constants, out float4 pos: SV_Position)` (removing the uniform qualifier) compiles ok, although I did not validate the emitted spirv code. Similarly, using `void main(uniform float push_constant, out float4 pos: SV_Position)` works ok.

The bug may disappear when commenting out different parts of the code due to compiler optimizations. I've found that I can also get rid of the bug by not actually using inPos, for example I can use, `pos = float4(0,0 0, 0);` in the simplified shader to make the bug go away.

[...]

Issue #5339, opened 2024-10-17
[CTS] SPIR-V Validation failure, "Expected Image Operand ConstOffset to be a const object"


jkwak-work:
Three CTS tests are occasionally causing the crash of deqp-vk.exe.
```
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.clamp_to_edge_repeat
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.mirrored_repeat_clamp_to_edge
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.repeat_mirrored_repeat
```
When the validation is enabled, those tests fail with the following message,
> Expected Image Operand ConstOffset to be a const objec

A simple repro is following,
```
#extension GL_EXT_gpu_shader5 : require

layout (location = 0) out mediump vec4 o_color;

layout (location = 0) in highp vec2 v_texCoord;

layout (binding = 0) uniform highp sampler2D u_sampler;
layout (binding = 1) uniform offset { highp ivec2 u_offset; };

void main(void)
{
                o_color = textureGatherOffset(u_sampler, v_texCoord, u_offset);
}
```
When the shader above is stored as `test.slang`, observe the validation error with the following command,
```
$ slangc.exe -target spirv -entry main -stage fragment -allow-glsl test.slang

error: line 53: Expected Image Operand ConstOffset to be a const object
  %29 = OpImageGather %v4float %24 %27 %int_0 ConstOffset %21

(0): internal error 99999: Validation of generated SPIR-V failed. SPIRV generated:
; SPIR-V
; Version: 1.5
; Generator: Khronos; 40
; Bound: 37
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %11 %offset %u_sampler %31 %entryPointParam_main_o_color %v_texCoord
               OpExecutionMode %main OriginUpperLeft

               ; Debug Information
               OpSource Slang 1
               OpName %v_texCoord "v_texCoord"  ; id %9
               OpName %SLANG_ParameterGroup_offset_std140 "SLANG_ParameterGroup_offset_std140"  ; id %13
               OpMemberName %SLANG_ParameterGroup_offset_std140 0 "u_offset"
               OpName %offset "offset"  ; id %17
               OpName %u_sampler "u_sampler"  ; id %26
[...]

PR #6998, opened 2025-05-05
Address issues with GLSL style global in/out vars (#6669)

sricker-nvidia:
Asserts and segfaults were observed trying to compile a simple vertex shader like:

===================
in int2 inPos;

[shader("vertex")]
main(uniform int2 test1, int2 test2, out float4 pos: SV_Position) void main()
{
    // Bogus use of all input vars to prevent optimizing out.
    pos = float4(inPos.x, test1.x, test2.y, 0);
}
===================

Further investigation found that while replacing "uniform int2 test1" with "int2 test1" allowed for successful compilation, the resulting output shader would have overlapping location qualifiers. For example, compiling the above with "int2 test1" to glsl might give:

===================
...
layout(location = 0) in ivec2 test1_0;
layout(location = 1) in ivec2 test2_0;
layout(location = 0) in ivec2 translatedGlobalParams_inPos_0; ...
===================

This was because Slang does not actually support mixing GLSL style global in/out vars and entry point params. However, this is never checked for or noted in documentation. Slang source also assumes input shaders do not mix these and these assumptions ultimately led to the observed asserts and seg faults when using uniform entry point params.

This change makes updates to throw an error when the compiler detects that it is trying to translate global in/out variables into entry point params when an entry point already contains parameters, allowing for compilation to fail gracefully. Change also makes minor updates to documentation to clarify allowed usage of GLSL-flavored global in/out variables.


source/slang/slang-ir-translate-global-varying-var.cpp
```diff
@@ -65,6 +65,13 @@ struct GlobalVarTranslationContext
             if (!entryPointFunc)
                 continue;
 
+            // It's an error to mix GLSL-flavored global in/out vars and entry point params.
+            // Before we continue, check for this here and throw an error if mixing is
+            // detected.
+
+            if (entryPointFunc->getParamCount() != 0)
+                SLANG_INVALID_OPERATION("Mixing of GLSL-flavored global in/out variables with entry point parameters is not supported.");
```
fairywreath, line 73: It's better to use diagnostic sink here instead of `SLANG_INVALID_OPERATION`, you can add a new error to `slang-diagnostics-defs.h`

source/slang/slang-ir-translate-global-varying-var.cpp
```diff
@@ -65,6 +65,13 @@ struct GlobalVarTranslationContext
             if (!entryPointFunc)
                 continue;
 
+            // It's an error to mix GLSL-flavored global in/out vars and entry point params.
+            // Before we continue, check for this here and throw an error if mixing is
+            // detected.
+
+            if (entryPointFunc->getParamCount() != 0)
```
csyonghe, line 72: I think this is likely too strict and will fail a lot of existing tests.

Can we just support this scenario instead of throwing an error?

csyonghe: At very least, we should only error when there are existing parameters that are varying in/out.

Existing uniform parameters, or builtin parameters are fine.

```
// should be allowed:
in float3 color;
void entry(float4 p: SV_Position, RWStructuredBuffer<float> b) {}
```

sricker-nvidia:
Might make sense to update the label to "pr: breaking" as this does effect the tests/bugs/gh-5339.slang test without additional edits and could have an effect on existing code (although it might be surprising if that code worked as intended).

jarcherNV:
> Might make sense to update the label to "pr: breaking" as this does effect the tests/bugs/[gh-5339](https://github.com/shader-slang/slang/issues/5339).slang test without additional edits and could have an effect on existing code (although it might be surprising if that code worked as intended).

Would it be possible to fix those tests in your PR?

source/slang/slang-ir-translate-global-varying-var.cpp
sricker-nvidia: The problem with allowing existing uniform parameters, or builtin parameters is that translateGlobalVaryingVar ends up replacing all of the entry point params before we try and collect entry point uniforms and make them global. The result is that collectEntryPointUniformParams ends up seg faulting because it gets a format for the entry point params that it doesn't expect after they have been transformed in translateGlobalVaryingVar. It might be possible to shift things around to avoid the seg fault, but it still would not solve the issue with the parameter offsets.

sricker-nvidia: I can use something like:
````
for (auto param : entryPointFunc->getParams())
{
    if (!param->findDecoration<IRSemanticDecoration>())
        // error
````
to allow something like
````
in float3 color;
void entry(float4 p: SV_Position, out float4 pos: SV_Position)
````
such that at least builtin parameters are fine, but working around uniform entry point parameters is more complicated to support.

**Comment by sricker-nvidia** [comment]
> > Might make sense to update the label to "pr: breaking" as this does effect the tests/bugs/[gh-5339](https://github.com/shader-slang/slang/issues/5339).slang test without additional edits and could have an effect on existing code (although it might be surprising if that code worked as intended).
> 
> Would it be possible to fix those tests in your PR?

Yeah, I'm working on it now for the next patch, I just wanted to note things in case someone wiser than myself thought the non-breaking label was incorrect.

File: source/slang/slang-ir-translate-global-varying-var.cpp (hunk 2 above)
Comment (csyonghe) [reply source/slang/slang-ir-translate-global-varying-var.cpp]: OK, let’s just disallow all parameters then.

**Comment by csyonghe** [comment]
Yes, this will need to be labeled as breaking change.

File: docs/user-guide/02-conventional-features.md [hunk 3]
```diff
@@ -582,6 +582,14 @@ int gCounter = 0;
 uniform float scaleFactor;
 ```
 
+Slang also supports GLSL-flavored global in/out variables:
```
csyonghe, line 585: Let's not advise it here. Anything we do for GLSL-flavored syntax is not part of the Slang language, and does not go into the user-guide. This should go into the GLSL on-ramp guide.

docs/user-guide/02-conventional-features.md
```diff
@@ -676,6 +684,10 @@ Entry-point parameters in Slang default to varying, but may be explicitly marked
 
 If an entry-point function declares a non-`void` result type, then its result behaves like an anonymous `out` parameter that is varying.
 
+> #### Note ####
```
csyonghe, line 687: Move to glsl onramp guide.

docs/user-guide/02-conventional-features.md
sricker-nvidia: ack, I'll move this and other notes to the onramp guide.

docs/user-guide/02-conventional-features.md
sricker-nvidia: ack

sricker-nvidia:
Noting something that might be worth discussing and potentially opening a separate issue for. In slang\source\slang\hlsl.meta.slang, there are a couple of intrinsic functions like WaveGetLaneIndex that we implement using GLSL style in/out variables:

````
/// Built-in values or system value semantics represented as in/out global variables.
/// This allows the built-ins to be arbitrarily used from a global scope without being
/// explicitly passed as entry point parameters.
in uint __builtinWaveLaneIndex : SV_WaveLaneIndex;
in uint __builtinWaveLaneCount : SV_WaveLaneCount;
in uint __builtinQuadLaneIndex : SV_QuadLaneIndex;
...
uint WaveGetLaneIndex()
{
...
    case metal:
        return __builtinWaveLaneIndex;
    case wgsl:
        __requireTargetExtension("subgroups");
        return __builtinWaveLaneIndex;
...
````
If using these functions, it's not immediately obvious that one could end up mixing GLSL style in/out vars with entry point vars depending on the platform. A resulting error message could be difficult to understand and I had to update a few of our tests in this change in order to account for this issue.

**Comment by fairywreath** [comment]
Yes that is indeed problematic - those builtins were added to support regular HLSL functions(eg. `WaveGetLaneIndex()`) and we should not error out when these functions are used alongside HLSL-style system value/builtin inputs. I am OK with having these addressed later.

**Comment by csyonghe** [comment]
I think we should definitely allow global vars with system semantics to co-exist with entry point parameters.

**Comment by csyonghe** [comment]
Using these intrinsics with entrypoint parameters should not be an error. We need to support this instead of changing the tests.

**Comment by sricker-nvidia** [comment]
> I think we should definitely allow global vars with system semantics to co-exist with entry point parameters.

@csyonghe  I thought we already decided against this above (https://github.com/shader-slang/slang/pull/6998#discussion_r2075807271) and decided to just disallow all parameters. I can change it back and use something like:

````
for (auto param : entryPointFunc->getParams())
{
    if (!param->findDecoration<IRSemanticDecoration>())
        // error
````

but this is allowing mixing entry point params with semantics, rather than allowing global vars with system semantics to co-exist with all entry point parameters.

**Comment by csyonghe** [comment]
I think what we have decided against previously is the opposite: we decided not to support mixing global varying in/out parameters with any entrypoint parameter.

However, I think we should have no problems supporting mixing global system value parameters with entrypoint parameters.

**Comment by sricker-nvidia** [comment]
I pushed an update commit that attempts to support mixing global system value parameters with entrypoint parameters. It's still a WAR as I need to do more testing on it and fix some issues, but I wanted to make it available to see if this was at least in the correct direction.

**Comment by sricker-nvidia** [comment]
Change is ready for re-review. Made updates to allow mixing global GLSL style global in/out vars with entry point params when either all global in/out vars or all entry point params have a system value binding semantic.

With these updates I no longer have to update any of the existing tests.

File: source/slang/slang-ir-translate-global-varying-var.cpp [hunk 5]
```diff
@@ -65,6 +65,47 @@ struct GlobalVarTranslationContext
...
+            // binding semantic. It may be possible to mix these in general, but binding offsets
+            // and other items would likely need to be recalculated.
+            bool allGlobalsHaveSVBindingSemantic = true;
+            bool allEntryParamsHaveSVBindingSemantic = true;
+
+            for (auto param : entryPointFunc->getParams())
+            {
+
+                auto semanticDecor = param->findDecoration<IRSemanticDecoration>();
+                if (semanticDecor &&
```
csyonghe, line 79: I think we should be able to do:
```
        if (auto layout = getVarLayout(param))
        {
            if (auto systemValueAttr = layout->findAttr<IRSystemValueSemanticAttr>())
            {
                  // we are a system value.
```

source/slang/slang-ir-translate-global-varying-var.cpp
```diff
@@ -65,6 +65,47 @@ struct GlobalVarTranslationContext
...
+                    semanticDecor->getSemanticName().startsWithCaseInsensitive(toSlice("sv_")))
+                    continue;
+                else
```
csyonghe, line 82: style: we can remove couple lines of code if we do:
```
if (!...)
{
    allEntryParamsHaveSVBindingSemantic = false;
    break;
}

**Review by csyonghe — APPROVED** [review]
Looks good, would be nice if we can put in some more comments explaining the layout updating logic.

**Comment by sricker-nvidia** [comment]
I started to add additional comments based on review feedback and in doing so thought I might be able to get mixing glsl global ins and entry point in vars working in general.

I made some updates that recalculate offsets for the global in vars when added to the recreated entry point layout and that seemed to get layout locations in the output shader lined up correctly. I then found what seemed to be some additional issues with how we were handling entry point uniform parameters and attempted to fix those as well.

With the latest changes I can run the repro shader from the issue and no longer see a seg fault or assert and the shader output looks generally correct.

I still need to improve comments for the new code and look into adding additional testing.

File: source/slang/slang-ir-glsl-legalize.cpp [hunk 7]
```diff
@@ -3575,28 +3575,40 @@ void legalizeEntryPointParameterForGLSL(
...
+                                        realGlobalVar = typeAdapterVal->val.irValue;
+                                    }
                                 }
+                                break;
                             }
-                            break;
                         }
                     }
                 }
+                else if (globalValue.flavor == ScalarizedVal::Flavor::address)
```
jkwak-work, line 3606: It might be better to use switch
```
switch (globalValue.flavor)
{
    case ScalarizedVal::Flavor::tuple:
        break;
    case ScalarizedVal::Flavor::address:
        break;
    default: continue;
}
```

jkwak-work (dismissed):
Looks good to me

sricker-nvidia:
Not sure what happened, but CI is now failing another two metal tests after rebasing, despite passing these tests before. Added "-v" again into the CI file to get print outs for metal to get an idea of what's happening.

sricker-nvidia:
hmm, seems like macos debug CI passed now? I guess there's still intermittent issues with debug?

jkwak-work (approved):

Issue #6968, opened 2025-04-30
Emit missing HLSL functions when using DXC1.6


jkwak-work:
[PR 6310](https://github.com/shader-slang/slang/pull/6310) turned out to be a breaking change when using an older version of DXC 1.6.
It was to resolve the following issue,
- https://github.com/shader-slang/slang/issues/6307

The problem is that slangc started using new functions that are not available with DXC 1.6 and below.
Those functions are `and`, `or` and `select` for vector arguments.

We can add a new command-line argument `-hlsl2018` and emit a few overloading for the missing functions.
```
float select(float a, float b, bool condition)
{
    return condition ? b : a;
}

int select(int a, int b, bool condition)
{
    return condition ? b : a;
}

float2 select(float2 a, float2 b, bool2 condition)
{
    return float2(condition.x ? b.x : a.x, condition.y ? b.y : a.y);
}

float3 select(float3 a, float3 b, bool3 condition)
{
    return float3(
        condition.x ? b.x : a.x,
        condition.y ? b.y : a.y,
        condition.z ? b.z : a.z);
}

float4 select(float4 a, float4 b, bool4 condition)
{
    return float4(
        condition.x ? b.x : a.x,
        condition.y ? b.y : a.y,
        condition.z ? b.z : a.z,
        condition.w ? b.w : a.w);
}

/// Same for `and` and `or`
```

csyonghe:
Instead of adding another top level compiler option, I suggest we add another capability atom `hlsl2018`, so users would opt-in to this behavior with `-capability hlsl2018`.

jkwak-work:
Yes, I will do that. That sounds like a clean way.

PR #7003, opened 2025-05-05
Add a new capability hlsl_2018 that avoid using select/and/or

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6968


jarcherNV (approved):
Thanks for doing this. Do you need me to test with my local OV build?

csyonghe (approved):

Issue #6974, opened 2025-05-01
CI autodiff tests are intermittently failing


jarcherNV:
It looks like some of the autodiff tests are failing intermittently on the CI.
Example: https://github.com/shader-slang/slang/actions/runs/14766459010/job/41460568277

FAILED test: 'tests/autodiff/bool-return-val-bwd.slang.1 (vk)'

This is causing issues due to people needing to re-run the CI to merge their PRs.

jkwak-work:
The failed test in the issue description is a test at an early stage and there aren't any tests expected to fail before it.
Note that they are mostly ran in alphabetic order and "autodiff/b" is at a pretty early step.

jkwak-work:
I am curious about the intermittency of "autodiff" tests.
And I started CI tests that repeats autodiff tests 100 times.
https://github.com/shader-slang/slang/pull/6908

Let's see how often they fail...

jkwak-work:
I was able to reproduce the following error on my local machine,
```
$ build/Release/bin/slang-test.exe tests/autodiff/auto-differential-type -api dx11 -use-test-server -server-count 8
Supported backends: fxc dxc glslang spirv-dis clang visualstudio genericcpp nvrtc llvm spirv-opt metal tint
Check vk,vulkan: Supported
Check dx12,d3d12: Supported
Check dx11,d3d11: Supported
Check cuda: Supported
Check wgpu,webgpu: Supported
passed test: 'tests/autodiff/auto-differential-type-generic.slang (dx11)' 
ignored test: 'tests/autodiff/auto-differential-type-generic.slang.1 (vk)' 
ignored test: 'tests/autodiff/auto-differential-type.slang (dx11)' 
error: ignored test: 'tests/autodiff/auto-differential-type.slang.1 (vk)' 
Unable to read render-test output: result code = 0
standard error = {
}
standard output = {
}
Retrying 1 failed tests...
FAILED test: 'tests/autodiff/auto-differential-type.slang (dx11)' 
error: Unable to read render-test output: result code = 0
standard error = {
}
standard output = {
}
50% of tests passed (1/2), 3 tests ignored
failing tests:
---
tests/autodiff/auto-differential-type.slang (dx11)
---
```
The repro rate appears to be about 10% and it requires "-use-test-server" option with at least 2 server count.
The issue had been seen only with dx11 so far.

jkwak-work:
I was able to do git-bisect with the repro method above; repeated 100 times for each commit.
[...]

PR #7006, opened 2025-05-05
Fix the intermittent failures of tests/autodiff/auto-differential-type

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6974

The use of `_wfopen_s()` was incorrect in a way the the result value had to be checked.

However, even with a proper handling of the failed case, the repro-rate was same.

The issue was reproduced at 5%~20% rate with the following commands,
  build/Release/bin/slang-test.exe tests/autodiff/auto-differential-type -api dx11 -use-test-server -server-count 8

_wfsopen appears to be a good alternative.
The intermittency failure is not observed with it; tried 1,000 times and no repro.


jarcherNV (approved):
LGTM

csyonghe (approved):

PR #7009, opened 2025-05-06
Destroy unused witness table after hoisting

### LINKED ISSUE #6933: Enable debug testing on linux and fix issues.

gtong-nv:
This PR turn on the debug tests in CI, which catches a validation error for the witness table using unseen inst.
The root cause is due to https://github.com/shader-slang/slang/commit/0c5eee we create a new witness table but not destroying the original one, and during the IR validation the old one is triggering errors.

Fixes #6933

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,7 @@ jobs:
             build-llvm: false
           # Self-hosted full gpu build
           - os: windows
+            config: release
             compiler: cl
             platform: x86_64
             test-category: full
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,6 @@ jobs:
             build-llvm: false
           # Self-hosted full gpu build
           - os: windows
-            config: release
             compiler: cl
             platform: x86_64
             test-category: full
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -175,7 +175,7 @@ function(set_default_compile_options target)
             ON
             # C++ standard
             CXX_STANDARD
-            17
+            20
             # pic
             POSITION_INDEPENDENT_CODE
             ON
--- a/docs/user-guide/02-conventional-features.md
+++ b/docs/user-guide/02-conventional-features.md
@@ -1065,7 +1065,7 @@ struct GenerateCtor1 : GenerateCtorInner1
 GenerateCtor1 val[2] = { { 3 }, { 2 } };
 ```
 
-In addition, Slang also provides compatibility support for C-style initializer lists with `struct`s. C-style initializer lists can use [Partial Initializer List's](#Partial-Initializer-List's) and [Flattened Array Initializer With Struct's](#Flattened-Array-Initializer-With-Struct)
+In addition, Slang also provides compatibility support for C-style initializer lists with `struct`s. C-style initializer lists can use [Partial Initializer List's](#Partial-Initializer-Lists) and [Flattened Array Initializer With Struct's](#Flattened-Array-Initializer-With-Structs)
 
 A struct is considered a C-style struct if:
 1. User never defines a custom constructor with **more than** 0 parameters
--- a/docs/user-guide/06-interfaces-generics.md
+++ b/docs/user-guide/06-interfaces-generics.md
@@ -806,7 +806,7 @@ void main()
 }
 
 ```
-See  [if-let syntax](convenience-features.html#if_let-syntax) for more details.
+See  [if-let syntax](03-convenience-features.html#if_let-syntax) for more details.
 
 
 Generic Interfaces
--- a/docs/user-guide/08-compiling.md
+++ b/docs/user-guide/08-compiling.md
@@ -861,7 +861,7 @@ program->link(linkedProgram.writeRef(), diagnosticBlob.writeRef());
 ```
 
 The linking step is also used to perform link-time specialization, which is a recommended approach for shader specialization
-compared to preprocessor based specialization. Please see [Link-time Specialization and Precompiled Modules](link-time-specialization) for more details.
+compared to preprocessor based specialization. Please see [Link-time Specialization and Precompiled Modules](10-link-time-specialization) for more details.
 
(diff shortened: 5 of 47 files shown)


gtong-nv:
The real issue for validation error "IR validation failed: def must come before use in same block" should have something to do with [Make IRWitnessTable HOISTABLE](https://github.com/shader-slang/slang/commit/0c5eee)  . Confirmed by reverting that commit. 

The error happens whenever we use `BackwardDifferentiable` with generics types.
Use `dynamic-dispatch-custom-bwd-derivative.slang` as an exmaple, 
 when we try to validate the witness table inst with
`validateIRInstOperand(IRValidateContext* context, IRInst* inst, IRUse* operandUse)`

operandUse is a StructType with name `s_bwd_prop_sqr_outter_Intermediates`, and it is not in `context->seenInsts` yet so the validation failed.
I noticed that that `StructType` (operand) and `WitnessTable`(inst) have the same `block` as the parent, whereas all the WitnessTables in passing tests have `module` as its parent - **should all WitnessTables set parent to `module` after hoisting?** 

I also noticed that for the failed witness tables are created with 
```
        builder.setInsertAfter(addMethod);
        auto diffTypeIsDiffWitness =
            builder.createWitnessTable(autodiffContext->differentiableInterfaceType, diffType);
```
and the builder has a `Before` mode in the `insertLoc` due to above. Does it mean the witness table is inserted to the list before the `StructType` (operand) ?

csyonghe:
It seems that in this case the struct type and the witness table are both inside a IRGeneric.

Does the struct type exist when the witness table is created?

gtong-nv:
> Does the struct type exist when the witness table is created?

Yes, the struct type has smaller UI and is created before the witness table.


> It seems that in this case the struct type and the witness table are both inside a IRGeneric

Looks like both have the following structs
IRModule -> IRGeneric  -> IRBlock ->IRWitnessTable / IRStructType

What concerns me is that
m_insertLoc is `AtEnd` when the `struct type` is inserted to the list, but `Before` when we are call addGlobalValue() for the the witness table.
As a reulst, IRWitnessTable is inserted before the struct type -
image

csyonghe:
Is the last instruction of the block returning the witness table or returning the struct type?

csyonghe:
If it is a bug in creation of that witness table (inserting into the incorrect location), why does this only surface when the hoisting PR is merged? I suspect it is the hoisting logic moving the witness table into the wrong location.

gtong-nv:
I think it's returnning the struct type. Dumped all children in the block when the valiation error happens. 

> let  %1 : BasicBlock    = block
> 
> child op: param (10096)
> [TypeConstraintDecoration(%IInterface)]
> [nameHint("T")]
> let  %T : type_t        = param
> 
> child op: param (10099)
> let  %1 : witness_table_t(%IInterface)  = param
> 
> child op: specialize (11065)
> 
> child op: specialize (11033)
> 
> child op: specialize (11031)
> 
> child op: specialize (11000)
> 
> child op: specialize (10987)
> 
> child op: specialize (10947)
> 
> child op: witness_table (10925)
> witness_table %1        : witness_table_t(%IDifferentiable)(specialize(%2, %T, %3))
> {
>         witness_table_entry(%4,specialize(%2, %T, %3))
> 
>         witness_table_entry(%5,%1)
> 
>         witness_table_entry(%6,specialize(%7, %T, %3))
> 
>         witness_table_entry(%8,specialize(%9, %T, %3))
> 
> }
> 
> 
> child op: witness_table (10927)
> witness_table %1        : witness_table_t(%IDifferentiable)(%2)
> {
>         witness_table_entry(%3,specialize(%4, %T, %5))
> 
>         witness_table_entry(%6,%7)
> 
>         witness_table_entry(%8,specialize(%9, %T, %5))
> 
>         witness_table_entry(%10,specialize(%11, %T, %5))
> 
> }
> 
> 
> child op: struct (10111)
> struct %1       : Type;
> 
> 
> child op: return_val (10112)
> return_val(%1)

csyonghe:
Is this witness table referenced by anything? If not it shouldn’t even be inserted.

gtong-nv:
>If it is a bug in creation of that witness table (inserting into the incorrect location), why does this only surface when the hoisting PR is merged? I suspect it is the hoisting logic moving the witness table into the wrong location.

I reverted the hoisting PR, and looks like the insertion order is the same. But when it comes to validating, the same witness table is using another witness table as the operand, instead of the unseen struct type. I guess it's the hoisting and dedpulication logic that surface the issue.

gtong-nv:
Debugging a little further, in the TOT with the hoisting logic, we have 2 witness tables in the same block. The first one is using IR_Specialize as operand, but the 2nd one is an unseen struct type as its opearnd.

> ======= validating the block ===== 
> -------------- child 0: param (10096)  
> -------------- child 1: param (10099)  
> -------------- child 2: specialize (11065)  
> -------------- child 3: specialize (11033)  
> -------------- child 4: specialize (11031)  
> -------------- child 5: specialize (11000)  
> -------------- child 6: specialize (10987)  
> -------------- child 7: specialize (10947)  
> -------------- child 8: witness_table (10925)  
> -------------- child 9: witness_table (10927)  
> -------------- child 10: struct (10111)  
> -------------- child 11: return_val (10112) 


In the old version, the 2 witness tables are still created, but only the first one is being validated  ( not inserted to the list or removed?) 

> ======= validating the block =====
> -------------- child 0: param (10096)
> -------------- child 1: param (10099)
> -------------- child 2: specialize (11059)
> -------------- child 3: specialize (11033)
> -------------- child 4: specialize (11031)
> -------------- child 5: specialize (11000)
> -------------- child 6: specialize (10987)
> -------------- child 7: specialize (10947)
> -------------- child 8: witness_table (10925)
> -------------- child 9: struct (10111)
> -------------- child 10: return_val (10112)
> 


I'm trying to find out the logic why we have 2 witness tables after the hositing PR

csyonghe (dismissed):

jkwak-work (approved):
Great job fixing this.

Looks good to me.

gtong-nv:
I'm not going to enable the debug test in this PR as there are still some other failures in the CI. I can't reproduce them locally, and I will try to find another way to debug it.

jkwak-work (approved):
Looks good to me.

Let's fix the other debug bugs later.

PR #7010, opened 2025-05-06
Address structured buffer `GetDimensions` issues for WGSL, GLSL and SPIRV

### LINKED ISSUE #6842: Slang exits without error and without generating wgls output (2025.6.3)
test.slang:
```
[[vk::binding(0)]]
StructuredBuffer<float3> Scene : register(t1);

[numthreads(8, 8, 1)]
void main(uint3 DTid : SV_DispatchThreadID)
{
    uint2 numVertsStride;
    Scene.GetDimensions(numVertsStride.x, numVertsStride.y);
}
```

Command:
`.\slangc .\test.slang -target wgsl -entry main -stage compute -o test.wgsl`

The command completes without any reported error, but there is no test.wgsl output.

fairywreath:
There are a few issues for structured buffer `GetDimensions`:
- Stride value relies from `IRSizeAndAlignmentDecoration` but this decoration is not always created. This causes the compiler to crash when targetting WGSL and silently giving incorrect stride results when targetting GLSL.
- Stride value is always zero when targetting SPIRV because this decoration is not checked properly.

This PR fixes these by always creating a `IRSizeAndAlignmentDecoration`  for structured buffers and properly accessing the size and alignment decoration when targetting SPIRV.

Fixes #6842.


csyonghe (dismissed):
Looks good to me!

tests/cross-compile/get-dimensions-stride.slang
```diff
@@ -0,0 +1,58 @@
+//TEST(compute):COMPARE_COMPUTE_EX(filecheck-buffer=CHECK):-vk -compute -shaderobj -emit-spirv-directly
```
csyonghe, line 1: I think you just need one -vk test without `-emit-spirv-directly`, and it will be run twice with both spirv and glsl backends.

fairywreath:
I am not sure why when I used `half` inside a test shader the test fails on WGPU on Windows CI but passes locally on my machine - I fixed it to only use 32 bit types.

csyonghe (approved):

Issue #7017, opened 2025-05-06
bitcast lowering issue: input has different width from result type


kaizhangNV:
In bitcast lowering, we emit invalid bit_cast IR inst where the input has different width from result type, e.g.:

```
%1: load ushort xxx
%2: bitcast Uint %1
```

This is invalid, but this issue is not exposed in bit_cast intrinsic because this IR is generated by lowerBitCast pass and we have typecheck system to capture this if this is user code.

PR #7018, opened 2025-05-06
bitcast require the input has same width with result type

kaizhangNV:
Fix the issue in lowering bitcast that it requires input has the same width with result type.

Close https://github.com/shader-slang/slang/issues/7017.

### DIFF
--- a/source/slang/slang-ir-extract-value-from-type.cpp
+++ b/source/slang/slang-ir-extract-value-from-type.cpp
@@ -190,7 +190,12 @@ IRInst* extractByteAtOffset(
     {
         uintType = builder.getUInt64Type();
     }
-    auto resultValue = builder.emitCast(uintType, leaf.leafValue);
+
+    auto resultValue = leaf.leafValue;
+    if (!as<IRPtrTypeBase>(leaf.leafValue->getDataType()))
+    {
+        resultValue = builder.emitCast(uintType, leaf.leafValue);
+    }
     resultValue = builder.emitBitCast(uintType, resultValue);
     if (leaf.offsetInValue != 0)
     {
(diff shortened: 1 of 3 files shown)


jkwak-work:
It looks like CI tests are failing.
I vaguely remember something like this last year.
We required the size to be same but it caused a problem so we relaxed it
I am not sure sure if it was the same thing as this one.

kaizhangNV:
I see, this is a problem that we either solved at spirv emit stage, or at IR stage.
Really depends how we define the rule of this IR.

kaizhangNV:
The error seems like we don't support int cast from pointer to integer, so I should rule that out from the integer cast.

source/slang/slang-ir-extract-value-from-type.cpp
```diff
@@ -190,7 +190,13 @@ IRInst* extractByteAtOffset(
     {
         uintType = builder.getUInt64Type();
     }
-    auto resultValue = builder.emitBitCast(uintType, leaf.leafValue);
+
+    auto resultValue = leaf.leafValue;
```
csyonghe, line 194: what if leafValue is a float? Cast here would be wrong.

kaizhangNV: Yes, I'm aware of that.
Then do you think it's fine to just exclude those types that are not able to int cast here?

csyonghe: I think we want to find the corresponding `uintN_t` type for each scalar type, and bit cast into that.

csyonghe: There is `getSameSizeUIntType` that you can use.

kaizhangNV: I come up with another solution, can you take a look?

csyonghe (approved):

source/slang/slang-ir-extract-value-from-type.cpp
csyonghe: yeah, that will work as well.

kaizhangNV: Basically what I'm trying to do is to cast it to a intermediate type first, then bitcast to either 4-byte or 8-byte integers.

kaizhangNV: Thanks.

Issue #6875, opened 2025-04-21
[CM2] Add new functions: Reduce and Transpose


jkwak-work:
A subtask from
- https://github.com/shader-slang/slang/issues/6398

Cooperative matrix 2 spec introduced three new functions:
- MatReduce
- MatPerElement
- MatTranspose

Note that the GLSL syntax for coopMatPerElementNV allows the users to define the behavior with a function pointer. Slang will need to have something equivalent.

jkwak-work:
I am going to split up the implementation of `MapElement` into a new issue, because it has a different level of complexity.

Issue #6907, opened 2025-04-25
[CP2] Support VulkanMemoryModel for cooperative vectors/matrices


gtong-nv:
Running  `tests/cooperative-matrix/add.slang` with Spirv validation (SLANG_RUN_SPIRV_VALIDATION=1), we saw some validation errors

```
If the CooperativeVectorNV capability is declared then the VulkanMemoryModel capability must be declared as well.
```

The reason is that we haven't added VulkanMemoryModel  support.

I tried to add `SpvCapabilityVulkanMemoryModel` for case **kIROp_CoopMatrixType** and case **kIROp_CoopVectorType**, and `emitOpMemoryModel SpvMemoryModelVulkan` for this cap with below.
```
        if (m_capabilities.contains(SpvCapabilityVulkanMemoryModel) || m_capabilities.contains(SpvCapabilityCooperativeMatrixKHR)) {
            emitOpMemoryModel(
                getSection(SpvLogicalSectionID::MemoryModel),
                nullptr,
                m_addressingMode,
                SpvMemoryModelVulkan);
        }
```

>  the proper implementation is to scan the entire module for things that needs VulkanMemoryModel first, and if we determined that we need VulkanMemoryModel (or if the user tells us to do so with a -profile/-capability setting), then we need to emit many things differently.
https://github.com/shader-slang/slang/pull/6893#discussion_r2059355474

jkwak-work:
https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_vulkan_memory_model.html

jkwak-work:
The description is little confusing that the validation error was observed with coop"mat" test but the error message says coop"vec".

jkwak-work:
A related document is https://docs.vulkan.org/spec/latest/appendices/memorymodel.html

jkwak-work:
I found a section about cooperative matrix from the [vulkan document](https://docs.vulkan.org/spec/latest/appendices/memorymodel.html#memory-model-cooperative-matrix):

> Cooperative Matrix Memory Access
> ---------------------------------------
> For each dynamic instance of a cooperative matrix load instruction (OpCooperativeMatrixLoadKHR , OpCooperativeMatrixLoadNV , OpCooperativeMatrixLoadTensorNV ), some implementation-dependent invocation(s) within the instance of the matrix’s scope perform a non-atomic load from each memory location that is defined to be accessed by the instruction.
>
[...]

PR #7027, opened 2025-05-07
Add new coopmat2 functions: Reduce and Transpose

jkwak-work:
This commit adds three new functions for CoopMat as described in the proposal document,
 - https://github.com/shader-slang/spec/pull/12

The new functions are:
- CoopMat<T,S,M,N,R>::Transpose
- CoopMat<T,S,M,N,R>::ReduceRow
- CoopMat<T,S,M,N,R>::ReduceColumn
- CoopMat<T,S,M,N,R>::ReduceRowAndColumn
- CoopMat<T,S,M,N,R>::Reduce2x2
- CoopMat<T,S,M,N,R>::MapElement *EDIT: this will be implemented in another PR
- Tuple<CoopMat<T,S,M,N,R>,...>::MapElement *EDIT: this will be implemented in another PR

In order to allow lambda and IFunc to be used with MapElement, the following helper functions are populated,
```
static T __functypeToIFunc_1(uint32_t row, uint32_t col, T v0, IFunc<T, uint32_t, uint32_t, T> func)
{ return func(row, col, v0); }
static T __functypeToIFunc_2(uint32_t row, uint32_t col, T v0, IFunc<T, uint32_t, uint32_t, T, T> func, T v1)
{ return func(row, col, v0, v1); }
static T __functypeToIFunc_3(uint32_t row, uint32_t col, T v0, IFunc<T, uint32_t, uint32_t, T, T, T> func, T v1, T v2)
{ return func(row, col, v0, v1, v2); }
static T __functypeToIFunc_4(uint32_t row, uint32_t col, T v0, IFunc<T, uint32_t, uint32_t, T, T, T, T> func, T v1, T v2, T v3)
{ return func(row, col, v0, v1, v2, v3); }
...
```
Note that they cannot have a same name as overloading because the SPIR-V function, `OpCooperativeMatrixPerElementOpNV`, cannot differentiate them by their parameters.
It has to use one explicitly based on the number of the argument like the following,
```
result:$$CoopMat<T, S, M, N, R> = OpCooperativeMatrixPerElementOpNV $this $This.__functypeToIFunc_1 $mapOp;
```
Also note that the ordering of the parameters is little confusing.
The first parameters to the function pointer has to come before the function pointer and the rest should follow after the function pointer.

Closes https://github.com/shader-slang/slang/issues/6875


csyonghe:
I would make `__functypeToIFunc_1` generic instead of using `IFunc` directly:
```
static T __functypeToIFunc_1<T, F:IFunc<T, u32, u32, T>>(uint32_t row, uint32_t col, T v0, F func)
{ return func(row, col, v0); }
```

But I think it is probably better to use a dedicated pass to generate this wrapper function instead of defining it in the stdlib.

source/slang/hlsl.meta.slang
```diff
@@ -22505,68 +22550,189 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
...
+        return spirv_asm
+        {
+            OpCapability CooperativeMatrixReductionsNV;
+            OpExtension "SPV_NV_cooperative_matrix2";
+            result:$$CoopMat<T, S, M / 2, N / 2, CoopMatMatrixUse.MatrixAccumulator> = OpCooperativeMatrixReduceNV $this 0x4 $combineOp;
+        };
+    }
+
+    [require(cooperative_matrix_map_element)]
+    This MapElement(functype(uint32_t, uint32_t, T)->T mapOp)
```
csyonghe, line 22668: Please don't do the expansion here. Map this intrinsic to an instruction and implement a dedicated pass for it.

source/slang/hlsl.meta.slang
```diff
@@ -22505,68 +22550,189 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
...
     }
 
+    //
+    // CoopMat operations
+    //
+
+    /// Returns the number of elements in the underlying storage of the cooperative matrix type.
+    [ForceInline]
+    [require(cooperative_matrix)]
+    static uint GetLength()
```
csyonghe, line 22585: Can this be different from `M*N`?

source/slang/hlsl.meta.slang
```diff
@@ -22819,77 +22992,178 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
 
 [ForceInline]
 [require(cooperative_matrix)]
-CoopMat<T, S, M, N, R> coopMatLoad<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, let N : int, let R : CoopMatMatrixUse>(ByteAddressBuffer buffer, uint element, uint stride, constexpr CoopMatMatrixLayout matrixLayout)
-{
-    return CoopMat<T, S, M, N, R>.load(buffer, element, stride, matrixLayout);
+__generic<
```
csyonghe, line 22995: why using `__generic` syntax? The trailing `<>` syntax should be preferred.

source/slang/hlsl.meta.slang
```diff
@@ -22898,63 +23172,149 @@ CoopMat<T, S, M, N, R> operator *(CoopMat<T, S, M, N, R> lhs, const T rhs)
...
-    MatrixBSigned = 0x2,
-    MatrixCSigned = 0x4,
-    MatrixResultSigned = 0x8,
-    SaturatingAccumulation = 0x10,
-};
 
-//
-// Cooperative Matrix multiply accumulate
-//
+${{{{
```
csyonghe, line 23239: Don't do expansion. Use dedicated pass to implement wrapper generation/transformation.

source/slang/slang-emit-spirv.cpp
```diff
@@ -1685,6 +1686,9 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
             }
         case kIROp_CoopMatrixType:
             {
+                m_memoryModel = SpvMemoryModelVulkanKHR;
```
csyonghe, line 1689: memory model needs to be determined earlier, during `determineSPIRVVersion`, so other emit logic can adapt to it. Note that use of vulkan memory model means that all globallycoherent resource delcarations and load/store/atomics will need to be emitted differently.

source/slang/hlsl.meta.slang
jkwak-work: Yes, it is explain in `tests/cooperative-matrix/length.slang`.

source/slang/hlsl.meta.slang
jkwak-work: I can change them back if it is the preferred way.

source/slang/slang-emit-spirv.cpp
jkwak-work: Yes, I have another issue for `globallycoherent`.
https://github.com/shader-slang/slang/issues/6907#issuecomment-2854715554

I didn't know we had `determineSPIRVVersion`. I will work on it.

source/slang/hlsl.meta.slang
```diff
@@ -22442,21 +22478,24 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
...
+    [require(cooperative_matrix)]
     __init(This x)
     {
         this = x;
     }
 
     // Required for `IArithmetic`.
-    [OverloadRank(-10)]
     [ForceInline]
+    [OverloadRank(-10)]
```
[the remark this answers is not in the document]
jkwak-work: Well.. It was there before and CoopVec also has it.
I am not sure why we have it.
I can take it down and see what happens.

source/slang/hlsl.meta.slang
jkwak-work: I was having multiple problems when I tried to use `expand` and `each` to void the expansion.

Here is one of the cases where I couldn't figure how to expand inside of `spriv_asm` block,
```
extension<
    each Ts : __BuiltinArithmeticType,
    let S : MemoryScope,
    let M : int,
    let N : int,
    let R : CoopMatMatrixUse
> Tuple<expand CoopMat<each Ts, S, M, N, R>>
{
    CoopMat<TR, S, M, N, R> MapElement<
        TR : __BuiltinArithmeticType
    >(functype(uint32_t, uint32_t, expand each Ts)->TR mapOp)
    {
        return spirv_asm
        {
            OpCapability CooperativeMatrixPerElementOperationsNV;
            OpExtension "SPV_NV_cooperative_matrix2";
            result:$$CoopMat<TR, S, M, N, R> = OpCooperativeMatrixPerElementOpNV $this._0 $mapOp expand each $this;
        };
    }
};
```
Is there a way to expand inside of `spirv_asm`?

jkwak-work: > Don't do expansion. Use dedicated pass to implement wrapper generation/transformation.

Oh, You meant that I should have __intrinsic_op for the implementation.
I see. I will do that.

source/slang/hlsl.meta.slang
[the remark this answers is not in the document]
jkwak-work: Without this line, I get an error like the following,
```
tests/cooperative-matrix/copyFrom.slang(13): note 99999: an internal error threw an exception while working on code near this location
    var result = CoopMat<int32_t, MemoryScope.Subgroup, 16, 16, CoopMatMatrixUse::MatrixAccumulator>(0);
```
I am going to keep it for now.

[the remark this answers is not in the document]
csyonghe: this feels like a legit problem to dig into.

jkwak-work:
This is ready for another review.

The vulkan memory model is handled in another PR more properly,
- https://github.com/shader-slang/slang/pull/7057

And I removed the implementation related to `MapElement` from this PR.
I will implement it in another PR.

jkwak-work:
The implementation for MapElement will be tracked in a new issue separately. 
- https://github.com/shader-slang/slang/issues/7058

source/slang/hlsl.meta.slang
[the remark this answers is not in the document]
jkwak-work: I will create a new issue later with a proper repro shader.

source/slang/hlsl.meta.slang
```diff
@@ -22505,25 +22552,13 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
     [__NoSideEffect]
     Ref<T> __indexRef(int index);
 
+    /// Returns the count as an integer value.
```
csyonghe, line 22555: I feel like we need to have more detailed documentation here explaining the meaning of `count`,
since it doesn't mean the total number of elements in the matrix, instead it refers to the number
of elements that should be processed by a single thread/lane.

source/slang/hlsl.meta.slang
```diff
@@ -22542,31 +22577,132 @@ struct CoopMat<T : __BuiltinArithmeticType, let S : CoopMatScope, let M : int, l
...
+        let matrixLayout : CoopMatMatrixLayout
+    >(RWByteAddressBuffer buffer, uint element, uint stride)
+    {
+        __Store(__getEquivalentStructuredBuffer<T>(buffer), element, stride, matrixLayout);
     }
 
-    [ForceInline]
     [require(cooperative_matrix)]
-    void store(RWStructuredBuffer<T> buffer, uint element, uint stride, constexpr CoopMatMatrixLayout matrixLayout)
+    void __Store(RWStructuredBuffer<T> buffer, uint element, uint stride, constexpr CoopMatMatrixLayout matrixLayout)
```
csyonghe, line 22705: We recently started supporting visibility modifiers in core module, so we can put `internal` here to make it invisible to user code.

csyonghe (approved):

csyonghe:
We still need to investigate the [OverloadRank(-1)] thing, and let's move things to `linalg` namespace.

source/slang/hlsl.meta.slang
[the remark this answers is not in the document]
jkwak-work: A issue tracking this is created
- https://github.com/shader-slang/slang/issues/7120

jkwak-work:
All of comments are address on the last push.
I will merge as soon as CI goes green.

PR #7039, opened 2025-05-08
Fix broken links in Slang docs

aidanfnv:
This change fixes some broken links/tags in the Slang docs that I found after running a link checker tool on the readthedocs site.
Some of these were broken on GitHub Pages as well, such as tags with apostrophes or docs that were moved.

The doc writer change adds an explicit .html extension to Parameter links, so that Sphinx/RTD does not treat it as a tag.

### DIFF
--- a/docs/user-guide/02-conventional-features.md
+++ b/docs/user-guide/02-conventional-features.md
@@ -1065,7 +1065,7 @@ struct GenerateCtor1 : GenerateCtorInner1
 GenerateCtor1 val[2] = { { 3 }, { 2 } };
 ```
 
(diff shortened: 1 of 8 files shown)


jkwak-work (approved):
Looks good to me.

I was wondering if we can have a CI to prevent regressions for the broken links?
If there are tool to check the links, it might be possible, isn't it?

csyonghe (approved):

aidanfnv:
A CI action in the website repo would be very useful for the link checking.
For checking these links I used this: https://github.com/filiph/linkcheck
You just pass it a URL as an argument, and it crawls the site. It does so much faster than the other link checkers that I tried, and it checked for broken anchors (which I keep calling tags by mistake) unlike the others.

PR #7046, opened 2025-05-09
Ensure to emit 32-bit type for bitfield operations for SPIRV

### LINKED ISSUE #7014: Always emit 32bit operand when emitting bit extracting instructions
Since spirv doesn't support operands other than 32-bit length in bit extracting instructions, slang has to ensure that all of those instructions casted to 32bit integer before emitting.

This restriction is not described in SPIRV spec, but in Standalone SPIR-V Validation: https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VUID-StandaloneSpirv-Base-04781.

There is a new feature to support this recently, so we should wait until the feature stable to apply this feature.
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7087/diffs

kaizhangNV:
Close #7014

### DIFF
--- a/source/slang/slang-ir-spirv-legalize.cpp
+++ b/source/slang/slang-ir-spirv-legalize.cpp
@@ -1471,6 +1471,39 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
         }
     }
 
+    // TODO: Currently SPIRV doesn't support non-32-bit integer types for bitfield extract and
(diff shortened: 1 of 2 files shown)


jkwak-work (approved):
Looks good to me.

What happens when the type is 64bit currently?
Do we emulate it?

kaizhangNV:
We don't support that, so I believe we will error that in IR stage.

Issue #7021, opened 2025-05-06
Move command line reference to readthedocs


aidanfnv:
The command line reference lacks title metadata, so if it gets included into the readthedocs site build, the page gets ignored.

So this issue has a 3-step solution:
1. In the slang repo, add a title to the command line reference by modifying the functions that generate it (in slang-options.cpp), and regenerate the doc
2. In the website repo, replace the external link in the readthedocs table of contents with the internal link to the markdown file
3. Also in the website repo, replace the link on the docs landing page to point to the Command Line Reference page on readthedocs

The title metadata should look like:
---
title: Command Line Reference
---
It will be invisible in the doc if viewed on GitHub, unless you view the raw markdown.

Once step 1 is done and the change is merged to slang, it will be picked up during the website repo's daily submodule update (or that action can be triggered manually, or it can be manually updated in step 2).

aidanfnv:
Strangely, Sphinx does not recognize that title format when the doc is not in the same directory as the file who's table of contents is referencing it. The alternative is to replace the "#" H1 marker with an RST-style underline "=".

aidanfnv:
The issue with the title is not actually due to the title itself, but rather a doc decoding issue.
For some reason this doc is encoded as UTF-16LE, whereas the rest of the docs are UTF-8. Maybe this is a quirk from me regenerating the doc on Windows?
In UTF-16LE, Sphinx will not recognize any titles in the doc because it's parsing all the docs as UTF-8.
It can be worked around by converting to UTF-8 in the readthedocs build, but that adds the strange quirk of either the metadata being rendered as `---title: Command Line Reference` if it is in the first line, or the existing H1 heading not being recognized as H1s. Adding a newline to the very start of the doc works around that.
The other H1s will need to be changed to H2s, and H2s changed to H3s, like was necessary for the other docs on ReadTheDocs.

PR #7048, opened 2025-05-09
Make Command Line Reference readthedocs compatible

aidanfnv:
For #7021 

This change modifies the code that generates the Command Line Reference doc to output H2 headings in place of H1 headings, and H3 in place of existing H2, so that readthedocs will not treat the additional H1 headings as titles.

This change also regenerates the Command Line Reference doc, as the current copy in the repo appears to be quite out-of-date. The existing copy is also encoded as UTF-16LE, whereas the other docs are all UTF-8. The regenerated doc is also UTF-8, and all I did to generate that was run `slangc.exe -help-style markdown -h > docs\command-line-slangc-reference.md 2>&1` after building slangc on Windows.

This change also adds GitHub actions workflows to check the contents of the doc, fail if a regenerated version needs to be checked in, and provide an option to regenerate it with a bot, all in a similar manner to User Guide TOC regeneration. The doc writer was producing different results from my local build until I changed how the writer sorts the shader stages. In the action, the order of `pixel` and `fragment` was reversed, despite the only difference from my local build being the OS.


.github/workflows/regenerate-cmdline-ref.yml
```diff
@@ -0,0 +1,110 @@
...
+          compiler: gcc
+          platform: x86_64
+          config: release
+          build-llvm: false
+
+      - name: Build Slang
+        id: build
+        run: |
+          cd pr-branch
+          cmake --preset default --fresh \
```
jkwak-work, line 45: Probably we can disable a few more to speed up like examples and wgpu

jkwak-work (approved):
Looks good to me

aidanfnv:
For https://github.com/shader-slang/slang/issues/7021

This change modifies the code that generates the Command Line Reference doc to output H2 headings in place of H1 headings, and H3 in place of existing H2, so that readthedocs will not treat the additional H1 headings as titles.

This change also regenerates the Command Line Reference doc, as the current copy in the repo appears to be quite out-of-date. The existing copy is also encoded as UTF-16LE, whereas the other docs are all UTF-8. The regenerated doc is also UTF-8, and all I did to generate that was run slangc.exe -help-style markdown -h > docs\command-line-slangc-reference.md 2>&1 after building slangc on Windows.

This change also adds GitHub actions workflows to check the contents of the doc, fail if a regenerated version needs to be checked in, and provide an option to regenerate it with a bot, all in a similar manner to User Guide TOC regeneration. The doc writer was producing different results from my local build until I changed how the writer sorts the shader stages. In the action, the order of pixel and fragment was reversed, despite the only difference from my local build being the OS.

PR #7051, opened 2025-05-09
Fix SPIRV unsigned to signed widening casts

### LINKED ISSUE #7035: uint8_t cast to int treats as int8_t
When converted to SPIR-V, casting a uint8_t to an int treats the bits as if they were initially for an int8_t. This creates the unexpected behavior that all the numbers above 127 become negative.

I first observed this with a very confusing out-of-bounds error that looked roughly like `colorPalette[colorPaletteIndex]` where `colorPaletteIndex` is a uint8_t. Upon further investigation I noticed, via a printf that looked like `printf("Example %d", (int)colorPaletteIndex);`, that the numbers were in fact becoming negative. This led me to believing that the out-of-bounds was likely an implicit cast to int as part of indexing (colorPalette is `const uint8_t4*`) and this seemed to be confirmed by that an explicit cast to a uint made my issue go away. I also did one extra test where I intentionally moved the colorPalette pointer 128 entries forward and confirmed that resolves the out-of-bounds issue, although obviously the colors that get selected are all from the wrong spots in the palette.

fairywreath:
Closes #7035.

Unsigned to signed casts are currently done using `OpSConvert` which is incorrect as no sign extension should be performed. `OpUConvert` is used instead.

### DIFF
--- a/source/slang/slang-emit-spirv.cpp
+++ b/source/slang/slang-emit-spirv.cpp
@@ -6599,11 +6599,11 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
         SLANG_ASSERT(!as<IRVectorType>(fromTypeV) == !as<IRVectorType>(toTypeV));
         const auto fromType = getVectorOrCoopMatrixElementType(fromTypeV);
         const auto toType = getVectorOrCoopMatrixElementType(toTypeV);
+        IRBuilder builder(inst);
 
         if (as<IRBoolType>(fromType))
         {
             // Cast from bool to int.
-            IRBuilder builder(inst);
             builder.setInsertBefore(inst);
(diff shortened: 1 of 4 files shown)


fivemtebex:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemhostingserver:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemserverhostings:
Impressive project structure! The architecture really demonstrates solid design principles.

csyonghe (approved):

Issue #7052, opened 2025-05-10
Spurious warning and compilation error when using constants inside functions (SPIR-V / Vulkan)


SaschaWillems:
With the following shader:

```slang
[shader("fragment")]
float4 fragmentMain(uint shadingRate : SV_ShadingRate)
{
    float4 color = float4(1.0, 1.0, 1.0, 1.0);

    const uint CONST_A = 0;
    const uint CONST_B = 1;

    switch (shadingRate) {
    case CONST_A:
        return color * 0.5;
    case CONST_B:
        return color * 0.7;
    }

    return col
``` 

I'm getting this odd warning:

```
warning 41016: use of uninitialized variable 'CONST_A'
    switch (shadingRate) {
    ^~~~~~
```

No error message on e.g. `CONST_B` or any other constant I've declared.

And the fragment shader also fail to compile.

If I replace `CONST_A` with `0` in that switch, everything is fine.

Also moving the constant outside of the function and declaring them as static works fine.

This happens using slangc 2025.7

juliusikkala:
I'm unable to reproduce the warning on 2025.7 or latest commit in master, but I can reproduce a compiler crash in a debug build. Could you share a complete example shader? I prepended it with the following:
```slang
struct VSOutput
{
    float3 Color;
    float2 UV;
};

Sampler2D samplerColorMap;
```

ArielG-NV:
Using the test shader provided I cannot repro the warnings (shader tested with is provided below):

```
[shader("fragment")]
float4 fragmentMain(uint shadingRate : SV_ShadingRate)
{
    float4 color = float4(1.0, 1.0, 1.0, 1.0);

    const uint CONST_A = 0;
    const uint CONST_B = 1;

    switch (shadingRate) {
    case CONST_A:
        return color * 0.5;
    case CONST_B:
        return color * 0.7;
    }

    return color;
}
```

ArielG-NV:
Is there potentially a different shader which produces the warning?

ArielG-NV:
If possible, @SaschaWillems please may you try to repro the warning using top-of-tree in-case this spurious warning issue has been resolved?

SaschaWillems:
Is there a way to download a nightly build? I tested with v2025.8.1 and still get that warning.

ArielG-NV:
Top of tree generally needs to be build from source. `v2025.9` should be released from what I see (although binaries are not yet provided).

SaschaWillems:
[...]

PR #7053, opened 2025-05-10
Fix local constants in switch cases

juliusikkala:
Partial fix for #7052. I couldn't reproduce the warning mentioned in the issue however, so this only fixes the crash I got with the example shader.

The problem is that with local constant values, `caseStmt->expr` lowers to a `kIROp_Var` instead of `kIROp_GlobalConstant`, and `getSimpleVal` would emit a load for it. However, since `SemanticsStmtVisitor::visitCaseStmt()` already finds a constant value, we can use that instead as a fallback.

### DIFF
--- a/source/slang/slang-lower-to-ir.cpp
+++ b/source/slang/slang-lower-to-ir.cpp
@@ -6814,20 +6814,11 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
             subBuilder.setInsertInto(info->initialBlock);
             subContext.irBuilder = &subBuilder;
 
-            auto caseValInfo = lowerRValueExpr(&subContext, caseStmt->expr);
-            if (caseValInfo.flavor != LoweredValInfo::Flavor::Simple)
-            {
-                // This expression would emit instructions other than a direct
-                // value, which we don't support yet here. If there was a
-                // constant value found during checking, we can use that value
-                // instead.
-                auto constVal = as<ConstantIntVal>(caseStmt->exprVal);
-                SLANG_ASSERT(constVal);
-                auto caseType = lowerType(context, constVal->getType());
-                caseValInfo = LoweredValInfo::simple(
+            auto constVal = as<ConstantIntVal>(caseStmt->exprVal);
+            SLANG_ASSERT(constVal);
+            auto caseType = lowerType(context, constVal->getType());
+            auto caseValInfo = LoweredValInfo::simple(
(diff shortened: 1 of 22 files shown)


source/slang/slang-lower-to-ir.cpp
```diff
@@ -6813,7 +6813,22 @@ struct StmtLoweringVisitor : StmtVisitor<StmtLoweringVisitor>
...
-            auto caseVal = getSimpleVal(context, lowerRValueExpr(&subContext, caseStmt->expr));
+
+            auto caseValInfo = lowerRValueExpr(&subContext, caseStmt->expr);
+            if (caseValInfo.flavor != LoweredValInfo::Flavor::Simple)
+            {
+                // This expression would emit instructions other than a direct
+                // value, which we don't support yet here. If there was a
+                // constant value found during checking, we can use that value
+                // instead.
+                auto constVal = as<ConstantIntVal>(caseStmt->exprVal);
```
csyonghe, line 6824: Should it always be the case that the const Val exists? If not this should have been a  semantic error, right?

juliusikkala: I thought there could be corner cases where a constant value may not be known at compile-time, like specialization constants in Vulkan. However, it appears that these don't work in switch cases yet anyway. So as it stands, I think `constVal` should always exist, with a diagnostic emitted from `slang-check-stmt.cpp` if not.

csyonghe: Yeah, I think we should just emit using the Val instead.

csyonghe: Just make sure when we check a switch case, we are folding the case value using CompileTime kind instead of LinkTime or SpecConst time.

juliusikkala: If we always just use `exprVal` instead of `expr`, wouldn't that mean that the names of the constant variables would no longer be preserved in the switch cases if emitting into a high-level language like GLSL? I'm personally completely fine with that though.

juliusikkala: It appears that the `CompileTime` kind is already being used (via `checkConstantIntVal`), so that's good.

csyonghe: Yeah we never preserve the names, so this change won’t make things worse.

csyonghe (approved):

Issue #6898, opened 2025-04-24
Using -emit-spirv-via-glsl in slang-test is a no-op


sricker-nvidia:
Using -emit-spirv-via-glsl in a subtest definition inside a slang-test slang shader like this:

//TEST(compute):COMPARE_COMPUTE_EX:-slang -vk -compute -emit-spirv-via-glsl

acts as a no-op instead of working as expected. We likely need to add a new flag and use it in tools/render-test/slang-support.cpp.

jkwak-work:
If this is a real issue, the priority of the task is pretty high. We don't want to skip the tests on ci.

sricker-nvidia:
Explored this a bit more. One option is to add a new flag like SLANG_TARGET_FLAG_EMIT_SPIRV_VIA_GLSL, but a note above the declaration of SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY reads, "This flag will be deprecated, use CompilerOption instead." so perhaps that is not the best approach.

I assume the comment is referring to using something like, "-Xslang... -emit-spirv-via-glsl -capability GL_EXT_shader_explicit_arithmetic_types_int64" to pass test options directly to the compiler, but that might mean modifying some old tests that still use -emit-spirv-via-glsl directly.

However, I have found that using either of the above reveals another issue in that test output does not appear to be captured for GLSL. This seems to occur using both filecheck and an expected output text file.

sricker-nvidia:
Fixed things up, but I'm now going through and dealing with a bunch of test failures resulting from -emit-spirv-via-glsl actually working as expected. Currently investigating an issue in tests/bugs/gh-4305.slang where one of the tests is somehow dropping a IRNumThreadsDecoration which leads to a layout like:

layout(local_size_x = 1, local_size_y = 1, local_size_z = 1)

instead of

layout(local_size_x = 2, local_size_y = 2, local_size_z = 1)

and leads to a glslang compile failure, "glslang:  tests/bugs/gh-4305.slang(38): error :  'derivative_group_quadsNV' : requires local_size_x and local_size_y to be multiple of two"

sricker-nvidia:
Fix has been checked in, closing issue.

PR #7054, opened 2025-05-10
Add a new option "-capability" to slang-test and render-test

jkwak-work:
This allows us to apply a certain capability while running slang-test.
You can use a command-line argument like,
```
slang-test.exe tests/compute/frem.slang -api vk -capability vk_mem_model
```
And you can also add an option for render-test like,
```
//TEST(compute):COMPARE_COMPUTE:-cpu -shaderobj -output-using-type -capability vk_mem_model
```

### DIFF
--- a/tools/render-test/options.cpp
+++ b/tools/render-test/options.cpp
@@ -170,6 +170,19 @@ static rhi::DeviceType _toRenderType(Slang::RenderApiType apiType)
         {
             outOptions.generateSPIRVDirectly = false;
         }
+        else if (argValue == "-capability" || argValue == "-capabilities")
+        {
+            String capabilities;
+            SLANG_RETURN_ON_FAIL(reader.expectArg(capabilities));
+
+            List<UnownedStringSlice> values;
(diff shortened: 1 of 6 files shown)


jkwak-work:
> Looks good to me. I wonder if we can add a test to prevent something like #6898.

Thanks for the review.
I think @sricker-nvidia  is working on it.

sricker-nvidia:
I'm a little late mentioning this, but I think a couple of these options like -emit-spirv-via-glsl and -capability can already be passed in via `-Xslang... <arg list> -X.` I ended up using something like, `-Xslang... -capability GL_KHR_shader_subgroup_rotate -X.` to fix a test failure in a different PR I was working on.

I opted to keep the -emit-spirv-via-glsl arg around because so many tests were already using it, but it might be good to print a warning when using it and suggest using `-Xslang -emit-spirv-via-glsl arg` instead (deprecating the arg).

I'm wondering if it would better to just remove the majority of slang-test args such that it becomes the expectation to use -Xslang, with any remaining args being specific to slang-test.

PR #7057, opened 2025-05-11
Support Vulkan memory model

jkwak-work:
Closes https://github.com/shader-slang/slang/issues/6907

Support Vulkan memory model

The user can explicitly use Vulkan memory model, or it will be
automatically used when cooperative-matrix is used.

When vulkan memory model is used, two keywords, "Coherent" and
Volatile", are not allowed.

There are many differences regarding atomic and texture but
this PR has changes limited to support `globallycoherent`
keyword. When variables with `globallycoherent` is used with `OpLoad`, it
will use additional options, `MakePointerAvailable|NonPrivatePointer`,
that will provide the same effect. For `OpStore`, it will use
`MakePointerVisible|NonPrivatePointer`.


kaizhangNV (commented):
So this PR for vulkan memory model part only implements the memory operations, not the image operations, right?

source/slang/slang-ir-spirv-legalize.cpp
```diff
@@ -1805,11 +1808,24 @@ struct SPIRVLegalizationContext : public SourceEmitterBase
...
+        {
+            m_sharedContext->m_memoryModel = SpvMemoryModelVulkan;
+        }
+
         if (m_sharedContext->m_spvVersion < 0x10300)
         {
             // Direct SPIRV backend does not support generating SPIRV before 1.3,
             // we will issue an error message here.
             m_sharedContext->m_sink->diagnose(SourceLoc(), Diagnostics::spirvVersionNotSupported);
+
```
csyonghe, line 1821: I wonder how one would actually trigger this case, because if coopmatrix capability requires spirv 1.6, then we can;'t end up with spirv version == 1.3 here, due to auto profile upgrade, right?

source/slang/slang-emit-spirv.cpp
```diff
@@ -1366,7 +1366,95 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
...
+        case SpvStorageClassStorageBuffer:
+            break;
+        default:
+            return false;
+        }
+
+        IRInst* baseObj = inst;
+        for (bool mayNeedToRetry = true; mayNeedToRetry && baseObj;)
+        {
+            switch (baseObj->getOp())
```
csyonghe, line 1415: We could reuse the existing `getRootAddr` to find the root of an address.

csyonghe (commented):
Let's also add a test for image load/store and image + atomics (InterlockedAdd etc.) and make sure we emit the correct spirv decorations.

jkwak-work:
The latest commit has my implementation for the texture read/write.
It is for the cases where RWTexture has `globallycoherent` where the write to the texture needs to be synchronized.

For atomic instructions, the currently implementation is to always use "device scope", but the vulkan memory model allows us to relax it down to a smaller scope. We currently don't have a way to specify the option for the atomic functions. 

Recently I added a new enum called "MemoryScope" in core.meta.slang. So we can use it when we support the scope option to atomic functions later. But that's not a part of this PR.

source/slang/slang-emit-spirv.cpp
```diff
@@ -1366,7 +1366,95 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
...
+
+            auto targetCaps = m_targetProgram->getTargetReq()->getTargetCaps();
+            if (targetCaps.implies(CapabilityAtom::spvVulkanMemoryModelDeviceScopeKHR))
+            {
+                requireSPIRVCapability(SpvCapabilityVulkanMemoryModelDeviceScope);
+            }
+        }
+    }
+
+    bool requiresCoherentLoadOrStore(IRInst* inst)
```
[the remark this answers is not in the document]
jkwak-work: I don't think it is worth bothering to cache from the performance perspective.

source/slang/slang-emit-spirv.cpp
```diff
@@ -6437,7 +6622,24 @@ struct SPIRVEmitContext : public SourceEmitterBase, public SPIRVEmitSharedContex
         }
         else
         {
-            return emitOpStore(parent, inst, inst->getPtr(), inst->getVal());
+            auto instPtr = inst->getPtr();
```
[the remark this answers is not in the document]
jkwak-work: I took this advice and made two functions with names similar to suggested.

csyonghe (dismissed):

csyonghe:
Can you update `/docs/user-guide/a2-01-spirv-target-specific.md` with a `Memory Model` section, documenting that by-default the SPIRV produced by Slang will use GLSL450 memory model. The user can opt-in to Vulkan Memory Model by specifying -capability vk_mem_model (also mention how to set via CompilerOptionName/API). If the code uses coop vector/matrix intrinsic, then the spirv will automatically be using vk_mem_model.

docs/user-guide/a2-01-spirv-target-specific.md
```diff
@@ -14,6 +14,21 @@ Experimental support for the older versions of SPIR-V
...
+    slang::CompilerOptionEntry entry;
+    entry.name = slang::CompilerOptionName::Capability;
+    entry.value.kind = slang::CompilerOptionValueKind::String;
+    entry.value.stringValue0 = "vk_mem_model";
+
+    slang::SessionDesc sessionDesc = {};
+    sessionDesc.compilerOptionEntries = &entry;
+    sessionDesc.compilerOptionEntryCount = 1;
+```
+
```
csyonghe, line 30: the Slang -> the Slang compiler

csyonghe (dismissed):

csyonghe (dismissed):

PR #7059, opened 2025-05-11
missing namespace std:: with nullptr_t (fails with clang) 

PierreEVEN:
Slang failed to compile using clang because of missing namespace std:: before nullptr_t

```
source/slang/slang-serialize.h:86:17: error: unknown type name 'nullptr_t'; did you mean 'std::nullptr_t'?
   86 |     void encode(nullptr_t) { encodeData(SerialBinary::kNullFourCC, nullptr, 0); }

source/slang/slang-serialize-ast.cpp:498:22: error: unknown type name 'nullptr_t'; did you mean 'std::nullptr_t'?
  498 |     void encodeValue(nullptr_t) { encoder->encode(nullptr); }
```

=> Added missing "std::" before nullptr_t : `nullptr_t` => `std::nullptr_t`

--------------------------------------------------
Context : 
- archlinux x86_64
- clang version 19.1.7
- cmake version 4.0.1


juliusikkala:
Good catch! It also appears that `slang-serialize.h` is expecting `stddef.h` to be transitively included through another system header, which is quite risky in case future compiler versions change what system headers include (at least GCC likes to do this often). Could you also add `#include <stddef.h>` while you're at it? Also, with the `.h` include, `nullptr_t` shouldn't need `std::` anymore, but it also doesn't hurt.

PierreEVEN:
> Good catch! It also appears that `slang-serialize.h` is expecting `stddef.h` to be transitively included through another system header, which is quite risky in case future compiler versions change what system headers include (at least GCC likes to do this often). Could you also add `#include <stddef.h>` while you're at it? Also, with the `.h` include, `nullptr_t` shouldn't need `std::` anymore, but it also doesn't hurt.

Thanks !
I've just checked include trees and I think `stddef.h` should already be included by `slang.h`
> `slang-serialize.h` > `slang-serializer-types.h` > `../core/slang-array-view.h` > `slang-common.h` > `slang.h`

https://github.com/shader-slang/slang/blob/b46c342f47b61119a0dc517ce6eb75eab3398504/include/slang.h#L504 
Also I saw nothing that defines `SLANG_NO_STDDEF` so it shouldn't be skipped.

juliusikkala (approved):

juliusikkala:
Yeah, you're right, I missed that. It's a bit interesting that Clang doesn't put `nullptr_t` in the global namespace even when included via `stddef.h`.

Thanks for contributing a fix!

Issue #7061, opened 2025-05-11
Slang doesn't allow function overloading by different sets of generic parameters


jkwak-work:
It might be an expected behavior depending on the language design.
But I was confused why my function overloading was not working properly until I figured that it has something to do with the generic parameters,

The following shader can reproduce the problem,
```
enum CoopMatMatrixLayout
{
    RowMajor = 0,
    ColumnMajor = 1,
};

int Load<let me : CoopMatMatrixLayout>(int inVal)
{
    return int(inVal % 10);
}

#if defined(REPRO)
int Load<let Dim : uint, let ClampMode : uint>(int inVal, int inVal2)
#else
int Load2<let Dim : uint, let ClampMode : uint>(int inVal, int inVal2)
#endif
{
    return Load<CoopMatMatrixLayout.RowMajor>(inVal);
}

[Shader("compute")]
[NumThreads(4, 1, 1)]
void computeMain(int3 dispatchThreadID : SV_DispatchThreadID)
{
    int tid = dispatchThreadID.x;

    int outVal = Load<CoopMatMatrixLayout.RowMajor>(tid);
}
```
When compiled with `REPRO` define, I observed the following error messages,
```
$ slangc.exe a.slang -target spirv-asm -entry computeMain -DREPRO
a.slang(18): error 30019: expected an expression of type 'uint', got 'CoopMatMatrixLayout'
    return Load<CoopMatMatrixLayout.RowMajor>(inVal);
                                    ^~~~~~~~
a.slang(18): note: explicit conversion from 'CoopMatMatrixLayout' to 'uint' is possible
a.slang(27): error 30019: expected an expression of type 'uint', got 'CoopMatMatrixLayout'
    int outVal = Load<CoopMatMatrixLayout.RowMajor>(tid);
                                          ^~~~~~~~
a.slang(27): note: explicit conversion from 'CoopMatMatrixLayout' to 'uint' is possible
```

jkwak-work:
It seems like the following workaround works, so the different number of generic arguments are allowed.

```
#if defined(WORK_AROUND)
enum CoopMatMatrixLayout : uint
#else
enum CoopMatMatrixLayout
#endif
{
    RowMajor = 0,
    ColumnMajor = 1,
};

int Load<let me : CoopMatMatrixLayout>(int inVal)
{
    return int(inVal % 10);
}

int Load<let Dim : uint, let ClampMode : uint>(int inVal, int inVal2)
{
[...]

PR #7064, opened 2025-05-12
Do not print errors in _coerce when "JustTrying".

jkwak-work:
While figuring out which generic-overload works best, `_coerce()` is printing errors and Slang compilation terminates prematurely.

When `TryCheckGenericOverloadCandidateTypes()` is calling `_coerce()` in "JustTrying" mode, the error messages should be snoozed.

The following logic shows the intention of how to silence the error messages, but the chain of `sink` was broken in the middle and `_coerce()` was using `getSink()` from the SemanticVisitor.

 val = ExtractGenericArgInteger(
   arg,
   getType(m_astBuilder, valParamRef),
   context.mode == OverloadResolveContext::Mode::JustTrying ? nullptr : getSink());

Closes https://github.com/shader-slang/slang/issues/7061

### DIFF
--- a/source/slang/slang-check-conversion.cpp
+++ b/source/slang/slang-check-conversion.cpp
@@ -137,6 +137,7 @@ bool SemanticsVisitor::_readValueFromInitializerList(
             outToExpr,
             firstInitExpr->type,
             firstInitExpr,
+            getSink(),
             nullptr);
     }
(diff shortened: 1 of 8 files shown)


sricker-nvidia (approved):
LGTM

PR #7067, opened 2025-05-12
Allow Link time constant array length sizing, warn on unsupported functionality

expipiplus1:


csyonghe:
Falcor test is breaking?

tests/bugs/link-time-constant-array-size-main.slang
```diff
@@ -0,0 +1,18 @@
+//TEST:COMPILE: tests/bugs/link-time-constant-array-size-lib.slang -o tests/bugs/link-time-constant-array-size-lib.slang-module
+//TEST:COMPILE: tests/bugs/link-time-constant-array-size-main.slang -o tests/bugs/link-time-constant-array-size-main.slang-module
+//TEST:COMPILE: tests/bugs/link-time-constant-array-size-main.slang-module tests/bugs/link-time-constant-array-size-lib.slang-module -target spirv -o out.spv
```
csyonghe, line 3: Can you turn this into //TEST:SIMPLE and file-check the SPIRV?

include/slang-deprecated.h
```diff
@@ -479,12 +479,17 @@ extern "C"
...
+    Use spReflectionType_GetSpecializedElementCount if the size is dependent on
+    a link time constant
     */
-    SLANG_API size_t spReflectionType_GetElementCount(
+    SLANG_API size_t spReflectionType_GetElementCount(SlangReflectionType* type);
+
+    /** The same as spReflectionType_GetElementCount except it takes into account specialization
+     * information from the given reflection info
+     */
+    SLANG_API size_t spReflectionType_GetSpecializedElementCount(
```
csyonghe, line 490: We should anvoid overloading in C API. Just name this spReflectionType_GetSpecializedElementCount2

tools/gfx-unit-test/link-time-constant-array-size-main.slang
```diff
@@ -0,0 +1,12 @@
+extern static const int N;
+
+struct S { int xs[N]; }
```
csyonghe, line 3: I think the test is too simplistic and missing important coverage.

We should call the reflection API explicitly and inspect the offset and sizes of relevant parameters/fields.


The type should contain additional fields following link time sized arrays, and we need to check the following field has the right offset when querying with a concrete program layout and had invalid offset (not 0) when querying without a concrete layout.


We should also test with types that contains resources, and nested link time sized arrays.

csyonghe:
We should also test declaring link time constants inside a namespace and a __included file.

include/slang-deprecated.h
expipiplus1: it's not overloaded, there's GetElementCount and GetSpecializedElementCount

expipiplus1:
We now warn instead of segfaulting on link time constant sized arrays, however the reflection API is still incomplete, we can tackle this in a follow up PR

kaizhangNV:
We should add more test cases here.
- reflection test `TEST:REFLECTION`
- thread local array: `static int a[link_time_const];`
- group shared array

kaizhangNV:
I'm OK with the change itself.

PR #7068, opened 2025-05-12
Remove readonly keyword from buffer pointer definitions

### LINKED ISSUE #6880: When using Ptr<type> and outputting to GLSL, the buffer_reference created is always readonly
**Description**

When using Ptr<type> to global memory (passed by push constants) with GLSL slang correctly creates buffer reference, but it is marked as read only and trying to assign to the pointer results in Slang internal error occurred in the playground.
This is not the case when using SPIRV, I could pass buffers pointers to push constants and do vertex skinning on them without problems. 

Steps to Reproduce

Code resulting in Slang internal error. (commenting the last line compiles the shader, buffer references are created but are read only)
```
struct TestStruct
{
    int value;
}

struct TestStructData
{
    TestStruct testDatas[];
}

struct PushConstants
{
    Ptr<TestStructData> test;
    int someCount;
};

[[vk_push_constant]] PushConstants pc;

RWStructuredBuffer<int> Output;

[numthreads(16,1,1)]
void main(uint2 dispatchThreadId : SV_DispatchThreadID)
{
    uint idx = dispatchThreadId.x;
    if(idx >= pc.someCount)
        return;

    uint someVal = pc.test.testDatas[0].value;
    pc.test.testDatas[0].value = someVal;
}
```

**Expected Behavior**
Slang compiles shader to GLSL, marking buffers as not read only and writing to them or producing a better error description if that is not supported.

**Actual Behavior**
Slang produces internal error when compiling to GLSL in playground.

Link to slang playground example [playground](https://shader-slang.org/slang-playground/?target=GLSL&code=eJx1kbFugzAQhnc_xY0gRaR06FAShoSlU6MmSgeEkIWNQA0G2WcUqcq790wtUhTVk3X-_o-7gzGD2lYIJ2nwOF3ZNwM6rUIY-cXKhN3YI5Vx5J68FwHp6l5MXvyNHaxp9r0yyBUanzqg3ix16RRP5q-bvpP73ioq3RLG8nz8KgcylZVXFcXSDENF3Mfnr9JqKXa2rqXekC2Fd4uDRSdStsNGSy5MEL-s4lUcFmzsWwEdb1VgiX4G0ZqBY9WcJvBNwCscz2W2rGahH8ZloBVX2D4Eo6ufqA4ckG6py2geLZze3NGSOlbU3uxz0JlfYEq41UT39T4Vkf85Dl-v_yco7kW0xh-op6S6)

aidanfnv:
For #6880

This change removes the `readonly` keyword from buffer pointer definitions from the GLSL source emitter, to allow for mutable buffer pointers.
Support for `readonly` will be readded when we add `const` pointer support later.


csyonghe (commented):
Let's not try to be clever here and just remove `readonly` for all cases.

We will support `readonly` when we add `const` pointer support later. For now we should assume all pointers are not mutable.

aidanfnv:
Understood. I have removed all of what I added and instead just removed readonly from the emitted GLSL, and updated the PR description and title accordingly.

csyonghe:
Need to update the expected test result for the existing const-buffer-pointer test.

jkwak-work:
Does this PR resolve the problem with playground?

csyonghe (approved):

PR #7072, opened 2025-05-12
Add explanation for SV_InstanceID usage difference from SPIR-V

### LINKED ISSUE #6805: Update "SPIRV target specific" document for SV_InstanceID
PR #6468 added an support for SV_InstanceID.
We need to update the document for the change.

The document lists all of SV_XXX keywords.
We will need to add the new one there.

And we may need to add more description of how to use it if needed.

aidanfnv:
Closes #6805

This change adds a note to the SPIR-V target specific doc that `SV_InstanceID` does not map directly to SPIR-V's `BuiltIn InstanceIndex`, and adds a more detailed explanation of the difference, its motivation, and how to derive the actual value equivalent to `BuiltIn InstanceIndex` with an example.


jkwak-work (approved):
Looks good to me.

PR #7074, opened 2025-05-12
Make CUDA version capabilities reach NVRTC

### LINKED ISSUE #7049: Slang-test.exe fails tests/diagnostics/local-line.slang.4 test
The following test failures are reported by github build.

failing tests:
---
tests/diagnostics/local-line.slang.4
tests/diagnostics/syntax-error-op-line-2.slang.4
tests/diagnostics/syntax-error-op-line.slang.4
tests/diagnostics/syntax-error-intrinsic.slang.4
tests/diagnostics/syntax-error-op-line-3.slang.4
---
Error: Process completed with exit code 1.

Repro steps:
With CUDA toolkit installed, run command:
$ ./build/Release/bin/slang-test.exe tests/diagnostics/local-line.slang
Supported backends: fxc dxc glslang spirv-dis visualstudio gcc genericcpp nvrtc llvm spirv-opt tint
passed test: 'tests/diagnostics/local-line.slang'
passed test: 'tests/diagnostics/local-line.slang.1'
passed test: 'tests/diagnostics/local-line.slang.2'
passed test: 'tests/diagnostics/local-line.slang.3'
error: ERROR:
EXPECTED{{{
40
}}}
ACTUAL{{{
0
}}}

ignored test: 'tests/diagnostics/local-line.slang.4'
passed test: 'tests/diagnostics/local-line.slang.5'
Retrying 1 failed tests...
error: ERROR:
EXPECTED{{{
40
}}}
ACTUAL{{{
0
}}}

FAILED test: 'tests/diagnostics/local-line.slang.4'

===
83% of tests passed (5/6), 1 tests ignored
===

failing tests:
---
tests/diagnostics/local-line.slang.4
---

tangent-vector:
Fixes #7049

The root cause of the problem in #7049 is simply that newer NVRTC versions produce a warning when asked to generate code for older CUDA SM versions, and the default that Slang was requesting compilation for was old enough to trigger that warning, and thus trip up the test case (which only looks at the first diagnostic produced by the downstream compiler).

Superficially, the fix was easy: change the test case in question (`tests/diagnostics/local-line.slang`) to request `-capability cuda_sm_8_0`, the minimum version supported by current NVRTC.

Unfortunately, the simple fix required some other fixes in order to actually work.

The capability system includes capability names of the form `cuda_sm_*_*`, but specifying such a capability had *no* impact on the CUDA SM version passed in when invoking NVRTC. Instead, only the CUDA SM versions requested in the implementation of intrinsics in the core module were affecting the version number passed down.

This change adds logic to `slang-compiler.cpp` to take explicitly requested capabilities into account when inferring the CUDA SM version to be passed downstream. A more complete fix would also add similar logic for all the other targets.

Unfortunately... yet again... that fix wasn't enough to make things work as expect. Now I had the problem that requesting `-capability cuda_sm_8_0` was actually causing the NVRTC invocation to request CUDA SM version **9.0**!

The underlying problem *there* was that the `slang-capabilities.capdef` file has defined certain capability names in a way that implies atomic capabilities much higher than one would expect. E.g., the `cuda_sm_8_0` alias was including HLSL `sm_5_0`, but then `sm_5_0` in turn included `_cuda_sm_9_0`.

The fix, for now, is to change the definitions in `slang-capabilities.capdef` to not have the counter-intuitive definitions for `cuda_sm_*_*`.

With this set of fixes, the test failure in the original bug report no longer occurs.

The work that went into this change suggests several larger-scope fixes that would be good to pursue:

* Ideally the capability definitions would have some sort of validation checking to make sure that counter-intuitive results like `cuda_sm_8_0` requesting CUDA SM 9.0 do not occur.

* The translation of capabilities over to version numbers for a downstream compiler should be expanded to cover other targets, and not just CUDA. It might be better/simpler to just pass the capabilities themselves to the downstream compiler, since it is possible that a downstream compiler could have more fine-grained enable/disable options than a simple version number.

* The entire approach to computing version numbers required for downstream compilation should be cleaned up so that we don't have this duplication between the capabilities that represent those versions and separate syntactic constructs that are used to "request" those versions as part of code generation.

* We are very much at the point where we should consider dropping the current behavior where a profile name or capability like `sm_5_0`, that is specific to a single target or a subset of targets, also implies a set of comparable capabilities for other targets.

### DIFF
--- a/source/slang/slang-capabilities.capdef
+++ b/source/slang/slang-capabilities.capdef
@@ -1812,43 +1812,43 @@ alias GLSL_430_SPIRV_1_0 = _GLSL_430 + GLSL_420_SPIRV_1_0 | GLSL_420_SPIRV_1_0;
 //
 /// cuda 1.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_1_0 = _cuda_sm_1_0 | sm_4_0;
+alias cuda_sm_1_0 = _cuda_sm_1_0;
 
 /// cuda 2.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_2_0 = _cuda_sm_2_0 | sm_4_1;
+alias cuda_sm_2_0 = _cuda_sm_2_0 | cuda_sm_1_0 | sm_4_0;
 
 /// cuda 3.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_3_0 = _cuda_sm_3_0 | sm_6_0;
+alias cuda_sm_3_0 = _cuda_sm_3_0 | cuda_sm_2_0;
 
 /// cuda 3.5 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_3_5 = _cuda_sm_3_5 | sm_6_0;
+alias cuda_sm_3_5 = _cuda_sm_3_5 | cuda_sm_3_0;
 
 /// cuda 4.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_4_0 = _cuda_sm_4_0 | sm_6_0;
+alias cuda_sm_4_0 = _cuda_sm_4_0 | cuda_sm_3_5;
 
 /// cuda 5.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_5_0 = _cuda_sm_5_0 | sm_6_0;
+alias cuda_sm_5_0 = _cuda_sm_5_0 | cuda_sm_4_0;
 
 /// cuda 6.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_6_0 = _cuda_sm_6_0 | sm_6_0;
+alias cuda_sm_6_0 = _cuda_sm_6_0 | cuda_sm_5_0 | sm_4_1;
 
 /// cuda 7.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_7_0 = _cuda_sm_7_0 | sm_5_1;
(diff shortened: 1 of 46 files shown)


source/slang/slang-capabilities.capdef
```diff
@@ -1812,43 +1812,43 @@ alias GLSL_430_SPIRV_1_0 = _GLSL_430 + GLSL_420_SPIRV_1_0 | GLSL_420_SPIRV_1_0;
 //
 /// cuda 1.0 and related capabilities of other targets.
 /// [Version]
-alias cuda_sm_1_0 = _cuda_sm_1_0 | sm_4_0;
+alias cuda_sm_1_0 = _cuda_sm_1_0;
```
juliusikkala, line 1815: Is there a reason why the `cuda_sm_1_0` does not need to include any `sm_*_*`, while others do have at least one (either directly or inherited)?

juliusikkala (dismissed):
Looks good to me, although I only have somewhat superficial experience with CUDA in general.

source/slang/slang-capabilities.capdef
tangent-vector: It isn't bothering to loop in anything else because even the lowest supported HLSL `sm_*_*` value (which is `sm_4_0`) would wrap around and require a higher CUDA SM version than `_cuda_sm_1_0`.

Putting this PR together has made me realize that there are some really questionable things going on around how the Slang compiler/codebase is handling capabilities/profiles/etc., and a more thorough scrubbing is really called for. I just don't want to have the issue that this PR was originally trying to fix get stalled on a major overhaul like that.

jkwak-work (approved):
Looks good to me.

PR #7076, opened 2025-05-13
Error out on invalid vector sizes

### LINKED ISSUE #6012: Invalid SPIR-V generation for vector<X, 8>/vector<X, 16> usage
Slightly related to https://github.com/shader-slang/slang/issues/5987 but the resolution might be different for the following:

When using vector<X, 8>/vector<X, 16>, Slang doesn't generate a warning nor error but generates invalid SPIR-V output instead.

```
StructuredBuffer<vector<float, 8>> bufferIn;
RWStructuredBuffer<float> resultOut;

[shader("compute")]
[numthreads(1,1,1)]
void computeMain(uint3 threadId : SV_DispatchThreadID)
{
	uint index = threadId.x;
	
	resultOut[index] = bufferIn[index][0] + bufferIn[index][4];
}
```

```
$ slangc test.slang -target spirv -o test.spv
$ spirv-val test.spv
error: line 37: Having 8 components for TypeVector requires the Vector16 capability
  %v8float = OpTypeVector %float 8
```

Marking the SPIR-V output with Vector16 capability is not an option as this is a kernel execution only capability.

Could one of the following be done?

- Generate an error for all output types where this is not supported.
- Split the vector<X,8>/vector<X,16> access into vector<X,4> access.

fairywreath:
Closes #6012. Slang currently only supports vector sizes supported by the compilations targets(i.e. no logic exists to lower long vectors), but no checks are done on the element count. Slang may generate invalid code such as `float13`. This PR adds a a check to error out on invalid vector element counts.

### DIFF
--- a/source/slang/slang-emit.cpp
+++ b/source/slang/slang-emit.cpp
@@ -614,105 +614,6 @@ static void unexportNonEmbeddableIR(CodeGenTarget target, IRModule* irModule)
     }
 }
 
-static void validateVectorOrMatrixElementType(
-    DiagnosticSink* sink,
-    SourceLoc sourceLoc,
-    IRType* elementType,
-    uint32_t allowedWidths,
-    const DiagnosticInfo& disallowedElementTypeEncountered)
-{
-    if (!isFloatingType(elementType))
-    {
-        if (isIntegralType(elementType))
-        {
-            IntInfo info = getIntTypeInfo(elementType);
-            if (allowedWidths == 0U)
-            {
-                sink->diagnose(sourceLoc, disallowedElementTypeEncountered, elementType);
-            }
-            else
-            {
-                bool widthAllowed = false;
-                SLANG_ASSERT((allowedWidths & ~(0xfU << 3)) == 0U);
-                for (uint32_t p = 3U; p <= 6U; p++)
-                {
-                    uint32_t width = 1U << p;
-                    if (!(allowedWidths & width))
-                        continue;
-                    widthAllowed = widthAllowed || (info.width == width);
-                }
-                if (!widthAllowed)
-                {
-                    sink->diagnose(sourceLoc, disallowedElementTypeEncountered, elementType);
-                }
-            }
-        }
-        else if (!as<IRBoolType>(elementType))
-        {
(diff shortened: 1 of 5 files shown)


source/slang/slang-emit.cpp
```diff
@@ -614,105 +614,6 @@ static void unexportNonEmbeddableIR(CodeGenTarget target, IRModule* irModule)
...
-        }
-        else if (auto vectorType = as<IRVectorType>(globalInst))
-        {
-            // Verify that the element type is a floating point type, or an allowed integral type
-            auto elementType = vectorType->getElementType();
-            uint32_t allowedWidths = 0U;
-            if (isWGPUTarget(targetRequest))
-                allowedWidths = 32U;
-            else
-                allowedWidths = 8U | 16U | 32U | 64U;
```
fairywreath, line 704: This code is just moved into the more appropriate location `slang-ir-validate.cpp`.

source/slang/slang-ir-validate.cpp
```diff
@@ -511,4 +512,124 @@ void validateAtomicOperations(bool skipFuncParamValidation, DiagnosticSink* sink
...
+    if (elementCount > maxCount)
+    {
+        sink->diagnose(
+            vectorType->sourceLoc,
+            Diagnostics::vectorWithInvalidElementCountEncountered,
+            elementCount,
+            "1",
+            maxCount);
+    }
+}
```
fairywreath, line 572: This is the only new piece of code.

source/slang/slang-ir-validate.cpp
```diff
@@ -511,4 +511,124 @@ void validateAtomicOperations(bool skipFuncParamValidation, DiagnosticSink* sink
...
+    if ((elementCount < minCount) || (elementCount > maxCount))
+    {
+        sink->diagnose(
+            vectorType->sourceLoc,
+            Diagnostics::vectorWithInvalidElementCountEncountered,
+            elementCount,
+            "1",
+            maxCount);
+    }
+}
```
fairywreath, line 571: This is the only piece of new code.

csyonghe (approved):
looks good, thank you for adding the check!

Issue #7077, opened 2025-05-13
VK_KHR_shader_bfloat16 support for SPIRV


fairywreath:
[SPIRV bfloat16 spec](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_bfloat16.html).
`bfloat16` use is limited to only cooperative matrices and dot products.

Issue #6966, opened 2025-04-30
Stop requiring IComparable for IArithmetic


jkwak-work:
Currently the interface, `IArithmetic`, requires `IComparable` to be implemented.
```
interface IArithmetic : IComparable
```
There are many cases this relationship doesn't make sense.

We should stop requiring IComparable for IArithmetic.
In other words, we should have just like this,
```
interface IArithmetic
```

When this change is made, the PR will be a breaking change.

jkwak-work:
Currently IComparable is defined as the following,
```
interface IComparable
{
    /// Returns true if two values of the conforming type are equal.
    __builtin_requirement($( (int)BuiltinRequirementKind::Equals))
    bool equals(This other);

    /// Returns true if `this` is less than `other`.
    __builtin_requirement($( (int)BuiltinRequirementKind::LessThan))
    bool lessThan(This other);

    /// Returns true if `this` is less than or equal to `other`.
    __builtin_requirement($( (int)BuiltinRequirementKind::LessThanOrEquals))
    bool lessThanOrEquals(This other);
}
```

I like to propose to break it down to three different interfaces
```
interface IEqualityComparable
{
    bool equals(This other);
}

interface ILessThanComparable
{
    bool lessThan(This other);
}

interface IComparable : IEqualityComparable, ILessThanComparable
{
    bool lessThanOrEquals(This other);
}
```

The reason is that equality checking makes sense for both scalar and vector types.
But "lessThan" doesn't make sense for vector types.

For example,
```
float4(0, 1, 1, 0 ) == float(1, 0, 0, 1) // Often used and makes sense.
float4(0, 1, 1, 0 ) < float(1, 0, 0, 1) // Not clear what it means.
```

filonik:
I am not sure why Slang chose to deviate from HLSL/GLSL with regard to the comparison operators on vectors. All the arithmetic operators are mapped component wise, only the comparison operators have the reduction baked in. How do I check if any of the components are equal/lessThan/greaterThan? Returning a `vector<bool, N>` and reducing using `all`/`any` would be more flexible and consistent with HLSL/GLSL.

PR #7078, opened 2025-05-13
Initial bfloat16 support

fairywreath:
Related to #7077.

Adds support for [SPV_KHR_bfloat16](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_bfloat16.html).
`bfloat16` use is limited to only cooperative matrices and dot products, arithmetic operations cannot be performed directly.  This limitation also applies to CUDA. 

This PR adds initial support for properly emitting the type and supporting dot operations. Coop matrix support will be added at a later time, after CM2 support is merged.. Also see implementation details section below for current issues adding `bfloat16` to `CoopMat`.

### Draft task List

- [x] Support type and dot ops for SPIRV
- [ ] Emit type for other targets?
- [ ]  Add tests
- [ ] Agree on implementation details

### Implementation details
Because `bfloat16` does not support arithmetic operations, the builtin type does not implement `__BuiltinArithmeticType`(and hence also does not implement `__BuiltinFloatingPointType`). The only issue here is that the currently `CoopMat`'s  element type argument requires `__BuiltinArithmeticType`. `__BuiltinArithmeticType` is used here as  a type-checking mechanism and there is nothing inherently required by the  `__BuiltinArithmeticType` interface for the element type as all operations are performed to the coop mat type itself, at least when targetting SPIRV. 

The only real use-case for the element type being an arithmetic is to implement the `IComparable` interface as `CoopMat` implements `IArithmetic` - an effort to separate `IComparable` from `IArithmetic` is tracked here https://github.com/shader-slang/slang/issues/6966.  IMO the cleanest solution is to have `CoopMat` drop the requirement for its element type to be `__BuiltinArithmeticType` so that `bfloat16` can be used as an element type, although this may pose some (engineering) difficulties if we ever plan to manually support other targets for cooperative matrices.

The other option is to fully "implement" `__BuiltinArithmeticType` for `bfloat16` and just error out when the arithmetic ops are actually used but this is not ideal. Any other ideas are welcome.

### DIFF
--- a/source/slang-core-module/slang-embedded-core-module-source.cpp
+++ b/source/slang-core-module/slang-embedded-core-module-source.cpp
@@ -56,6 +56,7 @@ enum BaseTypeConversionRank : uint8_t
     kBaseTypeConversionRank_Int32,
     kBaseTypeConversionRank_IntPtr,
     kBaseTypeConversionRank_Int64,
+    kBaseTypeConversionRank_Bfloat16,
     kBaseTypeConversionRank_Error,
 };
 
@@ -149,6 +150,12 @@ static const BaseTypeConversionInfo kBaseTypes[] = {
      UINT_MASK,
      kBaseTypeConversionKind_Unsigned,
      kBaseTypeConversionRank_IntPtr},
+
+    {"bfloat16",
+     BaseType::BFloat16,
+     FLOAT_MASK,
+     kBaseTypeConversionKind_Float,
+     kBaseTypeConversionRank_Bfloat16},
 };
 
 void Session::finalizeSharedASTBuilder()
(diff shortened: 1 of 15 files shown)


source/slang-core-module/slang-embedded-core-module-source.cpp
```diff
@@ -149,6 +150,12 @@ static const BaseTypeConversionInfo kBaseTypes[] = {
      UINT_MASK,
      kBaseTypeConversionKind_Unsigned,
      kBaseTypeConversionRank_IntPtr},
+
+    {"bfloat16",
```
csyonghe, line 154: should this be `bfloat16_t` to stay consistent with other explicitly sized types like `int64_t`?

PR #7079, opened 2025-05-13
Implement default initializer list for C-Style type member

### LINKED ISSUE #6189: Provide default initializer to C-Style type members
Due to SP004, lots of core module types are not default initializable anymore.

This creates an issue where if a struct type includes a field that is not c-style, the entire struct type can no longer be initialized with an empty initialization list.

For example:
```
struct NotCStyle { __init() {... } }

struct Foo {
NotCStyle nonCstyleMember;
int x;
int y;
}

Foo f = {}; // invalid, cannot call Foo.__init with 0 arguments.
```

This is causing a lot of friction in development, because it forces the user to put `={}` on all struct fields to enable the type to be default initialized with empty intiailizer list.

We should make it so that when we synthesize a ctor, if we encounter a field whose type is c-style, we will assign an empty initializer list `{}` to it. For `Foo` type, we currently synthesize ctor with this signature:
```
Foo.init(NotCStyle nonCstyleMember = NotCStyle(), int x, int y);
```

But instead we should be synthesising:

```
Foo.init(NotCStyle nonCstyleMember = NotCStyle(), int x = {}, int y = {});
```
because `={}` is valid on `int` since it is a cstyle type. 
Basically, as long as `={}` is valid on the field, the ctor parameter should have a default value.

kaizhangNV:
Close #6189.

Previsouly, for the C-Style member in a struct, if it doesn't have any initialize expression, when we synthesize the ctor, we will not associate the default value for the parameter corresponding to that member.

This bring some trouble that existing slang users has to add '= {}' to every struct fields in order to make all the parameters in the synthesized ctor having a default value, so people can still use `Struct a = {}` to create a struct.

To make this use case convenience, we will automatically associated a '= {}' as the default value for this case.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -2330,6 +2330,18 @@ static Expr* constructDefaultConstructorForType(SemanticsVisitor* visitor, Type*
         return invoke;
     }
 
+    // At the last, we will check if the type is a C-style type, if it is, we will use empty initializer
+    // list to construct the default constructor.
+    HashSet<Type*> isVisit;
+    if (visitor->isCStyleType(type, isVisit))
+    {
+        auto initListExpr = visitor->getASTBuilder()->create<InitializerListExpr>();
+        initListExpr->type = visitor->getASTBuilder()->getInitializerListType();
+        Expr* outExpr = nullptr;
+        if (visitor->_coerceInitializerList(type, &outExpr, initListExpr))
+            return outExpr;
+    }
+
     return nullptr;
 }
 
@@ -12853,7 +12865,13 @@ static Expr* _getParamDefaultValue(SemanticsVisitor* visitor, VarDeclBase* varDe
     if (!isDefaultInitializable(varDecl))
         return nullptr;
 
-    return constructDefaultConstructorForType(visitor, varDecl->type.type);
+    if (auto expr = constructDefaultConstructorForType(visitor, varDecl->type.type))
+    {
+        expr->loc = varDecl->loc;
+        return expr;
+    }
+
+    return nullptr;
 }
 
 bool SemanticsDeclAttributesVisitor::_synthesizeCtorSignature(StructDecl* structDecl)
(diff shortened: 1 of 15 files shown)


kaizhangNV:
The CI failure is mainly because we don't have link-size array/vector/matrix support to use "= {}". But on the ToT slang, it already has such support for array type.

But we still need to have vector/matrix support.

This test could be still failed.
tests/autodiff/self-differential-generic-type-synthesis.slang

kaizhangNV:
We could have a problem
```
struct A
{
    int a;
    int b;
}

A a[2] = {1, 2, 3, 4};
```

This case could not work anymore.

Because this will fall back to legacy init list,
when we read first element '1', it will find out that 1 can be coerced to `struct A`, because `struct A` has a ctor with signature `__init(int a = {}, int b = {})`.

csyonghe:
If we are falling back to legacy init list, why are we still calling ctors? We should just be filling in the fields directly, completely bypass any ctors?

kaizhangNV:
> If we are falling back to legacy init list, why are we still calling ctors? We should just be filling in the fields directly, completely bypass any ctors?

Because when we read each element of the initializer list, we check whether this element can be coerced to the target type, and that check will find out that `there is a ctor`, then it use that ctor.

```
    if (shouldUseInitializerDirectly(toType, firstInitExpr))                // -> this will return true
    {
        ioInitArgIndex++;
        return _coerce(
            CoercionSite::Initializer,
            toType,
            outToExpr,
            firstInitExpr->type,
            firstInitExpr,
            nullptr);
    }
```

csyonghe:
I feel like for C-style types we should completely bypass any ctors and do plain member initialization.

csyonghe:
Can you just use the existing `DefaultConstructExpr`?

source/slang/slang-check-decl.cpp
```diff
@@ -2330,6 +2330,23 @@ static Expr* constructDefaultConstructorForType(SemanticsVisitor* visitor, Type*
         return invoke;
     }
 
+    // At the last, we will check if the type is a C-style type, if it is, we will use empty
+    // initializer list to construct the default constructor.
+    HashSet<Type*> isVisit;
```
csyonghe, line 2335: `isVisit` doesn't read as correct english. consider `visitSet`?

csyonghe (dismissed):

Issue #6927, opened 2025-04-28
Enable full test suite on Windows/Debug CI.


csyonghe:
We are currently hitting a bunch of debug asserts that is not captured in the CI. We should enable full test suite with Windows/Debug config and fix all the failures.

gtong-nv:
The https://github.com/KhronosGroup/SPIRV-Tools/issues/5959 still causes spirv-val errors

error: line 107: NonSemantic.Shader.DebugInfo.100 DebugFunctionDefinition: must be in the entry basic block of the function
%27 = OpExtInst %void %2 DebugFunctionDefinition %23 %computeMain

Confirmed that running with -O0 the issue is gone. Should we add -O0 to those failed tests to bypass the error for now?

gtong-nv:
The debug test suite is still not stable in all CI machines, and the intermittent failures are hard to be located.
To narrow down the issue, I added a label to make it run only in the GCP runners, and was able to consistently show an failure - filed https://github.com/shader-slang/slang/issues/7089

PR #7085, opened 2025-05-13
Enable Windows full debug testsuite in CI

gtong-nv:
This PR turns on the full debug slang-test in CI

- Disables SPIRV optimization for DebugFunctionDefinition test in order to pass the debug build. Root cause might be the spirv-opt, bug tracked in https://github.com/KhronosGroup/SPIRV-Tools/issues/5959
- Avoid using test server for unit-test and gfx-test (e.g. `slang-unit-test-tool/RecordReplay.internal`), which always fail with non-zero exit code 
- Disable debug (validation) layer via the new added option to slang-test
- Limit the debug testsuite in "GCP" runners only
- Skip `Test Slang via glsl` for debug build to make sure the job finishes in ~30 min

Fixes: #6927


gtong-nv:
\format

kaizhangNV (dismissed):
LGTM

kaizhangNV (approved):

PR #7086, opened 2025-05-13
Stop escaping markdown between backticks

aidanfnv:
Fixes https://github.com/shader-slang/shader-slang.github.io/issues/95

The current Standard Modules Reference markdown generator escapes characters in code in the Conditional Conformances sections, which is not necessary, as characters between backticks in markdown are already escaped. On GitHub, the redundant escape slashes are ignored, but on both ReadTheDocs and GitHub Pages the escape slashes are still visible.
This change removes the redundant escaping.


jkwak-work (approved):
Looks good to me.

PR #7087, opened 2025-05-13
Implement C++ style default member initializer

### LINKED ISSUE #7069: Init List: C++ style default initializer
```
Struct MyStruct
{
     int a = 3;
     float2 b;
     __init() { b = 2;}
}

```

For above code, we currently don't initialize member `a`. 

We should support that to initialize all the members when there are default initializer attached in user defined constructor.

kaizhangNV:
close #7069.

We just insert the assignment expression to the beginning of the user-defined ctor, where the assginee is the member with init expression.

### DIFF
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -9392,8 +9392,7 @@ void SemanticsDeclBodyVisitor::maybeInsertDefaultInitExpr(StructDecl* structDecl
         if (auto varDeclBase = as<VarDeclBase>(member))
         {
             if (varDeclBase->hasModifier<HLSLStaticModifier>() ||
-                varDeclBase->getName() == nullptr ||
-                varDeclBase->initExpr == nullptr)
+                varDeclBase->getName() == nullptr || varDeclBase->initExpr == nullptr)
                 continue;
 
             MemberExpr* memberExpr = createMemberExpr(thisExpr, firstCtor->ownedScope, member);
--- a/source/slang/slang-check-decl.cpp
+++ b/source/slang/slang-check-decl.cpp
@@ -382,7 +382,7 @@ private:
 
     MemberExpr* createMemberExpr(ThisExpr* thisExpr, Scope* scope, Decl* member);
     Expr* createCtorParamExpr(ConstructorDecl* ctor, Index paramIndex);
-    void maybeInsertDefaultInitExpr(FunctionDeclBase* decl);
+    void maybeInsertDefaultInitExpr(StructDecl* structDecl);
 };
 
 template<typename VisitorType>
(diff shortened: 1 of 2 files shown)


source/slang/slang-check-decl.cpp
```diff
@@ -9368,6 +9369,71 @@ void SemanticsDeclBodyVisitor::synthesizeCtorBodyForMember(
 }
 
 
+void SemanticsDeclBodyVisitor::maybeInsertDefaultInitExpr(StructDecl* structDecl)
```
csyonghe, line 9372: Is this function inserting assignment statements to user defined ctors for fields with `=expr`? I wonder if we should be skipping synthesized ctors in the loop at 9429?

source/slang/slang-check-decl.cpp
```diff
@@ -9368,6 +9369,71 @@ void SemanticsDeclBodyVisitor::synthesizeCtorBodyForMember(
...
+        return;
+
+    auto seqStmtChild = m_astBuilder->create<SeqStmt>();
+    seqStmtChild->stmts.reserve(structDecl->members.getCount());
+    Dictionary<Decl*, Expr*> cachedDeclToCheckedVar;
+
+    // We use first ctor to create the `this` expression, so that we can modify the
+    // members in assign expression later.
+    auto firstCtor = structDecl->getMembersOfType<ConstructorDecl>().getFirst();
+    ThisExpr* thisExpr = m_astBuilder->create<ThisExpr>();
```
csyonghe, line 9384: I think reusing `thisExpr` created using the scope of first ctor and reuse it in other ctors is probably dangerous and can break in the future. Can we just duplicate this logic for every ctor, so the `this` expr gets the right scope?

kaizhangNV: OK, this is indeed a little bit of hacky.

source/slang/slang-check-decl.cpp
kaizhangNV: Sorry, I'm not sure I'm following.
> Is this function inserting assignment statements to user defined ctors for fields with `=expr`?
Yes

> wonder if we should be skipping synthesized ctors in the loop at 9429?
I'm not synthesizing ctors.
9429 just insert the assignment statement for every user-defined ctors.

csyonghe: I see that loop iterates over all ctors in the struct type, are we sure all ctors there are user-defined and not compiler-synthesized?

kaizhangNV: Yes, the first line in this function make sure of that.
```
    if (!_hasExplicitConstructor(structDecl, false))
        return;
```

If there is no explicit ctor defined, it will just return.

csyonghe: Ah so the logic is if the user defined any ctor, then we won’t be synthesizing any ctors, therefore all ctors in the struct member list must be user defined. Can you add a comment to explain this?

kaizhangNV: Yes, sure.

kaizhangNV:
update PR.

csyonghe (approved):

Issue #6704, opened 2025-03-31
`static const let` fails with cyclic reference error


Devon7925:
Trying to use `static const let` fails with an error for basic types.

```slang
import playground;

struct Data {}

static const let ABC = Data(); // works fine
static const let XYZ = 123; // cyclic reference

float4 imageMain(uint2 dispatchThreadID, int2 screenSize)
{
}
```
[Playground link](https://shader-slang.org/slang-playground/?target=SPIRV&code=eJxlyrEOgkAQRdF-v2JKSUyIaEcsVBoLKy3UbjIMMHHZJbtDDBL-XaS1fO8eaTsfFDqLQx1878rcmKihJ4UCFWGcfhtVCMi7qGBZ4XA8wX7pqySHNIW3D68IlTj-x_fHc8abbLtIGsjOOXDFgR2xMZX1qDuQFmu-oLhVL04zKCV2qNTcmsBYnos1LHekwOyu8uHEjGb6AikiQDc)

PR #7090, opened 2025-05-13
Infer type while constant folding causes failure

ArielG-NV:
fixes: #6704

Problem:
* Inferring type with `let` while using constant-foldable object's means we run in a circular `ensureDecl`'s loop (`SemanticsDeclHeaderVisitor::checkVarDeclCommon` to `_validateCircularVarDefinition` to `tryConstantFoldDeclRef`). This is a compiler error.

Fix:
* If we require to constant-fold and are a variable, we are ready to check definition. If we are not a constant to fold, we run `setCheckState` after anyways. Due to this, we can just set the decl state before our call to `_validateCircularVarDefinition`.

@kaizhangNV helped with the solution.


csyonghe (approved):
Looks good!

PR #7091, opened 2025-05-13
Fix broken -emit-spirv-via-glsl test option

sricker-nvidia:
Fixes issue #6898

The -emit-spirv-via-glsl slang-test option has been broken for some amount of time. Tests that were using it were operating as if using -emit-spirv-directly, leading to many duplicated tests.

After fixing the test option, there were an number of errors that appeared as a result.

This change fixes the broken test option and the resulting test errors. Some of the test errors revealed some legitimate issues, such as:

-The GLSL bitCount instrinsic only supports 32-bit integers and
 requires emulation for other bit widths.
-Emitting GLSL 8-bit and 16-bit glsl integer types did not emit
 the proper extension requirements
-Emitting GLSL and casting for 16-bit integers was missing a
 closing parenthesis.


sricker-nvidia:
Noting that I left a comment/question in the code denoted by "???" that I was hoping to discuss via this PR review.

include/slang.h
```diff
@@ -717,13 +717,8 @@ typedef uint32_t SlangSizeT;
 
         /* When set, will dump out the IR between intermediate compilation steps.*/
         SLANG_TARGET_FLAG_DUMP_IR = 1 << 9,
-
-        /* When set, will generate SPIRV directly rather than via glslang. */
-        // This flag will be deprecated, use CompilerOption instead.
-        SLANG_TARGET_FLAG_GENERATE_SPIRV_DIRECTLY = 1 << 10,
```
csyonghe, line 723: Let's not remove this to avoid breaking binary compatibility.

tools/render-test/slang-support.cpp
```diff
@@ -189,10 +189,26 @@ static SlangResult _compileProgramImpl(
...
-    else
-        sessionTargetDesc.flags = 0;
+
+    {
+        slang::CompilerOptionEntry entry;
+        entry.value.kind = slang::CompilerOptionValueKind::Int;
+        if (options.generateSPIRVDirectly)
+        {
+            entry.name = slang::CompilerOptionName::EmitSpirvDirectly;
+            // A value of 1 indicates SLANG_EMIT_SPIRV_DIRECTLY
```
csyonghe, line 199: I think it is really the existence of the entry means we want to emit spirv directly, the `intValue0` here doesn't matter and is ignored.

include/slang.h
sricker-nvidia: ack

tools/render-test/slang-support.cpp
sricker-nvidia: Are you sure the intValue0 is ignored? I thought it was getting used here in CompilerOptionSet::load:

````
value.intValue = entries[i].value.intValue0;
...
            set(slang::CompilerOptionName::EmitSpirvMethod,
                value.intValue != 0 ? SLANG_EMIT_SPIRV_DIRECTLY : SLANG_EMIT_SPIRV_VIA_GLSL);
````

File: tools/render-test/slang-support.cpp [hunk 3]
```diff
@@ -194,6 +194,22 @@ static SlangResult _compileProgramImpl(
...
+        entry.value.kind = slang::CompilerOptionValueKind::Int;
+        if (options.generateSPIRVDirectly)
+        {
+            entry.name = slang::CompilerOptionName::EmitSpirvDirectly;
+            entry.value.intValue0 = int(options.generateSPIRVDirectly);
+        }
+        else
+        {
+            entry.name = slang::CompilerOptionName::EmitSpirvViaGLSL;
+            entry.value.intValue0 = int(!options.generateSPIRVDirectly);
```
jkwak-work, line 208: Can we just assign 1? I am not sure why the indirection here?
```
entry.value.intValue0 = 1;
```

jkwak-work (approved):
Looks good to me.

tools/render-test/slang-support.cpp
sricker-nvidia: I thought it would make things more obvious in terms of intent vs just setting 1 or 0. The slang::CompilerOptionName::EmitSpirvDirectly technically accepts both 1 and 0. I had a comment in a previous revision noting that it was possible to set EmitSpirvViaGLSL both using slang::CompilerOptionName::EmitSpirvDirectly with intValue0 = 0 and with slang::CompilerOptionName::EmitSpirvViaGLSL and intValue0 =1

sricker-nvidia: I don't feel strongly either way and can switch it if the above does not make sense.

PR #7103, opened 2025-05-14
Add full support for SPV_NV_shader_subgroup_partitioned

### LINKED ISSUE #6839: Clean up WaveMask* implementations for SPIRV and GLSL
Slang has custom `WaveMask*` intrinsics which perform masked subgroup operations and are functionally the same as HLSL's `WaveMultiPrefix*` intrinsics. The `WaveMask*` functions are not properly implemented for SPIRV and GLSL however, `SPV_NV_shader_subgroup_partitioned`'s ops should be used but instead regular non-masked subgroup ops are used. `WaveMultiPrefix*` support is already (correctly) implemented and the `WaveMask*` implementation for SPIRV and GLSL should be fixed.

### LINKED ISSUE #6846: Implement GLSL GL_NV_shader_subgroup_partitioned
Most of the intrinsics can be aliases to the already implemented `WaveMultiPrefx*` intrinsics.

fairywreath:
Closes #6839, closes #6846.

Exposes all intrinsics provided by `SPV_NV_shader_subgroup_partitioned`. Added Slang-specific `WaveMulti` intrinsics to access SPIRV intrinsics through standard Slang (non-GLSL). Additionally refactor/clean-up a huge portion of `WaveMask` code and add proper implementations of them for SPIRV, GLSL and HLSL.

This is a breaking change because the current behavior of having the `mask` being ignored for the `WaveMask` intrinsics when targeting SPIRV/GLSL and HLSL is changed. Some intrinsics are supported by the target language and in this case the `mask` parameter is taken into account properly. This is technically a _fix_ but it is a breaking change nonetheless.


csyonghe (dismissed):
This is a very nice clean up. It would be great if we can completely remove the special case for CUDA target and just make everyone use WaveMultiXXX functions. We may want to consider delete WaveMask functions in the future, and maybe we mark all of them deprecated now?

fairywreath:
I made changes to have `WaveMask*` call the new `WaveMulti*` functions as much as possible, but there are still some cases where `WaveMask*` still has its own implementation:
- Some `WaveMask*` intrinsics are not supported by `SPV_NV_shader_subgroup_partitioned` or HLSL's `WaveMultiPrefix*`, and in this case there is no `WaveMulti*` function to call to.
- HLSL does fully support all intrinsics provided `SPV_NV_shader_subgroup_partitioned` while all `WaveMask*` operations  "support" HLSL (by ignoring the mask parameter :/). Some `WaveMulti*` intrinsics are not supported when targeting HLSL and in these cases the corresponding `WaveMask` intrinsic still maintains its incorrect HLSL implementation (where the mask is ignored).

csyonghe:
Yes, let's deprecate the WaveMask* functions that has SPIRV equivalent now.

fairywreath:
> Yes, let's deprecate the WaveMask* functions that has SPIRV equivalent now.

Can I do this in a separate PR? I am seeing some issues with the `[Deprecated(...)]` attribute when applied to some functions which needs a fix and a lot of the HLSL-standard `WaveActive*` functions still call `WaveMask*`. This PR is already very big as-is, specifically the changes inside `hlsl.meta.slang`.

csyonghe:
Sure, this PR is good as is.

csyonghe (approved):

PR #7105, opened 2025-05-14
Rename 'main' on some backends

### LINKED ISSUE #5542: [Optix]"main" entry point name not allowed on some backends
Some backends (cpu, cuda, metal) fail to compile a shader that has an entry point named `main`.

mkeshavaNV:
Fixes #5542

Some backends like cuda, metal, cpu do not allow 'main' as the entry point. This commit adds a new warning that is emitted when a program uses main as the entry point. In addition to emitting the warning, it internally renames the entry point to "main_". It also adds a test to check these for the three backends.


tests/compute/entry-point-main-warning.slang
```diff
@@ -0,0 +1,14 @@
+// Test to check if we emit a warning and rename main to main_ on CUDA, CPU, metal backends
```
szihs, line 1: _nit_ : Should we add this in tests/diagnostics

mkeshavaNV: yeah I did think that was the right place, but this does not affect all backends. Still, it maybe better than compute. Moved.

tests/diagnostics/entry-point-main-warning.slang
```diff
@@ -0,0 +1,14 @@
+// Test to check if we emit a warning and rename main to main_ on CUDA, CPU, metal backends
+
+//TEST:SIMPLE(filecheck=CHECK): -entry main -target cuda
```
szihs, line 3: s/TEST/DIAGNOSTIC_TEST

mkeshavaNV: fixed

szihs (commented):
LGTM

source/slang/slang-emit-c-like.cpp
```diff
@@ -34,6 +34,8 @@ namespace Slang
 bool isCPUTarget(TargetRequest* targetReq);
 bool isCUDATarget(TargetRequest* targetReq);
 
+Dictionary<String, String> CLikeSourceEmitter::s_entryPointRenameMap = {{"main", "main_"}};
```
expipiplus1, line 37: I think that we can probably just put these strings directly in the maybe rename function, seems a bit extra to make a dict here for just one mapping

expipiplus1 (changes requested):
What happens if there's another entry point already called 'main_'?

mkeshavaNV:
> What happens if there's another entry point already called 'main_'?

Ah, fixed this. I didn't realize we can have multiple entry points for the same stage.

source/slang/slang-emit-c-like.cpp
mkeshavaNV: removed this now.

expipiplus1 (approved):

Issue #6758, opened 2025-04-07
Support CoopVec with a coming DXC April preview release


jkwak-work:
## Background

Microsoft is planning to release a preview of SM 6.9 by End of April.
And it is expected to include some implementations for Cooperative Vector.

The proposal document can be found here,
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0029-cooperative-vector.md
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0031-hlsl-vector-matrix-operations.md

The proposal doc describes that there will be the end-user facing API functions.
And there will be built-in functions that the API functions use.

For the preview release, DXC may include only the built-in functions described in the proposal doc above.
The WIP implementation can be found here,
https://github.com/microsoft/DirectXShaderCompiler/pull/7290

The notable differences from the CoopVec for SPIR-V is that the type is not an opaque type but a native vector type like vector<T,N> where N can be an any value from 1 to 1024.
This is also called "Long vectors" and its proposal can be found here,
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0030-dxil-vectors.md

## Goal for Slang

Slang currently uses DXC build built from a custom DXC branch created by NVIDIA.
https://github.com/NVIDIA-RTX/DirectXShaderCompiler/tree/CooperativeVector

The NVIDIA branch implemented the "built-in" functions for Cooperative Vector.
And it had been the only way to use CoopVec features with DX12 backend.
When the preview of DXC is released, Slang should be able to use the official DXC builds rather than the NVIDIA branch.

The function names used in the NVIDIA branch is different from what was proposed to DXC.
And there might be more differences than just the name differences but they are expected to be trivial differences.

jkwak-work:
It appears that I need to add a new SM profile version 6.9 to support this.

And because it is no longer an opaque type, some of functions are missing for the construction of CoopVec such as `fill()`.

jkwak-work:
Setting this as Blocked.
Currently DXIL side is not implemented and I cannot test my changes yet.

jkwak-work:
Changing the status to "In progress", because "Blocked" status draws unnecessary attention.

jkwak-work:
Pushing the issue to the next sprint because DXIL compiler will not be ready until next week.

jkwak-work:
Some of required changes for this issue are merged:
- https://github.com/shader-slang/slang/pull/6879
- https://github.com/shader-slang/slang-rhi/pull/325

[...]

PR #7108, opened 2025-05-14
Support the new CoopVec builtins

jkwak-work:
**NOTE: This is a breaking change for users who were using POC variant of DXC.
In order to keep the compatibility, the users will have to use `-capability hlsl_coopvec_poc` to their command line.

This PR adds a new capability "hlsl_coopvec_poc".
- When it is used, the HLSL for CoopVec will be emitted for [the POC variant of DXC](https://github.com/NVIDIA-RTX/DirectXShaderCompiler/tree/CooperativeVector).
- When it is not used, the HLSL for CoopVec will be emitted for the DXC that officially supports the cooperative vector.

Closes https://github.com/shader-slang/slang/issues/6758

### DIFF
--- a/source/slang/slang-capabilities.capdef
+++ b/source/slang/slang-capabilities.capdef
@@ -224,7 +224,7 @@ def hlsl_nvapi : hlsl;
 /// [Version]
 def hlsl_2018 : _sm_5_1;
 
-/// Represet HLSL compatibility support for the deprecated POC DXC
+/// Represet compatibility support for the deprecated POC DXC
 /// [Version]
 def hlsl_coopvec_poc : _sm_6_8;
 
--- a/source/slang/slang-capabilities.capdef
+++ b/source/slang/slang-capabilities.capdef
@@ -224,6 +224,10 @@ def hlsl_nvapi : hlsl;
 /// [Version]
 def hlsl_2018 : _sm_5_1;
 
+/// Represet HLSL compatibility support for the deprecated POC DXC
+/// [Version]
+def hlsl_coopvec_poc : _sm_6_8;
+
 /// Represents capabilities required for DXIL Library compilation.
 /// [Version]
 alias dxil_lib = _sm_6_3;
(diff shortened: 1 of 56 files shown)


source/slang/slang-capabilities.capdef
```diff
@@ -1118,7 +1122,7 @@ alias bufferreference_int64 = bufferreference + GL_EXT_shader_explicit_arithmeti
 /// Note that cpp and cuda are supported via a fallback non-cooperative implementation
 /// No HLSL shader model bound yet
 /// [Compound]
-alias cooperative_vector = _sm_6_8 | cpp | _cuda_sm_9_0 | spvCooperativeVectorNV;
+alias cooperative_vector = _sm_6_9 | cpp | _cuda_sm_9_0 | spvCooperativeVectorNV;
```
ArielG-NV, line 1125: Question about design:
D3D CoopVectors seem to have [tiers and optional features as per section `Support Tiers`](https://microsoft.github.io/hlsl-specs/proposals/0029-cooperative-vector.html).

Do we maybe want to break down the capability of cooperative_vector into these subsets (`D3D12_COOPERATIVE_VECTOR_TIER_1_1` and `D3D12_COOPERATIVE_VECTOR_TIER_1_0`)?

ArielG-NV (commented):
Changes look good to me. I have 1 question.
I will let someone else review before approval.

source/slang/slang-capabilities.capdef
jkwak-work: I am not sure if we want to go by the fine granularity on HLSL like how we are doing for SPIRV.
On Spirv side, the tier_1_1 is same to the training instructions.
```
alias cooperative_vector_training = spvCooperativeVectorTrainingNV;
```
But I think we have been going with just shader-model version numbers for D3D backend rather than the finer names.

I prefer to not differentiate them on D3D side for now; if nobody tells me otherwise.
We can add them back later if needed.

csyonghe (approved):
Looks good, my understanding is we will get rid of all the poc specific logic in next month or so, when the official DXC support for coop vector is available, right?

source/slang/slang-capabilities.capdef
csyonghe: sm_6_9 for HLSL target is fine I think.

jkwak-work:
> Looks good, my understanding is we will get rid of all the poc specific logic in next month or so, when the official DXC support for coop vector is available, right?

We could do that and I prefer to do that.
But I am not sure if anybody will yell at us for not supporting it, although the chance is very low.

Issue #5570, opened 2024-11-15
Mismatched qualifier on link-time-variables causes incorrect IR which will fail on SPIR-V targets


ccummingsNV:
When linking together the following 2 modules:

```
    //xx.slang
    export static float VAL = 42.0;

    //yy.slang
    extern static const float VAL;
    RWStructuredBuffer<float> result;
    [shader("compute")]
    [numthreads(1, 1, 1)]
    void main(uint3 tid: SV_DispatchThreadID)
    {
        result[0] = VAL;
    }
```

The omission of the 'const' in xx.slang causes Vulkan to silently generate incorrect SPRI-V and 'result' is not populated. On D3D12 the generated hlsl correctly populates 'result' with the value '42.0'.

Changing the line to `export static const float VAL=42.0` results in correct behaviour on both platforms.

Expected behaviour: Omission of const should either work on both platforms, or fail with a link time error on both platforms.

Incorrect SPIRV:
[spriv.zip](https://github.com/user-attachments/files/17773222/spriv.zip)

ArielG-NV:
Update to the Issue:
* Mismatched qualifier on link-time-variables causes incorrect IR which will fail on SPIR-V targets

Flaw causing problems found:
1. we allow non-const-link-time-var to be linked to a const-link-time-var.
2. problem is that: module use site has `const` var, so, we emit `OpStore %Ptr %Const` in IR, this is expected, this is good. We fail because we in reality have a `OpStore %Ptr %Var` (fails since we need a `OpLoad` in-between) in IR since the module with our link-time-variable-value is a regular variable.
3. we loose the float_litteral talked about inside the github issue since, we technically don't use our variable "VAL" (we never `OpLoad` from it), so spirv-opt removes the float_litteral, this is a byproduct of the actual issue.

Solution we will implement:
* Export of link-time-variables must always be const

PR #7113, opened 2025-05-14
Enforce rule that `export`/`extern` (non cpp) must be `const`

ArielG-NV:
fixes: #5570

Problem:
1. we allow non-const-link-time-var to be linked to a const-link-time-var.
2. problem is that: module use site has const var, so, we emit `OpStore %Ptr %Const` in IR, this is expected, this is good. We fail because we in reality have a `OpStore %Ptr %Var` in IR since the module with our link-time-variable is a regular variable (not `const`). This is an error since `OpLoad` is needed to load from `%Var`

Solution:
* `export`/`extern` variables must always be `const`. This excludes `__extern_cpp` since `cpp` does not exhibit this issue and works differently.


source/slang/slang-check-decl.cpp
```diff
@@ -42,6 +42,34 @@ struct SemanticsDeclModifiersVisitor : public SemanticsDeclVisitorBase,
...
+                hasExportOrExtern = true;
+            else if (as<ConstModifier>(m))
+                hasConst = true;
+            else if (as<ExternCppModifier>(m))
+            {
+                isExternCPP = true;
+                break;
+            }
+        }
+        if (!isExternCPP && hasExportOrExtern && !hasConst)
```
csyonghe, line 66: extern_cpp should be irrelevant to this check.

The logic should be, if there is found `hasExportOrExtern`, then the var must be `static const`.

ArielG-NV: `slang\tools\slang-unit-test\unit-test-com-host-callable.slang` explicitly makes use of writeable `export __extern_cpp __global int`. Am I to assume that this test is no-longer legal Slang?

csyonghe: We could say an export/ extern is either static const, or not. Let’s forbid static or const.

csyonghe (dismissed):

source/slang/slang-check-decl.cpp
```diff
@@ -42,6 +42,31 @@ struct SemanticsDeclModifiersVisitor : public SemanticsDeclVisitorBase,
...
+        // Export'd/Extern'd variables must be `const`, otherwise we may have a mismatch
+        // causing errors.
+        bool hasConst = false;
+        bool hasExportOrExtern = false;
+        bool hasStatic = false;
+        for (auto m : decl->modifiers)
+        {
+            if (as<ExternModifier>(m) || as<HLSLExportModifier>(m))
+                hasExportOrExtern = true;
+            else if (as<ConstModifier>(m))
```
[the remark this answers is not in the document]
ArielG-NV: `m` is 1 modifier inside the modifier list, `else if (as<ConstModifier>(m) && as<HLSLStaticModifier>(m))` would always 
 be false.

steps are:
1. do we have export/extern.
2. we must find in the list `static` and `const` or neither, otherwise fail.

csyonghe (approved):

jkwak-work:
I am not sure if this is a breaking change.
Wasn't it already broken and this PR fixes it?

ArielG-NV:
> I am not sure if this is a breaking change.

The original problem that was looked into was that with a mismatch of export/extern qualifiers we fail since we cannot `OpLoad` a Variable directly (case where we are missing `const`) but we can `OpLoad` a constant (case where we have a `const`).

Due to this causing the failure, it is possible to have a case where both a variable and constant can be used interchangeably and still produce valid SPIR-V. This makes the PR a breaking change since someone may accidently be using this case.

PR #7118, opened 2025-05-14
Add checking for hlsl register semantic.

### LINKED ISSUE #7112: HLSL input missing validation for register u/b/c/t
slang accepts a shader that incorrectly assigns a RWStructuredBuffer to a register(t#).  For example: https://shader-slang.org/slang-playground/?target=SPIRV&code=eJwtjcEKwjAQRO_7FXNsQaSCJ6MevHtRwYNIKW1aAtqEzUYQ6b-bNlkWdlhm5pm3syxwr-Y7sA1jp4i8cGgFN-2lFvoR4gQzyha9tYomRZf7dfEE1t0p9L3mfbbjiKGeJXZgPRgvmgupVvCuafWmVPSxpoPj2HduzFiUGZDjEo9aHrNaRyAOCV4gtuRFmTwJ9aie0ZSS0x982EFp

```
import playground;

struct Test_t
{
    uint4 foo;
};
RWStructuredBuffer< Test_t > g_Test : register(t0, space1);
void printMain()
{
    Test_t test;
    test.foo = uint4( 0, 0, 0, 0 );
    g_Test[0] = test;
}
```

This should be an error, the spaces for each type is spelled out in https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#implicit-binding-number-assignment.  RWStructuredBuffer needs to go into the u space.  DXC will throw an error if you give the wrong register.

The problem this gave for us is that we use auto shifted bindings and basically what happens is that it ignores the register syntax and the binding ends up in the wrong descriptor set and binding leading to very difficult debugging :)

csyonghe:
Closes #7112.


source/slang/slang-diagnostic-defs.h
```diff
@@ -1366,6 +1366,13 @@ DIAGNOSTIC(
     Error,
     variableCannotBePushAndSpecializationConstant,
     "'$0' cannot be a push constant and a specialization constant at the same time")
+
+DIAGNOSTIC(31221, Error, invalidHLSLRegisterName, "invalid HLSL register name '$0'.")
+DIAGNOSTIC(
+    31222,
+    Error,
+    invalidHLSLRegisterNameForType,
+    "invalid HLSL register name '$0' for type '$1'.")
```
fairywreath, line 1375: Can we also output the expected/proper register name here? I guess this is tricky to do when the incorrect usage comes from the `if (auto resType = as<ResourceType>(varType))` branch but this should be easy for the other invalid cases.

csyonghe: it is possible, but then there are tricky cases that will make the code a lot messier (e.g. combined samplers), so I decided against it for simplicity.

fairywreath (approved):

PR #7128, opened 2025-05-15
Fix slangc path used by cmdline reference doc regeneration action

aidanfnv:
This change uses a different path for accessing the slangc binary during the command line reference doc regeneration action, as the environment variable referenced for the bin directory may not be accurate.

This change also makes some exclusions from the slang build in the command line reference actions, to speed up those builds for checking/regenerating.

### DIFF
--- a/.github/workflows/check-cmdline-ref.yml
+++ b/.github/workflows/check-cmdline-ref.yml
@@ -31,7 +31,13 @@ jobs:
       - name: Build Slang
         run: |
           cmake --preset default --fresh \
-            -DSLANG_SLANG_LLVM_FLAVOR=DISABLE
+            -DSLANG_SLANG_LLVM_FLAVOR=DISABLE \
+            -DSLANG_ENABLE_TESTS=OFF \
+            -DSLANG_ENABLE_EXAMPLES=OFF \
+            -DSLANG_ENABLE_GFX=OFF \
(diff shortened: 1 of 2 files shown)


.github/workflows/check-cmdline-ref.yml
```diff
@@ -31,7 +31,13 @@ jobs:
...
         run: |
           cmake --preset default --fresh \
-            -DSLANG_SLANG_LLVM_FLAVOR=DISABLE
+            -DSLANG_SLANG_LLVM_FLAVOR=DISABLE \
+            -DSLANG_ENABLE_TESTS=OFF \
+            -DSLANG_ENABLE_EXAMPLES=OFF \
+            -DSLANG_ENABLE_GFX=OFF \
+            -DSLANG_ENABLE_SLANGD=OFF \
+            -DSLANG_EXCLUDE_DAWN=ON \
+            -DSLANG_EXCLUDE_TINT=ON
```
jkwak-work, line 40: You can also skip
```
-DSLANG_ENABLE_REPLAYER=OFF
-DSLANG_ENABLE_CUDA=OFF
-DSLANG_ENABLE_OPTIX=OFF
-DSLANG_ENABLE_AFTERMATH=OFF
```

jkwak-work (approved):
looks good to me.

PR #7129, opened 2025-05-15
Allow lambda exprs without captures to coerce to `functype`.

csyonghe:


csyonghe:
Closes #7130.

jkwak-work (approved):
Looks great. I am surprised by how few lines of code can do such a complex thing.

I have tested it on my local machine and it has worked out strange use cases as well.

I was doing a similar thing yesterday but I didn't realize that it had to be done at the AST level. I was working in lower-to-ir, but fortunately I didn't spend too much time there.

csyonghe:
Yeah, I am able to do it with few lines of code because a lot of the building blocks (e.g. function synthesis logic) are already in the compiler, so the real work is to figure out how to string together existing components of the compiler to achieve something conceptually complex.

Issue #7132, opened 2025-05-15
Allow debug layer/validation layer to be disabled in debug build


gtong-nv:
We have quite a lot of valiation layer violations in slang-test, which causes some tests to fail intermittently in the debug build.
A master issue is created to track https://github.com/shader-slang/slang/issues/4798.
Meanwhile, we should allow the VVL to be disabled so that we can turn on the debug testsuite in CI to provide some protection. 
We should remove the "-disable-debug-layers" option after all the violations have been fixed.

Issue #4798, opened 2024-08-06
Enable Vulkan Validation Layer for slang-test by default


jkwak-work:
Currently when slang-test runs, the Vulkan Validation Layer is not used.
It can be manually enabled but it will be better to enable by default.

Also when the Vulkan Validation Layer is enabled and it prints errors, slang-test is currently not treating the case as a failure.
We should change slang-test to fail when the Vulkan Validation Layer printed errors.

jkwak-work:
I am not sure if this is a right link to it, but we probably should enable the debug layer for DX12 as well.
https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-d3d12-debug-layer-gpu-based-validation

gtong-nv:
The VVL is enabled in debug build of slang-test via RHI, 
but if using test server with option`-user-test-server`,  we disabled debug layer even in Debug build, see [here](https://github.com/shader-slang/slang/blob/49de1e8f60c698e9d524befacc988fb06574b234/tools/slang-test/slang-test-main.cpp#L3611)
Right now we have about 60 vk tests failed consistently when running in serial in debug slang-test

sricker-nvidia:
Noting that this seems to have a lot of overlap with https://github.com/shader-slang/slang/issues/6793. I think we should consider closing one of them.

gtong-nv:
As of 5/20, there are 10 distinct VUID violations we should address.

```
VUID Occurrences:
  error: -1110149302 - VUID-vkCmdPipelineBarrier-srcStageMask-09556: 18
  error: -1221877677 - VUID-vkCmdDispatch-None-06479: 14
  error: -1441689486 - VUID-vkCmdPipelineBarrier-dependencyFlags-07891: 9
  error: -1520283006 - VUID-VkShaderModuleCreateInfo-pCode-08737: 6
  error: -1742625639 - VUID-vkCmdCopyBufferToImage-dstImage-07973: 21
  error: -2035756682 - VUID-RuntimeSpirv-shaderSubgroupRotateClustered-09566: 40
  error: -60244330 - VUID-VkShaderModuleCreateInfo-pCode-08742: 39
  error: -609835766 - VUID-RuntimeSpirv-samples-08726: 2
  error: 115483881 - VUID-VkShaderModuleCreateInfo-pCode-08740: 36
  error: 1729412364 - VUID-vkCmdDispatch-format-07753: 10

Total unique VUIDs: 10
```

[...]

PR #7133, opened 2025-05-15
Unify Debug Layer Control Logic and Add Disable Option for Debug Builds

gtong-nv:
This PR refactors and unifies the debug layer control logic in slang-test. A new `-disable-debug-layers` option is introduced, allowing debug builds to skip enabling the validation (debug) layer. This is currently needed to ensure stability in the debug test suite.

Previously, different toggles such as `ENABLE_VALIDATION_LAYER`, `ENABLE_DEBUG_LAYER`, and `debugLayerEnabled` were used inconsistently across different components of slang-test. This PR standardizes the logic by using a single variable, debugLayerEnabled, to control the enabling/disabling of the debug layer internally.

Notes:
By default, the debug/validation layer is enabled in debug builds and is not supported in release builds of slang-test.

Fixes: #7132

### DIFF
--- a/tools/gfx/render.cpp
+++ b/tools/gfx/render.cpp
@@ -9,327 +9,333 @@
 
 namespace gfx
 {
-using namespace Slang;
+    using namespace Slang;
 
-Result SLANG_MCALL createD3D11Device(const IDevice::Desc* desc, IDevice** outDevice);
-Result SLANG_MCALL createD3D12Device(const IDevice::Desc* desc, IDevice** outDevice);
-Result SLANG_MCALL createVKDevice(const IDevice::Desc* desc, IDevice** outDevice);
-Result SLANG_MCALL createMetalDevice(const IDevice::Desc* desc, IDevice** outDevice);
-Result SLANG_MCALL createCUDADevice(const IDevice::Desc* desc, IDevice** outDevice);
-Result SLANG_MCALL createCPUDevice(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createD3D11Device(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createD3D12Device(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createVKDevice(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createMetalDevice(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createCUDADevice(const IDevice::Desc* desc, IDevice** outDevice);
+    Result SLANG_MCALL createCPUDevice(const IDevice::Desc* desc, IDevice** outDevice);
 
-Result SLANG_MCALL getD3D11Adapters(List<AdapterInfo>& outAdapters);
-Result SLANG_MCALL getD3D12Adapters(List<AdapterInfo>& outAdapters);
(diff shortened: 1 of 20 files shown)


tools/gfx/render.cpp
```diff
@@ -428,9 +434,9 @@ extern "C"
         return SLANG_OK;
     }
 
-    SLANG_GFX_API void SLANG_MCALL gfxEnableDebugLayer()
+    SLANG_GFX_API void SLANG_MCALL gfxSetDebugLayerEnabled(bool enable)
```
jkwak-work, line 437: I think the previous name was fine with the additional parameter like,
```
void gfxEnableDebugLayer(bool enable)
```

tools/gfx/vulkan/vk-device.cpp
```diff
@@ -308,7 +309,7 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
         return SLANG_FAIL;
     SLANG_RETURN_ON_FAIL(m_api.initInstanceProcs(instance));
 
-    if ((enableRayTracingValidation || useValidationLayer) && m_api.vkCreateDebugReportCallbackEXT)
+    if ((enableRayTracingValidation && useValidationLayer) && m_api.vkCreateDebugReportCallbackEXT)
```
jkwak-work, line 312: I think this should stay as `||`.
I think there is another command-line argument to enable the ray-tracing validation.
If you change this to `&&`, both options must be enabled to get the same behavior as before.
I think it can be unintuitive like "why is ray tracing validation not working when I use XXX option??"

tools/slang-test/options.cpp
```diff
@@ -88,6 +88,10 @@ static bool _isSubCommand(const char* arg)
         "  -use-test-server               Run tests using test server\n"
         "  -use-fully-isolated-test-server  Run each test in isolated server\n"
         "  -capability <name>             Compile with the given capability\n"
+#if _DEBUG
+        "  -disable-debug-layers          Disable the debug layers (default enabled in debug "
+        "build)\n"
+#endif
```
jkwak-work, line 94: We should have this option also for Release build.
The name is "debug"-layer but it is really "validation-layer" that should work for both release and debug.

If it is not too much work, I think we should rename the all occurrences of "debug-layer" or "debugLayer" to "validationLayer".

jkwak-work (commented):
I have a few comments.

sricker-nvidia:
Are we intending to use this to pass debug CI? If so I'd be a bit concerned that hiding the VUID warnings could lead to legit failures down the line, even if they are benign at the moment.

tools/gfx/vulkan/vk-device.cpp
gtong-nv: You are right. I thought `useValidationLayer` is a master toggle and `enableRayTracingValidation` is the finegrained control. But looks like we do have a dedicated desc `RayTracingValidationDesc` for it. I will change this back

gtong-nv:
> Are we intending to use this to pass debug CI? If so I'd be a bit concerned that hiding the VUID warnings could lead to legit failures down the line, even if they are benign at the moment.

This is a temporary WAR to make the debug testsuite in CI more stable. We will run without "-disable-debug-layers" in CI once we fixed all VVL issues in https://github.com/shader-slang/slang/issues/4798

gtong-nv:
Close as this is merged as part of https://github.com/shader-slang/slang/pull/7085

PR #7142, opened 2025-05-16
Update ubuntu runners to use 22.04

aasgaonkar:
- Use ubuntu-22.04 runners instead of latest
- Leverage glibc-2.35 for release pkg


mkeshavaNV:
@aasgaonkar - The windows error seems unrelated?

aasgaonkar:
\format

aasgaonkar:
used another PR to test out the changes as they require hacks in workflow yaml file to run verification pipeline.
Release yml verified: https://github.com/shader-slang/slang/actions/runs/15285813904/job/42995610641 
centos-gcc release is failing, but it seems to be failing at current TOT as well: https://github.com/shader-slang/slang/actions/runs/15267984287/job/42937041435

.github/workflows/ci.yml
```diff
@@ -123,6 +123,11 @@ jobs:
               sudo apt-get update
               sudo apt-get install -y libx11-dev
           fi
+      - name: Setup Node.js
```
aasgaonkar, line 126: this step is needed so that the nodejs version is newer, else WASM build fail.

.github/workflows/release-linux-glibc-2-35.yml
```diff
@@ -2,11 +2,10 @@ on:
   push:
     tags:
       - "v*"
-
 name: centos7-gcc9 Release
 jobs:
   build:
-    runs-on: ubuntu-latest
```
kaizhangNV, line 9: I don't understand this, this job is using docker, and everything is built in docker, so why it has to change to 22.04?

kaizhangNV: I think what we really should update is the docker image, instead of runner image.

aasgaonkar:
- Use ubuntu-22.04 runners instead of latest
- Leverage glibc-2.35 for release pkg
- delete centos/glibc2.17 release workflow yaml

kaizhangNV (approved):
LGTM

jkwak-work (approved):
Looks good to me.

We may need to bring back the deleted file depending on the response from the customer.

PR #7148, opened 2025-05-16
Make example and unit test more robust

gtong-nv:
- gpu-printing
The original handling seems wrong
`dataCursor - dataEnd` is most likely a negative number and casting it to size_t results in a huge postive number.
The condition `payloadWordCount > size_t(dataCursor - dataEnd)` would not be met.

- slang-unit-test-tool/compileBenchmark.internal
Use the same stage (compute) as defined in the shader - doesn't really matter in current implementation though.
Check the return value.


jkwak-work (approved):
Looks good to me.

It could have been a buffer overflow.

Issue #7058, opened 2025-05-11
[CM2] Implement MapElement with IFunc and Lambda


jkwak-work:
The proposal document describes what needs to be implemented for `MapElement`.
- https://github.com/shader-slang/slang/issues/6398

The difficult part is that it should work with a recent language feature in Slang, Lambda.
And it is little difficult to handle the "this" memory pointer through the SPIRV instruction PerElement.

jkwak-work:
We call the SPIRV function for `MapElement()` as following,
```
OpCooperativeMatrixPerElementOpNV %returnType %mat0 %func %mat1 %mat2
```
The function signature of `%func` must take `uint` for the first two parameters like,
```
int func(uint, uint, int, int, int) // first two arguments are for coordination in the matrix.
```
When we want to send `this` of the functor as one of the parameters, it should be like the following,
```
OpCooperativeMatrixPerElementOpNV %returnType %mat0 %func %funcThis %mat1 %mat2
```
And now the function signature must be,
```
int func(uint, uint, TFuncThis*, int, int, int)
```
Note that `%funcThis` will be sent to the 3rd parameter and not the first.

So the lambda function must synthesize a new function like the following,
```
struct __Slang_lambda_1
{
    __init()
    {
        ... initialization ...
    }
    static int operator()(uint c0, uint c1, This* funcThis, int a, int b, int c)
    {
        return funcThis(c0, c1, a, b, c);
    }
}
```

csyonghe:
Don’t synthesize the function at ast level, instead do it in spirv-legalize pass to synthesize a wrapper function in the IR.

PR #7159, opened 2025-05-19
Implement MapElement for CoopMat

jkwak-work:
With this PR, MapElement works for the following signatures:
 - CoopMat<...>::MapElement(functype(...));
 - CoopMat<...>::MapElement(capturing-lambda);
 - CoopMat<...>::MapElement(not-capturing-lambda);

 - Tuple<CoopMat<...>,...>::MapElement(functype(...));
 - Tuple<CoopMat<...>,...>::MapElement(capturing-lambda);
 - Tuple<CoopMat<...>,...>::MapElement(not-capturing-lambda);

Note: New tests are effectively disabled because currently the CI runners don't have a graphics driver that supports CoopMat2 features.

### DIFF
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -23415,6 +23415,24 @@ struct CoopMat
         };
     }
 
+    __intrinsic_op($(kIROp_CoopMatMapElementIFunc))
+    internal static This __MapElement<
+        TOperator,
+        TFunc : IFunc<T, uint32_t, uint32_t, T>
+    >(
+        This coopMat,
+        TOperator mapOp,
+        Ptr<TFunc> pMapOp
+    );
+
+    This MapElement<
+        TFunc : IFunc<T, uint32_t, uint32_t, T>
+    >(TFunc mapOp)
+    {
+        Ptr<TFunc> pMapOp = &mapOp;
+        return __MapElement(this, mapOp.operator(), pMapOp);
+    }
+
     //
     // Store
     //
@@ -24276,7 +24294,44 @@ CoopMat<T, S, M, N, CoopMatMatrixUse.MatrixAccumulator> coopMatMulAdd<
     };
 }
 
+extension<
+    let S : MemoryScope,
+    let M : int,
+    let N : int,
+    let R : linalg.CoopMatMatrixUse,
+    each Ts : __BuiltinArithmeticType
+> Tuple<expand linalg.CoopMat<each Ts, S, M, N, R>>
+{
+#if 0
+    // The type of `mapOp` gets ConcreteTypePack due to `expand each`
+    // and it cannot match to the regular flat function type.
+    // TODO: We need to fix it.
+    __intrinsic_op($(kIROp_TupleCoopMatMapElementFunc))
+    CoopMat<TR, S, M, N, R> MapElement<
+        TR : __BuiltinArithmeticType,
+    >(functype(uint32_t, uint32_t, expand each Ts)->TR mapOp);
+#endif
+
+    __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
+    static CoopMat<TR, S, M, N, R> __MapElement<
+        TR : __BuiltinArithmeticType,
+        TOperator,
+        TFunc : IFunc<TR, uint32_t, uint32_t, expand each Ts>
+    >(This tuple, TOperator mapOp, Ptr<TFunc> pMapOp);
+
+    [ForceInline]
+    CoopMat<TR, S, M, N, R> MapElement<
+        TR : __BuiltinArithmeticType,
+        TFunc : IFunc<TR, uint32_t, uint32_t, expand each Ts>
+    >(TFunc mapOp)
+    {
+        Ptr<TFunc> pMapOp = &mapOp;
+        return __MapElement<TR>(this, mapOp.operator(), pMapOp);
+    }
+};
+
 } // namespace linalg
+
 
 //
 // Cooperative Vector
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -24303,7 +24303,7 @@ extension<
     each Ts : __BuiltinArithmeticType
 > Tuple<linalg.CoopMat<T, S, M, N, R>, expand linalg.CoopMat<each Ts, S, M, N, R>>
 {
-    __intrinsic_op($(kIROp_TupleCoopMatMapElementFunc))
+    __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
     CoopMat<T, S, M, N, R> MapElement(functype(uint32_t, uint32_t, T, expand each Ts)->T mapOp);
 
     __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -24295,38 +24295,30 @@ CoopMat<T, S, M, N, CoopMatMatrixUse.MatrixAccumulator> coopMatMulAdd<
 }
 
 extension<
+    T : __BuiltinArithmeticType,
     let S : MemoryScope,
     let M : int,
     let N : int,
     let R : linalg.CoopMatMatrixUse,
     each Ts : __BuiltinArithmeticType
-> Tuple<expand linalg.CoopMat<each Ts, S, M, N, R>>
+> Tuple<linalg.CoopMat<T, S, M, N, R>, expand linalg.CoopMat<each Ts, S, M, N, R>>
 {
-#if 0
-    // The type of `mapOp` gets ConcreteTypePack due to `expand each`
-    // and it cannot match to the regular flat function type.
-    // TODO: We need to fix it.
     __intrinsic_op($(kIROp_TupleCoopMatMapElementFunc))
-    CoopMat<TR, S, M, N, R> MapElement<
-        TR : __BuiltinArithmeticType,
-    >(functype(uint32_t, uint32_t, expand each Ts)->TR mapOp);
-#endif
+    CoopMat<T, S, M, N, R> MapElement(functype(uint32_t, uint32_t, T, expand each Ts)->T mapOp);
 
     __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
-    static CoopMat<TR, S, M, N, R> __MapElement<
-        TR : __BuiltinArithmeticType,
+    static CoopMat<T, S, M, N, R> __MapElement<
         TOperator,
-        TFunc : IFunc<TR, uint32_t, uint32_t, expand each Ts>
+        TFunc : IFunc<T, uint32_t, uint32_t, T, expand each Ts>
     >(This tuple, TOperator mapOp, Ptr<TFunc> pMapOp);
 
     [ForceInline]
-    CoopMat<TR, S, M, N, R> MapElement<
-        TR : __BuiltinArithmeticType,
-        TFunc : IFunc<TR, uint32_t, uint32_t, expand each Ts>
+    CoopMat<T, S, M, N, R> MapElement<
+        TFunc : IFunc<T, uint32_t, uint32_t, T, expand each Ts>
     >(TFunc mapOp)
     {
         Ptr<TFunc> pMapOp = &mapOp;
-        return __MapElement<TR>(this, mapOp.operator(), pMapOp);
+        return __MapElement(this, mapOp.operator(), pMapOp);
     }
 };
 
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -23422,15 +23422,14 @@ struct CoopMat
     >(
         This coopMat,
         TOperator mapOp,
-        Ptr<TFunc> pMapOp
+        TFunc mapObj
     );
 
     This MapElement<
         TFunc : IFunc<T, uint32_t, uint32_t, T>
     >(TFunc mapOp)
     {
-        Ptr<TFunc> pMapOp = &mapOp;
-        return __MapElement(this, mapOp.operator(), pMapOp);
+        return __MapElement(this, mapOp.operator(), mapOp);
     }
 
     //
@@ -24310,15 +24309,14 @@ extension<
     static CoopMat<T, S, M, N, R> __MapElement<
         TOperator,
         TFunc : IFunc<T, uint32_t, uint32_t, T, expand each Ts>
-    >(This tuple, TOperator mapOp, Ptr<TFunc> pMapOp);
+    >(This tuple, TOperator mapOp, TFunc mapObj);
 
     [ForceInline]
     CoopMat<T, S, M, N, R> MapElement<
         TFunc : IFunc<T, uint32_t, uint32_t, T, expand each Ts>
     >(TFunc mapOp)
     {
-        Ptr<TFunc> pMapOp = &mapOp;
-        return __MapElement(this, mapOp.operator(), pMapOp);
+        return __MapElement(this, mapOp.operator(), mapOp);
     }
 };
 
--- a/source/slang/hlsl.meta.slang
+++ b/source/slang/hlsl.meta.slang
@@ -24302,10 +24302,10 @@ extension<
     each Ts : __BuiltinArithmeticType
 > Tuple<linalg.CoopMat<T, S, M, N, R>, expand linalg.CoopMat<each Ts, S, M, N, R>>
 {
-    __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
+    __intrinsic_op($(kIROp_CoopMatMapElementIFunc))
     CoopMat<T, S, M, N, R> MapElement(functype(uint32_t, uint32_t, T, expand each Ts)->T mapOp);
 
-    __intrinsic_op($(kIROp_TupleCoopMatMapElementIFunc))
+    __intrinsic_op($(kIROp_CoopMatMapElementIFunc))
     static CoopMat<T, S, M, N, R> __MapElement<
         TOperator,
         TFunc : IFunc<T, uint32_t, uint32_t, T, expand each Ts>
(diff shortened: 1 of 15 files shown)


jkwak-work:
Closes https://github.com/shader-slang/slang/issues/7058

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1261,80 @@ bool SemanticsVisitor::TryUnifyTypes(
         return TryUnifyConjunctionType(constraints, unifyCtx, fst, snd);
     }
 
+    // Use our static helper function to flatten both type packs
+    ShortList<Type*> flattenedFirst;
+    ShortList<Type*> flattenedSecond;
+
+    flattenConcreteTypePack(fst, flattenedFirst);
```
csyonghe, line 1268: I wonder why this is necessary. The concrete type packs should have been unwrapped during subsitution and we should never need to actively unwrap things here.

source/slang/slang-ast-type.cpp
```diff
@@ -578,7 +578,19 @@ Val* FuncType::_substituteImplOverride(ASTBuilder* astBuilder, SubstitutionSet s
     List<Type*> substParamTypes;
     for (Index pp = 0; pp < getParamCount(); pp++)
     {
-        substParamTypes.add(as<Type>(getParamType(pp)->substituteImpl(astBuilder, subst, &diff)));
+        auto substParamType = as<Type>(getParamType(pp)->substituteImpl(astBuilder, subst, &diff));
```
csyonghe, line 581: This should be added to `ConcreteTypePack::_substituteImplOverride` too.

source/slang/slang-check-constraint.cpp
jkwak-work: Hmmm. I will try again to see if I can unwrap in _substituteImpl().
I tried it once but I didn't know better of what was happening.

source/slang/slang-ast-type.cpp
jkwak-work: I will try it

source/slang/hlsl.meta.slang
```diff
@@ -23179,6 +23179,24 @@ struct CoopMat
...
     }
 
+    __intrinsic_op($(kIROp_CoopMatMapElementIFunc))
+    internal static This __MapElement<
+        TOperator,
+        TFunc : IFunc<T, uint32_t, uint32_t, T>
+    >(
+        This coopMat,
+        TOperator mapOp,
+        Ptr<TFunc> pMapOp
```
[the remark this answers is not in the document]
jkwak-work: This one was little tricky for me.
Because I have to use a pointer value to `SpvOpCooperativeMatrixPerElementOpNV`, I needed a pointer. I can probably legalize inside somewhere if that is a more preferred way.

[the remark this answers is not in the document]
jkwak-work: Hmm. You were right. I didn't need this.
Because I was already legalizing it...

source/slang/slang-ast-type.cpp
jkwak-work: I tried it and here is what I am seeing.

When I applied unwrapping of `ConcreateTypePack` inside of `ConcreateTypePack::_substituteImplOverride()`, it removes the "nested" `ConcreateTypePack`, but it still returns one `ConcreteTypePack`, which needs to be still unwrapped here.

source/slang/slang-check-constraint.cpp
jkwak-work: Because my new change unwraps `ConcreteTypePack` in `ConcreteTypePack::_substituteImpl()`, I don't need to unwrap here as you suggested.

jkwak-work:
It appears that my lasted commit is failing the tests. I will fix it tomorrow.

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -2282,6 +2282,33 @@ static LegalVal legalizeGlobalParam(
     IRTypeLegalizationContext* context,
     IRGlobalParam* irGlobalParam);
 
+static LegalVal legalizeCoopMatMapElementIFunc(IRTypeLegalizationContext* context, IRInst* inst)
+{
+    // When the functor object is a lambda with no captures,
+    // it will be removed as a part of the legalization process,
+    // because it is a zero-sized struct.
+    // We need to explicitly remove it from its useser.
```
csyonghe, line 2290: typo: useser

source/slang/slang-ir-legalize-types.cpp
```diff
@@ -2293,6 +2320,10 @@ static LegalVal legalizeInst(IRTypeLegalizationContext* context, IRInst* inst)
     // Special-case certain operations
     switch (inst->getOp())
     {
+    case kIROp_CoopMatMapElementIFunc:
+    case kIROp_TupleCoopMatMapElementIFunc:
```
csyonghe, line 2324: why do we need different opcodes for these two cases?

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1244,88 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+    if (sndTypeCount > 0)
+        sndExpandType = as<ExpandType>(sndTypePack->getElementType(sndTypeCount - 1));
+    else
+        sndExpandType = as<ExpandType>(snd);
+
+    if ((fstTypePack && sndTypePack) || (fstTypePack && sndExpandType) ||
+        (fstExpandType && sndTypePack))
     {
-        if (auto sndTypePack = as<ConcreteTypePack>(snd))
+        if (fstExpandType && sndExpandType)
```
csyonghe, line 1272: this can be moved outside the `if` at 1269?

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1244,88 @@ bool SemanticsVisitor::TryUnifyTypes(
         return TryUnifyConjunctionType(constraints, unifyCtx, fst, snd);
     }
 
-    // If one of the types is a type pack, we need to recursively unify the element types.
-    if (auto fstTypePack = as<ConcreteTypePack>(fst))
+    auto fstTypePack = as<ConcreteTypePack>(fst);
```
csyonghe, line 1247: Can use some more comments here on the unification process.

My understanding of what the code is doing here is that it attempts to do element-wise unification all the way up to the second last element of whatever side that has smaller size. Then unify the last element with the remaining elements of the other side.

However, this may not be correct.

Consider this case:
left = ConcreteTypePack(ExpandType, ExpandType)
right = ConcreteTypePack(int, bool, float, double).

In this case, we shouldn't be mapping the first `ExpandType` to `int` and the second `ExpandType` to `bool, float, double`. Instead, they should evenly divide the second type pack, so we map first ExpandType with `int, bool`, and second ExpandType to `float, double`.

csyonghe: I think we need to generalize this loop a little bit, to make sure it can handle the case where one ConcreteTypePack ends with more than one ExpandType, and the remaining elements from the other side just got evenly divided to map to each ExpandType. If things can't be evenly divided, unification should fail.

We want to make sure this case does work:

```
struct X<B, each T, each U> {
   int getTSize() { return countof(T); }
   int getUSize() { return countof(U); }   
}

func foo<each T, each U>() -> X<bool, expand Ptr<each T>, int, expand Ptr<each U>, float> // unify
{
    return {};
}

void main()
{
    let x = foo<int, float>();
    printf("%d %d\n", x.getTSize(), x.getUSize());  // should print 2 2
}
```

jkwak-work: I see.
Thanks for the explanation.
I wasn't able to see the big picture and I was hoping to get some help like this one.

It makes more sense to me now and I will get the multiple ExpandType to work.

source/slang/slang-check-constraint.cpp
jkwak-work: I think that makes sense to do that.
The only reason I put it there is because the previous code didn't have any handling for it.
And I wasn't sure what to do about that.
But I think we should simply return false when both are ExpandType, because there is no way to resolve the case.

csyonghe: When both sides are ExpandType, it should be as simple as recursively unifying the two expand types, right?

source/slang/slang-ast-type.cpp
csyonghe: yes, we need the unwrapping logic in both FuncType and ConcreteTypePack. I was suggesting adding the logic to ConcreteTypePack, not moving it.

source/slang/slang-check-constraint.cpp
jkwak-work: Hmmm. I think that means I should just remove the line 1272 and 1273. And then handle to resolve the case here.
I will try that.

jkwak-work:
I implemented cases where Slang can handle the `Extend` and `GenericTypePackParamDecl` better.

Unfortunately the code block looks very much duplicated. I spend a few hours to simplify and reduce the duplication but I failed. It was only making the code harder to read. They look very similar but each case has its own uniqueness which made my refactoring ineffective.

A new problem I discovered while implementing it was that I also had to handle `GenericTypePackParamDecl` and treat it like how I was treating `ExpandType`.
Because this PR unwraps ConcreteTypePack in `ConcreteTypePack::_substituteImpl()`, the following case that worked before was no longer working,
```
left = ConcreteTypePack(int, ConcreteTypePack(int, int))
right = ConcreteTypePack(int, GenericTypePackParamDecl).
```
The data above now became,
```
left = ConcreteTypePack(int, int, int)
right = ConcreteTypePack(int, GenericTypePackParamDecl).
```
When the last two types were wrapped with ConcreteTypePack like `ConcreteTypePack(int, int)`, it was easier to call the function itself recursively and let it handle a pair of types; one in `ConcreteTypePack` and another in `GenericTypePackParamDecl`.
But now it became more dynamic sized matching of "int, int" against "GenericTypePackParamDecl".
That's where I had to handle the case with the new for-loop; rather than calling the function recursively.
This is same as how `ExpandType` was handled.

source/slang/slang-check-constraint.cpp
```diff
@@ -1210,6 +1210,58 @@ void SemanticsVisitor::maybeUnifyUnconstraintIntParam(
...
+            {
+                if (typePackParamDecl->parentDecl == genericDecl)
+                    return typePackParamDecl;
+            }
+        }
+    }
+    return nullptr;
+}
+
+static bool isExpandableType(Type* type, GenericDecl* genericDecl)
```
csyonghe, line 1229: There is already a `isAbstractTypePack` for this.

source/slang/slang-check-constraint.cpp
```diff
@@ -1210,6 +1210,58 @@ void SemanticsVisitor::maybeUnifyUnconstraintIntParam(
...
+{
+    return as<ExpandType>(type) || asGenericTypePackParamDecl(type, genericDecl);
+}
+
+// Static helper function to unwrap a ConcreteTypePack and count how many ExpandType were in.
+static int UnwrapConcreteTypePackAndCountExpandable(
```
csyonghe, line 1235: `unwrapConcreteTypePack(..., int& outAbstractTypePackElementCount)`

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1296,217 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+        // Consider this case:
+        // left = ConcreteTypePack(ExpandType, ExpandType)
+        // right = ConcreteTypePack(int, bool, float, double).
+        //
+        // In this case, we shouldn't be mapping the first ExpandType to int and the second
+        // ExpandType to bool, float, double. Instead, they should evenly divide the second type
+        // pack, so we map first ExpandType with int, bool, and second ExpandType to float, double.
+        //
+        int howManyTypesPerEachExpand = 1;
+        int countDifference = fstCount - sndCount;
```
csyonghe, line 1328: I think this should be `fstExpandableCount - sndExpandableCount`, and always map the side with less expandable to the side with more.

Consider:
```
first = (Expand, Expand)
snd = ()
```
This should still work.

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1296,217 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+        // Unwrap ConcreteTypePack to make the logic simpler
+        ShortList<Type*> flattenedFirst;
+        ShortList<Type*> flattenedSecond;
+
+        int fstExpandableCount =
+            UnwrapConcreteTypePackAndCountExpandable(fst, flattenedFirst, constraints.genericDecl);
+        int sndExpandableCount =
+            UnwrapConcreteTypePackAndCountExpandable(snd, flattenedSecond, constraints.genericDecl);
+
+        int fstCount = (int)flattenedFirst.getCount();
```
csyonghe, line 1313: Now that you are here, you can have a `bool matchTypeArgMapping(ShortList firstTypes, ShortList sndTypes, out ShortList<IndexSpanPair> mapFirstToSecond)` that returns the mapped pairs of elements in first and second. If mapping fails, the function returns `false`.

Then we can just iterate through `mapFirstToSecond` and recursively apply unification. Each `IndexSpanPair` should be:
```
struct IndexSpanPair { IndexSpan first, second; };
```
and one of first.count or second.count must be 1.



We would then try to map first to second, if that succeeds, great. If not, we try map second to first, if that also fails, return false.

`matchTypeArgMapping` should fail if first contains more expandable types than second.

This will allow you to unify all the cases down below and avoid the duplication.

The existing `matchArgumentsToParams` is implementing this for match call arguments to params, and is following the same idea.

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1296,217 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+        // left = ConcreteTypePack(ExpandType, ExpandType)
+        // right = ConcreteTypePack(int, bool, float, double).
+        //
+        // In this case, we shouldn't be mapping the first ExpandType to int and the second
+        // ExpandType to bool, float, double. Instead, they should evenly divide the second type
+        // pack, so we map first ExpandType with int, bool, and second ExpandType to float, double.
+        //
+        int howManyTypesPerEachExpand = 1;
+        int countDifference = fstCount - sndCount;
+        if (countDifference > 0)
```
csyonghe, line 1329: I think the correct way to do is to match the side less expandables to the side with more expandables.

Consider the following case:

```
first = (Expand, Expand)
second = ()
```

This should still success, resulting first Expand to match to (index = 0, count = 0) and second Expand to match to (index = 0, count = 0)

jkwak-work: I see. I will fix it.

source/slang/slang-check-constraint.cpp
jkwak-work: This suggestion worked out very well.
Thank you very much.

jkwak-work:
This PR is ready for another review, @csyonghe

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1430,139 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+
+            if (firstIsSingle)
             {
-                if (!TryUnifyTypes(
-                        constraints,
-                        unifyCtx,
-                        QualType(fstTypePack->getElementType(i), fst.isLeftValue),
-                        QualType(sndTypePack->getElementType(i), snd.isLeftValue)))
-                    return false;
+                singleType = flattenedFirst[mapping.first.index];
```
csyonghe, line 1474: We can use the astBuilder on the semantic visitor to create temporary `ConcreteTypePack` from the mapped range of type elements, and then recursively call `TryUnifyTypes` using the constructed type pack to get rid of the duplication.

source/slang/slang-check-constraint.cpp
```diff
@@ -1210,6 +1210,192 @@ void SemanticsVisitor::maybeUnifyUnconstraintIntParam(
...
+    // Unwrap and flatten the types
+    ShortList<Type*>& firstTypes = outFlattenedFirst;
+    ShortList<Type*>& secondTypes = outFlattenedSecond;
+
+    // Count expandable types as we unwrap
+    int firstExpandableCount = 0;
+    int secondExpandableCount = 0;
+
+    // Unwrap first type
+    if (auto concretePack = as<ConcreteTypePack>(firstType))
```
csyonghe, line 1273: can use a single helper function to unwrap a type pack to a list.

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1423,152 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+            {
+                // Single GenericTypePackParamDecl unifies with multiple types as a pack - one pair
+                // total For GenericTypePackParamDecl, singleType always goes to first argument (no
+                // swapping)
+                allowSwap = false;
+
+                ShortList<Type*> multipleTypes;
+                for (Index i = 0; i < multipleCount; ++i)
+                {
+                    multipleTypes.add((*multipleSideList)[multipleStartIndex + i]);
```
csyonghe, line 1523: Instead of copying, you can form an ArrayView directly from `multipleSideList`.

source/slang/slang-check-constraint.cpp
```diff
@@ -1244,58 +1423,152 @@ bool SemanticsVisitor::TryUnifyTypes(
...
+
+            // Get the single type and the range for multiple types
+            Type* singleType;
+            Index multipleStartIndex;
+            Index multipleCount;
+            ShortList<Type*>* multipleSideList;
+            bool singleSideIsLeftValue;
+            bool multipleSideIsLeftValue;
+
+            if (firstIsSingle)
```
csyonghe, line 1465: Now that we already have the mapping, why isn't the logic here just as simple as:

1. leftType = firstIsSingle?flattenedFirst[mapping.first.index]:createTypePack(mapping.first);
2. rightType = secondIsSingle?flattenedFirst[mapping.second.index]:createTypePack(mapping.second);
3. TryUnifyTypes(leftType, rightType)

csyonghe: There are details about propagating isLValue, but you can just grab that from the `flattenedFirst[mapping.first.index]` in both single and multiple cases.

csyonghe (approved):
Looks good!

PR #7173, opened 2025-05-20
Add inverse hyperbolic derivatives

fairywreath:
Adds derivatives for `asinh`, `acosh` and `atanh`. They are defined [here](https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions#Derivatives).


csyonghe (dismissed):
Looks good, however I didn’t double check check the actual derivative definition, and I assume they are correct.

csyonghe (approved):

PR #7175, opened 2025-05-20
Fix retry logic and skip high intermittent test failure

gtong-nv:
- Fix the retry logic. Before it doesn't add to failedUnitTests if the test fails on retry, making the 2nd retry attempt impossible. Remove the bool check.
- Add print to show when a test is about to run; making it easy to identify which test fails next.
-  Update the CI yml file to allow manual dispatch, so that we can test in its own branch before creating PR.
-  Mark `slang-unit-test-tool/RecordReplay.internal` expected failure until we get 90% passing for it in CI.

#7148

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
             full-gpu-tests: true
             runs-on: [Windows, self-hosted, "GCP-T4"]
             has-gpu: true
-            server-count: 8
+            server-count: 2
       fail-fast: false
     runs-on: ${{ matrix.runs-on }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -191,6 +192,7 @@ jobs:
               -category ${{ matrix.test-category }} \
               -api all-cpu \
               -expected-failure-list tests/expected-failure-github.txt \
+              -expected-failure-list tests/expected-failure-record-replay-tests.txt \
```
sricker-nvidia, line 195: Are we able to specify -expected-failure-list multiple times and have each list appended or will the previous -expected-failure-list be overwritten here?

gtong-nv: I think it's allowed. I followed patten below and tested locally and it worked fine.

sricker-nvidia (approved):
Looks ok to me

PR #7178, opened 2025-05-20
Add command line option for separate debug info

jarcherNV:
Add command line arg -separate-debug-info which, if provided, produces both a .spv and a .dbg.spv file. The .dbg.spv file contains full debug info and the .spv file has all debug info stripped out.


source/slang/slang-emit.cpp
```diff
@@ -122,6 +122,7 @@
 #include "slang-vm-bytecode.h"
 
 #include <assert.h>
+#include <unordered_set>
```
csyonghe, line 125: no need for std, use `HashSet<uint>` instead.

source/slang/slang-emit.cpp
```diff
@@ -2091,22 +2092,103 @@ SlangResult emitSPIRVFromIR(
...
+    // Standard debug opcodes to strip out:
+    // OpName, OpMemberName, OpString, OpLine, OpNoLine.
+    static const std::unordered_set<uint16_t> debugOpCodes = {
+        5, 6, 7, 8, 317
+    };
+    // If the instruction is an extended instruction, then we also need
+    // to check if the instruction number is for a debug instruction as
+    // listed in slang-emit-spirv-ops-debug-info-ext.h
+    static const std::unordered_set<uint32_t> debugExtInstNumbers = {
+        1, 35, 102, 103, 107, 20, 101, 8, 10, 11, 5, 2, 6, 108, 3, 4, 23, 24,
```
csyonghe, line 2112: Can you use SPIRV enums instead of raw values here?

source/slang/slang-emit.cpp
jarcherNV: Done.

jarcherNV:
I have updated this to include a DebugBuildIdentifier instruction, which is kept in both output spv files.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -11959,6 +11971,25 @@ static void ensureAllDeclsRec(IRGenContext* context, Decl* decl)
...
+    ComPtr<ISlangBlob> hashBlob;
+    module->getEntryPointHash(0, 0, hashBlob.writeRef());
+
+    const uint8_t* data = reinterpret_cast<const uint8_t*>(hashBlob->getBufferPointer());
+    size_t size = hashBlob->getBufferSize();
+    StringBuilder sb;
+    char buf[3]; // 2 hex digits + null terminator
+    for (size_t i = 0; i < size; ++i)
+    {
+        snprintf(buf, sizeof(buf), "%02x", data[i]);
```
kaizhangNV: why can't we use "sb<<data" syntax?

source/slang/slang-ir.cpp
```diff
@@ -3393,20 +3393,40 @@ IRInst* IRBuilder::emitDebugSource(UnownedStringSlice fileName, UnownedStringSli
...
     return emitIntrinsicInst(getVoidType(), kIROp_DebugSource, 2, args);
 }
+IRInst* IRBuilder::emitDebugBuildIdentifier(UnownedStringSlice buildIdentifier, IRIntegerValue flags)
+{
+    IRInst* args[] = {getStringValue(buildIdentifier), getIntValue(flags)};
+    return emitIntrinsicInst(getVoidType(), kIROp_DebugBuildIdentifier, 2, args);
+}
 IRInst* IRBuilder::emitDebugLine(
     IRInst* source,
+    IRInst* debugBuildIdentifier,
```
kaizhangNV: As discussed, this should not appear in DebugLine.
`debugBuildIdentifier` should be an separate inst appear at debug info section, which DebugLine will be non-semantics inst interleaved with normal spirv inst.

source/slang/slang-ir-insts.h
```diff
@@ -3438,6 +3438,13 @@ struct IRDebugSource : IRInst
     IRInst* getSource() { return getOperand(1); }
 };
 
+struct IRDebugBuildIdentifier : IRInst
+{
+    IR_LEAF_ISA(DebugBuildIdentifier)
+    IRInst* getBuildIdentifier() { return getOperand(0); }
+    IRInst* getFlags() { return getOperand(1); }
```
kaizhangNV, line 3469: What's the usage of this flag?

jarcherNV: I don't know how we should use this flag, but I think it's part of the spec for the instruction: https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugBuildIdentifier

source/slang/slang-lower-to-ir.cpp
jarcherNV: Perhaps we can. I'll see if I can clean up this logic in a future version.

jarcherNV:
Current version is my WIP implementation, which just emits the DebugBuildIdentifier IR instruction by itself without using DebugLine. The IR for this is present through all the IR stages and emitOpDebugBuildIdentifier is called inside slang-emit-spirv.cpp but the instruction seems to not be present in the SPIRV output (although wierdly the string containing the hash is). I will continue debugging this.

source/slang/slang-lower-to-ir.cpp
jarcherNV: I've played around with this, but I'm not too sure how to use that to give a hexadecimal format.

jarcherNV:
Current version should be ok to review now, but it relies on a change to spirv-tools otherwise the added DebugBuildIdentifier instruction will be optimised out by dead code elimination. I will prepare a fix for this.

jarcherNV:
I have prepared a fix to prevent elimination of the new instruction here: https://github.com/KhronosGroup/SPIRV-Tools/pull/6166

source/slang/slang-ir-insts.h
kaizhangNV: OK, I see.

source/slang/slang-emit.cpp
```diff
@@ -2101,22 +2102,164 @@ SlangResult emitSPIRVFromIR(
...
+        return SLANG_FAIL;
 
-    auto irModule = linkedIR.module;
-    auto irEntryPoints = linkedIR.entryPoints;
+    const uint32_t* spirvWords = reinterpret_cast<const uint32_t*>(spirvBlob->getBufferPointer());
+    size_t wordCount = spirvBlob->getBufferSize() / sizeof(uint32_t);
+
+    // The first 5 uints are header data which does not need to be checked
+    // for debug instructions, but should still be added to the new artifact.
+    size_t firstWord = 5;
```
kaizhangNV: For those magic numbers, are there any macros/const var define for them?
I'm a little scared using those number literal.

source/slang/slang-emit.cpp
```diff
@@ -2101,22 +2102,164 @@ SlangResult emitSPIRVFromIR(
...
+    // Standard debug opcodes to strip out:
+    // OpName, OpMemberName, OpString, OpLine, OpNoLine.
+    static const uint16_t debugOpCodeVals[] = {
+        SpvOpName, SpvOpMemberName, SpvOpString, SpvOpLine, SpvOpNoLine
+    };
+    // If the instruction is an extended instruction, then we also need
+    // to check if the instruction number is for a debug instruction as
+    // listed in slang-emit-spirv-ops-debug-info-ext.h
+    static const uint32_t debugExtInstVals[] = {
+        NonSemanticShaderDebugInfo100DebugCompilationUnit,
```
kaizhangNV, line 2268: Is this list matched the strip logic in spirv-opt?
I want to make sure we are doing the same thing as spirv-opt.

source/slang/slang-ir-inst-defs.h
```diff
@@ -1372,6 +1372,7 @@ INST(DebugFunction, DebugFunction, 5, 0)
 INST(DebugInlinedVariable, DebugInlinedVariable, 2, 0)
 INST(DebugScope, DebugScope, 2, 0)
 INST(DebugNoScope, DebugNoScope, 1, 0)
+INST(DebugBuildIdentifier, DebugBuildIdentifier, 2, HOISTABLE)
```
kaizhangNV: why this is HOISTABLE?
This inst will only be a global inst.

source/slang/slang-lower-to-ir.cpp
kaizhangNV: sb << CharUtil::getHexChar(data[i])

this might be working.

not saying what you do is wrong, it's just that we try to avoid calling any c/c++ lib functions through our code base, which could break ABI.

kaizhangNV: Or you can use
`StringUtil::makeStringWithFormat(... )`

So always check source/core/ to see what we already have first.

jarcherNV: Ahh yeah I missed those. I'll update this.

source/slang/slang-ir.cpp
jarcherNV: Also as discussed, just leaving a note that this is fixed now.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -12008,6 +12030,16 @@ RefPtr<IRModule> generateIRForTranslationUnit(
         }
     }
 
+    // Set the debug build identifier if using separate debug info, to be emitted later.
+    if (context->useSeparateDebugInfo)
```
csyonghe: I think we can include the identifier for all configs, and not just for separate debug info.

source/slang/slang-options.cpp
```diff
@@ -3062,6 +3066,19 @@ SlangResult OptionsParser::_parse(int argc, char const* const* argv)
...
+        case OptionKind::EmitSeparateDebug:
+            {
+                // This will emit a separate debug file, containing all debug info in
+                // a .dbg.spv file.
+                // This needs to override any obfuscation and debug info flags.
+                // We now want to output SPIRV with full debug info and no obfuscation.
+                // The main output SPIRV will have all debug info stripped, and obfuscation
+                // is not needed as this only changes names which will now not be present.
+                m_compileRequest->setDebugInfoLevel(SLANG_DEBUG_INFO_LEVEL_MAXIMAL);
+                linkage->m_optionSet.set(OptionKind::Obfuscate, false);
```
csyonghe: Let's not do the silent translation here.

csyonghe (commented):
We should also consider adding a unit test to look for the debug identifier in the spirv that has the debug info.

source/slang/slang-ir-inst-defs.h
jarcherNV: I guess it doesn't need to be hoistable. Ultimately I think it ends up being emitted next to the DebugSource instruction either way. That particular instruction is hoistable so I thought we may want to do something similar for this. I will change it.

source/slang/slang-emit.cpp
jarcherNV: It looks like spirv-opt excludes some more instructions than what I included. I will update this.

source/slang/slang-emit.cpp
jarcherNV: As discussed in the code review meeting, I will update this logic to use a helper class, which will hopefully not use magic numbers.

source/slang/slang-options.cpp
jarcherNV: Fixed.

source/slang/slang-lower-to-ir.cpp
jarcherNV: Fixed.

jarcherNV:
I have updated I have updated this with the following:
- Updated SPIRV parsing to use a helper class rather than the previous duplicated loops.
- Added a unit test using the Slang API.
- Added a new API function to allow getting the debug data and identifier.

jarcherNV:
I have updated this again as I forgot to include a comment describing the behavior of the added function.

source/slang/slang-emit.cpp
```diff
@@ -2104,22 +2105,250 @@ SlangResult emitSPIRVFromIR(
...
+        // extract the string value.
+        String getStringFromInst() const
+        {
+            String result;
+            for (uint32_t i = 1; i < getWordCountForInst(); i++)
+            {
+                SpvWord op = getOperand(i);
+                for (int b = 0; b < 4; ++b)
+                {
+                    char c = (char)((op >> (b * 8)) & 0xFF);
```
kaizhangNV, line 2174: can you change this to uint8_t instead?

jarcherNV: I could, but I would need to cast it to a char regardless so that it is appended as a char and not an integer value.

jarcherNV:
I have updated this with API changes suggested by Yong. I was unable to have the new interface extend ICastable as that caused compile errors when IArtifact extended it. I'm not sure if we need this behavior.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -12006,6 +12025,11 @@ RefPtr<IRModule> generateIRForTranslationUnit(
                 source->getContent());
             context->shared->mapSourceFileToDebugSourceInst.add(source, debugSource);
         }
+
+        // Build identifier is a hash of the source code and compile options.
+        String buildIdentifier = getBuildIdentifierString(translationUnit->getModule());
+        int buildIdentifierFlags = 0;
+        builder->emitDebugBuildIdentifier(buildIdentifier.getUnownedSlice(), buildIdentifierFlags);
```
csyonghe: Producing a debug identifier for each module during lower-to-ir, and then include them during linking process will create ambiguity when modules are compiled separately.

consider:
```
module E;
import moduleA;
void main() {}
```

If `moduleA` is compiled to `moduleA.slang-module` first, the `slang-module` will contain a debug identifier.
Now when I compile module `E`, we will get another debug identifier.
When linking the two modules, we will include two debug identifiers. Which one should triumph?

The cause of this issue is that it doesn't make sense to include a debug identifier without doing a full spirv compilation.
I think the right solution is, instead of generating a debug build identifier during lower-to-ir, it should be inserted during linkIR, and you get the identifier by calling `codeGenContext->targetProgram->buildHash()` there.

include/slang.h
```diff
@@ -4163,9 +4164,29 @@ struct IMetadata : public ISlangCastable
...
+    */
+    virtual const char* getDebugBuildIdentifier() = 0;
 };
     #define SLANG_UUID_IMetadata IMetadata::getTypeGuid()
 
+/** Compile result for storing and retrieving multiple output blobs.
+    This is needed for features such as separate debug compilation which
+    output both base and debug spirv.
+ */
+struct ICompileResult
```
kaizhangNV: change to class.

jarcherNV:
Current version should fix most test failures except tests\language-feature\modules\import-in-include\c.slang which I am still trying to resolve.

include/slang.h
jarcherNV: The others around it are structs so I was following that when I added this. Are you sure I should change this?

source/slang/slang-lower-to-ir.cpp
jarcherNV: This makes sense. I have moved the logic for this to after the linking. Unfortunately there is still a failing test. I have sync'd with Joel about this and while he is not blocked yet, he will be soon. We think it may be best to submit this with the check to only emit the instruction if separate debug info is requested so as not to break anything and allow him to work with the new API, and we create a follow-up issue to unconditionally emit the instruction, if that's ok with you.

source/slang/slang-emit.cpp
jarcherNV: Confirming that I have updated this.

source/slang/slang-lower-to-ir.cpp
```diff
@@ -603,6 +603,7 @@ struct IRGenContext
     FunctionDeclBase* funcDecl;
 
     bool includeDebugInfo = false;
+    bool useSeparateDebugInfo = false;
```
csyonghe: this should no longer be needed anymore.

jarcherNV: Fixed.

csyonghe (dismissed):

include/slang.h
```diff
@@ -4163,9 +4164,28 @@ struct IMetadata : public ISlangCastable
...
+    */
+    virtual const char* getDebugBuildIdentifier() = 0;
 };
     #define SLANG_UUID_IMetadata IMetadata::getTypeGuid()
 
+/** Compile result for storing and retrieving multiple output blobs.
+    This is needed for features such as separate debug compilation which
+    output both base and debug spirv.
+ */
+struct ICompileResult : public ISlangCastable
```
csyonghe, line 4179: Should this be `ISlangUnknown`?

jarcherNV: Unfortunately that did not work. The IArtifact extends ISlangCastable, and I ended up with a lot of problems trying to have that extend both ISlangCastable and ICompileResult (when the ICompileResult was extending either of Unknown/Castable). The easiest fix seemed to be to have this extend ISlangCastable, and then have IArtifact only extend this.

kaizhangNV (approved):
LGTM now, you might need Yong to take another look.

jkwak-work:
This turned out to be a breaking change.
I changed the label accordingly.

PR #7203, opened 2025-05-21
List all source files in debug source file list

lujinwangnv:
The source file which does not participate in the line table is missing from the debug source file list. Always copy IRDebugSource instruction in linkIR() to fix the issue.


source/slang/slang-ir-link.cpp
```diff
@@ -2158,14 +2158,17 @@ LinkedIR linkIR(CodeGenContext* codeGenContext)
...
     for (IRModule* irModule : userModules)
     {
         for (auto inst : irModule->getGlobalInsts())
         {
-            if (auto bindInst = as<IRBindGlobalGenericParam>(inst))
+            if (as<IRBindGlobalGenericParam>(inst) ||
+                as<IRDebugSource>(inst))
             {
-                cloneValue(context, bindInst);
+                cloneValue(context, inst);
```
jkwak-work, line 2171: I don't think we should do this when the option is `-g0`, is it?

lujinwangnv: How do we check -gX in the slang code?

source/slang/slang-ir-link.cpp
```diff
@@ -2158,14 +2158,17 @@ LinkedIR linkIR(CodeGenContext* codeGenContext)
...
     //
+    // Need to list all source files in the debug source file list,
+    // regardless if the source files participate in the line table or not.
 
     for (IRModule* irModule : userModules)
     {
         for (auto inst : irModule->getGlobalInsts())
         {
-            if (auto bindInst = as<IRBindGlobalGenericParam>(inst))
+            if (as<IRBindGlobalGenericParam>(inst) ||
```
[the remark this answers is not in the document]
lujinwangnv: Good point, updated the code.

source/slang/slang-ir-link.cpp
```diff
@@ -2146,26 +2146,33 @@ LinkedIR linkIR(CodeGenContext* codeGenContext)
...
+                // as stand-alone global-scope instructions in the IR module for
+                // `SpecializedComponentType`s. These instructions are required for
+                // correct codegen, and so we must make sure to copy them all over,
+                // even though they are not directly referenced.
+                //
+                // TODO: We should change these to decorations, akin to how
+                // `[bindExistentialSlots(...)]` works, so that they can be attached
+                // to the relevant parameters and cloned via `cloneExtraDecorations`.
+                // In the long run we do not want to *ever* iterate over all the
+                // instructions in all the input modules.
```
jkwak-work, line 2170: Can we have `[[fallthrough]];` ?
The comment is making it harder to understand that the logic here is to fall-through.

lujinwangnv: yes, updated.

PR #7167, opened 2025-05-19
Update build to allow setting external lua path

lujinwangnv:
Update the build to allow setting user-specific path for the external module lua.

PR #7207, opened 2025-05-22
Fix lua header file path

lujinwangnv:
Add two missed files in #7167

### DIFF
--- a/tools/slang-fiddle/slang-fiddle-script.cpp
+++ b/tools/slang-fiddle/slang-fiddle-script.cpp
@@ -1,9 +1,9 @@
 // slang-fiddle-script.cpp
 #include "slang-fiddle-script.h"
 
-#include "../external/lua/lapi.h"
-#include "../external/lua/lauxlib.h"
-#include "../external/lua/lualib.h"
+#include "lapi.h"
+#include "lauxlib.h"
+#include "lualib.h"
 
 namespace fiddle
 {
--- a/tools/slang-fiddle/slang-fiddle-script.h
+++ b/tools/slang-fiddle/slang-fiddle-script.h
@@ -1,8 +1,8 @@
 // slang-fiddle-script.h
 #pragma once
 
-#include "../external/lua/lapi.h"
-#include "../external/lua/lauxlib.h"
+#include "lapi.h"
+#include "lauxlib.h"
 #include "compiler-core/slang-source-loc.h"
 #include "core/slang-list.h"
 #include "core/slang-string.h"


lujinwangnv:
Is slang-fiddle-script.cpp not built by default? I am surprised this did not cause any build failure on github slang.

tools/slang-fiddle/slang-fiddle-script.cpp
```diff
@@ -1,9 +1,9 @@
 // slang-fiddle-script.cpp
 #include "slang-fiddle-script.h"
 
-#include "../external/lua/lapi.h"
```
csyonghe, line 4: can we at least keep the `lua` folder here? We don't want to add everything in a third party library in the default include search list as it can introduce conflicts.

lujinwangnv: yes, added lua/ back. Thanks

csyonghe: What I was saying is we shouldn't be adding `external/lua` to the default search list. Instead, just add "external" to it, so we can write `#include "lua/lapi.h"`, no `"../"`.

csyonghe: I don't know the logic behind removing "external", but in general it is a good practice to keep external references explicit, by naming "external" in `#include` directive.

lujinwangnv: I followed the code which were added (SLANG_OVERRIDE_xxx_PATH) to override path of tools in external/ to define SLANG_OVERRIDE_LUA_PATH. We put those tools in a third_party/ folder outside slang/, SLANG_OVERRIDE_xxx_PATH has been pointed to path/to/third_party/xxx. If you prefer to leave "xxx/" in #include, I think we have to use "../xxx/".

csyonghe: I think all of the OVERRIDE_PATHs shouldn’t include xxx in it. They should all be pointing to that third_party folder outside slang, so slang can do include “lua/lua.h”.

Adding all root folders of external projects into the search list is dangerous, and we shouldn’t be doing that.

csyonghe: If both libraryA and libraryB has something like “math_util.h” or “com_ptr.h” in its root directory, this setup will break compile. Since we don’t have control on third party code, we shouldn’t rely on them following any name convention on their file names, and assume file names will clash.

lujinwangnv: Ok, I can define a SLANG_OVERRIDE_EXTERNAL_PATH and get rid of all individual SLANG_OVERRIDE_xxx_PATH and add xxx back to #include. Should I do that?

csyonghe: That will work, as long as you only need to configure this for all external dependencies consistently, and don’t require setting the path differently for each external dependency, or turn on/off overriding for each individual dependency.

csyonghe: I think some of our users may want to link against llvm and spirv headers to a different folder but keep other dependencies default.

It is probably safest to just remove xxx from all the override paths for now.

lujinwangnv: Ok, I will do that, it is a bit weird to not include /xxx in SLANG_OVERRIDE_xxx_PATH though.

csyonghe (approved):

csyonghe (approved):

PR #7213, opened 2025-05-22
Make sizeof(T) & alignof(T) of generic types work as compile-time constants

### LINKED ISSUE #7196: sizeof(T) of a generic type does not produce a compile-time constant
Hello, for some reason this case fails to compile even though the generic type should be known at compile time on use:

```c
struct Foo1<T> {
    uint8_t data[sizeof(uint)]; // ok
};

struct Foo2<let N : uint> {
    uint8_t data[N]; // ok
};

struct Foo3<T> {
    uint8_t data[sizeof(T)]; // error 39999: expression does not evaluate to a compile-time constant
};
```

juliusikkala:
Fixes #7196.

### DIFF
--- a/source/slang/slang-ast-val.cpp
+++ b/source/slang/slang-ast-val.cpp
@@ -3,6 +3,7 @@
 
 #include "slang-ast-builder.h"
 #include "slang-ast-dispatch.h"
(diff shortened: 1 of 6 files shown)


csyonghe (approved):
Looks good!

PR #7214, opened 2025-05-22
Add default constructor for Ptr type

fairywreath:
Make `Ptr` type a c-tyle type to be able to initialize pointer variables with the `{}` syntax.

### DIFF
--- a/source/slang/core.meta.slang
+++ b/source/slang/core.meta.slang
@@ -1279,6 +1279,12 @@ struct Ptr
         [nonmutating]
         ref;
     }
+
+    [ForceInline]
+    __init()
+    {
+        return __default<Ptr<T, addrSpace>>();
+    }
 };
 
 //@hidden:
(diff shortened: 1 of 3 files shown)


source/slang/core.meta.slang
```diff
@@ -1279,6 +1279,12 @@ struct Ptr
         [nonmutating]
         ref;
     }
+
+    [ForceInline]
```
csyonghe, line 1283: Instead of doing this, we should just make pointer type a C-Style type so they can be default initialized to 0.

csyonghe: The difference is that is ptr is a c-style types, then the following will also work:
```
struct M
{
       int* x;
};

M m = {}; // should init.
```
If we only have `__init` on `Ptr` type, at the moment `x` won't be initialized unless it is defined as `int* x = {}`.

csyonghe (approved):
Looks good!

PR #7218, opened 2025-05-23
Enable slang-rhi integration testing

aasgaonkar:


.github/workflows/ci.yml
```diff
@@ -246,6 +246,14 @@ jobs:
             -expected-failure-list tests/expected-failure.txt \
             -skip-reference-image-generation \
             -show-adapter-info
+      - name: slang-rhi tests
+        if: steps.filter.outputs.should-run == 'true' && matrix.platform != 'wasm' && matrix.full-gpu-tests
+        run: |
+          cd external/slang-rhi
+          cmake --preset msvc -S . -B build
```
aasgaonkar, line 253: there are 2 presets in rhi CI - msvc and clang. I picked msvc here, but we can do clang if needed as well. I think wrt unit testing, they are the same. but we may want coverage with clang usage?

csyonghe: just msvc is fine.

csyonghe: We are already building slang-rhi during slang build, so I think the build step is unnecessary.

aasgaonkar: doesn't seem like there is a build folder, I tried just the ctest cmd. https://github.com/shader-slang/slang/actions/runs/15201660823/job/42756807464?pr=7218 
what am I missing?

.github/workflows/ci.yml
```diff
@@ -246,6 +246,10 @@ jobs:
             -expected-failure-list tests/expected-failure.txt \
             -skip-reference-image-generation \
             -show-adapter-info
+      - name: slang-rhi tests
+        if: steps.filter.outputs.should-run == 'true' && matrix.platform != 'wasm' && matrix.full-gpu-tests
+        run: |
+          ctest --test-dir ${{matrix.config}}/bin --build-config ${{matrix.config}} --verbose
```
skallweitNV, line 252: you could instead just run `slang-rhi-tests`

aasgaonkar: done in new patches

Issue #6982, opened 2025-05-02
partial-resident-texture-combined fails with dx12


jkwak-work:
The following test is always failing with dx12 backend
```
tests/hlsl-intrinsic/texture/partial-resident-texture-combined.slang (dx12)
```
Instead of keeping it in "expected-failure" list, we are disabling the test until it can be properly fixed.

This issue is to track the disabled test so that we don't keep it disabled forever.

When I compiled with `slangc -target dxil`, I got the following error message,
```
$ slangc.exe -target dxil -profile cs_6_5 -entry computeMain
dxc 1.9: hlsl.hlsl(986): error :  no matching member function for call to 'Load'
dxc 1.9: note :     float4 _S126 = s2D_texture_0.Load(s2D_sampler_0, iuvs_0, offset_0, status_3);
dxc 1.9: note :                    ~~~~~~~~~~~~~~^~~~
dxc 1.9: hlsl.hlsl(986): error :  candidate function template not viable: requires single argument 'x', but 4 arguments were provided
dxc 1.9: hlsl.hlsl(986): error :  candidate function template not viable: requires 2 arguments, but 4 were provided
dxc 1.9: hlsl.hlsl(986): error :  candidate function template not viable: requires 3 arguments, but 4 were provided
dxc 1.9: hlsl.hlsl(1000): error :  no matching member function for call to 'Load'
dxc 1.9: note :         float4 _S129 = s2DMS_texture_0.Load(s2DMS_sampler_0, iuv_0, int(0), offset_0, status_3);
dxc 1.9: note :                        ~~~~~~~~~~~~~~~~^~~~
dxc 1.9: hlsl.hlsl(1000): error :  candidate function template not viable: requires 2 arguments, but 5 were provided
dxc 1.9: hlsl.hlsl(1000): error :  candidate function template not viable: requires 3 arguments, but 5 were provided
dxc 1.9: hlsl.hlsl(1000): error :  candidate function template not viable: requires 4 arguments, but 5 were provided
```

jkwak-work:
The error message tells us that the shader was trying to use "sampler" for "Load", which makes no sense.
The use of "sampler" is to apply some kind of filtering but "Load" means we want to read a value without filter.
The meanings from both conflicts.

It makes sense that HLSL doesn't have a function overloading on `Load` with sampler support.

This is a user error but Slang should handle this better with `[require]` if possible.

jkwak-work:
It seems like there is a problem when a combined texture is declared.
[...]

PR #7235, opened 2025-05-25
Properly error out when Image::Load is called with sampler

jkwak-work:
This can resolve the issue described in 
- https://github.com/shader-slang/slang/issues/6982

With this PR, when `Load` function in HLSL is used with a Sampler, Slang will print an error message due to `static_assert`.

But the disabled test cannot be re-enabled yet.
I am preparing another fix for it.


jkwak-work:
Changing to a draft until the fix on slang-rhi lands.

PR #7244, opened 2025-05-27
Add check for subscript return type

### LINKED ISSUE #6987: `__subscript` without defined return type leads to an internal error
If the return type in the `__subscript` syntax is not defined, compilation will fail with an internal error, instead of a proper error message.

Tested against the 7f9283a commit.

Example code (derived from the documentation):
```slang
struct MyType
{
    int val[12];
    __subscript(int x, int y)
    {
        get { return val[x*3 + y]; }
    }
}

```

slangc output:
```
test15.slang(6): note 99999: an internal error threw an exception while working on code near this location
        get { return val[x*3 + y]; }
        ^~~
(0): error 99999: Slang compilation aborted due to an exception of N5Slang13InternalErrorE: unexpected: ErrorType
```

mkeshavaNV:
Fixes #6987

### DIFF
--- a/docs/command-line-slangc-reference.md
+++ b/docs/command-line-slangc-reference.md
@@ -1128,6 +1128,7 @@ A capability describes an optional feature that a target may or may not support.
 * `SPV_NV_ray_tracing_motion_blur` : enables the SPV_NV_ray_tracing_motion_blur extension 
 * `SPV_NV_shader_invocation_reorder` : enables the SPV_NV_shader_invocation_reorder extension 
 * `SPV_NV_cluster_acceleration_structure` : enables the SPV_NV_cluster_acceleration_structure extension 
+* `SPV_NV_linear_swept_spheres` : enables the SPV_NV_linear_swept_spheres extension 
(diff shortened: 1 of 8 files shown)


expipiplus1 (approved):

source/slang/slang-check-decl.cpp
```diff
@@ -10405,7 +10405,16 @@ void SemanticsDeclHeaderVisitor::visitAbstractStorageDeclCommon(ContainerDecl* d
 
 void SemanticsDeclHeaderVisitor::visitSubscriptDecl(SubscriptDecl* decl)
 {
-    decl->returnType = CheckUsableType(decl->returnType, decl);
+    // __subscript needs to have a return type specified. Check if return type
```
csyonghe, line 10408: Why special case here in subscript? It should be an error whenever CheckUsableType is called on an IncompleteExpr.

mkeshavaNV: @csyonghe oh this got merged since it was approved. Are you suggesting there are other use cases that trigger IncompleteExpr? This was added for SubscriptDecl since that was the bug report.

PR #7255, opened 2025-05-27
Add command line option for separate debug info

jarcherNV:
Add command line arg -separate-debug-info which, if provided, produces both a .spv and a .dbg.spv file. The .dbg.spv file contains full debug info and the .spv file has all debug info stripped out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,6 +246,31 @@ jobs:
             -expected-failure-list tests/expected-failure.txt \
             -skip-reference-image-generation \
             -show-adapter-info
+      - name: slang-rhi tests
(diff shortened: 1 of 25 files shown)


jarcherNV:
I have updated the patch to point spirv-tools at the Slang fork, which has the needed fix for this to work. We will need to change this back once the patch is upstreamed (assuming all the tests pass).

jarcherNV:
After some recent discussion with Aftermath team it sounds like we may want to use the https://github.com/shader-slang/slang/pull/7178 PR instead. I will leave this open for now in case we change our minds, but likely to close without submitting.

csyonghe:
We should also consider adding a unit test to look for the debug identifier in the spirv that has the debug info.

jarcherNV:
We have decided not to use this approach so I am closing this PR.

PR #7260, opened 2025-05-28
Make interface types non c-style in Slang2026.

### LINKED ISSUE #7261: Types with interface typed fields shouldn't be treated as C-Style.
Given 
```
interface IFoo {}

struct Bar
{
    IFoo f;
}
```

`Bar` shouldn't be treated as a C-style type and shouldn't be construtable with `{}`.

### LINKED ISSUE #7274: Make `Optional<IFoo>` consume the same amount of storage as `IFoo`.
We will disallow initializing `IFoo` with `{}`, such as:
```
interface IFoo{}

IFoo f = {};
```

Because this code has no well-defined meaning and shouldn't be allowed.

To allow initialization to "invalid" value, the user should write:

```
Optional<IFoo> f = none;
```

To make this possible without incurring any runtime overhead, we need to make sure `Optional<IFoo>` occupies the same amount of storage as `IFoo`, similar to what we do for `Optional<void*>`.

csyonghe:
Previously we didn't check for existential types inside a struct and treat such types as non-c-style. That will lead to the compiler generating invalid code from ill-formed user code.

This PR adds the checks in Slang 2026, which means what used to compile (into invalid code) is now diagnosing an error.

The following cascading changes are added to make sure the feature is usable.

To allow an `IFoo f` to be in an "empty"/"null" state, the user is expected to write `Optional<IFoo>`. An `Optional<IFoo>` is made to take up exact same amount of bytes as `IFoo`.

This PR also adds a new `ISession::getDynamicObjectRTTIBytes` to allow an application to query what bytes to append to a dynamic object when filling in buffers for `IFoo`, and applications are expected to call this function to get the 16byte header, instead of assembling the header themselves with the sequential witness ID.

In order to test the changes, this PR also adds a `-conformance "Bar:IFoo=1"` slangc option to specify additional type conformances.

The PR also fixes optional type lowering and move it to be after generic specialization. This means that optional types are lowered after autodiff, and autodiff passes are modified to work with optional types explicitly. Doing so also revealed that the type system never supported optional types for autodiff, so the PR also adds that, and added testing for using Optional types in autodiff.

Moving optional type lowering to after specialization means we also need to move `Com-Interface` lowering pass, because that pass assumes optional types are lowered first. Therefore, the PR also moves that pass to be after optional type lowering.

Closes #7261.
Closes #7274.


fairywreath (dismissed):

source/slang/slang-ir-lower-optional-type.cpp
```diff
@@ -81,15 +79,20 @@ struct OptionalTypeLoweringContext
         case kIROp_NativeStringType:
         case kIROp_PtrType:
         case kIROp_ClassType:
+            outKind = LoweredOptionalTypeKind::PtrValue;
```
ArielG-NV, line 82: should `kIROp_StructType` also be here?

csyonghe: No, it shouldn't. A class type is a reference, so it should be treated as a pointer. but a struct type is not.

source/slang-record-replay/record/slang-session.cpp
```diff
@@ -369,6 +369,34 @@ SLANG_NO_THROW SlangResult SessionRecorder::getTypeConformanceWitnessMangledName
     return result;
 }
 
+SLANG_NO_THROW SlangResult SessionRecorder::getDynamicObjectRTTIBytes(
+    slang::TypeReflection* type,
+    slang::TypeReflection* interfaceType,
+    uint32_t* outRTTIDataBuffer,
+    uint32_t bufferSizeInBytes)
```
kaizhangNV, line 376: doesn't `bufferSizeInBytes` need to be recorded?

tests/autodiff/optional.slang
```diff
@@ -0,0 +1,50 @@
...
+
+    // CHECK: 1.0
+    dpa = diffPair<Optional<float>>(3.0f, none);
+    dpb = diffPair<Optional<float>>(4.0f, none);
+    bwd_diff(sumSquare)(dpa, dpb, none);
+    if (dpa.d.value == 0.0 && dpb.d.value == 0.0)
+    {
+        outputBuffer[1] = 1.0f;
+    }
+}
```
kaizhangNV, line 50: for this test, should we also add the path like
```
var dpa = diffPair<Optional<float>>(none, none);
```
so it can cover more paths in `sumSquare`.

kaizhangNV (approved):
There is nothing blocking this PR.
I just have few suggestions and a question.

source/slang-record-replay/record/slang-session.cpp
csyonghe: didn't pay attention to this, yes they need to be recorded, but I don't know how `outRTTIDataBuffer` should be handled.

tests/autodiff/optional.slang
csyonghe: done.

source/slang-record-replay/record/slang-session.cpp
csyonghe: I think this function doesn't need to be recorded at all, since it is just a reflection-like function that won't affect compilation.

kaizhangNV: If you think we don't need to record this function, then it's fine, just leave it empty.

But if we do need to record this function, your currently code is correct, don't need to record `outRTTIDataBuffer`. But we need to record bufferSizeInBytes, so replayer will know how large the buffer to allocate when calling this function.

tests/language-feature/interfaces/optional-none.slang
```diff
@@ -0,0 +1,47 @@
...
+Optional<T> process<T>(Optional<T> opt)
+{
+    return opt;
+}
+
+[numthreads(1, 1, 1)]
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+{
+    MyType t = {};
+    t.foo = process(t.foo);
```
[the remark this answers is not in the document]
kaizhangNV: I see.

kaizhangNV (approved):
Except the recording thing, everything else looks good to me.

PR #7264, opened 2025-05-28
Move IncompleteExpr check to CheckUsableType

mkeshavaNV:
PR https://github.com/shader-slang/slang/pull/7244 added check for IncompleteExpr for `__subscript`. It can be made a more generic check as was suggested in that PR.

### DIFF
--- a/docs/design/serialization.md
+++ b/docs/design/serialization.md
@@ -5,6 +5,147 @@ Slang's infrastructure for serialization is currently in flux, so there exist a 
 
 This document is curently minimal, and primarily serves to provide a replacement for an older draft that no longer reflects the state of the codebase.
 
+The Fossil Format
+=================
+
+The "fossil" format is a memory-mappable binary format for general-purpose serialization.
+
+Goals
+-----
+
+The main goals of the fossil format are:
+
+* Data can be read from memory as-is.
+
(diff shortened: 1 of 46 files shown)


szihs (commented):
Looks good to me.

expipiplus1 (approved):

source/slang/slang-check-type.cpp
```diff
@@ -447,6 +447,11 @@ TypeExp SemanticsVisitor::CoerceToUsableType(TypeExp const& typeExp, Decl* decl)
 
 TypeExp SemanticsVisitor::CheckUsableType(TypeExp typeExp, Decl* decl)
 {
+    auto callableDecl = as<CallableDecl>(decl);
```
csyonghe, line 450: I think the right place is to diagnose an error immediately upon creation of the `IncompleteExpr`.

Search for all `create<IncompleteExpr>` revealed that we are diagnosing an error before creating `IncompleteExpr` in all places, except for `SubscriptDecl` in `parseSubscriptDecl` (in slang-parser.cpp). Should just add the diagnostic there, and keep this function unchanged.

mkeshavaNV: thanks, fixed comment.

csyonghe (approved):

Issue #5161, opened 2024-09-26
Investigate why CTS Nightly started failing


jkwak-work:
VK-CTS nightly CI started failing from Sep/20.
The output log looks like the process was crashed while running the CTS test.
And it appears to be crashing always at the following test,
```
Test case 'dEQP-VK.glsl.texture_gather.offset.min_required_offset.2d_array.depth32f.size_npot.compare_greater.mirrored_repeat_clamp_to_edge'..
  Pass (Pass)

Test case 'dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.clamp_to_edge_repeat'..
Error: Process completed with exit code 1.
```

jkwak-work:
I was not able to reproduce the issue on my local machine nor on the runner machine.
When I tried to reproduce the issue on the runner machine, I also used the slang.dll built by the runner action.
I am puzzled about how to reproduce it.

I am going to disable a few tests and trigger the CI.
Since CI is only why to reproduce the issue at the moment, this may allow me to narrow down or bisect to where to problem is.

I can also setup a local runner on my machine and see if I can reproduce it via CI process.

jkwak-work:
The failing CTS nightly tests can be found from [here](https://github.com/shader-slang/slang/actions/workflows/vk-gl-cts-nightly.yml)

aleino-nv:
Do we have some kind of asserts enabled in Release mode?

I ask because the process exit code on Windows is never just `1` on a "crash" (e.g. segfault) -- it's usually some 'large' number.

This makes me think the exit is controlled. Maybe even `exit(1)` is being called.
If so, then probably there is a more detailed error message somewhere. Do we capture stderr? Is the full (.xml format) dEQP log available?

jkwak-work:
I was able to reproduce the issue only through the CI workflow.
And the following three tests are related to the problem,
```
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.clamp_to_edge_repeat
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.mirrored_repeat_clamp_to_edge
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.repeat_mirrored_repeat
```

When I removed them from the tests, the CI went green.
When I tested only three of them, the CI was also passing, which is strange.

I believe this is related to the recent upgrade of SPIR-V headers.
[...]

PR #7269, opened 2025-05-28
Allow regular params to propagate constexpr-ness

### LINKED ISSUE #6370: Loop fails to unroll and invalid SPIR-V is generated for OpImageFetch

gtong-nv:
The https://github.com/shader-slang/slang/pull/3139 added a check to prevent the op_param from propagating constexpr-ness unless it's loopPhi. 
I'm not sure if that is intentional, but removing it fixes #6370, and still keep the same behavior that the loop counters can be used as constexpr - somehow bypassing the constexpr-ness requirement from the callee functions.

Fixes #6370

### DIFF
--- a/source/slang/slang-ir-constexpr.cpp
+++ b/source/slang/slang-ir-constexpr.cpp
@@ -177,8 +177,6 @@ IRLoop* isLoopPhi(IRParam* param)
 
 bool opCanBeConstExprByBackwardPass(IRInst* value)
 {
-    if (value->getOp() == kIROp_Param)
-        return isLoopPhi(as<IRParam, IRDynamicCastBehavior::NoUnwrap>(value));
     if (opCanBeConstExpr(value->getOp()))
         return true;
     if (auto callInst = as<IRCall>(value))


tests/bugs/gh-5339.slang
```diff
@@ -21,7 +20,8 @@ buffer MyBlockName
...
 {
     //SPV:OpImageGather %
-    //SPV-NOT:Const
-    //SPV-SAME: Offset %
-    outputBuffer.result = textureGatherOffset(u_sampler, v_texCoord, u_offset);
+    //SPV-SAME: ConstOffset %
+    // Can textureGatherOffset take non-const offset? 
+    constexpr ivec2 offset2D = ivec2(2);
+    outputBuffer.result = textureGatherOffset(u_sampler, v_texCoord, offset2D);
 }
```
gtong-nv: This test is added for the github issue #5339 , where it claims textureGatherOffset() takes a non-cosnt offset. I couldn't find anywhere specifing that in the glsl specs, only meantioning the offset should be static.
@jkwak-work , can you point me to the specs stating that? 
If we do need textureGatherOffset to take both const and non-const offests, I'm not sure if we can overload that. Might need to add something like `textureGatherDynamicOffset()` for the non-const version.

source/slang/slang-diagnostic-defs.h
```diff
@@ -2274,11 +2274,8 @@ DIAGNOSTIC(
     "\"index\"] attribute to provide a binding location.")
 DIAGNOSTIC(40006, Error, unimplementedSystemValueSemantic, "unknown system-value semantic '$0'")
 
-
 DIAGNOSTIC(49999, Error, unknownSystemValueSemantic, "unknown system-value semantic '$0'")
 
-DIAGNOSTIC(40006, Error, needCompileTimeConstant, "expected a compile-time constant")
```
gtong-nv, line 2448: 40006 is conflict with the `unimplementedSystemValueSemantic` error code above, so I renamed it to 40012 below.

source/slang/slang-ir-constexpr.cpp
```diff
@@ -481,6 +493,14 @@ bool propagateConstExprBackward(PropagateConstExprContext* context, IRGlobalValu
...
                                         changedThisIteration = true;
                                     }
+                                    // If arg is not constexpr after this, meaning it can't be marked constexpr for some reason,
+                                    // but the param requires that. This is not expected.
+                                    if (!isConstExpr(arg)) {
+                                        context->getSink()->diagnose(
+                                            callInst->sourceLoc,
+                                            Diagnostics::argIsNotConstexpr, pp+1, calleeFunc);
+                                        return false;
+                                    }
```
gtong-nv: The `if (calleeFunc && isDefinition(calleeFunc))` and the `else` block seems doing the exact the same thing, and removing one doesn't affect the tests at all, including the tests that have external calls.
Can we combine the branches into one?

gtong-nv: Disucessed with Tess and looks like there is no distinct diefference with the current implementation for external/internal functions. I combined the two banches into one.

tests/bugs/gh-5339.slang
jkwak-work: My change was based on the CTS failure.
The CTS task case was using the offset value coming from the uniform variable, which is not a constant value but a runtime value.
I don't see the spec requiring it to be a constant value or static here.
https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureGatherOffset.xhtml
So it should work with both const and none-const, I think.
If it work with non-const, wouldn't it work with constexpr automatically?

gtong-nv: >So it should work with both const and none-const, I think.

yes, I think the `textureGatherOffset` takes both const and non-const offset

>  If it work with non-const, wouldn't it work with constexpr automatically

With this PR that adds the const-ness requirement, it must be const, given we don't have a good way to overload them and tell them apart in the frontend. 
Tess suggested for this PR we can just leave as it is. So I might need to update this test.
But I don't know if that would the CTS failure again..

gtong-nv:
It's likely this PR will break the CTS tests that were disabled previously for #5161

dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.clamp_to_edge_repeat
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.mirrored_repeat_clamp_to_edge
dEQP-VK.glsl.texture_gather.offset.implementation_offset.2d.rgba8.size_pot.repeat_mirrored_repeat

jhelferty-nv:
@gtong-nv @jkwak-work  Is this PR still in review? There has been no movement in a month.

source/slang/slang-ir-constexpr.cpp
```diff
@@ -60,7 +61,10 @@ bool isConstExpr(IRInst* value)
     case kIROp_WitnessTable:
     case kIROp_Generic:
         return true;
-
+    //case kIROp_Param:
+    //    if (isGenericParam(value))
+    //        return true;
+    //    break;
```
gtong-nv: with this, `tests\language-feature\types\conditional-varying.slang` failed because `IR_intCast(bool)`, where the bool is a generic param, is marked constexpr, but its operand is not

gtong-nv: The resolution is to keep this check, and for the test `tests\language-feature\types\conditional-varying.slang`, the opearnd of the `IR_intCast` is an `IR_GlobalConstant`, which should also be a `constexpr`. Updated in the new commit.

gtong-nv:
https://github.com/shader-slang/slang/actions/runs/17664177055
VK CTS is fully passing after dropping the `constexpr` for offsets in the `textureGather*` intrinsics.
This is needed because of https://github.com/shader-slang/slang/pull/7269#discussion_r2136188327
Given that, I think this PR is ready to be merged

source/slang/hlsl.meta.slang
```diff
@@ -1064,7 +1064,7 @@ extension _Texture<T,Shape,isArray,isMS,sampleCount,0,isShadow,1,format>
     [ForceInline]
     __glsl_extension(GL_ARB_sparse_texture_clamp)
     [require(cpp_glsl_hlsl_metal_spirv, texture_sm_4_1_clamp_fragment)]
-    T Sample(vector<float, Shape.dimensions+isArray> location, vector<int, Shape.planeDimensions> offset, float clamp)
+    T Sample(vector<float, Shape.dimensions+isArray> location, constexpr vector<int, Shape.planeDimensions> offset, float clamp)
```
csyonghe, line 1067: why do we need to add constexpr here?

gtong-nv: This is to be consistent with the Sample functions in line 1984 and 2048 etc. 
Without this, I would get error `arg 4 in 'Sample' is not a compile-time constant`

gtong-nv: Another option is to remove all `constexpr` from offset in the _Texture/Sample chain, but I'm afraid that would introduce large consequence.

csyonghe: OK, this is fine for now, but I think we need to remove all these constexpr modifiers on sample methods because it is now allowed for offset to be non-constexpr on vulkan and d3d. This can be for another day.

csyonghe (approved):

